PR #93, opened 2026-02-12
fix: prevent UTF-8 panics on multi-byte characters

polaminggkub-debug:
## Summary

- Fix 10 byte-indexed string slicing locations (`&s[..n]`) that panic on multi-byte UTF-8 characters (Thai, emoji, CJK)
- Replace with char-aware operations: `chars().take()`, `.get()`, `is_char_boundary()` 
- Add `safe_char_boundary()` utility in `utils.rs`
- Add 20 regression tests covering Thai, emoji, and CJK input

## Problem

RTK crashes on non-ASCII content because Rust strings are UTF-8 encoded. Thai characters are 3 bytes, emoji are 4 bytes. `&line[..77]` can land mid-character and panic with `byte index is not a char boundary`.

## Files Changed

| File | Fix |
|------|-----|
| `src/git.rs` | `filter_log_output`: `chars().take(77)` instead of `&line[..77]`; `format_status_output`: `.get()` instead of `&line[0..2]` |
| `src/log_cmd.rs` | Error/warning truncation: `chars().take(97)` instead of `&original[..97]` (2 locations) |
| `src/env_cmd.rs` | Long value display: `chars().take(50)`; `mask_value`: char-based prefix/suffix |
| `src/parser/mod.rs` | `truncate_output`: `is_char_boundary()` loop instead of `&output[..max_chars]` |
| `src/grep_cmd.rs` | `clean_line`: char-boundary snapping for pattern context; `chars().take()` for fallback |
| `src/wget_cmd.rs` | `compact_url`, `truncate_line`, `parse_error`: all switched to `chars().take()` |
| `src/utils.rs` | New `safe_char_boundary()` helper + multi-byte truncation tests |

## Test plan

- [x] `cargo test` — 284 tests pass (20 new UTF-8 regression tests)
- [x] `cargo clippy` — no new warnings (all pre-existing)
- [x] `cargo build --release` — builds successfully
- [ ] Manual: create file with Thai name, run `rtk git status` — should not panic

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #110, opened 2026-02-13
feat(docker): add docker compose support

### LINKED ISSUE #101: feat: Add support for docker compose
Love the tool, thanks for the effort.

I noticed you had support for docker commands.
Support for `docker compose` commands would also be handy.

polaminggkub-debug:
## Summary
- Adds token-optimized output for `docker compose ps`, `docker compose logs`, and `docker compose build`
- Unsupported compose subcommands (`up`, `down`, `restart`, etc.) pass through directly
- Includes 11 unit tests (8 compose + 3 for existing `compact_ports`)

## Commands Added

| Command | Strategy | Expected Savings |
|---------|----------|-----------------|
| `rtk docker compose ps` | Compact service table (name, image, status, ports) | ~70% |
| `rtk docker compose logs` | Deduplication via existing log engine | ~60-80% |
| `rtk docker compose build` | Summary line + service names + step count | ~80% |
| Other compose subcommands | Passthrough | 0% (tracked) |

## Test plan
- [x] 11 unit tests pass (`cargo test container::tests::`)
- [x] Full suite: 324 tests pass
- [x] `cargo fmt --all --check` clean
- [x] `cargo clippy --all-targets` no new warnings
- [x] Manual test: `cargo run -- docker compose ps` produces compact output

Fixes #101

### DIFF
--- a/.github/workflows/validate-docs.yml
+++ b/.github/workflows/validate-docs.yml
@@ -1,0 +1,69 @@
+name: Documentation Validation
+
+on:
+  pull_request:
+    paths:
+      - 'src/**/*.rs'
+      - 'Cargo.toml'
+      - '**.md'
+      - '.claude/hooks/*.sh'
+  push:
+    branches:
+      - master
+      - feat/**
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Validate documentation consistency
+        run: bash scripts/validate-docs.sh
+
+      - name: Check module count consistency
+        run: |
+          MAIN_MODULES=$(grep -c '^mod ' src/main.rs)
+
+          # Check ARCHITECTURE.md if it exists
+          if [ -f "ARCHITECTURE.md" ]; then
+            ARCH_MODULES=$(grep 'Total:.*modules' ARCHITECTURE.md | grep -o '[0-9]\+' | head -1)
+            if [ -n "$ARCH_MODULES" ] && [ "$MAIN_MODULES" != "$ARCH_MODULES" ]; then
+              echo "❌ Module count mismatch"
+              echo "main.rs: $MAIN_MODULES modules"
+              echo "ARCHITECTURE.md: $ARCH_MODULES modules"
+              exit 1
+            fi
+          fi
+
+          echo "✅ Module count consistent: $MAIN_MODULES modules"
+
+      - name: Verify Python/Go commands documented
+        run: |
+          for cmd in ruff pytest pip go golangci; do
+            if ! grep -q "$cmd" README.md; then
+              echo "❌ README.md missing Python/Go command: $cmd"
+              exit 1
+            fi
+            if ! grep -q "$cmd" CLAUDE.md; then
+              echo "❌ CLAUDE.md missing Python/Go command: $cmd"
+              exit 1
+            fi
+          done
+          echo "✅ All Python/Go commands documented"
+
+      - name: Verify hook coverage
+        run: |
+          HOOK_FILE=".claude/hooks/rtk-rewrite.sh"
+          if [ ! -f "$HOOK_FILE" ]; then
+            echo "❌ Hook file not found: $HOOK_FILE"
+            exit 1
+          fi
+
+          for cmd in ruff pytest pip "go " golangci; do
+            if ! grep -q "$cmd" "$HOOK_FILE"; then
+              echo "❌ Hook missing rewrite for: $cmd"
+              exit 1
+            fi
+          done
+          echo "✅ Hook rewrites present for Python/Go commands"


src/container.rs
```diff
@@ -385,6 +385,152 @@ fn kubectl_logs(args: &[String], _verbose: u8) -> Result<()> {
...
+                    .unwrap_or(line_str.len());
+                line_str[idx..end].trim()
+            } else {
+                "unknown"
+            };
+
+            // Extract port from end of line
+            let port_str = if let Some(port_idx) = line_str.rfind("0.0.0.0:") {
+                let chars: Vec<char> = line_str.chars().collect();
+                let port_chars: Vec<char> = chars[port_idx..].to_vec();
```
pszymkowiak: Bug: `rfind()` returns a **byte offset**, but it's used here as an index into `Vec<char>`. If the line contains any multi-byte UTF-8 characters before the port section, this will index into the wrong position or panic with out-of-bounds.

Same pattern flagged in PR #93.

Fix: either work entirely in bytes (use `&line_str[port_idx..]`) or convert the byte offset to a char offset:
```rust
let port_str = if let Some(port_idx) = line_str.rfind("0.0.0.0:") {
    let compact = compact_ports(line_str[port_idx..].trim());
    format!(" [{}]", compact)
} else {
    String::new()
};
```
No need for the `chars().collect()` → `Vec<char>` dance here — byte slicing works fine since `rfind` returns a byte offset and string slicing expects byte offsets.

src/container.rs
```diff
@@ -385,6 +385,152 @@ fn kubectl_logs(args: &[String], _verbose: u8) -> Result<()> {
...
+        } else {
+            result.push_str("🐳 Build:\n");
+        }
+    }
+
+    // Collect unique service names from build steps like "[web 1/4]"
+    let mut services: Vec<String> = Vec::new();
+    for line in raw.lines() {
+        if let Some(start) = line.find('[') {
+            let after: String = line.chars().skip(start + 1).collect();
```
pszymkowiak: Same bytes/chars bug: `find('[')` returns a byte offset, but `chars().skip(start + 1)` treats it as a char count.

Simpler fix — use byte slicing throughout:
```rust
if let Some(start) = line.find('[') {
    if let Some(end) = line[start+1..].find(']') {
        let bracket = &line[start+1..start+1+end];
        let svc = bracket.split_whitespace().next().unwrap_or("");
        // ...
    }
}
```
`'['` and `']'` are single-byte ASCII, so byte arithmetic is safe here.

src/container.rs
```diff
@@ -385,6 +385,152 @@ fn kubectl_logs(args: &[String], _verbose: u8) -> Result<()> {
...
+        // We split by 2+ spaces to handle multi-word fields
+        let parts: Vec<&str> = line.split_whitespace().collect();
+        if parts.len() >= 6 {
+            let name = parts[0];
+            let image = parts[1];
+
+            // Find SERVICE column (usually index 3, after quoted COMMAND)
+            // Find STATUS by looking for "Up" or "Exited" keywords
+            let line_str = *line;
+            let status = if line_str.contains("Up") {
```
pszymkowiak: `contains("Up")` is too greedy — it will false-positive on image names like `setup-service:latest`, commands like `"python startup.py"`, etc.

Two options:
1. Search for `"  Up "` (with surrounding spaces, matching docker's column formatting)
2. Better: use `docker compose ps --format` (like the existing `docker_ps()` does) to get structured output instead of parsing the human-readable table. This would avoid all the fragile column-parsing logic.

src/container.rs
```diff
@@ -431,6 +577,113 @@ pub fn run_docker_passthrough(args: &[OsString], verbose: u8) -> Result<()> {
...
 
+/// Run `docker compose ps` with compact output
+pub fn run_compose_ps(verbose: u8) -> Result<()> {
+    let timer = tracking::TimedExecution::start();
+
+    let raw = Command::new("docker")
+        .args(["compose", "ps"])
+        .output()
+        .map(|o| String::from_utf8_lossy(&o.stdout).to_string())
+        .unwrap_or_default();
```
pszymkowiak: Inconsistent error handling: `unwrap_or_default()` silently swallows Docker errors. If Docker isn't installed or `compose` isn't available, the user gets "🐳 0 compose services" with no indication something is wrong.

Compare with `run_compose_logs` (line 610) and `run_compose_build` (line 641) which use `.context("Failed to run ...")?`.

Should be consistent:
```rust
let output = Command::new("docker")
    .args(["compose", "ps"])
    .output()
    .context("Failed to run docker compose ps")?;
let raw = String::from_utf8_lossy(&output.stdout).to_string();
```

src/container.rs
```diff
@@ -385,6 +385,152 @@ fn kubectl_logs(args: &[String], _verbose: u8) -> Result<()> {
...
+                let port_chars: Vec<char> = chars[port_idx..].to_vec();
+                let port_text: String = port_chars.iter().collect();
+                let compact = compact_ports(port_text.trim());
+                format!(" [{}]", compact)
+            } else {
+                String::new()
+            };
+
+            // Shorten image name: keep only the last segment
+            let short_image = image.split('/').last().unwrap_or(image);
```
pszymkowiak: Clippy warning: `.last()` on a `DoubleEndedIterator` iterates the entire iterator needlessly. Use `.next_back()` instead:
```rust
let short_image = image.split('/').next_back().unwrap_or(image);
```

src/container.rs
```diff
@@ -454,3 +707,115 @@ pub fn run_kubectl_passthrough(args: &[OsString], verbose: u8) -> Result<()> {
...
+
+    #[test]
+    fn test_format_compose_logs_basic() {
+        let raw = "\
+web-1  | 192.168.1.1 - GET / 200
+web-1  | 192.168.1.1 - GET /favicon.ico 404
+api-1  | Server listening on port 3000
+api-1  | Connected to database";
+        let out = format_compose_logs(raw);
+        assert!(!out.is_empty(), "should produce output");
```
pszymkowiak: Nit: this test only checks `!out.is_empty()` — it doesn't validate any actual content. At minimum, assert it contains the expected prefix:
```rust
assert!(out.contains("Compose logs"), "should have compose logs header");
```

src/container.rs
```diff
@@ -385,6 +385,152 @@ fn kubectl_logs(args: &[String], _verbose: u8) -> Result<()> {
...
+    if services.is_empty() {
+        return "🐳 0 compose services".to_string();
+    }
+
+    let mut result = format!("🐳 {} compose services:\n", services.len());
+
+    for line in services.iter().take(20) {
+        // docker compose ps columns are whitespace-separated but STATUS can contain spaces
+        // Columns: NAME IMAGE COMMAND SERVICE CREATED STATUS PORTS
+        // We split by 2+ spaces to handle multi-word fields
```
pszymkowiak: Nit: comment says "We split by 2+ spaces" but the code uses `split_whitespace()` which splits on any whitespace (single space included). Either update the comment or actually split by 2+ spaces with a regex/custom split.

polaminggkub-debug:
Sorry for having you review so many issues. I really appreciate your time and feedback. I’ll make sure to double-check everything more carefully next time.

PR #128, opened 2026-02-15
feat(gain): add per-project token savings with -p flag

heAdz0r:
## Summary

Adds per-project scoping to `rtk gain` so users can see token savings for the current project instead of global aggregates.

**Problem:** Users working on multiple projects see combined statistics in `rtk gain`. There's no way to answer "how much does rtk save in _this_ project?"

**Solution:** `rtk gain -p` filters all statistics to the current working directory (and subdirectories).

### Usage

```bash
# Global view (default, unchanged)
rtk gain

# Project-scoped view (new)
rtk gain -p
rtk gain --project

# Works with all existing flags
rtk gain -p --daily
rtk gain -p --format json
rtk gain -p --history
```

### Files Changed

| File | Changes | Purpose |
|------|---------|---------|
| `src/tracking.rs` | +120 | project_path column, migration, filtered query methods |
| `src/gain.rs` | +80 | project scope resolution, scope-aware display |
| `src/main.rs` | +15 | `--project/-p` Clap flag |

**Total:** 215 lines added, 50 lines modified

### tracking.rs

- **Schema migration**: Adds `project_path TEXT` column via `ALTER TABLE` (idempotent, backward-compatible)
- **Auto-recording**: `current_project_path_string()` captures canonical cwd on every command execution
- **Index**: `idx_project_path_timestamp` for fast project-scoped queries
- **NULL normalization**: Pre-existing rows get empty string (migration runs once)
- **Filtered API**: All query methods get `_filtered(project_path: Option<&str>)` variants:
  - `get_summary_filtered()`
  - `get_all_days_filtered()`
  - `get_by_week_filtered()`
  - `get_by_month_filtered()`
  - `get_recent_filtered()`
- **SQL filter**: `WHERE (?1 IS NULL OR project_path = ?1 OR project_path LIKE ?2)` — exact match + subdirectory prefix match
- Original unfiltered methods delegate to `_filtered(None)` (zero behavior change)

### gain.rs

- `resolve_project_scope()`: Resolves `--project` flag to canonical cwd path
- `shorten_path()`: Abbreviates long paths for display (`/Users/foo/bar/project` → `/.../bar/project`)
- Scope-aware header: "RTK Token Savings (Project Scope)" with path display
- All exports (JSON, CSV) pass project scope through

### Backward Compatibility

- Existing databases: Migration adds column with empty default. All existing data remains accessible.
- Existing CLI: `rtk gain` without `-p` behaves identically (filtered with `None` returns all rows).
- No new dependencies added.

### Critical files check
- `src/tracking.rs` is a critical file (per `security-check.yml`)
- **No shell execution**: All changes are SQLite queries via rusqlite parameterized statements
- **No user input in SQL**: Project path comes from `std::env::current_dir()`, not CLI arguments
- **No Cargo.toml changes**: No new dependencies

## Verification

- [x] `cargo build` — compiles clean
- [x] `cargo clippy` — no errors
- [x] `cargo test tracking` — 7 passed, 1 pre-existing failure (unrelated `test_default_db_path`)
- [x] `cargo fmt` — formatted

---

### DIFF
--- a/src/tracking.rs
+++ b/src/tracking.rs
@@ -49,12 +49,13 @@ fn current_project_path_string() -> String {
 }
 
 /// Build SQL filter params for project-scoped queries.
-/// Returns (exact_match, prefix_like) for WHERE clause.
+/// Returns (exact_match, glob_prefix) for WHERE clause.
+/// Uses GLOB instead of LIKE to avoid `_` and `%` in paths acting as wildcards. // changed: GLOB
 fn project_filter_params(project_path: Option<&str>) -> (Option<String>, Option<String>) {
     match project_path {
         Some(p) => (
             Some(p.to_string()),
-            Some(format!("{}{}", p, std::path::MAIN_SEPARATOR)),
+            Some(format!("{}{}*", p, std::path::MAIN_SEPARATOR)), // changed: GLOB pattern with * wildcard
         ),
         None => (None, None),
     }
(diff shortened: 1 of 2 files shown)


src/tracking.rs
```diff
@@ -340,17 +378,28 @@ impl Tracker {
...
     /// ```
     pub fn get_summary(&self) -> Result<GainSummary> {
+        self.get_summary_filtered(None) // delegate to filtered variant
+    }
+
+    /// Get summary statistics filtered by project path. // added
+    ///
+    /// When `project_path` is `Some`, matches the exact working directory
+    /// or any subdirectory (prefix match with path separator).
+    pub fn get_summary_filtered(&self, project_path: Option<&str>) -> Result<GainSummary> {
```
pszymkowiak, line 401: LIKE ?2 without a trailing % wildcard is an exact match — subdirectory prefix matching doesn't work. project_path LIKE '/home/user/project/' will never match /home/user/project/src.      
                                                                            
  Fix: Some(format!("{}{}%", p, std::path::MAIN_SEPARATOR))                                                                                                                                  
                                                                                                                                                                                             
  But then _ and % in path names become LIKE wildcards (e.g. my_project matches myXproject). Consider using GLOB instead of LIKE, or add an ESCAPE clause.

src/tracking.rs
```diff
@@ -251,6 +274,18 @@ impl Tracker {
             "ALTER TABLE commands ADD COLUMN exec_time_ms INTEGER DEFAULT 0",
             [],
         );
+        // Migration: add project_path column for per-project scoping // added
+        let _ = conn.execute("ALTER TABLE commands ADD COLUMN project_path TEXT", []);
+        // Normalize NULL values for pre-project-aware rows // added
+        let _ = conn.execute(
+            "UPDATE commands SET project_path = '' WHERE project_path IS NULL",
```
pszymkowiak: The UPDATE commands SET project_path = '' WHERE project_path IS NULL in Tracker::new() runs on every startup, not just once. After the first migration it's a no-op query on every      
  command invocation. Consider wrapping it in a schema version check or checking if the column already has a default.

PR #141, opened 2026-02-16
feat(hooks): add cross-platform Node.js/Bun hook for Windows support

schizoidcock:
## What
- Add `rtk-rewrite.js` - a cross-platform version of the hook that works on Windows, macOS, and Linux
- Update `rtk-awareness.md` with installation instructions for both platforms

## Why
The original `rtk-rewrite.sh` bash script doesn't work on Windows because:
1. Windows can't execute `.sh` files directly
2. Even with Git Bash, the path resolution and stdin handling differs

## Key Features
1. **Cross-platform**: Works with Node.js or Bun on any OS
2. **Chained command support**: Handles `cd "..." && git status` patterns
   - Claude Code often generates commands like `cd "/path" && git status`
   - Extracts the last command after `&&` for matching
   - Rewrites to: `cd "/path" && rtk git status`
3. **No external dependencies**: Only requires Node.js or Bun (no jq needed on Windows)
4. **Same pattern matching**: Equivalent regex logic to the bash version (grep/sed)

## Platform Recommendations
- **macOS/Linux**: Use `rtk-rewrite.sh` (bash + grep/sed + jq) - battle-tested
- **Windows**: Use `rtk-rewrite.js` (Node.js/Bun) - native compatibility

## Usage (Windows)
```json
{
  "hooks": {
    "PreToolUse": [{
      "matcher": "Bash",
      "hooks": [{ "type": "command", "command": "bun C:/Users/YOU/.claude/hooks/rtk-rewrite.js" }]
    }]
  }
}
```

### DIFF
--- a/hooks/rtk-rewrite.js
+++ b/hooks/rtk-rewrite.js
@@ -6,6 +6,446 @@
 
 const fs = require('fs');
 
+// ============================================================================
+// LEXER - Quote-aware tokenizer for shell commands
+// ============================================================================
+
+const TokenKind = {
+  ARG: 'arg',
+  OPERATOR: 'operator',  // &&, ||, ;
+  PIPE: 'pipe',          // |
+  REDIRECT: 'redirect',  // >, >>, <, 2>, etc.
+  SHELLISM: 'shellism',  // *, ?, $(), ``, etc.
+};
+
+/**
+ * Tokenize a shell command string, respecting quotes.
+ * Returns array of { kind, value } tokens.
+ */
+function tokenize(input) {
+  const tokens = [];
+  let i = 0;
+  let current = '';
+  let inSingleQuote = false;
+  let inDoubleQuote = false;
+
+  const pushArg = () => {
+    if (current) {
+      // Check for shellisms in unquoted content
+      // Skip if the token is fully quoted (starts and ends with same quote)
+      const isFullyQuoted = (current.startsWith('"') && current.endsWith('"')) ||
+                            (current.startsWith("'") && current.endsWith("'"));
+      if (!isFullyQuoted && /[*?]|\$\(|`/.test(current)) {
+        tokens.push({ kind: TokenKind.SHELLISM, value: current });
+      } else {
+        tokens.push({ kind: TokenKind.ARG, value: current });
+      }
+      current = '';
+    }
+  };
+
+  while (i < input.length) {
+    const ch = input[i];
+    const next = input[i + 1];
+
+    // Handle escape sequences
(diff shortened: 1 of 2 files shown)


FlorianBruniaux:
Thanks for tackling Windows support — that's a real gap today.

A few points:
1. **Overlap with #131**: @ahundt's PR also implements chained command rewriting (`cd && git status` patterns). Please coordinate to avoid duplicate work — ideally one approach wins and the other adapts
2. **CI is failing** — could you check what's broken?
3. **Node.js/Bun as dependency**: The current bash hook has zero external deps. Adding a Node.js/Bun requirement is a significant change for the install story. Is there a way to make it optional (Windows-only) while keeping the bash hook as default on Unix?

The cross-platform goal is valid, but let's make sure we don't regress the zero-dependency Unix experience.

pszymkowiak:
Thanks for the thorough discussion and the quality of this PR — the lexer, the test suite, and the edge case handling are genuinely solid work.

You make a fair point about auditability vs signatures. For the local threat model you describe, you're right that a signed binary doesn't protect against someone who already has filesystem access.

That said, we're going with the binary approach (`rtk hook claude` from PR #156) as the single path forward. The main reasons:

1. **Speed** — Rust gives us sub-millisecond hook execution. The hook runs on every single Bash command Claude Code generates, so latency matters. A Node.js/Bun cold start adds 30-80ms overhead per invocation, which adds up fast in agentic loops with dozens of commands.
2. **One implementation to maintain** — keeping bash + JS + binary hooks in sync as we add features (like the secrets vault we're working on, which adds a decrypt pass to the hook) would be a real maintenance burden
3. **Consistent architecture** — everything in Rust, same build/test/release pipeline

Regarding your point about contribution barrier and extensibility — we agree that rewrite rules shouldn't require Rust knowledge to modify. The plan is to externalize the rule definitions into TOML configuration, so users and contributors can add/customize rewrite patterns without touching Rust code or recompiling. The binary handles the lexer, chain parsing, and execution; TOML handles the "what to rewrite" part.

If you'd like to explore that direction, we'd love a PR that designs the TOML schema for custom rewrite rules — something like:

```toml
[[rewrite]]
match = "terraform"
subcommands = ["plan", "apply", "init"]
rewrite = "rtk {command}"
```

That would let anyone add patterns for their stack without touching Rust. Your understanding of the rewrite logic from this PR puts you in a great position to design that. Happy to discuss in [GitHub Discussions](https://github.com/rtk-ai/rtk/discussions) if you prefer.

Closing in favor of #156, but genuinely appreciate the effort and hope to see you on another PR!

PR #215, opened 2026-02-19
fix(playwright): fix JSON parser and binary resolution

edsai:
Fixes three bugs that cause `rtk playwright test` to fall through to passthrough mode (Tier 3) instead of parsing JSON output (Tier 1):

**1. JSON struct mismatch**

The parser expected `suites[].tests[]` but Playwright's JSON reporter outputs `suites[].specs[].tests[].results[]`. Added the missing `PlaywrightSpec` intermediate struct, use `spec.ok` for failure detection, and changed `error` from `Option<String>` to `errors: Vec<PlaywrightError>` to match the array format Playwright emits.

**2. Float duration**

Playwright emits `stats.duration` as a float (e.g. `3519.7039999999997`). The struct had `u64`, causing deserialization to fail. Changed to `f64`.

**3. pyenv shim collision**

`which playwright` can resolve to a pyenv shim (Python's playwright package), which doesn't have a `test` subcommand. This causes the spawned process to error out before producing any JSON. Fixed by always resolving through the detected package manager (`npx`/`pnpm exec`/`yarn exec`) instead of `which`.

Also strips any user-supplied `--reporter` flags so they don't conflict with the forced `--reporter=json`.

Tested against real Playwright test suites — parser now reliably hits Tier 1 (full JSON parse).

### DIFF
--- a/src/playwright_cmd.rs
+++ b/src/playwright_cmd.rs
@@ -1,5 +1,5 @@
 use crate::tracking;
-use crate::utils::{package_manager_exec, strip_ansi};
+use crate::utils::{detect_package_manager, strip_ansi};
 use anyhow::{Context, Result};
 use regex::Regex;
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


pszymkowiak:
LGTM — tested locally on the branch, all good.                                                                                                                                             
                                                                                                                                                                                           
  - 398 tests pass, no conflicts with master                                                                                                                                                 
  - 82/83 smoke tests pass (1 false positive: test script hardcodes expected test count)                                                                                                     
  - 23 commands tested manually: git status/log/diff/branch, ls, read, find, grep, gh pr/issue list, deps, env, summary, err, test, gain, cargo build, tsc/lint/vitest/docker/playwright help
   — zero regressions
  - rtk playwright test without a Playwright project: falls back cleanly to passthrough (no crash)
  - All 3 fixes are solid: JSON struct matches real Playwright output (suites→specs→tests→results), float duration handled, pyenv shim collision avoided
  - Good test coverage added: float duration, failure with error messages, nested suites

PR #216, opened 2026-02-19
feat: add AWS CLI and psql modules

itai-sagi:
## Summary

- **`rtk aws`**: Specialized filters for 8 high-frequency AWS CLI commands (STS identity, S3 ls, EC2 describe-instances, ECS list/describe-services, RDS describe-db-instances, CloudFormation list/describe-stacks). Generic fallback forces `--output json` and compresses via `json_cmd::filter_json_string()` at depth 4. Achieves 60%+ token savings on verbose AWS table/text output.
- **`rtk psql`**: Detects psql table format (strips `----+----` borders, `(N rows)` footer, column padding → tab-separated) and expanded format (`-[ RECORD N ]-` blocks → `[N] key=val` one-liners). Passthrough for COPY/SET/notices. Overflow limits: 30 table rows, 20 expanded records.
- Hook rewrite patterns added for both `aws` and `psql` in distributable and dev hooks.

## Test plan

- [x] 32 new inline tests (16 aws_cmd + 16 psql_cmd) all passing
- [x] Full test suite: 428 tests pass, 0 failures
- [x] `cargo fmt --all` clean
- [ ] Manual test with real AWS CLI (if configured): `rtk aws sts get-caller-identity`
- [ ] Manual test with real psql (if available): `rtk psql -c "SELECT 1" mydb`

### DIFF
--- a/.claude/hooks/rtk-rewrite.sh
+++ b/.claude/hooks/rtk-rewrite.sh
@@ -195,6 +195,14 @@ elif echo "$MATCH_CMD" | grep -qE '^go[[:space:]]+vet([[:space:]]|$)'; then
   REWRITTEN="${ENV_PREFIX}$(echo "$CMD_BODY" | sed 's/^go vet/rtk go vet/')"
 elif echo "$MATCH_CMD" | grep -qE '^golangci-lint([[:space:]]|$)'; then
   REWRITTEN="${ENV_PREFIX}$(echo "$CMD_BODY" | sed 's/^golangci-lint/rtk golangci-lint/')"
+
+# --- AWS CLI ---
+elif echo "$MATCH_CMD" | grep -qE '^aws[[:space:]]+'; then
+  REWRITTEN="${ENV_PREFIX}$(echo "$CMD_BODY" | sed 's/^aws /rtk aws /')"
+
+# --- PostgreSQL ---
+elif echo "$MATCH_CMD" | grep -qE '^psql([[:space:]]|$)'; then
+  REWRITTEN="${ENV_PREFIX}$(echo "$CMD_BODY" | sed 's/^psql/rtk psql/')"
 fi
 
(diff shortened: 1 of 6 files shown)


pszymkowiak:
Thanks for adding AWS CLI and psql support — great additions to RTK's ecosystem! Tests pass (428/428), build is clean, and existing commands are unaffected. Here are some items to address
   before merge:

  Critical

  1. Revert commit polluting the PR: There's a revert commit that adds noise to the diff. Please clean up the git history (interactive rebase) so only AWS/psql changes are included.
  2. run_generic --output injection: The --output json flag is injected into all AWS commands via run_generic, but not all AWS subcommands support --output. This could break commands like
  aws s3 cp or aws s3 sync. Consider only injecting --output json for commands that support structured output (describe, list, get, etc.), or catch the error and fallback to raw execution.
  3. psql tracking on failed commands: Token savings are recorded even when the psql command fails (non-zero exit code). Tracking should only happen on successful executions to avoid
  skewing metrics.

  Improvements

  4. Duplicate lazy_static! blocks: Several identical regex patterns are defined in multiple lazy_static! blocks across both modules. Factor them into shared constants or a single block.
  5. Test fixtures should use real command output: Per project guidelines (see CLAUDE.md > Testing Strategy), tests should use include_str!("../tests/fixtures/...") with real command
  output, not synthetic/hardcoded strings. This ensures filters work against actual CLI output formats.
  6. Token savings threshold too low: Tests assert ≥30% savings, but RTK's project standard is ≥60%. Please raise the threshold or justify why these specific filters have lower savings
  expectations.

  Minor

  - Consider adding snapshot tests (assert_snapshot!) for output format validation
  - AWS module is +845 lines — could some shared patterns (JSON flattening, table formatting) be extracted to utils.rs?

  Overall solid work, looking forward to the next iteration! 🔧

PR #235, opened 2026-02-20
fix(gh): show fallback note for filtered body; fix api truncation message

polaminggkub-debug:
## Problems

Two related issues in \`gh_cmd.rs\`, same category as the body-display fix from #214.

### 1. Silent body loss in \`view_pr()\` and \`view_issue()\`

When \`filter_markdown_body()\` removes all content (body contains only badges, images, or HTML comments), the body section is silently skipped — no output, no indication that content existed.

**Example body that disappears completely:**
\`\`\`markdown

[CI](url)
screenshot
---
\`\`\`

The user sees PR metadata but has no way to know a body was present and filtered.

**Fix:** Show \`(body contained only badges/images/comments)\` so the user is aware.

### 2. Naive \`.take(20)\` with double iteration in \`run_api()\`

The non-JSON fallback in \`run_api()\` used the same naive truncation pattern that #214 fixed for \`view_pr\`:
- Calls \`raw.lines()\` twice (once to take 20, once to count total)
- Shows \`"... (truncated)"\` with no line count

**Fix:** Single pass, message shows exact count: \`"... N more lines (use gh api for full output)"\`, consistent with \`list_prs()\` and the rest of the module.

## Testing

All 412 existing tests pass.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/gh_cmd.rs
+++ b/src/gh_cmd.rs
@@ -382,6 +382,10 @@ fn view_pr(args: &[String], _verbose: u8, ultra_compact: bool) -> Result<()> {
                     filtered.push_str(&formatted);
                     print!("{}", formatted);
                 }
+            } else {
+                let note = "  (body contained only badges/images/comments)\n";
+                filtered.push_str(note);
+                print!("{}", note);
             }
         }
     }
@@ -678,6 +682,10 @@ fn view_issue(args: &[String], _verbose: u8) -> Result<()> {
                     filtered.push_str(&formatted);
                     print!("{}", formatted);
                 }
+            } else {
+                let note = "  (body contained only badges/images/comments)\n";
+                filtered.push_str(note);
+                print!("{}", note);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Hi @polaminggkub-debug — closing as stale, but **we'll re-implement
your fix on `develop` with `Co-authored-by: polaminggkub-debug`
attribution.** The bugs you spotted are still live in v0.42.0, so the
work won't be lost.

What's making a fresh PR cleaner than rebasing:
- target = `master` (would need re-target)
- CLA still unsigned (https://cla-assistant.io/rtk-ai/rtk) — blocks
  merge even after rebase
- `src/gh_cmd.rs` moved to `src/cmds/git/gh_cmd.rs` after the layout
  refactor (#826), and `view_pr`/`view_issue` signatures changed
  (return `Result<i32>`, body now flows through `format_pr_view` /
  `format_issue_view` returning a `String` instead of `print!()`
  inline)
- Truncation CAPS were standardized in #1998 via `LimitsConfig`, and
  the API-truncation message should now match that convention

Your fix logic carries over cleanly:
1. `format_pr_view` (`src/cmds/git/gh_cmd.rs:411-421`) and
   `format_issue_view` (`:658`) need an `else` branch when
   `body_filtered.is_empty()` to push the
   `(body contained only badges/images/comments)` note.
2. `run_api` truncation message gets the single-pass refactor +
   `... N more lines (use gh api for full output)` formatting.

Closing as superseded-by-reimplementation. Thanks for finding both
bugs — your `run_api` single-pass version is genuinely cleaner than
the original double-iteration.

PR #267, opened 2026-02-25
ci: replace x86_64-unknown-linux-gnu with musl target for static binaries

vanackere:
The current x86_64-unknown-linux-gnu release binary dynamically links against glibc, which causes failures on systems with an older glibc version:

rtk: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by rtk)

This PR replaces the x86_64-unknown-linux-gnu build target with x86_64-unknown-linux-musl, producing a fully statically-linked binary with no libc dependency.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,9 +35,10 @@ jobs:
             os: macos-latest
             archive: tar.gz
           # Linux
-          - target: x86_64-unknown-linux-gnu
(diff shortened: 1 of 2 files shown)


ofek:
FYI, I just opened https://github.com/rtk-ai/rtk/pull/2089 to provide an equivalent fix for Linux aarch64.

Issue #222, opened 2026-02-19
bug: rtk proxy does not stream long-running command output progressively


pavelbe:
## Summary

`rtk proxy` does not stream stdout/stderr progressively for long-running commands.
In TTY usage, output is often delivered in one block only after command completion ("silent hang" UX).

## Environment

- rtk: `0.22.0`
- OS: Linux (WSL2 Ubuntu)
- checked on: `19 February 2026`

## Reproduction

```bash
/home/pavelbe/bin/rtk proxy bash -lc 'for i in 1 2 3; do date +%s; echo probe-$i; sleep 3; done; date +%s; echo done'
```

Observed in TTY polling:

- command runs ~9 seconds,
- no intermediate output during execution,
- then full output appears at once at the end.

Example output block received only after completion:

```text
1771540196
probe-1
1771540199
probe-2
1771540202
probe-3
1771540205
done
```

## Expected

- progressive output streaming (first lines visible quickly),
- no long silent period for long-running proxy commands.

## Actual

- delayed/block output delivery until process end (or near end).

## Impact

- hard to distinguish "still running" vs "stuck",
- poor UX for deploy/test/log-tail workflows,
- causes extra retries/splits for chained operational commands.

## Notes

No duplicate open/closed issue was found with queries:
- `rtk proxy long running output`
- `rtk proxy no output until command finishes`
- `streaming stdout proxy`

pavelbe:
Update (19 February 2026): reproduced on rtk 0.22.1 as well. Same command still shows long silent period and flushes output in one block near completion.

aeppling:
Hello, sorry was not careful when i first read passing by this, i just tried this out, not an issue anymore

Closing this, fixed since :  0.27.2

aeppling:
Hey, should be solved on https://github.com/rtk-ai/rtk/releases/tag/v0.37.0

PR #268, opened 2026-02-25
stream proxy output while command is running

dragonGR:
The rtk proxy was buffering command output until the child process finished which made long-running commands look stuck in TTY mode.

So this commit switches proxy execution from buffered Command::output() to a streamed spawn() flow with piped stdout/stderr so output is forwarded incrementally while the command is still running.

What was improved:
stream stdout and stderr in real time (chunked reads)
flush each chunk immediately for better interactive UX
keep captured output for tracking/analytics
preserve child exit status behavior (still returns the same failure code when command fails)

Ref:
https://github.com/rtk-ai/rtk/issues/222

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1696,7 +1696,9 @@ fn main() -> Result<()> {
         }
 
         Commands::Proxy { args } => {
-            use std::process::Command;
+            use std::io::{Read, Write};
+            use std::process::{Command, Stdio};
+            use std::thread;
 
             if args.is_empty() {
                 anyhow::bail!(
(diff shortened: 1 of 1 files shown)


pszymkowiak:
The PR doesn't compile — utils::status_code_error is referenced but never defined anywhere in the codebase.

  Easy fix: replace the return Err(utils::status_code_error(...)) with the original pattern:
  std::process::exit(status.code().unwrap_or(1));
  The streaming logic itself looks correct. Just needs this one fix to build.

dragonGR:
Hello, yes, it’s most likely due to the recent changes you pushed. I’ve pushed an update and however I force-pushed it, so I hope that won’t cause any problems.

PR #270, opened 2026-02-27
feat(pnpm): Add filter argument support

### LINKED ISSUE #259: `pnpm --filter` command results in rtk usage output

```sh

```

KuSh:
- Add support for pnpm global filter arguments (`--filter`) across all pnpm commands (`list`, `outdated`, `install`, etc.).
- Merge pnpm filters with native command arguments (`--filter` + custom args like `--depth`).
- Implement proper handling for `Vec<String>` (text filters) and `Vec<OsString>` (OS-specific filters).
- Add unit tests for `merge_pnpm_args` and CLI parsing to validate expected behavior.

This enables users to apply workspace-specific filtering directly via `--filter @scope` before running pnpm commands, reducing token overhead by excluding irrelevant packages/dependencies from output.

Fixes: #259


aeppling (changes requested):
## Review: feat(pnpm) Add filter argument support

Good addition overall — the approach is clean, additive, follows the existing `Git` global-args pattern, and all 623 tests pass with no new clippy warnings. A few things to address before merging:

---

### Required

**`Build` and `Typecheck` silently ignore `--filter`**

When a user runs `rtk pnpm --filter @app1 build`, the filter is destructured but never used — no warning, no error. This will confuse users who expect filtering to apply.

At minimum, emit a warning when filters are provided with these subcommands:

```rust
if !filter.is_empty() {
    eprintln!("[rtk] warning: --filter is not yet supported for pnpm build, filters will be ignored");
}
```

Same for `Typecheck`. The FIXME comments acknowledge the gap but the silent discard is a UX issue that should not ship as-is.

---

### Recommended (not blocking, but would strengthen the PR)

**Missing test coverage for edge cases:**

- Empty filters: verify `merge_pnpm_args(&[], &args)` returns `args` unchanged (backward-compat)
- Multiple `--filter` flags in CLI parsing: `["rtk", "pnpm", "--filter", "@app1", "--filter", "@app2", "list"]` — only single filter is tested today
- `Other` passthrough with filters: the `OsString` path is only tested at the helper level, not through CLI parsing → dispatch
- `Build`/`Typecheck` with filters: document the expected behavior (filters ignored + warning)

---

### Nits

- The doc comment on `merge_pnpm_args_os` is a copy-paste of `merge_pnpm_args` — should mention it's the `OsString` variant
- `args.iter().map(|arg| arg.to_string())` in `merge_pnpm_args` could be `.chain(args.iter().cloned())` — more idiomatic since `String: Clone`

aeppling (approved):
After reviewing the PR we could merge this, user has informed with // FIXME and this is not blocking nor an real issue but a limitation.
Limitation is flagged and PR is valid.

src/main.rs
```diff
@@ -1404,33 +1451,49 @@ fn main() -> Result<()> {
...
+                }
+                PnpmCommands::Build { args } => {
+                    let mut build_args: Vec<String> = vec!["build".into()];
+                    build_args.extend(args);
+                    let os_args: Vec<OsString> =
+                        build_args.into_iter().map(OsString::from).collect();
+                    pnpm_cmd::run_passthrough(&merge_pnpm_args_os(&filter, &os_args), cli.verbose)?;
+                }
+                PnpmCommands::Typecheck { args } => {
+                    tsc_cmd::run(&args, cli.verbose)?;
```
KuSh: NOTE: Readd comment

src/main.rs
```diff
@@ -1182,6 +1186,49 @@ fn shell_split(input: &str) -> Vec<String> {
...
+                let cmd_name = match command {
+                    PnpmCommands::Build { .. } => "build",
+                    PnpmCommands::Typecheck { .. } => "tsc",
+                    _ => unreachable!(),
+                };
+                let msg = format!(
+                    "[rtk] warning: --filter is not yet supported for pnpm {}, filters preceding the subcommand will be ignored",
+                    cmd_name
+                );
+                eprintln!("{}", msg);
```
KuSh: Note: remove, now that it's printed by main

test-results/.last-run.json
KuSh, line 1: note: remove

get
```diff
@@ -0,0 +1,12 @@
+{
```
KuSh: note: remove

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
KuSh: note: remove

KuSh:
Thanks for lending a helping hand @pszymkowiak. I've made some fixes and additions:

- Now that PnpmCommand::Build is simply another passthrough I've removed it completely
- I've also removed the eprintln! call in validate_pnpm_filters to avoid double printed warning
- There were some files added inadvertently, I've removed them
- A git test was failing on my linux computer. I've fixed it (`cmd.get_program()` is returning `/usr/bin/git` instead of just `git`)

PR #274, opened 2026-02-28
fix: surface build failures in go test summary

riclib:
## Summary

- `go test -json` emits build failures as `build-output`/`build-fail` actions with `ImportPath` (not `Package`/`Test`), followed by a package-level `fail` with a `FailedBuild` field
- The JSON parser only counted failures when `event.test.is_some()`, so **build errors were silently dropped** from the summary — showing "891 passed" when a package failed to compile
- Now parses `build-output`, `build-fail`, and `FailedBuild` fields from the Go test JSON protocol
- Build failures count toward the failure total and display with `[build failed]` plus compiler error lines

**Before:**
```
✓ Go test: 891 passed in 74 packages
```

**After:**
```
Go test: 891 passed, 1 failed in 74 packages

📦 stream [build failed]
  handler_test.go:64:9: undefined: undefinedVariable
```

## Test plan

- [x] Introduced intentional compile error (`var _ = undefinedVariable`) in a test file
- [x] Verified rtk previously showed `✓ 891 passed` (masking the failure)
- [x] Verified rtk now shows `891 passed, 1 failed` with the build error details
- [x] Verified clean run still shows `✓ 904 passed in 73 packages`
- [x] `cargo build --release` succeeds with no new warnings

### DIFF
--- a/src/go_cmd.rs
+++ b/src/go_cmd.rs
@@ -21,6 +21,10 @@ struct GoTestEvent {
     output: Option<String>,
     #[serde(rename = "Elapsed")]
     elapsed: Option<f64>,
+    #[serde(rename = "ImportPath")]
+    import_path: Option<String>,
+    #[serde(rename = "FailedBuild")]
+    failed_build: Option<String>,
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Good catch. go test -json emits build-output/build-fail events (with ImportPath, not Package) when compilation fails — these were silently dropped before. The fix correctly handles both
  the buffering of build error lines and the package-level failure flag.

  39 go_cmd tests pass. LGTM.

PR #293, opened 2026-03-01
feat(kubectl): add global flags --context, --kubeconfig, -n, -l, --as

mariusvniekerk:
## Summary
- Add `KubectlGlobalArgs` struct with `--context`, `--kubeconfig`, `-n/--namespace`, `-l/--selector`, and `--as` flags
- Flatten into `pods`, `services`, and `logs` subcommands via `#[command(flatten)]`
- Forward all global args to the underlying `kubectl` binary
- Previously only `-n` was supported on pods/services, and logs had no namespace/context support

## Test plan
- [x] 5 new CLI parsing tests covering all flags and combinations
- [x] All 435 tests pass
- [x] Manual verification: `rtk kubectl get --context foo -n blah somecrd a` correctly passes through

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -924,6 +924,22 @@ fn rewrite_segment_inner(
         return Some(rewritten);
     }
 
+    if rule.rtk_cmd == "rtk kubectl" {
+        for (resource, alias) in [("pods", "pods"), ("services", "services")] {
+            let prefix = format!("kubectl get {resource}");
+            if let Some(rest) = strip_word_prefix(cmd_part, &prefix) {
+                if kubectl_alias_args_supported(rest) {
+                    let rewritten = if rest.is_empty() {
+                        format!("rtk kubectl {alias}{redirect_suffix}")
+                    } else {
+                        format!("rtk kubectl {alias} {rest}{redirect_suffix}")
+                    };
+                    return Some(rewritten);
+                }
+            }
+        }
+    }
+
     // #196: gh with --json/--jq/--template produces structured output that
     // rtk gh would corrupt — skip rewrite so the caller gets raw JSON.
     if rule.rtk_cmd == "rtk gh" {
@@ -985,6 +1001,47 @@ fn strip_word_prefix<'a>(cmd: &'a str, prefix: &str) -> Option<&'a str> {
     } else {
         None
     }
+}
+
+fn kubectl_alias_args_supported(args: &str) -> bool {
+    let tokens = tokenize(args);
+    let mut index = 0;
+
+    while index < tokens.len() {
+        let token = &tokens[index];
+        if token.kind != TokenKind::Arg {
+            return false;
+        }
+
+        match token.value.as_str() {
+            "-A" | "--all-namespaces" => index += 1,
+            "-n" | "--namespace" | "--context" | "--kubeconfig" | "--as" | "-l" | "--selector" => {
+                let Some(value) = tokens.get(index + 1) else {
+                    return false;
+                };
+                if value.kind != TokenKind::Arg || value.value.starts_with('-') {
+                    return false;
+                }
+                index += 2;
+            }
+            value
+                if [
+                    "--namespace=",
+                    "--context=",
+                    "--kubeconfig=",
+                    "--as=",
+                    "--selector=",
+                ]
+                .iter()
+                .any(|prefix| value.strip_prefix(prefix).is_some_and(|v| !v.is_empty())) =>
+            {
+                index += 1;
+            }
+            _ => return false,
+        }
+    }
+
+    true
 }
 
 #[cfg(test)]
@@ -2286,6 +2343,38 @@ mod tests {
         assert_eq!(
             rewrite_command_no_prefixes("kubectl describe pod mypod", &[]),
             Some("rtk kubectl describe pod mypod".into())
+        );
+    }
+
+    #[test]
+    fn test_rewrite_kubectl_get_pods_to_direct_alias() {
+        assert_eq!(
+            rewrite_command_no_prefixes("kubectl get pods -n default", &[]),
+            Some("rtk kubectl pods -n default".into())
+        );
+    }
+
+    #[test]
+    fn test_rewrite_kubectl_get_services_to_direct_alias() {
+        assert_eq!(
+            rewrite_command_no_prefixes("kubectl get services -A", &[]),
+            Some("rtk kubectl services -A".into())
+        );
+    }
+
+    #[test]
+    fn test_rewrite_kubectl_get_pods_with_output_flag_preserves_get() {
+        assert_eq!(
+            rewrite_command_no_prefixes("kubectl get pods -o json", &[]),
+            Some("rtk kubectl get pods -o json".into())
+        );
+    }
+
+    #[test]
+    fn test_rewrite_kubectl_get_named_pod_preserves_get() {
+        assert_eq!(
+            rewrite_command_no_prefixes("kubectl get pods my-pod", &[]),
+            Some("rtk kubectl get pods my-pod".into())
         );
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,7 +27,7 @@ use cmds::system::{
 
 use anyhow::{Context, Result};
 use clap::error::ErrorKind;
-use clap::{Parser, Subcommand, ValueEnum};
+use clap::{Args, Parser, Subcommand, ValueEnum};
 use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 
@@ -1035,6 +1035,45 @@ enum ComposeCommands {
     Other(Vec<OsString>),
 }
 
+#[derive(Args, Clone, Debug)]
+struct KubectlGlobalArgs {
+    /// The name of the kubeconfig context to use
+    #[arg(long)]
+    context: Option<String>,
+    /// Path to the kubeconfig file
+    #[arg(long)]
+    kubeconfig: Option<String>,
+    /// Namespace scope for this request
+    #[arg(short, long)]
+    namespace: Option<String>,
+    /// Username to impersonate
+    #[arg(long = "as")]
+    impersonate: Option<String>,
+}
+
+impl KubectlGlobalArgs {
+    fn to_args(&self) -> Vec<String> {
+        let mut args = Vec::new();
+        if let Some(ref ctx) = self.context {
+            args.push("--context".to_string());
+            args.push(ctx.clone());
+        }
+        if let Some(ref kc) = self.kubeconfig {
+            args.push("--kubeconfig".to_string());
+            args.push(kc.clone());
+        }
+        if let Some(ref ns) = self.namespace {
+            args.push("-n".to_string());
+            args.push(ns.clone());
+        }
+        if let Some(ref imp) = self.impersonate {
+            args.push("--as".to_string());
+            args.push(imp.clone());
+        }
+        args
+    }
+}
+
 #[derive(Debug, Subcommand)]
 enum KubectlCommands {
     /// Get Kubernetes resources (compact for pods/services)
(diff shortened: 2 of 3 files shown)


mariusvniekerk:
@aeppling rebased this pr.

src/main.rs
```diff
@@ -851,35 +851,111 @@ enum ComposeCommands {
...
+    kubeconfig: Option<String>,
+    /// Namespace scope for this request
+    #[arg(short, long)]
+    namespace: Option<String>,
+    /// Label selector (e.g. -l app=web)
+    #[arg(short = 'l', long)]
+    selector: Option<String>,
+    /// Username to impersonate
+    #[arg(long = "as")]
+    impersonate: Option<String>,
```
KuSh, line 1053: Are you sure those args have **optional** values?

Also `-l`/`--selector` doesn't seem to be a global arg, only available for delete/get and describe subcommands

src/main.rs
```diff
@@ -851,35 +851,111 @@ enum ComposeCommands {
...
+        #[command(flatten)]
+        global: KubectlGlobalArgs,
         #[arg(short, long)]
         container: Option<String>,
     },
+    /// Get a resource (kubectl get pods, kubectl get services, ...)
+    Get {
+        #[command(subcommand)]
+        resource: KubectlGetResource,
+    },
```
KuSh: This should be handled by rewrite rules:

`rtk rewrite kubectl get pods`, should return `rtk kubectl pods` instead of the actual `rtk kubectl get pods`

src/main.rs
```diff
@@ -851,35 +851,111 @@ enum ComposeCommands {
...
 
+#[derive(Subcommand)]
+enum KubectlGetResource {
+    /// List pods with filtered output
+    Pods {
+        #[command(flatten)]
+        global: KubectlGlobalArgs,
+        /// All namespaces
+        #[arg(short = 'A', long)]
+        all: bool,
```
KuSh, line 1136: long seems to be --all-namespaces according to kubectl doc

KuSh (changes requested):
Hi, thanks for your PR. There seems to be errors on kubectl args, and the "get xxx" part should be handled by rewrite.

src/main.rs
mariusvniekerk: Fixed in 16227b7. `--context`, `--kubeconfig`, `--namespace`, and `--as` require values and are now true kubectl-level global flags. `-l`/`--selector` is scoped only to the pods and services aliases. `-A` now exposes the correct `--all-namespaces` long form.

<sup>generated by a clanker</sup>

src/main.rs
mariusvniekerk: Fixed in 16227b7. The rewrite engine now maps safe `kubectl get pods/services` forms to `rtk kubectl pods/services`, while preserving generic forms such as `-o json` or named resources on `rtk kubectl get ...`.

<sup>generated by a clanker</sup>

src/main.rs
mariusvniekerk: Fixed in 16227b7. The aliases now accept `-A` and `--all-namespaces`.

<sup>generated by a clanker</sup>

src/discover/registry.rs
```diff
@@ -924,6 +924,22 @@ fn rewrite_segment_inner(
         return Some(rewritten);
     }
 
+    if rule.rtk_cmd == "rtk kubectl" {
+        for (resource, alias) in [("pods", "pods"), ("services", "services")] {
```
KuSh, line 928: question (non-blocking): `resource` and `alias` always carry the same value here, so the destructuring never does anything — `for resource in ["pods", "services"]` would read identically.

If the tuple was meant to map short forms, then the aliases are precisely what's missing: `k8s_get_target` in `src/cmds/cloud/container.rs` already resolves `po | pod | pods` and `svc | service | services`, so today `kubectl get po` rewrites to `rtk kubectl get po` rather than `rtk kubectl pods`. No behaviour is lost (both paths end up in `k8s_pods`), which is exactly why the alias half of the tuple is dead.

So either collapse it to a plain array, or complete it:

```rust
for (resource, alias) in [
    ("po", "pods"), ("pod", "pods"), ("pods", "pods"),
    ("svc", "services"), ("service", "services"), ("services", "services"),
] {
```

src/main.rs
KuSh: You fixed `--all-namespaces` for `kubectl` but not for `oc` — `OcCommands::Pods` and `OcCommands::Services` still declare `#[arg(short = 'A', long)]`, which clap expands to `--all`.

Verified against a `kubectl`/`oc` shim on 16227b7:

- `rtk kubectl pods --all-namespaces` → `kubectl get pods -o json -A` ✅
- `rtk oc pods --all-namespaces` → clap rejects the flag, RTK falls back to raw passthrough and runs `oc pods --all-namespaces`, which `oc` itself rejects ❌

Same fix as the kubectl side, on both `OcCommands::Pods` and `OcCommands::Services`:

```rust
#[arg(short = 'A', long = "all-namespaces")]
all: bool,
```

src/discover/registry.rs
```diff
@@ -987,6 +1003,47 @@ fn strip_word_prefix<'a>(cmd: &'a str, prefix: &str) -> Option<&'a str> {
...
+                .any(|prefix| value.strip_prefix(prefix).is_some_and(|v| !v.is_empty())) =>
+            {
+                index += 1;
+            }
+            _ => return false,
+        }
+    }
+
+    true
+}
```
KuSh, line 1045: This whitelist has to stay in lockstep with the clap definitions in `main.rs`, and nothing enforces it.

Every token accepted here must be parseable by `rtk kubectl pods` / `rtk kubectl services`. Right now the two lists match exactly, so the rewrite is correct. But if they ever drift — a flag added here without a matching field on `KubectlGlobalArgs`, or a `long = ` renamed on the alias subcommand — the rewriter emits `rtk kubectl pods <flag>`, clap fails to parse it, and RTK silently degrades to a raw `kubectl` passthrough, losing the compaction this rewrite exists to provide. The failure is invisible: no error, just unfiltered output.

Worth a test that feeds every whitelisted token through `Cli::try_parse_from(["rtk", "kubectl", "pods", ...])` and asserts it parses, so the two sides can't drift apart unnoticed.

src/main.rs
```diff
@@ -1900,21 +1973,54 @@ fn run_cli() -> Result<i32> {
...
+                };
+                if let Some(selector) = selector {
+                    args.push("-l".to_string());
+                    args.push(selector);
+                }
+                if all {
+                    args.push("-A".to_string());
+                }
                 container::run(container::ContainerCmd::KubectlServices, &args, cli.verbose)?
             }
```
KuSh, line 2010: These two arms are byte-for-byte identical apart from `KubectlPods` / `KubectlServices`, and because the assembly is inlined in the match it can't be tested: the five new tests all assert on `KubectlGlobalArgs::to_args()`, never on the vector actually handed to `container::run`.

Extracting it alongside `build_kubectl_get_args` removes the duplication and makes the behaviour assertable:

```rust
fn build_kubectl_alias_args(
    global: &KubectlGlobalArgs,
    selector: Option<String>,
    all: bool,
) -> Vec<String>
```

That would also pin down a rule which is currently silent and untested: when `-A` is passed, `-n` is dropped (`rtk kubectl pods -A -n foo` → `kubectl get pods -o json -A`). That matches kubectl's own precedence so it's the right call, but it deserves an explicit test rather than being an emergent property of two separate `if all` branches.

src/main.rs
```diff
@@ -1035,6 +1037,51 @@ enum ComposeCommands {
...
+            args.push(imp.clone());
+        }
+        args
+    }
+
+    fn to_args_without_namespace(&self) -> Vec<String> {
+        let mut global = self.clone();
+        global.namespace = None;
+        global.to_args()
+    }
```
KuSh, line 1082: nit: this clones the whole struct just to null a single field. A single entry point avoids both the clone and the second method:

```rust
fn to_args_with(&self, include_namespace: bool) -> Vec<String>
```

with the `-n` push guarded by the flag, and `to_args()` delegating as `self.to_args_with(true)`.

KuSh (changes requested):
Thanks for the follow-up. I re-verified the three open threads against 16227b7 with a `kubectl`/`oc` shim, and two of them are genuinely fixed — I've resolved those:

- **Optional values / `-l` scoping** — fixed. The four global flags now require values, and `-l`/`--selector` is scoped to the `pods`/`services` aliases only. ✅
- **`get pods` handled by rewrite** — fixed. `rtk rewrite "kubectl get pods"` → `rtk kubectl pods`, while `-o json` and named-resource forms correctly keep `rtk kubectl get`. ✅
- **`--all-namespaces`** — fixed for `kubectl`, but `oc` was left behind. Details inline. ❌

The branch itself is clean: `cargo fmt --check` OK, `cargo clippy --all-targets` zero warnings, 2449 + 6 + 11 tests pass.

**Blocking**

1. `oc` still exposes `--all` instead of `--all-namespaces` — `rtk oc pods --all-namespaces` fails to parse and degrades to a raw passthrough that `oc` itself rejects.

**Non-blocking, but worth folding in before merge**

2. The rewrite whitelist can silently drift out of sync with the clap definitions, and the failure mode is invisible.
3. The `Pods` / `Services` arms are duplicated, and the arg assembly they contain isn't covered by any test.
4. The `resource` / `alias` tuple is dead as written.
5. `to_args_without_namespace` clones the struct to null one field.

**Two points that don't map onto a diff line**

- `docs/usage/FEATURES.md:902-904` still documents `rtk kubectl pods [-n ns] [-A]`. The new `--context`, `--kubeconfig`, `--as` and `-l`/`--selector` flags aren't mentioned there or anywhere else in the docs.
- Context on the rewrite change, since I'm the one who asked for it: `rtk kubectl get pods` and `rtk kubectl pods` produce byte-identical output today — both end up running `kubectl get pods -o json` through `format_kubectl_pods` — so the rewrite is a cosmetic change to the emitted command string rather than a behavioural one. That's fine and it's what I wanted, but it's also why point 2 matters: the whitelist buys presentation, and pays for it with a silent-degradation risk. Keeping the two sides pinned by a test makes that trade worth it.

src/main.rs
```diff
@@ -1045,18 +1092,20 @@ enum KubectlCommands {
...
     Pods {
-        #[arg(short, long)]
-        namespace: Option<String>,
+        /// Label selector (e.g. -l app=web)
+        #[arg(short = 'l', long)]
+        selector: Option<String>,
         /// All namespaces
-        #[arg(short = 'A', long)]
+        #[arg(short = 'A', long = "all-namespaces")]
         all: bool,
```
KuSh, line 1100: Renaming `long` to `long = "all-namespaces"` silently removes `--all`, which this crate accepted before.

On `develop` the attribute was `#[arg(short = 'A', long)]`, so clap derived the long form from the field name and `rtk kubectl pods --all` was valid. It no longer is — and it doesn't fail loudly. `pods` still matches the known subcommand, so `--all` becomes an unknown-argument error *inside* `Pods`; `main()` catches it, `run_fallback` (`src/main.rs:1298`) records the clap error via `record_parse_failure_silent` and never prints it, then execs `args` verbatim.

Verified against a `kubectl` shim on 16227b7:

- before: `rtk kubectl pods --all` → `kubectl get pods -o json -A`
- after:  `rtk kubectl pods --all` → `kubectl pods --all`

Real kubectl answers that with `error: unknown command "pods" for "kubectl"` — an error naming neither `--all` nor the actual problem — and the compaction is lost. Same for `rtk kubectl services --all`.

Worth noting the two halves of this PR now disagree: after the rename, `--all` works on `rtk oc pods` (still `long`) but not on `rtk kubectl pods`. Whichever way you settle the `oc` comment, the two should end up spelling it the same way.

Two ways out, either is fine by me:

1. Keep back-compat — `#[arg(short = 'A', long = "all-namespaces", alias = "all")]`. `alias` is hidden from `--help`, so the correct kubectl spelling is the one advertised while existing `--all` callers keep working.
2. Accept the break, but mark it. `CHANGELOG.md` is release-please-generated from the commits, and neither commit here carries a `!` or a `BREAKING CHANGE:` footer, so as it stands this ships as a plain `fix:` with nothing in the changelog telling users their `--all` scripts stopped working.

To be clear, the silent-passthrough-on-parse-error behaviour is pre-existing and not yours to fix — it's just what turns a flag rename into an invisible break, which is why the rename deserves one of the two treatments above.

KuSh (commented):
One more, found while re-checking the `--all-namespaces` rename — it silently drops the previously-valid `--all` spelling. Details inline.

PR #306, opened 2026-03-02
Replace lazy_static with std::sync::LazyLock

jbgriesner:
Replaces all uses of the `lazy_static` crate with `std::sync::LazyLock`, which has been stable since Rust 1.80.

  **Motivation:**                              
  - Removes one external dependency (`lazy_static = "1.4"`)
  - Uses std

  **No behavior change.** LazyLock initializes on first access, exactly like lazy_static. All regex patterns remain identical.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -589,7 +589,6 @@ dependencies = [
  "colored",
  "dirs",
  "ignore",
- "lazy_static",
  "regex",
  "rusqlite",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ anyhow = "1.0"
 ignore = "0.4"
 walkdir = "2"
 regex = "1"
-lazy_static = "1.4"
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["preserve_order"] }
 colored = "2"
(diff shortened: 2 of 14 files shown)


jbgriesner:
> 2. CI shows 0/0 checks. Could you check why the workflow isn't triggering? We need `cargo fmt --all && cargo clippy --all-targets && cargo test` to pass before merge.


The 0/0 CI checks are a GitHub limitation with fork PRs: workflows run from the fork's own .github/workflows/ directory, not from the upstream repo. Since my fork doesn't replicate the upstream CI setup, no checks trigger automatically.                                                                                                           
                                                                   
I've verified locally that all three gates pass: cargo fmt --all && cargo clippy --all-targets && cargo test                                                                                                                 
  430 tests pass, no clippy errors.

pszymkowiak:
Hi! Two things needed before we can review:

1. **Retarget to `develop`** — this PR targets `master`, but all PRs should target `develop`. You can change the base branch in the PR settings (right sidebar).
2. **Sign the CLA** — if not already done, please sign at https://cla-assistant.io/rtk-ai/rtk

Thanks!

KuSh:
Thanks for the cleanup. This landed as PR #3244 (merged 2026-07-28), which replaced every lazy_static! block with std::sync::LazyLock and removed the dependency from Cargo.toml, so develop no longer references lazy_static anywhere. The file layout also changed since this branch was opened, with the command modules moved under src/cmds and shared code under src/core. If you find a spot that still compiles a regex eagerly or was missed by that pass, please comment here or reopen.

PR #367, opened 2026-03-06
Add support for CLAUDE_CONFIG_DIR

GertBurger:
Claude will honor the environmental variable CLAUDE_CONFIG_DIR (if specified) to locate its config directory, see https://code.claude.com/docs/en/settings#environment-variables

This adapts RTK to match that behaviour. Only tested on linux.

Adds `temp-env` dev dependency to properly test env variable usage in new and some existing tests.

A few pre-existing tests that directly or indirect use `RTK_DB_PATH` result in flaky tests, I didn't fix those.
edit: maybe this PR should not adjust existing tests...


GertBurger:
Done, that was added after I posted this :P

PR #390, opened 2026-03-06
fix: normalize venv/path-prefixed commands in rewrite hook

mhcoen:
Commands invoked via venv or full path (e.g., .venv/bin/pytest, /usr/local/bin/ruff) were not matched by the rewrite hook because the patterns only recognize bare command names. Claude Code non-deterministically chooses between bare command names and full venv paths depending on what it has seen in the project (e.g., after discovering a .venv/ directory), so both forms must be handled.

This adds path normalization in the shell hook that strips directory prefixes for matching only, preserving the original path in the rewritten command so the correct binary is executed.

Handles both cases:

Simple prepend: .venv/bin/pytest -v → rtk .venv/bin/pytest -v
Transformation: /usr/bin/cat file → rtk read file

### DIFF
--- a/hooks/cursor/rtk-rewrite.sh
+++ b/hooks/cursor/rtk-rewrite.sh
@@ -38,10 +38,31 @@ if [ -z "$CMD" ]; then
   exit 0
 fi
 
+# Normalize path-prefixed commands for matching only.
+# Claude Code often invokes tools via venv or full path, e.g.
+# ".venv/bin/pytest -v" or "/usr/local/bin/ruff check .".
+# Strip the directory prefix from the first word so rtk rewrite
+# can match the bare command name.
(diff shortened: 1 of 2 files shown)


pszymkowiak:
Hi! Thanks for the contribution! Since March 6, all PRs should target the **`develop`** branch instead of `master` (see [CONTRIBUTING.md](https://github.com/rtk-ai/rtk/blob/master/CONTRIBUTING.md#5-merge-into-develop)).

Could you update the base branch? Click **Edit** at the top right of this PR and change it from `master` to `develop`.

Thanks!

mhcoen:
Sure. Done.

mhcoen:
I'm closing this as there are some more serious path handling problems in the architecture that need to be dealt with.

Issue #384, opened 2026-03-06
Windows: rtk init --uninstall does not work


FlorianBruniaux:
## Summary

The `rtk init -g --uninstall` command does not work on Windows. The uninstall process likely assumes a Unix-style environment (bash, Unix paths) and fails silently or errors out.

## Environment

- OS: Windows (reported with Cygwin)
- RTK version: latest

## Expected behavior

```
rtk init -g --uninstall
```

Should cleanly remove:
- `%USERPROFILE%\.claude\hooks\rtk-rewrite.sh` (or equivalent)
- `%USERPROFILE%\.claude\RTK.md`
- `@RTK.md` reference from `%USERPROFILE%\.claude\CLAUDE.md`
- RTK hook entry from `%USERPROFILE%\.claude\settings.json`

## Actual behavior

Uninstall does not complete successfully on Windows.

## Additional context

Related to the broader Windows support gap: the hook system (`rtk-rewrite.sh`) is a bash script and is already flagged as unsupported on Windows during `rtk init -g`. The uninstall path likely has the same assumptions about the shell environment.

Reported by Benjamin on Discord (Clean Coder community).

pszymkowiak:
> [w] **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 90% |

### Summary

The `rtk init -g --uninstall` command fails on Windows due to Unix-specific assumptions in the uninstall path; needs Windows-compatible path and shell handling.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

PR #392, opened 2026-03-06
fix(init-uninstall): uninstall removes --claude-md artifacts on Windows

aeppling:
Fix for : https://github.com/rtk-ai/rtk/issues/384

  Core fix (src/init.rs — uninstall()):
  - Added step after existing @RTK.md removal: checks for <!-- rtk-instructions marker in CLAUDE.md and removes the block using the existing remove_rtk_block() helper
  - Handles both artifacts in a single pass (covers edge case where both @RTK.md and the instructions block exist, e.g. after running both rtk init -g and rtk init -g --claude-md)
  - If CLAUDE.md becomes empty after cleanup, deletes the file

  Hardening:
  - Extracted RTK_BLOCK_START and RTK_BLOCK_END constants — all marker matching across uninstall(), upsert_rtk_block(), remove_rtk_block(), patch_claude_md(), and show_config() now uses these instead of duplicated string literals
  - When uninstall finds nothing to remove, prints which paths were checked to help users diagnose path issues
  - resolve_claude_dir() error message now shows %USERPROFILE% on Windows instead of $HOME


KuSh (approved):
LGTM 🚀️

PR #394, opened 2026-03-06
feat(xcodebuild): add support

### LINKED ISSUE #376: Feature Request: Add support for `xcodebuild` and subcommands
iOS/macOS projects rely heavily on xcodebuild for compilation, testing, archiving, and code signing. These commands produce extremely verbose output that would benefit significantly from RTK's token-optimized filtering.

## Suggested commands

- `xcodebuild build` — build a project/workspace
- `xcodebuild test` — run unit and UI tests
- `xcodebuild archive` — create an archive for distribution
- `xcodebuild clean` — clean build artifacts
- `xcodebuild -showBuildSettings` — display build settings
- `xcodebuild -list` — list schemes, targets, and configurations
- `xcodebuild build-for-testing` / `xcodebuild test-without-building` — split build and test phases
- `xcodebuild -exportArchive` — export an archive to IPA/PKG

xcodebuild is notoriously verbose — a typical build can produce thousands of lines of output, most of which is redundant for an LLM context window. Filtering down to warnings, errors, and build results would yield massive token savings.

rursache:
## Summary
- add a dedicated `rtk xcodebuild` command with mode-aware filtering for build-like, test, info, and fallback invocations
- preserve high-value `xcodebuild test` context while stripping repetitive build noise, and keep `-list` / `-showBuildSettings` near-lossless
- wire `xcodebuild` into CLI discovery, auto-rewrite hooks, docs, changelog, smoke coverage, and hook rewrite tests

## Details
- added `src/xcodebuild_cmd.rs` with conservative filtering, raw fallback behavior, `-v` passthrough, and exit-code propagation
- added fixture-driven and fake-binary test coverage for build success/failure, test success/failure, archive/export, and info output
- registered `xcodebuild` in `src/main.rs`, `src/discover/rules.rs`, and `src/discover/registry.rs`
- updated `.claude/hooks/rtk-suggest.sh` and `hooks/test-rtk-rewrite.sh`
- updated `README.md`, `INSTALL.md`, `ARCHITECTURE.md`, `CLAUDE.md`, `CHANGELOG.md`, and validation/smoke scripts

## Verification
- `cargo fmt --all --check`
- `cargo test --quiet`
- `bash hooks/test-rtk-rewrite.sh`
- `bash scripts/validate-docs.sh`
- `cargo run --quiet -- xcodebuild --help`

## Notes
- I also ran `cargo clippy --all-targets -- -D warnings` per the contributing guide.
- It currently fails on pre-existing `develop`-branch warnings and clippy findings outside this feature (for example in `cc_economics.rs`, `container.rs`, `gain.rs`, `git.rs`, `init.rs`, `tracking.rs`, and other older modules). I did not fold a repo-wide cleanup into this PR.

Closes #376

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,7 +299,6 @@ After install, **restart Claude Code**.
 | `pytest` | `rtk pytest` |
 | `pip list/install` | `rtk pip ...` |
 | `go test/build/vet` | `rtk go ...` |
-| `golangci-lint run` | `rtk golangci-lint run` |
 | `xcodebuild ...` | `rtk xcodebuild ...` |
 | `golangci-lint` | `rtk golangci-lint` |
 | `docker ps/images/logs` | `rtk docker ...` |


aeppling:
This is a good implementation but there is some missing requirements and some out of scope actions.

## Testing 

CONTRIBUTING.md PR checklist requires:
Token savings >=60% verified

None of the 12 tests in xcodebuild_cmd.rs verify token savings. A test like this is required to ensure >=60%

## Scope

- This PR modifies expected outputs for commands completely unrelated to xcodebuild in hooks/test-rtk-rewrite.sh 

- Lines 134-156 of hooks/test-rtk-rewrite.sh overhaul the HOOK resolution logic (adding SCRIPT_DIR, REPO_ROOT, PATH manipulation). While potentially useful, this is infrastructure work unrelated to xcodebuild.

Those look like enhancements but this does not belong to the PR which focus on adding support for xcodebuild.

- Version bump doesn't belong in feature PR                               

## Bug 

test-without-building classified as BuildLike instead of Test

## Contribution Requirements  (to be validated)

We want to enforce commit DCO signing for PR authors, i need to check with maintainers team if this is taking effect by now, if so you'll need to sign this. (sign latest commit).

README.md
aeppling, line 1: adding "golangci-lint run" in README.md is out of scope (and already defined below)

aeppling: @rursache  I just noticed this last out-of-scope modification, then we should be good once removed.

Thanks for contributing

rursache: @aeppling fixed, sorry for missing this!

PR #419, opened 2026-03-08
fix: preserve gh run view passthrough args

### LINKED ISSUE #416: Issue accepting arguments for `gh run`
Tried various combinations of `rtk gh run view --log-failed --job 123456789` but get:

```
flag needs an argument: --job

Usage:  gh run view [<run-id>] [flags]
```

tsubasakong:
## Summary
- preserve raw passthrough for flag-only forms like `gh run view --log-failed --job <job-id>` by checking passthrough flags before extracting a positional run ID
- treat short `-j` / `-a` forms the same as `--job` / `--attempt` when parsing `gh run view` arguments
- add regression tests covering short-flag and flag-only argument combinations

Fixes #416.

## Validation
- `cargo fmt --all --check`
- `cargo test`

### DIFF
--- a/src/gh_cmd.rs
+++ b/src/gh_cmd.rs
@@ -128,7 +128,9 @@ fn extract_identifier_and_extra_args(args: &[String]) -> Option<(String, Vec<Str
         "--jq",
         "-t",
         "--template",
+        "-j",
         "--job",
+        "-a",
         "--attempt",
     ];
     let mut identifier = None;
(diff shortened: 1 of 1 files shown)


saimonmoore (approved):
Cool. Thanks a lot!

pszymkowiak:
Thanks @tsubasakong — the fix is correct and still needed: I checked current `develop` and the bug is unchanged (`-j`/`-a` missing from the value-flag list, and `view_run` extracts the run-id before the passthrough check, which reorders the args). Nicely tested too.

It needs a rebase onto `develop` — it's currently conflicting because the `src/` reorganization (#826) moved the file to `src/cmds/git/gh_cmd.rs`. Heads-up while rebasing: `view_run` now returns `Result<i32>` (not `Result<()>`) and uses `resolved_command("gh")` instead of `Command::new("gh")`, so the passthrough `return` needs to be adapted to the new return type.

```bash
git fetch origin && git rebase origin/develop
```

Also the CLA still shows unsigned — please sign it so CI can run.

Ping me once it's rebased and I'll do a final pass.

PR #452, opened 2026-03-10
feat: add Windows hook support for rtk init --global

daixudk96-max:
## Summary

- Removes Unix-only `#[cfg(unix)]` guards from `src/init.rs`, enabling hook-based mode on Windows
- Adds `hooks/rtk-rewrite.js`: a Node.js PreToolUse hook that calls `rtk rewrite` via `execFileSync`, with fail-open behavior (silently passes through on error)
- Platform-aware helpers: `hook_filename()`, `hook_content()`, `build_hook_command()` select `.sh` on Unix and `.js` on Windows
- Windows hook command format: `node "C:\Users\...\.claude\hooks\rtk-rewrite.js"` (object-style `{"type":"command","command":"..."}`)
- Adds 3 Windows-specific tests: `test_windows_hook_uses_rtk_rewrite`, `test_insert_hook_entry_windows_command`, `test_hook_already_present_windows_command`

## Motivation

Currently `rtk init --global` falls back to `--claude-md` mode on Windows, which injects ~137 lines into CLAUDE.md and relies on the LLM to "remember" to prefix commands with `rtk`. This is unreliable and wastes tokens. With this PR, Windows users get the same transparent hook-based interception that Unix users enjoy.

## Test plan

- [ ] `cargo test init::tests` passes (all existing + new Windows tests)
- [ ] `cargo check` clean (no new warnings from this PR)
- [ ] On Windows: `rtk init --global --auto-patch` installs `rtk-rewrite.js` to `~/.claude/hooks/` and patches `settings.json`
- [ ] On Unix: behavior unchanged (still uses `rtk-rewrite.sh`)

## Known issue

On Windows with default stack size, `cargo run` may hit `STATUS_STACK_OVERFLOW` due to large embedded string constants. Adding `.cargo/config.toml` with `rustflags = ["-C", "link-arg=/STACK:8388608"]` resolves this. Not included in this PR as it may be environment-specific.

### DIFF
--- a/hooks/rtk-rewrite.js
+++ b/hooks/rtk-rewrite.js
@@ -1,0 +1,55 @@
+#!/usr/bin/env node
+const { execFileSync } = require("node:child_process");
+
+function readInput() {
+  const chunks = [];
+  process.stdin.on("data", (chunk) => chunks.push(chunk));
+  process.stdin.on("end", () => {
+    const input = Buffer.concat(chunks).toString("utf8");
+    processInput(input);
+  });
+  process.stdin.resume();
+}
+
+function processInput(inputRaw) {
+  let input;
+  try {
+    input = JSON.parse(inputRaw);
+  } catch {
+    return;
+  }
+
+  const command = input?.tool_input?.command;
+  if (!command) {
+    return;
+  }
+
+  let rewritten;
+  try {
+    rewritten = execFileSync("rtk", ["rewrite", command], {
+      encoding: "utf8",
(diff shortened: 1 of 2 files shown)


aeppling:
Thanks for contributing @daixudk96-max — the Node.js hook-rewrite approach was a creative cross-platform solution.

This is now superseded by the v0.37.x releases: master uses the RTK binary itself as the hook command (`rtk hook claude`), requiring zero external dependencies. Closing as resolved.

PR #459, opened 2026-03-10
fix(ci): switch aarch64 Linux build to musl for static linking

### LINKED ISSUE #265: rtk install fails on Ubuntu 22.04 LTS due to requiring GLIBC 2.39
Installing with `curl -fsSL https://raw.githubusercontent.com/rtk-ai/rtk/refs/heads/master/install.sh | sh` on Ubuntu 22.04 LTS doesn't work.

```bash
$ rtk
rtk: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by rtk)
```

  Environment:
  - OS: Ubuntu 22.04.5 LTS
  - Kernel: 6.8.0-100-generic
  - Arch: x86-64
  - System glibc: 2.35

The binary was built against glibc 2.39, which is only available on newer distros (Ubuntu 24.10+, Fedora 41+, etc.). Ubuntu 22.04 is still a supported LTS release (EOL April 2027) and ships glibc 2.35.

Consider either building against an older glibc for broader compatibility, or providing a statically linked binary.

davidjosefson-neo4j:
Fixes install failures on systems with glibc < 2.39 (e.g. Ubuntu 22.04, devcontainers) by building a statically-linked musl binary via the cross tool instead of dynamically linking against glibc.

Mirrors the x86_64 fix from #267.

Fixes #265

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,10 +39,11 @@ jobs:
             os: ubuntu-latest
             archive: tar.gz
             musl: true
-          - target: aarch64-unknown-linux-gnu
+          # aarch64 musl: uses cross tool + Docker image (no host musl-tools needed)
+          - target: aarch64-unknown-linux-musl
             os: ubuntu-latest
             archive: tar.gz
-            cross: true
+            use_cross: true
           # Windows
           - target: x86_64-pc-windows-msvc
             os: windows-latest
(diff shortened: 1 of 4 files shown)


Znuff:
The install script still pulls the `glibc` version on `aarch64` on `Ubuntu 22.04`.

ofek:
I opened https://github.com/rtk-ai/rtk/pull/2089 to fix Linux aarch64.

PR #517, opened 2026-03-11
fix: preserve -- for cargo clippy rewrites

### LINKED ISSUE #496: cargo clippy -- -D warnings fails: args after -- treated as filenames
## Description

When rtk rewrites `cargo clippy ... -- -D warnings`, the flags after `--` get mangled and passed to rustc as positional arguments (filenames) instead of lint flags.

## Command

```
rtk cargo clippy -p my-service -p my-crate -- -D warnings
```

## Error

From full log at `~/Library/Application Support/rtk/tee/`:

```
error: multiple input filenames provided (first two filenames are `crates/my-crate/build.rs` and `-D`)
error: could not compile `my-crate` (build script) due to 1 previous error

error: multiple input filenames provided (first two filenames are `crates/my-macro/src/lib.rs` and `-D`)
error: could not compile `my-macro` (lib) due to 1 previous error
```

## Expected Behavior

`-D warnings` should be forwarded to rustc as a lint flag, identical to running `cargo clippy` directly.

## Workaround

```bash
rtk proxy cargo clippy -p my-service -p my-crate -- -D warnings
# Succeeds cleanly
```

## Root Cause Hypothesis

rtk's `cargo clippy` rewrite does not preserve the `--` separator, causing subsequent args (`-D`, `warnings`) to be passed as positional arguments to rustc rather than as compiler flags. Reproduces consistently on v0.28.2.

## Environment

- rtk v0.28.2
- macOS arm64
- Claude Code hook (auto-rewrite `cargo` → `rtk cargo`)

## Related

#361 Bug 4 covers wrong command routing; this is a separate arg-passing bug specific to `cargo clippy -- <flags>`.

zerone0x:
## Summary
Fixes #496

- detect the original position of `--` relative to the cargo subcommand
- rebuild the argument list with the separator restored before subcommand-specific flags
- add regression coverage for multi-package clippy invocations

## Testing
- not run (rust toolchain unavailable in environment)

### DIFF
--- a/src/cargo_cmd.rs
+++ b/src/cargo_cmd.rs
@@ -30,29 +30,50 @@ pub fn run(cmd: CargoCommand, args: &[String], verbose: u8) -> Result<()> {
 /// Reconstruct args with `--` separator preserved from the original command line.
 /// Clap strips `--` from parsed args, but cargo subcommands need it to separate
 /// their own flags from test runner flags (e.g. `cargo test -- --nocapture`).
-fn restore_double_dash(args: &[String]) -> Vec<String> {
+fn restore_double_dash(subcommand: &str, args: &[String]) -> Vec<String> {
     let raw_args: Vec<String> = std::env::args().collect();
-    restore_double_dash_with_raw(args, &raw_args)
+    restore_double_dash_with_raw(subcommand, args, &raw_args)
 }
 
 /// Testable version that takes raw_args explicitly.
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Thanks for the PR @zerone0x! This issue (#496) has already been fixed on `develop` via the `restore_double_dash()` function in `cargo_cmd.rs`. It detects the original `--` position from `std::env::args()` and restores it before passing to cargo.

You can verify with `develop`:
```
rtk cargo clippy -- -D warnings
```

Closing as duplicate — but thanks for investigating!

PR #520, opened 2026-03-11
fix: handle pnpm --filter/-F flag (#259)

### LINKED ISSUE #259: `pnpm --filter` command results in rtk usage output

ousamabenyounes:
## Summary

Fixes #259 — `rtk pnpm --filter mymodule test` fails with "unexpected argument '--filter' found".

**Root cause**: pnpm's `--filter` is a global flag that precedes the subcommand (`pnpm --filter <pkg> <cmd>`). Clap's `PnpmCommands` subcommand enum expects a subcommand name first, so `--filter` is rejected as an unknown flag.

**Fix**: Intercept `--filter` / `--filter=` / `-F` in raw args **before** `Cli::try_parse()` and route directly to `pnpm_cmd::run_passthrough()`.

## Before/After

```bash
# BEFORE
$ rtk pnpm --filter mymodule test

# AFTER — all variants work
$ rtk pnpm --filter mymodule test     # ok
$ rtk pnpm --filter=mymodule test     # ok
$ rtk pnpm -F mymodule test           # ok
```

## Impact

In pnpm monorepos (very common in modern JS/TS), almost every command uses `--filter`. Without this fix, **every pnpm command in a monorepo** bypasses RTK filtering = 0% savings instead of 70-90%.

## Test plan

- [x] New test: `test_pnpm_filter_detected` — verifies detection logic for all 3 variants
- [x] Full suite: 765 passed, 0 failed
- [x] Manual verification: `--filter`, `--filter=`, `-F` all pass through to pnpm
- [x] `cargo fmt && cargo clippy && cargo test` all green
- [x] `validate-docs.sh` passes

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1235,6 +1235,22 @@ fn main() -> Result<()> {
     // Fire-and-forget telemetry ping (1/day, non-blocking)
     core::telemetry::maybe_ping();
 
+    // Intercept `pnpm --filter`/`-F` before Clap parsing.
+    // pnpm's --filter is a global flag that precedes the subcommand, which Clap
+    // cannot handle in a subcommand enum. Route directly to pnpm passthrough.
+    {
+        let raw_args: Vec<String> = std::env::args().skip(1).collect();
+        if raw_args.first().map(|s| s.as_str()) == Some("pnpm")
+            && raw_args
+                .iter()
+                .any(|a| a == "--filter" || a == "-F" || a.starts_with("--filter="))
+        {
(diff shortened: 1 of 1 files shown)


ousamabenyounes:
Closing as superseded by #270 (merged in commit a39ac43), which natively handles `--filter`/`-F` on the pnpm Clap subcommand with a cleaner architecture (no pre-parse hack) and explicitly fixes #259. Thanks!

PR #559, opened 2026-03-13
fix: forward exit codes in err, test, and summary commands

### LINKED ISSUE #557: bug: rtk does not preserve the underlying command's exit code
## Description

When a command executed through `rtk` returns a non-zero exit code, `rtk` always returns `0` instead of forwarding the original exit code. This breaks any script or tool that relies on exit codes to detect failures.

## Reproduction

```zsh
# Direct command — exit code is preserved
zsh -c 'exit 1'; echo "exit: $?"
# → exit: 1

zsh -c 'exit 2'; echo "exit: $?"
# → exit: 2

# Through rtk — exit code is always 0
rtk err zsh -c 'exit 1'; echo "exit: $?"
# → ✅ Command completed successfully (no errors)
# → exit: 0

rtk err zsh -c 'exit 2'; echo "exit: $?"
# → ✅ Command completed successfully (no errors)
# → exit: 0

rtk summary zsh -c 'exit 2'; echo "exit: $?"
# → exit: 0

rtk test zsh -c 'exit 2'; echo "exit: $?"
# → exit: 0
```

## Expected behavior

`rtk` should forward the exit code of the underlying command. In the examples above, all `rtk` invocations should return the same exit code as the wrapped command (`1` or `2`).

## Why this matters

- **CI/CD pipelines** rely on exit codes to detect failures
- **Claude Code hooks** using `rtk` as a proxy won't detect command failures
- **Shell scripts** using `&&` or `set -e` will silently continue past errors wrapped in `rtk`

## Environment

- rtk version: 0.29.0
- OS: macOS (Darwin 25.2.0)
- Shell: zsh

FlorianBruniaux:
## Problem

`rtk err`, `rtk test`, and `rtk summary` always returned exit code 0, even when the wrapped command failed. The exit code was computed but never forwarded.

Reproduction from #557:
```
rtk err zsh -c 'exit 1'  # → exit: 0 (wrong)
rtk test false            # → exit: 0 (wrong)
rtk summary false         # → exit: 0 (wrong)
```

## Root cause

All three commands called `std::process::exit` nowhere after computing `exit_code`. They returned `Ok(())` unconditionally.

## Fix

Added `if exit_code != 0 { std::process::exit(exit_code); }` before returning in `runner::run_err`, `runner::run_test`, and `summary::run`.

## Verification

```
rtk err false       # → exit: 1 ✓
rtk err true        # → exit: 0 ✓
rtk summary false   # → exit: 1 ✓
rtk test false      # → exit: 1 ✓
```

Integration tests added (marked `#[ignore]`, require `rtk` on PATH):
- `test_err_forwards_exit_code_nonzero`
- `test_err_forwards_exit_code_zero`
- `test_test_forwards_exit_code_nonzero`

Closes #557

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -293,7 +293,7 @@ SHARED            utils.rs          Helpers                N/A        ✓
                   tee.rs            Full output recovery   N/A        ✓
 ```
 
-**Total: 60 modules** (38 command modules + 22 infrastructure modules)
+**Total: 64 modules** (42 command modules + 22 infrastructure modules)
 
 ### Module Count Breakdown
 
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -59,6 +59,9 @@ pub fn run_err(command: &str, verbose: u8) -> Result<()> {
         println!("{}", rtk);
     }
     timer.track(command, "rtk run-err", &raw, &rtk);
+    if exit_code != 0 {
+        std::process::exit(exit_code);
+    }
     Ok(())
 }
 
@@ -100,6 +103,9 @@ pub fn run_test(command: &str, verbose: u8) -> Result<()> {
         println!("{}", summary);
     }
     timer.track(command, "rtk run-test", &raw, &summary);
+    if exit_code != 0 {
+        std::process::exit(exit_code);
+    }
     Ok(())
 }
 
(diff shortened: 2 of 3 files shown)


aeppling:
fixed by the standardization refacto (commit 32ff466 )

PR #563, opened 2026-03-13
fix: skip rewriting find/fd in pipes (#439)

### LINKED ISSUE #439: Trouble running xargs as second bash command.
After configuring rtk tried doing some changes and claude wanted to run a command in the style `find path | xargs grep `, and the grep was failing trying to look up presumably compressed folder names/paths. Unfortunately I didn't grab a screenshot and had to reset the machine, so all I can give you is a poor witness report. `rtk init --show` is all green, so presumably claude knew about prefixing rtk in commands on top of having it as a hook.

pszymkowiak:
## Summary
- Skip rewriting `find`/`fd` when followed by a pipe (`|`)
- `rtk find` outputs a grouped format (files space-separated) incompatible with pipe consumers like `xargs`, `grep`, `wc`, `sort` which expect one path per line
- `find` without a pipe is still rewritten to `rtk find` (savings preserved)

## Before
```
find . -name "*.rs" | xargs grep "fn run"
→ rtk find . -name "*.rs" | xargs grep "fn run"
→ xargs receives "73F 4D: ./ main.rs lib.rs..." → FAILS
```

## After
```
find . -name "*.rs" | xargs grep "fn run"
→ find . -name "*.rs" | xargs grep "fn run"  (not rewritten)
→ works correctly
```

## Tests
- `test_rewrite_find_pipe_skipped` — find with xargs not rewritten
- `test_rewrite_find_pipe_xargs_wc` — find with wc not rewritten
- `test_rewrite_find_no_pipe_still_rewritten` — find alone still rewritten

Fixes #439

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -355,8 +355,18 @@ fn rewrite_compound(cmd: &str, excluded: &[String]) -> Option<String> {
                 } else {
                     // `|` pipe — rewrite first segment only, pass through the rest unchanged
                     let seg = cmd[seg_start..i].trim();
-                    let rewritten =
-                        rewrite_segment(seg, excluded).unwrap_or_else(|| seg.to_string());
+                    // Skip rewriting `find`/`fd` in pipes — rtk find outputs a grouped
+                    // format that is incompatible with pipe consumers like xargs, grep,
+                    // wc, sort, etc. which expect one path per line (#439).
+                    let is_pipe_incompatible = seg.starts_with("find ")
+                        || seg == "find"
+                        || seg.starts_with("fd ")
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Tested manually:

- `rtk rewrite "find src -name *.rs | xargs grep fn"` → exit 1 (not rewritten) ✅
- `rtk rewrite "find src -name *.rs"` → `rtk find src -name *.rs` (rewritten) ✅
- `/usr/bin/find src -name "*.rs" | /usr/bin/xargs grep -l "fn run"` → works correctly with native find ✅
- 903 tests pass, 3 new tests added

PR #571, opened 2026-03-13
fix: handle benchmark-only go test output

DevNewbie1826:
## Problem
`rtk go test` treated benchmark-only runs as if nothing had executed.

For a command like `go test -json -run=^$ -bench=.`, Go emits benchmark result lines but does not emit a test-level `pass` event for the benchmark itself. The old parser only counted test-scoped pass/fail/skip events, so benchmark-only output could collapse to:

```text
Go test: No tests found
```

## What changed
- detect benchmark result lines from `go test -json` output and store them separately from test pass counts
- handle both observed output shapes:
  - benchmark result attached to `Test`
  - benchmark result emitted as package-level `output`
- render a dedicated benchmark summary instead of falling back to `Go test: No tests found`
- keep benchmark results out of the `N passed` test counter so the summary stays semantically accurate

## Before / After
Given a benchmark-only run, the old behavior could report:

```text
Go test: No tests found
```

With this change, `rtk` now produces a benchmark-aware summary like:

```text
✓ Go test: benchmark results in 1 packages

📦 benchfixture [benchmarks]
  BenchmarkFib10-10           9558400       128.3 ns/op
```

## Why this approach
Go's `test2json` benchmark event model is incomplete and does not consistently provide benchmark lifecycle events the same way it does for tests. This change takes the smallest safe approach:
- preserve the existing test/failure accounting
- use benchmark output lines as the source of truth for benchmark-only runs
- avoid overcounting benchmarks as test passes

## Tests
- add regression coverage for benchmark-only output with test-scoped benchmark lines
- add regression coverage for benchmark-only output with package-level benchmark lines

## Verification
Run from the repository root.

- [x] `docker run --rm -v "$PWD:/work" -w /work rust:1.85 cargo test benchmark_only -- --nocapture`
- [x] `docker run --rm -v "$PWD:/work" -w /work rust:1.85 cargo test go_cmd::tests -- --nocapture`
- [x] `docker run --rm -v "$PWD:/work" -w /work rust:1.85 cargo build`

### DIFF
--- a/src/cmds/go/go_cmd.rs
+++ b/src/cmds/go/go_cmd.rs
@@ -37,6 +37,7 @@ struct PackageResult {
     skip: usize,
     build_failed: bool,
     build_errors: Vec<String>,
+    benchmark_outputs: Vec<String>,
     failed_tests: Vec<(String, Vec<String>)>, // (test_name, output_lines)
 }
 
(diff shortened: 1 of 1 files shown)


DevNewbie1826:
Rebased onto latest develop and force-updated this branch.

What I did:
- replayed the PR change on top of current develop
- resolved merge conflicts in src/cmds/go/go_cmd.rs
- re-ran verification in Docker (go_cmd tests, benchmark_only tests, and cargo build)

Ready for review again.

DevNewbie1826:
@pszymkowiak @aeppling  quick ping for review when you have a moment. This PR has been rebased onto latest develop and verification rerun.

PR #588, opened 2026-03-14
fix(pytest): -q mode summary line not detected

### LINKED ISSUE #565: bug: `rtk pytest -q` incorrectly reports "Pytest: No tests collected" when tests are actually executed
## Summary

`rtk pytest -q` reports:

```text
Pytest: No tests collected
[full output: ~/.local/share/rtk/tee/<timestamp>_pytest.log]
```

but the underlying pytest run actually executes the full test suite.

In my case, the full output shows:

```text
5 failed, 1698 passed, 2 skipped
```

and both plain `pytest -q` and `rtk pytest -q` exit with code `1`.

Since pytest exit code `5` is the "no tests collected" case, the RTK summary appears to be incorrect.

## Version

```text
rtk 0.29.0
```

## Reproduction

Environment:
- Linux Ubuntu 24.04
- Python/pytest in an active Conda environment (Python 3.12.11)
- command run directly, without Codex/VS Code wrappers involved

Commands:

```bash
pytest -q; echo $?
/home/niccolo/.local/bin/rtk pytest -q; echo $?
rtk --version
```

Observed output shape:

```text
# plain pytest
... lots of test progress ...
5 failed, 1698 passed, 2 skipped in 108.89s
1

# rtk pytest
Pytest: No tests collected
[full output: ~/.local/share/rtk/tee/1773398989_pytest.log]
1

rtk 0.29.0
```

## Expected behavior

`rtk pytest -q` should emit a summary consistent with the actual pytest run.

For example, something like:

```text
FAILED: 5 tests
PASSED: 1698
SKIPPED: 2
[full output: ...]
```

or any equivalent summary, but not `Pytest: No tests collected`.

## Actual behavior

RTK reports `Pytest: No tests collected` even though:
- tests are clearly executed,
- failures are present in the tee log,
- the exit code is `1`, not `5`.

## Notes

- This reproduces when calling `rtk` directly, so it does not appear to be caused by Codex CLI, VS Code extension, shell aliases, or PATH shims.
- The issue seems to be in the pytest parsing/summarization layer rather than in test discovery itself.
- I am intentionally ignoring the actual failing tests here; the bug is about RTK's summary being inconsistent with the real pytest outcome.

GokhanKabar:
Fixes #565

In quiet mode, pytest emits the final summary without `===` wrappers. The parser only matched `===`-prefixed lines, so the summary was never captured and RTK fell back to the wrong message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
-* **pytest:** fix `rtk pytest -q` incorrectly reporting "No tests collected" when tests ran ([#565](https://github.com/rtk-ai/rtk/issues/565)) — quiet mode summary line (no `===` wrapper) was not captured by the parser, causing `parse_summary_line("")` to return `(0, 0, 0)` and trigger the wrong message. Also fix false "No tests collected" when only skipped tests exist.
 * **diff:** correct truncation overflow count in condense_unified_diff ([#833](https://github.com/rtk-ai/rtk/pull/833)) ([5399f83](https://github.com/rtk-ai/rtk/commit/5399f83))
 * **git:** replace vague truncation markers with exact counts in log and grep output ([#833](https://github.com/rtk-ai/rtk/pull/833)) ([185fb97](https://github.com/rtk-ai/rtk/commit/185fb97))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


mfronczyk:
I see same issue with printing "No tests collected" even when invoking pytest without the "-q" quite option. Will this PR fix this also?

pszymkowiak:
Tested locally — fix is correct.

**Before (current develop):**
- `rtk pytest -q` with 3 pass + 2 fail + 1 skip → "No tests collected"
- `rtk pytest -q -k skipped` (only skipped) → "No tests collected"

**After (this PR):**
- `rtk pytest -q` → "3 passed, 2 failed, 1 skipped" ✓
- `rtk pytest -q -k skipped` → "0 passed, 0 failed, 1 skipped" ✓
- `rtk pytest` (normal mode) → still works correctly ✓

Needs a rebase — only conflict is CHANGELOG (trivial).

aeppling (approved):

PR #627, opened 2026-03-16
feat(hook-script): Improve Claude code hook script

TheoBrigitte:
This PR makes some minor improvements to the Bash script used as Claude code hook for tools rewriting.

It does:
- Replace the version check for `rtk rewrite` support with a direct check if the command exists. This reduces the logic to a single command a condition check for faster execution
- Add double quotes protection to avoid any unexpected Bash expansions
- Use double square brackets `[[` native Bash conditions rather than `[` for faster execution


TheoBrigitte:
Since no one is interested, I am not rebasing this.

PR #628, opened 2026-03-16
fix(contributing): fix contributing guidelines

TheoBrigitte:
This PR fixes some inconsistencies in the CONTRIBUTING.md guidelines.

This addresses 2 points:

- It fixes invalid suggested git branch names, which should rather be something like `feat(scope)/description', otherwise the following error is thrown

```
$ git checkout -b 'feat(contributing): fix-contributing-guidelines'
fatal: 'feat(contributing): fix-contributing-guidelines' is not a valid branch name
hint: See `man git check-ref-format`
hint: Disable this message with "git config set advice.refSyntax false"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to rtk (Rust Token Killer) will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


KuSh:
Thanks for catching the invalid branch names. CONTRIBUTING.md was corrected on develop by commits c5e1d7f5 and c253c016 (2026-03-12): the guide now uses `git checkout -b feat/scope-your-clear-description` and the table lists conventional commit types rather than branch prefixes with parentheses, so the examples git rejected are gone. The hook file you also touched moved to hooks/claude/rtk-rewrite.sh in 46fa31c4 and its version guard was reworked in b3a7e753, which covers the same ground your rewrite did. If you spot anything else in the guidelines that still does not work as written, please comment here or reopen.

PR #644, opened 2026-03-16
feat(aws): add serverless filters for DynamoDB, CloudWatch, S3 transfer, Secrets Manager

JamieCressey:
## Summary

Adds specialized RTK filters for high-frequency AWS serverless commands, plus scoped pipe-safety to prevent breaking downstream consumers.

### New filters

- **DynamoDB scan/query/get-item**: Recursive type flattening strips `{S/N/BOOL/NULL/L/M/SS/NS/BS}` wrappers. Preserves `LastEvaluatedKey` (pagination token) and `ConsumedCapacity` (RCU cost) so LLMs know when results are truncated (~40%+ savings)
- **CloudWatch Logs filter-log-events**: Timestamps include date (`MM-DD HH:MM:SS`), `logStreamName` shown per event, consecutive duplicates collapsed with `[xN]` counts, metadata stripped (~40%+ savings)
- **CloudWatch Logs get-query-results**: Compact `field=value` format per row, internal `@ptr` field filtered out
- **S3 sync/cp**: Summarize upload/download/delete/copy counts from text output, preserve error and warning lines verbatim. Pass through short output (<10 lines) unchanged (~60%+ savings)
- **Secrets Manager get-secret-value**: Extract `Name` + `SecretString` only, compact-print JSON secrets, strip `ARN`, `VersionId`, `VersionStages`, `CreatedDate` (~60%+ savings)

### Removed

- **Lambda invoke**: Removed — no meaningful token savings possible without stripping essential debugging info (`LogResult` contains base64-encoded execution logs, memory/duration stats). Falls through to the generic JSON schema compressor instead.

### Pipe safety

- Scoped pipe rewriting via `PIPE_UNSAFE_PREFIXES`: commands like `aws dynamodb scan | python` are **not** rewritten (RTK's compressed output would break `json.load()`), while text-based commands like `git log | grep` are still rewritten for savings
- This replaces the previous global "skip all pipe rewrites" approach with targeted per-command scoping

All new filters are added as match arms in `aws_cmd::run()`, following the existing pattern for STS/S3/EC2/ECS/RDS/CloudFormation. Unmatched subcommands continue to fall through to the generic JSON schema compressor.

## Test plan

- [x] 46 unit tests for AWS filters (type flattening, filter output, edge cases, empty results, token savings)
- [x] 158 registry tests including pipe-safety assertions for AWS
- [x] Token savings assertions verified per filter
- [x] `cargo fmt --all --check` passes
- [x] `cargo clippy --all-targets` — 0 errors
- [x] `cargo test --all` — 963 tests pass, 0 failures
- [x] Manual test: `rtk aws dynamodb scan --table-name <table>`
- [x] Manual test: `rtk aws logs filter-log-events --log-group-name <group>`
- [x] Manual test: `rtk aws s3 sync <src> <dst>`
- [x] Manual test: `aws dynamodb scan ... | python` passes raw JSON through (pipe-safe)

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -353,15 +353,22 @@ fn rewrite_compound(cmd: &str, excluded: &[String]) -> Option<String> {
                     }
                     seg_start = i;
                 } else {
-                    // `|` pipe — rewrite first segment only, pass through the rest unchanged
+                    // `|` pipe — rewrite first segment only if the command's
+                    // output format is pipe-safe (text-based). Commands that
+                    // transform structured output (JSON → compressed text)
+                    // must NOT be rewritten, as downstream programs (jq,
+                    // python json.load, etc.) expect the original format.
                     let seg = cmd[seg_start..i].trim();
-                    let rewritten =
-                        rewrite_segment(seg, excluded).unwrap_or_else(|| seg.to_string());
-                    if rewritten != seg {
-                        any_changed = true;
+                    if is_pipe_safe(seg) {
+                        let rewritten = rewrite_segment(seg, excluded)
+                            .unwrap_or_else(|| seg.to_string());
+                        if rewritten != seg {
+                            any_changed = true;
+                        }
+                        result.push_str(&rewritten);
+                    } else {
+                        result.push_str(seg);
                     }
-                    result.push_str(&rewritten);
-                    // Preserve the space before the pipe that was lost by trim()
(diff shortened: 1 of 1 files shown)


aeppling:
I'm discussing with maintainers about the filters i mentioned in the review, if we should keep those or not .

We are thinking about different level of filtering options, for now i need to review with maintainers.

PR #681, opened 2026-03-18
fix(pnpm): allow flag-first args to pass through

### LINKED ISSUE #259: `pnpm --filter` command results in rtk usage output

TechWizard9999:
## Description
This fixes `rtk pnpm` argument parsing for flag-first invocations so RTK no longer fails before passthrough can run.
Commands like `rtk pnpm --filter mymodule test` are now routed to native `pnpm` safely.

## Related Issue
Closes #259

## Changes Made
- Reworked `rtk pnpm` CLI parsing to accept raw pnpm args (instead of strict clap subcommands)
- Added runtime routing for known optimized paths (`list`, `outdated`, `install`, `typecheck`)
- Added explicit passthrough for flag-first pnpm forms and unsupported subcommands
- Added regression tests for pnpm flag-first parsing and list-depth extraction
- Updated changelog unreleased bug fixes

## Testing
- `cargo fmt --all --check`
- `rtk cargo clippy --all-targets`
- `rtk cargo test --all`
- Manual reproduction:
  - `rtk cargo run -- pnpm --filter mymodule test --help`
  - Result now reaches pnpm (no RTK clap usage error)

@aeppling please review when you have a moment.


TechWizard9999:
Rebased on latest develop per the refactor notice and force-updated this branch.

@pszymkowiak could you please re-review when convenient?

FlorianBruniaux:
This is architecturally the right approach — removing the `PnpmCommands` enum and handling flag-first args generically is cleaner than pre-Clap interception.

CI shows unstable status. Could you rebase on `develop` to retrigger checks? If CI passes, we will merge this over #520.

TechWizard9999:
Updated this branch to the latest develop to clear conflict state and refresh CI status.\n\n@pszymkowiak @FlorianBruniaux please recheck and let me know if you want me to restore any specific delta from the prior commit set.

TechWizard9999:
Quick maintainer ping for visibility after latest updates. @pszymkowiak @FlorianBruniaux please recheck when convenient.

PR #750, opened 2026-03-20
feat: add Pi agent support for hook installation

zogzog26:
## Summary

This PR adds support for the Pi agent (oh-my-pi / omp) in RTK, following the same pattern as the existing Cursor agent support.

## Changes

- Add  to the  enum in 
- Create  - a shell script hook that intercepts bash commands and rewrites them to use  for token savings (similar to Cursor's approach)
- Add installation logic in  to install the hook to 
- Add  flag support to the CLI
- Update uninstall to support removing Pi hooks

## Usage

The hook will be installed to  where Pi should automatically detect it.

## Testing

- All 979 existing tests pass
- Code compiles successfully
-  has been run

### DIFF
--- a/hooks/pi-rtk-rewrite.sh
+++ b/hooks/pi-rtk-rewrite.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+# rtk-hook-version: 1
+# RTK Pi Agent hook — rewrites shell commands to use rtk for token savings.
+# Works with Pi (oh-my-pi / omp) CLI.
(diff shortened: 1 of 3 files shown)


pszymkowiak:
Thanks @zogzog26 for the early Pi support! Pi integration has since been picked up in #1741, which takes the modern approach (a TypeScript extension delegating to `rtk rewrite`, like the OpenCode/Hermes hooks) with install/uninstall, global+local scopes and tests.

Closing this as superseded by #1741 — there's been no activity here for ~8 weeks. Thanks again, and feel free to chime in on #1741.

PR #752, opened 2026-03-20
feat(sbt): add SBT (Scala Build Tool) support

randomBrainstormer:
## Summary

- Adds `rtk sbt test` — ScalaTest output filtering with 90% token reduction (compact single-line on pass, failure details on fail)
- Adds `rtk sbt compile` — strips SBT noise, keeps errors with 75% token reduction (source count + time on success)
- Adds `rtk sbt run` — light filtering, strips SBT preamble, passes through program output
- Adds `rtk sbt <other>` — passthrough for any unsupported subcommand

## Implementation

Follows the established `go_cmd.rs` pattern:
- `lazy_static!` regex for noise/summary/error line detection
- `crate::tee::tee_and_hint` for full output recovery on failure
- Exit code propagation for CI/CD compatibility
- Discovery rules updated (`src/discover/rules.rs`) with 80% savings estimate

## Test plan

- [x] 11 unit tests with token savings assertions (>=60% on test pass, >=40% on test fail, >=30% on compile error)
- [x] 3 real-output fixtures: `sbt_test_pass.txt`, `sbt_test_fail.txt`, `sbt_compile_error.txt`
- [x] `cargo test --all` — 990 passed, 0 failed
- [x] Manual: `rtk sbt test` in a real Scala project — pending, will validate on a Scala workspace

## Documentation

- `README.md` — added `rtk sbt` to build commands section and command rewrite table
- `CHANGELOG.md` — added entry under `[Unreleased]`

Signed-off-by: Ivan Severino <4858703+randomBrainstormer@users.noreply.github.com>


khoanguyenxuan (approved):

Angel-O:
hey @aeppling any update on this ? do you think it can be merged ?

dannylinden:
Please merge before my wallet getting empty due to our scala project :D

pszymkowiak:
Hi @randomBrainstormer! Re-ran CI on the rebased branch and it hit two regressions from `develop` that landed after your last green build (2026-04-17):

1. **Missing `Debug` derive** on `SbtCommands` — `Commands` enum on develop now requires Debug on all sub-enums:
```
error[E0277]: `SbtCommands` doesn't implement `std::fmt::Debug`
```
2. **Clippy 0.1.95** flags 5 `unnecessary >= y + 1` patterns — clippy expanded that lint since you submitted. Should rewrite as `x > y` instead of `x >= y + 1`.

Could you push a quick fix? After that CI should go green and we can merge.

Full job log: https://github.com/rtk-ai/rtk/actions/runs/24583551517/job/76078806371

randomBrainstormer:
@pszymkowiak Done! Rebased onto develop and pushed a fix for both:

  - Added #[derive(Debug)] to SbtCommands
  - The 5 int_plus_one warnings in stream.rs were already fixed by 1bc4713 on develop, they came in automatically with the rebase

  CI should be green now. Let me know if anything else comes up before merge.

pszymkowiak:
Thanks for the quick rebase fix — CI is green now (verified, 11 checks pass on all 3 OS).

I built the branch and tested `rtk sbt` against real Scala projects in a container. `sbt compile` and `sbt run` work well. One real gap on `sbt test`:

**`filter_sbt_test` only recognizes ScalaTest output.** Tested against `typelevel/cats` (`kernelLawsJVM/test`, 1701 tests) and a fresh munit project — both use **munit** (via `discipline-munit`), whose summary format is:
```
[info] Passed: Total 1701, Failed 0, Errors 0, Passed 1701
```
`TEST_SUMMARY_RE` does not match this, so `has_summary` stays false and the output falls back to **`sbt test: parse error`** (`sbt_cmd.rs:409`) — even though it then recovers the summary from the `[error]` lines. munit is the default test framework in many modern Scala 3 templates, so this is not an edge case.

Two suggestions:
1. Add a munit branch to `filter_sbt_test` (`Total N, Failed N, Errors N, Passed N`), and ideally specs2/ZIO Test later.
2. When the fallback *does* recover a summary, it should not print `parse error` — that header is misleading.

Minor: `sbt update` falls through to passthrough (`run_other`) — fine for now, but on a real project dependency resolution is very verbose and would benefit from a filter later.

`compile`/`run` are solid. Up to you whether you want to extend `test` framework coverage in this PR or land it as-is and follow up — your call.

randomBrainstormer:
@pszymkowiak good catch on munit , it's the default in Scala 3 templates so definitely not an edge case. Fix was small enough to fold in here:
  
- Added `MUNIT_SUMMARY_RE` for the `Passed: Total N, Failed N, Errors N, Passed N format. Maps succeeded = Passed, failed = Failed + Errors`, sets `has_summary = true`. The ```*** FAILED ***``` detail blocks work the same as ScalaTest so those came for free. Added pass/fail fixtures to cover both frameworks.

- Also renamed the fallback label from `parse error` to `errors`, agreed it was misleading.

`sbt update` and specs2/ZIO Test I'll leave for a follow-up as you suggested.

randomBrainstormer:
@pszymkowiak hey, just checking in — munit fix is in, CI is green. Any blockers on your end before this can be merged?

max-leuthaeuser:
Thanks for this PR! I am a heavy Scala/SBT user and would really like to finally get this feature.

randomBrainstormer:
Hi @pszymkowiak @aeppling , bumping this one. All your review points are addressed (munit fix in, last CI green). Anything still blocking, or is it good to merge?

pszymkowiak (approved):
Re-approving on the current head (`a521234`) after a full local verification — the earlier approval predated later commits, so I wanted to confirm the code as it stands now.

**Verified locally:**
- `cargo build` ✅, `cargo test` (30 sbt tests) ✅
- The 8 `unwrap()` calls are all `lazy_static! { Regex::new(...).unwrap() }` — the sanctioned RTK pattern, not production panics.
- **End-to-end on a real Scala/SBT project** (installed SBT, ran a project with one passing + one failing ScalaTest): `rtk sbt test` correctly summarized `1 passed, 1 failed`, surfaced the failing test (`add is broken on purpose` / `CalcSpec`), propagated the non-zero exit code, and wrote the full output to the tee log. Token savings ~88% (154 → 18 words).

Clean, well-tested, real fixtures, no scope creep. LGTM.

Minor non-blocking follow-up idea: including the assertion detail (e.g. `4 did not equal 5`) alongside the failing test name would help agents debug without opening the tee log — but the log covers it, so not a blocker.

randomBrainstormer:
Thanks for the review and approval! @pszymkowiak 
The CI workflow is sitting in `action_required` state, it needs a maintainer to click approve. Could someone with write access kick that off?

src/discover/rules.rs
```diff
@@ -506,6 +506,16 @@ pub const RULES: &[RtkRule] = &[
         subcmd_savings: &[],
         subcmd_status: &[],
     },
+    // Scala/SBT
+    RtkRule {
+        pattern: r"^sbt\s+(test|compile|run|clean|assembly|package)",
```
KuSh: Nitpick: we generally enforce that the command ends there with `(\s|$)` (or `\b` but that also match `_` which sometimes is problematic) to make sure we don’t rewrite anything beyond those subcommands.

src/main.rs
```diff
@@ -1539,11 +1571,83 @@ where
...
+///
+/// SBT accepts global JVM/session flags before the task name. Claude routinely runs
+/// `sbt -batch Test/compile` or `sbt -batch it/Test/test`. Clap can't parse `-batch`
+/// (single-dash long flag) so these fall through to unfiltered passthrough without this.
+///
+/// Examples:
+///   `sbt -batch compile`          → `sbt compile`
+///   `sbt -batch Test/compile`     → `sbt compile Test/compile`
+///   `sbt -Dsbt.log.format=false Test/test` → `sbt test Test/test`
+fn preprocess_sbt_args(args: Vec<OsString>) -> Vec<OsString> {
```
KuSh: These global args aren't accounted for by the rewrite rule, so it shouldn’t happen

CHANGELOG.md
```diff
@@ -524,6 +524,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug Fixes
 
 * **cargo:** preserve compile diagnostics when `cargo test` fails before any test suites run
+* **sbt:** add `rtk sbt` command for SBT (Scala Build Tool) with ScalaTest filtering (90% token reduction on test output, 75% on compile)
```
KuSh: Misplaced and not needed, the changelog is generated on release

KuSh (changes requested):
Please follow [CONTRIBUTING.md](https://github.com/rtk-ai/rtk/blob/develop/CONTRIBUTING.md), especially the [Transparency](https://github.com/rtk-ai/rtk/blob/develop/CONTRIBUTING.md#transparency) section.

Also, two tests (`test_filter_sbt_compile_empty_input` and `test_filter_sbt_test_empty_input`) appear to conflict with `guard.rs`’s own test suite (`never_worse("", "0 matches")` returns `""`), so empty input must yield empty output

TaKO8Ki:
Following up on pszymkowiak’s real-project testing, I also tested `sbt test`, `sbt compile`, and `sbt run` with Cats and official Scala/Typelevel example projects against the current head a521234.

Each command was run once directly with SBT and once through RTK. The filtered outputs and reduction figures below are produced by the SBT filter implemented in this PR.

Results:

- `typelevel/cats` `kernelLawsJVM/test`: 1,701 tests passed; native 831 words → filtered 4 words
- `typelevel/cats` `kernelJVM/compile`: passed; native 288 words → filtered 4 words
- `scala/scala3-example-project` `run`: passed and preserved the application output; native 179 words → filtered 174 words

I also ran the SBT-related unit tests, and they all passed.

Once the changes requested by @KuSh are addressed, LGTM.

src/cmds/scala/sbt_cmd.rs
```diff
@@ -0,0 +1,935 @@
...
+
+pub fn run_run(args: &[String], verbose: u8) -> Result<()> {
+    let timer = tracking::TimedExecution::start();
+
+    let mut cmd = resolved_command("sbt");
+    cmd.arg("run");
+
+    for arg in args {
+        cmd.arg(arg);
+    }
```
TaKO8Ki: Is it unnecessary to preserve a scoped task similarly to `run_test` and `run_compile`? I tried `rtk sbt exampleJVM/run` and rtk transformed it into `sbt run exampleJVM/run` and failed.

```suggestion
    let (sbt_task, rest) = match args.first() {
          Some(a) if is_scoped_task(a) => (a.as_str(), &args[1..]),
        _ => ("run", args),
    };
    cmd.arg(sbt_task);
    for arg in rest {
        cmd.arg(arg);
    }
```

src/discover/rules.rs
randomBrainstormer: Good catch — anchored it with `(?:\s|$)` (matching the golangci-lint rule just above), so it no longer rewrites `sbt testify` and friends. Fixed in ef38103.

src/main.rs
randomBrainstormer: You're right — the rewrite rule only matches `sbt <subcommand>`, so `sbt <global-flag> <task>` never gets rewritten to an `rtk` invocation and this preprocessing was unreachable. Removed `preprocess_sbt_args` together with `is_sbt_global_flag`, `normalize_sbt_task` and their tests. `sbt -batch …` now simply passes through raw (RTK's normal fallback). ef38103.

CHANGELOG.md
randomBrainstormer: Removed — thanks, I didn't realize the changelog is generated at release. ef38103.

src/cmds/scala/sbt_cmd.rs
randomBrainstormer: Thanks, applied your suggestion — `run_run` now preserves a scoped task via `is_scoped_task`, mirroring `run_test`/`run_compile`. `rtk sbt exampleJVM/run` now runs `sbt exampleJVM/run` instead of `sbt run exampleJVM/run`. ef38103.

randomBrainstormer:
Thanks for the careful review @KuSh, and @TaKO8Ki for the scoped-task catch! All fixed in ef38103.

Everything passes locally: `cargo fmt --all --check`, `cargo clippy --all-targets` (no warnings), `cargo test --all` (2456 passed), and semgrep is clean against the PR base.

Back to you when you get a chance, mind kicking off the CI run (it needs a maintainer to approve it) and taking another look?

aeppling:
Hey @randomBrainstormer 

Following up on the Transparency point KuSh raised, here are some point to address for better standardization, otherwise LGTM :)

- run_test / run_compile / run_run: replace the manual .output() + track + exit blocks with the shared runner, same as go_cmd.rs
```rust
  runner::run_filtered(cmd, "sbt test", &args.join(" "), filter_sbt_test, RunOptions::with_tee("sbt_test"))
```

- run_other it:test branch appends the tee hint without the never_worse cap. When filter_sbt_test doesn't recognize the output it returns it unchanged, so what gets printed is the full raw output plus the hint line, more than the raw command produced. print_with_hint (already used in the other three handlers) caps the total; going through the runner fixes it for free.

- Propagate the exit codes to main (use `runner::run_filtered`  and return its `Result<i32>`)


Refer to CONTRIBUTING.md and CODING_PRACTICES.md for the transparence guidelines

randomBrainstormer:
Thanks @aeppling, all three addressed in 59b604b:

- `run_test` / `run_compile` / `run_run` now go through `runner::run_filtered` (via a small `run_task` helper) instead of the manual `.output()` + track + exit blocks — same shape as `go_cmd.rs`.
- `run_other`'s `it:test` branch also routes through the runner now, so the `never_worse` cap applies: an unrecognized output no longer gets reprinted verbatim plus the hint. The catch-all branch uses `runner::run_passthrough`.
- Handlers return `Result<i32>` and the exit code is propagated to `main` (no more direct `std::process::exit`).

The net result is ~190 fewer lines in the SBT module 🚀 ,  also all green locally: `cargo fmt --all --check`, `cargo clippy --all-targets` (no warnings), `cargo test --all` (2456 passed), semgrep clean against the PR base.

aeppling (approved):

TaKO8Ki (approved):

KuSh (approved):

aeppling:
Thanks for contributing to RTK @randomBrainstormer  !

greenhost87:
Minor problem - single test execution via `sbt testOnly` falls into `rtk sbt <other>`

aeppling:
@greenhost87  https://github.com/rtk-ai/rtk/pull/3140

PR #760, opened 2026-03-21
fix: track ls savings against baseline output

### LINKED ISSUE #561: fix: ls tracking baseline should compare against `ls` not `ls -la`
## Problem

`rtk ls` internally runs `ls -la` to get file metadata for grouping/filtering, then compresses the output. The savings % is tracked against the `ls -la` output, not plain `ls` (which is what the agent would have run without RTK).

This inflates the reported savings. Example from user report:
- `ls` = 332 tokens
- `ls -la` = 4173 tokens  
- `rtk ls` = 993 tokens
- Reported: 76.2% savings (vs `ls -la`)
- Actual: -199% vs plain `ls`

## Fix

Track against the baseline command the agent would have run (`ls`), not the internal enriched command (`ls -la`).

In `src/ls.rs` line 113-118, change the tracking input from the `ls -la` raw output to a plain `ls` capture (or estimate).

## Reference

https://www.reddit.com/r/ClaudeAI/comments/1rjkf9r/comment/oa3trnn/

cnaples79:
## Summary
- measure ls savings against the baseline command output instead of the enriched ls -la
- record the baseline command string (ls plus user flags/targets) for tracking

## Rationale
- rtk ls should compare compression to what the user would have run (plain ls), not the internal expanded ls -la, to avoid inflated savings

## Changes
- run an additional baseline ls using user flags/paths
- feed baseline output and command string into tracking instead of the enriched output

Fixes #561

No tests added; behavior unchanged aside from corrected tracking metrics.

### DIFF
--- a/src/ls.rs
+++ b/src/ls.rs
@@ -49,7 +49,28 @@ pub fn run(args: &[String], verbose: u8) -> Result<()> {
         .map(|s| s.as_str())
         .collect();
 
-    // Build ls -la + any extra flags the user passed (e.g. -R)
+    // Build baseline command (what the user would have run: plain `ls` + their flags)
+    let mut baseline_cmd = resolved_command("ls");
+    for flag in &flags {
+        baseline_cmd.arg(flag);
+    }
+    if paths.is_empty() {
+        baseline_cmd.arg(".");
+    } else {
+        for p in &paths {
+            baseline_cmd.arg(p);
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Closing this PR for inactivity — the CLA was never signed and there's been no activity since March. This isn't a judgment on the change itself.

If you'd still like to land it, just sign the CLA (https://cla-assistant.io/rtk-ai/rtk) and ping a maintainer to reopen — happy to pick it back up. Thanks for the contribution!

PR #762, opened 2026-03-21
fix(npx): use npx (not npm) for unrecognized npx commands

### LINKED ISSUE #713: npx -p @scope/pkg binary fails: RTK treats scoped package name as npm subcommand
## Bug

`rtk npx` does not handle the `-p <scoped-package> <binary>` form of npx. When the package name starts with `@`, RTK appears to parse it as a npm subcommand rather than a package specifier.

## Reproduction

```bash
# Fails
npx -y -p @playwright/cli playwright-cli list
# Error: Unknown command: "@playwright/cli"

# Also fails
npx --package=@playwright/cli playwright-cli list
# Error: Unknown command: "playwright-cli"

# Works (bypasses RTK)
rtk proxy npx --package=@playwright/cli -- playwright-cli list
# Output: ### Browsers\n  (no browsers)
```

## Context

- RTK version: 0.29.0
- npm version: 10.9.3
- Node: v22.19.0
- macOS (Apple Silicon)

The rewrite hook converts `npx -y -p @playwright/cli playwright-cli list` to `rtk npx -y -p @playwright/cli playwright-cli list`, which then fails. Using `rtk proxy npx` works correctly.

This affects any skill or tool that invokes CLI tools from scoped npm packages (e.g. `@playwright/cli`, `@anthropic-ai/sdk`, etc.) via the standard npx package flag syntax.

## Related

Issue #197 covers passthrough of known tools via npx but does not cover the `-p @scope/package binary` invocation pattern.

lineardevon:
## Summary

- The `Commands::Npx` catch-all fallback called `npm_cmd::run()`, which invoked `npm` instead of `npx`
- When npx flags preceded the tool name (e.g. `npx -y -p @playwright/cli playwright-cli list`), routing fell to the catch-all and ran the wrong binary
- Replace with direct npx passthrough matching the existing prisma passthrough pattern

Fixes #713

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` (1056 passed on develop)
- [x] Manual testing: verified `rtk npx tsc --version` and `rtk npx playwright test` still route to specialized handlers
- [x] Verified existing passthrough pattern consistency (prisma at lines 1933-1946)

> **Note:** Flag-aware routing (`npx -y tsc` → `tsc_cmd`) and rewrite-layer regex updates are follow-up opportunities.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1871,8 +1871,22 @@ fn run_cli() -> Result<i32> {
                 "prettier" => prettier_cmd::run(&args[1..], cli.verbose)?,
                 "playwright" => playwright_cmd::run(&args[1..], cli.verbose)?,
                 _ => {
-                    // Generic passthrough with npm boilerplate filter
-                    npm_cmd::run(&args, cli.verbose, cli.skip_env)?
+                    // Passthrough via npx (not npm) — unknown tools need npx directly.
+                    let timer = core::tracking::TimedExecution::start();
+                    let mut cmd = core::utils::resolved_command("npx");
+                    cmd.args(&args);
+                    if cli.skip_env {
+                        cmd.env("SKIP_ENV_VALIDATION", "1");
+                    }
+                    let cmd_label = format!("npx {}", args.join(" "));
+                    let status = cmd.status().with_context(|| {
+                        format!("Failed to run `{cmd_label}`. Is npx on PATH?")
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as https://github.com/rtk-ai/rtk/pull/1458 has taken care of the problem

PR #767, opened 2026-03-21
fix: rewrite swift test commands

### LINKED ISSUE #765: Add swift test to hook rewrite registry
## Summary

`swift build` is in the hook rewrite registry (`src/discover/registry.rs`), but `swift test` is not. This means the PreToolUse hook rewrites `swift build` → `rtk swift build` automatically, but `swift test` passes through unfiltered.

Since `swift-build.toml` already exists as a built-in TOML filter, and `swift test` output is structurally similar (build noise + test results), adding `swift test` to the rewrite registry would allow project-local or user-global TOML filters to handle it.

## Context

- `swift test` produces very verbose output (~76KB for ~580 tests using Swift Testing framework)
- A custom TOML filter achieves 99.4% token savings (76KB → ~200 bytes)
- But without a rewrite rule, the hook doesn't intercept `swift test`, so the filter never runs automatically
- Currently requires manually patching `rtk-rewrite.sh`, which breaks integrity verification

## Proposed Change

Add `swift test` to the rewrite registry in `src/discover/registry.rs`, similar to how `swift build` is already registered.

Optionally, a built-in `swift-test.toml` filter would be valuable — Swift Testing framework output follows a consistent pattern:
- `◇ Test/Suite ... started.` (noise)
- `✔ Test ... passed after N seconds.` (noise)
- `✘ Test ... failed/skipped` (keep)
- `✔/✘ Test run with N tests passed/failed` (summary, keep)

## Environment

- macOS, Swift 6.x, SPM (no Xcode)
- Swift Testing framework (not XCTest)
- rtk v0.31.0

zerone0x:
## Summary
- add `swift test` to the hook rewrite registry alongside `swift build`
- route rewritten Swift test commands through `rtk swift`
- add registry coverage for Swift test classification and rewrite behavior

## Test plan
- [x] `cargo test swift_test`
- [x] `cargo test`

Fixes #765

---
Generated by Orbit 🛸

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1519,6 +1519,27 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_classify_swift_test() {
+        assert!(matches!(
+            classify_command("swift test"),
+            Classification::Supported {
+                rtk_equivalent: "rtk swift",
+                category: "Build",
+                estimated_savings_pct: 90.0,
+                status: RtkStatus::Existing,
+            }
+        ));
+    }
+
+    #[test]
+    fn test_rewrite_swift_test() {
+        assert_eq!(
+            rewrite_command("swift test --parallel", &[]),
+            Some("rtk swift test --parallel".into())
+        );
+    }
+
     // --- #336: docker compose supported subcommands rewritten, unsupported skipped ---
 
     #[test]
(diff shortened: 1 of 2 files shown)


pszymkowiak (approved):
LGTM — rewrite rule for `swift test` is correct. Regex, RULES entry, and tests are properly aligned.

Note: the TOML filter `swift-build.toml` only matches `swift build`, so `swift test` output passes through unfiltered for now. This is fine — the rewrite is the first step, filtering will follow in a separate issue.

Issue #720, opened 2026-03-19
rtk gh issue view --comments silently drops comments, misleading AI agents


cderv:
`rtk gh issue view --comments` and `rtk gh pr view --comments` silently discard all comments. The output shows only the issue/PR body with no comments section, which causes Claude Code to conclude "no comments" — a false claim when comments exist.

This is worse than truncation (#354, #188) because the data is never fetched at all.

## Reproduction

On any issue with comments:

```
# Raw gh — shows 2 comments
gh issue view 187 --repo Dicklesworthstone/beads_rust --comments

# Through rtk — no comments visible
rtk gh issue view 187 --repo Dicklesworthstone/beads_rust --comments
```

Claude Code with the rtk hook sees the second output and reports:
> No comments at all — the maintainer never responded.

The issue actually has 2 comments including a maintainer reply.

## Cause

`view_issue` hardcodes `--json number,title,state,author,body,url` — `comments` is not in the field list:

https://github.com/rtk-ai/rtk/blob/4792008fc15553cbb9aeaa602f773a5f8f7f7afe/src/gh_cmd.rs#L688-L689

The `--comments` flag is appended as an extra arg, but in `--json` mode gh ignores it and only returns the listed fields. The formatting code then only processes those 6 fields, so comments are structurally absent from the output.

`view_pr` has the same issue — its JSON fields don't include `comments` either.

For PR view, `should_passthrough_pr_view` triggers passthrough on `--json`, `--jq`, and `--web`, but not `--comments`:

https://github.com/rtk-ai/rtk/blob/4792008fc15553cbb9aeaa602f773a5f8f7f7afe/src/gh_cmd.rs#L286-L289

`view_issue` has no passthrough check at all.

## Impact

With the rtk hook active, Claude Code has no way to know it's running through rtk. It trusts the output and makes false conclusions about issues having no responses. This is particularly harmful when triaging issues or following up on bug reports — the agent acts on incorrect information.

## Suggestion

The simplest fix would be to treat `--comments` as a passthrough trigger (like `--json`/`--jq`/`--web`), for both `issue view` and `pr view`. When a user explicitly asks for comments, correctness matters more than token savings.

## Related

- #354 — gh PR comments/reviews truncated (different: truncation vs never fetched)
- #188 — GitHub issue content truncation (different: body truncation, not comments)

## Environment

- rtk 0.30.1 (master, commit 48e53cb)
- gh 2.74.1
- Claude Code
- Windows 11

PR #773, opened 2026-03-21
fix(gh): skip compact_diff for --name-only/--stat in pr diff

### LINKED ISSUE #730: rtk gh pr diff --name-only returns empty output — diff filter applied to non-diff data
`rtk gh pr diff --name-only` returns empty output. The `--name-only` flag changes `gh pr diff` output from a unified diff to a plain filename list, but the filter applies `compact_diff()` regardless — which expects `diff --git` headers, `@@` hunks, and `+`/`-` lines. A plain filename list matches none of these patterns, so the filter produces empty output.

## Reproduction

On any PR with changed files:

```
# Raw gh — lists 39 changed files
gh pr diff 1815 --repo quarto-dev/quarto-web --name-only

# Through rtk — empty output
rtk gh pr diff 1815 --repo quarto-dev/quarto-web --name-only
```

Claude Code with the rtk hook sees empty output and concludes "no changed files" or retries with different flags — all hitting the same empty result.

## Cause

`pr_diff` passes `--name-only` through to `gh` (line 1125-1127), so the raw output is a filename list. But line 1144 unconditionally runs `git::compact_diff(&raw, 500)` on whatever comes back:

https://github.com/rtk-ai/rtk/blob/48e53cb4792e4ae8e9fb50f4e908b11a9c55a3b1/src/gh_cmd.rs#L1139-L1147

`compact_diff` only emits lines matching diff structure (`diff --git`, `@@`, `+`, `-`). Plain filenames match nothing, so `result` is empty.

## Suggestion

Flags like `--name-only`, `--stat`, and `--name-status` change the output format away from unified diff. When any of these are present, passthrough unfiltered — the output is already compact (just filenames) and doesn't benefit from diff compaction.

## Broader Pattern

This is the same class of bug as #720 — a filter that doesn't account for a flag silently produces wrong output, and the hook makes it invisible to the AI agent. The agent trusts the output and draws false conclusions.

Both could be mitigated by a general principle: when a filter produces empty output from non-empty input, emit a warning to stderr rather than silently returning nothing.

## Related

- #720 — `--comments` silently dropped on `gh issue/pr view` (same pattern: flag ignored by filter)
- #354 — gh output truncation (different: partial output vs empty output)

## Environment

- rtk 0.30.1 (master, commit 48e53cb)
- gh 2.74.1
- Claude Code
- Windows 11

mvanhorn:
## Summary

- Skip `compact_diff` filtering when `gh pr diff` is called with output-format-changing flags (`--name-only`, `--stat`, `--name-status`, `--numstat`, `--shortstat`)

## Why this matters

`compact_diff()` expects unified diff format with `diff --git` headers and `@@` hunks. When `--name-only` is passed, gh outputs a plain filename list that matches none of these patterns, producing empty output. Claude Code with the rtk hook sees empty output and concludes "no changed files" ([#730](https://github.com/rtk-ai/rtk/issues/730)).

Same class of bug as [#720](https://github.com/rtk-ai/rtk/issues/720) - a filter that doesn't account for a flag silently produces wrong output. @cderv works around both by excluding `gh` from the rtk hook entirely via `config.toml`.

## Changes

Added `has_non_diff_format_flag()` in `src/gh_cmd.rs` that checks for flags producing non-unified-diff output. When any of these flags are present, `pr_diff()` now takes the early passthrough path (same as `--no-compact`), matching the existing pattern at line 1117-1118.

7 unit tests cover all flag variants and the no-flag case.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` (1079 tests pass)
- [x] Unit tests for `has_non_diff_format_flag` with all 5 flag variants
- [x] Verified no-flag case still uses `compact_diff` path

Fixes #730

### DIFF
--- a/src/gh_cmd.rs
+++ b/src/gh_cmd.rs
@@ -1118,6 +1118,18 @@ fn pr_merge(args: &[String], _verbose: u8) -> Result<()> {
     Ok(())
 }
 
+/// Flags that change `gh pr diff` output from unified diff to a different format.
+/// When present, compact_diff would produce empty output since it expects diff headers.
+fn has_non_diff_format_flag(args: &[String]) -> bool {
+    args.iter().any(|a| {
+        a == "--name-only"
+            || a == "--name-status"
+            || a == "--stat"
+            || a == "--numstat"
+            || a == "--shortstat"
+    })
+}
+
 fn pr_diff(args: &[String], _verbose: u8) -> Result<()> {
     // --no-compact: pass full diff through (gh CLI doesn't know this flag, strip it)
     let no_compact = args.iter().any(|a| a == "--no-compact");
@@ -1127,7 +1139,9 @@ fn pr_diff(args: &[String], _verbose: u8) -> Result<()> {
         .cloned()
         .collect();
 
-    if no_compact {
+    // Passthrough when --no-compact or when a format flag changes output away from
+    // unified diff (e.g. --name-only produces a filename list, not diff hunks).
+    if no_compact || has_non_diff_format_flag(&gh_args) {
         return run_passthrough_with_extra("gh", &["pr", "diff"], &gh_args);
     }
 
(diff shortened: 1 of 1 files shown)


mvanhorn:
Rebased onto develop in 2ef0690.

aeppling (approved):

PR #776, opened 2026-03-21
feat(config): add suppress_hook_warning option

### LINKED ISSUE #682: Need config to suppress warnings like "No hook installed — run `rtk init -g` for automatic token savings"
I'm using rtk in Windows with my Claude Code. According to the guidance, I added rtk's instructions to my `CLAUDE.md` instead of configuring Claude Code Hooks.

But every time when Claude Code run some bash command with rtk, it will print a warning message like below:
```bash
● Let me check the current state of these files.
● Bash(rtk git log --oneline -20)
  ⎿  [rtk] /!\ No hook installed — run `rtk init -g` for automatic token savings               !!!!!! Unexpected warnings.
     7cb555668 Fix some package versions
     f214d1b2e Fix csproj
     … +18 lines (ctrl+o to expand)
● Searched for 1 pattern (ctrl+o to expand)
```

mvanhorn:
## Summary

- Add `hooks.suppress_hook_warning` config option in `config.toml`
- Add `RTK_SUPPRESS_HOOK_WARNING=1` env var as zero-config alternative

## Why this matters

Users running rtk via CLAUDE.md instructions (instead of hooks) or with tools like OpenCode get `[rtk] /!\ No hook installed` on every command ([#682](https://github.com/rtk-ai/rtk/issues/682)). The warning wastes tokens and confuses AI agents - @btbxbob reports it printing on every `git add`, `git commit`, and `git status`. 5 users across Windows and OpenCode report the same problem.

Currently the warning is rate-limited to once per day, but there's no way to suppress it permanently.

## Changes

`src/config.rs`:
- Added `suppress_hook_warning: bool` field to `HooksConfig` (defaults to `false`)
- Added `hook_warning_suppressed()` helper that checks the env var first, then config file

`src/hook_check.rs`:
- `check_and_warn()` now returns early when `hook_warning_suppressed()` is true

The `rtk gain` diagnostic output (`src/gain.rs`) still shows hook status since that's an explicit diagnostic command.

### Usage

Config file (`~/.config/rtk/config.toml`):
```toml
[hooks]
suppress_hook_warning = true
```

Or env var (useful for CI, OpenCode, non-Claude tools):
```bash
export RTK_SUPPRESS_HOOK_WARNING=1
```

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` (1074 tests pass)
- [x] `suppress_hook_warning` deserializes correctly from TOML
- [x] Default is `false` (existing behavior unchanged)
- [x] Missing field in `[hooks]` section defaults to `false`

Fixes #682


pszymkowiak (approved):
LGTM — clean and useful. Config field + env var, serde(default) for backward compat, 4 tests. Exactly what OpenCode and non-hook setups need.

One minor suggestion: consider also checking the env var in `runtime_check()` (integrity.rs) for consistency, but not a blocker.

mvanhorn:
Rebased onto develop in 65795c6 - resolved a conflict in src/core/config.rs (kept the helper functions I originally added) and fixed a stale import path in hook_check.rs. The two suppress_hook_warning tests pass locally.

Let me know if anything else is needed before merge.

src/core/config.rs
```diff
@@ -128,6 +133,24 @@ pub fn limits() -> LimitsConfig {
...
+    }
+    Config::load()
+        .map(|c| c.hooks.suppress_hook_warning)
+        .unwrap_or(false)
+}
+
+/// Check if telemetry is enabled in config. Returns None if config can't be loaded.
+pub fn telemetry_enabled() -> Option<bool> {
+    Config::load().ok().map(|c| c.telemetry.enabled)
+}
```
KuSh: **Blocker: this function is dead code and will fail the build.**

`telemetry_enabled()` has zero callers anywhere in the crate — `telemetry.rs:46` and `telemetry_cmd.rs:44` both read `config.telemetry.enabled` directly. `src/main.rs` declares `mod core;` privately and there is no `lib.rs`, so this item is unreachable and `rustc` fires `dead_code`. `Cargo.toml` sets `[lints.rust] warnings = "deny"`, which turns that warning into a hard error. Applying this PR onto `develop` and running `cargo check --all-targets` gives:

```
error: function `telemetry_enabled` is never used
   --> src/core/config.rs:170:8
```

This looks like a conflict-resolution artifact: `telemetry_enabled()` and `telemetry_consent()` were deliberately removed from this file by `8156081 fix(telemetry): clean code`, and the rebase note on this PR mentions keeping the helper functions originally added. It's also unrelated to the `suppress_hook_warning` feature.

Please drop this function.

src/core/config.rs
```diff
@@ -128,6 +133,24 @@ pub fn limits() -> LimitsConfig {
     Config::load().map(|c| c.limits).unwrap_or_default()
 }
 
+/// Check if hook warnings are suppressed via config or env var.
+pub fn hook_warning_suppressed() -> bool {
+    if std::env::var("RTK_SUPPRESS_HOOK_WARNING")
+        .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
+        .unwrap_or(false)
```
KuSh: The env var can force suppression **on** but never **off**, which doesn't match the PR description ("the env var is checked first").

Because the early return only fires on a truthy value and anything else falls through to the config file, a user with `suppress_hook_warning = true` in `config.toml` who runs `RTK_SUPPRESS_HOOK_WARNING=0 rtk git status` still gets the warning suppressed — there's no escape hatch to get the warning back for a single command. Matching on `Some(false)` explicitly would give a real override; otherwise the docs should be reworded.

Minor, separately: the parse diverges from the convention used everywhere else in the repo, which is `std::env::var("RTK_...").as_deref() == Ok("1")` (`trust.rs:104`, `hook_cmd.rs:547`, `toml_filter.rs:402`, `tee.rs:184`).

src/hooks/hook_check.rs
```diff
@@ -104,6 +104,11 @@ fn check_and_warn() -> Option<()> {
         HookStatus::Outdated => "[rtk] /!\\ Hook outdated — run `rtk init -g` to update",
     };
 
+    // Suppress if configured via config.toml or RTK_SUPPRESS_HOOK_WARNING env var
+    if crate::core::config::hook_warning_suppressed() {
+        return Some(());
+    }
```
KuSh: This check saves nothing for the users it's meant to help, because it runs *after* the expensive work.

By the time we reach here, `status()` has already resolved the Claude dir, read and JSON-parsed `settings.json`, and read the hook script — and this then adds a *new* `Config::load()` (another file read + TOML parse) on top. For exactly the population in #682 (no hook installed, `maybe_warn()` called from `main.rs:1638` on every single command), each invocation now pays the full probe *plus* a config read just to print nothing.

Moving `hook_warning_suppressed()` to the top of `check_and_warn()`, before `status()`, makes the suppressed path nearly free and keeps the <10ms startup budget intact.

src/core/config.rs
```diff
@@ -29,6 +29,11 @@ pub struct HooksConfig {
     /// Survives `rtk init -g` re-runs since config.toml is user-owned.
     #[serde(default)]
     pub exclude_commands: Vec<String>,
+    /// Suppress "No hook installed" and "Hook outdated" warnings.
+    /// Useful when running rtk via CLAUDE.md instructions instead of hooks,
+    /// or with tools like OpenCode that don't use Claude Code hooks.
+    #[serde(default)]
+    pub suppress_hook_warning: bool,
```
KuSh, line 58: Worth documenting the new option while you're here — a user hitting #682 won't discover it otherwise.

- `docs/guide/getting-started/configuration.md:51` — the `[hooks]` sample block lists only `exclude_commands`
- `docs/guide/getting-started/configuration.md:57-65` — the "Environment variables" table lists `RTK_DISABLED`, `RTK_TEE_DIR`, `RTK_TELEMETRY_DISABLED`, `RTK_HOOK_AUDIT` but not `RTK_SUPPRESS_HOOK_WARNING`
- `README.md:416` — same config snippet

The triage bot already has "Docs updated" unchecked on this PR.

KuSh (changes requested):
Thanks for the rebase work on this — the feature itself is the right shape, and the `[hooks]` config + env-var pair is the correct place for it. One blocker before this can go in, plus three smaller things worth folding in while the branch is being touched.

## Blocker

`telemetry_enabled()` in `src/core/config.rs` has no callers and will fail the build (see inline comment). It's unrelated to this PR's scope and appears to be a conflict-resolution artifact.

## This PR needs another rebase

`develop` has since added `transparent_prefixes: Vec<String>` to `HooksConfig`, and GitHub now reports this branch as `CONFLICTING` / `DIRTY`.

## Note: `develop` itself is currently broken — not your fault, and not something you need to fix here

While verifying this PR I applied both hunks onto `develop` and found that `develop` does not build on its own, independently of this PR: `src/cmds/php/phpt_cmd.rs:162` uses `lazy_static::lazy_static!` but `lazy_static` is not in `Cargo.toml`, and `src/discover/rules.rs` is missing the `pipeline_final_safe` field. Both regressions came in with the `.phpt` filter merge (#1503, `e8541d1`).

That is already fixed by #3793. So when you rebase and `cargo clippy --all-targets` explodes with `E0433 cannot find crate lazy_static` and `E0063 missing field pipeline_final_safe`, those seven errors are not yours — ignore them, or rebase on top of #3793 to get a clean run. The only error attributable to this PR is the `telemetry_enabled` one flagged inline.

## Note on the existing approval

The 2026-03-26 approval predates two rebases and was given on `02ea499`, before the conflict resolution that introduced `telemetry_enabled()`, so it doesn't cover the current head.

On the suggestion to also check the env var in `integrity.rs::runtime_check()` — I'd leave that one alone. That path only prints on `IntegrityStatus::Tampered`, which is a security signal rather than recurring noise, so it shouldn't be suppressible by the same switch.

mvanhorn:
All four are in dfb8886, rebased onto current develop.

The dead telemetry_enabled is gone. It was a conflict-resolution artifact exactly as you guessed, and dropping it is what makes the branch build under warnings = deny.

The env var is a real override now: Ok("1") suppresses, any other value forces the warning back on, and only an unset var falls through to config.toml. That also moves it onto the as_deref() == Ok("1") shape the rest of the repo uses.

hook_warning_suppressed() runs before status() now, so a suppressed user pays neither the hook probe nor the config read that used to follow it.

Documented in configuration.md, both the [hooks] sample and the environment variable table, plus the README snippet.

On develop being broken: it builds clean at f83c0cf. cargo fmt --check, cargo clippy --all-targets with zero warnings, and the config and hook tests all pass locally against that base, so the lazy_static and pipeline_final_safe errors you hit are gone.

src/core/config.rs
KuSh: Confirmed fixed on `dfb8886f` — `telemetry_enabled()` is gone from `src/core/config.rs`. Verified in a clean worktree at the PR head: `cargo clippy --all-targets` is clean (zero warnings), so the `warnings = "deny"` blocker is resolved. Resolving.

src/core/config.rs
KuSh: Confirmed fixed — now `Ok("1") => true, Ok(_) => false`, so `RTK_SUPPRESS_HOOK_WARNING=0` genuinely overrides a config-level `suppress_hook_warning = true`, and the parse matches the `as_deref()` convention used in `trust.rs` / `hook_cmd.rs` / `toml_filter.rs` / `tee.rs`. Resolving.

One small non-blocking note: the two new unit tests cover TOML deserialization only, so the env-var override arm isn't exercised. Not worth holding the PR for — env-var tests are awkward under parallel test execution.

src/hooks/hook_check.rs
KuSh: Confirmed fixed — the suppression check now runs at the top of `check_and_warn()`, before `status()`, so the suppressed path no longer pays for the Claude-dir resolution, the `settings.json` read/parse, or the hook-script read. Resolving.

src/core/config.rs
KuSh: Confirmed fixed — documented in `docs/guide/getting-started/configuration.md` (both the `[hooks]` sample at line 53 and the environment-variable table at line 66) and in `README.md:417`. Resolving.

src/core/config.rs
```diff
@@ -184,6 +189,17 @@ pub(crate) fn cached_config() -> &'static Config {
...
 }
 
+/// Check if hook warnings are suppressed via config or env var.
+pub fn hook_warning_suppressed() -> bool {
+    match std::env::var("RTK_SUPPRESS_HOOK_WARNING").as_deref() {
+        Ok("1") => true,
+        Ok(_) => false,
+        Err(_) => Config::load()
+            .map(|c| c.hooks.suppress_hook_warning)
+            .unwrap_or(false),
```
KuSh, line 199: `Ok(_) => false` goes further than my previous comment intended, and the blast radius is wider than the override it buys.

Any non-`"1"` value now vetoes the config file entirely:

- `RTK_SUPPRESS_HOOK_WARNING=true` is the obvious thing a user tries — and it was accepted by the *previous* revision of this PR. It now returns `false`, so the warning prints **and** an explicit `suppress_hook_warning = true` in `config.toml` is silently ignored, with nothing to indicate why.
- More painful in CI: a GitHub Actions `env:` block exports an **empty string** when the repository variable isn't configured. So `RTK_SUPPRESS_HOOK_WARNING: ${{ vars.SOMETHING }}` silently unsets the user's config setting. This repo already documents that exact hazard at `src/core/telemetry.rs:22-26` and ships `core::utils::env_is_some` (`src/core/utils.rs:548`) precisely for it.

I don't want to send you back to enable-only, though — that would re-create the "can force on but never off" problem from my last comment. A three-way match satisfies both rounds:

```rust
match std::env::var("RTK_SUPPRESS_HOOK_WARNING").as_deref() {
    Ok("1") | Ok("true") => true,
    Ok("0") | Ok("false") => false,
    _ => Config::load()
        .map(|c| c.hooks.suppress_hook_warning)
        .unwrap_or(false),
}
```

That keeps the explicit off-override, restores `=true`, and lets an empty or unrecognised value fall through to config instead of silently overriding it.

Separately: the two new tests cover TOML deserialization only. The env-var/config precedence is the part most likely to be wrong and currently has no test at all — worth adding one now that the arms are non-trivial.

src/hooks/hook_check.rs
```diff
@@ -93,6 +93,11 @@ pub fn maybe_warn() {
 
 /// Single source of truth: delegates to `status()` then rate-limits the warning.
 fn check_and_warn() -> Option<()> {
+    // Suppress before probing the hook installation to keep this path cheap.
+    if crate::core::config::hook_warning_suppressed() {
+        return Some(());
+    }
```
KuSh, line 99: Moving this to the top was my suggestion, but I under-weighted the cost to the common path — this now makes things slower for everyone to save work for the few who set the flag.

`maybe_warn()` runs on every command except `Gain` (`src/main.rs:1637-1639`). Before this change, the overwhelmingly common case — hook installed, `status()` returns `Ok` — hit `HookStatus::Ok => return Some(())` in the match and **never read `config.toml`**. Now every single `rtk` invocation pays a `Config::load()`: `stat` + file read + full TOML parse. So the comment "to keep this path cheap" is currently backwards for ~100% of users.

Cheapest fix that keeps this ordering: use `cached_config()` instead of `Config::load()`. The repo already establishes that precedent explicitly — see the doc comment on `hook_rewrite_params` (`src/core/config.rs:158-170`), which calls out this exact double-parse concern on the hook hot path. It's safe here: nothing reads the cached config after a same-process config write on this path.

(The alternative — moving the check back below the `status()` match — also works and costs nothing, but then the suppressed path pays for the hook probe again, which is what my last round was about. `cached_config()` gets you both.)

src/core/config.rs
```diff
@@ -51,6 +51,11 @@ pub struct HooksConfig {
     /// not anything else.
     #[serde(default)]
     pub transparent_prefixes: Vec<String>,
+    /// Suppress "No hook installed" and "Hook outdated" warnings.
+    /// Useful when running rtk via CLAUDE.md instructions instead of hooks,
+    /// or with tools like OpenCode that don't use Claude Code hooks.
+    #[serde(default)]
+    pub suppress_hook_warning: bool,
```
KuSh, line 58: One flag suppresses both `HookStatus::Missing` and `HookStatus::Outdated`, which quietly disables the upgrade signal too.

Concrete path: an OpenCode / CLAUDE.md user sets this to kill the "No hook installed" noise (the #682 case). Later they do install the Claude Code hook, then upgrade rtk to a version that bumps `CURRENT_HOOK_VERSION`. They never see "Hook outdated — run `rtk init -g`", so they keep running stale rewrite logic indefinitely with no indication anything is wrong.

#682 only asks for the *missing*-hook case. Either scope the suppression to `HookStatus::Missing`, or extend this doc comment (and the docs you added) to say the flag also mutes upgrade prompts, so the trade-off is at least discoverable.

KuSh (changes requested):
Second pass on `dfb8886f`. All four threads from the previous round are genuinely addressed and I've resolved them — the rebase is clean, `telemetry_enabled()` is gone, docs are in, and I verified `cargo clippy --all-targets` is clean with 3074 tests passing at this head.

Three new points below. Up front: **the first one is partly my fault** — my previous comment asked for an env-var override and the literal implementation of that ask has a sharper edge than I anticipated. The suggested fix below reconciles both rounds, so please don't just revert to what you had.

Also worth noting my earlier pointer to #3793 for the `develop` build break is now stale — that PR was closed, but the `lazy_static` breakage is gone from your base anyway, so it's moot.

mvanhorn:
All three are in 4e53f76, on top of a rebase onto develop. The env var now parses truthy and falsy values and only overrides when it is actually set, so `true` works and an unrecognised value no longer vetoes config.toml. check_and_warn() probes status() first and returns on Ok before touching Config::load(), so the common path pays nothing new. And the flag is scoped to Missing: an outdated hook still prints the upgrade prompt, which is what #682 asked for and no more. Docs and README say so now.

KuSh (approved):
## Round 3 of 3 — APPROVED

**Claim**: let users permanently silence the recurring missing-hook warning via `hooks.suppress_hook_warning` and `RTK_SUPPRESS_HOOK_WARNING`.
**Scope**: accept.
**CI** @`d2e906a0`: all green — clippy, fmt, doc review, test presence, ubuntu/macos/windows, Security Scan, semgrep, benchmark, CLA.

### Blocks merge (0)
Everything from the first two rounds is verified fixed **by execution**, not by reading. I built this head and the previously reviewed one with separate target dirs (hashes confirmed distinct) and ran a 90-cell matrix — 2 hook states x 3 config states x 15 env values — each cell in a fresh `HOME`/`XDG_CONFIG_HOME`/`XDG_DATA_HOME`/`CLAUDE_CONFIG_DIR` under `LC_ALL=C`:

| | before | after |
|---|---|---|
| `cfg=true`, env `""` / `"   "` (the CI empty-string veto) | WARNS | silent |
| `cfg=*`, env `true`/`TRUE`/`" true "`/`yes`/`on` | WARNS | silent |
| `cfg=true`, env `maybe` / `2` | WARNS | silent |
| Outdated hook + suppression | silent | **WARN-outdated** |

23 of 90 cells changed, all in the intended direction, and the falsy force-off (`0`/`false`/`no`/`off` beating `cfg=true`) is intact. The hot-path fix is confirmed by syscall count: on `HookStatus::Ok` the previous head opened `config.toml` twice, this one opens it once. Merge onto latest `develop` is clean and green.

Nicely done on `parse_suppress_hook_warning_env` — extracting the pure parser is what made the precedence testable without touching process env or the config cache.

### Fixup pushed — `d2e906a0`
Two things I folded in rather than sending back for a fourth round:

1. **`rtk gain` ignored the flag.** It prints its own missing-hook line from a bare `status()` match, outside `maybe_warn`, so neither the config option nor the env var reached it — on the report the no-hook audience reads most, which is the audience `#682` is about. Its `Missing` arm is now gated on `hook_warning_suppressed()`; the outdated prompt stays visible, consistent with the scoping in this PR.
2. **The env tests were tautologies.** They exercise the parser alone, so `hook_warning_suppressed()` — which composes env and config — had no coverage. Rewriting it as `config || env.unwrap_or(false)`, dropping the falsy force-off this PR just added, kept the entire suite green while the binary silently regressed. `tests/hook_warning_scope_test.rs` now drives the composition through the real binary, and both gaps are mutation-checked: reverting either fix fails a test.

The suite also now pins `XDG_CONFIG_HOME`/`XDG_DATA_HOME` to the temp home and seeds the config in both the XDG and macOS locations. My first attempt at these tests passed locally and failed on both Unix runners — `dirs::config_dir()` is `~/.config` on Linux but `~/Library/Application Support` on macOS. That was my bug, fixed in the same commit; it incidentally hardens the pre-existing tests in that file, which were reading the developer's real config too.

### Next
Nothing — approving. Merges as-is.

src/core/config.rs
KuSh: Fixed in `4e53f760`. `parse_suppress_hook_warning_env` now accepts a truthy and a falsy set and returns `None` for anything else, so an empty or unrecognised value falls through to config instead of vetoing it. Verified by execution across a 90-cell matrix: `cfg=true` with env `""`, `"   "`, `maybe` or `2` went from WARNS to silent, while `0`/`false`/`no`/`off` still force the warning back on. Resolving.

src/hooks/hook_check.rs
KuSh: Fixed in `4e53f760`, and better than what I suggested: `status()` is probed first with an early return on `Ok`, and the helper now reads `cached_config()`. Measured with `strace` on the `HookStatus::Ok` path — the previous head opened `config.toml` twice, this one opens it once. Resolving.

src/core/config.rs
KuSh: Fixed in `4e53f760`. Suppression is scoped to `HookStatus::Missing` and `warning_text` has a full truth-table test. Verified end to end: with suppression on, an outdated hook went from silent to printing the upgrade prompt. The same scoping now holds for `rtk gain` after the fixup. Resolving.

mvanhorn:
Thanks for the merge, @KuSh. The env var should cover the CLAUDE.md and OpenCode paths without anyone needing to touch `config.toml`.

PR #792, opened 2026-03-24
fix(diff): check modified count in identical-files guard

### LINKED ISSUE #781: `diff` reports "Files are identical" when lines differ by only a few characters
### Description

`rtk diff` incorrectly reports files as identical when modified lines share a long common prefix. The underlying `/usr/bin/diff` correctly detects the differences.

### Version

Built from source at commit 97b6878

### Minimal reproduction

```bash
# FAILS — reports identical (4+ char common prefix)
printf "abcd\n" > /tmp/a.txt
printf "abce\n" > /tmp/b.txt
# → [ok] Files are identical

# WORKS — correctly detects difference (3 char prefix)
printf "abc\n" > /tmp/a.txt
printf "abd\n" > /tmp/b.txt
# → +1 added, -1 removed

# Verification with system diff
/usr/bin/diff /tmp/a.txt /tmp/b.txt
# → 1c1 < abcd --- > abce
```

### Additional examples

```bash
# Also fails with multi-word lines
printf "foo bar\n" > /tmp/a.txt
printf "foo baz\n" > /tmp/b.txt
# → [ok] Files are identical

# Also fails with longer content
printf "abc def\n" > /tmp/a.txt
printf "abc deg\n" > /tmp/b.txt
# → [ok] Files are identical
```

### Expected behavior

`rtk diff` should report the differences, consistent with `/usr/bin/diff`.

### Suspected cause

A similarity threshold in rtk's diff filtering appears to treat lines that are "close enough" as identical, when it should only filter output formatting — not discard actual differences.

gghez:
## Summary

`rtk diff` incorrectly reports "Files are identical" when lines differ by only a few characters. Two root causes:

1. The identical-files guard at line 24 only checked `added == 0 && removed == 0`, ignoring the `modified` counter. Lines classified as modifications (similarity > 0.5) incremented `modified` but not `added`/`removed`, so the guard passed and the diff silently swallowed actual differences.

2. The `similarity()` function used Jaccard similarity on unique character sets, which over-estimated similarity for lines sharing common characters regardless of position. Replaced it with a positional prefix+suffix ratio that better reflects actual line differences.

## Changes

- Added `&& diff.modified == 0` to the identical-files guard in `run()`
- Replaced `similarity()` implementation: Jaccard set-based -> positional prefix+suffix metric
- Added 4 regression tests covering all reproduction cases from the issue
- Updated existing `test_similarity_partial_overlap` for the new metric

## Test plan

- [x] All 983 existing tests pass (`cargo test --all`)
- [x] 4 new regression tests cover exact reproduction cases from #781
- [x] `cargo check` compiles without warnings

Fixes #781

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -22,7 +22,7 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<()> {
     let diff = compute_diff(&lines1, &lines2);
     let mut rtk = String::new();
 
-    if diff.added == 0 && diff.removed == 0 {
+    if diff.added == 0 && diff.removed == 0 && diff.modified == 0 {
         rtk.push_str("[ok] Files are identical");
         println!("{}", rtk);
         timer.track(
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Fixes a bug where `rtk diff` falsely reported files as identical when lines differed by only a few characters. The identical-files guard now checks the `modified` counter in addition to `added` and `removed`, and the similarity function was replaced with a positional prefix+suffix metric to avoid over-estimating similarity.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #781

---

KuSh:
Closing this one as overtaken by events rather than as a duplicate — sorry it sat this long.

The headline fix (adding `modified` to the identical-files guard) has since landed by another route: `render_file_diff` now branches on `diff.changes.is_empty()`, which covers modified-only diffs by construction. Verified against the exact repro from #781/#2811 on 0.46.0:

```
$ printf 'alpha beta\n' > v1; printf 'alpha zzzz\n' > v2
$ rtk diff v1 v2; echo "exit=$?"
exit=1
```

The branch is also conflicting after the `src/` reorganisation you rebased for back in March, so there's nothing here that applies cleanly any more.

The part that did **not** land is your `similarity()` rewrite. The current implementation is still Jaccard over character *sets*, which is order-insensitive — `"abcd"` and `"dcba"` score 1.0 — and your prefix/suffix version is strictly better for deciding whether two lines are a modification or an unrelated pair. That's still worth having, and it's small enough to stand alone. A fresh PR with just that change would be welcome.

PR #797, opened 2026-03-24
fix: strip trailing stderr redirects before rewrite matching (#530)

### LINKED ISSUE #530: feat(rewrite): strip trailing stderr redirects before pattern matching
## Description

`rtk rewrite` doesn't match commands that have trailing `2>&1` or `2>/dev/null`, because the redirect suffix is part of the string passed to pattern matching.

## Reproduction

```bash
# These match:
rtk rewrite "git status"        # -> rtk git status ✓
rtk rewrite "find . -name '*.ts'"  # -> rtk find . -name '*.ts' ✓

# These don't match (exit 1):
rtk rewrite "git status 2>&1"              # -> no match ✗
rtk rewrite "find . -name '*.ts' 2>&1"     # -> no match ✗
rtk rewrite "ls -la 2>/dev/null"           # -> no match ✗
```

## Impact

Claude Code appends `2>&1` to the vast majority of Bash commands. This means the hook-based rewrite misses most commands that would otherwise match.

From `rtk discover` on my project: ~975K tokens missed, many from commands that would match if not for the `2>&1` suffix.

## Suggested fix

In `rtk rewrite`, strip trailing `2>&1` / `2>/dev/null` before pattern matching, then re-append to the rewritten output.

**Workaround**: I've added pre-processing to my hook script to strip these before calling `rtk rewrite`.

## Environment

- rtk 0.28.2
- macOS Darwin 25.3.0

pszymkowiak:
Closes #530

## Problem

`rtk rewrite "git status 2>&1"` returned nothing because `2>&1` was part of the string passed to pattern matching.

## Fix

Strip trailing redirects (`2>&1`, `2>/dev/null`, etc.) before matching, re-append after rewrite.

```
git status 2>&1       → rtk git status 2>&1
cargo test 2>/dev/null → rtk cargo test 2>/dev/null
git log 2> /dev/null  → rtk git log 2> /dev/null
```

## Test plan

- [x] 1111 tests passing
- [x] Tested with `rtk rewrite` on all redirect patterns
- [x] Existing rewrites unaffected

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -327,9 +327,34 @@ pub fn strip_disabled_prefix(cmd: &str) -> &str {
     trimmed[prefix_len..].trim_start()
 }
 
-/// Rewrite a raw command to its RTK equivalent.
-///
-/// Returns `Some(rewritten)` if the command has an RTK equivalent or is already RTK.
+lazy_static! {
+    // Match trailing shell redirections:
+    // Alt 1: N>&M or N>&- (fd redirect/close): 2>&1, 1>&2, 2>&-
+    // Alt 2: &>file or &>>file (bash redirect both): &>/dev/null
+    // Alt 3: N>file or N>>file (fd to file): 2>/dev/null, >/tmp/out, 1>>log
+    // Note: [^(\\s] excludes process substitutions like >(tee) from false-positive matching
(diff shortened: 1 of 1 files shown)


FlorianBruniaux (approved):
ok

PR #806, opened 2026-03-25
feat(hook): extend Gemini BeforeTool hook to intercept read_file tool…

### LINKED ISSUE #796: Gemini CLI --yolo to read files, it uses the read_file tool
RTK only registers a BeforeTool hook on run_shell_command. But when Gemini CLI uses --yolo to read files, it uses the read_file tool 

Can we extend rtk to hook into read_file also?

Ross-cripto:
… #796

- Route `read_file` alongside `run_shell_command` in `run_gemini()`
- Pre-apply `start_line`/`end_line` slice before filtering, then override both in the response to prevent Gemini's merge from re-applying original (now wrong) line numbers to the filtered file
- Write filtered content to `~/.local/share/rtk/filtered/{ts}-{stem}.rtk` (timestamp-first so lexicographic sort in rotation is chronological)
- Rotate filtered dir to keep last 20 files, mirroring tee.rs pattern
- Fallback to allow for unreadable/binary files, slices <500 chars, and slices with <10% savings

Closes #796

## Summary

-

## Test plan

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -172,15 +172,18 @@ pub fn run_gemini() -> Result<()> {
             }
         }
         "read_file" => {
-            let file_path = json
+            let file_path = match json
                 .pointer("/tool_input/file_path")
                 .and_then(|v| v.as_str())
-                .unwrap_or("");
-
-            if file_path.is_empty() {
-                print_allow();
-                return Ok(());
-            }
+                .filter(|s| !s.is_empty())
+            {
+                Some(p) => p,
+                None => {
+                    eprintln!("rtk: read_file hook: missing or empty file_path, allowing through");
+                    print_allow();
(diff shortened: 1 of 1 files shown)


pszymkowiak (changes requested):
Thanks @Ross-cripto — intercepting `read_file` for Gemini is a good idea. The approach of pre-applying line range, filtering, then writing to a temp file is clever.

**Please retarget to `develop`** — all PRs must target `develop`, not `master`.

A few concerns:
1. **Temp file cleanup** — are the temp files cleaned up after Gemini reads them? Leaking filtered copies of source files could be a concern
2. **`unwrap_or("")` on file_path** — if the path is empty we should probably error, not silently allow
3. **Single file changed** — good scope, easy to review once retargeted

Please retarget and we'll do a deeper review. Thanks!

Issue #823, opened 2026-03-25
feat: rtk init --agent copilot (automated Copilot VS Code + CLI setup)


pszymkowiak:
## Context

The hook `rtk hook copilot` is implemented and works for both VS Code Copilot Chat and Copilot CLI. However, there is no automated setup — users must manually create the required files.

The README documents `rtk init -g` as supporting Copilot, but in reality it only installs the Claude Code hook. `AgentTarget` enum has Claude, Cursor, Windsurf, Cline — but not Copilot.

## Proposal

Add `rtk init -g --agent copilot` that:

1. Creates `.github/hooks/rtk-rewrite.json` — PreToolUse hook config pointing to `rtk hook copilot`
2. Creates/patches `.github/copilot-instructions.md` — prompt-level RTK awareness
3. Verifies `rtk` binary is in PATH

## Files to modify

- `src/main.rs` — add `Copilot` to `AgentTarget` enum
- `src/init.rs` — add `run_copilot_mode()` function
- `hooks/copilot-rtk-awareness.md` — template for instructions (already exists)

## References

- Hook implementation: `src/hook_cmd.rs` (run_copilot, detect_format)
- VS Code hooks spec: https://code.visualstudio.com/docs/copilot/copilot-extensibility-overview
- Existing patterns: `run_cursor_mode()`, `run_windsurf_mode()`, `run_cline_mode()` in `src/init.rs`

nachovoss:
ok so how do i use it in copilot?

zac1wz:
Bumping @nachovoss question

lgrawet:
Since v0.34 :
```rtk init -g --copilot```

See`rtk init --help`
```
     --copilot
          Install GitHub Copilot integration (VS Code + CLI)
```

PR #605, opened 2026-03-15
feat(copilot): cross-platform PreToolUse hook via rtk hook (VS Code + Copilot CLI + Windows)

jeziellopes:
Closes #593

## What this adds

Cross-platform `PreToolUse` hook for GitHub Copilot (VS Code Copilot Chat + Copilot CLI) via `rtk hook` — a native Rust binary. No bash scripts, no `jq`, works on Windows natively.

### How it works

`rtk hook` reads `PreToolUse` JSON from stdin, detects the agent format, responds appropriately:

**VS Code Copilot Chat** (`updatedInput` — transparent rewrite):
```
git status → rtk hook → updatedInput: "rtk git status"  (silent, no denial)
```

**GitHub Copilot CLI** (deny-with-suggestion — CLI ignores `updatedInput` today, [copilot-cli#2013](https://github.com/github/copilot-cli/issues/2013)):
```
git status → rtk hook → deny: "Token savings: use `rtk git status` instead"
             Copilot reads reason → re-runs rtk git status
```

When Copilot CLI adds `updatedInput` support, only `rtk hook` needs a one-line change.

### Also fixed

The old `.github/hooks/rtk-rewrite.json` had wrong key names (`preToolUse` → `PreToolUse`, `bash`/`timeoutSec` → `command`/`timeout`) that prevented it from loading in VS Code.

## Measured gains (validated Copilot CLI session)

| Command | Savings | Tokens saved |
|---------|---------|----------------------------|
| `rtk cargo test` | **99.9%** | 15.8K |
| `rtk cargo clippy --all-targets` | **93.3%** | 5.9K |
| `rtk git diff HEAD~1` | **81.2%** | 2.6K |
| `rtk gh pr list` | **67.0%** | 0.8K |
| **Session total (24 commands)** | **83.5% avg** | **~23.3K** |

## Files

| File | Role |
|------|------|
| `src/hook_cmd.rs` | New `rtk hook` subcommand (13 unit tests) |
| `.github/hooks/rtk-rewrite.json` | Fixed key casing, points to `rtk hook` |
| `hooks/copilot-rtk-rewrite.sh` | **Deleted** — replaced by `rtk hook` |
| `hooks/test-copilot-rtk-rewrite.sh` | Rewritten to test `rtk hook` |
| `hooks/copilot-rtk-awareness.md` | Updated: VS Code row, both output paths |
| `src/init.rs` | Remove bash script checks, update status message |
| `README.md` | VS Code Copilot Chat support, Windows note |
[...]

PR #728, opened 2026-03-19
feat(init): complete Copilot lifecycle - uninstall, idempotency, show_config

jeziellopes:
## Context

PR #859 (merged) shipped a basic `--copilot` flag from PR #824. This PR completes what was shipped — adding the full agent lifecycle that every other agent in rtk has.

Upstream's current `run_copilot()` installs files and exits. No idempotency. No uninstall. `rtk init --show` is blind to Copilot files. This PR fills all of those gaps.

## What this adds on top of what shipped

| Feature | `--copilot` (upstream, merged) | `--agent copilot` (this PR) |
|---------|-------------------------------|------------------------------|
| Installs `.github/` files | ✅ | ✅ |
| Enum pattern (`--agent copilot`) | ❌ standalone `--copilot: bool` | ✅ consistent with `--agent cline/windsurf/cursor` |
| Idempotency | ❌ always re-runs | ✅ early-return with clear status message |
| PATH check for `rtk` binary | ❌ none | ✅ `utils::tool_exists()` — cross-platform, no false positives on Windows |
| `--uninstall --agent copilot` | ❌ no uninstall path exists | ✅ strips RTK section, preserves pre-existing user content |
| `rtk init --show` Copilot status | ❌ `show_config()` has no Copilot awareness | ✅ `[ok]/[--]` for both `.github/` files |
| Unit tests | 0 | 6 new tests |

## Also fixes #836 / supersedes #837

Issue #836 reports that users who ran `rtk init --copilot` early on got `.github/hooks/rtk-rewrite.json` with `"command": "rtk hook"` (missing subcommand), causing a `parse_failure` entry on every Bash tool call.

- This PR's uninstall path removes the broken file cleanly
- This PR's `show_config()` exposes the Copilot files so users can inspect/verify the state
- PR #837 (also open, targets same bug) adds orphaned hook detection to `--show` and `--uninstall`; this PR's broader lifecycle support already covers that surface

## Changes

| File | Change |
|------|--------|
| `src/main.rs` | `AgentTarget::Copilot` enum variant; `"hook"` added to `RTK_META_COMMANDS`; routing to `run_copilot_mode()` |
| `src/hooks/init.rs` | `run_copilot_mode()` (idempotent install), full uninstall branch (content-safe), `show_config()` checks, 6 unit tests |
| `.github/hooks/rtk-rewrite.json` | `"rtk hook"` → `"rtk hook copilot"` (embedded via `include_str!`, single source of truth) |
[...]

PR #837, opened 2026-03-25
fix(init): detect and clean up orphaned .github/hooks/rtk-rewrite.json with broken hook command

### LINKED ISSUE #836: [bug] .github/hooks/rtk-rewrite.json with broken "rtk hook" command causes spurious parse_failure entries per session
## Summary

Users who ran an earlier version of `rtk init --copilot` (or `rtk init` when Copilot support was first introduced in #605) end up with a `.github/hooks/rtk-rewrite.json` containing:

```json
{
  "hooks": {
    "PreToolUse": [
      {
        "type": "command",
        "command": "rtk hook",
        "cwd": ".",
        "timeout": 5
      }
    ]
  }
}
```

`"rtk hook"` requires a subcommand (`gemini` or `copilot`). Without one, it prints help and exits non-zero. This fires on **every Bash tool call**, silently logging a `parse_failure` with `raw_command = "hook"` in `history.db`.

## Reproduction

1. Have `.github/hooks/rtk-rewrite.json` with `"command": "rtk hook"` in your project
2. Run any Bash command through Claude Code
3. Run `rtk gain --failures` → see `Nx hook` entries with `fallback_succeeded = 0`
4. Query the DB directly:
   ```
   sqlite3 ~/Library/Application\ Support/rtk/history.db \
     "SELECT error_message FROM parse_failures WHERE raw_command = 'hook' LIMIT 1;"
   ```
   → Returns the full `rtk hook --help` output as the error message

## Impact

- Inflated failure count in `rtk gain --failures`
- Spurious `fallback_succeeded = 0` entries (no actual fallback occurs — the hook returns non-zero, Claude Code runs the original command unchanged)
- Silent — no visible error to the user

## Root cause

Commit #605 (`feat(copilot): add Copilot hook support`) included `.github/hooks/rtk-rewrite.json` with `"command": "rtk hook"` instead of `"command": "rtk hook copilot"`. The fix in the pending Copilot integration (see #823) generates the correct `"rtk hook copilot"` command, but does not migrate or clean up existing broken files.

Neither `rtk init --show` nor `rtk init --uninstall` check `.github/hooks/` for orphaned files.

## Note: different from #712

Issue #712 reports that `rtk init` doesn't wire up `rtk hook` at all.
This issue is the inverse: `rtk init` **did** write `.github/hooks/rtk-rewrite.json`, but with an incomplete command that silently breaks on every invocation.

## Proposed fix

1. `rtk init --show` detects `.github/hooks/rtk-rewrite.json` with `"rtk hook"` and warns with actionable guidance
2. `rtk init -g --uninstall` removes the broken file as part of cleanup

PR with the fix is attached.

rolf-quadra:
Fixes #836

## What

- **`rtk init --show`**: warns when `.github/hooks/rtk-rewrite.json` exists with `"command": "rtk hook"` (missing subcommand), with actionable guidance pointing to `rtk gain --failures` and the fix
- **`rtk init -g --uninstall`**: removes the broken file as part of standard cleanup (only when the command is the broken `"rtk hook"` — a correctly configured `"rtk hook copilot"` file is left untouched)

## Why

Users who ran an earlier `rtk init --copilot` (introduced in #605) got `.github/hooks/rtk-rewrite.json` with `"command": "rtk hook"` (no subcommand). Since `rtk hook` requires `gemini` or `copilot`, it prints help and exits non-zero on every Bash tool call. RTK silently logs a `parse_failure` with `raw_command = "hook"` and `fallback_succeeded = 0` in `history.db`, inflating `rtk gain --failures` stats with no visible error to the user.

The pending Copilot integration (#823) correctly generates `"rtk hook copilot"`, but neither `rtk init --show` nor `rtk init --uninstall` check `.github/hooks/` for these orphaned files.

## Changes

`src/init.rs` only — 66 lines added, no deletions, no new dependencies.

## Test

```bash
# Reproduce
echo '{"hooks":{"PreToolUse":[{"type":"command","command":"rtk hook","cwd":".","timeout":5}]}}' \
  > .github/hooks/rtk-rewrite.json

# Should warn
rtk init --show

# Should remove
rtk init -g --uninstall
ls .github/hooks/rtk-rewrite.json  # should be gone
```


jeziellopes:
Hi @rolf-quadra — just wanted to flag that PR #728 (`feat(init): complete Copilot lifecycle`) already covers the scenario you're fixing here:

- `--uninstall --agent copilot` removes `.github/hooks/rtk-rewrite.json` cleanly
- `rtk init --show` gains `[ok]/[--]` status for both Copilot files

If #728 gets merged first this PR may become redundant, but if yours lands first the orphaned-hook detection would be a nice complement. Either way, thanks for tackling #836 — it's the same root cause I ran into.

PR #843, opened 2026-03-26
fix(config): add serde(default) to TeeConfig fields

fkztw:
## Summary

- `TeeConfig` fields lack `#[serde(default)]`, causing the entire `Config` deserialization to fail when users specify a `[tee]` section in `config.toml` but omit some fields (e.g. `max_file_size`)
- Since call sites use `Config::load().map(...).unwrap_or_default()` to silently handle errors, all user settings (including `hooks.exclude_commands`) are silently ignored with no warning
- Add `#[serde(default = "...")]` to every `TeeConfig` field so partial `[tee]` sections deserialize correctly
- Add tests for partial and empty deserialization

## Reproduction

```toml
# ~/.config/rtk/config.toml (Linux) or ~/Library/Application Support/rtk/config.toml (macOS)
[hooks]
exclude_commands = ["curl", "helm"]

[tee]
enabled = true
mode = "failures"
max_files = 20
# max_file_size is omitted — this causes the ENTIRE config to fail silently
```

```bash
$ rtk config
Config: ~/Library/Application Support/rtk/config.toml

Error: TOML parse error at line 35, column 1
   |
35 | [tee]
   | ^^^^^
missing field `max_file_size`
```

When `rtk config` is not called directly, `Config::load()` fails silently and falls back to defaults — all user settings are ignored without any warning.

## Test plan

- [x] `cargo fmt --all --check` — passed
- [x] `cargo test --all` — 1116 passed, 3 ignored
- [x] New `test_tee_config_partial_deserialize` — verifies omitting `max_file_size` does not fail
- [x] New `test_tee_config_empty_deserialize` — verifies a completely empty `[tee]` section uses defaults

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+* **config:** add `serde(default)` to `TeeConfig` fields to prevent silent parse failure when `[tee]` section is partially specified ([#843](https://github.com/rtk-ai/rtk/pull/843)) ([08068447](https://github.com/rtk-ai/rtk/commit/08068447))
 * **diff:** correct truncation overflow count in condense_unified_diff ([#833](https://github.com/rtk-ai/rtk/pull/833)) ([5399f83](https://github.com/rtk-ai/rtk/commit/5399f83))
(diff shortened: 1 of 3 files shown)


KuSh:
Thanks for digging into this one. The [tee] section was reworked by #3278, merged 2026-09-09, which moved recovery settings to [retriever] and kept [tee] as a legacy section parsed into a struct where every field is optional and the struct carries serde(default), so a partial [tee] block no longer fails the whole config load. The regression case you added is covered on develop by test_legacy_tee_disabled_maps_to_disabled_mode and test_legacy_tee_never_mode_maps_to_disabled_mode, both of which parse a [tee] section with a single field. Closing as already fixed, but please comment or reopen if you find a config shape that still parses badly.

Issue #154, opened 2026-02-16
refactor: migrate rtk-rewrite.sh hook from shell to Rust subcommand


pszymkowiak:
## Problem

The Claude Code rewrite hook (`rtk-rewrite.sh`) is currently a ~200-line Bash script that:
- Requires `jq` as an external dependency
- Uses fragile regex/case pattern matching
- Has a separate shell-based test suite (`test-rtk-rewrite.sh`)
- Forks shell + jq twice per invocation (read stdin JSON, write stdout JSON)

This works but has become technical debt as the project grows.

## Proposal

Migrate the hook logic to a native Rust subcommand: `rtk hook-rewrite`

The Claude Code hook config would change from:
```json
"command": "bash ~/.claude/hooks/rtk-rewrite.sh"
```
to:
```json
"command": "rtk hook-rewrite"
```

## Benefits

| | Shell (current) | Rust (proposed) |
|---|---|---|
| **Dependencies** | Requires `jq` | None (rtk binary only) |
| **Testing** | Separate bash test script | `#[cfg(test)]` integrated |
| **Performance** | Fork shell + fork jq ×2 | Single binary, serde_json |
| **Maintainability** | Bash regex, case statements | Rust pattern matching |
| **Adding rules** | Copy-paste case blocks | Typed structs + match |
| **Error handling** | Silent failures | Result<> propagation |

## Scope

- New subcommand `rtk hook-rewrite` that reads JSON from stdin, applies rewrite rules, outputs JSON to stdout
- Port all existing rewrite rules from `rtk-rewrite.sh`
- Port all test cases from `test-rtk-rewrite.sh` to Rust unit tests
- Keep the shell script as a fallback for users who haven't updated their hook config
- Update `rtk init` to configure the Rust-based hook by default
- Integrate audit logging (PR #151) directly into the Rust implementation

## Non-goals

- Changing the Claude Code hooks protocol (stdin JSON → stdout JSON)
- Removing shell script support immediately (deprecate, don't break)

pszymkowiak:
> [w] **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 95% |

### Summary

[...]

PR #862, opened 2026-03-26
fix(hook): replace rtk-rewrite.sh with native rtk hook claude (no jq dependency)

### LINKED ISSUE #430: Hook silently fails when jq is not installed
The rtk-rewrite.sh hook silently exits when jq is not installed on the system, causing rtk to appear as if it's working while all commands bypass the rewrite pipeline entirely.

## Steps to Reproduce

1. Install rtk (>= 0.23.0)
2. Set up the Claude Code hook via rtk setup
3. Do not have jq installed (or uninstall it)
4. Run any command through Claude Code (e.g. git status)

- Expected: Some indication that the hook is not functioning
- Actual: Hook exits silently, commands run without rtk rewrite, and the user has no way of knowing

## Root Cause

In rtk-rewrite.sh:

```sh
if ! command -v jq &>/dev/null; then
  exit 0
fi
```

This exits cleanly with no warning.

## Suggestions

A few possible approaches:

1. Print a warning when jq is missing, similar to the version guard
2. Check for jq during rtk setup and warn the user before installing the hook
3. Remove the jq dependency entirely

## Environment

- rtk 0.27.2
- macOS (Apple Silicon)
- Claude Code with PreToolUse hook

FlorianBruniaux:
## Problem

\`rtk init -g\` installs a shell script (\`rtk-rewrite.sh\`) as a Claude Code PreToolUse hook. That script has a hard dependency on \`jq\` to parse and build JSON.

When \`jq\` is not installed, the hook silently exits 0 with no output — Claude Code sees no rewrite, no error, nothing. Users end up with RTK installed but zero commands being rewritten, with no indication of why.

The warning is printed to stderr inside the hook process, which is swallowed by Claude Code's hook runner. So the user never sees it:

\`\`\`bash
if ! command -v jq &>/dev/null; then
  echo "[rtk] WARNING: jq is not installed..." >&2
  exit 0  # silent passthrough — user never knows
fi
\`\`\`

This affects anyone on macOS who did not install jq via Homebrew, and Linux users on minimal installs.

## Solution

Replace the shell script with a native Rust subcommand \`rtk hook claude\` that uses \`serde_json\` for JSON parsing. No external dependencies — if \`rtk\` is in PATH, the hook works.

The hook command registered in \`settings.json\` changes from:
\`\`\`json
{ "command": "/Users/you/.claude/hooks/rtk-rewrite.sh" }
\`\`\`
to:
\`\`\`json
{ "command": "rtk hook claude" }
\`\`\`

## Changes

- Add \`rtk hook claude\` subcommand (\`hook_cmd.rs\`): reads Claude Code PreToolUse JSON from stdin, rewrites Bash tool commands, preserves all \`tool_input\` fields (description, etc.)
- Extract core logic into \`process_claude_hook(input: &str) -> Option<String>\` for direct unit testability
- Update \`rtk init -g\` to register \`rtk hook claude\` instead of writing/installing \`rtk-rewrite.sh\`
- Auto-migrate: if \`rtk-rewrite.sh\` is already in \`settings.json\`, re-running \`rtk init -g\` replaces it in-place with \`rtk hook claude\`
- Remove \`prepare_hook_paths\` and \`ensure_hook_installed\` (dead code after this change)
- Legacy \`hooks/rtk-rewrite.sh\` stays in the repo for reference and for users who installed it manually

## Upgrade path for existing users

\`\`\`bash
cargo install rtk   # or: brew upgrade rtk-ai/tap/rtk
rtk init -g         # auto-migrates rtk-rewrite.sh → rtk hook claude in settings.json
# Restart Claude Code
\`\`\`

## Test plan

- [ ] \`rtk hook claude\` rewrites \`git status\` → \`rtk git status\`
- [ ] Non-Bash tools (Edit, Write...) pass through silently
- [ ] Already-\`rtk\`-prefixed commands pass through silently
- [ ] \`rtk init -g\` on fresh install registers \`rtk hook claude\` in settings.json
- [ ] \`rtk init -g\` on existing install with \`rtk-rewrite.sh\` migrates automatically
- [ ] 1139 unit tests pass, 0 clippy warnings

Closes #430. Implements #154.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Features
+
+* **hook:** add `rtk hook claude` native Rust subcommand — replaces `rtk-rewrite.sh`, no `jq` dependency required ([#862](https://github.com/rtk-ai/rtk/pull/862))
+* **init:** `rtk init -g` now installs `rtk hook claude` (native binary) instead of the shell script; auto-migrates existing `rtk-rewrite.sh` entries on re-run ([#862](https://github.com/rtk-ai/rtk/pull/862))
+
 ### Bug Fixes
 
+* **hook:** fix silent failure when `jq` is not installed — `rtk-rewrite.sh` would exit 0 without rewriting any commands ([#430](https://github.com/rtk-ai/rtk/issues/430), [#862](https://github.com/rtk-ai/rtk/pull/862))
 * **diff:** correct truncation overflow count in condense_unified_diff ([#833](https://github.com/rtk-ai/rtk/pull/833)) ([5399f83](https://github.com/rtk-ai/rtk/commit/5399f83))
 * **git:** replace vague truncation markers with exact counts in log and grep output ([#833](https://github.com/rtk-ai/rtk/pull/833)) ([185fb97](https://github.com/rtk-ai/rtk/commit/185fb97))
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -93,11 +93,12 @@ rtk gain  # MUST show token savings, not "command not found"
 
 ```bash
 rtk init -g
-# → Installs hook to ~/.claude/hooks/rtk-rewrite.sh
+# → Registers "rtk hook claude" in settings.json (native Rust, no jq required)
 # → Creates ~/.claude/RTK.md (10 lines, meta commands only)
 # → Adds @RTK.md reference to ~/.claude/CLAUDE.md
 # → Prompts: "Patch settings.json? [y/N]"
 # → If yes: patches + creates backup (~/.claude/settings.json.bak)
+# → If rtk-rewrite.sh was previously installed: auto-migrates to native hook
 
 # Automated alternatives:
 rtk init -g --auto-patch    # Patch without prompting
@@ -113,7 +114,7 @@ rtk init --show  # Check hook is installed and executable
 Claude Code's hook registry. RTK adds a PreToolUse hook that rewrites commands transparently. Without this, Claude won't invoke the hook automatically.
 
 ```
-  Claude Code          settings.json        rtk-rewrite.sh        RTK binary
+  Claude Code          settings.json        rtk hook claude       RTK binary
        │                    │                     │                    │
        │  "git status"      │                     │                    │
        │ ──────────────────►│                     │                    │
@@ -247,7 +248,7 @@ rtk vitest run
 rtk init -g --uninstall
 
 # What gets removed:
-#   - Hook: ~/.claude/hooks/rtk-rewrite.sh
+#   - Hook: "rtk hook claude" entry from settings.json
 #   - Context: ~/.claude/RTK.md
 #   - Reference: @RTK.md line from ~/.claude/CLAUDE.md
 #   - Registration: RTK hook entry from settings.json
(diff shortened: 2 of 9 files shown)


pszymkowiak (changes requested):
Excellent work @FlorianBruniaux — this is significantly better than #785. The migration logic (`migrate_legacy_hook`), the 10 tests, the `description` field preservation, and the correct absence of `permissionDecision` in the output all show careful attention to the Claude Code protocol.

The jq removal solves a real pain point (#430) and the native Rust path is the right direction.

**P0 — Security: permissions deny/ask not called**

`process_claude_hook()` calls `get_rewritten()` → `rewrite_command()` but never `check_command()` from `src/permissions.rs`. The current bash script calls `rtk rewrite` which internally runs `check_command()` and exits with:
- Exit 2 → deny (pass through unchanged, Claude Code native deny handles it)
- Exit 3 → ask (rewrite but omit `permissionDecision`, Claude Code prompts user)
- Exit 0 → allow (rewrite + auto-allow)

Without this, any command matching a user's `permissions.deny` or `permissions.ask` rules in `settings.json` is silently auto-allowed. This is a security regression from PR #576.

Fix sketch:
```rust
pub(crate) fn process_claude_hook(input: &str) -> Option<String> {
    // ... parse JSON, extract cmd ...
    
    // Check permissions BEFORE rewriting
    match crate::permissions::check_command(cmd) {
        PermissionVerdict::Deny => return None,  // pass through, Claude Code handles deny
        PermissionVerdict::Ask => {
            let rewritten = get_rewritten(cmd)?;
            // Rewrite but omit permissionDecision — Claude Code will prompt
            return Some(json!({
                "hookSpecificOutput": {
                    "updatedInput": build_updated_input(&v, &rewritten)
                }
            }).to_string());
        }
        PermissionVerdict::Allow => {}  // continue to normal rewrite
    }
    
    let rewritten = get_rewritten(cmd)?;
    // ... current code with permissionDecision omitted (correct) ...
}
```

**P1 — Audit logging dropped**

`RTK_HOOK_AUDIT=1` writes to `~/.local/share/rtk/hook-audit.log` in the bash script. The native command doesn't replicate this. Users relying on `rtk hook-audit` for rewrite metrics will get an empty log. Gate on `std::env::var("RTK_HOOK_AUDIT")` and write the same pipe-separated format.

**P1 — Integrity check orphaned**

`store_hash()` is now `#[cfg(test)]` only, but `runtime_check()` in `integrity.rs` still looks for `~/.claude/hooks/rtk-rewrite.sh`. After migration, the file is gone → `runtime_check()` returns `NotInstalled` forever. Either disable `runtime_check` when using native hook mode, or adapt it to verify the `"rtk hook claude"` entry in `settings.json` instead.

**P2 — Uninstall path**

Does `rtk uninstall` also clean up the `"rtk hook claude"` entry from settings.json? The current uninstall removes `rtk-rewrite.sh` file — verify it also handles the new command string.

**P2 — Scope: other agents**

`rtk hook gemini` and `rtk hook copilot` already exist as native Rust commands. Do they also lack the permission check? Worth verifying for consistency — if Claude gets deny/ask, Gemini and Copilot should too.

---

**Summary**: fix the permissions check (P0), add audit logging (P1), and this is ready. The core implementation is solid — much cleaner than #785. Also closes the discussion on #785 since this PR supersedes it.

aeppling:
Hey @FlorianBruniaux ,

Covered by https://github.com/rtk-ai/rtk/pull/956 which is up to date with latest versions

Will close this once 956 released

PR #869, opened 2026-03-26
fix(refacto-codebase): technical docs & sub folders

aeppling:
## Summary

Part 1 of codebase refacto for better onboarding, cleaner code for an extensible and collaborative codebase


aeppling:
Hello contributors !

We are missing some CLA signs by the following contributors to be able to release this :

@timeshift92
@tobwen 
@TechWizard9999

You should be able to sign this by clicking the link in the git check status just below this message !

Thanks for contributing to RTK

PR #870, opened 2026-03-26
fix(err): propagate exit code from underlying command (#846)

### LINKED ISSUE #846: bug: rtk err swallows exit code (returns 0 instead of underlying exit code)
## Problem

`rtk err false` returns exit 0 even though `false` returns exit 1.

```bash
rtk err false
# Output: [FAIL] Command failed (exit code: Some(1))
# Exit: 0  ← should be 1
```

The error is detected and reported in the output, but the exit code is not propagated.

## Expected

`rtk err <cmd>` should propagate the exit code of the underlying command, like all other RTK filters do.

## Location

`src/runner.rs` — the `err` subcommand path likely returns `Ok(())` instead of calling `std::process::exit(code)`.

## Found by

Multipass integration test suite (PR #830).

ousamabenyounes:
## Problem

`rtk err <cmd>` and `rtk test <cmd>` always returned exit code **0**, even when the underlying command failed. This broke CI/CD pipelines that rely on non-zero exit codes for error detection.

```bash
rtk err false
# Output: [FAIL] Command failed (exit code: Some(1))
# Exit:   0  ← should be 1
```

## Root Cause

`run_err()` and `run_test()` in `src/cmds/rust/runner.rs` computed `exit_code` but only passed it to `tee_and_hint()` — they never called `std::process::exit()`.

## Fix

Add `std::process::exit(exit_code)` after output when the child exits non-zero, matching the established pattern in `git.rs`, `gt_cmd.rs`, and all other command modules.

## Tests

Two `#[ignore]` subprocess regression tests added:
- `test_run_err_propagates_nonzero_exit_code` — `rtk err false` → exits 1
- `test_run_err_zero_exit_on_success` — `rtk err true` → exits 0

```
cargo test cmds::rust::runner -- --include-ignored
# 3 passed ✅
```

Fixes #846

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -61,6 +61,9 @@ pub fn run_err(command: &str, verbose: u8) -> Result<()> {
         println!("{}", rtk);
     }
     timer.track(command, "rtk run-err", &raw, &rtk);
+    if exit_code != 0 {
+        std::process::exit(exit_code);
+    }
     Ok(())
 }
 
@@ -102,6 +105,9 @@ pub fn run_test(command: &str, verbose: u8) -> Result<()> {
         println!("{}", summary);
     }
     timer.track(command, "rtk run-test", &raw, &summary);
+    if exit_code != 0 {
+        std::process::exit(exit_code);
+    }
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


ousamabenyounes:
Closing as superseded by upstream commit `32ff466` ("fix(cmds): exit code unified return code flow"), which refactored `runner.rs` and 30+ filter modules to thread child exit codes via `Result<i32>` → `std::process::exit(code)` in `main.rs::run_cli`. The #846 bug is fixed via a stronger architecture. Thanks!

PR #871, opened 2026-03-26
docs: update grep descriptions for passthrough behavior

BadassBison:
## Summary

Doc-only follow-up to #791 (grep false negatives, output mangling, and truncation annotations).

Updates all relevant documentation to reflect the behavioral changes made in that PR:

- `rtk grep` now passes through raw `file:line:content` output for <=50 matches, and uses the grouped format only for >50 matches
- `--no-ignore-vcs` is used (not `--no-ignore`) so `.ignore`/`.rgignore` are still respected
- Savings range updated to `0-90%` to reflect passthrough for small result sets

## Files changed

| File | Change |
|------|--------|
| `README.md` | grep description: "Grouped search results" → "Search results (raw <=50, grouped >50)" |
| `README_es.md` / `fr` / `ja` / `ko` / `zh` | Same update in each language |
| `ARCHITECTURE.md` | SYSTEM savings range: `50-90%` → `0-90%` |
| `INSTALL.md` | grep description updated |
| `docs/AUDIT_GUIDE.md` | grep savings note updated |
| `docs/FEATURES.md` | Full grep objective updated, `--no-ignore` → `--no-ignore-vcs` |

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -281,7 +281,7 @@ cp ~/.claude/settings.json.bak ~/.claude/settings.json
 ```bash
 rtk ls .              # Compact tree view
 rtk read file.rs      # Optimized reading
-rtk grep "pattern" .  # Grouped search results
+rtk grep "pattern" .  # Filtered search results (file:line:content, --no-ignore-vcs)
 ```
 
 ### Git
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ rtk read file.rs                # Smart file reading
 rtk read file.rs -l aggressive  # Signatures only (strips bodies)
 rtk smart file.rs               # 2-line heuristic code summary
 rtk find "*.rs" .               # Compact find results
-rtk grep "pattern" .            # Grouped search results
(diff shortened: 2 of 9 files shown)


KuSh:
Thanks for taking the time to update the grep wording across all the language files. The behaviour these lines describe has changed twice since this was opened: #2641 (merged 2026-06-26) removed the rg substitution and the --no-ignore-vcs flag, and #3182 (merged 2026-07-28) replaced the 50-match passthrough threshold with a rule that only emits the grouped form when capping actually shrinks the output. Because of that, the new wording here, file:line:content and --no-ignore-vcs, no longer matches what rtk grep does, so the text cannot land as written. The docs on develop are still stale in the other direction, since README.md, INSTALL.md, the translated READMEs, docs/usage/AUDIT_GUIDE.md and docs/usage/FEATURES.md all still call the output grouped by file, and a fresh pass describing the current never-worse model would be very welcome. Please comment or reopen if I have missed a part of this.

Issue #827, opened 2026-03-25
bug(critical): silent truncation causes data loss — LLM takes decisions on incomplete diffs


pszymkowiak:
## User report

> I did a diff between 2 large files (+1000 lines of data/yaml) and asked Claude for a summary.
> RTK truncated the diff, Claude didn't tell me or didn't notice.
> I made important decisions based on this diff that I can't undo (very frustrating).
> Claude confirmed the next day that RTK had truncated the data.
> I lost trust, I don't have time for this, I'm uninstalling.

## Root cause

`rtk diff` silently truncates output at 3 levels:
1. **Max 50 changes** shown (`.take(50)`) — no warning to LLM
2. **Lines truncated at 70-80 chars** — content silently cut
3. **No machine-readable truncation marker** — LLM cannot detect data loss

Same pattern exists in `rtk git diff`, `rtk read`, and potentially other commands.

## Impact

- **P0 severity**: Users make irreversible decisions on incomplete data
- **Trust killer**: Once bitten, users uninstall RTK
- **LLM blind spot**: Claude/Copilot cannot tell the user that data was truncated

### Short term (must-fix)
Add a clear, machine-readable truncation warning at the end of any truncated output:

```
[warn] Output truncated: showing 50/847 changes. Run `diff file1 file2` for full output.
```

This lets the LLM:
1. Tell the user the diff is incomplete
2. Suggest running the raw command for full output
3. Avoid making decisions on partial data

### Long term
- Add `--full` flag to disable truncation for critical commands (diff, read)
- Consider making `rtk diff` of large files pass through to raw `diff` with a warning
- Tee integration: save full output to disk, print hint

## Affected commands

| Command | Truncation | Current warning |
|---------|-----------|-----------------|
| `rtk diff` | 50 changes, 70-80 char lines | "... +N more changes" (easy to miss) |
| `rtk git diff` | hunk limits | None |
| `rtk read` | filter levels strip content | None |
| `rtk git log` | body truncated to 3 lines | None |

## Related

- #822 — file size returned by RTK is truncated
- #781 — diff reports files identical when they differ
- #621 — signal truncation in git diff/log

## Priority

**P0** — This is a trust and safety issue. Users who experience silent data loss will uninstall RTK permanently.

PR #881, opened 2026-03-27
fix(grep): read stdin when piped instead of searching filesystem

### LINKED ISSUE #838: bug(grep): stdin pipe silently ignored — searches codebase instead of piped input
First off — RTK is an amazing tool 👍. Saves a ton of tokens daily. Found this edge case while debugging test logs.
`rtk grep` silently ignores stdin when used in a pipe. Instead of filtering the piped input, it searches the codebase, producing completely unrelated results. There's no warning that stdin is being discarded.
```bash
cat /tmp/wdio17.log | grep "Error in"
cat /tmp/wdio17.log | rtk grep "Error in"
```
Passing the file as a positional argument works correctly:
```bash
rtk grep "Error in" /tmp/wdio17.log
```
One of:
1. **Read stdin** when piped (like `grep` / `rg` do) — filter the piped input
2. **Warn/error** that stdin is not supported, suggesting the positional argument form
3. **Document** in `--help` that stdin is not supported
- Stdin is silently discarded
- Searches current working directory instead
- Returns unrelated matches from codebase files
- Panics with `Broken pipe (os error 32)` when pipe closes
This is a **silent correctness issue** — the user gets plausible-looking output that comes from a completely different source than intended. An LLM agent piping command output through `rtk grep` would act on wrong data without any signal that something went wrong.
Similar in spirit to #827 (silent truncation causing decisions on wrong data), but for stdin rather than output truncation.
- rtk version: 0.33.x
- OS: macOS 15.4 (Darwin 25.3.0, Apple Silicon)
- Shell: zsh
- Is stdin support intentionally omitted, or is this a gap?
- If intentional, would you accept a PR adding a warning when stdin is a pipe but gets ignored?

qte77:
Fixes #838

  ## Summary

  - `rtk grep` silently ignored piped stdin, always searching the filesystem instead
  - Detect TTY via `std::io::IsTerminal` — when stdin is piped, read from it instead of passing a path to `rg`
  - Also includes upstream fixes from #833 (truncation accuracy) and additional improvements:
    - `rtk read` default filter changed from `minimal` to `none` (prevents silent content loss)
    - `rtk read` safety fallback when filter empties a non-empty file
    - `rtk diff` no longer truncates at 50 changes (#827) — all changes shown in full
    - Swift `test` subcommand support added to discover registry
    - Copilot VS Code integration added to `rtk init`

  ## Test plan

  - [x] `echo "hello world" | rtk grep hello` reads stdin (was: searched codebase)
  - [x] `rtk grep pattern src/` still works (TTY mode)
  - [ ] Rebase on `develop` after #826 refactor

### DIFF
--- a/src/grep_cmd.rs
+++ b/src/grep_cmd.rs
@@ -4,6 +4,7 @@ use crate::utils::resolved_command;
 use anyhow::{Context, Result};
 use regex::Regex;
 use std::collections::HashMap;
+use std::io::IsTerminal;
 
 #[allow(clippy::too_many_arguments)]
 pub fn run(
@@ -25,8 +26,14 @@ pub fn run(
     // Fix: convert BRE alternation \| → | for rg (which uses PCRE-style regex)
     let rg_pattern = pattern.replace(r"\|", "|");
 
+    let stdin_is_tty = std::io::stdin().is_terminal();
+
     let mut rg_cmd = resolved_command("rg");
-    rg_cmd.args(["-n", "--no-heading", &rg_pattern, path]);
+    if stdin_is_tty {
+        rg_cmd.args(["-n", "--no-heading", &rg_pattern, path]);
+    } else {
+        rg_cmd.args(["-n", "--no-heading", &rg_pattern]);
+    }
 
     if let Some(ft) = file_type {
         rg_cmd.arg("--type").arg(ft);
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Closing this PR for inactivity — the CLA was never signed and there's been no activity since March. This isn't a judgment on the change itself.

If you'd still like to land it, just sign the CLA (https://cla-assistant.io/rtk-ai/rtk) and ping a maintainer to reopen — happy to pick it back up. Thanks for the contribution!

gene1wood:
@qte77 Is not a person (it's an AI) and so it may not have seen the asks here for the update to the branch and to have the CLA signed.

PR #894, opened 2026-03-28
feat(rewrite): consult user TOML filters as fallback

### LINKED ISSUE #820: rtk rewrite should consult user TOML filters
## Problem

`rtk rewrite` only checks the built-in Rust registry. It doesn't look at user TOML filters in `~/.config/rtk/filters.toml` (or platform equivalent like `~/Library/Application Support/rtk/filters.toml` on macOS).

This means the Claude Code hook (`rtk-rewrite.sh`) can't auto-route commands that only have user-defined TOML filters. The hook delegates to `rtk rewrite`, gets exit 1, and passes through without filtering.

### Reproduction

```bash
# Add a user filter for ssh
cat >> ~/Library/Application\ Support/rtk/filters.toml << 'TOML'
[filters.ssh]
description = "Strip SSH connection noise"
match_command = "^ssh\\b"
strip_ansi = true
strip_lines_matching = [
  "^Warning: Permanently added",
  "^Connection to .* closed",
]
TOML

# Filter works when called directly
echo "Warning: Permanently added 'host' to known hosts." | rtk ssh root@host ls
# -> filters correctly

# But rtk rewrite doesn't know about it
rtk rewrite "ssh root@host ls"
# -> exit 1 (no rewrite)
```

### Current workaround

Patch the hook to maintain a manual list of user TOML commands as fallback:

```bash
if [ $? -ne 0 ]; then
  USER_TOML_CMDS="ssh|scp|nono|python3"
  FIRST_WORD=$(echo "$CMD" | awk '{print $1}')
  if echo "$FIRST_WORD" | grep -qE "^($USER_TOML_CMDS)$"; then
    REWRITTEN="rtk $CMD"
  else
    exit 0
  fi
fi
```

Works but requires keeping the hook list in sync with `filters.toml` manually.

### Proposed fix

Have `rtk rewrite` check user TOML filters as a fallback when the built-in registry has no match. Built-in priority stays the same, it just extends coverage to user-defined commands.

Happy to send a PR if this approach makes sense.

tmchow:
## Summary

- `rtk rewrite` now checks user TOML filters when no built-in Rust handler matches
- Commands with TOML-only filters (e.g. `ssh`, `jq`, `just`) get rewritten to `rtk <cmd>` so the hook routes them through the TOML filter engine
- Built-in Rust registry retains priority over TOML filters

## Test plan

- [x] `cargo test --all` (1131 passed)
- [x] New tests: TOML-only command matched, no filter returns None, registry priority preserved

## What changed

In `src/hooks/rewrite_cmd.rs`, the `None` branch of `registry::rewrite_command()` now calls `toml_filter::find_matching_filter(cmd)` before exiting with code 1. If a TOML filter matches, the command is rewritten as `rtk <cmd>` and printed to stdout with the same permission verdict handling as built-in commands.

Three new unit tests verify:
- `ssh root@host ls` (TOML-only) is matched by `find_matching_filter` but not by `rewrite_command`
- `htop` has neither registry nor TOML match
- `git status` (registry command) is handled by the registry, never reaching the TOML fallback

Fixes #820

### DIFF
--- a/src/hooks/rewrite_cmd.rs
+++ b/src/hooks/rewrite_cmd.rs
@@ -1,6 +1,7 @@
 //! Translates a raw shell command into its RTK-optimized equivalent.
 
 use super::permissions::{check_command, PermissionVerdict};
+use crate::core::toml_filter;
 use crate::discover::registry;
 use std::io::Write;
 
@@ -42,8 +43,24 @@ pub fn run(cmd: &str) -> anyhow::Result<()> {
             PermissionVerdict::Deny => unreachable!(),
         },
         None => {
-            // No RTK equivalent. Exit 1 = passthrough.
-            // Claude Code independently evaluates its own ask rules on the original cmd.
+            // Fallback: check user TOML filters (project-local and global).
+            if toml_filter::find_matching_filter(cmd).is_some() {
+                let rewritten = format!("rtk {}", cmd);
+                match verdict {
+                    PermissionVerdict::Allow => {
+                        print!("{}", rewritten);
+                        let _ = std::io::stdout().flush();
+                        return Ok(());
+                    }
+                    PermissionVerdict::Ask | PermissionVerdict::Default => {
+                        print!("{}", rewritten);
+                        let _ = std::io::stdout().flush();
+                        std::process::exit(3);
+                    }
+                    PermissionVerdict::Deny => unreachable!(),
+                }
+            }
+            // No RTK equivalent (built-in or TOML). Exit 1 = passthrough.
             std::process::exit(1);
         }
     }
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for this. The TOML filter fallback landed on develop in #2748 (merged 2026-07-03), which implements it inside the rewrite registry in src/discover/registry.rs so `rtk rewrite "ssh root@host ls"` now prints `rtk ssh root@host ls`, with additional handling for exclude lists, absolute paths, reserved command names and pipelines. Your test cases are covered by the test_rewrite_toml_* suite in the same file. Note that filter files are now trust gated, so `rtk trust` is needed before a user filter is consulted. If some part of this PR was missed, please comment or reopen.

PR #895, opened 2026-03-28
fix(discover): expand rules coverage for git, kubectl, jq and ignore ln

fkztw:
## Description
This PR expands the token savings discovery coverage (`rtk discover`) for several common command line patterns:
- **`git`**: Broadened matching criteria to accurately capture complex arguments including `git clone`, `git commit -am`, and strings wrapped in quotes.
- **`kubectl`**: Enhanced support to detect multi-flag statements typical in Kubernetes operational manifests.
- **`jq`**: Corrected the filter assignment to properly resolve coverage savings calculations for `jq` queries.
- **`ln`**: Added the local link command to the global ignore list (`IGNORED_PREFIXES`), to prevent trivial commands from cluttering standard output optimization reports.

### DIFF
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -15,7 +15,7 @@ pub struct RtkRule {
 
 // Patterns ordered to match RULES indices exactly.
 pub const PATTERNS: &[&str] = &[
-    r"^git\s+(?:-[Cc]\s+\S+\s+)*(status|log|diff|show|add|commit|push|pull|branch|fetch|stash|worktree)",
+    r"^git\s+(?:-[Cc]\s+\S+\s+)*(status|log|diff|show|add|commit|push|pull|branch|fetch|stash|worktree|checkout|switch|remote|config|ls-files|tag|merge|rebase|reset|clean|clone|init|submodule|cherry-pick|revert|rm|mv|bisect|blame|reflog|shortlog|describe|archive|bundle|notes)",
     r"^gh\s+(pr|issue|run|repo|api|release)",
     r"^cargo\s+(build|test|clippy|check|fmt|install)",
     r"^pnpm\s+(list|ls|outdated|install)",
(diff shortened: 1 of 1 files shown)


src/discover/rules.rs
```diff
@@ -88,6 +88,8 @@ pub const PATTERNS: &[&str] = &[
     r"^trunk\s+build",
     r"^uv\s+(sync|pip\s+install)\b",
     r"^yamllint\b",
+    // jq — TOML filter exists (src/filters/jq.toml) but was missing from discover
```
KuSh, line 91: This should be handled by https://github.com/rtk-ai/rtk/pull/894

src/discover/rules.rs
```diff
@@ -706,6 +717,8 @@ pub const IGNORED_PREFIXES: &[&str] = &[
     "while ",
     "if ",
     "case ",
+    "ln ",
+    "ln\t",
```
KuSh, line 721: I'm not sure this should be added. Tab is a completion key in most terminals, it shouldn't be easy to use it in command lines and I'm not sure LLM would use it either.

If it's a real case however, this should be handled globally and not just for that particular command

src/discover/rules.rs
```diff
@@ -15,7 +15,7 @@ pub struct RtkRule {
 
 // Patterns ordered to match RULES indices exactly.
 pub const PATTERNS: &[&str] = &[
-    r"^git\s+(?:-[Cc]\s+\S+\s+)*(status|log|diff|show|add|commit|push|pull|branch|fetch|stash|worktree)",
+    r"^git\s+(?:-[Cc]\s+\S+\s+)*(status|log|diff|show|add|commit|push|pull|branch|fetch|stash|worktree|checkout|switch|remote|config|ls-files|tag|merge|rebase|reset|clean|clone|init|submodule|cherry-pick|revert|rm|mv|bisect|blame|reflog|shortlog|describe|archive|bundle|notes)",
```
KuSh, line 18: Could you sort commands in alphabetical order to ease maintenance burden 🙏

PR #899, opened 2026-03-28
feat(codex): add native hooks integration

sf-jin-ku:
## Summary

- add native Codex `PreToolUse` integration via `rtk hook codex`
- extend `rtk init --codex` to patch `.codex/config.toml` and `.codex/hooks.json`
- fall back to prompt-only Codex setup on Windows and honor `CODEX_HOME` for global installs

## Problem

`rtk init --codex` was prompt-only. It wrote `RTK.md` and patched `AGENTS.md`, but it did not install Codex lifecycle hooks, so Codex had no hard guardrail to route supported raw Bash commands through `rtk`.

### Hook runtime

- add `rtk hook codex`
- reuse the existing rewrite source of truth via `get_rewritten()`
- for supported raw Bash commands, return Codex-native `permissionDecision: "deny"` with the exact `rtk ...` replacement
- pass through unchanged for unsupported commands, excluded commands, heredocs, and already-`rtk` commands

### Codex install / uninstall / show

- `rtk init --codex` now patches:
  - `RTK.md`
  - `AGENTS.md`
  - `.codex/config.toml`
  - `.codex/hooks.json`
- ensure `features.codex_hooks = true`
- install one `PreToolUse` matcher group for `Bash`
- `rtk init --show --codex` now reports local/global Codex hook state
- `rtk init --codex --uninstall` now supports local uninstall
- leave `codex_hooks = true` in place on uninstall, but remove RTK-managed hook entries

### Compatibility / edge cases

- use `CODEX_HOME` for global Codex path resolution instead of hardcoding `~/.codex`
- on Windows, fall back to the old prompt-only Codex setup because upstream Codex disables lifecycle hooks there
- local `.codex` installs still depend on Codex project trust
- update Codex-facing docs and hook README/test layout to match the current hook structure under `src/hooks` and `hooks/codex`

## Test plan

- [x] `cargo test --all`
- [x] `cargo build && RTK=target/debug/rtk bash hooks/codex/test-rtk-rewrite.sh`

## Notes

Codex currently does not support transparent `updatedInput` rewrites for `PreToolUse`, so this integration uses deny-and-retry instead of the Claude/Cursor transparent rewrite flow.


sf-jin-ku:
@rolandkakonyi addressed in e633de6.

I changed the Codex install path so it no longer relies on `@RTK.md` resolution for either global or project-local Codex installs. `rtk init --codex` now writes RTK-managed guidance inline into `AGENTS.md`, migrates legacy `@RTK.md` installs forward, and `rtk verify` now reports Codex `AGENTS.md` health explicitly alongside the Codex hook/config status.

That means global `$CODEX_HOME/AGENTS.md` / `~/.codex/AGENTS.md` no longer depends on upstream `@file` include support.

callme-sy:
this looks good.

one thing i was wondering about is the rewrite-miss path.

from the PR it looks like commands that `rtk rewrite` can map get the deny-and-retry flow with the exact `rtk ...` replacement, and everything else passes through unchanged.

i’ve been testing a local codex hook prototype around the same idea, and i ran into a couple cases where `rtk rewrite` missed but direct `rtk <cmd>` still worked better. 

do you think there’s eventually room for a lighter fallback for some rewrite misses before leaving the original command unchanged?

KuSh:

Issue #844, opened 2026-03-26
RTK exhibits unusual behavior when using ls and wildcards.


anjiahao1:
```
ajh@Asuka:~$ ls /dev/ttyACM*
/dev/ttyACM0
ajh@Asuka:~$ rtk ls /dev/ttyACM*
(empty)
ajh@Asuka:~$
```

rtk can't use `*` to match ls output?

pszymkowiak:
> [w] **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 88% |

### Summary

RTK's ls filter or glob handling drops wildcard-expanded results; likely the ls filter strips output that doesn't match expected patterns. Related to #1162 (ls filter removes metadata) and #1105 (glob_matches fails on middle-wildcard patterns).

> This looks like a **simple fix**. Only repo maintainers can trigger auto-fix.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

Issue #839, opened 2026-03-26
Empirical benchmark: 5 repos, 2,100 measurements — how do actual savings compare to claims?


qte77:
# Empirical benchmark: 5 repos, 2,100 measurements — how do actual savings compare to claims?

Related: #590, #538, #545, #827

## Context

RTK is a genuinely useful tool — `ls`, `docker ps`, and verbose git output compress well, and the Rust implementation is fast. Impressive work for a small team shipping at this pace.

This issue isn't about whether RTK has value (it does), but about closing the gap between what the README promises and what users actually experience. We ran a benchmark to quantify that gap — and we'd like to help close it.

## Setup

RTK v0.33.1 tested across **5 repos** (83–74K files, Shell/Bash/Python), **9 categories**, **10 iterations**, **3 independent runs** (2,100 total measurements). Results are deterministic: 699/700 ops showed <10-byte variance per run, and cross-run results are byte-identical when repo state is unchanged.

## Results

| Category | N | Actual | Claimed | Notes |
|----------|--:|-------:|--------:|-------|
| `ls` | 100 | **72%** | 80% | Matches well, especially `ls -laR` on larger repos |
| `git-log` | 150 | **98%** | 80% | Truncation rather than compression — 544K→1.8K |
| `git-status` | 100 | **46%** | 80% | Verbose format compresses; `-s`/`--porcelain` pass through |
| `git-diff` | 100 | **20%** | 75% | Full diff compresses somewhat; `--stat` passes through |
| `docker` | 20 | **38%** | 80% | `docker ps` = 75%; `docker ps -a` = 0% |
| `tree` | 50 | **4%** | 80% | Minimal compression |
| `cat` | 50 | **0%** | 70% | Passes through unchanged |
| `grep` | 100 | **-0%** | 80% | Slightly larger output (RTK overhead) |
| `ruff` | 30 | **-1%** | 80% | Slightly larger output |

Also tested separately: `pytest --co -q` = **0%** (185,962B unchanged, claimed -90%).

## The gap in the aggregate claim

The README's savings table includes several categories that show no compression in practice. These account for over half the claimed total:

| Category | Claimed saved/session | Measured |
|----------|----------------------:|:--------:|
| cat/read (20x) | 28,000 | 0% |
| grep/rg (8x) | 12,800 | -0% |
| pytest (4x) | 7,200 | 0% |
| ruff (3x) | 2,400 | slightly negative |
[...]

PR #900, opened 2026-03-28
feat(gain): add --quality flag for filter quality analysis

tmchow:
## Summary

- Add `rtk gain --quality` / `rtk gain -Q` that analyzes tracking data for filter quality signals
- Surfaces retry detection, low-savings commands, parse failures, and net savings estimate
- Follows the existing `--failures` flag pattern, no new dependencies

## Why this matters

RTK has 12 open issues tagged `filter-quality` where filtering removes too much information, causing AI agents to retry commands and waste more tokens than RTK saves. The tracking database records every command execution but provides no quality analysis.

- [#831](https://github.com/rtk-ai/rtk/issues/831) - AI agents burn extra tokens on retry loops from over-filtering
- [#839](https://github.com/rtk-ai/rtk/issues/839) - Request for empirical benchmarks (2 thumbs up)
- [#846](https://github.com/rtk-ai/rtk/issues/846), [#844](https://github.com/rtk-ai/rtk/issues/844), [#838](https://github.com/rtk-ai/rtk/issues/838), [#803](https://github.com/rtk-ai/rtk/issues/803), [#759](https://github.com/rtk-ai/rtk/issues/759) - additional filter-quality issues

## What it does

`rtk gain --quality` reports four sections:

1. **Retry Detection** - Identical commands re-run within 60 seconds. Uses a LAG window function partitioned by full command string for O(n log n) performance.
2. **Low Savings** - Filters consistently delivering <30% savings, excluding proxy/passthrough commands that intentionally have 0%.
3. **Parse Failures** - Reuses existing `parse_failures` table to show filters that fell back to raw output.
4. **Net Savings** - Gross savings minus estimated retry overhead.

The 60-second retry window matches on the full command string, so `ls src/` and `ls docs/` are tracked separately. Only exact re-runs within 60s are flagged.

## Demo

CleanShot 2026-04-08 at 14 52 25@2x

## Changes

- `src/main.rs`: Add `quality: bool` field to `Gain` variant, wire early return to `gain::show_quality()`
- `src/tracking.rs`: Add `RetryRecord`/`LowSavingsRecord` structs, `get_retry_commands()`, `get_low_savings_commands()`, `get_gross_savings()` methods
- `src/gain.rs`: Add `show_quality()` function following the `show_failures()` pattern

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` (1138 passed, 0 failed)
- [x] Manual testing: `rtk gain --quality` against real tracking DB (151M tokens saved, 99% efficiency)

Relates to #831, #839

### DIFF
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -835,7 +835,7 @@ pub fn show_quality(tracker: &Tracker) -> Result<()> {
     );
     if gross > 0 {
         let efficiency = (net.max(0) as f64 / gross as f64) * 100.0;
-        println!("  Efficiency:        {:.0}%", efficiency);
+        println!("  Efficiency:        {:.1}%", efficiency);
     }
     println!();
 
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -725,3 +725,129 @@ fn show_failures(tracker: &Tracker) -> Result<()> {
 
     Ok(())
 }
+
+pub fn show_quality(tracker: &Tracker) -> Result<()> {
+    let retries = tracker
+        .get_retry_commands()
+        .context("Failed to load retry data")?;
+    let low_savings = tracker
+        .get_low_savings_commands()
+        .context("Failed to load low-savings data")?;
+    let pf_summary = tracker
+        .get_parse_failure_summary()
+        .context("Failed to load parse failure data")?;
+    let gross = tracker
+        .get_gross_savings()
+        .context("Failed to load gross savings")?;
+
+    println!("{}", styled("RTK Filter Quality Report", true));
+    println!("{}", "═".repeat(60));
+    println!();
+
+    if retries.is_empty() {
+        println!("{}", styled("Retry Detection", true));
+        println!("{}", "─".repeat(60));
+        println!("  No retries detected (commands re-run within 60s).");
+        println!();
+    } else {
+        println!(
+            "{}",
+            styled("Retry Detection (commands re-run within 60s)", true)
+        );
+        println!("{}", "─".repeat(60));
+        for r in &retries {
+            let rate = if r.total_runs > 0 {
+                (r.retry_count as f64 / r.total_runs as f64) * 100.0
+            } else {
+                0.0
+            };
+            let retry_word = if r.retry_count == 1 {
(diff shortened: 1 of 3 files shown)


aeppling:
Hello, thanks for contributing

Just tried this, the retry detection look broken:


  ls                   4638 retries / 4663 runs  (99.5% retry rate)


Please check how do you get retry because you logic with time looks sketchy for this detection.

tmchow:
@aeppling After latest fixes:

CleanShot 2026-04-08 at 14 52 25@2x

aeppling:
Hello @tmchow ,


Thank you for trying to identify retries like those and address this, sadly this is not solid enough to really detect retries caused by RTK.
This could provide wrong informations, not helping us identify real token save loss.

(eg: i have followed long session on my v1 benchmarking solution, when AI agent retries, they mostly retries with different args / flags to try retrieve the data they want another way, even sometimes trying bypassing rtk binary, so all those retries are not identified , plus, some commands may run multiples times and be the same, and not being a RTK retry but agent will, another case is also that command could fail for another reason)

We agree that those retries are costly and must be transparent for users, that is one of our current challenge to ensure high saving for all users, we want something more precise on this.

We are working on an internal benchmark that will run across hundreds or parallels session, where we can directly identify token saves and loss from every aspects (inputs tokens, cache, api calls, turns, ...)

We really appreciate your contribution and you trying to solve this problem; i'll keep this open but as "invalid" for now, so i can get back to you here and see how we could implement this here so users can troubleshoot this by themselves.

We will soonly be able to provide better analytics, with real data analysis on long sessions and complexe task.
This PR should be updated at one point to match behavior of this benchmark if possible.


I see your active on Discord, i can propose you to discuss about it here if you want to keep dig this problem with us.

Thanks for contributing !

tmchow:
@aeppling Fair call. You're right that a timing-window heuristic can't distinguish RTK-caused retries from agents choosing different flags or bypassing the binary, and false positives there are worse than no signal.

Keeping an eye on the benchmark work and ping me on Discord once the retry signal from that dataset is clearer, and if there's a real signal to surface from tracking data I can revisit this PR against it. Thanks for the thoughtful writeup.

PR #904, opened 2026-03-28
fix(refacto-p1): unified cmds execution flow  (+ rm dead code)

aeppling:
Part 2 of codebase refacto 

#### Summary

- Standardize exit code handling across all modules: Result<()> → Result<i32> with a single process::exit() in main.rs — modules
  never call process::exit() directly
- Standardize filter execution flow across commands
- Fix critical bug in runner.rs where rtk err and rtk test always exited 0 regardless of command failure
- Add exit_code_from_status() helper for passthrough commands and centralize signal handling (128+signal) in
  exit_code_from_output()

#### Test plan

- cargo fmt --all && cargo clippy --all-targets && cargo test — 1123 passed, 0 failed
- Zero std::process::exit() in src/cmds/ and src/core/ (verified via grep), exactly 1 in main.rs
- golangci-lint exit 1 exception preserved (lint issues → exit 0, matching old behavior)
- src/cmds/README.md updated with new Result<i32> contract, examples, and exit code tables
- Manual testing: rtk <command> ; echo $? for success (0) and failure (non-zero) paths

### DIFF
--- a/src/cmds/js/vitest_cmd.rs
+++ b/src/cmds/js/vitest_cmd.rs
@@ -262,7 +262,7 @@ fn run_vitest(args: &[String], verbose: u8) -> Result<i32> {
         }
     };
 
-    let exit_code = output.status.code().unwrap_or(1);
+    let exit_code = crate::core::utils::exit_code_from_output(&output, "vitest");
     if let Some(hint) = crate::core::tee::tee_and_hint(&combined, "vitest_run", exit_code) {
         println!("{}\n{}", filtered, hint);
     } else {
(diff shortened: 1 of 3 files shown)


src/cmds/js/vitest_cmd.rs
```diff
@@ -271,8 +271,10 @@ fn run_vitest(args: &[String], verbose: u8) -> Result<()> {
 
     timer.track("vitest run", "rtk vitest run", &combined, &filtered);
 
-    // Propagate original exit code
-    std::process::exit(exit_code)
+    if !output.status.success() {
+        return Ok(exit_code);
```
FlorianBruniaux, line 275: This `exit_code` was computed with `.code().unwrap_or(1)` at line 265, not with `exit_code_from_output()`. Signal kills on vitest return 1 instead of 128+signal, which breaks the zero-`.code().unwrap_or(1)` invariant the PR description claims.

One-line fix at line 265:
```rust
let exit_code = exit_code_from_output(&output, "vitest");
```

PR #910, opened 2026-03-29
fix(refacto-p2):  more standardize

aeppling:
## Summary

- constant extract of repetitive patterns
- more covering for run_filtered wrapper and sub wrappers for specifics cmds environments
- automod for cmds environments (auto discover)
- merge pattern & rules for routing

## Test plan
- all test pass
- all edge cases kept (exclusions not covered by global wrappers)


FlorianBruniaux (changes requested):
Three things to fix before merging.

**1. Silent JSON fallback in `container.rs` (run_kubectl_json)**
When `serde_json::from_str` fails, the closure returns `stdout.to_string()` with no warning. Users get raw kubectl JSON (potentially thousands of tokens) with no indication the filter failed. Add `eprintln!("[rtk] kubectl: JSON parse failed: {}", e)` in the `Err` branch.

**2. Savings inflation in `runner.rs` (filter_stdout_only path)**
When `filter_stdout_only` is true, `text_to_filter = &stdout` but `timer.track()` is called with `&raw` (stdout + stderr combined) as the input. Savings % gets inflated by stderr length. On noisy commands like `docker build`, stderr can be substantial. The fix: track `&stdout` as raw input when `filter_stdout_only` is true. This is pre-existing but it is now centralized in one place, so the right time to fix it is here.

**3. automod contributor contract change**
Every `.rs` file dropped into `src/cmds/system/` is now automatically a public module. The previous `pub mod foo;` declarations were a forcing function. A developer who adds a WIP file directly into the directory will silently expose it. Low risk, but worth a comment or a note in the README explaining that automod is in use.

The `run_aws_filtered` deduplication (180 lines to 60) and the `run_passthrough` centralization are both good. Depends on #904, merge that first.

PR #925, opened 2026-03-30
fix(pytest): handle --collect-only output correctly

tangi-vass:
## Summary
- RTK's pytest filter misclassified `--collect-only` output as "No tests collected" because the summary line says "N tests collected" not "N passed/failed"
- Detect `--collect-only`/`--co` flag: skip injecting `-q`/`--tb=short`, route to dedicated `filter_collect_only_output()` that preserves the test tree
- Fix misleading debug `eprintln` that hardcoded injected flags

## Context
Agents using RTK spent ~7 extra turns investigating phantom "test discovery is broken" before bypassing RTK with `rtk proxy`.

## Test plan
- [x] `test_filter_collect_only` — verifies tree preserved and count correct
- [x] `test_filter_collect_only_zero_items` — edge case with 0 items
- [x] All 11 pytest tests pass, 1130 total tests pass
- [x] `cargo fmt && cargo clippy --all-targets` clean

### DIFF
--- a/src/cmds/python/pytest_cmd.rs
+++ b/src/cmds/python/pytest_cmd.rs
@@ -25,15 +25,20 @@ pub fn run(args: &[String], verbose: u8) -> Result<()> {
         c
     };
 
-    // Force short traceback and quiet mode for compact output
-    let has_tb_flag = args.iter().any(|a| a.starts_with("--tb"));
-    let has_quiet_flag = args.iter().any(|a| a == "-q" || a == "--quiet");
+    // --collect-only shows test tree, not test results — skip -q/--tb injection
+    let collect_only = args.iter().any(|a| a == "--collect-only" || a == "--co");
 
-    if !has_tb_flag {
-        cmd.arg("--tb=short");
-    }
-    if !has_quiet_flag {
-        cmd.arg("-q");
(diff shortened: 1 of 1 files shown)


mattbuot:
Confirming this is still live on `develop` five months on, in case it reads as stale.

`src/cmds/python/pytest_cmd.rs` on `develop` still has no `--collect-only` / `--co` branch and still injects `--tb=short` and `-q` unconditionally (lines 35 and 38). Reproduced on rtk 0.45.0, macOS arm64, Homebrew:

```
$ uv run rtk proxy pytest --collect-only -q | tail -1
3971/4018 tests collected (47 deselected) in 9.77s

$ uv run rtk pytest --collect-only -q
Pytest: No tests collected
```

One thing worth folding in if this gets picked up: the real summary line carries a `collected / total` pair and a `(N deselected)` clause, so matching on a bare `N tests collected` will miss projects that deselect.

This is also the actual root cause behind #3565, which was filed against the `uv run` rewrite path instead — I have commented there pointing here.

The PR is `mergeable_state: dirty` and not rebaseable in place. @tangi-vass, are you still on this? If not I can rebase it onto current `develop` with your commit preserved.

PR #934, opened 2026-03-30
fix(refacto): wrappers for standardization, exit codes lexer tokenizer, constants, code clean

aeppling:
## Bug Fixes

  - cmds: unified command execution flow across all filter modules
  - cmds: standardized exit code propagation (single exit point)
  - runner: fixed token savings inflation when filtering stdout only
  - container: kubectl warns on JSON parse failure instead of silent fallback
  - registry: env prefix rewriting handles quoted values correctly (FOO="bar baz")
  - psql: no longer misinterprets error messages as table output

## Features

  - lexer: shell lexer for robust command tokenization (quotes, escapes, redirects, pipes, operators)
  - automod: automatic module discovery (no more manual mod.rs maintenance)
  - runner: RunOptions builder for filter execution (tee, stdout-only, early-exit, formatting)
  - rules: added wc rewrite rule

## Refactoring

  - aws/gh/container/go: deduplicated command handlers via shared run_filtered
  - main: simplified routing (-385/+246)
  - registry: consolidated head/tail rewriters into single function
  - registry: moved inline regex to module-level lazy_static
  - deps: removed unused thiserror dependency

 ## Documentation

  - technical: full rewrite pipeline walkthrough in TECHNICAL.md
  - maintainers: application guide (Ecosystem + Core roles)
  - cmds: updated filter development checklist

### DIFF
--- a/docs/maintainers/MAINTAINERS_APPLY.md
+++ b/docs/maintainers/MAINTAINERS_APPLY.md
@@ -22,7 +22,7 @@ To apply, you should have:
 
 ### ✍️ How to Apply
 
-1. Open a discussion in [rtk-ai/rtk Maintainers Applications · Discussions · GitHub]() titled **Maintainer Application: [Your GitHub Handle]**
+1. Open a discussion in [rtk-ai/rtk Maintainers Applications · Discussions · GitHub](https://github.com/rtk-ai/rtk/discussions/categories/maintainers-applications) titled **Maintainer Application: [Your GitHub Handle]**
 2. In your application, include:
    - The ecosystem(s) you're interested in
    - Your experience with those ecosystems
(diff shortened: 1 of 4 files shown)


aeppling:
@pszymkowiak  Thanks for the review

### Fixed

- P1:  .unwrap() in lexer.rs: All 9 production unwrap() calls replaced with safe patterns. Loop cases use while let Some(&nc) = chars.peek() with break. All 78 lexer tests pass.
- P2: strip_quotes dead code in lexer.rs: Removed function and its 4 tests. Verified zero production callers.
- P2: Empty link in MAINTAINERS_APPLY.md: Fixed with correct URL.
- P2: Registry $() test: Added tests for command substitution passthrough (classify, rewrite, split).
- P1: eprint! inconsistancies replaced with eprintln!

### Not fixed

- P1: PR description: I do not see any mention of this in the PR description
- P0: psql savings threshold (40%): The filter and its 40% threshold are pre-existing on master. This PR only migrated psql_cmd.rs to runner::run_filtered for exit code handling.
- P2: is_table_format false positive on |: is_table_format checks -+- and ---+--- separators, not |. The | splitting happens inside filter_table, which is only reached after is_table_format passes.

PR #940, opened 2026-03-31
feat(discover): treat yadm commands like git

### LINKED ISSUE #567: Handle yadm commands in the same way as git (yadm is a git alias)
We could easily save tokens for people who are using yadm (https://yadm.io), since the same handling as for the respective git commands would apply (i.e. `yadm add` = `git add` with a specific GIT_DIR).

niklasmarderx:
## Problem

yadm (`https://yadm.io`) is a dotfile manager that wraps git. Running
`yadm status`, `yadm diff`, etc. produces the same output as the
equivalent git commands, but rtk doesn't recognize them.

## Fix

Extended the git discovery rule to also match `yadm` prefixes:

- `rules.rs`: pattern now matches `^(?:git|yadm)\s+...`, added `"yadm"` to rewrite prefixes
- 3 new tests: classify yadm status, classify yadm diff, rewrite yadm status

After:
```
yadm status  →  rtk git status  (70% savings)
yadm diff    →  rtk git diff    (80% savings)
```

Closes #567

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -594,6 +594,40 @@ mod tests {
     }
 
     #[test]
+    fn test_classify_yadm_status() {
+        assert_eq!(
+            classify_command("yadm status"),
+            Classification::Supported {
+                rtk_equivalent: "rtk git",
+                category: "Git",
+                estimated_savings_pct: 70.0,
+                status: RtkStatus::Existing,
+            }
+        );
+    }
+
+    #[test]
+    fn test_classify_yadm_diff() {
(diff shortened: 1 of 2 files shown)


pszymkowiak (approved):
Thanks @niklasmarderx — clean addition. yadm is a thin git wrapper so routing through rtk git makes sense. The regex change is minimal and the 3 tests cover classify + rewrite. 

LGTM.

niklasmarderx:
Just a gentle ping — this has been approved for a while. Let me know if there's anything else needed before merge.

PR #943, opened 2026-03-31
ci: enforce test presence on new/modified filter modules

FlorianBruniaux:
## Summary

- Add `scripts/check-test-presence.sh`: fails CI if a new or modified `*_cmd.rs` file is missing `#[cfg(test)]`. Uses `--diff-filter=AM` to catch both additions and test deletions on existing files.
- New CI job (`check-test-presence`) runs with no dependencies — pure grep, no Rust toolchain, <10s, parallel to all other jobs.
- Add 17 unit tests to `wget_cmd.rs` (compact_url, format_size, parse_error, extract_filename, truncate_line)
- Add 12 unit tests to `env_cmd.rs` (mask_value, is_lang_var, is_cloud_var, is_tool_var, is_interesting_var, sensitive_patterns)
- Fixes enforcement gap: CONTRIBUTING.md required tests but CI did not check. Now 34/34 `*_cmd.rs` modules have `#[cfg(test)]`.

## Test plan
- [ ] `bash scripts/check-test-presence.sh origin/develop` → exit 0
- [ ] `bash scripts/check-test-presence.sh --self-test` → exit 0
- [ ] `cargo test wget` → 17 passed
- [ ] `cargo test env_cmd` → 12 passed
- [ ] `cargo test --all` → all green

### DIFF
--- a/scripts/check-test-presence.sh
+++ b/scripts/check-test-presence.sh
@@ -1,7 +1,15 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# check-test-presence.sh — CI guard: new/modified *_cmd.rs files must have #[cfg(test)]
+# check-test-presence.sh — CI guard: new functions in *_cmd.rs must have test_<fn> coverage
+#
+# For each *_cmd.rs changed in this PR, parses newly added function signatures from the diff
+# and verifies each non-trivial function has a corresponding fn test_<name>[_variant] in the file.
+#
+# A function is skipped if:
+#   - It starts with test_ or _ (already a test or private helper)
+#   - It's an entry point or trait impl (run, run_*, new, default, fmt, clone, ...)
+#   - Its definition only appears inside the #[cfg(test)] block (test-module helpers)
 #
 # Usage:
 #   bash scripts/check-test-presence.sh [BASE_BRANCH]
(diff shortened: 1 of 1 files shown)


aeppling:
Hey @FlorianBruniaux 

This script is not checking if new functions has tests, but only if _cmd files contain a test function tag, which is not accurate because one test will be enough for a whole file.

If we want this test, we need to parse new functions in _cmd files from diff, then check if associated test function name (will require clean function naming and test function name suffix)

Will not merge as it is, i let this open in case you want to address this

FlorianBruniaux:
Hey @aeppling — reworked the script to be function-granular.

**What changed:**

The script now parses newly added function signatures directly from the diff (via `git diff --unified=0`), then verifies each non-trivial function has a corresponding `fn test_<name>[_variant]` in the file.

**Skip rules** (no test required):
- Entry points: `run`, `run_*`, `main`
- Constructors/trait impls: `new`, `default`, `fmt`, `clone`, `from_*`, `into_*`, `hash`, `eq`, `drop`, ...
- Already tests: `test_*`
- Private helpers: `_*`
- Functions defined only inside `#[cfg(test)]` (e.g. `count_tokens` test helpers)

**Example output when a function is missing its test:**
```
check-test-presence: scanning 1 filter module(s) for untested functions...

  FAIL  src/cmds/cloud/wget_cmd.rs::detect_protocol()
        Expected: fn test_detect_protocol[_variant] in src/cmds/cloud/wget_cmd.rs

check-test-presence: FAILED — add tests before merging.
```

One test block covering 10 functions would no longer pass — each function needs its own named test.

PR #962, opened 2026-04-01
fix(ls): strip -1 flag to prevent empty output (#803)

### LINKED ISSUE #803: `rtk ls -1` returns empty output
I let Claude explain itself what went wrong...

  ## Bug                                                                                                                                                                    
                                                                                                                                                                          
  `rtk ls -1 <directory>` returns empty output even when the directory contains files.                                                                                      
                                                                                      
  ## Reproduction                                                                                                                                                           
                                                                                                                                                                          
  ```bash                                                                                                                                                                 
  # Standard ls works
  ls -1 .concept/plan/
  # Shows 11 files                                                                                                                                                          
                  
  # RTK ls without flags works                                                                                                                                              
  rtk ls .concept/plan/                                                                                                                                                   
  # Shows 11 files with sizes                                                                                                                                               
                                                                                                                                                                          
  # RTK ls -1 fails                                                                                                                                                       
  rtk ls -1 .concept/plan/                                                                                                                                                  
  # Returns empty         
  ```                                                                                                                                                                       
                                                                                                                                                                          
  ## Expected                                                                                                                                                             
             
  `rtk ls -1` should return one filename per line, same as `ls -1`.
                                                                                                                                                                            
  ## Impact
                                                                                                                                                                            
  When Claude Code hooks rewrite `ls -1` → `rtk ls -1`, the command silently returns no output, causing Claude to incorrectly report directories as empty.

ousamabenyounes:
## Summary

- `rtk ls -1 <dir>` returned empty output even when the directory had files
- Root cause: the `-1` flag was passed through to `ls -la -1`. On both Linux (GNU coreutils) and macOS (BSD ls), `-la -1` switches to single-column format (overriding `-l`). The `compact_ls` parser expects long-format output (≥9 fields per line) and silently skipped every line, yielding `(empty)`
- Fix: extract `strip_ls_format_chars()` and add `'1'` to the existing set of format-control chars (`l`, `a`, `h`) stripped before passing extras to the underlying `ls`. RTK ls output is already one-entry-per-line, so `-1` is a no-op from the user's perspective

## Test plan

- [x] New unit test `test_strip_ls_format_chars` exercises the production helper directly
- [x] Regression test `test_flag_minus_one_does_not_produce_empty_output` verifies long-format input never yields `(empty)`
- [x] Manual: `rtk ls -1 .` now shows files instead of `(empty)`

Closes #803

---

### DIFF
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -18,6 +18,21 @@ lazy_static! {
     .unwrap();
 }
 
+/// Strip format-control characters from a short-flag bundle before passing
+/// extras to the underlying `ls` command.
+///
+/// We always invoke `ls -la` internally for parsing, so `l`, `a`, and `h`
+/// (human-readable, already implied) are redundant. `1` (single-column) is
+/// also stripped: `ls -la -1` overrides `-l` on both Linux and macOS and
+/// returns single-column output that `compact_ls` cannot parse (issue #803).
+/// RTK's own output is already one-entry-per-line, so `-1` is a no-op here.
+fn strip_ls_format_chars(flag_chars: &str) -> String {
+    flag_chars
+        .chars()
+        .filter(|c| !matches!(c, 'l' | 'a' | 'h' | '1'))
+        .collect()
+}
+
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let show_all = args
         .iter()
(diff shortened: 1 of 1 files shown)


pszymkowiak (changes requested):
Thanks @ousamabenyounes — the `ls.rs` fix for #803 is correct. Stripping `-1` from the flag bundle is the right approach since RTK already outputs one-entry-per-line.

However, please split this PR into two:

1. **ls -1 fix** (just `ls.rs` changes) — ready to merge
2. **dangerous-actions-blocker.sh** — separate PR for review (new hook, different scope)

The hook script is an interesting idea but needs its own review cycle — it adds a new security layer with regex-based command matching that deserves careful analysis.

ousamabenyounes:
Friendly ping @pszymkowiak — the PR was force-pushed to only the `ls.rs` fix as you requested (the unrelated `dangerous-actions-blocker.sh` is gone, branch rebased on latest `develop`). Ready for a re-review when you have a moment.

PR #964, opened 2026-04-01
fix: not merging git show with contradictory flags

ericbsantana:
## Summary

- **Problem:** Commands such as `rtk git show --name-only HEAD | head -20` could fail with Git’s error: `fatal: options '--name-only', '--name-status', '--check', and '-s' cannot be used together`.
- **Cause:** The compact `git show` implementation runs several `git show` steps (summary line, stat, diff, etc.) and **appends** fixed flags like `--no-patch` and `--pretty=format:…` to the same argument list as the user’s options. Git treats `--name-only` / `--name-status` / `--check` / `-s` as incompatible with that combination, so the inner `git` invocation exits with the fatal above—not because of `head`, but because the merged command line is invalid.
- **Fix:** If the user already asked for `--name-only`, `--name-status`, `--check`, `--no-patch`, or `-s`, **skip** the multi-step compact flow and run a **single** `git show` with exactly the user’s arguments (passthrough). That matches plain `git show` behavior and avoids mixing those options with the extra flags the compact path adds.

## Test plan

- [ ] Manual: `rtk git show --name-only HEAD` — should list paths only, no fatal.
- [ ] Manual: `rtk git show --name-status HEAD` — same idea, no fatal.
- [ ] Regression: `rtk git show HEAD` — still uses the compact multi-step output when none of the passthrough flags are present.


em0t (commented):
I tested this against a minimal repo and could reproduce the current fatal on master for `rtk git show --name-only HEAD`, `--name-status`, and `--check`.

On this branch those cases match raw `git show` again, and plain `rtk git show HEAD` still keeps the compact path. No blockers from my side.

One non-blocking suggestion: adding an end-to-end regression around `run_show` would make this safer than helper-only coverage.

ericbsantana:
thanks @em0t 

unfortunately I think adding a regression e2e testing would require maybe a broader discussion on how this testing structure would be - not only specifically for git but also for other CLI apps

bump on this PR @pszymkowiak and if we could open a discussion for e2e testing maybe

PR #970, opened 2026-04-02
fix(hook): reduce hook spawn overhead

### LINKED ISSUE #968: rtk-rewrite.sh PreToolUse hook causes EAGAIN (posix_spawn resource exhaustion) under heavy tool use
## Bug

The `rtk-rewrite.sh` PreToolUse hook causes `EAGAIN: resource temporarily unavailable, posix_spawn '/bin/sh'` errors when Claude Code is making rapid Bash tool calls.

## Root cause

The hook runs on **every** `Bash` tool invocation and spawns multiple child processes:

```
/bin/sh → rtk-rewrite.sh → jq (×3) + rtk rewrite + grep + cut
```

Under heavy tool use (e.g., parallel agent calls, rapid-fire commands), this saturates the OS process spawn capacity momentarily, causing `posix_spawn` to fail with `EAGAIN`.

## Environment

- macOS 15 (Darwin 25.3.0), Apple Silicon
- Claude Code (CLI)
- rtk installed via cargo
- `ulimit -u` = 8000, `kern.maxproc` = 12000

## Reproduction

1. Configure the rtk-rewrite hook per the docs
2. Run a Claude Code session with heavy Bash tool use (e.g., launching background tasks, parallel searches, agent subprocesses)
3. Intermittent `EAGAIN: resource temporarily unavailable, posix_spawn '/bin/sh'` errors appear

## Suggestion

Consider one or more of:

1. **Cache the `rtk --version` check** — the version guard runs `rtk --version | grep | cut` on every invocation. This could be checked once and cached in a file.
2. **Combine the jq calls** — the script calls `jq` 3 separate times (parse input, build updated input, build output). These could be a single jq invocation.
3. **Use a long-running sidecar** instead of spawning fresh processes per hook invocation — e.g., a simple socket/pipe server that the hook script queries.
4. **Rate-limit or debounce** — if the same command pattern was just checked, skip the rewrite.

zerone0x:
## Summary
- cache the rtk version guard to avoid spawning multiple processes on every hook run
- parse version in bash to remove grep/cut/head subprocesses
- reduce jq invocations by building updated input directly from the original payload

Fixes #968

## Test plan

### DIFF
--- a/hooks/claude/rtk-rewrite.sh
+++ b/hooks/claude/rtk-rewrite.sh
@@ -78,28 +78,24 @@ case $EXIT_CODE in
     ;;
 esac
 
-UPDATED_INPUT=$(jq -c --arg cmd "$REWRITTEN" '.tool_input | .command = $cmd' <<<"$INPUT")
-
 if [ "$EXIT_CODE" -eq 3 ]; then
   # Ask: rewrite the command, omit permissionDecision so Claude Code prompts.
-  jq -n \
-    --argjson updated "$UPDATED_INPUT" \
-    '{
+  jq -c --arg cmd "$REWRITTEN" \
+    '.tool_input.command = $cmd | {
       "hookSpecificOutput": {
         "hookEventName": "PreToolUse",
-        "updatedInput": $updated
+        "updatedInput": .tool_input
       }
-    }'
+    }' <<<"$INPUT"
 else
   # Allow: rewrite the command and auto-allow.
-  jq -n \
-    --argjson updated "$UPDATED_INPUT" \
-    '{
+  jq -c --arg cmd "$REWRITTEN" \
+    '.tool_input.command = $cmd | {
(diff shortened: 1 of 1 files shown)


pszymkowiak (approved):
Thanks @zerone0x — solid optimization. The version cache, herestring instead of echo|jq, and single-pass jq are all good improvements for reducing hook overhead (#968).

Two notes (non-blocking):
1. The cache file (`~/.cache/rtk-hook-version-ok`) has no expiry — after an `rtk` upgrade, the old cache persists. Consider adding the version to the filename (e.g., `rtk-hook-version-ok-0.34`) or clearing it on `rtk init`.
2. This PR modifies `rtk-rewrite.sh` which is also modified by PR #981 (permission model). Whichever merges first will cause a conflict on the other — just a heads-up for rebase.

LGTM.

Issue #913, opened 2026-03-29
Persistent "No hook installed" warning on Windows after rtk init -g --opencode


amaldevice:
### Description

After running `rtk init -g --opencode` (and also `rtk init -g`), the command `rtk gain` still shows this warning:

> [warn] No hook installed — run `rtk init -g` for automatic token savings

This warning appears every time even though:
- The OpenCode plugin has been installed successfully (`rtk init --show` shows `[ok] OpenCode: plugin installed`)
- `rtk --version` works correctly (currently v0.34.1)
- `rtk gain` successfully shows token savings statistics
- Cursor hook is also installed and working

### Steps to Reproduce

1. Install RTK on Windows (`rtk-x86_64-pc-windows-msvc.zip`)
2. Add `C:\rtk` to PATH
3. Run `rtk init -g --opencode`
4. Restart OpenCode
5. Run `rtk gain`

→ Warning still appears.

### Expected Behavior

`rtk gain` should recognize that the OpenCode plugin is installed and not show the "No hook installed" warning (or show a more accurate message for OpenCode/Windows).

### Actual Behavior

The warning persists, which is confusing even though token savings are working and the plugin is active.

### Environment

- **OS**: Windows 11
- **RTK Version**: 0.34.1
- **Tool**: OpenCode (with custom plugin `rtk.ts`)
- **Other hooks**: Cursor hook also installed

### Additional Context

According to the troubleshooting docs, OpenCode uses a custom `tool.execute.before` plugin instead of native Unix-style hooks. The detection logic in `rtk gain` seems to only check for native hooks, causing this misleading warning on Windows.

It would be great if RTK could:
- Better detect OpenCode plugin installation on Windows, or
- Show a more specific message like "OpenCode plugin installed (native hook not available on Windows)" when the plugin is present.

<img width="668" height="535" alt="Image" src="https://github.com/user-attachments/assets/0f513041-b649-4ac4-a1b5-e65aa3e02208" />

Thanks!

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 88% |

### Summary

[...]

Issue #897, opened 2026-03-28
grep subprocesses leak memory catastrophically, causing macOS kernel panic


javiersantaella-coder:
Summary
                                                                                                                     
  RTK spawns grep subprocesses (via rtk grep rewrites) that accumulate memory without releasing it. On a 96GB Mac
  Studio M2 Ultra, 8 concurrent grep processes in RTK's coalition consumed a combined ~514GB (RAM + swap), exhausting
   all memory and triggering a macOS kernel panic.                   
                                                                                                                     
  Environment                                                        

  - RTK version: 0.29.0
  - macOS: 26.3.1 (25D2128), Apple Silicon (Mac15,14 / M2 Ultra)
  - RAM: 96GB                                                                                                        
  - Integration: Claude Code PreToolUse hook via rtk-rewrite.sh                                                      
                                                                                                                     
  What happened                                                                                                      
                                                                                                                     
  While Claude Code was running bash commands (all routed through the RTK hook), grep subprocesses spawned by RTK    
  accumulated memory progressively over the session. The system triggered 5 JetsamEvents between 09:48–10:03 UTC
  before crashing.                                                                                                   
                                                                     
  Kernel panic log:
  panic(cpu 3 caller 0xfffffe0038d680e4): userspace watchdog timeout:
  no successful checkins from WindowServer in 204 seconds            
                                                                                                                     
  WindowServer could not allocate memory and hung, causing a hard reboot.                                            
                                                                                                                     
  Data from JetsamEvent at crash time (coalition 707 = RTK)                                                          
                                                                                                                     
  ┌─────────┬───────────┬───────────────────────────┐                                                                
  │ Process │ Instances │ Total memory (incl. swap) │                
  ├─────────┼───────────┼───────────────────────────┤                                                                
  │ grep    │ 8         │ ~505 GB                   │                
  ├─────────┼───────────┼───────────────────────────┤
  │ rtk     │ 8         │ ~9 GB                     │                                                                
  ├─────────┼───────────┼───────────────────────────┤                                                                
  │ Total   │           │ ~514 GB                   │                                                                
  └─────────┴───────────┴───────────────────────────┘                                                                
                                                                     
  Worst individual grep process:                                                                                     
  name: grep
  pid:  97560                                                                                                        
  rpages: 12,126,704  (× 16384 bytes = ~198 GB)                      
  cpuTime: 1048.6s                                                                                                   
  physicalPages.internal: [2010, 12118749]                                                                           
[...]

Issue #892, opened 2026-03-28
Codex in worktrees giving "sed: RTK.md: No such file or directory"


tmchow:
I'm running codex in Conductor so in worktrees and i'm seeing error:

```
sed: RTK.md: No such file or directory
```

in every session since initializing for codex on rtk 0.34.0

aeppling:
https://github.com/rtk-ai/rtk/pull/978

aeppling:
Fixed in 0.35.0

PR #978, opened 2026-04-02
fix(main): kill zombie processes + path for rtk md + missing intergrations

aeppling:
https://github.com/rtk-ai/rtk/issues/892
https://github.com/rtk-ai/rtk/issues/897
https://github.com/rtk-ai/rtk/issues/913
https://github.com/rtk-ai/rtk/issues/989

## Test plan


pszymkowiak (changes requested):
Thanks @aeppling — good set of improvements. Build & tests pass (1260 passed, 3 ignored). Reviewed code and ran manual tests on the release binary.

**P0 — must fix:**

**1. `ChildGuard` calls `kill()` on already-reaped process**
`src/main.rs` — After the normal `child.0.wait()` succeeds, the `Drop` still calls `kill()` + `wait()` on the dead process. Harmless on Unix today but incorrect semantics + theoretical PID recycling race. Fix: disarm the guard after successful wait:

```rust
struct ChildGuard(Option<std::process::Child>);
impl Drop for ChildGuard {
    fn drop(&mut self) {
        if let Some(mut child) = self.0.take() {
            let _ = child.kill();
            let _ = child.wait();
        }
    }
}
// After wait() succeeds:
let status = child.0.take().unwrap().wait()?;
```

**2. `rtk read - -` silently produces empty output on second stdin read**
`src/main.rs` Read loop — The second `run_stdin()` call consumes EOF, returning empty output with exit 0. Should warn on stderr:

```rust
let mut stdin_seen = false;
for file in &files {
    if file == Path::new("-") {
        if stdin_seen {
            eprintln!("rtk: warning: stdin specified more than once");
        }
        stdin_seen = true;
        // ...
    }
}
```

**P1 — important:**

**3. `hook_check.rs` — `other_integration_installed` masks Claude Code missing-hook warning**
When `.claude/` exists but the hook is missing, the presence of Cursor/Codex/OpenCode suppresses the "No hook installed" warning. A Claude Code user with Cursor installed for another project will never see the alert. The early exit at line 29 (no `.claude` dir) already handles the "user doesn't use Claude Code" case.

**4. `docs/contributing/ARCHITECTURE.md` — broken relative links after move**
`hooks/README.md` (line 38) and `CONTRIBUTING.md` (line 1034) still reference paths relative to root, but the file is now at `docs/contributing/`. These should be `../../hooks/README.md` and `../../CONTRIBUTING.md`.

**5. `init.rs` — misleading confirmation message in global mode**
Line 1291 prints `@RTK.md reference added` but in global mode the actual reference is an absolute path. Should use the `rtk_md_ref` variable.

**6. No tests for multi-file read loop**
The main behavioral change (single file to multi-file) has no test coverage for: two valid files concatenated, mix of valid + nonexistent, stdin dedup.

**Notes:**
- `ROADMAP.md` deletion: fine, it was obsolete (15 lines, all items completed)
- `filter-workflow.md` deletion: this was the only mermaid diagram of the TOML pipeline — intentional?
- `print!` vs `println!` change: verified correct with `xxd`, no trailing newline regression

Fix P0s and it's ready.

pszymkowiak (approved):

PR #986, opened 2026-04-03
chore(deps): update safe Rust crates

Coldaine:
Updates the lockfile for the smallest low-risk dependency set found in the crate scan:\n\n- clap 4.5.60 -> 4.6.0\n- tempfile 3.26.0 -> 3.27.0\n- which 8.0.1 -> 8.0.2\n\nThe update also pulled in small transitive lockfile refreshes (for example anstream/anstyle-parse) required by clap 4.6.0. No source changes were needed, and cargo test --locked passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,9 +40,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


pszymkowiak (approved):
Thanks @Coldaine — lockfile-only update, all within existing Cargo.toml semver ranges.

Verified: clap 4.6.0, tempfile 3.27.0, which 8.0.2. The anstream/anstyle-parse major bumps are transitive deps internal to clap, no impact on RTK APIs. The which simplification (drops rustix/winsafe, adds libc) is a net improvement.

LGTM.

PR #548, opened 2026-03-12
feat: add Bun and Deno runtime support

bz00qa:
## Summary

Add token-optimized filtering for Bun and Deno CLI commands, extending RTK's JavaScript/TypeScript runtime coverage.

### Bun commands
- `rtk bun install/add/remove` — strip progress bars, version headers (~80% reduction)
- `rtk bun test` — failures only via runner (~90% reduction)
- `rtk bun build` — errors only via runner
- `rtk bun run` — passthrough (script output unpredictable)
- `rtk bun pm ls` — JSON parsing with sorted compact dependency tree
- `rtk bunx <tool>` — smart routing: `tsc` → tsc filter, `eslint` → lint filter, others → error filter

### Deno commands
- `rtk deno test` — failures only via runner (~90% reduction)
- `rtk deno lint/check` — strip download lines, tee recovery on failure
- `rtk deno run/task` — passthrough with download line stripping
- `rtk deno compile` — errors only via runner
- `rtk deno install` — passthrough

### Hook integration
- Added `bun`, `bunx`, `deno` patterns to discover registry (PATTERNS + RULES)
- Commands auto-rewritten by `rtk rewrite` hook
- Added to `is_operational_command` security whitelist

### Files changed
- **New:** `src/bun_cmd.rs` (261 lines), `src/deno_cmd.rs` (254 lines)
- **Modified:** `src/main.rs` (mod declarations, Commands enum, subcommand enums, dispatch logic)
- **Modified:** `src/discover/rules.rs` (3 patterns + 3 rules)

## Test plan
- [x] 7 bun_cmd unit tests (install filtering, JSON parsing, text truncation, error preservation)
- [x] 4 deno_cmd unit tests (download stripping, empty output, error preservation)
- [x] All 898 existing tests pass (1 pre-existing binlog CRLF failure on Windows unrelated)
- [x] `cargo clippy` — no new warnings
- [x] `cargo fmt` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #197, opened 2026-02-18
Handle supported commands through the bunx / bun x / pnpx / pnpm dlx / npx


art-shen:
It's relatively common for AI models to run some linting or formatting using tools that aren't available in the current project directly through commands like bunx, npx, etc.

I think it would be logical to support the commands that are already supported by RTK when they are called through these package managers. 

I guess something like that could also exist for Python or Rust packages as well. 

E.g.,:
```
TOP UNHANDLED COMMANDS -- open an issue?
----------------------------------------------------
bunx tsc                   107    bunx tsc --noEmit --strict scripts/lib..
```

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 91% |

### Summary

Feature request to recognize commands run via package-manager proxies (bunx, npx, pnpm dlx, etc.) and apply existing RTK command handling — PR #1979 is in progress but appears to address only git subcommands, not the broader proxy pattern.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

KuSh:
This looks resolved on the current code (PR #988 (bunx) plus npx/pnpm dlx rewrite rules on develop), but it could not be fully reproduced here. Marking it resolved-pending-close: it will be closed in a while unless someone reports it still happening on the latest version.

KuSh:
This has carried the resolved-pending-close label with no new report of it happening, so closing it. Commands run through package-manager proxies now reach the matching RTK filter: `npx`, `pnpx` and `pnpm dlx` are rewritten to the underlying filter (`npx tsc --noEmit` -> `rtk tsc --noEmit`, `pnpm dlx eslint .` -> `rtk lint .`), and `bunx`/`bun x` route tsc and eslint to their filters via #988, which shipped in v0.48.0. Your original example `bunx tsc --noEmit ...` is handled. `yarn dlx` and the Python side (`uvx`) are still not covered, so please open a separate issue if you need those. If a proxy form still passes through unfiltered on the latest version, comment here and we'll reopen.

PR #988, opened 2026-04-03
feat(bun,deno): add Bun and Deno runtime support

geofflittle:
## Summary

Token-optimized filtering for **Bun** and **Deno** CLI commands.

- **Bun**: `install`/`add`/`remove` (strip the version header and resolution lines, ~70% measured), `test` (failures only, ~90%), `pm ls` (compact dependency list, ~70%), `run`, `build` and `pm <non-ls>` (passthrough), `bunx` (routes `tsc` and `eslint` to their filters, everything else through the npm line filter)
- **Deno**: `test` (failures only, ~90%), `lint` (~40%) and `check` (~50%) (strip ANSI, download and blank lines + tee recovery), `run`/`task`/`install`/`compile` (passthrough)

Based on #548 by @bz00qa and carried forward by @geofflittle. Supersedes #548 and #1374.

### Why some subcommands are passthrough

`bun build` writes its bundle to **stdout** when no output flag is given, and its emitted-file summary is the point of the run when one is. `deno compile` is the same: the artifact path and, on failure, the type-check diagnostics are the payload. An errors-only filter kept neither, so both run unfiltered and are declared 0% in the discover rules rather than claiming savings they cannot deliver.

Watch mode (`--watch`) is passthrough on every filtered entry point: the filtered runners buffer the whole stream until the child exits, and a watched child never does — so a filtered watch run printed nothing at all and lost the buffer on Ctrl-C. A deno reporter the user named (`--reporter=junit`, `--junit-path`) is likewise left alone, since it writes a machine-readable report to stdout.

### Savings figures

Every percentage above is measured against a captured fixture of real `bun 1.2.20` / `deno 2.9.6` output, and asserted by that module's own test — not against a fixture shaped to the filter. `deno lint` and `check` keep every diagnostic; their reduction is ANSI and blank lines, which is why the numbers are lower than a summarizing filter's.

### Known limitation

`bunx <tool>` forwards the caller's stdin only when it is a pipe, and `rtk npx` does not forward it at all. That is #2431, which covers every filtered entry point rather than this one; the `bunx` half arrives with this PR and should be fixed there.

## Test plan

- [x] `cargo fmt --all --check` clean
- [x] `cargo clippy --all-targets` — 0 warnings
- [x] `cargo test --all` — 3058 tests pass
- [x] Savings measured against captured fixtures, asserted per filter
- [x] Exercised against real `bun 1.2.20` and `deno 2.9.6`: pass/fail/timeout/module-load-error runs, type-check failures, leak failures, watch mode, `pm ls` nested trees, piped stdin, exit-code propagation

## Maintainer takeover

@geofflittle has stepped back from this PR. I have taken ownership: rebased onto `develop` and pushed fixes for five rounds of review findings on top of his work, which is preserved in the history.

**This needs review by another maintainer** — I authored the remediation commits, so I should not be the one approving them.


geofflittle:
I had to remove the co-authored-by @bz00qa trailers to get the CLA check to pass (they still have credit in the PR descr)

geofflittle:
Hi @aeppling — picking up #548 since it went quiet. Rebased, fixed 4 routing bugs (e.g. bun remove ran bun install), aligned with conventions, added token savings tests. Mind taking a look?

pszymkowiak:
Thanks for picking up #548 and doing the cleanup — this is a solid PR. RTK has **no bun/deno support at all** today, so it fills a real gap.

It's been open a while and `develop` has moved on quite a bit. Could you rebase onto `develop`? Two things to handle along with it:

1. **`BunCommands` and `DenoCommands` need `#[derive(Debug, Subcommand)]`.** `develop`'s `Commands` enum now requires `Debug` on every sub-enum — without it the build fails with `error[E0277]: BunCommands doesn't implement Debug`. (Same regression #752 hit.)

2. **Please drop the `CHANGELOG.md` change** — release-please manages the changelog automatically; manual entries cause merge conflicts. (The `README.md` Runtimes section is great, keep that.)

Once rebased with CI green, this looks ready to merge. Thanks again!

gandazgul:
@geofflittle I would love this, I use deno a lot and was looking to do this exact same thing. Let me know if I can help.

geofflittle:
@pszymkowiak
Ready!

pszymkowiak:
@geofflittle — first off, apologies for leaving this hanging. You addressed both points I raised (the `Debug` derives on `BunCommands`/`DenoCommands` and dropping the `CHANGELOG.md` change) and rebased back in June, and I never circled back to move it forward. That's on me. 🙏

The PR is in good shape and I'd like to get it in. Two small things:

1. `develop` has moved again since your last rebase, so GitHub now shows conflicts — could you do one more rebase onto `develop`?
2. One gap against #197: the space form `bun x <tool>` (e.g. `bun x tsc`) isn't routed to the tool filter the way `bunx tsc` is — the `Bun` rule pattern excludes `x`, so it falls through to a plain passthrough. If it's a quick add, great; otherwise it can be a follow-up.

Once it's rebased I'll review and get it merged. Thanks for your patience, and sorry again for the silence.

geofflittle:
@pszymkowiak ready x2!

pszymkowiak:
Tested this extensively with real `bun`/`deno` binaries (installed both to validate end-to-end, not just unit tests). Found a real bug in the `bun test` path worth fixing before merge.

**`is_bun` branch of `extract_test_summary` (core/runner.rs) loses failure detail and duplicates source context on multi-failure runs**

Ran `bun test` on a real 10-test suite (4 failing). The `SUMMARY:` section fills with duplicated source-code context (`2 | test("t1 passes", ...)`, `3 | test("t2 fails", ...)`, repeated and growing before each subsequent failure), while the actual diagnostic info — `error: expect(received).toBe(expected)`, `Expected: X`, `Received: Y` — is dropped entirely. Measured with a real BPE tokenizer (tiktoken): only ~38% savings on this case, vs 77%+ for the equivalent `deno test` case, and the output is *less* useful to an agent, not more — the one thing needed to fix a failing test (expected vs received) never survives the filter.

Root cause:

```rust
if is_bun {
    let trimmed = line.trim_start();
    if line.contains(" pass") || line.contains(" fail") || trimmed.starts_with("Ran ") {
        result.push(line.to_string());
    }
    if line.contains('✗') || line.contains("(fail)") {
        failures.push(line.to_string());
    }
}
```

`line.contains(" fail")`/`" pass"` is an unanchored substring match — it also matches bun's own echoed source lines whenever a test's *name* contains "passes"/"fails" (e.g. `test("t2 fails", ...)`), which is exactly what bun prints as growing context before each subsequent failure. That's why `SUMMARY:` fills with duplicate snippets as failure count grows.

Separately, `is_bun` has no equivalent to `is_cargo`'s/`is_deno`'s `in_failure && line.starts_with("    ")` state machine that captures the indented detail lines after a failure marker — so `Expected:`/`Received:` never make it into `failures`/`failure_lines`.

`is_deno` doesn't share this bug — verified clean output on the equivalent 5-test/2-failure case (exact failure list, correct `3 passed | 2 failed`, no duplication), because it reuses the same robust state-machine approach as `is_cargo`.

Suggested fix: give `is_bun` the same failure-marker + indented-detail capture used by `is_cargo`/`is_deno`, anchored on bun's actual output shape (`(fail) <name>` as the marker, then capture the `error:`/`Expected:`/`Received:` lines that follow) instead of the loose substring check.

Everything else I tested (install/add/remove, pm ls with JSON+text fallback, build success/failure, run passthrough, bunx/`bun x` routing to tsc/eslint, deno lint/check/test/run/compile/install) worked correctly against real binaries — this is the one thing I'd fix before merge, since test-failure output is the most common thing an agent needs to act on correctly.

geofflittle:
Fixed in 92aa5dd. I used the same methodology as your review (real `bun`/`deno` binaries, real failing suites, tiktoken measurements) for everything below.

**The `bun test` fix.** Both parts of the root cause are addressed: summary lines are now anchored on bun's exact count shape (` 4 fail`), so echoed source context can't match even when a test name contains "fails", and there's now a state machine that captures the `error:` / `Expected:` / `Received:` block per failure. One correction from fixturing: bun prints the diagnostic **before** the `(fail)` marker, not after it, so the capture anchors on `error:` and closes at `(fail)`. Reproduced your 10-test/4-fail case with real bun 1.3.6 output (committed as a fixture, with a byte-exact golden test): tiktoken savings went from ~38% to **74.9%**, and Expected/Received now survives for every failure.

Running the same end-to-end approach over the rest of the branch found four more real issues. All fixed before merge:

1. **`rtk bun add "lodash@^4.17.21"` failed before bun even ran.** An rtk-side character whitelist rejected `^`, `~`, `:`, `#`, so semver ranges, `github:`/`npm:` specifiers, and `workspace:*` were all broken. Since args go to bun as an argv vector (never through a shell), the validation had no security value, so it's gone. A test now locks verbatim spec passthrough.

2. **`deno test` had the same bug class as `bun test`.** The parser matched deno's old cargo-shaped output (`test result:`, `failures:`), which deno 2.x no longer emits, so the summary fell back to a raw tail and the assertion diffs were dropped entirely. Rewritten against real deno 2.9.1 output (` FAILURES ` section, `FAILED | 9 passed | 4 failed` footer), with the legacy anchors kept for older denos. Measured on a 13-test/4-fail suite: **90.2%** tiktoken savings, with the `[Diff] Actual / Expected` blocks intact. While in there I replaced the per-ecosystem substring flags with a `TestEcosystem` enum picked once at the boundary ("cargo test" contains "go test" as a substring, so the flags could co-fire).

3. **`bun pm ls --json` is silently ignored by bun 1.x**, so the JSON path never fired and every real run hit a 500-char text truncation. There's now a tree parser for what bun actually prints, and the filter picks a parser based on the output shape rather than the flag we passed.

4. **`deno lint`/`deno check` bypassed the `never_worse` guard** by hand-rolling their print path. They now go through the shared core runner like the other filters.

Everything is backed by committed real-output fixtures and regression tests, and CI is green at 2209a04. Three smaller things I left for a follow-up PR, per your earlier note: bunx/npx tool-dispatch parity (bunx currently routes only tsc/eslint to dedicated filters), compacting deno lint's verbose per-diagnostic frames, and `deno install` being classified as filtered in discover when it's actually passthrough.

pszymkowiak (approved):
Thanks, tested this again after your fix — all good. `bun test`/`deno test` failures now show the actual error details, no more duplication, exit codes and savings all check out on my end too. LGTM 👍

pszymkowiak (approved):
Re-confirming my approval after a full end-to-end verification on the current head (`2209a04`).

**Verified locally:**
- `cargo build` ✅, `cargo test` → 43 tests pass (32 bun + 11 deno), all against real command-output fixtures.
- **Live e2e on real runtimes** (installed bun 1.3.14 and deno 2.9.3, each with a one-pass/one-fail suite):
  - `rtk bun test`: surfaced the failing test + assertion detail (`Expected: 5 / Received: 4`), correct `1 pass / 1 fail` summary, propagated exit code 1. ~62% token savings.
  - `rtk deno test`: surfaced the failing test + diff (`Actual 4 / Expected 5`), correct summary, exit code 1, tee log for full output. ~76% token savings.

Covers both Bun and Deno, real fixtures, exit-code propagation is correct, and it shows assertion detail (nice for agent debugging). LGTM.

@TaKO8Ki could you give this a maintainer pass when you have a moment? It supersedes the narrower Bun-only #1374 and #548. Thanks!

pszymkowiak:
Thanks for this — the bun/deno filters look solid and I verified them end-to-end on macOS with real `bun`/`deno` (test summaries + failure detail, exit codes propagated, `bun pm ls` compression, argv-vector safety). The blocker right now is that the branch has drifted: it's ~108 commits behind `develop` and currently **CONFLICTING**.

I reproduced the rebase locally to de-risk it — it's small, only **2 things** to resolve:

1. **`src/main.rs`** — two purely *additive* conflicts. `develop` added `enum SbtCommands` right where this PR adds `enum BunCommands`; keep **both** (each with its own `Other(...)` variant), and keep the PR's `BunPmCommands` / `DenoCommands` / `run_bunx_tool` block.

2. **`src/discover/rules.rs`** — the non-obvious one: `develop` added a `pipeline_final_safe: bool` field to `RtkRule` (via #3128 pipe-rewrite-safety). The three new rules need it:
   ```rust
   pipeline_final_safe: false,   // matches cargo/npm/pnpm/jest/vitest/go
   ```
   The compiler flags this as `E0063 missing field` — it's what a mechanical rebase misses.

Everything else (README, `registry.rs`, `core/runner.rs`, `cmds/rust/runner.rs`) auto-merges cleanly. After those two fixes: `cargo build` + `clippy --all-targets` clean, full suite green, and `cat x | bun test` correctly falls through to raw (the `pipeline_final_safe: false` integrates with the new pipe-safety).

A `git merge --squash origin/develop`-style resolution is much easier here than replaying all 19 commits. Could you rebase on `develop` when you get a chance? Happy to help if anything's unclear.

zacgoodwin:
Looking forward to this!

pszymkowiak:
This is approved on my end, but `develop` has moved and GitHub now shows conflicts — could you rebase/resync when you get a chance so we can merge? Thanks for your patience.

geofflittle:
@pszymkowiak ready x3!

src/core/runner.rs
```diff
@@ -284,3 +287,586 @@ pub fn run_streamed(
...
+            output.push_str(&format!("  {}\n", f.trim()));
+        }
+        if failure_lines.len() > MAX_RUNNER_LINES {
+            output.push_str(&format!(
+                "  ... +{} more\n",
+                failure_lines.len() - MAX_RUNNER_LINES
+            ));
+        }
+        output.push('\n');
+    }
```
KuSh: **Blocking.** Bun diagnostics are dropped entirely when there's no `(fail)` marker.

When bun can't load a test file it prints `error: Cannot find module …` followed by `0 pass` / `0 fail` / `Ran 0 tests` — and **no** `(fail)` line. So `failures` stays empty, this whole block is skipped, and `failure_lines` (which for bun is where *every* diagnostic goes) is thrown away. Actual output:

```
SUMMARY:
 0 pass
 0 fail
 Ran 0 tests across 1 files…
```

rtk exits 1 with no indication of why. The `OUTPUT (last 5 lines)` fallback further down doesn't rescue it either, because the count lines already made `result` non-empty — cargo and deno do fall back here, bun doesn't.

Fix: emit `failure_lines` when `failures` is empty too, or push module-level `error:` lines into `failures` so the block is entered.

src/core/runner.rs
```diff
@@ -284,3 +287,586 @@ pub fn run_streamed(
...
+                //   (fail) t2 fails [1.86ms]
+                if trimmed.starts_with("(fail)") || line.contains('✗') {
+                    failures.push(line.to_string());
+                    in_failure = false;
+                } else if trimmed.starts_with("error:") {
+                    in_failure = true;
+                    failure_lines.push(line.to_string());
+                } else if in_failure && !trimmed.is_empty() && !trimmed.starts_with("at ") {
+                    failure_lines.push(line.to_string());
+                }
```
KuSh: This `error:` anchor isn't restricted to bun's own diagnostics. Any line a *test* writes to stdout starting with `error:` flips `in_failure` on, and it only clears on the next `(fail)` — so echoed source context and `(pass)` lines get swept into the failure block. A fully passing run that logs `error: could not connect to db (logged by test)` produces:

```
[FAIL] FAILURES:
  (pass) t1 logs [1ms]
  1 | test("t2", () => { … })
```

Same leak class as the test-name one you already fixed, just reached through test stdout instead. Deno guards its equivalent with `trimmed != "error: Test failed"` — bun needs a comparable anchor, e.g. only accept `error:` when the next line is an `Expected:`/`Received:`/`at ` frame, or only while a test-file block is open.

src/core/runner.rs
```diff
@@ -284,3 +287,586 @@ pub fn run_streamed(
...
+                    } else if !trimmed.is_empty()
+                        && trimmed != "^"
+                        && !trimmed.starts_with("throw ")
+                    {
+                        // "throw new AssertionError(...)" is deno-std's echoed
+                        // internal throw site; the failing test's location is
+                        // already in the FAILURES marker.
+                        failure_lines.push(line.to_string());
+                    }
+                }
```
KuSh: `in_failure` is only cleared by an `at ` stack frame, but deno's leak and sanitizer failures carry no frames — `error: Leaks detected:` opens the capture and nothing ever closes it, so the rest of the run gets appended. Verified output duplicates the failure entry and then appends `FAILURES`, `FAILED | 1 passed | 1 failed (10ms)` and `error: Test failed` into the failure block.

Cheapest fix: also clear `in_failure` on the section boundaries you already detect a few lines up (`FAILURES`, `FAILED |`, `ok |`).

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,484 @@
...
+use anyhow::{Context, Result};
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::ffi::OsString;
+
+/// JSON structure for `bun pm ls --json` output.
+#[derive(Debug, Deserialize)]
+struct BunPmPackage {
+    version: Option<String>,
+}
```
KuSh, line 19: `BunPmPackage` has a single `Option<String>` field and serde ignores unknown keys, so `HashMap<String, BunPmPackage>` (line 73) deserializes *any* object-of-objects — including a grouped shape:

```json
{"dependencies": {"express": {…}}, "devDependencies": {"vitest": {…}}}
```
→ `2 deps\ndependencies\ndevDependencies`

Group names reported as packages. And since `filter_bun_pm_ls` tries JSON first, any bun version that emits grouped JSON produces a confidently wrong dependency list instead of falling through to the tree parser you added for exactly this reason.

Require `version` to be present (or reject values whose keys don't look like package metadata) before accepting the JSON path.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,484 @@
...
+        &format!("bun_{}", subcmd),
+        exit_code,
+    );
+
+    timer.track(
+        &format!("bun {} {}", subcmd, args.join(" ")),
+        &format!("rtk bun {} {}", subcmd, args.join(" ")),
+        &combined,
+        &filtered,
+    );
```
KuSh: Two things here (both also apply to `run_pm_ls` below, lines 208-224):

1. `format!("{}{}", stdout, stderr)` concatenates the streams instead of interleaving them. `bun install` writes progress to stderr, so all of it lands after all of stdout and the error/progress ordering is scrambled. `core::runner::run_filtered` / `stream::run_streaming` preserves ordering — this is the alignment with the shared runner asked for in the earlier round, and `deno_cmd.rs` already does it.

2. `print_with_hint` returns the string that was *actually shown*, and it's discarded here — `timer.track` then records `&filtered`. When `never_worse` falls back to raw output, savings get over-reported. Track the returned value instead.

(`uv_cmd.rs:78` has the same shape, so #2 is a pre-existing convention rather than something this PR introduced — worth fixing in both, but I won't block on the `uv` side.)

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,484 @@
...
+        // Skip version headers like "bun install v1.1.0" / "bun add v1.1.0" / "bun remove v1.1.0"
+        if (trimmed.starts_with("bun install v")
+            || trimmed.starts_with("bun add v")
+            || trimmed.starts_with("bun remove v"))
+            && trimmed.split_whitespace().count() <= 4
+        {
+            continue;
+        }
+
+        result.push(trimmed);
```
KuSh: `result.push(trimmed)` drops the original indentation on every kept line, so bun's nested error blocks (`    at …`, `    hint: …`) come out fully left-aligned. On a multi-error `bun install` failure that makes it hard to tell which hint belongs to which package.

Push `line` and keep `trimmed` only for the emptiness/prefix checks above.

src/discover/rules.rs
```diff
@@ -655,6 +655,39 @@ pub const RULES: &[RtkRule] = &[
...
+    RtkRule {
+        pattern: r"^bunx\s+",
+        rtk_cmd: "rtk bunx",
+        rewrite_prefixes: &["bunx"],
+        category: "PackageManager",
+        savings_pct: 70.0,
+        ..RtkRule::DEFAULT
+    },
+    RtkRule {
+        pattern: r"^deno\s+(test|lint|check|run|task|compile|install)",
```
KuSh: Missing trailing `\b` — the bun rule at line 660 has one. As written this matches prefixes: `deno taskfoo` is rewritten to `rtk deno taskfoo` (confirmed with `rtk rewrite`). Harmless today because it lands in `DenoCommands::Other` passthrough, but it will mis-route any future subcommand that shares a prefix, and it inflates discover counts.

src/discover/rules.rs
```diff
@@ -655,6 +655,39 @@ pub const RULES: &[RtkRule] = &[
...
+        rtk_cmd: "rtk deno",
+        rewrite_prefixes: &["deno"],
+        category: "Build",
+        savings_pct: 75.0,
+        subcmd_savings: &[("test", 90.0), ("lint", 80.0)],
+        subcmd_status: &[
+            ("run", RtkStatus::Passthrough),
+            ("task", RtkStatus::Passthrough),
+        ],
```
KuSh, line 719: `DenoCommands::Install` goes to `run_passthrough` (0% reduction) but isn't listed here, so `rtk discover` credits every `deno install` in the user's history with this rule's 75%. Add `("install", RtkStatus::Passthrough)`.

Same problem for `bun pm <non-ls>` on the rule above — `BunPmCommands::Other` is passthrough too, but that case can't be expressed with the current single-capture pattern, so at minimum don't claim savings for it.

KuSh (changes requested):
Requesting changes. Thanks for the rework — everything from the previous round is addressed (`Debug` derives, the `bun x <tool>` space form, argv passthrough so `^`/`~`/`npm:`/`workspace:` specs survive, the `is_bun_count_line` state machine, current-deno `FAILURES` / `FAILED |` parsing, the `bun pm ls` tree parser, `deno lint`/`check` routed through `never_worse`, and `pipeline_final_safe` via `..RtkRule::DEFAULT`).

**Is this still needed? Yes.** `develop` has no bun/deno support whatsoever — no `bun_cmd.rs`/`deno_cmd.rs`, no `bun`/`bunx`/`deno` tokens in `src/main.rs`, no discover rules. This supersedes the narrower #548 and #1374, which I'm closing in favour of it.

Two things left before merge:

**1. Rebase onto `develop`** — the PR is currently `CONFLICTING`. It's mechanical: only `src/main.rs` conflicts, and only inside the `mod tests` block (develop's new grep-routing tests landed on the same lines as your bun/deno parse tests). Both sides are purely additive — keep both. I reproduced the resolution locally: `cargo build` clean, `cargo test --all` → 2918 passed, 0 failed.

**2. Eight inline findings below.** The first three are the ones that matter — they all hit `bun test` / `deno test` failure output, which is the highest-value path in this PR. The rest are small.

Unrelated heads-up so you don't chase it: `develop` itself doesn't compile right now (the phpt filter from #1503 uses `lazy_static!` without the crate, and `src/discover/rules.rs` is missing a `pipeline_final_safe`). CI will be red on any rebase until that lands separately — it's being fixed.

geofflittle:
@pszymkowiak @KuSh ready. Rebased onto current `develop`, all eight findings fixed.

The three test-output ones are verified end to end against real bun 1.3.6 and deno 2.9.6, with the captured output committed as fixtures. A module-load failure now shows its `error:` line instead of exiting 1 silently, a passing test that logs its own `error:` no longer manufactures a failure block, and deno leak failures close at the section boundary so the `FAILED |` footer survives.

One correction on the suggested bun anchor: bun 1.3.6 prints no `(pass)` markers at all, so keying on them cannot work. A block is now kept only with proof bun wrote it (an `Expected:` / `Received:` / `at ` frame, or an unhandled-error banner), which covers both the dropped-diagnostic and the swept-stdout case in one rule.

`uv_cmd.rs` left alone, since you flagged it as pre-existing.

CI is sitting at `action_required` and needs a maintainer to release the workflow run.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,476 @@
...
+    )
+}
+
+/// Run `bun build` with error-only filtering. Args are passed as a vector, never via a shell.
+pub fn run_build(args: &[String], verbose: u8) -> Result<i32> {
+    let mut cmd = resolved_command("bun");
+    cmd.arg("build").args(args);
+    let display = format!("bun build {}", args.join(" "));
+    crate::core::runner::run_err_cmd(cmd, display.trim_end(), verbose)
+}
```
KuSh, line 209: **Blocking.** `rtk bun build` destroys the bundle.

`run_err_cmd` installs `ErrorStreamFilter`, which drops every non-error line and, on a clean exit with nothing emitted, replaces the whole stream with `[ok] Command completed successfully (no errors)`. But `bun build <entry>` with no `--outdir` writes the bundled JS **to stdout**:

```console
$ bun build ./index.ts
// index.ts
var x = 1;
console.log(x);
export { x };

$ rtk bun build ./index.ts
```

`src/discover/rules.rs` rewrites `bun build …` → `rtk bun build …`, so this isn't opt-in: `bun build ./index.ts > bundle.js` now writes that literal `[ok]` line into `bundle.js`. Silent data loss in a file the user will not think to check.

`bun build` needs stdout passthrough, or at minimum a bypass when neither `--outdir` nor `--compile` is present (those are the forms that write to disk and produce a summary worth filtering).

Related: `README.md:235` documents this as `# Errors only`, so the docs currently describe the destructive behaviour as intended.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,476 @@
...
+    )
+}
+
+/// Run `bunx <tool>` with error-only filtering. Args are passed as a vector, never via a shell.
+pub fn run_bunx(args: &[String], verbose: u8) -> Result<i32> {
+    let mut cmd = resolved_command("bunx");
+    cmd.args(args);
+    let display = format!("bunx {}", args.join(" "));
+    crate::core::runner::run_err_cmd(cmd, display.trim_end(), verbose)
+}
```
KuSh: **Blocking, same root cause as `run_build`.** This applies the errors-only filter to *arbitrary* executables, and the new `^bunx\s+` rule rewrites every `bunx` invocation:

```console
$ bunx cowsay hello
 _______
< hello >
 -------
        \   ^__^

$ rtk bunx cowsay hello
```

Any bunx tool whose entire value is its stdout — a codegen script, a formatter printing to stdout, a CLI that reports results — returns nothing.

The npx path already solved this: unknown tools go to `npm_cmd::exec` (`filter_npm_output`, a light line filter that preserves content), deliberately **not** an errors-only filter. `bunx` should mirror that rather than `rtk err`.

src/main.rs
```diff
@@ -1323,6 +1342,120 @@ enum SbtCommands {
...
+
+/// Route `bunx <tool>` and `bun x <tool>` to the matching tool filter,
+/// falling back to the generic bunx runner for unrecognized tools.
+fn run_bunx_tool(args: &[String], verbose: u8) -> Result<i32> {
+    if args.is_empty() {
+        anyhow::bail!("bunx requires a command argument");
+    }
+    match args[0].as_str() {
+        "tsc" | "typescript" => tsc_cmd::run(&args[1..], verbose),
+        "eslint" => lint_cmd::run(&args[1..], verbose),
```
KuSh: `&args[1..]` drops `eslint`, but `lint_cmd::run` does its own tool detection — `detect_linter` treats `args[0]` as an explicit linter name unless it starts with `-` or contains `/` or `.`. So a bare directory becomes the "linter":

```console
$ rtk -v bunx eslint src
Running: src with structured output
npm error could not determine executable to run

$ rtk -v bunx eslint src/
Running: eslint with structured output      # works only because of the slash
```

`lint_cmd::strip_pm_prefix` already knows about `bunx`, so the fix is to keep the tool name:

```suggestion
        "eslint" => lint_cmd::run(args, verbose),
```

Heads-up on provenance: this is the same bug that already exists on `develop` for npx at `src/main.rs:2361`, which I filed as #3812 yesterday after @HenriqueSchroeder flagged the pattern. That one is latent maintenance; this copy is new and hook-rewritten, so it ships the bug to a second path. If you fix `:2361` in the same commit, #3812 closes with this PR.

src/main.rs
```diff
@@ -1323,6 +1342,120 @@ enum SbtCommands {
...
+}
+
+/// Route `bunx <tool>` and `bun x <tool>` to the matching tool filter,
+/// falling back to the generic bunx runner for unrecognized tools.
+fn run_bunx_tool(args: &[String], verbose: u8) -> Result<i32> {
+    if args.is_empty() {
+        anyhow::bail!("bunx requires a command argument");
+    }
+    match args[0].as_str() {
+        "tsc" | "typescript" => tsc_cmd::run(&args[1..], verbose),
```
KuSh: Both bunx tool arms re-execute through npx or pnpm, never through bun:

- `tsc_cmd::run` falls back to `npx tsc` when `tsc` isn't on `$PATH`
- `lint_cmd` uses `package_manager_exec`, whose `detect_package_manager` only knows `pnpm-lock.yaml` / `yarn.lock` / npm — there's no bun branch

```console
$ rtk bun x tsc --noEmit
Running: npx tsc --noEmit
```

In a bun-only project (`bun.lockb`, no npm/npx installed) a working `bunx tsc` becomes "npx not found" the moment the hook rewrites it. Teaching `detect_package_manager` about `bun.lockb`/`bun.lock` would fix both arms at once and benefits the existing npx routing too.

src/core/runner.rs
```diff
@@ -284,3 +287,693 @@ pub fn run_streamed(
...
+pub fn run_err_cmd(cmd: Command, display: &str, verbose: u8) -> Result<i32> {
+    if verbose > 0 {
+        eprintln!("Running: {}", display);
+    }
+    run_streamed(
+        cmd,
+        "err",
+        display,
+        Box::new(ErrorStreamFilter::new()),
+        RunOptions::with_tee("err"),
```
KuSh: `run_err_cmd` hardcodes `tool_name = "err"` and `run_test_cmd` hardcodes `"test"` (line 452), and `run()` builds `cmd_label = format!("{} {}", tool_name, args_display)` — so everything routed through them is recorded under a tool that isn't the one that ran. From a scratch tracking DB after using the new commands:

```
1. rtk test bun test
2. rtk err bunx cowsay hi
3. rtk bun install            <- correct, run_pkg passes "bun"
4. rtk err bun build ./i...
```

`rtk gain --history` then lists invocations the user never typed and that aren't valid rtk commands, and bun/deno adoption is invisible in analytics. `run_pkg` and `run_filtered_subcmd` already pass the real tool name, so this is an inconsistency inside the PR rather than a pre-existing convention.

The tee slug has the same problem — every ecosystem shares the `err` / `test` key, so recovery files collide across tools.

Threading the tool name through both helpers (as the other call sites already do) is the fix.

README.md
```diff
@@ -228,6 +228,17 @@ rtk bundle install              # Ruby gems (strip Using lines)
...
 
+### Runtimes
+```bash
+rtk bun install                  # Strip progress bars (~80% reduction)
+rtk bun test                     # Failures only (-90%)
+rtk bun build                    # Errors only
+rtk bunx tsc                     # Smart routing to tsc filter
+rtk deno test                    # Failures only (-90%)
+rtk deno lint                    # Strip download lines + tee recovery
+rtk deno check                   # Type check errors only
```
KuSh: Overstated. `filter_deno_output` only strips ANSI, `Download ` lines and blank lines — everything else is kept, including `Check file:///…` progress. "Type check errors only" promises something the filter doesn't do.

Line 238 (`rtk deno lint  # Strip download lines + tee recovery`) is accurate; this one should describe the same treatment. And see my note on line 235 — `rtk bun build  # Errors only` is accurate about the *current* code but that behaviour is itself the blocker above, so it'll need rewording once `bun build` stops eating stdout.

KuSh (changes requested):
Second round. First: all eight threads from the last review are resolved — I checked the code, not just the commit messages, and each one is properly fixed. The `pending`/`pending_proven` buffering is a better answer than what I suggested, the `pm\s+ls` alternative plus the `registry.rs` whitespace collapse solves the bun `pm` accounting I'd written off as inexpressible, and the rebase is clean (`MERGEABLE`, `cargo test --all` green on the branch).

Requesting changes again on a new set, found by exercising the branch against a real bun install rather than re-reading the diff. Six inline below.

**The two blockers are the same mistake in two places:** `run_build` and `run_bunx` both route through `run_err_cmd`, whose errors-only filter replaces a clean run's entire output with `[ok] Command completed successfully (no errors)`. That's right for a compiler and wrong for anything that emits real stdout — and both are hook-rewritten, so users hit them without opting in.

One finding has no inline anchor because the file isn't in this diff:

**`src/core/tracking.rs` — `categorize_command` was not extended.** The `"npm" | "npx" | "pnpm" | … => "js"` arm has no `bun` / `bunx` / `deno` case, so every new command falls into the `_ => "other"` telemetry bucket despite living in `src/cmds/js/`. One-line addition, but without it this whole feature is invisible in per-ecosystem reporting.

Things I checked and found correct, so you don't re-verify them: `--` survives `trailing_var_arg` (`rtk bun run dev -- --port 3000`), exit codes propagate on both `bun test` and `bun build`, the routing fixes in the PR description all reconstruct the right argv, `bun pm ls --json` really is ignored by bun 1.2.20 so the tree parser is the live path, and `TestEcosystem::detect` ordering genuinely fixes the old `"cargo test".contains("go test")` co-firing.

geofflittle:
@KuSh all seven fixed at `31194b6`.

`bun build` now passes through unless `--outdir`, `--outfile` or `--compile` sends the bundle to disk, and `bunx` uses the same light npm filter as the npx path. Verified against real bun: `rtk bun build ./index.ts` emits the bundle, `rtk bunx cowsay hello` emits the cow.

Two places I went further than asked. `detect_package_manager` alone does not fix both bunx arms, because `tsc_cmd` hardcoded its own npx fallback rather than going through `package_manager_exec`, so that arm is converted too. With npx off `$PATH` in a bun project, `rtk -v bunx tsc` now resolves through bunx instead of failing. I also took the invitation on the npx copy in `main.rs`, so both eslint call sites keep the tool name.

`rtk gain --history` lists `rtk bun test` and `rtk bunx cowsay hello` instead of `rtk err ...` and `rtk test ...`, and tee slugs are per tool.

CI is still sitting at `action_required`.

src/core/runner.rs
```diff
@@ -284,3 +287,710 @@ pub fn run_streamed(
...
+                } else if trimmed == "FAILURES" || line.starts_with("failures:") {
+                    in_failures_list = true;
+                    in_failure = false;
+                } else if in_failures_list && !trimmed.is_empty() {
+                    failures.push(line.to_string());
+                }
+                // Diagnostics live in the ERRORS section: an "error:" line, then
+                // [Diff] and +/- lines, then a stack that ends the block.
+                // "error: Test failed" is deno's generic footer, not a diagnostic.
+                if trimmed.starts_with("error:") && trimmed != "error: Test failed" {
```
KuSh: **Blocking.** Deno opens a diagnostic block on any `error:` line with no proof requirement — the guard bun received last round was never applied here.

Deno prints a test's own stdout inline between `------- output -------` and `----- output end -----`, so a **passing** test that logs `console.error("error: could not reach cache…")` produces:

```
[FAIL] ERRORS:
  error: could not reach cache, using fallback
  retrying in 1s
  ----- output end -----
  ok (5ms)

SUMMARY:
  ok | 1 passed | 0 failed (12ms)
```

Exit code is 0, the summary says `0 failed`, and the output still leads with `[FAIL] ERRORS:`. For an agent reading rtk's output — the entire point of the tool — that reads as a failed suite.

This is the most severe finding of the round because it's a false positive on a **green** run, not a degraded message on a red one. It's also the case the unified open/prove/close policy from the review body closes by construction: deno's proof column is empty, and it shouldn't be.

src/core/runner.rs
```diff
@@ -284,3 +287,710 @@ pub fn run_streamed(
...
+                    pending_proven = unhandled_section;
+                } else if is_count {
+                    unhandled_section = false;
+                } else if pending_open {
+                    if trimmed.starts_with("at ") {
+                        pending_proven = true;
+                    } else if trimmed.starts_with("Expected:") || trimmed.starts_with("Received:") {
+                        pending_proven = true;
+                        pending.push(line.to_string());
+                    } else if !trimmed.is_empty() && !trimmed.chars().all(|c| c == '-') {
```
KuSh: Bun's proof guard is narrower than it looks. `pending_proven` is set by *any* indented `at ` line following an `error:` line, so a passing test that logs a caught error's stack (`console.error(err.stack)`) proves the block — and since `(pass) …` isn't a `closes_block` token, it gets absorbed into it. On a `1 pass / 0 fail` run:

```
[FAIL] ERRORS:
  error: could not reach cache
  (pass) t1 works [1ms]

SUMMARY:
   1 pass
   0 fail
```

Same false-positive-on-green shape as the deno finding above. The guard narrowed the original leak without closing it, because "an `at ` frame appeared" is evidence that *a stack* was printed, not that *bun* printed it — bun's own frames sit under its diagnostic and are followed by a marker; a logged stack isn't.

Adding `(pass)` to `closes_block` fixes this instance. Requiring the block to terminate in a marker before it counts as proven fixes the class.

src/cmds/js/tsc_cmd.rs
```diff
@@ -79,20 +81,20 @@ fn clean_line(line: &str) -> Cow<'_, str> {
...
-    let mut cmd = if tsc_exists {
-        resolved_command("tsc")
-    } else {
-        let mut c = resolved_command("npx");
-        c.arg("tsc");
-        c
-    };
+    // Fall back through the project's own package manager rather than always
+    // npx: a bun-only project has no npx to fall back to.
+    let mut cmd = package_manager_exec("tsc");
```
KuSh: **This one is my fault** — my round-2 comment asked for it, and the suggestion was wrong. Two regressions from `resolved_command("npx")` → `package_manager_exec("tsc")`:

**1. Engine substitution.** The runner now comes from lockfile detection instead of from what the user typed. In a directory with `pnpm-lock.yaml`:

```console
$ rtk bunx tsc --noEmit
Running: pnpm tsc --noEmit
 ERR_PNPM_NO_PKG_MANIFEST  No package.json found … run pnpm install
```

Before this PR that ran `npx tsc`. The user asked for `bunx` and got `pnpm`. Same path via `rtk npx tsc`, and via `rtk bunx eslint` through `lint_cmd`.

**2. `--no-install`** — see my separate note on `src/core/utils.rs:372`.

RTK forwards what was typed rather than substituting an engine for it, and my suggestion violated that rule. The problem I was actually pointing at was narrower: `tsc_cmd` hardcoded `npx` *even when invoked through bunx*. So the runner should come from the invocation — `bunx tsc` → `bunx`, `npx tsc` → `npx` — falling back to `detect_package_manager()` only for a bare `rtk tsc`, where the user specified nothing. That still fixes the bun-only-project case I raised, without overriding an explicit choice.

src/core/utils.rs
```diff
@@ -358,6 +362,11 @@ pub fn package_manager_exec(tool: &str) -> Command {
...
                 c
             }
+            "bun" => {
+                let mut c = resolved_command("bunx");
+                c.arg(tool);
+                c
+            }
             _ => {
                 let mut c = resolved_command("npx");
                 c.arg("--no-install").arg("--").arg(tool);
```
KuSh: The npm arm adds `--no-install`, which `tsc_cmd` did not have before it started routing through here. `rtk tsc` in a project with no local `typescript` now fails where `npx tsc` would previously have fetched and run it.

Worth deciding deliberately rather than inheriting by accident: `--no-install` is right for this function's original callers (lint, playwright — where a silent network fetch is surprising), but it's a behaviour change for anything newly routed through it. Keeping it is a defensible call; just say so in the PR description, since it changes what `rtk tsc` does on a cold checkout.

src/core/runner.rs
```diff
@@ -284,3 +287,710 @@ pub fn run_streamed(
...
+                if trimmed.starts_with("error:") && trimmed != "error: Test failed" {
+                    in_failure = true;
+                    failure_lines.push(line.to_string());
+                } else if in_failure {
+                    if trimmed.starts_with("at ") {
+                        in_failure = false;
+                    } else if !trimmed.is_empty()
+                        && trimmed != "^"
+                        && !trimmed.starts_with("throw ")
+                    {
```
KuSh: `in_failure` is cleared only by an `at ` frame or a section boundary — but the leak and sanitizer failures the comment just above cites carry no frames. So with **two** of them, the second failure's `name => file:line:col` header and the `To get more details…` hint are absorbed into the first block. Verified: `t3 leaks => ./x_test.ts:6:6` appears twice, once in `FAILURES` and once inside the diagnostics.

Round 1's fix added the section boundaries, which covers one such failure followed by the footer. It doesn't cover two in a row, because nothing closes a block when the *next* block opens.

Bun already gets this right — `opens_block` is part of its `closes_block` set. That asymmetry is the review-body point in miniature.

src/discover/rules.rs
```diff
@@ -671,6 +671,45 @@ pub const RULES: &[RtkRule] = &[
         savings_pct: 75.0,
         ..RtkRule::DEFAULT
     },
+    // Bun/Deno
+    RtkRule {
+        pattern: r"^bun\s+(install|add|remove|test|build|run|pm\s+ls|pm|x)\b",
```
KuSh, line 676: `bun build` is credited the rule's default 75% and `RtkStatus::Existing`, but as of this PR `run_build` passes through unfiltered unless `--outdir` / `--outfile` / `--compile` is present (`bun_cmd.rs`, `build_writes_to_disk`). The common `bun build ./index.ts` form is 0% reduction, so `rtk discover` over-reports it.

This is the same accounting bug you fixed for `deno install` and `bun pm <non-ls>` in round 2. It reappeared because the round-2 passthrough fix *created a new passthrough form* without revisiting the table — which is why the review body asks for one pass over the whole table rather than per-entry patches.

`build` doesn't split on a flag the way `pm ls` did, so the honest options are dropping it from `subcmd_savings` to the passthrough default, or keeping the rule and not claiming the headline number for it.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,526 @@
...
+        return tree_result;
+    }
+    filter_bun_pm_ls_text(raw)
+}
+
+/// Text fallback for `bun pm ls`.
+pub fn filter_bun_pm_ls_text(raw: &str) -> String {
+    let lines: Vec<&str> = raw.lines().filter(|l| !l.trim().is_empty()).collect();
+
+    truncate(&join_or_ok(&lines), 500)
```
KuSh, line 143: The JSON and tree paths both `strip_ansi` first; this fallback doesn't — and it's the path error output takes. `bun pm ls` colorizes, so the fallback emits raw escape sequences and then `truncate(…, 500)` spends the budget on them. It can also cut mid-sequence, leaving a severed CSI in the output.

src/core/utils.rs
```diff
@@ -1387,4 +1407,15 @@ mod tests {
...
+    #[test]
+    fn test_detect_package_manager_recognizes_bun() {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let prev = std::env::current_dir().expect("cwd");
+        std::env::set_current_dir(dir.path()).expect("chdir");
+        std::fs::write("bun.lockb", "").expect("write lockfile");
+        let detected = detect_package_manager();
+        std::env::set_current_dir(prev).expect("restore cwd");
+        assert_eq!(detected, "bun");
+    }
```
KuSh, line 1662: This chdirs the **whole process** into a tempdir. It shares a test binary (and cargo's default thread pool) with `test_detect_package_manager_default` at `utils.rs:851`, which asserts the result is one of `pnpm`/`yarn`/`npm` — if that runs while the cwd is this tempdir, it observes `"bun"` and fails. The five `set_current_dir` tests in `src/hooks/init.rs` race the other direction.

I did not reproduce it across 8 full-suite runs, so this is a latent flake rather than a live failure — but cwd is process-global and the window is real. `#[serial]`, or a `detect_package_manager_in(path)` variant the test can call without chdir, closes it.

KuSh (changes requested):
Round 3. All fourteen threads from rounds 1 and 2 are resolved — I verified each fix in the code, and several are better than what I asked for: the `build_writes_to_disk` gate errs toward passthrough, which is the safe direction, and `exec_with` generalizes the npx filter instead of duplicating it. `cargo test --all` (2982 tests) and `cargo clippy --all-targets` are clean on the branch.

**First, a correction that's on me.** My round-2 comment on `src/main.rs:1453` suggested teaching `detect_package_manager` about `bun.lockb`. That suggestion was wrong and it introduced a regression — details inline at `tsc_cmd.rs:86`. RTK forwards what the user typed rather than substituting an engine for it; my suggestion violated that and I should have caught it before proposing it.

## The structural ask

Seven findings this round, but three of them are one problem, and the history is what makes the case:

| Fixed in | Found this round |
|---|---|
| Round 1 — bun's `error:` anchor swallowing test stdout | deno has the identical defect, with no proof guard at all |
| Round 1 — deno's block not closing on `at ` frames | still doesn't close for *consecutive* stack-free diagnostics |
| Round 2 — `deno install` credited savings it doesn't deliver | `bun build` now has the same accounting bug, introduced by the round-2 passthrough fix |

Three times a defect has been fixed on one runtime while its mirror stayed live — and once, a fix on one side created the bug on the other. That's not a bun problem or a deno problem. `extract_test_summary` has bun on a buffered `pending`/`pending_proven` model and deno on a flat `in_failure` flag: two state machines answering the same question — *which lines belong to a real failure rather than the test's own stdout*. So every fix has to be independently rediscovered on the far side, and one of them hasn't been.

Both arms already decompose into the same three predicates:

| | opens a block | proves it | closes it |
|---|---|---|---|
| **bun** | `error:` | `at `, `Expected:`/`Received:`, unhandled banner | count line, `(fail)`, `✗` |
| **deno** | `error:` (≠ `error: Test failed`) | — | `at `, `FAILURES`, `FAILED \|`, `ok \|`, `test result:` |

One buffered engine plus a per-ecosystem policy of those three predicates collapses findings 1, 2 and 5 below into a single change, and turns deno's empty proof column from something a reviewer has to notice into a cell you can't leave blank. Deliberate divergences then live in that table instead of in prose spread through a 100-line match arm.

Two smaller asks in the same spirit:

- **Paired fixtures.** Every test-output fixture gets a bun *and* a deno counterpart for the same scenario, so an unfixed mirror fails CI instead of waiting for round 4.
- **Audit the discover table once, across both runtimes**, against actual runtime status — rather than patching entries as they're spotted. That's exactly how `bun build` slipped through right after `deno install` was fixed.

To be clear on scope: I am **not** asking you to split this into per-runtime PRs. The shared runner is the coupling point, so splitting it is what *creates* the drift, and the rebase is finally clean. The genuinely asymmetric surface should stay asymmetric too — `bunx` has no deno equivalent, `deno check` has no bun one. The factoring argument is about `extract_test_summary` and the discover accounting, not the command surface.

The remaining three findings (tsc runner substitution, `pm ls` ANSI, the cwd-mutating test) are one-sided and stand alone.

geofflittle:
@KuSh all seven fixed at `75857a1`, and thanks for the correction on the round-2 suggestion.

`extract_test_summary` is now one buffered engine plus a per-runtime policy table (opens, proves, closes, keeps, section), so findings 1, 2 and 5 close by construction rather than as three more one-sided patches. Both byte-exact goldens still pass, so the unification changed nothing that already worked. Deno's proof column is the ERRORS section, not a marker: deno fences test stdout in `------- output -------` and prints nothing per failure that could vouch for a block, so bun's marker rule would not have transferred. Verified on real runtimes, a green run that logs `error:` is clean on both, and two consecutive leak failures no longer merge.

Fixtures are paired now. Every scenario asserts on both runtimes in one loop, so an unfixed mirror fails CI.

On `--no-install`, I did not take the documented-behavior-change option. Adding it changes what the command does rather than what it prints, which is the same rule that killed the engine substitution, so writing it down would have excused a second instance of the bug I was fixing. There is now an explicit `MissingTool::Fetch` or `Fail` per caller: `tsc` fetches exactly as `npx tsc` always did, lint, vitest, prettier and format keep `--no-install`, and the next caller routed through there has to choose rather than inherit. Net result is no behavior change to document.

The discover table was audited across both runtimes in one pass, which is what surfaced `bun build`. CI is still at `action_required`.

src/discover/rules.rs
KuSh: Reopening — I resolved this prematurely, and the fix I suggested was incomplete.

Adding `("install", RtkStatus::Passthrough)` sets the *status*, but `classify_command` resolves status and savings **independently** (`registry.rs:181-196`): a subcommand present in `subcmd_status` but absent from `subcmd_savings` falls back to `rule.savings_pct`, and `discover/mod.rs:148` multiplies output tokens by `estimated_savings_pct` without ever consulting status. So `deno install` is still credited the rule's 75% — which was the actual complaint, not the status flag.

`run` and `task` on this rule have it too, as do `run`, `pm` and `build` on the bun rule (reopened separately).

The fix has to reach the savings side. Either add a `0.0` entry to `subcmd_savings` for every passthrough subcommand, or — better — have `classify_command` force savings to 0 when status is `Passthrough`, so the two can't drift apart again. The second also fixes the pre-existing `cargo fmt` entry at `rules.rs:78`, which has the same shape.

Worth noting `test_passthrough_subcommands_claim_no_savings` (`registry.rs:3437`) asserts only `status` — it never checks a savings number, so it passes despite its name. Extending it would have caught this, and would stop it recurring.

src/discover/rules.rs
KuSh: Reopening for the same reason as the deno rule thread — resolved prematurely on my side, and my suggested fix was incomplete.

`("build", RtkStatus::Passthrough)` sets the status, but savings resolve independently (`registry.rs:189-194`), so `bun build ./index.ts` still carries this rule's 75% into `discover/mod.rs:148`. My round-3 wording — "dropping `build` from `subcmd_savings` to the passthrough default" — was simply wrong: the default **is** the 75%.

`run` and `pm` on this rule have the same problem. See the deno thread for the shape of the real fix.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,538 @@
...
+/// Run `bun test` showing only failures. Args are passed as a vector, never via a shell.
+pub fn run_test(args: &[String], verbose: u8) -> Result<i32> {
+    let mut cmd = resolved_command("bun");
+    cmd.arg("test").args(args);
+    let display = format!("test {}", args.join(" "));
+    crate::core::runner::run_test_cmd(
+        cmd,
+        "bun",
+        display.trim_end(),
+        "bun_test",
```
KuSh, line 226: `bun test --watch` produces **no output at all** and looks like a hang.

`run_test_cmd` goes through `core::runner::run_filtered` → `stream::run_streaming(..., FilterMode::CaptureOnly)` (`src/core/stream.rs:501`), which buffers the whole stream and prints nothing until the child exits. A watch process never exits, so nothing is ever printed; Ctrl-C discards the buffer, and the tee/recovery hint also only fires at exit, so there's nothing to fall back to either.

Same shape in `deno_cmd.rs` for `deno test --watch` (`:60`) and `deno lint --watch` (`:43`).

The repo already solved this twice: `src/cmds/js/vitest_cmd.rs:219-233` forces `--no-watch` and strips a user-supplied `--watch`, and `src/cmds/cloud/container.rs:804` rejects `-w`/`--watch` outright. Either approach is fine here — but `--watch` is common enough on both runtimes that doing neither means a user's first `rtk bun test --watch` looks like rtk broke their test runner.

src/core/runner.rs
```diff
@@ -284,3 +287,858 @@ pub fn run_streamed(
...
+}
+
+/// Bun writes a diagnostic before the marker it belongs to, so a block only
+/// counts once a `(fail)` marker claims it. A test logging its own `error:`,
+/// even with a stack, gets claimed by `(pass)` or by the counts instead.
+/// Module-level failures carry no marker at all, which is what the
+/// unhandled-error banner is for.
+const BUN_POLICY: BlockPolicy = BlockPolicy {
+    opens: |t| t.starts_with("error:"),
+    proves: |t| t.starts_with("at ") || t.starts_with("Expected:") || t.starts_with("Received:"),
```
KuSh: `proves` requires `at `, `Expected:` or `Received:`, so a genuine bun failure carrying none of them loses its diagnostic completely.

A timeout is the clearest case:
```
error: Test "slow" timed out after 5000ms
(fail) slow [5001ms]
```
The block opens untrusted (no `# Unhandled error` banner), `evidence` stays false, and although `(fail)` is a `Vouched` close, `close()` requires `evidence && Vouched` — so `pending` is discarded. Output shows `(fail) slow` with nothing saying *why*.

Blocks opened outside the unhandled-error section are the exposed ones, which is the ordinary in-run case. Note this isn't a regression from the refactor — the old code had the same gap — but the policy table makes it fixable in one place now.

A `Vouched` close arguably *is* the proof: `(fail)` is bun claiming the block, which is stronger evidence than a stack frame that a test could have printed itself. Treating a vouched close as sufficient on its own, with `proves` only rescuing blocks that close plainly, would cover timeouts without reopening the stdout leak. Worth a frameless-failure fixture either way.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,538 @@
...
+                    let bracket_content = &trimmed[1..close];
+                    if bracket_content.contains('/') {
+                        let parts: Vec<&str> = bracket_content.split('/').collect();
+                        if parts.len() == 2
+                            && parts[0].trim().parse::<u32>().is_ok()
+                            && parts[1].trim().parse::<u32>().is_ok()
+                        {
+                            continue;
+                        }
+                    }
```
KuSh: This branch never fires against real bun output, which leaves the `("install", 80.0)` claim in `rules.rs:681` resting on nothing.

It matches a leading `[n/m]` bracket followed by `...` — `[1/5] Resolving packages...` — which is **yarn's** progress format. Bun renders install progress only on a TTY, and as `Resolving [12/45]` with the bracket trailing, so when rtk pipes the child there is no progress line to strip at all. The only part of `filter_bun_pkg` that matches real output is the version-header skip (`bun install v1.3.6 (d530ed99)`, 4 tokens).

Every `bun install`/`add`/`remove` test here is an inline synthetic string in that yarn-shaped format, so the tests confirm the filter matches the fixture rather than that the fixture matches bun. `.claude/rules/cli-testing.md` calls this out by name.

Capturing a real `tests/fixtures/bun_install_raw.txt` would settle whether there's anything to strip on the piped path — and if there isn't, the honest move is to drop the 80% claim rather than keep a branch that only fires on yarn.

src/discover/rules.rs
```diff
@@ -671,6 +671,51 @@ pub const RULES: &[RtkRule] = &[
...
+        savings_pct: 70.0,
+        ..RtkRule::DEFAULT
+    },
+    RtkRule {
+        pattern: r"^deno\s+(test|lint|check|run|task|compile|install)\b",
+        rtk_cmd: "rtk deno",
+        rewrite_prefixes: &["deno"],
+        category: "Build",
+        savings_pct: 75.0,
+        subcmd_savings: &[("test", 90.0), ("lint", 80.0)],
```
KuSh: `("lint", 80.0)` and the 75% headline reaching `check` overstate what the filter does.

`filter_deno_output` (`deno_cmd.rs:8`) drops blank lines and lines starting with `Download `, and nothing else. Real `deno lint` output contains no `Download ` lines at all, and `deno check` emits them only on a cold cache — so on a warm project the actual reduction is close to 0% while `rtk discover` recommends the command promising 80%.

`test_filter_deno_output_token_savings` doesn't catch it because its input is synthetic and 15 of 18 lines are `Download` lines: it verifies the fixture, not the command. A captured `deno_lint_raw.txt` / `deno_check_raw.txt` would give these numbers something real to stand on, or show they need lowering.

src/main.rs
```diff
@@ -2270,11 +2403,11 @@ fn run_cli() -> Result<i32> {
             }
         },
 
-        Commands::Tsc { args } => tsc_cmd::run(&args, cli.verbose)?,
+        Commands::Tsc { args } => tsc_cmd::run(None, &args, cli.verbose)?,
 
         Commands::Next { args } => next_cmd::run(&args, cli.verbose)?,
 
-        Commands::Lint { args } => lint_cmd::run(&args, cli.verbose)?,
+        Commands::Lint { args } => lint_cmd::run(None, &args, cli.verbose)?,
```
KuSh, line 2410: `runner: None` here defeats the invariant the previous commit just established.

`lint_cmd::strip_pm_prefix` (`lint_cmd.rs:63`) exists specifically to consume a leading `bunx`/`npx`/`pnpm`/`yarn` from these args. So `rtk lint bunx eslint .` strips `bunx`, then — because no runner was threaded in — falls back to lockfile detection and runs eslint through `pnpm` in a pnpm project. That's the exact "a user who types `bunx` must not get `pnpm`" case `tool_exec` was introduced to prevent, still live on this path.

The prefix `strip_pm_prefix` removes is the runner the user named, so it should be recovered there and passed through rather than discarded.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,538 @@
...
+        crate::core::runner::RunOptions::with_tee(&tee_label),
+    )
+}
+
+pub fn run_pm_ls(args: &[String], verbose: u8) -> Result<i32> {
+    let mut cmd = resolved_command("bun");
+    cmd.arg("pm").arg("ls");
+    if !args.iter().any(|a| a == "--json") {
+        cmd.arg("--json");
+    }
```
KuSh: Injecting `--json` into a command the user typed can change the outcome, not just the output.

The doc comment on `filter_bun_pm_ls` says bun 1.x silently ignores the flag — which makes this dead weight on the happy path, but on any bun version that rejects unknown flags `rtk bun pm ls` exits non-zero where the bare command succeeds. Since the tree parser is already the live path and selection is by output shape rather than by the flag, the injection isn't buying anything.

It also cuts against the rule this PR just applied to runners: forward what was typed rather than substituting.

Secondary, same function: when neither the JSON nor the tree parser matches, `filter_bun_pm_ls_text` truncates at 500 chars, so a long dependency list is silently cut mid-list with no indication anything was dropped.

src/core/utils.rs
```diff
@@ -331,10 +331,18 @@ pub fn count_tokens(text: &str) -> usize {
...
+/// Lockfile detection against an explicit directory, so callers (and tests) do
+/// not have to move the process's current directory to ask the question.
+pub fn detect_package_manager_in(dir: &std::path::Path) -> &'static str {
+    if dir.join("pnpm-lock.yaml").exists() {
         "pnpm"
-    } else if std::path::Path::new("yarn.lock").exists() {
+    } else if dir.join("yarn.lock").exists() {
         "yarn"
+    } else if dir.join("bun.lockb").exists() || dir.join("bun.lock").exists() {
+        "bun"
```
KuSh, line 397: This new `bun` arm leaves two things stale above and below it:

- The doc comment at `:323-330` still says *"Returns `pnpm`, `yarn`, or `npm` based on lockfile presence"* and the example repeats the same three.
- `test_detect_package_manager_default` at `:880-885` asserts `["pnpm", "yarn", "npm"].contains(&pm)`. That now fails for anyone running `cargo test` from a checkout with a `bun.lock`/`bun.lockb` at the repo root. The rtk repo has none, so CI stays green — but the test encodes an invariant the code no longer holds, and it'll fail on a contributor's machine rather than in CI.

src/core/runner.rs
```diff
@@ -284,3 +287,858 @@ pub fn run_streamed(
...
+    },
+};
+
+/// Deno groups every diagnostic it writes under an ERRORS section, and fences
+/// the test's own stdout between output rules. That section is the proof: deno
+/// prints no per-failure marker that could vouch for a block, so one opened
+/// anywhere else is the test talking.
+const DENO_POLICY: BlockPolicy = BlockPolicy {
+    opens: |t| t.starts_with("error:") && t != "error: Test failed",
+    proves: |t| t.starts_with("[Diff]"),
```
KuSh: `proves` can never fire for deno, so this line is dead.

`FailureBlocks::close` keeps a block on `opened_trusted || (evidence && close == Some(Close::Vouched))`, but `DENO_POLICY.closes` only ever returns `Close::Plain` or `None` — never `Vouched`. So `evidence` cannot participate in the decision, and deno's correctness comes entirely from the `ERRORS` → `Section::Trusted` path.

No wrong output today. But the table advertises a deno proof rule that cannot fire, which is misleading in a struct whose whole purpose is to make each runtime's policy legible side by side — and if someone later gives deno a vouching close, `[Diff]` would silently start mattering.

Either drop it, or give deno a `Vouched` close so the rule earns its place.

KuSh (dismissed):
Round 4. All 22 threads from rounds 1-3 verified fixed, and the `BlockPolicy` refactor is a faithful implementation of the structural ask — one `FailureBlocks` engine, both runtimes reduced to policy tables, deno's missing proof expressed as `Section::Trusted` on `ERRORS`, blocks closing when the next one opens, and paired fixtures for every scenario. That's exactly the shape I was arguing for, and it's better than what I sketched.

**Two threads reopened, and the incomplete fix is mine.** I asked for `RtkStatus::Passthrough` entries on `deno install` and `bun build`, checked that they appeared, and resolved. But `classify_command` resolves status and savings *independently* (`registry.rs:181-196`), and `discover/mod.rs:148` multiplies by `estimated_savings_pct` without consulting status — so every passthrough subcommand still carries the rule's 75%. The status flag was never the fix; I should have traced through to the savings path before closing those out. Details on the two reopened threads, including why I'd push the invariant into `classify_command` rather than patch entries.

Eight further findings inline. Two are worth doing first:

- **`--watch` produces zero output and appears to hang** on `bun test`, `deno test` and `deno lint`. `FilterMode::CaptureOnly` buffers until the child exits, and a watch process never does; Ctrl-C then discards the buffer, and the tee hint only fires at exit so there's nothing to recover. `vitest_cmd.rs:219-233` and `container.rs:804` already guard this — the new paths don't.
- **Bun failures with no stack frame lose their diagnostic entirely.** A timeout prints `error: Test "slow" timed out after 5000ms` then `(fail) slow`, with nothing matching `BUN_POLICY.proves`, so the block is discarded and the user sees a failure with no reason. That's a gap in the engine we just landed, not a regression from it.

A theme worth naming across three of the rest: **the savings claims that can't be reproduced are exactly the ones with no captured fixture.** `tests/fixtures/` has real bun 1.3.6 output for `bun test` and `bun pm ls` — and those filters hold up. There's no `bun_install_raw.txt`, no `deno_lint_raw.txt`, no `deno_check_raw.txt`, and those three are precisely where the numbers don't survive contact with real output. The `bun install` progress branch matches yarn's `[1/5] …` shape rather than bun's, and `filter_deno_output` only drops `Download ` lines that real `deno lint` never emits. Capturing those three fixtures would settle it either way.

Finally: the branch is `CONFLICTING` again after these commits.

geofflittle:
Hey @KuSh, while I was trying to be a good steward of this change and promptly respond to feedback, this is getting into never-ending territory. I won't be responding to requests for changes any more.

If the work done here is useful to you and desired, feel free to take up the rest of the changes you want in-hand.

KuSh:
> Hey @KuSh, while I was trying to be a good steward of this change and promptly respond to feedback, this is getting into never-ending territory. I won't be responding to requests for changes any more.
> 
> If the work done here is useful to you and desired, feel free to take up the rest of the changes you want in-hand.

Understood, thanks for your time and your work on this. I’ll take it from here.

KuSh:
Resolving the remaining review threads. Each was verified against the pushed tree and, where the behaviour is observable, against real `bun 1.2.20` / `deno 2.9.6` rather than only by reading the diff.

| Thread | Fix |
|---|---|
| `rules.rs:687` — `deno install` credited savings | `classify_command` derives savings from status, so a passthrough subcommand is 0% and the two fields cannot drift |
| `rules.rs:676` — `bun build` credited 75% | Same derivation, plus `("build", Passthrough)`; `bun build` now runs unfiltered |
| `bun_cmd.rs:244` — `bun test --watch` hangs silently | `is_watch_mode` guard on every filtered entry point |
| `runner.rs:610` — frameless failures lose their reason | A `(fail)`/`✗` marker vouches for the block it closes, so a timeout keeps `timed out after 200ms` |
| `bun_cmd.rs:56` — progress branch matched yarn's shape | Replaced with the lines bun actually prints when piped; claim now 70%, measured against a captured fixture |
| `rules.rs:708` — `deno lint` claimed 80% | Measured: lint 40%, check 50%, each backed by a real capture and asserted by the module's test |
| `main.rs:2410` — `rtk lint bunx eslint` ignored the runner | The stripped prefix is recovered and wins over lockfile detection |
| `bun_cmd.rs:184` — `--json` injected into `bun pm ls` | Removed; the tree parser selects on output shape |
| `utils.rs:345` — stale doc and default test | Both now cover `bun` |
| `runner.rs:638` — `DENO_POLICY.proves` was dead | The field is gone; vouching moved into `closes`, trust into `section` |

Four further rounds of review ran after these threads were opened; those findings are fixed in the same two commits. The ones deliberately left alone, with reasons, are in the PR description and in #2431 (`bunx`/`npx` stdin), #3738 and #3844 (`rtk err` dropping diagnostics — not reachable from this PR, since `bun build` and `deno compile` are passthrough here).

As noted in the description: I authored the remediation, so this still needs a second maintainer's review.

KuSh:
Validated by @pszymkowiak, merging

PR #990, opened 2026-04-03
fix(ls): filter quality wrong truncation

### LINKED ISSUE #948: [BUG] 'rtk ls' parsing issue
Related to https://github.com/rtk-ai/rtk/issues/822

 rtk ls misparses ls -la output on systems where owner/group columns contain spaces, causing
  incorrect filenames and 0B sizes

  Description

  rtk ls currently parses the plain text output of ls -la by splitting each line on whitespace and
  assuming fixed column positions.

  That assumption breaks on systems where the owner or group field contains spaces. In that case,
  the columns shift, so RTK reads the wrong token as the file size and also reconstructs the
  filename incorrectly.

  As a result:

  - some files are shown with a size of 0B even when they are not empty
  - some entry names include trailing date/time fragments such as 31 16:18
  - directory/file names can be corrupted in the compact output

  Current implementation

  In dev/projets/rtk/src/cmds/system/ls.rs:162, the code does:

  - split each ls -la line with split_whitespace()
  - assume parts[4] is the size
  - assume parts[8..] is the filename

  This is fragile because ls output is not a stable machine-readable format.

  Relevant lines:

  - dev/projets/rtk/src/cmds/system/ls.rs:162
  - dev/projets/rtk/src/cmds/system/ls.rs:167
  - dev/projets/rtk/src/cmds/system/ls.rs:185

  Environment where reproduced

  On this machine, ls -la shows a group name with spaces, for example:

  -rw-r--r--  1 fjeanne utilisa. du domaine     0 Mar 31 16:18 empty.txt

  RTK output for the same directory becomes:

  31 16:18 empty.txt  0B

  This demonstrates that the parser is consuming part of the date/time as part of the filename and
  defaulting the size to 0.

  Steps to reproduce

  1. Use a system where the owner or group name contains spaces in ls -la output.
  2. Create a test directory with at least one file.
  3. Run ls -la <dir>.
  4. Run rtk ls <dir>.

TropicalDog17:
## Summary

Might fixes #948

**Problem**: The old code split ls -la lines on whitespace and used fixed column indices (parts[4] for size, parts[8..] for filename). This broke when owner/group names contained spaces, causing the columns to shift — sizes became 0 and date/time fragments leaked into filenames.

**Fix**: The new parse_ls_line() function uses a regex (LS_DATE_RE) to find the date field as a stable anchor, which allow correctly parsing.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` passed for old and newly added tests


aeppling (approved):

TropicalDog17:
Is it ready to merge @aeppling?

aeppling:
was waiting for previous release, merged

aeppling:
Thanks for the contribution :)

PR #992, opened 2026-04-03
fix(go): avoid false build errors from download logs

em0t:
## Summary
- tighten `rtk go build` error detection so dependency download/progress lines are not treated as compiler failures
- keep real compiler errors such as `file.go:line:col` diagnostics and explicit toolchain error lines
- add regression coverage for package names containing `errors`, `xerrors`, and `multierror`

## Minimal Case
Raw
```text
go: downloading github.com/pkg/errors v0.9.1
go: downloading golang.org/x/xerrors v0.0.0-20231012003039-104605ab7028
```

Previous RTK
```text
Go build: 2 errors
  go: downloading github.com/pkg/errors v0.9.1
  go: downloading golang.org/x/xerrors v0.0.0-20231012003039-104605ab7028
```

This PR
```text
Go build: Success
```

## Testing
- cargo fmt --all
- cargo clippy --all-targets
- cargo test --all

## Notes
- `cargo clippy --all-targets` still reports pre-existing warnings outside this change set

### DIFF
--- a/src/cmds/go/go_cmd.rs
+++ b/src/cmds/go/go_cmd.rs
@@ -534,11 +534,14 @@ fn is_go_build_error_line(line: &str) -> bool {
         "no required module provides package ",
         "missing go.sum entry for module providing package ",
         "found packages ",
+        "go: go.mod file not found in current directory or any parent directory",
+        "go: cannot load module ",
         "go: build failed",
         "go: error ",
         "error: ",
         "go: updates to go.mod needed",
         "go: inconsistent vendoring",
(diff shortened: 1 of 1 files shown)


em0t:
Thanks @aeppling, good catch. I pushed an update that keeps the download-line false-positive fix but restores non-file build errors, including `undefined: ...`, `cannot find package ...`, import cycles, build constraint failures, and `go.mod`/`go.work` parse errors. I also added regression tests for these cases.

aeppling (approved):

PR #993, opened 2026-04-03
fix(git): preserve full diff hunk headers

em0t:
## Summary
- preserve the full unified diff hunk header in `rtk git diff` compact output instead of reconstructing it without trailing context
- keep function / symbol context that appears after the second `@@` marker
- add regression coverage for hunk headers such as `@@ -10,3 +10,4 @@ fn important_context() {`

## Minimal Case
Raw
```diff
@@ -10,3 +10,4 @@ fn important_context() {
```

Previous RTK
```diff
@@ -10,3 +10,4 @@
```

This PR
```diff
@@ -10,3 +10,4 @@ fn important_context() {
```

## Testing
- cargo fmt --all
- cargo clippy --all-targets
- cargo test --all

## Notes
- `cargo clippy --all-targets` still reports pre-existing warnings outside this change set

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -330,8 +330,9 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
             }
             in_hunk = true;
             hunk_shown = 0;
-            let hunk_info = line.split("@@").nth(1).unwrap_or("").trim();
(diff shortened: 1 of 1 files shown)


em0t:
Could a maintainer please click `Approve and run workflows` for this PR?

Issue #996, opened 2026-04-03
[Codex MacOS App] : RTK.md is not opened automatically


Valmonzo:
I configured with `rtk init -g --codex` like the doc said. 
I use the codex app for MacOS. After 2 days i saw that codex reads the AGENTS.md but doesn't go to the reference if i don't ask him explicitly.
Maybe we should copy paste all the RTK.md on AGENTS.md ?

jiemojiemo:
https://github.com/rtk-ai/rtk/pull/1002

I'm trying to fix this issue

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟢 **Priority** | `low` |
| 🎯 **Confidence** | 68% |

### Summary

After `rtk init -g --codex`, Codex reads AGENTS.md but does not automatically follow references to RTK.md; init may need to inline RTK.md content for Codex compatibility.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

Valmonzo:
Yeah, that’s exactly what I did before creating the issue. To be honest, instead of creating an issue next time, I could just get off my lazy arse and suggest it myself instead of bothering everyone else. I’m sorry.

Thanks again for the quick response and the work on this repo – it’s brilliant!!

Lucartas:
Still having issues even when using the absolute path in ~/.codex/AGENTS.md.
When asking which instructions it's following I get this response:
- AGENTS.md also references {path_to_RTK.md}, but its contents were not included in the prompt context.
I'm using codex-cli@0.118 on macOS 26.

akimabs:
any update how to fix this?

msladek:
not related to MacOS, also happens on linux.

codex does not expand the FQ include `@/home/<user>/.codex/RTK.md` in `AGENTS.md`.

ammoradi:
The same issue happens on Windows as well.

lyz508:
This looks related to an upstream Codex limitation rather than RTK itself:

https://github.com/openai/codex/issues/6038

[...]

PR #1002, opened 2026-04-03
Fix Codex global RTK reference path on macOS

jiemojiemo:
Description:
## Summary

This PR fixes a Codex global init issue on macOS where `rtk init -g --codex` wrote `@RTK.md` into `~/.codex/AGENTS.md`, but Codex resolved that reference relative to the current project instead of `~/.codex`. As a result, Codex could fail to find the global `RTK.md` when opened in another repository.

## What changed

- Update Codex global init to write the absolute `~/.codex/RTK.md` path into `~/.codex/AGENTS.md`
- Keep local Codex init behavior unchanged and continue using the project-local `@RTK.md` reference
- Make AGENTS.md patching migrate legacy global `@RTK.md` references to the new absolute-path format
- Make Codex uninstall and config detection handle both legacy relative references and the new absolute reference
- Update Codex documentation to describe the correct global behavior

## Tests

Added and updated unit tests covering:

- adding the RTK reference only once
- creating a missing `AGENTS.md`
- migrating inline RTK blocks
- migrating a legacy global `@RTK.md` reference to an absolute path
- removing the absolute-path reference during Codex uninstall
- preserving uninstall idempotency

Validated with:

```bash
cargo test patch_agents_md -- --nocapture
cargo test uninstall_codex_at -- --nocapture
```

## Why this matters

Codex global instructions should remain valid no matter which project is opened. Writing an absolute path for global Codex setup makes the configuration stable across repositories, while preserving the existing local-project workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,7 @@ Creates `~/.gemini/hooks/rtk-hook-gemini.sh` + patches `~/.gemini/settings.json`
 rtk init -g --codex
 ```
 
-Creates `~/.codex/RTK.md` + `~/.codex/AGENTS.md` with `@RTK.md` reference. Codex reads these as global instructions.
+Creates `~/.codex/RTK.md` + `~/.codex/AGENTS.md`. Global Codex installs write the absolute `~/.codex/RTK.md` path into `AGENTS.md` so the instructions still resolve when Codex is opened from another project.
 
 ### Windsurf
 
--- a/hooks/codex/README.md
+++ b/hooks/codex/README.md
@@ -5,5 +5,5 @@
 ## Specifics
 
 - Prompt-level guidance via awareness document -- no programmatic hook
-- `rtk-awareness.md` is injected into `AGENTS.md` with an `@RTK.md` reference
+- Global installs inject the absolute `~/.codex/RTK.md` path into `AGENTS.md`; local installs keep the project-local `@RTK.md` reference
 - Installed to `~/.codex/` by `rtk init --codex`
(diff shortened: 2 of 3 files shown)


em0t (commented):
I tested this on macOS with a temporary HOME and the behavior matches the PR description.

On this branch:
- global `rtk init -g --codex` writes a plain absolute RTK path into `~/.codex/AGENTS.md`
- legacy global references migrate correctly
- uninstall removes the RTK reference cleanly and stays idempotent
- local `rtk init --codex` still keeps the project-local `@RTK.md` flow

No blockers from my side.

aeppling:
Thanks for the quick fix @jiemojiemo — you were one of the first to jump on issue #996 and the approach was sound at the time.

However, this PR is now **outdated** and we'll close it. Here's why:

**1. Fix already landed in `develop`**
The absolute-path resolution for global Codex (`@{absolute_path}` format) was integrated into `develop` as part of the issue #892 / #996 fixes. The current `develop` branch already has `codex_rtk_md_ref()`, `has_rtk_reference()`, `run_codex_mode_with_paths()`, and all associated tests — covering the same ground as this PR. This is why the branch now shows merge conflicts.

**2. `@` prefix difference**
This PR writes the bare path (`/home/user/.codex/RTK.md`) whereas `develop` writes `@/home/user/.codex/RTK.md` (with the `@` include prefix). The `@` form is safer since it preserves Codex's file-include syntax.

**3. Native Codex hooks incoming (PR #899)**
The prompt-only Codex integration (AGENTS.md + RTK.md) is being superseded by a native `PreToolUse` hook approach in #899. That PR adds `rtk hook codex`, patches `.codex/hooks.json`, and provides a hard guardrail instead of relying on prompt-level references. Once that lands, the AGENTS.md reference format becomes secondary.

Closing this PR — thank you again for the contribution and the responsiveness on the issue. The fix is live in `develop` and a more robust Codex integration is on the way.

PR #1013, opened 2026-04-04
fix(init): skip redundant hook warning during rtk init

### LINKED ISSUE #1011: `rtk init -g` shows redundant "run `rtk init -g`" warning
## Description

Running `rtk init -g` prints a confusing warning before it has a chance to install the hook:

```
$ rtk init -g
```

The warning tells the user to run the exact command they're already running.

## Steps to Reproduce

1. Uninstall any existing hooks: `rtk init -g --uninstall`
2. Reset the rate-limit marker: `rm ~/Library/Application\ Support/rtk/.hook_warn_last`
3. Run `rtk init -g`
4. Observe the redundant warning before the hook is installed

## Root Cause

`maybe_warn()` in `hook_check.rs` fires for all commands except `Gain` before the command dispatches. When the command is `Init`, the hook hasn't been installed yet, so the check sees a missing hook and warns.

The warning is also rate-limited to once per day, making it hard to reproduce consistently.

## Expected Behavior

`rtk init -g` should not warn missing hooks nor redundantly recommend running `rtk init -g` again. If anything, it should indicate that it's about to install the hooks.

andeart:
## Summary

This fixes #1011.
- Skip `maybe_warn()` hook check when the command is `Init`, since it's about to install/update hooks itself
- Suppresses the misleading "No hook installed" warning that printed during `rtk init` (no replacement message, `init` already announces its own actions)
- Matches existing pattern where `Gain` already skips the check

## Test plan

- [x] Pre-test: clean up any previous markers first, with `rtk init -g --uninstall` and `rm ~/Library/Application\ Support/rtk/.hook_warn_last`, then rebuild the binary.
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`. All pass.
- [x] Manual testing: `rtk init -g` no longer prints the misleading "No hook installed" warning. `rtk init --uninstall` and `--show` stay clean too.


pszymkowiak:
Thanks @andeart — #1011 is a real bug (verified, still present on `develop`), and skipping the hook-check for `init` is the right call.

I built and tested the branch. The fix works: `rtk init -g` no longer prints the misleading "No hook installed" warning. Two things on the implementation though:

1. **`eprintln!("[rtk] Installing RTK hook")` also fires for `rtk init --uninstall`.** `--uninstall` is a flag on the `Init` command, so `matches!(.., Commands::Init { .. })` matches it too. Tested — `rtk init -g --uninstall` prints:
   ```
   [rtk] Installing RTK hook
   RTK uninstalled:
     - Hook: ... removed
   ```
   "Installing" during an *un*install.

2. The message sits in the generic hook-check block of `run_cli`. An "Installing…" notice belongs in `init`'s own code (which already prints its output), not in this block.

The minimal fix is just to extend the skip condition, no extra message:

```rust
if !matches!(cli.command, Commands::Gain { .. } | Commands::Init { .. }) {
    hooks::hook_check::maybe_warn();
}
```

`init` already announces what it does. Could you simplify it that way? The diagnosis is spot-on — it's just the fix that's a bit over-built.

andeart:
Thanks for the suggestion, @pszymkowiak . This is cleaner and more direct. I just updated the code if you could take another look when you have a moment please.

pszymkowiak:
Tested against a compiled build of the current head (052c283). Small PR, one real gap.

The exemption at `src/main.rs:1399` is a blanket match on the whole `Init` command:

```rust
// Warn if installed hook is outdated/missing (1/day, non-blocking).
// Skip for Gain (own inline hook warning) and Init (about to install/update hooks itself).
if !matches!(cli.command, Commands::Gain { .. } | Commands::Init { .. }) {
    hooks::hook_check::maybe_warn();
}
```

But `Init` covers every `--agent` target (cursor, windsurf, cline, kilocode, antigravity, hermes, ...), and only the Claude Code path actually touches the hook this warning is about. The other agent-init subflows (`run_hermes_mode`, `run_kilocode_mode`, etc. in `src/hooks/init.rs`) never write `.claude/settings.json`'s PreToolUse entry or the legacy hook script — so the comment's rationale ("about to install/update hooks itself") doesn't hold for them.

Repro with a compiled build (`~/.claude` exists but has no hook registered, so `status()` returns `Missing`):

```bash
# baseline — warns correctly
rtk git status
# [rtk] /!\ No hook installed — run `rtk init -g` for automatic token savings

# init for an unrelated agent — no warning, even though it never touches the Claude hook
rtk init --agent hermes
# (no warning printed)

# confirms the Claude hook is still missing right after
rtk git log
# [rtk] /!\ No hook installed — run `rtk init -g` for automatic token savings
```

Low severity — it's a single missed notification per day (the rate-limit marker is only written when the warning actually fires, so the next non-Init command still warns normally), not a correctness/security issue. But it doesn't match the stated rationale for non-Claude agent targets. Might be worth narrowing the skip to `Commands::Init { agent: None, .. }` / the Claude-specific branch, or checking `agent` inside the condition, rather than exempting the whole command.

PR #1016, opened 2026-04-04
fix(curl): preserve valid JSON output for piping (#1015)

### LINKED ISSUE #1015: bug(json): curl filter outputs invalid JSON — object keys lose double quotes
## Description

When `rtk curl` filters JSON API responses, the output has **unquoted object keys**, producing invalid JSON that breaks downstream parsers (`python3 -c 'json.load(...)'`, `jq`, etc.).

## Reproduction

```bash
# With rtk:
rtk curl -s https://any-api.example/endpoint | python3 -c 'import sys,json; json.load(sys.stdin)'
# => JSONDecodeError: Expecting property name enclosed in double quotes

# Without rtk (works fine):
rtk proxy curl -s https://any-api.example/endpoint | python3 -c 'import sys,json; json.load(sys.stdin)'
# => OK
```

## Expected vs Actual

**Expected** (valid JSON):
```json
{"id": 123, "status": "open"}
```

**Actual** (invalid — keys unquoted):
```
{id: 123, status: "open"}
```

## Root Cause

In `src/cmds/system/json_cmd.rs`, the `extract_schema()` function formats object keys without wrapping them in double quotes:

```rust
// Lines 251-258
lines.push(format!("{}  {}: {},", indent, key, val_trimmed));  // missing quotes around key
lines.push(format!("{}  {}: {}", indent, key, val_trimmed));   // missing quotes around key
lines.push(format!("{}  {}:", indent, key));                   // missing quotes around key
```

This is called from `curl_cmd.rs` line 61 via `filter_json_string()` → `extract_schema()`.

## Impact

- Any pipeline that does `rtk curl ... | python3 -c 'json.load(...)'` or `| jq .` breaks
- Workaround: `rtk proxy curl` bypasses filtering but defeats the purpose of rtk
- Affects all JSON API calls through curl (Gitea, GitHub, REST APIs, etc.)

## Environment

- rtk v0.34.3 (stable) and dev-0.35.0-rc.108 — both affected
- macOS ARM64

## Fix

Wrap `key` in double quotes at lines 253, 255, and 258 of `json_cmd.rs`.

pszymkowiak:
## Summary
- **curl filter** no longer transforms JSON into schema (type names), preserving valid JSON for downstream parsers
- **json_cmd extract_schema** now quotes object keys for valid JSON output
- Closes #1015

## Problem
`rtk curl https://api.example/endpoint | python3 -c 'json.load(sys.stdin)'` failed because:
1. `extract_schema()` replaced values with type names (`int`, `string`, `bool`)
2. Object keys were unquoted (`{name: string}` instead of `{"name": "string"}`)

## Fix
- `curl_cmd`: JSON responses pass through unchanged (no schema transform)
- `json_cmd`: keys are now double-quoted in `extract_schema()` for other consumers
- Non-JSON output still gets truncated as before

## Test plan
- [x] `cargo test curl_cmd` — 4 tests pass
- [x] `cargo test json_cmd` — 9 tests pass
- [x] Manual: `rtk curl <api> | python3 -c 'json.load(...)'` works

### DIFF
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -54,35 +54,30 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
 fn filter_curl_output(output: &str) -> String {
     let trimmed = output.trim();
 
-    // Try JSON detection: starts with { or [
+    // JSON output: pass through unchanged to preserve validity for piping (#1015)
     if (trimmed.starts_with('{') || trimmed.starts_with('['))
         && (trimmed.ends_with('}') || trimmed.ends_with(']'))
     {
-        if let Ok(schema) = json_cmd::filter_json_string(trimmed, 5) {
-            // Only use schema if it's actually shorter than the original (#297)
-            if schema.len() <= trimmed.len() {
-                return schema;
-            }
-        }
+        return trimmed.to_string();
     }
 
-    // Not JSON: truncate long output
+    // Non-JSON: truncate long output
     let lines: Vec<&str> = trimmed.lines().collect();
-    if lines.len() > 30 {
-        let mut result: Vec<&str> = lines[..30].to_vec();
+    if lines.len() > 50 {
+        let mut result: Vec<&str> = lines[..50].to_vec();
         result.push("");
         let msg = format!(
             "... ({} more lines, {} bytes total)",
-            lines.len() - 30,
+            lines.len() - 50,
             trimmed.len()
         );
         return format!("{}\n{}", result.join("\n"), msg);
     }
 
-    // Short output: return as-is but truncate long lines
+    // Short non-JSON output: truncate long lines
     lines
         .iter()
-        .map(|l| truncate(l, 200))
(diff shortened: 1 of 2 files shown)


pszymkowiak:
Closing in favor of #1602, which addresses the same underlying problem (#1536 — `rtk curl` truncating large JSON mid-stream and producing invalid output) but on the current `develop` shape of `curl_cmd.rs`. Since this PR was opened, the schema-conversion path it targeted has been refactored away and replaced with a hardcoded 500-byte truncation + tee hint, so a clean rebase wasn't practical.

#1602 keeps the spirit of the JSON-passthrough idea here, adds an `IsTerminal` gate so pipes/redirects also receive the full body (also closes #1282), and ports the test coverage.

The `json_cmd.rs::extract_schema` key-quoting change isn't carried over: even with quoted keys, the surrounding pseudo-schema (`int`, `string`, `array`) isn't valid JSON, so the change adds tokens without making the output parseable. If we want a real "valid JSON schema" output mode for `rtk json` it should be a separate, deliberate feature.

PR #1019, opened 2026-04-04
fix(read): fallback to raw bytes for non-UTF-8 reads

em0t:
## Summary
- preserve plain raw-read semantics for non-UTF-8 files by falling back to raw byte passthrough instead of failing early in `rtk read`
- keep the fallback narrow: it only applies to unfiltered reads with no line windows and no line numbers
- leave text-oriented modes unchanged, so filtered reads and `-n` still require valid UTF-8
- add regression coverage for the binary fallback gate and the non-UTF-8 decode path

### non-UTF-8 raw read
Raw
```text
/bin/cat nonutf8.bin >/dev/null
exit 0
```

Previous RTK
```text
rtk read nonutf8.bin >/dev/null
stderr contains: stream did not contain valid UTF-8
exit 1
```

This PR
```text
rtk read nonutf8.bin >/dev/null
exit 0
```

## Testing
- `cargo fmt --all`
- `cargo test test_supports_binary_passthrough`
- `cargo test test_binary_passthrough`
- `cargo test --all`

### DIFF
--- a/src/cmds/system/read.rs
+++ b/src/cmds/system/read.rs
@@ -4,6 +4,7 @@ use crate::core::filter::{self, FilterLevel, Language};
 use crate::core::tracking;
 use anyhow::{Context, Result};
 use std::fs;
+use std::io::Write;
 use std::path::Path;
 
 pub fn run(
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Thanks @em0t — the non-UTF-8 read bug is real and still present on `develop`. This branch is ~6 weeks behind though (329 commits, never rebased). Could you rebase onto `develop` first? Once it's up to date I'll do a full review on the current code — no point reviewing line-by-line against a 6-week-old base.

em0t:
@pszymkowiak Thanks for taking a look.

I have rebased this branch onto current `develop`. Could you please continue the review when convenient?

em0t:
@aeppling Sorry to bother you here.

This PR has also been rebased onto current `develop` after the previous review feedback, and it is now mergeable. @pszymkowiak had asked for the rebase before doing the full review.

Since you previously helped review and merge my RTK filter fixes, could you please take a look when you have time, or help route this to the right maintainer?

Thanks again.

PR #1021, opened 2026-04-04
fix(proxy): kill child process on SIGINT/SIGTERM to prevent orphans

aeppling:
## Summary
- Register SIGINT/SIGTERM signal handler in proxy mode that kills the child process via its PID stored in an `AtomicU32`
- Fixes orphan processes when `rtk proxy` is killed (by Claude Code, process managers, or manual `kill`)
- Drop-based ChildGuard doesn't run on signals with `panic=abort` (release profile) — this is the actual fix for issue #897
- Uses `libc` (already in dependency tree via `dirs`) for `kill(2)` syscall — zero new downloads

## Reproduction
```bash
# Before: child survives
rtk proxy sleep 300 &; kill $!; sleep 2; pgrep sleep  # orphan alive

# After: child killed
rtk proxy sleep 300 &; kill $!; sleep 2; pgrep sleep  # nothing
```

## Test plan
- [x] `cargo test --all` — 1350 passed, 0 failed
- [x] SIGTERM kills child (verified manually)
- [x] Normal proxy exit still works (`rtk proxy echo hello`)
- [x] Exit code propagation preserved (`rtk proxy false` → exit 1)
- [x] Large output passthrough unaffected (`rtk proxy seq 1 100000`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -916,6 +916,7 @@ dependencies = [
  "hostname",
  "ignore",
  "lazy_static",
+ "libc",
  "quick-xml",
  "regex",
  "rusqlite",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,9 @@ quick-xml = "0.37"
 which = "8"
 automod = "1"
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [build-dependencies]
 toml = "0.8"
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -1951,6 +1951,7 @@ fn run_cli() -> Result<i32> {
         Commands::Proxy { args } => {
             use std::io::{Read, Write};
             use std::process::Stdio;
+            use std::sync::atomic::{AtomicU32, Ordering};
             use std::thread;
 
             if args.is_empty() {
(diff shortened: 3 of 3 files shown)


pszymkowiak (approved):
Thanks @aeppling — tested the release binary manually. Signal handling works for direct child processes.

Build: 1350 tests passed, 6 ignored. 2 new clippy warnings on the `libc::signal` cast (easy fix: `handle_signal as *const () as libc::sighandler_t`).

**Manual test results:**

| Test | Result |
|------|--------|
| Proxy normal exit | PASS |
| SIGINT kills child (sleep) | PASS |
| SIGTERM kills child (sleep) | PASS |
| Exit code propagation (1, 42) | PASS |
| Sequential proxy (static PID reuse) | PASS |
| Stress 10 rapid proxies (no leak) | PASS |
| Multi-child (sh -c "sleep & sleep & wait") + SIGINT | WARN — direct child killed, grandchildren survive |

**One edge case for follow-up** (not blocking):

The signal handler kills the direct child PID, but if the child is a shell with background subprocesses (`sh -c "cmd1 & cmd2"`), the grandchildren become orphans. This is the exact pattern from #897 where grep subprocesses accumulated. The fix would be to use process groups:

```rust
// At spawn: make child its own process group leader
use std::os::unix::process::CommandExt;
cmd.process_group(0);

// In signal handler: kill the process group (negative PID)
libc::kill(-(pid as libc::pid_t), libc::SIGTERM);
```

This is a follow-up improvement — the current PR already fixes the common case (direct child orphan on SIGINT/SIGTERM).

LGTM.

PR #1023, opened 2026-04-04
fix(json): quote object keys in schema output

### LINKED ISSUE #1015: bug(json): curl filter outputs invalid JSON — object keys lose double quotes

CastawayEGR:
Object keys in `compact_json()` and `extract_schema()` were missing double quotes, producing invalid JSON that breaks `jq`, `python3 json.load()`, etc.

Wraps all key formatting in `"key":` instead of `key:`.

Fixes #1015

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+* **json:** quote object keys in extract_schema and compact_json output ([#1015](https://github.com/rtk-ai/rtk/issues/1015))
 * **security:** missing toml pkg ([51f9c88](https://github.com/rtk-ai/rtk/commit/51f9c888b81169309df92f7fa3a6f705d44adcab))
 * **security:** salt device hash for telemetry ([32fdbbb](https://github.com/rtk-ai/rtk/commit/32fdbbbb6923c70d343fab14b4b0ce70424e610f))
 * **security:** set 0600 permissions on salt file ([5eae11d](https://github.com/rtk-ai/rtk/commit/5eae11d16410dc4ff26e97672e5367b14efaab76))
--- a/src/cmds/system/json_cmd.rs
+++ b/src/cmds/system/json_cmd.rs
@@ -158,9 +158,9 @@ fn compact_json(value: &Value, depth: usize, max_depth: usize) -> String {
 
                     if is_simple {
                         let val_str = compact_json(val, 0, max_depth);
-                        lines.push(format!("{}  {}: {}", indent, key, val_str.trim()));
+                        lines.push(format!("{}  \"{}\": {}", indent, key, val_str.trim()));
                     } else {
-                        lines.push(format!("{}  {}:", indent, key));
+                        lines.push(format!("{}  \"{}\":", indent, key));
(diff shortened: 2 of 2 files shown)


em0t (changes requested):
I tested this locally against current master and this branch, and I still see blocking issues before merge.

First, `rtk json <file>` still does not produce valid JSON for object output because object entries are not comma-separated, so `jq` and `python -c 'import json,sys; json.load(sys.stdin)'` still fail on simple cases.

Second, keys are wrapped in quotes but not escaped, so inputs containing `"` or `\` in object keys still produce invalid output.

I also don’t think `--schema` currently satisfies the “valid JSON” claim, because tokens like `int`, `string`, and `[int] (3)` are not JSON values.

I’d either narrow the PR description to “quote keys in display output” or extend the implementation/tests to cover actual JSON-valid output.

pszymkowiak:
@CastawayEGR — gentle ping: still on this one? em0t's review has 3 points to address before it can merge (comma-separated object entries, key escaping, and the `--schema` "valid JSON" claim). Happy to move it forward once those are sorted — just let us know if you'd like to continue.

pszymkowiak:
Closing: already fixed on develop. The scenario in #1015 (`rtk curl` returning invalid JSON) was resolved by 71baf2b8 and 02da3d07, which stopped routing curl's JSON bodies through the schema formatter; `rtk curl ... | jq` parses on current develop. What this diff still touches, `rtk json`, is a lossy summary format by design and would not become valid JSON with quoted keys alone (entries are not comma-separated, keys are not escaped). Thanks for the PR.

PR #1026, opened 2026-04-04
Add Codex hook installation and guidance

### LINKED ISSUE #1003: Add hook to Codex
Codex now supports plugins and hooks. You can update your Codex installation with hook.

ThewindMom:
## Summary
- add Codex-specific `rtk hook codex` handling and install it via Codex `hooks.json`
- enable `features.codex_hooks = true` during `rtk init --codex`, while preserving AGENTS/RTK guidance
- make Codex hooks point at the exact RTK binary that ran `rtk init --codex`, so the live hook works even when an older `rtk` exists on PATH
- update Codex docs to explain the current deny-with-suggestion behavior and add coverage for install/uninstall helpers

## Why
Codex CLI now exposes hooks, but the current official Codex hooks docs say `updatedInput` is parsed but not supported yet for `PreToolUse`, so it fails open. RTK therefore installs a Codex hook that blocks matching raw Bash commands and tells Codex to retry with the `rtk ...` equivalent.

During live testing I also found a real install bug: a bare `rtk hook codex` could hit an older RTK binary on PATH, causing the hook to silently do nothing. The hook now stores the exact RTK executable path used during install.

Docs reference: https://developers.openai.com/codex/hooks

## Validation
- `cargo fmt --all`
- `cargo test --all`
- temp-home smoke test for `rtk init -g --codex`, generated `hooks.json` / `config.toml`, `rtk hook codex`, and `codex features list`
- live Codex exec with temporary deny hook: command blocked
- live Codex exec with temporary allow + `updatedInput`: original command still ran
- live Codex exec after reinstall using absolute RTK hook path: real RTK hook blocked `git status` and suggested `rtk git status`

Closes #1003.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,9 @@ rtk init --agent cline          # Cline / Roo Code
 git status  # Automatically rewritten to rtk git status
 ```
 
-The hook transparently rewrites Bash commands (e.g., `git status` -> `rtk git status`) before execution. Claude never sees the rewrite, it just gets compressed output.
+> Codex note: current Codex hooks block + suggest `rtk git status` rather than rewriting in place.
+
+Most supported hook integrations transparently rewrite Bash commands (e.g., `git status` -> `rtk git status`) before execution. Codex currently cannot apply `updatedInput` in hooks yet, so RTK uses a deny-with-suggestion hook there instead.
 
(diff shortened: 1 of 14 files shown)


rebroad:
why was this closed?

PR #809, opened 2026-03-25
feat: enable hook-based mode on Windows

MauroDeryckere:
## Summary
- Remove `#[cfg(not(unix))]` fallbacks in `run_default_mode` and `run_hook_only_mode` that forced Windows into legacy CLAUDE.md injection mode
- Make `ensure_hook_installed` cross-platform by wrapping only the `chmod` call in `#[cfg(unix)]`
- Windows users can now run `rtk init -g` and get the same hook-based setup as macOS/Linux (zero context cost, deterministic rewriting)

## Test plan
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [x] Manual testing: `rtk init -g` hook script written and settings.json patched on Windows 11
- [x] Manual testing: PreToolUse hook fires and rewrites commands in a live Claude Code session on Windows (Git Bash)
- [x] Manual testing: `rtk rewrite "git status"` returns `rtk git status` on Windows

PR #727, opened 2026-03-19
feat: Windows hook support with --windows flag (opt-in Node.js hooks)

Pamacea:
## Summary

Adds Windows support for `rtk init` via an opt-in `--windows` flag that installs Node.js hooks instead of shell scripts.

### Problem

RTK's `rtk init` only worked on Unix (macOS/Linux) because:
- Hooks were written in bash (`rtk-rewrite.sh`)
- Code used `#[cfg(unix)]` guards to block Windows
- `chmod +x` and `PermissionsExt` are Unix-only

Windows users got this error:
```
[warn] Hook-based mode requires Unix (macOS/Linux).
    Windows: use --claude-md mode for full injection.
```

### Solution

**Conservative approach** - keeps existing behavior as default and adds opt-in flag:

**New files:**
- `hooks/rtk-rewrite.cjs` - Claude Code Node.js hook (cross-platform)
- `hooks/cursor-rtk-rewrite.cjs` - Cursor IDE Node.js hook (cross-platform)

**Changes to `src/init.rs`:**
- Add `--windows` flag to Init command
- Add `REWRITE_HOOK_CJS` and `CURSOR_REWRITE_HOOK_CJS` constants
- Modify `prepare_hook_paths()` to accept `use_windows` parameter
- Modify `ensure_hook_installed()` to write .cjs when `use_windows=true`
- Update `run_default_mode()` and `run_hook_only_mode()` to pass flag through
- Update `install_cursor_hooks()` to support Windows mode
- Update tests to include new `windows` parameter

### Usage

```bash
# Default behavior (Unix/macOS/Linux) - uses .sh hooks
rtk init -g

# New Windows mode - uses .cjs hooks (works on all platforms)
rtk init -g --windows
```

### Key Differences from PR #726

| Aspect | PR #726 (replacement) | This PR (opt-in flag) |
|--------|----------------------|------------------------|
| Default behavior | .cjs for everyone | .sh for Unix, error for Windows |
| User choice | None | `--windows` flag opts in |
| Backward compatibility | Breaking (removes .sh) | Non-breaking |
| Migration path | Forced (auto-switch to .cjs) | User-controlled |

### Token Savings

Same 60-90% savings on all platforms (with `--windows` flag).

### Testing

```bash
cargo check    # ✅ passes
cargo test     # ✅ 981 passed, 1 unrelated failure (Windows path issue)
cargo test init::  # ✅ 42 passed
```

### Checklist

- [x] Code compiles (`cargo check`)
[...]

PR #726, opened 2026-03-19
feat: Windows hook support - cross-platform Node.js hooks

Pamacea:
## Summary

Adds full Windows support for `rtk init` by replacing Unix-only shell scripts with cross-platform Node.js hooks.

### Problem

RTK's `rtk init` command only worked on Unix (macOS/Linux) because:
- Hooks were written in bash (`rtk-rewrite.sh`)
- Code used `#[cfg(unix)]` guards to block Windows
- `chmod +x` and `PermissionsExt` are Unix-only

Windows users got this error:
```
[warn] Hook-based mode requires Unix (macOS/Linux).
    Windows: use --claude-md mode for full injection.
```

### Solution

Replace shell scripts with Node.js (`.cjs`) files that work on all platforms:

**New files:**
- `hooks/rtk-rewrite.cjs` - Claude Code hook (Windows/macOS/Linux)
- `hooks/cursor-rtk-rewrite.cjs` - Cursor IDE hook (Windows/macOS/Linux)

**Changes to `src/init.rs`:**
- Removed `#[cfg(unix)]` guards - functions now work on all platforms
- Removed `chmod`/`PermissionsExt` code (not needed for .cjs)
- Updated `prepare_hook_paths()` to use `.cjs` extension
- Updated tests to validate Node.js format

### Token Savings

Same 60-90% savings on ALL platforms (previously Unix-only).

### Testing

```bash
cargo check    # ✅ passes
cargo test     # ✅ 983 tests passed
```

### Breaking Changes

None for end users - the .cjs hooks provide identical functionality.
The shell scripts are no longer installed (replaced by .cjs).

### Checklist

- [x] Code compiles (`cargo check`)
- [x] All tests pass (`cargo test`)
- [x] No clippy warnings
- [x] Cross-platform compatible
- [x] Tests updated for new format

---

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #551, opened 2026-03-12
feat: native cross-platform hook for Windows support

bz00qa:
## Summary

- **`rtk hook-rewrite` command**: Native Claude Code PreToolUse hook that reads JSON from stdin, rewrites commands via the registry, and outputs hook response JSON — replaces the bash+jq dependency for cross-platform support
- **Windows `rtk init -g`**: Now registers native hook instead of falling back to `--claude-md` mode
- **Windows `rtk init -g --hook-only`**: Uses native hook instead of bailing with an error
- **`#[cfg(unix)]` guards**: Bash-only code paths (REWRITE_HOOK, prepare_hook_paths) properly gated
- **Clippy fixes**: Unnecessary borrows, dead code annotations in init.rs

## Why

Windows users currently get a degraded experience — `rtk init -g` falls back to `--claude-md` mode because the bash hook (`rtk-rewrite.sh`) requires Unix. The native `rtk hook-rewrite` command eliminates the bash/jq dependency entirely, making the hook work on all platforms.

## Changes

| File | Description |
|------|-------------|
| `src/hook_rewrite_cmd.rs` | New native hook: parse stdin JSON, rewrite via registry, output hook response |
| `src/main.rs` | Add `mod hook_rewrite_cmd`, `HookRewrite` command variant + routing |
| `src/init.rs` | Windows default/hook-only modes use native hook, `#[cfg(unix)]` guards, clippy fixes |

## Test plan

- [x] All 891 tests pass (1 pre-existing binlog CRLF failure)
- [x] 4 new tests: hook JSON structure, field preservation, unsupported commands, already-rtk passthrough
- [x] `cargo fmt --all --check` passes
- [x] Manual: `rtk init -g` on Windows, verify settings.json gets `rtk hook-rewrite` entry
- [x] Manual: pipe hook JSON to `echo '{"tool_input":{"command":"git status"}}' | rtk hook-rewrite`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1027, opened 2026-04-04
fix(init): enable hook-based mode on Windows

### LINKED ISSUE #913: Persistent "No hook installed" warning on Windows after rtk init -g --opencode

daywhirls:
## Summary

`rtk init -g` on Windows falls back to `--claude-md` mode with a warning ("Hook-based mode requires Unix"), and `rtk rewrite` emits a persistent "Hook outdated" warning — even when a working bash hook exists. This is because the entire `ensure_hook_installed`, `run_default_mode`, and `run_hook_only_mode` functions are gated behind `#[cfg(unix)]`, but the only Unix-specific operation is `chmod +x`.

This PR removes the platform gates and makes hook installation cross-platform:

- **`ensure_hook_installed`**: Remove `#[cfg(unix)]` from the function; wrap only the `chmod` call in `#[cfg(unix)]`
- **`run_default_mode` / `run_hook_only_mode`**: Delete the `#[cfg(not(unix))]` fallback stubs; remove `#[cfg(unix)]` from the real implementations
- **`patch_settings_json`**: On Windows, prefix the hook command with `bash` so the `.sh` script executes via Git Bash/MSYS2 (e.g. `bash "C:\Users\...\.claude\hooks\rtk-rewrite.sh"`)

## Testing

- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — all 1355 tests pass, 0 failures
- New test: `test_ensure_hook_installed_cross_platform` — verifies hook creation, content, and idempotency on all platforms
- New test: `test_hook_already_present_bash_wrapped` — verifies `bash "..."` wrapped commands are detected by `hook_already_present`
- End-to-end: built release binary on Windows 11, ran `rtk init -g`, confirmed hook installed and `rtk rewrite` no longer warns

## Context

There are several open PRs for Windows hook support (#150, #452, #551, #726, #727, #809, #905, #954). This PR takes the minimal approach — 1 file, +56/-36 lines — rather than introducing alternative hook implementations (Python, Node.js, native). The bash hook works on Windows via Git Bash, which is near-universal on Windows dev machines with git installed.

Fixes #913

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -307,7 +307,6 @@ fn prepare_hook_paths() -> Result<(PathBuf, PathBuf)> {
 }
 
 /// Write hook file if missing or outdated, return true if changed
-#[cfg(unix)]
 fn ensure_hook_installed(hook_path: &Path, verbose: u8) -> Result<bool> {
     let changed = if hook_path.exists() {
         let existing = fs::read_to_string(hook_path)
@@ -335,10 +334,13 @@ fn ensure_hook_installed(hook_path: &Path, verbose: u8) -> Result<bool> {
         true
     };
 
-    // Set executable permissions
-    use std::os::unix::fs::PermissionsExt;
-    fs::set_permissions(hook_path, fs::Permissions::from_mode(0o755))
-        .with_context(|| format!("Failed to set hook permissions: {}", hook_path.display()))?;
+    // set executable permissions (unix only -- not needed on windows)
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        fs::set_permissions(hook_path, fs::Permissions::from_mode(0o755))
+            .with_context(|| format!("Failed to set hook permissions: {}", hook_path.display()))?;
+    }
 
     // Store SHA-256 hash for runtime integrity verification.
     // Always store (idempotent) to ensure baseline exists even for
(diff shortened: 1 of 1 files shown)


aeppling:
Hey, solved on latest releases 0.37.x

Thanks for contributing and trying to address this issue.

Closing this for clean up

PR #1047, opened 2026-04-06
fix(windows): wrap .cmd/.bat wrappers with cmd.exe /C for reliable execution

### LINKED ISSUE #950: rtk cannot spawn .cmd/.sh wrapper commands on Windows (pnpm, npm, tsc)
## Summary

When `rtk` wraps commands like `pnpm`, `npm`, `npx`, or `tsc` on Windows, they fail with:

```
Error: Failed to run pnpm
```

This is because rtk (Rust binary) spawns these as subprocesses, but on Windows they are **not native executables** — they are shell script wrappers or `.cmd` batch files that require a shell to interpret.

## Root Cause

Windows Node.js toolchain installs wrapper scripts instead of real binaries:

| Command | Location | Actual file type |
|---------|----------|-----------------|
| `pnpm` | `Program Files/nodejs/` | POSIX shell script + `.CMD` batch |
| `npm`, `npx` | `Program Files/nodejs/` | bash script + `.cmd` batch |
| `tsc` | `AppData/Roaming/npm/` | sh script + `.cmd` batch |
| `corepack` | `Program Files/nodejs/` | `.cmd` batch |

When Git Bash runs `pnpm install` directly, the shell interprets the wrapper script. But when rtk intercepts and tries to spawn `pnpm` as a subprocess (presumably via `std::process::Command`), it fails because:

1. POSIX shell scripts need a shell interpreter
2. `.cmd` files need `cmd.exe /c` to execute

## Current Workaround

In our PreToolUse hook (see #502), we added a passthrough list for known wrapper commands:

```bash
CMD_WRAPPER_RE="^(pnpm|pnpx|npm|npx|tsc|tsserver|corepack)([[:space:]]|$)"
if printf '%s' "$COMMAND" | sed 's/&&/\n/g; s/||/\n/g; s/;/\n/g' | sed 's/^[[:space:]]*//' | grep -qE "$CMD_WRAPPER_RE"; then
  exit 0  # skip rtk for these commands
fi
```

This sacrifices rtk's token savings for these commands in exchange for execution reliability.

## Suggested Fix

When spawning a command on Windows, rtk could:

1. Check if the resolved path is a `.cmd` file → use `cmd.exe /c <cmd>` 
2. Check if it's a shell script → use `sh -c <cmd>` or the Git Bash `bash.exe`
3. Or simply always spawn via `cmd.exe /c` on Windows (similar to how Node.js `child_process` handles it with `shell: true`)

## Related

- #502 — Windows hook support request (same environment)

## Environment

- OS: Windows 11 Pro (10.0.26200)
- RTK: v0.27.2
- Node.js: installed via official installer (Program Files/nodejs/)
- pnpm: installed via corepack
- Shell: Git Bash (MINGW64)

reverinudog:
## Summary

Fixes #950

On Windows, Node.js tools (`npm`, `pnpm`, `npx`, `tsc`, `vitest`, etc.) are installed as `.cmd`/`.bat` shims. While `which::which()` correctly resolves these paths via PATHEXT, `Command::new(path)` via `CreateProcessW` can fail to execute complex `.cmd` scripts depending on environment and script content.

This PR modifies `resolved_command()` to detect when the resolved binary path has a `.cmd` or `.bat` extension and transparently wraps the execution with `cmd.exe /C`, matching the standard Windows pattern used by Node.js's `child_process` with `shell: true`.

## Changes

- **`command_for_path()`** (new): Builds a `Command` from a resolved path, wrapping `.cmd`/`.bat` with `cmd.exe /C` on Windows. On non-Windows platforms, this is simply `Command::new(path)`.
- **`is_cmd_or_bat()`** (new, `#[cfg(target_os = "windows")]`): Case-insensitive extension check for `.cmd`/`.bat`.
- **`resolved_command()`** (modified): Now delegates to `command_for_path()` instead of calling `Command::new(path)` directly, so all callers — JS ecosystem commands, passthrough mode, TOML filter mode, and proxy mode — benefit automatically.

## Why `cmd.exe /C`?

Rust's `std::process::Command::new()` uses `CreateProcessW` internally. While `CreateProcessW` can handle simple `.cmd` files, it may fail with complex scripts that use `SETLOCAL`, `FOR /F`, or other advanced batch constructs (which npm/pnpm `.cmd` shims use). Wrapping with `cmd.exe /C` is the standard reliable approach, matching what Node.js, Python (`subprocess`), and Go (`exec.Command`) all do on Windows.

## Test plan

7 new Windows-specific tests (`#[cfg(target_os = "windows")]`):

- [x] `test_command_for_path_wraps_cmd_with_cmd_exe` — `.cmd` executes via `cmd.exe /C`
- [x] `test_command_for_path_wraps_bat_with_cmd_exe` — `.bat` executes via `cmd.exe /C`
- [x] `test_command_for_path_passes_args_through_cmd_exe` — arguments forwarded correctly through `cmd.exe /C`
- [x] `test_command_for_path_does_not_wrap_exe` — `.exe` files are NOT wrapped
- [x] `test_is_cmd_or_bat` — extension detection (case-insensitive, covers `.cmd`, `.CMD`, `.bat`, `.BAT`, `.exe`, no-ext, `.ps1`)
- [x] Existing `test_resolved_command_executes_cmd_wrapper` continues to pass
- [x] Existing `test_resolved_command_fallback_on_unknown_binary` continues to pass

## Impact

- **Zero impact on non-Windows**: `command_for_path()` and `is_cmd_or_bat()` are `#[cfg(target_os = "windows")]`-gated; on Unix, it's simply `Command::new(path)`.
- **Centralized fix**: All command execution paths that go through `resolved_command()` benefit automatically — no per-module changes needed.
- **Backward compatible**: Native executables (`.exe`, extensionless binaries) are unaffected.


pszymkowiak:
Thanks @reverinudog — #950 is real (`resolved_command` still does a plain `Command::new` on `develop`, no `.cmd`/`.bat` wrapping; #269 fixed PATH *resolution* but not *execution*). The `utils.rs` fix looks right.

Three things before merge:

1. **Scope creep** — the PR touches `.release-please-manifest.json`, `CHANGELOG.md`, `Cargo.toml` (version bump) and `Cargo.lock`. Those are release-please's job — please drop them; the PR should only change `src/core/utils.rs`.
2. **CLA isn't signed** — please sign it (link in the CLA assistant comment above).
3. The branch is 353 commits behind `develop` — please rebase (that should also clear the stale `check-target` failure).

The Windows execution path can only be verified by the Windows CI runner, so a clean rebase + green CI is what we'll go on.

KuSh:
Thanks for digging into this and for the Windows test coverage. Since this PR was opened, rtk moved to edition 2024 with rust-version 1.91 in #3782 (merged 2026-09-14), and the standard library has spawned .bat and .cmd targets through cmd.exe with its own escaping since 1.77.2, so Command::new on a resolved shim no longer needs a manual wrap. #3728 (merged 2026-09-19) then built the child argument quoting on top of that and deliberately keeps batch programs on the standard library path, detecting them by the program extension, so replacing the program with cmd.exe would defeat that detection and change how quoted arguments reach the shim. What issue #950 actually reports is still open, since the failure in that trace is resolution rather than execution, with which::which not finding pnpm under Git Bash, and src/core/utils.rs:615 resolve_binary is where a fix would go. A narrow PR or an issue on that resolution path would be very welcome. Please comment or reopen if I have missed something.

PR #1057, opened 2026-04-06
fix: preserve caller binary path for path-prefixed command rewrites

### LINKED ISSUE #1053: path-prefixed commands pass through unfiltered; rewriting them needs a binary-override mechanism
RTK’s rewrite pipeline handles path-prefixed commands inconsistently.

Commands like `.venv/bin/pytest -v`, `.venv/bin/ruff check .`, and `./node_modules/.bin/vitest run` are recognized as supported by `classify_command()`, which normalizes `argv[0]` via `strip_absolute_path()`. However, `rewrite_segment()` does not normalize before prefix replacement, so these commands can classify as supported and still fail rewrite.

Relevant code in `src/discover/registry.rs`:
- `classify_command()` normalizes via `strip_absolute_path()` (around line 100)
- `rewrite_segment()` matches against the unnormalized token (around line 555)
- `exclude_commands` matching also uses the unnormalized base token (around line 512)

Current behavior is accidentally safe. Path-prefixed commands fail rewrite, fall through to raw execution, and preserve the original executable path. No filtering happens, but execution semantics do not change.

A naive fix that normalizes in `rewrite_segment()` and emits `rtk pytest -v` would change semantics from “run this exact binary” to “run whatever PATH resolves.” RTK handlers generally discard the caller’s executable path and re-resolve via `resolved_command()` (for example `pytest_cmd.rs:16` and `ruff_cmd.rs:44`). When the venv is not activated, or the path was specified intentionally, this can produce wrong-binary execution or command-not-found failures.

A correct fix requires three things:
1. Normalize path-prefixed commands in `rewrite_segment()` for matching only.
2. Preserve the original executable path in the rewritten RTK invocation, for example via a `--rtk-bin` flag or environment variable.
3. Teach relevant handlers to honor an explicit binary override instead of always resolving via PATH.

Without all three pieces, making path-prefixed commands rewrite successfully would turn safe passthrough into incorrect execution.

mhcoen:
rewrite_segment() now normalizes path-prefixed commands (e.g.                
  .venv/bin/pytest, ./node_modules/.bin/vitest) for matching, and carries the  
  original binary path via RTK_BIN so handlers invoke the correct executable   
  instead of re-resolving via PATH.                                            
                                                            
  Also fixes exclude_commands matching to compare against the basename rather  
  than the raw path-prefixed token.
                                                                               
  ## Summary                                                
                                         
  - Normalize path-prefixed commands in rewrite_segment() for matching only,   
  preserving the original executable path via a single-quoted RTK_BIN env var
  in the rewrite output                                                        
  - Teach resolved_command() and tool_exists() to honor RTK_BIN when its
  basename matches the requested binary, so handler branch decisions (e.g.     
  pytest_cmd.rs) use the correct executable
  - Fix exclude_commands to compare against the basename so exclusions work for
   path-prefixed commands                                                      
                                         
  ## Test plan                                                                 
                                                            
  - [x] Manual testing: `rtk rewrite '.venv/bin/pytest -v'` outputs
  `RTK_BIN='.venv/bin/pytest' rtk pytest -v`                                   
  - [x] Manual testing: `rtk rewrite 'pytest -v'` outputs `rtk pytest -v` (no
  RTK_BIN)                                                                     
  - [x] Manual testing: `RTK_BIN='.venv/bin/pytest' rtk pytest --version`
  executes the venv binary                                                     
  - [x] 13 new unit tests covering rewrite output, quoting, exclusion matching,
   resolved_command override, and tool_exists override                         

Fixes #1053

### DIFF
--- a/src/core/utils.rs
+++ b/src/core/utils.rs
@@ -337,6 +337,18 @@ pub fn resolve_binary(name: &str) -> Result<PathBuf> {
 /// # Returns
 /// A `Command` configured with the resolved binary path.
 pub fn resolved_command(name: &str) -> Command {
+    // Honor RTK_BIN override when the basename matches the requested binary.
+    // This preserves the caller's original binary path (e.g. .venv/bin/pytest)
+    // when the rewrite pipeline normalized it for matching purposes.
+    if let Ok(bin) = std::env::var("RTK_BIN") {
+        let bin_basename = std::path::Path::new(&bin)
+            .file_name()
+            .and_then(|f| f.to_str())
+            .unwrap_or("");
+        if bin_basename == name {
+            return Command::new(&bin); // nosemgrep: dynamic-command-execution
+        }
+    }
     match resolve_binary(name) {
         Ok(path) => Command::new(path),
         Err(e) => {
@@ -359,6 +371,17 @@ pub fn resolved_command(name: &str) -> Command {
 ///
 /// Replaces manual `Command::new("which").arg(tool)` checks that fail on Windows.
 pub fn tool_exists(name: &str) -> bool {
+    // Check RTK_BIN override first: if it points to this binary and the file
+    // exists, the tool is available even if it's not on PATH.
+    if let Ok(bin) = std::env::var("RTK_BIN") {
+        let bin_basename = std::path::Path::new(&bin)
+            .file_name()
+            .and_then(|f| f.to_str())
+            .unwrap_or("");
+        if bin_basename == name && std::path::Path::new(&bin).exists() {
+            return true;
+        }
+    }
     which::which(name).is_ok()
 }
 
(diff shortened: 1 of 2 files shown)


mhcoen:
@pszymkowiak Can this please be considered for merging? It's been in limbo for a long time and this is a major issue for Python users. I just updated it again for conflicts with later releases.

mhcoen:
Reproduced on **v0.42.0** (current release) and **develop**, both built clean from source.

`$ rtk --version
rtk 0.42.0`

**`rtk rewrite` (CLI):**

| `rtk rewrite "<cmd>"` | stdout | exit | result |
| --- | --- | --- | --- |
| `pytest -v` | `rtk pytest -v` | 0 | rewritten |
| `python -m pytest -v` | `rtk pytest -v` | 0 | rewritten |
| `.venv/bin/pytest -v` | (empty) | 1 | no rewrite |
| `./venv/bin/pytest -v` | (empty) | 1 | no rewrite |
| `/usr/bin/pytest -v` | (empty) | 1 | no rewrite |
| `.venv/bin/python -m pytest -v` | (empty) | 1 | no rewrite |

**`rtk hook claude` (the production PreToolUse entry point):**

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"pytest -v"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk pytest -v"},"permissionDecision":"allow"}}

$ echo '{"tool_name":"Bash","tool_input":{"command":".venv/bin/pytest -v"}}' | rtk hook claude
(empty — no hookSpecificOutput)

$ echo '{"tool_name":"Bash","tool_input":{"command":"/usr/bin/pytest -v"}}' | rtk hook claude
(empty — no hookSpecificOutput)

$ echo '{"tool_name":"Bash","tool_input":{"command":".venv/bin/python -m pytest -v"}}' | rtk hook claude
(empty — no hookSpecificOutput)
```

Only the bare, PATH-resolved forms (`pytest`, `python -m pytest`) are rewritten. Every path-prefixed form — virtualenv binaries, `./`-relative, and absolute `/usr/bin` — fails to match: `rewrite` exits 1, and the production hook emits no `hookSpecificOutput`, so Claude Code runs the command verbatim and rtk is never invoked. This is general to path-prefixing, not pytest-specific (`.venv/bin/ruff` and `/usr/bin/grep` behave identically). The result is 0% savings on these invocations, silently.

mhcoen:
Superseded by a rebased and split version, closing this one to keep #1053 readable.

This branch no longer merges: `develop` has moved 436 commits since its base, and the PHP
tooling work rewrote the same region of `rewrite_segment_inner` that it patched. Rather than
force this back into a mergeable state as one 207-line change to the rewrite core, I rebased
it onto current `develop` and split it into three pieces that can be reviewed and merged
independently:

| PR | What | Size |
| --- | --- | --- |
| #3590 | `resolved_command()` / `tool_exists()` honor `RTK_BIN` | +71, 1 file. Inert on its own, nothing sets the variable yet |
| #3592 | the rewrite emits `RTK_BIN` for path-prefixed commands | +194/-11, 2 files. Depends on #3590 |
| #3591 | `exclude_commands` matches path-prefixed invocations | +30/-1, 1 file. Independent of both |

Two deliberate changes from what was in this PR, both narrowing it:

- The new matching runs **only after** the existing literal prefixes fail, so every command
  that rewrites today is byte-identical. `./gradlew` no longer picks up an `RTK_BIN` prefix
  the way it did here, and the Composer tools keep their own narrower normalization, which
  their tests require.
- `.venv/bin/python -m pytest` is deliberately left unrewritten. The path names python while
  the rule routes to `rtk pytest`, so preserving it would hand pytest an interpreter path and
  silently resolve the wrong binary. Covering that form means teaching the python handlers
  about the interpreter, which belongs in its own PR.

All three are green on fmt, clippy, semgrep, Security Scan and tests across ubuntu, macos and
windows. The `benchmark` job is red on all of them and on `develop` itself, from the `find`
filter, unrelated to any of this.

Issue #640, opened 2026-03-16
I asked claude --effort high to do a security review of rtk


the-michael-toy:
I love this idea and I am constantly trying to squeeze more value of a session because once a compression happens, in a complex collaborative session, important knowledge is always lost.

I know nothing, I just pulled the repo and ask Claude how safe it was. I was taken aback by the answer. Just because it seems like such an obvious thing to do, and the output was was more scary than I expected ...

Feel free to close this, I'm actually running rtk right now, but I felt like I need to at least mention this.

# RTK Security Review

**Date**: 2026-03-16
**Reviewer**: Claude Opus 4.6 (automated)
**Scope**: Full codebase at commit 4467296 (master)

| Critical | High | Medium | Low | Info |
|:---:|:---:|:---:|:---:|:---:|
| 1 | 3 | 5 | 4 | 3 |

**Overall verdict**: Mostly sound architecture with real security engineering (SHA-256 hook pinning, trust boundaries), but some significant issues. The project has done more security work than the typical CLI tool.

---

## Policy vs Implementation Gap

The project has an existing `SECURITY.md` with a thorough PR review policy. Several findings in this review directly contradict the project's own stated security rules:

| Policy rule (SECURITY.md) | Violation found |
|---|---|
| `Command::new("sh")` listed as dangerous pattern (line 109) | `runner.rs` and `summary.rs` use exactly this pattern (C-1) |
| "No shell, direct binary execution" (line 159) | `sh -c` with unsanitized user input in 3 code paths |
| "NEVER use `.unwrap()` in `src/`" (line 172) | `tee.rs` byte-index slicing panics on multi-byte UTF-8 (L-3) |
| "Data exfiltration" listed as a risk (line 24) | Telemetry sends device hash + top commands with no opt-in consent (H-1) |
| "Canonicalize file paths to prevent traversal attacks" (line 181) | `RTK_TEE_DIR` accepted without validation (M-1) |
| "tracking.db exposure" listed as a concern (line 24) | Full command strings with secrets stored for 90 days (M-2) |

The security policy describes the right approach — the implementation doesn't fully match it.

---

### C-1: Shell injection via `sh -c` in `runner.rs` / `summary.rs`

**Files**: `src/runner.rs` (lines 14-27, 73-86), `src/summary.rs` (lines 15-28)

[...]

PR #1072, opened 2026-04-07
fix(tee): prevent panic on UTF-8 multi-byte truncation boundary

pszymkowiak:
## Summary
- `&raw[..max_file_size]` in `write_tee_file` panics if the byte offset falls inside a multi-byte UTF-8 character (Japanese, emoji, etc.)
- Now uses `char_indices()` to find the nearest safe boundary before slicing
- 2 new tests: Japanese 3-byte chars + emoji 4-byte chars at truncation boundary

Ref: #640 (L-3 finding from security audit)

## Test plan
- [x] 1360 tests pass, 6 ignored
- [x] `test_write_tee_file_truncation_utf8_boundary` — 3-byte Japanese chars, cut at 998/999
- [x] `test_write_tee_file_truncation_emoji` — 4-byte emoji, cut at 201/400
- [x] Python proof: old code would decode-error at byte 996-997

### DIFF
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -121,11 +121,17 @@ fn write_tee_file(
     let filename = format!("{}_{}.log", epoch, slug);
     let filepath = tee_dir.join(filename);
 
-    // Truncate at max_file_size
+    // Truncate at max_file_size (find a safe UTF-8 char boundary)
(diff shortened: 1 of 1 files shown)


FlorianBruniaux (approved):
LGTM — security fix, reviewed diff, tests pass.

PR #1075, opened 2026-04-07
fix(hook): cursor hook fails to rewrite commands due to incorrect exit code handling

### LINKED ISSUE #1145: preToolUse hook broken: rtk rewrite exits code 3 on success, || fallback swallows all rewrites
## Bug

`rtk init -g --agent cursor` installs the following in `~/.cursor/hooks/rtk-rewrite.sh`:

```bash

## Root Cause

$ rtk rewrite "git status"; echo "exit: $?"
rtk git status
exit: 3

## Fix

REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null)
if [ -z "$REWRITTEN" ]; then
  echo '{}'
  exit 0
fi


## Environment
rtk 0.35.0
Cursor 3.0.13
macOS Apple Silicon

### PR DESCRIPTION
**Pull request by sebastiancichos**
## Summary

The Cursor agent hook (`~/.cursor/hooks/rtk-rewrite.sh`) fails to rewrite shell commands because it incorrectly handles the exit code from `rtk rewrite`. The hook treats exit code 3 (which indicates a successful rewrite) as a failure condition.

Fixes #1145.

### Current Behavior (Broken)
All shell commands pass through unchanged - RTK never rewrites them.

### Expected Behavior
Shell commands like `git status`, `ls`, `cat`, etc. should be automatically rewritten to their RTK equivalents.

## Root Cause

In `hooks/cursor/rtk-rewrite.sh` (line 43):

```bash
# Delegate all rewrite logic to the Rust binary.
# rtk rewrite exits 1 when there's no rewrite — hook passes through silently.
```

**The problem**: The comment says "exits 1 when there's no rewrite", but `rtk rewrite` actually returns **exit code 3** when it successfully rewrites a command.

The `||` operator treats ANY non-zero exit code as failure, so:
- Exit 0 (no rewrite needed) → OK, hook continues
- Exit 3 (successfully rewritten) → TREATED AS FAILURE, hook exits with `{}`
- Exit 1 (actual error) → TREATED AS FAILURE, hook exits with `{}`

**Result**: Commands are never rewritten because the hook always exits early.

## Evidence

Test showing `rtk rewrite` exit codes:

```bash
# Command that should be rewritten
$ rtk rewrite "git status"
rtk git status
$ echo $?  
3  # ← This is "success", not failure!

# Command with no rewrite needed  
$ rtk rewrite "unknown-command"
unknown-command
$ echo $?
0
```

Hook behavior with current code:
```bash
$ echo '{"tool_input": {"command": "git status"}}' | ~/.cursor/hooks/rtk-rewrite.sh
{}  # ← Always returns empty object, never rewrites
```

## Fix

Replace the incorrect exit code handling:

**Before:**
```bash
```

**After:**
```bash
# rtk rewrite returns exit 3 on successful rewrite, 0 when no rewrite needed.
REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null)
RTK_EXIT=$?

# Exit 0 = no rewrite needed, exit 3 = successfully rewritten
# Any other exit code indicates an actual error
if [ "$RTK_EXIT" -ne 0 ] && [ "$RTK_EXIT" -ne 3 ]; then
  echo '{}'
  exit 0
fi
```

## Testing

After applying the fix:

```bash
$ echo '{"tool_input": {"command": "git status"}}' | ~/.cursor/hooks/rtk-rewrite.sh
{
  "permission": "allow",
  "updated_input": {
    "command": "rtk git status"
  }
}
```

Hook now correctly rewrites commands and RTK tracking works:

```bash
$ rtk gain

Total commands:    1
Input tokens:      77
Output tokens:     76
Tokens saved:      1 (1.3%)
```

## Impact

**Without this fix:**
- Cursor users get 0% token savings from the auto-rewrite feature
- Hook appears installed but does nothing
- Users must manually prefix commands with `rtk`

**With this fix:**
- 60-90% token savings on shell commands as advertised
- Transparent command rewriting works as designed
- Full RTK adoption for Cursor users

## Follow-up (review feedback)

Incorporating [@CamilleGuillory](https://github.com/CamilleGuillory)'s suggestions on this PR:

- **Empty output guard** — if `rtk rewrite` exits 0 or 3 but prints nothing, the hook returns `{}` and does not emit malformed `updated_input`.
- **`rtk-hook-version: 2`** — bumps the Cursor hook script version so installs and tooling can distinguish this revision from the broken v1 line.
- **Comments before the `jq` block** — documents `rtk rewrite` exit codes (0 vs 3) and why the response uses `"permission": "allow"` for Cursor.

## Acknowledgments

Thanks to [@CamilleGuillory](https://github.com/CamilleGuillory) for the review and for noting the overlap with [#1100](https://github.com/rtk-ai/rtk/pull/1100). The follow-up commit on this branch includes a `Co-authored-by` trailer attributing the review-driven hardening and documentation.

## Checklist

- [x] Fix implemented in `hooks/cursor/rtk-rewrite.sh`
- [x] Tested locally with `rtk rewrite` commands
- [x] Verified hook output format matches Cursor requirements
- [x] Review follow-up: empty-output guard, hook version 2, jq/Cursor comments
- [x] Attribution: `Co-authored-by` for [@CamilleGuillory](https://github.com/CamilleGuillory) on the follow-up commit

### DIFF
--- a/hooks/cursor/rtk-rewrite.sh
+++ b/hooks/cursor/rtk-rewrite.sh
@@ -39,8 +39,16 @@ if [ -z "$CMD" ]; then
 fi
 
 # Delegate all rewrite logic to the Rust binary.
-# rtk rewrite exits 1 when there's no rewrite — hook passes through silently.
-REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null) || { echo '{}'; exit 0; }
+# rtk rewrite returns exit 3 on successful rewrite, 0 when no rewrite needed.
+REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null)
+RTK_EXIT=$?
+
+# Exit 0 = no rewrite needed, exit 3 = successfully rewritten
+# Any other exit code indicates an actual error
+if [ "$RTK_EXIT" -ne 0 ] && [ "$RTK_EXIT" -ne 3 ]; then
+  echo '{}'
+  exit 0
+fi
 
 # No change — nothing to do.
 if [ "$CMD" = "$REWRITTEN" ]; then
--- a/hooks/cursor/rtk-rewrite.sh
+++ b/hooks/cursor/rtk-rewrite.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# rtk-hook-version: 1
+# rtk-hook-version: 2
 # RTK Cursor Agent hook — rewrites shell commands to use rtk for token savings.
 # Works with both Cursor editor and cursor-cli (they share ~/.cursor/hooks.json).
 # Cursor preToolUse hook format: receives JSON on stdin, returns JSON on stdout.
@@ -50,11 +50,17 @@ if [ "$RTK_EXIT" -ne 0 ] && [ "$RTK_EXIT" -ne 3 ]; then
   exit 0
 fi
 
-# No change — nothing to do.
-if [ "$CMD" = "$REWRITTEN" ]; then
+# No change — nothing to do (same text, or empty stdout despite exit 0/3).
+if [ -z "$REWRITTEN" ] || [ "$CMD" = "$REWRITTEN" ]; then
   echo '{}'
   exit 0
 fi
+
+# Cursor expects JSON with "permission" and optional "updated_input".
+# rtk rewrite: exit 0 = no rewrite needed; exit 3 = rewritten successfully (RTK uses 3
+# as a success code, unlike typical shell conventions). Either way, when we emit
+# updated_input, Cursor requires "permission": "allow" so the tool runs with the
+# new command (rather than ask/deny/default pass-through).
 
 jq -n --arg cmd "$REWRITTEN" '{
   "permission": "allow",


sebastiancichos:
@aeppling Please review this PR to provide solution for Cursor users

CamilleGuillory:
Hi @sebastiancichos, nice catch — I opened #1100 for the same bug before seeing yours (now closed).

A few suggestions to make this more robust:

1. **Empty output guard** — add `[ -z "$REWRITTEN" ]` to the "no change" check, in case `rtk rewrite` returns exit 0/3 but produces empty output
2. **Version bump** — bump `rtk-hook-version` from `1` to `2` so users/tooling can detect the fixed version
3. **Comment on the jq block** — clarify that exit 0 = Allow and exit 3 = Ask/Default, and that Cursor requires `"permission": "allow"`

Happy to help if needed. Let me know!

sebastiancichos:
Hi @CamilleGuillory — thank you for the review and for calling out the overlap with #1100.

I have applied your three suggestions on this branch:

1. **Empty output guard** — the no-op path now treats empty `$REWRITTEN` the same as unchanged text, so we never emit `updated_input` with an empty command if `rtk rewrite` misbehaves.
2. **`rtk-hook-version`** — bumped from `1` to `2` on the Cursor hook so the fixed revision is easy to spot.
3. **Comments before the `jq` block** — short note on `rtk rewrite` exit `0` vs `3` and why the hook sets Cursor JSON `permission` to `allow` when returning `updated_input`.

The follow-up commit uses **`Co-authored-by: Camille Guillory <45099363+CamilleGuillory@users.noreply.github.com>`** so GitHub attributes the review-driven hardening to you. The PR description is updated with a **Follow-up** section, **Acknowledgments**, and checklist items.

Let me know if you would like any wording tweaks.

Eji4h:
Are any update?

eilone:
any update?

almogbhl:
any update?

OlegLustenko:
Our team is waiting for it

Eji4h:
I'm still waiting too.

KuSh:
Thanks for the detailed diagnosis. The exit-code handling in hooks/cursor/rtk-rewrite.sh was fixed on develop by commits ff0b9ac2 (2026-06-25) and a0c16ef0 (2026-06-29), which capture the return code of rtk rewrite and treat 3 as a successful rewrite rather than a failure, along with the guard against empty output that this PR also added. develop goes slightly further and maps exit 3 to a permission of ask. Closing as already covered, but please comment or reopen if you see a case the current hook still gets wrong.

PR #1076, opened 2026-04-07
fix(read): don't count --max-lines truncation as token savings (#1045)

### LINKED ISSUE #1045: rtk read: truncation via --max-lines incorrectly counted as token savings, producing 250M+ saved tokens in one command
## Bug                                                                                                                                          
                                                                                                                                               
  When rtk read truncates output via --max-lines the dropped content is recorded as "saved tokens" in the        
  tracking database.                                                                                                                                                                                                                                                         
  This inflates rtk gain numbers. In my case, a handful of large bioinformatics files (FASTA, SAM) pushed reported savings to 278M tokens — almost entirely fake.
  Real data from my history — a few bioinformatics file reads dominate the entire dashboard:
```
rtk gain --tier pro -q

Total commands:    92
Input tokens:      278.1M
Output tokens:     6.0K
Tokens saved:      278.1M (100.0%)
Total exec time:   25.6s (avg 277ms)
Efficiency meter: ████████████████████████ 100.0%

By Command
────────────────────────────────────────────────────────────────────────
  #  Command                   Count   Saved    Avg%    Time  Impact    
────────────────────────────────────────────────────────────────────────
 1.  rtk read                     17  269.2M   88.1%   150ms  ██████████
 2.  rtk curl -sL https://...      1    8.8M  100.0%    1.8s  ░░░░░░░░░░
 3.  rtk ls                       27    2.8K   65.5%     1ms  ░░░░░░░░░░
...
```
                                                                                                                        
                                                                                                                                               
##   Reproduction          
  Environment                                                                                                                                  
                                                                 
  - rtk 0.34.2                                                                                                                                 
 - Linux 6.12.0                         
 - Claude Code with PreToolUse hook (rtk-rewrite.sh v3)                                                                                                                         
  ```                                                                                                                                                                                                                                                                                        
 # 1. Create test file: 1000 lines, ~90KB                                                                                                     
 python3 -c "for i in range(1000): print(f'line {i}: ' + 'x'*80)" > /tmp/test.txt                                                             
 wc -c /tmp/test.txt   # 90890 bytes                                                                                                          
                                                                                                                                               
 # 2. Read with --max-lines                                                                                        
 rtk read --max-lines 1 /tmp/test.txt > /dev/null                                                                                             
                                                                                                                                               
 # 3. Check tracking DB                                         
 sqlite3 ~/.local/share/rtk/history.db \
    "SELECT input_tokens, output_tokens, saved_tokens, savings_pct
     FROM commands ORDER BY id DESC LIMIT 1;"                                                                                                  
  
 # Result:                                                                                                                                    
 #   input_tokens  = 22723    ← ceil(file_size / 4), from stat()
 #   output_tokens = 9        ← actual output after truncation                                                                                
 #   saved_tokens  = 22714    ← WRONG: this is discarded data, not savings                                                                    
 #   savings_pct   = 99.96    ← meaningless    
```                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                               
 ## Expected behavior
Does not count dropped data with --max-lines as saved, input_tokens should be derived from the content actually read

### PR DESCRIPTION
**Pull request by ousamabenyounes**
## Summary

- `rtk read file --max-lines N` was comparing the full raw file against the truncated output in `timer.track()`, inflating reported savings by the entire skipped portion — 250M+ tokens reported for a single command on large files
- Fix: capture the post-filter, pre-truncation content and use it as the tracking baseline when `--max-lines` or `--tail-lines` is active
- Same fix applied to `run_stdin()`

## Root cause

```rust
// BEFORE — bug
timer.track("cat file", "rtk read", &content, &rtk_output);
//                                   ^^^^^^^^   ^^^^^^^^^^
//                                   11MB file  5 lines → fake 2.8M token "savings"

// AFTER — fix
let tracking_output = filtered_for_tracking.as_deref().unwrap_or(&rtk_output);
timer.track("cat file", "rtk read", &content, tracking_output);
//                                             ^^^^^^^^^^^^^^
//                                             pre-truncation filtered content → real savings only
```

## Test plan

4 new regression tests in `src/cmds/system/read.rs`:

- [x] `test_max_lines_tracking_uses_pretruncation_content` — tracking baseline equals full filtered content, not truncated slice
- [x] `test_tail_lines_tracking_uses_pretruncation_content` — same invariant for `--tail-lines`
- [x] `test_no_truncation_tracking_unchanged` — without truncation flags, behaviour is identical to before
- [x] `test_savings_inflation_is_zero_with_fix` — numerical proof: buggy code reports >1k fake tokens, fix reports 0

```
cargo test: 1347 passed, 0 failed
```

Fixes #1045

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug Fixes
 
 * **git:** remove `-u` short alias from `--ultra-compact` to fix `git push -u` upstream tracking ([#1086](https://github.com/rtk-ai/rtk/issues/1086))
+* **read:** stop counting `--max-lines` / `--tail-lines` truncation as token savings ([#1045](https://github.com/rtk-ai/rtk/issues/1045)). Affected users will see lower (and accurate) numbers in `rtk gain` going forward.
 
 ## [0.35.0](https://github.com/rtk-ai/rtk/compare/v0.34.3...v0.35.0) (2026-04-06)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -243,6 +243,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Bug Fixes
 
 * **git:** remove `-u` short alias from `--ultra-compact` to fix `git push -u` upstream tracking ([#1086](https://github.com/rtk-ai/rtk/issues/1086))
+* **read:** stop counting `--max-lines` / `--tail-lines` truncation as token savings ([#1045](https://github.com/rtk-ai/rtk/issues/1045)). Affected users will see lower (and accurate) numbers in `rtk gain` going forward.
 
 ## [0.35.0](https://github.com/rtk-ai/rtk/compare/v0.34.3...v0.35.0) (2026-04-06)
 
--- a/src/cmds/system/read.rs
+++ b/src/cmds/system/read.rs
@@ -66,22 +66,27 @@ pub fn run(
     // Capture post-filter content before user-requested truncation.
     // --max-lines/--tail-lines is a user choice, not RTK compression — savings
     // are tracked against filtered (not truncated) output to avoid inflation.
-    let filtered_for_tracking = if max_lines.is_some() || tail_lines.is_some() {
-        Some(filtered.clone())
+    // mem::take avoids cloning the entire filtered buffer (can be large).
+    let pre_truncation: Option<String> = if max_lines.is_some() || tail_lines.is_some() {
+        Some(std::mem::take(&mut filtered))
     } else {
         None
     };
-
-    filtered = apply_line_window(&filtered, max_lines, tail_lines, &lang);
+    let windowed = apply_line_window(
+        pre_truncation.as_deref().unwrap_or(&filtered),
+        max_lines,
+        tail_lines,
+        &lang,
+    );
 
     let rtk_output = if line_numbers {
-        format_with_line_numbers(&filtered)
+        format_with_line_numbers(&windowed)
     } else {
-        filtered.clone()
+        windowed
     };
-    println!("{}", rtk_output);
+    print!("{}", rtk_output);
 
-    let tracking_output = filtered_for_tracking.as_deref().unwrap_or(&rtk_output);
+    let tracking_output = pick_tracking_output(pre_truncation.as_deref(), &rtk_output);
     timer.track(
         &format!("cat {}", file.display()),
         "rtk read",
@@ -138,22 +143,26 @@ pub fn run_stdin(
         );
     }
 
-    let filtered_for_tracking = if max_lines.is_some() || tail_lines.is_some() {
-        Some(filtered.clone())
+    let pre_truncation: Option<String> = if max_lines.is_some() || tail_lines.is_some() {
+        Some(std::mem::take(&mut filtered))
     } else {
         None
     };
-
-    filtered = apply_line_window(&filtered, max_lines, tail_lines, &lang);
+    let windowed = apply_line_window(
+        pre_truncation.as_deref().unwrap_or(&filtered),
+        max_lines,
+        tail_lines,
+        &lang,
+    );
 
     let rtk_output = if line_numbers {
-        format_with_line_numbers(&filtered)
+        format_with_line_numbers(&windowed)
     } else {
-        filtered.clone()
+        windowed
     };
     print!("{}", rtk_output);
 
-    let tracking_output = filtered_for_tracking.as_deref().unwrap_or(&rtk_output);
+    let tracking_output = pick_tracking_output(pre_truncation.as_deref(), &rtk_output);
     timer.track("cat - (stdin)", "rtk read -", &content, tracking_output);
     Ok(())
 }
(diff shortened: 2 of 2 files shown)


pszymkowiak (changes requested):
Thanks for tackling this — issue #1045 has been a real pain point in our telemetry data, and the diagnosis is spot-on. The core fix (capturing post-filter / pre-truncation content for tracking) is the right approach, well-bounded, and symmetrically applied to both `run()` and `run_stdin()`. The numerical proof test (`test_savings_inflation_is_zero_with_fix`) is a particularly nice touch.

A few things to address before merge — happy to discuss any of them.

## 🔴 Blocking — please split the PR

This PR currently contains 3 commits, only one of which relates to #1045:

- `a6dee12` feat(hooks): add dangerous-actions-blocker PreToolUse hook
- `db556a3` test(hooks): add test suite for dangerous-actions-blocker
- `65de749` fix(read): don't count --max-lines truncation as token savings

The dangerous-actions-blocker hook (102 lines + 115 lines of tests) is a meaningful security feature that deserves its own PR, its own review, and its own discussion. Could you `git rebase -i develop` to keep only `65de749` here, and open a separate PR for the hook? Happy to review that one too.

## 🟠 Tests — three of the four don't actually test the production code

`simulate_tracking()` re-implements the exact branching logic from `run()` (the `Some(filtered.clone())` if-condition, the `as_deref().unwrap_or(...)` fallback). The first three tests then assert against `simulate_tracking`, which means a regression in `run()` itself wouldn't fail them — they validate their own helper.

Two ways to fix this:

1. **Extract the selection into a tiny pure function** that both `run()` and the tests call:
   ```rust
   fn pick_tracking_output<'a>(pre_truncation: Option<&'a str>, rtk_output: &'a str) -> &'a str {
       pre_truncation.unwrap_or(rtk_output)
   }
   ```
   Then the tests exercise the real function instead of a copy.

2. **Or drive `run()` end-to-end** with a `tempfile::NamedTempFile`, then read back from the tracking DB and assert on `saved_tokens`. This is heavier but catches more.

The 4th test (`test_savings_inflation_is_zero_with_fix`) does call `estimate_tokens` from the production module, so it's already in good shape — just needs the same shift to exercising real code paths instead of `simulate_tracking`.

## 🟡 Minor — `print!` → `println!` change

```diff
-    print!(\"{}\", rtk_output);
+    println!(\"{}\", rtk_output);
```

This isn't mentioned in the description and silently adds a trailing newline to all `rtk read` output. It may break downstream pipes or scripts that count on the original behavior. Could you either revert it (orthogonal to #1045) or note the rationale in the commit message?

## 🟡 Minor — avoidable allocation

`Some(filtered.clone())` duplicates the entire filtered content in memory whenever truncation is active. For a power user running `rtk read --max-lines 1` on a 100 MB log file, that's 100 MB held twice for no reason. A `std::mem::take` (or restructuring to keep `filtered` immutable) would avoid the copy:

```rust
let pre_truncation = if max_lines.is_some() || tail_lines.is_some() {
    Some(std::mem::take(&mut filtered))
} else {
    None
};
let windowed = apply_line_window(
    pre_truncation.as_deref().unwrap_or(&filtered),
    max_lines, tail_lines, &lang,
);
// …
let tracking_output = pre_truncation.as_deref().unwrap_or(&windowed);
```

Not a blocker, but the bug-affected workload (large files + truncation) is exactly where this matters.

## 🟡 Note — same pattern likely exists in `git diff` / `git show`

`src/cmds/git/git.rs:200,314` track against the compacted output. There, `compact_diff` is real RTK compression so it's legitimate — **but** when the user passes `--max-lines N` to `rtk git diff`, the same inflation pattern applies. Worth opening a follow-up issue rather than expanding the scope here.

## 📋 Out of scope but worth flagging

- **Already-polluted historical data.** Devices that have been over-reporting for weeks (we have several reporting 4-72 G tokens/day in the API DB) will keep skewing lifetime totals until they re-ping with the fixed binary. Probably needs a server-side cap or a one-off migration in a follow-up.
- **CHANGELOG.md** isn't updated. Since this fix will visibly lower `rtk gain` numbers for affected users, it's worth a user-facing line.

---

Again, the diagnosis and the core fix are good — just need the scope split and the tests pointed at the real code path. Let me know if you'd like me to push the test refactor as a suggestion commit instead.

ousamabenyounes:
@pszymkowiak Thanks for the review — all points addressed. Summary of the final state:

## 🔴 PR scope
Rebased onto `develop`, dropped the dangerous-actions-blocker hook commits. Only the #1045 fix remains here. That hook has its own PR (#1008).

## 🟠 Tests exercise real production code
- Extracted `pick_tracking_output()` helper, called from both `run()` / `run_stdin()` and the regression tests — `simulate_tracking()` removed.
- Added `test_run_does_not_inflate_savings_with_max_lines_e2e`: points `RTK_DB_PATH` at a temp SQLite (serialized via static `Mutex` like `tracking.rs::test_db_path_env_and_default`), drives real `run(..., max_lines=Some(5), ...)` against a 1000-line file (`Language::Unknown` filter = identity, so any reported savings can only come from a tracking-baseline bug), reads back the persisted row and asserts `input_tokens ≈ output_tokens` and `saved_tokens ≈ 0`. Catches call-site regressions, not just helper-level ones.

## 🟡 `print!` → `println!`
Reverted in `run()` — unintentional drift in the original commit, orthogonal to #1045.

## 🟡 Avoidable allocation
Replaced `Some(filtered.clone())` with `std::mem::take(&mut filtered)` and plumbed through `pre_truncation.as_deref().unwrap_or(&filtered)`. No double-allocation on the bug-affected path.

## 🟡 Same pattern in `git diff` / `git show`
Tracked as follow-up: #1561.

## 📋 CHANGELOG
Added under `## [Unreleased]`:
> * **read:** stop counting `--max-lines` / `--tail-lines` truncation as token savings (#1045). Affected users will see lower (and accurate) numbers in `rtk gain` going forward.

`cargo fmt` ✅ · `cargo clippy --all-targets` ✅ · `cargo test --all`: **1692 passed, 0 failed**.

Ready for re-review when you have a moment.

PR #1083, opened 2026-04-08
fix(hook_check): skip hook warning on Windows when --claude-md mode is installed

SrHuevo:
## Summary

- On Windows, `rtk init -g` cannot install a PreToolUse hook — it falls back to `--claude-md` mode and writes the RTK instructions block into `~/.claude/CLAUDE.md`. That's the only valid integration on this platform.
- But `status()` still returns `Missing` in that configuration, so `maybe_warn()` keeps printing ``/!\ No hook installed — run `rtk init -g` for automatic token savings`` every 24h.
- The advice is impossible to follow: running `rtk init -g` on Windows just re-emits the `--claude-md` fallback message (`Hook-based mode requires Unix (macOS/Linux). Windows: use --claude-md mode`). The user is nagged forever about an integration they can never install.
- This PR teaches `status()` to recognize the `--claude-md` integration on Windows by looking for the `<!-- rtk-instructions` marker in `~/.claude/CLAUDE.md`, and return `HookStatus::Ok` in that case.
- The check is narrowly gated on `cfg!(windows)` to stay minimal and avoid reopening the broader "hook masking" discussion from d85fe33 — on macOS/Linux, behavior is unchanged.

## Reproduction (before this patch, on Windows)

```text
> rtk init -g --no-patch
[rtk] Hook-based mode requires Unix (macOS/Linux).
       Windows: use --claude-md mode (automatic fallback).
> rtk init -g --claude-md
[rtk] Installed RTK instructions into ~/.claude/CLAUDE.md
> rtk git status
...
```

After the patch, once `<!-- rtk-instructions` is present in `~/.claude/CLAUDE.md`, the warning stops — on Windows only.

## Test plan

- [x] New unit tests cover the three shapes of `claude_md_mode_installed`: marker present, marker absent, file missing.
- [x] `test_status_returns_valid_variant` extended with a Windows + `--claude-md` branch asserting `HookStatus::Ok`.
- [ ] `cargo fmt --all && cargo clippy --all-targets && cargo test` — I don't have a Rust toolchain on the Windows machine where I reproduced this, so relying on CI for the pre-commit gate.
- [ ] Manual testing on Windows: confirmed the warning fires in v0.35.0 with `~/.claude/CLAUDE.md` containing the RTK block; after this patch it should go silent.

### DIFF
--- a/src/hooks/hook_check.rs
+++ b/src/hooks/hook_check.rs
@@ -9,6 +9,8 @@ use std::path::PathBuf;
 
 const CURRENT_HOOK_VERSION: u8 = 3;
 const WARN_INTERVAL_SECS: u64 = 24 * 3600;
+const CLAUDE_MD: &str = "CLAUDE.md";
+const RTK_INSTRUCTIONS_MARKER: &str = "<!-- rtk-instructions";
 
 /// Hook status for diagnostics and `rtk gain`.
 #[derive(Debug, PartialEq, Clone)]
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the careful Windows reproduction. Hook based mode now works on Windows since PR #1399 (merged 2026-04-19), which dropped the Unix only guard and registers the native rtk hook in settings.json, so rtk init -g installs a real PreToolUse hook there and the "No hook installed" warning is both accurate and actionable. The claude-md fallback this patch recognized is no longer the only integration available on Windows, so the special case is not needed, and treating it as Ok would now hide the prompt to migrate to the hook. If you still see the warning after running rtk init -g on current develop, please comment and we can reopen.

PR #1100, opened 2026-04-09
fix(cursor): handle exit code 3 in rtk-rewrite.sh

### LINKED ISSUE #1099: Cursor hook fails to rewrite commands due to unhandled exit code 3
The generated `rtk-rewrite.sh` hook for Cursor fails to rewrite commands because it expects `rtk rewrite` to exit with code `0`. However, `rtk rewrite` exits with code `3` by default (when no explicit allow/deny rules are matched, falling back to Ask/Default).

Because the Cursor hook script uses:
```bash
```
The non-zero exit code causes the hook to abort and return `{}`, falling back to the original uncompressed command.

I have prepared a PR to fix this by explicitly handling exit code 3 in the Cursor hook script.

camguillory:
Fixes #1099

The Cursor `rtk-rewrite.sh` hook was failing to rewrite commands because `rtk rewrite` exits with code `3` by default (when no explicit allow/deny rules are matched). The hook script was using `|| { echo '{}'; exit 0; }`, which caused it to abort on any non-zero exit code.

This PR updates the Cursor hook to explicitly handle exit code `3` (Ask/Default) and exit code `0` (Allow) as successful rewrites, while ignoring exit codes `1` (No RTK equivalent) and `2` (Deny rule matched), similar to how the Claude Code hook handles it.

### DIFF
--- a/hooks/cursor/rtk-rewrite.sh
+++ b/hooks/cursor/rtk-rewrite.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# rtk-hook-version: 1
+# rtk-hook-version: 2
 # RTK Cursor Agent hook — rewrites shell commands to use rtk for token savings.
 # Works with both Cursor editor and cursor-cli (they share ~/.cursor/hooks.json).
 # Cursor preToolUse hook format: receives JSON on stdin, returns JSON on stdout.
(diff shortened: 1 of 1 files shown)


sebastiancichos:
that's duplicate of https://github.com/rtk-ai/rtk/pull/1075

KuSh:
Thanks for tracking this down. The Cursor hook on develop already handles the exit code 3 case: commits ff0b9ac2 (2026-06-25) and a0c16ef0 (2026-06-29) capture the return code, pass through on 1 and 2, accept 0 and 3 as a rewrite, guard against empty output, and additionally map code 3 to a Cursor ask permission. The empty rewrite guard you added is present there too. If you still see the Cursor hook dropping rewrites on the current develop, please comment here or reopen with the command and the hook output.

PR #1115, opened 2026-04-09
fix(dotnet): stop prepending raw stdout on build failure (#914)

### LINKED ISSUE #914: dotnet build can be more noisy with rtk
I created a simple dotnet console app with a deliberate compilation error:

```csharp
Console.Write("Enter your name: ");
string name = Console.ReadLine();
Console.WriteLine($"Hello, {name}!");
error
```

`dotnet build > output.txt`

```
  Determining projects to restore...
  All projects are up-to-date for restore.
E:\tempconsole\Program.cs(4,6): error CS1001: Identifier expected [E:\tempconsole\tempconsole.csproj]
E:\tempconsole\Program.cs(4,6): error CS1002: ; expected [E:\tempconsole\tempconsole.csproj]

Build FAILED.

E:\tempconsole\Program.cs(4,6): error CS1001: Identifier expected [E:\tempconsole\tempconsole.csproj]
E:\tempconsole\Program.cs(4,6): error CS1002: ; expected [E:\tempconsole\tempconsole.csproj]
    0 Warning(s)
    2 Error(s)

Time Elapsed 00:00:00.69
```
**215 tokens** 

`rtk dotnet build > output.txt`

```
Determining projects to restore...
  All projects are up-to-date for restore.
E:\tempconsole\Program.cs(4,6): error CS1001: Identifier expected [E:\tempconsole\tempconsole.csproj]
E:\tempconsole\Program.cs(4,6): error CS1002: ; expected [E:\tempconsole\tempconsole.csproj]

Build FAILED.

E:\tempconsole\Program.cs(4,6): error CS1001: Identifier expected [E:\tempconsole\tempconsole.csproj]
E:\tempconsole\Program.cs(4,6): error CS1002: ; expected [E:\tempconsole\tempconsole.csproj]
    0 Warning(s)
    2 Error(s)

Time Elapsed 00:00:00.63

fail dotnet build: 1 projects, 2 errors, 0 warnings (00:00:00.63)
---------------------------------------

Errors:
  E:\tempconsole\Program.cs(0,0) error CS1001: Identifier expected
  error ; expected
```

**280 tokens**

It looks like `rtk` produces the summary but also keeps the output. It only happens if the compilation fails. When the compilation succeeds, `rtk` only produces the summary, as expected.

ousamabenyounes:
## Summary

Fixes #914

- On `dotnet build` failure, RTK was prepending the full raw stdout before the filtered summary — producing **more tokens than the original command**
- The filtered summary from `format_build_output` already contains all errors and warnings; the raw output is redundant
- On success, only the filtered summary was shown (correct). This fix makes failure consistent with success

## Verification

- [x] Baseline tests: 1350 pass, 0 pre-existing failures
- [x] Post-fix tests: 1351 pass, 0 regressions
- [x] New test: `test_format_build_output_on_failure_is_more_concise_than_raw` verifies ≥60% token savings and error info preserved
- [x] Review agent: issue alignment verified

## Files changed

| File | Change |
|------|--------|
| `src/cmds/dotnet/dotnet_cmd.rs` | Remove raw-prepend block on failure; always use `filtered`; add savings test |

Generated by Claude Code
Vibe coded by ousamabenyounes

---

### DIFF
--- a/src/cmds/dotnet/dotnet_cmd.rs
+++ b/src/cmds/dotnet/dotnet_cmd.rs
@@ -213,27 +213,16 @@ fn run_dotnet_with_binlog(subcommand: &str, args: &[String], verbose: u8) -> Res
         _ => raw.clone(),
     };
 
-    let output_to_print = if !command_success {
-        let stdout_trimmed = result.stdout.trim();
-        let stderr_trimmed = result.stderr.trim();
-        if !stdout_trimmed.is_empty() {
-            format!("{}\n\n{}", stdout_trimmed, filtered)
-        } else if !stderr_trimmed.is_empty() {
-            format!("{}\n\n{}", stderr_trimmed, filtered)
-        } else {
-            filtered
-        }
-    } else {
-        filtered
-    };
-
-    println!("{}", output_to_print);
+    // Always use the filtered summary — it already includes errors and warnings.
+    // Prepending raw stdout on failure was producing more tokens than the original
+    // command output, defeating the purpose of RTK filtering (issue #914).
+    println!("{}", filtered);
 
     timer.track(
         &format!("dotnet {} {}", subcommand, args.join(" ")),
         &format!("rtk dotnet {} {}", subcommand, args.join(" ")),
         &raw,
-        &output_to_print,
+        &filtered,
     );
 
     cleanup_temp_file(&binlog_path);
@@ -1221,6 +1210,71 @@ mod tests {
         assert!(output.contains("dotnet build: 2 projects, 1 errors, 1 warnings"));
         assert!(output.contains("error CS0103"));
         assert!(output.contains("warning CS0219"));
+    }
+
+    #[test]
+    fn test_format_build_output_on_failure_is_more_concise_than_raw() {
+        // Regression test for issue #914: rtk dotnet build was producing MORE tokens than
+        // the original on failure because it prepended raw stdout before the filtered summary.
+        // The filtered summary already contains errors — raw output must NOT be prepended.
+        let summary = binlog::BuildSummary {
+            succeeded: false,
+            project_count: 1,
+            errors: vec![
+                binlog::BinlogIssue {
+                    code: "CS1001".to_string(),
+                    file: "Program.cs".to_string(),
+                    line: 4,
+                    column: 6,
+                    message: "Identifier expected".to_string(),
+                },
+                binlog::BinlogIssue {
+                    code: "CS1002".to_string(),
+                    file: "Program.cs".to_string(),
+                    line: 4,
+                    column: 6,
+                    message: "; expected".to_string(),
+                },
+            ],
+            warnings: vec![],
+            duration_text: Some("00:00:00.69".to_string()),
+        };
+        let filtered = format_build_output(&summary, Path::new("/tmp/build.binlog"));
+        // Simulated raw output — realistic dotnet build failure output (215 tokens per issue report)
+        let raw_stdout = "  Determining projects to restore...\n\
+              All projects are up-to-date for restore.\n\
+            Build started 4/9/2026 10:00:00 AM.\n\
+            1>Project \"tempconsole.csproj\" on node 1 (Build target(s)).\n\
+            1>CoreCompile:\n\
+            1>  C:\\Program Files\\dotnet\\sdk\\8.0.303\\Roslyn\\bincore\\csc.dll Program.cs\n\
+            Program.cs(4,6): error CS1001: Identifier expected [tempconsole.csproj]\n\
+            Program.cs(4,6): error CS1002: ; expected [tempconsole.csproj]\n\
+            1>Done Building Project \"tempconsole.csproj\" (Build target(s)) -- FAILED.\n\
+            Build FAILED.\n\
+            Program.cs(4,6): error CS1001: Identifier expected [tempconsole.csproj]\n\
+            Program.cs(4,6): error CS1002: ; expected [tempconsole.csproj]\n\
+                0 Warning(s)\n\
+                2 Error(s)\n\
+            Time Elapsed 00:00:00.69";
+
+        // The filtered summary must save >=60% tokens vs raw (project standard)
+        let count_tokens = |s: &str| s.split_whitespace().count();
+        let savings =
+            1.0 - (count_tokens(&filtered) as f64 / count_tokens(raw_stdout) as f64);
+        assert!(
+            savings >= 0.60,
+            "Expected >=60% token savings, got {:.1}%",
+            savings * 100.0
+        );
+        // The filtered summary must contain error info (not empty)
+        assert!(filtered.contains("CS1001"), "Filtered must include error code");
+        assert!(filtered.contains("fail dotnet build"), "Filtered must include status");
+        // Prepending raw to filtered would exceed raw alone — this is what we fixed
+        let buggy_output = format!("{}\n\n{}", raw_stdout.trim(), filtered);
+        assert!(
+            buggy_output.len() > raw_stdout.len(),
+            "Prepending raw+filtered is always longer than raw alone"
+        );
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


danielmarbach (approved):
Nice thanks for fixing this!

Originally I had this in there to be super safe in case of failures. but yeah this makes sense

F0rty-Tw0 (commented):
Reviewed end-to-end and ran a fresh sub-review against the repo's rust/cli-testing rules. The change is correct: `filtered` is already derived from raw text via `parse_build_from_text(&raw)`, and `format_build_output` always emits at least the verdict line, so it can never be empty when `command_success=false`. Combined with `DOTNET_CLI_UI_LANGUAGE` forcing English locale (line 120), the prepend was redundant insurance, not a fallback. @danielmarbach's approval as the original author of the removed block is the strongest signal you can get here.

Four non-blocking observations follow as inline comments. Green on three OSes since April 9 with original-author approval. Would be great to see this land.

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1223,6 +1212,71 @@ mod tests {
...
+            "Expected >=60% token savings, got {:.1}%",
+            savings * 100.0
+        );
+        // The filtered summary must contain error info (not empty)
+        assert!(filtered.contains("CS1001"), "Filtered must include error code");
+        assert!(filtered.contains("fail dotnet build"), "Filtered must include status");
+        // Prepending raw to filtered would exceed raw alone — this is what we fixed
+        let buggy_output = format!("{}\n\n{}", raw_stdout.trim(), filtered);
+        assert!(
+            buggy_output.len() > raw_stdout.len(),
```
F0rty-Tw0: This assertion is tautological for the inputs this test cares about. The `format!()` line above concatenates `raw_stdout.trim()`, a 2-char separator, and `filtered`. With `format_build_output` always emitting a non-empty summary and the test fixture having no trailing whitespace, `buggy_output.len()` will exceed `raw_stdout.len()` by construction, so this `.len()` check can never fail in practice.

The meaningful regression check is the >=60% `count_tokens()` savings assertion above. Either drop this `.len()` line, or rewrite it in tokens to actually demonstrate the #914 bug:

```rust
assert!(
    count_tokens(&buggy_output) > count_tokens(raw_stdout),
    "Prepending raw+filtered must exceed raw alone in tokens (the #914 regression)"
);
```

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -213,27 +213,16 @@ fn run_dotnet_with_binlog(subcommand: &str, args: &[String], verbose: u8) -> Res
...
-        }
-    } else {
-        filtered
-    };
-
-    println!("{}", output_to_print);
+    // Always use the filtered summary — it already includes errors and warnings.
+    // Prepending raw stdout on failure was producing more tokens than the original
+    // command output, defeating the purpose of RTK filtering (issue #914).
+    println!("{}", filtered);
```
F0rty-Tw0, line 214: The block this `println!` replaces was after the subcommand match, so the fix correctly applies to all three of `dotnet build`, `dotnet test`, and `dotnet restore`. Only `build` has explicit regression coverage in the new test. Worth either:

- a one-liner test for `test`/`restore` reusing the same fixture, or
- a PR-title bump to `fix(dotnet): stop prepending raw stdout on build/test/restore failure`

so the changelog reflects the full scope.

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -213,27 +213,16 @@ fn run_dotnet_with_binlog(subcommand: &str, args: &[String], verbose: u8) -> Res
...
-            format!("{}\n\n{}", stderr_trimmed, filtered)
-        } else {
-            filtered
-        }
-    } else {
-        filtered
-    };
-
-    println!("{}", output_to_print);
+    // Always use the filtered summary — it already includes errors and warnings.
```
F0rty-Tw0: This is true for `build`/`test`/`restore`, but the wildcard arm in the subcommand match (`_ => raw.clone()`) sets `filtered = raw` for unknown subcommands. Optional tweak so the comment matches the actual code path:

```rust
// for build/test/restore, filtered already includes errors and warnings.
```

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1223,6 +1212,71 @@ mod tests {
...
+                    column: 6,
+                    message: "; expected".to_string(),
+                },
+            ],
+            warnings: vec![],
+            duration_text: Some("00:00:00.69".to_string()),
+        };
+        let filtered = format_build_output(&summary, Path::new("/tmp/build.binlog"));
+        // Simulated raw output — realistic dotnet build failure output (215 tokens per issue report)
+        let raw_stdout = "  Determining projects to restore...\n\
```
F0rty-Tw0: `.claude/rules/cli-testing.md` calls for `include_str!("../tests/fixtures/...")` on real captured output, and explicitly flags hand-built synthetic strings as an anti-pattern. This literal is hand-built and short; a real `dotnet build` failure ships with ~10× more MSBuild boilerplate. Non-blocking, but a real captured fixture (e.g. `tests/fixtures/dotnet/build_failure_raw.txt`) would make the savings ratio less dependent on how compact the literal happens to be.

ousamabenyounes:
@F0rty-Tw0 Thanks for the thorough end-to-end review and the four inline observations — all addressed in `f3e841e` (force-pushed after a self-review caught a tautology in my first pass):

**1. Tautological `.len()` assertion (line 1275)** — replaced with the real RTK invariant rather than the token-form you suggested. The token suggestion `count_tokens(raw + filtered) > count_tokens(raw)` is mathematically true whenever `filtered` is non-empty (which `format_*_output` always guarantees), so it doesn't actually guard against the #914 regression class. The current assertion is:

```rust
assert!(
    count_tokens(&filtered) < count_tokens(raw_stdout),
    "filtered must compress vs raw — if violated, even emitting filtered alone \
     is worse than emitting raw, which is the #914 class of bug"
);
```

This fails if a future change makes filtered ≥ raw — exactly the regression class.

**2. Scope: only `build` had explicit regression coverage (line 219)** — added two new regression tests with the same real-invariant assertion:
- `test_format_test_output_on_failure_is_more_concise_than_raw`
- `test_format_restore_output_on_failure_is_more_concise_than_raw`

The shared `println!()` path is exercised on all three subcommands.

**3. Comment scope (line 216)** — tightened to `For build/test/restore, filtered already includes errors and warnings.` (the wildcard arm `_ => raw.clone()` sets `filtered = raw` for unknown subs, as you noted.)

**4. Hand-built synthetic fixture (line 1244)** — moved to two real fixture files:
- `tests/fixtures/dotnet/build_failure_raw.txt` — MSBuild boilerplate + CS1001/CS1002
- `tests/fixtures/dotnet/restore_failure_raw.txt` — nuget NU1101 (semantically correct for the restore code path, not a copy of build output)

`cargo fmt --all && cargo clippy --all-targets && cargo test --all`: 1690 passed, 0 failed.

ousamabenyounes:
Hey @aeppling — gentle nudge, all green and reviews addressed. Anything else before merge?

ousamabenyounes:
@aepling — rebased on `develop` + CI fix landed in `05154ce`. All checks green.

**RED → GREEN proof** (develop now emits the status line LAST in `format_restore_output` after #1574 — verdict comes after Errors/Warnings sections, not before):

```
# BEFORE fix (parent commit 3c92925)
thread 'test_format_restore_output_on_failure_is_more_concise_than_raw' panicked at
src/cmds/dotnet/dotnet_cmd.rs:1516:9:
assertion failed: filtered.starts_with("fail dotnet restore")
test result: FAILED. 1930 passed; 1 failed

# AFTER fix (05154ce) — assertion now checks last_line.starts_with(…)
test result: ok. 5 passed; 0 failed (all test_format_restore_output_* tests)
```

Matches the convention already used by `test_format_build_output_status_line_is_last` on develop.

Ready for re-review 🙏

PR #1117, opened 2026-04-09
fix: augment PATH in hook to fix silent failure on Homebrew installs (#685)

### LINKED ISSUE #685: Hook silently fails when rtk not in Claude Code's restricted PATH (Homebrew installs)
## Problem

When RTK is installed via Homebrew, the `rtk-rewrite.sh` hook silently stops working because Claude Code runs hooks with a **restricted PATH** that does not include `/opt/homebrew/bin`.

The hook does:
```bash
if ! command -v rtk &>/dev/null; then
  exit 0  # silent failure — no rewrite happens, no warning
fi
```

This means **all Bash commands pass through unmodified** with zero indication anything is wrong. `rtk gain` still works fine from the terminal (because the user's shell has Homebrew in PATH), so the breakage is invisible.

## How it breaks

1. User installs RTK via `cargo install rtk` → binary at `~/.cargo/bin/rtk`
2. Hook works (cargo bin may be in PATH, or user never notices)
3. User switches to `brew install rtk` → binary moves to `/opt/homebrew/bin/rtk`
4. Hook silently fails — `command -v rtk` returns nothing in Claude Code's hook environment
5. `rtk gain --history` stops updating, but user assumes hook is fine

## Verification

You can reproduce by running the hook with a clean environment:
```bash
env -i HOME="$HOME" echo '{"tool_input":{"command":"git status"}}' | env -i HOME="$HOME" bash ~/.claude/hooks/rtk-rewrite.sh
# Output: (nothing) — exits 0 silently, no rewrite
```

## Root cause

`rtk init -g` installs a hook that uses `rtk` by name and relies on it being in PATH. But Claude Code's hook subprocess does not inherit the user's full shell PATH (no Homebrew, no cargo, etc. on macOS).

## Suggested fixes

**Option A (best):** `rtk init -g` should inject the **absolute path** to the current binary into the hook at install time:
```bash
# Generated by rtk init -g
RTK_BIN="/opt/homebrew/bin/rtk"  # resolved at install time
```

**Option B:** The hook should probe known install locations as fallback:
```bash
export PATH="/opt/homebrew/bin:/usr/local/bin:$HOME/.cargo/bin:$PATH"
```

**Option C:** Emit a visible warning (not just `exit 0`) when `rtk` is not found — at minimum `>&2 echo "[rtk] WARNING: rtk not found in PATH, hook disabled"` so users know something is wrong.

Note: Option C has already been partially addressed (warnings added to the hook), but the silent failure on Homebrew installs is the main UX problem.

## Environment

- macOS (Apple Silicon) — Homebrew installs to `/opt/homebrew/bin`
- RTK installed via `brew install rtk`
- Claude Code hook PATH does not include `/opt/homebrew/bin`

## Workaround

In `~/.claude/settings.json`, change the hook command from:
```
/Users/<you>/.claude/hooks/rtk-rewrite.sh
```
to:
```
PATH="/opt/homebrew/bin:/usr/local/bin:$PATH" bash /Users/<you>/.claude/hooks/rtk-rewrite.sh
```

ousamabenyounes:
## Summary

Fixes #685

- Claude Code hooks run with a restricted `PATH` that excludes Homebrew (`/opt/homebrew/bin` on Apple Silicon, `/usr/local/bin` on Intel) and cargo (`~/.cargo/bin`) directories
- `command -v rtk` silently returns non-zero, causing the hook to print a warning and exit — no commands are rewritten, token savings are lost silently
- Fix adds `export PATH="$HOME/.cargo/bin:/opt/homebrew/bin:/usr/local/bin:$PATH"` at the top of the hook, before the `command -v rtk` guard
- Hook version bumped from 3 → 4; existing installations will be upgraded on the next `rtk init -g`

## Verification

- [x] Baseline tests: 1351 pass, 6 ignored, 0 pre-existing failures
- [x] Post-fix tests: no regressions
- [x] New test: `test_hook_augments_path_before_rtk_check` — asserts PATH export exists and precedes the rtk guard, and that `/opt/homebrew/bin` and `.cargo/bin` are covered
- [x] Review agent: issue alignment verified

## Files changed

| File | Change |
|------|--------|
| `hooks/claude/rtk-rewrite.sh` | Bump version 3→4, add PATH augmentation (6 lines) |
| `src/hooks/hook_check.rs` | Update `CURRENT_HOOK_VERSION` constant to 4 |
| `src/hooks/init.rs` | Add `test_hook_augments_path_before_rtk_check` test |

Generated by Claude Code
Vibe coded by ousamabenyounes

---

### DIFF
--- a/hooks/claude/rtk-rewrite.sh
+++ b/hooks/claude/rtk-rewrite.sh
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
-# rtk-hook-version: 3
+# rtk-hook-version: 4
+# Augment PATH so rtk is found in Claude Code's restricted hook environment.
+# Common locations: Homebrew (Apple Silicon /opt/homebrew/bin, Intel /usr/local/bin)
+# and cargo installs ($HOME/.cargo/bin). Safe to add even if already in PATH.
+export PATH="$HOME/.cargo/bin:/opt/homebrew/bin:/usr/local/bin:$PATH"
 # RTK Claude Code hook — rewrites commands to use rtk for token savings.
 # Requires: rtk >= 0.23.0, jq
 #
--- a/src/hooks/hook_check.rs
+++ b/src/hooks/hook_check.rs
@@ -9,7 +9,7 @@ use super::constants::{CODEX_DIR, CURSOR_DIR, GEMINI_DIR, GEMINI_HOOK_FILE, OPEN
 use crate::core::constants::RTK_DATA_DIR;
 use std::path::PathBuf;
 
-const CURRENT_HOOK_VERSION: u8 = 3;
+const CURRENT_HOOK_VERSION: u8 = 4;
 const WARN_INTERVAL_SECS: u64 = 24 * 3600;
 
 /// Hook status for diagnostics and `rtk gain`.
--- a/src/hooks/hook_check.rs
+++ b/src/hooks/hook_check.rs
@@ -7,7 +7,7 @@ use super::constants::{
 use crate::core::constants::RTK_DATA_DIR;
 use std::path::PathBuf;
 
-const CURRENT_HOOK_VERSION: u8 = 3;
+const CURRENT_HOOK_VERSION: u8 = 4;
 const WARN_INTERVAL_SECS: u64 = 24 * 3600;
 
 /// Hook status for diagnostics and `rtk gain`.
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -2729,6 +2729,32 @@ mod tests {
     }
 
     #[test]
+    fn test_hook_augments_path_before_rtk_check() {
+        // PATH augmentation must exist so that rtk is found in Claude Code's restricted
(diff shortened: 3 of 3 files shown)


ousamabenyounes:
@aepling — rebased on `develop` + CI fix landed in `bccbabf`. All checks green.

**RED → GREEN proof** (inline `REWRITE_HOOK` const was removed on develop in favor of the external `hooks/claude/rtk-rewrite.sh` file):

```
# BEFORE fix (parent commit 0840839)
error[E0425]: cannot find value `REWRITE_HOOK` in this scope
  --> src/hooks/init.rs:3990:13
  (×5 occurrences in test_hook_augments_path_before_rtk_check)

# AFTER fix (bccbabf) — assertion now reads the script via include_str!
test hooks::init::tests::test_hook_augments_path_before_rtk_check ... ok
```

Same invariants asserted: PATH export precedes `command -v rtk` guard, `.cargo/bin` + `/opt/homebrew/bin` present.

Ready for re-review 🙏

PR #1121, opened 2026-04-10
feat(read): add --offset and --limit for reading specific line ranges

### LINKED ISSUE #1104: `rtk read` does not support reading specific lines
I instruct Claude Code to always use `rtk read`, `rtk grep` etc instead of the built-in read tool, to make sure it doesn't simply default to using its own built-in tools and thus not save any tokens. The problem is that the Claude Code built-in read tool supports reading a specific range of line numbers, so it tries to e.g. run `rtk read <file> <start_line> <end_line>`, which first returns the full file contents and then throws two errors. It would be nice if `rtk read` would support reading a range of line numbers, so that I don't need to explcitly instruct the LLM to use its built-in reading tool if and only if it reads a specific range.

I'm only a newcomer to rtk.ai so if a clean solution already exists, please let me know and I will close this issue!

sebastianbreguel:
## Summary
- Add `--offset` (1-based start line) and `--limit` (number of lines) options to `rtk read`
- Enables `rtk read file --offset 10 --limit 20` to read lines 10-29
- Matches Claude Code's Read tool semantics so LLMs can use `rtk read` for line-range reads

Closes #1104

## Test plan
- [x] `cargo test cmds::system::read` — 6 tests pass (including new offset/limit test)
- [x] `cargo check` — compiles clean
- [x] Existing tests unaffected

### DIFF
--- a/src/cmds/system/read.rs
+++ b/src/cmds/system/read.rs
@@ -12,6 +12,8 @@ pub fn run(
     max_lines: Option<usize>,
     tail_lines: Option<usize>,
     line_numbers: bool,
+    offset: Option<usize>,
+    limit: Option<usize>,
     verbose: u8,
 ) -> Result<()> {
     let timer = tracking::TimedExecution::start();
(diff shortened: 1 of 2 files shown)


sebastianbreguel:
@pszymkowiak could you review this?

jneeven:
Amazing!

PR #1122, opened 2026-04-10
fix(gh): don't early exit on failure for "pr checks" 

carlosdagos:
## Summary

if there are failing checks in the PR, then `gh` will return a status of `1` and `rtk` will refuse to parse the output due to using `early_exit_on_failure`.

remove the use of `early_exit_on_failure` so it can still parse the output.

now if there is an error, we will still get the summarised output with all `0`s, and we'll also pass through the stderr.

example if a repository does not exist:

```
CI Checks Summary:
  [ok] Passed: 0
  [FAIL] Failed: 0
GraphQL: Could not resolve to a Repository with the name 'some-org/some-repo'. (repository)
```

with the error passed through to the output regardless, the error handling downstream by the LLM should still work well.

## Test plan

### DIFF
--- a/src/cmds/git/gh_cmd.rs
+++ b/src/cmds/git/gh_cmd.rs
@@ -418,9 +418,7 @@ fn pr_checks(args: &[String], _verbose: u8, _ultra_compact: bool) -> Result<i32>
         "gh",
         &format!("pr checks {}", pr_number),
         format_pr_checks,
-        RunOptions::stdout_only()
-            .early_exit_on_failure()
-            .no_trailing_newline(),
+        RunOptions::stdout_only().no_trailing_newline(),
     )
 }
 


lucashungaro:
+1 on merging this. It currently makes it so agents can't use `gh pr checks`:

CleanShot 2026-04-16 at 15 10 46@2x

lucashungaro:
Bump 🆙

KuSh (approved):
## Round 1 of 3 — APPROVED

**Claim**: stop skipping the `gh pr checks` filter when `gh` exits non-zero, so failing and pending CI gets summarised instead of dumped raw.
**Scope**: accept (round 1, frozen)
**Ran**: real `gh` 2.46 against PRs 3982 (red) / 4010 (green) / 99999 / a bad repo slug / `--required` / `--watch`, develop `0cd3eba8` vs merged. Red PR: 1432 B raw → 255 B, **0% → 83%** savings, exit 1 preserved. Green path and every error path byte-identical to develop. Floor is 20% (CONTRIBUTING.md).
**CI** @`d1052271`: all 11 green, including ubuntu / macos / windows.

### Blocks merge (0 — closed by fixup `d1052271`)

#3542 landing fixed two of the four problems this turned up: the `pass`-in-description misclassification, and the `--watch` count inflation (#3541, now closed). One residual survived, and it was a regression this PR introduced: `parse_pr_check_line` mapped every bucket other than `pass`/`fail`/`pending` to `_ => return None`, so a cancelled run — the normal outcome when a new push supersedes CI — summarised to `Passed: 0 / Failed: 0` with every row dropped, where develop had shown the raw table. That is #4021.

Fixup `d1052271` counts the remaining buckets and reports them, plus five regression tests. Real PR 3982 now accounts for all 15 checks (8 passed / 2 failed / 5 skipped) at 81%. Mutation-checked both ways: reverting your one-line change fails 3 of the tests, reverting the bucket fix fails exactly the cancelled-run test.

### Filed as follow-ups
- #4025 — `gh run view --exit-status` loses its filter the same way (250 B raw / exit 1 with the flag, 26 B filtered without it). Outside this PR's hunks.
- #4021 — fixed by the fixup here; it will close when this merges.

Exit-code propagation (1 and 8); stderr forwarding; the error paths; the green path.

### Note on the PR body
The downside you flagged — "we will still get the summarised output with all `0`s" — does not actually happen. `guard::never_worse` returns the empty raw stdout, so a failed lookup is byte-identical to raw `gh`, stderr and exit code included. Worth correcting if you touch the body.

### Next
Nothing — approving.

Apologies for the five-month wait, and for the two CI churns on your branch just now: my first fixup push carried a stray committer identity from a scratch worktree, which is what briefly knocked the CLA check sideways. Both are sorted, and this is the first time CI has actually built the branch.

PR #1175, opened 2026-04-10
feat(ls): add --metadata flag to preserve permissions, sizes, and dates

### LINKED ISSUE #1162: enhancement: ls filter removes file metadata (permissions, sizes, dates) needed by agents
## Problem

\`rtk ls\` converts standard ls output into a compact tree view that strips file permissions, sizes, modification dates, and other metadata. AI agents often need this metadata to:

1. **Diagnose permission issues**: Is the file executable? Who owns it?
2. **Identify stale files**: When was this file last modified?
3. **Check file sizes**: Is this a large binary or a small config?
4. **Understand symlinks**: Is this a real file or a symlink?

### Example

\`\`\`bash
$ ls -la
-rw-r--r--  1 user staff  1234 Jan 15 10:30 config.json
lrwxr-xr-x  1 user staff    12 Jan 10 09:00 link -> /other/path
drwxr-xr-x  8 user staff   256 Jan 14 14:22 src

$ rtk ls -la
my-project/
+-- config.json
+-- link
+-- src/ (8 files)
\`\`\`

All metadata is lost.

## Proposed fix

1. **Add a --metadata or -l flag** that preserves permissions, sizes, and dates in the tree output:
   \`\`\`
   my-project/
   +-- config.json (1.2K, Jan 15, -rw-r--r--)
   +-- link -> /other/path (Jan 10, lrwxr-xr-x)
   +-- src/ (8 files, 256B, Jan 14, drwxr-xr-x)
   \`\`\`

2. **Detect when -la or -l flags are used** and include metadata automatically (the flags are passed through but the filter strips the raw output)

3. **Add an env var** like \`RTK_LS_METADATA=1\` that agents can set to always include metadata

## Relevant code

- src/cmds/system/ls.rs: ls filtering implementation
- src/discover/rules.rs line 92: ls rewrite rule

iRonin:
## Summary

Add `--metadata` (`-m`) flag to the `ls` command that preserves file permissions, sizes, dates, and other metadata in the compact output.

## Changes

- **`src/main.rs`**: Added `--metadata` (`-m`) flag to the Ls command
- **`src/cmds/system/ls.rs`**: 
  - Extended `compact_ls` to parse and store permissions, dates, and sizes
  - When metadata mode is on, each entry appends metadata like `(1.2K, Jan 15, -rw-r--r--)`
  - Auto-detects `-l` flag in args and enables metadata mode automatically
  - Added 5 new tests for metadata functionality

## Example

```
# Normal mode (default)
rtk ls
src/
Cargo.toml  1.2K
README.md   5.5K

# Metadata mode
rtk ls -m
src/  (64B, Jan 1 12:00, drwxr-xr-x)
Cargo.toml  1.2K  (Jan 1 12:00, -rw-r--r--)
README.md   5.5K  (Jan 1 12:00, -rw-r--r--)

# Auto-enabled with -l
rtk ls -la  # automatically uses metadata mode
```

Fixes #1162

### DIFF
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -4,24 +4,19 @@ use super::constants::NOISE_DIRS;
 use crate::core::runner::{self, RunOptions};
 use crate::core::utils::resolved_command;
 use anyhow::Result;
-use lazy_static::lazy_static;
-use regex::Regex;
 use std::io::IsTerminal;
 
-lazy_static! {
-    /// Matches the date+time portion in `ls -la` output, which serves as a
-    /// stable anchor regardless of owner/group column width.
-    /// E.g.: " Mar 31 16:18 " or " Dec 25  2024 "
-    static ref LS_DATE_RE: Regex = Regex::new(
-        r"\s+(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s+\d{1,2}\s+(?:\d{4}|\d{2}:\d{2})\s+"
-    )
-    .unwrap();
-}
-
-pub fn run(args: &[String], verbose: u8) -> Result<i32> {
+pub fn run(args: &[String], verbose: u8, metadata: bool) -> Result<i32> {
     let show_all = args
         .iter()
         .any(|a| (a.starts_with('-') && !a.starts_with("--") && a.contains('a')) || a == "--all");
+
+    // Auto-detect -l flag and enable metadata mode automatically
+    let has_long = args.iter().any(|a| a == "-l" || a == "--long" || a == "--format=long");
+    let has_short_l = args
+        .iter()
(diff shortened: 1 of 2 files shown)


KuSh:
Thanks for this. Metadata in `rtk ls` landed on develop through #1675 on 2026-05-25, which detects the long listing automatically from -l and also from -g, -n, -o, --full-time and --format=long or verbose, and renders permissions as octal next to the size, so the explicit -m flag is no longer needed. One thing this PR had that develop does not is the modification date in the compact listing, and a small PR adding the date to the existing long mode would be welcome. Worth noting the parsing here goes back to fixed whitespace columns, which is what #1675 replaced with a date-anchored parser so that owner and group names containing spaces do not shift the size and filename fields. If you think part of this was missed, please comment or reopen.

PR #1180, opened 2026-04-10
fix(rewrite): handle tail/head -n<N> joined flag form

### LINKED ISSUE #1108: Claude Code hook: `tail -N` rewrite produces invalid `rtk read -N` (should be `-n N`)
## Summary

The Claude Code Bash tool hook that transparently rewrites `tail <file>` to `rtk read <file>` mishandles the short numeric flag form `tail -N`. It produces `rtk read -N`, which is not a valid rtk `read` invocation — rtk's `read` subcommand expects `-n N` (separate flag + value).

## Environment

- Claude Code hook: the one that transparently rewrites common CLI commands (`git`, `tail`, `head`, etc.) to `rtk <cmd>` for token savings

## Reproduction

Inside a Claude Code session with the rtk hook active, run any of these:

```bash
tail -20 /some/file
tail -5 /some/file
tail -100 /var/log/foo.log
```

Observed: the hook rewrites these to `rtk read -20 /some/file` (or -5, -100), which rtk rejects as an invalid argument.

Expected: the hook should rewrite to `rtk read -n 20 /some/file` (or the rtk equivalent of "last N lines"), separating the short flag letter from the numeric value.

## Root cause (speculative)

GNU/BSD `tail` accepts the combined form `tail -20 file` (where `-20` means "last 20 lines") as a legacy shorthand for `tail -n 20 file`. The hook's rewrite regex likely copies the argument verbatim into rtk's argv, but rtk's argument parser does not accept `-20` as a valid flag — it expects `-n 20`.

## Suggested fix

In the hook's rewrite logic for `tail`, detect the combined-short-form `-<digits>` pattern and normalize it to `-n <digits>` before inserting the `rtk read` prefix. Or: normalize to `-n <digits>` as part of a general `tail` argv canonicalization step, independent of the rtk substitution.

## Impact

Low severity in isolation (user sees an rtk error and falls back to manual invocation), but it's a papercut that happens every time a user types a legacy-form `tail -N`. The failure mode is "command errors" rather than "command silently does the wrong thing," which is the right failure mode — but fixing it would eliminate the papercut.

1llum1n4t1s:
## Summary
- Fix `tail -n20 file` (no space between `-n` and number) not being rewritten to `rtk read file --tail-lines 20`
- Add missing `head -n 20`, `head -n20`, and `head --lines 20` rewrite patterns (only `head -20` and `head --lines=20` were supported)
- Add 4 new tests covering all new patterns

## Root Cause
The `TAIL_N_SPACE` regex required a mandatory space between `-n` and the digit (`\s+`), so `tail -n20` fell through unmatched. For `head`, the `-n` and `--lines <space>` forms were never implemented at all.

## Test plan
- [x] `cargo test test_rewrite_tail` — 7 tests pass
- [x] `cargo test test_rewrite_head` — 7 tests pass
- [x] `cargo fmt && cargo clippy --all-targets && cargo test --all` — 1357 tests pass, 0 failures
- [x] End-to-end binary verification of all tail/head flag forms

Closes #1108

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -67,10 +67,14 @@ lazy_static! {
     // invocations like `head -3 a b c` fail to match so the segment is passed through
     // to the native `head`/`tail` binary — which already handles multi-file with
     // `==> name <==` banners that `rtk read --max-lines` cannot reproduce.
+    // Issue #1108: also accept the joined `-n<N>` form (`-n\s*`) and the
+    // `head -n` / `head --lines <space>` variants.
     static ref HEAD_N: Regex = Regex::new(r"^head\s+-(\d+)\s+(\S+)$").unwrap();
+    static ref HEAD_N_FLAG: Regex = Regex::new(r"^head\s+-n\s*(\d+)\s+(\S+)$").unwrap();
     static ref HEAD_LINES: Regex = Regex::new(r"^head\s+--lines=(\d+)\s+(\S+)$").unwrap();
+    static ref HEAD_LINES_SPACE: Regex = Regex::new(r"^head\s+--lines\s+(\d+)\s+(\S+)$").unwrap();
     static ref TAIL_N: Regex = Regex::new(r"^tail\s+-(\d+)\s+(\S+)$").unwrap();
-    static ref TAIL_N_SPACE: Regex = Regex::new(r"^tail\s+-n\s+(\d+)\s+(\S+)$").unwrap();
+    static ref TAIL_N_FLAG: Regex = Regex::new(r"^tail\s+-n\s*(\d+)\s+(\S+)$").unwrap();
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes tail and head command rewrite patterns to handle the joined flag form (e.g., `tail -n20` without a space) and adds missing `head -n` and `head --lines <space>` rewrite variants. The TAIL_N_SPACE regex is updated to use `\s*` instead of `\s+`, and three new regexes are added for head patterns. Four new tests cover all added patterns.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1108

---

1llum1n4t1s:
Closing this PR as it doesn't look like it'll be merged. Feel free to reopen if there's interest. Thanks! 🙏

PR #1181, opened 2026-04-10
fix(telemetry): RGPD-compliant, consent gate, erasure, privacy controls

aeppling:


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🔴 **Risk** | `high` |

### Summary

This PR makes the telemetry system GDPR/RGPD-compliant by changing telemetry from opt-out (enabled by default) to opt-in (requiring explicit consent). It adds consent tracking fields, a 'forget' command for data erasure (GDPR Art. 17), removes personally identifiable data (hostname/username) from device hashing, and documents user rights, retention policies, and erasure procedures.

### Review Checklist
- [x] Tests present
- [x] Breaking change
- [x] Docs updated

---

pszymkowiak:
Hey @adrien — nice work on the opt-in migration, the architecture is clean and the RGPD doc is solid. A few things to address before merge:

## P0 — Blockers

**1. Consent nagging (init.rs ~L450-460)**
If the user declined (`consent_given = Some(false)`), re-asking every 14 days violates GDPR Art. 7(3). CNIL/EDPB explicitly sanction this pattern. Remove the re-ask mechanism — a "no" should be final.

**2. `rtk telemetry forget` doesn't purge tracking.db**
The local command history stays intact after `forget`. Art. 17 right to erasure applies here too. Either drop/truncate the telemetry-related tables in tracking.db, or document explicitly that the user must delete it manually.

**3. `send_erasure_request` missing auth token**
`maybe_ping()` sends `RTK_TELEMETRY_TOKEN` via `Authorization` header, but the erasure request doesn't. This needs to be aligned with the server side (rtk-ai/rtk-telemetry#3) — both PRs should use the same auth mechanism.

## P1 — Should fix

**4. Race condition in `run_forget`** — if the process is interrupted between salt deletion and the erasure HTTP call, the device hash is lost forever. The truncated hash displayed (`8...8` chars) is useless for support. Show the full hash in the fallback message.

**5. `prompt_telemetry_consent` displays 8 lines before checking `is_terminal()`** — in CI/CD, this prints noise to stderr before defaulting to N. Move the terminal check before the prompt output.

**6. Zero tests on `telemetry_cmd.rs`** (165 new lines) — at minimum: test that `maybe_ping()` is no-op when `consent_given` is `None` or `Some(false)`, and a unit test for `send_erasure_request`.

## Cross-PR note

This PR and rtk-ai/rtk-telemetry#3 need to be aligned on erasure auth: client must send the token, server must require it. Currently neither side does auth on `/erasure`.

pszymkowiak (changes requested):
## Review — fix(telemetry): RGPD-compliant

| P0 | P1 | P2 |
|:--:|:--:|:--:|
| 2  | 3  | 2  |

### P0

**1. Non-terminal `rtk init` permanently writes `consent_given = false`**

In `prompt_telemetry_consent()`, the non-terminal path calls `save_telemetry_consent(false)` — this permanently records a rejection in `config.toml`. A user running `rtk init` in CI/Docker/pipe is silently opted-out forever. Next interactive `rtk init` sees `Some(false)` and never prompts.

Fix: don't write anything in non-terminal context — just return. The user will be prompted on their next interactive `rtk init`.

```rust
if !io::stdin().is_terminal() {
    return Ok(());  // no consent recorded
}
```

**2. Double `Config::load()` per command**

`maybe_ping()` calls `config::telemetry_consent()` then `config::telemetry_enabled()` — two `Config::load()` calls (2x filesystem I/O on every command). For a <10ms startup target this matters. The old `telemetry_enabled()` check is now redundant since `consent_given = Some(false)` already implies disabled.

Fix: single `Config::load()`, check both fields.

### P1

**3.** `send_erasure_request` chains `.timeout()` on Request — verify ureq version compatibility (v1 vs v2 API differs).

**4.** `run_status()` uses `Config::load().unwrap_or_default()` — silently shows wrong consent state if config is corrupted. Should warn.

**5.** `consent_date: Option<String>` — no RFC 3339 validation. For GDPR Art. 7(1) audit, a malformed date undermines the record.

### P2

- `generate_device_hash()` hashes a 32-byte random salt with SHA-256 — the SHA step adds no entropy (vestigial)
- Double `Config::load()` should be merged into one call

aeppling:
Cargo.toml pins ureq = "2", semver won't resolve to 3.x. The chained API (post().set().timeout().send_string()) is the correct ureq 2.x API. No action needed unless we decide to upgrade to ureq 3, which would be a separate migration.

Intentional unwrap_or_default. Config::load().unwrap_or_default() means a corrupted config.toml (user typo, encoding issue) falls back to safe defaults instead of crashing the CLI. Since RTK is a transparent proxy, a crash on config parse would block all shell commands

pszymkowiak (approved):
All P0 and P1 items from my earlier review are addressed:

**P0:**
- ✅ Consent nagging removed — `Some(false)` now returns immediately, no 14-day re-ask
- ✅ `run_forget` deletes the local tracking database (`history.db`)
- ✅ `send_erasure_request` now sends the `X-RTK-Token` header (aligned with the server side in rtk-ai/rtk-telemetry#3)

**P1:**
- ✅ Full device hash shown in erasure fallback message (not truncated) — users can email support with the complete hash
- ✅ `is_terminal()` check moved before prompt output — no more stderr noise in CI/non-interactive mode

**Remaining:** no unit tests on `telemetry_cmd.rs` (~165 lines). Not a blocker for this PR since the code was manually verified, but tracked as #1254 for follow-up given the GDPR sensitivity.

LGTM.

PR #1186, opened 2026-04-10
fix(discover): expand npm rewrite rule to match install and other subcommands

### LINKED ISSUE #1148: bug: npm rule in rewrite registry misses install — filter already exists in npm_cmd.rs
## Bug

The npm rewrite rule in `src/discover/rules.rs` only matches `run|exec`:

```rust
pattern: r"^npm\s+(run|exec)",
```

This means `npm install` (and `npm ci`, `npm list`, `npm outdated`, etc.) are never rewritten by the hook, so they bypass RTK entirely.

## The filter already handles it

`src/cmds/js/npm_cmd.rs` has full install support — `filter_npm_output()` strips progress bars and keeps the summary. The `NPM_SUBCOMMANDS` const lists 30+ subcommands including `install`, `ci`, `list`, `outdated`, `audit`, etc.

The rule just doesn't route them there.

## Reproduction

```bash
$ rtk rewrite "npm install"; echo "exit: $?"
# exit: 1  (no match — falls through uncompressed)
```

## Suggested fix

```rust
// rules.rs line 56
pattern: r"^npm\s+(run|exec|install|ci|list|ls|outdated|audit|test)",
```

## Environment
- rtk 0.35.0
- macOS Apple Silicon
- `rtk discover` shows 43 `npm install` calls with 0% savings

Yanhu007:
## Summary

Fixes #1148 — npm rewrite rule only matches `run|exec`, causing `npm install`, `npm ci`, `npm test`, etc. to bypass RTK.

## Problem

```rust
// rules.rs L56 — only 2 of 30+ subcommands matched
pattern: r"^npm\s+(run|exec)",
```

`npm_cmd.rs` supports 30+ subcommands via `NPM_SUBCOMMANDS` const and already filters their output (strips progress bars, keeps summaries). The rewrite rule just doesn't route them there.

## Fix

Expand the regex to include the most commonly used npm subcommands that benefit from output filtering:

```rust
pattern: r"^npm\s+(run|exec|install|i|ci|uninstall|remove|rm|update|up|list|ls|outdated|init|create|publish|pack|link|audit|fund|test|t|start|stop|restart)",
```

## Verification

```bash
rtk rewrite "npm install"  # Before: exit 1 (no match) | After: rewrites to `rtk npm install`
```

### DIFF
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -53,7 +53,7 @@ pub const RULES: &[RtkRule] = &[
         subcmd_status: &[],
     },
     RtkRule {
-        pattern: r"^npm\s+(run|exec)",
+        pattern: r"^npm\s+(run|exec|install|i|ci|uninstall|remove|rm|update|up|list|ls|outdated|init|create|publish|pack|link|audit|fund|test|t|start|stop|restart)",
         rtk_cmd: "rtk npm",
         rewrite_prefixes: &["npm"],
         category: "PackageManager",


pszymkowiak:
Hey! Thanks for the fix.

We've moved to a `develop` branch as the default target for PRs — `master` is now reserved for releases. Could you retarget this PR to `develop`?

You can do it via the PR UI (Edit → change base branch) or with:
```bash
gh pr edit --base develop
```

See [CONTRIBUTING.md](https://github.com/rtk-ai/rtk/blob/develop/CONTRIBUTING.md) for the workflow.

KuSh:
Closing in favour of #1204, which makes the same npm rewrite-rule change, ships unit tests for it, and has a signed CLA. It also leaves the interactive subcommands such as login and init out of the pattern on purpose. If your version covers a case #1204 misses, please say so there and it can be folded in.

PR #1200, opened 2026-04-10
fix(permissions): glob_matches middle-wildcard matches commands with no trailing arguments

### LINKED ISSUE #1105: glob_matches fails to match middle-wildcard patterns when command has no trailing arguments
`command_matches_pattern` in `permissions.rs` fails to match patterns with a middle wildcard when the command ends at the middle segment — no trailing arguments.

## Reproduction

Pattern `git -C * diff:*` should match both of these commands:

```
git -C /path diff --stat   ← matches ✓
git -C /path diff          ← no match ✗
```

The simple (no-wildcard-in-prefix) path handles this correctly — `git diff:*` matches bare `git diff` via the word-boundary fast path at line 210. The bug is specific to `glob_matches`.

## Cause

When `glob_matches` normalizes `git -C * diff:*`, it becomes `git -C * diff *`, which splits on `*` into:

```
["git -C ", " diff ", ""]
```

The middle segment ` diff ` (with trailing space from the split) is searched as a substring in the remaining command text. For `git -C /path diff`, the remaining text after matching `git -C ` is `/path diff` — which does not contain ` diff ` (with trailing space). It contains ` diff` at the end, but the trailing space is missing.

For `git -C /path diff --stat`, the remaining text is `/path diff --stat`, which does contain ` diff ` — so it matches.

https://github.com/rtk-ai/rtk/blob/41a6c6bf6da78a4754794fdc6a1469df2e327920/src/hooks/permissions.rs#L227-L265

## Impact

Any `permissions.allow` entry with a middle wildcard fails to match commands that end at the middle segment. This causes `check_command` to fall through to the Default (ask) verdict, so the hook returns exit 3 instead of exit 0 for commands that should be auto-allowed.

Concrete example: a user with `Bash(git -C * status:*)` in their allow list still gets prompted for `git -C /path status` (no flags), but not for `git -C /path status --short`.

## Suggestion

Middle segments (not first, not last) could have their trailing whitespace trimmed before the substring search, or the search could also accept end-of-string as a valid boundary.

ousamabenyounes:
## Summary

- `glob_matches` failed to match commands that terminated at the middle-wildcard token with no trailing arguments
- Example: pattern `git -C * diff:*` matched `git -C /path diff --stat` (has trailing flag) but NOT bare `git -C /path diff`
- Root cause: normalization turns `git -C * diff:*` → `git -C * diff *`, splitting into `["git -C ", " diff ", ""]`. Middle segment `" diff "` has a trailing space; `/path diff` ends with `" diff"` (no trailing space), so `str::find(" diff ")` returns `None` → the command is incorrectly rejected
- Fix: in the middle-segment branch, also check `remaining.ends_with(part.trim_end())` as fallback — this handles commands that end at the middle token with no trailing args

## Test plan

- [x] `cargo clippy --all-targets` — no new errors
- [x] `cargo test --all` — all 1374 tests pass
- [x] New test: `test_middle_wildcard_at_end_of_command` — covers both positive cases (bare command, command with args) and a negative case

Closes #1105

🤖 Generated with [Ora Studio](https://studio.oratelecom.net)

### DIFF
--- a/src/hooks/permissions.rs
+++ b/src/hooks/permissions.rs
@@ -253,10 +253,20 @@ fn glob_matches(cmd: &str, pattern: &str) -> bool {
                 return false;
             }
         } else {
-            // Middle segment: find next occurrence
-            match cmd[search_from..].find(*part) {
-                Some(pos) => search_from += pos + part.len(),
-                None => return false,
+            // Middle segment: find next occurrence.
+            // Also accept end-of-string when the segment ends with whitespace — this
+            // handles commands that terminate at the middle token without trailing args,
+            // e.g. "git -C * diff:*" should match bare "git -C /path diff" (#1105).
+            let remaining = &cmd[search_from..];
+            if let Some(pos) = remaining.find(*part) {
+                search_from += pos + part.len();
+            } else {
+                let trimmed = part.trim_end();
+                if !trimmed.is_empty() && remaining.ends_with(trimmed) {
+                    search_from += remaining.len();
+                } else {
+                    return false;
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


FlorianBruniaux (approved):
LGTM — clean fix, tests pass.

Issue #1155, opened 2026-04-10
security: hook auto-allow bypasses agent permission model for rewritten commands


iRonin:
## Problem

The RTK hook outputs \`"permissionDecision": "allow"\` for rewritten commands, which **bypasses Claude Code's native permission model** for any command RTK recognizes.

### How it works

\`.claude/hooks/rtk-rewrite.sh\` lines 95-107:

\`\`\`bash
jq -n \
  --argjson updated "$UPDATED_INPUT" \
  '{
    "hookSpecificOutput": {
      "hookEventName": "PreToolUse",
      "permissionDecision": "allow",
      "permissionDecisionReason": "RTK auto-rewrite",
      "updatedInput": $updated
    }
  }'
\`\`\`

When the hook detects a rewrite match (exit code 0 from \`rtk rewrite\`), it outputs \`allow\`. Claude Code trusts this and executes the rewritten command **without prompting the user**.

### Risk scenarios

1. **Wrong rewrite**: If a regex rule in \`src/discover/rules.rs\` matches incorrectly, the command is silently rewritten and auto-executed.

2. **Argument corruption**: The rewrite changes the command. The user intended \`git status\` but the hook rewrites it to \`rtk git status\`. The output format changes (condensed, filtered). For operations where the user needs full output, this is a correctness issue masquerading as a permission bypass.

3. **Trust boundary**: The rewrite logic (70+ regex patterns in \`rules.rs\`) becomes a critical trust boundary. A bug in any pattern can silently alter command behavior.

## Current permission flow

\`src/hooks/rewrite_cmd.rs\` already implements a permission check system:
- Exit 0: Allow (no permission rules matched -> auto-allow)
- Exit 2: Deny (pass through to Claude's native deny)
- Exit 3: Ask (rewrite but let Claude prompt)

The issue is that **exit code 0 (default/no rule matched) maps to auto-allow**, which means any command RTK can rewrite gets executed without Claude prompting.

## Proposed fix

1. **Change the default from allow to ask**: When no explicit permission rule matches (the \`Default\` case in \`PermissionVerdict\`), the hook should output the rewritten command **without** \`"permissionDecision": "allow"\`, letting Claude prompt the user.

2. **Only auto-allow when an explicit allow rule matches**: In \`src/hooks/permissions.rs\`, the \`PermissionVerdict::Allow\` variant should be the **only** path to auto-allow. \`PermissionVerdict::Default\` should behave like \`Ask\`.

[...]

PR #1214, opened 2026-04-11
fix(hooks): require all segments to match allow rules (#1213)

### LINKED ISSUE #1213: security: compound command permission escalation — single allowed segment grants auto-allow to entire chain
## Summary

In `src/hooks/permissions.rs`, `check_command_with_rules` uses an `any_allow` flag
for compound commands. If **any single segment** matches an allow rule, the entire
compound command receives `PermissionVerdict::Allow` — even when other segments
have no matching allow rule and should default to ask.

## Reproduction

Given `~/.claude/settings.json`:
```json
{
  "permissions": {
    "allow": ["Bash(git status *)", "Bash(git status)", "Bash(cargo *)"]
  }
}
```

```bash
$ rtk rewrite "git add ."                    # exit 3 (ask) — correct
$ rtk rewrite "git status && git add ."      # exit 0 (allow) — BUG
$ rtk rewrite "cargo test && git add . && git commit -m foo"  # exit 0 (allow) — BUG
```

`git add .` alone correctly returns exit 3 (ask), but chaining it after an allowed
command escalates the entire chain to exit 0 (auto-allow).

## Root cause

```rust
// src/hooks/permissions.rs — check_command_with_rules()
if !any_allow && !any_ask {
    for pattern in allow_rules {
        if command_matches_pattern(segment, pattern) {
            any_allow = true;  // set by first segment, never unset
            break;
        }
    }
}
```

Once the first segment sets `any_allow = true`, subsequent segments skip the allow
check entirely. The final verdict `any_allow → Allow` grants auto-allow to the
whole chain.

## Expected behavior

A compound command should only receive `Allow` when **all** non-empty segments
individually match an allow rule. If any segment lacks an allow match, the verdict
should be `Default` (ask).

## Suggested fix

Replace `any_allow` with `all_allow` logic:

```rust
let mut all_segments_allowed = true;

for segment in &segments {
    let segment = segment.trim();
    if segment.is_empty() { continue; }

    // deny — unchanged (any deny → immediate Deny)
    for pattern in deny_rules {
        if command_matches_pattern(segment, pattern) {
            return PermissionVerdict::Deny;
        }
    }

    // ask — unchanged
    if !any_ask {
        for pattern in ask_rules {
            if command_matches_pattern(segment, pattern) {
                any_ask = true;
                break;
            }
        }
    }

    // allow — check every segment independently
    if all_segments_allowed {
        let matched = allow_rules.iter()
            .any(|p| command_matches_pattern(segment, p));
        if !matched {
            all_segments_allowed = false;
        }
    }
}

if any_ask { PermissionVerdict::Ask }
else if all_segments_allowed && !allow_rules.is_empty() { PermissionVerdict::Allow }
else { PermissionVerdict::Default }
```

## Impact

An LLM agent can construct `<allowed_cmd> && <unapproved_cmd>` to bypass user
confirmation on the second command. This is a permission escalation via compound
command chaining.

## Related

- #1155 (general hook auto-allow bypass)
- #1105 (glob_matches bug in permissions.rs)

pszymkowiak:
## Summary

Fixes #1213 — a compound command permission escalation where a single allowed segment granted auto-allow to the entire chain.

**Before** (broken):
```bash
rtk rewrite "git status"                    # exit 3 (ask) ✓
rtk rewrite "git add ."                     # exit 3 (ask) ✓
rtk rewrite "git status && git add ."       # exit 0 (allow) ❌ BUG
```

**After** (fixed):
```bash
rtk rewrite "git status && git add ."       # exit 3 (ask) ✓
rtk rewrite "cargo test && git add . && git commit -m foo"  # exit 3 (ask) ✓
```

## Root cause

`check_command_with_rules` used an `any_allow` flag that was set by the first matching segment and never unset. Once set, subsequent segments skipped the allow check entirely, escalating the whole chain.

## Fix

Replace `any_allow` with `all_segments_allowed` logic. Every non-empty segment must independently match an allow rule — otherwise the chain demotes to `Default` (ask). Also add:
- `saw_segment` guard to prevent vacuous Allow on empty commands
- `!allow_rules.is_empty()` check in the final verdict

Precedence is unchanged: **Deny > Ask > Allow > Default**.

## Impact

Closes a permission escalation vector: an LLM agent could construct `<allowed_cmd> && <unapproved_cmd>` to bypass user confirmation on the second command.

## Test plan

- [x] 6 new regression tests covering:
  - Reproduction case from #1213 (`git status && git add .`)
  - Three-segment chain with middle unallowed
  - Reversed ordering (unallowed first)
  - All 4 compound separators: `&&`, `||`, `|`, `;`
  - Ask-wins-over-partial-allow precedence
- [x] All 35 permissions tests pass (29 existing + 6 new)
- [x] Full suite: 1377 tests pass, 0 failures
- [x] `cargo fmt` clean

## Related

- #1155 (general hook auto-allow bypass)
- #1105 (glob_matches bug in permissions.rs)

### DIFF
--- a/src/hooks/permissions.rs
+++ b/src/hooks/permissions.rs
@@ -34,21 +34,27 @@ pub(crate) fn check_command_with_rules(
 ) -> PermissionVerdict {
     let segments = split_compound_command(cmd);
     let mut any_ask = false;
-    let mut any_allow = false;
+    // Every non-empty segment must independently match an allow rule for the
+    // compound command to receive Allow. See issue #1213: previously a single
+    // matching segment escalated the entire chain to Allow, enabling bypass.
+    let mut all_segments_allowed = true;
+    let mut saw_segment = false;
 
     for segment in &segments {
         let segment = segment.trim();
         if segment.is_empty() {
             continue;
         }
+        saw_segment = true;
 
-        // Deny takes highest priority
+        // Deny takes highest priority — any segment matching Deny blocks the whole chain.
         for pattern in deny_rules {
             if command_matches_pattern(segment, pattern) {
                 return PermissionVerdict::Deny;
             }
         }
 
+        // Ask — if any segment matches an ask rule, the final verdict is Ask.
         if !any_ask {
             for pattern in ask_rules {
                 if command_matches_pattern(segment, pattern) {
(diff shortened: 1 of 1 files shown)


FlorianBruniaux (approved):
LGTM — security fix, reviewed diff, tests pass.

PR #1217, opened 2026-04-11
fix(git): re-insert -- separator when clap consumes it from git diff args (#1215)

### LINKED ISSUE #1215: bug: git diff -- <path> fails from subdirectory — clap trailing_var_arg consumes  when first positional
## Bug

`rtk git diff -- <path>` fails with git ambiguity error when run from a subdirectory of the repo, while bare `git diff -- <path>` succeeds.

## Root cause

`clap`'s `trailing_var_arg = true` on the `Diff { args: Vec<String> }` subcommand consumes `--` as the option terminator when it appears **before any other positional argument**. Once at least one positional has been seen, clap preserves subsequent `--` as a positional value.

Result:
- `rtk git diff -- file` → `args = ["file"]` (clap ate `--`)
- `rtk git diff HEAD -- file` → `args = ["HEAD", "--", "file"]` (preserved after HEAD)

RTK then builds `git diff --stat file` (no `--`). Git tries the path as a revision when it can't resolve it relative to CWD → ambiguous argument error.

## Reproduction

```bash
# From a subdirectory of the repo
cd apps/client/frontend

# This fails (-- consumed by clap, git gets `git diff --stat apps/client/.../file`)
rtk git diff -- apps/client/frontend/src/components/MyComponent.tsx
# fatal: ambiguous argument '...': unknown revision or path not in the working tree

# This works (HEAD precedes --, so clap preserves it)
rtk git diff HEAD -- apps/client/frontend/src/components/MyComponent.tsx
# ✓

# Bare git works fine (-- not consumed)
git diff -- apps/client/frontend/src/components/MyComponent.tsx
# ✓
```

## Affected forms

Any `rtk git diff` invocation where `--` is the **first positional argument** (i.e., no flags or refs precede it).

## Suggested fix

In `run_diff`, after receiving `args`, normalize them by re-inserting `--` before path-like arguments when `--` is absent:

```rust
fn normalize_diff_args(args: &[String]) -> Vec<String> {
    if args.iter().any(|a| a == "--") {
        return args.to_vec(); // already correct
    }
    let path_start = args.iter().position(|arg| {
        !arg.starts_with('-') && looks_like_path(arg)
    });
    match path_start {
        Some(idx) => {
            let mut out = args[..idx].to_vec();
            out.push("--".to_string());
            out.extend_from_slice(&args[idx..]);
            out
        }
        None => args.to_vec(),
    }
}
```

Alternatively, change the clap definition to preserve `--` verbatim (e.g., using `raw = true` or restructuring the subcommand to not use `trailing_var_arg`).

## Environment

- RTK 0.34.1
- clap 4.x (derive feature)
- Windows 11, git 2.x
- Reproducible on any platform when CWD is a subdirectory of the git repo

ousamabenyounes:
## Summary

Fixes #1215

- Adds `normalize_diff_args()` which re-inserts `--` before the first path-like argument when clap's `trailing_var_arg = true` silently consumed it
- Adds `looks_like_path()` heuristic: args containing `/`, `\`, or starting with `.`/`~` are treated as paths
- No-op when `--` is already present — so `rtk git diff HEAD -- file` (where clap preserves the separator) is unaffected

## Root cause

`clap`'s `trailing_var_arg = true` treats `--` as the option terminator and drops it when it appears as the **first** positional argument. Result:

```
rtk git diff -- apps/.../MyComponent.tsx
  → args = ["apps/.../MyComponent.tsx"]   # -- was consumed
  → git diff --stat apps/.../MyComponent.tsx
  → fatal: ambiguous argument 'apps/...': unknown revision or path
```

The fix normalises `args` at the start of `run_diff` before any git command is built.

## Test plan

- [x] 7 new unit tests for `normalize_diff_args`: no-op, path-with-slash, ref+path, flag+path, pure-flags, dotfile, bare-ref
- [x] 1380 tests pass (baseline: 1373), 0 regressions

## Files changed

| File | Change |
|------|--------|
| `src/cmds/git/git.rs` | Add `looks_like_path`, `normalize_diff_args`, call in `run_diff`, 7 tests |

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -59,6 +59,44 @@ pub fn run(
     }
 }
 
+/// Returns true if `arg` looks like a file-system path rather than a git revision.
+///
+/// Used by `normalize_diff_args` to decide where to inject `--`.
+fn looks_like_path(arg: &str) -> bool {
+    // Path separators are the strongest signal
+    arg.contains('/') || arg.contains('\\') || arg.starts_with('.') || arg.starts_with('~')
+}
+
+/// Re-insert `--` before the first path-like argument when clap has consumed it.
+///
+/// clap's `trailing_var_arg = true` silently drops `--` when it appears as the
+/// first positional argument (before any other positional).  This means:
+///   `rtk git diff -- file` → args = ["file"]   (clap ate `--`)
+///   `rtk git diff HEAD -- file` → args = ["HEAD", "--", "file"]  (preserved)
+///
+/// Without the `--` separator git may treat an unambiguous path as a revision and
+/// emit "fatal: ambiguous argument".  We re-insert `--` before the first
+/// path-like argument when `--` is absent so git always gets the correct intent.
+fn normalize_diff_args(args: &[String]) -> Vec<String> {
+    // Already has `--` — nothing to do
+    if args.iter().any(|a| a == "--") {
+        return args.to_vec();
+    }
+    // Find the first non-flag arg that looks like a path
+    let path_start = args
+        .iter()
+        .position(|arg| !arg.starts_with('-') && looks_like_path(arg));
+    match path_start {
+        Some(idx) => {
+            let mut out = args[..idx].to_vec();
+            out.push("--".to_string());
+            out.extend_from_slice(&args[idx..]);
+            out
+        }
+        None => args.to_vec(),
+    }
+}
+
 fn run_diff(
     args: &[String],
     max_lines: Option<usize>,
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Fixes a bug where clap's trailing_var_arg silently consumes the `--` separator in git diff args, causing git to misinterpret file paths as revisions. Also fixes a secondary issue where user-format git log output was being incorrectly truncated, corrupting programmatic output like SHA hashes.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1215

---

FlorianBruniaux (approved):
LGTM — clean fix, tests pass.

Issue #893, opened 2026-03-28
Hook updatedInput silently ignored when using bypassPermissions mode in Claude Code


BingBingHe:
Problem

  When Claude Code is configured with bypassPermissions mode (in ~/.claude/settings.json), the RTK rewrite hook silently fails to rewrite commands. The hook runs, but the updatedInput has no effect.

  Root Cause

  Claude Code has two internal paths for handling hook output:

  ┌──────────────────────┬─────────────────────────────────────────────────────────┬─────────────────────────────────────┐
  │         Path         │                         Trigger                         │  Effect in bypassPermissions mode   │
  ├──────────────────────┼─────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ hookUpdatedInput     │ Hook returns only updatedInput, no permissionDecision   │ ✅ Command is rewritten             │
  ├──────────────────────┼─────────────────────────────────────────────────────────┼─────────────────────────────────────┤
  │ hookPermissionResult │ Hook returns permissionDecision: "allow" + updatedInput │ ❌ updatedInput is silently ignored │
  └──────────────────────┴─────────────────────────────────────────────────────────┴─────────────────────────────────────┘

  The current RTK hook outputs permissionDecision: "allow" alongside updatedInput. In bypassPermissions mode, Claude Code skips permission handling entirely — and discards the updatedInput that came with it.

  Fix

  Remove permissionDecision and permissionDecisionReason from the hook output:

  # ❌ Before (broken in bypassPermissions mode)
  jq -n --argjson updated "$UPDATED_INPUT" '{
    "hookSpecificOutput": {
      "hookEventName": "PreToolUse",
      "permissionDecision": "allow",
      "permissionDecisionReason": "RTK auto-rewrite",
      "updatedInput": $updated
    }
  }'

  # ✅ After (works correctly)
  jq -n --argjson updated "$UPDATED_INPUT" '{
    "hookSpecificOutput": {
      "hookEventName": "PreToolUse",
      "updatedInput": $updated
    }
  }'

aeppling:
Fixed in 0.35.0

PR #1218, opened 2026-04-11
fix(hook): emit explicit allow for non-rewritten commands (#1033)

### LINKED ISSUE #1033: Hook produces unexpected permission prompts in --dangerously-skip-permissions mode for non-rewritten commands
## Summary

When RTK's `PreToolUse` hook is installed and Claude Code is running in `--dangerously-skip-permissions` mode, commands that RTK does **not** rewrite still trigger permission prompts — as if no bypass flag were set.

## Environment

- RTK version: 0.34.2
- Claude Code: latest
- OS: macOS (arm64)
- Hook: `~/.claude/hooks/rtk-rewrite.sh` (hook-version: 3), registered as `PreToolUse` Bash matcher

## Steps to Reproduce

1. Install RTK and run `rtk init -g`
2. Start Claude Code with `--dangerously-skip-permissions`
3. Ask Claude to run any command that RTK has no rewrite rule for (e.g. `which rtk`, `rtk --version`, `claude mcp list`)

## Expected Behavior

Commands should execute without a permission prompt, consistent with `--dangerously-skip-permissions` behavior when no hook is installed.

## Actual Behavior

Every Bash command that the hook does not rewrite produces a permission prompt, making `--dangerously-skip-permissions` effectively non-functional when the RTK hook is active.

## Analysis

The hook exits 0 with no stdout for commands it has no opinion on (no RTK equivalent, early exits for missing tools, empty command, unknown exit codes). Claude Code appears to interpret "hook ran and produced no output" differently from "no hook exists" — instead of falling back to bypassPermissions, it falls through to its default permission prompt behaviour.

This affects all the hook's "no opinion" exit paths.

## References

- Related: #886 (permission model interaction)
- Related: #361 (hook architecture rewrite)

ousamabenyounes:
## Summary

Fixes #1033

- When RTK's `PreToolUse` hook is installed and a command has no RTK rewrite rule, the hook exits 0 with no stdout output
- Ora Studio interprets "hook ran with no output" differently from "no hook installed" — it falls through to its default permission prompt, defeating `--dangerously-skip-permissions`
- Fix: add `_rtk_allow_passthrough()` helper that emits `permissionDecision: "allow"` and call it in all "no opinion" exit paths

## Root Cause

The hook protocol distinguishes two states:
1. **Hook absent / exit 0 + no output** → Ora Studio applies its own permission model (may prompt)
2. **`permissionDecision: "allow"` in stdout** → Ora Studio unconditionally allows the command

The old hook used path (1) for passthrough cases, which silently broke `--dangerously-skip-permissions` whenever a registered hook was present. The fix moves all passthrough cases to path (2).

## Changes

| File | Change |
|------|--------|
| `hooks/claude/rtk-rewrite.sh` | Add `_rtk_allow_passthrough()`, call it for exit codes 1 (no match), 0-no-change (already RTK), unknown, and early guards (missing jq/rtk/old version). Deny-rule exit 2 retains bare `exit 0`. Bump hook-version to 4. |
| `hooks/claude/test-rtk-rewrite.sh` | Update `test_rewrite` to verify `permissionDecision=allow` + no `updatedInput` for passthrough cases. Add `test_passthrough_allows` section reproducing the exact commands from the bug report. |

## Verification

- [x] New tests: 5 added, all pass (`which rtk`, `rtk --version`, `claude mcp list`, `echo hello`, `cd /tmp`)
- [x] No regressions: pre-existing 17 failures unchanged (unrelated to this fix), 0 new failures
- [x] Rewrite cases unaffected: `git status → rtk git status` still works + emits `allow`
- [x] Deny rule path (exit 2) unchanged: bare `exit 0`, lets Ora Studio handle natively
- [x] `cargo fmt && cargo clippy && cargo test`: 1373 pass, 0 fail

---

### DIFF
--- a/hooks/claude/rtk-rewrite.sh
+++ b/hooks/claude/rtk-rewrite.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# rtk-hook-version: 3
+# rtk-hook-version: 4
 # RTK Claude Code hook — rewrites commands to use rtk for token savings.
 # Requires: rtk >= 0.23.0, jq
 #
@@ -13,14 +13,22 @@
 #   2           Deny rule matched → pass through (Claude Code native deny handles it)
 #   3 + stdout  Ask rule matched → rewrite but let Claude Code prompt the user
 
+# Emit an explicit allow decision so Claude Code honours --dangerously-skip-permissions
+# even when the hook has no rewrite to apply.  Using printf avoids a jq dependency
+# for the early-exit paths (missing tools, old version).
+_rtk_allow_passthrough() {
+  printf '{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"allow"}}\n'
+  exit 0
+}
+
 if ! command -v jq &>/dev/null; then
   echo "[rtk] WARNING: jq is not installed. Hook cannot rewrite commands. Install jq: https://jqlang.github.io/jq/download/" >&2
-  exit 0
+  _rtk_allow_passthrough
 fi
 
 if ! command -v rtk &>/dev/null; then
   echo "[rtk] WARNING: rtk is not installed or not in PATH. Hook cannot rewrite commands. Install: https://github.com/rtk-ai/rtk#installation" >&2
-  exit 0
+  _rtk_allow_passthrough
 fi
 
 # Version guard: rtk rewrite was added in 0.23.0.
@@ -37,7 +45,7 @@ if [ ! -f "$CACHE_FILE" ]; then
     # Require >= 0.23.0
     if [ "$MAJOR" -eq 0 ] && [ "$MINOR" -lt 23 ]; then
       echo "[rtk] WARNING: rtk $RTK_VERSION is too old (need >= 0.23.0). Upgrade: cargo install rtk" >&2
-      exit 0
+      _rtk_allow_passthrough
     fi
   fi
   mkdir -p "$CACHE_DIR" 2>/dev/null
@@ -59,11 +67,15 @@ case $EXIT_CODE in
   0)
     # Rewrite found, no permission rules matched — safe to auto-allow.
     # If the output is identical, the command was already using RTK.
-    [ "$CMD" = "$REWRITTEN" ] && exit 0
+    if [ "$CMD" = "$REWRITTEN" ]; then
+      _rtk_allow_passthrough
+    fi
     ;;
   1)
-    # No RTK equivalent — pass through unchanged.
-    exit 0
+    # No RTK equivalent — pass through with explicit allow so that
+    # --dangerously-skip-permissions is respected (Claude Code treats a hook
+    # that exits 0 with no output differently from no hook being installed).
+    _rtk_allow_passthrough
     ;;
   2)
     # Deny rule matched — let Claude Code's native deny rule handle it.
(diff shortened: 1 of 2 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Fixes a bug where RTK's PreToolUse hook would emit no output for passthrough commands, causing Claude Code to fall back to its permission prompt instead of honoring --dangerously-skip-permissions. The fix adds an explicit permissionDecision:allow emission for all no-opinion exit paths in the hook script, with corresponding test coverage.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [x] Docs updated

**Linked issues:** #1033

---

roeex5:
## This fixes #1033 — but #893 (rewrites discarded in bypass mode) is still open

Great fix for the passthrough case, and the `printf` instead of `jq` for the early-exit paths is a nice touch.

One gap worth flagging: the rewrite path is unchanged — it still emits `permissionDecision: "allow"` + `updatedInput` together. In `--dangerously-skip-permissions` mode, Claude Code routes that through its `hookPermissionResult` path, which **discards `updatedInput`**. The command gets allowed but runs unrewritten — so token savings are lost entirely in bypass mode.

### The dual-swap approach would fix both

| Case | Current output | Proposed output |
|---|---|---|
| RTK rewrites command | `permissionDecision: allow` + `updatedInput` | `updatedInput` only |
| RTK has no opinion (passthrough) | *(empty)* → `allow` after this PR | `permissionDecision: allow` only ✅ already in this PR |

Removing `permissionDecision` from the rewrite path routes it through the `hookUpdatedInput` path instead, which correctly applies `updatedInput` in both normal and bypass modes.

### Open question before expanding the scope

Does returning **only `updatedInput`** (no `permissionDecision`) in **normal mode** (no bypass) still auto-allow the rewritten command — or does Claude Code fall through to its default permission prompt?

If it auto-allows: the dual-swap is a clean complete fix with no regressions.  
If it prompts: the rewrite path needs a different approach (e.g. detecting bypass mode via an env var, or waiting for a Claude Code fix).

Happy to help test or extend this PR if that's useful.

ousamabenyounes:
Thanks for the review @roeex5 — the gap you're flagging on the rewrite path in bypass mode is real, and the dual-swap table makes it easier to reason about.

I'd rather not guess the answer to your open question on spec alone — the hook routing between `hookPermissionResult` and `hookUpdatedInput` deserves an empirical test in both normal and bypass modes before committing to the change.

Can we land this PR as-is (the passthrough fix is cleanly in scope for #1033) and track the rewrite path in a dedicated follow-up issue? Happy to pair on that follow-up.

ousamabenyounes:
Quick update on the dual-swap discussion: noticed #893 was closed by @aeppling on 2026-04-06 with *"Fixed in 0.35.0"* — that was before your review @roeex5, so worth re-checking against current `develop`.

If the 0.35.0 fix already routes the rewrite path correctly in bypass mode, then this PR (passthrough fix only, scoped to #1033) is complete on its own and the dual-swap concern may already be moot.

If you've reproduced the bypass-mode issue against ≥0.35.0, please reopen #893 with the repro — happy to extend this PR or open a follow-up.

@aeppling @pszymkowiak — would appreciate a maintainer pass on this when convenient. Mergeable, in scope for #1033, and the rewrite-path concern from @roeex5 looks like it's already addressed by whatever shipped in 0.35.0.

PR #1227, opened 2026-04-11
fix(ccusage): add --yes flag and warn when falling back to npx

### LINKED ISSUE #1226: bug: the command cc-economics is hanging
I left like this for half a minute and nothing was outputed. Not sure if this is expected.

Image

Version:
rtk 0.35.0

OS:
MacOS Tahoe 26

davidalecrim1:
## Summary

- Fixes silent hang in `rtk cc-economics` when `ccusage` is not installed globally
- `npx` prompts "Ok to proceed? (y)" before installing the package — RTK's subprocess blocks on this forever since it has no interactive stdin
- Adding `--yes` auto-confirms the install
- Added an `[info]` warning to stderr so users know why it may take a moment

## Root Cause

`build_command()` falls back to `npx ccusage` when the binary isn't in PATH. Without `--yes`, npx waits for interactive confirmation that never comes.

Closes #1226

### DIFF
--- a/src/analytics/ccusage.rs
+++ b/src/analytics/ccusage.rs
@@ -95,7 +95,9 @@ fn build_command() -> Option<Command> {
     }
 
     // Fallback: try npx
+    eprintln!("[info] ccusage not installed globally, fetching via npx...");
     let npx_check = resolved_command("npx")
+        .arg("--yes")
         .arg("ccusage")
         .arg("--help")
         .stdout(std::process::Stdio::null())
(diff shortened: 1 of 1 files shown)


src/analytics/ccusage.rs
```diff
@@ -95,7 +95,9 @@ fn build_command() -> Option<Command> {
     }
 
     // Fallback: try npx
+    eprintln!("[info] ccusage not installed globally, fetching via npx...");
```
davidalecrim1, line 98: We can remove this if doesn't make sense for the CLI UX, let me know if that's the case.

FlorianBruniaux (approved):
LGTM — clean fix, tests pass.

PR #1273, opened 2026-04-13
fix(pytest,grep): harden quiet summary and grep match parsing

siubing05:
## Summary
- harden pytest quiet-mode summary detection and summary parsing
- surface quiet-mode error/xfailed/xpassed/deselected outcomes correctly
- harden grep match parsing for single-file output, Windows paths, and content containing colons
- keep filenames in rg/grep output with `-H` and fix the output label typo

## Why
While integrating RTK auto-rewrite into Hermes Agent, two parser edge cases showed up:
- `pytest -q` can end with summaries like `25 passed in 3.92s`, `1 error in 0.20s`, or `2 deselected in 0.02s`; the old logic mainly relied on the `=== ... ===` style summary and could fall back to `Pytest: No tests collected`
- single-file grep output can omit the filename, and Windows paths / colon-rich content make naive `split(':')` parsing unsafe

### pytest
- add `SummaryCounts` parsing for passed / failed / errors / skipped / xfailed / xpassed / deselected
- detect quiet summaries and `no tests ran` lines via `is_summary_line()`
- preserve failure/error detail rendering in the compact RTK summary

### grep
- add `parse_match_line()` to safely parse `file:line:text` and fallback `line:text`
- force filenames with `-H` for both `rg` and `grep`
- add regression coverage for Windows paths and content containing colons

## Validation
- `cargo test`
- `cargo fmt --all --check`

### DIFF
--- a/src/cmds/python/pytest_cmd.rs
+++ b/src/cmds/python/pytest_cmd.rs
@@ -12,6 +12,32 @@ enum ParseState {
     Summary,
 }
 
+fn is_summary_line(line: &str) -> bool {
+    let normalized = line.trim().trim_matches('=').trim().to_lowercase();
+    if normalized.is_empty() {
+        return false;
+    }
+
+    if normalized.starts_with("no tests ran") {
+        return true;
+    }
+
+    let Some(first_token) = normalized.split_whitespace().next() else {
+        return false;
+    };
+    if first_token.parse::<usize>().is_err() {
+        return false;
+    }
+
+    normalized.contains(" passed")
+        || normalized.contains(" failed")
+        || normalized.contains(" skipped")
+        || normalized.contains(" xfailed")
+        || normalized.contains(" xpassed")
+        || normalized.contains(" error")
+        || normalized.contains(" errors")
+}
+
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let mut cmd = if tool_exists("pytest") {
         resolved_command("pytest")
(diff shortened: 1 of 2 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes two parsing bugs: pytest quiet-mode output (e.g. '25 passed in 3.92s') was not recognized as a summary line, causing incorrect 'No tests collected' results, and ripgrep single-file searches emitting 'LINE:text' without a filename were being misparsed. Adds the -H flag to rg/grep for consistent filename output and introduces a robust parse_match_line() helper.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

---

F0rty-Tw0:
Hey @siubing05, heads up that I just opened #1554 which covers the same single-file grep parser fix as the grep portion of this PR. Yours has been open longer; I didn't notice it before starting, apologies for the duplicate effort.

For reviewers' context, the two approaches differ on Windows path handling:

- This PR uses `splitn(3, ':')` with a 2-format fallback (handles `line:content` if `-H` is ever absent). On Windows abs paths like `C:\foo\file.rs:42:content` the splitn picks `file="C"` and the line-number parse fails, so the line gets dropped.
- #1554 uses a greedy regex `^(.+):(\d+):(.*)$` that picks the LAST `:digits:` boundary, handling drive letters correctly.

Both fix the Linux/macOS reproducer in #1436. Whichever shape maintainers prefer, happy to defer or fold a Windows-path test fixture into this one if that's the path forward. Also no objection if you'd rather merge yours and I close #1554, your work was first.

siubing05:
Thanks for the heads-up — no worries on the overlap.

I’ve since updated this PR to cover the Windows-path issue you pointed out as well. The grep parser now handles drive-letter paths robustly, and I added regression coverage for that case in addition to the original single-file parsing fix.

This PR still also includes the pytest quiet-summary fix, so at this point it covers:
- quiet pytest summary parsing
- single-file grep output parsing
- Windows absolute path parsing in grep matches

Latest head is `502d037` if maintainers want to review the updated version.

If maintainers prefer to keep everything in one place, I’m happy for this PR to stand as the combined fix.

PR #1281, opened 2026-04-13
fix(ls): support ISO date format from GNU coreutils

caoer:
## Summary

- `rtk ls` produces empty output on systems where `ls` outputs ISO-format dates (`2026-04-13 17:07`), common with GNU coreutils / Nix
- The `LS_DATE_RE` regex only matched BSD-style dates (`Mar 31 16:18` / `Dec 25  2024`)
- Extended the regex to match both BSD and ISO date formats

## Test plan

- [x] Added `test_parse_ls_line_iso_date` — parses file with ISO date
- [x] Added `test_parse_ls_line_iso_date_dir` — parses directory with ISO date
- [x] Added `test_compact_iso_dates` — full compact_ls with ISO-dated input
- [x] All 1449 existing tests pass, no new clippy warnings

### DIFF
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -11,9 +11,11 @@ use std::io::IsTerminal;
 lazy_static! {
     /// Matches the date+time portion in `ls -la` output, which serves as a
     /// stable anchor regardless of owner/group column width.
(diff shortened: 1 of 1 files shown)


caoer:
check?

PR #1286, opened 2026-04-14
fix(git): exclude branch refs from looks_like_path heuristic

FlorianBruniaux:
## Problem

PR #1217 fixed issue #1215 (clap consuming `--` before path args) but introduced a regression: the `looks_like_path` heuristic treated any arg containing `/` as a file path. This broke `rtk git diff origin/main` — the command became `git diff -- origin/main`, git looked for a file literally named `origin/main`, found nothing, and silently returned an empty diff.

## Root cause

```rust
// Before — too loose
fn looks_like_path(arg: &str) -> bool {
    arg.contains('/') || arg.contains('\\') || arg.starts_with('.') || arg.starts_with('~')
}
```

`origin/main`, `upstream/develop`, `feature/my-branch` all contain `/` and were misclassified as paths.

## Fix

Layered detection strategy, cheapest signals first:

1. Explicit path prefixes (`./`, `../`, `~/`, `~`, backslash) → path, no git call.
2. Git revision syntax (`..` range, `^` parent) → ref, no git call.
3. Dotfile (`.gitignore`, `.env`) → path, no git call.
4. Bare word without `/` or `@` (HEAD, main, develop) → ref, no git call.
5. Ambiguous (`origin/main`, `src/main.rs`, `HEAD@{1}`) → ask git via `git rev-parse --verify --quiet`. If git knows it, it's a ref. If not, it's a path.

One `git rev-parse` subprocess only fires on ambiguous args. For the vast majority of `git diff` invocations it never runs.

## Tested

```bash
rtk git diff origin/master       # now shows real diff (was empty)
rtk git diff develop             # still works
rtk git diff src/cmds/git/git.rs # path diff still works
rtk git diff .gitignore          # dotfile still works
```

All 7 original tests from #1217 pass + 6 new regression tests added.

Fixes regression from #1217. Preserves fix for #1215.

/cc @ousamabenyounes

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1951,13 +1951,24 @@ mod tests {
     }
 
     /// Regression #1217: remote branch refs must not be treated as paths.
-    /// `origin/master` is a valid git ref — no `--` injection.
+    /// Skips gracefully when the remote ref is not available (shallow CI clones).
     #[test]
     fn test_normalize_diff_args_no_injection_for_remote_ref() {
-        // origin/master contains `/` but is a valid ref in this repo.
+        // Probe whether origin/master is reachable in this checkout.
+        // CI often uses shallow clones without all remote tracking refs.
+        let has_ref = std::process::Command::new("git")
+            .args(["rev-parse", "--verify", "--quiet", "origin/master"])
+            .output()
+            .map(|o| o.status.success())
+            .unwrap_or(false);
+
+        if !has_ref {
+            // Ref not available in this environment — skip rather than false-fail.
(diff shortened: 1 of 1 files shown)


LeGEC:
I see that other commands (`cargo` commands for example) have a "restore the '--'" feature:
https://github.com/rtk-ai/rtk/blob/a69935746172d913fcbd282d720d2daf5025e5e9/src/cmds/rust/cargo_cmd.rs#L35-L69

shouldn't that be the way to go with `git` commands ? check if the original command line arguments had a `--` in them ?

PR #1293, opened 2026-04-14
fix(npx): route unknown packages directly via npx, not npm run

### LINKED ISSUE #1269: 【Bug】rtk npx vue-tsc --noEmit（MacOS）
Problem confirmation
After detailed troubleshooting, I confirmed the problem:
Root cause
There are bugs in the processing of vue tsc by rtk npx:
Command Result Reason
npx vue-tsc --version  ✅  Normal native npx direct execution package
rtk npx vue-tsc --version  ✅  Normal rtk correct proxy
npx vue-tsc --noEmit  ✅  Normal native npx passes parameters to vue tsc
rtk npx vue-tsc --noEmit  ❌  An error is reported. The rtk internally erroneously executes vue tsc as an npm script
The error message npm error Missing script: "vue tsc" clearly indicates that the rtk npx calls npm run vue tsc internally instead of npx vue tsc.

NEOFIT-dev:
## Problem

`rtk npx vue-tsc --noEmit` was calling `npm run vue-tsc --noEmit` instead of `npx vue-tsc --noEmit`, producing `npm error Missing script: "vue-tsc"`.

Root cause: the default `Commands::Npx` handler called `npm_cmd::run()` which injects `run` before any argument not found in `NPM_SUBCOMMANDS`. Since `vue-tsc` is not in that list, it became `npm run vue-tsc`.

Note: `rtk npx vue-tsc --version` appeared to work by accident — npm silently consumes `--version` as its own flag rather than passing it to the script.

## Fix

Add `npm_cmd::run_npx()` that calls `npx <args>` directly without any `run` injection, and use it as the default handler for unknown packages in `Commands::Npx`.

## Testing

```
# Before (system rtk 0.35.0)
$ rtk npx vue-tsc --noEmit
npm error Missing script: "vue-tsc"

# After
$ rtk npx vue-tsc --noEmit
Version 6.0.2
tsc: The TypeScript Compiler - Version 6.0.2
...
```

Fixes #1269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
+* **npx:** fix `rtk npx <package> --flags` calling `npm run <package>` instead of `npx <package>` for unknown packages like `vue-tsc` ([#1269](https://github.com/rtk-ai/rtk/issues/1269))
 * **git:** remove `-u` short alias from `--ultra-compact` to fix `git push -u` upstream tracking ([#1086](https://github.com/rtk-ai/rtk/issues/1086))
 
 ## [0.35.0](https://github.com/rtk-ai/rtk/compare/v0.34.3...v0.35.0) (2026-04-06)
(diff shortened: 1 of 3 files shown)


KuSh:
Closing as https://github.com/rtk-ai/rtk/pull/1458 already fixed the problem

Issue #1152, opened 2026-04-10
bug(rtk curl): JSON schema conversion destroys API response values, breaks browser CDP connections


iRonin:
## Problem

When `rtk curl` receives JSON output, it runs it through `json_cmd::filter_json_string()` which converts JSON values to their types (schema mode). This **destroys actual data values** that AI agents and scripts need to read.

### Real-world impact: AI browser tool breakage

When an AI agent runs `curl -s http://localhost:9222/json/version` to get the Chrome DevTools Protocol WebSocket URL, Chrome returns:

\`\`\`json
{
  "webSocketDebuggerUrl": "ws://127.0.0.1:9222/devtools/browser/abc-123",
  "browser": "Chrome/131.0.0.0"
}
\`\`\`

RTK converts this to a schema:

\`\`\`
{
  webSocketDebuggerUrl: string,
  browser: string
}
\`\`\`

The actual WebSocket URL is **destroyed**. The browser tool cannot connect because it never receives the URL value.

### Root cause

\`src/cmds/cloud/curl_cmd.rs\` lines 58-67:

\`\`\`rust
if (trimmed.starts_with('{') || trimmed.starts_with('['))
    && (trimmed.ends_with('}') || trimmed.ends_with(']'))
{
    if let Ok(schema) = json_cmd::filter_json_string(trimmed, 5) {
        // Only use schema if it is actually shorter than the original
        if schema.len() <= trimmed.len() {
            return schema;  // SCHEMA replaces actual values!
        }
    }
}
\`\`\`

This behavior is correct for reducing tokens when showing an API structure to an LLM, but **wrong for commands where values matter** (CDP endpoints, auth tokens, IDs, URLs).

## Proposed fix

1. **Add a `--raw` / `--no-filter` flag** to `rtk curl` that skips JSON schema conversion and returns raw output
2. **Skip schema conversion for localhost/127.0.0.1 URLs** -- these are typically internal API responses where values are needed
3. **Default to keeping original JSON when it is already compact** (the current length check helps but still converts if schema is shorter, even if the original was small and readable)

## Relevant code

- \`src/cmds/cloud/curl_cmd.rs\` (lines 54-88: \`filter_curl_output\`)
- \`src/cmds/system/json_cmd.rs\` (schema conversion logic)

## Workaround

[...]

PR #1310, opened 2026-04-14
 Simplified `rtk curl` Filter

JBF1991:
## Summary
Refactored `rtk curl` to use a simple, predictable truncation strategy with tee hints instead of JSON schema conversion. This eliminates edge cases, reduces code complexity by ~36%, and delivers consistent 73-99% token savings across all response types greater than 500 bytes.

## Test plan
I tested against various local and remote endpoints with responses >500 bytess, <500 bytes and  both successful and error responses.

### Savings

| Response Type | Original Size | Savings | Behavior |
|--------------|---------------|---------|----------|
| npm registry (express) | 801 KB | **99.9%** | 500B truncated + tee hint |
| GitHub API (releases) | 36 KB | **98.5%** | 500B truncated + tee hint |
| httpbin HTML | 3.7 KB | **86.0%** | 500B truncated + tee hint |
| nginx 404 HTML | 2 KB | **73.8%** | 500B truncated + tee hint |
| httpbin UUID | 53 B | 0% | Below threshold, passed through |
| httpbin 404 (empty) | 0 B | 0% | Empty body, no output, timer not called|

### Tiny JSON — passed through (< 500 bytes)
```bash
$ rtk curl -s https://httpbin.org/ip
{
  "origin": "75.143.98.179"
}
```
### Large API response — truncated + tee hint
```bash
$ rtk curl -s "https://api.github.com/repos/rust-lang/rust/releases?per_page=5"
[
  {
    "url": "https://api.github.com/repos/rust-lang/rust/releases/301852655",
    "assets_url": "https://api.github.com/repos/rust-lang/rust/releases/301852655/assets",
    "upload_url": "https://uploads.github.com/repos/rust-lang/rust/releases/301852655/assets{?name,label}",
    ... (36049 bytes total)
[full output: ~/.local/share/rtk/tee/1776209074_curl.log]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -892,7 +892,7 @@ dependencies = [
 
 [[package]]
 name = "rtk"
-version = "0.34.3"
+version = "0.35.0"
 dependencies = [
  "anyhow",
  "automod",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -892,7 +892,7 @@ dependencies = [
 
 [[package]]
 name = "rtk"
-version = "0.35.0"
+version = "0.34.3"
 dependencies = [
  "anyhow",
  "automod",
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ rtk json config.json            # Structure without values
 rtk deps                        # Dependencies summary
 rtk env -f AWS                  # Filtered env vars
 rtk log app.log                 # Deduplicated logs
-rtk curl <url>                  # Auto-detect JSON + schema
+rtk curl <url>                  # Truncate + save full output
 rtk wget <url>                  # Download, strip progress bars
 rtk summary <long command>      # Heuristic summary
 rtk proxy <command>             # Raw passthrough + tracking
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -964,13 +964,13 @@ Les lignes repetees sont fusionnees : `[ERROR] Connection refused (x42)`.
 
 ---
 
-### `rtk curl` -- HTTP avec detection JSON
+### `rtk curl` -- HTTP avec troncature
 
 ```bash
 rtk curl [args...]
 ```
 
-Auto-detecte les reponses JSON et affiche le schema au lieu du contenu complet.
+Tronque les reponses longues et sauvegarde la sortie complete dans un fichier pour recuperation.
 
 ---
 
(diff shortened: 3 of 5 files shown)


JBF1991:
I am working on documentation for this. Apologies.

aeppling:
Hello @JBF1991 

RTK rewrite commands transparently, it catch cmd by hooks and then rewrite with rtk. Some implementation use context because hook not available, but in reality, for RTK to be efficient, it should rewrite transparently.

For me this fix is not the correct approach since it require LLM to be aware of RTK and add flag.

You may want to check https://github.com/rtk-ai/rtk/blob/master/CONTRIBUTING.md#design-philosophy this part of contributing guide.

Thanks for the contribution ! If any questions or want to discuss i'm reachable here or on Discord

JBF1991:
I have updated to using the tee/hint form from other areas of the project as well as re-updated the docs.

aeppling:
Hey @JBF1991 

Sorry for response time, lot of work beside, here is some issues found to be addressed before merging:

### P0: Data loss when tee is unavailable

`filter_curl_output` truncates at 500 bytes unconditionally, but `force_tee_hint` can return `None` in at least 5 scenarios: `RTK_TEE=0`, `tee.enabled = false`, broken config TOML, directory creation failure, or disk full. When any of these hit, everything beyond byte 500 is permanently lost with no recovery path. The old code would have passed through the full response.

### P1: Magic number 500 decoupled from `MIN_TEE_SIZE`

`tee.rs` defines `MIN_TEE_SIZE = 500` privately. `curl_cmd.rs` hardcodes the same `500` separately.

Use already existing constant for consistence with codebase.

### Coding practices

**Hardcoded threshold**: The `500` byte threshold appears inline. Coding practice requires named constants at the top of the file for fixed limits.

**Missing edge case comment**: The UTF-8 char boundary loop is a non-obvious edge case that needs a comment explaining why it's there.

**Fallback pattern**: If a filter fails, fall back to raw output. When tee is unavailable, truncation still destroys data with no fallback. Directly related to the P0.

**Stale header comment**: The file header still says `//! Runs curl and auto-compresses JSON responses`, JSON compression is removed and should reflect the new behavior.

aeppling:
Hello, rectifying review, for MIN_TEE_SIZE do not reuse from tee you can just define your constant locally it will be better for this purpose; thanks for pointing this out

aeppling:
https://github.com/rtk-ai/rtk/issues/1152

JBF1991:
Added MAX_RESPONSE_SIZE comment.
Corrected doc comment.
Added comment for utf-8 boundary edge case.
If force_tee_hint returns none, default to trimmed raw.

aeppling (approved):

PR #1304, opened 2026-04-14
docs: Adrien review — hierarchy, counts, consistency, gaps

FlorianBruniaux:
## Summary

Full doc review pass based on Adrien's feedback.

**Hierarchy**
- Move `troubleshooting.md` and `what-rtk-covers.md` to `guide/resources/` (new folder alongside existing `getting-started/` and `analytics/`)
- Add `guide/resources/telemetry.md` — user-facing page (consent, opt-out, GDPR rights) adapted from `docs/TELEMETRY.md`

**Content fixes**
- Remove all hardcoded counts (`9 ecosystems`, `12 agents`, `60+`, `7 more`) — qualitative language only
- Unify DB filename: `tracking.db` → `history.db` everywhere (canonical: `src/core/constants.rs`)
- `installation.md`: replace bare `cargo install rtk` with warning callout + explicit git URL (Rust Type Kit name collision)
- `supported-agents.md` / `quick-start.md`: counts and hardcoded lists removed

**New content**
- `gain.md`: `--quota` section explaining `pro`/`5x`/`20x` tier meanings
- `gain.md`: callout linking to `discover.md` (find missed savings)
- `index.md`: "Analyze your usage" section (mentions `rtk discover` + `rtk session`)
- `configuration.md`: `ignore_dirs`/`ignore_files` scope clarified + prose link to telemetry
- `what-rtk-covers.md`: note that `--ultra-compact` long form should be preferred over `-u` with git commands (conflict with git's own `-u`)

**Cross-references**
- All internal links updated for new `resources/` paths
- Landing repo (`rtk-landing`) already has `astro.config.mjs` redirects so old URLs (`/guide/troubleshooting/`, `/guide/what-rtk-covers/`) 301 to new paths

**README**
- Core team section added (Patrick Szymkowiak — Founder, Florian Bruniaux — Core contributor, Adrien Eppling — Core contributor)

## Test plan

- [ ] `pnpm build` in `rtk-landing` with `RTK_REPO_PATH` pointing to this branch — 0 Starlight broken-link warnings
- [ ] `/guide/` sidebar shows 3 groups: Getting Started, Analytics, Resources
- [ ] `/guide/troubleshooting/` redirects to `/guide/resources/troubleshooting/` (301)
- [ ] `/guide/what-rtk-covers/` redirects to `/guide/resources/what-rtk-covers/` (301)
- [ ] `rtk-landing` Cmd+K search — docs results navigate to `/guide/…` (not 404)
[...]

PR #1312, opened 2026-04-14
fix(ci): add write permissions to pr-target-check workflow

FlorianBruniaux:
## Summary

- `pull_request_target` doesn't inherit write permissions by default (GitHub security model)
- Workflow was crashing with `403 Resource not accessible by integration` when trying to add the `wrong-base` label and post a comment
- Added explicit `permissions: issues: write / pull-requests: write` at the workflow level

## Root cause

`pull_request_target` runs in the context of the base repo (not the fork), so it *can* have write access — but only if explicitly granted. Without the `permissions` block, the default `GITHUB_TOKEN` is read-only.

Fixes the error seen on PR #1304.

### DIFF
--- a/.github/workflows/pr-target-check.yml
+++ b/.github/workflows/pr-target-check.yml
@@ -4,6 +4,10 @@ on:
   pull_request_target:
     types: [opened, edited]
 
+permissions:
+  issues: write
+  pull-requests: write
+
 jobs:
   check-target:
     runs-on: ubuntu-latest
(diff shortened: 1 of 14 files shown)


KuSh:
Thanks for chasing down the 403 on the pr-target-check workflow, the diagnosis was right. The same failure was fixed in #1792 (merged 2026-05-08), which moved the job onto a GitHub App token with permission-pull-requests: write, and #1871 (merged 2026-05-13) then set permissions: {} on the job so no ambient token write is granted on a pull_request_target workflow. That covers the label and comment steps without a workflow-level write block. The other files in this branch, the discover weighted savings rate, the README core team section and the docs reorganisation, all landed already through #1304. If you think the workflow still lacks a permission it needs, please comment here or reopen.

PR #1330, opened 2026-04-15
fix (copilot): hook syntax for copilot integration

visttux:
## Summary
- Fix rtk-rewrite.json Copilot hook format generated by rtk init --copilot (is using an outdated syntax)
 - Corrected 4 schema issues: 
     - Added "version": 1
     - Fixed casing PreToolUse → preToolUse
     - Renamed field command → bash
     - Renamed timeout → timeoutSec

## Test plan
- [x] Manual testing: hook output inspected using copilot CLI

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -2526,13 +2526,14 @@ fn uninstall_gemini(verbose: u8) -> Result<Vec<String>> {
 // ── Copilot integration ─────────────────────────────────────
 
 const COPILOT_HOOK_JSON: &str = r#"{
+  "version": 1,
   "hooks": {
-    "PreToolUse": [
+    "preToolUse": [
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for catching the missing schema version. `"version": 1` landed with #2101 on 2026-05-25, and the casing and field-name question was settled the other way in #3212 on 2026-08-02: an earlier revision did declare a camelCase preToolUse entry alongside the PascalCase one, but Copilot CLI registers both keys as independent hooks and runs them in sequence, so the config now keeps only the PascalCase PreToolUse form with `command` and `timeout`, which Copilot CLI was confirmed to honour directly. Closing since the file already matches, but please comment or reopen if you hit a Copilot CLI version that rejects it.

PR #1334, opened 2026-04-15
feat(pulumi): add CLI filters for preview/up/destroy/refresh/stack

wlami:
## Summary

Adds 5 TOML filters for the Pulumi CLI, matching the existing OpenTofu/ Ansible/Terraform TOML-based pattern. Filters strip known noise (header banners, View Live URLs, policy pack loading, @ progress spinners, intermediate creating/updating/deleting rows, Duration, Node.js stack trace frames) while preserving resource change rows, policy violations, diagnostics, outputs, and summaries.

Registers matching RtkRule in src/discover/rules.rs so the rewrite hook transparently proxies `pulumi <subcmd>` through `rtk pulumi`. Bumps BUILTIN_TOML filter count 59 -> 64 and updates the expected-filter list and concat-discoverability test.

## What the filters strip

  Verified against real Pulumi Cloud backend output:

  - Header banners (`Previewing update`, `Updating (stack)`, `Destroying (stack)`, `Refreshing (stack)`)
  - `View Live:` / `View in Browser:` URL lines
  - `Loading policy packs...` loader lines
  - `@ Previewing update...` / `@ Updating...` progress spinners
  - `Duration:` timing footer
  - `More information at:` / `Use `pulumi ...`` hint lines
  - Intermediate `creating (0s)` / `updating (0s)` / `deleting (0s)` / `refreshing (0s)` progress rows (final `created (0.03s)` rows preserve same info)
  - Node.js stack trace frames inside Diagnostics sections (`at /path/to/node_modules/...`, `promise: Promise`, `[Circular *1]`)
  - Redundant per-resource error-count summary rows (duplicated in `Resources:` block)
  - `Current stack is dev:` / `Owner:` / `Last updated:` metadata (stack command)

## What the filters preserve

  Per RTK's correctness-over-compression design philosophy:

  - Resource change rows (`+`/`~`/`-`/`+-`/`++`)
  - Policy pack violations (advisory/mandatory)
  - Diagnostics error messages (first line of each error)
  - Stack outputs
  - Final `Resources: + N to create` summary
  - Stack resource tree (`pulumi stack` command)

  ## Measured savings (neo-migration, ~30 AWS resources, Pulumi Cloud backend)

  | Command | Raw | Filtered | Savings |
  |---|---|---|---|
  | `pulumi up` (create from empty) | 1052w / 213L | 360w / 81L | **65.8%** |
  | `pulumi destroy` (teardown 30 resources) | 966w / 153L | 268w / 61L | **72.3%** |
  | `pulumi preview --refresh` | 861w / 123L | 511w / 81L | **40.7%** |
  | `pulumi refresh` (drift detection) | 850w / 119L | 553w / 61L | **34.9%** |
  | `pulumi stack` | 147w / 45L | 104w / 39L | **29.3%** |
  | `pulumi preview` (clean, no noise) | 348w / 76L | 316w / 60L | **9.2%** |

Savings are **input-dependent**: create/destroy lifecycle operations hit >60% because they emit substantial intermediate progress noise. Steady-state operations (clean preview, `stack`) produce mostly-signal output where aggressive compression would drop correctness — the filter deliberately stops there.

 ## Test plan

  - [x] 10 inline `[[tests.pulumi-*]]` blocks (2 per filter) — all pass via `rtk verify --filter pulumi-*`
  - [x] `BUILTIN_TOML` filter count bumped 59 → 64
  - [x] `test_builtin_all_expected_filters_present` updated with 5 new names
  - [x] `test_new_filter_discoverable_after_concat` updated 60 → 65
  - [x] `cargo clippy --all-targets` — 0 errors (only pre-existing warnings untouched)
  - [x] `cargo test` — 1449 passed, 0 failed
  - [x] Manual smoke test: `rtk pulumi preview|up|destroy|refresh|stack` against real neo-migration AWS project with 33 resources
  - [x] CHANGELOG.md updated
  - [x] README.md command list updated (new "Infrastructure as Code" section)


aeppling:
Hey @wlami , glad to have you here contributing to RTK

Here are few things to be addressed before merging :

1. Changelog changes not required, a clear commit message is enough and will be handled automatically by cicd

2. pulumi-stack match scope match_command = "^pulumi\\s+stack(\\s|$)" also matches stack export (full JSON state), stack graph (DOT). Could you narrow this to bare pulumi stack or an explicit subcommand list

3. savings_pct: 70.0 in rules.rs: Since subcmd_savings is empty, rtk discover uses 70% flat for all subcommands, but your measured data ranges 9-72% (avg ~42%). Consider populating subcmd_savings with your per-command measurements and setting savings_pct closer to ~50%.



Thanks for this addition !

wlami:
Hey, thanks for the quick review! Pushed 888d1a8, which addresses all 3 points:

  1. Reverted CHANGELOG.md: didn't know about release-please, makes sense.
  2. Narrowed `pulumi-stack` match to a subcommand whitelist. `export` and `graph` now pass through unfiltered.
  3. Lowered `savings_pct` to 45 and filled `subcmd_savings` with real numbers (up=66, destroy=72, refresh=35, preview=25, stack=29).

One extra thing I caught while looking at point 2: every pulumi subcommand supports `-j/--json`, and Rust's regex crate can't do negative lookahead. So I removed `max_lines` from all filters, the strip patterns don't match JSON structure anyway, so `pulumi preview --json` etc. now pass through clean. Verified on the real stack.

Let me know if there is anything else!

aeppling:
Hey @wlami, thanks for this PR.

One change before merge: pulumi-stack strips Current stack is <name>: and Last updated:. Confirmed on real output, stack is read-only (exit 0), so there's no tee fallback and the active-stack identity is irreversibly dropped.

It also orphans the indented Managed by <user> line under the removed header. Please keep Current stack is X: (and ideally Last updated:).

I'm not a pulumi user so open to any thoughts about this, but RTK principle is stripping noise without loosing signals.

wlami:
Good catch, fixed in 1f6e36b.

`pulumi-stack` now keeps `Current stack is X:` and `Last updated:`. Since the command is read-only (exit 0) there's no tee fallback, so dropping the active-stack identity was irreversible - agreed it's signal, not noise. Keeping the header also resolves the orphaned indented `Managed by <user>` line.

`Owner:` and `Pulumi version used:` stay stripped as noise. Updated the inline test to lock the new output. Let me know if you'd prefer `Owner:` kept too.

aeppling:
Honestly hard to know like this since i'm not working often with pulumi CLI , since you are an pulumi any thoughts on this ?

Missing signals can trigger more api turns because agents will retry with "rtk proxy" to get the raw output, which cost more in the end.

Waiting for your call, if we already filter enough, we may want to keep this to avoid this

aeppling:
@wlami sorry little missclick on the close button, PR LGTM , about the Owner filter we can drop it if rtk already correct saves on pulumi

since we have many signal lost issue and that is a challenge for rtk, i'm in favor of not taking this risk for now and keeping Owner: information

wlami:
Agree on the signal-loss risk. pulumi-stack now keeps the full identity block.

aeppling (approved):

aeppling:
@wlami LGTM, thanks for contributing to RTK!

PR #1337, opened 2026-04-16
fix: Python single-line docstring and trailing brace on function signature

### LINKED ISSUE #1322: MinimalFilter: single-line Python docstring ("""...""") leaves in_docstring=true, suppressing subsequent comment stripping
## Description

When a Python line both opens and closes a docstring on the same line (e.g. `"""Short docstring."""`), `MinimalFilter` flips `in_docstring` to `true` and emits the line but never processes the closing `"""` on that same line. Every line after it is treated as inside a docstring until the next `"""` line appears, meaning `#` comments that should be stripped are instead preserved.

## Reproduction

```rust
#[test]
fn test_minimal_python_docstring_toggle() {
    let filter = MinimalFilter;
    let input = r#"def foo():
    """Short docstring."""
    # this comment should be stripped
    x = 1
    # another comment to strip
    return x
"#;
    let result = filter.filter(input, &Language::Python);
    // Bug: in_docstring is left true after the single-line docstring,
    // so the # comments below it are preserved instead of being stripped.
    assert!(
        !result.contains("# this comment should be stripped"),
        "single-line docstring should not leave in_docstring=true"
    );
    assert!(
        !result.contains("# another comment to strip"),
        "comment after single-line docstring should be stripped"
    );
}
```

Running this test exposes the bug: both `# this comment should be stripped` and `# another comment to strip` appear in the output because `in_docstring` is `true` after the `"""Short docstring."""` line, so the `# ...` lines hit the `if in_docstring { keep line }` branch rather than the comment-stripping branch.

## Root cause

In `MinimalFilter::filter` (`src/core/filter.rs`), the Python docstring branch is:

```rust
if *lang == Language::Python && trimmed.starts_with("\"\"\"") {
    in_docstring = !in_docstring;
    result.push_str(line);
    result.push('\n');
    continue;
}
```

This unconditionally toggles `in_docstring` whenever a line *starts with* `"""`. For a single-line docstring like `"""Short docstring."""`, the opening `"""` flips the flag to `true`, the line is emitted, and execution moves on. The closing `"""` on the same line is never seen because the code already called `continue`. Every subsequent line is then considered inside a docstring.

The fix should check whether the same line also *contains a closing* `"""` after the opening one (i.e. the docstring both opens and closes on one line) and, if so, keep `in_docstring` at `false`.

## Notes

Found while running [tailtest](https://github.com/avansaber/tailtest) on this codebase. The generated test suite documents the current (buggy) behavior -- the assertions will fail once the bug is fixed.

### LINKED ISSUE #1323: AggressiveFilter: trailing { on function signature skips brace counting, leaking let-bindings from function bodies
## Description

When a function signature includes the opening brace on the same line (e.g. `fn compute(x: i32) -> i32 {`), the `FUNC_SIGNATURE` branch emits the line and calls `continue` before the brace-counting code runs. This leaves `brace_depth` at 0 after the signature, causing the first body line to immediately reset `in_impl_body = false`, so subsequent `let` bindings leak through the "Keep type definitions" branch.

## Reproduction

```rust
#[test]
fn test_aggressive_brace_tracking() {
    let filter = AggressiveFilter;
    let input = r#"fn compute(x: i32) -> i32 {
    let result = x * 2;
    let extra = result + 1;
    extra
}
"#;
    let result = filter.filter(input, &Language::Rust);
    // Bug: brace_depth stays 0 after the signature line because the
    // FUNC_SIGNATURE branch calls `continue` before brace counting runs.
    // The body's `let` lines then match the "Keep type definitions" branch
    // and leak into the output.
    assert!(
        !result.contains("let result = x * 2;"),
        "let binding inside function body should be filtered out"
    );
    assert!(
        !result.contains("let extra = result + 1;"),
        "let binding inside function body should be filtered out"
    );
}
```

Running this test exposes the bug: both `let result` and `let extra` appear in the aggressive-filtered output because they escape through the `trimmed.starts_with("let ")` branch that is intended only for top-level bindings.

## Root cause

In `AggressiveFilter::filter` (`src/core/filter.rs`), the signature branch is:

```rust
if FUNC_SIGNATURE.is_match(trimmed) {
    result.push_str(line);
    result.push('\n');
    in_impl_body = true;
    brace_depth = 0;
    continue;   // <-- jumps past brace-counting below
}

// Track brace depth for implementation bodies
let open_braces = trimmed.matches('{').count();
let close_braces = trimmed.matches('}').count();
```

When the signature line ends with `{` (e.g. `fn compute(x: i32) -> i32 {`), `continue` is called before `open_braces` is counted. `brace_depth` stays at 0. The very next line hits:

```rust
if brace_depth <= 0 {
    in_impl_body = false;
    ...
}
```

...which turns off `in_impl_body` immediately. Subsequent `let` lines fall through to the "Keep type definitions" section and are emitted.

The fix is to count the braces on the signature line itself before (or instead of) calling `continue`, so that a trailing `{` is reflected in `brace_depth`.

## Notes

Found while running [tailtest](https://github.com/avansaber/tailtest) on this codebase. The generated test suite documents the current (buggy) behavior -- the assertions will fail once the bug is fixed.

ekkoitac:
## Description

Fix two bugs in `src/core/filter.rs`:

### 1. MinimalFilter: Python single-line docstrings

Single-line docstrings like `\`\`\`"Short docstring."\`\`\`` now correctly keep `in_docstring=false`. Previously, the opening `\`\`\`` toggled the flag to `true` and the line was emitted via `continue`, so the closing `\`\`\`` on the same line was never processed. Every subsequent line was treated as inside a docstring, suppressing comment stripping.

**Root cause:** The Python docstring branch unconditionally toggled `in_docstring` on any line starting with `\`\`\``, without checking if the same line also closed the docstring.

**Fix:** Check if the line both starts and ends with `\`\`\`` — if so, it is a single-line docstring and we skip the toggle.

### 2. AggressiveFilter: Trailing `{` on function signature

Function signatures with trailing `{` on the same line (e.g. `fn compute(x: i32) -> i32 {`) now correctly count the brace before `continue`. Previously, the `FUNC_SIGNATURE` branch called `continue` before brace counting ran, leaving `brace_depth=0`. This caused the first body line to set `in_impl_body=false`, letting `let` bindings leak through the "Keep type definitions" branch.

**Root cause:** The signature branch reset `brace_depth = 0` and called `continue` before the brace-counting code, so a trailing `{` on the signature line was never accounted for.

**Fix:** Count opening and closing braces on the signature line before `continue`, so `brace_depth` reflects any trailing `{`.

## Tests

Both fixes include regression tests:
- `test_minimal_python_single_line_docstring`
- `test_aggressive_filter_trailing_brace_on_signature`

## Related Issues

- Fixes #1322
- Fixes #1323

### DIFF
--- a/src/core/filter.rs
+++ b/src/core/filter.rs
@@ -188,6 +188,15 @@ impl FilterStrategy for MinimalFilter {
 
             // Handle Python docstrings (keep them in minimal mode)
             if *lang == Language::Python && trimmed.starts_with("\"\"\"") {
+                // Single-line docstring: both opens and closes on this line
+                // e.g. """Short docstring."""
+                if trimmed.ends_with("\"\"\"") {
+                    // Single-line docstring — don't toggle, just keep the line
+                    result.push_str(line);
+                    result.push('\n');
+                    continue;
+                }
+                // Multi-line docstring start or end — toggle the flag
                 in_docstring = !in_docstring;
                 result.push_str(line);
                 result.push('\n');
@@ -265,10 +274,14 @@ impl FilterStrategy for AggressiveFilter {
 
             // Always keep function/struct/class signatures
             if FUNC_SIGNATURE.is_match(trimmed) {
+                // Count braces on the signature line (e.g. "fn foo() {")
+                // before continuing, so a trailing '{' is reflected in brace_depth
+                let sig_open = trimmed.matches('{').count() as i32;
+                let sig_close = trimmed.matches('}').count() as i32;
                 result.push_str(line);
                 result.push('\n');
                 in_impl_body = true;
-                brace_depth = 0;
+                brace_depth = sig_open - sig_close;
                 continue;
             }
 
@@ -467,6 +480,33 @@ mod tests {
     }
 
     #[test]
+    fn test_aggressive_filter_trailing_brace_on_signature() {
+        // When a function signature has trailing '{' on the same line,
+        // brace_depth must be counted before continuing, so the function body
+        // is correctly identified as implementation.
+        let input = r#"fn compute(x: i32) -> i32 {
+    let result = x * 2;
+    let extra = result + 1;
+    extra
+}
+"#;
+        let filter = AggressiveFilter;
+        let result = filter.filter(input, &Language::Rust);
+        assert!(
+            !result.contains("let result = x * 2;"),
+            "let binding inside function body should be filtered out"
+        );
+        assert!(
+            !result.contains("let extra = result + 1;"),
+            "let binding inside function body should be filtered out"
+        );
+        assert!(
+            result.contains("fn compute(x: i32) -> i32 {"),
+            "function signature with trailing brace should be kept"
+        );
+    }
+
+    #[test]
     fn test_minimal_filter_removes_comments() {
         let code = r#"
 // This is a comment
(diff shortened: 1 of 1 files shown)


KuSh:
Closing in favour of #1328 and #1329, which fix the same two bugs in src/core/filter.rs and have their CLA signed. The single-line docstring check here uses ends_with, so a bare triple-quote line that opens or closes a multi-line docstring is also treated as single-line and in_docstring never toggles back. If you would rather carry the work here, comment on #1328 and we can compare.

PR #1357, opened 2026-04-17
fix(ls): force C locale for consistent date parsing across locales

xingzhicn:
## Problem
The `ls` command returns "(empty)" on systems with non-English locales (Chinese, French, German, etc.) because the date parsing regex only matches English month names.

## Root Cause
The regex in `ls.rs` only matches English month abbreviations:
```rust
r"\s+(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s+..."
```

On systems with non-English locales, `ls -la` outputs localized month names (e.g., "4月" in Chinese, "fév" in French), causing the regex to fail. This makes `parse_ls_line()` return `None` for all files, resulting in an empty directory listing.

## Solution
Force `LC_ALL=C` environment variable when executing the `ls` command to ensure consistent English output format across all locales.

```rust
let mut cmd = resolved_command("ls");
cmd.env("LC_ALL", "C");  // Force C locale for consistent date format
```

## Testing
- Added test case `test_locale_independence()` to verify English date format parsing
- Manually tested on Chinese locale (LANG=zh_CN.UTF-8) - confirmed fix works
- **Note**: Unable to run full test suite locally (no Rust environment), but changes are minimal and follow established patterns

## Known Limitation
This fix assumes `ls` respects `LC_ALL=C` on all platforms. This should work on macOS, Linux, and most Unix-like systems. Windows users typically don't use `ls` directly.

## Checklist
- [x] Code changes made
- [x] Test case added
- [x] CHANGELOG.md updated
- [ ] cargo fmt, clippy, test (unable to run locally - will be verified by CI)

## Related
Fixes the issue where RTK `ls` shows "(empty)" on non-English systems while `rtk proxy ls` shows files correctly.


xingzhicn:
Closing this PR due to merge conflicts. Reopening with clean branch in new PR.

PR #1374, opened 2026-04-17
feat(js): add Bun runtime support (bun test, run, bunx)

HenriqueSchroeder:
## Summary

Adds Bun runtime support to RTK with filters for `bun test`, `bun run`, `bun install`, `bun build`, and `bunx`.

### What's new

- **`bun test` filter**: strips passing tests, keeps failures with error context, compact summary. Handles both TTY (`✓`/`✗`) and piped (`(pass)`/`(fail)`) formats. Buffers lines preceding a fail marker since Bun prints error context _before_ the marker, then flushes them on match.
- **`bun install` filter**: strips progress bars, keeps summary.
- **`bun build` filter**: compact build output.
- **`bun run <script>` filter**: passes through script output with error-focused filtering.
- **`bunx` passthrough** with intelligent routing (tsc, eslint, prisma → specialized filters).
- Discovery rules so `bunx tsc`, `bunx eslint`, etc. get rewritten to specialized RTK filters.

### Token savings

All filters verified `>= 60%` savings against real `bun test` output (see `tests/fixtures/bun_test_real.txt`, 140 lines of real bun output with 79 pass / 2 fail).

### Tests

- Snapshot tests for all filter variants
- Token savings assertions with real fixtures (`bun_test_real.txt`)
- Clap routing tests for `bun test/run/install/build` subcommands and `bunx`

## Test plan

- [ ] `cargo test` — 1372 tests pass locally
- [ ] `cargo clippy --all-targets` — no new warnings in `bun_cmd.rs`
- [ ] `cargo fmt --all --check` — clean
- [ ] Manual: `rtk bun test` in a Bun project verifies end-to-end filtering

## Notes

Branch is based on an older point in history (pre-0.37.0). Merge conflicts with current `develop` (notably `src/discover/rules.rs` where Bun entries need to be combined with the expanded `npx/pnpm exec` patterns, and `src/main.rs` PrismaCommands derive) are expected and will need resolution before merge.


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟡 **Risk** | `medium` |

### Summary

Adds comprehensive Bun runtime support to RTK, including filters for bun test, bun install, bun build, bun run, and bunx commands. Implements intelligent routing for bunx to specialized filters (tsc, eslint, prisma) and includes snapshot tests, token savings assertions, and clap routing tests.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

---

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,539 @@
+//! Filters Bun output — test results, install logs, build summaries, script output.
+
+use crate::core::runner;
+use crate::core::utils::{resolved_command, strip_ansi};
+use anyhow::Result;
+use lazy_static::lazy_static;
+use regex::Regex;
+use std::ffi::OsString;
+
+lazy_static! {
```
KuSh, line 10: All those regex seems fragile and requires bun not to change its console output format.
bun doesn't have ([for now?](https://github.com/oven-sh/bun/issues/6250)) a json output format, but it has a junit one we could use to make that much robust

src/discover/rules.rs
```diff
@@ -70,6 +70,24 @@ pub const RULES: &[RtkRule] = &[
...
+        pattern: r"^bun\s+(test|run|install|build|add|remove)",
+        rtk_cmd: "rtk bun",
+        rewrite_prefixes: &["bun"],
+        category: "PackageManager",
+        savings_pct: 80.0,
+        subcmd_savings: &[("test", 88.0), ("install", 85.0), ("build", 80.0)],
+        subcmd_status: &[],
+    },
+    RtkRule {
+        pattern: r"^bunx\s+",
```
KuSh: We could also handle `bun x` which is an alias to bunx
```suggestion
        pattern: r"^(bun\s+x|bunx)(\s|$)",
```

Same goes for specific command bellow

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,539 @@
...
+    Build,
+    Run { script: String },
+}
+
+pub fn run(cmd: BunCommand, args: &[String], verbose: u8, skip_env: bool) -> Result<i32> {
+    match cmd {
+        BunCommand::Test => run_test(args, verbose),
+        BunCommand::Install => run_install(args, verbose, skip_env),
+        BunCommand::Build => run_build(args, verbose),
+        BunCommand::Run { script } => run_script(&script, args, verbose, skip_env),
```
KuSh: why not just make those function public and call them from main as it's done elsewhere ? This avoids passing BunCommand down

Take this opportunity to shorten function naming and just call them, test, install, build or run (for run_script)

src/main.rs
```diff
@@ -1916,6 +1960,103 @@ fn run_cli() -> Result<i32> {
...
+        },
+
+        Commands::Bunx { args } => {
+            if args.is_empty() {
+                anyhow::bail!("bunx requires a command argument");
+            }
+
+            // Intelligent routing: delegate to specialized filters
+            match args[0].as_str() {
+                "tsc" | "typescript" => tsc_cmd::run(&args[1..], cli.verbose)?,
```
KuSh: you seem to have missed at least jest, vitest, biome

KuSh (changes requested):

src/cmds/js/bun_cmd.rs
HenriqueSchroeder: Thanks for the suggestion! I investigated the JUnit reporter approach.
Unfortunately, bun's JUnit output (tested on 1.2.x and 1.3.x) produces empty <failure> elements:

```xml
 <testcase name="...">
   <failure type="AssertionError" />
 </testcase>
```
 
No message, no Expected/Received diff, no stack trace — just the exception type.

The current regex parser captures the actual failure context (test name, error message, expected vs received, surrounding lines) that would be lost with the JUnit approach. For now, I believe the best approach is to keep things as they are until there's proper support for it.

src/core/utils.rs
```diff
@@ -287,6 +289,11 @@ pub fn package_manager_exec(tool: &str) -> Command {
     } else {
         let pm = detect_package_manager();
         match pm {
+            "bun" => {
+                let mut c = resolved_command("bunx");
+                c.arg(tool);
```
KuSh:
```suggestion
                c.arg("--").arg(tool);
```

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,522 @@
...
+        }
+
+        // Failure marker — flush error buffer as context
+        if is_fail_marker(trimmed) {
+            has_failures = true;
+            for buf_line in &error_buffer {
+                failures.push(buf_line.to_string());
+            }
+            error_buffer.clear();
+            failures.push(line.to_string());
```
KuSh:
```suggestion
            failures.push(trrimmed.to_string());
```

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,522 @@
...
+        {
+            error_buffer.clear();
+            summary_lines.push(trimmed);
+            continue;
+        }
+
+        // Skip empty lines (don't buffer)
+        if trimmed.is_empty() {
+            continue;
+        }
```
KuSh: Why is this not done upfront directly after the trimming? It could avoid a condition in `if in_failed_summary` block

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,522 @@
...
+    if !summary_lines.is_empty() {
+        for line in &summary_lines {
+            result.push(line.to_string());
+        }
+    } else {
+        result.push("bun test: completed".to_string());
+    }
+
+    let output = result.join("\n");
+    if output.trim().is_empty() {
```
KuSh:
```suggestion
    let output = result.join("\n").trim();
    if output.is_empty() {
```

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,522 @@
...
+        }
+        // Skip progress indicators
+        if trimmed.contains("⸩") || trimmed.contains("⸨") {
+            continue;
+        }
+        // Skip resolution/download progress
+        if trimmed.starts_with("Resolving")
+            || trimmed.starts_with("Downloading")
+            || trimmed.starts_with("Extracting")
+        {
```
KuSh, line 295: Could you test bun in non English environment? Something like LC_ALL=de_DE.UTF-8 to see if bun has i18n output.

If that's the case, we should run it with `env("LC_ALL", "C")` to ensure consistent output whatever user's config

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,522 @@
...
+        // Skip bun run boilerplate ($ command echo)
+        if trimmed.starts_with("$") && trimmed.len() < 200 {
+            continue;
+        }
+        // Skip bun version header
+        if trimmed.starts_with("bun run v") {
+            continue;
+        }
+
+        result.push(line.to_string());
```
KuSh, line 381:
```suggestion
        result.push(trimmed.to_string());
```

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,522 @@
...
+    }
+
+    if result.is_empty() {
+        "ok".to_string()
+    } else {
+        result.join("\n")
+    }
```
KuSh:
```suggestion
    let output = result.join("\n").trim();
    if output.is_empty() {
        "ok".to_string()
    } else {
        output
    }
```

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,522 @@
...
+        if trimmed.contains("built") || trimmed.contains("Bundle") || trimmed.starts_with("Done") {
+            result.push(trimmed.to_string());
+        }
+    }
+
+    if result.is_empty() {
+        "bun build: ok".to_string()
+    } else {
+        result.join("\n")
+    }
```
KuSh:
```suggestion
    let output = result.join("\n").trim();
    if output.is_empty() {
        "build: ok".to_string()
    } else {
        output
    }
```

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,522 @@
...
+
+        // Keep: package count, warnings, errors, lockfile info
+        result.push(trimmed.to_string());
+    }
+
+    if result.is_empty() {
+        "bun install: ok".to_string()
+    } else {
+        result.join("\n")
+    }
```
KuSh:
```suggestion
    let output = result.join("\n").trim();
    if output.is_empty() {
        "install ok".to_string()
    } else {
        output
    }
```

src/discover/rules.rs
```diff
@@ -107,9 +125,11 @@ pub const RULES: &[RtkRule] = &[
         subcmd_status: &[],
     },
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?tsc(\s|$)",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx|bunx|bun\s+x)\s+)?tsc(\s|$)",
         rtk_cmd: "rtk tsc",
         rewrite_prefixes: &[
+            "bun x tsc",
+            "bunx tsc",
```
KuSh, line 152: Missing `bun tsc` and `bun run tsc`
Same goes for other commands

src/main.rs
```diff
@@ -932,6 +945,37 @@ enum KubectlCommands {
...
+    },
+    /// Run script with filtered output
+    Run {
+        /// Script name and arguments
+        #[arg(trailing_var_arg = true, allow_hyphen_values = true)]
+        args: Vec<String>,
+    },
+    /// Passthrough: runs any unsupported bun subcommand directly
+    #[command(external_subcommand)]
+    Other(Vec<OsString>),
```
KuSh, line 1158: the rewrite should take care of most cases but you should special case `bun x` in case a tool directly call it

src/main.rs
```diff
@@ -1916,6 +1963,96 @@ fn run_cli() -> Result<i32> {
...
+                        }
+                    } else {
+                        let timer = core::tracking::TimedExecution::start();
+                        let status = core::utils::resolved_command("bunx")
+                            .arg("prisma")
+                            .status()
+                            .context("Failed to run bunx prisma")?;
+                        timer.track_passthrough("bunx prisma", "rtk bunx prisma (passthrough)");
+                        core::utils::exit_code_from_status(&status, "bunx prisma")
+                    }
```
KuSh: Could you factorize that to avoid duplicating code?

src/main.rs
```diff
@@ -1916,6 +1963,96 @@ fn run_cli() -> Result<i32> {
...
+                        cmd.arg(arg);
+                    }
+                    let status = cmd.status().context("Failed to run bunx")?;
+                    let args_str = args.join(" ");
+                    timer.track_passthrough(
+                        &format!("bunx {}", args_str),
+                        &format!("rtk bunx {} (passthrough)", args_str),
+                    );
+                    core::utils::exit_code_from_status(&status, "bunx")
+                }
```
KuSh: A `bun_cmd::exec()` like in npm_cmd would be much appreciated

KuSh (changes requested):

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,538 @@
...
+        }
+
+        // Keep: package count, warnings, errors, lockfile info
+        result.push(trimmed.to_string());
+    }
+
+    let joined = result.join("\n");
+    let output = joined.trim();
+    if output.is_empty() {
+        "bun install: ok".to_string()
```
KuSh: we could reduce output to just ok everywhere. the llm knows what command is running

KuSh (changes requested):

src/discover/rules.rs
```diff
@@ -242,9 +284,15 @@ pub const RULES: &[RtkRule] = &[
         subcmd_status: &[],
     },
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?jest(\s+run)?(\s|$)",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx|bunx|bun\s+x|bun)\s+)?jest(\s+run)?(\s|$)",
```
KuSh: why not `bun(\s+run)?run\s+jest(\s+run)?`

src/discover/rules.rs
```diff
@@ -282,9 +330,15 @@ pub const RULES: &[RtkRule] = &[
         subcmd_status: &[],
     },
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?vitest(\s+run)?(\s|$)",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx|bunx|bun\s+x|bun)\s+)?vitest(\s+run)?(\s|$)",
```
KuSh: missing `bun(\s+run)?run\s+vitest(\s+run)?`

src/discover/rules.rs
```diff
@@ -347,9 +405,11 @@ pub const RULES: &[RtkRule] = &[
         subcmd_status: &[],
     },
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?prisma",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx|bunx|bun\s+x)\s+)?prisma",
```
KuSh: missing `bun(\s+run)?run\s+prisma`

KuSh (changes requested):

bZichett:
Bumping.  Great to see this coming soon.  Much appreciated library btw.  Been using for a few weeks with success.

Cargo.lock
KuSh, line 1: Please avoid messing with lock file

KuSh (changes requested):
Some previous comments have not been taken into account or answered, conflicts have to be handled

src/cmds/js/bun_cmd.rs
HenriqueSchroeder: Didn't spin up a de_DE box, but added `LC_ALL=C` to test/install/build defensively since it's free and removes the locale-dependent parsing risk. Left it off `bun run` on purpose: that runs an arbitrary user script, and forcing the locale down the child tree changes its behavior (a Python step dropping to ASCII stdio, etc.). We don't parse that output structurally anyway.

src/cmds/js/bun_cmd.rs
HenriqueSchroeder: Kept `line` here on purpose. `bun run` just forwards a script's own stdout, and trimming flattens indentation that carries meaning (stack traces, trees, tables). The install/build filters trim because their output is structured; this one is arbitrary passthrough.

src/discover/rules.rs
HenriqueSchroeder: All four forms are covered (`bun jest`, `bun run jest`, `bunx jest`, `bun x jest`) through the shared prefix group (`bunx|bun\s+x|bun(\s+run)?`). Didn't take `bun(\s+run)?run\s+jest` literally because `(\s+run)?run` double-matches (`bunrun` / `bun run run`); `bun(\s+run)?` handles both `bun jest` and `bun run jest` cleanly.

src/discover/rules.rs
HenriqueSchroeder: Covered. Same `bunx|bun\s+x|bun(\s+run)?` prefix on the vitest rule, so `bun run vitest` and the other three forms match.

src/discover/rules.rs
HenriqueSchroeder: Same story. The prisma rule got the `bunx|bun\s+x|bun(\s+run)?` prefix, covering `bun prisma`, `bun run prisma`, `bunx prisma`, `bun x prisma`.

pszymkowiak:
Reviewed this locally (build + the 17 bun tests pass, clippy clean) — solid work, and it's the most complete of the open Bun PRs: the dedicated `bun test` failure extraction (buffering the error context Bun prints *before* the fail marker) is exactly where the token savings are, and the real-output fixtures back it up. Two mechanical things before merge:

**1. Rebase needed — the branch is stale.** On the current branch `cargo test` fails to compile the test binary on unrelated dead-code-as-error (`FILTERS_TOML` in `constants.rs`, `load` in `toml_filter.rs`) that develop has since resolved. I confirmed it's purely staleness: after merging `develop` in, the build is clean and all 17 bun tests pass. A rebase onto `develop` fixes it.

**2. Coordinate the `vitest_cmd.rs` refactor with #2418.** This PR decouples `run_test` from the `Commands` enum via a new `TestFramework` enum — #2418 does the same refactor. Whichever merges first, the other will need to drop/reconcile that part to avoid a conflict.

Substance is good — happy to re-review once rebased.

aeppling:
Thanks for the rebase. The review threads are in good shape: I re-checked the 19 resolved ones against the code and they have all landed (`bun x` alias and `bun run <tool>` prefixes, public `test/install/build/run`, `route_bunx` coverage, `"ok"` outputs, `LC_ALL=C`, lock file untouched).

Two blockers before this can go green, both from develop moving under the merge:

1. **`lazy_static` was removed from the crate** (`5269df7`). `bun_cmd.rs:6` fails to import it and all the `BUN_*_RE` statics are undefined. Use `std::sync::LazyLock` like `playwright_cmd.rs:167`:
   ```rust
   static BUN_FAIL_MARKER_RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"...").unwrap());
   ```
2. **`RtkRule` gained `pipeline_final_safe`** (`523c803`) and now has a `Default`. The two new bun rules in `rules.rs` (around lines 105 and 114) need `..RtkRule::default()` like their neighbors.

With those the test binary should build again. Please run `cargo fmt --all --check && cargo clippy --all-targets && cargo test` on the merged head before pushing; it currently fails with 15 errors.

Two threads are still open with a reply but no resolution; a one-line confirmation each would let them be closed:
- non-English locale: `LC_ALL=C` on test/install/build without having tested `de_DE` is fine by me, just say so explicitly in the thread.
- `bun run` keeping untrimmed lines: agreed, it is passthrough; note it in the thread.

Coordination: #2418 carries the same `TestFramework` refactor of `vitest_cmd.rs`. Whichever merges first, the other rebases and drops that hunk; worth a note on #2418 so it is not a surprise.

Once the build is green I am happy to re-review. The `bun test` failure extraction is the valuable part here and the real-output fixtures back it.

aeppling:
## Final pass — CONTRIBUTING / CODING_PRACTICES + fuzz testing

Thanks for addressing the previous review — public fns, `route_bunx`, `bun x` handling, LazyLock regexes and the rewrite coverage are all in good shape now.

Reviewed head `c3c2257` one last time against the guidelines and fuzz-tested the binary against a real Bun 1.3.14 project (pass/fail/todo/skip, 250 failures, timeouts, snapshots, SIGKILL, `--bail`, `--coverage`, bare `bun run`/`bunx`, build/install). fmt, clippy, tests green. Routing, rewrite rules and refactors are solid.

### Blockers

**1. Shared infrastructure not used.** `bun_cmd.rs` bypasses the core building blocks every filter relies on: no `with_tee` on any `run_filtered`, no `CAP_*` cap, no `force_tee_hint`. Result: data is dropped with no way to recover it (250 failures → 2996 lines uncapped; build output file name, coverage table, `$`-prefixed script lines, `Bailed out`, snapshot/todo counts all silently gone). CONTRIBUTING "Never Block": hidden data requires a recovery hint. See `pytest_cmd.rs:286-294` and `tsc_cmd.rs:40` for the pattern.

**2. Bare `bun run` / `bunx` / `bun x` become rtk errors.** Real commands list scripts / print usage; rtk bails. Pass through like `rtk npm run` does.

**3. `ok` printed on crash.** SIGKILL in a test → Bun prints header only, exit 137 → rtk prints `ok`. Filter never sees the exit code (`run_filtered` vs `run_filtered_with_exit`).

### Must fix

4. `bun_build.txt` / `bun_install.txt` fixtures are not real Bun output; build regex never matches the real `name  N bytes  (entry point)` line. Replace with real captures.
5. Summary regexes miss ` 1 snapshots, 6 expect() calls`, ` 1 todo`, `snapshots: +1 added`, `Bailed out after N failure`.
6. `filter_bun_test` trims lines: caret `^` loses column, `toEqual` diff loses indentation.
7. `filter_bun_run` drops every `$`-prefixed line; Bun echoes one `$ cmd` line, strip only that.
8. `--coverage` requested → table dropped.

Requesting changes. Blocker 1 gates this on its own. Happy to re-run the same fuzz set on the next revision.

HenriqueSchroeder:
All eight are in, pushed as `5804705`. Ran the same kind of pass you did: fmt, clippy and the 2757 tests are green, and I drove the built binary against a real bun 1.3.14 project.

**Blocker 1.** Every run goes through `run_filtered_with_exit` with a tee label now. Failures are grouped into blocks (context + marker) and capped at `CAP_WARNINGS`, with the remainder behind a tee hint; install caps its package list and build its artifact list at `CAP_LIST`. A 40-failure run gives 10 blocks, the overflow line and the hint, with the summary intact.

**Blocker 2.** Bare `bun run`, `bunx` and `bun x` pass through, so they print bun's own script list and usage instead of an rtk error.

**Blocker 3.** `finish()` reports the exit code when the filter kept nothing and the command failed, for all four filters. A SIGKILLed run prints `bun test exited with code 137` plus the tee hint, with 137 propagated. One nuance: when the raw output is just the banner, `never_worse` keeps bun's raw line instead, which is fine, it still isn't `ok` and the code still propagates.

**4.** Replaced every bun fixture with a real capture from 1.3.14, not just build and install. The build regex now matches `index.js  72 bytes  (entry point)`. Worth recording: 1.3.14 no longer prints a line per passing test, so the old 1.3.10 fixture with its 79 `(pass)` lines is kept to cover that format.

**5.** The summary parser picks up `1 todo`, `snapshots: +1 added`, the `N snapshots, N expect() calls` variant and `Bailed out after N failure`.

**6.** Context lines are kept verbatim. There is a test that measures the caret indentation, so a future trim will not slip back in.

**7.** Only the single `$ cmd` line bun echoes is dropped; a script printing its own `$`-prefixed lines keeps them.

**8.** The coverage table survives.

While I was in there I answered the locale question with actual data: on 1.3.14 the output of `bun test`, `bun install` and `bun build` is byte identical under `LC_ALL=C`, `pt_BR.utf8` and `de_DE.UTF-8`, decimals dot separated in all three. `de_DE` is not generated on my box so it falls back to C; `pt_BR` is the real one. Keeping `LC_ALL=C` as cheap insurance.

Note dropped on #2418.

aeppling:
Thanks, re-ran the same fuzz set on `5804705`. Blockers 1-3 and points 4-6, 8 verified fixed: tee hints, caps, `run_filtered_with_exit`, SIGKILL propagates 137 without `ok`, bare `bun run`/`bunx` pass through, caret/diff context intact, todo/snapshot/bail/coverage kept, real fixtures.

One must-fix left, #7: the "drop first `$` line" heuristic drops the wrong line. Bun writes the `$ cmd` echo to stderr and the script to stdout, and stdout lands first in the capture. 8/8 runs dropped the script's `$HOME is not set` and kept the echo. Inject `--silent` before the script (only suppresses the echo, script output verified identical) and remove the `$` heuristic.

Approve once that's in.

HenriqueSchroeder:
Good catch on the ordering, that is exactly it. Reproduced before touching anything: bun writes the echo to stderr and the script to stdout, stdout lands first in the capture, and `rtk bun run` dropped the script line 5/5.

Fixed in `3d30c06` your way: `bun run --silent` before the script name, and the `$` heuristic is gone, so the filter no longer looks at `$`-prefixed lines at all. `run_args` builds the argv and a test pins `--silent` ahead of the script, since after it bun would forward the flag to the script instead.

Verified 8/8 with a script printing `$HOME is not set`: output is now byte identical to `bun run --silent` run directly. fmt, clippy and 2758 tests green.

aeppling:
All blockers and must-fixes resolved. LGTM.

@KuSh still pending changes

src/main.rs
```diff
@@ -1579,6 +1623,39 @@ where
     }
 }
 
+fn route_bunx(args: &[String], verbose: u8) -> Result<i32> {
```
KuSh, line 1626: Re-raising the factorization point — my earlier comment on this is now collapsed as outdated, and the duplication has since drifted.

`route_bunx` and the `Commands::Npx` arm (~line 2425) are the same tool→filter table, and they no longer agree:

| | `route_bunx` | `Commands::Npx` |
|---|---|---|
| jest / vitest / biome | present | **missing** |
| prisma fallthrough | `bun_cmd::run_tool` | ~30 lines of inlined passthrough |
| final fallback | `bun_cmd::run_tool` | `npm_cmd::exec` |

The drift is the point: the missing arms are exactly what I flagged earlier, and the `&args[1..]` slicing bug on the `eslint`/`biome` arm below exists identically in the npx copy — one bug, copied.

Something like `route_js_tool(args, verbose, fallback)` with npx passing `npm_cmd::exec` and bunx passing `bun_cmd::run_tool` fixes both call sites at once and stops the next tool from being registered in only one of them. The npx prisma branch should call that fallback too rather than hand-rolling its own passthrough.

src/main.rs
```diff
@@ -1579,6 +1623,39 @@ where
...
+    let Some(tool) = args.first().map(|s| s.as_str()) else {
+        // Bare `bunx` / `bun x` prints its usage; don't turn that into an rtk error.
+        return bun_cmd::run_tool(args, verbose);
+    };
+    match tool {
+        "tsc" | "typescript" => tsc_cmd::run(&args[1..], verbose),
+        "eslint" | "biome" => lint_cmd::run(&args[1..], verbose),
```
KuSh, line 1633: This strips the linter name, but `lint_cmd::run` expects it as `args[0]` — `detect_linter` reads `args[0]`, and `rtk lint biome` is the documented form.

So `rtk bunx biome check .` calls `lint_cmd::run(["check", "."])`, `detect_linter` treats `check` as the linter, and rtk shells out to `bunx -- check .`. Verified end-to-end: bunx answers `could not determine executable to run for package check`, and rtk then reports `Lint: 1 errors, 0 warnings`. biome never runs and the failure is masked.

```suggestion
        "eslint" | "biome" => lint_cmd::run(args, verbose),
```

Note `bunx eslint src` misroutes the same way (`bunx -- src`), mirroring the same slicing in the `Commands::Npx` arm — see the factorization thread.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,878 @@
...
+        || BUN_TEST_EXPECT_RE.is_match(line)
+        || BUN_SNAPSHOT_RE.is_match(line)
+        || line.starts_with("Ran ")
+        || line.starts_with("Bailed out after ")
+}
+
+/// A row of the `--coverage` table. Every row (header, separator and per-file)
+/// carries the three column separators, which ordinary test output does not.
+fn is_coverage_row(line: &str) -> bool {
+    line.matches('|').count() >= 3
```
KuSh, line 327: This misfires on bun's own source frames. bun prints failure context as `NN | <source>`, so any frame whose source contains `||` (two pipes plus the gutter pipe = 3) or a TS union type is classified as a `--coverage` row: `buffer.clear()` throws away every preceding context line of that failure, and the line is then re-emitted *after* the FAILURES block.

Reproduced with a test failing on `const name = user.name || user.nick || "anon";` — frame lines 10-13 were dropped, the failure block started at line 14, and line 13 reappeared under the coverage section. `||` is ubiquitous in JS, so this fires often.

The coverage table is a contiguous block with a `---|---` separator and fixed headers; keying on that (or on being inside a detected coverage section) would be much safer than a bare pipe count.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,878 @@
...
+
+fn is_test_file_header(line: &str) -> bool {
+    line.ends_with(':') && (line.contains(".test.") || line.contains(".spec."))
+}
+
+fn is_summary_line(line: &str) -> bool {
+    BUN_TEST_COUNT_RE.is_match(line)
+        || BUN_TEST_EXPECT_RE.is_match(line)
+        || BUN_SNAPSHOT_RE.is_match(line)
+        || line.starts_with("Ran ")
```
KuSh, line 320: `BUN_TEST_COUNT_RE` was deliberately anchored with `$` so app output isn't read as the `N pass` summary, but this prefix check (and `"Bailed out after "` below) is still open.

Reproduced: a test logging `Ran 5 migrations before assert` before failing had that line hoisted into the summary section, and `buffer.clear()` then dropped the failure's code frame and caret. Worth anchoring the same way, e.g. `^Ran \d+ tests?\b`.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,878 @@
...
+        let trimmed = line.trim();
+
+        if trimmed.is_empty() {
+            continue;
+        }
+        if BUN_BANNER_RE.is_match(trimmed) {
+            continue;
+        }
+        // Keep errors and warnings
+        if trimmed.contains("error")
```
KuSh, line 396: A build failure loses its location. Only lines containing `error`/`warn`/`Error`/`Warn`, artifact lines and summary lines survive; bun's code frame, the caret, and the `at /path/file.ts:1:15` line match none of those and are dropped.

Verified: frame + caret + `error: Could not resolve: "missing"` + `at /home/u/p/index.ts:1:15` filters down to just the `error:` line. The file and line number are exactly what the LLM needs to act, so keeping the `at <path>:<line>:<col>` line (and ideally the frame) would be worth the tokens.

src/cmds/js/bun_cmd.rs
```diff
@@ -0,0 +1,878 @@
...
+/// trimmed either, this is passthrough of arbitrary output where indentation
+/// is content.
+fn filter_bun_run(output: &str, exit_code: i32) -> String {
+    let clean = strip_ansi(output);
+    let mut result = Vec::new();
+
+    for line in clean.lines() {
+        let trimmed = line.trim();
+
+        if trimmed.is_empty() {
```
KuSh, line 447: This drops all blank lines, which contradicts the reasoning used to keep `line` rather than `trimmed` a few lines down ("this is passthrough of arbitrary output where indentation is content"). If indentation is content here, so are the paragraph breaks.

`Section A\n\n  item 1\n  item 2\n\nSection B` collapses into four consecutive lines, destroying the structure of any report or diff a script prints.

src/discover/rules.rs
```diff
@@ -102,6 +102,23 @@ pub const RULES: &[RtkRule] = &[
         savings_pct: 70.0,
         ..RtkRule::DEFAULT
     },
+    RtkRule {
+        pattern: r"^bun\s+(test|run|install|build|add|remove)",
```
KuSh, line 106: Missing the terminator that every sibling rule has (`^npm\s+(exec|run|…)(\s|$)`, `^(bun\s+x|bunx)(\s|$)`, `^ls(\s|$)`), so this matches on a prefix.

I ran `rewrite_command` over a few forms and these all rewrite:

```
bun build.ts     -> Some("rtk bun build.ts")
bun test.ts      -> Some("rtk bun test.ts")
bun install.ts   -> Some("rtk bun install.ts")
bun removeall    -> Some("rtk bun removeall")
```

`bun <file.ts>` is a normal invocation since bun executes files directly, so `bun build.ts` isn't contrived. Execution stays correct (those land in `BunCommands::Other` → `run_passthrough`), but the hook rewrites commands that gain nothing and discover reports 80% savings for them.

```suggestion
        pattern: r"^bun\s+(test|run|install|build|add|remove)(\s|$)",
```

src/discover/rules.rs
```diff
@@ -102,6 +102,23 @@ pub const RULES: &[RtkRule] = &[
...
+        rtk_cmd: "rtk bun",
+        rewrite_prefixes: &["bun"],
+        category: "PackageManager",
+        savings_pct: 80.0,
```
KuSh, line 110: Separately from the anchoring: `savings_pct: 80.0` over-reports for part of this rule's own alternation. `add` and `remove` land in `BunCommands::Other` → `run_passthrough` (0% reduction), and `bun run <script>` is near-passthrough. `bun add zod` currently classifies as 80% estimated savings, so `rtk discover` will over-promise. The `subcmd_savings` list already carries the real numbers for test/install/build — the base rate is the one that's off.

README.md
```diff
@@ -219,6 +222,9 @@ rtk sbt run                     # Strip SBT preamble noise
 ### Package Managers
 ```bash
 rtk pnpm list                   # Compact dependency tree
+rtk uv run pytest               # Preserve uv env, errors only
```
KuSh, line 225: This duplicates the pre-existing line 228 (`rtk uv run pytest  # Preserve uv env, keep program output`) with a contradictory description, three lines apart. Looks like a rebase artifact from the pre-0.37.0 base — probably wants dropping.

KuSh (commented):
Another pass over the current head. Most of the earlier threads are genuinely addressed — the `bun x` alias, `bun tsc` / `bun run tsc`, the `bun x` special case in main, `c.arg("--").arg(tool)`, the `finish()` consolidation, trim-upfront, the public `test`/`install`/`build`/`run` fns, `LC_ALL=C` and the Cargo.lock revert all check out. I ran the rewrite engine over the bun command forms and all four spellings (`bun X`, `bun run X`, `bunx X`, `bun x X`) resolve to the right specialized filter, so the jest/vitest/prisma/tsc threads are correctly closed.

Two things below are blocking (the `biome` arm never runs biome, and `is_coverage_row` eats failure context on any line containing `||`). The rest are smaller.

Two more that I can't attach inline because the files aren't in the diff:

- **`src/core/tracking.rs`, `categorize_command`** — not extended with `"bun" | "bunx"`, so every `rtk bun …` / `rtk bunx …` invocation is bucketed as `"other"` instead of `"js"` in telemetry and `rtk gain`.
- **`src/core/toml_filter.rs`, `RUST_HANDLED_COMMANDS`** — not extended with `bun`/`bunx`. A project TOML filter whose `match_command` matches `bun` will compile with no "already handled in Rust" shadow warning and then silently never fire, since clap now routes `rtk bun` first.

KuSh:
Thanks for this — solid work, especially the fixture coverage.

Closing in favour of #988, which is further along in review and covers both runtimes (bun *and* deno) rather than bun alone. Keeping two overlapping bun implementations in flight isn't serving anyone, and #988 is the one about to land.

That said, this PR has things #988 doesn't: `bun test --coverage` and `--bail` handling, and a much richer fixture set. Those would be very welcome as a focused follow-up on top of #988 once it merges — please do open one if you're up for it.

HenriqueSchroeder:
Fair call, no argument from me. Two overlapping implementations is worse than one.

I'll open the follow-up once #988 lands: `--coverage` and `--bail` handling, plus the fixtures. The fixture set is real captures from bun 1.3.14 and covers the cases #988's don't: todo/skip/snapshot counts, a SIGKILLed run for exit code propagation, and a 40-failure run for the cap. Happy to hand those over earlier if @geofflittle wants them before the merge, they are on my branch and independent of the rest.

One thing from your last pass is worth rescuing, since it outlives this PR: the `&args[1..]` slicing you flagged on the eslint/biome arm exists in the `Commands::Npx` arm on develop today, so `rtk npx eslint src` misroutes to `npx -- src` and masks the failure the same way. Not mine to fix here anymore, but it is live.

KuSh:
Thanks for the reminder, @HenriqueSchroeder. I’ve opened #3812 for that. I’ll ping you here if and when #988 is merged so we can follow up. Thanks again!

Issue #1206, opened 2026-04-10
Normalize .venv/bin/ and virtualenv prefixes before registry lookup


ReinierKoops:
## Problem

RTK does not match virtualenv-prefixed tool invocations against its registry, even when a filter exists for the tool.

**Repros (rtk 0.35.0):**
```
$ rtk rewrite ".venv/bin/pytest --tb=short"
# exit 1 — no rewrite, even though rtk pytest exists

$ rtk rewrite ".venv/bin/ruff check ."
# exit 1 — no rewrite, even though rtk ruff exists
```

Note: `rtk discover -a` classifies these as "MISSED SAVINGS" with status `existing` (implying RTK can handle them), but `rtk rewrite` fails. The discover classification is based on tool-name recognition, not actual rewrite-engine matching — this discrepancy is misleading.

**Paths to normalize:**
- `.venv/bin/<tool>` — relative virtualenv (most common)
- `./.venv/bin/<tool>` — explicit relative
- `/home/user/project/.venv/bin/<tool>` — absolute path to venv bin directory
- `venv/bin/<tool>`, `virtualenv/bin/<tool>` — alternate venv directory names
- On Windows: `.venv/Scripts/<tool>.exe`

## Expected behavior

`.venv/bin/pytest --tb=short` → strip prefix, look up `pytest` in registry → `rtk pytest --tb=short`

If the tool has no RTK filter after normalization, fall through to passthrough as today.

## Impact

In this project: 550 `.venv/bin/pytest` and 376 `.venv/bin/ruff` invocations per `rtk discover -a` — ~926 total currently passing through unfiltered (~178K estimated saveable tokens).

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 88% |

### Summary

RTK fails to match virtualenv-prefixed tool invocations (e.g. .venv/bin/pytest) against the registry; strip prefix before lookup. PR #1378 is in progress.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

rooberttdev:
Confirming reproduction on rtk 0.35.0:

```
[...]

PR #1378, opened 2026-04-18
fix(discover): normalize venv/absolute paths with binary preservation (#1053)

### LINKED ISSUE #1053: path-prefixed commands pass through unfiltered; rewriting them needs a binary-override mechanism

Timber125:
## Summary

Fixes the path-prefixed command rewrite issue described in #1053 and #1206.

The naive approach (strip `.venv/bin/` prefix and emit bare `rtk pytest`) changes execution semantics: handlers call `resolved_command("pytest")` which resolves via PATH, not the original venv binary. When the venv isn't activated, this executes the wrong binary or fails.

This PR implements the **three-part fix** from #1053:

1. **Normalize for matching only** — `strip_absolute_path()` in `rewrite_segment_inner()` normalizes path-prefixed commands to find the matching registry rule, but doesn't discard the original path
2. **Preserve original binary via `RTK_BIN_PATH`** — when a path prefix is stripped, the rewritten command includes `RTK_BIN_PATH=<original_bin>` as an env var prefix (e.g. `RTK_BIN_PATH=.venv/bin/python rtk pytest tests/`)
3. **Handlers honor the override** — `override_command()` in `utils.rs` checks `RTK_BIN_PATH`; if the basename starts with "python", it uses `-m <tool>` module invocation, otherwise direct execution. All Python handlers (pytest, ruff, mypy, pip) use this before falling back to PATH resolution.

## Paths handled

| Input | Rewritten |
|-------|-----------|
| `.venv/bin/pytest tests/` | `RTK_BIN_PATH=.venv/bin/pytest rtk pytest tests/` |
| `.venv/bin/python -m pytest tests/` | `RTK_BIN_PATH=.venv/bin/python rtk pytest tests/` |
| `/usr/bin/grep -rn pattern src/` | `RTK_BIN_PATH=/usr/bin/grep rtk grep -rn pattern src/` |
| `pytest tests/` (no path) | `rtk pytest tests/` (no RTK_BIN_PATH) |

## Test plan

- [x] All 1606 existing tests pass
- [x] New test: bare commands produce no `RTK_BIN_PATH`
- [x] Updated tests: venv paths emit correct `RTK_BIN_PATH` values
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` clean

Closes #1053
Refs #1206

### DIFF
--- a/src/cmds/python/mypy_cmd.rs
+++ b/src/cmds/python/mypy_cmd.rs
@@ -1,19 +1,21 @@
 //! Filters mypy type-checking output, grouping errors by file.
 
 use crate::core::runner;
-use crate::core::utils::{resolved_command, strip_ansi, tool_exists, truncate};
+use crate::core::utils::{override_command, resolved_command, strip_ansi, tool_exists, truncate};
 use anyhow::Result;
 use regex::Regex;
 use std::collections::HashMap;
 
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
-    let mut cmd = if tool_exists("mypy") {
-        resolved_command("mypy")
-    } else {
-        let mut c = resolved_command("python3");
-        c.arg("-m").arg("mypy");
-        c
-    };
+    let mut cmd = override_command("mypy").unwrap_or_else(|| {
+        if tool_exists("mypy") {
+            resolved_command("mypy")
+        } else {
+            let mut c = resolved_command("python3");
+            c.arg("-m").arg("mypy");
+            c
+        }
+    });
 
     for arg in args {
         cmd.arg(arg);
--- a/src/cmds/python/pip_cmd.rs
+++ b/src/cmds/python/pip_cmd.rs
@@ -17,8 +17,12 @@ struct Package {
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
 
-    // Auto-detect uv vs pip
-    let use_uv = tool_exists("uv");
+    // Honor explicit binary path from rewrite engine (#1053), otherwise auto-detect
+    let has_bin_override = std::env::var("RTK_BIN_PATH")
+        .ok()
+        .filter(|s| !s.is_empty())
+        .is_some();
+    let use_uv = !has_bin_override && tool_exists("uv");
     let base_cmd = if use_uv { "uv" } else { "pip" };
 
     if verbose > 0 && use_uv {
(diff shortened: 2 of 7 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes a bug where commands invoked via absolute or virtualenv paths (e.g., `.venv/bin/python -m pytest`) were correctly classified but not rewritten to their `rtk` equivalents. The root cause was that `rewrite_segment_inner()` did not call `strip_absolute_path()` before prefix matching, unlike `classify_command()`. The fix is a 2-line insertion to normalize paths before the rewrite loop.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [x] Docs updated

---

Timber125:
⚠️ Converting to draft — this fix is naive and hits exactly the wrong-binary trap described in #1053.

**Problem**: Normalizing `.venv/bin/pytest` → `pytest` in rewrite output means `rtk pytest` calls `resolved_command("pytest")` which resolves via PATH, not the original venv binary. If the venv isn't activated, this executes the wrong binary or fails.

Reworking to implement the three-part fix from #1053:
1. Normalize in `rewrite_segment()` **for matching only**
2. Preserve the original executable path via `RTK_ORIGINAL_BIN` env var
3. Teach handlers to honor the explicit binary override

Will reopen when ready.

PR #1389, opened 2026-04-18
fix(openclaw): preserve rewrite stdout on exit 3

### LINKED ISSUE #1388: bug(openclaw): plugin discards rewrite stdout on non-zero exit
## Summary

The OpenClaw plugin currently discards any successful `rtk rewrite` output if the child process exits non-zero.

The plugin uses `execSync(...)`, so Node throws on any non-zero exit code and the plugin returns `null`. That means if `rtk rewrite` ever emits a valid rewritten command on stdout while exiting non-zero, the rewrite is silently lost.

## Current plugin logic

```ts
function tryRewrite(command: string): string | null {
  try {
    const result = execSync(`rtk rewrite ${JSON.stringify(command)}`, {
      encoding: "utf-8",
      timeout: 2000,
    }).trim();
    return result && result !== command ? result : null;
  } catch {
    return null;
  }
}
```

## Why this is a bug

The plugin assumes that a useful rewrite can only come back on exit `0`.

That is brittle. If RTK emits a rewritten command on stdout together with a non-zero status for any classification or environment-specific reason, the OpenClaw plugin will drop the rewrite entirely.

## Expected behavior

If `rtk rewrite` emits a rewritten command on stdout, the OpenClaw plugin should preserve and use that rewrite even when the subprocess exits non-zero.

## Suggested fix direction

Handle non-zero exits without discarding stdout. For example:

- inspect `error.stdout` in the `catch`, or
- switch to a subprocess API that always exposes stdout/status together

## Notes

I originally observed a case where `rtk rewrite` returned rewritten stdout with a non-zero exit, but I have not re-confirmed that exact exit behavior on every environment / version combination.

So the most precise framing is:

- this is definitely a robustness hole in the OpenClaw plugin
- whether it is a currently reproducible regression on latest RTK may be environment/version dependent

jhenderiks:
## Summary
- preserve `rtk rewrite` stdout when the child process exits with status `3`
- fix the OpenClaw plugin path where `execSync(...)` throws and the rewrite is discarded
- closes #1388

## Why this change
The current OpenClaw plugin drops rewritten stdout if `rtk rewrite` exits with status `3`.

That matters because RTKs hook-side rewrite contract uses exit `3` as the "rewritten + ask" path in `src/hooks/rewrite_cmd.rs`. The current `execSync(...)` path throws on that non-zero exit and the plugin returns `null`, so the rewrite is lost.

This PR makes the plugin preserve rewritten stdout on exit `3` instead of treating it as unrecoverable.

## Validation
- Ran: `bun --eval 'await import("./openclaw/index.ts")'`
- Ran: `cargo fmt --all --check`
- Ran: `cargo clippy --all-targets`
- Ran: `cargo test`
- Result: `1592 passed; 0 failed; 6 ignored`
- Not re-confirmed locally before opening this PR: manual verification in an environment where `rtk rewrite` emits rewritten stdout with exit `3`

## Notes
I am intentionally framing this as a correctness fix against the current exit-code contract.

The direct protocol evidence for exit `3` lives in `src/hooks/rewrite_cmd.rs`.

### DIFF
--- a/openclaw/index.ts
+++ b/openclaw/index.ts
@@ -24,6 +24,29 @@ function checkRtk(): boolean {
   return rtkAvailable;
 }
 
+function getRewriteExitStatus(error: unknown): number | null {
+  if (!error || typeof error !== "object" || !("status" in error)) {
+    return null;
+  }
+
+  const status = error.status;
+  return typeof status === "number" ? status : null;
+}
+
+function getRewriteStdout(error: unknown): string {
+  if (!error || typeof error !== "object" || !("stdout" in error)) {
+    return "";
+  }
+
+  const stdout = error.stdout;
+
+  if (typeof stdout === "string") {
+    return stdout.trim();
+  }
+
+  return Buffer.isBuffer(stdout) ? stdout.toString("utf-8").trim() : "";
+}
+
 function tryRewrite(command: string): string | null {
   try {
     const result = execSync(`rtk rewrite ${JSON.stringify(command)}`, {
@@ -31,8 +54,11 @@ function tryRewrite(command: string): string | null {
       timeout: 2000,
     }).trim();
     return result && result !== command ? result : null;
-  } catch {
-    return null;
+  } catch (error) {
+    const stdout = getRewriteStdout(error);
+    return getRewriteExitStatus(error) === 3 && stdout && stdout !== command
+      ? stdout
+      : null;
   }
 }
 


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes the OpenClaw plugin to preserve stdout from `rtk rewrite` when the child process exits with status 3. Previously, `execSync` would throw on non-zero exit codes, discarding the rewritten output. The fix extracts the exit status and stdout from the caught error object and returns the rewritten command when appropriate.

### Review Checklist
- [ ] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1388

---

KuSh:
Thanks for this. Develop already preserves the rewritten stdout when `rtk rewrite` exits 3: commit 487a2e7b (2026-06-30) added the exit 3 branch to openclaw/index.ts along with the exit 2 deny branch, and commit f525ceec (2026-06-03) had already replaced the `execSync` call with `execFileSync` so the command is no longer interpolated into a shell. The exit 3 handling this PR adds is therefore present, in a form that also covers deny and avoids the shell round trip. If you think a case is still mishandled on develop, please comment here or reopen.

Issue #1392, opened 2026-04-19
[Bug] RTK Mangles PyPi JSON Calls


avishj:
Eg.

```bash
rtk curl -s "https://pypi.org/pypi/just/json"
{
  info:
  {
    author: null,
    author_email: string,
    bugtrack_url: null,
    classifiers:
    [string] (18)
    description: string[60],
    description_content_type: date?,
    docs_url: null,
    download_url: null,
    downloads:
    {
      last_day: int,
      last_month: int,
      last_week: int
    }
    dynamic:
    [
      string
    ]
    home_page: null,
    keywords: null,
    license: string,
    license_expression: null,
    license_files:
    [
      string
    ]
    maintainer: null,
    ... +14 more keys
  }

<cut short for brevity but original output continues>
```

This breaks any commands that use something like:

```bash
rtk curl -s "https://pypi.org/pypi/just/json" | python3 -c "import sys,json; print(json.load(sys.stdin)['info']['summary'])" 2>&1
```

pszymkowiak:
Closing as duplicate of #1324. See the original issue for updates.

Issue #1313, opened 2026-04-14
Feature request: lossless-only mode + honest account of silent failures truncation causes in agent contexts


ChrisEdwards:
## Feature Request: `no_truncation = true` config flag

RTK's internal architecture already distinguishes between **lossless summarization** (whitespace stripping, ANSI removal, deduplication, reformatting) and **lossy truncation** (hard result caps that silently drop data). That distinction is encoded in `FilterResult.truncated`, `FormatMode`, and `FilterLevel` — but it's never surfaced to users.

**Request:** expose a single config flag that disables all lossy truncation while keeping lossless compression:

```toml
[safety]
no_truncation = true
```

When set, RTK would only apply operations that preserve all facts: strip ANSI codes, collapse blank lines, reformat JSON, deduplicate repeated log lines. It would never hard-cap result counts or truncate line content.

---

## Why this matters: real failure scenarios

These aren't hypothetical. Every one of these can happen silently today with default config.

### 1. Incomplete git commits (`git status` capped at 15 modified + 10 untracked)

An AI agent reviews `git status` output to decide which files to stage. RTK shows 15 modified files — but 20 were actually modified. The agent stages and commits the 15 it can see. The remaining 5 files — potentially including a config change, migration file, or security fix — are silently left out of the commit. No error. No warning. The commit goes out broken.

### 2. Missed security vulnerabilities in grep results (200 total, 25/file hard cap)

An agent is tasked with finding all usages of a deprecated/vulnerable API pattern across the codebase. `grep` returns what looks like a complete result set, but RTK has silently capped it at 200 total and 25 per file. The agent updates every match it sees, declares the job done, and ships — with the remaining instances of the vulnerable pattern still in production code.

### 3. Approved diffs with hidden changes (diff capped at 10 changes/file, hunks at 100 lines)

An agent performs a pre-commit review of a diff. RTK shows 10 changes per file. There are actually 25. The agent approves the diff having never seen the other 15 changes — which might include a logic error, a hardcoded credential, or a broken test. The `... +15 more` marker is easy to miss or ignore.

### 4. Missing breaking change notices in git log (commit body truncated to 3 lines)

An agent analyzes commit history to understand a change before cherry-picking or reverting it. The commit body contains a `BREAKING CHANGE:` notice, a required migration step, or a known caveat — but it's on line 4 or later. RTK drops it. The agent proceeds without the context it needed, causing a broken deployment or missed migration.

[...]

Issue #1282, opened 2026-04-14
Silent output corruption when stdout is piped or redirected (suggest isatty passthrough)


panwudi:
## Summary                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                          
  When `rtk` rewrites a command (e.g. `grep`, `find`, `ls`) and the output is piped into another program or redirected to a file, the compressed human-readable format silently corrupts the downstream consumer. The caller usually has no way to detect this until a downstream assertion fails.          
  
  We suggest RTK follow the same convention as `ls --color=auto`, `git diff`, `grep --color=auto`, etc.: **check `isatty(stdout)` at runtime, and fall back to raw passthrough whenever stdout is not a TTY** (pipe, redirect, or process substitution).                                                    
                                                                                                                                                                                                                                                                                                          
  ## Reproduction                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                          
  Environment: rtk 0.36.0, Claude Code PreToolUse hook (`rtk-rewrite.sh` v3), Linux.                                                                                                                                                                                                                        
  
  ```bash                                                                                                                                                                                                                                                                                                   
  # A file with exactly 83 lines matching the pattern                                                                                                                                                                                                                                                     
  $ grep -cE '^- \[ \]' TODO.md                                                                                                                                                                                                                                                                             
  83                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                            
  # Naive redirection through RTK hook rewrite                                                                                                                                                                                                                                                              
  $ grep -n '^- \[ \]' TODO.md > /tmp/pending.txt                                                                                                                                                                                                                                                         
  $ wc -l /tmp/pending.txt                                                                                                                                                                                                                                                                                  
  136 /tmp/pending.txt   # <-- WRONG, expected 83                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                            
  $ head -3 /tmp/pending.txt                                                                                                                                                                                                                                                                              
  83 matches in 36F:                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                            
  // ... 134 more lines (total: 136)...                                                                                                                                                                                                                                                                     
  ```                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                            
[...]

PR #1396, opened 2026-04-19
feat(config): support multi-word entries in exclude_commands for subcommand-level exclusion

atgim:
## Summary

Extends `[hooks].exclude_commands` matching from first-word-only equality to whitespace-aware token matching. Entries containing whitespace match the command's first N tokens (where N is the entry's token count), enabling **subcommand-level exclusion** without blocking the parent command.

```toml
[hooks]
exclude_commands = ["curl", "git diff", "kubectl describe"]
```

- `"curl"` behaves exactly as before (excludes all `curl` invocations)
- `"git diff"` now excludes `git diff` with any args, but **not** `git status` / `git log`
- Word boundaries use whitespace, so `"git diff"` never matches `git difference` or `git diffoo`
- Whitespace in both entry and command is normalized (handled by `split_whitespace` on both sides)

## Motivation

First-word-only matching forces all-or-nothing exclusion: users cannot exclude `git diff` from rewriting while keeping `git status` filtered, because both start with `git`. This is particularly harmful for **diff/patch-producing subcommands** where silent output corruption (stripped context lines, removed headers, dropped blob hashes) leads LLM-based agents to misinterpret file state and produce wrong edits.

The existing `RTK_DISABLED=1` env-var escape hatch requires the user to remember it on every invocation, which is not viable for agent-driven pipelines. A persistent, config-level knob at subcommand granularity is what was missing.

## Related issues

- #1282 — Silent output corruption when stdout is piped or redirected (directly relevant)
- #1313 — Feature request: lossless-only mode + honest account of silent failures (same concern, broader scope)
- #1392 — `rtk curl` breaks downstream parsers (motivating the silent-corruption concern)

This PR addresses a concrete slice of #1313: users who know which specific subcommands they need raw output for can now declare it in config instead of patching the binary or sprinkling `RTK_DISABLED=1`.

## Behavior

Implementation is a small helper function called from the existing exclude check:

```rust
fn is_command_excluded(cmd: &str, entry: &str) -> bool {
    let entry_tokens: Vec<&str> = entry.split_whitespace().collect();
    if entry_tokens.is_empty() {
        return false;
    }
    let cmd_tokens = cmd.split_whitespace().take(entry_tokens.len());
    cmd_tokens.eq(entry_tokens.iter().copied())
}
```

The call site in `rewrite_segment_inner` changes from:

```rust
let base = cmd_part.split_whitespace().next().unwrap_or("");
if excluded.iter().any(|e| e == base) { return None; }
```

to:

```rust
if excluded.iter().any(|e| is_command_excluded(cmd_part, e)) { return None; }
```

## Backward compatibility

- All existing single-word entries behave exactly as before (single-token `split_whitespace().eq()` reduces to first-word equality)
- No config schema changes; no migration needed
- All existing tests pass unchanged

## Test plan

- [x] `test_rewrite_multi_word_excludes_git_diff` — two-word entry blocks subcommand + args
- [x] `test_rewrite_multi_word_allows_other_git_subcommands` — other subcommands still filter
- [x] `test_rewrite_multi_word_no_partial_word_match` — `git diff` entry does not match `git diffoo` / `git difference`
- [x] `test_rewrite_multi_word_and_single_word_coexist` — mixed entries work independently
- [x] `test_is_command_excluded_whitespace_normalization` — multi-space handled correctly
- [x] `test_is_command_excluded_empty_entry_never_matches` — empty/whitespace-only entries safe
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` — all 1596 tests pass, no new clippy warnings introduced

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -595,6 +595,27 @@ const SHELL_PREFIX_BUILTINS: &[&str] = &["noglob", "command", "builtin", "exec",
 
 const MAX_PREFIX_DEPTH: usize = 10;
 
+/// Check whether a command should be excluded from rewriting based on a single
+/// `exclude_commands` entry.
+///
+/// - **Single-word entry** (no whitespace, e.g. `"curl"`): matches when the first
+///   whitespace-separated token of the command equals the entry.
+/// - **Multi-word entry** (contains whitespace, e.g. `"git diff"`): matches when
+///   the command's first N tokens equal the entry's tokens, where N is the
+///   entry's token count. This lets users exclude specific subcommands without
+///   blocking the entire parent command.
+///
+/// Word boundaries use whitespace, so `"git diff"` matches `"git diff HEAD"` and
+/// `"git diff"` exactly, but never `"git difference"` or `"git diffoo"`.
+fn is_command_excluded(cmd: &str, entry: &str) -> bool {
+    let entry_tokens: Vec<&str> = entry.split_whitespace().collect();
+    if entry_tokens.is_empty() {
+        return false;
+    }
+    let cmd_tokens = cmd.split_whitespace().take(entry_tokens.len());
+    cmd_tokens.eq(entry_tokens.iter().copied())
+}
+
 fn rewrite_segment(seg: &str, excluded: &[String]) -> Option<String> {
     rewrite_segment_inner(seg, excluded, 0)
 }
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟢 **Risk** | `low` |

### Summary

Extends the `exclude_commands` configuration to support multi-word entries, enabling subcommand-level exclusion (e.g., excluding `git diff` without blocking `git status`). The implementation adds a small helper function that performs whitespace-aware token matching against the command's first N tokens, with full backward compatibility for existing single-word entries.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1282, #1313, #1392

---

KuSh:
Hi, this is already the case. You can already exclude sub commands or even arguments using prefix string or regex starting with `^`, see [docs/guide/getting-started/configuration.md#excluding-commands-from-auto-rewrite](https://github.com/rtk-ai/rtk/blob/master/docs/guide/getting-started/configuration.md#excluding-commands-from-auto-rewrite)

Issue #1335, opened 2026-04-15
bug(hooks): exclude_commands has no effect on rtk rewrite or hook output in 0.36.0


dbl-0:
### Environment
- rtk 0.36.0 (Homebrew, /opt/homebrew/bin/rtk)
- macOS darwin 25.3.0 (arm64)
- Claude Code 2.1.109, hook `rtk-rewrite.sh` (shipped by `rtk init -g`)
- Config path: `~/Library/Application Support/rtk/config.toml` (macOS `dirs::config_dir()`)

### Repro
1. Write `~/Library/Application Support/rtk/config.toml`:
   ```toml
   [hooks]
   exclude_commands = ["^curl", "curl"]
   ```
2. `rtk config` → prints the value back, confirming config is loaded.
3. `strings $(which rtk) | grep exclude_commands` → symbol present in binary.
4. `rtk rewrite "curl http://example.com"` → prints `rtk curl http://example.com` (exit 0).

### Expected
`rtk rewrite "curl ..."` exits 1 (command excluded) per PR #342 test `test_rewrite_excludes_curl`, and the Claude Code hook passes `curl` through unchanged.

### Actual
Both the CLI `rewrite` subcommand and the installed Claude Code hook rewrite `curl` despite `exclude_commands = ["curl"]`. Only `rtk proxy curl ...` reliably bypasses filtering.

### Also tried
Project-local `.rtk/filters.toml` with `match_command = "^curl .*/health"` — verified + trusted via `rtk trust`, but only applies to unhandled commands; does not override the built-in `curl → rtk curl` rewrite.

### Likely cause
`src/hooks/rewrite_cmd.rs` reads `config.hooks.exclude_commands` and passes it to `registry::rewrite_command`. The top-level `src/rewrite_cmd.rs` (the `rtk rewrite` CLI subcommand) does not appear to thread the exclude list through the same path — two separate code paths, only one wired. Same shape as #1134 (config surface present, wiring absent).

### Related
- #243 (open — @nhumrich's 2026-03-27 comment matches this exactly)
- #219 (open — purportedly "fixed in v0.26.0" but still reported broken)
- #917 (open — adjacent tokenizer bug on env-var-prefixed commands)

### Downstream tracking
dbl-0/ha-config#1907

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

[...]

Issue #917, opened 2026-03-29
psql -h flag conflict + exclude_commands ignores env-var prefixed commands


timwagnerjr-athleticops:
## Summary

Two related issues with `rtk psql` and the hook rewrite system:

### 1. `psql -h` (host) conflicts with `rtk psql -h` (help)

When the hook rewrites `psql -h localhost -p 54322 ...` → `rtk psql -h localhost -p 54322 ...`, RTK interprets `-h` as its own `--help` flag instead of passing it through to psql as the host flag.

**Repro:**
```bash
# This prints RTK help instead of connecting to postgres:
PGPASSWORD=postgres rtk psql -h localhost -p 54322 -U postgres -d postgres -c 'SELECT 1'

# Workaround — use `--` separator:
PGPASSWORD=postgres rtk psql -- -h localhost -p 54322 -U postgres -d postgres -c 'SELECT 1'
# ✅ Works correctly
```

The `--` workaround works, but the hook-generated rewrite doesn't include it, so psql is broken out of the box when used with `-h` for host.

### 2. `exclude_commands` in config.toml doesn't match env-var-prefixed commands

Setting `exclude_commands = ["psql"]` in `config.toml` correctly excludes bare `psql` commands:
```bash
rtk rewrite "psql -h localhost"  # EXIT: 1 ✅ excluded
```

But the common pattern `PGPASSWORD=x psql ...` is still rewritten:
```bash
rtk rewrite "PGPASSWORD=postgres psql -h localhost -p 54322 -U postgres -d postgres"
# Output: PGPASSWORD=postgres rtk psql -h localhost ...  (EXIT: 0 — still rewritten)
```

It seems `rtk rewrite` parses the first token as the command name. When the command is prefixed with env vars (`PGPASSWORD=postgres psql ...`), the exclude check doesn't match `psql` because the first token is `PGPASSWORD=postgres`.

## Expected behavior

1. `rtk psql` should not consume `-h` as its own help flag — psql's `-h` means host, not help. Either use a different flag for RTK help or auto-insert `--` before psql args.
2. `exclude_commands` should match the actual command name even when preceded by inline env vars (e.g., `VAR=val psql` should still match the `psql` exclusion).

## Environment

- rtk 0.34.1
- macOS (Darwin 25.4.0, arm64)
- Claude Code hook v3
- psql (PostgreSQL) 18.0

pszymkowiak:
[...]

Issue #243, opened 2026-02-21
Support exclude_commands config for auto-rewrite hook


nwparker:
## Problem

The auto-rewrite hook (`rtk-rewrite.sh`) rewrites all supported commands unconditionally. There's no way to exclude specific commands from being rewritten without manually editing the hook script — which gets overwritten on the next `rtk init -g`.

## Proposal

Add an `exclude_commands` option to `config.toml` that the hook script reads at runtime:

```toml
[hooks]
exclude_commands = ["playwright", "curl"]
```

The hook script would check this list before rewriting and skip excluded commands.

### Alternative approaches

- **Environment variable**: `RTK_EXCLUDE="playwright,curl"` — simpler but less discoverable
- **Per-command disable in config**: `[commands.playwright] enabled = false`
- **Hook flag**: `rtk init -g --exclude playwright` that generates the hook without those patterns

Any of these would work. The key requirement is that the exclusion survives `rtk init -g` re-runs.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 91% |

### Summary

Feature request: add exclude_commands config option so specific commands survive rtk init -g re-runs without manual hook edits

<details>
<summary>📁 Relevant files</summary>

- `config.toml`
- `rtk-rewrite.sh`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

pszymkowiak:
`exclude_commands` was implemented in #342 (2026-03-05), so this issue looks resolved at a glance — but testing it end-to-end reveals a real gap that undermines the exact example this issue gives.

**`exclude_commands = ["playwright"]` does not exclude `npx playwright test`** — only a direct `playwright ...` invocation.

Root cause: `compile_exclude_patterns` (registry.rs) turns every entry into an anchored regex `^{pattern}($|\s)`, and `is_excluded` runs it against the command *before* wrapper prefixes (`npx`, `pnpm exec`, `yarn dlx`, ...) are stripped. Since `npx playwright test` doesn't start with `playwright`, the pattern never matches.

Verified live with the real config file + real Playwright:

```toml
[...]

PR #1398, opened 2026-04-19
fix(discover): exclude_commands bypass for env-prefix, sub cmd + regex

aeppling:
## Summary

- #917  part 2: `PGPASSWORD=x psql -h localhost` now correctly excluded by `["psql"]` , env prefix is stripped before matching
- #243  : `["git push"]` now excludes `git push origin main` : matches against full command, not just first token
- #1335  : `["^curl"]` works as regex , patterns are compiled as regex instead of literal `==` comparison
- Invalid regex patterns (e.g. `["curl["]`) fall back to prefix match with a stderr warning
- Non-regex patterns auto-anchor with `^` and `\b` to prevent substring false positives (`"go"` won't exclude `golangci-lint`)

## Changes

Single file: `src/discover/registry.rs`

- `ExcludePattern` enum , compiled regex or prefix fallback, precompiled once per `rewrite_command` call
- `compile_exclude_patterns()` , anchoring logic + `eprintln!` on invalid regex
- `is_excluded()` , matches against env-stripped command using precompiled patterns
- Modified `rewrite_segment_inner()` to strip env prefix before exclude check

## Test plan

- [x] `test_exclude_env_prefixed_command` , `PGPASSWORD=x psql` excluded by `["psql"]`
- [x] `test_exclude_subcommand_pattern` , `git push origin main` excluded by `["git push"]`
- [x] `test_exclude_regex_pattern` , `curl http://...` excluded by `["^curl"]`
- [x] `test_exclude_invalid_regex_fallback` , invalid regex doesn't crash, falls back to prefix
- [x] `test_exclude_does_not_substring_match` , `"go"` doesn't exclude `golangci-lint`
- [x] All 4 pre-existing exclude tests still pass
- [x] Full suite: 1597 passed, 0 failed

### DIFF
--- a/docs/guide/getting-started/configuration.md
+++ b/docs/guide/getting-started/configuration.md
@@ -91,6 +91,19 @@ Prevent specific commands from being rewritten by the hook:
 exclude_commands = ["git rebase", "git cherry-pick", "docker exec"]
 ```
 
+Patterns match against the full command after stripping env prefixes (`sudo`, `VAR=val`), so `"psql"` excludes both `psql -h localhost` and `PGPASSWORD=x psql -h localhost`.
+
+Subcommand patterns work too: `"git push"` excludes `git push origin main` but not `git status`.
+
+Patterns starting with `^` are treated as regex:
+
+```toml
+[hooks]
+exclude_commands = ["^curl", "^wget", "git rebase"]
+```
+
+Invalid regex patterns fall back to prefix matching.
+
 Or for a single invocation:
 
 ```bash
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -184,7 +184,7 @@ Example: `cargo fmt --all && cargo test` becomes `rtk cargo fmt --all && rtk car
 ### Override Controls
 
 - **`RTK_DISABLED=1`**: Per-command override (`RTK_DISABLED=1 git status` runs raw)
-- **`exclude_commands`**: In `~/.config/rtk/config.toml`, list commands to never rewrite
+- **`exclude_commands`**: In `~/.config/rtk/config.toml`, list commands to never rewrite. Matches against the full command after stripping env prefixes. Subcommand patterns work (`"git push"` excludes `git push origin main`). Patterns starting with `^` are treated as regex.
 - **Already-RTK**: `rtk git status` passes through unchanged (no `rtk rtk git`)
 
 ## Exit Code Contract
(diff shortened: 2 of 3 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Fixes the exclude_commands matching logic in the discover registry to handle env-prefixed commands (e.g., PGPASSWORD=x psql), multi-token subcommand patterns (e.g., "git push"), and regex patterns (e.g., "^curl"). Non-regex patterns are auto-anchored with ^ and \b to prevent substring false positives. Invalid regex patterns gracefully fall back to prefix matching with a warning.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #917, #243, #1335

---

pszymkowiak (approved):
LGTM. Clean fix for 3 real user-visible regressions (#917, #243, #1335). `ExcludePattern` enum with Regex/Prefix fallback is a good design, `regex::escape` + auto-anchoring `^{esc}($|\s)` correctly prevents substring false-positives (`"go"` no longer excludes `golangci-lint`). Env-prefix stripping via existing `ENV_PREFIX` regex is consistent with `classify_command`.

Verified locally:
- `cargo test discover::registry` → 232/232
- Rewrite still wraps env correctly: `FOO=bar git status` → `FOO=bar rtk git status`
- Docs updated in `configuration.md` + `hooks/README.md`

Non-blocking follow-ups:
1. `test_exclude_invalid_regex_fallback` passes for the wrong reason. `"curl["` is not `^`-prefixed, so it goes through `regex::escape` → `^curl\[($|\s)` which is a **valid** regex matching literal `curl[`. The `Prefix` fallback path is never exercised. To actually test the fallback, use an anchored invalid pattern like `"^curl["` (no escape → regex compile fails → `ExcludePattern::Prefix` branch).
2. `compile_exclude_patterns` runs on every `rewrite_command` call, so an invalid pattern in `config.toml` will `eprintln!` on every Bash command the hook intercepts. Consider log-once or compile-at-config-load for a follow-up.
3. Minor perf: regex re-compiled per call (~10-50µs for 3-5 patterns, negligible vs 10ms startup, but cache at `Config` level would be cleaner).

Issue #1250, opened 2026-04-12
Path traversal risk in install.sh tar extraction


ether-btc:
## Summary

The install.sh extraction pattern is insufficient for path traversal prevention.

## Issue

Current code uses:

```bash
find -name '..*'
```

This is a glob matching files **starting with** `..`, not a path traversal check. Additionally, the check runs **after** extraction — damage is already done by then.

## Correct approach

Use `tar -tf` to list contents **before** extraction:

```bash
# List all files without extracting
if tar -tf "$archive" | grep -qE '\.\./|\./|\.\.$'; then
  echo "Blocked: path traversal attempt in archive"
  exit 1
fi
tar -xf "$archive"
```

## Why this matters

Users who download RTK via tarball from releases are exposed to path traversal if a malicious mirror is compromised. The fix is a single pre-extraction check with no architectural changes.

## References

- CWE-22: Path Traversal
- Similar issue in curl's `--remote-name-all` behavior

pszymkowiak:
> [w] **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 82% |

### Summary

install.sh tar extraction checks for path traversal after extraction rather than before, and uses insufficient glob pattern; recommends pre-extraction check via tar -tf

> This looks like a **simple fix**. Only repo maintainers can trigger auto-fix.

<details>
<summary>📁 Relevant files</summary>

- `install.sh`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

ether-btc:
Fix submitted as PR #1686 — pre-extraction `tar -tf` check blocking `../`, `/./`, `..` patterns. CWE-22 addressed.

Issue #1159, opened 2026-04-10
enhancement: sign release binaries with GPG for integrity verification


iRonin:
## Problem

RTK releases on GitHub contain pre-built binaries (\`.tar.gz\` archives) with no cryptographic signature. Users who download these binaries have no way to verify:

1. **Authenticity**: The binary was actually built by RTK maintainers
2. **Integrity**: The binary was not modified in transit or on GitHub's servers

This is particularly important for a CLI tool that:
- Sits between the user and their shell commands
- Executes system commands on behalf of AI agents
- Has access to file system operations via its hooks

### 1. Generate a maintainer GPG key

Create a dedicated signing key for RTK releases (not a personal key).

### 2. Sign release artifacts

In the CI/CD release workflow (`.github/workflows/`), after building binaries:

\`\`\`bash
gpg --batch --yes --armor --detach-sign rtk-x86_64-unknown-linux-musl.tar.gz
gpg --batch --yes --armor --detach-sign rtk-aarch64-apple-darwin.tar.gz
# etc.
\`\`\`

### 3. Publish signatures with releases

The \`.asc\` signature files should be uploaded alongside the binaries in each GitHub Release.

### 4. Update installer

\`install.sh\` should download and verify the signature:

\`\`\`bash
# Download signature
curl -fsSL "$DOWNLOAD_URL.asc" -o "$ARCHIVE.asc"

# Verify (requires importing the public key first)
gpg --verify "$ARCHIVE.asc" "$ARCHIVE"
\`\`\`

### 5. Publish public key

Host the public signing key on the RTK website and a keyserver. Include the key fingerprint in the README.

## Alternative: Sigstore/cosign

Instead of GPG, use Sigstore's cosign for keyless signing. This is the modern approach used by many Rust projects:

\`\`\`bash
cosign sign-blob --yes rtk-x86_64-unknown-linux-musl.tar.gz
cosign verify-blob --certificate-identity=<maintainer> rtk-x86_64-unknown-linux-musl.tar.gz
\`\`\`

## Acceptance criteria

- Each release has signed artifacts or a verifiable checksum
- CI workflow automatically signs during release
- Public verification key is accessible
- Installer supports verification (or documents manual verification steps)

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
[...]

Issue #1158, opened 2026-04-10
security: install.sh uses curl|sh pattern without binary integrity verification


iRonin:
## Problem

The recommended quick install method downloads and executes a script directly from GitHub:

\`\`\`bash
curl -fsSL https://raw.githubusercontent.com/rtk-ai/rtk/refs/heads/master/install.sh | sh
\`\`\`

### Risks

1. **Supply chain attack surface**: If the GitHub repo is compromised (or a maintainer account is hijacked), every new install gets a backdoored binary. There is no integrity verification between download and execution.

2. **MITM risk**: While HTTPS protects against most MITM attacks, a compromised CA or DNS hijack could serve a malicious script. The script then downloads and installs a binary without verification.

3. **No checksum verification**: The installer downloads the tar.gz but does not verify a checksum or signature before extracting and installing.

\`install.sh\` lines 75-85:

\`\`\`bash
info "Downloading from: $DOWNLOAD_URL"
if ! curl -fsSL "$DOWNLOAD_URL" -o "$ARCHIVE"; then
    error "Failed to download binary"
fi

info "Extracting..."
tar -xzf "$ARCHIVE" -C "$TEMP_DIR"

mkdir -p "$INSTALL_DIR"
mv "${TEMP_DIR}/${BINARY_NAME}" "${INSTALL_DIR}/"
\`\`\`

### Short term: Add checksum verification

Publish SHA-256 checksums alongside each release and verify in the installer:

\`\`\`bash
# Download checksum
CHECKSUM=$(curl -fsSL "$DOWNLOAD_URL.sha256")
echo "$CHECKSUM  $ARCHIVE" | sha256sum --check
\`\`\`

### Medium term: Add GPG signature verification

Sign releases with a maintainer GPG key and verify in the installer.

### Long term: Provide OS package manager installation as primary

The Homebrew installation (\`brew install rtk\`) already has integrity verification via Homebrew's formula system. Making this the primary recommendation would be more secure.

## Comparison with industry best practices

- **Rustup**: Downloads a script but verifies GPG signatures on the binaries
- **Homebrew**: Verifies SHA-256 checksums in formula
- **Nix**: Cryptographic content-addressed store paths
- **asdf**: Verifies checksums against published hashes

## Acceptance criteria

- Installer verifies binary integrity before installation
- Checksums or signatures are published with each release
- Documentation recommends verified installation methods first

pszymkowiak:
[...]

Issue #450, opened 2026-03-10
Failed to get latest version


alfinsalsabil:
[INFO] Installing rtk...
curl: (22) The requested URL returned error: 403
[ERROR] Failed to get latest version

PR #1414, opened 2026-04-20
fix(install): resolve version via redirect to avoid GitHub API rate limits

xdm67x:
## Summary

- `install.sh` fails with `Failed to get latest version` when the GitHub REST API returns 403 (anonymous rate limit of 60 req/hour, easily hit from shared NATs, CI, or corporate networks).
- Switch the primary lookup to the `https://github.com/<repo>/releases/latest` 302 redirect, whose `Location` header contains the tag — **no API quota consumed**.
- Keep the REST API call as a fallback for defense in depth.
- Add an `RTK_VERSION` env var as an escape hatch to pin a specific version (the error message now points users to it).

## Problem

```sh
# Before
VERSION=$(curl -fsSL "https://api.github.com/repos/${REPO}/releases/latest" \
  | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
```

When unauthenticated requests exceed 60/hour (GitHub's documented rate limit for anonymous API access), the API returns HTTP 403. With `curl -f`, that produces an empty pipeline result, and the installer bails with `Failed to get latest version`. Users sharing an outbound IP (CI, NAT, coworking spaces) can trivially exhaust the quota for everyone behind that IP.

## Fix

Primary: parse the 302 redirect from `github.com/<repo>/releases/latest`. This doesn't hit the API at all and has no documented rate limit.

```sh
VERSION=$(curl -sI "https://github.com/${REPO}/releases/latest" \
    | grep -i '^location:' \
    | sed -E 's|.*/tag/([^[:space:]]+).*|\1|' \
    | tr -d '\r')
```

Fallback: if the redirect parsing yields nothing (unexpected GitHub change, network oddity), fall back to the original API call with a warning.

Escape hatch: `RTK_VERSION=v0.37.1 curl -fsSL ... | sh` lets users pin a version and skip both lookups entirely.

## Verification

- POSIX syntax check: `sh -n install.sh` ✅
- Redirect lookup tested against `rtk-ai/rtk`:
  ```
  $ curl -sI https://github.com/rtk-ai/rtk/releases/latest | grep -i '^location:'
  location: https://github.com/rtk-ai/rtk/releases/tag/v0.37.1
  ```
- Resolved version: `v0.37.1` ✅
- Fallback path preserved (still parses the API JSON the same way).
- `RTK_VERSION` override path tested by setting the env var and short-circuiting `get_latest_version`.

## Test plan

- [x] `curl -fsSL https://raw.githubusercontent.com/rtk-ai/rtk/master/install.sh | sh` installs successfully on macOS
- [x] Same on Linux (bash, `/bin/sh` = dash)
- [x] `RTK_VERSION=v0.37.0 curl -fsSL ... | sh` installs v0.37.0 specifically

## Related

- No existing open issue or PR covers this (#450 was a different problem — missing binaries on the `latest` tag, fixed in v0.28.0).
- Security-adjacent issues on the installer (#1158, #1159, #1250) are separate concerns and unaffected by this change.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -46,10 +46,25 @@ detect_arch() {
 }
 
 # Get latest release version
+# Primary: parse the 302 redirect on /releases/latest (no API call, no rate limit).
+# Fallback: the GitHub REST API (subject to 60 req/hour anonymous limit).
 get_latest_version() {
-    VERSION=$(curl -fsSL "https://api.github.com/repos/${REPO}/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
+    # Try the web redirect first — does not count against the API rate limit.
+    VERSION=$(curl -sI "https://github.com/${REPO}/releases/latest" \
+        | grep -i '^location:' \
+        | sed -E 's|.*/tag/([^[:space:]]+).*|\1|' \
+        | tr -d '\r')
+
+    # Fallback to the REST API if the redirect didn't yield a tag.
     if [ -z "$VERSION" ]; then
-        error "Failed to get latest version"
+        warn "Redirect lookup failed, falling back to GitHub API..."
+        VERSION=$(curl -fsSL "https://api.github.com/repos/${REPO}/releases/latest" \
+            | grep '"tag_name":' \
+            | sed -E 's/.*"([^"]+)".*/\1/')
+    fi
+
+    if [ -z "$VERSION" ]; then
+        error "Failed to get latest version (GitHub API may be rate-limited; set RTK_VERSION=vX.Y.Z to pin)"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


xdm67x:
I don't know why, but I can't sign for the CLA :(

image

pszymkowiak (approved):
Tested the redirect approach live — `curl -sI https://github.com/rtk-ai/rtk/releases/latest | grep location:` returns the tag cleanly, full parsing pipeline extracts `v0.37.2` correctly with no trailing chars.

The fallback to REST API and the `RTK_VERSION` escape hatch are both well-placed. Fixes a real blocker for CI environments and shared NATs hitting the 60 req/hr anonymous limit.

LGTM

PR #1422, opened 2026-04-20
fix: suppress hook warning during rtk init and verify

### LINKED ISSUE #1011: `rtk init -g` shows redundant "run `rtk init -g`" warning

JackDanger:
Fixes #1011

## Problem

Running `rtk init -g` printed a spurious warning:

```
```

This happened because `maybe_warn()` fired before the init completed.

## Fix

Skip `maybe_warn()` for `Init` and `Verify` commands — they manage the hook themselves and don't need the reminder.

```rust
if !matches!(
    cli.command,
    Commands::Gain { .. } | Commands::Init { .. } | Commands::Verify { .. }
) {
    hooks::hook_check::maybe_warn();
}
```

## Test plan

- [ ] `rtk init -g` no longer shows the hook warning
- [ ] `rtk verify` no longer shows the hook warning  
- [ ] Other commands (e.g. `rtk git status`) still show the warning when hook is missing
- [ ] `cargo test --all` passes (1606 tests)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1331,8 +1331,11 @@ fn run_cli() -> Result<i32> {
     };
 
     // Warn if installed hook is outdated/missing (1/day, non-blocking).
-    // Skip for Gain — it shows its own inline hook warning.
-    if !matches!(cli.command, Commands::Gain { .. }) {
+    // Skip for Gain (shows its own inline warning), Init/Verify (manage the hook themselves).
+    if !matches!(
+        cli.command,
+        Commands::Gain { .. } | Commands::Init { .. } | Commands::Verify { .. }
+    ) {
         hooks::hook_check::maybe_warn();
     }
 


KuSh (approved):
## Round 1 of 3 — APPROVED
**Claim**: `rtk init` and `rtk verify` should not print the daily "No hook installed" reminder.
**Scope**: accept (round 1, frozen).
**Ran**: merge-base `da3c4322` vs head vs head-merged-into-`develop`, isolated `HOME` per case, `LC_ALL=C`; 12 meta commands probed for the reminder and for the daily-marker side effect.
**CI** @`06d53872`: all 11 green (fmt, clippy, test on ubuntu/macos/windows, benchmark, semgrep, Security Scan, doc review, test presence, CLA).

### Blocks merge (0)

The fix does what it claims, and the reason it is right is slightly better than the description says. `maybe_warn()` does not only print — it *touches the once-a-day marker*. So on `develop` a meta command spends the reminder on itself and leaves the next filtered command silent:

| | `develop` | this PR |
|---|---|---|
| `rtk init -g` | warns (tells you to run what you are running) | quiet |
| `rtk verify` | warns **and** prints its own `SKIP RTK hook not installed` — twice | own line only |
| `rtk verify --filter X` | warns, burns the marker | quiet, marker intact |
| `rtk ls`, `rtk git status` | warns | warns |

Measured: on `develop`, `rtk verify --filter X` followed by `rtk ls` gives **0** warnings — the reminder was spent on the meta command. On this branch the same sequence gives **1**, on the command that should have been hooked.

- *"`verify --filter X` loses the hook signal, so the guard should be `Verify { filter: None, .. }`."* Raised independently by a deep review pass and by me, both from reading the code. Execution drops it: narrowing that way would restore the marker burn shown above, which is the worse outcome. Not filed.

`is_operational_command` allowlist unaffected; both variants are struct variants so `{ .. }` compiles; `maybe_warn()` has no side effect beyond the stderr line and the marker; merges cleanly into current `develop`, where the full suite is green.

### Added by fixup — `06d53872`
`tests/hook_warning_scope_test.rs`: pins both directions (exempt commands stay quiet, filtered commands still warn) and pins the marker behaviour above, which is the part worth protecting. Reverting this PR's change fails two of the three tests. Unix-gated, because `hook_check::status()` resolves the Claude directory through the home directory and only Unix takes that from `$HOME` — Windows CI is green on that basis.

### Filed as follow-ups
Found while probing, all pre-existing and outside this PR's hunks — issues will be opened after merge:
- `rtk gain` reports a missing hook only on the default view **with** tracking data. `--daily`, `--weekly`, `--monthly`, `--all` and the empty report stay silent — and the empty report is where the advice matters most, since a missing hook is the likeliest reason there is nothing to show. A fix is ready.
- Six more meta commands (`config`, `trust`, `hook-audit`, `discover`, `session`, `cc-economics`) still burn the daily marker — the same class this PR fixes two members of.
- `dotnet_trx::tests::test_parse_trx_files_in_dir_since_ignores_older_files` is flaky (1/5 failures here, 2/6 on the merge-base); it depends on file mtimes.

### Heads-up for whoever merges next
**#3582** and **#3614** both edit this same `matches!` expression — each adds `Commands::Hook`, and #3614 also `Rewrite`. They duplicate each other; whichever follows this one needs a one-line rebase.

### Next
Nothing — approving. Thanks for the patience on the five-month wait; this was correct as submitted, and CI had never actually built it until now.

PR #1443, opened 2026-04-22
fix(go): preserve benchmark and fuzz output in go test filter

matheus-meneses:
## Summary

- Fix go test filter returning "No tests found" for benchmark-only runs — Go does not emit individual `pass` events with `Test` field for benchmarks, so the filter saw zero passes
- Preserve fuzz test final stats (execs/sec, new interesting) instead of silently discarding them on success — only the last progress line is kept, intermediate lines are filtered as noise
- Surface non-JSON error lines instead of silently dropping them when the filter would return "No tests found"

## Test plan

- [x] Manual testing: `rtk go test -json -bench=. -benchmem -run='^$' ./...` shows benchmark data
- [x] Manual testing: `rtk go test -json -fuzz=Fuzz -fuzztime=10s ./pkg/` shows final fuzz stats only


matheus-meneses:
Closing: will re-submit from a branch based on develop.

PR #1444, opened 2026-04-22
fix(go): preserve benchmark and fuzz output in go test filter

### LINKED ISSUE #1445: bug(go): go test filter returns 'No tests found' for benchmarks and silently drops fuzz output
## Summary

The `go test` filter has three related issues when handling non-unit-test output types (benchmarks, fuzz, build errors).

### 1. Benchmark-only runs show "No tests found"

When running benchmarks with `-run='^$'` (skip unit tests), the filter returns "No tests found" instead of showing benchmark results.

**Root cause:** Go does not emit individual `pass` events with a `Test` field for benchmarks — only package-level passes. The filter only counts passes when `event.test.is_some()`, making benchmarks invisible.

```bash
rtk go test -json -bench=. -benchmem -run='^$' ./...
# Actual:   "No tests found"
# Expected: BenchmarkFoo-10  5000  200000 ns/op  1024 B/op  8 allocs/op
```

### 2. Fuzz test progress silently discarded on success

Active fuzzing emits multiple `fuzz: elapsed:` progress lines with execution stats (total execs, exec/s, new interesting inputs). The filter discards all of them on success, showing only "N passed" with no fuzz data.

```bash
rtk go test -json -fuzz=FuzzFoo -fuzztime=10s ./pkg/
# Actual:   "1 passed"
# Expected: "1 passed, 1 fuzz" + final fuzz stats line
```

### 3. Non-JSON error lines silently dropped

When `go test -json` emits non-JSON lines (e.g., build errors before JSON streaming starts), the filter does `Err(_) => continue`, which can mask build failures as "No tests found".

## Expected behavior

- Benchmark result lines (ns/op, B/op, allocs/op) should be detected and displayed
- Fuzz final progress line should be preserved (only the last one, intermediate lines filtered)
- Non-JSON lines should be surfaced when the filter would otherwise return "No tests found"

## Environment

- rtk version: 0.37.2
- Go version: 1.26.1
- OS: macOS (darwin)

matheus-meneses:
## Summary

Fixes #1445

- Fix go test filter returning "No tests found" for benchmark-only runs — Go does not emit individual `pass` events with `Test` field for benchmarks, so the filter saw zero passes
- Preserve fuzz test final stats (execs/sec, new interesting) instead of silently discarding them on success — only the last progress line is kept, intermediate lines are filtered
- Surface non-JSON error lines instead of silently dropping them when the filter would return "No tests found"

## Test plan

- [x] Manual testing: `rtk go test -json -bench=. -benchmem -run='^$' ./...` shows benchmark data
- [x] Manual testing: `rtk go test -json -fuzz=Fuzz -fuzztime=10s ./pkg/` shows final fuzz stats only


matheus-meneses:
The code is ready for review. Thanks for the attention.

matheus-meneses:
Anything, Guys?

matheus-meneses:
Anything, Guys?

PR #1446, opened 2026-04-22
fix: align rewrite regression tests and hook audit behavior

josh87523:
## Summary
- add Claude hook audit logging and skip already-RTK/heredoc commands early
- update hook regression expectations to match current rewrite surface
- make smoke assertions tolerate rewrite exit code 3 and validate current `rtk json --schema` behavior

## Verification
- `bash hooks/claude/test-rtk-rewrite.sh`
- `bash scripts/test-all.sh`

### DIFF
--- a/hooks/claude/rtk-rewrite.sh
+++ b/hooks/claude/rtk-rewrite.sh
@@ -43,6 +43,31 @@ if [ -z "$CMD" ]; then
   exit 0
 fi
 
+audit_log() {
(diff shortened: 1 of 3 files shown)


hooks/claude/rtk-rewrite.sh
KuSh, line 1: That would be cool for rtk bin to be "overridable". Like to test the current debug build of the repository without having to add it to the current PATH:

RTK=./target/debug/rtk hooks/claude/rtk-rewrite.sh

Otherwise LGTM!

PR #1454, opened 2026-04-22
fix(ls): fall back to raw output when ls binary is eza/exa/lsd (#1321)

### LINKED ISSUE #1321: Bug: `rtk ls` returns empty output when shell alias `ls` is mapped to `eza`
### Description
When using zsh with `alias ls='eza --icons --group-directories-first'`, running `rtk ls` returns empty output, while `rtk proxy ls` works correctly.
### Environment
- OS: Linux
- Shell: zsh
- RTK version: 0.36.0
- eza version: (run `eza --version` to get)
- Alias: `ls: aliased to eza --icons --group-directories-first`
### Reproduction
```bash
# 1. Set up eza alias in zsh
alias ls='eza --icons --group-directories-first'
# 2. Run rtk ls - returns empty
rtk ls .
# Output: (empty)
# 3. Run rtk proxy ls - works correctly  
rtk proxy ls .
# Output: shows files correctly
# 4. Direct eza works
eza --icons --group-directories-first .
# Output: shows files with icons
Expected Behavior
rtk ls should handle cases where shell ls is aliased to eza, or at least:
- Fallback to raw output when parsing fails
- Provide clear error/warning instead of silently returning empty
Root Cause Analysis
1. RTK's ls filter parses output from underlying ls command
2. When ls is aliased to eza, output format differs (icons, colors, different column structure)
3. Parser fails to parse eza's output, returns empty silently
4. rtk proxy ls bypasses parser, so works correctly
Related Issues
- #803: rtk ls -1 returns empty output (similar parser bug)
- PR #962: fix(ls): strip -1 flag (in progress)
Proposed Solutions
1. Detect shell alias: Before parsing, check if ls is aliased to non-standard tool (eza, exa, lsd)
2. Fallback gracefully: If parser returns empty on non-empty directory, fall back to raw output
3. Add warning: Emit warning when detecting incompatible ls alias
4. Force native ls: Use /usr/bin/ls or command ls internally instead of shell alias
Impact
This breaks RTK's hook integration when users have modern ls replacements (eza/exa/lsd) aliased. AI agents silently get empty directory listings, causing incorrect reasoning about project structure.

pszymkowiak:
## Summary

- `rtk ls` returned `(empty)` for non-empty directories when the `ls` binary on PATH was eza, exa, or lsd
- Root cause: eza uses day-first dates (`22 Apr 14:30`) vs POSIX month-first (`Apr 22 14:30`) — the regex anchor never matched, every line was silently dropped
- Fix: track `any_content_line` + `parse_hit` in `compact_ls`; return a sentinel when output was present but nothing parsed, triggering a raw passthrough with a warning

## Behaviour change

| Scenario | Before | After |
|---|---|---|
| eza as `ls` on PATH | `(empty)` ❌ | eza raw output + warning ✅ |
| Standard `/bin/ls` | filtered output ✅ | unchanged ✅ |
| Truly empty directory | `(empty)` ✅ | unchanged ✅ |

No token savings on eza output (passthrough). A follow-up PR could add an eza-aware parser.

## Test plan

- [x] 19 existing ls unit tests pass (`cargo +nightly test cmds::system::ls`)
- [x] New regression test `test_compact_eza_format_returns_sentinel` added
- [x] Manually tested with `PATH=/tmp/eza-test-path:$PATH rtk ls src/` (eza symlinked as `ls`)
- [x] Truly empty directory still returns `(empty)`
- [x] `cargo +nightly fmt` and `cargo +nightly clippy` clean

Fixes #1321


ashokgelal:
I can reproduce this as well.

In my shell, `ls` is provided by `lsd` (`ls='lsd -l'`), and `rtk ls` returns `(empty)` in non-empty directories for some flag combinations.

Examples that fail for me:
- `rtk ls -1A .`
- `rtk ls -1 .`
- `rtk ls -ltT .`

Examples that work:
- `rtk ls -A .`
- `rtk ls -lt .`

This looks consistent with the bug fixed in this PR.

ashokgelal:
Additional data point from my side:

Even after removing `lsd` from PATH and confirming `ls` resolves to `/bin/ls`, I still see this:

```sh
which ls
# /bin/ls

rtk ls -1A
# (empty)

rtk /bin/ls -1A
# <lists files correctly>

aeppling:
Thanks for the contribution! This bug is already fixed on `develop` since commit `0d70760ef987` (2026-04-16), which introduced the `parsed_count == 0 && has_real_content` raw fallback as part of fixing #1276.

Tested on current `develop` with both **eza** (v0.23.4) and **lsd** (v1.2.0) symlinked as `ls` — no combination returns `(empty)` for non-empty directories, and truly empty directories still return `(empty)` correctly.

Issue #1452, opened 2026-04-22
bug(grep): `-c` count flag ignored — returns match summary + context instead of line count


sajupingdev-collab:
## Description

`rtk grep -c` does not preserve `grep -c` semantics. Plain `grep -c` prints only a single integer (the number of matching lines); `rtk grep -c` instead prints the full match summary (`N matches in MF:`) and the per-line context, the same as without `-c`.

This breaks scripts that rely on `grep -c` returning a parseable integer, e.g. `count=$(grep -c pattern file)`.

## Reproduction

```bash
printf 'foo\nbar\nfoo\nbaz\n' > /tmp/rtktest.txt

/usr/bin/grep -c foo /tmp/rtktest.txt
# → 2

rtk grep -c foo /tmp/rtktest.txt
# → 2 matches in 1F:
#
#   [file] rtktest.txt (2):
#        1: foo
#        3: foo
```

## Expected

`rtk grep -c <pattern> <file>` should return the bare integer `2`, matching plain `grep -c`.

## Actual

The `-c` flag is effectively ignored: output is the normal rtk-compressed match listing, not the count. Any `$(rtk grep -c ...)` capture yields a multi-line blob rather than an integer.

## Environment

- rtk version: 0.36.0
- OS: macOS 15.x (Darwin 25.2.0, Apple Silicon)
- Shell: zsh

## Relation to other issues

- #1436 (`The rtk grep is not grep-compatible`): same class — rtk grep diverges from GNU grep semantics for specific flags. This report is for `-c` specifically; #1436 is for `-n` + alternation.
- #838 (`stdin pipe silently ignored`): another grep-compat gap, but for stdin rather than flag semantics.

Summary: `-c` is a widely-used flag in agent scripts and CI; silent divergence produces plausibly-wrong data without warning.

pszymkowiak:
> [w] **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 92% |

### Summary

rtk grep -c flag ignores grep-compatible count semantics, outputting full match listing instead of bare integer; fix the -c flag handler in the grep filter to emit only the line count.

> This looks like a **simple fix**. Only repo maintainers can trigger auto-fix.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

PR #1456, opened 2026-04-22
fix(grep): adjust the command to fall through if the output would already be as small as possible

JBF1991:
## Summary
Fix `rtk grep` compatibility with standard grep format flags (`-c`, `-l`, `-L`, `-o`, `-Z`). These flags change output shape so fundamentally that RTK's grouped formatter produced garbage or ignored the flag entirely.

### Root Cause
- Issue: #1452 
- `rtk grep` hardcoded `-n --no-heading` into `rg` regardless of user flags.
- The output parser assumed `file:line:content` format, so `-c` output (`file:count`) was misinterpreted.
- Clap's `context_only: bool` had `#[arg(short, long)]`, which captured `-c` before it reached `extra_args`.

## Test plan

**All tests pass:**
```bash
$ rtk cargo test --all
cargo test: 1680 passed, 6 ignored
```

**Format flags now produce raw output:**

| Flag | Command | Output |
|------|---------|--------|
| `-c` | `rtk grep -c "fn run" src/cmds/system/grep_cmd.rs` | `1` |
| `-l` | `rtk grep -l "fn run" src/cmds/system/grep_cmd.rs` | `src/cmds/system/grep_cmd.rs` |
| `-L` | `rtk grep -L "nonexistent" src/cmds/system/grep_cmd.rs` | `src/cmds/system/grep_cmd.rs` |
| `-o` | `rtk grep -o "fn" src/cmds/system/grep_cmd.rs` | `fn\nfn\nfn...` |

**Normal path unchanged:**
```bash
$ rtk grep "fn run" src/cmds/system/grep_cmd.rs
1 matches in 1F:

[file] src/cmds/system/grep_cmd.rs (1):
    12: pub fn run(
```

### DIFF
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -215,11 +215,11 @@ rtk grep <pattern> [chemin] [options]
 |--------|-------|--------|-------------|
 | `--max-len` | `-l` | 80 | Longueur maximale de ligne |
 | `--max` | `-m` | 50 | Nombre maximum de resultats |
-| `--context-only` | `-c` | non | Afficher uniquement le contexte du match |
+| `--context-only` |  | non | Afficher uniquement le contexte du match (pas de raccourci, `-c` est reserve a `grep --count`) |
 | `--file-type` | `-t` | tous | Filtrer par type (ts, py, rust, etc.) |
 | `--line-numbers` | `-n` | oui | Numeros de ligne (toujours actif) |
 
-Les arguments supplementaires sont transmis a `rg` (ripgrep).
+Les arguments supplementaires sont transmis a `rg` (ripgrep). Les flags qui changent le format de sortie (`-c`, `-l`, `-L`, `-o`, `-Z`) passent directement a `rg`/`grep` sans filtrage RTK.
 
 **Economies :** ~80%
 
--- a/src/cmds/system/README.md
+++ b/src/cmds/system/README.md
@@ -5,7 +5,7 @@
 ## Specifics
 
 - `read.rs` uses `core/filter` for language-aware code stripping (FilterLevel: none/minimal/aggressive)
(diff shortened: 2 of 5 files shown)


aeppling (commented):
## Blockers

### 1. Missing `--no-ignore-vcs` in passthrough path

The duplicated rg command construction forgot `--no-ignore-vcs` (added by #791 to fix false negatives on gitignored files).

Result: `rtk grep "pattern" .` searches gitignored files, `rtk grep -c "pattern" .` does not. Same pattern, different results depending on output flag.

### 2. `-l` / `--files-with-matches` is dead code

Clap captures `-l` as `max_len` (`#[arg(short = 'l', ...)]`) before it reaches `extra_args`. Users get a parse error, not file listing. Same class of bug this PR fixed for `-c`.

### Root cause: both bugs come from duplicating the command construction

The passthrough path copies ~25 lines of rg command setup instead of reusing it. Fix both bugs at once by moving the `has_format_flag` check to *after* execution, *before* formatting:

```rust
// existing rg command construction (unchanged, includes --no-ignore-vcs)
let result = exec_capture(&mut rg_cmd).or_else(|_| { /* grep fallback */ })?;

if has_format_flag(extra_args) {
    print!("{}", result.stdout);
    if !result.stderr.is_empty() {
        eprintln!("{}", result.stderr.trim());
    }
    timer.track_passthrough(...);
    return Ok(result.exit_code);
}

// existing formatter (unchanged)
```

No duplication, no flag drift, `--no-ignore-vcs` inherited automatically.

JBF1991:
I wasn't aware of #791 when I pushed this up. Sorry! I will update it accordingly.

JBF1991:
@aeppling I went ahead and made the change with slight tweaks to avoid empty spaces in the db when there was a pass-through.

aeppling (approved):

aeppling:
LGTM, thanks for contributing @JBF1991 

Will be released during the week

PR #840, opened 2026-03-26
fix(npx): route unknown tools through npx instead of npm

gghez:
## Summary

- `rtk npx <unknown-tool>` was dispatching to `npm` instead of `npx` because the
  catch-all in the Npx match arm delegated to `npm_cmd::run()`, which hardcodes
  `resolved_command("npm")`
- Replaced the catch-all with a direct `npx` invocation that still applies npm
  boilerplate filtering via `filter_npm_output`

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [x] 3 new tests covering npx argument parsing
- [x] Manual testing: `rtk npx cowsay hello` output inspected

Fixes #815

PR #819, opened 2026-03-25
fix(npx): use npx instead of npm for generic passthrough

soof-golan:
## Summary

- `rtk npx` fallback for unrecognized commands called `npm_cmd::run()`, which executes `npm` instead of `npx`
- Replaced with a direct `npx` passthrough, matching the existing prisma passthrough pattern

Fixes #815

PR #1458, opened 2026-04-22
fix(npx): dispatch unknown tools to npx instead of npm

### LINKED ISSUE #815: `rtk npx` executes `npm` instead of `npx`
## Bug

`rtk npx` dispatches to `npm` instead of `npx`, breaking all `rtk npx` invocations.

## Reproduction

```bash
# npx works
$ npx cowsay hello
 _______
< hello >
 -------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

# rtk npx doesn't — verbose confirms it runs npm instead of npx
$ rtk npx -v cowsay hello
Running: npm cowsay hello
npm error Missing script: "cowsay"
npm error
npm error To see a list of scripts, run:
npm error   npm run
npm error A complete log of this run can be found in: ~/.npm/_logs/2026-03-25T11_53_23_189Z-debug-0.log
```

## Expected behavior

`rtk npx <cmd> [args]` should execute `npx <cmd> [args]`, applying only output filtering.

## Actual behavior

`rtk npx` executes `npm` instead of `npx`. Since `npm` is a different binary with different argument parsing, subcommands and their flags are misinterpreted.

## Context

- rtk: 0.31.0
- npm/npx: 11.3.0
- Node: v24.0.0
- macOS (Apple Silicon)

The rewrite hook converts `npx <cmd> [args]` → `rtk npx <cmd> [args]`, which then fails. `rtk proxy npx <cmd> [args]` works correctly as a workaround.

## Related

- #713 — scoped package parsing in `rtk npx` (different root cause, same subsystem)
- #739 — npx fallback hang on first-run approval (different symptom)
- #197 — passthrough for tools invoked via npx/bunx/pnpx

tmchow:
## Summary

- Fixes `rtk npx` dispatching to `npm` instead of `npx` for unrouted tools. The fallback arm of `Commands::Npx` called `npm_cmd::run`, so `rtk npx cowsay hello` tried to run `npm cowsay hello` and failed with "Missing script".
- Replaces the fallback with a plain `npx` passthrough modeled on the existing prisma generic-subcommand arm a few lines up in the same file.
- Preserves `--skip-env` (sets `SKIP_ENV_VALIDATION=1` on the spawned process, matching what `npm_cmd::run` did).

## Evidence

Demo: rtk npx cowsay hello before vs after the fix

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` pass locally (1679 tests, 0 failed). Pre-existing clippy warnings are unrelated to this change.
- [x] Manual: installed `rtk 0.37.2` still reproduces with `npm error code ENOENT`. Rebuilt `target/debug/rtk` from this branch routes through npx and prints the cowsay output (see demo above).
- [x] Added `test_npx_unknown_tool_passthrough` next to the existing clap parsing tests to pin the parse-level contract.

## Notes

- Targets `develop` per CONTRIBUTING.md.
- No `CHANGELOG.md` edit since release-please generates entries from commit messages.
- Aware of #819 and #840 which target the same bug. Happy to close this one if a maintainer prefers either of those. This version mirrors the in-file prisma passthrough pattern, is rebased on current `develop` (post #826 layout), and preserves `--skip-env`.

Fixes #815

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/src/core/runner.rs
+++ b/src/core/runner.rs
@@ -169,28 +169,16 @@ where
 }
 
 pub fn run_passthrough(tool: &str, args: &[std::ffi::OsString], verbose: u8) -> Result<i32> {
+    if verbose > 0 {
+        eprintln!("{} passthrough: {:?}", tool, args);
+    }
     let mut cmd = crate::core::utils::resolved_command(tool);
     cmd.args(args);
     let args_str = tracking::args_display(args);
-    run_passthrough_cmd(cmd, tool, &args_str, verbose)
-}
-
-/// Passthrough variant for callers that need to build the Command themselves
-/// (e.g., to set env vars like `SKIP_ENV_VALIDATION`). Handles the verbose log
-/// line and delegates to `run` with `RunMode::Passthrough`.
-pub fn run_passthrough_cmd(
-    cmd: Command,
-    tool_name: &str,
-    args_display: &str,
-    verbose: u8,
-) -> Result<i32> {
-    if verbose > 0 {
-        eprintln!("{} passthrough: {}", tool_name, args_display);
-    }
     run(
         cmd,
-        tool_name,
-        args_display,
+        tool,
+        &args_str,
         RunMode::Passthrough,
         RunOptions::default(),
     )
--- a/src/core/runner.rs
+++ b/src/core/runner.rs
@@ -169,16 +169,28 @@ where
 }
 
 pub fn run_passthrough(tool: &str, args: &[std::ffi::OsString], verbose: u8) -> Result<i32> {
-    if verbose > 0 {
-        eprintln!("{} passthrough: {:?}", tool, args);
-    }
     let mut cmd = crate::core::utils::resolved_command(tool);
     cmd.args(args);
     let args_str = tracking::args_display(args);
+    run_passthrough_cmd(cmd, tool, &args_str, verbose)
+}
+
+/// Passthrough variant for callers that need to build the Command themselves
+/// (e.g., to set env vars like `SKIP_ENV_VALIDATION`). Handles the verbose log
+/// line and delegates to `run` with `RunMode::Passthrough`.
+pub fn run_passthrough_cmd(
+    cmd: Command,
+    tool_name: &str,
+    args_display: &str,
+    verbose: u8,
+) -> Result<i32> {
+    if verbose > 0 {
+        eprintln!("{} passthrough: {}", tool_name, args_display);
+    }
     run(
         cmd,
-        tool,
-        &args_str,
+        tool_name,
+        args_display,
         RunMode::Passthrough,
         RunOptions::default(),
     )
--- a/src/main.rs
+++ b/src/main.rs
@@ -2057,20 +2057,7 @@ fn run_cli() -> Result<i32> {
                 "next" => next_cmd::run(&args[1..], cli.verbose)?,
                 "prettier" => prettier_cmd::run(&args[1..], cli.verbose)?,
                 "playwright" => playwright_cmd::run(&args[1..], cli.verbose)?,
-                _ => {
-                    // Generic npx passthrough: unknown tools run through npx, not npm.
-                    // Build the Command here so we can set SKIP_ENV_VALIDATION, then hand
-                    // it to the shared runner so tracking + streaming match pnpm_cmd.
-                    let mut cmd = core::utils::resolved_command("npx");
-                    for arg in &args {
-                        cmd.arg(arg);
-                    }
-                    if cli.skip_env {
-                        cmd.env("SKIP_ENV_VALIDATION", "1");
-                    }
-                    let args_str = args.join(" ");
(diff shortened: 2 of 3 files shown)


tmchow:
Note: the failing benchmark test is pre-existing

src/main.rs
```diff
@@ -2058,8 +2058,22 @@ fn run_cli() -> Result<i32> {
                 "prettier" => prettier_cmd::run(&args[1..], cli.verbose)?,
                 "playwright" => playwright_cmd::run(&args[1..], cli.verbose)?,
                 _ => {
-                    // Generic passthrough with npm boilerplate filter
-                    npm_cmd::run(&args, cli.verbose, cli.skip_env)?
```
KuSh, line 2062: Instead of duplicating code, loosing some parts (verbose at least) and adding maintenance burden, could't we have a `npm_cmd::exec` function with npm_run::run splitted between run specifics (the part computing effective_args) and a reusable part (the rest) that would be used by npm_cmd::exec too ?

tmchow: Pushed 09aaf55: added `runner::run_passthrough_cmd(cmd, tool, args_display, verbose)` that takes a pre-built Command (so callers can set env vars like `SKIP_ENV_VALIDATION`), and refactored the existing `run_passthrough` to delegate to it. The npx arm now builds the Command, sets the env when needed, and hands it to the shared helper.

As a bonus this also upgrades the npx passthrough from `cmd.status()` to the shared streaming + tracking path that `pnpm_cmd::run_passthrough` already uses, so verbose logging and telemetry match across tools. Tests: 1679 pass, including `test_npx_unknown_tool_passthrough`.

tmchow:
@KuSh all fixed up

src/main.rs
```diff
@@ -2058,8 +2058,18 @@ fn run_cli() -> Result<i32> {
...
+                    // it to the shared runner so tracking + streaming match pnpm_cmd.
+                    let mut cmd = core::utils::resolved_command("npx");
+                    for arg in &args {
+                        cmd.arg(arg);
+                    }
+                    if cli.skip_env {
+                        cmd.env("SKIP_ENV_VALIDATION", "1");
+                    }
+                    let args_str = args.join(" ");
+                    core::runner::run_passthrough_cmd(cmd, "npx", &args_str, cli.verbose)?
```
KuSh: Thanks for the change but that's still not what I had in mind.

I'd prefer to have everything in src/cmds/js/npm_cmd.rs within an exec function. And here just have a `npm_cmd::exec(&args, cli.verbose, cli.skip_env)?` call.

Constructing args, handling skip_env can be mutualized. The only thing that change compared to `run` is the effective_args building and the `resolved_command` argument

So what I had in mind was:
- create a `fn run_filtered` that calls `resolved_command` with command name from params, add args from params to cmd, handle skip_env and verbose then calls `runner::run_filtered`
- update `pub fn run` to construct its effective_args and calls npm_cmd::run_filtered with 'npm' command name
- create a new `pub fn exec` that calls npm_cmd::run_filtered with 'npx' command name
- main calls npm_cmd::exec instead of npm_cmd::run for npx passthrough

KuSh (changes requested):

tmchow:
Pushed a5e6bf5: moved everything into npm_cmd. Added a private `run_filtered(name, args, verbose, skip_env)` helper that builds the command, applies `SKIP_ENV_VALIDATION`, emits the verbose log, and routes through `runner::run_filtered` with `filter_npm_output`. `pub fn run` keeps its existing `effective_args` building (run-injection logic) and calls `run_filtered("npm", ...)`. New `pub fn exec` just calls `run_filtered("npx", args, verbose, skip_env)`. The fallback in main.rs is now `npm_cmd::exec(&args, cli.verbose, cli.skip_env)?`.

One call-out: I kept `exec` routing through `runner::run_filtered` (matching npm) instead of `runner::run_passthrough_cmd`, so both share streaming + tracking and the same output filter. If you want raw passthrough on npx, swap the runner call in `run_filtered` and pass a passthrough flag, happy to do that.

Verified: `cargo test` 1679 passing (no change), `cargo clippy` clean for the changed files. Manual: rebuilt `target/release/rtk`, `rtk npx cowsay "ok"` still works, `-v` shows `Running: npx cowsay ok`.

src/core/runner.rs
KuSh, line 1: This part could be revert now that you don't use `run_passthrough_cmd` anymorre

KuSh (changes requested):
Just one unneeded change left, otherwise LGTM!

src/core/runner.rs
tmchow: Reverted in d8bfca8: `src/core/runner.rs` is back to its develop state. `cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test` all pass (1679 tests, 0 failed).

KuSh (approved):

PR #1472, opened 2026-04-23
fix(discover): encode '.' as '-' in project path slug

### LINKED ISSUE #1457: rtk discover: default project filter fails when cwd contains a dot (e.g. /Users/first.last)
## Summary

`rtk discover` (no args) returns `Scanned: 0 sessions` when run from a
directory whose path contains a `.`, even though session files exist
for that project. `rtk discover -a` and `rtk discover -p <hyphenated-substring>`
both work correctly.

## Environment

- rtk: **0.37.1** (Homebrew, `/opt/homebrew/bin/rtk`)
- OS: macOS (Darwin 25.2.0), zsh
- Claude Code with `rtk hook claude` wired as `PreToolUse > Bash`
- Username contains a dot: `/Users/triok.t`

## Repro

1. cwd: `/Users/triok.t` (path contains `.`)
2. Claude Code stores sessions at `~/.claude/projects/-Users-triok-t/` —
   Claude's slug escaper converts **both** `/` and `.` into `-`.
3. Run `rtk discover` with no args.

**Actual**:
```
```

**Expected**: same ~600+ sessions that `rtk discover -p triok-t` finds.

## Evidence of the root cause

Running explicit substring filters:

| Filter | Result |
|---|---|
| `-p triok-t`          | 614 sessions ✅ |
| `-p "Users-triok.t"`  | 0 sessions ❌ |
| `-p "/Users/triok.t"` | 0 sessions ❌ |

Only the hyphenated form matches — i.e. the actual directory name
uses `-` for both `/` and `.`, but rtk's implicit current-project
filter seems to preserve the dot.

## Hypothesis

rtk's path-to-slug conversion for the implicit "current project"
filter replaces `/` → `-` but not `.` → `-`. For cwd `/Users/triok.t`
it computes `-Users-triok.t`, which doesn't match Claude's actual
`-Users-triok-t` directory. Since `-p` is a substring match, no
sessions match.

## Workaround

```bash
ln -s -- /Users/triok.t/.claude/projects/-Users-triok-t \
         /Users/triok.t/.claude/projects/-Users-triok.t
```

This makes `rtk discover` find sessions from cwd `/Users/triok.t`, but
it causes `rtk discover -a` to double-count the project (~6% inflation),
since directory walk visits both the real dir and the symlink.

## Proposed fix

Align rtk's cwd→slug conversion with Claude Code's escaping rule:
replace both `/` and `.` with `-`. (Possibly other special chars too —
worth checking Claude Code's escaper to be safe.)

niklasmarderx:
Claude Code replaces both `/` and `.` with `-` when computing a project directory name from the working directory path. `rtk` only replaced `/`, so for users with a dot in their path (e.g. `/Users/first.last`) the computed slug didn't match Claude's directory name and `rtk discover` returned zero sessions.

The fix is a one-character change in `encode_project_path`: replace `['/', '.']` instead of just `'/'`.

Two tests added to cover the dot case and multiple consecutive dots.

Closes #1457

### DIFF
--- a/src/discover/provider.rs
+++ b/src/discover/provider.rs
@@ -57,8 +57,9 @@ impl ClaudeProvider {
 
     /// Encode a filesystem path to Claude Code's directory name format.
     /// `/Users/foo/bar` → `-Users-foo-bar`
+    /// `/Users/first.last/bar` → `-Users-first-last-bar`
     pub fn encode_project_path(path: &str) -> String {
-        path.replace('/', "-")
+        path.replace(['/', '.'], "-")
     }
 }
 
@@ -334,6 +335,26 @@ mod tests {
         assert_eq!(
             ClaudeProvider::encode_project_path("/Users/foo/bar/"),
             "-Users-foo-bar-"
+        );
+    }
+
+    #[test]
+    fn test_encode_project_path_dot_in_username() {
+        // Claude Code replaces both '/' and '.' with '-'.
+        // A cwd like /Users/first.last must produce the same slug as
+        // Claude's projects directory (-Users-first-last), otherwise
+        // `rtk discover` finds zero sessions for that project.
+        assert_eq!(
+            ClaudeProvider::encode_project_path("/Users/first.last/my-project"),
+            "-Users-first-last-my-project"
+        );
+    }
+
+    #[test]
+    fn test_encode_project_path_multiple_dots() {
+        assert_eq!(
+            ClaudeProvider::encode_project_path("/Users/a.b.c/proj"),
+            "-Users-a-b-c-proj"
         );
     }
 
--- a/src/discover/provider.rs
+++ b/src/discover/provider.rs
@@ -56,10 +56,24 @@ impl ClaudeProvider {
     }
 
     /// Encode a filesystem path to Claude Code's directory name format.
-    /// `/Users/foo/bar` → `-Users-foo-bar`
-    /// `/Users/first.last/bar` → `-Users-first-last-bar`
+    ///
+    /// Claude Code replaces `/`, `.`, `_`, `\`, and any non-ASCII character
+    /// with `-` when computing the project directory slug under `~/.claude/projects/`.
+    ///
+    /// `/Users/foo/bar`          → `-Users-foo-bar`
+    /// `/Users/first.last/bar`   → `-Users-first-last-bar`
+    /// `/home/chris/2_project`   → `-home-chris-2-project`
+    /// `C:\Users\foo\bar`        → `C:-Users-foo-bar`
     pub fn encode_project_path(path: &str) -> String {
-        path.replace(['/', '.'], "-")
+        path.chars()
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes project path encoding in the discover module to replace '.' characters with '-', matching Claude Code's directory naming convention. Without this fix, users with dots in their filesystem path (e.g., /Users/first.last) would get zero sessions from `rtk discover` because the computed slug didn't match Claude's actual directory name.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1457

---

src/discover/provider.rs
```diff
@@ -57,8 +57,9 @@ impl ClaudeProvider {
 
     /// Encode a filesystem path to Claude Code's directory name format.
     /// `/Users/foo/bar` → `-Users-foo-bar`
+    /// `/Users/first.last/bar` → `-Users-first-last-bar`
     pub fn encode_project_path(path: &str) -> String {
-        path.replace('/', "-")
+        path.replace(['/', '.'], "-")
```
KuSh: according to https://github.com/anthropics/claude-code/issues/24067, `_` is also replaced. Could you test and update the code if that's true?

according to https://github.com/anthropics/claude-code/issues/40946, non ASCII characters are also normalized

I don't found references but on windows, `\\` should also be normalized IMHO

niklasmarderx:
Good catch — pushed a follow-up commit that extends the encoding to cover `_`, `\`, and non-ASCII characters. Used `chars().map()` instead of chained `replace()` calls since we now handle four cases plus the Unicode check. Added three tests referencing the upstream issues.

src/discover/provider.rs
```diff
@@ -56,9 +56,24 @@ impl ClaudeProvider {
...
-        path.replace('/', "-")
+        path.chars()
+            .map(|c| {
+                if c == '/' || c == '.' || c == '_' || c == '\\' || !c.is_ascii() {
+                    '-'
+                } else {
+                    c
+                }
+            })
+            .collect()
```
KuSh, line 78: nitpick:

```rust
const SANITIZED_CHARS: &[char] = &['/', '.', '_', '\\'];

path.chars()
    .map(|c| {
        if !c.is_ascii() || SANITIZED_CHARS.contains(&c) {
            '-'
        } else {
            c
        }
    })
    .collect()
```

KuSh (changes requested):

niklasmarderx:
Done — extracted `SANITIZED_CHARS` constant and flipped the condition to `!c.is_ascii() || SANITIZED_CHARS.contains(&c)` as suggested.

KuSh (approved):
Thanks!

PR #1481, opened 2026-04-23
feat(hooks): add `transparent_prefixes` config for wrapper commands

vsumner:
## Problem

When a shell command is wrapped in a transparent env-wrapper, rtk can't see past the wrapper to route the inner command to a filter:

```bash
$ rtk rewrite "docker exec mycontainer git status"
exit=1   # no rewrite — passthrough

$ rtk rewrite "git status"
rtk git status   # exit=3 ask
```

This affects project layouts where every command flows through a wrapper, such as `docker exec mycontainer`, `direnv exec .`, `poetry run`, or `bundle exec`. In those layouts, test runners, linters, build tools, and git commands bypass rtk and the agent sees raw output.

rtk already handles the same shape for shell builtins via `SHELL_PREFIX_BUILTINS` (`noglob`, `command`, `builtin`, `exec`, `nocorrect`): strip the prefix, recurse on the inner command, then re-prepend the prefix. Third-party wrappers have the same transparent semantics, but need to be user-configurable.

## Fix

Adds a `[hooks].transparent_prefixes` config field:

```toml
[hooks]
transparent_prefixes = ["docker exec mycontainer", "direnv exec ."]
```

With this configured:

```bash
$ rtk rewrite "docker exec mycontainer cargo test"
docker exec mycontainer rtk cargo test   # exit=3 ask
```

Matching is strict and literal: a configured prefix `"foo bar"` matches a command that starts with `"foo bar "` or strictly equals `"foo bar"`, not anything else.

## Implementation

- Keeps the production registry entry point named `rewrite_command(cmd, excluded, transparent_prefixes)`.
- Uses local test helpers for no-prefix cases so existing registry tests stay readable without keeping a dead production wrapper.
- Normalizes configured prefixes once at the rewrite boundary: trim, drop empty entries, dedupe, then sort longest-first so overlapping prefixes choose the most specific match.
- Threads transparent prefixes through compound command rewriting and segment rewriting.
- Applies transparent-prefix stripping with the same strip-recurse-reprepend contract as shell builtins.
- Checks `RTK_DISABLED` from the already-computed env prefix before recursing, preserving the existing warning comments and avoiding duplicate prefix scans.
- Preserves `--` in `rtk hook check ...` command capture, with a regression test for Clap parsing.

## Tests

Focused coverage includes:

- config deserialization and backwards-compatible missing-field behavior
- strip/reprepend for wrapper prefixes
- unknown inner command passthrough
- unmatched prefix passthrough
- shell builtin + transparent prefix composition
- env prefix + transparent prefix composition
- multiple and overlapping prefixes
- strict whole-word matching
- blank-prefix normalization
- compound command rewriting
- excluded inner command behavior
- recursion bound for pathological prefix configs
- `rtk hook check shadowenv exec -- git status` preserving `--`

## Verification

```bash
cargo test transparent_prefix --bin rtk
cargo test hook_check --bin rtk
```

Full gate passes: `1706 passed; 0 failed; 6 ignored`. Existing unrelated clippy warnings remain on this branch.

### DIFF
--- a/src/core/config.rs
+++ b/src/core/config.rs
@@ -41,6 +41,10 @@ pub struct HooksConfig {
     /// `direnv exec`, `nix develop --command`, `docker exec <container>`,
     /// `poetry run`, or `bundle exec`.
     ///
+    /// Matching is literal, not pattern-based. `docker exec <container>` is an
+    /// example shape, not a wildcard; configure the exact concrete prefix you
+    /// actually use, such as `docker exec app`.
+    ///
     /// Extends the built-in `SHELL_PREFIX_BUILTINS` list (`noglob`, `command`,
     /// `builtin`, `exec`, `nocorrect`) with user- or organization-specific
     /// wrappers. Matching is whole-word: a configured prefix `"foo bar"` matches
--- a/src/core/config.rs
+++ b/src/core/config.rs
@@ -29,6 +29,25 @@ pub struct HooksConfig {
     /// Survives `rtk init -g` re-runs since config.toml is user-owned.
     #[serde(default)]
     pub exclude_commands: Vec<String>,
+
+    /// Wrapper prefixes that should be transparently stripped before routing
+    /// to a filter, then re-prepended on the rewrite. For example, with
+    /// `transparent_prefixes = ["shadowenv exec --"]`, the command
+    /// `shadowenv exec -- git status` rewrites to
+    /// `shadowenv exec -- rtk git status` instead of passing through unrewritten.
+    ///
+    /// Useful for any per-project env wrapper that sits in front of every
+    /// command — e.g. [shadowenv](https://github.com/Shopify/shadowenv),
+    /// `direnv exec`, `nix develop --command`, `docker exec <container>`,
+    /// `poetry run`, or `bundle exec`.
+    ///
+    /// Extends the built-in `SHELL_PREFIX_BUILTINS` list (`noglob`, `command`,
+    /// `builtin`, `exec`, `nocorrect`) with user- or organization-specific
+    /// wrappers. Matching is whole-word: a configured prefix `"foo bar"` matches
+    /// a command that starts with `"foo bar "` (or equals `"foo bar"`), not
+    /// `"foobarbaz"`.
+    #[serde(default)]
+    pub transparent_prefixes: Vec<String>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -201,6 +220,32 @@ exclude_commands = ["curl", "gh"]
     fn test_hooks_config_default_empty() {
         let config = Config::default();
         assert!(config.hooks.exclude_commands.is_empty());
+        assert!(config.hooks.transparent_prefixes.is_empty());
+    }
+
+    #[test]
+    fn test_hooks_config_transparent_prefixes_deserialize() {
+        let toml = r#"
+[hooks]
+transparent_prefixes = ["direnv exec .", "nix develop --command"]
+"#;
+        let config: Config = toml::from_str(toml).expect("valid toml");
+        assert_eq!(
+            config.hooks.transparent_prefixes,
+            vec!["direnv exec .", "nix develop --command"]
+        );
+    }
+
+    #[test]
+    fn test_hooks_config_transparent_prefixes_missing_is_empty() {
+        // Older configs that predate this field must still parse.
+        let toml = r#"
+[hooks]
+exclude_commands = ["curl"]
+"#;
+        let config: Config = toml::from_str(toml).expect("valid toml");
+        assert_eq!(config.hooks.exclude_commands, vec!["curl"]);
+        assert!(config.hooks.transparent_prefixes.is_empty());
     }
 
     #[test]
--- a/src/core/config.rs
+++ b/src/core/config.rs
@@ -32,24 +32,23 @@ pub struct HooksConfig {
 
     /// Wrapper prefixes that should be transparently stripped before routing
     /// to a filter, then re-prepended on the rewrite. For example, with
-    /// `transparent_prefixes = ["shadowenv exec --"]`, the command
-    /// `shadowenv exec -- git status` rewrites to
-    /// `shadowenv exec -- rtk git status` instead of passing through unrewritten.
+    /// `transparent_prefixes = ["docker exec mycontainer"]`, the command
+    /// `docker exec mycontainer git status` rewrites to
+    /// `docker exec mycontainer rtk git status` instead of passing through
+    /// unrewritten.
     ///
     /// Useful for any per-project env wrapper that sits in front of every
-    /// command — e.g. [shadowenv](https://github.com/Shopify/shadowenv),
-    /// `direnv exec`, `nix develop --command`, `docker exec <container>`,
-    /// `poetry run`, or `bundle exec`.
+    /// command — e.g. `docker exec mycontainer`, `direnv exec .`, `poetry run`,
+    /// or `bundle exec`.
     ///
-    /// Matching is literal, not pattern-based. `docker exec <container>` is an
-    /// example shape, not a wildcard; configure the exact concrete prefix you
-    /// actually use, such as `docker exec app`.
+    /// Matching is literal, not pattern-based. Configure the exact concrete
+    /// prefix you actually use, such as `docker exec mycontainer`.
     ///
     /// Extends the built-in `SHELL_PREFIX_BUILTINS` list (`noglob`, `command`,
     /// `builtin`, `exec`, `nocorrect`) with user- or organization-specific
-    /// wrappers. Matching is whole-word: a configured prefix `"foo bar"` matches
-    /// a command that starts with `"foo bar "` (or equals `"foo bar"`), not
-    /// `"foobarbaz"`.
+    /// wrappers. Matching is strict: a configured prefix `"foo bar"` matches
+    /// a command that starts with `"foo bar "` (or strictly equals `"foo bar"`),
+    /// not anything else.
     #[serde(default)]
     pub transparent_prefixes: Vec<String>,
 }
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -391,14 +391,18 @@ pub fn has_rtk_disabled_prefix(cmd: &str) -> bool {
     prefix_part.contains("RTK_DISABLED=")
 }
 
-/// Strip RTK_DISABLED=X and other env prefixes, return the actual command.
-pub fn strip_disabled_prefix(cmd: &str) -> &str {
+fn split_env_prefix(cmd: &str) -> (&str, &str) {
     let trimmed = cmd.trim();
     let stripped = ENV_PREFIX.replace(trimmed, "");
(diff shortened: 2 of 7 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟢 **Risk** | `low` |

### Summary

Adds a new `transparent_prefixes` config field under `[hooks]` that lets users specify wrapper commands (e.g. `shadowenv exec --`, `direnv exec .`) to be transparently stripped before routing to a filter and re-prepended on rewrite. This extends the existing built-in shell prefix builtins pattern to support third-party env wrappers, ensuring commands run through wrappers are properly rewritten by rtk.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

---

src/discover/registry.rs
```diff
@@ -438,7 +442,36 @@ fn strip_trailing_redirects(cmd: &str) -> (&str, &str) {
...
 /// For pipes (`|`), only rewrites the left-hand command (pipe targets stay raw),
 /// but continues rewriting segments after subsequent `&&`/`||`/`;` operators.
+/// Back-compat entry point preserved for call sites (including the ~150
+/// existing unit tests in this module) that never needed wrapper-prefix
+/// stripping. New production callers should prefer
+/// [`rewrite_command_with_prefixes`].
+#[allow(dead_code)]
 pub fn rewrite_command(cmd: &str, excluded: &[String]) -> Option<String> {
+    rewrite_command_with_prefixes(cmd, excluded, &[])
+}
```
KuSh: Just move the rewrite_command in the `tests` module as a simplification for tests that don't need to handle transparent prefixes. Allowing dead_code to avoid impacting tests isn't a good option. Checking rapidly it seems you could even strip the excluded parameter as almost no tests use it.

src/core/config.rs
```diff
@@ -29,6 +29,29 @@ pub struct HooksConfig {
     /// Survives `rtk init -g` re-runs since config.toml is user-owned.
     #[serde(default)]
     pub exclude_commands: Vec<String>,
+
+    /// Wrapper prefixes that should be transparently stripped before routing
+    /// to a filter, then re-prepended on the rewrite. For example, with
+    /// `transparent_prefixes = ["shadowenv exec --"]`, the command
```
KuSh: Are you sure `--` isn't swallowed by clap? It seemed to me it didn't preserve arguments separator

src/core/config.rs
```diff
@@ -29,6 +29,29 @@ pub struct HooksConfig {
...
+    ///
+    /// Useful for any per-project env wrapper that sits in front of every
+    /// command — e.g. [shadowenv](https://github.com/Shopify/shadowenv),
+    /// `direnv exec`, `nix develop --command`, `docker exec <container>`,
+    /// `poetry run`, or `bundle exec`.
+    ///
+    /// Matching is literal, not pattern-based. `docker exec <container>` is an
+    /// example shape, not a wildcard; configure the exact concrete prefix you
+    /// actually use, such as `docker exec app`.
+    ///
```
KuSh: Just use `docker exec mycontainer` as example

src/core/config.rs
```diff
@@ -29,6 +29,29 @@ pub struct HooksConfig {
...
+    /// Extends the built-in `SHELL_PREFIX_BUILTINS` list (`noglob`, `command`,
+    /// `builtin`, `exec`, `nocorrect`) with user- or organization-specific
+    /// wrappers. Matching is whole-word: a configured prefix `"foo bar"` matches
+    /// a command that starts with `"foo bar "` (or equals `"foo bar"`), not
+    /// `"foobarbaz"`.
```
KuSh:
```suggestion
    /// wrappers. Matching is strict: a configured prefix `"foo bar"` matches
    /// a command that starts with `"foo bar "` (or strictly equals `"foo bar"`), not
    /// anything else.
```

src/discover/registry.rs
KuSh, line 406: Just make strip_disabled_prefix returns the stripped part as the second member of the 2-tuple, and make necessary changes to code

src/core/config.rs
KuSh: There's other occurrences to fix too.

src/discover/registry.rs
KuSh: Can you take the opportunity to correct this piece of code and use trimmed instead?

src/discover/registry.rs
```diff
@@ -634,8 +676,25 @@ fn compile_exclude_patterns(patterns: &[String]) -> Vec<ExcludePattern> {
...
-    rewrite_segment_inner(seg, excluded, 0)
+fn normalize_transparent_prefixes(prefixes: &[String]) -> Vec<String> {
+    let mut normalized: Vec<String> = prefixes
+        .iter()
+        .map(|prefix| prefix.trim())
+        .filter(|prefix| !prefix.is_empty())
+        .map(str::to_string)
+        .collect();
+
+    normalized.sort_unstable_by(|a, b| b.len().cmp(&a.len()).then_with(|| a.cmp(b)));
```
KuSh: Why not just call sort?

src/discover/registry.rs
```diff
@@ -655,15 +719,43 @@ fn rewrite_segment_inner(seg: &str, excluded: &[ExcludePattern], depth: usize) -
         return None;
     }
 
+    let (env_prefix, rest_after_env) = split_env_prefix(trimmed);
+    if !env_prefix.is_empty() {
+        let rewritten =
+            rewrite_segment_inner(rest_after_env, excluded, transparent_prefixes, depth + 1)?;
+        if has_rtk_disabled_prefix(trimmed) {
```
KuSh: You have already computed the env prefix, it's a shame to do it a second time in `has_rtk_disabled_prefix`. Also check that before calling `rewrite_segment_inner`

And please retain original comments

src/discover/registry.rs
```diff
@@ -655,15 +719,43 @@ fn rewrite_segment_inner(seg: &str, excluded: &[ExcludePattern], depth: usize) -
...
+        }
+    }
+
+    // User-configured wrapper prefixes (e.g. `shadowenv exec --`). Same
+    // strip-recurse-reprepend contract as the builtin list above.
+    for prefix in transparent_prefixes {
+        let p = prefix.trim();
+        if p.is_empty() {
+            continue;
+        }
```
KuSh: You have already trimmed them and filtered the empty ones when building transparent_prefixes from config (or you should have)

src/discover/registry.rs
```diff
@@ -387,14 +387,18 @@ pub fn has_rtk_disabled_prefix(cmd: &str) -> bool {
...
-pub fn strip_disabled_prefix(cmd: &str) -> &str {
+fn split_env_prefix(cmd: &str) -> (&str, &str) {
     let trimmed = cmd.trim();
     let stripped = ENV_PREFIX.replace(trimmed, "");
-    // stripped is a Cow<str> that borrows from trimmed when no replacement happens.
-    // We need to return a &str into the original, so compute the offset.
     let prefix_len = trimmed.len() - stripped.len();
-    trimmed[prefix_len..].trim_start()
+    let prefix_part = &trimmed[..prefix_len];
+    let rest = trimmed[prefix_len..].trim_start();
```
```suggestion
    let rest = stripped.trim();
```

src/discover/registry.rs
```diff
@@ -706,29 +798,15 @@ fn rewrite_segment_inner(seg: &str, excluded: &[ExcludePattern], depth: usize) -
...
-    // #345: RTK_DISABLED=1 in env prefix → skip rewrite entirely
-    // #508: warn on stderr so agents learn to stop overusing it
-    if has_rtk_disabled_prefix(cmd_part) {
-        eprintln!(
-            "[rtk] RTK_DISABLED=1 detected — skipping filter for this command. \
-             Remove RTK_DISABLED=1 to restore token savings."
-        );
-        return None;
-    }
+    let cmd_clean = cmd_part;
```
KuSh: Just use cmd_part

KuSh (changes requested):

src/main.rs
```diff
@@ -2117,12 +2117,12 @@ fn run_cli() -> Result<i32> {
                 0
             }
             HookCommands::Check { agent: _, command } => {
-                use crate::discover::registry::rewrite_command;
+                use crate::discover::registry::rewrite_command_with_prefixes;
```
KuSh: After a second thought, I'm not sure preserving the tests but refactoring main code isn't the right move.
Keep the `rewrite_command` name, and refactor tests or use dedicated test functions that wraps rewrite_command if you want to avoid passing empty arguments everywhere

src/discover/registry.rs
KuSh, line 405: Again just make strip_disabled_prefix returns the stripped part as the second member of the 2-tuple, and make necessary changes to code

src/discover/registry.rs
```diff
@@ -627,15 +658,33 @@ fn compile_exclude_patterns(patterns: &[String]) -> Vec<ExcludePattern> {
...
+    let mut normalized: Vec<String> = prefixes
+        .iter()
+        .map(|prefix| prefix.trim())
+        .filter(|prefix| !prefix.is_empty())
+        .map(str::to_string)
+        .collect();
+
+    normalized.sort();
+    normalized.dedup();
+    normalized.sort_unstable_by_key(|prefix| std::cmp::Reverse(prefix.len()));
```
KuSh: This seems not needed anymore now that you sort just before deduping, or did I miss something?

src/discover/registry.rs
```diff
@@ -781,6 +845,10 @@ mod tests {
     use super::super::report::RtkStatus;
     use super::*;
 
+    fn rewrite_command(cmd: &str, excluded: &[String]) -> Option<String> {
```
KuSh: that's a bit misleading as it shadows parent function. This is a good first step in the right direction but an explicit name would be better, and easy to refactor now that you have done that first step. Something like rewrite_command_simple or rewrite_command_no_prefixes (I'm bad at naming things but you get the idea)

src/hooks/hook_cmd.rs
```diff
@@ -510,6 +510,10 @@ fn run_cursor_inner_with_rules(
 mod tests {
     use super::*;
 
+    fn rewrite_command(cmd: &str, excluded: &[String]) -> Option<String> {
```
KuSh: Same remark, find a better name to avoid shadowing the real function

src/hooks/rewrite_cmd.rs
```diff
@@ -53,20 +53,24 @@ pub fn run(cmd: &str) -> anyhow::Result<()> {
 mod tests {
     use super::*;
 
+    fn rewrite_command(cmd: &str) -> Option<String> {
```
KuSh: name

KuSh (changes requested):
LGTM except for one comment you missed and minor function shadowing!
Thanks for that

vsumner:
@KuSh This should be good if you have a moment

KuSh (approved):

KuSh:
Thanks @vsumner !

PR #678, opened 2026-03-18
fix(rewrite): route pnpm lint through rtk pnpm

TechWizard9999:
## Description
Fixes rewrite routing for `pnpm lint` so it no longer gets forced into `rtk lint` (which can cause ESLint-oriented retry loops in non-ESLint projects).

## Related Issue
Closes #665

## Changes Made
- updated discovery pattern matching so `pnpm lint` is recognized by the `pnpm` rule path
- removed `pnpm lint` from `rtk lint` rewrite prefixes
- added rewrite regression tests:
  - `pnpm lint` -> `rtk pnpm lint`
  - `pnpm lint --fix` -> `rtk pnpm lint --fix`
- added changelog note under Unreleased bug fixes

## Files Changed
- `src/discover/rules.rs` - rewrite/classification adjustments
- `src/discover/registry.rs` - regression tests for pnpm lint rewrite behavior
- `CHANGELOG.md` - unreleased bug-fix note for #665

## Testing
- [x] `cargo fmt --all --check`
- [x] `rtk cargo clippy --all-targets` (2 pre-existing warnings in untouched files)
- [x] `rtk cargo test`
- [x] `rtk cargo test discover::registry::tests::test_rewrite_pnpm_lint_routes_to_pnpm`
- [x] `rtk cargo test discover::registry::tests::test_rewrite_pnpm_lint_with_args_routes_to_pnpm`
- [x] `./target/debug/rtk rewrite "pnpm lint"` -> `rtk pnpm lint`

## Checklist
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Documentation updated
- [x] No breaking changes introduced

cc @pszymkowiak for review

PR #443, opened 2026-03-09
feat(npm): structured NpmCommands sub-enum with smart script routing

jackfreem:
## Why
The flat `rtk npm <args>` design had no way to distinguish `npm run <script>` from other npm subcommands (`install`, `ci`, `audit`, …), and provided no script-aware filter routing. PR #440 patched the immediate double-run bug; this PR replaces the design root-cause.

## What
- Add `NpmCommands` sub-enum (`Run { script, args }` + `Other` passthrough), mirroring the `PnpmCommands` pattern from PR #232
- Route well-known scripts to specialised filters in `main.rs`:
  - `build` → `next_cmd`
  - `typecheck` / `tsc` → `tsc_cmd`
  - `lint` → `lint_cmd`
  - everything else → `npm_cmd::run_script` (boilerplate stripping)
- Add `npm_cmd::run_passthrough` for non-`run` subcommands (`install`, `ci`, `audit`, …)
- Make `filter_npm_output` `pub` so it can be reused by other modules
- Remove the old `npm_cmd::run` catch-all that caused the double-`run` bug

## References
- Fixes #438 (structurally — explicit subcommand parsing makes the double-run impossible)
- Supersedes / builds on PR #440 (minimal Path A fix)
- PR #232 is the direct pnpm precedent this implementation mirrors

Issue #316, opened 2026-03-04
Support Biome for lint command


ysulyma:
I use [Biome](https://biomejs.dev/) for linting instead of ESLint. This can be detected either through my `package.json`, which contains

```jsonc
  "scripts": {
    "lint": "biome check --fix"
  },
```
or via the presence of the `biome.json` file (adjacent to `package.json`).

In this case, `rtk lint` should invoke Biome instead of ESLint.

PR #1491, opened 2026-04-24
fix(rewrite): route npm lint scripts through rtk npm (biome-safe)

### LINKED ISSUE #1489: bug(rewrite): npm run lint forced into rtk lint on Biome projects
### Environment

- rtk 0.37.2 (Homebrew, `/opt/homebrew/bin/rtk`)
- macOS darwin 25.4.0 (arm64)
- Claude Code 2.1.x with the `rtk hook claude` PreToolUse hook shipped by `rtk init`

### Repro

Project uses Biome, not ESLint. `package.json`:

```json
{
  "scripts": {
    "lint": "biome check"
  }
}
```

1. `rtk rewrite "npm run lint"` -> `rtk lint` (exit 0).
2. Claude Code hook rewrites `npm run lint` to `rtk lint` before execution.
3. `rtk lint` shells out to biome, biome exits 0 with plain-text output, and the lint adapter attempts to parse biome's output as ESLint JSON.
4. Observed output: `ESLint output (JSON parse failed: EOF while parsing a value at line 1 column 0)`, process exit 0.

### Expected

`npm run lint` passes through to the project's actual lint script without being forced into the ESLint adapter. Either:

- route `npm run lint` through `rtk npm run lint` (mirrors the fix in PR #678 for `pnpm lint`), or
- detect Biome from `package.json` / presence of `biome.json` and dispatch to the existing `rtk biome` filter added in #316.

Either way, the adapter's JSON parse should never run on biome output.

### Actual

`rtk rewrite` sends every `npm run lint` into the ESLint adapter regardless of what the script actually invokes. The adapter fails to parse biome output, prints a JSON-parse-failed warning, and exits 0. Agents see exit 0 and move on; the real lint result is discarded silently.

Reproduced cleanly with an empty-stdout target:

```
$ rtk lint ./no-op.sh
ESLint output (JSON parse failed: EOF while parsing a value at line 1 column 0)
$ echo $?
0
```

### Also tried

- `[hooks] exclude_commands = ["npm run lint"]` in `config.toml`: no effect on `rtk rewrite` or the Claude hook (matches #243 / #1335).
- `.rtk/filters.toml` with a matching `match_command`: only applies to unhandled commands; cannot override the built-in rewrite (also noted in #1335).
- `RTK_DISABLED=1 rtk lint`: still runs the ESLint-JSON parse and still prints the warning. Not an escape hatch for this collision.
- `rtk proxy npm run lint`: works per-command but strips all downstream filtering if the agent reflexively applies it more broadly.

### Related

- #665 (open) `pnpm lint` shows the same symptom on Biome projects.
- PR #678 (open, checks green) routes `pnpm lint` through `rtk pnpm lint`. Does not cover `npm run lint`.
- PR #443 (open) is the active `NpmCommands` sub-enum restructure for smart script routing. As currently written it wires `npm run lint -> lint_cmd` (the ESLint adapter), which codifies this bug rather than fixing it. Addressing the Biome case here, before #443 merges, seems cheapest.
- #316 (closed, 0.30.0) added the Biome output filter, but only on the direct `rtk biome lint` path, not through the `rtk lint` rewrite.
- #243 / #1335 (both open) document that `[hooks] exclude_commands` is not threaded into the Claude Code hook rewrite path, so config-level opt-out is not currently an option.

### Likely cause

`src/discover/rules.rs` maps any `npm run lint` (and `pnpm lint`, `biome check`, `npm run check`) to the `rtk lint` adapter, which assumes ESLint JSON. PR #678 patches the `pnpm lint` row; the `npm run lint` row should follow the same routing change. The natural landing spot is PR #443's `NpmCommands::Run { script: "lint", .. }` branch: either detect Biome from `package.json` / `biome.json` and dispatch to the existing biome filter, or fall through to the generic `npm_cmd::run_script` boilerplate-stripper when detection is uncertain. Either beats running the ESLint JSON parser on biome output.

### Impact

AI coding agents see exit 0 with a suspicious-looking warning and often ignore it. Real lint failures (and anything downstream that gates on lint) can pass silently on Biome projects for multiple commits before the regression is noticed by a human.

LukaPrebil:
## Summary

- Fixes #1489: route `npm [run|run-script|rum|urn|x] lint`, `npm lint`, and `npx lint` through `rtk npm` / `rtk npx` instead of `rtk lint`, so the ESLint JSON adapter never runs on Biome projects (was silently exiting 0 on parse failure)
- Mirrors PR #678's pnpm-side fix: early-return in `rewrite_segment_inner` + remove stale npm/npx entries from the lint rule's `rewrite_prefixes` + extend the npm rule regex so bare `npm lint` still classifies as `rtk npm`
- Coordination note for PR #443: its `"lint" => lint_cmd::run(...)` dispatch in `main.rs` would re-introduce this bug if merged as-is; please fall through to `npm_cmd::run_script` for the `lint` script before merging

## Test plan


LukaPrebil:
## Smoke test (buggy 0.37.2 vs this PR)

Scratch `package.json` with `"lint": "<mock>"` that emits diagnostics and exits 1:

| Scenario | Installed rtk 0.37.2 | This PR |
|---|---|---|
| Biome-shaped diagnostics, exit 1 | `ESLint output (JSON parse failed: EOF ...)`, **exit 0** (silent) | Real biome diagnostics, **exit 1** (correct) |
| ESLint-shaped diagnostics, exit 1 | n/a (same class of parse failure) | Real ESLint diagnostics, **exit 1** (correct) |
| Clean pass with message, exit 0 | unchanged | "All files pass linting." preserved, exit 0 |
| Silent success, exit 0 no output | unchanged | No false warning, exit 0 |

And rewrite coverage (`rtk rewrite "<cmd>"`):

| Command | Before | After |
|---|---|---|
| `npm run lint` | `rtk lint` | `rtk npm run lint` |
| `npm run lint --fix` | `rtk lint` | `rtk npm run lint --fix` |
| `npm run-script lint` | `rtk lint` | `rtk npm run-script lint` |
| `npm lint` | `rtk lint` | `rtk npm lint` |
| `npm rum/urn/x lint` | `rtk lint` (`npm x lint` was not rewriteable) | `rtk npm <alias> lint` |
| `npx lint` | `rtk lint` | `rtk npx lint` |
| `FOO=1 npm run lint` | `FOO=1 rtk lint` | `FOO=1 rtk npm run lint` |
| `npm run lint 2>&1` | n/a | `rtk npm run lint 2>&1` |
| `npm lint-staged` (regression) | not rewritten | not rewritten (word-boundary protected) |
| `biome check` (out of scope) | `rtk lint check` | unchanged |
| `eslint src/` (out of scope) | `rtk lint src/` | unchanged |
| `pnpm lint` (#678's scope) | `rtk lint` | unchanged |

## Accepted tradeoff: classify/rewrite drift

`classify_command("npm run lint")` still returns `rtk lint` because the lint rule's regex pattern was intentionally left alone (last-match wins, and the lint rule comes after the npm rule in `RULES[]`). Only the rewrite short-circuits via the new early-return. This affects `rtk discover` analytics output, not any real command execution. Tightening classify would require removing the `p?np(m|x)` alternation from the lint rule's pattern, which widens scope to the `biome` / `eslint` bare-PM variants too — left for a follow-up if maintainers want classify/rewrite parity.

The same drift already exists for `pnpm lint` after PR #678 ships, so the behaviour is consistent across the two sides.

## Token-savings tradeoff

For projects that really do use ESLint, `npm run lint` previously ran through `rtk lint` which added `-f json` and applied the structured `filter_eslint_json` (high compression). After this PR, `rtk npm run lint` runs the user's literal script (no `-f json` injection) and applies only the generic `filter_npm_output` (boilerplate strip). That is a savings regression for pure-ESLint projects in exchange for correctness on Biome projects. Aligns with the "Correctness VS Token Savings" principle in CONTRIBUTING.md, and is recoverable via package.json-aware filter dispatch (see follow-ups).

## Follow-up issues (will file separately)

1. `rtk lint` silently exits 0 when auto-defaulted ESLint produces non-JSON output — adapter-level hardening. The reproducer from #1489 (`rtk lint ./no-op.sh` → "ESLint output (JSON parse failed: EOF ...)" with exit 0) still reproduces after this PR, because this PR only fixes the rewrite path.
2. `biome check` currently routes `rtk lint biome check` → `filter_generic_lint`. It should route through the existing biome TOML filter added in #316, either by a new rewrite rule direct to `rtk biome` or by teaching `rtk lint` to consult the TOML registry when it detects `biome` as the explicit linter.
3. `rtk npm run <script>`: extend script-body detection in `package.json` to eslint/prettier/tsc. My next PR (stacked on this one) seeds the detector + wires the biome filter only; the other tools are additive follow-ups.

pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Fixes a bug where `npm [run|run-script|rum|urn|x] lint`, `npm lint`, and `npx lint` commands were incorrectly routed through `rtk lint` (the ESLint JSON adapter), which silently swallowed parse failures on Biome projects. Now these commands route through `rtk npm` / `rtk npx` instead, with an early-return in `rewrite_segment_inner` and updated rule patterns.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1489

---

PR #1500, opened 2026-04-24
fix(git): passthrough `git branch -v`/`-vv`/`--verbose` in list mode (#1499)

### LINKED ISSUE #1499: `rtk git branch -vv` strips commit hashes, tracking markers, and remote metadata — filter has no `-vv` awareness
`rtk git branch -vv` drops commit hashes, the `[origin/...]` upstream tracking marker, and commit subjects — the very information `-vv` exists to show. `filter_branch_output` is designed around bare `git branch -a` output and has no handling for the verbose format, so metadata the user explicitly requested is silently discarded.

## Reproduction

```
$ RTK_DISABLED=1 git branch -vv
  feature-a  abc12345 Some commit subject
  feature-b  def67890 Another commit subject
* main       12345678 [origin/main] Latest commit on main

$ rtk git branch -vv
* main
  feature-a
  feature-b
  remote-only (3):
    topic-x
    agent/topic-y-abc123
    topic-z
```

The upstream marker `[origin/main]`, the commit hashes, and the subjects are gone. From the agent's point of view, `git branch -vv` is indistinguishable from `git branch -a` — the flag had no effect.

## Cause

`-vv` isn't in the list-flag set at `run_branch`, so the command falls through to List mode:

https://github.com/rtk-ai/rtk/blob/80a6fe606f73b19e52b0b330d242e62a6c07be42/src/cmds/git/git.rs#L1144-L1164

List mode adds `-a --no-color`, passes `-vv` through to `git`, then pipes the result through `filter_branch_output`:

https://github.com/rtk-ai/rtk/blob/80a6fe606f73b19e52b0b330d242e62a6c07be42/src/cmds/git/git.rs#L1234-L1261

`filter_branch_output` was designed for bare `git branch -a` output (see the test fixtures at lines 1885-1925 — none use `-vv` format) and reshapes everything into the compact `* current` + locals + `remote-only` block shown above. There's no code path that preserves `[origin/...]` markers or hash + subject when `-vv` is present:

https://github.com/rtk-ai/rtk/blob/80a6fe606f73b19e52b0b330d242e62a6c07be42/src/cmds/git/git.rs#L1273-L1328

## Impact

`-vv` is the canonical way to check upstream tracking state — "does my local branch have an upstream? is it ahead/behind?". When RTK answers the agent's `git branch -vv` with bare names, the agent has no way to see tracking configuration and may conclude branches have no upstream when they do. `RTK_DISABLED=1 git branch -vv` works as a workaround, but as with #1474 and #730, the agent only reaches for it once something else tips it off that the output is incomplete.

## Suggestion

The project's own design philosophy covers this case:

> When a user or LLM explicitly requests detailed output via flags (e.g., `git log --comments`, `cargo test -- --nocapture`, `ls -la`), respect that intent. Compressing explicitly-requested detail defeats the purpose.

`-v` / `-vv` / `--verbose` fit that rule exactly. One option is to detect those flags in `run_branch` and passthrough raw. Another is to make `filter_branch_output` aware of the verbose format and preserve the hash + tracking marker + subject per line. Passthrough is simpler and matches the principle more literally.

## Broader Pattern

Same class of bug as #1474, #730, #720 — a filter that isn't flag-aware silently strips content the user asked for, and the hook makes the loss invisible to the agent.

## Related

- #1474 — `gh pr comment --help` outputs canned `ok commented` (same class: flag ignored)
- #730 — `gh pr diff --name-only` returns empty (same class: filter applied to non-matching data)
- #720 — `--comments` silently dropped on `gh issue/pr view` (same class: flag ignored)

## Environment

- rtk 0.37.2
- Windows 11 (Git Bash)

truffle-dev:
Closes #1499.

## Cause

`run_branch` list mode always runs `filter_branch_output`, which compresses `git branch -a`-shaped output into a compact `* current` + locals + `remote-only` block. When the user asks for `-v`/`-vv`/`--verbose`, the filter still ran, stripping the commit hash, `[origin/...]` upstream marker, and subject that `-v` exists to show. From git's point of view the flag had no effect.

The `-v` flag reaches `run_branch` through two different paths depending on position:

1. **trailing position** (`rtk git branch --list -v`): clap's `trailing_var_arg = true` on `Branch { args }` captures `-v` inside `args`.
2. **leading position** (`rtk git branch -vv`): rtk's own `-v` / `--verbose` is declared `global = true` on `Cli`, so clap consumes the `-vv` before the subcommand is parsed. `args` arrives empty and `cli.verbose == 2`.

The original fix I wrote only covered case 1, but the repro in #1499 is case 2, so I had to add the second branch. A small debug build in `/tmp` confirmed the split:

```
$ rtk git branch -vv          → args=[], verbose=2
$ rtk git branch --verbose    → args=[], verbose=1
$ rtk git branch -a -v        → args=["-a","-v"], verbose=0
```

## Fix

`is_verbose_branch_list(args)` recognizes `--verbose` and short-flag groups (so `-v`, `-vv`, `-va`, `-av`, `-avv` all match) while rejecting look-alikes such as `--vim`, `--verify-signatures`, and positional branch names containing `v`. For the case-2 path we inject `-v<N>` before forwarding to `git` so the child process actually sees the flag. Either signal flips `has_verbose_intent`, which bypasses `filter_branch_output` and forwards git's raw output unchanged.

Trade-off: `rtk -v git branch` (rtk-verbose-as-trace) now also forwards `-v` to git and shows verbose branch info. I think this is an improvement; the prior behavior (trace line plus compact list) was an accidental product of the clap collision, and the new behavior matches what anyone typing `-v` near `git branch` actually wants.

## Before / after

Against a small test repo with one local tracking `origin/main` (gone) and one local tracking `origin/feature-1` (gone):

```
$ target/release/rtk-master git branch -vv
git branch
* main
  feature-1

$ target/release/rtk-fix git branch -vv
git branch
  feature-1 7b78477 [origin/feature-1: gone] feature work
* main      8c70500 [origin/main: gone] init

$ RTK_DISABLED=1 git branch -vv
  feature-1 7b78477 [origin/feature-1: gone] feature work
* main      8c70500 [origin/main: gone] init
```

Negative path verified: `rtk git branch` and `rtk git branch -a` still produce the compact filter output.

## Tests

Six new unit tests for `is_verbose_branch_list` covering canonical spellings, combined short-flag groups, no-verbose cases, long flags containing `v`, and positional names containing `v`:

```
test cmds::git::git::tests::test_is_verbose_branch_list_canonical_flags ... ok
test cmds::git::git::tests::test_is_verbose_branch_list_combined_with_list_flags ... ok
test cmds::git::git::tests::test_is_verbose_branch_list_combined_short_flag_group ... ok
test cmds::git::git::tests::test_is_verbose_branch_list_no_verbose ... ok
test cmds::git::git::tests::test_is_verbose_branch_list_does_not_match_long_flags_with_v ... ok
test cmds::git::git::tests::test_is_verbose_branch_list_does_not_match_positional_containing_v ... ok
```

Full suite: `cargo test --bin rtk` → **1612 passed, 6 ignored, 0 failed** on `rust:1.95`. `cargo fmt --all -- --check` clean. `cargo clippy --bin rtk` introduces zero new issues vs master.

## Scope

Only `run_branch` list mode. `--show-current`, write-path (`-d` / `-D` / `-m` / `-M` / rename / upstream edit), and branch creation are untouched. The compact `filter_branch_output` is unchanged so the 80%+ savings on bare `git branch -a` stay intact.

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1163,6 +1163,17 @@ fn run_branch(args: &[String], verbose: u8, global_args: &[String]) -> Result<i3
     // Detect positional arguments (not flags) — indicates branch creation
     let has_positional_arg = args.iter().any(|a| !a.starts_with('-'));
 
+    // Detect verbose-list intent: the user asked for commit hashes,
+    // `[origin/...]` upstream markers, and subjects. There are two ways
+    // that intent reaches us: either a `-v`/`-vv`/`--verbose` survived
+    // in `args` (trailing position, e.g. `rtk git branch --list -v`), or
+    // rtk's global `--verbose` flag consumed it before the subcommand
+    // was parsed (leading position, e.g. `rtk git branch -vv`). Either
+    // source means the compact `filter_branch_output` would defeat the
+    // flag, so we skip the filter and forward git's output as-is.
+    let verbose_in_args = is_verbose_branch_list(args);
+    let has_verbose_intent = verbose_in_args || verbose > 0;
+
     // --show-current: passthrough with raw stdout (not "ok")
     if has_show_flag {
         let mut cmd = git_cmd(global_args);
@@ -1238,6 +1249,14 @@ fn run_branch(args: &[String], verbose: u8, global_args: &[String]) -> Result<i3
         cmd.arg("-a");
     }
     cmd.arg("--no-color");
+    // If rtk's global `--verbose` swallowed the user's `-v`/`-vv`, forward
+    // the equivalent flag to git. Without this the user's intent never
+    // reaches the child process. `args` is checked first so explicit
+    // `--verbose` / `-v` in args wins and avoids duplicating the flag.
+    if !verbose_in_args && verbose > 0 {
+        let count = verbose.min(2);
+        cmd.arg(format!("-{}", "v".repeat(count as usize)));
+    }
     for arg in args {
         cmd.arg(arg);
     }
@@ -1257,7 +1276,13 @@ fn run_branch(args: &[String], verbose: u8, global_args: &[String]) -> Result<i3
         return Ok(result.exit_code);
     }
 
-    let filtered = filter_branch_output(&result.stdout);
+    let filtered = if has_verbose_intent {
+        // Passthrough: the user asked for verbose format. Trim the
+        // trailing newline so println! doesn't double it.
+        result.stdout.trim_end().to_string()
+    } else {
+        filter_branch_output(&result.stdout)
+    };
     println!("{}", filtered);
 
     timer.track(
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes git branch -v/-vv/--verbose passthrough in list mode, which previously had its verbose output stripped by the compact filter. The fix detects verbose intent from both trailing args and rtk's global --verbose flag, bypasses the filter, and injects the flag into the child git process when needed.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1499

---

truffle-dev:
Superseded by #1502, which retargets the same fix to `develop` per CONTRIBUTING. Same diff, cleanly rebased on `develop`, full suite passes (1684/1684).

PR #1171, opened 2026-04-10
feat(cli): add --raw flag for unfiltered command output

iRonin:
## Summary

Adds a `--raw` global flag that bypasses all output filtering and returns raw command output directly. This implements issue #1156.

### src/main.rs
- Added `--raw` global CLI argument to the `Cli` struct
- Sets `RTK_RAW_MODE=1` env var when flag is present, before command dispatch

### src/core/runner.rs
- Modified `run_filtered()` to check for `RTK_RAW_MODE` env var
- When set, skips the filter function and returns output as-is
- Added unit tests for raw mode detection and filter bypass behavior

## Usage

```bash
rtk --raw git log -20          # Full git log, no truncation\nrtk --raw cargo build          # Full cargo build output\nrtk --raw ls -la               # Full ls output, no filtering\n```

## Implementation Details

Uses an environment variable (`RTK_RAW_MODE`) to communicate the flag state from CLI parsing to the command execution pipeline. This approach:
- Works across all command handlers without changing their signatures
- Is simple and efficient (single env var check)\n- Does not affect proxy mode or other existing behavior
\nCloses #1156

PR #1167, opened 2026-04-10
feat(hooks): add RTK_DISABLED=1 env var to bypass all hooks

iRonin:
## Summary

Adds the `RTK_DISABLED=1` environment variable that bypasses all hook rewrites, providing a simple opt-out mechanism for the rewrite system.

## Changes

- **src/hooks/rewrite_cmd.rs**: Added early exit check at the start of `run()` — when `RTK_DISABLED=1`, the function calls `process::exit(1)` (passthrough mode) before any rewrite logic runs.
- **src/hooks/rewrite_cmd.rs**: Added test `test_rtk_disabled_env_var_bypasses_rewrite` to verify the env var check logic.
- **README.md**: Added Environment Variables section documenting `RTK_DISABLED` and `RTK_TELEMETRY_DISABLED`.

## Usage

```bash
export RTK_DISABLED=1
# All hook rewrites are now bypassed (passthrough mode)
```

Closes #1153

PR #1508, opened 2026-04-25
fix(ls): preserve .env in default output

nghiahsgs:
## Problem

`.env` is listed in `NOISE_DIRS` (`src/cmds/system/constants.rs`), so `rtk ls` silently drops it unless the agent passes `-a`. This creates a data-loss hazard:

1. Agent runs `rtk ls` to scan a project's setup state.
2. Output shows `.env.example` but no `.env`.
3. Agent concludes the project isn't configured and writes a fresh `.env` from the template.
4. Real production secrets (DB passwords, API keys) are clobbered.

A user reproduced this directly:

```
$ touch .env
$ rtk ls
.claude/
demo-agents/
README.md  2.6K
notes.md  19.0K
slides.md  31.1K
Summary: 3 files, 2 dirs (3 .md)
```

`.env` is gone from the output. The agent has no way to know it exists.

## Why this matters

`.env` is **configuration the agent must reason about**, not a build artifact like `node_modules/`, `.git/`, or `target/`. The other entries in `NOISE_DIRS` are regenerable; `.env` typically is not.

CONTRIBUTING.md's *Correctness VS Token Savings* principle applies directly:

> When in doubt, preserve correctness.

A few tokens for one filename are cheap. Restoring rotated production credentials is not.

## Fix

Remove `.env` from `NOISE_DIRS`. Sibling entries `.env.local`, `.env.production`, `.env.example` were never matched (the filter uses exact equality), so behavior for those is unchanged — they were already visible. This change only restores `.env` itself.

## Test

Adds `test_compact_preserves_dotenv` covering `.env`, `.env.local`, and `.env.example` to lock in the expected behavior.

```
running 19 tests
test result: ok. 19 passed; 0 failed
```

## Scope

Intentionally narrow. Does not touch:
- `env`, `.venv`, `venv` (Python virtualenv directories — those are regenerable)
- The broader question of whether *any* dotfile filtering belongs in `ls` (see #810)
- `RTK_DISABLED` / `--raw` discussions (#1167, #1171)

Just `.env`. One line in constants, one regression test.

### DIFF
--- a/src/cmds/system/constants.rs
+++ b/src/cmds/system/constants.rs
@@ -15,7 +15,6 @@ pub const NOISE_DIRS: &[&str] = &[
     ".venv",
     "venv",
     "env",
-    ".env",
     "coverage",
     ".nyc_output",
     ".DS_Store",
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -274,6 +274,24 @@ mod tests {
     }
 
     #[test]
+    fn test_compact_preserves_dotenv() {
+        // .env files MUST be visible by default. Hiding them silently caused
+        // agents to assume the project was unconfigured and overwrite the file
+        // with template defaults — destroying production secrets.
+        // See discussion: r/ClaudeCode "Token Optimizers...Silently Dangerous".
+        let input = "total 8\n\
+                     -rw-r--r--  1 user  staff  64 Jan  1 12:00 .env\n\
+                     -rw-r--r--  1 user  staff  64 Jan  1 12:00 .env.local\n\
+                     -rw-r--r--  1 user  staff  64 Jan  1 12:00 .env.example\n\
+                     -rw-r--r--  1 user  staff  100 Jan  1 12:00 main.rs\n";
+        let (entries, _summary) = compact_ls(input, false);
+        assert!(entries.contains(".env"), "default ls must show .env");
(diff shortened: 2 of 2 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Removes `.env` from the `NOISE_DIRS` list so that `rtk ls` no longer silently hides `.env` files in its default output. This prevents agents from mistakenly concluding a project is unconfigured and overwriting real `.env` files containing production secrets. A regression test is added to lock in the corrected behavior.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

---

KuSh:
Thanks for raising this, the reasoning in the description is exactly right. .env was taken out of NOISE_DIRS by #1895 (merged 2026-05-19), which also left a note on the neighbouring entry saying that .env is deliberately not filtered because agents must see it, so rtk ls and rtk tree no longer drop it. Nothing else in the branch differs from develop. One gap is left: no test pins that behaviour, and the test here cannot be carried over because compact_ls now takes a show_all and a show_long argument, so a small PR adding a regression test on the current signature would be welcome. Please comment or reopen if you still see .env disappear on a recent build.

PR #1513, opened 2026-04-25
fix(hook): preserve existing Copilot instructions

### LINKED ISSUE #1512: `rtk init -g --copilot` overwrites `copilot-instructions.md`
Instead of appending or preprending, running `rtk init -g --copilot` simply replaces existing `copilot-instructions.md` with its own instructions. Lucky if `copilot-instructions.md` had already been commited and this change can be easily spotted and reverted. Goofy move otherwise.

Image

season179:
## Summary

- Preserve existing `.github/copilot-instructions.md` content when installing Copilot support.
- Add an RTK instruction block with `<!-- rtk-instructions v2 -->` markers so future installs update idempotently.
- Migrate legacy unmarked RTK Copilot instructions without duplicating the RTK section.

Fixes #1512

## Root Cause

`run_copilot()` used `write_if_changed()` for `.github/copilot-instructions.md`. When the file already existed with user-authored Copilot rules, any content mismatch caused RTK to atomically replace the
whole file with RTK instructions.

## Test plan

- [x] `rtk cargo +1.92.0 test patch_copilot_instructions`
- [x] `rtk cargo fmt --all -- --check`
- [x] `rtk git diff --check`
- [x] Manual temp-project check: running `init --copilot` twice preserved existing Copilot rules and produced one RTK marker block

## Notes

Full `rtk cargo +1.92.0 test` was attempted and reached the suite, but failed in unrelated existing areas: curl truncation expectations, tracking DB creation in this sandbox, and a global uninstall test
touching `/Users/season/.config/opencode/plugins/rtk.ts`. The new Copilot tests passed.


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

This PR fixes the Copilot instructions installation to preserve existing user-authored content in `.github/copilot-instructions.md` instead of overwriting it. It introduces HTML comment markers (`<!-- rtk-instructions v2 -->`) for idempotent upserts and includes migration logic to strip legacy unmarked RTK instruction blocks.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1512

---

kiljacken:
Drive by comment, but I think the more idiomatic way to manage this would be to create a .github/instructions/rtk.instruction.md.

Avoids messing with the users existing instructions, and reduces the risk of clobbering the RTK instructions if the user runs /init to regenerate their copilot-instructions.

season179:
Thanks @kiljacken, appreciate the drive-by — and you're right, .github/instructions/rtk.instructions.md is the more idiomatic home for this. It avoids touching the user's main copilot-instructions.md entirely, survives a Copilot /init regen, and uninstall becomes a single rm. Cleaner all around.

That said, I'd still like to merge this PR first as the immediate fix for #1512 — users hitting the clobber bug today need the v2 marker path, and the migration logic added here is what we'd reuse to clean up marker blocks when we move to the separate-file approach in a follow-up "v3".

One thing I'd love your input on before committing to the move: do you know how well .github/instructions/*.instructions.md is supported outside VS Code Copilot (JetBrains, Visual Studio, Copilot CLI)? If it's broadly supported, v3 is a no-brainer. If it's VS Code-only, we may want to keep copilot-instructions.md as a fallback.

Either way, happy to open a follow-up issue and update accordingly. Thanks again for the suggestion!

xxradzip:
@season179 @kiljacken placing the instructions in `.github/instructions/rtk.instructions.md` does **not** work in Copilot CLI 1.0.37. It still can only see this on a fresh start:
```
Custom Instructions

3/3 enabled · Changes apply to the current session only.

User (1/1)
❯ [✓] Home copilot-instructions.md (/Users/cut_it_out_gdpr_cut_n_burn/.copilot/copilot-instructions.md)

Repository (2/2)
  [✓] .github/copilot-instructions.md (.github/copilot-instructions.md)
  [✓] AGENTS.md (AGENTS.md)
```

kiljacken:
@xxradzip Might need to add markdown frontmatter with files it should apply to:
```markdown
---
applyTo: "**"
---
Rest of the instruction goes here...
```

KuSh:
Thanks for this. Preserving user content in .github/copilot-instructions.md landed on develop through #1976 on 2026-05-20, and commit 35a540c3 then pulled the read, upsert and atomic write into a shared write_rtk_block helper that every agent target uses, with the same Added, Updated, Unchanged and Malformed handling, so patch_copilot_instructions here has been replaced rather than merged. One case is still missing: a file written by a pre-marker rtk init keeps its old unmarked RTK text alongside the freshly appended block, because upsert_rtk_block does not strip a legacy body. That affects every target, not only Copilot, so it is worth an issue or a small PR against the shared helper. If you think part of this was missed, please comment or reopen.

PR #1528, opened 2026-04-25
fix(grep): support -v invert-match (#1477)

### LINKED ISSUE #1477: rtk grep "-v" does not invert the search
Sometimes an LLM wants to invert the search and will use the `grep -v` command. RTK doesn't support this because the `-v` argument controls verbosity

# Steps to reproduce

```
rtk grep -v foo /etc/hosts
```

# Expected result

`rtk grep` should return all lines in the file which do not contain the string "foo"

# Actual result

`rtk grep` returns all lines which **do** contain the string "foo"

ousamabenyounes:
## Problem

\`rtk grep -v foo /etc/hosts\` returned matching lines instead of inverting the search. The \`-v\` flag was silently dropped.

## Root cause

The global \`--verbose\` count flag is registered with \`global = true\` and short \`-v\` in \`Cli\`. clap consumes globals before subcommand-specific positionals, so \`-v\` was eaten by verbose and never reached the \`Grep\` handler. \`extra_args\` ended up empty, so \`rg\` got no \`-v\`/\`--invert-match\` and returned matching lines.

A second latent issue: the grep fallback path (when \`rg\` is unavailable) was passing only \`-rn pattern path\` and discarding \`extra_args\` entirely. That would have broken the same flag for users without ripgrep installed.

## Fix

1. **\`preprocess_argv\`** runs before \`Cli::try_parse_from\` and rewrites a literal \`-v\` token into \`--invert-match\` only when the active subcommand is \`grep\`. Every other subcommand keeps \`-v\` as the verbose short alias.

2. **\`Commands::Grep\`** gains an explicit \`--invert-match\` boolean field. When set, the handler prepends \`--invert-match\` to \`extra_args\` so the flag reaches both the rg primary path and the grep fallback path.

3. **\`grep_cmd::run\`** fallback now forwards \`extra_args\` to GNU grep instead of dropping them (skipping \`-r\`/\`--recursive\` since \`grep -rn\` already implies recursion).

## Test plan

- [x] \`test_preprocess_grep_v_becomes_invert_match\` — argv rewrite happens.
- [x] \`test_preprocess_only_rewrites_for_grep_subcommand\` — \`rtk git -v status\` keeps \`-v\` as verbose.
- [x] \`test_grep_invert_match_flag_parses\` — clap accepts \`--invert-match\`.
- [x] \`test_grep_dash_v_full_pipeline_inverts_search\` — full argv → preprocess → clap pipeline lands on \`Grep::invert_match=true\` with \`cli.verbose==0\`.
- [x] \`cargo test --all\` → 1685 passed, 0 failed.
- [x] Manually verified post-fix on \`/etc/hosts\`:
  - \`rtk grep -v localhost /etc/hosts\` → 13 lines (all non-localhost lines, including comments)
  - \`rtk grep localhost /etc/hosts\` → 2 lines (still inverts correctly when -v absent)
  - \`rtk grep --invert-match localhost /etc/hosts\` → 13 lines (long form also works)

Closes #1477

---

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -49,9 +49,18 @@ pub fn run(
 
     let result = exec_capture(&mut rg_cmd)
         .or_else(|_| {
+            // Fallback when rg is unavailable: invoke GNU grep with the same
+            // user-supplied extra_args (e.g. --invert-match for #1477) so
+            // forwarded flags like -v still take effect.
             let mut grep_cmd = resolved_command("grep");
-            //When we fall back to grep,include all args, not just -rn.
-            grep_cmd.args(["-rn", pattern, path]).args(extra_args);
+            grep_cmd.arg("-rn");
+            for arg in extra_args {
+                if arg == "-r" || arg == "--recursive" {
+                    continue;
+                }
+                grep_cmd.arg(arg);
+            }
+            grep_cmd.args([pattern, path]);
             exec_capture(&mut grep_cmd)
         })
         .context("grep/rg failed")?;
--- a/src/main.rs
+++ b/src/main.rs
@@ -318,6 +318,11 @@ enum Commands {
         /// Show line numbers (always on, accepted for grep/rg compatibility)
         #[arg(short = 'n', long)]
         line_numbers: bool,
+        /// Invert match — select lines that do NOT match the pattern (grep/rg -v).
+        /// `-v` is reserved for global --verbose; `rtk` translates a literal `-v`
+        /// after the `grep` subcommand into this flag (issue #1477).
+        #[arg(long = "invert-match")]
+        invert_match: bool,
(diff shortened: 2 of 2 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Fixes a bug where `rtk grep -v` had the `-v` flag silently consumed by the global `--verbose` flag instead of being passed as `--invert-match` to the grep/rg backend. Adds a pre-clap argv preprocessor that rewrites `-v` to `--invert-match` only for the `grep` subcommand, adds an explicit `--invert-match` field to the Grep command, and fixes the GNU grep fallback to forward extra_args.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1477

---

PR #1541, opened 2026-04-26
fix(grep): return bare integer for -c/--count flag

### LINKED ISSUE #1452: bug(grep): `-c` count flag ignored — returns match summary + context instead of line count

CtriXin:
## Summary

`rtk grep -c <pattern> <file>` now outputs just the bare integer count, matching plain `grep -c` semantics.

**Root cause**: the `-c` flag was passed through to `rg` (which outputs `file:count` format), but the code split on `:` expecting `file:linenum:content`, misinterpreting the output. The full grouped match listing was returned instead of the count.

**Fix**: detect `-c`/`--count` in `extra_args`, route to dedicated `run_count()` path that sums all per-file counts from `rg -c` and prints only the integer.

Fixes: rtk-ai/rtk#1452

## Test Plan

- [x] `cargo test grep_cmd` — 11 tests pass
- [ ] Manual: `rtk grep -c foo /tmp/file` returns single integer
- [ ] Manual: `count=$(rtk grep -c pattern file)` captures parseable integer

## Notes

- Existing grep output (without `-c`) unchanged
- `rg` is used when available, `grep` as fallback (unchanged)

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -8,6 +8,67 @@ use anyhow::{Context, Result};
 use regex::Regex;
 use std::collections::HashMap;
 
+fn has_count_flag(extra_args: &[String]) -> bool {
+    extra_args.iter().any(|a| a == "-c" || a == "--count")
+}
+
+struct CountResult {
+    exit_code: i32,
+    count: i32,
+    raw_output: String,
+}
+
+fn run_count(
+    pattern: &str,
+    path: &str,
+    file_type: Option<&str>,
+    extra_args: &[String],
+    verbose: u8,
+) -> Result<CountResult> {
+    if verbose > 0 {
+        eprintln!("grep -c: '{}' in {}", pattern, path);
+    }
+
+    let rg_pattern = pattern.replace(r"\|", "|");
+    let mut rg_cmd = resolved_command("rg");
+    rg_cmd.args(["-c", "--no-heading", &rg_pattern, path]);
+
+    if let Some(ft) = file_type {
+        rg_cmd.arg("--type").arg(ft);
+    }
+
+    for arg in extra_args {
+        if arg == "-r" || arg == "--recursive" || arg == "-c" || arg == "--count" {
+            continue;
+        }
+        rg_cmd.arg(arg);
+    }
+
+    let result = exec_capture(&mut rg_cmd).or_else(|_| {
+        let mut grep_cmd = resolved_command("grep");
+        grep_cmd.args(["-c", "-r", pattern, path]);
+        exec_capture(&mut grep_cmd)
+    }).context("grep -c failed")?;
+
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes the `rtk grep -c` flag to output a bare integer count instead of misinterpreting ripgrep's `file:count` format. A dedicated `run_count()` code path detects the `-c`/`--count` flag, sums per-file counts from `rg -c` output, and prints only the total integer, matching standard `grep -c` semantics.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1452

---

KuSh:
Thanks for digging into this one. The -c handling landed a different way in #1456 (merged 2026-05-01), which routes count and other format flags straight through to the invoked search engine instead of reparsing its output, with follow-ups in #2333 and #2641. On current develop, rtk grep -c foo file.txt prints just the count and rtk grep -c -r foo . prints native per-file counts, so the behaviour your patch targeted is in place. The bare-integer case you described is covered there, and grep_cmd.rs itself has since been folded into search.rs. If you spot a count case that still misbehaves on develop, please comment here or reopen.

PR #1547, opened 2026-04-26
fix(stream): passthrough watch commands

### LINKED ISSUE #1545: Streaming/watch-mode commands still buffer or suppress stdout outside proxy mode
## Summary

`rtk proxy` streaming was fixed by the `src/core/stream.rs` work, but the fix is not global. Several filtered or nominally passthrough RTK command routes still do not emit stdout for long-running/watch/dev commands until the child exits, or until the parent is killed.

This is a different failure mode than the already-closed proxy-only issue: `rtk proxy <cmd>` streams, while `rtk <recognized-filter-route> <watch/dev command>` can still hang silently.

## Verified behavior on current master

I tested current `master` (`rtk 0.37.2` source tree) with fake long-running executables placed first in `PATH`. Each fake tool prints one line immediately and then sleeps forever. I measured whether RTK emitted the first stdout line within 2 seconds.

### Streams correctly

```text
STREAMS  rtk proxy npm run dev
STREAMS  rtk pnpm run dev
STREAMS  rtk pnpm dev
STREAMS  rtk docker logs -f web
STREAMS  rtk docker compose logs -f web
STREAMS  rtk kubectl logs -f web
```

### Still broken before this fix

```text
NO_STDOUT_WITHIN_2S  rtk npm run dev
NO_STDOUT_WITHIN_2S  rtk npm start
NO_STDOUT_WITHIN_2S  rtk prettier --watch .
NO_STDOUT_WITHIN_2S  rtk playwright codegen
NO_STDOUT_WITHIN_2S  rtk gh pr checks 123 --watch
NO_STDOUT_WITHIN_2S  rtk dotnet watch run
NO_STDOUT_WITHIN_2S  rtk go run .
```

The user-visible effect is that an agent or human sees no progress and may kill/retry the command even though the child process is alive and already wrote output.

## Minimal real-world repro

```bash
mkdir /tmp/rtk-stream-repro
cd /tmp/rtk-stream-repro
cat > package.json <<'JSON'
{
  "scripts": {
    "dev": "node -e \"console.log('dev-start'); setInterval(() => {}, 1000);\""
  }
}
JSON

npm run dev          # prints immediately
rtk proxy npm run dev # prints immediately
rtk npm run dev       # no stdout before forced termination
```

## Root cause

The merged streaming infrastructure is present, but not all command paths use it.

Important paths:

- `src/core/runner.rs`: `RunMode::Passthrough` uses `FilterMode::Passthrough`, which inherits stdout/stderr and streams.
- `src/core/runner.rs`: `RunMode::Filtered` calls `stream::run_streaming(..., FilterMode::CaptureOnly)`, which captures stdout/stderr until EOF before applying a filter.
- `src/core/stream.rs`: `exec_capture()` still uses `cmd.output()`, which waits for the child process to exit.
- Some functions named `run_passthrough` were not actually streaming passthrough. For example, unsupported `go` and `dotnet` routes captured output and printed it only after completion.

So `#222` was fixed for `rtk proxy`, but filtered routes and some passthrough-named routes remained EOF-buffered.

## Related issues and PRs

- #222: fixed the narrow `rtk proxy` long-running streaming bug. This issue is not the same because `rtk proxy npm run dev` streams while `rtk npm run dev` does not.
- #361: broad hook/streaming issue. It correctly identified `.output()` buffering, but the merged stream infrastructure did not eliminate all buffered filter paths.
- #956: merged `src/core/stream.rs`, `FilterMode`, and `rtk proxy` streaming. This introduced the needed primitives, but many routes still use `CaptureOnly` or `exec_capture()`.
- #1267: related Docker logs/progress problem. Container follow paths now tend to stream via passthrough, but filtered log summarization can still hide progress.
- #963 and PR #1531: `git push` timeout has the same root cause pattern (`Command::output()` hiding progress/prompts), but PR #1531 only fixes `git push`.
- #1156 and PR #1171: `--raw` / `--no-filter` is related, but users should not need to know to opt out manually for commands that are inherently streaming.

## Suggested fix

RTK should detect inherently streaming invocations before choosing a buffered filtering path and run them with inherited stdout/stderr instead.

Practical detection heuristics:

- `--watch`, `--watchAll`, `-w`
- `--follow`, `-f` for log/follow tools such as Docker, kubectl, tail, journalctl
- streaming/dev subcommands such as `watch`, `dev`, `serve`, `start`
- known interactive/long-running commands such as `playwright codegen`, `playwright show-report`, `go run`, `dotnet watch`
- inherently streaming binaries such as `tail`, `nodemon`, `watchman`

Also, routes called `run_passthrough` should use actual passthrough execution (`Stdio::inherit()` / `FilterMode::Passthrough`), not `Command::output()`.

## Acceptance criteria

- `rtk npm run dev` emits output immediately and does not wait for EOF.
- `rtk npm start` emits output immediately and does not wait for EOF.
- `rtk prettier --watch .` emits output immediately.
- `rtk playwright codegen` emits output immediately.
- `rtk gh pr checks <pr> --watch` emits output immediately.
- `rtk dotnet watch run` emits output immediately.
- `rtk go run .` emits output immediately.
- Finite commands such as `rtk npm run build` and `rtk gh pr checks <pr>` remain filterable.

nkootstra:
## Summary
Fixes watch/dev-style commands that previously looked like they were streaming but still buffered through RTK filtering paths until the child process exited.

This routes known long-running invocations through passthrough execution so stdout/stderr stay attached to the terminal, while still keeping normal filtered commands on the existing buffered/streaming pipeline. It also updates unsupported `go` and `dotnet` passthrough paths to actually stream instead of capturing via `.output()`.

## Changes
- Detect watch/dev/follow/codegen-style invocations centrally before applying filtered execution.
- Preserve `gh pr checks --watch` detection when extra args are present.
- Stream interactive Playwright, unsupported Go, and unsupported Dotnet passthrough routes instead of buffering them.
- Include small clippy cleanups required by the current develop branch.

## Verification
- `smll cargo fmt --all`
- `smll cargo clippy --all-targets`
- `smll cargo test --all` (`1691p 0f`)

Closes rtk-ai/rtk#1545


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Fixes watch/dev-style commands (e.g. --watch, dev, serve, codegen) that were being buffered through RTK's filtered execution pipeline instead of streaming directly to the terminal. Adds centralized detection of long-running/streaming invocations in the runner module and routes them through passthrough execution, while also converting unsupported Go and Dotnet passthrough paths from buffered capture to proper streaming.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1545

---

pszymkowiak:
Thanks for the very thorough fix — the diagnosis in #1545 and the centralized `is_streaming_invocation()` helper are exactly the right shape.

**Looks great:**
- Centralized detection in `runner.rs` before the filtered branch, with unit tests for watch / follow / dev / finite cases ✅
- `--follow` / `-f` correctly gated on `docker | kubectl | tail | journalctl` to avoid false positives like `git log -f` ✅
- `go run` and unsupported `dotnet` routes now actually stream instead of capturing via `.output()` ✅
- `gh pr checks --watch` detection preserved when extra args are present ✅

After a deeper pass (with help from a code-review pass on my side), here's what I'd want addressed before merge.

---

### 🔴 Blocking — `-w` is treated as a universal watch flag

```rust
if args.iter().any(|arg| matches!(*arg, "--watch" | "--watchAll" | "-w")) {
    return true;
}
```

`-w` is overloaded outside the JS test ecosystem:
- `--word-regexp` in `grep` / `rg`
- `--width` in `cargo test` and `wc`
- not actually documented for Jest (Jest uses `--watch`); the existing test `is_streaming_invocation("jest", "run -w")` asserts a behavior that's already shaky for Jest itself

The current `grep_cmd.rs` happens to bypass `runner::run_filtered`, so the live blast radius is small today — but any future filter route that accepts `-w` will silently switch to passthrough and skip filtering. Suggest scoping `-w` to a known allowlist (`jest`, `vitest`, `webpack`, `nodemon`) the same way `--follow` is scoped to log tools.

### 🟡 Test bug to fix while you're in there

```rust
assert!(!is_streaming_invocation("go test", "./..."));
```

This passes only because `tool_name.split_whitespace().next()` drops `"test"` from the tool name. Real callers in `main.rs` pass `tool_name="go"` and `args_display="test ./..."`. Recommend rewriting as `("go", "test ./...")` so the test reflects the actual call shape.

### 🟡 Coverage scope is undocumented

The check lives only inside the `RunMode::Filtered` arm — `RunMode::Streamed` ignores it entirely. A future contributor adding a filter via `run_streamed` won't get watch-mode protection. A short doc comment on `is_streaming_invocation` describing (a) which `RunMode` it guards and (b) the contract on `args_display` (structured subcommand tokens only, never user-controlled prose like PR titles) would prevent surprises.

### 🟡 Missing test cases

While we're here:
- `is_streaming_invocation("cargo", "test -w")` should be `false` — currently returns `true`
- `pnpm run dev` (pnpm has its own filter module, worth confirming the same path is exercised)

---

### My earlier 3 concerns — resolved on second look

1. **Bundled clippy cleanups.** The `develop` branch runs stricter clippy (`-D warnings` + dead-code lint), which actually requires the signal cast, `collapse_if`, `match → map`, and the six `#[allow(dead_code)]` annotations to compile. They belong in this PR — only the PR description should call this out explicitly so future reviewers don't burn time on it.

2. **`stream.rs` RAW_CAP boundary.** Recomputed on all 5 sites: old condition allows the buffer to reach exactly `RAW_CAP`; new condition caps at `RAW_CAP - 1`. One byte more conservative against a multi-MiB cap — invisible in practice, no regression. Worth a one-line commit message clarifying the intent.

3. **`Cargo.lock` `0.36.0 → 0.34.3`.** This is the self-referential `rtk` package version from your fork's `Cargo.toml`. `master` is at `0.37.2` and `develop` is ahead of you — this **will conflict on merge**. A rebase onto current `develop` and a fresh `cargo build` should regenerate the lockfile cleanly.

---

**Edge-case note** (no action needed, just visible): the bare-token list `watch | dev | serve | start | codegen | show-report` applies to all tools positionally on `args_display`. As long as `args_display` is composed of structured subcommand tokens (which all current callers respect), this is safe. Documenting that contract on the helper is enough.

Once the `-w` scoping is tightened and the lockfile is rebased, happy to approve. Thanks again — this is a high-quality fix.

PR #1548, opened 2026-04-26
fix(git): clarify stash push no-op output

### LINKED ISSUE #1535: rtk git stash {push,pop,apply,drop} returns exit 0 on failure and produces ambiguous success output
## Summary

```bash
$ cd /any/git/repo
git stash Some("push")
FAILED: git stash push
Did you forget to 'git add'?
$ echo $?
0
```

## Repro 2 — "ok" on no-op

```bash
git stash Some("push")
ok stash push
$ echo $?
0
```

## Why this matters

## Source location

```rust

    println!("{}", msg);
    msg
} else { ... }
```

## Suggested fix

## Environment

- git 2.x (system)

MestreY0d4-Uninter:
## Summary

- make explicit `rtk git stash push` report `ok (nothing to stash)` when git reports no local changes
- keep existing non-zero exit propagation for failing stash subcommands unchanged
- add a regression unit test for the no-op stash push message

Closes #1535

## Test plan

- [x] `cargo test test_stash_push_noop_success_message -- --nocapture`
- [x] `cargo clippy --all-targets -- -D unsafe_code`
- [x] `cargo test --all`
- [x] `bash scripts/check-test-presence.sh origin/develop`
- [x] Manual testing: in a clean temp git repo, `rtk git stash push` now returns rc=0 with `ok (nothing to stash)`; `rtk git stash pop/apply/drop` still return rc=1 with `FAILED: git stash <sub>`.

Note: on the current `develop` branch, the non-zero exit status handling for failed `pop/apply/drop/push` was already present. This PR fixes the remaining ambiguous success output for explicit `stash push` when there is nothing to stash.

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1463,7 +1463,7 @@ fn run_stash(
             let combined = result.combined();
 
             let msg = if result.success() {
-                let msg = format!("ok stash {}", sub);
+                let msg = stash_success_message(sub, &result.stdout);
                 println!("{}", msg);
                 msg
             } else {
@@ -1555,6 +1555,14 @@ fn run_stash(
     }
 
     Ok(0)
+}
+
+fn stash_success_message(sub: &str, stdout: &str) -> String {
+    if sub == "push" && stdout.contains("No local changes") {
+        "ok (nothing to stash)".to_string()
+    } else {
+        format!("ok stash {}", sub)
+    }
 }
 
 fn filter_stash_list(output: &str) -> String {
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Clarifies the output of `rtk git stash push` when there are no local changes to stash, changing the ambiguous success message to an explicit 'ok (nothing to stash)'. Extracts the success message logic into a dedicated helper function and adds a unit test for the no-op case.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1535

---

KuSh:
Hi @MestreY0d4-Uninter thanks for your PR.

I've pushed https://github.com/rtk-ai/rtk/pull/1586 which is more complete and handle all corner cases, at least I hope.
I'm closing yours in favor of that one. Don't hesitate to review and test it to see if that works for your case

Issue #1436, opened 2026-04-21
The rtk grep is not grep-compatible


Sediug:
# rtk grep is not grep-compatible here: extra matches and malformed output for a single-file `-n` search

## Description

`rtk grep` is not preserving `grep` compatibility in this case.

I can reproduce this with the same pattern and the same single file:

```bash
grep -n "private function delete\|try\|catch\|finally\|delete()" externalImportShell.class.php
rtk grep -n "private function delete\|try\|catch\|finally\|delete()" externalImportShell.class.php
```

Plain `grep` returns matches from a single file with the expected line numbers.

`rtk grep` instead reports `23 matches in 2F`, broadens the match set, and also mangles one valid match into a synthetic file bucket.

This seems to break the project's correctness/transparency goal: the output is no longer a valid shortened subset of the original `grep` output.

## Steps to reproduce

1. Run:

```bash
grep -n "private function delete\|try\|catch\|finally\|delete()" /path/to/externalImportShell.class.php
```

2. Run:

```bash
rtk grep -n "private function delete\|try\|catch\|finally\|delete()" /path/to/externalImportShell.class.php
```

3. Compare the outputs.

## Expected behavior

- `rtk grep` should preserve the same match semantics as plain `grep` for the same command.
- Output can be shortened, but it should remain a valid/useful subset of the original `grep` output.
- The match on line `81` should stay attached to the real file and line number.
- A single-file search should not be reported as `2F`.
- Not removing part of the code, this has been removed from line 81 -> `$this->queueProcessModel = ClassRegistry::`

## Actual behavior

- `rtk grep` reports `23 matches in 2F` for a single-file search.
- It creates a synthetic block:

```text
[file] 81 (1):
     0: :init('Collections.QueueProcess');
```

- It broadens the results and includes extra `delete`-related lines that plain `grep` does not return for the same pattern.
- A valid matching line containing `ClassRegistry::init(...)` is rendered incorrectly.

## Relevant output

Expected (`grep`):

```shell
10:use Cbi\Core\Lib\Utility\ClassRegistry;
46:    /** Seconds to wait before retrying to save a file */
49:    /** SQL duplicate entry error code */
81:        $this->queueProcessModel = ClassRegistry::init('Collections.QueueProcess');
108:        try {
124:        } catch (\Error|\Exception $e) {
166:            $this->delete();
442:    private function delete()
487:        try {
499:        } catch (\Exception $e) {
658:        try {
660:        } catch (\Exception $e) {
[...]

PR #1554, opened 2026-04-27
fix(grep): parse single-file output containing colons

F0rty-Tw0:
## Summary

- Fix `rtk grep` synthesizing a phantom `[file] N` bucket when a single-file match contained `:` (e.g. `ClassRegistry::init(...)`); ripgrep emits `<line>:<content>` for single-file searches and the parser used `splitn(3, ':')` length to disambiguate, which broke whenever content also had `:`.
- Force `--with-filename` on rg (`-H` on the grep fallback) so output is always `file:line:content`, and replace the inline parser with a regex-based `parse_match_line` that also handles paths with embedded `:` (Windows drive letters like `C:\...`).
- Scoped to the parser/rendering portion of #1436. The BRE→regex match-set behavior (e.g. `delete()` matching every `delete`) is a separate concern and ships in a follow-up PR (now also pushed).

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --bin rtk` — 1692 passed, 0 failed
- [x] 5 new unit tests on `parse_match_line`: simple, content-with-`::` (#1436 reproducer), Windows drive path, malformed line, empty content
- [x] Manual end-to-end: synthetic PHP file with the issue's exact pattern + a `ClassRegistry::init('...')` line; rtk output now matches plain `grep -nH` line-for-line, no phantom bucket
- [x] CI green on `develop`

Closes the parser/rendering portion of #1436.

### Related work

[PR #1273](https://github.com/rtk-ai/rtk/pull/1273) by @siubing05 (open since 2026-04-13) addresses the same parser issue alongside a pytest summary fix. Their approach uses `splitn(3, ':')` with a 2-format fallback that tolerates missing `-H`. This PR uses a regex parser (`^(.+):(\d+):(.*)$`) that additionally handles paths with embedded `:` (Windows drive letters). Either approach fixes the Linux/macOS reproducer in #1436 — happy to defer or fold a Windows-path test into #1273 if maintainers prefer that path.

### Notes for reviewers

- **`-H` is load-bearing, not noise.** Without it, GNU grep's single-file output drops the filename, the new regex parser can't extract three fields, and every match gets silently dropped (rendering shows `N matches in 0 files`). It's an input flag for grep, rtk re-formats its own output, so user-facing token cost is zero. Verified empirically (with-`-H` vs without-`-H` build).
- **Greedy regex on path is intentional.** `^(.+):(\d+):(.*)$` picks the LAST `:digits:` boundary, which correctly handles `C:\path\file.rs:42:content`. A pathological filename containing `:digits:` would mis-split; that edge is documented in the parser's doc comment.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -34,14 +34,14 @@ pub fn run(
     // false negatives that make AI agents draw wrong conclusions.
     // Using --no-ignore-vcs (not --no-ignore) so .ignore/.rgignore are still respected.
     //
-    // --with-filename: force rg to always include the filename, even for
-    // single-file invocations. Without it, rg emits `<line>:<content>` for a
-    // single file, and the parser cannot tell `<line>:<content>` from
-    // `<file>:<line>:<content>` when content itself contains ':' (issue #1436).
+    // -H: force rg to always include the filename, even for single-file searches.
+    // --null: separate file from line:content with a NUL byte so the parser is
+    // unambiguous even when the filename or content contains `:digits:`
+    // patterns (issue #1436).
     rg_cmd.args([
-        "-n",
+        "-nH",
         "--no-heading",
-        "--with-filename",
+        "--null",
         "--no-ignore-vcs",
         &rg_pattern,
         path,
@@ -62,9 +62,10 @@ pub fn run(
     let result = exec_capture(&mut rg_cmd)
         .or_else(|_| {
             let mut grep_cmd = resolved_command("grep");
-            // -H: always emit the filename so the parser sees a uniform
-            // file:line:content shape (parity with rg's --with-filename).
-            grep_cmd.args(["-rnH", pattern, path]).args(extra_args);
+            // -H: always emit the filename; -Z: NUL-separate filename from
+            // line:content so the parser can disambiguate even for filenames
+            // or content containing `:digits:` (parity with rg's --null).
+            grep_cmd.args(["-rnHZ", pattern, path]).args(extra_args);
             exec_capture(&mut grep_cmd)
         })
         .context("grep/rg failed")?;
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes the `rtk grep` parser that incorrectly created phantom file buckets when a single-file match line contained colons (e.g. `ClassRegistry::init(...)`). Forces `--with-filename` on rg and `-H` on grep so output is always `file:line:content`, and replaces the inline `splitn`-based parser with a greedy regex-based `parse_match_line` function that also handles Windows drive-letter paths.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1436

---

F0rty-Tw0:
Follow-up: the BRE-translation portion of #1436 is queued up in [F0rty-Tw0/rtk#1](https://github.com/F0rty-Tw0/rtk/pull/1) (stacked on this branch in my fork). Once this PR merges, I'll rebase that one onto upstream `develop` and reopen it here. Keeping the two changes separate so each is reviewable on its own:

- This PR (#1554): parser fix only - single-file output containing `:` no longer creates phantom `[file] N` buckets.
- Follow-up: dialect-aware routing - `delete()` matches literal `()` in BRE again instead of being treated as an empty regex group.

F0rty-Tw0:
Since PR #1273 is closed, i would like to continue with this one. Summoning @KuSh, if you can take a look on it whenever you got a free moment. Not urgent tho.

siubing05:
Sounds good to me — thanks for continuing this.

Just noting that #1273 also included a pytest quiet-summary fix. If maintainers want that part separately, I’m happy to split it into a smaller follow-up PR.

F0rty-Tw0:
@siubing05 that would be amazing.

src/cmds/system/grep_cmd.rs
```diff
@@ -33,7 +33,19 @@ pub fn run(
...
+    // single file, and the parser cannot tell `<line>:<content>` from
+    // `<file>:<line>:<content>` when content itself contains ':' (issue #1436).
+    rg_cmd.args([
+        "-n",
+        "--no-heading",
+        "--with-filename",
+        "--no-ignore-vcs",
+        &rg_pattern,
+        path,
+    ]);
```
KuSh: nitpick:
```suggestion
    rg_cmd.args(["-nH", "--no-heading", "--no-ignore-vcs", &rg_pattern, path]);
```

src/cmds/system/grep_cmd.rs
```diff
@@ -147,6 +152,29 @@ pub fn run(
...
+///   - content with `::` (e.g. `ClassRegistry::init(...)`) does not split into
+///     a phantom file bucket (issue #1436);
+///   - paths with embedded `:` (Windows drive letters like `C:\foo\file.php`)
+///     parse correctly — the LAST `:digits:` before content is the boundary.
+///
+/// Returns `None` for lines that do not match the expected shape (e.g. rg
+/// `-A`/`-B` context lines that use `-` as separator).
+fn parse_match_line(line: &str) -> Option<(String, usize, &str)> {
+    lazy_static::lazy_static! {
+        static ref MATCH_LINE_RE: Regex = Regex::new(r"^(.+):(\d+):(.*)$").unwrap();
```
KuSh: Unfortunately this won't work with badly named files like in that example:

```sh
$ echo "xxx" > "badly_named:52:file.txt"
$ grep -rnH xxx . 2>/dev/null
./badly_named:52:file.txt:1:xxx
./src/cmds/git/git.rs:503:    // (BREAKING CHANGE, Closes #xxx, design notes)
./scripts/benchmark/cloud-init.yaml:282:    printf "DATABASE_URL=postgres://user:pass@localhost:5432/db\nAPI_KEY=sk-1234567890abcdef\nSECRET_TOKEN=ghp_xxxx\nNODE_ENV=production\nPORT=3000\n" > /tmp/.env
```

You need to use the `-Z` (`-0` on rg) arg to have a file sep that will always be valid.

With that you can use a regex like this one which should always work, whatever the filename:
```suggestion
        static ref MATCH_LINE_RE: Regex = Regex::new(r"^[^\x00]+\x00(\d+):(.*)$").unwrap();
```

You'll have to add a test for that too

src/cmds/system/grep_cmd.rs
```diff
@@ -147,6 +152,29 @@ pub fn run(
...
+    }
+    let caps = MATCH_LINE_RE.captures(line)?;
+    let file = caps.get(1)?.as_str().to_string();
+    let line_num: usize = caps.get(2)?.as_str().parse().ok()?;
+    let content_start = caps.get(3)?.start();
+    Some((file, line_num, &line[content_start..]))
```
KuSh:
```suggestion
    MATCH_LINE_RE.captures(line).and_then(|caps| {
        let (_, [file, line, content]) = caps.extract();
        let line_num: usize = line.parse().ok()?;
        Some((file.to_string(), line_num, content))
    })
```

KuSh (changes requested):

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Thanks @KuSh, will take a look.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Done, went with `-nH`. Left a note above so nobody strips it later.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Yeah good one. Fixed, also noticed the same thing could trip on `:42:` style stuff in content, so I added tests for both.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Done, folded in with the regex change above.

KuSh (approved):

PR #1558, opened 2026-04-27
fix(json): use char boundary when truncating long string values

swithek:
## Summary

- Fix UTF-8 panic in `rtk json` when a multi-byte char crosses byte 77 of a string value (e.g. JSON containing Japanese, Cyrillic, accented Latin, or emoji).
- `compact_json` now rounds the truncation index down to the nearest valid char boundary with `str::floor_char_boundary`, matching the existing pattern in `pipe_cmd::auto_detect_filter`.
- Add regression test covering the multi-byte boundary case.

## Test plan

  ```sh
  python3 -c 'import json; print(json.dumps({"key": "a"*76 + "日本語"*5}, ensure_ascii=False))' \
    | ./target/release/rtk json -
Before: panicked at `src/cmds/system/json_cmd.rs:109`: `end byte index 77 is not a char boundary; it is inside '日'`
After: pretty-printed compact JSON with the value truncated to "aaaa…aaaa...".

### DIFF
--- a/src/cmds/system/json_cmd.rs
+++ b/src/cmds/system/json_cmd.rs
@@ -330,11 +330,8 @@ mod tests {
         assert!(schema.contains("(3)"));
     }
 
-    #[test]
-    fn test_compact_long_multibyte_string_does_not_panic() {
-        let payload = "日本語テスト".repeat(85); // well above the 80-byte threshold
+    fn assert_value_truncated(payload: &str) {
         let json = format!(r#"{{"key": "{}"}}"#, payload);
-
         let output = filter_json_compact(&json, 5)
             .expect("filter_json_compact must not error on valid JSON");
 
@@ -343,12 +340,25 @@ mod tests {
             output.contains("..."),
             "long string should be truncated, got: {output}"
         );
-        for line in output.lines() {
-            assert!(
-                line.chars().count() <= 80,
-                "output line exceeds 80 chars ({} chars): {line}",
-                line.chars().count()
-            );
-        }
+
+        let value = output
+            .split('"')
+            .nth(1)
+            .expect("output should contain a quoted string value");
+        assert!(
+            value.len() <= 80,
+            "truncated value is {} bytes: {value}",
+            value.len()
+        );
+    }
+
+    #[test]
+    fn test_compact_truncates_pure_multibyte_string() {
+        assert_value_truncated(&"日本語テスト".repeat(85));
+    }
+
+    #[test]
+    fn test_compact_truncates_mixed_ascii_multibyte_string() {
+        assert_value_truncated(&("a".repeat(76) + &"日本語".repeat(5)));
     }
 }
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes a UTF-8 panic in the JSON compact display when truncating long string values containing multi-byte characters. Uses `str::floor_char_boundary` to round the truncation index down to a valid char boundary instead of slicing at a fixed byte offset. Includes a regression test for the multi-byte boundary case.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

---

swithek:
hey @KuSh @aeppling I think this might be a somewhat urgent issue. I'm hitting this panic quite often with emoji/non-latin content and it's pretty much breaking my rtk workflows at the moment

src/cmds/system/json_cmd.rs
```diff
@@ -328,4 +329,19 @@ mod tests {
...
+    fn test_compact_long_multibyte_string_does_not_panic() {
+        let payload = "a".repeat(76) + &"日本語".repeat(5);
+        let json = format!(r#"{{"key": "{}"}}"#, payload);
+
+        let output = filter_json_compact(&json, 5)
+            .expect("filter_json_compact must not error on valid JSON");
+
+        assert!(output.contains("key"));
+        assert!(
+            output.contains("..."),
```
KuSh, line 340: Nitpick, at least check output lenght to be <=80 characters

If i'm not mistaken this test will truncate after the last `a` and before the first `日`, it would be better to test that it can truncate within a Japanese sentence or any other non latin1 utf8 string.

KuSh (changes requested):

src/cmds/system/json_cmd.rs
swithek: Thanks for the review @KuSh :) I've pushed an update, let me know if there's anything else that can be improved here.

src/cmds/system/json_cmd.rs
```diff
@@ -328,4 +329,26 @@ mod tests {
...
+            "long string should be truncated, got: {output}"
+        );
+        for line in output.lines() {
+            assert!(
+                line.chars().count() <= 80,
```
KuSh: This works by "chance" as you're counting the total line chars whereas the code is truncating the string value at 77 byte length and terminating it with "...". With your previous test `"a".repeat(76) + &"日本語".repeat(5)` this test would fail. You should check that the "key" value IS <= 80 byte (double quote excluded)

KuSh (changes requested):

src/cmds/system/json_cmd.rs
swithek: Good point 👍

KuSh (approved):

PR #1569, opened 2026-04-28
fix(ls): parse ls -la output on en_GB / en_IN locales (#1566)

### LINKED ISSUE #1566: rtk ls returns empty output for non-empty directories (100% information loss)
### Summary

`rtk ls <dir>` produces zero file names regardless of the directory's actual content. The verbose mode reports `Chars: 458 → 8 (99% reduction)` as if it were a successful compression, but the output is in fact unusable for any downstream consumer (LLM agent, script, human). All `ls` flags I tried (`-l`, `-la`, `--ultra-compact`) reproduce the issue.

### Reproduction

```bash
TMP=$(mktemp -d)
for n in alpha bravo charlie delta echo; do touch "$TMP/$n.dart"; done

rtk ls "$TMP/"                  # → empty
rtk ls -l "$TMP/"               # → empty
rtk ls --ultra-compact "$TMP/"  # → empty
rtk -v ls "$TMP/"               # → "Chars: 458 → 8 (99% reduction)" then empty

rtk proxy ls "$TMP/"            # → 5 lines, the actual file names
```

### Observed

`rtk ls` strips out every file name. Even with `-v`, the only thing surviving is the compression banner — the file list itself is gone.

### Expected

`rtk ls` should preserve at least the file/directory names. A reasonable budget could be: list the first N entries verbatim, then summarize the tail (e.g. `… and 12 more`). A 100% reduction on a 5-file directory is never the right answer.

### Why this matters for LLM agents

The whole point of RTK is to keep enough signal in the context for an agent to make decisions. `ls` output is one of the most decision-load-bearing surfaces:

- "does this file exist?"
- "what's in this directory?"
- "how many scripts are there?"

When `rtk ls` returns empty, an agent concludes the directory is empty and proceeds incorrectly. In our project (Phoenix/Alter, ~40 scripts) we hit this immediately when an agent tried to list `scripts/` and concluded it was empty.

### Environment

- `rtk 0.37.2`
- `Darwin 25.4.0 arm64` (macOS, Apple Silicon)
- Shell: `zsh`
- Installed via `brew`

### Suggested triage

1. Confirm the issue reproduces on Linux as well.
2. Check whether the regression is recent (bisect against earlier 0.3x releases).
3. Even if compression is desired, a hard floor — "always keep at least the first N names" — would prevent the worst failure mode.

Happy to test a fix or provide more context.

### LINKED ISSUE #1652: [BUG] rtk ls parsing relies on POSIX/C locale date ordering and can return false (empty) output
## Summary

`rtk ls` can return `(empty)` for a non-empty directory when the underlying `ls -la` output uses day-month date ordering instead of the POSIX/C locale month-day ordering.

This appears to be because the parser in `src/cmds/system/ls.rs` assumes dates in the form:

```text
Apr 29 10:42
```

but my normal shell locale produces:

```text
29 Apr 10:42
```

When that happens, RTK drops valid lines and emits `(empty)` instead of either:
- parsing successfully, or
- falling back to raw output

## Environment

- RTK version: `0.37.2`
- Install method: Homebrew
- Binary: `/opt/homebrew/bin/rtk`
- Shell: `zsh`
- Locale involved: `LANG=en_GB.UTF-8`

## Reproduction

In a non-empty repo:

```bash
ls
```

shows normal output, e.g.

```text
build.rs                   docs                       README_es.md               scripts
Cargo.lock                 Formula                    README_fr.md               SECURITY.md
Cargo.toml                 hooks                      README_ja.md               src
CHANGELOG.md               INSTALL.md                 README_ko.md               tests
CLAUDE.md                  install.sh                 README_zh.md
CONTRIBUTING.md            LICENSE                    README.md
DISCLAIMER.md              openclaw                   release-please-config.json
```

But RTK returns false empty output:

```bash
rtk ls
```

```text
(empty)
```

Verbose mode shows that RTK is definitely seeing data and then collapsing it:

```bash
rtk -v ls .
```

```text
Chars: 2072 → 8 (100% reduction)
(empty)
```

Same for a matching file glob:

```bash
rtk -v ls *.sh
```

```text
Chars: 60 → 8 (87% reduction)
(empty)
```

## Locale-sensitive difference

This appears to depend on the date ordering produced by `ls`.

With POSIX/C locale:

```bash
LC_ALL=C /bin/ls -la .
```

includes lines like:

```text
-rw-r--r--@   1 donvince  staff  48140 Apr 29 10:42 Cargo.lock
```

With my normal locale:

```bash
LANG=en_GB.UTF-8 /bin/ls -la .
```

includes lines like:

```text
-rw-r--r--@   1 donvince  staff  48140 29 Apr 10:42 Cargo.lock
```

And RTK works if I force POSIX/C locale:

```bash
LC_ALL=C rtk -v ls .
```

```text
Chars: 2072 → 552 (74% reduction)
.claude/
.github/
.rtk/
Formula/
docs/
hooks/
openclaw/
scripts/
src/
tests/
.gitignore  503B
.release-please-manifest.json  20B
CHANGELOG.md  76.1K
CLAUDE.md  7.2K
CONTRIBUTING.md  13.9K
Cargo.lock  47.0K
Cargo.toml  1.6K
DISCLAIMER.md  2.2K
INSTALL.md  11.3K
LICENSE  10.5K
README.md  20.6K
README_es.md  5.3K
README_fr.md  6.5K
README_ja.md  5.7K
README_ko.md  5.3K
README_zh.md  5.2K
SECURITY.md  7.3K
build.rs  2.3K
install.sh  2.8K
release-please-config.json  181B

Summary: 20 files, 10 dirs (12 .md, 2 .json, 1 .sh, 1 .toml, 1 no ext, +3 more)
```

```bash
LC_ALL=C rtk -v ls *.sh
```

```text
Chars: 60 → 51 (15% reduction)
install.sh  2.8K

Summary: 1 files, 0 dirs (1 .sh)
```

A stripped environment also works:

```bash
env -i PATH="/usr/bin:/bin:/opt/homebrew/bin" HOME="$HOME" /opt/homebrew/bin/rtk -v ls .
```

## Expected behavior

`rtk ls` should not emit false `(empty)` output for a non-empty directory.

If parsing assumptions do not hold, I would expect RTK to either:
1. parse the locale-specific `ls` output correctly, or
2. fall back to raw output

but not silently suppress valid data.

## Likely code path

`LC_ALL=C` forces the POSIX/C locale and makes `rtk ls` work. This matches the current parser implementation in `src/cmds/system/ls.rs`, which explicitly expects month-day date ordering and therefore does not match day-month output like `29 Apr 10:42`.

The parser in `src/cmds/system/ls.rs` appears to anchor on a month-day date shape and does not handle day-month ordering.

## Possible fix directions

I’m not attached to a specific implementation, but these seem plausible:

- force `LC_ALL=C` for the child `ls` process to stabilize input format
- make the parser robust against locale-variant `ls` date formats (regardless of language and ordering)
- in any case, avoid returning false `(empty)` when parsing fails and fall back to raw output instead

Happy to open a PR if maintainers have a preferred direction.

SAY-5:
Closes #1566. Also fixes #1652.

## Reproduction (matches the issue)
```bash
LANG=en_IN.UTF-8 # or en_GB.UTF-8, or any non-US locale BSD ls honors
TMP=$(mktemp -d)
for n in alpha bravo charlie delta echo; do touch \"$TMP/$n.dart\"; done

rtk ls \"$TMP/\"     # before this PR: \"(empty)\"
rtk -v ls \"$TMP/\"  # before this PR: \"Chars: 458 → 8 (99% reduction)\" then \"(empty)\"
```

## Root cause
BSD `ls` honors `LC_TIME` and emits `28 Apr 02:47` (`DD Mon HH:MM`) on locales such as `en_GB.UTF-8`, `en_IN.UTF-8`, etc. The `LS_DATE_RE` regex in `cmds/system/ls.rs` only anchored on `Mon DD HH:MM`, so `parse_ls_line` returned `None` for every entry and `compact_ls` fell into the empty-result branch, printing `(empty)` *and* a misleading 99% \"compression\" banner.

## Fix
Two minimal, complementary changes:

1. **Pin the subprocess locale.** `cmd.env(\"LC_TIME\", \"C\")` is set before spawning `ls`, so the date column is always emitted as `Mon DD HH:MM` regardless of the contributor's shell locale. This is the *correctness* fix.
2. **Relax the regex** to accept both orderings (`Mon DD …` and `DD Mon …`). Belt-and-suspenders so any future caller that pipes localized output through the parser still works.

## Test plan
Added three regression tests in `cmds/system/ls::tests`:

- [x] `test_compact_locale_dd_mon_time`, the exact 5-file shape from the issue (`28 Apr 02:47`); asserts file names, file count, and `Summary: 3 files`.
- [x] `test_compact_locale_dd_mon_year`, year-format variant (`25 Dec  2024`).
- [x] `test_parse_ls_line_locale_dd_mon_time`, unit-level check that type / size / filename are extracted correctly from the new shape.

Results:
- [x] `cargo test --bin rtk -- cmds::system::ls` → **21 passed; 0 failed** (3 new + 18 existing).
- [x] End-to-end smoke test on macOS Darwin 25.0 with `LANG=en_IN.UTF-8`:
    ```
    === rtk ls ===
    alpha.dart  0B
    bravo.dart  0B
    charlie.dart  0B
    delta.dart  0B
    echo.dart  0B
    === rtk -v ls ===
    Chars: 398 → 76 (81% reduction)
    alpha.dart  0B
    ...
    ```
- [x] `cargo build --release` clean.
- [x] `cargo fmt` clean.

## Notes
- No filter behavior change for users on US/POSIX locales (existing tests pass unchanged).
- `LC_TIME=C` only affects the spawned subprocess; the user's shell locale is untouched.
- This is the smallest possible fix; it does not touch the noise filter, the summary line, or the output format. The \"never returns empty for non-empty dirs\" hard floor suggested in the issue's triage section is left for follow-up since the immediate bug is the locale-induced parse miss.


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes a bug where `rtk ls` produced empty output on non-US locales (e.g. en_GB, en_IN) because BSD `ls` emits dates in `DD Mon HH:MM` format instead of `Mon DD HH:MM`. The fix pins `LC_TIME=C` on the spawned subprocess and relaxes the date-parsing regex to accept both orderings.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1566

---

donvince:
This PR also covers #1652 (rtk ls false `(empty)` output with `en_GB.UTF-8` locale) — the day-month date ordering fix and `LC_TIME=C` pinning address that case exactly. Flagging in case it helps with prioritisation.

SAY-5:
Good catch @donvince — updated the PR body to also reference #1652.

PR #1570, opened 2026-04-28
fix(gh): accept `gh run view --job <id>` without positional run-id (#1568)

### LINKED ISSUE #1568: `gh run view --job <id>` rejected with "Run ID required"
`gh run view` accepts `--job <id>` as an alternative to a positional run ID, but rtk rejects this form with `rtk: Run ID required`.

Reproduction:

```
$ gh run view --job 72893237475 --repo owner/repo --log
rtk: Run ID required
```

The same command works with `RTK_DISABLED=1`.

Per the gh CLI, the signature is `gh run view [<run-id>] [flags]` and `--job` is documented as a way to view a specific job — the positional run ID becomes optional in that case.

`view_run` requires a positional identifier and errors when none is found, even though `--job` and its value are correctly consumed into `extra_args`:

https://github.com/rtk-ai/rtk/blob/e1f2845df06a8d8b8325945dc4940ec5f530e4cc/src/cmds/git/gh_cmd.rs#L745-L749

`--job` is already listed in `flags_with_value` so its value is not mistaken for the run ID, but the `--job`-only invocation path was not handled:

https://github.com/rtk-ai/rtk/blob/e1f2845df06a8d8b8325945dc4940ec5f530e4cc/src/cmds/git/gh_cmd.rs#L124-L133

Environment: rtk 0.37.2, Windows 11, Git Bash.

Related: #416 / #477 fixed the case where `--job`'s value was being eaten as the run ID, but did not cover the case where `--job` is the sole identifier.

SAY-5:
Closes #1568.

## Reproduction
```
$ gh run view --job 72893237475 --repo owner/repo --log
rtk: Run ID required
```
(Same command works under `RTK_DISABLED=1`, confirming the issue is in the rtk wrapper.)

## Root cause
`gh run view` treats `--job <id>` as an alternative to the positional run-id: when `--job` is supplied, the run-id becomes optional. rtk's `view_run` (`src/cmds/git/gh_cmd.rs:745`) calls `extract_identifier_and_extra_args` and returns `Err(\"Run ID required\")` whenever no positional was found, even though `--job <value>` is already correctly listed in `flags_with_value` and consumed into `extra`.

So the user's intent never reaches gh, the wrapper rejects the call up front.

## Fix
Two small, focused changes in `view_run`:

1. New helper `extract_job_id(args)` that scans windows of two for the literal token `--job` and returns the next argument. It does not match substrings like `--job-foo`.
2. When `extract_identifier_and_extra_args` returns `None`, fall back to `extract_job_id`. If `--job` is present, pass the original args to gh via `run_passthrough_with_extra(\"gh\", &[\"run\", \"view\"], &args.to_vec())` instead of erroring. If neither a positional nor `--job` is present, the existing `Run ID required` error stands.

This matches the behavior gh already documents and matches the existing pattern: `--log`, `--log-failed`, and `--json` already trigger passthrough; this just stops rejecting the call before that branch is reached.

## Tests
Added four test cases under `cmds::git::gh_cmd::tests`:
- [x] `test_extract_job_id_when_present`, resolves the value from the exact arg shape in the issue.
- [x] `test_extract_job_id_absent`, returns `None` when no `--job` is given.
- [x] `test_extract_job_id_does_not_match_substrings`, `--job-foo` is not treated as `--job`.
- [x] `test_run_view_only_job_flag_no_positional_resolves_to_none`, pins the contract: identifier extractor → `None`, job extractor → `Some(id)`.

Results:
- [x] `cargo test --bin rtk -- cmds::git::gh_cmd` → **61 passed; 0 failed** (4 new + 57 existing, including the pre-existing `test_extract_identifier_with_job_flag_after/before/and_log_failed` which still pass).
- [x] `cargo fmt` clean.

## Notes
- This is a complement to #416 / #477, which fixed `--job <value>` being eaten as the run-id when both were given. That work added `--job` to `flags_with_value`. This PR closes the remaining gap: the `--job`-only invocation.
- No filtering behavior change for existing call shapes.


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes a bug where `gh run view --job <id>` without a positional run-id was rejected by the rtk wrapper with 'Run ID required', even though the gh CLI treats `--job` as a valid alternative identifier. Adds an `extract_job_id` helper to detect the `--job` flag and falls back to passthrough mode instead of erroring.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1568

---

PR #1571, opened 2026-04-28
feat: rtk extract — proximity windows for minified files

### LINKED ISSUE #1567: feat: proximity-based extraction for minified files (`rtk extract`)
## Problem

`rtk grep` is line-oriented. On minified bundles (single-line JS, compressed JSON, log files without newlines) a single match line can be 100K+ characters. The current behavior truncates to `max_line_len` (default 80) and destroys the surrounding semantic context that an LLM needs to reason about the match.

Concretely, running `rtk grep 'fetch' bundle.min.js` on a 1 MB minified file either:
- truncates the matching line to 80 chars (losing the URL, payload, options object), or
- returns the entire 1 MB line if `--max-len` is raised, blowing the token budget.

Neither outcome is useful, and this is exactly the kind of token-waste RTK is meant to prevent.

## Proposal

Add a new subcommand `rtk extract` that finds regex hits and emits a configurable **character window** around each match — independent of line boundaries — with optional secondary-keyword filtering to drop irrelevant windows.

```
rtk extract <PATTERN> [PATH] [-w/--window N] [-b/--before N] [-a/--after N]
                              [-r/--require KEYWORD]... [-i/--ignore-case]
                              [-m/--max N] [--no-dedupe]
```

| Flag | Default | Purpose |
|---|---|---|
| `-w, --window` | 100 | Symmetric window (chars). |
| `-b, --before` / `-a, --after` | from `--window` | Asymmetric overrides. |
| `-r, --require` | none, repeatable | Window must contain ALL of these substrings. |
| `-i, --ignore-case` | false | Case-insensitive primary regex and `--require`. |
| `-m, --max` | 100 | Token-budget cap on emitted windows. |
| `--no-dedupe` | false | Disable collapsing of identical windows (`(xN)`). |

### Example

```
$ rtk extract 'fetch\([^)]{0,300}\)' bundle.min.js -w 80 -r '/api/v1/login'
bundle.min.js (2 matches)
  @184523: ...auth.Ka.send(«fetch(\"/api/v1/login\",{method:\"PUT\",body:JSON.stringify(t)})»);return r...
  @201117: ...refresh=()=>«fetch(\"/api/v1/login\",{method:\"POST\",body:e})»;...
1 file, 2 windows shown (of 14 matches, 12 filtered by --require)
```

### Why a new command instead of grep flags

Keeping `grep` line-oriented is valuable for the 99% case of normal source files. Folding `--window` and `--require` into `grep` would either change its default semantics (breaking) or add flags that only make sense in a different mental model. A separate command keeps both clean.

## Expected savings

On a typical minified bundle, a 1 MB match-line collapses to a ~300-char window — roughly 99.9% reduction. Even on multi-match cases capped at `--max=100`, output stays in the low-KB range vs. raw multi-MB. Comfortably above RTK's 60% bar.

## Prior art

I have a working Python proof-of-concept that uses the same model (`re.finditer` + character slicing + secondary substring filter) and it has been the most useful pattern for analyzing reverse-engineered minified bundles. Porting it natively into RTK keeps the workflow inside the existing `rtk gain` tracking pipeline.

## Out of scope (deliberately deferred)

- Streaming for files > 5 MB (initial version reads whole file; > 5 MB will fail loud).
- Folding `--window` / `--require` back into `rtk grep`.
- Binary-file detection beyond the read-as-utf8 short-circuit.

I have an implementation ready and will open a PR shortly.

furkankoykiran:
## Summary

- Adds `rtk extract <PATTERN> [PATH]`, a regex matcher that emits a configurable **character window** around each hit instead of returning whole lines. Aimed at minified bundles where `rtk grep` either truncates to 80 chars (losing context) or returns a multi-MB single line (blowing the token budget).
- Optional `-r/--require KEYWORD` (repeatable, AND-combined) drops windows that lack secondary substrings — useful for narrowing wide patterns like `fetch\([^)]{0,300}\)` down to the call sites that touch a specific endpoint.
- New file `src/cmds/system/extract_cmd.rs` (~315 lines + 18 tests). Routing wired through `src/main.rs` (Clap variant, dispatcher arm, integrity-check whitelist).

Closes #1567

## Why a new command instead of `grep` flags

`grep` is line-oriented and that's the right model for normal source. Folding `--window`/`--require` into it would either change defaults (breaking) or add flags that only make sense in a different mental model. Keeping both clean.

## Behavior

```
$ rtk extract 'fetch\([^)]{0,300}\)' bundle.min.js -w 80 -r '/api/v1/login'
bundle.min.js (2 matches)
  @184523: ...auth.Ka.send(«fetch("/api/v1/login",{method:"PUT",body:JSON.stringify(t)})»);return r...
  @201117: ...refresh=()=>«fetch("/api/v1/login",{method:"POST",body:e})»;...
1 file, 2 windows shown (of 14 matches, 12 filtered)
```

| Flag | Default | Purpose |
|---|---|---|
| `-w, --window` | `100` | Symmetric chars on each side. |
| `-b, --before` / `-a, --after` | from `--window` | Asymmetric overrides. |
| `-r, --require` | none, repeatable | All substrings must be in the window. |
| `-i, --ignore-case` | `false` | Case-insensitive pattern *and* `--require`. |
| `-m, --max` | `100` | Token-budget cap on emitted windows. |
| `--no-dedupe` | `false` | Disable collapsing of identical windows (`(xN)`). |

- Newlines inside windows are replaced with `␤` so multi-line context fits one logical line.
- Match wrapped in `«…»` for unambiguous boundaries without ANSI.
- File walk skips `.git/`, `node_modules/`, `target/`, `dist/`, `build/` and silently skips files >5MB; a single oversize file passed directly errors loud rather than silently truncating.
- Char-boundary safe (multibyte UTF-8 input doesn't panic — covered by test).
- No new dependencies; everything via existing `regex`, `walkdir`, `anyhow`.
- Tracked via `TimedExecution::track` under the `rtk extract` label so it shows up in `rtk gain --history`.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` — all 1711 tests pass (18 new).
- [x] Manual smoke (release build):
  ```
  printf 'xxx...fetch("/api/v1/login",{method:"PUT"})...yyy' > /tmp/min.js
  rtk extract 'fetch\([^)]{0,300}\)' /tmp/min.js -w 30 -r login
  rtk extract 'NOPE' /tmp/min.js  # exit 1
  ```
- [x] Token-savings test (`end_to_end_minified_savings_above_95pct`) asserts ≥95% byte savings on a 50KB synthetic minified blob.
- [x] Multibyte test (`slice_window_multibyte_no_panic`) verifies char-boundary slicing on Japanese input.
- [x] Boundary tests verify no panic at offset 0 and at end-of-file.

## Out of scope (deliberately deferred)

- Streaming for files >5 MB (current implementation reads whole file; >5 MB fails loud).
- Folding `--window`/`--require` back into `rtk grep`.
- Binary-file detection beyond the read-as-utf8 short-circuit.

### DIFF
--- a/src/cmds/system/extract_cmd.rs
+++ b/src/cmds/system/extract_cmd.rs
@@ -1,0 +1,566 @@
+//! Proximity-based regex extraction for minified files.
+//!
+//! `rtk grep` is line-oriented and useless on single-line minified bundles.
+//! This module extracts a configurable character window around each regex
+//! match instead, with optional secondary-keyword filtering.
+
+use crate::core::tracking;
+use anyhow::{Context, Result};
+use regex::RegexBuilder;
+use std::collections::BTreeMap;
+use std::fs;
+use std::path::{Path, PathBuf};
+use walkdir::WalkDir;
+
+const MAX_FILE_BYTES: u64 = 5 * 1024 * 1024;
+const SKIP_DIRS: &[&str] = &[".git", "node_modules", "target", "dist", "build"];
+
+#[derive(Clone)]
+struct Window {
+    offset: usize,
+    text: String,
+}
+
+#[allow(clippy::too_many_arguments)]
+pub fn run(
+    pattern: &str,
+    path: &str,
+    before: usize,
+    after: usize,
+    require: &[String],
+    ignore_case: bool,
+    max_results: usize,
+    dedupe: bool,
+    verbose: u8,
+) -> Result<i32> {
+    let timer = tracking::TimedExecution::start();
+
+    if verbose > 0 {
+        eprintln!(
+            "extract: '{}' in {} (before={}, after={}, require={:?})",
+            pattern, path, before, after, require
+        );
+    }
+
+    let re = RegexBuilder::new(pattern)
+        .case_insensitive(ignore_case)
+        .build()
+        .with_context(|| format!("invalid regex: {pattern}"))?;
+
+    let raw_cmd = format!("extract '{pattern}' {path}");
+    let files = collect_files(Path::new(path))?;
+
+    let mut by_file: BTreeMap<String, Vec<Window>> = BTreeMap::new();
+    let mut total_matches = 0usize;
+    let mut filtered = 0usize;
+    let mut emitted = 0usize;
+    let mut input_bytes = 0usize;
+
+    'outer: for file in &files {
+        let content = match fs::read_to_string(file) {
+            Ok(c) => c,
+            Err(_) => continue, // binary / unreadable — skip silently
+        };
+        input_bytes = input_bytes.saturating_add(content.len());
+
+        for m in re.find_iter(&content) {
+            total_matches += 1;
+            let win = slice_window(&content, m.start(), m.end(), before, after);
+            if !passes_require(&win, require, ignore_case) {
+                filtered += 1;
+                continue;
+            }
+            by_file
+                .entry(file.display().to_string())
+                .or_default()
+                .push(Window {
+                    offset: m.start(),
+                    text: win,
+                });
+            emitted += 1;
+            if emitted >= max_results {
+                break 'outer;
+            }
+        }
+    }
+
+    let output = format_output(&by_file, total_matches, filtered, emitted, dedupe);
+    print!("{output}");
+
+    // estimate_tokens reads only .len(); use a length-proxy to avoid holding
+    // every file's bytes in memory simultaneously.
+    let raw_proxy: String = " ".repeat(input_bytes);
+    timer.track(&raw_cmd, "rtk extract", &raw_proxy, &output);
+
+    Ok(if total_matches == 0 { 1 } else { 0 })
+}
+
+fn collect_files(path: &Path) -> Result<Vec<PathBuf>> {
+    let meta = fs::metadata(path).with_context(|| format!("cannot stat {}", path.display()))?;
+    if meta.is_file() {
+        if meta.len() > MAX_FILE_BYTES {
+            anyhow::bail!(
+                "{} is {} bytes (>5MB cap). Streaming for large files is not yet supported.",
+                path.display(),
+                meta.len()
+            );
+        }
+        return Ok(vec![path.to_path_buf()]);
+    }
+
+    let mut out = Vec::new();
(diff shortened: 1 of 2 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟡 **Risk** | `medium` |

### Summary

Adds a new `rtk extract` command that performs regex matching with configurable character-window context around hits, designed for minified files where line-oriented grep is ineffective. Includes optional keyword filtering (--require), deduplication, and integrates with the existing tracking system. New module ~315 lines with 18 tests.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [x] Docs updated

**Linked issues:** #1567

---

furkankoykiran:
Quick real-world check before this sits waiting for review — ran against a 2MB minified bundle (single-line, the worst case for `grep`):

```
$ rtk extract 'fetch\([^)]{0,300}\)' bundle.min.js -w 60 -r '/api/v1' -m 6
bundle.min.js (3 matches)
  @1679899: ...switch(a.label){case 0:return[4,«fetch(e?"/api/v1/configuration/client/"+e:"/api/v1/configuration",{headers:Qa,method:"GET"})»];case 1:return[2,a.sent...
  @1680687: ...this,function(n){return[2,this.«fetch("/api/v1/console-link/verify",{method:"POST",body:JSON.stringify({code:e,token:a})»})]})})},e.prototype.parse=fu...
  @1917801: ...trys.push([1,6,,7]),a(null),[4,«fetch("/api/v1/tos?locale="+g,{credentials:"include",headers:{Accept:"text/html","Content-Type":"text/html"}})»];case...
1 file, 3 windows shown (of 8 matches, 5 filtered)
```

2,125,604 bytes in, ~1.2KB out — three actionable API call sites from a bundle that `grep` would either truncate to 80 chars or dump whole. `--require '/api/v1'` dropped five non-API `fetch()`es (telemetry, asset loaders) so what's left is what you'd actually open the file to read.

Matches the shape of the use case in #1567 (tracing login flows in reverse-engineered bundles).

PR #1573, opened 2026-04-28
fix(git): don't count --max-lines truncation as savings (#1561)

### LINKED ISSUE #1561: rtk git diff/show --max-lines may inflate savings same way as #1045
## Context

Follow-up flagged by @pszymkowiak in [PR #1076 review](https://github.com/rtk-ai/rtk/pull/1076#pullrequestreview-4183717289).

#1045 was about `rtk read` counting user-requested line truncation as RTK token savings. The same pattern likely exists in:

- `src/cmds/git/git.rs:200` — `git diff` tracking
- `src/cmds/git/git.rs:314` — `git show` tracking

Both currently track against the **compacted** output. `compact_diff` itself is real RTK compression so that part is legitimate — **but** when a user passes `--max-lines N` to `rtk git diff` / `rtk git show`, the truncated portion gets counted as savings same as the original bug.

## Reproduce (likely)

```bash
rtk git diff --max-lines 5  # large diff
rtk gain --history          # check if the saved-tokens row for this command is suspiciously high
```

## Suggested fix

Apply the same pattern as #1076:

1. Extract a `pick_tracking_output(pre_truncation: Option<&str>, rtk_output: &str) -> &str` helper (already lives in `src/cmds/system/read.rs` after #1076 merges — could be promoted to `src/core/utils.rs`).
2. Capture the post-compact / pre-truncation content via `std::mem::take` when `--max-lines` is active.
3. Pass the pre-truncation slice to `timer.track()` instead of the windowed output.

## Out of scope here

- Polluted historical data (separate concern — needs server-side cap or one-off migration, also flagged in #1076 review).

cc @pszymkowiak — opening as you suggested.

SAY-5:
Closes #1561.

## Background
Sibling of #1045 (already addressed for \`rtk read\` in #1076). \`rtk git diff\` and \`rtk git show\` previously called:
```rust
compact_diff(stdout, max_lines.unwrap_or(500))
```
and then tracked savings against the resulting (possibly truncated) output. \`compact_diff\` performs *both* real RTK compression and the early-break truncation at \`result.len() >= max_lines\`. When the user passed \`--max-lines N\`, the truncated tail was counted as RTK savings — the same inflation pattern as #1045.

## Fix
Apply the same shape #1076 used for read:

1. Promote the implicit RTK budget to a named constant \`RTK_DIFF_BUDGET = 500\`. \`compact_diff\` is now always called with this budget, so its early break represents real RTK compression and nothing else.
2. Add \`truncate_to_lines(&str, Option<usize>) -> Cow<'_, str>\` as a separate post-step that honors the user's \`--max-lines\`. \`Cow::Borrowed\` on the no-truncation fast path keeps the common case zero-alloc.
3. In \`run_diff\` and \`run_show\`, display the user-truncated slice but track against the RTK-compressed (pre-user-truncation) string. When \`max_lines\` is \`None\`, the tracking string equals \`final_output\` exactly as before.

## Behavior
| Invocation | Displayed | Tracked |
|---|---|---|
| \`rtk git diff\` (no flag) | compact_diff with 500-line cap (unchanged) | same as displayed (unchanged) |
| \`rtk git diff --max-lines 5\` | compact_diff(500) → first 5 lines + \"more changes truncated\" | compact_diff(500) — full ~500-line view |
| Same for \`rtk git show\` | … | … |

## Test plan
Four regression tests under \`cmds::git::git::tests\`:
- [x] \`test_truncate_to_lines_no_truncation_when_none\` — pins the \`Cow::Borrowed\` fast path.
- [x] \`test_truncate_to_lines_under_limit_returns_owned_unchanged_body\` — body preserved, no breadcrumb when under the limit.
- [x] \`test_truncate_to_lines_over_limit_emits_breadcrumb\` — over-limit truncates and emits the breadcrumb.
- [x] \`test_max_lines_does_not_inflate_tracking\` — end-to-end shape: 200-line compacted diff + \`--max-lines 5\` → displayed ≤ 6 lines with breadcrumb, tracking has all 200 and no breadcrumb.

Results:
- [x] \`cargo test --bin rtk -- cmds::git::git\` → **62 passed; 0 failed** (4 new + 58 existing).
- [x] \`cargo build --bin rtk\` clean (pre-existing libc::sighandler_t warning is unrelated).
- [x] \`cargo fmt\` clean.

## Notes
- Independent of #1076 (which is still in review): this PR introduces its own \`truncate_to_lines\` helper local to \`git.rs\` rather than depending on a future shared \`pick_tracking_output\` in \`core/utils.rs\`. If maintainers prefer that helper to land first, this can be rebased to use it.
- Polluted historical data is out of scope (per the issue) — this fix only makes future tracking accurate.


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes inflated RTK token savings tracking when users pass --max-lines to `rtk git diff` or `rtk git show`. The truncation from --max-lines was being counted as RTK compression savings. The fix separates RTK compression (compact_diff with a fixed 500-line budget) from user-requested truncation (new truncate_to_lines helper), tracking against the pre-truncation content.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1561

---

PR #1575, opened 2026-04-28
fix(dotnet): move build/test/restore status line to the bottom (#1574)

SAY-5:
Closes #1574.

## Reproduction (from the issue)
```bash
rtk dotnet build 2>&1 | tail -5
```
Native `dotnet` puts `Build succeeded.` / `Build FAILED.` at the bottom; consumers that read the end of the stream (`tail -N`, IDE log followers, agent watch loops) reliably see the verdict. RTK put it at the *top*, so the agent saw error noise + EOF with no `ok`/`fail` line anywhere near the end. This silently broke any agent loop that gates the next step on build success.

## Fix (Option 2 from the issue)
Reorder each of `format_build_output`, `format_test_output`, and `format_restore_output` to:

1. build the body (errors / warnings / failed tests),
2. emit the `---` separator only if the body is non-empty,
3. append the verdict header last.

So the final line of the formatted output is always the verdict — matching native `dotnet`.

## Behavior
| Invocation | Last non-empty line (before) | Last non-empty line (after) |
|---|---|---|
| `rtk dotnet build` (success, no warnings) | `ok dotnet build: …` | `ok dotnet build: …` |
| `rtk dotnet build` (failure with errors) | last error line | `fail dotnet build: …` |
| `rtk dotnet test` (failures) | last warning / error line | `fail dotnet test: …` |
| `rtk dotnet restore` (errors) | last error line | `fail dotnet restore: …` |

No information loss: error / warning / failed-test sections still print in full, just above the verdict instead of below.

## Test plan
Four regression tests under `cmds::dotnet::dotnet_cmd::tests`:
- [x] `test_format_build_output_status_at_bottom_failure` — `fail dotnet build:` is the last line even when errors are present.
- [x] `test_format_build_output_status_at_bottom_success` — `ok dotnet build:` is the last line in the no-issues case.
- [x] `test_format_test_output_status_at_bottom` — `fail dotnet test:` last when there are failed tests.
- [x] `test_format_restore_output_status_at_bottom` — `fail dotnet restore:` last when restore has errors.

Each test calls a shared `last_non_empty_line` helper to assert the contract.

Results:
- [x] `cargo test --bin rtk -- cmds::dotnet` → **109 passed; 0 failed** (4 new + 105 existing).
- [x] `cargo build --bin rtk` clean.
- [x] `cargo fmt` clean.

## Notes
[...]

Issue #1560, opened 2026-04-27
bug(rewrite): hook rewrites left-hand side of piped commands, breaking downstream filters


kotofos:
## Summary

`rtk rewrite` (and `rtk hook claude`) rewrites piped commands like `ps aux | grep foo | grep -v grep` into `rtk ps aux | grep foo | grep -v grep`. Since `rtk ps aux` compresses/reformats output, the downstream `grep` no longer matches — the pipe silently returns empty results.

## Reproduction

```bash
$ rtk rewrite 'ps aux | grep python | grep -v grep'
rtk ps aux | grep python | grep -v grep    # exit 0 — rewrite applied

$ ps aux | grep python | grep -v grep      # returns matching processes
$ rtk ps aux | grep python | grep -v grep  # returns nothing
```

Also confirmed via hook:

```bash
$ echo '{"tool_input":{"command":"ps aux | grep python | grep -v grep"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk ps aux | grep python | grep -v grep"}}}
```

## Expected behavior

`rtk rewrite` should detect pipes in the command string and either:

1. Skip rewriting entirely for piped commands, or
2. Only rewrite if downstream stages are rtk-compatible

## Workaround

Wrapper script around `rtk hook claude` that skips piped commands:

```bash
#!/usr/bin/env bash
INPUT=$(cat)
CMD=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null)

case "$CMD" in
  *"|"*) exit 0 ;;
esac

echo "$INPUT" | rtk hook claude
```

Wire in `settings.json`:

```json
"command": "~/.claude/hooks/rtk-hook-wrapper.sh"
```

## Environment

- rtk 0.37.2 (also reproduced on 0.34.2)
- Hook: `rtk hook claude` (v0.37.2 built-in)
- OS: Linux x86_64

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 96% |

### Summary

[...]

PR #1576, opened 2026-04-28
fix(dotnet): move build/test/restore status line to the bottom (#1574)

### LINKED ISSUE #1574: bug(dotnet): build/test summary line at top of output is cut off by `| tail -N`
## TL;DR

`rtk dotnet build` (also `test`, `restore`) puts the success/failure status line at the **top** of the output. Native `dotnet` puts it at the **bottom**. Anything that reads the *end* of the stream loses the verdict: `| tail -N`, watch/monitor mode in agents, IDE log tails, chunked-output readers.

## Repro

```bash
rtk dotnet build 2>&1 | tail -5
```

Expected: the last 5 lines include `ok dotnet build: 0 errors, 0 warnings (…)`.
Actual: the last 5 lines are warnings or empty; the status line is gone.

## Why agents in particular get hurt

Agents (Claude Code, Codex, Copilot CLI, etc.) commonly read build output through:

- `| tail -N` to keep token usage low
- streaming / watch consumers (`/watch`, `/monitor`, `tail -f`, log followers)
- a bounded recent-output context window

In all three cases the agent only sees the last N lines or the tail of the stream. Native `dotnet` writes `Build succeeded.` or `Build FAILED.` last, so the agent gets a clear verdict. RTK writes the verdict first, then errors and warnings, so the agent sees error noise followed by EOF, with no `ok` or `fail` line anywhere near the end.

Result: the agent cannot tell whether the build passed, even though the information was technically there 200 lines ago. This silently breaks any agent loop that gates the next step on build success.

## Why this happens (code)

`src/cmds/dotnet/dotnet_cmd.rs::format_build_output` (and the test/restore equivalents) emit the status line first, then errors, then warnings.

Two things compound the problem:

1. The hook rewrites the left side of pipes. `dotnet build 2>&1 | tail -5` becomes `rtk dotnet build … | tail -5`, so `tail` operates on the filtered stream, not raw `dotnet` output. Tracked in #1560.
2. On failure, raw stdout/stderr is prepended before the filtered summary (`dotnet_cmd.rs` lines 216–228), pushing the verdict even further from the tail. PR #1115 (issue #914) drops the prepend on failure but does not move the summary line.

## Suggested fixes

Two options, not mutually exclusive:

1. Hook side (root fix, #1560): stop rewriting the left side when piping into `tail`, `head`, `wc`, `grep`, or any explicit "read-the-end" consumer. A pipe into `| tail -N` is an explicit user signal for raw, unreordered output.
2. Filter side (defensive): move the status line to the bottom of `format_{build,test,restore}_output`, matching native `dotnet`. Cheap, unblocks streaming consumers and tail/head pipes today, even before #1560 lands.

Option 2 alone fixes the agent watch/monitor case described above. Option 1 alone fixes the explicit-pipe case but not streaming consumers that read RTK's stdout directly.

## Related

- #1560: hook rewrites left side of piped commands (root cause for the pipe case)
- #914 / PR #1115: sibling dotnet output-shape issue (failure path)
- #1545: streaming / watch-mode commands buffer or suppress stdout outside proxy mode (sibling for the streaming case)

## Environment

- Code reference: `src/cmds/dotnet/dotnet_cmd.rs:986` (`format_build_output` builds the header first)
- Reproduces on `master` (commit 4113696)

F0rty-Tw0:
Closes #1574.

## Reproduction

```bash
rtk dotnet build 2>&1 | tail -5
```

Native `dotnet` ends with `Build succeeded.` / `Build FAILED.`, so consumers reading the end of the stream (`tail -N`, IDE log followers, agent watch/monitor loops, bounded context windows) reliably see the verdict. RTK put it at the *top*, so agents reading via `| tail -N` or with limited recent-output windows saw error noise followed by EOF, with no `ok` / `fail` line anywhere near the end. This silently broke any agent loop that gates the next step on build success.

## Fix (Option 2 from the issue)

Reorder `format_build_output`, `format_test_output`, and `format_restore_output` to:

1. build the body (errors / warnings / failed tests),
2. emit the `---` separator only when the body is non-empty,
3. append the verdict header last.

The final line of the formatted output is now always the verdict, matching native `dotnet`.

## Behavior

| Invocation | Last line (before) | Last line (after) |
|---|---|---|
| `rtk dotnet build` (success) | `ok dotnet build: …` | `ok dotnet build: …` |
| `rtk dotnet build` (errors) | last error / warning | `fail dotnet build: …` |
| `rtk dotnet test` (failures) | last error / warning | `fail dotnet test: …` |
| `rtk dotnet restore` (errors) | last error / warning | `fail dotnet restore: …` |

No information loss. Error / warning / failed-test sections still print in full, just above the verdict.

## Output contract change

This is a breaking change for any downstream consumer that parses the *first* line of dotnet output. RTK's filtered output is not a stable parse target by design (it includes runtime-dependent project counts and durations), and the issue explicitly motivates moving the verdict to the tail, but flagging it for anyone relying on line-1 positioning.

## Token impact

Neutral to −1 byte across all paths (clean build, errors only, warnings only, errors + warnings). The `!out.is_empty()` guard avoids emitting a useless separator on clean one-line builds.

## Test plan

Three regression tests in `src/cmds/dotnet/dotnet_cmd.rs`:

- [x] `test_format_build_output_status_line_is_last_for_tail_consumers` — asserts `output.lines().last()` starts with `ok dotnet build:` AND that the verdict is within the last 5 lines (`tail -5`).
- [x] `test_format_test_output_status_line_is_last_for_tail_consumers` — asserts `ok dotnet test:` is the last line.
- [x] `test_format_restore_output_status_line_is_last_for_tail_consumers` — asserts `fail dotnet restore:` is the last line.

The contract uses strict `lines().last()` rather than "last non-empty line" so a future regression that appends a trailing context line under the verdict gets caught. The `tail -5` inclusion assertion mirrors the exact user-reported scenario.

Results:
- [x] `cargo fmt --all` clean
- [x] `cargo clippy --all-targets` clean (no new warnings in `dotnet_cmd.rs`)
- [x] `cargo test --all` → 1690 passed, 0 failed, 6 ignored

## Notes

- Independent of #1560 (hook left-side rewrite). Per the issue: "Option 2 alone fixes the agent watch/monitor case described above. Option 1 alone fixes the explicit-pipe case but not streaming consumers that read RTK's stdout directly."
- Composes with #1115 (failure-path output shape). Traced both interleavings: the verdict stays at the bottom whether `output_to_print = filtered` (post-#1115) or `<raw>\n\n<filtered>` (current). No merge conflict expected at the formatter layer.
- Implements the same Option 2 approach as the concurrently-opened #1575. Tests in this PR use strict `lines().last()` plus a `tail -5` inclusion check, which catches a wider class of trailing-content regressions. Maintainers should pick whichever PR fits their preference.

### DIFF
--- a/src/cmds/dotnet/dotnet_cmd.rs
+++ b/src/cmds/dotnet/dotnet_cmd.rs
@@ -987,20 +987,6 @@ fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> S
     let status_icon = if summary.succeeded { "ok" } else { "fail" };
     let duration = summary.duration_text.as_deref().unwrap_or("unknown");
 
-    let mut warnings = String::new();
-    if !summary.warnings.is_empty() {
-        warnings.push_str("Warnings:\n");
-        for issue in summary.warnings.iter().take(10) {
-            warnings.push_str(&format!("{}\n", format_issue(issue, "warning")));
-        }
-        if summary.warnings.len() > 10 {
-            warnings.push_str(&format!(
-                "  ... +{} more warnings\n",
-                summary.warnings.len() - 10
-            ));
-        }
-    }
-
     let mut errors = String::new();
     if !summary.errors.is_empty() {
         errors.push_str("Errors:\n");
@@ -1011,6 +997,20 @@ fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> S
             errors.push_str(&format!(
                 "  ... +{} more errors\n",
                 summary.errors.len() - 20
+            ));
+        }
+    }
+
+    let mut warnings = String::new();
+    if !summary.warnings.is_empty() {
+        warnings.push_str("Warnings:\n");
+        for issue in summary.warnings.iter().take(10) {
+            warnings.push_str(&format!("{}\n", format_issue(issue, "warning")));
+        }
+        if summary.warnings.len() > 10 {
+            warnings.push_str(&format!(
+                "  ... +{} more warnings\n",
+                summary.warnings.len() - 10
             ));
         }
     }
@@ -1103,17 +1103,6 @@ fn format_test_output(
         }
     }
 
-    let mut warnings_section = String::new();
-    if !warnings.is_empty() {
-        warnings_section.push_str("Warnings:\n");
-        for issue in warnings.iter().take(10) {
-            warnings_section.push_str(&format!("{}\n", format_issue(issue, "warning")));
-        }
-        if warnings.len() > 10 {
-            warnings_section.push_str(&format!("  ... +{} more warnings\n", warnings.len() - 10));
-        }
-    }
-
     let mut errors_section = String::new();
     if !errors.is_empty() {
         errors_section.push_str("Errors:\n");
@@ -1122,6 +1111,17 @@ fn format_test_output(
         }
         if errors.len() > 10 {
             errors_section.push_str(&format!("  ... +{} more errors\n", errors.len() - 10));
+        }
+    }
+
+    let mut warnings_section = String::new();
+    if !warnings.is_empty() {
+        warnings_section.push_str("Warnings:\n");
+        for issue in warnings.iter().take(10) {
+            warnings_section.push_str(&format!("{}\n", format_issue(issue, "warning")));
+        }
+        if warnings.len() > 10 {
+            warnings_section.push_str(&format!("  ... +{} more warnings\n", warnings.len() - 10));
         }
     }
 
@@ -1163,17 +1163,6 @@ fn format_restore_output(
     let status_icon = if has_errors { "fail" } else { "ok" };
     let duration = summary.duration_text.as_deref().unwrap_or("unknown");
 
-    let mut warnings_section = String::new();
-    if !warnings.is_empty() {
-        warnings_section.push_str("Warnings:\n");
-        for issue in warnings.iter().take(10) {
-            warnings_section.push_str(&format!("{}\n", format_issue(issue, "warning")));
-        }
-        if warnings.len() > 10 {
-            warnings_section.push_str(&format!("  ... +{} more warnings\n", warnings.len() - 10));
-        }
-    }
-
     let mut errors_section = String::new();
     if !errors.is_empty() {
         errors_section.push_str("Errors:\n");
@@ -1182,6 +1171,17 @@ fn format_restore_output(
         }
         if errors.len() > 20 {
             errors_section.push_str(&format!("  ... +{} more errors\n", errors.len() - 20));
+        }
+    }
+
+    let mut warnings_section = String::new();
+    if !warnings.is_empty() {
+        warnings_section.push_str("Warnings:\n");
+        for issue in warnings.iter().take(10) {
+            warnings_section.push_str(&format!("{}\n", format_issue(issue, "warning")));
+        }
+        if warnings.len() > 10 {
+            warnings_section.push_str(&format!("  ... +{} more warnings\n", warnings.len() - 10));
         }
     }
 
--- a/src/cmds/dotnet/dotnet_cmd.rs
+++ b/src/cmds/dotnet/dotnet_cmd.rs
@@ -979,8 +979,11 @@ fn format_issue(issue: &binlog::BinlogIssue, kind: &str) -> String {
     )
 }
 
+/// Format the build summary for stdout.
+///
+/// `_binlog_path` is intentionally unused — the binlog is a temporary file
+/// that has already been cleaned up by the time this runs.
 fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> String {
-    // Binlog path omitted from output (temp file, already cleaned up)
     let status_icon = if summary.succeeded { "ok" } else { "fail" };
     let duration = summary.duration_text.as_deref().unwrap_or("unknown");
 
@@ -1013,16 +1016,11 @@ fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> S
     }
 
     let sep = if !warnings.is_empty() || !errors.is_empty() {
-        "---------------------------------------".to_string()
+        "---------------------------------------"
     } else {
-        String::new()
+        ""
     };
 
-    // Status line is emitted last so consumers that read the tail of the stream
-    // (`| tail -N`, agent watch/monitor modes, bounded context windows) get a
-    // definitive verdict. Mirrors native `dotnet build`, which ends with
-    // `Build succeeded.` / `Build FAILED.`. See issue #1574.
-    // Warnings before errors: errors survive `| tail -N` immediately above the verdict.
     let verdict = format!(
         "{} dotnet build: {} projects, {} errors, {} warnings ({})",
         status_icon,
@@ -1032,13 +1030,22 @@ fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> S
         duration
     );
 
-    [warnings, errors, sep, verdict]
+    // Status line is emitted last so consumers that read the tail of the stream
+    // (`| tail -N`, agent watch/monitor modes, bounded context windows) get a
+    // definitive verdict. Mirrors native `dotnet build`, which ends with
+    // `Build succeeded.` / `Build FAILED.`. See issue #1574.
+    // Warnings before errors: errors survive `| tail -N` immediately above the verdict.
+    [warnings, errors, sep.into(), verdict]
         .into_iter()
         .filter(|s| !s.is_empty())
         .collect::<Vec<_>>()
         .join("\n")
 }
 
+/// Format the test summary for stdout.
+///
+/// `_binlog_path` is intentionally unused — the binlog is a temporary file
+/// that has already been cleaned up by the time this runs.
 fn format_test_output(
     summary: &binlog::TestSummary,
     errors: &[binlog::BinlogIssue],
@@ -1078,7 +1085,6 @@ fn format_test_output(
         )
     };
 
-    // Binlog path omitted from output (temp file, already cleaned up)
     let mut failed_tests_section = String::new();
     if has_failures && !summary.failed_tests.is_empty() {
         failed_tests_section.push_str("Failed Tests:\n");
@@ -1119,28 +1125,40 @@ fn format_test_output(
         }
     }
 
-    let sep = if !failed_tests_section.is_empty() || !warnings_section.is_empty() || !errors_section.is_empty() {
-        "---------------------------------------".to_string()
+    let sep = if !failed_tests_section.is_empty()
+        || !warnings_section.is_empty()
+        || !errors_section.is_empty()
+    {
+        "---------------------------------------"
     } else {
-        String::new()
+        ""
     };
 
     // Status line emitted last; see format_build_output (issue #1574).
     // Warnings before errors: errors survive `| tail -N` immediately above the verdict.
-    [failed_tests_section, warnings_section, errors_section, sep, header]
+    [
+        failed_tests_section,
+        warnings_section,
+        errors_section,
+        sep.into(),
+        header,
+    ]
         .into_iter()
         .filter(|s| !s.is_empty())
         .collect::<Vec<_>>()
         .join("\n")
 }
 
+/// Format the restore summary for stdout.
+///
+/// `_binlog_path` is intentionally unused — the binlog is a temporary file
+/// that has already been cleaned up by the time this runs.
 fn format_restore_output(
     summary: &binlog::RestoreSummary,
     errors: &[binlog::BinlogIssue],
     warnings: &[binlog::BinlogIssue],
     _binlog_path: &Path,
 ) -> String {
-    // Binlog path omitted from output (temp file, already cleaned up)
     let has_errors = summary.errors > 0;
     let status_icon = if has_errors { "fail" } else { "ok" };
     let duration = summary.duration_text.as_deref().unwrap_or("unknown");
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Reorders the formatted output of `format_build_output`, `format_test_output`, and `format_restore_output` so the status verdict line (ok/fail) is emitted last instead of first. This ensures consumers reading the tail of the stream (e.g. `tail -N`, agent monitors) reliably see the build/test/restore verdict. Three regression tests are added to enforce the new contract.

### Review Checklist
- [x] Tests present
- [x] Breaking change
- [ ] Docs updated

**Linked issues:** #1574

---

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -983,17 +983,10 @@ fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> S
...
-        "{} dotnet build: {} projects, {} errors, {} warnings ({})",
-        status_icon,
-        summary.project_count,
-        summary.errors.len(),
-        summary.warnings.len(),
-        duration
-    );
+    let mut out = String::new();
 
     if !summary.errors.is_empty() {
```
KuSh, line 991: Order should also be reversed to have errors after warnings as those are most important than warnings. They should survive before the warnings in case of piping to tail

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1006,7 +999,10 @@ fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> S
     }
 
     if !summary.warnings.is_empty() {
-        out.push_str("\nWarnings:\n");
+        if !out.is_empty() {
```
KuSh: Another way would be to have four strings : errors, warnings, sep that is not empty if having errors or warnings and summary then iterate over the 4, filter out empty ones and joining them with `\n`. That way, we don't have to refactor everything when the order needs to be changed, or if something else needs to be inserted in between

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1018,7 +1014,23 @@ fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> S
         }
     }
 
-    // Binlog path omitted from output (temp file, already cleaned up)
```
KuSh, line 1021: That one could have been kept

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1089,7 +1106,10 @@ fn format_test_output(
     }
 
     if !warnings.is_empty() {
```
KuSh: Same remark, order should be reverted too

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1098,7 +1118,13 @@ fn format_test_output(
         }
     }
 
-    // Binlog path omitted from output (temp file, already cleaned up)
```
KuSh, line 1101: keep existing comments please

KuSh (changes requested):

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: Good point, fixed. Warnings now emit before errors in all three formatters so errors sit directly above the verdict, closer to what `| tail -N` captures.

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: Thats actually cool idea. Each formatter now builds independent section strings (`warnings, errors, sep, verdict`), filters out empty ones, and joins with `\n`. The `!out.is_empty()` guards are gone and reordering is a one-line change.

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: same fix applied

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: the comment sat right next to the verdict push we were moving, and got swept out with it, restored at the right place.

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: the comment sat right next to the verdict push we were moving, and got swept out with it, restored at the right place.

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -980,46 +980,63 @@ fn format_issue(issue: &binlog::BinlogIssue, kind: &str) -> String {
 }
 
 fn format_build_output(summary: &binlog::BuildSummary, _binlog_path: &Path) -> String {
+    // Binlog path omitted from output (temp file, already cleaned up)
```
KuSh: This seems misplaced. Preferably add a function comment description and add that information as a detail

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -980,46 +980,63 @@ fn format_issue(issue: &binlog::BinlogIssue, kind: &str) -> String {
...
+    }
 
+    let mut errors = String::new();
     if !summary.errors.is_empty() {
-        out.push_str("\n---------------------------------------\n\nErrors:\n");
+        errors.push_str("Errors:\n");
         for issue in summary.errors.iter().take(20) {
-            out.push_str(&format!("{}\n", format_issue(issue, "error")));
+            errors.push_str(&format!("{}\n", format_issue(issue, "error")));
         }
```
KuSh, line 999: nitpick: verdict could be created here and warnings could be moved down after errors to avoid to much noise in the PR

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -980,46 +980,63 @@ fn format_issue(issue: &binlog::BinlogIssue, kind: &str) -> String {
...
-                "  ... +{} more warnings\n",
-                summary.warnings.len() - 10
-            ));
-        }
-    }
+    let sep = if !warnings.is_empty() || !errors.is_empty() {
+        "---------------------------------------".to_string()
+    } else {
+        String::new()
+    };
```
KuSh, line 1022:
```suggestion
    let sep = if !warnings.is_empty() || !errors.is_empty() {
        "---------------------------------------"
    } else {
        ""
    };
```
Then use `sep.into()` when building the vec

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -980,46 +980,63 @@ fn format_issue(issue: &binlog::BinlogIssue, kind: &str) -> String {
...
 
-    // Binlog path omitted from output (temp file, already cleaned up)
-    out
+    // Status line is emitted last so consumers that read the tail of the stream
+    // (`| tail -N`, agent watch/monitor modes, bounded context windows) get a
+    // definitive verdict. Mirrors native `dotnet build`, which ends with
+    // `Build succeeded.` / `Build FAILED.`. See issue #1574.
+    // Warnings before errors: errors survive `| tail -N` immediately above the verdict.
```
KuSh: nitpick: That comment would be better placed over the final output

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1061,45 +1078,60 @@ fn format_test_output(
         )
     };
 
+    // Binlog path omitted from output (temp file, already cleaned up)
```
KuSh: Again this seems totally misplaced

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1061,45 +1078,60 @@ fn format_test_output(
...
                 summary.failed_tests.len() - 15
             ));
         }
     }
 
-    if !errors.is_empty() {
-        out.push_str("\nErrors:\n");
-        for issue in errors.iter().take(10) {
-            out.push_str(&format!("{}\n", format_issue(issue, "error")));
+    let mut warnings_section = String::new();
```
KuSh: nitpick: If you can reorder the creation of the different strings to avoid too much noise that would be great

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1061,45 +1078,60 @@ fn format_test_output(
...
         }
     }
 
-    // Binlog path omitted from output (temp file, already cleaned up)
-    out
+    let sep = if !failed_tests_section.is_empty() || !warnings_section.is_empty() || !errors_section.is_empty() {
+        "---------------------------------------".to_string()
+    } else {
+        String::new()
+    };
```
KuSh:
```suggestion
    let sep = if !failed_tests_section.is_empty() || !warnings_section.is_empty() || !errors_section.is_empty() {
        "---------------------------------------"
    } else {
        ""
    };
```

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1108,37 +1140,51 @@ fn format_restore_output(
     warnings: &[binlog::BinlogIssue],
     _binlog_path: &Path,
 ) -> String {
+    // Binlog path omitted from output (temp file, already cleaned up)
```
KuSh: again misplaced

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1108,37 +1140,51 @@ fn format_restore_output(
...
     let has_errors = summary.errors > 0;
     let status_icon = if has_errors { "fail" } else { "ok" };
     let duration = summary.duration_text.as_deref().unwrap_or("unknown");
 
-    let mut out = format!(
-        "{} dotnet restore: {} projects, {} errors, {} warnings ({})",
-        status_icon, summary.restored_projects, summary.errors, summary.warnings, duration
-    );
+    let mut warnings_section = String::new();
```
KuSh: nitpick: keep previous order to avoid noise

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -1108,37 +1140,51 @@ fn format_restore_output(
...
-        }
-        if warnings.len() > 10 {
-            out.push_str(&format!("  ... +{} more warnings\n", warnings.len() - 10));
-        }
-    }
+    let sep = if !warnings_section.is_empty() || !errors_section.is_empty() {
+        "---------------------------------------".to_string()
+    } else {
+        String::new()
+    };
```
KuSh, line 1192:
```suggestion
    let sep = if !warnings_section.is_empty() || !errors_section.is_empty() {
        "---------------------------------------"
    } else {
        ""
    };
```

KuSh (changes requested):
Mostly nitpicks, this is nearly ready to get merged. Thanks

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: sure :) moved it above

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: Kept the current source order on purpose. It mirrors the output order (warnings, errors, sep, verdict), so reading the function top-to-bottom matches what hits stdout. The whole point of #1574 is errors landing right above the verdict for `tail -N`, so I'd rather take the diff noise here than split source order from output order.

Happy to flip if you'd still prefer the smaller diff.

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: Done.

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: done :) moved above

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: Kept the current source order for the same reason as `format_build_output`. Source mirrors output

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: done

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: done

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: Kept the current source order for the same reason as `format_build_output`. Source mirrors output

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: fixed :)

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: done

src/cmds/dotnet/dotnet_cmd.rs
KuSh: Yeah I prefer smaller diff and better blame ;)

F0rty-Tw0: gotcha, done :)

src/cmds/dotnet/dotnet_cmd.rs
F0rty-Tw0: reordered after the small talk

KuSh (approved):

KuSh:
LGTM thanks !

PR #1578, opened 2026-04-28
feat(git): support git sparse-checkout subcommand

zackharley:
## Summary

Adds explicit handling for `git sparse-checkout` so the RTK hook system auto-rewrites it and the noisy progress output emitted by mutating subcommands collapses to a short `ok` line.

Today `git sparse-checkout` falls through to `GitCommands::Other` (the `external_subcommand` passthrough), which means:

- The hook system doesn't rewrite it (no `rtk_cmd` mapping in `discover/rules.rs`).
- Mutating subcommands stream raw `Updating files: 50% (5/10)…` progress noise into the LLM context with zero filtering.
- `list` output is unbounded — sparse-checkout configs in large monorepos can hit hundreds of patterns.

## What changed

| Subcommand | Behavior |
|---|---|
| `list` | Prints all patterns verbatim, capped at 50 lines with `... +N more`. Patterns are **never** reordered or merged — they are semantically meaningful (they decide which paths exist on disk). |
| `init` / `set` / `add` / `disable` / `reapply` | Collapses to `ok sparse-checkout <sub>` on success; surfaces stderr + git's real exit code on failure. |
| `check-rules`, no subcommand, any future/unknown subcommand | Passthrough — never block a future git release. |

The discover rule pattern is extended so the hook auto-rewriter picks `git sparse-checkout …` up:

```diff
-pattern: r"^(?:git|yadm)\s+(?:-[Cc]\s+\S+\s+)*(status|log|diff|show|...|stash|worktree)",
+pattern: r"^(?:git|yadm)\s+(?:-[Cc]\s+\S+\s+)*(status|log|diff|show|...|stash|worktree|sparse-checkout)",
```

## Token savings

Verified end-to-end against a real sparse repo with 200 cone patterns:

```
raw   git sparse-checkout list: 1700 chars
rtk   git sparse-checkout list:  397 chars   (-76.6%)
```

Well above the 60% bar from `CONTRIBUTING.md`.

For action subcommands the savings come from collapsing git's progress lines (and any warnings) into a single `ok sparse-checkout <sub>` token sequence, matching the existing pattern used by `add`/`commit`/`push`/`pull`/`stash`/`worktree`.

## Design notes

- **Patterns are preserved verbatim.** Sparse-checkout patterns are not advisory — they decide which paths get materialized in the working tree. Sorting, deduping, or grouping them would change which files an LLM sees on disk. The filter only trims trailing whitespace, drops blank lines, and applies a top-N cap with a `... +N more` summary.
- **Truncation threshold is 50.** Same order of magnitude as the truncation thresholds already used by `format_status_output`. Tuned high enough that typical project configs never truncate, but low enough that a 200+ pattern monorepo cone list compresses meaningfully.
- **Action list is a `const &[&str]`.** A small unit test asserts the contents to guard against accidental rename/typo.
- **Empty stdout on `list` → friendly message.** When a worktree has sparse-checkout enabled but no patterns yet, raw git prints nothing; we surface "No sparse-checkout patterns (sparse-checkout disabled or empty)" so the LLM doesn't see an unexplained empty response.
- **`check-rules` is intentionally passthrough.** Its output is structured (path + match status) and primarily consumed by scripts; reformatting risks breaking valid use cases for marginal token savings.
- **Exit codes propagate.** `git sparse-checkout list` on a non-sparse worktree exits 128 — RTK forwards that exit code unchanged for CI/CD compatibility.

## Manual verification

```
$ rtk git sparse-checkout init --cone
ok sparse-checkout init
$ rtk git sparse-checkout set src docs
ok sparse-checkout set
$ rtk git sparse-checkout list
docs
src
$ rtk git sparse-checkout add scripts
ok sparse-checkout add
$ rtk git sparse-checkout reapply
ok sparse-checkout reapply
$ rtk git sparse-checkout disable
ok sparse-checkout disable
$ echo "src/a.txt" | rtk git sparse-checkout check-rules    # passthrough
src/a.txt
$ rtk hook check "git sparse-checkout list"                 # auto-rewrite
rtk git sparse-checkout list
```

## Tests

Six new unit tests in `src/cmds/git/git.rs`:

- `test_filter_sparse_checkout_list_short` — happy path, no truncation
- `test_filter_sparse_checkout_list_drops_blank_and_trailing_whitespace` — Windows CRLF / trailing-space tolerance
- `test_filter_sparse_checkout_list_truncates_when_over_limit` — 120 patterns → 50 + `... +70 more`
- `test_filter_sparse_checkout_list_at_exact_limit` — boundary: exactly 50 must not get a `+0 more` suffix
- `test_filter_sparse_checkout_list_preserves_negation_patterns` — non-cone `!/path/` patterns preserved verbatim and in order
- `test_sparse_checkout_action_subcommands_are_recognized` — guards the `SPARSE_CHECKOUT_ACTION_SUBCOMMANDS` constant against accidental rename

All 1693 existing tests still pass. `cargo fmt --all --check && cargo clippy --all-targets && cargo test` is green.

## Docs

- `README.md`: added two `rtk git sparse-checkout …` examples to the Git section.
- `src/cmds/git/README.md`: documented routing rules under "Specifics".

## Scope

Single feature, single PR, per `CONTRIBUTING.md` scope rules. No drive-by refactors, no `CHANGELOG.md` edit (release-please handles it from the conventional commit).

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -23,6 +23,7 @@ pub enum GitCommand {
     Fetch,
     Stash { subcommand: Option<String> },
     Worktree,
+    SparseCheckout { subcommand: Option<String> },
 }
 
 /// Create a git Command with global options (e.g. -C, -c, --git-dir, --work-tree)
@@ -57,6 +58,9 @@ pub fn run(
             run_stash(subcommand.as_deref(), args, verbose, global_args)
         }
         GitCommand::Worktree => run_worktree(args, verbose, global_args),
+        GitCommand::SparseCheckout { subcommand } => {
+            run_sparse_checkout(subcommand.as_deref(), args, verbose, global_args)
+        }
     }
 }
 
@@ -1669,6 +1673,146 @@ fn filter_worktree_list(output: &str) -> String {
     result.join("\n")
 }
 
+/// Maximum number of sparse-checkout patterns to print before truncating to a
+/// `... +N more` summary. Mirrors the truncation thresholds used by
+/// `format_status_output` so very large monorepo cone lists don't blow up the
+/// LLM context.
+const SPARSE_CHECKOUT_LIST_MAX_LINES: usize = 50;
+
+/// Subcommands of `git sparse-checkout` that mutate state.
+///
+/// On success they typically produce only progress noise (`Updating files: ...`)
+/// which we collapse to a short `ok` line. Unknown subcommands fall through to
+/// passthrough so we never block a future git release.
+const SPARSE_CHECKOUT_ACTION_SUBCOMMANDS: &[&str] = &[
+    "init", "set", "add", "disable", "reapply",
+];
+
+fn run_sparse_checkout(
+    subcommand: Option<&str>,
+    args: &[String],
+    verbose: u8,
+    global_args: &[String],
+) -> Result<i32> {
+    let timer = tracking::TimedExecution::start();
+
+    if verbose > 0 {
+        eprintln!("git sparse-checkout {:?}", subcommand);
+    }
+
+    match subcommand {
+        Some("list") => {
+            let mut cmd = git_cmd(global_args);
+            cmd.args(["sparse-checkout", "list"]);
+            for arg in args {
+                cmd.arg(arg);
+            }
+            let result = exec_capture(&mut cmd)
+                .context("Failed to run git sparse-checkout list")?;
+
+            if !result.success() {
+                eprintln!("FAILED: git sparse-checkout list");
+                if !result.stderr.trim().is_empty() {
+                    eprintln!("{}", result.stderr);
+                }
+                return Ok(result.exit_code);
+            }
+
+            let filtered = if result.stdout.trim().is_empty() {
+                "No sparse-checkout patterns (sparse-checkout disabled or empty)".to_string()
+            } else {
+                filter_sparse_checkout_list(&result.stdout, SPARSE_CHECKOUT_LIST_MAX_LINES)
+            };
+            println!("{}", filtered);
+            timer.track(
+                "git sparse-checkout list",
+                "rtk git sparse-checkout list",
+                &result.stdout,
+                &filtered,
+            );
+        }
+        Some(sub) if SPARSE_CHECKOUT_ACTION_SUBCOMMANDS.contains(&sub) => {
+            let mut cmd = git_cmd(global_args);
+            cmd.args(["sparse-checkout", sub]);
+            for arg in args {
+                cmd.arg(arg);
+            }
+            let result = exec_capture(&mut cmd)
+                .context("Failed to run git sparse-checkout")?;
+            let combined = result.combined();
+
+            let msg = if result.success() {
+                let msg = format!("ok sparse-checkout {}", sub);
+                println!("{}", msg);
+                msg
+            } else {
+                eprintln!("FAILED: git sparse-checkout {}", sub);
+                if !result.stderr.trim().is_empty() {
+                    eprintln!("{}", result.stderr);
+                }
+                combined.clone()
+            };
+
+            timer.track(
+                &format!("git sparse-checkout {}", sub),
+                &format!("rtk git sparse-checkout {}", sub),
+                &combined,
+                &msg,
+            );
+
+            if !result.success() {
+                return Ok(result.exit_code);
+            }
+        }
+        Some(sub) => {
+            // Unrecognized subcommand (e.g. `check-rules`, future additions):
+            // passthrough unfiltered so we never break valid git usage.
+            let mut passthrough_args: Vec<OsString> =
+                Vec::with_capacity(args.len() + 2);
+            passthrough_args.push(OsString::from("sparse-checkout"));
+            passthrough_args.push(OsString::from(sub));
+            for arg in args {
+                passthrough_args.push(OsString::from(arg));
+            }
+            let _ = timer; // tracking handled inside run_passthrough
+            return run_passthrough(&passthrough_args, global_args, verbose);
+        }
+        None => {
+            // `git sparse-checkout` with no subcommand prints usage to stderr
+            // and exits non-zero. Pass through so the user sees the real help.
+            let passthrough_args: Vec<OsString> = vec![OsString::from("sparse-checkout")];
+            let _ = timer;
+            return run_passthrough(&passthrough_args, global_args, verbose);
+        }
+    }
+
+    Ok(0)
+}
+
+/// Compress `git sparse-checkout list` output: trim, drop blank lines, and cap
+/// at `max_lines` patterns followed by a `... +N more` summary. Each pattern is
+/// preserved verbatim — sparse-checkout patterns are semantically meaningful
+/// (they decide which paths exist on disk), so we never reorder or merge them.
+pub(crate) fn filter_sparse_checkout_list(output: &str, max_lines: usize) -> String {
+    let patterns: Vec<&str> = output
+        .lines()
+        .map(str::trim_end)
+        .filter(|l| !l.is_empty())
+        .collect();
+
+    if patterns.len() <= max_lines {
+        return patterns.join("\n");
+    }
+
+    let mut out: Vec<String> = patterns
+        .iter()
+        .take(max_lines)
+        .map(|s| (*s).to_string())
+        .collect();
+    out.push(format!("... +{} more", patterns.len() - max_lines));
+    out.join("\n")
+}
+
 /// Runs an unsupported git subcommand by passing it through directly
 pub fn run_passthrough(args: &[OsString], global_args: &[String], verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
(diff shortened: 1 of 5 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟢 **Risk** | `low` |

### Summary

Adds explicit handling for the `git sparse-checkout` subcommand in the RTK hook system. The `list` subcommand output is capped at 50 lines with a truncation summary, mutating subcommands (`init`, `set`, `add`, `disable`, `reapply`) collapse noisy progress output to a short `ok` line, and unrecognized subcommands pass through unchanged. This achieves ~77% token savings for large sparse-checkout configs.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [x] Docs updated

---

src/cmds/git/git.rs
```diff
@@ -1669,6 +1673,146 @@ fn filter_worktree_list(output: &str) -> String {
     result.join("\n")
 }
 
+/// Maximum number of sparse-checkout patterns to print before truncating to a
+/// `... +N more` summary. Mirrors the truncation thresholds used by
```
zackharley: 🎨 **Code Style**: The doc claims `SPARSE_CHECKOUT_LIST_MAX_LINES = 50` 'mirrors the truncation thresholds used by `format_status_output`'. In reality, `format_status_output` reads its limits from `config::limits()` — `status_max_files` (default 15) and `status_max_untracked` (default 10), both user-configurable. This new constant is hard-coded at 50 and ignores config, so neither the value nor the configuration knob matches. The PR description more accurately calls this 'same order of magnitude'. As written, the comment will mislead a future reader who bumps `status_max_files` into thinking sparse-checkout truncation is also affected — it isn't.

**Suggestion:** Tone down the doc to match reality, optionally noting that the fixed-vs-configurable divergence is intentional:

```suggestion
/// Maximum number of sparse-checkout patterns to print before truncating to a
/// `... +N more` summary. Same order of magnitude as `format_status_output`'s
/// configurable thresholds (`status_max_files`, `status_max_untracked`, defaults
/// 15/10) but intentionally fixed here — typical project configs never truncate,
/// while a 200+ pattern monorepo cone list compresses meaningfully without
/// blowing up the LLM context.
```

src/cmds/git/git.rs
```diff
@@ -1669,6 +1673,146 @@ fn filter_worktree_list(output: &str) -> String {
...
+                &filtered,
+            );
+        }
+        Some(sub) if SPARSE_CHECKOUT_ACTION_SUBCOMMANDS.contains(&sub) => {
+            let mut cmd = git_cmd(global_args);
+            cmd.args(["sparse-checkout", sub]);
+            for arg in args {
+                cmd.arg(arg);
+            }
+            let result = exec_capture(&mut cmd)
```
zackharley, line 1771: 🐛 **Bug**: `git sparse-checkout set --stdin` and `git sparse-checkout add --stdin` are documented git modes that read patterns from stdin until EOF. The new action branch routes mutating subcommands through `core::stream::exec_capture`, which unconditionally calls `cmd.stdin(Stdio::null())` (src/core/stream.rs:496). Before this PR, `git sparse-checkout` fell through to `GitCommand::Other` → `run_passthrough`, which uses `.status()` and inherits stdin. After this PR, `cat patterns.txt | rtk git sparse-checkout set --stdin` will see immediate EOF and either error or silently apply an empty pattern set — effectively wiping the user's sparse-checkout config while still reporting success. The discover rule change in this same PR makes the hook auto-rewrite `git sparse-checkout` to `rtk git sparse-checkout`, so this regression becomes automatically reachable for any hook-installed user. The PR's manual verification only covers the positional-argument form.

**Suggestion:** Detect `--stdin` for `set`/`add` and route to `run_passthrough` so stdin is inherited (or add a capture helper that inherits stdin). Add a regression test asserting `--stdin` routes through passthrough.

```suggestion
        Some(sub) if SPARSE_CHECKOUT_ACTION_SUBCOMMANDS.contains(&sub) => {
            // `set --stdin` / `add --stdin` read patterns from stdin; exec_capture
            // would null stdin and silently clear the pattern set. Fall through
            // to passthrough so stdin is inherited.
            if (sub == "set" || sub == "add") && args.iter().any(|a| a == "--stdin") {
                let mut passthrough_args: Vec<OsString> =
                    Vec::with_capacity(args.len() + 2);
                passthrough_args.push(OsString::from("sparse-checkout"));
                passthrough_args.push(OsString::from(sub));
                for arg in args {
                    passthrough_args.push(OsString::from(arg));
                }
                return run_passthrough(&passthrough_args, global_args, verbose);
            }
            let mut cmd = git_cmd(global_args);
```

src/cmds/git/git.rs
```diff
@@ -1669,6 +1673,146 @@ fn filter_worktree_list(output: &str) -> String {
...
+            // Unrecognized subcommand (e.g. `check-rules`, future additions):
+            // passthrough unfiltered so we never break valid git usage.
+            let mut passthrough_args: Vec<OsString> =
+                Vec::with_capacity(args.len() + 2);
+            passthrough_args.push(OsString::from("sparse-checkout"));
+            passthrough_args.push(OsString::from(sub));
+            for arg in args {
+                passthrough_args.push(OsString::from(arg));
+            }
+            let _ = timer; // tracking handled inside run_passthrough
```
zackharley: 🎨 **Code Style**: `tracking::TimedExecution` is a plain struct wrapping an `Instant` with no `Drop` impl (src/core/tracking.rs:1305-1329), so dropping it does nothing. `let _ = timer;` at lines 1777 and 1784 has zero runtime effect and does not silence any unused-variable warning (the binding is consumed in the other match arms, which is enough). The accompanying comment 'tracking handled inside run_passthrough' is correct, but the discard line itself reads like an explicit consume/release of a resource — a foreign idiom in this file (run_stash, run_commit, run_worktree don't use this pattern).

**Suggestion:** Move `let timer = tracking::TimedExecution::start();` into the `Some("list")` and action-subcommand arms, and drop the two `let _ = timer;` lines. The `verbose` print can be hoisted before the match if you want it unconditional. Example shape:

```rust
match subcommand {
    Some("list") => {
        let timer = tracking::TimedExecution::start();
        // ...existing list arm...
    }
    Some(sub) if SPARSE_CHECKOUT_ACTION_SUBCOMMANDS.contains(&sub) => {
        let timer = tracking::TimedExecution::start();
        // ...existing action arm...
    }
    Some(sub) => {
        // tracking handled inside run_passthrough
        // ...build passthrough_args and return run_passthrough(...)
    }
    None => {
        // tracking handled inside run_passthrough
        // ...return run_passthrough(...)
    }
}
```

src/cmds/git/git.rs
```diff
@@ -1669,6 +1673,146 @@ fn filter_worktree_list(output: &str) -> String {
...
+    }
+
+    Ok(0)
+}
+
+/// Compress `git sparse-checkout list` output: trim, drop blank lines, and cap
+/// at `max_lines` patterns followed by a `... +N more` summary. Each pattern is
+/// preserved verbatim — sparse-checkout patterns are semantically meaningful
+/// (they decide which paths exist on disk), so we never reorder or merge them.
+pub(crate) fn filter_sparse_checkout_list(output: &str, max_lines: usize) -> String {
```
zackharley: 🎨 **Code Style**: `filter_sparse_checkout_list` is declared `pub(crate)`, but the only callers are `run_sparse_checkout` and the in-file test module — there are no cross-module imports. The two stylistically closest siblings, `filter_stash_list` (line 1564) and `filter_worktree_list` (line 1649), are private. The only `pub(crate)` filter in this file is `filter_log_output`, which is genuinely consumed by `src/cmds/system/pipe_cmd.rs`. Granting `pub(crate)` here breaks the file's local convention and widens API surface for no reason.

**Suggestion:** Drop the `pub(crate)` qualifier so it lines up with the rest of the file's private helpers:

```suggestion
fn filter_sparse_checkout_list(output: &str, max_lines: usize) -> String {
```

src/cmds/git/git.rs
```diff
@@ -2023,6 +2167,74 @@ mod tests {
...
+        let result = filter_sparse_checkout_list(output, 50);
+        assert_eq!(result, "/*\n!/docs/\n!/src/lib/");
+    }
+
+    #[test]
+    fn test_sparse_checkout_action_subcommands_are_recognized() {
+        // Guard against accidental rename: each action must produce an `ok`
+        // success message in run_sparse_checkout. The list also drives the
+        // dispatch in the match arm above.
+        for sub in SPARSE_CHECKOUT_ACTION_SUBCOMMANDS {
```
zackharley: 💡 **Improvement**: The current test checks that every existing entry is one of the expected action names, but it would still pass if an action were accidentally removed from `SPARSE_CHECKOUT_ACTION_SUBCOMMANDS` — or even if the slice were emptied entirely. Since this list drives dispatch behavior, a direct equality assertion would better guard additions, removals, typos, and ordering changes.

**Suggestion:** Consider asserting the whole slice against the expected actions.

```suggestion
        assert_eq!(
            SPARSE_CHECKOUT_ACTION_SUBCOMMANDS,
            &["init", "set", "add", "disable", "reapply"]
        );
```

src/main.rs
zackharley, line 1: 💡 **Improvement**: The new hyphenated nested `sparse-checkout` command adds parsing behavior similar to other git command routing cases already covered by clap regression tests in this file. Without targeted coverage, regressions in splitting the optional `subcommand` from trailing `args` could slip through even if the git execution tests continue to pass.

**Suggestion:** Consider adding tests under the existing `#[cfg(test)] mod tests` for representative forms such as `rtk git sparse-checkout list`, `rtk git sparse-checkout set --stdin`, and `rtk git sparse-checkout check-rules -z`, asserting the expected `GitCommands::SparseCheckout { subcommand, args }` values.

src/cmds/git/git.rs
zackharley: Fixed in 41ea67d. Toned down the doc to drop the 'mirrors thresholds' claim — agreed it was misleading since `format_status_output` reads `status_max_files`/`status_max_untracked` from `core::config` (defaults 15/10) while this constant is intentionally fixed at 50. New wording calls out the 'same order of magnitude' / 'intentionally fixed' relationship explicitly so a future reader bumping the status limits doesn't expect this to follow.

src/cmds/git/git.rs
zackharley: Fixed in 41ea67d. Confirmed the bug end-to-end (`echo patterns | rtk git sparse-checkout set --stdin` reported `ok` while wiping the user's pattern set) and added the `sparse_checkout_uses_stdin()` predicate to route `set`/`add` + `--stdin` to `run_passthrough` so stdin is inherited via `.status()`.

The match arm guard now reads:
```rust
Some(sub)
    if SPARSE_CHECKOUT_ACTION_SUBCOMMANDS.contains(&sub)
        && !sparse_checkout_uses_stdin(sub, args) =>
```
with the `--stdin` form falling through to the `Some(sub)` passthrough arm (same path now also carries `check-rules`, which also reads stdin).

Verified end-to-end against a real sparse repo:
```
$ printf 'docs\nscripts\n' | rtk git sparse-checkout set --stdin
$ rtk git sparse-checkout list
docs
scripts          ← actually applied (was empty before the fix)
```
Also added two unit tests (`test_sparse_checkout_uses_stdin_only_for_set_and_add`, `test_git_sparse_checkout_set_stdin_flag_lands_in_args`) to lock in both the predicate and the clap parsing contract.

src/cmds/git/git.rs
zackharley: Fixed in 41ea67d. Confirmed `TimedExecution` has no `Drop` impl (just an `Instant` field at `core/tracking.rs:1305-1329`), so the discards were genuinely no-ops. Moved `let timer = TimedExecution::start();` into the two arms that actually call `timer.track()` (the `list` arm and the action arm), and dropped the two `let _ = timer;` lines. The `verbose` print stays at the top of `run_sparse_checkout` since it's unconditional. Now matches the shape used by `run_stash`/`run_worktree`.

src/cmds/git/git.rs
zackharley: Fixed in 41ea67d. Dropped the `pub(crate)` qualifier — agreed there's no cross-module caller, so `filter_sparse_checkout_list` should be private like `filter_stash_list` and `filter_worktree_list`. (`filter_log_output` keeps `pub(crate)` because `cmds/system/pipe_cmd.rs` actually imports it.)

src/cmds/git/git.rs
zackharley: Fixed in 41ea67d. Replaced the per-element membership check with a full-slice equality assertion:

```rust
assert_eq!(
    SPARSE_CHECKOUT_ACTION_SUBCOMMANDS,
    &["init", "set", "add", "disable", "reapply"]
);
```
Now catches accidental removals, additions, typos, and ordering changes — the old form would have passed for the empty slice.

src/main.rs
zackharley: Done in 41ea67d. Added 4 `Cli::try_parse_from` tests in `main.rs` covering the optional-subcommand + trailing-args split for the hyphenated `sparse-checkout` variant:

- `test_git_sparse_checkout_list_no_args` — bare `list`, asserts `args.is_empty()`
- `test_git_sparse_checkout_set_with_patterns` — positional patterns land in `args`
- `test_git_sparse_checkout_set_stdin_flag_lands_in_args` — `--stdin` reaches `args` (so the runtime `--stdin` carve-out can detect it)
- `test_git_sparse_checkout_check_rules_with_z_flag` — hyphenated `check-rules` parses with hyphen-prefixed trailing args
- `test_git_sparse_checkout_no_subcommand_is_none` — bare `rtk git sparse-checkout` parses with `subcommand=None` so the runtime can passthrough to git's usage error

While writing these I also confirmed via a small clap reproducer that `rtk git sparse-checkout -z` (hyphen-prefixed first positional, no subcommand value) is rejected by clap and falls through to `run_fallback` (raw exec) — so my `None` arm only triggers in the bare-no-args case, which is what the last test pins down.

PR #1587, opened 2026-04-28
fix(curl): emit valid JSON envelope when truncating large responses

### LINKED ISSUE #1536: P0 bug(rtk curl): large JSON responses truncated mid-stream producing invalid JSON, breaks all API agent pipelines
## Summary

When `rtk curl` receives a JSON response larger than the internal `passthrough_max_chars` threshold (~510 chars), it truncates the raw response body **mid-stream** — inserting `... (N bytes total)` directly inside a JSON string value. This produces **invalid, unparseable JSON** on stdout, silently breaking every agent pipeline that pipes `rtk curl` to a JSON parser.

This is a **distinct bug from #1152** (which covers schema conversion of small JSON). This truncation happens at a lower layer — the raw body is cut before schema conversion can run — and is far more destructive because it produces invalid JSON rather than schema-typed JSON.

## Environment

- RTK version: 0.37.2
- OS: macOS Darwin 25.4.0
- Hook mode: Claude Code PreToolUse hook (`rtk hook claude`)

## Reproduction

```bash
# Small JSON response (~200 bytes) — passes through fine:
rtk curl -s 'http://127.0.0.1:3101/api/health'
# stdout: {"status":"ok","version":"0.3.1","deploymentMode":"local_trusted",...}
# Valid JSON ✅

# Large JSON response (~4000 bytes) — truncated mid-stream:
rtk curl -s 'http://127.0.0.1:3101/api/companies/COMPANY_ID/issues?status=in_progress&limit=2'
# stdout: [{"id":"9366a21a-...","title":"...","description":"Infrastructure scope...\n\n**What to do:**... (4048 bytes total)
#         [full output: ~/Library/Application Support/rtk/tee/1777194358_curl.log]
# INVALID JSON ❌

# Confirm breakage:
rtk curl -s '...' | python3 -c "import sys,json; json.load(sys.stdin)"
# json.decoder.JSONDecodeError: Invalid control character at: line 1 column 521 (char 520)
```

## What actually appears on stdout

```
[{"id":"9366a21a-...","companyId":"...","title":"DEA-187 follow-up: ...","description":"Infrastructure scope extracted from DEA-187 (Prometheus alerting rules).\n\n**Selected approach: Option A**\n\n**What to do:**... (4048 bytes total)
[full output: ~/Library/Application Support/rtk/tee/1777194358_curl.log]
```

The truncation marker `... (4048 bytes total)` is inserted **inside a JSON string value** (the `description` field), immediately breaking JSON syntax. The full response is only in the tee log file; stdout is non-recoverable.

## Root Cause

The `passthrough_max_chars` limit in `LimitsConfig` is applied to the raw HTTP body before any JSON processing. When the body exceeds ~510 chars, the body is cut and the truncation marker is appended mid-stream — directly into whatever JSON string was being written, regardless of structure.

This is **different from the schema conversion path** in `src/cmds/cloud/curl_cmd.rs` (tracked in #1152). That path converts small-enough JSON to schema form. This truncation fires when the response is too large to enter the schema path at all.

## Impact

**P0 — breaks all agent JSON API pipelines:**

- `rtk curl ... | python3 -c "import json,sys; json.load(sys.stdin)"` → `JSONDecodeError`
- `rtk curl ... | jq '.'` → `parse error (Invalid numeric literal at line 1, column 521)`
- Any agent using the Claude Code hook (`rtk hook claude`) that calls a local REST API with non-trivial responses gets **silent data loss** — the pipeline appears to run but yields no parseable data
- Currently blocks RTK pilot rollout to autonomous coding agents — agents cannot call their orchestration APIs (Paperclip, local services) via the hook

## Expected Behaviour

When `rtk curl` produces JSON output that exceeds `passthrough_max_chars`:

**Option A (preferred):** Pass JSON through unmodified. JSON is already structured and compact; mid-stream truncation provides zero token savings and silently corrupts the pipeline.

**Option B:** Emit a valid JSON document describing the truncation:
```json
{"_rtk": {"truncated": true, "total_bytes": 4048, "tee_log": "~/Library/Application Support/rtk/tee/1777194358_curl.log"}}
```

**Option C:** Respect `passthrough_max_chars = 0` or the `no_truncation = true` flag proposed in #1313.

## Workaround

```bash
# Bypass rtk filtering entirely (loses token savings):
rtk proxy curl -s 'http://127.0.0.1:3101/api/.../issues?status=in_progress'
```

## Related Issues

- #1152 — JSON schema conversion destroys values (schema conversion path, distinct bug)
- #1313 — Feature request: lossless-only mode (overlapping solution space)
- #219 — Config for curl filter bypass (resolved-pending-close, solution would help)
- #1311 — curl truncation breaks piped bash install (CLOSED, same truncation mechanism)

truffle-dev:
Closes #1536.

Today, on any curl response above 500 bytes the truncation produces
output like:

```
{"items":[{"id":1,"name":"alpha"},{"id":2,"name"... (1234 bytes total)
[full output: ~/.local/share/rtk/curl-...log]
```

That is no longer valid JSON. Anything piping rtk through `jq`, parsing
structured tool output, or feeding the response back to an LLM hits a
parse error on every response above the threshold. Reproducer in #1536
(`rtk curl https://jsonplaceholder.typicode.com/users` → `jq` fails).

## Approach

I considered three options and went with the smallest one that fits the
existing design:

- **A — passthrough (PR #1016):** drop truncation entirely for JSON. This
  regresses the token-savings intent of the simplified filter that landed
  in #1310, and makes large responses expensive again for the LLM caller.
- **B — JSON envelope (this PR):** when the response is JSON-shaped and
  above the threshold, replace the body with a small envelope:
  ```json
  {"_rtk":{"truncated":true,"total_bytes":1234,"tee_log":"~/path"}}
  ```
  Downstream agents can parse it, branch on `_rtk.truncated`, and read
  the full body from the tee log. Keeps the token-saving guarantee.
- **C — config flag:** adds surface area for a behavior that should
  always be correct.

The envelope path also folds the tee log into the JSON object and sets
`tee_hint = None`, so callers print exactly one valid JSON document and
not a separate `[full output: ...]` trailer line.

Non-JSON responses keep the existing prefix-with-byte-count format. The
shape detection is a one-byte check (`{` or `[`); responses curl produces
that aren't actually JSON but happen to start with those bytes are
extremely rare and will surface as a parse error on `_rtk` lookup, which
is at worst the same outcome as today.

## Verification

Stash-bisect against this branch's two new tests:

Without the fix (existing implementation, new tests applied):
```
test test_filter_curl_large_json_object_emits_valid_envelope ... FAILED
test test_filter_curl_large_json_array_emits_valid_envelope  ... FAILED
  panicked: 'envelope must be valid JSON for downstream agents to parse:
  Error("EOF while parsing a string", line: 1, column: 521)'
```

With the fix:
```
test result: ok. 7 passed; 0 failed; 0 ignored
```

The four pre-existing tests (`test_filter_curl_json_small_no_tee_hint`,
`test_filter_curl_non_json`, `test_filter_curl_long_output_truncated`,
`test_filter_curl_multibyte_boundary`, `test_filter_curl_exact_500_bytes`)
all still pass — non-JSON truncation behavior is unchanged.

### DIFF
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -58,20 +58,51 @@ fn filter_curl_output(raw: &str) -> FilterResult {
     let trimmed = raw.trim();
     let tee_hint = force_tee_hint(raw, "curl");
 
-    // If the output is too long and we have a tee hint, truncate the output.
-    let content = if trimmed.len() >= MAX_RESPONSE_SIZE && tee_hint.is_some() {
-        let mut end = MAX_RESPONSE_SIZE;
-        // Ensure we don't cut in the middle of a UTF-8 character.
-        // .len() counts bytes, not chars.
-        while !trimmed.is_char_boundary(end) {
-            end -= 1;
-        }
-        format!("{}... ({} bytes total)", &trimmed[..end], trimmed.len())
-    } else {
-        trimmed.to_string()
-    };
+    if trimmed.len() < MAX_RESPONSE_SIZE || tee_hint.is_none() {
+        return FilterResult {
+            content: trimmed.to_string(),
+            tee_hint,
+        };
+    }
 
+    let total_bytes = trimmed.len();
+
+    // For JSON-shaped responses, mid-stream truncation produces invalid JSON
+    // that downstream agents can't parse. Emit a parseable envelope instead and
+    // fold the tee path into it so the caller doesn't print a separate trailer.
+    if is_json_shape(trimmed) {
+        let tee_log = tee_hint
+            .as_deref()
+            .and_then(|h| h.strip_prefix("[full output: "))
+            .and_then(|h| h.strip_suffix(']'))
+            .unwrap_or("");
+        let envelope = serde_json::json!({
+            "_rtk": {
+                "truncated": true,
+                "total_bytes": total_bytes,
+                "tee_log": tee_log,
+            }
+        })
+        .to_string();
+        return FilterResult {
+            content: envelope,
+            tee_hint: None,
+        };
+    }
+
+    // Non-JSON: keep the existing prefix-with-byte-count truncation.
+    let mut end = MAX_RESPONSE_SIZE;
+    // Ensure we don't cut in the middle of a UTF-8 character.
+    // .len() counts bytes, not chars.
+    while !trimmed.is_char_boundary(end) {
+        end -= 1;
+    }
+    let content = format!("{}... ({} bytes total)", &trimmed[..end], total_bytes);
     FilterResult { content, tee_hint }
+}
+
+fn is_json_shape(s: &str) -> bool {
+    matches!(s.as_bytes().first(), Some(b'{' | b'['))
 }
 
 struct FilterResult {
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes invalid JSON output when curl responses exceed the 500-byte truncation threshold. Instead of mid-stream truncation that breaks JSON parsing, large JSON responses now emit a valid JSON envelope containing truncation metadata and a path to the full output log.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1536

---

KuSh:
rtk aims is to reduce LLM's tokens usage, passing content to other tools doesn't consume tokens and shouldn't be impacted by rtk. it should passthrough for such cases. I'm closing this PR in favor of https://github.com/rtk-ai/rtk/pull/1602

PR #1602, opened 2026-04-29
fix(curl): JSON passthrough + IsTerminal gate to prevent invalid JSON output

### LINKED ISSUE #1536: P0 bug(rtk curl): large JSON responses truncated mid-stream producing invalid JSON, breaks all API agent pipelines

pszymkowiak:
## Summary

Fixes #1536 — `rtk curl` was truncating any response ≥500 bytes mid-stream, inserting `... (N bytes total)` directly inside JSON string values and producing invalid, unparseable output. This silently broke every agent pipeline that piped `rtk curl` to `jq` / `python -m json.tool` / similar, and was reproducible on the issue's public repro (`https://jsonplaceholder.typicode.com/posts`).

Also closes the related silent-data-loss path tracked in #1282 — `curl > file` and `curl | tool` were truncated even though `curl -o file` worked, because the rewrite layer captured curl's stdout regardless of `isatty(stdout)`.

## Changes

`src/cmds/cloud/curl_cmd.rs`

- Added `use std::io::IsTerminal;` and `let is_tty = std::io::stdout().is_terminal();` in `run()`
- `filter_curl_output(raw, is_tty)` now skips truncation when:
  - body looks like a JSON document (starts/ends with `{}` or `[]`) — matters for downstream parsers
  - stdout is not a terminal — matters for pipes / shell redirects
- Tee file is still written in those cases (preserves observability), but the hint line is suppressed so it never leaks into pipes
- The legacy 500-byte truncation + `[full output: …]` hint is preserved for human TTY mode on non-JSON bodies

## Tests

- Existing 5 unit tests updated to pass `is_tty=true` (preserve their TTY-mode assertions)
- 4 new unit tests:
  - `test_filter_curl_large_json_object_passthrough` — TTY + JSON object > 500 B → no truncation
  - `test_filter_curl_large_json_array_passthrough` — TTY + JSON array > 500 B → no truncation
  - `test_filter_curl_pipe_no_truncation_for_non_json` — pipe + plain text 1000 B → full body
  - `test_filter_curl_pipe_no_truncation_for_json` — pipe + JSON > 500 B → full body, no hint

`cargo fmt --all && cargo clippy --all-targets && cargo test --all` → 1691 passed, 6 ignored.

## Verification

End-to-end repro from the issue, before vs after:

```bash
# Before (master 0.37.2)
$ rtk curl -s 'https://jsonplaceholder.typicode.com/posts' | python3 -c 'import sys,json; json.load(sys.stdin)'
json.decoder.JSONDecodeError: Invalid control character at: line 12 column 153 (char 523)

# After (this PR)
$ rtk curl -s 'https://jsonplaceholder.typicode.com/posts' | python3 -c 'import sys,json; print(len(json.load(sys.stdin)))'
100
```

## Notes on related PRs

- #1016 was an earlier attempt at the JSON-passthrough idea but is currently in conflict with develop (the `curl_cmd.rs` shape it targeted has been refactored away). It also bundles an unrelated `extract_schema` quoting fix in `json_cmd.rs`. I'll comment on #1016 to either rebase its `json_cmd.rs` part as a separate scoped PR or close it — happy to follow whatever you prefer.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` green
- [x] End-to-end smoke test against `https://jsonplaceholder.typicode.com/posts` returns parseable JSON
- [ ] CI matrix (Ubuntu / macOS / Windows) green
- [ ] Manual check: `rtk curl` to a non-JSON endpoint in a TTY still shows the truncated form + tee hint

### DIFF
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -1,15 +1,20 @@
-//! Runs curl and applies a simple truncation with tee hint if the output is too long.
+//! Runs curl and condenses long output for human consumption.
+//!
+//! For pipes / redirects (non-TTY) and JSON bodies the full response is passed
+//! through unchanged — truncating mid-stream would break downstream parsers.
+//! The condensed-form-with-tee-hint path is reserved for non-JSON bodies on
+//! a real terminal where a human reads the output and the tee file gives the
+//! LLM a way to recover the raw response.
 
 use crate::core::tee::force_tee_hint;
 use crate::core::tracking;
 use crate::core::{stream::exec_capture, utils::resolved_command};
 use anyhow::{Context, Result};
+use std::borrow::Cow;
 use std::io::IsTerminal;
 
 const MAX_RESPONSE_SIZE: usize = 500;
 
-/// Not using run_filtered: on failure, curl can return HTML error pages (404, 500)
-/// that the JSON schema filter would mangle. The early exit skips filtering entirely.
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
     let mut cmd = resolved_command("curl");
@@ -25,7 +30,8 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
 
     let result = exec_capture(&mut cmd).context("Failed to run curl")?;
 
-    // Early exit: don't feed HTTP error bodies (HTML 404 etc.) through JSON schema filter
+    // Skip filtering on failure: curl can return HTML error bodies that would
+    // be misleading to summarize, and we want the real exit code surfaced.
     if !result.success() {
         let msg = if result.stderr.trim().is_empty() {
             result.stdout.trim().to_string()
@@ -36,13 +42,13 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         return Ok(result.exit_code);
     }
 
-    let raw = result.stdout.clone();
-
+    let exit_code = result.exit_code;
+    let raw = result.stdout;
     let is_tty = std::io::stdout().is_terminal();
-    let result = filter_curl_output(&result.stdout, is_tty);
+    let filtered = filter_curl_output(&raw, is_tty);
 
-    println!("{}", result.content);
-    if let Some(hint) = &result.tee_hint {
+    println!("{}", filtered.content);
+    if let Some(hint) = &filtered.tee_hint {
         println!("{}", hint);
     }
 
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

Fixes JSON output truncation in `rtk curl` where responses >= 500 bytes were being cut mid-stream, producing invalid JSON that broke downstream parsers like `jq` and `python -m json.tool`. Adds an `IsTerminal` gate so pipes/redirects always receive full output, and skips truncation for JSON-shaped bodies even on TTY. Legacy truncation behavior is preserved only for non-JSON content displayed in a terminal.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1536, #1282

---

src/cmds/cloud/curl_cmd.rs
```diff
@@ -35,47 +42,79 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
...
+            tee_hint: None,
+        };
     };
 
-    FilterResult { content, tee_hint }
+    let mut end = MAX_RESPONSE_SIZE;
+    // Don't cut in the middle of a UTF-8 character — .len() counts bytes.
+    while !trimmed.is_char_boundary(end) {
+        end -= 1;
+    }
```
KuSh, line 105:
```suggestion
    let end = trimmed.floor_char_boundary(MAX_RESPONSE_SIZE);
```

KuSh (approved):
LGTM!

src/cmds/cloud/curl_cmd.rs
KuSh: My bad it counts in bytes not in char count, forget about that

PR #1603, opened 2026-04-29
feat(discover): wire xcodebuild into rewriter registry (#1551)

### LINKED ISSUE #1551: rewriter: auto-rewrite xcodebuild → rtk xcodebuild (filter exists since v0.30.0)
## Context

Issue #376 was closed COMPLETED in v0.30.0 with a `xcodebuild.toml` TOML filter — `rtk xcodebuild build` correctly strips `CompileSwift` lines etc.

However, the **rewriter registry** (`rtk rewrite`) does not auto-rewrite raw `xcodebuild` invocations to `rtk xcodebuild`. This means when `xcodebuild` is called from a Claude Code Bash hook (or any wrapper that uses `rtk rewrite`), the command falls through unchanged and the filter never fires.

## Reproduction (rtk 0.34.2)

```
$ rtk rewrite "xcodebuild build -project Foo.xcodeproj"; echo $?
1                                       # exit 1 = "no RTK equivalent" → passthrough

$ rtk rewrite "git status"; echo $?
rtk git status
0                                       # exit 0 = rewrite applied

$ rtk xcodebuild build -project Foo.xcodeproj   # explicit prefix works fine
... (filtered output)
```

## Impact

From a single Claude Code user's `rtk discover` over 30 days:

```
xcodebuild   3,231 invocations    (top unhandled command after ssh)
```

That's ~3,231 calls/month per user that bypass the v0.30.0 filter purely because the rewriter doesn't add the prefix.

## Suggested fix

Add `xcodebuild` (and probably `xcrun`, `swift`-already-supported, etc.) to the rewriter registry in `src/discover/registry.rs` so it returns exit 0 with `"rtk xcodebuild $args"`. The filter is already there — this is just wiring the rewriter side.

## Verification of fix

```
$ rtk rewrite "xcodebuild build -project Foo.xcodeproj"
rtk xcodebuild build -project Foo.xcodeproj
$ echo $?
0
```

niklasmarderx:
The xcodebuild TOML filter has existed since v0.30.0 but the rewriter registry had no matching entry. `rtk rewrite "xcodebuild build -project Foo.xcodeproj"` exits 1 (no RTK equivalent), so Claude Code hooks never route xcodebuild through the filter and `rtk discover` shows xcodebuild invocations as missed savings even when the filter is active.

The fix is a single registry rule: pattern `^xcodebuild\b`, rtk_cmd `rtk xcodebuild`, category Build, 85% savings (consistent with other build tools that eliminate verbose compiler noise). This wires the existing filter into the rewriter the same way swift, systemctl, yamllint, etc. are wired.

Closes #1551

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1933,6 +1933,34 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_classify_xcodebuild() {
+        assert!(matches!(
+            classify_command("xcodebuild build -project Foo.xcodeproj"),
+            Classification::Supported {
+                rtk_equivalent: "rtk xcodebuild",
+                category: "Build",
+                ..
+            }
+        ));
+    }
+
+    #[test]
+    fn test_rewrite_xcodebuild() {
+        assert_eq!(
+            rewrite_command("xcodebuild build -project Foo.xcodeproj", &[]),
+            Some("rtk xcodebuild build -project Foo.xcodeproj".into())
+        );
+    }
+
+    #[test]
+    fn test_rewrite_xcodebuild_test() {
+        assert_eq!(
+            rewrite_command("xcodebuild test -scheme MyApp -destination 'platform=iOS Simulator,name=iPhone 15'", &[]),
+            Some("rtk xcodebuild test -scheme MyApp -destination 'platform=iOS Simulator,name=iPhone 15'".into())
+        );
+    }
+
     // --- #336: docker compose supported subcommands rewritten, unsupported skipped ---
 
     #[test]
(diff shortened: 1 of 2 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟢 **Risk** | `low` |

### Summary

Adds a new RtkRule entry for xcodebuild to the rewriter registry in rules.rs, wiring the existing xcodebuild TOML filter so that `rtk rewrite` and `rtk discover` correctly recognize and route xcodebuild commands. Three tests are added to verify classification and rewriting behavior.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1551

---

KuSh:
Closing this one in favour of #2531, which adds the same xcodebuild rewrite rule and has its CLA signed and its head building. If there is something in this version that #2531 misses, please say so there and it can be folded in.

PR #1586, opened 2026-04-28
refactor(git): Fix stash status detection for all cases

KuSh:
## Summary

- Ensure "No local changes" status is detected for all stash operations, not just the default push case
- Consolidate duplicate message formatting into `format_stash_message()` helper
- Improve maintainability and reduce code duplication

Closes https://github.com/rtk-ai/rtk/issues/1535

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [x] Manual testing: `rtk git stash` and `rtk git stash push myfile.txt` output inspected

PR #1636, opened 2026-04-30
fix(git): distinguish stash push no-op from successful stash (#1535)

### LINKED ISSUE #1535: rtk git stash {push,pop,apply,drop} returns exit 0 on failure and produces ambiguous success output

officialasishkumar:
## Summary

`git stash push` exits 0 even when **no stash entry is created** — for example on a clean tree, or when pathspec-restricted invocations match nothing. rtk's generic `ok stash push` collapsed that no-op into the same message it prints for a real stash, which led to lost WIP in the report at #1535 (the user assumed work had been preserved that wasn't).

This change keeps the success-case wire format users rely on, but adds a distinct `noop stash push (no local changes)` line for the no-op case so callers and AI agents can tell whether anything was actually stashed.

## Repro before

```
$ cd /clean/repo
$ rtk -vv git stash push -m "test"
git stash Some("push")
ok stash push      # ← misleading — nothing was stashed
$ echo \$?
0
```

## After

```
$ rtk -vv git stash push -m "test"
git stash Some("push")
noop stash push (no local changes)
$ echo \$?
0
```

A real change still produces the existing `ok stash push` message; the failure path (bad pathspec with real changes elsewhere) still produces `FAILED: git stash push` with a non-zero exit code.

## Notes on Repro 1 (exit 0 on bad pathspec)

The reporter's first repro (exit 0 when the underlying git fails) does not reproduce on current master — `result.exit_code` already propagates correctly through the `pop`/`apply`/`drop`/`push` arm of `run_stash`. I verified this against rtk@master with `git stash push DOES_NOT_EXIST.txt` on a tree with real changes elsewhere: rtk now exits 1 and prints `FAILED: git stash push`. So this PR focuses on Repro 2, which is still present.

## Scope

Only the `Some("pop") | Some("apply") | Some("drop") | Some("push")` arm in `run_stash` is touched. The `None` arm (default `git stash`) already detects "No local changes" separately, and the unrecognized-subcommand and `list`/`show` arms have no equivalent no-op semantics.

## Test plan

- [x] New unit tests cover `stash_push_is_noop` against real `git stash push` output (clean tree, and pathspec-only no-op).
- [x] Manual verification on a local repo:
  - clean tree: prints `noop stash push (no local changes)`, exit 0
  - bad pathspec with real changes elsewhere: prints `FAILED: git stash push`, exit 1
  - real change: prints `ok stash push`, exit 0
- [x] \`cargo fmt --all && cargo clippy --all-targets && cargo test --all\` passes (1689 tests, 0 new warnings).

Closes #1535

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1463,9 +1463,19 @@ fn run_stash(
             let combined = result.combined();
 
             let msg = if result.success() {
-                let msg = format!("ok stash {}", sub);
-                println!("{}", msg);
-                msg
+                // `git stash push` exits 0 even when no stash entry is created
+                // (e.g. clean tree, or pathspecs that match nothing). The generic
+                // "ok stash push" message hid that case and led to lost WIP in
+                // #1535 — surface it explicitly instead.
+                if sub == "push" && stash_push_is_noop(&combined) {
+                    let msg = "noop stash push (no local changes)".to_string();
+                    println!("{}", msg);
+                    msg
+                } else {
+                    let msg = format!("ok stash {}", sub);
+                    println!("{}", msg);
+                    msg
+                }
             } else {
                 eprintln!("FAILED: git stash {}", sub);
                 if !result.stderr.trim().is_empty() {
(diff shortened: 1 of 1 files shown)


KuSh:
Should be fixed by https://github.com/rtk-ai/rtk/pull/1586 which be in the next release. Thanks for your PR

PR #1638, opened 2026-04-30
fix(go): pass through unrecognized go build output instead of "Success" (#1599)

### LINKED ISSUE #1599: go build: reports "Success" when build fails with unrecognized error
## Bug

`rtk go build` reports "Go build: Success" when the underlying `go build` command fails with an unrecognized error. The exit code is correctly propagated as 1, but the filtered output says "Success", which is misleading.

## Reproduction

```bash
# From any Go project root:
rtk go build -o /tmp/test ./cmd/nonexistent
# Output: "Go build: Success"
# Exit code: 1
# Binary: not created

# Bypass filter to see real error:
rtk proxy go build -o /tmp/test ./cmd/nonexistent
# Output: "stat .../cmd/nonexistent: directory not found"
# Exit code: 1
```

## Root Cause

`filter_go_build()` in `src/cmds/go/go_cmd.rs` determines success/failure by scanning output for known error patterns via `is_go_build_error_line()`. When no patterns match, it returns `"Go build: Success"` — even if the output contains unrecognized error text.

The error `stat ...: directory not found` doesn't match any of the recognized patterns (`.go:` locations, `go.mod:` errors, `undefined:` prefixes, etc.), so it's silently dropped and the filter reports success.

## Expected Behavior

When `go build` produces output that doesn't match any known progress/download pattern and doesn't match any known error pattern, the output should be passed through as a failure rather than swallowed.

A truly successful `go build` produces no output (or only `go: downloading ...` progress lines). Any other unrecognized output indicates an error the filter doesn't understand.

officialasishkumar:
## Summary

\`filter_go_build\` called any line not matching \`is_go_build_error_line\` a non-error and reported \`Go build: Success\` if no errors were found. That hid real failures whose output the filter doesn't recognize — e.g. \`stat .../cmd/nonexistent: directory not found\` from the report at #1599 — while the exit code was already correct.

A truly successful \`go build\` produces no output, or only known progress lines (\`go: downloading|finding|extracting\`, package-context headers like \`# example.com/foo\`). Anything else is suspect and is now surfaced as a failure with the unrecognized lines passed through, so the user sees what actually happened.

## Behavior

Before:
\`\`\`
\$ rtk go build -o /tmp/test ./cmd/nonexistent
Go build: Success
\$ echo \$?
1
\`\`\`

After:
\`\`\`
\$ rtk go build -o /tmp/test ./cmd/nonexistent
Go build: failed (1 unrecognized line)
1. stat /tmp/gotest/cmd/nonexistent: directory not found
\$ echo \$?
1
\`\`\`

## Notes

- Existing recognized-error path (\`Go build: N errors\` table) is untouched. Only the previously-silent \"unrecognized output but exit nonzero\" path changes.
- Added a regression test covering the original repro, plus a positive test for \"progress lines + package header only\" to make sure the package-context (\`# example.com/foo\`) noise still maps to \"Success\" when there is no actual error.
- The new helper \`is_go_build_progress_line\` factors the existing benign-line list out of \`is_go_build_error_line\` so both call sites stay in sync.

## Test plan

- [x] New unit tests cover the repro and the progress-only success case.
- [x] All existing \`filter_go_build\` tests still pass.
- [x] \`cargo fmt --all && cargo test --all\` passes (1689 tests).
- [x] Manual repro against \`go build -o /tmp/test ./cmd/nonexistent\` produces the expected failure output.

Closes #1599

### DIFF
--- a/src/cmds/go/go_cmd.rs
+++ b/src/cmds/go/go_cmd.rs
@@ -563,16 +563,47 @@ fn is_go_test_failure_line(line: &str) -> bool {
 /// Filter go build output - show only errors
 pub(crate) fn filter_go_build(output: &str) -> String {
     let mut errors: Vec<String> = Vec::new();
+    let mut unrecognized: Vec<String> = Vec::new();
 
     for line in output.lines() {
         let trimmed = line.trim();
+        if trimmed.is_empty() {
+            continue;
+        }
         if is_go_build_error_line(trimmed) {
             errors.push(trimmed.to_string());
+        } else if !is_go_build_progress_line(trimmed) {
+            // Output we don't classify as either an error or known progress
+            // is almost certainly a failure mode the filter doesn't know about
+            // (#1599: e.g. "stat ...: directory not found"). Don't swallow it.
+            unrecognized.push(trimmed.to_string());
         }
     }
 
-    if errors.is_empty() {
+    if errors.is_empty() && unrecognized.is_empty() {
         return "Go build: Success".to_string();
+    }
+
+    if errors.is_empty() {
+        // Pass the unrecognized output through so the user sees what actually
+        // happened rather than a misleading "Success".
+        let mut result = String::new();
+        result.push_str(&format!(
+            "Go build: failed ({} unrecognized line{})\n",
+            unrecognized.len(),
+            if unrecognized.len() == 1 { "" } else { "s" }
+        ));
+        result.push_str("═══════════════════════════════════════\n");
+        for (i, line) in unrecognized.iter().take(20).enumerate() {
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the fix and the careful repro. Issue #1599 was resolved on develop by #2188 (merged 2026-06-01, commit f69ad6ed), which takes the go build exit status into the filter: when no known error pattern matched but the process exited nonzero, filter_go_build now prints the failure header and the raw output rather than Success, and test_filter_go_build_nonzero_exit_never_reports_success in src/cmds/go/go_cmd.rs covers the stat directory-not-found case you reported. Keying off the exit code rather than line shapes also avoids the risk of calling a successful build failed because it printed an unfamiliar line. Please comment or reopen if you still see Success on a failing build.

PR #1639, opened 2026-04-30
fix(rewrite): leave pipe groups raw end-to-end (#1560)

### LINKED ISSUE #1560: bug(rewrite): hook rewrites left-hand side of piped commands, breaking downstream filters

officialasishkumar:
## Summary

\`rewrite_compound\` only treated \`find\`/\`fd\` as pipe-incompatible (#439) and rewrote every other LHS. RTK filters reformat output — compress, group, drop columns — and silently break downstream pipe consumers. The reported case is

\`\`\`
ps aux | grep python | grep -v grep
\`\`\`

which used to be rewritten to

\`\`\`
rtk ps aux | grep python | grep -v grep
\`\`\`

and returns nothing because \`rtk ps aux\` restructures the output before \`grep\` sees it.

This change generalizes the pipe-incompatible rule: when a segment is followed by a pipe, the entire pipe group stays raw end-to-end. Non-piped segments around a pipe group (separated by \`&&\` / \`||\` / \`;\`) are still rewritten as before, so chained workflows keep their token savings.

## Behavior

Before:
\`\`\`
\$ rtk rewrite 'ps aux | grep python | grep -v grep'
rtk ps aux | grep python | grep -v grep
\$ echo \$?
3
\`\`\`

After:
\`\`\`
\$ rtk rewrite 'ps aux | grep python | grep -v grep'
\$ echo \$?
1   # no rtk equivalent — hook passes through unchanged
\`\`\`

Hook (\`rtk hook claude\`) for the same command produces no \`updatedInput\` after this PR, so Claude Code runs the command verbatim.

Compound commands still rewrite their non-piped halves:
\`\`\`
git status && cargo test | grep FAIL
  → rtk git status && cargo test | grep FAIL    # pipe group raw, &&-segment rewritten
\`\`\`

## Notes

- Several existing tests (\`test_rewrite_pipe_first_only\`, \`test_rewrite_compound_pipe_*\`, \`test_rewrite_*_pipe_*\`, \`test_rewrite_redirect_2_gt_amp_1_with_pipe\`) codified the old \"LHS gets rewritten\" behavior. They are updated to reflect the new safe-by-default rule, with comments pointing back to #1560 and the existing #439 carve-out.
- The find/fd carve-out is no longer special-cased — it falls out of the universal rule.
- New test \`test_rewrite_pipe_in_compound_keeps_pipe_group_raw_only\` pins down the compound-with-pipe semantics.

## Test plan

- [x] All updated and new pipe-related tests pass.
- [x] \`cargo fmt --all && cargo test --all\` passes (1688 tests).
- [x] Manual verification:
  - \`rtk rewrite 'ps aux | grep python | grep -v grep'\` → exit 1, no rewrite
  - \`rtk hook claude\` for same command → no \`updatedInput\` in JSON
  - \`rtk rewrite 'cargo test'\` (no pipe) → \`rtk cargo test\` exit 3 (still works)

Closes #1560

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -462,8 +462,8 @@ fn collapse_line_continuations(s: &str) -> std::borrow::Cow<'_, str> {
 /// Returns `None` if the command is unsupported or ignored (hook should pass through).
 ///
 /// Handles compound commands (`&&`, `||`, `;`) by rewriting each segment independently.
-/// For pipes (`|`), only rewrites the left-hand command (pipe targets stay raw),
-/// but continues rewriting segments after subsequent `&&`/`||`/`;` operators.
+/// For pipes (`|`), leaves the whole pipe group raw, but continues rewriting
+/// segments after subsequent `&&`/`||`/`;` operators.
 /// Also strips user-configured transparent wrapper prefixes
 /// (`[hooks].transparent_prefixes` in `config.toml`) before routing.
 ///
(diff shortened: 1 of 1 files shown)


reneleonhardt:
Thank you so much for your very important fix! ❤️

Do the maintainers still want to review and merge this, so the conflicts would have to be fixed?

officialasishkumar:
Rebased onto current master and fixed the conflicts in e52d50a.

reneleonhardt:
How can reviewers be requested?

I guess the unmerged fixes are needed for Claude Code to reduce token usage instead of increasing it.
https://github.com/headroomlabs-ai/headroom/issues/1237

> I also often stumble with claude saying "output is compressed" and then it proceeds to read local files with python.

Can there be something done to improve the interplay with Claude and headroom?
More realistic unit or integration tests?

KuSh:
Thanks for digging into this one. The pipe rewrite was fixed on develop by #3128 (merged 2026-07-23), which leaves any command that feeds a pipe raw and only rewrites the final stage when that filter is safe to read from stdin, with a follow-up on 2026-07-31 that rewrites the producer only when every consumer is head, tail or cat. Your repro now comes back as 'ps aux | grep python | rtk grep -v grep', and the find and fd carve-out this PR generalised is covered by the same rule. Closing as already addressed, but please comment or reopen if you still see a rewrite that breaks a pipeline.

PR #1646, opened 2026-04-30
feat(trust): fall back to content-hash lookup when path is unknown

artile:
## Summary
- The current trust store keys entries by canonical absolute path. The same byte-for-byte \`.rtk/filters.toml\` at a different path is treated as untrusted and re-prompts \`rtk trust\` even though the user has already reviewed that exact content.
- Real cases this hurts:
  - **git worktrees** — a second checkout sits at a different absolute path, so the worktree gets a noisy untrusted warning.
  - **Multiple self-hosted CI runners** — each runner has its own \`_work/<repo>/<repo>\` path; trust on one runner doesn't carry to another.
  - **Repo rename / move** — invalidates trust even though content is identical.
  - **Vendor / shared filters copied between projects** — second project re-triggers review of identical bytes.
- Add a content-hash fallback to \`check_trust\`. Exact path lookup still wins (preserves \`ContentChanged\` alarm + audit trail at known locations). If the path is unknown, the file's sha256 is compared against every entry in the store; a match means \"the user already reviewed these exact bytes somewhere\" → Trusted.

## Lookup order (after this PR)
1. \`RTK_TRUST_PROJECT_FILTERS=1\` + recognized CI env → \`EnvOverride\` (unchanged)
2. Exact canonical-path match → \`Trusted\` / \`ContentChanged\` (unchanged)
3. **NEW:** Any stored entry's \`sha256\` matches the file's hash → \`Trusted\`
4. Otherwise → \`Untrusted\` (unchanged)

## Security note
sha256 collision resistance is the boundary. If the file's hash matches a previously-trusted entry, the content is byte-identical to something the user explicitly reviewed. Reproducing a known-trusted hash with different content would require a sha256 preimage attack. The \`rtk trust\` review flow (content dump + risk summary) is unchanged — first-time trust still requires explicit review.

## Schema
Unchanged. Existing trust stores work as-is; the fallback just consults the existing \`sha256\` field on every \`TrustEntry\`. No migration.

## Test plan
- [x] \`cargo test --bin rtk hooks::trust\` — 13/13 pass (10 existing + 3 new).
- [x] \`cargo test --bin rtk\` — 1609/1609 pass.
- New cases:
  - \`test_content_hash_fallback_trusts_identical_bytes_at_new_path\` — trust at \`repo-a\`, identical bytes at \`repo-b\` are auto-trusted.
  - \`test_content_hash_fallback_does_not_trust_different_bytes\` — different bytes at a new path stay untrusted.
  - \`test_path_match_takes_precedence_over_hash_fallback\` — modifying a known path's file to bytes that match some other trusted entry still surfaces \`ContentChanged\` (alarm not masked by fallback).


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟡 **Risk** | `medium` |

### Summary

Adds a content-hash fallback to the trust store lookup. When a file's canonical path has no existing trust entry, the system now checks whether any previously trusted entry shares the same SHA-256 hash, treating byte-identical content as already reviewed. This eliminates redundant trust prompts across git worktrees, CI runners, repo renames, and copied filter files.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

---

artile:
Pushing directly to develop instead.

PR #1110, opened 2026-04-09
feat(php): add PHPStan analyse filter module

LucianoVandi:
## Summary

Adds a new `php` ecosystem module with a filter for [PHPStan](https://phpstan.org/).

### What it does

- injects `--error-format=json` for `analyse` / `analyze`
- groups errors by file, sorted by error count descending
- shows up to 10 files × 5 messages each, with truncation hints
- auto-detects `vendor/bin/phpstan` before falling back to `phpstan` on PATH
- falls back to text parsing when a custom `--error-format` is requested
- falls back safely to `fallback_tail()` if JSON parsing fails
- leaves utility commands like `--version`, `list`, and `clear-result-cache` as real passthrough

### Rewrite coverage

Adds rewrite support for:

- `phpstan analyse`
- `vendor/bin/phpstan analyse`
- `./vendor/bin/phpstan analyse`
- `php vendor/bin/phpstan analyse`

## Tests

Added module tests covering:

- zero-error and error JSON output
- truncation behavior
- token savings on a realistic fixture
- malformed / empty output handling
- text fallback behavior
- classify / rewrite behavior, including negative coverage for utility commands

## Validation

```bash
cargo fmt --all --check
cargo clippy --all-targets
cargo test --all
```

## Notes

Snapshot tests were not added because `insta` is not currently a repo dependency. This PR uses explicit assertions and fixture-based coverage instead.

PR #874, opened 2026-03-26
feat(php): add PHPUnit support with compact output (65%+ token savings)

Beninho:
## Summary

- New `rtk phpunit` command with a state-machine output filter (65%+ token savings)
- Supports all PHPUnit invocation variants: `phpunit`, `vendor/bin/phpunit`, `bin/phpunit` (Symfony), `php vendor/bin/phpunit`, `php bin/phpunit`

## Changes

| File | Change |
|------|--------|
| `src/cmds/mod.rs` | Add entry for PHP |
| `src/cmds/php/mod.rs` | New PHP module |
| `src/cmds/php/phpunit_cmd.rs` | New module — state-machine filter, binary detection, tee integration |
| `src/main.rs` | Routing, doc comment, `is_operational_command` registration |
| `src/discover/rules.rs` | PATTERNS regex + RtkRule with 5 rewrite prefixes |
| `src/discover/registry.rs` | 10 classify/rewrite tests for all variants |
| `tests/fixtures/phpunit_raw.txt` | Realistic fixture (PHPUnit 10, 110 tests, 1 failure) |
| `scripts/test-all.sh` | PHP section (conditional on phpunit/vendor/bin presence) |
| `README.md` | Quick-ref line + hook mapping table entry |

## Filter design

State machine: `Header → Failures → Done`

```
PHPUnit 10.5.0 ...          ← stripped (Header state)
Runtime: PHP 8.2.0          ← stripped
.F.......  100/100 (100%)   ← stripped
Time: 00:01:23, Memory: ... ← stripped
There was 1 failure:        ← triggers Failures state
1) UserTest::testEmail      ← captured
Failed asserting ...        ← captured (up to 2 detail lines)
/path/to/test:42            ← captured
FAILURES!                   ← triggers Done state
Tests: 110, Assertions: 340, Failures: 1.  ← parsed for summary
```

Output:
```
PHPUnit: 110 tests, 340 assertions, 1 failures
═══════════════════════════════════════
Failures:
1. 1) UserTest::testEmail
   Failed asserting that false is true.
   /path/to/test:42
```

## Test plan

- [X] `cargo test --all` — 1143 passed, 0 failed
- [X] `cargo fmt --all && cargo clippy --all-targets` — clean for updated files
- [X] Token savings test asserts ≥65% on realistic fixture
- [X] All 5 invocation variants tested for classify + rewrite
- [X] Manual: `rtk phpunit --help` shows description
[...]

PR #1649, opened 2026-04-30
feat(php): consolidated PHP tooling (php, artisan, phpunit, phpstan, pest, paratest, ecs, pint)

iliaal:
## Summary

Consolidates the open and closed PHP-tooling work into a single PR against `develop`, omitting `.phpt` (which has its own PR #1503).

This PR adds eight `rtk` subcommands for PHP test runners and code tooling: `php`, `artisan`, `phpunit`, `phpstan`, `pest`, `paratest`, `ecs`, `pint`. Output compression in the 60% to 95% range per tool, structured parsers where possible (PHPUnit state machine, PHPStan typed JSON, Pint per-file rule counts).

## Credits to original authors

- **@aaronflorey** ([rtk-ai/rtk#1246](https://github.com/rtk-ai/rtk/pull/1246), self-closed 2026-04-26): `php`, `artisan`, `ecs`, `pest`, `paratest`, `test_output`, `utils`, `composer_bin_dirs`, registry normalization for Composer custom-bin-dir layouts.
- **@Beninho** ([rtk-ai/rtk#874](https://github.com/rtk-ai/rtk/pull/874), open): `phpunit` state-machine parser. I ported it into the consolidated module, switched it to `runner::run_filtered`, and stripped emoji from output.
- **@LucianoVandi** ([rtk-ai/rtk#1110](https://github.com/rtk-ai/rtk/pull/1110), open): `phpstan` typed serde::Deserialize parser for `--error-format=json`. Groups errors by file, sorts by count descending; passes utility commands (`--version`, `list`, `clear-result-cache`) through unchanged. I stripped emoji from the success output.
- **New work**: `pint` (Laravel Pint code-style fixer) using `--format=json` for structured per-file rule counts.

If the maintainers prefer this consolidation, close the three separate PRs in favor of this one. If they prefer the originals, close this PR and let the per-tool PRs continue independently. Either is fine with me.

## Composer custom-bin-dir support

`composer_bin_dirs()` reads `COMPOSER_BIN_DIR` and `composer.json`'s `config.bin-dir`, so `tools/bin/phpunit` classifies identically to `vendor/bin/phpunit`. `registry.rs` normalizes tool paths before matching, so a single rule covers the standard Composer layouts.


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟡 **Risk** | `medium` |

### Summary

Adds eight new rtk subcommands for PHP tooling (php, artisan, phpunit, phpstan, pest, paratest, ecs, pint) with structured output parsers and compression. Consolidates work from three prior PRs by different authors into a single module, including shared test output filtering, Composer custom-bin-dir support, and registry normalization.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [x] Docs updated

**Linked issues:** #1503, #1246, #874, #1110

---

pierresh:
Hi, 
Thanks a lot for this consolidated PHP tooling. I tested it, and it works well. Just 2 suggestions: to add Behat and Rector as they are both popular tools in the PHP world.

iliaal:
> Hi, Thanks a lot for this consolidated PHP tooling. I tested it, and it works well. Just 2 suggestions: to add Behat and Rector as they are both popular tools in the PHP world.

Can be added maybe as a follow-up, PR is already quite big, but 100% a good idea.

iliaal:
Added f9e139a on top of this branch: swaps the `make` filter from `max_lines = 50` (head-only) to `head_lines = 10 + tail_lines = 40`, so `make test` runs (e.g. PHPT, which puts ~17000 progress lines before the pass/fail summary) keep both the prologue and the trailing summary instead of dropping the latter.

Companion fix at the recovery layer: #1696 — same head-bias bug in `write_tee_file`, where the tee log capped at 1MB by keeping `raw[..1MB]`. With both landed, tail-heavy output survives at the filter and at the fallback log.

webagil-kevin:
Tested this branch on a production Symfony monorepo (27 packages, PHP 8.4/8.5, Pest 4.6.3 + PHPStan level 9), built from source and wired through the Claude Code `PreToolUse` hook.

Works well across the toolchain:
- `rtk pest` / `rtk phpunit` — failures + summary only, drops the passing noise while preserving diffs and `file:line`
- `rtk phpstan` — compact `file:line` errors instead of the full table + preamble
- `rtk pint`, `rtk php artisan test` — resolve and filter as expected
- `vendor/bin/*` rewrites and `php artisan*` routing all behave correctly
- `cargo test` is green on the branch (2035 passed)

Coverage matches our PHP stack exactly and the filtered output stays faithful to each tool's native format. Would love to see this land — happy to help test further. Thanks for the thorough work here.

EliW:
Really looking forward to this landing as part of rtk.    It's needed to get real benefits from rtk for those of us on PHP monoliths at the moment :) 

Pretty please?

KuSh:
Hi @iliaal,

Since you're aggregating contributions from others in this PR, please ensure you include the "Co-authored-by:" trailer in the commit message to properly credit them.

Example:
````
Co-authored-by: Name <email@example.com>
````

This helps maintain proper attribution for all contributors. Thanks!

**Comment by evaldnet** [comment]
Tried this branch locally against current PHP tool versions (**Pint 1.27.1, PHPStan 2.1.40, ECS 13.0.4**) — great work, the savings are substantial. I hit two parser bugs where the expected schema no longer matches what current tooling emits. Both stem from test fixtures using an older schema, so the suite stays green. Details + fixes below.

**1. `pint` — JSON field names changed (`name`/`appliedFixers` → `path`/`fixers`)**

`PintFile` requires `name` + `appliedFixers`, but Pint ≥ ~1.14 (confirmed on 1.27.1) emits:

```json
{"result":"fail","files":[{"path":"app/Foo.php","fixers":["concat_space"]}]}
```

The fields are required with no aliases, so serde rejects it and `filter_pint_json` falls back to raw output — no compression. Fix is backward-compatible aliases:

```rust
#[serde(alias = "path")]
name: String,
#[serde(rename = "appliedFixers", alias = "fixers")]
applied_fixers: Vec<String>,
```

**2. `phpstan` — reports `phpstan: ok` while errors exist** ⚠️

`filter_phpstan_json` gates the "ok" case on `totals.errors == 0`, but PHPStan puts per-file errors in `totals.file_errors` and leaves `totals.errors` (global/config errors only) at `0`. A real run with 78 file errors — `{"totals":{"errors":0,"file_errors":78}, ...}` — returns `phpstan: ok`, silently hiding the failures (the summary line under-reports for the same reason). Note `file_errors` is currently `#[allow(dead_code)]`. Fix:

```rust
if phpstan.totals.file_errors == 0 && phpstan.totals.errors == 0 { return "phpstan: ok".to_string(); }
// and report phpstan.totals.file_errors in the summary line
```

Both fixtures set `errors == file_errors` (phpstan) / use the old field names (pint), which is why tests don't catch these. Happy to open a PR against this branch with both fixes + regression tests using the current-version schemas if that's useful.

iliaal:
Thanks for the precise report, @evaldnet — both confirmed and fixed on this branch (184ced3).

- `pint`: added backward-compatible serde aliases (`path`/`fixers`) so current Pint output compresses again instead of falling back to raw.
- `phpstan`: the `ok` gate and summary now read `file_errors`, so runs with file-level errors no longer report `phpstan: ok`.

Both ship with regression tests on the current-version schemas — the old fixtures set `errors == file_errors` and used the old pint keys, which is why the suite stayed green.

EliW:
@evaldnet Thanks for finding those. --- out of curiousity @iliaal --- Does this mean that the code here now may depend on which version of phpstan you are using for example?   Or did I grok that incorrectly.   Some folks end up stuck on older phpstan's due to framework incompatibilities at times...

iliaal:
@EliW No. Both old and current tool output parse; I checked each schema against the new code.

**pint:** the fix adds serde aliases, not a rename. The struct accepts both the old keys (`name`/`appliedFixers`) and the current ones (`path`/`fixers`), so it parses Pint from either side of the ~1.14 rename. Before, only the old schema parsed and current Pint fell back to raw. Confirmed: old-key and new-key JSON both compress.

**phpstan:** no new version dependency. `totals.errors` and `totals.file_errors` have both been in the `--error-format=json` output since at least 0.12, unchanged through 1.x and current 2.1.40. `errors` counts non-file-specific errors, `file_errors` counts file-specific ones. The parser already required `file_errors`, it was just unused. The bug read `errors` where it should read `file_errors`, which is correct on old and current PHPStan alike.

Safety net under both: if the JSON ever fails to parse, rtk falls back to raw tool output. You lose the compression for that run, not the result. So anyone stuck on an older phpstan or pint won't get broken output; worst case is unfiltered passthrough.

evaldnet:
Rebuilt on 184ced3 and re-ran against current versions. Both fixes hold:

- phpstan (2.1.40): file-level errors now report instead of "phpstan: ok".
- pint (1.29.1, the new path/fixers keys): an 8-fixer file compresses to "pint: 8 changes in 1 files" with the rule list.

One separate thing I noticed while testing: the leading `./` form isn't matched for most tools, so it runs raw with no compression.

- `./vendor/bin/pint` runs raw, while `vendor/bin/pint` rewrites to `rtk pint`. Same for pest, paratest, ecs, phpunit.
- The phpstan pattern already handles this with an optional `\.?/?` prefix (it also requires the `analyse` subcommand). The other five patterns are `^(?:vendor/bin/)?<tool>`, so a leading `./` defeats the match.

Since `./vendor/bin/...` is how most people invoke these in Laravel, adding the same `\.?/?` prefix would close it.

iliaal:
Fixed on the branch (bd91048). One correction on the mechanism: the regex isn't the blocker. `classify_command` already normalizes a leading `./` (`normalize_php_tool_path` strips it), so `./vendor/bin/pint` classifies fine without `\.?/?` in the pattern. The gap is the rewrite step, which strips *literal* `rewrite_prefixes` from the raw command. The five rules only carried `vendor/bin/<tool>` and bare `<tool>`; phpstan works because `./vendor/bin/phpstan` is in its prefix list, not its regex. Fix adds `./vendor/bin/<tool>` to the five lists, with a regression test covering `./vendor/bin/{pint,pest,paratest,ecs,phpunit}`.

iliaal:
Rebased onto current `develop` (this branch was ~50 commits behind) and fixed the two CI failures:

- **`test` (`test_every_subcommand_is_classified`)**: `develop` added this test plus the `PASSTHROUGH` const after this branch forked, so the merged tree had the php subcommands with no classification entry. Added `php`/`phpunit`/`phpstan`/`pest`/`paratest`/`ecs`/`pint` to `PASSTHROUGH`.
- **semgrep (`dynamic-command-execution`)**: `php_tool_command` built the command via `Command::new` on a resolved path. Routed it through `resolved_command` (the sanctioned PATHEXT-aware constructor) instead — no behavior change.

The six php-tooling commits replayed with zero conflicts. Locally: `test_every_subcommand_is_classified` passes, fmt/clippy clean, full suite green. The semgrep fix I verified by inspection (no `Command::new(<var>)` left in the php dir); CI will confirm.

Heads up: the workflow run on the new head is sitting in `action_required` — it needs a maintainer to approve before `test`/`semgrep` execute.

src/cmds/php/phpstan_cmd.rs
```diff
@@ -0,0 +1,502 @@
...
+    // Build list of files with errors, sorted by error count descending
+    let mut files_vec: Vec<(&String, &PhpstanFile)> = phpstan.files.iter().collect();
+    files_vec.sort_by(|a, b| b.1.errors.cmp(&a.1.errors).then(a.0.cmp(b.0)));
+
+    let max_files = 10;
+    let max_messages_per_file = 5;
+
+    for (path, file) in files_vec.iter().take(max_files) {
+        let short = compact_php_path(path);
+        result.push_str(&format!("\n{} ({} errors)\n", short, file.errors));
```
KuSh, line 181: **🔴 Bug: `"not found"` matches real PHPStan analysis output**

```rust
|| t.contains("not found")
```

This is meant to detect a missing binary, but PHPStan's own static-analysis messages routinely contain this substring. Real examples from PHPStan text output:

```
Method 'findById' not found in class 'UserRepository'.
Class 'NotFoundHttpException' not found.
Property $user not found in class 'Controller'.
```

When any of these appear (which they will on a failing run), the function short-circuits and returns `"PHPStan error:\n..."` instead of the analysis summary — **swallowing the real errors**.

Fix: replace the broad substring check with shell-exit-style prefixes only:

```rust
|| t.starts_with("phpstan: command not found")
|| t.starts_with("sh: ")
|| t.starts_with("bash: ")
```

KuSh (changes requested):
## Code Review — feat(php): consolidated PHP tooling

> **Review performed using [Claude Code](https://claude.ai/code) (claude-sonnet-4-6) by the RTK maintainers.**

This is a solid and well-scoped contribution. The PHPUnit state machine, the PHPStan `errors` vs `file_errors` distinction, and the Pint dual-schema handling via serde aliases are all done correctly and are well-tested. The `php_tool_command()` abstraction for Composer-aware binary resolution is the right design.

That said, there are **three blocking bugs** and **two maintainability issues** to address before merge.

---

### 🔴 Blockers

**1. `filter_phpstan_text`: `"not found"` is dangerously broad** (inline comment on `phpstan_cmd.rs:165`)

**2. `phpstan_cmd.rs` bypasses `php_tool_command()`** (inline comment on `phpstan_cmd.rs:48`)

**3. `is_numbered_failure_heading()` has false positives** (inline comment on `phpunit_cmd.rs:89`)

---

### 🟡 Should-fix

**4. `compact_php_path()` is Laravel-only** (inline comment on `phpstan_cmd.rs:192`)

**5. `short_path()` calls `current_dir()` in a per-file loop** (inline comment on `pint_cmd.rs:126`)

---

### What's working well

- PHPStan `errors` vs `file_errors` bug is correctly identified and tested (`test_filter_phpstan_file_errors_not_hidden`) — this was a real prior bug fixed here
- PHPUnit state machine correctly handles success, multi-failure, errors, skipped, and truncation cases with 9 test cases
- Pint dual-schema serde aliases (`name`/`appliedFixers` vs `path`/`fixers`) cleanly handle the Pint ≥1.14 rename
- `php_tool_command()` centralises Composer bin-dir resolution correctly — pity PHPStan doesn't use it (see inline comment)
- Token savings are verified against a real 380-line `phpstan_raw.json` fixture, not synthetic data
- No `unwrap()` in production paths, fallback pattern respected throughout

src/cmds/php/phpstan_cmd.rs
```diff
@@ -0,0 +1,502 @@
...
+    #[serde(default)]
+    #[allow(dead_code)]
+    ignorable: bool,
+}
+
+// ── Public entry point ───────────────────────────────────────────────────────
+
+pub fn run(args: &[String], verbose: u8) -> Result<i32> {
+    // Check for vendor/bin/phpstan first
+    let mut cmd = if Path::new("vendor/bin/phpstan").exists() {
```
KuSh: **🔴 Bug: PHPStan ignores `COMPOSER_BIN_DIR` and custom `config.bin-dir`**

```rust
let mut cmd = if Path::new("vendor/bin/phpstan").exists() {
    resolved_command("vendor/bin/phpstan")
} else {
    resolved_command("phpstan")
};
```

Every other PHP command in this module uses `php_tool_command("tool")`, which calls `composer_tool_paths()` and correctly respects `COMPOSER_BIN_DIR` and `composer.json`'s `config.bin-dir`. PHPStan hard-codes `vendor/bin/` here, diverging silently from the pattern.

On a project with a custom Composer bin dir (e.g. `config.bin-dir: "bin"`), this falls through to the global `phpstan` binary — likely a different version than the project pin.

Fix:
```rust
let mut cmd = php_tool_command("phpstan");
```
(and `use super::utils::php_tool_command;` in the imports)

src/cmds/php/phpunit_cmd.rs
```diff
@@ -0,0 +1,337 @@
...
+            return format!("PHPUnit: {} tests, {} assertions", tests, assertions);
+        }
+        return "PHPUnit: ok".to_string();
+    }
+
+    build_phpunit_summary(output, &failures)
+}
+
+fn is_numbered_failure_heading(line: &str) -> bool {
+    // PHPUnit formats each failure as "N) Class::method"
```
KuSh: **🔴 Bug: `is_numbered_failure_heading()` matches too broadly**

```rust
fn is_numbered_failure_heading(line: &str) -> bool {
    let mut chars = line.chars();
    let first_digit = chars.next().is_some_and(|c| c.is_ascii_digit());
    first_digit && line.contains(')')
}
```

The intent is to match PHPUnit's `N) ClassName::testMethod` heading format. But the predicate — _starts with a digit_ AND _contains `)`_ — is too loose. Detail lines within a failure block can plausibly match:

```
Failed asserting that Array(3 items) is identical to Array(5 items).
                                   ^--- starts with digit? No, starts with 'F'
```

More concretely, data-provider traces or exception messages like:
```
5 of 10 assertions passed in Foo::bar()
```
start with a digit and contain `)`, splitting what should be a single failure into two.

Fix: anchor to the exact `N) ` PHPUnit format:
```rust
lazy_static! {
    static ref FAILURE_HEADING_RE: Regex = Regex::new(r"^\d+\) \S").unwrap();
}

fn is_numbered_failure_heading(line: &str) -> bool {
    FAILURE_HEADING_RE.is_match(line)
}
```

src/cmds/php/phpstan_cmd.rs
```diff
@@ -0,0 +1,502 @@
...
+
+pub(crate) fn filter_phpstan_text(output: &str) -> String {
+    // Check for errors first
+    for line in output.lines() {
+        let t = line.trim();
+        if t.contains("cannot load such file")
+            || t.contains("not found")
+            || t.starts_with("phpstan: command not found")
+            || t.starts_with("phpstan: No such file")
+        {
```
KuSh, line 223: **🟡 Maintainability: `compact_php_path()` is Laravel-only and will need ongoing maintenance**

```rust
for prefix in &[
    "app/Models/",
    "app/Http/Controllers/",
    "app/Http/Middleware/",
    "app/Services/",
    "app/Repositories/",
    "src/",
    "tests/",
    "config/",
    "database/",
] {
```

These are Laravel's directory conventions. Symfony projects use `src/Entity/`, `src/Controller/`, etc. Plain PHP projects can use anything. For non-Laravel codebases, none of these prefixes match, and the function falls through to the last-two-components heuristic anyway.

The fallback at the bottom of this function (last 2 path components) is actually a good universal default for all frameworks. The prefix list adds maintenance burden for little marginal benefit over just using the fallback everywhere.

Consider removing the prefix list and using only:
```rust
fn compact_php_path(path: &str) -> String {
    let path = path.replace('\\', "/");
    // Keep last 2 path components: dir/File.php is enough context
    if let Some(pos) = path.rfind('/') {
        if let Some(prev) = path[..pos].rfind('/') {
            return path[prev + 1..].to_string();
        }
        return path[pos + 1..].to_string();
    }
    path
}
```

src/cmds/php/pint_cmd.rs
```diff
@@ -0,0 +1,218 @@
...
+    let mut files = pint.files;
+    files.sort_by(|a, b| {
+        b.applied_fixers
+            .len()
+            .cmp(&a.applied_fixers.len())
+            .then(a.name.cmp(&b.name))
+    });
+
+    let mut result = format!("pint: {} changes in {} files\n", total_rules, total_files);
+
```
KuSh, line 104: **🟢 Performance: `current_dir()` syscall inside a per-file loop**

`short_path()` calls `std::env::current_dir()` on every iteration of this loop (up to `MAX_FILES_SHOWN = 15` calls per `pint` invocation). The cwd doesn't change between iterations.

Fix: compute it once before the loop:

```rust
let cwd_prefix = std::env::current_dir().ok()
    .and_then(|p| p.into_os_string().into_string().ok())
    .map(|s| format!("{}/", s));

for file in files.iter().take(MAX_FILES_SHOWN) {
    let name = cwd_prefix.as_deref()
        .and_then(|c| file.name.strip_prefix(c))
        .unwrap_or(&file.name);
    result.push_str(&format!("\n{} ({})\n", name, file.applied_fixers.len()));
    ...
}
```

(Minor — doesn't block merge, but it's a trivial fix.)

src/cmds/php/phpstan_cmd.rs
iliaal: Fixed in `2211f3b`. `filter_phpstan_text` now matches only shell-level binary-missing lines (`phpstan: command not found`, `sh:`/`bash:`). Dropped the broad `not found` substring and a stray `cannot load such file` (a Ruby `LoadError` string that never applied to PHP). Regression test feeds analysis output containing `Class 'NotFoundHttpException' not found` and asserts the summary survives.

src/cmds/php/phpstan_cmd.rs
iliaal: Fixed in `2211f3b`. `run()` now calls `php_tool_command("phpstan")`, so `COMPOSER_BIN_DIR` / `config.bin-dir` resolve like the other tools. Removed the now-unused `resolved_command`/`Path` imports.

src/cmds/php/phpunit_cmd.rs
iliaal: Fixed in `4c31a9d`. Anchored to `^\d+\) \S` via a lazy_static regex. Added `test_numbered_failure_heading_anchored` covering `1) Class::method` and multi-digit (match) plus `5 of 10 assertions passed in Foo::bar()` and a bare `1)` (no match).

src/cmds/php/phpstan_cmd.rs
iliaal: Done in `2211f3b` — dropped the prefix list, kept the last-two-components fallback for all frameworks; tests updated to the compacted output. (Minor: `src/`/`tests/`/`config/`/`database/` in the old list were generic rather than Laravel-only, but the universal fallback is the simpler call — agreed.)

src/cmds/php/pint_cmd.rs
iliaal: Fixed in `9d05729`. Resolved once into a `cwd_prefix` before the loop; `short_path()` removed.

iliaal:
Thanks for the thorough review, @KuSh — all five addressed in `2211f3b`, `4c31a9d`, `9d05729` (on top of the develop rebase).

- **Blocker 1** (`not found`): text fallback now matches only shell-level binary-missing lines; regression test guards against swallowing analysis output.
- **Blocker 2** (`php_tool_command`): phpstan now resolves Composer-aware like the other tools.
- **Blocker 3** (failure heading): anchored to `^\d+\) \S` with a regression test for the `5 of 10 ...` false positive.
- **Should-fix 4** (`compact_php_path`): simplified to last-two-components for all frameworks.
- **Perf 5** (`current_dir`): hoisted out of the per-file loop.

Per-comment notes inline. fmt/clippy clean, php suite green.

src/filters/make.toml
KuSh, line 1: Could you move this non-PHP-related change to a dedicated PR (or merge them with your other PR [#1696])?

This PR is already quite large, so keeping it focused on a single subject would make it easier to review.

src/discover/rules.rs
```diff
@@ -530,6 +530,91 @@ pub const RULES: &[RtkRule] = &[
...
+        pattern: r"^php\s+-l(?:\s|$)",
+        rtk_cmd: "rtk php",
+        rewrite_prefixes: &["php"],
+        category: "Build",
+        savings_pct: 60.0,
+        subcmd_savings: &[],
+        subcmd_status: &[],
+    },
+    RtkRule {
+        pattern: r"^(?:php\s+)?(?:(?:vendor/bin|bin)/)?phpunit(?:\s|$)",
```
KuSh: It seems the regex patterns aren’t consistent, even when the requirements seem identical. For consistency, the regex should be standardized as:
```regex
^(?:php\s+)?(?:\./)?(?:(?:vendor/)?bin/)?phpunit(?:\s|$)
```

The same approach should apply to phpstan, while a simpler variant can be used for pest, paratest, ecs, and pint.

Additionally, your rewrite_prefixes don’t cover all possible regex variations, which could lead to issues. Please ensure:
- The regex covers all intended cases and only those.
- The rewrite_prefixes list is exhaustive.

src/filters/make.toml
iliaal: Moved out — opened #2707 against develop for the `make.toml` change and dropped it from this branch, so #1649 is now PHP-only. (Note: #1696 is the tee tail-preservation PR, so I gave make its own PR rather than folding it there.)

src/discover/rules.rs
iliaal: Done. Patterns standardized: `^(?:php\s+)?(?:\./)?(?:(?:vendor/)?bin/)?<tool>` for phpunit/phpstan, `^(?:\./)?(?:vendor/bin/)?<tool>` for pest/paratest/ecs/pint.

On exhaustiveness: rather than enumerate every literal in `rewrite_prefixes`, I made the rewrite path normalize the leading invocation (php wrapper, `./`, vendor/bin, composer bin-dir) the same way `classify_command` already does. Any spelling the pattern accepts now normalizes to the same canonical form before prefix-matching, so the list collapses to the residual forms (`bin/<tool>` + bare name for phpunit/phpstan, bare name for the rest) and can't drift out of sync with the regex. `test_rewrite_php_tool_invocation_forms` asserts all 7 phpunit spellings (incl. `./bin/phpunit`, `php phpunit`) map to one rewrite, and that the simpler tools reject `php vendor/bin/pint` / `bin/pint`.

src/cmds/php/pint_cmd.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+#[derive(Deserialize)]
+struct PintFile {
+    // Pint ≥ ~1.14 renamed the JSON keys: name→path, appliedFixers→fixers.
+    // Aliases keep both schemas parsing so output stays compressed across versions.
+    #[serde(alias = "path")]
+    name: String,
+    #[serde(rename = "appliedFixers", alias = "fixers")]
+    applied_fixers: Vec<String>,
```
KuSh, line 33: From what I can see, [applied_fixers is optional in PHP CS Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/8054bfe12f3b34fc3b4779175247459bc79aa03f/src/Console/Report/FixReport/JsonReporter.php#L40), so this case should be handled as well.
Minimal reproduction case:
```sh
echo '{"files":[{"name":"x.php"}]}' | ./target/debug/rtk pipe -f pint
[rtk] pint (JSON parse error): output format not recognized, showing last 20 lines
{"files":[{"name":"x.php"}]}

**Review by KuSh — CHANGES_REQUESTED** [review]

File: src/cmds/php/pint_cmd.rs (hunk 7 above)
Comment (iliaal) [reply src/cmds/php/pint_cmd.rs:33]: Fixed in `0cc15dc`. `applied_fixers` is now `#[serde(default)]`, so a file entry without the key deserializes to an empty vec instead of failing the whole parse. Added `test_pint_file_without_fixers_key` covering your `{"files":[{"name":"x.php"}]}` case.

Heads-up on the minimal repro: with that exact single-file input it still echoes the raw JSON — but not from a parse error. The `never_worse` guard fires because the summary (`pint: 0 changes in 1 files` + `x.php (0)`) is longer than the 28-byte input, so raw is genuinely the smaller output. On a realistically-sized report the fixers-less entry now parses and compresses:

```
$ printf '{"files":[{"name":"app/Models/User.php","appliedFixers":["no_unused_imports","ordered_imports"]},{"name":"app/Services/PaymentGateway.php"}]}' | rtk pipe -f pint
pint: 2 changes in 2 files

app/Models/User.php (2)
  - no_unused_imports
  - ordered_imports

app/Services/PaymentGateway.php (0)
```

File: src/cmds/php/phpunit_cmd.rs [hunk 8]
```diff
@@ -0,0 +1,358 @@
...
+    runner::run_filtered(
+        cmd,
+        "phpunit",
+        &args.join(" "),
+        filter_phpunit_output,
+        runner::RunOptions::stdout_only().tee("phpunit"),
+    )
+}
+
+pub(crate) fn filter_phpunit_output(output: &str) -> String {
```
KuSh, line 43: `filter_phpunit_output` never strips ANSI, breaking all pattern matches when `--colors=always`

Pattern checks operates on raw bytes. With --colors=always it will fail and fall to "PHPUnit: ok" with zero counts — losing failure information entirely. Other PHP filters call strip_ansi_and_controls as their first step; it seems it is needed here too (and perhaps do a check to see if it's missing anywhere else)

src/cmds/php/phpstan_cmd.rs
```diff
@@ -0,0 +1,496 @@
...
+
+pub fn run(args: &[String], verbose: u8) -> Result<i32> {
+    // Composer-aware resolution (COMPOSER_BIN_DIR / config.bin-dir), matching
+    // the other php tools, with PATH fallback for a global install.
+    let mut cmd = php_tool_command("phpstan");
+
+    // Utility commands (--version, list, clear-result-cache, worker, …): real passthrough.
+    // Only analyse/analyze subcommands get filtered and token-tracked.
+    let is_analyse = args
+        .first()
```
KuSh: Seems like this will fail if any global flag is used. Minimal reproduction case: `phpstan -c phpstan.neon analyse src/`

src/cmds/php/utils.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PhpTestRunner {
+    Pest,
+    Phpunit,
+    Unknown,
+}
+
+pub fn detect_php_test_runner() -> PhpTestRunner {
+    if composer_tool_exists("pest") || Path::new("pest.php").exists() {
```
KuSh: According to Claude, Pest does not create or require a file named `pest.php`. The canonical indicators are the `vendor/bin/pest` binary (already covered by `composer_tool_exists`) and `pestphp/pest` in `composer.json`. Any project with a utility file named `pest.php` (compatibility shim, bootstrap, custom runner) causes `detect_php_test_runner()` to return `Pest` even in a PHPUnit-only project. Consequence: `rtk php artisan test` applies `filter_test_runner_output` instead of `filter_artisan_output`, leaving box-drawing characters and multiple blank lines unstripped. Fix: remove the `pest.php` file check; `composer_tool_exists("pest")` alone is sufficient.

src/cmds/system/pipe_cmd.rs
```diff
@@ -154,6 +179,11 @@ pub fn auto_detect_filter(input: &str) -> fn(&str) -> String {
         return crate::cmds::python::pytest_cmd::filter_pytest_output;
     }
 
+    // phpunit banner: "PHPUnit X.Y.Z by Sebastian Bergmann and contributors."
+    if first_1k.contains("by Sebastian Bergmann") {
```
KuSh: This seems a bit too broad; anything mentioning the author (a LICENSE file, `git log` output, etc.) would be routed to `filter_phpunit_output`. Could we tighten this with `first_trimmed.starts_with("PHPUnit ")` or an equivalent check?

src/core/utils.rs
```diff
@@ -355,6 +357,54 @@ pub fn resolved_command(name: &str) -> Command {
...
+}
+
+pub fn composer_tool_paths(tool: &str) -> Vec<PathBuf> {
+    composer_bin_dirs()
+        .into_iter()
+        .map(|dir| dir.join(tool))
+        .collect()
+}
+
+fn composer_bin_dirs_from(env_bin_dir: Option<&str>, composer_json: Option<&str>) -> Vec<PathBuf> {
```
KuSh, line 387: This is called many time per `rewrite_segment_inner` call (called by `classify_command` and directly by `rewrite_segment_inner`) through `normalize_php_tool_command`. Hoisting the result to a once_cell or passing it as a parameter would reduce this to one read per invocation.

src/cmds/php/phpunit_cmd.rs
```diff
@@ -0,0 +1,358 @@
...
+        result.push_str(&format!(
+            "\n... +{} more failures\n",
+            failures.len() - MAX_FAILURES_SHOWN
+        ));
+    }
+
+    result.trim().to_string()
+}
+
+fn parse_counts(output: &str) -> (usize, usize, usize, usize) {
```
KuSh: nitpick: This function accumulates both `Failures:` and `Errors:` as failures. Could be misleading

src/cmds/php/phpstan_cmd.rs
```diff
@@ -0,0 +1,496 @@
...
+                }
+                break;
+            }
+            if a.starts_with("--error-format=") && a != "--error-format=json" {
+                found = true;
+                break;
+            }
+        }
+        found
+    };
```
KuSh: nitpick: instead of returning a boolean we could directly return the custom format used so that if it is already json we don't insert another `--error-format json` arg

KuSh (changes requested):
Hi @iliaal, this looks mostly good to me, just a few minor nitpicks and suggestions. Sorry the PR is a bit huge; I wasn’t able to review it all in one go

src/core/utils.rs
iliaal: Done in `5d2928c`. Wrapped `composer_bin_dirs()` in a `OnceLock` — resolution depends only on env + cwd `composer.json`, both constant per process, so it reads once now instead of per segment. Went with the cache over threading a param since `classify_command` has many callers.

src/cmds/php/phpstan_cmd.rs
iliaal: Done in `7cc46b4`. Replaced the bool with a 3-state `ErrorFormat` enum (`Unspecified`/`Json`/`Custom`): an explicit `--error-format=json` is now kept as-is with no duplicate flag appended, and only `Custom` routes to the text filter.

src/cmds/php/utils.rs
iliaal: Done in `28366ce`. Removed the `pest.php` check — confirmed Pest has no root `pest.php` (bootstrap is `tests/Pest.php`), so it never matched real Pest projects and only false-positived. `composer_tool_exists("pest")` is now the sole marker.

src/cmds/php/phpunit_cmd.rs
iliaal: Fixed in `8825480`. `filter_phpunit_output` now calls `strip_ansi_and_controls` first, matching the other PHP filters. Audited the rest: ecs/artisan/php/test_output already strip; `filter_phpstan_text` was the other gap (it substring-matches `[OK]`/`Found N errors`), so I added stripping there too. The JSON paths (pint, phpstan json) don't need it. Regression test feeds `\x1b[30;42mOK\x1b[0m (9 tests, 20 assertions)` and asserts the count survives.

src/cmds/php/phpunit_cmd.rs
iliaal: Good catch — addressed in `8825480`. `parse_counts` now tracks errors separately and the summary reads `N failures, M errors` (errors shown only when present) instead of lumping exceptions under "failures".

src/cmds/system/pipe_cmd.rs
iliaal: Fixed in `50a8743`. Now requires `first_trimmed.starts_with("PHPUnit ")` in addition to the author string, so a LICENSE/composer/`git log` mentioning Sebastian Bergmann passes through untouched. Added a negative test (git-log-style input → identity).

src/cmds/php/phpstan_cmd.rs
iliaal: Fixed in `7cc46b4`. Detection now scans all args (`is_analyse_command`) instead of just `first()`, so `phpstan -c phpstan.neon analyse src/` is filtered. Test covers the global-flag-before-subcommand case plus `--version`/`clear-result-cache` staying passthrough.

KuSh (changes requested):
Could you add tests for paratest? That’s what’s causing the CI failure. See [.claude/rules/cli-testing.md](https://github.com/rtk-ai/rtk/blob/develop/.claude/rules/cli-testing.md) for the full testing guide. After that, LGTM!

iliaal:
Added in `fe75595`. `paratest_cmd.rs` was missing the mandatory `#[cfg(test)]` module — that's what tripped the `check-test-presence` gate. The two tests exercise the ParaTest-specific branch of `filter_test_runner_output` (banner, `Random Seed:`, and dot-progress stripped; result summary and failure details preserved), rather than duplicating pest's coverage.

Verified locally: `bash scripts/check-test-presence.sh origin/develop` now passes all 9 modules, fmt/clippy clean.

KuSh (approved):

KuSh:
LGTM, thanks for all the work!

PR #1657, opened 2026-05-01
fix(tee): render absolute log path in format_hint

### LINKED ISSUE #1644: Tee hint path with ~ shorthand breaks on paths with spaces (not shell-safe)
The path shown in rtk output is not directly usable when it contains spaces and uses ~.

Example:
`rtk curl -fsSL https://raw.githubusercontent.com/rtk-ai/rtk/refs/heads/master/install.sh`

Output includes:
`[full output: ~/Library/Application Support/rtk/tee/1777536940_curl.log]`

Trying to use this path directly fails:
`cat "~/Library/Application Support/rtk/tee/1777536940_curl.log"`

or even:
`cat ~/Library/Application Support/rtk/tee/1777536940_curl.log`

Both do not work as expected in bash/zsh.

However, using the absolute path works:
`cat "$HOME/Library/Application Support/rtk/tee/1777536940_curl.log"`

This is problematic for LLM/agent-based workflows, where the path is often copied verbatim and reused in subsequent commands. Since ~ does not expand inside quotes and spaces are not escaped, the file cannot be accessed, which may lead the agent to incorrectly assume the file does not exist.

File to fix: src/core/tee.rs
Location: Lines 172–183, function format_hint()
                                                                                                                                                                                                                                                                                                                                                                                                                                                                            Current code:                                                     
```
/// Format the hint line with ~ shorthand for home directory.                                                                                                                                                                           LSP                                     
     fn format_hint(path: &std::path::Path) -> String {                                                                                                                                                                                      LSPs will activate as files are read    
         let display = if let Some(home) = dirs::home_dir() {                                                                                                                                                                                                                        
             if let Ok(relative) = path.strip_prefix(&home) {                                                                                                                                                                                                                        
                 format!("~/{}", relative.display())                                                                                                                                                                                                                                 
             } else {                                                                                                                                                                                                                                                                
                 path.display().to_string()                                                                                                                                                                                                                                          
             }                                                                                                                                                                                                                                                                       
         } else {                                                                                                                                                                                                                                                                    
             path.display().to_string()                                                                                                                                                                                                                                              
         };                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                     
         format!("[full output: {}]", display)                                                                                                                                                                                                                                       
     }
```
Problem:                                                                                                                                                                                                                                                                        
The format_hint() function formats tee file paths using ~ shorthand (e.g., ~/Library/Application Support/rtk/tee/123_curl.log). This causes issues in shell environments when:                                                                                                  
1. The path contains spaces (e.g., Application Support)                                                                                                                                                                                                                         
2. The ~ does not expand inside double quotes in bash/zsh                                                                                                                                                                                                                       
3. LLM/agent workflows copy the path verbatim, making it unusable directly

Proposed fix:
Replace ~ shorthand with the absolute path using $HOME or full absolute path:

```
/// Format the hint line with absolute path (shell-safe, no ~ shorthand).
    let absolute_path = std::fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf());
    format!("[full output: {}]", absolute_path.display())
}
```

Benefits:
- Absolute paths work reliably in all shell contexts
- No ambiguity with ~ expansion rules
- LLM/agents can copy-paste the path directly without modification
- Fixes workflows where the hint path is reused in subsequent commands

truffle-dev:
## Summary

Tee writes the captured output to a file under `dirs::data_local_dir()` and prints a hint line so a downstream consumer can read it back. When the file landed under `$HOME`, the hint was rendered with the `~` shorthand:

```
[full output: ~/Library/Application Support/rtk/tee/7_curl.log]
```

That string is not safe to copy verbatim into a shell:
- `cat "~/Library/Application Support/rtk/tee/..."` does not expand `~` inside double quotes
- `cat ~/Library/Application Support/rtk/tee/...` splits on the space and fails

Both paths the hint advertises are paths the consumer is expected to use as-is, so the shorthand is a trap on macOS (default data dir is `~/Library/Application Support`) and on any Linux user that puts spaces in their home path.

## Fix

Drop the `~` strip-prefix logic and render the absolute path. The path passed in is already absolute (built from `data_local_dir()`), so this is a one-line change to `format_hint`. The `[full output: ...]` enclosure stays.

`canonicalize` would also work but requires the file to exist at hint time, and the hint is built before the writer flushes, so absolute-only is the simpler fit.

## Tests

Two regression tests cover the home-prefixed and space-containing shapes. `cargo test --all` 1689 passed.

Closes #1644

### DIFF
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -168,19 +168,12 @@ pub fn tee_raw(raw: &str, command_slug: &str, exit_code: i32) -> Option<PathBuf>
     )
 }
 
-/// Format the hint line with ~ shorthand for home directory.
+/// Format the hint line with the absolute log path so downstream
+/// consumers can use it verbatim. `~` shorthand does not expand
+/// inside double quotes and an unquoted `~/path with spaces` splits
+/// on the space, so paths under `$HOME` are rendered absolute.
 fn format_hint(path: &std::path::Path) -> String {
-    let display = if let Some(home) = dirs::home_dir() {
-        if let Ok(relative) = path.strip_prefix(&home) {
-            format!("~/{}", relative.display())
-        } else {
-            path.display().to_string()
-        }
-    } else {
-        path.display().to_string()
-    };
-
-    format!("[full output: {}]", display)
+    format!("[full output: {}]", path.display())
 }
 
 /// Convenience: tee + format hint in one call.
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Removes the `~` home directory shorthand from the tee hint line, rendering the absolute path instead. The `~` shorthand broke shell copy-paste because it doesn't expand inside double quotes and splits on spaces when unquoted. Two regression tests are added to verify absolute paths and paths with spaces.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1644

---

truffle-dev:
Superseded by #2187. The  shorthand was extracted into a new `display_path` helper in #1895, so rebasing this branch into the current shape was simpler than chasing the diff. #2187 applies the same fix to the new helper and keeps the same regression tests.

PR #1660, opened 2026-05-01
fix: report ok for silent npm scripts

sdwolf4103:
## Summary
- Treat npm 7+ leading lifecycle echo lines as boilerplate so successful silent scripts fall back to `ok`.
- Preserve substantive script output, including `> ` lines after real output and a single leading `> ` line that may be user output.
- Add regression tests for lifecycle-only output and over-filtering edge cases.

## Verification
- `git diff --check -- src/cmds/js/npm_cmd.rs`
- Static logic verification of the added npm filter cases

Note: local environment did not have `cargo`/`rustc`, so Rust tests should run in CI.

### DIFF
--- a/src/cmds/js/npm_cmd.rs
+++ b/src/cmds/js/npm_cmd.rs
@@ -135,17 +135,24 @@ fn run_filtered(name: &str, args: &[String], verbose: u8, skip_env: bool) -> Res
 /// Filter npm run output - strip boilerplate, progress bars, npm WARN
 fn filter_npm_output(output: &str) -> String {
     let mut result = Vec::new();
+    let mut seen_substantive_output = false;
+    let mut leading_gt_lines = Vec::new();
 
     for line in output.lines() {
-        // Skip npm boilerplate
-        if line.starts_with('>') && line.contains('@') {
+        let trimmed = line.trim_start();
+
+        // npm 7+ emits script name and command as leading `> ` lines without
+        // package@version. Treat 2+ leading lines as lifecycle echo so silent
+        // scripts can fall back to `ok`, but preserve one as possible output.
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟡 **Risk** | `medium` |

### Summary

This PR fixes the npm output filter to treat npm 7+ leading lifecycle echo lines (e.g. script name and command) as boilerplate, so that silent scripts with no substantive output correctly report "ok". It preserves a single leading "> " line as possible user output and keeps "> " lines that appear after real output. Regression tests cover lifecycle-only output, substantive output preservation, and edge cases.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

---

aeppling:
Hello, and thanks for the contribution.

Blocker: the count heuristic (`leading_gt_lines.len() > 1` then drop) can swallow real error output from tools that prefix their stdout with `>` (vite/webpack verbose, jest, some custom scripts):

```
> build
> vite build
> Build failed: module not found
```

All three lines get classified as boilerplate, the LLM sees `ok` on a failed build. That is exactly the over-filtering case RTK has to avoid.

Also worth running `cargo test test_filter_npm_output` locally — the existing assertion that `> project@version` lines are stripped looks like it regresses when only a single workspace header arrives (no `> cmd` echo to make len ≥ 2).

Suggested: match the known npm 7+ header shape (`> name@version <verb>` or bare `> scriptname`) rather than counting consecutive `>` lines. Real fixtures under `tests/fixtures/` would catch the ANSI / CRLF / workspace cases the synthetic tests miss.

PR #1661, opened 2026-05-01
fix(gh): passthrough --help/-h on pr subcommands (#1474)

### LINKED ISSUE #1474: rtk gh pr comment --help outputs `ok commented` — filter discards stdout unconditionally
`rtk gh pr comment --help` (and `rtk gh pr edit --help`) output `ok commented` / `ok edited` instead of the actual `gh` help text. The `pr_action` filter unconditionally discards stdout and synthesizes a confirmation message, regardless of what `gh` actually produced.

## Reproduction

```
# Raw gh — shows the help text
RTK_DISABLED=1 gh pr comment --help

# Through rtk — just "ok commented"
gh pr comment --help
```

I reproduced this with `rtk 0.37.2` on Windows.

## Cause

`pr_action` runs the `gh` command and then replaces stdout with `ok_confirmation(action, pr_num)` — the closure ignores its `_stdout` argument:

https://github.com/rtk-ai/rtk/blob/80a6fe606f73b19e52b0b330d242e62a6c07be42/src/cmds/git/gh_cmd.rs#L911-L931

`gh`'s help output is thrown away and replaced with `"ok {action} #{N}"` — with an empty PR number, since `--help` starts with `-` and no positional identifier is found.

Both `gh pr comment` and `gh pr edit` go through this path:

https://github.com/rtk-ai/rtk/blob/80a6fe606f73b19e52b0b330d242e62a6c07be42/src/cmds/git/gh_cmd.rs#L214-L215

`pr_create` has a structurally similar pattern — its closure assumes stdout is a PR URL and extracts a number from it. I haven't tested `gh pr create --help`, but the code would parse help text as a URL:

https://github.com/rtk-ai/rtk/blob/80a6fe606f73b19e52b0b330d242e62a6c07be42/src/cmds/git/gh_cmd.rs#L835-L857

## Impact

`--help` is how agents (and humans) discover what flags a command supports. When RTK replaces help output with a fake confirmation, the agent has no way to learn the right flags — it sees `ok commented`, concludes the command worked, and moves on. `RTK_DISABLED=1 gh <cmd> --help` works as a workaround, but the agent only knows to reach for it once something else tips it off that the output is wrong.

## Suggestion

`should_passthrough_pr_status` already treats `--help` / `-h` as a passthrough trigger for `gh pr status`:

https://github.com/rtk-ai/rtk/blob/80a6fe606f73b19e52b0b330d242e62a6c07be42/src/cmds/git/gh_cmd.rs#L305-L312

The same check could be applied before `pr_action` and `pr_create` (and possibly other gh subcommands) — when `--help` or `-h` is present, passthrough raw so `gh` can print its own help.

## Broader Pattern

This is the same class of bug as #720 and #730 — a filter that doesn't account for a flag silently produces wrong output, and the hook makes it invisible to the AI agent. `pr_action` goes one step further and fabricates a success confirmation from no data, which is the most misleading variant so far.

## Related

- #720 — `--comments` silently dropped on `gh issue/pr view` (same class: flag ignored by filter)
- #730 — `--name-only` produces empty output on `gh pr diff` (same class: filter applied to non-matching data)

## Environment

- rtk 0.37.2
- gh 2.x
- Claude Code
- Windows 11

season179:
Closes #1474.

`rtk gh pr comment --help` returns ``ok commented``, `pr edit --help` returns ``ok edited``, and `pr create --help` stitches ``ok created`` onto mangled help text. `pr view`/`checks`/`diff --help` are also broken — they error with ``PR number required`` or drop output silently. Six of nine `gh pr` subcommands corrupt their help output.

Cause: `pr_action` and `pr_create` fabricate confirmations from `ok_confirmation()` regardless of stdout; `view_pr`/`pr_checks` require a positional PR number; `pr_diff` runs `compact_diff` over non-diff text. None of them check for `--help`/`-h`.

Fix: add a `has_help_flag` helper next to the existing `has_json_flag`, and short-circuit `run_pr` to passthrough when the flag is present. Same family as #720 / #730 — flag-aware passthrough at the dispatcher.

Verified locally on macOS / rtk 0.38.0:

| subcommand | before | after |
|---|---|---|
| comment / edit | ``ok commented`` / ``ok edited`` | real help |
| create | ``ok created`` + mangled help | real help |
| view / checks | ``PR number required`` | real help |
| diff | empty stdout | real help |
| status / merge / list | ✓ already worked | ✓ unchanged |

`-h` short flag also works. Filter path for non-help calls is unaffected (`rtk gh pr list -L 3` still shows the compact format).

## Test plan
- [x] 5 unit tests on `has_help_flag` (long, short, absent, substring resistance, `--body=--help` value form)
- [x] `cargo fmt --all && cargo clippy --all-targets -- -D unsafe_code && cargo test --all` — 1691 + 5 new pass
- [x] Live binary check across all 9 `gh pr` subcommands
- [x] Regression sanity: filtered `rtk gh pr list` output unchanged


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes six of nine `gh pr` subcommands whose `--help`/`-h` output was corrupted by rtk filters fabricating confirmations or requiring positional args. Adds a `has_help_flag` helper and short-circuits `run_pr` to passthrough when the flag is present, letting `gh` print its native help text.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1474

---

dergachoff:
bumping this pr since issue is very annoying

PR #1662, opened 2026-05-01
fix(learn): drop parallel-tool-call cancellations from corrections

### LINKED ISSUE #1659: rtk learn: parallel-tool-call cancellations misclassified as CLI errors, swamping output
## Problem

`rtk learn -w` on a 30-day window (795 sessions, this is a Claude Code project) produced a 5,060-line / 251 KB `.claude/rules/cli-corrections.md`. The vast majority of \"corrections\" are not corrections at all — they are user-initiated cancellations of fan-out parallel tool calls, which Claude Code surfaces as:

```
<tool_use_error>Cancelled: parallel tool call Bash(...) errored</tool_use_error>
```

When the user Ctrl-Cs one branch of an N-way parallel call, RTK reads the next adjacent command as \"the working alternative\" and emits a rule of the form `Use X not Y`. Both X and Y are usually fine — the user just stopped one branch.

## Repro

A typical entry from the generated file:

```
[7x] git show <sha> --no-stat -p -- backend/app/features/memory/data/memory_repository.py
   →  git show <sha> --no-stat -p -- backend/app/features/memory/data/schema.py
   Error: <tool_use_error>Cancelled: parallel tool call Bash(git show <sha> --no-stat -p -- ...)
```

There is no error in either command. The user fired both `git show` calls in parallel for two different files and cancelled the slower one. RTK now believes \"don't use the memory_repository.py path, use schema.py instead.\"

## Impact

- Output is dominated by these false positives — manual triage is impractical at 5K lines.
- Loading the file into every session would blow the context budget for negligible signal.
- Real corrections (e.g. `uv run ruff check` → `uv run pyrefly check` for typecheck) are buried under hundreds of fake ones.
- File is unusable as-is for any project that uses parallel tool calls heavily (i.e., most modern Claude Code projects).

## Suggested fixes

1. **Filter `tool_use_error: Cancelled` from the input set entirely.** It's a user signal, not a CLI failure. The error text is consistent and easy to match.
2. Treat \"parallel tool call errored\" specifically as cancellation — the sibling commands in the same `parallel` group are co-equal, not before/after.
3. Consider filtering by exit code semantics: only treat exits caused by the *callee* (non-zero from the tool) as corrections, not exits caused by the *caller* (Claude / user cancel).
4. Raising `--min-confidence` / `--min-occurrences` defaults won't fix this — the cancellations are highly repetitive and would clear any threshold.

## Related

- #654 (generalizable patterns) — adjacent but different concern.
- #651 (sanitize sensitive data) — also relevant; the file leaked internal pod names, AWS account IDs, spreadsheet IDs.

Versions: rtk 0.38.0 / Claude Code (current).

season179:
Closes #1659.

`rtk learn` treated Claude Code's parallel-tool-call cancellations as CLI errors, so the next adjacent command got emitted as the "fix" — dominating the output (reporter: 5060 lines / 251 KB on a 30-day window).

Filter the cancellation marker at both sides of the correction-pair search: `is_command_error` rejects it as wrong-side; the inner-loop iterator filters before `take(WINDOW)` so cancellations don't consume lookahead slots and hide a real fix after them. Gated on `is_error=true` so a successful command whose stdout merely mentions the phrase stays eligible.

5 new unit tests cover the issue's repro plus edge cases (window-consumption, success-with-phrase, all-caps, mid-stream mention, leading whitespace).

### DIFF
--- a/src/learn/detector.rs
+++ b/src/learn/detector.rs
@@ -73,11 +73,30 @@ lazy_static! {
     static ref USER_REJECTION_RE: Regex = Regex::new(
         r"(?i)(user (doesn't want|declined|rejected|cancelled)|operation (cancelled|aborted) by user)"
     ).unwrap();
+
+    // Claude Code emits this when the user Ctrl-Cs one branch of an N-way parallel
+    // tool call. Both the cancelled and surviving branches are usually fine commands —
+    // treating either as "wrong" or "right" yields bogus corrections. See issue #1659.
+    //
+    // Anchored with ^\s* so a build log or error message that merely echoes the
+    // phrase mid-stream is NOT matched — only outputs that ARE the cancellation
+    // event (Claude emits the marker as the entire content of the tool_result).
+    static ref PARALLEL_TOOL_CALL_CANCELLATION_RE: Regex = Regex::new(
+        r"(?i)^\s*(?:<tool_use_error>\s*)?(?:tool_use_error:\s*)?cancelled:\s*parallel tool call\b"
+    ).unwrap();
 }
 
-/// Filters out user rejections - requires actual error-indicating content
+fn is_parallel_tool_call_cancellation(output: &str) -> bool {
+    PARALLEL_TOOL_CALL_CANCELLATION_RE.is_match(output)
+}
+
+/// Filters out user rejections and parallel-tool-call cancellations - requires actual error-indicating content
 pub fn is_command_error(is_error: bool, output: &str) -> bool {
     if !is_error {
+        return false;
+    }
+
+    if is_parallel_tool_call_cancellation(output) {
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Filters out Claude Code's parallel-tool-call cancellation markers from the correction-pair detection in `rtk learn`. The cancellation events were being misclassified as CLI errors, causing the next adjacent command to be emitted as a bogus "fix" and dominating output. The fix adds a regex-based filter at both the error-detection and lookahead-window stages of correction finding.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1659

---

PR #1673, opened 2026-05-02
refactor(json): Single-line output with truncation hints

KuSh:
## Summary
- Consolidate json_cmd output to single-line format and add user-facing truncation feedback.
- Also fixes UTF-8 correctness by using char-based truncation instead of byte boundaries.

## Test plan
- [x] Manual testing: `target/debug/rtk json openclaw/package.json` output inspected

### DIFF
--- a/src/cmds/cloud/aws_cmd.rs
+++ b/src/cmds/cloud/aws_cmd.rs
@@ -256,7 +256,7 @@ fn run_generic(subcommand: &str, args: &[String], verbose: u8, full_sub: &str) -
         return Ok(crate::core::utils::exit_code_from_output(&output, "aws"));
     }
 
-    let filtered = match json_cmd::filter_json_string(&raw, JSON_COMPRESS_DEPTH) {
+    let filtered = match json_cmd::filter_json_schema(&raw, JSON_COMPRESS_DEPTH) {
         Ok(schema) => {
             println!("{}", schema);
             schema
--- a/src/cmds/system/json_cmd.rs
+++ b/src/cmds/system/json_cmd.rs
@@ -6,6 +6,7 @@ use serde_json::Value;
 use std::fs;
 use std::io::{self, Read};
 use std::path::Path;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 /// Reject non-JSON files with a clear error before doing any I/O.
 fn validate_json_extension(file: &Path) -> Result<()> {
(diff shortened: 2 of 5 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ♻️ **Type** | `refactor` |
| 🟡 **Risk** | `medium` |

### Summary

Refactors JSON command output from multi-line indented format to single-line compact format, adds truncation tracking with user-facing hints, and fixes UTF-8 correctness by switching from byte-boundary truncation to char-based truncation. The function `filter_json_string` is renamed to `filter_json_schema` and the compact output now returns a truncation flag.

### Review Checklist
- [ ] Tests present
- [x] Breaking change
- [ ] Docs updated

---

aeppling:
### Blocker 1: TeeMode::Always is broken for tee_and_hint (regression, reproduced)

content_hint::save_output_on_failure returns None on exit_code == 0 before it reads config.tee.mode, so the Always arm of the mode match is unreachable. The old tee::should_tee proceeded for Always regardless of exit code.

Reproduced with a real mode = "always" config.

So with mode = "always", every tee_and_hint caller that succeeds (runner::print_with_hint, the streamed runner, aws_cmd.rs:369 which passes 0 literally) silently stops teeing, while force_tee_hint callers (curl, aws-truncated) still do, the setting is now half-applied. Not mentioned in the PR.

Fix: check mode before the exit_code == 0 shortcut, or deliberately drop Always here and update the mode match + TeeConfig tests to say so.


### Blocker 2: the move deleted the tests that catch Blocker 1

tee.rs had 18 test fns. PR keeps 5 in tee.rs, migrates 3 to content_hint.rs. 10 are deleted with no equivalent anywhere:

- 6 × test_should_tee_*: the mode × exit-code matrix, including test_should_tee_always_mode_success (exactly the case Blocker 1 breaks → CI stays green).
- 4 × test_write_tee_file_*: including the UTF-8-boundary and emoji truncation cases. content_hint::write_content_file still does that char_indices().take_while(...) truncation at max_file_size, now with zero coverage.

CONTRIBUTING.md requires tests for every change; relocated logic must bring its tests. Re-adding the should_tee-equivalent ones will fail until Blocker 1 is fixed; which is the point.

KuSh:
@aeppling sorry about that, I've restored the behavior and associated tests. Thanks for the review

KuSh:
obsolete, will need to be entirely redone if still needed

PR #1675, opened 2026-05-02
fix(ls): preserve permission info as octal when -l/-la is passed

### LINKED ISSUE #1672: The result of `rtk ls -la` is equivalent to `rtk ls`, which leads to the loss of some info.
### Problem

When running `rtk ls -la`, the permissions column (`-rw-r--r--`) and ownership columns (`user group`) are entirely stripped from output. The result:

```
# Native ls -la
-rw-r--r--@  1 user  staff  6144  May  1 23:00 .DS_Store
-rw-r--r--   1 user  staff   984  May  1 22:00 __init__.py

# rtk ls -la
.DS_Store  6.0K
__init__.py  984B
```

Users pass `-la` specifically to **check permissions and ownership**. Removing these columns makes `-la` equivalent to plain `ls` and forces a re-run without rtk when permissions actually matter.

### Proposed Solution

Convert the permissions column to octal notation instead of deleting it:

```
# Expected
644  .DS_Store  6.0K
644  __init__.py  984B
```

Octal (`644`, `755`, `600`) is:
- **More compact** than `-rw-r--r--` (still saves tokens)
- **More precise** — no ambiguity about setuid/setgid/sticky bits
- **Machine-parseable** — consistent 3-4 digit format
- **What programmers already think in** — nobody mentally translates `rwx` to bits

### Comparison

| Approach | Token count | Info preserved |
|----------|-------------|----------------|
| Raw `-rw-r--r--@` | ~13 chars | Full |
| Strip entirely (current) | 0 | None |
| Octal `644` | 3-4 chars | Permission bits |
| Octal + owner | 3-4 + N chars | Permission + ownership |

The octal approach is the sweet spot — compact yet informative.

ryana-0154:
## Summary
  - Fixes #1672 — `rtk ls -la` was producing output identical to `rtk ls` because the filter stripped permission/ownership entirely.
  - When the user passes `-l` (or `-la`), each entry is now prefixed with octal perms (e.g. `644`, `755`), matching the proposal in the issue.
  - Special bits (setuid/setgid/sticky) emit a 4-digit form (`4755`, `1777`, `7777`).
  - `rtk ls` (no `-l`) is unchanged.

  ## Example
  Before:
  .DS_Store  6.0K
  init.py  984B

  After (`rtk ls -la`):
  644  .DS_Store    6.0K
  644  init.py  984B

  ## Test plan
  - [x] `cargo clippy --all-targets` — 0 errors (pre-existing warnings unchanged)
  - [x] `cargo test --all` — 1692 passed, 6 ignored
  - [x] New unit tests for `perms_to_octal` (common perms, setuid/setgid/sticky, garbage input) and for long vs. short output formatting

### DIFF
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -25,10 +25,10 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         .any(|a| (a.starts_with('-') && !a.starts_with("--") && a.contains('a')) || a == "--all");
 
     // Per `man ls`, the long listing is triggered by `-l` and also implied by
-    // `-g`, `-n`, `-o`, and `--full-time`. In any of those cases we preserve
-    // permission info as octal.
+    // `-g`, `-n`, `-o`, `--full-time` or GNU `--format=long` and `--format=verbose`.
+    // In any of those cases we preserve permission info as octal.
     let show_long = args.iter().any(|a| {
-        if a == "--full-time" {
+        if a == "--full-time" || a == "--format=long" || a == "--format=verbose" {
             return true;
         }
         if a.starts_with('-') && !a.starts_with("--") {
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -24,6 +24,19 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         .iter()
         .any(|a| (a.starts_with('-') && !a.starts_with("--") && a.contains('a')) || a == "--all");
 
+    // Per `man ls`, the long listing is triggered by `-l` and also implied by
+    // `-g`, `-n`, `-o`, and `--full-time`. In any of those cases we preserve
+    // permission info as octal.
+    let show_long = args.iter().any(|a| {
+        if a == "--full-time" {
+            return true;
+        }
+        if a.starts_with('-') && !a.starts_with("--") {
+            return a.chars().any(|c| matches!(c, 'l' | 'g' | 'n' | 'o'));
+        }
+        false
+    });
+
     let flags: Vec<&str> = args
         .iter()
         .filter(|a| a.starts_with('-'))
@@ -74,7 +87,7 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         "ls",
         &format!("-la {}", target_display),
         |raw| {
-            let (entries, summary, parsed_count) = compact_ls(raw, show_all);
+            let (entries, summary, parsed_count) = compact_ls(raw, show_all, show_long);
 
             // If no lines were parsed (e.g., unrecognized locale), fall back to raw output.
             // This is safer than returning "(empty)" for a non-empty directory.
@@ -124,14 +137,17 @@ fn human_size(bytes: u64) -> String {
     }
 }
 
-/// Parse a single `ls -la` line, returning `(file_type_char, size, name)`.
+/// Parse a single `ls -la` line, returning `(file_type_char, perms, size, name)`.
+///
+/// `perms` is the raw 10-char string from ls (e.g. `-rw-r--r--`); use
+/// [`perms_to_octal`] to render it.
 ///
 /// Uses the date field as a stable anchor — the date format in `ls -la` is
 /// always three tokens (`Mon DD HH:MM` or `Mon DD  YYYY`), so we locate it
 /// with a regex, then extract size (rightmost number before the date) and
 /// filename (everything after the date). This handles owner/group names that
 /// contain spaces, which break the old fixed-column approach.
-fn parse_ls_line(line: &str) -> Option<(char, u64, String)> {
+fn parse_ls_line(line: &str) -> Option<(char, String, u64, String)> {
     // Skip . and .. entries before date parsing (works for non-English locales too)
     if is_dotdir(line) {
         return None;
@@ -146,7 +162,7 @@ fn parse_ls_line(line: &str) -> Option<(char, u64, String)> {
         return None;
     }
 
-    let perms = before_parts[0];
+    let perms = before_parts[0].to_string();
     let file_type = perms.chars().next()?;
 
     // Size is the rightmost parseable number before the date.
@@ -160,7 +176,7 @@ fn parse_ls_line(line: &str) -> Option<(char, u64, String)> {
         }
     }
 
-    Some((file_type, size, name))
+    Some((file_type, perms, size, name))
 }
 
 /// Returns true if the line represents a . or .. directory entry.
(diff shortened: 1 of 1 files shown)


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| 🐛 **Type** | `bug-fix` |
| 🟢 **Risk** | `low` |

### Summary

Fixes the `rtk ls -la` command which was producing output identical to `rtk ls` by preserving permission information as octal notation when `-l` is passed. Adds a `perms_to_octal` function to convert ls-style permission strings to octal format, handling special bits (setuid/setgid/sticky) with a 4-digit form. Existing `rtk ls` behavior without `-l` is unchanged.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #1672

---

src/cmds/system/ls.rs
```diff
@@ -23,6 +23,12 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         .iter()
         .any(|a| (a.starts_with('-') && !a.starts_with("--") && a.contains('a')) || a == "--all");
 
+    // `-l` (or any short cluster containing 'l') means the user explicitly asked
+    // for the long listing, so we preserve permission info as octal.
+    let show_long = args
+        .iter()
+        .any(|a| a.starts_with('-') && !a.starts_with("--") && a.contains('l'));
```
KuSh: according to [man ls](https://www.man7.org/linux/man-pages/man1/ls.1.html) this should also be true with `-g`, `-n`, `-o` and `--full-time`

src/cmds/system/ls.rs
```diff
@@ -144,26 +153,73 @@ fn parse_ls_line(line: &str) -> Option<(char, u64, String)> {
...
-/// Parse ls -la output into compact format:
-///   name/  (dirs)
-///   name  size  (files)
+/// Convert an `ls`-style permission string (e.g. `-rw-r--r--`, `drwxr-xr-x`,
+/// `-rwsr-xr-t`) into octal notation (e.g. `644`, `755`, `4755`).
+///
+/// Returns an empty string if the input does not look like a permission field.
+/// Special bits (setuid/setgid/sticky) are encoded as a leading 4th digit when
+/// any are set; otherwise we emit a 3-digit value to stay compact.
+fn perms_to_octal(perms: &str) -> String {
```
KuSh: An Option<String> would be better. Returning None instead of an empty string if input doesn't look like a permission field

src/cmds/system/ls.rs
```diff
@@ -144,26 +153,73 @@ fn parse_ls_line(line: &str) -> Option<(char, u64, String)> {
...
+
+    let owner = (if b[1] == b'r' { 4 } else { 0 })
+        + (if b[2] == b'w' { 2 } else { 0 })
+        + (if owner_x { 1 } else { 0 });
+    let group = (if b[4] == b'r' { 4 } else { 0 })
+        + (if b[5] == b'w' { 2 } else { 0 })
+        + (if group_x { 1 } else { 0 });
+    let other = (if b[7] == b'r' { 4 } else { 0 })
+        + (if b[8] == b'w' { 2 } else { 0 })
+        + (if other_x { 1 } else { 0 });
```
KuSh: nitpick, more readable:

```suggestion
    fn perm_value(read: bool, write: bool, exec: bool) -> u32 {
        ((read as u32) << 2) | ((write as u32) << 1) | (exec as u32)
    }

    let owner = perm_value(b[1] == b'r', b[2] == b'w', owner_x);
    let group = perm_value(b[4] == b'r', b[5] == b'w', group_x);
    let other = perm_value(b[7] == b'r', b[8] == b'w', other_x);
    ...
    let special = perm_value(setuid, setgid, sticky);
```

src/cmds/system/ls.rs
```diff
@@ -197,13 +255,21 @@ fn compact_ls(raw: &str, show_all: bool) -> (String, String) {
     let mut entries = String::new();
 
     // Dirs first, compact
-    for d in &dirs {
-        entries.push_str(d);
+    for (name, octal) in &dirs {
+        if show_long && !octal.is_empty() {
```
KuSh: octal should not have been computed in the first time if show_long is false

KuSh (changes requested):
Thanks for the PR, that's a great addition but it needs some minor changes first!

KuSh (changes requested):
Thanks for the PR, that's a great addition but it needs some minor changes first!

src/cmds/system/ls.rs
ryana-0154: Good catch — fixed in 3f5279d6114c09d5df5facbf1ae9f90513de8a2d: `show_long` now also triggers on `-g`, `-n`, `-o`, and `--full-time`.

src/cmds/system/ls.rs
ryana-0154: Done in 3f5279d6114c09d5df5facbf1ae9f90513de8a2d — `perms_to_octal` now returns `Option<String>`, with `None` for malformed input.

src/cmds/system/ls.rs
ryana-0154: Applied in 3f5279d6114c09d5df5facbf1ae9f90513de8a2d — extracted `perm_value(read, write, exec)` and used it for owner/group/other plus the special-bits digit.

src/cmds/system/ls.rs
ryana-0154: Fixed in 3f5279d6114c09d5df5facbf1ae9f90513de8a2d — octal is now only computed when `show_long` is true (returns `None` otherwise).

src/cmds/system/ls.rs
```diff
@@ -411,18 +486,19 @@ mod tests {
...
         assert_eq!(ft, '-');
+        assert_eq!(perms, "-rw-r--r--");
         assert_eq!(size, 1234);
         assert_eq!(name, "file.txt");
     }
 
     #[test]
     fn test_parse_ls_line_multiline_group() {
-        let (ft, size, name) = parse_ls_line(
+        let (ft, _, size, name) = parse_ls_line(
```
KuSh: Could you check perms in every existing tests instead of ignoring it? that will allow us to catch regression if any.
Otherwise LGTM!

KuSh (changes requested):

src/cmds/system/ls.rs
ryana-0154: Addressed in https://github.com/rtk-ai/rtk/commit/f0160ba83e98a3611266097eed5b6d91d01b95fd: existing parse_ls_line tests now assert the perms field instead of ignoring it, so any regression in permission parsing will be caught.

KuSh (approved):

KuSh (changes requested):
Hi @ryana-0154, you'll need to resolve conflicts before we can merge that one, otherwise LGTM!

ryana-0154:
> Hi @ryana-0154, you'll need to resolve conflicts before we can merge that one, otherwise LGTM!

Hi - there doesn't seem to be any merge conflicts to develop - can you confirm?

KuSh:
> > Hi @ryana-0154, you'll need to resolve conflicts before we can merge that one, otherwise LGTM!
> 
> Hi - there doesn't seem to be any merge conflicts to develop - can you confirm?

Let me recheck, I did have one when I tried earlier

KuSh:
@ryana-0154 there's a conflict, please rebase:

```sh
❯ git rebase upstream/develop
Auto-merging src/cmds/system/ls.rs
CONFLICT (content): Merge conflict in src/cmds/system/ls.rs
error: could not apply 39dee40... fix(ls): preserve permission info as octal when -l/-la is passed
hint: Resolve all conflicts manually, mark them as resolved with
hint: "git add/rm <conflicted_files>", then run "git rebase --continue".
hint: You can instead skip this commit: run "git rebase --skip".
hint: To abort and get back to the state before "git rebase", run "git rebase --abort".
hint: Disable this message with "git config set advice.mergeConflict false"
Could not apply 39dee40... # fix(ls): preserve permission info as octal when -l/-la is passed

KuSh:
You seem to have rebased on the wrong branch (certainly master) you have to fix that. 57 files changed all over the place is too much for the subject 😉

ryana-0154:
> You seem to have rebased on the wrong branch (certainly master) you have to fix that. 57 files changed all over the place is too much for the subject 😉

duh... what an idiot :D Fixed!

KuSh (approved):

PR #196, opened 2026-02-18
fix: passthrough --json/--jq/--template flags in all gh subcommands

shadowofdoom:
## Summary

- `gh pr view --json fields` failed with `Unknown JSON field: "--json"` because `view_pr()` treated `--json` as the PR identifier
- The v0.21.1 fix (#159) only covered `gh run view` via `should_passthrough_run_view()`
- This generalizes the fix with `has_output_format_flags()` and applies it to all handlers that hardcode `--json`: `list_prs`, `view_pr`, `pr_status`, `list_issues`, `view_issue`, `list_runs`, and `run_repo`
- Also fixes `view_pr` to handle omitted PR number (gh defaults to current branch)

## Reproduction

```
rtk gh pr view --json number,title,state
# Before: Unknown JSON field: "--json"
# After: passes through to gh correctly
```

## Test plan

- [x] All 383 tests pass (5 new tests for `has_output_format_flags`)
- [x] `rtk gh pr view --json number,title,state` passes through correctly
- [x] `rtk gh pr list` still uses RTK's compressed output
- [x] `rtk gh pr view <number>` still uses RTK's compressed output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1677, opened 2026-05-02
feat(curl): config-driven URL allowlist to bypass schema-mode rewrite

globalsecurepayments:
## Problem

`rtk curl` rewrites every `curl URL` to `rtk curl URL` and pipes the response
through `rtk json --schema`, which produces field-type literals (`field: int`,
`field: string`) and a `(N)` array-length suffix. That's a token-savings win for
human-facing exploration of arbitrary third-party APIs, but it actively breaks
downstream JSON parsing for **private or internal APIs** whose responses are
consumed by parsers (jq, `python json.load`, agents that round-trip JSON,
anything that needs the literal payload).

Today the only escape hatch is `[hooks] exclude_commands = ["curl"]`, which
disables curl rewriting entirely — losing the token savings on every other
`curl` invocation in the same session.

## Proposal

Add a narrow opt-in mechanism: a substring allowlist that bypasses the rewrite
on a per-URL basis. Default empty, so users who don't configure anything see
identical behavior to today.

```toml
[curl]
bypass_url_markers = [
  "localhost:8080/api/",
  "//internal.example.com/v1/",
]
```

Each marker is a substring match against the full command segment (after
env-prefix stripping, e.g. `PROXY=… curl …`), so it composes cleanly with
curl's flag positioning (`-X POST`, `-H`, `-d`, etc). Bypass is **per-segment**:
in `a && curl <internal> | jq`, the curl segment passes through unchanged
while `a` and the pipe target rewrite as normal.

Mirrors the existing #196 bypass shape for `gh --json/--jq/--template` (detect
a structured-output consumer, skip schema-mode filtering).

## API surface

- New `crate::core::config::CurlConfig { bypass_url_markers: Vec<String> }` in
  `src/core/config.rs`. Default empty.
- New `crate::discover::registry::RewriteOptions { curl_bypass_url_markers }`.
- New `rewrite_command_with_options(cmd, excluded, &opts)` for callers that
  want explicit control (tests, callers that already have config in hand).
- Existing `rewrite_command(cmd, excluded)` is unchanged for callers — it now
  reads `Config.curl.bypass_url_markers` from disk internally and forwards.

## Tests

- 4 new config tests: default-empty, `[curl]` roundtrip, missing-section
  tolerance.
- 9 new registry tests: localhost / loopback / hostname marker bypass; POST
  with headers + payload; multi-marker OR; default-empty preserves rewrite;
  unmatched URL still rewrites; port-specific narrowness; per-segment behavior
  in compound commands.
- Full suite: `cargo test` → **1699 passed, 0 failed, 6 ignored** on
  `aarch64-apple-darwin` against `master @ v0.38.0`.

## Real-world motivation

Hit while running `rtk` against an internal API server: agents reading
`curl … | jq` got `field: string` instead of the actual JSON values, leading
to three misfiled "the API is broken" reports before realising the schema
output was the rewrite, not the upstream response. The narrow allowlist
preserves token savings for the 95% case while letting users opt private
endpoints out cleanly.


pszymkowiak:

## 📊 Automated PR Analysis

| | |
|---|---|
| ✨ **Type** | `feature` |
| 🟢 **Risk** | `low` |

### Summary

Adds a config-driven URL substring allowlist (`[curl] bypass_url_markers`) that lets users opt specific curl invocations out of RTK's automatic `rtk curl ... | rtk json --schema` rewrite. This preserves raw JSON output for private/internal APIs consumed by downstream parsers, while keeping the default schema-mode rewrite for all other curl calls.

### Review Checklist
- [x] Tests present
- [ ] Breaking change
- [ ] Docs updated

**Linked issues:** #196

---

KuSh:
Thanks for the detailed write-up. Two changes on develop cover this: exclude_commands has matched regular expressions per command segment since #1398 (merged 2026-04-19), so exclude_commands = ["^curl .*localhost:8080/api/"] bypasses just those URLs while every other curl still gets filtered, including the env-prefixed and compound-command forms; and curl gained an IsTerminal gate on 2026-04-29 that passes the full body through whenever stdout is a pipe or a redirect, so jq and other parsers see the real JSON. The narrow per-URL escape hatch you wanted is therefore available without a curl-specific config key. Closing for that reason, and please comment or reopen if your setup needs something the regex form cannot express.

PR #1678, opened 2026-05-02
fix(grep): support --pcre2 flag

### LINKED ISSUE #1651: rtk fails to handle --pcre2 flag when agent tries to use rg
```
$ rtk grep -n --pcre2 "e2e-tests(?!-v2)"
/usr/bin/grep: unrecognized option '--pcre2'
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```

vyctorbrzezowski:
## Summary

Fixes #1651 by teaching `rtk grep` to parse `--pcre2` before the pattern and forward it to ripgrep instead of falling through to system `grep`.

The existing grep compatibility handling for `-r`/`--recursive` is preserved.

## Validation

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test pcre2`
- `cargo test grep_cmd`
- `cargo test test_grep_pcre2_before_pattern_parses`
- `cargo test` (`1689 passed, 6 ignored`)
- Manual smoke with `rg 15.1.0 features:+pcre2`

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -16,6 +16,7 @@ pub fn run(
     max_results: usize,
     context_only: bool,
     file_type: Option<&str>,
+    pcre2: bool,
     extra_args: &[String],
     verbose: u8,
 ) -> Result<i32> {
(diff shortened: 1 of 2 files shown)


KuSh:
Thanks for the fix and the thorough validation notes. This was solved more broadly in #2641 (merged 2026-06-26), which stopped substituting one search engine for the other: rtk grep now runs real grep and a new rtk rg runs real ripgrep with your flags passed through untouched, so --pcre2 and every other ripgrep flag work without needing individual clap definitions. Your test case runs correctly on current develop, and issue #1651 was closed against that change. The grep_cmd.rs file this patch edits has since been replaced by src/cmds/system/search.rs. If you find a flag that still does not make it through on the latest develop, please comment here or reopen.

Issue #1431, opened 2026-04-21
`rtk git diff` produces empty output when branch name contains `/`


binsee:
## Summary

`rtk git diff <branch>` silently produces no output when the branch name contains a forward slash (e.g. `feature/user-auth`, `main...feature/user-auth`). Native `git diff` works correctly. Exit code is 0, so there is no error signal.

## Root Cause

`normalize_diff_args` uses a string heuristic to decide whether to inject `--` before positional arguments:

```rust
fn looks_like_path(arg: &str) -> bool {
    arg.contains('/') || arg.contains('\\') || arg.starts_with('.') || arg.starts_with('~')
}
```

Any argument containing `/` is treated as a file path. This causes `rtk git diff feature/user-auth` to be silently rewritten as `git diff -- feature/user-auth`, which tells git to treat the branch name as a pathspec — no matching file exists, so output is empty.

## Reproduction

```bash
# Setup a fresh repo
mkdir /tmp/rtk-repro && cd /tmp/rtk-repro
git init
git commit --allow-empty -m "initial commit"
git checkout -b feature/user-auth
echo 'export function login(user, pass) { return auth(user, pass); }' > auth.js
echo 'export function logout() { clearSession(); }' >> auth.js
git add auth.js && git commit -m "add user auth module"
git checkout main
```

```bash
# Branch name with slash — native works, rtk empty
$ git diff feature/user-auth
diff --git a/auth.js b/auth.js
deleted file mode 100644
index 3f362f7..0000000
--- a/auth.js
+++ /dev/null
@@ -1,2 +0,0 @@
-export function login(user, pass) { return auth(user, pass); }
-export function logout() { clearSession(); }

$ rtk git diff feature/user-auth
(no output, exit code 0)

# Range syntax with slash — native works, rtk empty
$ git diff main...feature/user-auth
diff --git a/auth.js b/auth.js
new file mode 100644
index 0000000..3f362f7
--- /dev/null
+++ b/auth.js
@@ -0,0 +1,2 @@
+export function login(user, pass) { return auth(user, pass); }
+export function logout() { clearSession(); }

$ rtk git diff main...feature/user-auth
(no output, exit code 0)
```

## Expected Behavior

`rtk git diff feature/user-auth` and `rtk git diff main...feature/user-auth` should produce the same output as their native `git diff` equivalents.

## Workaround

Use a commit hash instead of a branch name:
```bash
rtk git diff $(git rev-parse feature/user-auth)
```

## Environment

[...]

PR #1679, opened 2026-05-02
fix(git): inject `--` for bare-word-with-extension paths in git diff

### LINKED ISSUE #1669: rtk does not handle git diff correctly in some cases
`rtk git diff -- "package.json" "playwright.config.ts" ".harness/e2e_pipeline.yaml" "e2e-tests/config/loadEnv.ts" "e2e-tests/config/test.env"`

output: `fatal: bad revision 'package.json'`

`git diff -- "package.json" "playwright.config.ts" ".harness/e2e_pipeline.yaml" "e2e-tests/config/loadEnv.ts" "e2e-tests/config/test.env"`

output: works fine

truffle-dev:
Closes #1669.

Reporter ran:

```
rtk git diff -- "package.json" "playwright.config.ts" ".harness/e2e_pipeline.yaml" "e2e-tests/config/loadEnv.ts" "e2e-tests/config/test.env"
```

and got `fatal: bad revision 'package.json'` while raw `git diff -- ...` worked.

## What was happening

clap's `trailing_var_arg = true` drops `--` when it's the first positional argument (existing comment at `src/cmds/git/git.rs:74-83`). `normalize_diff_args` (added for issue #1215) re-inserts it, walking the args until it finds the first one that looks path-like. Three rules existed before this PR:

1. Explicit prefix `.` or `~` → always a path.
2. Contains `/` or `\` → use `path_exists`.
3. Bare word → never a path (added for issue #1431 to keep `rtk git diff main` resolving as a branch).

For the reporter's command, the first path-looking arg by rule 1 is `.harness/e2e_pipeline.yaml` at index 2. `--` lands at index 2, leaving `package.json` (index 0) and `playwright.config.ts` (index 1) before the separator, where git treats them as revisions.

## The fix

A new rule between the slash-path rule and the plain-bare-word rule: a bare word with a mid-string `.` (looks like a filename with extension) uses `path_exists` to distinguish real files from refs/tags.

```rust
if arg.find('.').is_some_and(|i| i > 0) {
    return path_exists(arg);
}
```

`package.json` and `playwright.config.ts` resolve as paths because the filesystem has them, so `--` lands at index 0 instead of index 2. Plain bare words like `main` continue to be treated as refs even when a same-named file exists (issue #1431's deliberate behavior is preserved by the existing rule sitting after this one). A version-tag-shape arg like `v1.2.3` that doesn't exist as a file remains a ref candidate.

## Verification

Stash-bisect on a five-file repro that mirrors the reporter's exact command:

Before the patch:
```
$ rtk git diff -- "package.json" "playwright.config.ts" ".harness/e2e_pipeline.yaml" "e2e-tests/config/loadEnv.ts" "e2e-tests/config/test.env"
fatal: bad revision 'package.json'
```

After the patch:
```
$ rtk git diff -- "package.json" "playwright.config.ts" ".harness/e2e_pipeline.yaml" "e2e-tests/config/loadEnv.ts" "e2e-tests/config/test.env"
.harness/e2e_pipeline.yaml  | 2 +-
 e2e-tests/config/loadEnv.ts | 2 +-
 e2e-tests/config/test.env   | 2 +-
 package.json                | 2 +-
 playwright.config.ts        | 1 +
 5 files changed, 5 insertions(+), 4 deletions(-)
```

## Tests

Four new unit tests next to the existing ones in `mod tests`:

- `bare_word_with_extension_treated_as_path` — single `package.json` that exists.
- `bare_word_with_extension_no_injection_when_not_file` — `v1.2.3` tag pattern that doesn't exist on disk stays a ref.
- `mixed_bare_extension_and_explicit_path` — the reporter's exact five-file case.
- `ref_before_bare_word_extension` — `["HEAD", "Cargo.toml"]` → `["HEAD", "--", "Cargo.toml"]`.

`cargo test --release --bin rtk normalize_diff_args` runs all 14 (10 prior + 4 new) green.

## Scope notes

- `run_log` and `run_show` (`src/cmds/git/git.rs:447`, `:240`) do not call `normalize_diff_args`, so `rtk git log -- file.json` and `rtk git show -- file.json` likely have the same shape. Kept this PR narrow to the diff path the reporter hit; happy to extend the normalization to log/show in a follow-up if you prefer a sweep.
- The wshm AI-triage comment on the issue diagnosed this as "arg-forwarding bug ... arg splitting logic" at 95% confidence. The actual cause is more specific — the splitting works, but the rule for which leading args qualify as paths missed the bare-word-with-extension category.

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -91,8 +91,16 @@ fn normalize_diff_args(args: &[String]) -> Vec<String> {
 /// 1. Explicit path prefixes (`.`, `~`) → always a path, no filesystem check needed.
 /// 2. Contains path separator (`/`, `\`) → use `path_exists` to distinguish branch names
 ///    (e.g. `feature/auth`) from real paths (e.g. `src/main.rs`).
-/// 3. Bare word with no separator → never a path (avoids injecting `--` when a file
-///    happens to share a name with a branch or ref, e.g. a file named `main`).
+/// 3. Bare word with a mid-string `.` (looks like a filename with extension, e.g.
+///    `package.json`, `playwright.config.ts`) → use `path_exists` to distinguish
+///    refs/tags (e.g. `v1.2.3`) from real files. Required for issue #1669: when a
+///    user-passed `--` is dropped by clap and the file list mixes bare-word-with-
+///    extension args (`package.json`) with explicit-path args (`.harness/foo.yaml`),
+///    rule 1 alone places `--` after the bare-word-with-extension args, leaving them
+///    misinterpreted as revisions.
+/// 4. Bare word with no separator and no `.` → never a path (avoids injecting `--`
+///    when a file happens to share a name with a branch or ref, e.g. a file named
+///    `main`).
 fn normalize_diff_args_impl<F>(args: &[String], path_exists: F) -> Vec<String>
 where
     F: Fn(&str) -> bool,
@@ -114,7 +122,14 @@ where
         if arg.contains('/') || arg.contains('\\') {
             return path_exists(arg);
         }
-        // Bare word (no separator, no special prefix) — never inject `--`
+        // Bare word with mid-string `.` (looks like filename with extension) —
+        // use filesystem check to distinguish refs/tags (v1.2.3) from real
+        // files (package.json). The `.` must not be at position 0 because
+        // that prefix is already handled by rule 1 (.gitignore as dotfile).
+        if arg.find('.').is_some_and(|i| i > 0) {
+            return path_exists(arg);
+        }
+        // Plain bare word (no separator, no `.`, no special prefix) — never inject `--`
         // This avoids misidentifying a ref/branch as a path even if a same-named
         // file happens to exist on disk.
         false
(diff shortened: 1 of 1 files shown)


KuSh:
Hi, thanks for your PR, I'm willing to wait to see if clap allows us to keep escapes `--` to allow transmitting it transparently. There seems to be [traction](https://github.com/clap-rs/clap/issues/5055) outside our own needs.

truffle-dev:
Sounds good. Subscribed to clap-rs/clap#5055. Happy to keep this rebased as a fallback, or close whenever you prefer.

KuSh:
@truffle-dev After thinking more about this and seeing that clap's issue has not seen any updates since the spring of 2024, I pushed a more generic solution than stacking more heuristic on already existing ones, in https://github.com/rtk-ai/rtk/pull/1823. If you want to give it a try, it would help, thanks

truffle-dev:
Tried it. Built #1823, ran the args_utils tests (15/15 pass), and walked through the original cases:

- `rtk cargo clippy -p test_pkg -- -D warnings` — pass-through clean.
- `rtk cargo test -- --nocapture` — pass-through clean.
- `rtk git diff -- existing.rs` — `--` survives.
- `rtk git diff existing.rs` — no `--` injected, which is the deliberate scope cut.

Closing this in favor of #1823. The preserve-what-the-user-typed shape is cleaner than the path-existence heuristic; users who want a bare-word-with-extension treated as a path can still spell `--` themselves, same as upstream git.

truffle-dev:
Superseded by #1823.

PR #1532, opened 2026-04-25
chore: Fix clippy warnings

KuSh:
## Summary

- **Path Centralization:** Hardcoded directory and file paths across all hook logic () are replaced with dedicated, exported constants in . This prevents magic strings and simplifies maintenance when system directories change.
- **Code Cleanup:** Move all code only used by tests behind cfg(test) attribute
- **Refactoring:** Apply Clippy fixes and address remaining warnings

## Test plan
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [ ] Manual testing: `rtk <command>` output inspected

PR #1692, opened 2026-05-03
Next Release

### LINKED ISSUE #384: Windows: rtk init --uninstall does not work

### LINKED ISSUE #844: RTK exhibits unusual behavior when using ls and wildcards.

### LINKED ISSUE #1276: rtk ls returns "(empty)" for non-empty directories on non-English locales

### LINKED ISSUE #1362: [BUG] head -N with multiple files rewrites to invalid `rtk read file1 file2 ... --max-lines N`
## Summary

`rtk rewrite` turns `head -N file1 file2 ...` into `rtk read file1 file2 ... --max-lines N`, but `rtk read` only accepts a single `<FILE>` argument. When invoked with multiple files, `rtk read` falls through to the system `read` command (`/usr/bin/read` on macOS, which is a POSIX wrapper around the shell builtin), which interprets the first filename as a variable name and fails with "not a valid identifier".

Same class of bug as #989 (cat multi-file, fixed) and #1108 (tail -N).

## Repro

```shell
$ rtk --version
rtk 0.37.0   # also confirmed on 0.31.0

$ rtk rewrite "head -3 /tmp/a /tmp/b /tmp/c"
rtk read /tmp/a /tmp/b /tmp/c --max-lines 3

$ rtk read /tmp/a /tmp/b /tmp/c --max-lines 3
/usr/bin/read: line 4: read: `/tmp/a': not a valid identifier
$ echo $?
0   # silent failure — exit 0 despite error
```

Real-world Claude Code failure:

```
$ head -5 .claude/skills/generate/SKILL.md .claude/skills/enclosure-design/SKILL.md
Exit code 1
/usr/bin/read: line 4: read: `.claude/skills/generate/SKILL.md': not a valid identifier
```

## Environment

- OS: macOS (Darwin 25.2.0)
- rtk: 0.37.0 (also 0.31.0)
- Claude Code hook v2 (auto-patched via `rtk init -g --auto-patch`)

## Expected

Either:
1. `rtk rewrite` returns exit 1 (no rewrite) when `head -N` has >1 file argument, so the shell runs native `head` — consistent with the `cat` fix in #989.
2. `rtk read` natively supports multiple files.

## Workaround

Use `rtk proxy head -N file1 file2 ...` to bypass the rewrite, or use editor/IDE file-read tools that don't go through Bash.

## Related

- #989 Cat multiple files result in an error (closed, same class)
- #1108 Claude Code hook: \`tail -N\` rewrite produces invalid \`rtk read -N\` (same class)

### LINKED ISSUE #1452: bug(grep): `-c` count flag ignored — returns match summary + context instead of line count

### LINKED ISSUE #1535: rtk git stash {push,pop,apply,drop} returns exit 0 on failure and produces ambiguous success output

aeppling:
## Fix

- bug(grep): -c count flag ignored, returns match summary + context instead of line count https://github.com/rtk-ai/rtk/pull/1456 close https://github.com/rtk-ai/rtk/issues/1452

- fix(ls): LC_ALL=C + fallback to raw on unrecognized locale https://github.com/rtk-ai/rtk/pull/1338 close https://github.com/rtk-ai/rtk/issues/1276 https://github.com/rtk-ai/rtk/issues/1475 https://github.com/rtk-ai/rtk/issues/1652  https://github.com/rtk-ai/rtk/issues/1321 https://github.com/rtk-ai/rtk/issues/1566 https://github.com/rtk-ai/rtk/issues/803 https://github.com/rtk-ai/rtk/issues/1715 https://github.com/rtk-ai/rtk/issues/1723

- [fix(ls): handle all file types (device, pipe, socket) in ls filter](https://github.com/rtk-ai/rtk/pull/1451#top) https://github.com/rtk-ai/rtk/pull/1451 close #844 

- fix: head/tail multi-file rewrite falls back to native command https://github.com/rtk-ai/rtk/pull/1371 close https://github.com/rtk-ai/rtk/issues/1362

- fix(pnpm): install don't take a list of packages https://github.com/rtk-ai/rtk/pull/1009 

- fix(tests): dotnet cmd test flakiness https://github.com/rtk-ai/rtk/pull/1238 solve an known test env issue 

- refactor(git): Fix stash status detection for all case https://github.com/rtk-ai/rtk/pull/1586 close https://github.com/rtk-ai/rtk/issues/1535

- feat(cicd): add auto next release parser  https://github.com/rtk-ai/rtk/pull/1462

- [fix(init-uninstall): uninstall removes --claude-md artifacts on Windows](https://github.com/rtk-ai/rtk/pull/392#top) https://github.com/rtk-ai/rtk/pull/392 close https://github.com/rtk-ai/rtk/issues/384

## Others

- chore: Fix clippy warnings https://github.com/rtk-ai/rtk/pull/1532 code cleaning


aeppling:
LGTM !

pszymkowiak (approved):
LGTM — solid release.

PR #1696, opened 2026-05-03
fix(tee): keep head + tail when raw output exceeds max_file_size

iliaal:
## Before
`write_tee_file` truncated raw output at `max_file_size` (1MB default) by keeping the first 1MB of bytes and dropping the rest. The tee log is the recovery hatch agents fall back to when the live filter dropped a needed signal, but for tail-heavy commands the head is the part that doesn't matter.

Triggered by `rtk make test` on a php-src checkout: ~17000 phpt progress lines followed by a pass/fail summary. The make filter kept only the first ~16 progress lines (separate fix on `feat/php-tooling`). The tee log captured 1MB of progress lines and also dropped the summary. Same head-bias on a tail-heavy command, two layers down.

## After
`write_tee_file` keeps a 25/75 head/tail slice with a `--- N bytes truncated from middle ---` marker between them. Default 1MB budget gives ~256KB of head plus ~768KB of tail. The slicer rounds both cuts to UTF-8 char boundaries, so multi-byte input never panics.

Tail gets the larger share because tee mostly captures build and test output, where summaries and failure trailers live at the end. The head stays because the prologue often shows the command echo and config that explain what was running.

## Scope rationale
- Not adding a config knob for head/tail/balanced selection. 25/75 beats head-only for the cases that matter; can follow up if someone asks for selectable shapes.
- Not changing the public API: `max_file_size`, `RTK_TEE=0`, `force_tee_hint` behave the same.
- Not fixing the make filter itself. Lives on `feat/php-tooling` because that branch already touches PHP tooling. This PR fixes the recovery layer for every tail-heavy command, not just make.

Backward compatibility: marker text changed from `--- truncated at N bytes ---` to `--- N bytes truncated from middle ---`. Visible if anyone parses tee logs programmatically.

## Test plan
- [x] `cargo test --bin rtk`: 1721 passed, 0 failed
- [x] `cargo clippy --all-targets`: no new warnings
- [x] `cargo fmt --all`: no churn
- [x] New `test_write_tee_file_keeps_tail`: 3000-line phpt-style run, verifies head + marker + tail all survive
- [x] Updated `test_write_tee_file_truncation_utf8_boundary` and `test_write_tee_file_truncation_emoji`: new split exercised with multi-byte chars at both cuts

### DIFF
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -121,18 +121,37 @@ fn write_tee_file(
     let filename = format!("{}_{}.log", epoch, slug);
     let filepath = tee_dir.join(filename);
 
-    // Truncate at max_file_size (find a safe UTF-8 char boundary)
+    // Truncate to fit in max_file_size, keeping head + tail and dropping the middle.
+    // Tail-heavy commands (test runners, build summaries) put their useful data at the
+    // end; head-only truncation throws that away. 25/75 split keeps the prologue while
+    // giving the summary three quarters of the budget.
     let content = if raw.len() > max_file_size {
-        let boundary = raw
+        let head_budget = max_file_size / 4;
+        let tail_budget = max_file_size - head_budget;
+
+        // Last UTF-8 char boundary at or before head_budget.
+        let head_end = raw
             .char_indices()
-            .take_while(|(i, _)| *i < max_file_size)
+            .take_while(|(i, _)| *i < head_budget)
             .last()
             .map(|(i, c)| i + c.len_utf8())
             .unwrap_or(0);
+
+        // First UTF-8 char boundary at or after raw.len() - tail_budget.
+        let tail_target = raw.len().saturating_sub(tail_budget);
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into the recovery path. The mechanism this patch edits was replaced on develop by #3278 (merged 2026-09-09), which moved output recovery to a content addressed sqlite store selected by the [retriever] mode setting and left the tee file writer as the legacy path in src/core/tee_file.rs. That store already keeps the hidden tail instead of the prefix the agent has already seen, so the head bias you hit no longer applies on the default path, and write_tee_file no longer lives in src/core/tee.rs, so the change cannot land as written. Two gaps are still worth a small issue or a narrow PR against the current code: the legacy tee writer still keeps only the head when output exceeds max_file_size, and the store cuts at the first max_entry_bytes of the hidden region, so a very long hidden tail still loses its end. Please comment or reopen if a part of this was missed.

Issue #1272, opened 2026-04-13
Cursor hook: rtk rewrite success exit code causes rewrites to be dropped; optional PATH note for Linux


arthur-b-renaud:
# Cursor hook: `rtk rewrite` exit code drops rewrites; PATH on Linux

## Summary

After `rtk init -g --agent cursor`, the global Cursor hook looked correctly configured (`~/.cursor/hooks.json` + `~/.cursor/hooks/rtk-rewrite.sh`), but **shell commands from the agent were not rewritten** to `rtk …`. Two separate causes showed up:

1. **Hook logic vs `rtk rewrite` exit codes** — the hook treats a non-zero exit from `rtk rewrite` as failure and returns `{}`, which **discards successful rewrites**.
2. **(Linux) PATH** — the hook subprocess may not see `~/.local/bin`, so `command -v rtk` fails and the hook exits without rewriting (stderr warning).

## Environment

- **OS:** Linux
- **RTK:** e.g. `0.35.0` (`rtk --version`)
- **Editor:** Cursor (global hooks: `~/.cursor/hooks.json`)
- **`rtk` install location:** e.g. `~/.local/bin/rtk`

## Expected behavior

When the Cursor agent issues a Shell tool call that RTK can rewrite (e.g. `ls -la`), `preToolUse` should return JSON that updates the command to the `rtk`-wrapped form (e.g. `rtk ls -la`).

## Actual behavior

- Agent still runs the **original** command (no `rtk` prefix), or the hook appears to no-op.
- `rtk gain` may only reflect commands run **manually** with `rtk`, not agent rewrites — so it’s easy to think the hook isn’t active.

## Root cause 1: Exit code handling in `rtk-rewrite.sh`

The hook (as installed) does something equivalent to:

```bash
REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null) || { echo '{}'; exit 0; }
```

Observed behavior of `rtk rewrite`:

| Outcome    | stdout              | exit code (observed) |
|------------|---------------------|----------------------|
| Rewrote    | e.g. `rtk ls -la`   | **3**                |
| No rewrite | empty               | **1**                |

Because **both** outcomes use non-zero exit codes, the `|| { echo '{}'; exit 0; }` branch runs after a **successful** rewrite as well: stdout is captured into `REWRITTEN`, but the script then prints `{}` and exits, so the rewrite is never applied.

[...]

PR #1718, opened 2026-05-05
fix(hooks): make Cursor preToolUse rewrites work and stay visible

### LINKED ISSUE #1463: Bug Report: Cursor agents doesn't use RTK on Ubuntu 24.04.4 LTS
## Summary
The Cursor agents are not using RTK even after running `rtk init -g --agent cursor` and `rtk hook cursor`.

## Environment
OS: Ubuntu 24.04.4 LTS
RTK version: 0.37.1, 0.37.2  
Editor: Cursor  3.1.17

## Expected behavior
After running `rtk init -g --agent cursor` and registering the hook, Cursor agents should use RTK when executing shell commands (e.g., `git status`).

The hook is correctly registered in `~/.cursor/hooks.json`:

```json
{
  "version": 1,
  "hooks": {
    "preToolUse": [
      {
        "command": "rtk hook cursor",
        "matcher": "Shell"
      }
    ]
  }

### PR DESCRIPTION
**Pull request by kamilkaczmareksolutions**
## TL;DR

On Windows + Cursor, RTK's `preToolUse` hook silently no-ops on **every** Shell command — and the panel shows `Output: {}`, which looks like the hook fired correctly. It didn't. This PR fixes the actual rewrite (root cause: Cursor sends UTF-8 BOM in stdin and serde_json rejects it) and the two visibility gaps that were hiding the breakage.

## What users saw vs. what was actually happening

| | Before this PR | After this PR |
|---|---|---|
| `preToolUse` panel | `Output: {}` for every Shell call | full JSON: `continue`, `permission`, `updated_input` |
| Command Cursor actually ran | original (e.g. raw `git status`) | rewritten (`rtk git status`) |
| `rtk gain --history` | no entries from Cursor | rewrites with 30–94% token reduction |
| User signal | "looks like the hook fires but does nothing" | "I can see exactly what RTK rewrote, output got smaller" |

This was particularly confusing because `Output: {}` is **also** the legitimate response when RTK intentionally skips a command (e.g. `htop`, `kill`, already-`rtk`-prefixed). So users had no signal that the hook was actually broken — it looked indistinguishable from "RTK chose not to rewrite this one".

### Before — `rtk hook cursor` (raw config from README)

before — Output: {} for git status, hook command is `rtk hook cursor`

### Investigation — wrapped `rtk hook cursor` with a Python tracer

The tracer logged stdin byte-for-byte and forwarded to RTK unchanged. Same `Output: {}`, but the log proved **what** Cursor was actually sending.

before — same Output: {} via tracer wrapper

```
STDIN_RAW first bytes (hex): ef bb bf ef bb bf 7b 22 63 6f 6e 76 65 72 73 ...
                              ^^^^^^^^ ^^^^^^^^ {  "  c  o  n  v  e  r  s
                              BOM #1   BOM #2

STDIN_PARSE_ERROR: JSONDecodeError('Unexpected UTF-8 BOM (decode using utf-8-sig)')
RTK_STDOUT_RAW: {}
```

Cursor on Windows ships hook payloads with **two leading UTF-8 BOMs**. `serde_json::from_str` refuses to parse them, so `run_cursor` falls into the "no command" branch and returns `{}` — silently, on every single hook call.

### After — same binary path, same Cursor, just patched RTK

after — full JSON output: continue: true, permission: allow, updated_input.command shows `rtk` injected into a compound command

The panel above is for a real compound command Cursor sent (`export PATH=... && cd ... && rtk cargo build --release 2>&1 | tail -8`). RTK correctly identifies the rewritable segment (`cargo build`), preserves the rest, returns `continue: true` so the panel renders, and Cursor honors `permission: "allow"` so the rewrite actually executes.

## What this PR changes

Three issues, fixed together because each one alone produces the same user-visible symptom:

1. **`permission: "ask"` is silently ignored by Cursor.** Cursor preToolUse only enforces `allow`/`deny`. Returning `ask` with `updated_input` could drop the rewrite, so the original command would run unmodified. Always return `allow` for rewritten commands. Deny rules still take precedence and are evaluated first.

2. **Panel collapses to `Output: {}` without `continue: true`.** Every other Cursor hook (`afterShellExecution`, `beforeSubmitPrompt`, `stop`, ...) returns `{ "continue": true, ... }` and renders correctly. The flat `{ permission, updated_input }` payload was technically valid but the UI didn't surface it. Mirror the working shape so the rewrite is visible to the user.

3. **Cursor stdin on Windows is BOM-prefixed.** Strip up to N leading UTF-8 BOMs in `run_cursor` before parsing. **This is the actual root cause of the long-standing "RTK silently does nothing in Cursor" reports** — the rewrite path was never reached. Fixes \#1 and \#2 are correctness/UX gaps exposed once parsing succeeds.

## Test plan

- [x] `cargo test test_cursor_ -- --nocapture` — 13 passed (2 new: `test_cursor_compound_rewrite_includes_continue`, `test_cursor_strips_leading_utf8_bom`)
- [x] `cargo build --release` clean
- [x] Local Windows smoke against the patched binary:
  - `git status` (no BOM) → `{"continue":true,"permission":"allow","updated_input":{"command":"rtk git status"}}`
  - `git status` with `EF BB BF EF BB BF` prepended (matches what Cursor actually sends) → same output
  - Compound `cd "/tmp/proj" && git status` → rewrites only the right segment
  - Skipped commands (`htop`, `kill 12345`, already-`rtk`-prefixed) → still `{}` (intended no-op)
- [x] **End-to-end in Cursor 3.2.16 on Windows**: panel renders the rewrite payload (see screenshot above), and `rtk gain --history` confirms subsequent shell calls hit the rewrite path with 30–94% token reduction (e.g. `rtk git push -93%`, `rtk git commit -94%`, `rtk git status -73%`). Before this PR: zero entries from Cursor.

## Related issues

- **Refs #1272** (`bug, P1-critical, effort-medium`) — "Cursor hook: rtk rewrite success exit code causes rewrites to be dropped". The original report is Linux-focused (exit-code handling in the legacy `rtk-rewrite.sh` script + `~/.local/bin` PATH), but the comment thread already includes a Windows user (@lxy94, 2026-04-21: "windows系统也没有触发rewrite 命令" — "Windows also does not trigger the rewrite command"). This PR addresses the **Windows variant** of the same family: the native `rtk hook cursor` path silently no-ops because Cursor sends BOM-prefixed stdin. The Linux exit-code/PATH parts of #1272 remain, so this is `Refs` rather than `Closes`.
- **Refs #1463** (`bug, awaiting-information`) — "Cursor agents doesn't use RTK on Ubuntu 24.04.4 LTS". Same external symptom as the Windows case (`rtk init -g --agent cursor` runs cleanly, hook is registered, but agent commands never get the `rtk` prefix). Worth retesting on Ubuntu with this branch — if Cursor on Linux also prepends a UTF-8 BOM to hook stdin, this PR likely closes #1463 as well; if not, at least narrows the search.

## Notes for reviewers

- **Scope.** BOM-strip is intentionally Cursor-only (in `run_cursor`). I did not observe the same BOM behaviour on Claude/Copilot/Gemini stdin during this debugging, so I left those paths alone. Happy to widen to a shared helper across all hook handlers if maintainers prefer.
- **Cursor-side bug.** Sending double UTF-8 BOM on stdin is its own bug worth reporting upstream to Cursor — but RTK should be robust to it regardless, and shipping the fix here unblocks every existing Windows + Cursor user immediately rather than waiting on a Cursor release.
- **Why bundle the three changes.** They share the same user-visible symptom (`Output: {}` + commands not rewritten). Fixing only \#3 would make the rewrite work but the panel would still mislead users. Fixing only \#1+\#2 (which is what an earlier version of this PR did) doesn't help at all on Windows because parsing still fails before the new logic runs. The three together are the minimal change that actually closes the loop.
- **Screenshots** are hosted on a parking-lot branch in my fork (`pr-1718-assets`) so this PR's diff stays purely the code change.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -399,11 +399,23 @@ fn run_claude_inner(input: &str) -> Option<String> {
 
 // ── Cursor native hook ─────────────────────────────────────────
 
+/// Cursor on Windows ships hook payloads with one or more leading
+/// UTF-8 BOMs (`EF BB BF`, sometimes doubled), which serde_json
+/// refuses to parse. Strip them defensively so the rewrite path keeps
+/// working instead of silently returning `{}`.
+fn strip_leading_bom(input: &str) -> &str {
+    let mut s = input;
+    while let Some(rest) = s.strip_prefix('\u{feff}') {
+        s = rest;
+    }
+    s
+}
+
 /// Run the Cursor Agent hook natively.
 pub fn run_cursor() -> Result<()> {
     let input = read_stdin_limited()?;
 
-    let input = input.trim();
+    let input = strip_leading_bom(&input).trim();
     if input.is_empty() {
         let _ = writeln!(io::stdout(), "{{}}");
         return Ok(());
@@ -444,14 +456,20 @@ pub fn run_cursor() -> Result<()> {
         }
     };
 
-    let decision = match verdict {
-        PermissionVerdict::Allow => "allow",
-        _ => "ask",
-    };
+    // Cursor preToolUse currently enforces allow/deny only and can ignore
+    // updated_input when permission is "ask". Use "allow" for rewritten
+    // commands unless the command is explicitly denied above.
+    let decision = "allow";
 
     audit_log("rewrite", &cmd, &rewritten);
 
+    // `continue: true` mirrors the shape of every other Cursor hook
+    // (afterShellExecution, beforeSubmitPrompt, stop, ...). Cursor's
+    // preToolUse panel renders the JSON it received; without this field
+    // the panel collapses to `Output: {}` even though the rewrite ran,
+    // which makes the hook look broken to users.
     let output = json!({
+        "continue": true,
         "permission": decision,
         "updated_input": { "command": rewritten }
     });
@@ -471,6 +489,7 @@ fn run_cursor_inner_with_rules(
     ask_rules: &[String],
     allow_rules: &[String],
 ) -> String {
+    let input = strip_leading_bom(input);
     let v: Value = match serde_json::from_str(input) {
         Ok(v) => v,
         Err(_) => return "{}".to_string(),
@@ -492,11 +511,9 @@ fn run_cursor_inner_with_rules(
 
     match get_rewritten(&cmd) {
         Some(rewritten) => {
-            let decision = match verdict {
-                PermissionVerdict::Allow => "allow",
-                _ => "ask",
-            };
+            let decision = "allow";
             let output = json!({
+                "continue": true,
                 "permission": decision,
                 "updated_input": { "command": rewritten }
             });
@@ -777,10 +794,13 @@ mod tests {
     fn test_cursor_rewrite_flat_format() {
         let result = run_cursor_inner(&cursor_input("git status"));
         let v: Value = serde_json::from_str(&result).unwrap();
-        // Default permission (no explicit allow rule) → "ask"
-        assert_eq!(v["permission"], "ask");
+        // Cursor preToolUse expects allow/deny for rewrite application.
+        assert_eq!(v["permission"], "allow");
         assert_eq!(v["updated_input"]["command"], "rtk git status");
         assert!(v.get("hookSpecificOutput").is_none());
+        // `continue: true` keeps the Cursor preToolUse panel from collapsing
+        // to `Output: {}`; without it the rewrite is invisible to users.
+        assert_eq!(v["continue"], true);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


mhstudioos:
🙏

kamilkaczmareksolutions:
**Update — rebased onto current `develop` + improved test split**

Following @cristianrojas07's parallel rebase in #1807 (thanks!), this PR is now also rebased onto current `develop` HEAD. No merge conflicts — the `transparent_prefixes` work that landed in `hooks::hook_cmd` since this PR was opened touches `get_rewritten()` and `run_gemini()` but not the Cursor path, so the rebase was clean.

I also took the opportunity to adopt the test structure from #1807, which is genuinely better than my original:

- Split the single `test_cursor_strips_leading_utf8_bom` into separate `test_cursor_strips_single_utf8_bom` and `test_cursor_strips_double_utf8_bom` — clearer failure isolation.
- Added `test_strip_leading_bom_helper` as a direct unit test on the helper (covers 0 / 1 / 2 / 3 leading BOMs + mid-string BOM preservation), so future refactors can't regress the loop semantics without an obvious failure signal.

**Independent verification:** @cristianrojas07 confirmed the same fix works end-to-end on a notably newer stack (Cursor 3.3.27 + `rtk 0.39.0` vs my original Cursor 3.2.16 + 0.34.3). The BOM behavior has been stable across at least 3 Cursor releases and 5 RTK releases — this isn't a transient quirk.

**Current state of this PR:**
- Base: `develop` (fresh)
- Diff: still 1 file (`src/hooks/hook_cmd.rs`), now +91/-12
- Tests: 3 new (single BOM / double BOM / helper unit) + assertions added to 2 existing — all 43 `hooks::hook_cmd::tests` pass
- Mergeable: yes
- `cargo build --release`: clean

#1807 by @cristianrojas07 carries the same three fixes with equivalent intent. Maintainers can land either — they're functionally identical now. The other will be closed in deference to whichever lands first.

pszymkowiak (approved):
LGTM. Verified deny path short-circuits at hook_cmd.rs:432-437 before the `decision = "allow"` change, so safety model is preserved. Behavioral diff is limited to default "ask" mode → rewrite, matching how Claude/Copilot already behave. BOM-strip helper is loop-based (handles double BOM) and well unit-tested. Solid root-cause work with hex byte evidence and end-to-end Cursor screenshots.

Issue #1613, opened 2026-04-29
Bug: `rtk grep` misparses single-file rg output when content contains colons


frostime:
## Environment

- **RTK**: 0.37.2
- **ripgrep**: 15.1.0
- **OS**: Windows; rtk and rg are installed via scoop

## Description

`rtk grep` searching a single file produces **incorrect filenames and line numbers** when matched lines contain colons. The parser mistakenly treats the line number as a filename and the content gets truncated.

## Steps to Reproduce

```bash
# Create test file
cat > test.ts << 'EOF'
function readEnabledFromBranch(ctx: ExtensionContext, fallback: boolean): boolean {
    return true;
}
function simpleNoColon() {
    return 42;
}
EOF

# Run
rtk grep "function" test.ts
```

## Actual Output

```
2 matches in 2F:

[file] 1 (1):
  0: ExtensionContext, fallback: boolean): boolean {

[file] test.ts (1):
  4: function simpleNoColon() {
```

- Line with colons: filename=`2` (wrong — this is the line number), line_number=`0` (parse failure)
- Line without colons: correct

## Expected Output

```
2 matches in 1F:

[file] test.ts (2):
     1: function readEnabledFromBranch(ctx: ExtensionContext, fallback: boolean): boolean {
     4: function simpleNoColon() {
```

### rg Output Format Difference

In [`grep_cmd.rs`](https://github.com/rtk-ai/rtk/blob/master/src/cmds/system/grep_cmd.rs), rg is invoked:

```rust
rg_cmd.args(["-n", "--no-heading", &rg_pattern, path]);
```

ripgrep **omits the filename prefix** when searching a single file. RTK's parser assumes filename is always present.

```
# Multi-file: filename is ALWAYS included
src/a.ts:2:function foo() { }
src/b.ts:5:function bar() { }
   │         │         │
   ▼         ▼         ▼
 file     linenum    content

# Single-file: filename is OMITTED
2:function readEnabledFromBranch(ctx: ExtensionContext) { }
│         │                          │
▼         ▼                          ▼
linenum   content (contains colons!)
```

When RTK parses the single-file output with `splitn(3, ':')`:

```
Input:  2:function readEnabledFromBranch(ctx: ExtensionContext) { }
                   ↓ splitn(3, ':')
Part:   ["2"]  ["function readEnabledFromBranch(ctx"]  [" ExtensionContext) { }"]
            │              │                                     │
            ▼              ▼                                     ▼
[...]

PR #1725, opened 2026-05-05
fix(grep): force --with-filename to fix single-file rg output parsing

ShiraiEd:
Summary

  - rg omits the filename prefix when searching a single file, causing the parser to misread the line number as the filename and truncate content at the
  first colon
  - Added `--with-filename` to the rg invocation so output always includes the filename prefix regardless of how many files are searched
  - Extracted inline parsing logic into `parse_rg_line` helper for testability

  Test plan

  - cargo fmt --all && cargo clippy --all-targets && cargo test
  - Added `test_parse_rg_line_with_colons_in_content` — verifies filename and line number are correct when content contains colons
  - Added `test_parse_rg_line_single_file_fallback` — verifies two-part lines fall back to the provided path
  - Manual test: rtk grep "function" test.ts on a file with colon-heavy signatures now correctly returns 2 matches in 1 file with accurate filenames and line numbers 
  

(fixes [#1613])


ShiraiEd:
Already fixed

Issue #1763, opened 2026-05-07
how to update rtk self


wink-cloud-hub:
I want like this `rtk update`
but not support for now

fabis94:
+1, there's virtually no docs at all related to actually updating this

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟢 **Priority** | `low` |
| 🎯 **Confidence** | 91% |

### Summary

Feature request for a self-update command (`rtk update`); a PR #1768 is already in progress addressing this.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

KuSh:
Closing as a duplicate of #995, which tracks the same problem with the more complete discussion. Please follow up there; if your case differs in a way that issue does not cover, comment here and it can be reopened.

Issue #995, opened 2026-04-03
Add updater/auto-updater


aeppling:
RTK should have an update command to get latest version without having to use brew or re-install

We could use a auto-updater but we need user consent to update, so we need to figure a way to cleanly ask user for update or inform update is available, at least a update command.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟢 **Priority** | `low` |
| 🎯 **Confidence** | 97% |

### Summary

Feature request for a self-update command; already has an open PR (#1768) so priority is low pending merge.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

FlorianBruniaux:
Complements #725 which tracks the passive version-check notification layer. Issue #995 focuses on the explicit `rtk update` command (and optional auto-updater with consent), while #725 handles the one-line "update available" banner on startup. Together they form a Claude-Code-style update experience.

PR #614, opened 2026-03-16
feat: add rtk update command

ricatix:
## Summary
- add a new `rtk update` subcommand and wire it into the CLI meta-command handling
- update Homebrew installs automatically and pin default Quick Install updates to a tagged release via `install.sh`
- document the supported update paths and add a changelog entry

## Testing
- `sh -n install.sh`
- `. \"$HOME/.cargo/env\" && rtk cargo fmt --all --check`
- `. \"$HOME/.cargo/env\" && rtk cargo clippy --all-targets`
- `. \"$HOME/.cargo/env\" && rtk cargo test`
- `. \"$HOME/.cargo/env\" && cargo run -- update`
- Oracle read-only review of the final implementation

PR #1766, opened 2026-05-07
feat(update): add verified self-update command

othavioquiliao:
## Summary

Adds `rtk update`.

For direct binary installs, RTK now updates itself from GitHub Releases:

- `rtk update` installs the latest release
- `rtk update --check` checks without changing anything
- `rtk update --tag v0.39.0` installs a specific tag
- `rtk update --force` reinstalls even when versions match
- `rtk update --yes` skips the confirmation prompt

The updater verifies `checksums.txt`, rejects unsafe archive entries, and replaces the current binary through a temp file in the same directory.

Managed installs stay managed. If RTK detects Homebrew, Cargo, or a source-tree binary, it prints the right upgrade command instead of overwriting files it should not own.

## Why

Users keep asking for this. See #1763 and #995.

Today the answer is basically "reinstall it yourself". That works, but it is not much of an answer after you have shipped a CLI people use every day.

This keeps the boring path boring:

```bash
rtk update
```

No shell installer pipeline. No guessing which file to replace. No writing over Homebrew or Cargo installs.

## Related Work

Related to #614.

That PR adds an update command too. Main difference: this implementation keeps direct binary updates inside Rust and verifies the release artifact before replacing the executable.

Homebrew and Cargo remain package-manager paths here. I think that boundary matters. I have debugged enough broken self-updaters to prefer boring ownership rules.

## Docs

Updated:

- `README.md`
- `INSTALL.md`

`CHANGELOG.md` is not edited manually because release-please owns it.

## Testing

```bash
cargo fmt --all --check
cargo clippy --all-targets
cargo test
```

`cargo clippy --all-targets` exits successfully. It still reports existing warnings outside this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -375,6 +375,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
 
 [[package]]
+name = "filetime"
+version = "0.2.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f98844151eee8917efc50bd9e8318cb963ae8b297431495d3f758616ea5c57db"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "libredox",
+]
+
+[[package]]
 name = "find-msvc-tools"
 version = "0.1.9"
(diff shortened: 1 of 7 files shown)


othavi0:
Closing this one and reopening from a branch that follows the contributing branch naming convention (). Same change, cleaner PR metadata.

PR #1794, opened 2026-05-08
feat(discover): support uv run python tool wrappers

xronocode:
## Summary

Adds discover and rewrite coverage for common Python tooling invoked through `uv run`.

- treats `uv run` as a built-in transparent wrapper in the hook rewrite engine
- classifies `uv run pytest`, `uv run python -m pytest`, `uv run ruff check/format`, and `uv run python -m mypy`
- rewrites supported invocations to preserve the wrapper, e.g. `uv run pytest tests/` -> `uv run rtk pytest tests/`
- keeps unsupported arbitrary commands like `uv run python -c ...` unmodified

## Why

`rtk discover --all --since 7` showed `uv run` as a top unhandled developer-tool command. This closes the common Python-tooling cases without adding a broad passthrough wrapper for arbitrary `uv run` commands.

## Testing

- `cargo test discover::registry::tests`: 272 passed
- `cargo test`: 1818 passed, 0 failed, 6 ignored
- `cargo fmt -- --check`: passed
- manual local hook checks:
  - `uv run pytest tests/` -> `uv run rtk pytest tests/`
  - `uv run python -m pytest -x tests/` -> `uv run rtk pytest -x tests/`
  - `uv run ruff check .` -> `uv run rtk ruff check .`
  - `uv run python -c 'print(1)'` -> no rewrite

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -445,15 +445,15 @@ fn strip_trailing_redirects(cmd: &str) -> (&str, &str) {
 /// Handles compound commands (`&&`, `||`, `;`) by rewriting each segment independently.
 /// For pipes (`|`), only rewrites the left-hand command (pipe targets stay raw),
 /// but continues rewriting segments after subsequent `&&`/`||`/`;` operators.
-/// Also strips user-configured transparent wrapper prefixes
+/// Also strips built-in and user-configured transparent wrapper prefixes
 /// (`[hooks].transparent_prefixes` in `config.toml`) before routing.
 ///
 /// A transparent prefix is a wrapper command that doesn't change *what* is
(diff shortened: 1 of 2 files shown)


cjthompson:
I was just looking for this because all of my projects are using `uv` to manage the `venv` and correct python versions.

xronocode:
Thanks @cjthompson — uv-managed venvs with pinned Python versions are exactly the case this targets. The PR's been mergeable since early May with no conflicts. @aeppling, would a review be possible when you get a chance? Happy to rebase onto `develop` if anything has drifted.

xronocode:
Closing — `uv run` wrapper support landed in develop via a different design (`ROUTABLE_WRAPPER_PREFIXES: &["uv run"]` in `registry.rs`, with dedicated tests), so this is redundant. Thanks!

PR #1823, opened 2026-05-10
chore(args): introduce a more generic solution for restoring double dashes in args

### LINKED ISSUE #1669: rtk does not handle git diff correctly in some cases

KuSh:
## Summary
- Introduce a shared, generic solution to restore double dashed into args
- Replace custom solutions in cargo and git commands with that solution

Fixes #1669

## Test plan
- [x] Manual testing: `rtk git diff -- "src/core/args_utils.rs"` output inspected

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1,5 +1,6 @@
 //! Filters git output — log, status, diff, and more — keeping just the essential info.
 
+use crate::core::args_utils;
 use crate::core::stream::{
     self, exec_capture, CaptureResult, FilterMode, LineHandler, LineStreamFilter, StdinMode,
 };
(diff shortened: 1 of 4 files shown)


truffle-dev:
Tested at `4153ebb` against the original #1669 repros. All cases I'd tracked work:

- `rtk git diff -- <file>` (single file after `--`) — passes
- `rtk git diff -- file1 file2 file3` (multiple files after `--`) — passes
- `rtk git diff HEAD -- <file>` (ref before `--`, file after) — passes
- `rtk git diff <branch> -- <file-named-same-as-branch>` (file shares name with a ref) — passes
- `rtk git diff --cached -- <file>` — passes
- `rtk git diff` (no args) — passes
- `rtk git diff <bare-ref-with-no-dash>` — passes

`cargo test --release args_utils` is 15/15 green; the unit tests cover the swallowed-by-clap, preserved-by-clap, no-`--`-in-original, and interleaved-`--`-and-args shapes.

One behavioral note worth flagging in case it's intentional: with the old `normalize_diff_args` heuristic, `rtk git diff main` (where `main` is both a branch and a file on disk) would have had `--` injected based on the filesystem check and rendered the file diff. With this PR, the same command falls through to git which then exits with `fatal: ambiguous argument 'main'`. That looks like the right call to me — restoring what clap consumed is predictable, second-guessing the user with filesystem heuristics is not — but flagging it explicitly so the design choice is on the record.

Happy to close #1679 once this lands. Want me to do it now or wait until merge?

KuSh:
Thanks a lot for the detailled test and report, very much appreciated!

> One behavioral note worth flagging in case it's intentional: with the old normalize_diff_args heuristic, rtk git diff main (where main is both a branch and a file on disk) would have had -- injected based on the filesystem check and rendered the file diff. With this PR, the same command falls through to git which then exits with fatal: ambiguous argument 'main'.

The problem would have happen without rtk and should be handled by the LLM IMHO

> Happy to close #1679 once this lands. Want me to do it now or wait until merge?

As we have a release branch, merging doesn't mean available in a release, we will close it once released. Thanks again!

pszymkowiak (approved):
LGTM — solid refactor.

Reviewed in depth:
- The unified `args_utils::restore_double_dash` removes two divergent
  implementations (`normalize_diff_args` in git.rs, `restore_double_dash`
  in cargo_cmd.rs) and replaces them with a single orthogonal one.
- The behavior change (no more filesystem-existence heuristic to inject
  `--` in `rtk git diff`) is a deliberate design call — predictable
  passthrough over filesystem guessing, properly defended in the thread.
- 15 new tests in `args_utils` cover single/multiple swallowed dashes,
  preserved dashes, no-dash-in-original, compound commands, duplicate
  args, and git-diff specific cases.
- @truffle-dev's manual repro suite for #1669 passed end-to-end.
- All 10 CI checks green (Windows / macOS / Linux / clippy / semgrep /
  CodeQL).

A couple of minor follow-ups (non-blocking) if you ever revisit:
- Add a test for duplicate args traversing `--`
  (e.g. `raw = [..., "x", "--", "x"]`, `parsed = ["x", "x"]`) to lock
  in the behavior — `contains()` matches by value, edge worth pinning.
- A short note in the docstring that args present in `raw` but absent
  from `parsed_args` are silently skipped (e.g. the subcommand name).

Approving — none of these block merge.

aeppling (changes requested):
**Requesting changes.** The consolidation is the right direction, and dropping the filesystem heuristic genuinely fixes #1431 plus the bare-word miss. But the new reconstruction regresses a common `git diff` case. Verified end to end against a release build of `a5b4653`, not just the function in isolation.

## Blocker: positionals equal to a command/subcommand token get duplicated

`restore_double_dash_with_raw` rebuilds the list by scanning the whole raw command line and re-emitting any token where `parsed_args.contains(&token)`. That is a value test, and only `argv[0]` is skipped, so the command words (`git`, `diff`, `cargo`, `test`, ...) stay live candidates. When a user positional has the same value as one of them, the command word leaks back in.

All subcommands use `trailing_var_arg = true` (`src/main.rs:83` plus ~75 others), so clap swallows the leading `--` and user args arrive as the contiguous tail. That is exactly the shape this gets wrong.

Reproduced with this PR's binary in a real repo:

```
$ git diff -- diff       # native: shows the diff, RC 0
$ rtk git diff -- diff   # PR #1823 reconstructs `git diff diff -- diff`
fatal: bad revision 'diff'   (exit 128)
```

This is a regression. Installed `rtk 0.40.0` handles the same input correctly. It breaks Correctness over Token Savings for ordinary filenames (`diff`, `git`, `log`, `test`, `build`, `status`, ...) and violates "never block the user", since RTK makes a working command fail.

Root cause: value membership mapping with no positional anchor. `trailing_var_arg` guarantees the user args are the tail of `raw_args`, so restoration must align positionally from the right and stop at the first non-matching token (the command prefix). It must never scan back into the command words. The output should be byte for byte what the user typed, with only the `--` reinserted.

## test_compound_command_with_dashes asserts wrong behavior

```rust
let raw = vec!["rtk","cmd1","--","arg1","&&","cmd2","--","file"]; // asserts ["--","--","file"]
```

The shell splits on `&&`, so each `rtk` process gets its own `argv` and this input cannot reach the function. The test locks in the buggy behavior. Remove or replace it.

## Missing the test that would catch the blocker

No case uses a positional whose value equals a command token. Add `-- diff`, `-- git`, `-- test`, and `cargo test cargo -- --nocapture`.

## Doc comment overclaims

The header lists "duplicate args on both sides of `--`" as handled, but that only holds for the one tested shape. Make it match the chosen algorithm.

## Asks before merge

1. Replace value membership with positional tail alignment.
2. Add the collision tests above.
3. Fix or remove test_compound_command_with_dashes.
4. Re-confirm #1215, #1431, #1669 after the rewrite.

The fix is scoped to this one function. The rest of the refactor is good.

aeppling:
:c

PR #1835, opened 2026-05-11
feat: verify binary checksum after download

femrebora:
## What

Adds SHA256 checksum verification to the install script.

`checksums.txt` is already published alongside every release (generated by `sha256sum *` in the CI release workflow), but the install script never used it. This wires it up.

## How

A new `verify_checksum()` function is called immediately after the binary archive is downloaded and before it is extracted:

1. Downloads `checksums.txt` from the same release tag as the binary.
2. Looks up the expected SHA256 for the archive using `awk` (handles both plain filenames and `./`-prefixed entries).
3. Computes the actual SHA256 with `sha256sum` (Linux) or `shasum -a 256` (macOS).
4. Fails with a clear error on mismatch — **fails closed** if neither tool is found.

## Why

Without this, a MITM or compromised CDN response would be silently trusted. The checksums already exist; this just enforces them.

## Changes

- `install.sh` — adds `verify_checksum()` (~30 lines) and one call site inside `install()`


aeppling:
Hey @femrebora 

LGTM, just need rebase

PR #1845, opened 2026-05-11
fix(container): surface docker health status in rtk docker ps output

Donotavio:
## Problem

\`rtk docker ps\` fetches \`{{.Status}}\` which already includes health state when configured:
\`\`\`
Up 2 hours (healthy)
Up 1 hour (unhealthy)
\`\`\`

But \`docker_ps()\` parsed \`parts[2]\` (Status field) and **discarded it completely**. An AI agent monitoring containers had no way to identify unhealthy services.

## Fix

- Include Status field in output between name and image
- Prefix unhealthy containers with \`[!]\` for immediate visibility
- All other states (healthy, plain Up, Exited, Restarting) pass through verbatim

### Refactoring

- Extract \`format_docker_ps_line(parts) -> Option<String>\` as \`pub(crate)\` pure function — directly unit-testable without spawning docker
- Extract \`trim_image_registry()\` helper — eliminates duplication with \`format_compose_ps()\` which had the identical \`.split('/').next_back()\` pattern
- Add \`UNHEALTHY_PREFIX\` constant

## Tests

7 new unit tests:
- \`test_docker_ps_preserves_healthy_status\` — (healthy) shown verbatim, no [!]
- \`test_docker_ps_marks_unhealthy_prominently\` — [!] prefix present
- \`test_docker_ps_shows_status_without_health_check\` — plain Up shown
- \`test_docker_ps_exited_no_flag\` — Exited shown, no [!]
- \`test_docker_ps_restarting_no_flag\` — Restarting shown, no [!]
- \`test_docker_ps_line_too_few_parts_returns_none\` — None guard for malformed input
- \`test_docker_ps_formats_all_five_containers\` — all 5 states in one pass

---

> Identified via adversarial testing: https://www.reddit.com/r/ClaudeCode/comments/1spiy8t/

### DIFF
--- a/src/cmds/cloud/container.rs
+++ b/src/cmds/cloud/container.rs
@@ -86,24 +86,8 @@ fn docker_ps(_verbose: u8) -> Result<i32> {
 
     for line in stdout.lines().take(15) {
         let parts: Vec<&str> = line.split('\t').collect();
-        if parts.len() >= 4 {
-            let id = &parts[0][..12.min(parts[0].len())];
-            let name = parts[1];
-            let short_image = parts
-                .get(3)
-                .unwrap_or(&"")
-                .split('/')
-                .next_back()
-                .unwrap_or("");
-            let ports = compact_ports(parts.get(4).unwrap_or(&""));
-            if ports == "-" {
-                rtk.push_str(&format!("  {} {} ({})\n", id, name, short_image));
-            } else {
-                rtk.push_str(&format!(
-                    "  {} {} ({}) [{}]\n",
-                    id, name, short_image, ports
-                ));
-            }
+        if let Some(formatted) = format_docker_ps_line(&parts) {
+            rtk.push_str(&formatted);
         }
     }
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for this. The docker ps line formatting was pulled into its own helper on develop as format_container_line_from_parts, shared with `docker ps -a`, and it now prints the container status, first in #1895 on 2026-05-19 and then in commits 861a46de, 3d4189cc and b52db52d in June, which also added the never-worse guard and the tee hint around it. Because develop prints the fields in a different order (id, name, image, status), the tests here assert a shape that no longer exists. Two things from this PR are still missing on develop: there is no visual marker such as `[!]` when a health check is failing, so `(unhealthy)` reads like any other status text, and the registry trim is still duplicated at three call sites instead of a shared helper. Both would be welcome as a small issue or a narrow PR on the current helper. If I missed a part of this, please comment or reopen.

Issue #1153, opened 2026-04-10
enhancement: add RTK_DISABLED=1 environment variable to bypass all hooks


iRonin:
## Problem

AI coding agents (Claude Code, Codex, Gemini CLI, etc.) have no way to temporarily disable the RTK hook without uninstalling it. When an agent needs raw, unfiltered command output -- for debugging, API inspection, or commands RTK does not handle well -- it has no escape hatch.

## Use cases

1. **Debugging**: Agent needs to see full unfiltered output to diagnose a problem
2. **API inspection**: Agent is reading API responses where RTK schema conversion destroys values
3. **Temporary bypass**: Agent wants raw \`git diff\` for a specific file, not the condensed version
4. **Compatibility**: Some tools or scripts produce output that RTK filtering corrupts

## Proposed solution

Add an \`RTK_DISABLED=1\` environment variable check in the hook rewrite path. When set:
- \`rtk rewrite <cmd>\` returns exit code 1 (no RTK equivalent) without doing any rewriting
- The hook passes the original command through unchanged
- No tracking or telemetry is affected (those remain active)

### Implementation sketch

In \`src/hooks/rewrite_cmd.rs\`, add at the start of \`run()\`:

\`\`\`rust
if std::env::var("RTK_DISABLED").unwrap_or_default() == "1" {
    std::process::exit(1);  // No rewrite, passthrough
}
\`\`\`

This mirrors the existing pattern used for \`RTK_TELEMETRY_DISABLED\` in \`src/core/telemetry.rs\` (line 26).

## Why not just use \`exclude_commands\`?

\`exclude_commands\` requires editing a config file and only works for specific commands. \`RTK_DISABLED\` gives the agent a **temporary, session-scoped kill switch** that can be toggled per-conversation without file system changes.

## Related

- This complements a potential \`--raw\` flag for individual commands (separate issue)
- Similar to how \`NODE_ENV\`, \`DEBUG\`, and other env vars provide runtime control

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 95% |

### Summary

[...]

Issue #831, opened 2026-03-25
Output filtering causes AI agents (Claude Code) to burn extra tokens on retry loops


BadassBison:
https://github.com/rtk-ai/rtk/pull/791 (fix)
https://github.com/rtk-ai/rtk/pull/871 (doc updates)

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 76% |

### Summary

Output filtering is too aggressive, causing AI agents (e.g., Claude Code) to miss error signals and enter retry loops that waste tokens; fix PRs already merged/in-progress, and a new opt-out PR #1848 is open.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

bb1:
Related: #1848 adds a skill file that teaches agents about the `RTK_DISABLED=1` env prefix, so they can self-bypass rtk when compressed output causes retry loops — rather than repeating the same command with minor variations until a human intervenes.

Keesan12:
The failure mode here sounds less like "filtering is expensive" and more like "the system keeps paying to rediscover the same state."

If output filtering can cause another full retry, I would want one cheap fingerprint before the next attempt starts: same tool path, same target, same normalized failure, same output class. If that fingerprint matches the recent history, the default should be stop or downgrade, not try again at full price.

A lot of agent cost bugs are really admission bugs. The expensive part is not the first bad turn. It's that the next nearly-identical turn was allowed in at all.

Issue #732, opened 2026-03-19
`rtk git branch --show-current` loop in opencode


technoch1ef:
### Description

I've experienced these issues using Opencode on 3 different computers. When it tries to check the current branch, it just keeps going in an infinite loop.

* **Opencode v1.2.27**
* Model: **Claude Opus 4.6**

<img width="714" height="1005" alt="Image" src="https://github.com/user-attachments/assets/cb2cac91-e60f-4069-bc94-43966f9e4cd5" />

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 85% |

### Summary

RTK rewrites `git branch --show-current` in a way that triggers an infinite loop in OpenCode; a related PR (#1848) adding an opt-out is already open.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

ipatalas:
I have just installed rtk on my opencode 1.4.7 (on Windows 11) and I'm getting similar results for git status command:

<img width="1294" height="1572" alt="Image" src="https://github.com/user-attachments/assets/8017392b-00be-4562-98a4-cbabdc8a9a29" />

It figured out it's being intercepted but it's stuck in the loop anyway. Not sure that's the problem with rtk itself. Maybe it's the model taught to interpret plain command outputs. I gave it another try like this:

<img width="1320" height="806" alt="Image" src="https://github.com/user-attachments/assets/1e777f87-7eee-4afe-a1e5-57aeb4c90ec9" />

Now it handled that without any loop so maybe the problem is elsewhere. 
I use opencode on Windows 11 and I did `rtk init -g --opencode` which successfully installed a hook but did not update my `%USERPROFILE%\.config\opencode\AGENTS.md` so maybe that's the root cause?

bb1:
Related: #1848 adds a skill file that teaches agents about `RTK_DISABLED=1` to break out of exactly this kind of loop. When the agent detects compressed/unusable output, it can prefix the retry with `RTK_DISABLED=1` to get raw output instead of looping indefinitely.

Issue #345, opened 2026-03-05
Bug: RTK_DISABLED=1 is ignored by rtk rewrite


pszymkowiak:
## Description

`RTK_DISABLED=1 curl ...` is still rewritten to `RTK_DISABLED=1 rtk curl ...` by `rtk rewrite`. The env var prefix is preserved but never checked — `RTK_DISABLED` doesn't exist anywhere in the Rust codebase.

## Reproduction

```bash
rtk rewrite "RTK_DISABLED=1 curl https://example.com"
# Output: RTK_DISABLED=1 rtk curl https://example.com
# Expected: no rewrite (exit 1)
```

## Expected behavior

When a command is prefixed with `RTK_DISABLED=1`, `rtk rewrite` should exit 1 (no rewrite).

## Fix

Check for `RTK_DISABLED=1` prefix in `rewrite_command()` and skip rewrite.

PR #1848, opened 2026-05-12
feat(skills): add RTK_DISABLED=1 opt-out skill and install via rtk init

### LINKED ISSUE #1777: Feature request: per-invocation escape hatch (flag/env var) so an agent can self-disable rtk rewriting when stuck
## Summary

Request: a documented, per-invocation escape hatch that disables RTK's command rewriting/output condensation for a single command — ideally both a CLI flag and an env var — that an LLM agent can set itself when it determines RTK's rewritten output is hiding information it needs.

The motivating case: a Claude/Cursor agent ran `npm test` (jest), RTK's wrapper condensed the output, and the failing test detail the agent needed wasn't in the rewritten summary. The agent then retried the same command repeatedly with minor variations, getting the same condensed output each time, and got stuck in a loop until a human intervened.

If there were a known, agent-discoverable knob like `RTK_PASSTHROUGH=1 npm test` or `npm test --rtk-passthrough` (name TBD), the agent's loop-recovery logic could try it once and get the raw underlying output without the user having to uninstall RTK or reconfigure their shell.

## Why a flag/env var (not just \"uninstall rtk\")

- The whole point of RTK is that it's installed transparently — for humans and for the common agent path it's a net win. An agent shouldn't need to know how RTK was installed (Homebrew? PATH shim? hook?) to bypass it.
- Telling a user \"please disable rtk and rerun\" is exactly the kind of human-in-the-loop escalation RTK is supposed to reduce. A documented escape hatch keeps the agent unblocked.
- Existing wrappers in this space (e.g. `NO_COLOR`, `CI=1`, `FORCE_COLOR=0`, `npm_config_*`) have set the precedent that a single env var is the right surface area for tools that an agent might run unattended.

## Suggested shape

Any of these would solve it; ranked by how useful they'd be to an agent:

1. **`RTK_PASSTHROUGH=1` env var** — when set, RTK execs the underlying tool directly (or runs as a no-op wrapper that streams stdout/stderr unchanged and propagates exit code). One-shot via `RTK_PASSTHROUGH=1 npm test` is ideal because it doesn't pollute the parent shell.
2. **`--rtk-passthrough` / `--rtk-off` CLI flag** — RTK strips it from `argv` before invoking the wrapped tool. Useful when the agent can edit the command but not the env.
3. **`RTK_DISABLE=jest` (or `RTK_DISABLE=*`) env var** — finer-grained: disable rewriting for a specific wrapped command. Helpful for users who want RTK on for `tsc` but off for `jest`.
4. **Documented exit-code/marker in the rewritten output** — e.g. \"to see raw output, rerun with `RTK_PASSTHROUGH=1`\" appended to RTK's summary footer. This is what makes it agent-discoverable: the next time the agent reads condensed output and isn't satisfied, the recovery action is right there in the output it just read.

(4) is the cheapest and arguably the highest-leverage piece — without it, agents won't know the escape hatch exists; with it, they'll learn the pattern from a single observation.

## Acceptance criteria

- A single, documented mechanism to fully disable RTK rewriting for one command invocation.
- Documented in the README under a heading an LLM is likely to retrieve (e.g. \"Disabling RTK\" or \"Escape hatch for agents\").
- Mentioned (as a one-line hint) in RTK's own condensed output footer, so an agent reading the summary sees how to escalate to raw output.
- Exit code of the underlying tool is preserved exactly; stdout/stderr stream through unchanged.

## Related

This is adjacent to a few open issues that are essentially \"RTK's rewriting silently dropped/altered output I needed\" (#1772 tsc `--pretty` false-success, #1765 `go test -cover` coverage missing, #1773 empty hook output). Each of those is a real bug to fix on its own, but a generic escape hatch would also have unblocked the agent in each case as a fallback while the specific wrapper bug is being fixed.

bb1:
## Summary

- Add an agent-agnostic skill (`skills/raw-command-output/SKILL.md`) that teaches LLM agents to prefix commands with `RTK_DISABLED=1` when they need raw, uncompressed output (JSON/XML/CSV for data analysis)
- Install the skill to `~/.agents/skills/` during `rtk init --opencode` (and combined Claude+OpenCode modes)
- Show skill path in `rtk status` output; remove on uninstall

## Motivation

When rtk compresses command output for token savings, agents sometimes cannot parse the result (e.g., JSON data for analysis). This causes infinite retry loops where the agent keeps re-issuing the same command with slight variations.

The `RTK_DISABLED=1` env prefix already exists in the Rust binary (#345) but agents don't know about it. This skill teaches them to use it proactively.

## Related Issues

- Closes #1777 — per-invocation escape hatch so agents can self-disable rtk when stuck
- Relates to #831 — output filtering causes agents to burn tokens on retry loops
- Relates to #732 — `rtk git branch --show-current` loop in OpenCode
- Relates to #1153 — `RTK_DISABLED=1` env var to bypass hooks (already implemented in binary; this PR makes agents aware of it)

## Changes

| File | Change |
|------|--------|
| `skills/raw-command-output/SKILL.md` | New: agent-agnostic skill file |
| `src/hooks/constants.rs` | Add `AGENTS_DIR`, `SKILLS_SUBDIR`, `RTK_SKILL_DIR`, `RTK_SKILL_FILE` constants |
| `src/hooks/init.rs` | Add `rtk_skill_path()`, `ensure_rtk_skill_installed()`, update install/remove/status flows |

## Testing

- `cargo check` passes (verified via Docker `rust:latest`)
- Skill installs to `~/.agents/skills/raw-command-output/SKILL.md`
- Existing tests unaffected (skill install is home-relative, orthogonal to tempdir-based plugin tests)

### DIFF
--- a/skills/raw-command-output/SKILL.md
+++ b/skills/raw-command-output/SKILL.md
@@ -1,0 +1,28 @@
+---
+name: raw-command-output
+description: Use when shell command output appears compressed, summarized, or mangled, or when you need exact raw output (JSON/XML/CSV) for data analysis. Teaches RTK_DISABLED=1 prefix to bypass rtk output filtering.
+allowed-tools: Bash
+---
+
+# Raw Command Output
+
+## When to Use
+
+- Shell command output appears compressed, summarized, or truncated
+- You need to parse raw JSON/XML/CSV data from a command
+- A command succeeded but the output is unusable for analysis
+
+## How to Use
+
+Prefix your shell command with `RTK_DISABLED=1` to bypass output filtering:
+
+    RTK_DISABLED=1 aws ec2 describe-spot-price-history --output json
+    RTK_DISABLED=1 kubectl get pods -o json
+    RTK_DISABLED=1 curl -s https://api.example.com/data
+    RTK_DISABLED=1 terraform show -json
+
+This is a standard shell environment variable prefix. It is recognised by
+the rtk binary and all agent hooks (OpenCode, Claude Code, Cursor, etc.).
+
+The system also automatically detects retries and disables filtering,
+but use this prefix proactively when you know you need raw data.
--- a/src/hooks/constants.rs
+++ b/src/hooks/constants.rs
@@ -18,6 +18,11 @@ pub const OPENCODE_SUBDIR: &str = "opencode";
 pub const PLUGIN_SUBDIR: &str = "plugins";
 pub const OPENCODE_PLUGIN_FILE: &str = "rtk.ts";
 
+pub const AGENTS_DIR: &str = ".agents";
+pub const SKILLS_SUBDIR: &str = "skills";
+pub const RTK_SKILL_DIR: &str = "raw-command-output";
+pub const RTK_SKILL_FILE: &str = "SKILL.md";
+
 pub const CURSOR_DIR: &str = ".cursor";
 pub const CODEX_DIR: &str = ".codex";
 pub const GEMINI_DIR: &str = ".gemini";
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -8,7 +8,8 @@ use std::path::{Path, PathBuf};
 use tempfile::NamedTempFile;
 
 use crate::hooks::constants::{
-    CONFIG_DIR, CURSOR_DIR, GEMINI_DIR, OPENCODE_PLUGIN_FILE, OPENCODE_SUBDIR, PLUGIN_SUBDIR,
+    AGENTS_DIR, CONFIG_DIR, CURSOR_DIR, GEMINI_DIR, OPENCODE_PLUGIN_FILE, OPENCODE_SUBDIR,
+    PLUGIN_SUBDIR, RTK_SKILL_DIR, RTK_SKILL_FILE, SKILLS_SUBDIR,
 };
 
 use super::constants::{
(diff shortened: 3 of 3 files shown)


bb1:
@pszymkowiak — would appreciate your review on this when you get a chance.

Quick context: this adds an agent-agnostic `SKILL.md` that teaches agents to use `RTK_DISABLED=1` when they need raw command output (e.g. after retries or when parsing JSON/XML). `rtk init` installs it to `~/.agents/skills/` (OpenCode) and `~/.claude/skills/` (Claude Code).

One design question: currently only OpenCode and Claude Code support the `skills/<name>/SKILL.md` format for global skill discovery. Other agents use different conventions (Cursor `.cursor/rules/*.mdc`, Cline `.clinerules/`, Windsurf `.windsurfrules`, Codex `AGENTS.md`, etc.). Is it enough to support just Claude Code and OpenCode for now, or should we also convert the skill into agent-specific rule formats for the others?

huzvak (commented):
Hi @bb1 . Good initiative. I am only wondering: Is this the way to go? What if LLM/OpenCode/Claude decides to always use this prefix instead? Then the whole RTK misses the point. 

Maybe my thinking is wrong, and the agents/models are smart enough not to do shortcuts, and simply not add it each time. If so, please elaborate.

baldassarreFe:
Any progress on this PR? What's blocking it from being merged?

KuSh:
Thanks for this, and for spelling out the retry-loop failure mode. Teaching agents about the RTK_DISABLED=1 escape hatch now ships through the awareness files added by #3372 (merged 2026-09-07): rtk init writes an awareness block for every agent according to awareness.level, and both the high and the full file document RTK_DISABLED=1 <cmd>. The env var itself was already implemented in the binary, as your description notes. Two things are still open: the default awareness level does not mention the escape hatch, and rtk installs nothing into the agent skills directories, so an issue or a narrow PR for a skill channel on top of the awareness model would be welcome. Please comment or reopen if I missed a part of this.

PR #1855, opened 2026-05-12
fix(git): preserve context lines before first change in compact_diff

### LINKED ISSUE #1852: fix(git): compact_diff drops context lines before first change in hunk

## Description

`compact_diff()` in `src/cmds/git/git.rs` silently drops all context lines that appear **before** the first `+`/`-` line in a hunk. Only context lines **after** the first change are emitted.

This means AI agents (and users) lose the surrounding code above a change, making it harder to understand what the diff applies to.

## Root Cause

Line ~416 in `compact_diff()`:

```rust
    // Context line
    if hunk_shown > 0 {  // <-- BUG: skips context before first +/-
        result.push(format!("  {}", line));
        hunk_shown += 1;
    }
}
```

The `if hunk_shown > 0` guard means context lines are only emitted after the first `+`/`-` is seen. Context lines before the first change are silently dropped.

## Reproduction

```bash
# Create a file with context above the change
cat > test.txt <<'FILE'
line 1 - context above
line 2 - context above
line 3 - context above
line 4 - context above
line 5 - this will change
line 6 - context below
FILE
git add test.txt && git commit -m "initial"
sed -i 's/this will change/MODIFIED/' test.txt

# Raw git diff shows 3 context lines before the change
git diff test.txt | grep "context above"
# Output: 4 matches (lines 1-4)

# RTK drops them
rtk git diff test.txt | grep "context above"
# Output: 1 match (only the @@ header line)
```

## Expected vs Actual

| | Before change | After change |
|---|---|---|
| **Expected** | 3 context lines shown | 3 context lines shown |
| **Actual** | 0 context lines shown | 3 context lines shown |

## Fix

Remove the `if hunk_shown > 0` guard. Context lines should always be emitted (they already count toward the `max_hunk_lines` budget, so output remains bounded).

## Impact

- **Severity**: Medium — agents lose code context above changes, leading to incorrect edits
- **Token cost**: Adds ~3 lines per hunk on small diffs; on large diffs the 100-line/hunk and 500-line total caps dominate, so savings are preserved (measured 60% savings on a 200-change diff with the fix applied)
- **Version**: Verified on v0.39.0

## Related

- #827 (git diff truncation)
- #1669 (git diff filtering)

paolomainardi:

## Summary

- Removes the `if hunk_shown > 0` guard in `compact_diff()` that silently dropped context lines before the first `+`/`-` in each hunk
- Adds regression test `test_compact_diff_preserves_context_before_first_change`

## Before / After

| | Before fix | After fix |
|---|---|---|
| Context lines before change | Dropped (0 shown) | Preserved (3 shown) |
| Context lines after change | Shown | Shown (unchanged) |
| Token savings on large diffs | ~60% | ~60% (caps still active) |

All existing tests pass (1732 total).

Fixes #1852

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -412,11 +412,9 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
                     hunk_skipped += 1;
                 }
             } else if hunk_shown < max_hunk_lines && !line.starts_with("\\") {
-                // Context line
-                if hunk_shown > 0 {
-                    result.push(format!("  {}", line));
-                    hunk_shown += 1;
-                }
+                // Context line — always emit (before and after changes)
+                result.push(format!("  {}", line));
+                hunk_shown += 1;
             }
         }
 
@@ -2792,6 +2790,39 @@ no changes added to commit (use "git add" and/or "git commit -a")
         assert!(
             result.contains("+3 lines omitted"),
             "Expected '+3 lines omitted' when 6 body lines truncated to 3, got:\n{}",
+            result
+        );
+    }
+
+    #[test]
+    fn test_compact_diff_preserves_context_before_first_change() {
+        // Context lines BEFORE the first +/- in a hunk must be preserved.
+        // Regression test: previously `if hunk_shown > 0` dropped them.
+        let diff = r#"diff --git a/foo.rs b/foo.rs
+--- a/foo.rs
++++ b/foo.rs
+@@ -1,5 +1,5 @@
+ fn setup() {
+     let x = 1;
+     let y = 2;
+-    let z = 3;
++    let z = 999;
+ }
+"#;
+        let result = compact_diff(diff, 100);
+        assert!(
+            result.contains("fn setup()"),
+            "Expected context line 'fn setup()' before first change, got:\n{}",
+            result
+        );
+        assert!(
+            result.contains("let x = 1"),
+            "Expected context line 'let x = 1' before first change, got:\n{}",
+            result
+        );
+        assert!(
+            result.contains("let y = 2"),
+            "Expected context line 'let y = 2' before first change, got:\n{}",
             result
         );
     }


src/cmds/git/git.rs
```diff
@@ -412,11 +412,9 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
                     hunk_skipped += 1;
                 }
             } else if hunk_shown < max_hunk_lines && !line.starts_with("\\") {
-                // Context line
-                if hunk_shown > 0 {
-                    result.push(format!("  {}", line));
-                    hunk_shown += 1;
-                }
+                // Context line — always emit (before and after changes)
+                result.push(format!("  {}", line));
```
KuSh, line 416: Emitting every context line unconditionally means leading context now consumes the per-hunk `max_hunk_lines` budget (100). A hunk with 100 or more leading context lines therefore emits *only* context and truncates away every actual `+`/`-` line.

Repro: 120 context lines followed by one `-`/`+` pair. HEAD shows the change. With this patch the output is `let v98`, `let v99`, `let v100`, `... (2 lines truncated)` — the change itself is gone.

That shape is not exotic: `git diff -W`, `--function-context` and `-U<big>` all produce it, and it reaches users through `gh pr diff` and `glab mr diff` as well as `rtk git diff`.

The global `max_lines` cap is affected the same way — it now fills with context, truncating real changes earlier. On a real 13.8 KB diff from this repo at `max_lines=200`, changed lines shown drop from 183 to 168 and the file summary goes from `+166 -16` to `+151 -16`.

`never_worse()` doesn't rescue either case, since the filtered output here is *smaller* than raw, not larger — so the guard is happy while content is being dropped.

KuSh (changes requested):
Requesting changes, and apologies this sat so long without a response.

The bug you found is real: context lines before the first `+`/`-` in a hunk were silently dropped, and `if hunk_shown > 0` is the cause. That much is confirmed. The fix needs one more piece, though, because removing the guard without giving context its own budget trades one silent loss for a worse one — details inline. I built both HEAD and your patch and compared real output through `rtk pipe -f git-diff` to check this rather than reasoning about it; all 2659 unit tests pass with your patch applied.

Suggested direction: keep the fix, but bound leading context separately — cap it at ~3 lines per hunk, or reserve budget for `+`/`-` lines so a change can never be crowded out by its own context.

Two notes on the surrounding state, so you know what you're rebasing into:

- The branch is currently conflicting with `develop`.
- #3269 is an open PR touching these same lines (it removes the two-space indent so `grep '^-'` anchors again). It leaves the `hunk_shown > 0` guard alone, so the two changes are complementary — but whichever lands second will need a rebase.

A third thing, pre-existing but newly routine once leading context is emitted: `... (N lines truncated)` undercounts, because context lines dropped past `max_hunk_lines` never reach `hunk_skipped` (there's no `else` arm on the context branch). In my repro it reported "2 lines truncated" when 22 were dropped. A one-line `else { hunk_skipped += 1; }` fixes it and would be welcome in the same PR.

KuSh:
Closing in favour of #3022, which makes the same leading context fix and also removes the per hunk and diff wide truncation caps at every call site, with integration coverage. The leading context behaviour has meanwhile also landed on develop through a separate commit. Review and comments are welcome on #3022, or reply here if you think the scope differs.

PR #1860, opened 2026-05-12
Add Lua and common Lua commands support

LK4D4:
## Summary
- add rtk luacheck, and rtk busted routing
- compact luacheck and busted output with focused diagnostics and tests
- detect .lua files in rtk read and smart summaries

## Local verification
- cargo fmt --all --check
- cargo clippy --all-targets
- cargo test --all


LK4D4:
Closing this draft PR because we switched to a local release build/install instead of using GitHub Actions for this branch.

PR #1895, opened 2026-05-15
fix(filters): aggresivity batch fix

aeppling:
## Summary

- Some filter were too aggressive, lower filtering & more accurate results

### DIFF
--- a/src/cmds/cloud/container.rs
+++ b/src/cmds/cloud/container.rs
@@ -54,14 +54,18 @@ where
 fn docker_ps(_verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
 
+    // Baseline the LLM would otherwise see.
     let raw = exec_capture(resolved_command("docker").args(["ps"]))
         .map(|r| r.stdout)
         .unwrap_or_default();
 
+    // One structured call over *all* containers (`-a`) — splitting on the State
+    // field lets us list crashed/exited ones too, which plain `docker ps` hides.
     let result = exec_capture(resolved_command("docker").args([
         "ps",
+        "-a",
         "--format",
-        "{{.ID}}\t{{.Names}}\t{{.Status}}\t{{.Image}}\t{{.Ports}}",
+        "{{.State}}\t{{.ID}}\t{{.Names}}\t{{.Status}}\t{{.Image}}\t{{.Ports}}",
     ]))
     .context("Failed to run docker ps")?;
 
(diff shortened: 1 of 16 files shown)


src/cmds/cloud/container.rs
```diff
@@ -71,43 +75,66 @@ fn docker_ps(_verbose: u8) -> Result<i32> {
...
-    rtk.push_str(&format!("[docker] {} containers:\n", count));
+    let mut rtk = String::new();
+    rtk.push_str(&format!("[docker] {} running:\n", running.len()));
+    for parts in running.iter().take(20) {
+        if let Some(l) = format_line(parts, true) {
+            rtk.push_str(&l);
+        }
+    }
+    if running.len() > 20 {
+        rtk.push_str(&format!("  ... +{} more\n", running.len() - 20));
```
KuSh: nitpick, prefer utf8 character `…`
```suggestion
        rtk.push_str(&format!(" … +{} more\n", running.len() - 20));
```

src/cmds/cloud/container.rs
```diff
@@ -71,43 +75,66 @@ fn docker_ps(_verbose: u8) -> Result<i32> {
...
+        for parts in stopped.iter().take(20) {
+            if let Some(l) = format_line(parts, false) {
+                rtk.push_str(&l);
             }
         }
-    }
-    if count > 15 {
-        rtk.push_str(&format!("  ... +{} more", count - 15));
+        if stopped.len() > 20 {
+            rtk.push_str(&format!("  ... +{} more\n", stopped.len() - 20));
```
KuSh:
```suggestion
            rtk.push_str(&format!(" … +{} more\n", stopped.len() - 20));
```

src/cmds/cloud/container.rs
```diff
@@ -173,21 +200,22 @@ fn docker_images(_verbose: u8) -> Result<i32> {
...
-                image.to_string()
-            };
-            rtk.push_str(&format!("  {} [{}]\n", short, size));
+            rtk.push_str(&format!("  {} [{}]\n", image, size));
         }
     }
-    if lines.len() > 15 {
-        rtk.push_str(&format!("  ... +{} more", lines.len() - 15));
+    if lines.len() > MAX_IMAGES {
+        rtk.push_str(&format!("  ... +{} more\n", lines.len() - MAX_IMAGES));
```
KuSh:
```suggestion
        rtk.push_str(&format!(" … +{} more\n", lines.len() - MAX_IMAGES));
```

src/cmds/git/git.rs
```diff
@@ -911,7 +925,13 @@ fn run_status(args: &[String], verbose: u8, global_args: &[String]) -> Result<i3
         return Ok(result.exit_code);
     }
 
-    let formatted = format_status_output(&result.stdout);
+    let mut formatted = format_status_output(&result.stdout);
+
+    // Porcelain `-b` reduces a detached HEAD to "## HEAD (no branch)"; restore
+    // the explicit "HEAD detached at <sha>" from the plain status we captured.
+    if let Some(detached) = extract_detached_head(&raw_output) {
```
KuSh: Wouldn't it be better to check that before calling `format_status_output` and pass it `detached` instead of replacing the result afterward?

src/cmds/python/pip_cmd.rs
```diff
@@ -162,16 +166,21 @@ fn filter_pip_list(output: &str) -> String {
...
 
-        for pkg in pkgs.iter().take(10) {
+        for pkg in pkgs.iter().take(MAX_PER_LETTER) {
             result.push_str(&format!("  {} ({})\n", pkg.name, pkg.version));
         }
 
-        if pkgs.len() > 10 {
-            result.push_str(&format!("  ... +{} more\n", pkgs.len() - 10));
+        if pkgs.len() > MAX_PER_LETTER {
+            result.push_str(&format!("  ... +{} more\n", pkgs.len() - MAX_PER_LETTER));
```
KuSh, line 183: while you're at it:
```suggestion
            result.push_str(&format!(" … +{} more\n", pkgs.len() - MAX_PER_LETTER));
```

src/cmds/python/ruff_cmd.rs
```diff
@@ -185,6 +181,28 @@ pub fn filter_ruff_check_json(output: &str) -> String {
...
+            compact_path(&diag.filename),
+            diag.location.row,
+            diag.location.column,
+            diag.code,
+            truncate(diag.message.trim(), 100),
+        ));
+    }
+    if diagnostics.len() > MAX_VIOLATIONS {
+        result.push_str(&format!(
+            "  ... +{} more violations\n",
```
KuSh:
```suggestion
            " … +{} more violations\n",
```

src/parser/formatter.rs
```diff
@@ -112,6 +118,26 @@ impl TokenFormatter for TestResult {
...
+            && self.dependencies.iter().all(|d| d.latest_version.is_none());
+        if is_listing {
+            let total = self.total_packages.max(self.dependencies.len());
+            let mut lines = vec![format!("{} packages", total)];
+            for dep in self.dependencies.iter().take(50) {
+                let dev = if dep.dev_dependency { " (dev)" } else { "" };
+                lines.push(format!("  {} {}{}", dep.name, dep.current_version, dev));
+            }
+            if self.dependencies.len() > 50 {
+                lines.push(format!("  ... +{} more", self.dependencies.len() - 50));
```
KuSh, line 136:
```suggestion
                lines.push(format!(" … +{} more", self.dependencies.len() - 50));
```

src/parser/formatter.rs
```diff
@@ -112,6 +118,26 @@ impl TokenFormatter for TestResult {
...
+        let is_listing = self.outdated_count == 0
+            && !self.dependencies.is_empty()
+            && self.dependencies.iter().all(|d| d.latest_version.is_none());
+        if is_listing {
+            let total = self.total_packages.max(self.dependencies.len());
+            let mut lines = vec![format!("{} packages", total)];
+            for dep in self.dependencies.iter().take(50) {
+                let dev = if dep.dev_dependency { " (dev)" } else { "" };
+                lines.push(format!("  {} {}{}", dep.name, dep.current_version, dev));
+            }
```
KuSh, line 134: Isn't there a hierarchy in deps? dep > dev dep > opt dep > ... shouldn't we first print dependencies, then if < 50 pick dev dependencies until we reach 50?

KuSh (approved):
Minor nitpicks and questions. Otherwise LGTM!

pszymkowiak:
Tested the branch build against real commands — solid work, the filters no longer hide crashed containers, skipped/xfail tests, ruff locations, or report a false `up-to-date`. 11 commands verified end-to-end. Two points before merge:

**1. `git stash` no-op is mis-reported on non-English locales** (`git.rs:1510`)

`result.combined().contains("No local changes")` matches an English-only string. On a FR locale `git stash` prints *"Pas de modifications locales à sauver"*, so the check fails and a no-op stash prints **`ok stashed`** — telling the agent its changes were saved when nothing was. (Verified on a FR machine.)

This is the tip of a larger issue: `git.rs` matches localized git output in ~9 places (rebase/merge detection L750-762, `nothing to commit` L846/L1078, `Everything up-to-date` L1123, `Already up to date` L1183, shortstat parsing L1194-1210…). The robust fix is to run rtk's internal git invocations with `LC_ALL=C` so git emits stable English — worth a dedicated change. At minimum, L1510 should not regress a no-op into `ok stashed`.

**2. `constants.rs` — removing `env`/`.env` from `NOISE_DIRS`** is the only change with no test and no rationale. `env/` is a very common Python virtualenv dir; dropping it can flood `rtk tree`/`find`/`grep` with venv noise. Could you add a line explaining the intent (and ideally a test)?

Everything else is good to go.

PR #1906, opened 2026-05-16
fix(go): preserve coverage output from go test -cover on passing runs

### LINKED ISSUE #1765: bug: go test -cover didn't output test coverage
Bash(go test -count=1 -cover ./xxx/... ./xxx/...)
Go test: 348 passed in 9 packages

should show: coverage: 80.1% of statements

tylern91:
## Summary

- Detect `coverage:` lines in `filter_go_test_json()` and route them to a dedicated `coverage` field on `PackageResult`, not `package_fail_output`
- Append per-package coverage lines below the passing summary when present
- Preserves the compact format for runs without `-cover`

## Root Cause

Coverage lines arrive as package-level output JSON events (no `Test` field). The handler previously pushed all package-level output to `package_fail_output`, which is only emitted when `package_failed == true`. On passing runs, coverage data was silently discarded.

## Output Format

```
Go test: 348 passed in 9 packages
  foo: coverage: 80.1% of statements
  bar: coverage: 92.3% of statements
```

## Test plan

- [x] Added `test_filter_go_test_cover_output_preserved()` verifying coverage lines appear in passing-run output
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` — 1871 passed, 0 failures
- [x] Runs without `-cover` unchanged (coverage vec empty, indented block skipped)

Closes #1765

### DIFF
--- a/src/cmds/go/go_cmd.rs
+++ b/src/cmds/go/go_cmd.rs
@@ -40,6 +40,7 @@ struct PackageResult {
     failed_tests: Vec<(String, Vec<String>)>, // (test_name, output_lines)
     package_failed: bool,                     // package-level failure (timeout, signal, etc.)
     package_fail_output: Vec<String>,         // output lines collected before the package fail
+    coverage: Vec<String>,                    // per-package coverage lines from go test -cover
 }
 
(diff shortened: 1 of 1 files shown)


tylern91:

PR #1911, opened 2026-05-16
feat(discover): add jq and yq rewrite rules; add yq TOML filter

### LINKED ISSUE #1364: feat: add `rtk jq` command for compressing large JSON extraction output
## Summary

Add `rtk jq` to compress large JSON outputs from `jq` queries — currently a transparent fallback with 0% savings.

Searched existing issues (#229, #294, #538, PR #100) — didn't find existing `jq` coverage. Apologies if I missed one.

## Motivation

`rtk discover` on my workstation over the past 30 days (152 Claude Code sessions, 2,653 Bash invocations) shows `jq` as one of the top uncovered commands:

| Command | Count | Example |
| ------- | ----- | ------- |
| `jq`    | 24    | `jq -r '.paths \| keys[]' swagger_coverage/openapi.json` |

Caveat per #538: some of these may have gone through Claude Code's native tools rather than Bash. But `jq` specifically almost always runs via Bash (no native `jq` equivalent), so the count is likely close to real.

Typical use cases in my sessions — working with an `openapi.json` for a Go backend:
- `jq '.paths | keys[]' openapi.json` → can return hundreds of route paths
- `jq '.definitions | keys' openapi.json` → large type lists
- `jq '.' config.json` → pretty-print of multi-KB configs

All produce output that either floods context or gets truncated ad-hoc.

## Proposed behavior

- **Small results (<50 lines, configurable):** pass through verbatim. No compression, no loss.
- **Large arrays:** `first_N + "… (X more omitted) …" + last_N`, preserving valid-looking JSON structure so the LLM can still reason about shape.
- **Large object dumps:** same strategy — head/tail with ellipsis in the middle, keeping brace balance readable.
- **Respect `-r` / `--raw-output`:** when output is raw text rather than JSON, treat like `rtk read` (line-based head/tail).

Based on compression ratios I see for `rtk read` and `rtk grep` on similar output shapes, I'd expect 40–70% savings on these calls. Not a huge category (~24/month in my data), but easy win with no parser complexity — `jq` output is already structured.

## Offer

Happy to test patches against real sessions — `rtk gain` gives clean before/after and I can share anonymized `rtk discover` dumps if it helps prioritize. Let me know if there's a beta channel.

## Environment

- `rtk` 0.37.0 (Homebrew)
- macOS Darwin 23.6.0
- Claude Code with `rtk-rewrite.sh` hook in `~/.claude/settings.json` PreToolUse

### LINKED ISSUE #1910: feat: add yq rewrite rule and TOML filter for YAML query output compression
## Problem

`yq` is the YAML counterpart to `jq` — it produces verbose YAML output that can be large when querying multi-document manifests, Kubernetes configs, or Helm values files. Currently rtk has no hook rewrite rule for `yq`, so bare `yq` commands are never compressed even though a simple truncation filter would save significant tokens.

## Proposed Solution

Add a `yq` rewrite rule to `src/discover/rules.rs` and create `src/filters/yq.toml` following the same pattern as the existing `jq.toml`:

- `match_command = "^yq\\b"`
- `strip_lines_matching = ["^\\s*$"]` — remove blank lines
- `max_lines = 40`
- `truncate_lines_at = 120`

## Related

- Similar to #1364 (jq rewrite rule) — same compression pattern for structured query tools
- `jq.toml` already exists as the template for this approach

tylern91:
## Summary

- **jq**: The `jq.toml` filter (max_lines=40, truncate_lines_at=120) has existed but had no rewrite rule in `discover/rules.rs`. Bare `jq` commands were never routed through the filter by the hook. This PR adds the missing `RtkRule` so the hook auto-rewrites `jq` → `rtk jq`.

- **yq**: `yq` is the YAML equivalent of `jq` — same compression pattern. Creates `src/filters/yq.toml` (max_lines=40, truncate_lines_at=120, strips blank lines) and adds the corresponding `RtkRule`.

## Changes

| File | Action |
|------|--------|
| `src/discover/rules.rs` | Add jq and yq `RtkRule` entries (System category) |
| `src/filters/yq.toml` | CREATE — TOML filter for yq output, 2 inline tests |
| `src/core/toml_filter.rs` | Filter count 59→60; add jq+yq to expected array; concat assertion 60→61 |

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` — zero warnings
- [x] `cargo test --all` — 1870 passed, 6 ignored
- [x] `test_builtin_filter_count` passes with count=60
- [x] `test_builtin_all_expected_filters_present` — jq and yq verified present
- [x] yq inline tests: short output passthrough, empty input

## Closes

Closes #1364 (jq rewrite rule missing from discover)
Closes #1910 (yq filter + rewrite rule)


tylern91:

PR #1912, opened 2026-05-16
fix(discover): guard dbg/gdbg debugger commands from rewrite hook

### LINKED ISSUE #1387: Bash hook mangles dbg <pdb-verb> commands, subcommand swallowed or dbg prefix dropped.
When [dbg](https://redknightlois.github.io/dbg) or [gdbg](https://redknightlois.github.io/dbg/gdbg.html) (a unified debug CLI with subcommand dispatch) is invoked through Claude Code with the rtk bash hook active, subcommand names that collide with bash builtins or pdb/lldb/jdb verbs get mis-routed. The observed failure modes are either (a) the `dbg` prefix is dropped and the shell tries to execute the subcommand as a standalone binary, or (b) subsequent arguments lose context. Affects: `p`, `n`, `c`, `b`, `s`, `r`, `where`, `run`, `break`, `stack`, `exec`, `continue`, `step`. 

  ## Observed behavior

  Concrete traces collected across agent sessions:

  ```bash
  # dbg stack → "stack: command not found"
  # The `dbg` prefix was dropped before exec; only `stack` reached the shell.
  dbg stack
  # → bash: stack: command not found

  # Only worked when explicitly routed through rtk proxy:
  rtk proxy dbg stack
  # → (correct stack frame output)
  ```

  ```bash
  # Bare arguments shell-evaluated
  dbg p distance_miles
  # → empty / "command not found" / evaluates distance_miles as a bash token

  # Quoting the subcommand + arg as one string worked:
  dbg "p distance_miles"
  # → 1750
  ```

  ```bash
  # `dbg b` intercepted, breakpoint never set
  dbg b broken.py:27
  # → no breakpoint registered

  # `dbg break` worked:
  dbg break broken.py:27
  # → breakpoint set
  ```

  ```bash
  # `dbg where` fails; `dbg stack` also fails (see above)
  dbg where
  # → bash: where: command not found

  # Works via:
  rtk proxy dbg stack
  ```

  ```bash
  # Non-deterministic across invocations — sometimes intercepted, sometimes not:
  dbg run
  dbg continue
```

## Expected behavior

`dbg <anything>` should always reach the `dbg` binary with arguments intact. `dbg` is a single binary with internal subcommand dispatch the hook should not parse or rewrite past the first argument. Moreover, the tool is designed to be agent friendly and token efficient by removing noise, cutting out the responses actually hampers the use case.

tylern91:
## Problem

`dbg stack`, `dbg p variable`, `dbg where`, and `gdbg break main.py:27` were at risk of being mangled by the hook rewrite system. If a user added `dbg` to `transparent_prefixes` in their config, the prefix would be stripped **before** `IGNORED_PREFIXES` is checked — causing the subcommand to be routed or dropped incorrectly.

Root cause: `rewrite_segment_inner()` processes `transparent_prefixes` after `SHELL_PREFIX_BUILTINS` but before the `IGNORED_PREFIXES` check returns `None`. A user-configured `dbg` transparent prefix bypasses the ignore guard.

## Fix

Two-layer defense:

1. **`IGNORED_PREFIXES`** (rules.rs): Add `"dbg "` and `"gdbg "` to the prefix ignore list — covers the standard path.

2. **Early-return guard** (registry.rs): Add an explicit `if trimmed.starts_with("dbg ") || trimmed.starts_with("gdbg ")` guard inside `rewrite_segment_inner()` that fires **before** the `transparent_prefixes` loop — closes the ordering gap.

## Changes

| File | Action |
|------|--------|
| `src/discover/rules.rs` | Add `"dbg "` and `"gdbg "` to `IGNORED_PREFIXES` |
| `src/discover/registry.rs` | Early-return guard before `transparent_prefixes` loop + new test |

## Test plan

- [x] `cargo clippy --all-targets -- -D warnings` — zero warnings
- [x] `cargo test --all` — 1871 passed, 6 ignored
- [x] `test_dbg_commands_pass_through` — covers bare commands and transparent_prefix edge case

## Closes

Closes #1387

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -738,6 +738,12 @@ fn rewrite_segment_inner(
         }
     }
 
+    // Debugger prefixes must never be stripped even if listed in transparent_prefixes
+    // (#1387: dbg stack / gdbg break main.py:27 were being mangled).
+    if trimmed.starts_with("dbg ") || trimmed.starts_with("gdbg ") {
+        return None;
+    }
+
     // User-configured wrapper prefixes (e.g. `docker exec mycontainer`). Same
     // strip-recurse-reprepend contract as the builtin list above.
     for prefix in transparent_prefixes {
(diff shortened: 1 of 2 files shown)


tylern91:

PR #1497, opened 2026-04-24
Fix RTK shim recursion for Vitest passthrough

jeffpcbr:
## Summary
- bypass RTK shims for child commands launched through resolved_command
- pass through help/version-style Vitest and Jest invocations instead of parsing them as test output
- synthesize Vitest version output through Node when available to avoid recursive shim rewrites

## Tests
- cargo test test_should_passthrough_info_args
- cargo test test_should_synthesize_vitest_version
- cargo test test_resolved_command_sets_rtk_disabled

PR #1922, opened 2026-05-16
fix(rewrite): preserve vitest metadata commands

tangi-vass:
## Summary

Fixes `rtk rewrite` producing invalid Vitest/Jest rewrites for metadata and non-run commands.

Before:
- `npx vitest --version` -> `rtk vitest --version`
- `vitest --run` -> `rtk vitest --run`

After:
- version/help and explicit non-run subcommands pass through unchanged
- runnable forms normalize to `rtk vitest run ...`

This is related to #1497, but fixes the issue at the rewrite registry layer before `rtk vitest` execution begins.

## Tests

- `cargo fmt --check`
- `cargo test discover::registry -- --nocapture`
- `XDG_DATA_HOME=/tmp/rtk_validation_data rtk cargo test -- --test-threads=1`
- `cargo build`

## Notes

Plain `cargo test` in this sandbox tries to write the default tracking DB under `~/.local/share/rtk` after a tracking test mutates `RTK_DB_PATH`; redirecting `XDG_DATA_HOME` to `/tmp` keeps the test DB inside a writable path and the full suite passes.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -4,7 +4,7 @@ use lazy_static::lazy_static;
 use regex::{Regex, RegexSet};
 
 use super::lexer::{split_on_operators, tokenize, TokenKind};
-use super::rules::{IGNORED_EXACT, IGNORED_PREFIXES, RULES};
+use super::rules::{RtkRule, IGNORED_EXACT, IGNORED_PREFIXES, RULES};
 
 /// Result of classifying a command.
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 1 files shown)


ebroder:
Still reproduces on 0.44.2, and the user-visible failure is worse than an invalid rewrite: the summary is a false clean.

In a project with 2120 vitest tests:

```console
$ rtk vitest list
PASS (0) FAIL (0)
$ rtk proxy npx vitest list | wc -l
2120
```

An agent reading `PASS (0) FAIL (0)` concludes its new test file isn't being collected and goes hunting for a config problem that doesn't exist. That is how I found it, after a while spent chasing a phantom glob bug. (The exit code is 1, so it isn't strictly silent, but the text an agent reads says PASS.)

`--version` is the worse one, because it isn't a no-op:

```console
$ rtk vitest --version
PASS (2120) FAIL (0)
$ rtk proxy npx vitest --version
vitest/4.1.10 darwin-arm64 node-v24.15.0
```

Asking for a version number runs the whole suite. In CI or an agent loop that is real wall-clock and real tokens, and the answer still isn't the version.

Neither needs `rtk` to be typed, since the hook rewrites both forms:

```console
$ echo '{"tool_name":"Bash","tool_input":{"command":"npx vitest list"}}' | rtk hook claude
{"hookSpecificOutput":{...,"updatedInput":{"command":"rtk vitest list"}}}
```

The registry-layer fix in the description (pass through version/help and explicit non-run subcommands, normalize runnable forms to `rtk vitest run ...`) still looks like the right shape.

PR #1932, opened 2026-05-17
feat(cmds/make): grouping Rust filter, 98% savings on parallel builds

iliaal:
## Why

The bundled `make` TOML filter (`src/filters/make.toml`) strips `make[N]:` directory chrome and applies head/tail truncation, but leaves the bulk of a real build untouched: long runs of `gcc/clang/g++ -c ... foo.c -o foo.o` lines. Measured 12-13% savings on typical parallel builds (`make -j32` on a PHP-style codebase), well below rtk's 60%+ target.

## What

Replaces the TOML filter with a Rust module that:

1. Groups compiler invocations (`cc`, `gcc`, `g++`, `clang`, `c++`, `ar`, `ranlib`, `ld`, `libtool`, `/bin/sh ./libtool`) into a one-line summary per phase: `compiled N files (W warnings, E errors)`.
2. Preserves diagnostics verbatim: lines containing `: warning:`, `: error:`, `: note:`, plus their preceding `In file included from` chains.
3. Flushes groups at link/install boundaries (`/usr/bin/install`, `libtool: link:`, `Making install`, shared-library link steps) and keeps those lines verbatim, since they're already short.
4. Falls back to chrome-strip-only passthrough when no compiler lines are detected, so short incremental builds, custom targets, and non-build invocations aren't broken.
5. Strips chrome (`make[N]:` directory enter/leave, `Nothing to be done`, blank-line runs) consistent with the old TOML filter.

## Measurements

| Workload | Input tokens | Output tokens | Savings |
|---|---|---|---|
| 246-line parallel build with 2 warnings | 2,507 | 39 | **98.4%** |
| Short incremental build (1 file + warning) | passthrough | preserved | n/a |
| `make[N]:` chrome only | n/a | `make: ok` | n/a |

## Tests

17 unit tests in `src/cmds/system/make_cmd.rs`, including:

- Token-savings assertion (>=60% on the long fixture)
- Grouping of mixed `gcc`/`clang`/`ar`/`ranlib`
- Warning and error preservation, with include-chain context
- Link/install boundary flush
- Empty input and chrome-only fallback
- Below-threshold compiler runs still counted at flush

Two fixtures: `make_parallel_build_raw.txt` (251 lines, representative parallel build) and `make_incremental_raw.txt`.

## Notes

- Removes `src/filters/make.toml`; routes through `Commands::Make` in `main.rs` alongside other `system/` commands.
- Updates `src/discover/rules.rs` make-savings estimate from 65% to 90% to reflect the new behavior.
- Verified with `cargo fmt --all && cargo clippy --all-targets -D warnings && cargo test --all`: fmt and clippy clean, 1,899 tests pass (2 unrelated pre-existing flakes in `core::tracking::tests::test_timed_execution_*`).
- Branch built on top of current `develop`.


iliaal:
bump

Rebased onto current `develop` (9da7da1) and the conflicts are resolved. Four
things needed fixing beyond the merge itself:

- `lazy_static` is no longer a dependency, so the six regexes are now
  `LazyLock<Regex>`.
- `build.rs` now requires the module in `src/cmds/system/mod.rs`.
- `test_every_subcommand_is_classified` needs `make` in `PASSTHROUGH`.
- The no-compiler fallback was uncapped. It now caps at 10 head + 40 tail
  lines with an omission marker.

The fallback is the one worth a look, since it is a behaviour change rather
than a merge fix. `strip_chrome` returns every surviving line, and it is the
path `make test`, `make check`, and custom targets take. The `make.toml` this
replaces capped at 50 lines, so shipping it uncapped would have regressed
those workloads: a PHPT-style `make test` emits thousands of progress lines
and puts the verdict last, so a head-only cap would drop the pass/fail
summary. Two tests cover it (`test_fallback_keeps_tail_summary`,
`test_fallback_short_output_is_untouched`).

Built-in filter count drops 63 to 62 with `make.toml` removed.

`cargo fmt --check` and `clippy --all-targets --all-features -D warnings` are
clean. `cargo test --all` passes except
`git_log_malformed_digit_run_propagates_real_git_error` and
`git_diff_dash_dash_stat_pathspec_after_double_dash_is_not_stat_flag`, which
fail identically on pristine `develop` here (git 2.34.1), so they are not from
this branch.

PR #1939, opened 2026-05-18
fix(gradlew): anchor 'e:' in ERROR_LINE to avoid mid-line false positives

luinstra:
## Summary

- `BuildLineFilter`'s `ERROR_LINE` regex had an unanchored `e: ` alternation intended to match kotlinc errors (`e: file:///...`), parallel to `WARN_LINE`'s anchored `^w: `.
- Without the anchor, `e: ` matched any substring `e: ` in a line — producing false positives on structured log lines like `Received message: 1` or `Source: REDACTED` (the `e: ` falls inside `message: ` / `Source: `).
- Anchored to `^e: ` so only line-leading occurrences match. Adds a regression test covering both the kept kotlinc case and the now-rejected mid-line cases.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets -- -D warnings && cargo test` — all pass
- [x] New regression test `test_build_error_line_anchored_avoids_mid_line_false_positives` runs and passes
- [x] Verified the regex change does not affect existing build-filter tests


YOMXXX:
LGTM as an outside reviewer 👍

Anchoring `e: ` with `^` to mirror `^w: ` (kotlinc warnings) is the right symmetry — without it, any log line containing `e: ` mid-string was being matched as a compiler error, which would produce false positives across normal build logs. The regression test with `Received message: 1` and `Source: REDACTED` is exactly the case that motivates the anchor.

Maintainers: tiny safe fix, zero downside. Recommend merging.

luinstra:
bump

PR #1940, opened 2026-05-18
fix(gradlew): collapse consecutive blank lines in build filter output

luinstra:
## Summary

- `BuildLineFilter::feed_line` was stateless: every line `filter_build_line` accepted got emitted as-is, including blanks intentionally kept as section separators.
- On real kotlin/JVM builds — which emit groups of `w:` warnings separated by blank lines — this produced multi-screen runs of blank lines between groups of warnings, drowning the actual signal.
- Make `BuildLineFilter` stateful with `last_emitted_was_blank: bool`. Single-blank separators are preserved; runs of 2+ collapse to one. Non-blank lines unaffected.

**Note on the "fix" framing**: BPE tokenizers fully collapse whitespace runs, so this patch saves ~0 tokens. The win is human readability (tee files, debug sessions, scrollback). Filed as `fix` because it removes pathological output noise that hides the actual build signal.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets -- -D warnings && cargo test` — all pass
- [x] Four new tests use byte-exact `assert_eq!` via a `run_build_filter` helper:
  - `test_build_filter_collapses_consecutive_blanks` — runs of `\n\n\n\n` collapse to `\n\n`
  - `test_build_filter_preserves_single_blank_separator` — single `\n\n` survives
  - `test_build_filter_collapse_survives_dropped_lines_between_blanks` — `> Task :…` stripped between blanks still collapses (regression for the "state only tracks emitted lines" decision)
  - `test_build_filter_leading_blanks_collapsed` — initial run of blanks collapses to one

> Targets `develop` per CONTRIBUTING.md. Independent of #1939 (same file, unrelated bug).

### DIFF
--- a/src/cmds/jvm/gradlew_cmd.rs
+++ b/src/cmds/jvm/gradlew_cmd.rs
@@ -100,15 +100,29 @@ fn new_gradle_command(args: &[String]) -> Command {
 }
 
 /// `StreamFilter` for build mode: keeps lines for which `filter_build_line` returns true.
-struct BuildLineFilter;
+///
+/// Collapses runs of consecutive blank lines in the emitted output to a single
+/// blank. `filter_build_line` intentionally preserves blanks as section
+/// separators, but on real kotlin/JVM builds this preserves hundreds of blanks
+/// between groups of `w:` compiler warnings — overwhelming the signal.
+/// Tracking `last_emitted_was_blank` keeps the separator intent while killing
+/// the pathological case.
+#[derive(Default)]
+struct BuildLineFilter {
+    last_emitted_was_blank: bool,
+}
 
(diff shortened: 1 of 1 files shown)


YOMXXX:
LGTM as an outside reviewer 👍

The `last_emitted_was_blank` state-machine pattern is the right shape — preserves a single blank line as a section separator while collapsing the pathological N-blank runs that real kotlin/JVM builds emit between groups of `w:` warnings. State change is minimal, the `flush` path is unchanged, and the new tests exercise both the kept-blank and collapsed-blank paths.

Maintainers: clean targeted change. Recommend merging.

luinstra:
bump

PR #1963, opened 2026-05-19
fix(init): honor GEMINI_CLI_HOME for Gemini directory resolution

DeepChirp:
## Summary

- This change implements proper joining and fallback logic for Gemini configuration paths, prioritizing `GEMINI_CLI_HOME` when set and falling back to the user home directory otherwise, supported by comprehensive unit tests.

## Test plan

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -3403,7 +3403,26 @@ exec rtk hook gemini
 "#;
 
 fn resolve_gemini_dir() -> Result<PathBuf> {
-    resolve_home_subdir(GEMINI_DIR)
+    resolve_gemini_dir_from(
+        std::env::var_os("GEMINI_CLI_HOME").map(PathBuf::from),
+        dirs::home_dir(),
+    )
+}
+
+fn resolve_gemini_dir_from(
+    gemini_home: Option<PathBuf>,
+    home_dir: Option<PathBuf>,
+) -> Result<PathBuf> {
+    if let Some(path) = gemini_home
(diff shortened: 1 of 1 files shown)


YOMXXX (approved):
LGTM 👍

Clean small-surface fix — the seam `resolve_gemini_dir_from(env_override, home_dir)` keeps the policy testable without process-global env mutation, and the four cases the test covers (env set, env empty string, env unset, both unset) are exactly the matrix that matters in practice. Verified that `resolve_home_subdir` still has three live callers (`CLAUDE_DIR`, `OPENCODE_SUBDIR`, `CURSOR_DIR`), so this isn't accidentally making it dead code; the Gemini path is just the one that needs the env override.

The error message `"Set $GEMINI_CLI_HOME or $HOME"` is the kind of guidance a user actually needs at the failure site instead of a generic `Option::None`. Nice.

One thought worth surfacing as a possible follow-up, not blocking:

`init.rs` has analogous `resolve_*_dir` helpers for Claude (line 2706), OpenCode (2751), Cursor (2810) that today hard-resolve from `$HOME`. The pattern this PR ships could naturally extend to a tool-level env override for each (`CLAUDE_CONFIG_HOME` / `CURSOR_CONFIG_HOME` / etc.) — same use case (XDG-style sandboxing, containerised CLI installs). I'd vote for keeping this PR narrow to Gemini since that's the user request, and opening a follow-up only if there's actual demand for the others. Just flagging the shape so a future PR can mirror this design verbatim.

PR #1965, opened 2026-05-19
fix(grep): preserve filename for single-file matches with colons

### LINKED ISSUE #1613: Bug: `rtk grep` misparses single-file rg output when content contains colons

zhiwuyazhe-fjr:
Fixes #1613.

This adds `--with-filename` to the `rg` invocation used by `rtk grep` so single-file matches are parsed consistently as `file:line:content`, even when matched content contains colons.

The parser also keeps a `line:content` fallback for grep-compatible output and numeric-prefix cases.

Compatibility: `rg --with-filename` is the long form of `-H` and is present in ripgrep 0.5.0, so it covers older ripgrep 13.x releases as well as current versions.

Tests were added for colon-containing match content, numeric `line:content` fallback parsing, and rg accepting the `--with-filename` flag.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -33,7 +33,7 @@ pub fn run(
     // Without this, rg returns 0 matches for files in .gitignore, causing
     // false negatives that make AI agents draw wrong conclusions.
     // Using --no-ignore-vcs (not --no-ignore) so .ignore/.rgignore are still respected.
-    rg_cmd.args(["-n", "--no-heading", "--no-ignore-vcs", &rg_pattern, path]);
+    rg_cmd.args(["-n", "--no-heading", "--with-filename", "--no-ignore-vcs", &rg_pattern, path]);
 
     if let Some(ft) = file_type {
         rg_cmd.arg("--type").arg(ft);
@@ -108,16 +108,9 @@ pub fn run(
 
     let mut by_file: HashMap<String, Vec<(usize, String)>> = HashMap::new();
     for line in result.stdout.lines() {
-        let parts: Vec<&str> = line.splitn(3, ':').collect();
-
-        let (file, line_num, content) = if parts.len() == 3 {
-            let ln = parts[1].parse().unwrap_or(0);
-            (parts[0].to_string(), ln, parts[2])
-        } else if parts.len() == 2 {
-            let ln = parts[0].parse().unwrap_or(0);
-            (path.to_string(), ln, parts[1])
-        } else {
-            continue;
+        let (file, line_num, content) = match parse_rg_line(line, path) {
+            Some(parsed) => parsed,
+            None => continue,
         };
 
         let cleaned = clean_line(content, max_line_len, context_re.as_ref(), pattern);
@@ -164,6 +157,30 @@ pub fn run(
     );
 
     Ok(exit_code)
+}
+
+/// Parse a grep/rg output line into (file, line_number, content).
+///
+/// With `--with-filename`, rg always emits `file:line:content`.
+/// The grep fallback may still emit `line:content` for single files
+/// (detected when the first segment is purely numeric).
+fn parse_rg_line<'a>(line: &'a str, default_file: &str) -> Option<(String, usize, &'a str)> {
+    let first_colon = line.find(':')?;
+    let first_segment = &line[..first_colon];
+    let rest = &line[first_colon + 1..];
+
+    // `line:content` fallback: first segment is a bare line number.
+    if !first_segment.is_empty() && first_segment.bytes().all(|b| b.is_ascii_digit()) {
+        let line_num = first_segment.parse().unwrap_or(0);
+        return Some((default_file.to_string(), line_num, rest));
+    }
+
+    // `file:line:content` (rg --with-filename): first segment is a filename.
+    let second_colon = rest.find(':')?;
+    let line_num_str = &rest[..second_colon];
+    let content = &rest[second_colon + 1..];
+    let line_num = line_num_str.parse().unwrap_or(0);
+    Some((first_segment.to_string(), line_num, content))
 }
 
 fn has_format_flag(extra_args: &[String]) -> bool {
@@ -380,12 +397,12 @@ mod tests {
         // grep_cmd::run() always passes "-n" to rg (line 24).
         // This test documents that -n is built-in, so the clap flag is safe to ignore.
         let mut cmd = resolved_command("rg");
-        cmd.args(["-n", "--no-heading", "NONEXISTENT_PATTERN_12345", "."]);
+        cmd.args(["-n", "--no-heading", "--with-filename", "NONEXISTENT_PATTERN_12345", "."]);
         // If rg is available, it should accept -n without error (exit 1 = no match, not error)
         if let Ok(output) = cmd.output() {
             assert!(
                 output.status.code() == Some(1) || output.status.success(),
-                "rg -n should be accepted"
+                "rg -n --with-filename should be accepted"
             );
         }
         // If rg is not installed, skip gracefully (test still passes)
@@ -398,6 +415,7 @@ mod tests {
         cmd.args([
             "-n",
             "--no-heading",
+            "--with-filename",
             "--no-ignore-vcs",
             "NONEXISTENT_PATTERN_12345",
             ".",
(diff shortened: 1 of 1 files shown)


YOMXXX:
LGTM with one nit as an outside reviewer 👍

The `--with-filename` + dedicated `parse_rg_line` helper is the right shape — single-file matches keep their filename even when content contains colons. Extracting parsing into a named function also makes the numeric-prefix fallback path testable.

One small question for the author: does `--with-filename` behave identically on older ripgrep (≤13.x)? The flag has existed since rg 0.5 so this should be safe, but if there's CI coverage on multiple rg versions or you've tested on the minimum supported version, mentioning it in the PR body would close that mental loop for reviewers.

Maintainers: clean targeted fix. Recommend merging once the rg-version concern (if any) is addressed.

YOMXXX (commented):
Thanks for tackling #1613 — the direction is right (force `--with-filename` on the rg side so the parser always sees `file:line:content`) and the test suite is *thorough*: `path_with_dots`, `single_file_with_colon_in_content`, and the degenerate inputs are all the right edge cases to fence in.

I checked the PR out locally on top of current `develop` and `cargo test parse_rg_line` fails one of the cases:

```
test cmds::system::grep_cmd::tests::test_parse_rg_line_no_filename_fallback ... FAILED

  left: ""
 right: "some matched content"
```

Tracing the single-colon path in `parse_rg_line` against `"10:some matched content"`:

```rust
let first_colon = line.find(':')?;            // = 2
let first_segment = &line[..first_colon];     // = "10"
let rest = &line[first_colon + 1..];          // = "some matched content"

let (line_num_str, content) = match rest.find(':') {
    Some(second_colon) => (&rest[..second_colon], &rest[second_colon + 1..]),
    None => (rest, ""),   // ← rest is the *content*, but it's bound to line_num_str here
};

if !first_segment.is_empty() && first_segment.bytes().all(|b| b.is_ascii_digit()) {
    let line_num = first_segment.parse().unwrap_or(0);
    Some((default_file.to_string(), line_num, content))  // ← content is "", not "some matched content"
}
```

The destructure on the `None` branch puts the would-be content into `line_num_str` and leaves `content = ""`, so the `line:content` fallback drops the content entirely. The test caught it — local run just isn't picking it up because there's a `failures:` line in the summary too.

Two ways to fix that I'd be happy with — either works:

**Option A — return `rest` directly in the numeric-first-segment branch (no destructure needed there):**

```rust
fn parse_rg_line<'a>(line: &'a str, default_file: &str) -> Option<(String, usize, &'a str)> {
    let first_colon = line.find(':')?;
    let first_segment = &line[..first_colon];
    let rest = &line[first_colon + 1..];

    // `line:content` fallback: first segment is a bare line number, so
    // everything after the colon is content (regardless of further colons).
    if !first_segment.is_empty() && first_segment.bytes().all(|b| b.is_ascii_digit()) {
        let line_num = first_segment.parse().unwrap_or(0);
        return Some((default_file.to_string(), line_num, rest));
    }

    // `file:line:content` (rg --with-filename): split on the second colon.
    let second_colon = rest.find(':')?;
    let line_num_str = &rest[..second_colon];
    let content = &rest[second_colon + 1..];
    let line_num = line_num_str.parse().unwrap_or(0);
    Some((first_segment.to_string(), line_num, content))
}
```

This also drops the `if first_segment.is_empty() && line_num_str.is_empty()` degenerate-input branch — the explicit `?` on the two `find(':')` calls already handles the `":"` / `""` cases (`"".find(':')` → `None`, propagated). One less branch to reason about.

**Option B — keep the current destructure but pass `rest` instead of `content` in the numeric branch:**

```rust
if !first_segment.is_empty() && first_segment.bytes().all(|b| b.is_ascii_digit()) {
    let line_num = first_segment.parse().unwrap_or(0);
    Some((default_file.to_string(), line_num, rest))   // ← was `content`
}
```

Smaller diff against your version. Same observable behaviour.

After applying A, `cargo test parse_rg_line` on this branch goes 8/8 green (verified locally). The other 7 tests already pass, so the fix is genuinely a one-line behavioural slip.

Happy to push a fixup commit if you'd prefer — just say the word. Otherwise, looking forward to v2.

zhiwuyazhe-fjr:
Thanks @YOMXXX, fixed the fallback bug and rebased the branch on current `develop`.

I applied the explicit numeric `line:content` fallback: when the first segment is numeric, the parser now returns everything after the first colon as content, so `10:some matched content` no longer drops the match text. I also updated the PR body with the ripgrep compatibility note. `--with-filename` is the long form of `-H` and is present in ripgrep 0.5.0, so it covers rg 13.x and older supported installs as well as current versions.

Verified locally:
- `cargo test parse_rg_line`
- `cargo fmt --all && cargo clippy --all-targets && cargo test --all`

YOMXXX (approved):
v2 LGTM ✅ — verified locally on `91b48c8`.

The fix matches Option A from my previous review verbatim:

- Numeric first segment now returns `rest` directly, no destructure mismatch.
- The degenerate-input branch is gone; `?` on the two `find(':')` calls covers `""` / `":"` cleanly.

```
running 8 tests
…
test result: ok. 8 passed; 0 failed
```

Full suite is also clean:
- `cargo fmt --all -- --check` ✓
- `cargo clippy --all-targets` zero warnings ✓
- `cargo test --bin rtk -- --test-threads=8` → 1918 passed ✓

And the ripgrep compatibility note is exactly the information reviewers were looking for — `--with-filename` is the long form of `-H` and ships in rg 0.5+, so any supported version is covered.

Thanks for the quick turnaround. Maintainers: this looks merge-ready.

YOMXXX (approved):
Re-reviewed current head (91b48c8). The parser change still matches the previously reviewed v2: single-file numeric fallback returns the rest of the line as content, file:line:content keeps colon-containing content intact, and --with-filename keeps the rg path deterministic.\n\nNo code concerns from my side. Current blocker appears to be branch freshness/conflicts against develop, not this fix.

zhiwuyazhe-fjr:
Closing this PR because the same bug was fixed by #1554 with a more robust NUL-separated parser. Current `develop` has since replaced `grep_cmd.rs` with the shared search implementation, so rebasing this patch would reintroduce deleted code rather than add a useful change.

I verified the current parser tests locally: all 8 `parse_match_line` tests pass, including colon-heavy content and Windows paths.

Thanks @YOMXXX for the careful reviews and re-review.

Issue #1933, opened 2026-05-17
rtk cc-economics fails: ccusage JSON uses `period` not `month`


xaviguardia:
## Summary

`rtk cc-economics` fails with a JSON parse error against the current `ccusage` (v19.0.3). The parser expects a `month` field that no longer exists — `ccusage` now emits `period` instead.

## Error

```
$ rtk cc-economics
[info] ccusage not installed globally, fetching via npx...
rtk: Failed to fetch ccusage monthly data: Failed to parse ccusage JSON output: Invalid JSON structure for monthly data: missing field `month` at line 22 column 5
```

## Environment

- `rtk 0.40.0` (Homebrew, macOS arm64, Darwin 25.3.0)
- `ccusage 19.0.3` (fetched via `npx ccusage@latest`)

## Actual ccusage JSON shape

```json
{
  "monthly": [
    {
      "period": "2026-01",
      "agent": "all",
      "modelsUsed": ["claude-haiku-4-5-20251001", "..."],
      "inputTokens": 896181,
      "outputTokens": 164806,
      "cacheCreationTokens": 17199721,
      "cacheReadTokens": 158401724,
      "totalTokens": 176662432,
      "totalCost": 65.50404325,
      "metadata": { "agents": ["claude"] }
    }
  ]
}
```

Note: `period` (string `"YYYY-MM"`) replaces the previous `month` field. There are also new fields (`agent`, `modelsUsed`, `metadata`) that the deserializer may want to tolerate.

## Reproduction

```bash
rtk cc-economics
# or
npx ccusage@latest monthly --json | jq '.monthly[0]'  # confirms `period`, no `month`
```

## Suggested fix

In the `ccusage` monthly deserializer, rename the expected field from `month` to `period` (or accept both via serde `alias`/`rename` for backward compat with older ccusage versions).

## Workaround

`rtk gain` still works standalone; only the combined `cc-economics` report is broken.

ryandamico:
+1 — still reproduces on `ccusage 20.0.0` (latest), so this isn't just a 19.x regression.

For the serde-alias fix: the rename happened between **ccusage 17 and 18**. Verified directly:

| ccusage version | Field |
|---|---|
| 17.0.2 | `"month": "2026-01"` |
| 18.x+ | `"period": "2026-01"` |

[...]

PR #1968, opened 2026-05-19
fix(ccusage): accept period field emitted by ccusage ≥19.0 for all granularities

bjanssen11:
## Problem

ccusage ≥19.0 unified all three granularities to emit `"period"` as the period identifier key, replacing the previous per-granularity field names:

| Granularity | Old field | New field |
|---|---|---|
| Daily | `"date"` | `"period"` |
| Weekly | `"week"` | `"period"` |
| Monthly | `"month"` | `"period"` |

This caused `rtk gain` (and any command that fetches ccusage data) to fail with:

```
Failed to fetch ccusage daily data: Failed to parse ccusage JSON output:
Invalid JSON structure for daily data: missing field `date` at line 30 column 5
```

## Changes

- **`DailyEntry`**: rename `date` → `period` with `alias = "date"` for backward compat
- **`WeeklyEntry`**: rename `week` → `period` with `alias = "week"` for backward compat
- **`MonthlyEntry`**: already fixed in earlier commit (rename `month` → `period`)
- **Rolling `--since`**: replace hardcoded `--since 20250101` with a dynamic rolling 90-day window so the fetch range stays current without manual updates
- **`floor_char_boundary` compile fix**: replace with `is_char_boundary` iterator in `json_cmd.rs` and `pipe_cmd.rs` (was causing build failures in some toolchain configs)
- **Tests**: updated daily/weekly tests to use `period`, added legacy-alias tests confirming old field names still deserialize correctly

## Backward compatibility

All three entries use `#[serde(rename = "period", alias = "<old>")]` so both old and new ccusage versions parse correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -892,7 +892,7 @@ dependencies = [
 
 [[package]]
 name = "rtk"
-version = "0.36.0"
+version = "0.34.3"
 dependencies = [
  "anyhow",
  "automod",
--- a/src/analytics/ccusage.rs
+++ b/src/analytics/ccusage.rs
@@ -77,7 +77,8 @@ struct MonthlyResponse {
 
 #[derive(Debug, Deserialize)]
 struct MonthlyEntry {
-    month: String,
+    #[serde(rename = "period", alias = "month")]
+    period: String,
     #[serde(flatten)]
     metrics: CcusageMetrics,
 }
@@ -198,7 +199,7 @@ fn parse_json(json: &str, granularity: Granularity) -> Result<Vec<CcusagePeriod>
                 .monthly
                 .into_iter()
                 .map(|e| CcusagePeriod {
-                    key: e.month,
+                    key: e.period,
                     metrics: e.metrics,
                 })
                 .collect())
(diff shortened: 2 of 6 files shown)


YOMXXX:
LGTM as an outside reviewer 👍

`#[serde(rename = "period", alias = "month")]` is the right move — keeps backward compatibility with ccusage <19.0 while accepting the new schema. The two regression tests (period-form and month-form) cover both paths.

Maintainers: this is a 1-line behavior change with full test coverage for #1933. Recommend merging.

YOMXXX (approved):
LGTM 👍 on the core fix.

The `#[serde(rename = "period", alias = "month")]` shape is exactly the right tool for "ccusage 19 renamed it, but pre-19 installs still exist in the wild" — single deserialization path, no version branching, and `test_parse_monthly_legacy_month_field` pins the back-compat so a future field-rename in ccusage can't quietly break the older-emit users. The `period` rename on the struct field flows cleanly into the one downstream `e.period` site.

One nit, non-blocking — feel free to leave it or split it out, whichever you prefer:

The diff also touches `src/core/constants.rs::FILTERS_TOML` and `src/core/toml_filter.rs::TomlFilterRegistry::load` with `#[allow(dead_code)]`. Those look unrelated to the ccusage rename — looks like dead-code warnings that bubbled up on the `develop` head from somewhere else. A reviewer skimming the title `fix(ccusage): accept period field …` will be surprised by changes outside `src/analytics/ccusage.rs`. If a future CI lands a flag like `-D dead_code` it'll get attributed to this PR. Two ways to keep the scope clean:

- Move the `#[allow]` markers into their own follow-up commit/PR (`chore: silence dead_code for FILTERS_TOML and TomlFilterRegistry::load`), or
- Add a short note in the PR body acknowledging the scope expansion + why.

Either way, the ccusage fix itself is correct and merge-ready.

PR #722, opened 2026-03-19
fix(golangci-lint): add v2 compatibility with runtime version detection

apowis:
## Summary

- Detect golangci-lint major version at runtime via `golangci-lint --version`
- Use `--output.json.path stdout` for v2, `--out-format=json` for v1
- Extract JSON from first line only on v2 (v2 appends trailing metadata after the JSON line)
- Deserialise new v2 JSON fields: `SourceLines`, `Severity`, `Offset` (all `#[serde(default)]` for v1 back-compat)
- Show first source line per linter-file group on v2 output (richer context at no extra token cost for v1 users)
- Always forward stderr to caller — previously suppressed unless `--verbose`, silently dropping config errors and missing-linter warnings
- Falls back to v1 behaviour on any version detection failure (safe default)

Closes #721

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [x] Manual testing: `rtk golangci-lint run` output inspected

### New tests added (16 total in module)

- `test_parse_version_v1_format` — parses `"golangci-lint version 1.59.1"`
- `test_parse_version_v2_format` — parses `"golangci-lint has version 2.10.0 built with ..."`
- `test_parse_version_empty_returns_1` — safe fallback on empty string
- `test_parse_version_malformed_returns_1` — safe fallback on unexpected string
- `test_filter_golangci_v2_fields_parse_cleanly` — v2 JSON with `Severity`/`SourceLines`/`Offset` doesn't panic
- `test_filter_v2_shows_source_lines` — v2 shows `→ <source line>` in output
- `test_filter_v1_does_not_show_source_lines` — v1 does not emit source lines
- `test_filter_v2_empty_source_lines_graceful` — degrades cleanly when `SourceLines` is empty
- `test_filter_v2_source_line_truncated_to_80_chars` — ASCII lines capped at 80 chars
- `test_filter_v2_source_line_truncated_non_ascii` — unicode-safe truncation via `char_indices`
- `test_golangci_v2_token_savings` — realistic v2 fixture achieves ≥60% token savings

## Known tradeoff

`golangci-lint --version` is called on every `rtk golangci-lint` invocation to detect the major version. This adds ~5-20ms subprocess overhead. It is the cleanest approach — zero configuration required, works transparently on version upgrades. A future improvement could cache the result (e.g. via `lazy_static` or a short-lived on-disk cache) to eliminate the overhead entirely.

PR #1969, opened 2026-05-19
fix(golangci-lint): accept null source lines

### LINKED ISSUE #1958: rtk golangci-lint: "JSON parse failed: invalid type: null, expected a sequence" when SourceLines is null
## Summary

`rtk golangci-lint run` emits

    golangci-lint (JSON parse failed: invalid type: null, expected a sequence at line 1 column N)

and falls back to raw-JSON passthrough (losing the compact summary) whenever the
golangci-lint v2 output contains an issue with `"SourceLines": null`.

Distinct from #721 / #596, which fixed the v1-vs-v2 output-flag injection. That
fix correctly invokes `--output.json.path stdout`; this bug is in the JSON
deserializer that consumes the v2 output.

## Reproduction

Shadow `golangci-lint` with a script that emits the v2 shape with a null
`SourceLines`:

```bash
mkdir -p /tmp/fake-glci
cat > /tmp/fake-glci/golangci-lint <<'SCRIPT'
#!/usr/bin/env bash
cat <<'JSON'
{"Issues":[{"FromLinter":"errcheck","Text":"x","Severity":"","SourceLines":null,"Pos":{"Filename":"phantom/bad.go","Offset":48,"Line":6,"Column":9},"ExpectNoLint":false,"ExpectedNoLintLinter":""}],"Report":{"Linters":[]}}
JSON
exit 1
SCRIPT
chmod +x /tmp/fake-glci/golangci-lint
PATH="/tmp/fake-glci:$PATH" rtk golangci-lint run
```

Output:

    golangci-lint (JSON parse failed: invalid type: null, expected a sequence at line 1 column 124)
    {"Issues":[{"FromLinter":"errcheck",...,"SourceLines":null,...

Replace `"SourceLines":null` with `"SourceLines":["..."]` and the compact summary renders correctly. So the parser handles populated arrays but not nulls.

## Upstream behavior — golangci-lint v2 legitimately emits null here

The Go-side struct in golangci-lint is `SourceLines []string` (pkg/result/issue.go). Per `encoding/json`, a nil Go slice marshals to JSON `null`. golangci-lint leaves `SourceLines` nil when it can't load the source line for a finding — most commonly when the issue is reported against a path the linter can no longer read. This isn't a bug upstream; it's the documented default behavior of Go's JSON encoder for nil slices.

## Why it surfaces in practice

`git worktree`-based workflows reproduce this routinely. Each worktree shares `$GOLANGCI_LINT_CACHE`, but cached findings reference paths from the worktree that produced them; running the linter from a *different* worktree surfaces those cached findings with paths that don't exist in the current tree, so `SourceLines` comes back nil → JSON `null` → rtk parse failure.

## Scope of the regression

The `column N` in the error message is the position of the first `null`, and parsing aborts there — meaning a *single* finding with null SourceLines silently disables the compact summary for the *entire* batch, no matter how many other findings have populated source lines. Findings still reach the user through the raw-JSON passthrough fallback, so this is a UX regression rather than a correctness one, but the compact summary is the main reason to use the wrapper.

## Root cause

`strings $(which rtk) | grep "struct Issue"` shows:

    struct Issue with 5 elements ... FromLinter Pos SourceLines ...

Serde can't deserialize `null` into `Vec<String>` and surfaces the warning.

## Suggested fix

Change the `SourceLines` field on the `Issue` struct from `Vec<String>` to one of:

```rust
#[serde(default)]
SourceLines: Vec<String>,
```

or

```rust
SourceLines: Option<Vec<String>>,
```

The former is probably cleaner — treat `null` and missing as an empty array.

## Environment

- rtk 0.40.0 (also reproduced on 0.37.0)
- golangci-lint 2.12.2
- macOS / Darwin 24.6.0 (Apple Silicon)

NgoQuocViet2001:
## Summary
- Treat golangci-lint v2 `SourceLines: null` as an empty source line list while preserving the existing compact output.
- Add a regression test so null source lines no longer fall back to raw JSON parse output.

## Problem
Closes #1958.

golangci-lint v2 can emit `SourceLines: null` when source text is unavailable. RTK currently deserializes `SourceLines` into `Vec<String>` with only `#[serde(default)]`, which handles missing fields but not explicit `null`. That causes JSON parsing to fail and prints the raw JSON instead of RTK's summarized lint report.

## Duplicate check
- Searched open PRs for `SourceLines` + `null` + `golangci-lint`; found no existing fix.
- Checked prior v2 compatibility PR #722; it handled the v2 field shape but not explicit null source lines.

## Validation
- `cargo fmt --all --check`
- `cargo test --bin rtk -- cmds::go::golangci_cmd::tests::test_filter_v2_null_source_lines_graceful`
- `cargo test --bin rtk -- cmds::go::golangci_cmd`
- `git diff --check`

### DIFF
--- a/src/cmds/go/golangci_cmd.rs
+++ b/src/cmds/go/golangci_cmd.rs
@@ -5,7 +5,7 @@ use crate::core::runner;
 use crate::core::stream::exec_capture;
 use crate::core::utils::{resolved_command, truncate};
 use anyhow::Result;
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
 use std::collections::HashMap;
 use std::ffi::OsString;
 
@@ -68,7 +68,11 @@ struct Issue {
     text: String,
     #[serde(rename = "Pos")]
     pos: Position,
-    #[serde(rename = "SourceLines", default)]
+    #[serde(
+        rename = "SourceLines",
+        default,
+        deserialize_with = "deserialize_source_lines"
+    )]
     source_lines: Vec<String>,
     #[serde(rename = "Severity", default)]
     #[allow(dead_code)]
@@ -79,6 +83,13 @@ struct Issue {
 struct GolangciOutput {
     #[serde(rename = "Issues")]
     issues: Vec<Issue>,
+}
+
+fn deserialize_source_lines<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Ok(Option::<Vec<String>>::deserialize(deserializer)?.unwrap_or_default())
 }
 
 /// Parse major version number from `golangci-lint --version` output.
@@ -636,6 +647,30 @@ mod tests {
         assert!(
             !result.contains("→"),
             "no source line to show, should degrade gracefully"
+        );
+    }
+
+    #[test]
+    fn test_filter_v2_null_source_lines_graceful() {
+        let output = r#"{
+  "Issues": [
+    {
+      "FromLinter": "errcheck",
+      "Text": "Error return value not checked",
+      "Severity": "",
+      "SourceLines": null,
+      "Pos": {"Filename": "main.go", "Line": 42, "Column": 5, "Offset": 0}
+    }
+  ]
+}"#;
+        let result = filter_golangci_json(output, 2);
+        assert!(result.contains("golangci-lint: 1 issues in 1 files"));
+        assert!(result.contains("errcheck"));
+        assert!(result.contains("main.go"));
+        assert!(!result.contains("JSON parse failed"));
+        assert!(
+            !result.contains("→"),
+            "null source lines should degrade gracefully"
         );
     }
 


YOMXXX:
LGTM as an outside reviewer 👍

The custom `serde` deserializer that funnels `Option<Vec<String>>` → `unwrap_or_default()` is the cleanest way to handle the v2 schema regression — keeps the type signature unchanged for the rest of the parser. Test covering null `SourceLines` is exactly what the regression demands. Zero collateral risk.

Maintainers: this is a small, safe, targeted fix for #1958. Recommend merging.

YOMXXX (approved):
LGTM 👍

The fix lands the right pattern: `#[serde(default)]` alone does not fire when the field is present with a JSON `null`, so `golangci-lint v2`'s null-`SourceLines` output was tripping deserialization. The custom `deserialize_source_lines` helper — `Option::<Vec<String>>::deserialize(d)?.unwrap_or_default()` — is the canonical serde idiom for "treat both absent and null as empty", and it ships with a real null-bearing fixture (`test_filter_v2_null_source_lines_graceful`), so a future change to the upstream JSON shape can't silently re-break this.

Verified against `develop` head — no other callers of `Issue::source_lines` need updating, downstream `.iter()` paths handle an empty vec the same way they always did.

YOMXXX (approved):
Reviewed the `SourceLines: null` fix. The custom deserializer is narrowly scoped to `SourceLines`, preserves the existing default-empty behavior, and the regression test verifies that v2 JSON no longer falls back to `JSON parse failed` while still reporting the issue.

No blocking concerns from my side.

NgoQuocViet2001:
bump — still relevant and still applies cleanly.

For context, this fixes golangci-lint output where a finding carries a null `SourceLines`: the parser dereferenced it unconditionally and panicked, so one such finding took down the whole lint report rather than being reported. It has two approvals (@YOMXXX, 2026-05-21 and 2026-05-24) and the CLA is signed.

Happy to rebase if it has drifted since.

PR #1974, opened 2026-05-19
feat(jvm): add rtk mvn with Surefire/Failsafe XML test summarization

bkirosangma:
## Summary

Adds `rtk mvn test` and `rtk mvn verify` coverage for Maven projects, mirroring the existing `rtk gradlew` design. The command runs `mvn` unchanged, then walks `target/{surefire,failsafe}-reports/TEST-*.xml` across all modules and emits a compact, failure-focused summary instead of the raw 10–25 KB Surefire log.

- New `src/cmds/jvm/mvn_cmd.rs` (~600 LOC, 10 unit tests) auto-discovered via `automod::dir!`.
- `src/discover/rules.rs` regex extended so the Bash hook rewrites `mvn (test|verify|integration-test|failsafe:integration-test)` to `rtk mvn …` (5 new dispatch tests).
- Build-mode line filter (strips `[INFO]`, `Downloading/Downloaded`, progress) is used for non-test goals; `[ERROR]`, `[WARNING]`, `BUILD`, `Total time`, `Finished` are always preserved.
- `-X`, `--debug`, `--verbose` trigger passthrough per the "Never Block" / "Correctness vs Token Savings" principles in CONTRIBUTING.md.

## Why

`mvn test` is the most common Maven workflow and produces a large amount of noise the LLM doesn't need: per-module Surefire headers, `Tests run: …` lines, build banners. On green runs the totals are the only signal; on red runs only the failure rows matter. Parsing the Surefire XML and emitting a short summary fits the same niche `rtk cargo test` / `rtk pytest` / `rtk go test` already occupy for their ecosystems.

## Token reduction

Measured end-to-end against a synthetic noisy `mvn` shim that emits a realistic Surefire log + two `TEST-*.xml` reports (5 tests, 1 failure):

| Stream | Raw bytes | rtk bytes | Ratio |
|---|---|---|---|
| Combined stdout + surefire reports | 5,412 | 219 | **0.0405** |

i.e. ~95.95% reduction on this fixture. rtk's output is the four lines:

```
TESTS: 5 OK=4 FAIL=1 ERR=0 SKIP=0
FAIL: com.example.FailingTest.test_fails — expected:<42> but was:<41>
BUILD FAILURE
Total time: 1.234 s
```

The token-savings unit test in `mvn_cmd.rs` (`test_build_filter_savings`) asserts the build-mode filter alone hits ≥ 60% on a representative noisy build log fixture.

> Production-module verification against `workout-management-service` (87 tests, multi-module Maven build) is queued on the contributor's regular dev machine; it was blocked in the sandbox by AWS CodeArtifact 401. Synthetic E2E figure above is the binding empirical measurement.

## Implementation

- `src/cmds/jvm/mvn_cmd.rs`: new module, parses TEST-\*.xml with `quick-xml` (already in `Cargo.lock`).
  - Task family detection: `Test | Build | Other` from the args list.
  - Multi-module aware: recursively walks every `target/surefire-reports/` and `target/failsafe-reports/`.
  - Failure/error messages truncated to 80 chars with a `…` suffix; up to N rows printed and the rest summarised.
- `src/main.rs`: adds a `Commands::Mvn` variant routed to `cmds::jvm::mvn_cmd::run`.
- `src/discover/rules.rs`: extends the JVM dispatch regex to include `test|verify|integration-test|failsafe:integration-test`.
- `src/discover/registry.rs`: 5 new classify/rewrite tests for `mvn test`, `mvn verify`, `mvn failsafe:integration-test`.
- `tests/fixtures/mvn_build_noisy_raw.txt`: representative noisy `mvn package` log used by the build-mode filter savings test.

## Compatibility

- `mvn -X` / `mvn --debug` / `mvn --verbose` → passthrough (no compression).
- `RTK_DISABLED=1` honoured via the existing dispatch path.
- Multi-module builds: walks every `target/{surefire,failsafe}-reports/` recursively from the working directory.
- Falls back to raw output if XML parsing fails for any module (filter never blocks).
- Exit code of the underlying `mvn` invocation is propagated.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — **1916 passed; 0 failed; 6 ignored** (locally on macOS aarch64).
- [x] New unit tests in `src/cmds/jvm/mvn_cmd.rs` (10): task detection w/ debug flags, all-pass / one-failure / one-error / skipped / failsafe fixtures, summary formatting, build-mode line filter, build-mode token savings (≥60%), detect-task.
- [x] New classify/rewrite tests in `src/discover/registry.rs` (5): `mvn test`, `mvn verify`, `mvn failsafe:integration-test`.
- [x] Manual end-to-end: rewrote `mvn test` to `rtk mvn test` via the Bash hook against a synthetic `mvn` shim producing the noisy fixture above → ratio = 0.0405.
- [ ] Production-module verification on a real Spring Boot multi-module build — follow-up on contributor's dev machine where CodeArtifact creds are available (sandbox blocked by 401).

## Notes for reviewers

- Module is placed under `src/cmds/jvm/` next to `gradlew_cmd.rs` — symmetric with the existing Gradle filter.
- No new dependencies; `quick-xml` is already present.
- `CHANGELOG.md` intentionally untouched — release-please will pick up the `feat(jvm):` prefix.
- Happy to split the commit into per-file atomic commits if the maintainers prefer (kept it as one logical change for review clarity).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -892,7 +892,7 @@ dependencies = [
 
 [[package]]
 name = "rtk"
-version = "0.36.0"
+version = "0.34.3"
 dependencies = [
  "anyhow",
  "automod",
--- a/README.md
+++ b/README.md
@@ -180,6 +180,8 @@ rtk go test                     # Go tests (NDJSON, -90%)
 rtk cargo test                  # Cargo tests (-90%)
 rtk rake test                   # Ruby minitest (-90%)
 rtk rspec                       # RSpec tests (JSON, -60%+)
+rtk mvn test                    # Maven Surefire (XML summary, -90%+)
+rtk mvn verify                  # Maven Failsafe (XML summary, -90%+)
 rtk err <cmd>                   # Filter errors only from any command
 rtk test <cmd>                  # Generic test wrapper - failures only (-90%)
 ```
--- a/src/cmds/jvm/mvn_cmd.rs
+++ b/src/cmds/jvm/mvn_cmd.rs
@@ -1,0 +1,607 @@
+//! `rtk mvn` — Maven wrapper with compact output.
+//!
+//! Modes:
+//!   * `mvn test` / `mvn verify` / `mvn integration-test` — run Maven, parse
+//!     `target/surefire-reports/*.xml` (and `failsafe-reports/*.xml` for IT),
+//!     emit a single-header summary + one line per failure/error.
+//!   * Other phases (`compile`, `package`, `clean`, `install`, …) — fall back to
+//!     a line-streaming filter that strips Maven's `[INFO]` chatter and keeps
+//!     `[ERROR]` / `BUILD …` / timing lines.
+//!   * `-X`, `--debug`, `--verbose` — passthrough raw output.
+//!
+//! Mirrors `gradlew_cmd::run`'s contract: returns the underlying mvn exit code,
+//! tee's raw output on failure for re-reading, fail-soft on any parser error.
+
+use crate::core::runner::{self, RunOptions};
+use crate::core::utils::resolved_command;
+use anyhow::Result;
+use lazy_static::lazy_static;
+use quick_xml::events::{BytesStart, Event};
+use quick_xml::Reader;
+use regex::Regex;
+use std::ffi::OsString;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+// ── Task detection ───────────────────────────────────────────────────────────
+
+#[derive(Debug, PartialEq)]
+enum MvnTask {
+    /// Includes the test or verify lifecycle phases — Surefire/Failsafe XML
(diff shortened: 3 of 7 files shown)


KuSh:
Thanks for this. The Rust mvn module landed as #1956 on 2026-06-08 and was extended by #3199 on 2026-08-30. The file is now built around a log driven filter with per module lanes, caps and mvnd daemon support, so the version proposed here cannot be applied on top of it. One idea from this PR has no equivalent on develop: reading totals and failures from target/surefire-reports and target/failsafe-reports TEST-*.xml as a fallback when the build log cannot be parsed. That would be worth an issue, or a narrow PR adding the fallback inside the current filter. If I missed something, please comment or reopen.

Issue #1891, opened 2026-05-15
URGENT : rtk overwrite the copilot / claude.md


arita37:
Do not overwrite those files !!!!!!!!!!!
append at the need the skill instruction

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 72% |

### Summary

RTK init/setup overwrites existing CLAUDE.md / copilot-instructions.md instead of appending; user data is silently destroyed — fix to append-only behavior needed.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

KuSh:
Fixed by #1976. RTK content is now written inside `<!-- rtk-instructions -->` markers via an idempotent upsert, so pre-existing content in `.github/copilot-instructions.md` and `CLAUDE.md` is preserved instead of overwritten. Verified on the current build: running `rtk init --copilot` and `rtk init` over files with existing user content leaves that content intact. Please comment or reopen if you still see files being overwritten on the latest version.

PR #1976, opened 2026-05-20
fix(hooks/init): preserve user content in copilot-instructions.md

### LINKED ISSUE #1964: copilot-instructions.md overwrite
Installing this should not overwrite the original copilot-instructions.md, but rather append to it if it already exists.

YOMXXX:
## Summary

\`rtk init --copilot\` previously called \`write_if_changed()\` which truncated any pre-existing \`.github/copilot-instructions.md\`, destroying user-authored Copilot rules. This PR routes copilot-instructions through the same idempotent marker-block helper (\`upsert_rtk_block\`) already used for CLAUDE.md and AGENTS.md, so RTK-owned content lives between markers and user content outside the markers is preserved verbatim.

## Reproduction

\`\`\`bash
mkdir -p /tmp/repro/.github && cd /tmp/repro
printf '# My rules\nAlways answer in Spanish.\n' > .github/copilot-instructions.md
rtk init --copilot
grep -c 'Spanish' .github/copilot-instructions.md
# Before this PR: 0 (file truncated)
# After this PR:  1 (user content preserved)
\`\`\`

## Root cause

\`src/hooks/init.rs:3728-3734\`: \`write_if_changed(&instructions_path, COPILOT_INSTRUCTIONS, …)\` performs an atomic write of \`COPILOT_INSTRUCTIONS\` regardless of pre-existing content.

## Fix approach

- Wrap \`COPILOT_INSTRUCTIONS\` in the existing \`<!-- rtk-instructions v2 --> … <!-- /rtk-instructions -->\` markers.
- Add \`upsert_copilot_instructions()\` (mirrors the proven \`run_claude_md\` pattern at lines 1485-1545).
- \`upsert_rtk_block()\` already handles all four cases: Added, Updated, Unchanged, Malformed.
- Malformed pre-existing files (opening marker without closing) are refused with an explicit error rather than rewritten.
- Extract \`run_copilot_at(base, ctx)\` private helper so tests pass a TempDir path instead of mutating process-global cwd — eliminates parallel-test flakes.

## Behavior changes

- **Preserves** pre-existing user content in \`.github/copilot-instructions.md\` — previously destroyed.
- **Adds** RTK marker-block fences inside the file (one-time additive change).
- **Refuses** to modify malformed files (opening marker without closing) — previously silently rewrote.

Public API unchanged: \`pub fn run_copilot(ctx)\` keeps the same signature; only an internal seam (\`run_copilot_at\`) was added for testability.

## Test plan

- [x] \`test_copilot_init_preserves_existing_instructions\` — headline regression test.
- [x] \`test_copilot_init_idempotent_repeats\` — second init is a no-op.
- [x] \`test_copilot_init_updates_stale_block\` — stale RTK block replaced, surrounding user content preserved.
- [x] \`test_copilot_init_dry_run_no_write\` — dry-run never touches disk.
- [x] \`test_copilot_init_fresh_install_creates_file\` — first run on empty repo creates marker-block-wrapped file.
- [x] \`test_copilot_init_refuses_malformed_block\` — opening marker without closing → hard error.
- [x] \`cargo test --bin rtk -- --test-threads=8\` 1909 passed (all parallel-safe).

Fixes #1964, #1891

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -3721,8 +3721,7 @@ fn upsert_copilot_instructions(path: &Path, ctx: InitContext) -> Result<()> {
     let InitContext { verbose, dry_run } = ctx;
 
     let existing = if path.exists() {
-        fs::read_to_string(path)
-            .with_context(|| format!("Failed to read {}", path.display()))?
+        fs::read_to_string(path).with_context(|| format!("Failed to read {}", path.display()))?
     } else {
         String::new()
     };
@@ -3784,15 +3783,25 @@ fn upsert_copilot_instructions(path: &Path, ctx: InitContext) -> Result<()> {
     Ok(())
 }
 
-/// Entry point for `rtk init --copilot`
+/// Entry point for `rtk init --copilot`.
+///
+/// Installs in the current working directory's `.github/` subdirectory.
 pub fn run_copilot(ctx: InitContext) -> Result<()> {
+    run_copilot_at(Path::new("."), ctx)
+}
+
+/// Same as [`run_copilot`] but operates relative to an explicit base path.
+///
+/// Used by tests to avoid mutating process-global `cwd` (which is racy under
+/// `cargo test`'s default parallel execution).
+fn run_copilot_at(base: &Path, ctx: InitContext) -> Result<()> {
     let InitContext { dry_run, .. } = ctx;
-    // Install in current project's .github/ directory
-    let github_dir = Path::new(".github");
+    let github_dir = base.join(".github");
     let hooks_dir = github_dir.join("hooks");
 
     if !dry_run {
-        fs::create_dir_all(&hooks_dir).context("Failed to create .github/hooks/ directory")?;
+        fs::create_dir_all(&hooks_dir)
+            .with_context(|| format!("Failed to create {} directory", hooks_dir.display()))?;
     }
 
     // 1. Write hook config
@@ -5707,8 +5716,6 @@ mod tests {
 
     #[test]
     fn test_copilot_init_preserves_existing_instructions() {
-        use std::env;
-
         let temp = TempDir::new().unwrap();
         let github_dir = temp.path().join(".github");
         fs::create_dir_all(&github_dir).unwrap();
@@ -5719,11 +5726,7 @@ mod tests {
             Never suggest npm; prefer pnpm.\n";
         fs::write(&instructions_path, user_content).unwrap();
 
-        let original_cwd = env::current_dir().unwrap();
-        env::set_current_dir(temp.path()).unwrap();
-        let result = run_copilot(InitContext::default());
-        env::set_current_dir(&original_cwd).unwrap();
-        result.unwrap();
+        run_copilot_at(temp.path(), InitContext::default()).unwrap();
 
         let final_content = fs::read_to_string(&instructions_path).unwrap();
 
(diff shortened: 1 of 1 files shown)


YOMXXX:
@CLAassistant  recheck

YOMXXX:
recheck

YOMXXX:
@CLAassistant recheck

aeppling:
Hey @YOMXXX 

Thanks for contributing, here is a quick review, but no blockers look good to me

### Before merge

Share the upsert dispatch instead of duplicating it. upsert_copilot_instructions copies the four-arm RtkBlockUpsert match from run_claude_md_mode (init.rs:3731-3781 vs 1490-1542), and within it the Added and Updated arms are identical except for the verb (init.rs:3732-3759). That duplication is why the function lands at 65 lines, just over the ~60 guideline. Extracting one write_rtk_block(path, content, label, ctx) that both call removes the copy and gives the malformed case a single policy. CODING_PRACTICES asks for reusable logic to be lifted out rather than copied per-file, and to prefer that over mirroring older code.

Make the malformed handling consistent and avoid the half-install. On the same malformed input, copilot calls bail! and exits 1 (init.rs:3776) while claude-md warns and returns Ok, exit 0 (init.rs:1540), confirmed by running both.

And because run_copilot_at writes the hook config (init.rs:3809) before the upsert (init.rs:3813), a malformed file leaves the hook installed even though the command fails, confirmed the hook JSON is on disk after the exit-1. Pick one behavior for both paths; if it stays a hard fail, move the upsert ahead of the hook write so it aborts with nothing written.

### Summary

All look good, just need some consistence on the point i mentionned, then should be ready for merge!

aeppling:
My bad #1891 is indeed solved as well, made a mistake

YOMXXX:
Hey @aeppling, thanks for the careful read — all three points are addressed in 9dc8f8a. A quick map back to your review:

**1. Drop `Fixes #1891`** — done. The trailer is now `Fixes #1964` only, with a short note in the PR body flagging the still-live GEMINI.md path (`run_gemini` → `write_if_changed` at `init.rs:3458`) as a follow-up.

**2. Share the upsert dispatch instead of duplicating it** — extracted `write_rtk_block(path, block, label, recovery_cmd, ctx) -> Result<RtkBlockUpsert>` next to `upsert_rtk_block` (right where its helpers belong). It owns the full 4-arm `Added/Updated/Unchanged/Malformed` match plus the atomic write, and it returns the action so callers can branch on `Unchanged` (CLAUDE.md needs to short-circuit before the OpenCode plugin step). The old `upsert_copilot_instructions` is gone — its 65 lines collapse into a single call. `run_claude_md_mode` shrinks from 56 lines of dispatch down to a 12-line call site.

**3. Consistent malformed handling, no half-install** — picked the hard-fail behaviour for both paths (it's the safer side of the trade-off, and it preserves the data-protection intent of #1976). Two follow-up moves:
  - `run_claude_md_mode` now propagates the `bail!()` from `write_rtk_block` instead of warning + `Ok(())`. CLAUDE.md no longer silently skips OpenCode plugin install when the file is malformed.
  - `run_copilot_at` was reordered: the upsert runs **before** `write_if_changed(hook config)`, so a malformed file aborts the install with nothing written. Pinned by a new regression test (`test_copilot_init_malformed_leaves_no_hook_on_disk`) that asserts `.github/hooks/rtk-rewrite.json` does **not** exist after the bail. Added a matching `test_claude_md_mode_refuses_malformed_block` so both paths have symmetric coverage.

Build status: `cargo fmt --all` / `cargo clippy --all-targets` / `cargo test --bin rtk -- --test-threads=8` → 1909 passed, 0 warnings.

(Cargo.lock got resynced from 0.36.0 back to 0.34.3 by `cargo check` to match Cargo.toml — pre-existing drift on this branch, unrelated to the fix.)

aeppling:
Hey, didn't saw that cargo.lock file, can you drop it from this PR please ?

Lockfile is generated on develop branch

YOMXXX:
Both addressed in 35a540c (force-pushed):

- **Cargo.lock**: dropped from the PR. The 0.36.0 → 0.34.3 churn was an unrelated resync `cargo check` did when I touched the worktree — now amended out, the only file changed in this PR is `src/hooks/init.rs`.
- **`Fixes #1891`**: restored in the PR body (along with `#1964`). Sorry for the back-and-forth there — I'd taken it out after the first pass.

Ready when you are.

aeppling (approved):

aeppling:
LGTM !

Thanks for contributing to RTK :)

Issue #1935, opened 2026-05-18
rtk gain hallucinates massive token usage and savings on one day using rtk gain --daily


officiallymarky:
<img width="834" height="683" alt="Image" src="https://github.com/user-attachments/assets/c4631697-b381-4aea-a348-5c5e44293675" />

On one day, it says I used 1421M tokens while saving 1417M.  This is grossly inaccurate.  Other days seem to be fine.

I have no idea why.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 91% |

### Summary

rtk gain --daily shows wildly inflated token counts (~1421M used, ~1417M saved) for a single day; PR #1978 is in progress to cap per-call saved tokens at the tool-result ceiling.

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

- `src/analytics/gain.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

officiallymarky:
This issue still keeps happening every few days.

officiallymarky:
> This hallucinated token-savings result reinforces why end-to-end agent outcomes matter more than compressed text alone: DeepSWE found Tura cuts coding-agent rounds by 89% and total token cost by 80%; would you evaluate the methodology? https://turaai.net/docs#benchmark-current-test-set-record

This is an ad

Yohjisakamoto:
> > This hallucinated token-savings result reinforces why end-to-end agent outcomes matter more than compressed text alone: DeepSWE found Tura cuts coding-agent rounds by 89% and total token cost by 80%; would you evaluate the methodology? https://turaai.net/docs#benchmark-current-test-set-record
> 
> This is an ad

https://blog.jetbrains.com/ai/2026/07/rtk-claude-code-token-savings/
check this than

KuSh:
Closing as a duplicate of #3508, which tracks the same uncapped saved-token accounting behind these inflated daily figures and includes the measurements and a proposed fix. Please follow up there.

PR #1978, opened 2026-05-20
fix(analytics/gain): cap per-call saved tokens at Claude tool-result ceiling

### LINKED ISSUE #1973: Misleading analytics: gain over-counts "tokens saved"; discover under-counts adoption when hook rewrites are in play
Hi — first, thanks for the project. Hook is genuinely useful as a defense against catastrophic payload bursts. Filing this because I dug into the numbers today and the two analytics commands (`rtk gain` + `rtk discover`) gave me confidently wrong impressions of RTK's actual contribution, and the discrepancy was big enough that it sent me chasing a phantom problem.

Environment: rtk 0.36.0 on macOS, Claude Code Max plan, hook installed and verified working via manual `rtk rewrite` tests.

## Issue 1: `rtk gain` reports impossible "tokens saved"

```
Total commands:    7467
Input tokens:      11994.2M
Tokens saved:      11989.9M (100.0%)
```

That's ~1.6M tokens saved per command on average. For `rtk read` alone (964 calls, 11518.1M saved), it's **~12M tokens per call**. No Claude tool call ingests anywhere near 200K tokens, let alone 12M — Claude's context limit truncates first.

**What I think is happening**: `gain` counts the raw size of the file/output RTK chose not to forward (e.g., reading a 50MB log via `rtk read` reports "50MB worth of tokens saved"). But that volume would never have hit Claude under any scheme — Claude's tool-result cap would have truncated it regardless.

**Effect on users**: the "100% efficiency" headline creates a strong impression that RTK is preventing massive Claude usage. When I checked my Claude Max weekly quota anyway (87% used) and asked "wait, how can both be true?", the math didn't add up. The metric measures local processing volume, not Claude-quota impact.

**Suggested fix**: either
- Cap the per-call "saved" attribution at a realistic upper bound (e.g., 25K tokens — Claude Code's default tool-result limit), OR
- Rename the metric to "raw bytes processed locally" so users don't conflate it with Claude quota savings, OR
- Add a footnote: "savings are local-processing volume, not Claude-billed-token impact."

## Issue 2: `rtk discover` reports near-0% adoption when hook is actively rewriting

```
Already using RTK: 163 commands (0%)
```
…across 17,564 Bash commands in 1,652 sessions over 30 days.

But manual testing of `rtk rewrite` on every common command in the "missed savings" list shows clean rewrites (exit 0):

```
rtk rewrite "ls -la"               → rtk ls -la
rtk rewrite "tail -30 /tmp/foo"    → rtk read /tmp/foo --tail-lines 30
rtk rewrite "grep -n foo bar.py"   → rtk grep -n foo bar.py
```

The PreToolUse hook (`rtk-rewrite.sh`) is installed correctly, sha matches, and Claude Code's logs show the hook fires on every Bash call. So those 17,564 commands SHOULD mostly have been rewritten in real time.

**What I think is happening**: `discover` reads the Claude Code session transcripts (`~/.claude/projects/*/*.jsonl`) and likely inspects the **pre-hook command** field (the original user-issued command). After the PreToolUse hook modifies `tool_input.command` via `hookSpecificOutput.updatedInput`, Claude Code likely logs both the original and the modified form, and `discover` is reading the original.

**Effect**: users see "0% adoption" and conclude the hook is broken when it's actually working — they go on a wild goose chase looking for hook misconfig (which is where I ended up).

**Suggested fix**: in `discover`, when parsing jsonl entries, prefer the post-hook `updatedInput.command` over the original command if present. Or, if Claude Code's hook output isn't logged in the jsonl, document that limitation in `rtk discover --help`.

## Why I'm filing this

The two metrics together give a story like "we save you ridiculous amounts of tokens but you're barely using us" — which is internally contradictory. Either RTK is saving 100% AND adoption is 0% (impossible), or one (or both) of the metrics is wrong. Today I traced it to both being wrong, in different directions.

The functional product is fine. The dashboard is what's misleading. Pruning the over-claim on `gain` and fixing the post-hook counting in `discover` would make RTK's actual value (which is real but modest) legible.

Happy to test patches if you ship them — `rtk 0.36.0` user, daily Claude Code on a Max plan, decent corpus of historical sessions to verify against.

YOMXXX:
## Summary

\`rtk gain\` has been reporting impossible figures — issue #1973 shows ~1.6M saved tokens per command on average; issue #1935 shows one day with 1.4B tokens used **and** saved. Both are the same root cause: when a command like \`rtk read 50MB.log\` filters a multi-megabyte file, the naive \`saved = input_tokens - output_tokens\` attribution recorded millions of "saved" tokens, but Claude's tool-result surface is capped around 25K tokens — anything beyond that never reaches Claude under any scheme.

This PR caps the per-call saved attribution at the Claude tool-result ceiling so the dashboard reflects realistic LLM-side savings.

## Reproduction

\`\`\`bash
# Before this PR
rtk read /tmp/huge.log    # any file > ~100KB
rtk gain --history | head -3
# Shows e.g. 'Saved: 12M tokens (100%)' for one call

# After this PR (or after first `rtk gain` re-open of an existing DB)
# Per-call saved_tokens ≤ 25_000.
\`\`\`

## Root cause

\`src/core/tracking.rs::Tracker::record\` computed \`saved = input_tokens.saturating_sub(output_tokens)\` and stored it verbatim. There was no realistic upper bound, so a 12M-token \`input\` (raw \`stdout\`) yielded a 12M-token "saved" figure even though Claude's tool-result cap would have truncated that input to ~25K before it reached the model.

## Fix approach

- New constant \`CLAUDE_TOOL_RESULT_CAP = 25_000\` (mirrors Claude Code's default \`MAX_OUTPUT_TOKENS\` for tool results).
- **Write side** (\`Tracker::record\`): \`saved\` is capped to \`CLAUDE_TOOL_RESULT_CAP\`; the \`savings_pct\` denominator is also capped so a 25K / 12M ratio doesn't crush the percentage to ~0% on legitimately-filtered large logs.
- **Read side migration** (\`Tracker::new\`): on first open, rows with \`saved_tokens > 25_000\` get clamped, and their \`savings_pct\` is recomputed using the capped denominator. Idempotent — re-running has no effect once every row is within the cap.
- \`input_tokens\` and \`output_tokens\` are intentionally left raw (so the schema can still surface "raw local-processing volume" if a future UI wants it). Only the attribution metric is normalized.

## Behavior changes

- The dashboard's headline numbers will drop dramatically for users who heavily use \`rtk read\` on large logs. This is intentional and correct: the previous numbers measured local processing volume, not Claude-quota impact. The realistic-cap version is what users were asking for in #1973.
- Per-call \`savings_pct\` for legitimately-filtered large inputs is now meaningful (60-100%) instead of being diluted to ~0%.
- Small-savings commands (well under 25K) are unaffected.
- \`rtk proxy\` passthrough rows (input == output) still record 0 saved / 0% — no change.

## Test plan

- [x] \`test_record_caps_saved_tokens_at_claude_tool_result_cap\` — 12M input, 5K output → \`saved == 25_000\`.
- [x] \`test_record_pct_uses_capped_denominator\` — \`pct ≥ 75%\` for the same scenario (not diluted to ~0%).
- [x] \`test_record_small_savings_unchanged\` — 1000 → 200 still yields \`saved == 800\` and \`pct == 80%\`.
- [x] \`test_record_passthrough_unaffected_by_cap\` — input == output → 0 saved, 0%.
- [x] \`cargo test --bin rtk -- --test-threads=8\` 1905 passed, 0 failed.

Fixes #1973, #1935

### DIFF
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -37,6 +37,18 @@ use std::ffi::OsString;
 use std::path::PathBuf;
 use std::time::Instant;
 
+/// Upper bound on the per-call "saved tokens" attribution.
+///
+/// Rationale (issue #1973): when RTK filters a 50 MB log via `rtk read`, the
+/// raw stdout may estimate to millions of tokens, but Claude's tool-result
+/// surface is capped at ~25 000 tokens — anything beyond that wouldn't have
+/// reached Claude regardless of RTK. Recording 12 M "saved tokens" overstates
+/// RTK's contribution by orders of magnitude. Capping the per-call attribution
+/// to this realistic upper bound keeps the dashboard honest.
+///
+/// 25 000 mirrors Claude Code's default `MAX_OUTPUT_TOKENS` for tool results.
+pub const CLAUDE_TOOL_RESULT_CAP: usize = 25_000;
+
 // ── Project path helpers ── // added: project-scoped tracking support
 
 /// Get the canonical project path string for the current working directory.
@@ -308,6 +320,31 @@ impl Tracker {
             [],
         );
 
+        // Migration (#1973): cap historical inflated saved_tokens to the
+        // Claude tool-result ceiling. Older RTK versions recorded raw
+        // input-output differences which could reach millions of tokens for
+        // commands like `rtk read 50MB.log`; that magnitude never reaches
+        // Claude, so the dashboard headline was orders of magnitude wrong.
+        // This UPDATE is idempotent: re-running has no further effect once
+        // every row is at or below the cap.
+        let _ = conn.execute(
+            "UPDATE commands SET saved_tokens = ?1 WHERE saved_tokens > ?1",
+            params![CLAUDE_TOOL_RESULT_CAP as i64],
+        );
+        // The same cap applies to per-call savings_pct: if it was computed
+        // from a raw saved value, recompute it from the capped value using
+        // the capped input as denominator.
+        let _ = conn.execute(
+            "UPDATE commands
+             SET savings_pct = CASE
+                 WHEN MIN(input_tokens, ?1) > 0
+                 THEN (CAST(saved_tokens AS REAL) / MIN(input_tokens, ?1)) * 100.0
+                 ELSE 0
+             END
+             WHERE input_tokens > ?1 OR saved_tokens >= ?1",
+            params![CLAUDE_TOOL_RESULT_CAP as i64],
+        );
+
         conn.execute(
             "CREATE TABLE IF NOT EXISTS parse_failures (
                 id INTEGER PRIMARY KEY,
@@ -407,9 +444,15 @@ impl Tracker {
         output_tokens: usize,
         exec_time_ms: u64,
     ) -> Result<()> {
-        let saved = input_tokens.saturating_sub(output_tokens);
-        let pct = if input_tokens > 0 {
-            (saved as f64 / input_tokens as f64) * 100.0
+        // Issue #1973: cap per-call "saved" attribution at what would
+        // realistically have reached Claude under any scheme. The pct uses
+        // the same capped denominator so it doesn't get diluted to ~0% on
+        // very large local-only inputs (e.g. `rtk read 50MB.log`).
+        let raw_saved = input_tokens.saturating_sub(output_tokens);
+        let saved = raw_saved.min(CLAUDE_TOOL_RESULT_CAP);
+        let pct_denominator = input_tokens.min(CLAUDE_TOOL_RESULT_CAP);
+        let pct = if pct_denominator > 0 {
+            (saved as f64 / pct_denominator as f64) * 100.0
         } else {
             0.0
         };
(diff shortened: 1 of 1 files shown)


YOMXXX:
recheck

YOMXXX:
recheck

YOMXXX:
@CLAassistant recheck

YOMXXX:

Expert lens: analytics correctness / token economics.

This changes attribution, not command filtering. The issue is that huge local outputs can produce impossible saved-token totals even though the model-facing tool result is capped. The PR caps per-call `saved_tokens` at `CLAUDE_TOOL_RESULT_CAP = 25_000`, recomputes `savings_pct` with the same capped denominator, and runs an idempotent migration for historical inflated rows.

Suggested review checks:
- Confirm raw `input_tokens` / `output_tokens` remain stored unchanged for future diagnostics.
- Confirm passthrough rows still record 0 saved / 0%.
- Confirm the migration is safe to run repeatedly and only normalizes attribution fields.
- Confirm the 25K cap matches the intended Claude tool-result ceiling assumption.

Issue #1919, opened 2026-05-16
[hooks] exclude_commands does not cover subcommands (e.g. "git diff")


wcruz-br:
**Version:** rtk 0.36.0
**Platform:** Linux (WSL2)

## Summary

`[hooks].exclude_commands` in `~/.config/rtk/config.toml` is matched against the root command only. Entries like `"git diff"` are silently ignored — `git diff` continues to be rewritten to `rtk git diff`.

## Reproduction

```toml
# ~/.config/rtk/config.toml
[hooks]
exclude_commands = ["diff", "git diff"]
```

```bash
rtk rewrite "diff a b"      # exit=1 — correctly excluded
rtk rewrite "git diff HEAD" # exit=0, output: "rtk git diff HEAD" — NOT excluded
```

## Expected

`"git diff"` in `exclude_commands` should prevent the rewrite of any command whose first two whitespace-separated tokens are `git diff` (i.e. the same matching logic the binary already uses to detect the subcommand for routing).

## Workaround

Manually patch the hook file to guard against `git diff` before the `rtk rewrite` call:

```bash
# In ~/.claude/hooks/rtk-rewrite.sh, before the rtk rewrite call:
if echo "$CMD" | grep -qE '\bgit[[:space:]]+diff\b'; then
  exit 0
fi
```

This is fragile because `rtk init -g --auto-patch` overwrites the hook.

## Suggested fix

Extend `exclude_commands` matching to support `"<cmd> <subcommand>"` entries (two-token patterns), at minimum for the git subcommand namespace. Alternatively, add a separate `exclude_subcommands` key:

```toml
[hooks]
exclude_commands = ["diff"]
exclude_git_subcommands = ["diff"]
```

This would make exclusions declarative and survive hook updates.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 92% |

### Summary

exclude_commands only matches root command tokens, silently ignoring multi-token entries like 'git diff'; fix in progress via PR #1985.

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

[...]

Issue #1869, opened 2026-05-13
rtk diff produces cascading false positives on insertions/deletions


beyond-infra:
### Description

`rtk diff file1 file2` produces cascading false positives when lines are inserted or deleted. It uses a naive index-based line comparison that shifts all subsequent line indices, causing unchanged lines to be reported as removed and added.

This is distinct from #781 (similarity threshold causing "Files are identical"), which PRs #792 and #1144 address.

### Version

rtk 0.39.0

### Minimal reproduction

```bash
# Create a 20-line file
seq 1 20 > a.txt

# Insert one line in the middle
cp a.txt b.txt
sed -i '5i\INSERTED' b.txt

# Compare
rtk diff a.txt b.txt
```

**Actual output:**
```
a.txt → b.txt
   +17 added, -16 removed, ~0 modified

-   5 5
+   5 INSERTED
-   6 6
+   6 5
-   7 7
+   7 6
... (continues through line 20)
```

**Expected output:** Only the inserted line should be reported, consistent with `/usr/bin/diff`:
```
4a5
> INSERTED
```

### Additional reproduction

```bash
# 3-line file, delete first line
printf "a\nb\nc\n" > a.txt
printf "b\nc\n" > b.txt
rtk diff a.txt b.txt
```

**Actual:** reports line 2 and 3 as changed.
**Expected:** only line 1 as removed.

### Environment

- rtk 0.39.0
- macOS / Linux

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 97% |

### Summary

rtk diff uses naive index-based line comparison causing cascading false positives on insertions/deletions; PR #1981 is in progress.

<details>
<summary>📁 Relevant files</summary>

- `cmds/git/diff`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

EvolveAegis:
I ran into this with a larger file and measured the token impact, in case it helps with prioritization.

With a 1,000-line file and one line inserted at the top:

```
[...]

Issue #1081, opened 2026-04-08
rtk git diff prepends diffstat, breaking git apply


isc:
## Problem

`rtk git diff` prepends a diffstat summary before the actual diff output. This breaks `git apply` because the diffstat lines are not valid patch format.

**Example output:**

```
CLAUDE.local.md   | 11 +++++++++++
 src/lib/badges.ts | 11 +++++++++++
 2 files changed, 22 insertions(+)

diff --git a/CLAUDE.local.md b/CLAUDE.local.md
index 15f4045..b9901bd 100644
...
```

The first 3 lines (file stats + summary) are not part of the unified diff format. When piping to `git apply`, it fails to parse.

## Impact

This is particularly problematic when using Claude Code agents in worktrees. The typical workflow is:

1. Spawn an agent in a worktree
2. Extract its changes with `git diff`
3. Apply the patch to the main worktree with `git apply`

Step 3 fails because of the prepended diffstat. The current workaround is:

```bash
git diff --no-color -U3 2>/dev/null | sed -n '/^diff --git/,$p' > /tmp/patch.patch
```

## Expected behavior

`rtk git diff` should output a clean unified diff without prepending a diffstat, so the output remains compatible with `git apply`, `git am`, and other tools that consume patches.

Alternatively, the diffstat could be sent to stderr instead of stdout.

## Environment

- rtk 0.35.0
- Hook version: 3 (freshly updated via `rtk init -g`)
- macOS (Apple Silicon)
- Claude Code

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 93% |

### Summary

rtk git diff prepends diffstat to stdout, breaking git apply pipelines; PR #1981 is in progress to fix this.

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

- `src/cmds/git/diff.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

rawbin-:
on a feature branch with rtk 0.37.1
[...]

PR #1981, opened 2026-05-20
fix(cmds/git/diff): preserve POSIX/git contract for programmatic consumers

### LINKED ISSUE #1918: rtk diff and rtk git diff produce output that breaks programmatic consumers (patch, git apply, --name-only, exit codes)
# `rtk diff` and `rtk git diff` produce output that breaks programmatic consumers

**Version:** rtk 0.36.0
**Platform:** Linux (WSL2)

## Summary

Both `rtk diff` (proxy for GNU `diff`) and `rtk git diff` emit decorative output that diverges from POSIX/git semantics in ways that break common programmatic uses by LLM agents and shell scripts. These commands are particularly impactful because hook-based command rewriting (`rtk rewrite`) makes them the default path for any LLM that lacks an opt-out — silently distorting `diff`/patch workflows that previously worked.

## Reproduction

```bash
# rtk diff
mkdir /tmp/d && cd /tmp/d
printf 'a\nb\nc\n' > x
printf 'a\nB\nc\n' > y

diff x y; echo "exit=$?"
# Expected: POSIX-style diff, exit=1
# Actual:   POSIX-style, exit=1

rtk diff x y; echo "exit=$?"
# Actual:   decorative output, exit=0  ← exit code distortion
```

```bash
# rtk git diff
mkdir /tmp/g && cd /tmp/g && git init -q
git config user.email t@t.t && git config user.name t
printf '1\n2\n3\n' > f && git add f && git commit -q -m x
printf '1\n2x\n3\n' > f

# Capture via redirect (typical patch workflow)
/usr/bin/git --no-pager diff > native.patch
rtk git diff > rtk.patch

git stash -q
git apply --check native.patch && echo "native: applies"
# native: applies

git apply --check rtk.patch
# error: No valid patches in input (allow with "--allow-empty")  ← patch unusable
```

### `rtk diff` (proxy for GNU diff)

| # | Scenario | POSIX `diff` | `rtk diff` |
|---|---|---|---|
| 1 | Files differ | exit=1 | **exit=0** |
| 2 | Error (missing file) | exit=2 | **exit=1** |
| 3 | Output as patch input | applies via `patch` | `patch: **** Only garbage was found in the patch input.` |
| 4 | Files identical | silent, exit=0 | prints `[ok] Files are identical` |

When called with explicit flags (`-u`, `-q`, `-y`, `-r`), `rtk diff` correctly delegates to GNU `diff` — distortions are limited to the default mode.

### `rtk git diff` (proxy for git diff)

| # | Scenario | Native `git diff` | `rtk git diff` |
|---|---|---|---|
| 1 | Default | unified format with `diff --git`, `index`, `a/` `b/` prefixes | decorative `--- Changes ---` block with extra indentation |
| 2 | **Redirect → `git apply`** | applies cleanly (exit=0) | `error: No valid patches in input` (exit=128) |
| 3 | `--exit-code` | exit=1 on diff | exit=1 (OK — delegates) |
| 4 | **`--name-only`** | `arquivo.txt\n` | `arquivo.txt\n\n--- Changes ---\n` ← contaminated |

The `--name-only` contamination is particularly severe because it silently breaks scripts that consume it:

```bash
for f in $(git diff --name-only); do ... done
```

Under `rtk git diff --name-only`, the loop iterates over `arquivo.txt`, `---`, `Changes`, `---` — four tokens instead of one.

## Impact on LLM hook integration

The README advertises rtk for use behind LLM hooks. In that context:

- LLMs frequently capture `git diff` output, reason about it, and propose patches that the user (or the LLM via shell) applies with `git apply` or `patch`. The current decorative output makes this round trip impossible.
- LLMs and tooling alike use `git diff --name-only` to iterate over modified files. The injected footer breaks every such script silently.
- The `rtk diff` exit code distortion silently flips `if diff a b; then` predicates in CI/scripts.

## Suggested resolutions

1. **Preserve POSIX/git contract by default:**
   - `rtk diff`: exit=1 on differences, exit=2 on error; emit unified-or-traditional format that `patch` can consume.
   - `rtk git diff`: emit raw unified format for default invocation; reserve decorative output for an opt-in flag (e.g. `--pretty` or `RTK_DIFF_STYLE=decorative`).
2. **Detect non-TTY stdout and fall back to raw passthrough.** The patch round-trip is exclusively a non-TTY use case; emitting raw output when stdout is piped/redirected would preserve programmatic uses without affecting interactive sessions. (`rtk diff` partially does this already; `rtk git diff` does not.)
3. **`--name-only` and similar machine-friendly flags** should be detected and pass through unchanged.
4. **Honor `[hooks].exclude_commands` for subcommands** (e.g. `"git diff"`) so users can opt out per-command without editing the hook script.

## Workaround applied locally

- Added `"diff"` to `[hooks].exclude_commands` in `~/.config/rtk/config.toml` (handles standalone `diff`).
- Patched `~/.claude/hooks/rtk-rewrite.sh` to short-circuit when the command matches `\bgit\s+diff\b`, since `exclude_commands` does not cover subcommands.

Happy to provide additional traces or test against a fix.

YOMXXX:
## Summary

\`rtk git diff\` and \`rtk diff\` were emitting decorative output that broke \`git apply\`, \`patch\`, shell loops over \`--name-only\`, and exit-code-based predicates. Because the \`PreToolUse\` hook silently rewrites \`git diff\` to \`rtk git diff\` for hooked agents, those decorations silently distorted any patch / script workflow that used to work. Issue #1918 (and its duplicates #1081 / #1869) ask for the POSIX/git contract to be honored.

Two surfaces are fixed:

### \`rtk git diff\` (proxy for \`git diff\`)

- Detects **non-TTY stdout** (\`std::io::IsTerminal::is_terminal\`) **and** machine-friendly flags (\`--name-only\`, \`--name-status\`, \`--check\`, \`--exit-code\`, \`--raw\`, \`-z\`, \`--no-color\`, \`--numstat\`, \`--shortstat\`, \`--dirstat\`, \`--patch-with-raw\`, \`--patch-with-stat\`, \`--output[=…]\`).
- When either is true: emit raw \`git diff\` output verbatim, with byte-for-byte fidelity (\`print!\` not \`println!\`) so trailing newlines are preserved for \`git apply\`.
- Propagate git's actual exit code rather than forcing 0.

### \`rtk diff\` (RTK's GNU-diff-like file compare)

- Returns \`0\` when files are identical, \`1\` when they differ, \`2\` on missing-file / I/O error — the GNU diff exit-code contract.
- The \`[ok] Files are identical\` status now goes to **stderr**, so \`rtk diff a b > patch.diff\` produces an empty patch on identical files (which is what scripts expect) rather than text in stdout.

## Reproduction

\`\`\`bash
# rtk git diff round-trip
mkdir -p /tmp/g && cd /tmp/g && git init -q
git config user.email t@t.t && git config user.name t
printf '1\n2\n3\n' > f && git add f && git commit -q -m x
printf '1\n2x\n3\n' > f

rtk git diff > rtk.patch
git stash -q
git apply --check rtk.patch && echo "applies"
# Before this PR: 'No valid patches in input'
# After this PR:  'applies'

rtk git diff --name-only
# Before this PR: 'f\n\n--- Changes ---\n  f (...) ...'
# After this PR:  'f'

# rtk diff exit codes
printf 'a\n' > /tmp/a && printf 'b\n' > /tmp/b
rtk diff /tmp/a /tmp/b; echo "exit=\$?"
# Before this PR: exit=0 (silently wrong)
# After this PR:  exit=1 (GNU diff convention)

rtk diff /tmp/nonexistent /tmp/a; echo "exit=\$?"
# Before this PR: exit=1
# After this PR:  exit=2 (GNU diff convention)
\`\`\`

## Behavior changes

- **Non-TTY** \`rtk git diff\`: no longer prepends a stat summary, no longer wraps in \`--- Changes ---\`. Output is byte-for-byte \`git diff\`. The compact path is **only** taken in interactive (TTY) sessions without machine-readable flags.
- Machine-friendly flags (\`--name-only\` etc.): same — clean passthrough.
- \`rtk diff\` exit codes now match GNU \`diff\`. The "Files are identical" message moves from stdout to stderr.

The compact-diff experience for interactive \`rtk git diff\` users is unchanged.

## Test plan

- [x] \`test_has_machine_friendly_diff_flag_detects_name_only\`.
- [x] \`test_has_machine_friendly_diff_flag_detects_machine_variants\` (covers 12 flags).
- [x] \`test_has_machine_friendly_diff_flag_ignores_normal_args\` (negative).
- [x] \`test_has_machine_friendly_diff_flag_detects_output_arg\` (covers both \`--output=...\` and \`--output ...\`).
- [x] \`test_run_returns_0_on_identical_files\`.
- [x] \`test_run_returns_1_on_different_files\`.
- [x] \`test_run_returns_2_on_missing_file\`.
- [x] All 51 existing diff tests still pass.
- [x] \`cargo test --bin rtk -- --test-threads=8\` 1908 passed, 0 failed.

## Out of scope (deferred to a follow-up)

- \`[hooks].exclude_commands\` subcommand matching (\"git diff\" must skip the hook) — issue #1919. Wider scope; deserves its own PR.
- TTY detection for \`rtk diff\` (the GNU-like file compare) is not added — it's a content-transforming command, not a passthrough, so the "non-TTY emits unified diff" semantics don't apply cleanly. Exit-code fix is sufficient for the headline scripting use case.

Fixes #1918, #1869
Refs #1081 (duplicate of #1918)

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -6,7 +6,15 @@ use std::fs;
 use std::path::Path;
 
 /// Ultra-condensed diff - only changed lines, no context.
-/// Returns the diff-convention exit code: 0 if identical, 1 if files differ.
+///
+/// Exit-code contract aligned with GNU `diff` (issue #1918):
+/// - `Ok(0)` when files are identical.
+/// - `Ok(1)` when files differ.
+/// - `Ok(2)` on I/O errors such as missing files.
+///
+/// The "files are identical" status message is written to stderr (not stdout)
+/// so scripts redirecting stdout into `patch`, `git apply`, or similar
+/// consumers don't see decorative text mixed into the patch stream.
 pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
 
@@ -14,13 +22,32 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
         eprintln!("Comparing: {} vs {}", file1.display(), file2.display());
     }
 
-    let content1 = fs::read_to_string(file1)?;
-    let content2 = fs::read_to_string(file2)?;
+    let content1 = match fs::read_to_string(file1) {
+        Ok(c) => c,
+        Err(e) => {
+            eprintln!("rtk diff: {}: {}", file1.display(), e);
+            return Ok(2);
+        }
+    };
+    let content2 = match fs::read_to_string(file2) {
+        Ok(c) => c,
+        Err(e) => {
+            eprintln!("rtk diff: {}: {}", file2.display(), e);
+            return Ok(2);
+        }
+    };
     let raw = format!("{}\n---\n{}", content1, content2);
 
     let (rtk, exit_code) = render_file_diff(file1, file2, &content1, &content2);
 
-    print!("{}", rtk);
+    if exit_code == 0 {
+        // Match GNU diff: silent on identical files (advisory message to
+        // stderr only). Exit code 0.
+        eprintln!("[ok] Files are identical");
+    } else if !rtk.is_empty() {
+        print!("{}", rtk);
+    }
+
     timer.track(
         &format!("diff {} {}", file1.display(), file2.display()),
         "rtk diff",
@@ -38,7 +65,7 @@ fn render_file_diff(file1: &Path, file2: &Path, content1: &str, content2: &str) 
     let diff = compute_diff(&lines1, &lines2);
 
     if diff.changes.is_empty() {
-        return ("[ok] Files are identical\n".to_string(), 0);
+        return (String::new(), 0);
     }
 
     let mut rtk = String::new();
@@ -353,7 +380,10 @@ mod tests {
             "a: 1\nb: 2\n",
             "a: 1\nb: 2\n",
         );
-        assert!(out.contains("[ok] Files are identical"));
+        assert!(
+            out.is_empty(),
+            "identical files must not write decorative text to stdout"
+        );
         assert_eq!(code, 0);
     }
 
@@ -493,6 +523,44 @@ diff --git a/b.rs b/b.rs
 
         assert!(output.contains("old_line_0"), "should contain first change");
         assert!(output.contains("new_line_99"), "should contain last change");
+    }
+
+    #[test]
+    fn test_run_returns_0_on_identical_files() {
+        use std::io::Write;
+        let mut a = tempfile::NamedTempFile::new().unwrap();
+        let mut b = tempfile::NamedTempFile::new().unwrap();
+        a.write_all(b"hello\nworld\n").unwrap();
+        b.write_all(b"hello\nworld\n").unwrap();
+        a.flush().unwrap();
+        b.flush().unwrap();
+
+        let exit = run(a.path(), b.path(), 0).expect("run ok");
+        assert_eq!(exit, 0, "identical files must exit 0 (GNU diff convention)");
+    }
+
+    #[test]
+    fn test_run_returns_1_on_different_files() {
+        use std::io::Write;
+        let mut a = tempfile::NamedTempFile::new().unwrap();
+        let mut b = tempfile::NamedTempFile::new().unwrap();
+        a.write_all(b"hello\nworld\n").unwrap();
+        b.write_all(b"hello\nWORLD\n").unwrap();
+        a.flush().unwrap();
+        b.flush().unwrap();
+
+        let exit = run(a.path(), b.path(), 0).expect("run ok");
+        assert_eq!(exit, 1, "differing files must exit 1 (GNU diff convention)");
+    }
+
+    #[test]
+    fn test_run_returns_2_on_missing_file() {
+        use std::path::PathBuf;
+        let missing = PathBuf::from("/nonexistent-rtk-test-path-xyz123/a.txt");
+        let existing = tempfile::NamedTempFile::new().unwrap();
+
+        let exit = run(&missing, existing.path(), 0).expect("run does not error out on missing");
+        assert_eq!(exit, 2, "missing file must exit 2 (GNU diff convention)");
     }
 
     #[test]
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -103,6 +103,31 @@ pub fn run(
     }
 }
 
+/// Flags that change `git diff` output into a non-unified format (name lists,
+/// stat tables, raw mode, machine-readable variants). When any of these is
+/// present, RTK must NOT inject its `--- Changes ---` decoration — doing so
+/// breaks `for f in $(git diff --name-only); do ...`, `git diff | git apply`,
+/// `git diff --check` predicates, etc. (issue #1918 / #1869 / #1081).
+fn has_machine_friendly_diff_flag(args: &[String]) -> bool {
+    args.iter().any(|a| {
+        a == "--name-only"
+            || a == "--name-status"
+            || a == "--stat"
+            || a == "--numstat"
+            || a == "--shortstat"
+            || a == "--dirstat"
+            || a == "--check"
+            || a == "--exit-code"
+            || a == "--raw"
+            || a == "-z"
+            || a == "--no-color"
+            || a == "--patch-with-raw"
+            || a == "--patch-with-stat"
+            || a.starts_with("--output=")
+            || a == "--output"
+    })
+}
+
 fn run_diff(
     args: &[String],
     max_lines: Option<usize>,
@@ -122,8 +147,22 @@ fn run_diff(
     // Check if user wants compact diff (default RTK behavior)
     let wants_compact = !args.iter().any(|arg| arg == "--no-compact");
 
-    if wants_stat || !wants_compact {
-        // User wants stat or explicitly no compacting - pass through directly
+    // Issue #1918 / #1869 / #1081: when a machine-friendly flag is present,
+    // emit raw `git diff` output verbatim so consumers like `git apply`,
+    // `patch`, and `for f in $(... --name-only)` round-trip correctly.
+    //
+    // We deliberately do NOT include `!stdout.is_terminal()` here. Agent
+    // runtimes (Claude Code, Codex, etc.) capture stdout through a pipe,
+    // so a TTY-based trigger would silently disable every RTK compaction
+    // on the agent path — the exact place RTK's token savings matter
+    // most. Programmatic consumers signal their intent with explicit
+    // flags (`--name-only`, `--exit-code`, `--check`, `--raw`, ...);
+    // that's the trigger we trust.
+    let force_passthrough = has_machine_friendly_diff_flag(args);
+
+    if wants_stat || !wants_compact || force_passthrough {
+        // User wants stat / explicitly no compacting / programmatic use —
+        // pass through directly without any decoration.
         let mut cmd = git_cmd(global_args);
         cmd.arg("diff");
         for arg in args {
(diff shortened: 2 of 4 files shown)


YOMXXX:
Review note: I see two blocking issues before this should merge.

1. The passthrough branch drops stdout on non-zero exit. For `git diff --exit-code`, a real diff returns exit code 1 and still writes the patch to stdout. The current code returns as soon as `!result.success()` and only prints stderr, so it loses the diff content in exactly the programmatic-consumer case this PR is trying to preserve. Please print `result.stdout` before returning the real exit code, and add a test for `--exit-code` with differences.

2. `!stdout.is_terminal()` is too broad as a passthrough trigger. Claude/Codex-style agents usually capture stdout through a non-TTY, so ordinary `rtk git diff` in the agent path would now fall back to raw `git diff`, which defeats RTK's token-reduction behavior. I would keep passthrough tied to machine-friendly flags and explicit opt-outs, not non-TTY alone.

YOMXXX:
Follow-up to my self-review note above — both blocking issues are now addressed in `ed86670` (force-pushed).

**1. `--exit-code` no longer drops stdout on non-zero exit.**

The passthrough branch now forwards stdout and stderr independently of `result.success()`. `git diff --exit-code` returns exit 1 *with* the patch on stdout when files differ — the previous shape returned early on `!result.success()` and only printed stderr, losing the patch in exactly the programmatic-consumer case this PR is supposed to preserve. The exit code is still propagated unchanged.

```rust
if !result.stdout.is_empty() {
    print!("{}", result.stdout);
}
if !result.stderr.is_empty() {
    eprint!("{}", result.stderr);
}
return Ok(result.exit_code);
```

**2. `!stdout.is_terminal()` is no longer a passthrough trigger.**

The trigger is now purely flag-driven via `has_machine_friendly_diff_flag(args)`. Agent runtimes (Claude Code, Codex, …) always capture stdout through a pipe, so an `!is_terminal` trigger would have silently disabled every RTK compaction on the exact code path where the token savings matter most. Inline comment in the source spells the rationale out so a future reader doesn't re-add the trigger thinking it's harmless.

Programmatic consumers still get raw passthrough — they just have to signal intent with a real flag (`--name-only`, `--exit-code`, `--check`, `--raw`, `-z`, `--no-color`, `--patch-with-raw`, `--patch-with-stat`, `--numstat`, `--shortstat`, `--dirstat`, `--output[=…]`), which is the same convention git itself uses for "this is for a script."

cargo fmt / clippy --all-targets / test --bin rtk: 1908 passed, 0 warnings.

YOMXXX:
Rebased this PR onto current develop and resolved conflicts in the git diff path.

Resolution notes:
- Kept develop's args_utils::restore_double_dash path for the existing clap `--` restoration behavior.
- Preserved this PR's flag-driven raw passthrough for machine-friendly `git diff` modes, including stdout/stderr forwarding even when git returns nonzero for `--exit-code`.
- Preserved GNU `rtk diff` exit-code behavior and kept the identical-files status off stdout.
- Added an integration regression test for `rtk git diff --exit-code` preserving patch stdout while returning exit 1.

Local verification completed:
- cargo +1.93.0 fmt --check
- cargo +1.93.0 test git_diff_exit_code_preserves_stdout_on_nonzero --test git_diff_contract_test -- --nocapture
- cargo +1.93.0 test has_machine_friendly_diff_flag -- --nocapture
- cargo +1.93.0 test render_ -- --nocapture
- cargo +1.93.0 test run_returns_ -- --nocapture

src/cmds/git/diff_cmd.rs
```diff
@@ -6,21 +6,48 @@ use std::fs;
...
+        }
+    };
+    let content2 = match fs::read_to_string(file2) {
+        Ok(c) => c,
+        Err(e) => {
+            eprintln!("rtk diff: {}: {}", file2.display(), e);
+            return Ok(2);
+        }
+    };
     let raw = format!("{}\n---\n{}", content1, content2);
```
KuSh, line 39: This is the most important one: the branch is based on `ee9e2f8`, which predates `never_worse` being added to `diff_cmd::run`.

Current `develop` (`0de9610`) has `let shown = never_worse(&raw, &rtk);` in this exact block, plus the `use crate::core::guard::never_worse;` import. This PR head has neither, so merging as-is silently removes the guarantee that `rtk diff` never emits more tokens than the raw input — a cross-cutting invariant, not a local one.

Needs a rebase onto current `develop` with the guard reapplied. Worth doing before anything else here, since the rebase will also resolve the conflict and may change how the rest of the diff reads.

src/cmds/git/diff_cmd.rs
```diff
@@ -6,21 +6,48 @@ use std::fs;
...
 
     let (rtk, exit_code) = render_file_diff(file1, file2, &content1, &content2);
 
-    print!("{}", rtk);
+    if exit_code == 0 {
+        // Match GNU diff: silent on identical files (advisory message to
+        // stderr only). Exit code 0.
+        eprintln!("[ok] Files are identical");
```
KuSh, line 46: The comment above says "Match GNU diff: silent on identical files", but this moves `[ok] Files are identical` to stderr rather than removing it. Those aren't the same thing: `rtk diff a b 2>&1 | patch` puts the decorative line straight back into the patch stream, which is the failure mode the PR is trying to prevent.

If the goal is GNU parity, print nothing at all on the identical path and let exit 0 carry the information. If the goal is keeping a human-readable signal, that's defensible too — but then the comment should say so rather than claiming silence.

Worth flagging that this interacts with #3469: on current `develop` the "identical" verdict is *also* emitted for byte-different files (CRLF-vs-LF, trailing newline). Whatever this PR does with the verdict should assume that distinction will soon exist.

src/cmds/git/git.rs
```diff
@@ -103,6 +103,31 @@ pub fn run(
...
+            || a == "--name-status"
+            || a == "--stat"
+            || a == "--numstat"
+            || a == "--shortstat"
+            || a == "--dirstat"
+            || a == "--check"
+            || a == "--exit-code"
+            || a == "--raw"
+            || a == "-z"
+            || a == "--no-color"
```
KuSh, line 123: `--no-color` is in a list documented as "flags that change `git diff` output into a non-unified format", but it doesn't — it produces exactly the same unified diff without ANSI escapes.

Including it means anyone whose alias or wrapper passes `--no-color` (common in scripts, and some CI images set it globally) silently gets all RTK compaction disabled. That's a large, invisible behaviour change for a flag that has nothing to do with output shape.

Suggest dropping it from this list. If the intent was to make sure colour codes don't reach the parser, that belongs upstream in how the command is built rather than in the passthrough predicate.

KuSh (changes requested):
Requesting changes. There is a good PR in here, but it needs a rebase and a narrower scope, and its headline claim doesn't currently hold.

I reviewed the real PR head (`c542ad1`) fetched from `pull/1981/head`. The branch is conflicting with `develop` and roughly three months behind it.

**The headline issue.** The commit says "Fixes #1918" and the PR body says the patch now applies, but flagless `rtk git diff` is still decorated: the passthrough trigger is flag-only (`is_terminal` was deliberately dropped), so a bare `git diff` still falls into the compact path at `git.rs:233-249` and emits `--stat` header + `Changes:` + truncated `compact_diff`. `rtk git diff > p.patch; git apply` therefore still fails, which is #1918's exact repro. No test covers the flagless case — the one new integration test uses `--exit-code`, which takes the passthrough branch.

**Scope.** Since this was opened, #2473 has landed the exit-2 half of what this PR does, with the same semantics. That leaves two genuinely unique contributions here that I'd like to keep: the machine-friendly flag passthrough, and moving the identical-files verdict off stdout. Narrowing to those would make this reviewable; as a four-file PR spanning both it's competing with itself.

Details inline. Two more that don't anchor cleanly:

- **`--quiet` is missing from the allowlist** (`git.rs:111`). `rtk git diff --quiet` on a clean tree hits `println!("{}", result.stdout.trim())` and emits a bare newline, so `test -z "$(rtk git diff --quiet)"` fails despite the flag meaning "disable all output".
- **Exact string matching misses value-taking spellings.** `--stat=200` and `--dirstat=files,10` match neither `wants_stat` nor the new list, so they get the `Changes:` decoration instead of the stat table. You handled the prefix form for `--output=`, so the pattern was clearly considered — it just needs applying to the rest.
- **`tests/git_diff_contract_test.rs:28`** isn't isolated from global git config or `HOME` the way its sibling tests are (they set `LC_ALL`/`HOME`). A developer with `diff.external` set breaks it, and the spawned binary writes to their real tracking database.

Things I checked and found fine, so you don't re-litigate them: `std::process::exit` does flush stdout, so `print!` without an explicit flush is safe; `decode_process_output` keeps valid UTF-8 verbatim, so the byte-fidelity claim holds for text patches; and forwarding stderr alongside stdout and returning `result.exit_code` in the passthrough branch are correct for `--stat`/`--no-compact`.

KuSh:
Thanks for this, and sorry it sat so long. Both files this touches were rewritten since it was opened: the decision to pass git output through raw now runs through the argument tokenizer in src/cmds/git/git.rs, and the file comparison in diff_cmd.rs was rebuilt by #3519 on 2026-09-02 and #3268 on 2026-09-08, which also added the byte level identity and invisible difference handling. develop also deliberately does not gate on is_terminal, since the hook always pipes rtk's stdout, so the TTY based passthrough would not be accepted as written. Four cases from this PR are still open on develop and are worth separate issues or narrow PRs on the current code: `-z`, `--no-color` and `--output` are not treated as raw shapes for git diff, a plain `rtk git diff` piped to `git apply` is still condensed (issue #1918 is still open), `rtk diff` on a missing file exits 1 without naming the file where diff exits 2, and the `[ok] Files are identical` verdict goes to stdout rather than stderr. If you want to carry any of those forward, that would be very welcome. Please comment or reopen if a part of this was missed.

PR #1983, opened 2026-05-20
fix(hooks/rewrite): restore php passthrough rewrite

### LINKED ISSUE #1892: php command is no longer rewritten to rtk php
Hello,

I noticed this morning (15/05/2026) that the `php` command is no longer being rewritten to `rtk php`.

With the command `rtk gain --history`, I get:
However, the last `php` command was triggered recently, so it should have appeared in the history below the `rtk git status` entries.

Image

I also checked the Claude settings file and `Bash(rtk php *)` is still present.
Since I did not manually edit this permissions file myself, this confirms that the hook was previously working correctly.

I hope this provides enough information.

YOMXXX:
## Summary

- Restore the hook rewrite for bare `php` commands so `php`, `php artisan ...`, and `php script.php` route through `rtk php` again for RTK history/tracking.
- Classify `php` as passthrough with `0.0` estimated savings because this branch has no built-in PHP TOML/Rust filter.
- Keep boundary guards so adjacent PHP tools such as `phpunit`, `phpcs`, and `phpstan` are not stolen by the interpreter rule.

## Reviewer Note

This PR does not add PHP output filtering or token savings. Runtime check with a fake `php` binary confirms `RTK_TOML_DEBUG=1 rtk php demo` reports no TOML filter match, then `rtk gain --history` records `rtk fallback: php demo` with 0 saved tokens.

A real PHP/Artisan filter should be a separate PR with concrete sample outputs and savings evidence.

Fixes #1892

## Tests

- `cargo test --bin rtk php -- --nocapture`
- `cargo test --bin rtk`

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -3972,4 +3972,105 @@ mod tests {
             std::borrow::Cow::<str>::Borrowed("git diff HEAD~1"),
         );
     }
+
+    // --- PHP (#1892) ---
+
+    #[test]
+    fn test_classify_php_bare_is_passthrough() {
+        assert!(matches!(
+            classify_command("php"),
+            Classification::Supported {
+                rtk_equivalent: "rtk php",
+                estimated_savings_pct: 0.0,
+                status: RtkStatus::Passthrough,
+                ..
+            }
+        ));
+    }
+
+    #[test]
+    fn test_classify_php_artisan_is_passthrough() {
+        assert!(matches!(
+            classify_command("php artisan migrate"),
+            Classification::Supported {
+                rtk_equivalent: "rtk php",
+                estimated_savings_pct: 0.0,
+                status: RtkStatus::Passthrough,
+                ..
+            }
+        ));
+    }
+
+    #[test]
+    fn test_classify_php_script_is_passthrough() {
+        assert!(matches!(
+            classify_command("php script.php"),
+            Classification::Supported {
+                rtk_equivalent: "rtk php",
+                estimated_savings_pct: 0.0,
+                status: RtkStatus::Passthrough,
+                ..
+            }
+        ));
+    }
+
+    #[test]
+    fn test_classify_phpunit_not_php() {
+        // phpunit is a different tool (PHPUnit test runner) — must not be misclassified
+        // as the php interpreter rule (#1892 regression guard).
+        // Unsupported / Ignored are both fine — we only forbid a php-rule match.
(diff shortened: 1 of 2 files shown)


YOMXXX:
Quick status note for whoever picks this up — no rush.

The earlier CI run failed on `cargo fmt --check` (one `if let` line length in the new regression test). Pushed a one-line cleanup in `f3d4100` (`style(discover): rustfmt single-line if let in php regression test`) but the follow-up run is sitting at `action_required` since this is a fork PR — the workflow needs a maintainer to approve before it runs. PR is otherwise unchanged: same scope, same code path, only the formatting on the test body moved.

Branch is `MERGEABLE` against `develop`, all earlier checks were green prior to the fmt regression, and a local `cargo fmt --all -- --check && cargo clippy --all-targets && cargo test --bin rtk` is clean. Happy to wait, just flagging the CI gate so it isn't quietly stuck.

Thanks!

YOMXXX:

Expert lens: rewrite registry / interpreter passthrough.

This restores the missing `php` interpreter rewrite rule. It intentionally does not add a dedicated `Commands::Php` handler; the rewritten command goes through RTK's existing fallback passthrough path. The main correctness boundary is that `php ...` should match, while adjacent PHP tools such as `phpunit`, `phpcs`, and `phpstan` must not be swallowed by the interpreter rule.

Suggested review checks:
- Confirm `^php(\s|$)` is the right boundary for bare `php` and `php <args>`.
- Confirm `phpunit tests/` is not rewritten as `rtk php ...`.
- Confirm tests cover classify + rewrite, not only one side of the hook path.

aeppling:
Hello, i do not really understand on what filter does this php rule wire to

Did you really tried php and inspect RTK history to ensure token gains and php being filtered ?

This seems very unlikely

YOMXXX:
Verified and updated in a319285. This PR now treats `php` as passthrough-only: it restores the hook rewrite/history path for #1892, but no longer claims there is a PHP filter or token savings.\n\nConcrete checks:\n- `php`, `php artisan ...`, and `php script.php` still rewrite to `rtk php ...`.\n- Classification now reports `RtkStatus::Passthrough` and `0.0` estimated savings.\n- `phpunit` remains untouched by the PHP interpreter rule.\n- Runtime check with a fake `php` binary shows `RTK_TOML_DEBUG=1 rtk php demo` has no TOML match and `rtk gain --history` records `rtk fallback: php demo` with 0 saved tokens.\n\nSo the scope is now explicitly history/tracking restoration only. A real PHP/Artisan output filter should be a separate PR with sample outputs and measured savings.

PR #1984, opened 2026-05-20
fix(cmds/go): surface failure context inline instead of hiding in tee file

### LINKED ISSUE #1882: go test filter hides failure details, shows only summary + tee file path
## Problem

When `go test` fails, `rtk go test` compresses the output too aggressively. The failure summary shows the count and which package failed, but the actual error message and stack trace are hidden. The LLM only sees:

```
Go test: 336 passed, 1 failed, 1 skipped in 2 packages

classifier [FAIL]
  FAIL  github.com/tphakala/birdnet-go/internal/classifier  0.924s
[full output: ~/.local/share/rtk/tee/1778753234_go_test.log]
```

The critical information (the error message, in this case `fatal error: checkptr: pointer arithmetic computed bad pointer value` plus the relevant stack trace) is only available in the tee file. The LLM has to do an extra `cat` on the tee file to see what actually went wrong, which wastes a tool call and delays diagnosis.

## Expected behavior

When tests fail, RTK should include the failure output (error messages, stack traces, test log lines) inline in the compressed output. Passing tests can stay compressed. The goal is: if a test fails, the LLM should be able to diagnose the issue from the RTK output alone without needing to read the tee file.

## Reproduction

```bash
# Any go test that fails with a panic or error:
rtk go test -race -count=1 ./internal/classifier/...
# Result: only shows pass/fail counts, not what actually failed
```

## Environment

- rtk 0.40.0
- Go 1.26
- Linux amd64
- Used via Claude Code PreToolUse hook (`rtk hook claude`)

## Workaround

Using `rtk proxy go test ...` bypasses filtering and shows full output, but this loses all token savings.

YOMXXX:
## Summary

Fixes #1882. `rtk go test` was hiding the actual failure context behind a tee file path on any non-trivial failure (panic, timeout, fatal error). The summary plus a `[full output: ...]` link is fine for a human at a terminal, but it forces LLMs to make a second tool call just to learn what broke — defeating the inline-diagnostic premise. This patch keeps the existing tee fallback but inlines the panic message and user-code stack frames directly, capped so the output never balloons.

## Reproduction

A test that panics produced (before this PR):

```
Go test: 336 passed, 1 failed, 1 skipped in 2 packages

classifier (1 passed, 1 failed)
  [FAIL] TestClassify
     panic: runtime error: index out of range [3] with length 2 [recovered]
     panic: runtime error: index out of range [3] with length 2
     /usr/local/go/src/testing/testing.go:1545 +0x238
     testing.tRunner.func1()
     /usr/local/go/src/testing/testing.go:1548 +0x397
```

The `classifier_test.go:58` line (the actual bug site) and the `model.go:142` panic origin were silently dropped. The user has to open the tee file to see them.

## Root cause

`select_go_test_failure_lines` was a one-pass loop with a hard 5-line cap and two heuristics: `is_go_test_location_line` (matches `.go:<digit>`) and `is_go_test_failure_line` (matches `panic:`, `error:`, `expected`, `got`, etc.). A Go panic stack frame looks like:

```
github.com/x/y/z.(*Model).Predict(...)
	/home/runner/.../classifier/model.go:142 +0x1a4
```

The function-name row has no `.go:` and matches none of the failure keywords, so it's discarded. After 5 lines the loop exits, and the runtime-side frames win the cap because they appear first in the trace. The user-code frames the developer actually needs are never emitted.

## Fix

Detect when the captured output for a failed test contains a panic / fatal-error / goroutine header. If so, switch to a contiguous "trace mode":

- preserve every non-cosmetic line verbatim (drop only `=== RUN` / `=== PAUSE` / `=== CONT` / `=== NAME` / `--- PASS` and blank lines)
- cap at 50 lines per failure, appending `… +N more trace lines (see tee file)` when the cap fires
- widen per-line truncation from 100 → 160 chars (long CI paths fit)

Plain assertion failures still go through the old heuristic path, but with the per-failure cap bumped from 5 → 8 lines so an immediate follow-up context line after a `file:line` header isn't dropped.

## Behavior after this PR

```
Go test: 336 passed, 1 failed, 1 skipped in 2 packages

classifier (1 passed, 1 failed)
  [FAIL] TestClassify
     --- FAIL: TestClassify (0.02s)
     panic: runtime error: index out of range [3] with length 2 [recovered]
     panic: runtime error: index out of range [3] with length 2
     goroutine 17 [running]:
     testing.tRunner.func1.2({0x10a3ce0, 0xc000018090})
     /usr/local/go/src/testing/testing.go:1545 +0x238
     ...
     github.com/tphakala/birdnet-go/internal/classifier.(*Model).Predict(...)
     /home/runner/work/birdnet-go/internal/classifier/model.go:142 +0x1a4
     github.com/tphakala/birdnet-go/internal/classifier.TestClassify(0xc0001b6680)
     /home/runner/work/birdnet-go/internal/classifier/classifier_test.go:58 +0xf8
     ...
[full output: ~/.local/share/rtk/tee/...go_test.log]
```

The tee hint stays — full unfiltered NDJSON remains one path away if a developer wants it. Passing-only output is unchanged (still a one-line summary).

## Tradeoff

Failed tests with panics produce a larger inline section than before. On the synthetic panic fixture, RTK now emits 1267 bytes of filtered output vs. 502 bytes pre-fix — but the raw NDJSON is 8095 bytes, so we're still cutting ~84% of bytes. The cap of 50 trace lines guarantees no single failure can blow past tee territory.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo test --bin rtk -- --test-threads=8` — 1906 passed, 0 failed, 6 ignored
- [x] New fixture `tests/fixtures/go_test_panic_json.txt` (panic + stack trace, 2 packages, 1 skipped)
- [x] New fixture `tests/fixtures/go_test_pass_json.txt` (336 passing tests, 2 packages, regression guard)
- [x] `test_filter_go_test_panic_preserves_full_trace` — asserts panic msg, `goroutine 17 [running]:`, `classifier_test.go:58`, `model.go:142` all inline
- [x] `test_filter_go_test_panic_caps_stack_trace_size` — output ≤ 80 lines
- [x] `test_filter_go_test_panic_still_saves_meaningful_bytes` — ≥ 60% byte savings even with panic trace
- [x] `test_filter_go_test_all_pass_remains_compact` — 336-pass case stays summary-only, ≥ 60% token savings
- [x] All pre-existing go_cmd tests still pass (timeout/build-fail/double-count regression cases)

Fixes #1882

### DIFF
--- a/src/cmds/go/go_cmd.rs
+++ b/src/cmds/go/go_cmd.rs
@@ -486,8 +486,16 @@ pub(crate) fn filter_go_test_json(output: &str) -> String {
         for (test, outputs) in &pkg_result.failed_tests {
             result.push_str(&format!("  [FAIL] {}\n", test));
 
+            // Trace-mode (panic/fatal) needs wider lines for full stack frame
+            // paths -- truncating to 100 chars would chop the file the user
+            // actually wrote. Assertion-mode keeps the original 100-char cap.
+            let trace_mode = outputs
+                .iter()
+                .any(|line| is_go_trace_marker(line.trim()));
+            let line_cap = if trace_mode { 160 } else { 100 };
+
             for line in select_go_test_failure_lines(outputs) {
-                result.push_str(&format!("     {}\n", truncate(&line, 100)));
+                result.push_str(&format!("     {}\n", truncate(&line, line_cap)));
             }
         }
     }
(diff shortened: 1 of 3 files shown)


YOMXXX:
Rebased this PR onto current develop and resolved the test-module conflict in src/cmds/go/go_cmd.rs by keeping both sets of regression coverage: the newer go build nonzero/package-pattern tests from develop and this PR's go test panic/pass fixture tests.

Local verification completed:
- cargo +1.93.0 fmt --check
- cargo +1.93.0 test go_test -- --nocapture
- cargo +1.93.0 test go_build -- --nocapture

PR #1985, opened 2026-05-20
fix(hooks/exclude): match multi-token entries against command prefix

### LINKED ISSUE #1919: [hooks] exclude_commands does not cover subcommands (e.g. "git diff")

YOMXXX:
## Summary

`[hooks].exclude_commands` entries with more than one whitespace-separated
token (e.g. `"git diff"`, `"docker compose up"`) were matched by an
anchored regex built from the user input. The matching itself happened
to work for the well-formed case, but the construction was fragile and
the user-facing contract was undocumented — issue #1919 reports that a
config like

```toml
[hooks]
exclude_commands = ["diff", "git diff"]
```

does not reliably stop `git diff HEAD` from being rewritten. This PR
makes the contract explicit and the implementation literal: a
multi-token entry matches a command iff the command **equals the entry
exactly** or **starts with the entry followed by a whitespace
character**.

## Reproduction (before)

```toml
# ~/.config/rtk/config.toml
[hooks]
exclude_commands = ["diff", "git diff"]
```

```bash
rtk rewrite "diff a b"       # exit 1  — correctly excluded (single-token)
rtk rewrite "git diff HEAD"  # exit 0  — NOT excluded (multi-token broken)
```

## Root cause

`src/discover/registry.rs::compile_exclude_patterns` wrapped every entry
in `^<regex::escape>($|\s)` and then matched it as a regex. For
multi-token entries this:

1. relies on `regex::escape` turning the literal space into `\ ` — a
   detail that is easy to break in future refactors;
2. degrades silently to a substring `starts_with` whenever the regex
   fails to compile, which means an entry like `git[diff` would happily
   exclude `git-difftool` as well as `git diff`;
3. obscures the actual user-facing contract behind two layers of
   indirection.

## Fix approach

- Treat `^`-prefixed entries as regex (keeps the existing
  `^curl`-style escape hatch for power users).
- Treat every other entry as a literal command prefix and store it as
  `ExcludePattern::Prefix`.
- In `is_excluded`, a `Prefix` matches when the (left-trimmed) command
  is either equal to the entry or starts with the entry followed by a
  whitespace character — never on a substring boundary.
- Canonicalise interior whitespace in entries so `"git  diff"` (two
  spaces, typo) still matches `git diff HEAD`.
- Keep the invalid-regex fallback path so a malformed `^…` regex
  degrades to a literal prefix on the un-anchored body rather than
  silently doing nothing.

## Test plan

- [x] `test_exclude_single_token_still_matches` — `"diff"` still excludes `diff a b`
- [x] `test_exclude_two_token_subcommand_git_diff` — headline #1919 case
- [x] `test_exclude_two_token_with_many_trailing_args` — `git diff --staged path/to/file`
- [x] `test_exclude_three_token_subcommand` — `docker compose up -d`
- [x] `test_exclude_exact_match_with_no_trailing_args` — bare `git diff`
- [x] `test_exclude_two_token_does_not_match_sibling_subcommand` — `git status` still rewrites
- [x] `test_exclude_two_token_does_not_match_subcommand_prefix` — `git diffstat`, `git-difftool` not excluded
- [x] `test_exclude_extra_whitespace_in_entry_is_normalised` — `"git  diff"` typo tolerated
- [x] `cargo fmt --all` clean
- [x] `cargo test --bin rtk -- --test-threads=8`: 1909 passed / 6 ignored / 0 failed
- [x] Manual smoke test against the binary with a real `config.toml` confirms `rtk rewrite "git diff HEAD"` now exits 1.

Fixes #1919

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -633,7 +633,14 @@ const SHELL_PREFIX_BUILTINS: &[&str] = &["noglob", "command", "builtin", "exec",
 const MAX_PREFIX_DEPTH: usize = 10;
 
 enum ExcludePattern {
+    /// Anchored regex (entry started with `^`). Matched with `Regex::is_match`.
     Regex(Regex),
+    /// Literal command prefix. The entry can be one or more whitespace-separated
+    /// tokens (e.g. `"diff"`, `"git diff"`, `"docker compose up"`); a command
+    /// matches when it either equals the entry exactly or begins with the entry
+    /// followed by a whitespace character. This is what allows `"git diff"` to
+    /// exclude `"git diff HEAD"` without also excluding `"git diffstat"`
+    /// (#1919).
     Prefix(String),
 }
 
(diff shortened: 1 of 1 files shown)


YOMXXX:

Expert lens: hook permissions / rewrite exclusion semantics.

This clarifies `exclude_commands` matching. Non-`^` entries become literal command prefixes with a whitespace boundary: `git diff` excludes `git diff HEAD`, but not `git diffstat` or `git-difftool`. Entries starting with `^` remain regex patterns for power users. The key behavior is literal matching with clear token boundaries, not substring matching.

Suggested review checks:
- Confirm single-token excludes such as `diff` and existing regex excludes still behave as before.
- Confirm multi-token excludes match equality and trailing-argument forms.
- Confirm sibling commands and prefix-looking names do not get excluded accidentally.
- Confirm malformed regex fallback is acceptable and does not widen into unsafe substring matching.

Issue #1875, opened 2026-05-13
Privacy/security concerns: secret output filtering, local command history, tracking toggle


VictorJulianiR:
## Summary

I reviewed RTK v0.34.3 after installing it for Codex and found a few privacy/security issues that look worth tracking in one place. I searched existing public issues and saw related broad/data-directory issues such as #640 and #1014, but did not find a targeted issue covering these specific findings.

### 1. `aws secretsmanager get-secret-value` appears to print secret material

`src/cmds/cloud/aws_cmd.rs` routes `aws secretsmanager get-secret-value` through `filter_secrets_get`.

The filter extracts `SecretString` and emits it as `Secret: ...`:

- `filter_secrets_get` appends the raw/compacted secret string to output.
- The tests assert that `password":"secret123"` and `plain-text-password` are present in the filtered output.

For an LLM-facing command filter, this seems risky: it can place secret material directly into agent context and logs/transcripts. Expected behavior would be to redact or summarize the secret metadata by default, with an explicit opt-in/raw mode if users really want secret values shown.

### 2. RTK local tracking stores full command strings and project paths

The local SQLite tracking database stores `original_cmd`, `rtk_cmd`, and `project_path` in the `commands` table.

This can persist sensitive data when users run commands containing inline tokens, URLs, auth headers, env values, file contents, or private path names. In my local install, command history included full URLs and PowerShell command bodies.

Expected behavior could be one or more of:

- redact obvious secret-bearing arguments before storage
- store only normalized command categories by default
- make full command-history storage opt-in
- document the exact local persistence behavior very prominently in Codex/agent setup output

### 3. `[tracking] enabled = false` appears not to be enforced

`Config` exposes `tracking.enabled`, but source search did not show checks of `tracking.enabled` before calls to `Tracker::new()` / `Tracker::record(...)`. `Tracker::record` inserts rows unconditionally once called.

If this is intended to disable tracking, it looks like a bug. If it is not intended to disable tracking, the config key is misleading.

### 4. Raw output tee can persist sensitive command output

`tee` is enabled by default in failure mode and writes raw command output to the local data directory when output is large enough. This can preserve secrets from failing commands, especially cloud/auth/config commands.

Potential mitigations:

- disable tee by default for known sensitive commands
- avoid tee for commands involving env/auth/secret/key/config/token patterns
- redact before tee writes
- make tee opt-in rather than default-on

## Environment

- RTK version: 0.34.3
[...]

PR #1986, opened 2026-05-20
fix(aws): redact secretsmanager get-secret-value payload

YOMXXX:
## Summary

`filter_secrets_get` previously emitted the raw `SecretString` into LLM-bound stdout via `format!("Secret: {}", ...)`, and the two tests at `aws_cmd.rs:2495-2526` codified this leak as expected behaviour. This PR redacts the value by default, exposes only structural metadata (name + length + SHA-256 prefix + JSON-key shape), and adds opt-in escape hatches plus a one-shot stderr warning so the behaviour change is visible.

## Reproduction

```bash
# Before
$ rtk aws secretsmanager get-secret-value --secret-id prod/rds/app
Name: prod/rds/app
Secret: {"username":"admin","password":"secret123"}    # <-- secret value lands in the LLM context

# After (default)
$ rtk aws secretsmanager get-secret-value --secret-id prod/rds/app
rtk: redacted SecretString (set RTK_REVEAL_SECRETS=1 to disable)   # stderr, once per process
Name: prod/rds/app
SecretString: <redacted; 47 bytes; sha256=ab12cd34>
keys: [password, username]

# Opt-in (env)
$ RTK_REVEAL_SECRETS=1 rtk aws secretsmanager get-secret-value --secret-id prod/rds/app
Name: prod/rds/app
Secret: {"username":"admin","password":"secret123"}

# Opt-in (CLI flag, stripped before forwarding to the underlying `aws` binary)
$ rtk aws secretsmanager get-secret-value --reveal-secret --secret-id prod/rds/app
Name: prod/rds/app
Secret: {"username":"admin","password":"secret123"}
```

## Root cause

`src/cmds/cloud/aws_cmd.rs:1514-1541` — `filter_secrets_get` formats the raw `SecretString` into the LLM-bound output via `format!("Secret: {}", ...)`. The two tests at lines 2495-2526 had codified the leak as expected behaviour.

## Fix approach

- **Default behaviour**: emit `Name` + a `SecretString: <redacted; N bytes; sha256=XXXXXXXX>` line + (for JSON-shaped secrets) a sorted `keys: [...]` line listing top-level keys. Value is never printed.
- **SHA-256 prefix**: full SHA-256 of the secret string, hex-encoded, truncated to the first 8 hex characters. Enough to round-trip-verify between runs without exposing the value.
- **`--reveal-secret` flag**: detected and stripped in the `secretsmanager get-secret-value` router branch (`src/cmds/cloud/aws_cmd.rs`) so the underlying `aws` binary never sees it. A process-local `AtomicBool` (`REVEAL_SECRET_FLAG`) signals the filter; it is cleared at end-of-invocation so subsequent in-process calls (tests, hooks) do not inherit it.
- **`RTK_REVEAL_SECRETS=1`**: independent env-based opt-in checked by the filter directly.
- **One-shot warning**: `std::sync::Once` guarantees the stderr line fires at most once per process even under parallel test execution. Tests are serialised via a private `reveal_env_lock()` Mutex (same pattern as `tracking.rs::test_db_path_env_and_default`).

## Behaviour changes

- Default output of `rtk aws secretsmanager get-secret-value <id>` no longer contains the secret value.
- The two existing tests (`test_filter_secrets_get`, `test_filter_secrets_get_plain_text`) are updated to assert the new behaviour.
- CHANGELOG entry recommended under **Security** for the next release.

No new dependencies (project already uses `sha2`). No public Rust API change. No `unwrap()` in production code.

## Test plan

- [x] `test_filter_secrets_get` (modified — asserts redaction, no `secret123`/`admin`, contains `sha256=`)
- [x] `test_filter_secrets_get_plain_text` (modified — no `plain-text-password`)
- [x] `test_filter_secrets_get_invalid_json` (unchanged passthrough)
- [x] `test_filter_secrets_get_emits_json_keys` (new — JSON shape exposed, values do not)
- [x] `test_filter_secrets_get_reveal_via_env` (new — `RTK_REVEAL_SECRETS=1` restores legacy output; ENV_LOCK-style mutex)
- [x] `test_filter_secrets_get_reveal_via_flag` (new — `REVEAL_SECRET_FLAG` (set by `--reveal-secret`) restores legacy output)
- [x] `test_filter_secrets_get_token_savings` (new — multi-line `aws secretsmanager get-secret-value --output json` fixture, ≥60% savings target)
- [x] `cargo fmt --all` clean
- [x] `cargo test --bin rtk -- --test-threads=8`: 1905 passed / 0 failed

Refs #1875 (Finding 1)

### DIFF
--- a/src/cmds/cloud/aws_cmd.rs
+++ b/src/cmds/cloud/aws_cmd.rs
@@ -14,6 +14,9 @@ use anyhow::{Context, Result};
 use lazy_static::lazy_static;
 use regex::Regex;
 use serde_json::Value;
+use sha2::{Digest, Sha256};
+use std::sync::Once;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 const MAX_ITEMS: usize = 20;
 const JSON_COMPRESS_DEPTH: usize = 4;
@@ -187,12 +190,35 @@ pub fn run(subcommand: &str, args: &[String], verbose: u8) -> Result<i32> {
         "s3" if !args.is_empty() && (args[0] == "sync" || args[0] == "cp") => {
             run_s3_transfer(&args[0], &args[1..], verbose)
         }
-        "secretsmanager" if !args.is_empty() && args[0] == "get-secret-value" => run_aws_filtered(
-            &["secretsmanager", "get-secret-value"],
-            &args[1..],
-            verbose,
-            filter_secrets_get,
-        ),
+        "secretsmanager" if !args.is_empty() && args[0] == "get-secret-value" => {
+            // Strip the rtk-private --reveal-secret flag before forwarding to aws CLI;
+            // record it in REVEAL_SECRET_FLAG so filter_secrets_get can opt out of
+            // redaction for this invocation. RTK_REVEAL_SECRETS=1 has the same effect.
+            let mut forwarded: Vec<String> = Vec::with_capacity(args.len().saturating_sub(1));
+            let mut reveal_via_flag = false;
+            for a in &args[1..] {
+                if a == "--reveal-secret" {
+                    reveal_via_flag = true;
+                } else {
+                    forwarded.push(a.clone());
+                }
+            }
+            if reveal_via_flag {
+                REVEAL_SECRET_FLAG.store(true, Ordering::SeqCst);
+            }
+            let exit_code = run_aws_filtered(
+                &["secretsmanager", "get-secret-value"],
+                &forwarded,
+                verbose,
+                filter_secrets_get,
+            );
+            // Clear the per-invocation flag so subsequent process-internal calls
+            // (in tests, hooks, etc.) do not inherit it.
+            if reveal_via_flag {
+                REVEAL_SECRET_FLAG.store(false, Ordering::SeqCst);
+            }
+            exit_code
+        }
         _ => run_generic(subcommand, args, verbose, &full_sub),
     }
 }
@@ -1511,25 +1537,88 @@ fn filter_s3_transfer(output: &str) -> FilterResult {
     FilterResult::new(result_lines.join("\n"))
 }
 
+/// Process-local "reveal" flag set by the `--reveal-secret` CLI flag.
+/// Equivalent in effect to `RTK_REVEAL_SECRETS=1`. Either trigger restores
+/// the legacy verbatim-secret output of `filter_secrets_get`.
+static REVEAL_SECRET_FLAG: AtomicBool = AtomicBool::new(false);
+
+/// One-shot stderr warning the first time we redact a SecretString in this
+/// process. Implemented with `std::sync::Once` so the message fires at most
+/// once even under parallel test execution.
+static REDACTION_WARNED: Once = Once::new();
+
+fn should_reveal_secret() -> bool {
+    if REVEAL_SECRET_FLAG.load(Ordering::SeqCst) {
+        return true;
+    }
+    matches!(
+        std::env::var("RTK_REVEAL_SECRETS").ok().as_deref(),
+        Some("1")
+    )
+}
+
+fn warn_redacted_once() {
+    REDACTION_WARNED.call_once(|| {
+        eprintln!("rtk: redacted SecretString (set RTK_REVEAL_SECRETS=1 to disable)");
+    });
+}
+
+/// Render the structural metadata of a SecretString without revealing its
+/// value: byte length plus an 8-hex-char SHA-256 prefix that is stable
+/// across runs so the agent can round-trip-verify it saw the same secret.
+fn redact_secret_string(secret_str: &str) -> String {
+    let bytes = secret_str.len();
+    let mut hasher = Sha256::new();
+    hasher.update(secret_str.as_bytes());
+    let digest = format!("{:x}", hasher.finalize());
+    let prefix: String = digest.chars().take(8).collect();
+    format!("<redacted; {} bytes; sha256={}>", bytes, prefix)
+}
+
 fn filter_secrets_get(json_str: &str) -> Option<FilterResult> {
     let v: Value = serde_json::from_str(json_str).ok()?;
 
+    let reveal = should_reveal_secret();
     let mut lines = Vec::new();
 
-    // Extract Name
+    // Extract Name (the lookup key the agent asked for — not sensitive).
     if let Some(name) = v["Name"].as_str() {
         lines.push(format!("Name: {}", name));
     }
 
-    // Extract SecretString
+    // Extract SecretString.
     if let Some(secret_str) = v["SecretString"].as_str() {
-        // Try to parse as JSON and compact it
-        if let Ok(secret_json) = serde_json::from_str::<Value>(secret_str) {
-            let compact =
-                serde_json::to_string(&secret_json).unwrap_or_else(|_| secret_str.to_string());
-            lines.push(format!("Secret: {}", compact));
+        // Try to parse as JSON and compact it (preserves shape detection).
+        let parsed = serde_json::from_str::<Value>(secret_str).ok();
+
+        if reveal {
+            // Legacy behaviour: emit the raw secret value verbatim.
+            if let Some(secret_json) = parsed {
+                let compact = serde_json::to_string(&secret_json)
+                    .unwrap_or_else(|_| secret_str.to_string());
+                lines.push(format!("Secret: {}", compact));
+            } else {
+                lines.push(format!("Secret: {}", secret_str));
+            }
         } else {
-            lines.push(format!("Secret: {}", secret_str));
+            // Default behaviour: emit only structural metadata.
+            warn_redacted_once();
+            lines.push(format!("SecretString: {}", redact_secret_string(secret_str)));
+
+            // For JSON-shaped secrets, expose top-level key names (not values)
+            // so the agent knows the shape (e.g. {username, password}).
+            if let Some(Value::Object(map)) = parsed {
+                let mut keys: Vec<&String> = map.keys().collect();
+                keys.sort();
+                if !keys.is_empty() {
+                    let joined = keys
+                        .iter()
+                        .map(|k| k.as_str())
+                        .collect::<Vec<_>>()
+                        .join(", ");
+                    lines.push(format!("keys: [{}]", joined));
+                }
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/cmds/cloud/aws_cmd.rs
```diff
@@ -1511,25 +1537,88 @@ fn filter_s3_transfer(output: &str) -> FilterResult {
...
+                    .unwrap_or_else(|_| secret_str.to_string());
+                lines.push(format!("Secret: {}", compact));
+            } else {
+                lines.push(format!("Secret: {}", secret_str));
+            }
         } else {
-            lines.push(format!("Secret: {}", secret_str));
+            // Default behaviour: emit only structural metadata.
+            warn_redacted_once();
+            lines.push(format!("SecretString: {}", redact_secret_string(secret_str)));
```
KuSh, line 1606: **Blocking — the filter fails open, leaking the secret verbatim.**

This branch only runs when `v["Name"]`/`v["SecretString"]` resolve on the top-level object. With `--query`, they don't: both are `Null`, `lines` stays empty, and the `return None` at line 1625 hands control to the passthrough branch in `run_aws_filtered`, which prints the unfiltered response.

Reproduced on this branch (`03de7b0c`), stub `aws` on `PATH`:

```console
$ rtk aws secretsmanager get-secret-value --secret-id db --query SecretString
rtk: filter warning: aws filter returned None, passing through raw output
"pw"

$ rtk aws secretsmanager get-secret-value --secret-id db --query '[Name,SecretString]'
rtk: filter warning: aws filter returned None, passing through raw output
[
    "db",
    "pw"
]
```

`--query SecretString` is the idiom in AWS's own documentation, and `run_aws_json` rewrites `--output text` to `--output json`, so that variant lands here too.

A security filter has to fail **closed**: on this subcommand, an unrecognised shape should redact (or refuse) rather than fall back to raw. Locating the secret anywhere in the parsed value, instead of at two fixed top-level keys, would cover the `--query` forms.

src/cmds/cloud/aws_cmd.rs
```diff
@@ -187,12 +190,35 @@ pub fn run(subcommand: &str, args: &[String], verbose: u8) -> Result<i32> {
...
+                if a == "--reveal-secret" {
+                    reveal_via_flag = true;
+                } else {
+                    forwarded.push(a.clone());
+                }
+            }
+            if reveal_via_flag {
+                REVEAL_SECRET_FLAG.store(true, Ordering::SeqCst);
+            }
+            let exit_code = run_aws_filtered(
```
KuSh, line 209: **Blocking (after merge) — `run_aws_filtered` re-emits `raw`, undoing the redaction.**

This PR branches off a `develop` where the runner still `println!`s `result.text` directly. Current `develop` ends with `emit_guarded(&result.text, hint, &raw)` → `guard::never_worse(raw, body)`, which returns **`raw`** whenever the filtered body estimates more tokens. The redaction line is *longer* than a narrowed response, so the guard flips:

```console
$ rtk aws secretsmanager get-secret-value --secret-id db \
      --query '{Name:Name,SecretString:SecretString}'
rtk: redacted SecretString (set RTK_REVEAL_SECRETS=1 to disable)
{
    "Name": "db",
    "SecretString": "pw"
}
```

raw 46 B → 11 est. tokens; redacted body 59 B → 14. `14 > 11`, so `never_worse` returns raw. stderr claims redaction happened while stdout carries the plaintext — a silent failure of the control. Default config; no flags needed. The same command on this branch pre-merge redacts correctly, so CI here cannot catch it.

Second route, same function: with `tee.mode = "always"` the unfiltered response is written to disk and its path printed into the agent's stdout —

```console
SecretString: <redacted; 334 bytes; sha256=bd8dd5a7>
keys: [connection_url, dbname, engine, host, password, port, username]
[full output: ~/.local/share/rtk/tee/1788741469_aws_secr_733b51.log]
```

— and that file holds the plaintext, persisting after exit. `force_tee_hint` (taken when `result.truncated`) skips the mode check entirely.

Together with the `None` passthrough above, that's three routes around the filter. Suggest handling it here in the router branch: pass redacted text to tee/track and opt this subcommand out of `never_worse`, so the secret path can never emit `raw`.

src/cmds/cloud/aws_cmd.rs
```diff
@@ -1511,25 +1537,88 @@ fn filter_s3_transfer(output: &str) -> FilterResult {
...
+/// Render the structural metadata of a SecretString without revealing its
+/// value: byte length plus an 8-hex-char SHA-256 prefix that is stable
+/// across runs so the agent can round-trip-verify it saw the same secret.
+fn redact_secret_string(secret_str: &str) -> String {
+    let bytes = secret_str.len();
+    let mut hasher = Sha256::new();
+    hasher.update(secret_str.as_bytes());
+    let digest = format!("{:x}", hasher.finalize());
+    let prefix: String = digest.chars().take(8).collect();
+    format!("<redacted; {} bytes; sha256={}>", bytes, prefix)
```
KuSh, line 1575: **Note — the metadata is an offline verifier for low-entropy secrets.**

An unsalted SHA-256 truncated to 32 bits, published next to the *exact* byte length, is enough to confirm a guessed secret. Verified against the built binary:

```console
$ rtk aws secretsmanager get-secret-value --secret-id plain
SecretString: <redacted; 7 bytes; sha256=f52fbd32>
# recovered from a 6-word list: 'hunter2' (len=7, sha256=f52fbd32)
```

For JSON secrets the whole blob is hashed, so the bar is genuinely high there. The plaintext path is the exposed one — API keys, passwords, PINs — and this line ends up in transcripts and logs, exactly where the value was not supposed to go.

The stated goal (round-trip-verify the same secret across runs *within* a session) is met by an HMAC under a per-process random key: stable within a run, useless to anyone reading the transcript later. Bucketing the length (`~10s of bytes`) rather than printing it exactly would also cut the candidate set down less sharply.

src/cmds/cloud/aws_cmd.rs
```diff
@@ -1511,25 +1537,88 @@ fn filter_s3_transfer(output: &str) -> FilterResult {
...
+            if let Some(Value::Object(map)) = parsed {
+                let mut keys: Vec<&String> = map.keys().collect();
+                keys.sort();
+                if !keys.is_empty() {
+                    let joined = keys
+                        .iter()
+                        .map(|k| k.as_str())
+                        .collect::<Vec<_>>()
+                        .join(", ");
+                    lines.push(format!("keys: [{}]", joined));
```
KuSh, line 1619: **Note — key names are frequently identifiers, and this list is uncapped.**

The comment above says key names are exposed but values are not. In practice key names carry the identifying half of a credential — access key IDs, service-account addresses, hostnames:

```console
$ rtk aws secretsmanager get-secret-value --secret-id k
Name: keys-are-identifiers
SecretString: <redacted; 79 bytes; sha256=5433ae32>
keys: [AKIAIOSFODNN7EXAMPLE, svc-account@example.invalid]
```

Also unbounded — unlike the rest of this module, there's no `MAX_ITEMS`/`join_with_overflow` cap, so a wide secret prints every key. Worth capping, and worth reconsidering whether the shape hint is worth this on secrets whose keys are the sensitive half.

src/cmds/cloud/aws_cmd.rs
```diff
@@ -1511,25 +1537,88 @@ fn filter_s3_transfer(output: &str) -> FilterResult {
...
+    }
+    matches!(
+        std::env::var("RTK_REVEAL_SECRETS").ok().as_deref(),
+        Some("1")
+    )
+}
+
+fn warn_redacted_once() {
+    REDACTION_WARNED.call_once(|| {
+        eprintln!("rtk: redacted SecretString (set RTK_REVEAL_SECRETS=1 to disable)");
```
KuSh, line 1562: **Nit — the warning omits the flag, and the env check is stricter than it reads.**

The message advertises only `RTK_REVEAL_SECRETS=1` and never mentions `--reveal-secret`, which this PR adds as the other opt-in — a user who hits this line won't discover the flag.

Separately, `should_reveal_secret` matches exactly `Some("1")`, so `RTK_REVEAL_SECRETS=true` / `=yes` silently redact with no indication the value was rejected. Worth either accepting the usual truthy spellings or saying `=1` is required.

KuSh (changes requested):
Thanks for this — the leak is real and the fix works. I verified it end to end against an actual `aws` invocation (stub producer on `PATH`, real `rtk` binary, full `run()` → `run_aws_json` → filter path): the secret no longer reaches stdout, 82.5% byte reduction, 3.9 ms startup. It is also **not** superseded — `develop` still prints `Secret: {"username":"admin","password":"..."}` today.

Requesting changes on two paths where the redaction is bypassed, plus the merge conflict. Everything below was reproduced against a running binary, not read off the diff. All examples are synthetic (`example.invalid`, AWS's own `AKIAIOSFODNN7EXAMPLE` placeholder).

## Blocking

**A. The filter fails _open_ on shapes it doesn't recognise** (inline at the redaction branch). `--query` — the form AWS's docs use — makes both top-level lookups `Null`, so `lines` is empty, `return None` fires, and the runner passes the raw response through. Reproduced on this branch.

**B. After merge into current `develop`, the runner re-emits `raw` even when the filter redacted correctly** (inline at the router branch). Two independent routes, one of them silent and on default config.

**C. Merge conflict.** Imports only. I resolved it locally (`sha2` + `atomic` + `LazyLock` + `Once`); the merged tree builds clean, `fmt`/`clippy` clean, **3111 tests pass** — so this is mechanical, not a design problem.

The unifying point for A and B: there are **three** independent routes by which `run_aws_filtered` re-emits `raw` after the filter did its job — the `None` passthrough, `never_worse`, and the tee hint. Redaction implemented inside `filter_secrets_get` cannot hold while the shared runner has three ways around it. I'd suggest fixing this at the router branch — this subcommand should never emit `raw` under any condition — rather than patching the routes one at a time.

## Non-blocking notes

- **`--reveal-secret` is an opt-out the protected party can write itself.** The threat model is "keep secret values out of the LLM context", but the LLM composes the command, and the hook rewrites `aws …` → `rtk aws …` so the flag rides along. `RTK_REVEAL_SECRETS` at least has to come from the human's shell. Worth considering dropping the flag and keeping only the env opt-in.
- **`SecretBinary`** produces a bare `Name:` line with no warning — the agent cannot tell that anything was withheld. Pre-existing, but it is the one path that does not follow the redaction contract this PR introduces.
- **Env lock is module-local.** You followed the documented `tracking.rs` pattern, but that mutex is private to `tracking.rs`, so `reveal_env_lock` is a second disjoint mutex; there are 36 `set_var`/`remove_var` sites repo-wide, and concurrent `setenv`/`getenv` is a data race. You followed the stated convention — the convention is the problem, so I don't think this is yours to fix here.

## Not this PR's defect

`secretsmanager batch-get-secret-value` and `ssm get-parameter --with-decryption` emit the secret verbatim through `run_generic`. I verified both on **pristine `develop` without this PR**, so they are pre-existing and not a regression from your change. Flagging only because an agent blocked on `get-secret-value` gets the value from the adjacent command with no warning — worth a follow-up that keys redaction on response shape rather than one subcommand name.

## Suggested test

A differential test asserting the secret literal appears in **no** emitted stream — stdout, tee file, hint target — across the `--query` forms x tee modes. That closes all three re-emission routes at once; a unit test on `filter_secrets_get` alone cannot, because two of the three live in the runner.

PR #2002, opened 2026-05-21
chore(cargo): declare MSRV via rust-version = "1.91"

### LINKED ISSUE #1402: Add rust-version to Cargo.toml (MSRV = 1.91)
v0.37.1 uses `str::floor_char_boundary()` in `src/cmds/system/pipe_cmd.rs:135`, which was stabilized in Rust 1.91.0.

Without `rust-version` in Cargo.toml, building with Rust < 1.91 fails with a confusing error:

```
error[E0658]: use of unstable library feature `round_char_boundary`
 --> src/cmds/system/pipe_cmd.rs:135:21
  |
135 | let end = input.floor_char_boundary(end);
    |                 ^^^^^^^^^^^^^^^^^^^
```

Adding `rust-version = "1.91"` to `[package]` in Cargo.toml would give a clear error message at build time instead:

```
error: package `rtk v0.37.1` cannot be built because it requires rustc 1.91 or newer
```

This affects distribution packagers (e.g., Arch Linux AUR) where the system Rust version may lag behind.

YOMXXX:
## Summary

\`rtk\` already uses \`str::floor_char_boundary()\` (\`src/cmds/system/pipe_cmd.rs:140\`), stabilized in Rust 1.91.0, but \`Cargo.toml\` doesn't declare the MSRV. Distribution packagers on older toolchains see a confusing \`use of unstable library feature\` rustc error instead of a clear cargo-level MSRV diagnostic.

## Reproduction

\`\`\`bash
# Without this PR, on Rust 1.88.0:
$ cargo check
error[E0658]: use of unstable library feature \`round_char_boundary\`
 --> src/cmds/system/pipe_cmd.rs:140:21
\`\`\`

## Fix

Add \`rust-version = "1.91"\` to the \`[package]\` table.

\`\`\`toml
[package]
name = "rtk"
version = "0.34.3"
edition = "2021"
rust-version = "1.91"   # ← added
\`\`\`

## After

\`\`\`bash
$ cargo +1.88 check
error: rustc 1.88.0 is not supported by the following packages:
  rtk@0.34.3 requires rustc 1.91
\`\`\`

Cargo gives the right diagnostic at the \`[package]\`-validation step, before \`rustc\` ever runs.

## Test plan

- [x] \`cargo +1.93 fmt --all -- --check\` clean.
- [x] \`cargo +1.93 clippy --all-targets\` zero warnings.
- [x] \`cargo +1.93 test --bin rtk -- --test-threads=8\` → 1909 passed.
- [x] \`cargo +1.88 check\` reproduces the friendly MSRV error (intended behaviour).

## Notes

@DavidReque expressed interest a month ago but no PR materialized — happy to defer if they'd prefer to push their own; otherwise this lands the minimal one-line change so packagers stop hitting the unhelpful rustc message.

Fixes #1402

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 name = "rtk"
 version = "0.34.3"
 edition = "2021"
+rust-version = "1.91"
 authors = ["Patrick Szymkowiak"]
 description = "Rust Token Killer - High-performance CLI proxy to minimize LLM token consumption"
 license = "MIT"


YOMXXX:

Expert lens: Rust packaging / MSRV.

Scope is intentionally one line: `Cargo.toml` now declares `rust-version = "1.91"` because the code already uses `str::floor_char_boundary()`, which requires Rust 1.91. This does not change runtime behavior or dependency resolution on supported toolchains; it only makes older toolchains fail early with Cargo's clear MSRV diagnostic instead of a confusing rustc unstable-feature error.

Suggested review checks:
- Confirm Rust 1.91 is the actual minimum implied by current source usage.
- Confirm packagers are expected to see Cargo's MSRV error rather than compile into `pipe_cmd.rs` and fail later.
- No code-path review needed beyond the package metadata change.

YOMXXX:
@aeppling Could I ask for a focused maintainer pass on a small first batch?

I have several PRs open, so to avoid overloading reviewers I would like to prioritize only the narrow, mergeable, CI-clean ones first:

- #2002 — one-line MSRV declaration in `Cargo.toml`; no runtime behavior change.
- #2005 — narrow `bunx prisma` rewrite-table extension; routes into the existing Prisma handler.
- #1983 — restores the missing `php` interpreter rewrite rule with boundary tests for `phpunit`/adjacent tools.

The larger/security-sensitive PRs can wait. If this is not the preferred way to queue reviews, I am happy to follow whatever workflow maintainers prefer.

pszymkowiak (approved):
Thanks @YOMXXX — clean one-line metadata fix. The crate already requires `floor_char_boundary` (1.91+) in `pipe_cmd.rs:135` and `json_cmd.rs:109`, so declaring the MSRV explicitly is the right call — it turns the cryptic `E0658` rustc error into a clean cargo-level diagnostic for distribution packagers. CI green across the three OSes. Approving.

PR #2004, opened 2026-05-21
fix(discover): register ssh in the rewrite rule table so hook routes it

### LINKED ISSUE #1654: SSH filter passes `verify` (3/3) but `rtk rewrite` returns exit 1 — ssh commands unintercepted on Claude Code hook path
## Summary

The built-in `ssh` filter is registered (passes `rtk verify --filter ssh`)
but `rtk rewrite` never matches any `ssh ...` command shape, so the Claude
Code Bash hook path leaves all `ssh` invocations uncompressed. Other built-in
filters (`ps`, `grep`, `ls`) on the same machine continue to match correctly.

## Environment

- rtk **0.37.2** (Homebrew, `/opt/homebrew/Cellar/rtk/0.37.2/bin/rtk`)
- macOS (Darwin 25.4.0)
- Claude Code hook: `rtk hook claude` (configured in `~/.claude/settings.json`)
- `RTK_TELEMETRY=off`
- No project-local `.rtk/filters.toml` overrides for ssh

## Reproduction

```bash
$ rtk --version
rtk 0.37.2

$ rtk verify --filter ssh
3/3 tests passed

# SSH never matches rewrite (any shape):
$ for cmd in 'ssh root@host uptime' 'ssh user@example.test ls' \
             'ssh -o ConnectTimeout=5 host echo OK' 'ssh host'; do
    rtk rewrite "$cmd" >/dev/null 2>&1; echo "exit=$? cmd=$cmd"
  done
exit=1 cmd=ssh root@host uptime
exit=1 cmd=ssh user@example.test ls
exit=1 cmd=ssh -o ConnectTimeout=5 host echo OK
exit=1 cmd=ssh host

# Sanity check — other built-ins match correctly:
$ rtk rewrite "ps -ef"      # → exit 3, "rtk ps -ef"
$ rtk rewrite "grep -n x f" # → exit 0, "rtk grep -n x f"
$ rtk rewrite "ls -la"      # → exit 0, "rtk ls -la"
```

## Impact (30-day data)

`rtk discover` from this machine lists `ssh` as the #1 unhandled command,
with several hundred invocations in 30 days — far ahead of any other
unhandled command on the list.

Estimated savings missed at 95% compression: ~50K-100K tokens/month for
this single user (infra ops with heavy ssh use).

## Earlier in-the-wild evidence

`rtk gain --history` from a prior session captured 10 `rtk:toml ssh ...`
intercepts during a ~5-min window (86.5% savings) and zero intercepts since,
across multiple Claude Code session restarts. The current 30-day history
shows `rtk:toml ps -ef` and `rtk:toml ansible-playbook` firing normally —
ssh is the only built-in filter that has gone dark.

## Expected vs actual

- **Expected:** `rtk rewrite "ssh root@host uptime"` returns exit 0/3 with
  the rewritten `rtk ssh ...` command, mirroring `ps -ef` behavior, so the
  Claude Code hook can route ssh through compression.
- **Actual:** `rtk rewrite` returns exit 1 (passthrough) for every ssh form
  tested, despite `rtk verify --filter ssh` reporting 3/3 tests passing.

## Hypothesis

The `verify` and `rewrite` code paths appear to consult the SSH filter
differently — `verify` exercises the filter directly while `rewrite` may
have an exclusion or path bug that skips ssh. Possibly related to the
v0.35.0 security change (#886) that altered registry default behavior, but
that should produce exit 3, not exit 1.

## Ask

Could you confirm whether `rtk rewrite` is intentionally excluding `ssh`
in current builds, and if not, what diagnostic output would help track this
down? Happy to bisect across versions if useful.

---

_(Note: this is a clean refile. Original #1653 was withdrawn for a content-hygiene cleanup; same technical content, no internal hostnames in the example output.)_

YOMXXX:
## Summary

\`rtk verify --filter ssh\` reports 3/3 passing because the TOML filter at \`src/filters/ssh.toml\` is valid, but \`rtk rewrite "ssh …"\` exits 1 for every shape — the Claude Code hook's rewrite path therefore never produces \`rtk ssh …\` and the ssh filter is effectively dark in production, even though verify passes.

Reporter (#1654) saw ssh as the #1 unhandled command in 30 days of \`rtk discover\`, with ~50K–100K tokens/month of savings missed on a single user.

## Root cause

\`registry::rewrite_command\` → \`rewrite_segment_inner\` looks the base command up in \`RULES\` (\`src/discover/rules.rs\`). \`RULES\` is the source of truth for the hook rewrite path, but it did not list ssh. \`rtk verify\` consults the TOML filter registry; \`rtk rewrite\` consults the hard-coded RULES table — so a TOML-only filter with no RULES entry passes verify yet is invisible to the hook.

This split is the same shape as the historical \`ps\` / \`ansible-playbook\` setup that the reporter mentions still working — those have RULES entries; ssh did not.

## Fix

Add one \`RtkRule\` for ssh next to curl/wget (same \`Network\` category, same \`^cmd\\s+\` pattern shape). Once \`rtk rewrite\` produces \`rtk ssh …\`, main.rs's TOML-filter dispatch picks up \`src/filters/ssh.toml\` unchanged — no other code paths needed.

\`\`\`rust
RtkRule {
    pattern: r"^ssh\\s+",
    rtk_cmd: "rtk ssh",
    rewrite_prefixes: &["ssh"],
    category: "Network",
    savings_pct: 65.0,
    ...
}
\`\`\`

## Behaviour

| Input | After |
|---|---|
| \`ssh user@host uptime\` | \`rtk ssh user@host uptime\` (Allow + filter) |
| \`ssh -o ConnectTimeout=5 host echo OK\` | \`rtk ssh -o ConnectTimeout=5 host echo OK\` |
| \`ssh host\` (login only) | \`rtk ssh host\` (banners still filtered by \`strip_lines_matching\`) |
| bare \`ssh\` (no args) | unchanged (\`\\s+\` guard, nothing to filter) |

\`GIT_SSH_COMMAND="ssh -o …" git push\` continues to env-strip → \`rtk git push\` exactly as today; covered by existing \`test_classify_env_prefix_stripped\`.

## Test plan

- [x] \`test_classify_ssh_with_remote_command\` — Supported / Network classification, locks the rule-order invariant.
- [x] \`test_rewrite_ssh_shapes\` — three reproduction shapes from the issue verbatim.
- [x] \`test_rewrite_ssh_with_host_only\` — the no-remote-command shape from the repro.
- [x] \`test_rewrite_bare_ssh_not_rewritten\` — guards the \`\\s+\` boundary so a future loosened pattern doesn't start rewriting argv-less \`ssh\`.
- [x] \`cargo test --bin rtk -- --test-threads=8\` → **1913 passed** (was 1909).

## Notes

- \`src/filters/ssh.toml\` itself is unchanged; this PR only fills the RULES gap that kept the hook from ever reaching it.
- Token savings for an ssh command depend heavily on remote output, but \`ssh.toml\` strips connection banners + caps at \`max_lines = 200\`; \`savings_pct: 65\` follows wget's neighbour entry as a conservative estimate.
- \`rtk verify --filter ssh\` continues to pass (filter logic untouched).

Fixes #1654

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1011,6 +1011,67 @@ mod tests {
         );
     }
 
+    // --- #1654: ssh classify + rewrite roundtrip ---
+
+    #[test]
+    fn test_classify_ssh_with_remote_command() {
+        // Hook path must classify ssh-with-args as Supported so `rtk rewrite`
+        // returns the prefixed form instead of bailing with exit 1.
+        match classify_command("ssh user@host uptime") {
+            Classification::Supported {
+                rtk_equivalent,
+                category,
+                ..
+            } => {
+                assert_eq!(rtk_equivalent, "rtk ssh");
+                assert_eq!(category, "Network");
+            }
+            other => panic!("expected Supported, got {:?}", other),
+        }
+    }
+
+    #[test]
+    fn test_rewrite_ssh_shapes() {
+        // Spot-check the four shapes from #1654's reproduction script.
+        let cases = [
+            ("ssh root@host uptime", "rtk ssh root@host uptime"),
+            ("ssh user@example.test ls", "rtk ssh user@example.test ls"),
+            (
+                "ssh -o ConnectTimeout=5 host echo OK",
+                "rtk ssh -o ConnectTimeout=5 host echo OK",
+            ),
+        ];
+        for (input, expected) in cases {
+            assert_eq!(
+                rewrite_command_no_prefixes(input, &[]),
+                Some(expected.into()),
+                "rewrite for {:?} should be Some({:?})",
+                input,
+                expected
+            );
+        }
+    }
+
+    #[test]
+    fn test_rewrite_ssh_with_host_only() {
+        // `ssh host` (no remote command, just login) is the fourth shape
+        // from #1654's reproduction script. It still needs to flow through
+        // the filter — even login banners benefit from `strip_lines_matching`
+        // in `src/filters/ssh.toml`.
+        assert_eq!(
+            rewrite_command_no_prefixes("ssh host", &[]),
+            Some("rtk ssh host".into())
+        );
+    }
+
+    #[test]
+    fn test_rewrite_bare_ssh_not_rewritten() {
+        // `ssh` with no arguments at all is left alone (the regex requires
+        // `\s+` after `ssh`); without an argv it would just print usage to
+        // stderr and there's nothing to filter.
+        assert_eq!(rewrite_command_no_prefixes("ssh", &[]), None);
+    }
+
     #[test]
     fn test_classify_sudo_stripped() {
         assert_eq!(
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -425,6 +425,22 @@ pub const RULES: &[RtkRule] = &[
         subcmd_savings: &[],
         subcmd_status: &[],
     },
+    // #1654: ssh has a TOML filter (`src/filters/ssh.toml`) that passes
+    // `rtk verify --filter ssh` but the hook rewrite path never reached it
+    // because the registry didn't list a matching rule. Adding the rule lets
+    // `ssh user@host cmd` flow through `rtk rewrite` → `rtk ssh user@host cmd`
+    // → main.rs's TOML-filter dispatch. Interactive `ssh host` (no remote
+    // command, banners only) is intentionally not rewritten — `\s+` requires
+    // at least one argument, mirroring wget/curl above.
+    RtkRule {
+        pattern: r"^ssh\s+",
+        rtk_cmd: "rtk ssh",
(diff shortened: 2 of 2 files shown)


YOMXXX:
Review note: I would not merge this as-is because it rewrites interactive SSH shapes.

The tests intentionally rewrite `ssh host` to `rtk ssh host`, but that can be an interactive login session. Running it through RTK's TOML filter means stdout is captured/filtered, which can change or break interactive SSH, port forwarding, and no-command sessions like `ssh -N`, `ssh -L`, or `ssh -D`.

Please narrow the rewrite rule to remote-command forms only, or explicitly exclude interactive/forwarding shapes. Also, the new comment in `rules.rs` says interactive `ssh host` is intentionally not rewritten, while the test asserts that it is rewritten. That contradiction should be resolved with the behavior change.

PR #2005, opened 2026-05-21
feat(discover): rewrite bunx prisma → rtk prisma

### LINKED ISSUE #1401: feat: rewrite bunx prisma → rtk prisma (bun runner for Prisma CLI)
## Context

RTK already handles `npx prisma` → `rtk prisma`. When using Bun as the runtime, the equivalent invocation is `bunx prisma` — but the hook doesn't rewrite it, so output lands unfiltered.

## Usage data (rtk discover, 30-day scan, 13,730 Bash calls)

| Command | Monthly runs | Example |
|---------|-------------|---------|
| `bunx prisma` | **45** | `bunx prisma migrate dev --name add_venue_col` |

## Expected behavior

The hook rewrite registry should map `bunx prisma` → `rtk prisma`, identical to how `npx prisma` is already handled.

`bunx prisma` is semantically equivalent to `npx prisma` — same CLI, same output format, same subcommands (`migrate dev`, `migrate deploy`, `migrate status`, `generate`, `db seed`, `studio`). No new output-filtering logic is needed; just a new rewrite rule pointing at the existing Prisma handler.

## Proposed rewrite rules

```
bunx prisma migrate dev   → rtk prisma migrate dev
bunx prisma migrate deploy → rtk prisma migrate deploy
bunx prisma migrate status → rtk prisma migrate status
bunx prisma generate      → rtk prisma generate
bunx prisma db seed       → rtk prisma db seed
bunx prisma studio        → rtk prisma studio
```

Or simply: `bunx prisma <subcmd>` → `rtk prisma <subcmd>` as a prefix match.

## Stack context

Bun 1.2.19 + Elysia.js + Prisma + PostgreSQL monorepo. We run `bunx prisma migrate dev` and `bunx prisma migrate deploy` heavily during development and CI. Currently 45 calls/month flow through unoptimized because the rewrite rule only covers `npx prisma`.

RTK already has the hard part (Prisma output filtering). This is just a hook wiring change.

YOMXXX:
## Summary

Bun is the primary JS runtime on a growing share of Prisma projects. \`bunx prisma <sub>\` is semantically equivalent to \`npx prisma <sub>\` — same CLI, same output format, same subcommands — but the hook didn't recognise it, so the reporter's 45 monthly \`bunx prisma\` invocations (#1401) were flowing through the Bash hook completely unfiltered.

## Fix

One \`RtkRule\` change in \`src/discover/rules.rs\`:

\`\`\`diff
-    pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?prisma",
+    pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx|bunx)\s+)?prisma",
     rtk_cmd: "rtk prisma",
     rewrite_prefixes: &[
+        "bunx prisma",
         "npm exec prisma",
         …
     ],
\`\`\`

No new output-filtering logic — the entire Bun-Prisma stack goes through the same \`src/cmds/js/prisma_cmd.rs\` handler as npx/pnpm/npm; only the rewrite recognition was missing.

## Behaviour

| Input | After |
|---|---|
| \`bunx prisma migrate dev\` | \`rtk prisma migrate dev\` |
| \`bunx prisma migrate dev --name add_col\` | \`rtk prisma migrate dev --name add_col\` |
| \`bunx prisma migrate deploy\` | \`rtk prisma migrate deploy\` |
| \`bunx prisma migrate status\` | \`rtk prisma migrate status\` |
| \`bunx prisma generate\` | \`rtk prisma generate\` |
| \`bunx prisma db seed\` | \`rtk prisma db seed\` |
| \`bunx prisma studio\` | \`rtk prisma studio\` |

\`npx prisma\` / \`pnpm prisma\` / \`pnpm dlx prisma\` continue to work exactly as before — same rule, just one more alternation branch.

## Scope notes

- This PR is intentionally narrow: it only adds \`bunx\` to the **existing** Prisma rule. Other \`bun\` runner forms (\`bun x\`, \`bun run\`, bare \`bun prisma\`) are *not* covered here; if usage data shows they matter we can extend the alternation in a follow-up.
- Other RTK-handled binaries that ship as \`bunx <tool>\` (vitest, playwright, tsc, prettier, biome, jest, eslint…) are out of scope. The reporter only flagged Prisma — happy to do a sweep in a separate PR if maintainers want broad bun adoption.

## Test plan

- [x] \`test_classify_prisma\` and \`test_rewrite_prisma\` extended with \`bunx prisma\` (locks the invariant fence per runner).
- [x] New \`test_rewrite_bunx_prisma_subcommands\` covers the six subcommand shapes the reporter named.
- [x] \`cargo test --bin rtk -- --test-threads=8\` → **1910 passed**.

Fixes #1401

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -2851,6 +2851,7 @@ mod tests {
             "npx prisma",
             "pnpm prisma",
             "pnpx prisma",
+            "bunx prisma",
             "prisma",
         ];
         for command in commands {
@@ -2885,6 +2886,7 @@ mod tests {
             "npx prisma",
             "pnpm prisma",
             "pnpx prisma",
+            "bunx prisma",
             "prisma",
         ];
         for command in commands {
@@ -2893,6 +2895,34 @@ mod tests {
                 Some("rtk prisma migrate dev".into()),
                 "Failed for command: {}",
                 command
+            );
+        }
+    }
+
+    #[test]
+    fn test_rewrite_bunx_prisma_subcommands() {
+        // #1401: exercise the full Prisma surface that the reporter said hits
+        // unfiltered with bun. The classify_command/rewrite tests above already
+        // cover `migrate dev` for every runner; this nails down each subcommand
+        // shape verbatim so a future regex tweak doesn't silently drop one.
+        let cases = [
+            ("bunx prisma migrate dev", "rtk prisma migrate dev"),
+            (
+                "bunx prisma migrate dev --name add_col",
+                "rtk prisma migrate dev --name add_col",
+            ),
+            ("bunx prisma migrate deploy", "rtk prisma migrate deploy"),
+            ("bunx prisma migrate status", "rtk prisma migrate status"),
+            ("bunx prisma generate", "rtk prisma generate"),
+            ("bunx prisma db seed", "rtk prisma db seed"),
+            ("bunx prisma studio", "rtk prisma studio"),
+        ];
+        for (input, expected) in cases {
+            assert_eq!(
+                rewrite_command_no_prefixes(input, &[]),
+                Some(expected.into()),
+                "Failed for input: {}",
+                input
             );
         }
     }
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -346,10 +346,15 @@ pub const RULES: &[RtkRule] = &[
         subcmd_savings: &[],
         subcmd_status: &[],
     },
+    // #1401: Bun is now the primary JS runtime on a growing share of Prisma
+    // projects. `bunx prisma <sub>` is semantically equivalent to
+    // `npx prisma <sub>` (same CLI, same output format), so it routes through
+    // the same `rtk prisma` filter — no new output-handling logic required.
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?prisma",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx|bunx)\s+)?prisma",
         rtk_cmd: "rtk prisma",
         rewrite_prefixes: &[
+            "bunx prisma",
             "npm exec prisma",
             "npm prisma",
             "npm rum prisma",


YOMXXX:

Expert lens: rewrite registry / JS tooling.

This is a narrow rule-table change: `bunx prisma ...` now routes through the existing `rtk prisma` handler, same as `npx prisma`, `pnpm prisma`, and related runners. There is no new Prisma output parser and no new command implementation.

Suggested review checks:
- Verify the regex only adds the `bunx` runner branch and does not broaden matching beyond Prisma.
- Verify `rewrite_prefixes` includes `bunx prisma`, so the rewrite strips the runner and leaves the Prisma subcommand unchanged.
- Confirm existing non-bun Prisma runners remain covered by the expanded tests.

KuSh:
Heads-up rather than a review: #988 (bun/deno runtime support) is close to landing and adds a broad `^bunx\s+` → `rtk bunx` rewrite rule.

That does **not** supersede this PR — routing `bunx prisma` to the prisma filter is strictly better than routing it to the generic bunx path, which is what #988 alone would do. Keeping this open.

It does mean the two rules overlap, and `classify_command` takes the first match in `RULES`, so once #988 merges this will need the prisma rule ordered ahead of the generic `bunx` one (and a test pinning that order, since it is otherwise silent). Worth a rebase after #988 to confirm the ordering still resolves the way you intend.

KuSh:
Thanks for this, and for the scoping notes that made the Prisma surface explicit. #988 (merged 2026-09-03) added Bun and Deno support, including a general `^bunx\s+` rewrite rule, so `bunx prisma migrate dev` is now rewritten to `rtk bunx prisma migrate dev` instead of streaming raw, which is the unfiltered behaviour issue #1401 reported. Adding `bunx` to the prisma rule no longer has an effect on top of that: classify_command keeps the last matching rule in the RULES array, and the bunx rule sits after the prisma one, so the rebased hunk would be a no-op and the new test would fail. There is still something worth doing here, which is routing `bunx prisma` to the dedicated prisma filter rather than the generic npm line filter, and that needs the bunx rule to defer to tool-specific rules; an issue or a narrow PR on that precedence question would be very welcome. Happy to reopen if you read the rule ordering differently.

PR #2006, opened 2026-05-21
fix(pytest): surface diagnostic context when no tests are collected

### LINKED ISSUE #1417: pytest collection fails with no-args invocation
The problem is rtk pytest with no args — that returned "No tests collected" while rtk pytest tests/test_cli_dispatch.py -v worked. There's an rtk bug in how it handles the no-args pytest invocation, probably in its cwd handling or in how it discovers tests when the path isn't explicit.

YOMXXX:
## Summary

\`rtk pytest\` (no args) collapsed pytest's collection output to a single terse line — \`Pytest: No tests collected\` — whenever the summary parser couldn't find a \`passed/failed/skipped\` count. Reporter (#1417) saw this on a real project where \`rtk pytest tests/test_cli_dispatch.py -v\` worked but \`rtk pytest\` did not, with no way to tell whether it was a cwd issue, a pytest config issue, or an RTK bug.

Two real situations both hit this:

1. **No tests in cwd** — pytest emits \`no tests ran in X.XXs\` instead of a counted summary line. The user was left guessing why nothing was found.
2. **Collection error** — \`ImportError\` in a conftest, missing fixture import, etc. pytest emits \`errors during collection\` + an \`ERROR\` block, then \`no tests ran\`. The actionable error was being thrown away.

## Fix

Add \`diagnose_no_tests\`: when the existing summary collapses to \`No tests collected\`, scan the raw pytest output for the lines that explain *why* — \`rootdir:\`, \`configfile:\`, \`testpaths:\`, \`collected N items\`, any \`ERROR …\` line, plus the \`no tests ran\` / \`no tests collected\` / \`errors during collection\` markers — and emit them verbatim under a header so the LLM and the user can act on them.

\`\`\`
Pytest: no tests collected
rootdir: /Users/x/proj
configfile: pyproject.toml
testpaths: tests, integration
collected 0 items
no tests ran in 0.05s
\`\`\`

Capped at 15 lines (a tee-style overflow hint would be wrong here — diagnostic lines are bounded in practice).

**Falls back to the terse default** when *no* recognisable marker is present (e.g. RTK's parser broke but tests actually ran), so we don't flood the output with arbitrary noise on a parser regression.

## Behaviour

| Scenario | Before | After |
|---|---|---|
| Empty cwd, \`rtk pytest\` | \`Pytest: No tests collected\` | + rootdir / configfile / testpaths / "no tests ran" |
| Collection \`ImportError\` | \`Pytest: No tests collected\` | + the actual \`ERROR collecting …\` line |
| Quiet-only \`no tests ran in 0.05s\` (no rootdir present) | terse | terse (markers fence keeps it from over-firing) |
| Tests pass | unchanged | unchanged |
| Tests fail | unchanged | unchanged |

## Test plan

- [x] \`test_filter_no_tests_verbose_surfaces_rootdir_and_configfile\` — rootdir / configfile / testpaths / collected count / "no tests ran" all reach the consumer.
- [x] \`test_filter_no_tests_collection_error_surfaced\` — ImportError-during-collection scenario shows the \`ERROR\` line.
- [x] \`test_filter_no_tests_quiet_mode_falls_back_to_terse\` — bare \`no tests ran in 0.05s\` (no rootdir to anchor on) keeps the original short verdict.
- [x] Pre-existing \`test_filter_pytest_no_tests\` updated to assert the new enriched output while keeping the same verdict header for grep-ability.
- [x] \`cargo test --bin rtk -- --test-threads=8\` → **1912 passed** (was 1909).

## Notes

- The original reporter's underlying confusion (was the cwd wrong? was it pytest config? was it RTK?) is now answerable straight from the rtk output: rootdir tells you where pytest looked, configfile tells you which config it used, and testpaths tells you which subtree it scanned.
- Doesn't touch the \`-q\` defaults or the runner's cwd — those are still pytest's standard behaviour.

Fixes #1417

### DIFF
--- a/src/cmds/python/pytest_cmd.rs
+++ b/src/cmds/python/pytest_cmd.rs
@@ -151,7 +151,69 @@ pub(crate) fn filter_pytest_output(output: &str) -> String {
     }
 
     // Build compact output
-    build_pytest_summary(&summary_line, &test_files, &failures, &xfail_lines)
+    let summary = build_pytest_summary(&summary_line, &test_files, &failures, &xfail_lines);
+    // #1417: when the summary collapses to the terse "No tests collected"
+    // line (pytest -q's "no tests ran" never matched any of our summary
+    // regexes), surface pytest's own collection-time diagnostics — rootdir,
+    // configfile, testpaths, collected count, collection errors — so the
+    // user can see *why* pytest found nothing instead of guessing whether
+    // it was a cwd / config / RTK bug.
+    if summary == "Pytest: No tests collected" {
+        if let Some(diag) = diagnose_no_tests(output) {
+            return diag;
+        }
+    }
+    summary
+}
+
+/// Extracts pytest's collection-time diagnostic lines (rootdir, configfile,
+/// testpaths, collected count, collection errors, the `no tests ran` summary)
+/// from the raw pytest output. Returns `None` if no recognisable "no tests"
+/// marker is present — in that case callers should keep the terse default
+/// summary rather than echoing arbitrary noise.
+fn diagnose_no_tests(raw: &str) -> Option<String> {
+    let has_marker = raw.contains("no tests ran")
+        || raw.contains("no tests collected")
+        || raw.contains("ERROR collecting")
+        || raw.contains("errors during collection");
+    if !has_marker {
+        return None;
+    }
+
+    let mut diag_lines: Vec<&str> = Vec::new();
+    for line in raw.lines() {
+        let t = line.trim();
+        if t.starts_with("rootdir:")
+            || t.starts_with("configfile:")
+            || t.starts_with("testpaths:")
+            || t.starts_with("collected ")
+            || t.contains("no tests ran")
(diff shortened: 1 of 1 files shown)


YOMXXX:

Expert lens: Python test-runner output / diagnostic preservation.

This only changes the `No tests collected` branch. Passing and failing test summaries are intended to remain unchanged. When pytest collected zero tests or hit collection-time errors, RTK now preserves bounded diagnostic lines such as `rootdir`, `configfile`, `testpaths`, `collected 0 items`, and `ERROR ...`, so the user/agent can tell whether this is cwd/config/import failure rather than an RTK parser bug.

Suggested review checks:
- Confirm normal pass/fail output still uses the existing summary path.
- Confirm the diagnostic branch is bounded (`MAX_DIAG_LINES`) and does not dump arbitrary raw pytest output.
- Confirm collection errors surface the actionable `ERROR` lines instead of collapsing to a generic verdict.

PR #2014, opened 2026-05-21
fix(find): passthrough safe native predicates

danmunoz:
## Summary

- Allow `rtk find` to automatically passthrough unsupported read-only native `find` predicates instead of failing.
- Preserve RTK's compact filtered output for the currently supported simple cases (`-name`, `-iname`, `-type`, `-maxdepth`, and RTK syntax).
- Keep side-effecting native `find` actions out of RTK execution (`-delete`, `-exec`, `-execdir`, `-ok`, `-okdir`), so this PR does not introduce a new destructive execution path.

## Why

I kept running into native `find` expressions that RTK could not parse, especially compound predicates and output-shaping flags such as `-not`, `-o`, `-path`, and `-print0`.

Today those invocations fail first, then the agent has to rerun the equivalent raw command. That causes double calls and extra context output, which breaks RTK's core premise of saving tokens and keeping agent command execution efficient.

This change makes RTK behave more transparently for safe native `find` expressions: if RTK cannot faithfully compact the expression, it proxies the command through native `find` instead of forcing the user or agent to retry manually.

## Behavior after this change

- `rtk find . -name '*.rs' -type f -maxdepth 2` still uses RTK's compact filtered output.
- `rtk find '*.rs' src -m 5` still uses RTK syntax and compact filtered output.
- `rtk find . -name '*.rs' -not -path '*/target/*'` passes through to native `find` and preserves native semantics.
- `rtk find . -name '*.rs' -print0` passes through to native `find`, preserving NUL-delimited output for downstream consumers.
- `rtk find src tests -name '*.rs'` passes through to native `find`, preserving multi-start-path semantics.
- `rtk find . -name '*.tmp' -delete` remains blocked by RTK; users can run native `find` directly when they intentionally want side effects.

## Non-destructive safety note

This PR is intentionally non-destructive:

- It does not add shell invocation or evaluate embedded commands.
- It does not add dependencies.
- It does not change behavior for existing compact, read-only RTK `find` queries.
- It avoids silently executing destructive native `find` actions through RTK. Those actions already failed before this change; this PR narrows the failure behavior to side-effecting actions while allowing read-only unsupported predicates to work via passthrough.

## Test plan

- [x] Unit tests added/updated for changed code
- [x] Snapshot tests reviewed (`cargo insta review`) - not applicable; no snapshots changed
- [x] Token savings >=60% verified - not applicable for passthrough paths; the improvement avoids the previous fail-then-rerun double call
- [x] Any truncated list has a recovery hint (`force_tee_tail_hint` or `force_tee_hint`) - not applicable; no new truncation path
- [x] Edge cases covered
- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` passes

```bash
target/debug/rtk find "$tmpdir" -name '*.rs' -type f -maxdepth 3
target/debug/rtk find "$tmpdir" -name '*.rs' -not -path '*/tests/*'
target/debug/rtk find "$tmpdir" -name '*.rs' -print0 | wc -c
target/debug/rtk find "$tmpdir" -name '*.tmp' -delete
```

### DIFF
--- a/src/cmds/system/README.md
+++ b/src/cmds/system/README.md
@@ -6,6 +6,7 @@
 
 - `read.rs` uses `core/filter` for language-aware code stripping (FilterLevel: none/minimal/aggressive)
 - `grep_cmd.rs` reads `core/config` for `limits.grep_max_results` and `limits.grep_max_per_file`. Format-altering flags (`-c`, `-l`, `-L`, `-o`, `-Z`) bypass RTK filtering and run raw.
+- `find_cmd.rs` filters simple RTK/native queries (`-name`, `-iname`, `-type`, `-maxdepth`) and passes unsupported read-only native predicates through to `find` so compound expressions keep native semantics. Side-effecting native actions (`-delete`, `-exec`, `-execdir`, `-ok`, `-okdir`) are left for users to run with `find` directly.
 - `local_llm.rs` (`rtk smart`) uses `core/filter` for heuristic file summarization
 - `format_cmd.rs` is a cross-ecosystem dispatcher: auto-detects and routes to `prettier_cmd` or `ruff_cmd` (black is handled inline, not as a separate module)
 
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -4,6 +4,7 @@ use crate::core::tracking;
 use anyhow::{Context, Result};
 use ignore::WalkBuilder;
 use std::collections::HashMap;
+use std::ffi::OsString;
 use std::path::Path;
 
 /// Match a filename against a glob pattern (supports `*` and `?`).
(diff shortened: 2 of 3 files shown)


KuSh:
Thanks for this. #3603 (merged 2026-08-26) reworked rtk find around a dispatch on find's own grammar, so unmodeled predicates no longer fail: expressions like -not, -o, -path and multiple start paths are handed to real find and only its output is compressed, while -name/-iname/-type/-maxdepth keep the compact path, and I confirmed each of your examples against the current develop binary. One difference worth flagging: develop passes -delete and -exec verbatim to find instead of blocking them, following the rule that rtk never refuses a command the user could run directly. Closing as covered, but please comment or reopen if a case you had in mind still fails.

PR #2015, opened 2026-05-21
fix: honor explicit -n N limit for git log on merge commits

Kartalops:


aeppling:
Hey, pretty straightforward

aeppling (approved):

PR #2016, opened 2026-05-21
fix(git/log): stop dropping merge commits when user pins a selection

### LINKED ISSUE #2009: git log filter returns wrong SHA for merge commits (resolves to second parent)
## What's wrong

When rtk rewrites `git log`, it hands back the wrong commit SHA if the target is a merge commit. Ask for a merge commit `M` by any ref — `HEAD`, an annotated tag, or its own SHA — and you get one of `M`'s parents instead. Plain (non-merge) commits are fine.

## Setup

- rtk 0.40.0 (Homebrew bottle)
- macOS arm64 (Darwin 25.5.0)
- Running through a Claude Code hook that rewrites `git ...` to `rtk git ...`

## How to reproduce

Take a repo with a `--no-ff` merge commit `M` — parents `[P1, P2]`, where `P2` is the branch that got merged in:

```
$ git log -1 --format='%H' HEAD       # HEAD is M
P2                                     # wrong, should be M
$ git log -1 --format='%H' <M-sha>    # asked for M directly
P2                                     # wrong
$ git log -1 --format='%H' <tag>      # annotated tag pointing at M
P2                                     # wrong
$ git log -1 --format='%H' <P1-sha>   # a normal commit
P1                                     # fine
```

Run the same thing through `rtk proxy` and it comes back right:

```
$ rtk proxy git log -1 --format='%H' HEAD
M
```

`git rev-parse`, `git rev-list --parents`, and `git tag -n1` all report the right thing — it's only `git log` that's off. `git log --graph --oneline` has the same problem: the merge commit and the branch topology just don't show up.

## Why it matters

`git log` is usually the first thing you reach for to check a merge or a release. When it quietly returns the wrong SHA, it looks like your `--no-ff` merge fast-forwarded or never happened — easy to make a bad release call off that. And it's not an obvious error either; you get a real, recent commit back, so nothing tips you off that it's wrong.

## Guess at the cause

Looks like the `git log` filter treats merge commits as noise and drops them, so when you pass `-1` or an explicit ref it falls through to the next commit (the second parent) instead of the one you asked for.

Two ways out, roughly: never drop a commit that was named explicitly by ref or pinned with `-n`/`-1`. Or skip the merge-commit filtering entirely whenever `--format`, `-n`/`-1`, or an explicit revision shows up in the args.

YOMXXX:
## Summary

\`rtk git log -1 HEAD\` on a repo whose HEAD is a \`--no-ff\` merge commit M (parents \`[P1, P2]\`) returns **P2** instead of M. Same for explicit SHAs, annotated tags, \`--graph\`, and \`--graph --oneline\` — every shape that names a specific commit gets the second parent of the merge instead of the merge itself.

The caller can't tell the result is wrong, because P2 is a real recent commit. A release script that asks \"what's the SHA of the merge I just landed?\" gets a value that looks plausible and is wrong.

\`rtk proxy git log -1 HEAD\` returns M, confirming \`git\` itself is fine — the bug is in RTK's filter.

## Root cause

\`run_log\` in \`src/cmds/git/git.rs\` unconditionally injects \`--no-merges\` as a token-saving default. When the user passes \`-1\`, git walks past the dropped merge to the next non-merge commit (the second parent, because \`git log\` defaults to a topology-aware traversal), so the precise pin silently shifts the selection.

\`--no-merges\` is the right default only when RTK is fully driving output — bare \`git log\` with no flags, no ref, no format. As soon as the user signals \"I'm picking\", the default has to stand down.

## Fix

Extract a pure helper \`should_add_no_merges_default(args, has_format, has_limit)\`. It returns \`true\` only when **none** of these signals are present:

- \`--format\` / \`--pretty\` / \`--oneline\`  &nbsp; *(caller controls output shape)*
- \`-N\` / \`-n N\` / \`--max-count=N\`  &nbsp; *(caller picked a precise window)*
- \`--graph\`  &nbsp; *(caller wants topology — needs merges)*
- any positional argument  &nbsp; *(ref / SHA / tag / pathspec)*
- explicit \`--merges\` / \`--min-parents=2\`  &nbsp; *(caller already opted in)*

\`has_format\` and \`has_limit\` are passed in from the caller's existing scan so we don't re-walk the args list.

## Behaviour matrix

| invocation                              | \`--no-merges\` injected? | Notes |
|-----------------------------------------|--------------------------|-------|
| \`rtk git log\`                         | ✅ yes                   | RTK token-saving default (unchanged)
| \`rtk git log -1 HEAD\`                 | ❌ no                    | #2009 fix
| \`rtk git log -1\`                      | ❌ no                    | precise pin
| \`rtk git log --format=%H HEAD\`        | ❌ no                    | format + ref
| \`rtk git log <sha>\`                   | ❌ no                    | positional = ref
| \`rtk git log --graph\`                 | ❌ no                    | topology
| \`rtk git log --graph --oneline\`       | ❌ no                    | topology + compact
| \`rtk git log --merges\`                | ❌ no                    | explicit opt-in (pre-existing)
| \`rtk git log --min-parents=2\`         | ❌ no                    | explicit opt-in (pre-existing)
| \`rtk git log --author=foo\`            | ✅ yes                   | filter ≠ pin
| \`rtk git log --max-count=5\`           | ❌ no                    | explicit window

## Test plan

11 new tests pin each shape in the table above:

- \`test_no_merges_added_for_bare_git_log\` — token-saving default stays for bare invocation.
- \`test_no_merges_skipped_with_explicit_limit_one\` + \`_only\` — \`-1 HEAD\` and bare \`-1\` (headline reproduction).
- \`test_no_merges_skipped_with_format_flag_and_ref\` + \`_explicit_sha\` + \`_graph_flag\` + \`_graph_and_oneline\` — every other shape the issue body called out.
- \`test_no_merges_skipped_with_explicit_merges\` + \`_min_parents\` — pre-existing opt-in stays correct.
- \`test_no_merges_still_added_with_author_filter\` — documents the deliberate narrow scope: filter flags that don't pick a single commit still benefit from the token-saving default.
- \`test_no_merges_skipped_with_max_count_arg\` — short-circuit via \`has_limit\`.

\`cargo fmt --all\` clean / \`cargo clippy --all-targets\` zero warnings / \`cargo test --bin rtk -- --test-threads=8\` → **1920 passed** (was 1909).

## Scope notes

- The pure helper is unit-tested directly. The \`run_log\` integration path is unchanged otherwise, so existing \`filter_log_output\` / \`parse_user_limit\` tests stay green and document that the rest of the log filter is intact.
- The \`--author=foo\` case is deliberately left in the default-merges path: an author filter doesn't pick a single commit, so \`-N\` summary behaviour with the token-saving default is still the right shape. If usage data later shows the opposite, that's a follow-up.

Fixes #2009

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -520,12 +520,18 @@ fn run_log(
         (10, false)
     };
 
-    // Only add --no-merges if user didn't explicitly request merge commits
-    let wants_merges = args
-        .iter()
-        .any(|arg| arg == "--merges" || arg == "--min-parents=2" || arg == "--no-merges");
-    // Don't add --no-merges if user explicitly requested merges or an exact count (-n N / --max-count)
-    if !wants_merges && !has_limit_flag {
+    // Add --no-merges only when RTK is fully driving defaults. When the user
+    // pins to specific commits (--format / -N / --graph / explicit ref or
+    // pathspec), --no-merges silently shifts the selection — e.g.
+    // `git log -1 HEAD` on a merge commit M resolves to M's second parent
+    // because `-1` walks past the dropped merge to the next non-merge
+    // commit. See #2009.
+    //
+    // Extends the partial fix in #2015 (which only honoured `-n N` /
+    // `--max-count`) to also preserve merges for `-1 HEAD`, `<sha>`,
+    // `--graph`, `--format=%H HEAD`, etc. — the remaining shapes from
+    // #2009's reproduction.
+    if should_add_no_merges_default(args, has_format_flag, has_limit_flag) {
         cmd.arg("--no-merges");
     }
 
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for this, and for the careful test list. The exact reproduction from #2009, `git log -1 HEAD`, was fixed by commit b3adcb18 on 2026-05-21, and git log argument handling has since moved onto the shared argument tokenizer, so run_log decides on typed tokens rather than string prefixes and there is no helper for these tests to target. Three of the shapes you identified do still inject `--no-merges` on develop: a positional ref such as `rtk git log <sha>`, `--graph`, and a format flag with no limit like `rtk git log --format=%H HEAD`. Those are worth an issue or a narrow PR extending the token check in src/cmds/git/git.rs. If I missed something here, please comment or reopen.

PR #2049, opened 2026-05-23
fix(stream): replace map_while(Result::ok) with from_utf8_lossy to prevent silent output truncation

rosschurchill:
The 6 `lines().map_while(Result::ok)` sites in `stream.rs` silently terminate iteration on the first non-UTF-8 byte. A single bad byte in tool output means everything after it is silently dropped — the LLM never sees it, with no warning.

Replaced all 6 sites with `BufRead::read_until` + `String::from_utf8_lossy`, which replaces invalid bytes with `U+FFFD` and continues. The LLM sees a replacement character instead of losing the rest of the output.

Includes a regression test with a byte sequence containing `\xFF\xFE` mid-stream to verify all lines are preserved.

Free to use as-is.

### DIFF
--- a/src/core/stream.rs
+++ b/src/core/stream.rs
@@ -298,10 +298,14 @@ pub fn run_streaming(
             Some(std::thread::spawn(move || {
                 let mut writer = BufWriter::new(child_stdin);
                 let stdin_handle = io::stdin();
-                for line in BufReader::new(stdin_handle.lock())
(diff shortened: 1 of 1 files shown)


KuSh:
Hi @rosschurchill, thanks so much for this contribution! It looks like #2997 just got merged with similar work, so we're going to close this one out.

Issue #401, opened 2026-03-07
Pi support


maxim:
Would be great to add [pi-coding-agent](https://pi.dev) integration.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟢 **Priority** | `low` |
| 🎯 **Confidence** | 76% |

### Summary

Vague request to add pi-coding-agent (pi.dev) integration; needs more detail on what specific commands or behaviors should be supported before actionable.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

jtoyshopify:
I would like official support as well!

0serna:
+1

borrougagnou:
please ! we need a reviewer here !
https://github.com/rtk-ai/rtk/pull/1741

Jheysson:
Would love it

mrclrchtr:
In case someone is interested, I created https://github.com/mrclrchtr/supi/blob/main/packages/supi-rtk/README.md 
There are some commands, that are not working very well with PI or the LLMs. I added some safety guards for these cases. You can find it also here https://pi.dev/packages/@mrclrchtr/supi-rtk?name=mrclrchtr

aeppling:
Released in 0.42 :)

PR #1741, opened 2026-05-06
feat(hook): add pi support

gitbluf:
## Summary

 
 - Add Pi coding agent support via a thin TypeScript extension that delegates rewrites to rtk rewrite
 - Wire Pi into rtk init for install/uninstall, global and local scopes, plus Pi-specific docs
 - Add Pi tests and supporting hook/init plumbing

 ## Test plan

 
 - cargo fmt --all && cargo clippy --all-targets && cargo test
 - Manual testing: rtk <command> output inspected

Issue #401 

I also saw PR #750 is open, but it seems abandoned.

PR #2055, opened 2026-05-23
Next Release

aeppling:
## Features

- feat(hook): add pi support https://github.com/rtk-ai/rtk/pull/1741 closes https://github.com/rtk-ai/rtk/issues/401

## Fix

- fix: honor explicit -n N limit for git log on merge commits https://github.com/rtk-ai/rtk/pull/2015


pszymkowiak (approved):
**Reviewed and tested on a fresh build of the PR head — good to merge.** ✅

**Quality gate:** `cargo fmt --check` clean · `cargo clippy --all-targets` no issues · full suite **1928 passed / 0 failed**.

**Pi integration (#1741):** verified install/uninstall on real commands — local + global (dry-run, real, idempotent), `PI_CODING_AGENT_DIR` override, old `--pi` flag correctly rejected in favor of `--agent pi`, installed `rtk.ts` byte-identical to source, and the TS extension's exit-code contract (3 = rewrite, 1 = passthrough, fail-open) matches `rtk rewrite`.

**git log -n fix (#2015):** `rtk git log -n 8` now returns the exact same commits as raw `git log -n 8`, merges included, while the default `rtk git log` still strips merges (compaction preserved). `-N`, `-n N`, `--max-count=N` / `--max-count N` all honored.

Two non-blocking notes:
- Glued `-n5` form isn't detected as a limit flag, so it still strips merges (pre-existing, uncommon — worth a follow-up).
- The two-dot diff shows an apparent `0.41.0 → 0.40.0` change, but it's a false alarm: develop never touched the version vs. merge-base, so the 3-way merge keeps `0.41.0` (release-please re-bumps post-merge).

Good to merge.

KuSh (approved):

PR #2061, opened 2026-05-24
fix/ls 1 flag 2058

### LINKED ISSUE #2058: `rtk ls -1 path/` Always Return `(empty)`
My AI agent confused because when calling `rtk ls -1 path/` it always show `(empty)`. It should show list of files

pen-pal:
## Summary

- Fixes #2058: 
- `rtk ls -1 path/` returned `(empty)` on BSD `/bin/ls` because `ls -la -1` honors the last format flag, producing filename-only output the parser can't decode.
- Strips display-format flags (`-1`, `-C`, `-m`, `-x`) alongside the already-enforced `-l`/`-a`/`-h` before invoking ls. RTK's compact output is one-per-line anyway, so users get the layout they asked for.
- Split into 3 commits: refactor (extract `build_ls_args`), fix (+ primary regression test), test (14 more edge cases covering placement, sibling flags, compound splits, mixed groups, repetition).

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — 1943 pass, 0 fail, clippy clean
- [x] Manual testing: `PATH=/bin:/usr/bin rtk ls -1 /tmp/testdir/` returns compact output (`a.txt 0B` ...), not `(empty)` or raw filenames


YOMXXX (approved):
Reviewed the `ls -1` fix. The `build_ls_args` extraction is clean, the short-format flags that override `-l` on BSD are stripped before execution, and the regression coverage around `-1`, compound flags, and repeated format flags is solid.

One non-blocking follow-up: GNU `ls --format=single-column`, `--format=commas`, etc. can express the same class of display-format override through a long option and still pass through today. That can be tracked separately; this PR fixes the reported short-flag regression.

PR #2068, opened 2026-05-24
docs: add Codex PreToolUse adapter guide

xuegaoge:
## Summary

- Adds `docs/usage/CODEX_PRETOOLUSE_ADAPTER.md`, a focused guide for installing an optional Codex Desktop/CLI `PreToolUse` adapter.
- Keeps `hooks/codex/README.md` as the hook-directory overview and links it to the new usage guide.
- Provides an idempotent quick-install snippet that reuses an existing RTK install when `rtk gain` works, and installs RTK only when missing or incorrect.

## Why

`rtk init --codex` currently gives Codex prompt-level awareness through `AGENTS.md` and `RTK.md`, but it does not provide a native Codex hook. The documented adapter lets Codex rewrite Bash commands at tool-call time while preserving normal pass-through behavior when RTK has no rewrite.

## Validation

- Ran `git diff --check HEAD~1..HEAD`.
- Executed the quick-install snippet twice against a temporary `CODEX_HOME` and verified it remained idempotent with one `PreToolUse` entry.
- Verified the generated hook compiles with `python3 -m py_compile`.
- Simulated a Codex Bash payload and confirmed `git status` rewrites to `rtk git status`.
- Simulated a non-Bash payload and confirmed the hook emits zero bytes.

### DIFF
--- a/docs/usage/CODEX_PRETOOLUSE_ADAPTER.md
+++ b/docs/usage/CODEX_PRETOOLUSE_ADAPTER.md
@@ -1,0 +1,211 @@
+# Codex PreToolUse Adapter
+
+This guide adds an optional Codex Desktop/CLI `PreToolUse` hook that rewrites
+Bash commands through RTK before Codex executes them.
+
+## Why use this
+
+`rtk init --codex` writes lightweight RTK awareness into Codex's
+`AGENTS.md`/`RTK.md` context. That is useful, but it still depends on the model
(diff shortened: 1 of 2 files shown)


xuegaoge:
Closing in favor of #2069, which combines the native Codex plugin approach from #2063 with the safer tool_input-preserving behavior and user-facing setup guide from this PR.

PR #2069, opened 2026-05-24
feat(codex): add integrated PreToolUse plugin support

xuegaoge:
## Summary

This PR combines the stronger native Codex plugin direction from #2063 with the safer behavior and user-facing setup guidance from #2068.

- Adds the RTK Codex plugin package and `rtk hook codex` lifecycle integration from #2063.
- Preserves Codex's original Bash `tool_input` when rewriting commands, replacing only `command` so fields such as `timeout` and `description` are not dropped.
- Adds `docs/usage/CODEX_PRETOOLUSE_ADAPTER.md` with first-install guidance that reuses an existing RTK install when `rtk gain` works and installs RTK only when missing or incorrect.
- Updates Codex docs to describe the plugin hook, Windows command handling, hook trust, and `allow` + `updatedInput` behavior.

## Why

The previous prompt-only Codex setup depends on the model remembering to prefix shell commands with `rtk`. The native Codex plugin hook moves that rewrite to the tool-call layer. This PR keeps that product-level integration while carrying over the safer adapter behavior tested in #2068: preserve the rest of Codex's `tool_input` and provide a clear usage guide for new and already-installed RTK users.

Credit: this builds on the implementation direction from #2063 and folds in the manual-adapter safeguards from #2068.

## Validation

- `cargo test test_codex_rewrite_preserves_tool_input_fields -- --nocapture` initially failed before the preservation fix.
- `cargo test test_codex_rewrite_preserves_tool_input_fields -- --nocapture` passed after the fix.
- `cargo test codex -- --nocapture` passed: 28 passed.
- `cargo fmt --all -- --check` passed.
- `cargo clippy --all-targets -- -D warnings` passed.
- `cargo test --all` passed: 1947 passed, 6 ignored.
- `git diff --check HEAD~3..HEAD` passed.
- Ran `cargo run -- init -g --codex` with isolated `HOME` and `CODEX_HOME`; verified plugin package, marketplace entry, and `hooks/hooks.json` were written and `rtk init --show --codex` detected them.
- Verified the quick setup block in `docs/usage/CODEX_PRETOOLUSE_ADAPTER.md` with `bash -n`.


KuSh:

PR #2075, opened 2026-05-24
feat(init): add PowerShell hook for Claude Code on Windows

J-Pster:
## Problem

On Windows, Claude Code routes most shell invocations through its **PowerShell** tool rather than the **Bash** tool. The existing `rtk init -g` only registers a `PreToolUse` hook with `"matcher": "Bash"`, so Windows users get **zero automatic rewriting** for the majority of their shell calls.

## Solution

Register **both** matchers when patching `settings.json`:

```json
{ "matcher": "Bash",       "hooks": [{ "type": "command", "command": "rtk hook claude" }] }
{ "matcher": "PowerShell", "hooks": [{ "type": "command", "command": "rtk hook claude" }] }
```

The `rtk hook claude` handler already reads the tool-input JSON from stdin and pattern-matches on the `command` field — it doesn't inspect `tool_name` — so no changes to the hook runtime are required.

### Verified locally

```powershell
# Simulates the JSON Claude Code sends for a PowerShell tool call
echo '{"tool_name":"PowerShell","tool_input":{"command":"git status"}}' | rtk hook claude
# → {"hookSpecificOutput":{"updatedInput":{"command":"rtk git status"},"permissionDecision":"allow",...}}
```

Commands that RTK doesn't recognise (e.g. `Get-Process`) pass through with exit 0 and no output — unchanged behaviour.

## Changes

- **`insert_hook_entry()`** — pushes a second entry with `"matcher": "PowerShell"` after the existing `"Bash"` one
- **`print_manual_instructions()`** — shows both matchers in the manual-step output so users who decline auto-patch know to add both
- **Tests** — updated `test_insert_hook_entry_empty_root` and `test_insert_hook_entry_preserves_existing` to assert the new array lengths and verify both matchers are present

## Impact

- macOS / Linux: no behaviour change (both matchers are registered but Claude Code only fires the Bash one)
- Windows (native): full automatic rewriting now works for PowerShell tool calls
- Uninstall (`--uninstall`): `remove_hook_from_json` matches on command value, not matcher — both entries are removed correctly with no changes needed

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -500,12 +500,20 @@ fn prompt_telemetry_consent() -> Result<()> {
 fn print_manual_instructions(hook_command: &str, include_opencode: bool) {
     println!("\n  MANUAL STEP: Add this to ~/.claude/settings.json:");
     println!("  {{");
-    println!("    \"hooks\": {{ \"PreToolUse\": [{{");
-    println!("      \"matcher\": \"Bash\",");
-    println!("      \"hooks\": [{{ \"type\": \"command\",");
-    println!("        \"command\": \"{}\"", hook_command);
-    println!("      }}]");
-    println!("    }}]}}");
+    println!("    \"hooks\": {{ \"PreToolUse\": [");
+    println!("      {{");
+    println!("        \"matcher\": \"Bash\",");
+    println!("        \"hooks\": [{{ \"type\": \"command\",");
+    println!("          \"command\": \"{}\"", hook_command);
+    println!("        }}]");
+    println!("      }},");
+    println!("      {{");
+    println!("        \"matcher\": \"PowerShell\",");
+    println!("        \"hooks\": [{{ \"type\": \"command\",");
+    println!("          \"command\": \"{}\"", hook_command);
+    println!("        }}]");
+    println!("      }}");
+    println!("    ]}}");
     println!("  }}");
     if include_opencode {
         println!("\n  Then restart Claude Code and OpenCode. Test with: git status\n");
@@ -1074,6 +1082,13 @@ fn insert_hook_entry(root: &mut serde_json::Value, hook_command: &str) -> Result
 
     pre_tool_use.push(serde_json::json!({
         "matcher": "Bash",
+        "hooks": [{
+            "type": "command",
+            "command": hook_command
+        }]
+    }));
+    pre_tool_use.push(serde_json::json!({
+        "matcher": "PowerShell",
         "hooks": [{
             "type": "command",
             "command": hook_command
@@ -5064,10 +5079,17 @@ mod tests {
             .is_some());
 
         let pre_tool_use = json_content["hooks"]["PreToolUse"].as_array().unwrap();
-        assert_eq!(pre_tool_use.len(), 1);
+        assert_eq!(pre_tool_use.len(), 2);
 
-        let command = pre_tool_use[0]["hooks"][0]["command"].as_str().unwrap();
-        assert_eq!(command, hook_command);
+        let bash_matcher = pre_tool_use[0]["matcher"].as_str().unwrap();
+        assert_eq!(bash_matcher, "Bash");
+        let bash_command = pre_tool_use[0]["hooks"][0]["command"].as_str().unwrap();
+        assert_eq!(bash_command, hook_command);
+
+        let ps_matcher = pre_tool_use[1]["matcher"].as_str().unwrap();
+        assert_eq!(ps_matcher, "PowerShell");
+        let ps_command = pre_tool_use[1]["hooks"][0]["command"].as_str().unwrap();
+        assert_eq!(ps_command, hook_command);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for this — Windows support is genuinely wanted, and the JSON you're emitting is correct. But registering the matcher is the last step of this feature, not the first, and merging it as-is would ship broken behaviour to Windows users.

## First: as written, this is a no-op for every existing rtk user

`patch_settings_json_command` short-circuits on `hook_already_present` (src/hooks/init.rs:981), which returns true if *any* `PreToolUse` entry carries the rtk command — and the existing `Bash` entry does. So anyone who has already run `rtk init -g` (i.e. every current Windows user, the audience for this PR) re-runs it, gets `PatchResult::AlreadyPresent`, and never receives the PowerShell matcher. `rtk init --status` still cheerfully prints `[ok] settings.json: RTK hook configured`. Only a virgin `settings.json`, or a full uninstall + reinstall, picks it up.

`insert_hook_entry` would need to be per-matcher idempotent (push each matcher only if absent), and `hook_already_present` would need to require both. Worth knowing before anything else: the feature as merged would reach almost nobody.

## And when it does fire, the pipeline behind it is bash

The rewrite pipeline is a **bash** model end to end, and nothing in it is conditioned on which shell the command came from. `process_claude_payload` (src/hooks/hook_cmd.rs:574) reads `/tool_input/command` and rewrites blind — unlike the Gemini (:403), vibe (:466) and Copilot (:125) paths, it never inspects `tool_name`. So the moment the `PowerShell` matcher fires, PowerShell source text is fed to a bash lexer.

Concretely, on this branch:

**1. Multiline PowerShell is silently corrupted.** `collapse_line_continuations` (src/discover/registry.rs:587) treats `\` + newline as a continuation. In PowerShell `\` is a path separator:

```
in:  git status\ncd C:\repo\\ngit log -3
out: rtk git status\ncd C:\repo git log -3
```

The path loses its separator, `git log -3` becomes arguments to `cd`, and never runs. `cd C:\Users\me\proj\` is ordinary Windows usage, and the hook returns `updatedInput` — so the corrupted command is what actually executes.

**2. PowerShell aliases get hijacked into POSIX tools that don't exist on Windows.** Verified rewrites:

```
ps -Name git                              -> rtk ps -Name git            (Get-Process)
diff a.txt b.txt                          -> rtk diff a.txt b.txt        (Compare-Object)
ls -Recurse -Filter *.rs                  -> rtk ls -Recurse -Filter     (Get-ChildItem)
curl -Uri https://example.com -Method GET -> rtk curl -Uri ...           (Invoke-WebRequest)
wget https://example.com/x -OutFile x     -> rtk wget ... -OutFile x     (Invoke-WebRequest)
tree /F                                   -> rtk tree /F
find . -Name x                            -> rtk find . -Name x
```

`ps.exe` / `diff.exe` / `wget.exe` do not exist on Windows, and the filters parse GNU-style flags (e.g. src/cmds/system/ls.rs:22 handles `-l/-a/-g/-n/-o` then shells out to `ls`). Each of these turns a working PowerShell call into a failure.

**3. Permission rules are read only from the `Bash(` scope.** `append_bash_rules` (src/hooks/permissions.rs:164) keeps only `Bash(...)`-prefixed rules. A user's PowerShell-scoped **deny** is therefore invisible to rtk, while an unrelated `Bash(...)` allow that matches the command text makes the hook emit `permissionDecision: "allow"` — skipping the permission prompt entirely. That is the bypass the Deny-first ordering in `check_command_with_rules` exists to prevent.

**4. Most idiomatic PowerShell wouldn't be rewritten at all.** `contains_substitution` (src/discover/lexer.rs:328) returns true on any unquoted or double-quoted backtick, treating it as bash command substitution. The backtick is PowerShell's *escape* character and line-continuation, so backtick-quoted and backtick-continued commands all defer with no rewrite. That's fail-safe rather than harmful, but it means real-world savings on Windows would be far below what the PR describes.

**5. One unverified assumption.** The PR assumes the PowerShell tool's input schema uses a `command` field. If it doesn't, `pointer("/tool_input/command")` returns `None` -> `PayloadAction::Ignore`, and the feature is a silent no-op for a second reason. The description cites a hand-crafted JSON echo, not a captured payload.

## Where this leaves us

Closing rather than requesting changes, because the fix isn't in this file. Real PowerShell support needs, roughly:

- `tool_name` captured in `process_claude_payload` and threaded down into `registry::rewrite_command` and `permissions::check_command_for`
- a shell discriminator honoured by the lexer (backtick = escape, `\` = path separator, backtick + newline = continuation)
- a PowerShell rule set, or an alias denylist, so `ps` / `ls` / `diff` / `curl` / `wget` / `tree` / `find` aren't captured
- permission rules read from the PowerShell scope
- per-matcher idempotency in `insert_hook_entry` / `hook_already_present`
- fixture-driven tests built on a real captured PowerShell payload

That's a substantial piece of work, and we'd be glad to review it as a series of smaller PRs — starting with threading `tool_name` through, which is useful on its own.

Uninstall, for what it's worth, is fine: `remove_hook_from_json` matches on the command value, so both entries get removed.

PR #2078, opened 2026-05-24
fix: remove yadm from git rewrite rules

### LINKED ISSUE #2077: yadm incorrectly rewritten to rtk git (yadm is not a git alias)

## Bug

`rtk rewrite "yadm status"` outputs `rtk git status`, which fails because `yadm` is **not** a regular git alias.

## What is yadm

[yadm](https://yadm.io/) is a dotfile manager that wraps git with a separate bare repository (`~/.local/share/yadm/repo.git`) and `--work-tree=$HOME`. Running `rtk git status` in an arbitrary working directory fails with "Not a git repository" because there is no `.git` in the CWD — yadm manages this transparently via its own git config.

## Root cause

In `src/discover/rules.rs`, the first rule includes `yadm` in both the regex pattern and `rewrite_prefixes`:

```rust
RtkRule {
    pattern: r"^(?:git|yadm)\s+...",
    rtk_cmd: "rtk git",
    rewrite_prefixes: &["git", "yadm"],
    ...
}
```

## Expected behavior

`rtk rewrite "yadm status"` should exit 1 (no rewrite), passing the command through unchanged.

## Workaround

Adding `"^yadm(?:$| )"` to `hooks.exclude_commands` in `~/.config/rtk/config.toml` works, but the proper fix is removing `yadm` from `rewrite_prefixes` and the regex pattern.

## Environment

- rtk 0.42.0
- Arch Linux, Sway/Wayland
- yadm 3.x managing dotfiles via OpenCode plugin

paolomainardi:

## Problem

`rtk rewrite "yadm status"` outputs `rtk git status`, which fails because `yadm` is **not** a git alias.

[yadm](https://yadm.io/) is a dotfile manager that wraps git with a separate bare repository (`~/.local/share/yadm/repo.git`) and `--work-tree=$HOME`). Running `rtk git status` in an arbitrary CWD fails with "Not a git repository" since there's no `.git` in the working directory — yadm manages this transparently.

## Fix

- Remove `yadm` from the regex pattern in the first git rule (`rules.rs`)
- Remove `"yadm"` from `rewrite_prefixes`
- Update tests to expect `Classification::Unsupported` and `None` for yadm commands

The existing `src/filters/yadm.toml` (output filtering) is preserved — it only applies when users explicitly pipe yadm output through rtk.

Closes #2077

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -869,12 +869,7 @@ mod tests {
     fn test_classify_yadm_status() {
         assert_eq!(
             classify_command("yadm status"),
-            Classification::Supported {
-                rtk_equivalent: "rtk git",
-                category: "Git",
-                estimated_savings_pct: 70.0,
-                status: RtkStatus::Existing,
-            }
+            Classification::Unsupported
         );
     }
 
@@ -882,12 +877,7 @@ mod tests {
     fn test_classify_yadm_diff() {
         assert_eq!(
             classify_command("yadm diff"),
-            Classification::Supported {
-                rtk_equivalent: "rtk git",
-                category: "Git",
-                estimated_savings_pct: 80.0,
-                status: RtkStatus::Existing,
-            }
+            Classification::Unsupported
         );
     }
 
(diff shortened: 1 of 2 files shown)


jensenojs:
need this

paolomainardi:
> need this

i guess it's already implement by some other MR and already part of stable release, i can close this one.

PR #2081, opened 2026-05-24
feat: Rust filter for pre-commit/prek -- compact hook output

mariotaddeucci:
## Summary

Creates a Rust filter module for pre-commit / prek that transforms hook output:
- Removes ... dot separators between hook name and status
- Wraps status in [brackets]: Trim Trailing Whitespace [Passed]
- Failed hooks show hook-id [Failed] + full linter error output
- Strips [INFO] installation noise and empty lines
- Only applies to run subcommand; install, autoupdate, etc. passthrough
- Also adds prek support with the same filter

## Test plan

- [x] cargo fmt --all && cargo clippy --all-targets && cargo test - 1938 passed, clippy clean
- [x] 10 unit tests covering all-pass, mixed pass/fail, skipped with reason, hook modifications, INFO stripping, multiple failures, real-world 13-hook output

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,7 +75,7 @@ For the full architecture, component details, and module development patterns, s
 
 Module responsibilities are documented in each folder's `README.md` and each file's `//!` doc header. Browse `src/cmds/*/` to discover available filters.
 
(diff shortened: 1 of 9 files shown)


mariotaddeucci:
Bump

newtob:
Love this idea. 
I would add that `prek` has a `-q` quiet flag that only outputs errors. 

The real question is that as a hook harness it can throw errors from any underlying tool. Could the output be parsed to pickup the error formats output by the tools already supported? For example, Ruff fails and the output format is parsed?

PR #675, opened 2026-03-18
fix(init): preserve settings.json symlink and portable hook path

TechWizard9999:
## Description
This fixes two global `rtk init` problems when patching Claude settings:
1. symlinked `settings.json` was replaced by a regular file
2. hook command entries were written as machine-specific absolute paths

## Related Issue
Closes #653

## Changes Made
- updated atomic write path resolution to detect symlink targets and write through to the target file (preserving the symlink itself)
- switched hook command insertion to a portable shell expression:
  - `${CLAUDE_CONFIG_DIR:-$HOME/.claude}/hooks/rtk-rewrite.sh`
- updated manual settings patch snippet to use the same portable hook command
- added targeted tests for symlink-preserving atomic writes and portable hook command generation
- added changelog entry under Unreleased bug fixes

## Files Changed
- `src/init.rs` - symlink-safe write path + portable hook command + tests
- `CHANGELOG.md` - unreleased bug-fix note for #653

## Testing
- [x] `cargo fmt --all --check`
- [x] `rtk cargo clippy --all-targets` (2 pre-existing warnings in untouched files)
- [x] `rtk cargo test`
- [x] `rtk cargo test init::tests::test_atomic_write_preserves_symlink`
- [x] `rtk cargo test init::tests::test_hook_command_for_settings_uses_portable_path`

## Checklist
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Documentation updated
- [x] No breaking changes introduced

cc @pszymkowiak for review

PR #2088, opened 2026-05-25
fix(init): preserve settings.json symlink during atomic write

### LINKED ISSUE #653: "rtk init --global" breaks symlink
I share `settings.json` across multiple devices, but this command turned my symlink into a regular file.

I consider this a serious issue.

Another problem is that `settings.json` shouldn't contain absolute paths, and the patch content also includes absolute paths, making it impossible to share across multiple machines. I refactored them to relative paths.

As for why rtk.md is injected into CLAUDE.md, another issue already asked about it, so I won't ask here.

ChrisGVE:
## Summary

- `atomic_write()` replaced symlinked `settings.json` with a regular file, breaking setups that symlink config files (e.g. dotfile managers, multi-machine sync)
- Now resolves symlink targets before the tempfile+rename so the link itself is preserved

Closes #653

Supersedes #675 — rebased on current develop. The portable hook path change from #675 is no longer needed since the codebase moved from shell scripts to `rtk hook claude` binary command.

## Changes Made

- Added `resolve_atomic_target()` that detects symlinks and resolves to their target (absolute or relative)
- Updated `atomic_write()` to write through to the resolved target
- Added 2 unix-specific tests for absolute and relative symlink preservation

## Testing

- [x] `cargo test --all` — 1942 passed, 6 ignored
- [x] Specific tests: `cargo test init::tests::test_atomic_write`
- [x] Verified both fixes merge cleanly together (1947 tests pass combined)

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] No breaking changes introduced
- [x] Tests added for new functionality

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -384,31 +384,15 @@ fn write_if_changed(path: &Path, content: &str, name: &str, ctx: InitContext) ->
 
 /// Resolve the final write target: if `path` is a symlink, follow it so
 /// the atomic rename lands on the real file and the symlink is preserved.
-fn resolve_atomic_target(path: &Path) -> Result<PathBuf> {
-    if let Ok(meta) = fs::symlink_metadata(path) {
-        if meta.file_type().is_symlink() {
-            let link_target = fs::read_link(path)
-                .with_context(|| format!("Failed to read symlink target: {}", path.display()))?;
-            if link_target.is_absolute() {
-                return Ok(link_target);
-            }
-            let parent = path.parent().with_context(|| {
-                format!(
-                    "Cannot resolve relative symlink for {}: no parent directory",
-                    path.display()
-                )
-            })?;
-            return Ok(parent.join(link_target));
-        }
-    }
-    Ok(path.to_path_buf())
+fn resolve_atomic_target(path: &Path) -> PathBuf {
+    fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf())
 }
 
 /// Atomic write using tempfile + rename
 /// Prevents corruption on crash/interrupt
 /// Follows symlinks so the link itself is preserved.
 fn atomic_write(path: &Path, content: &str) -> Result<()> {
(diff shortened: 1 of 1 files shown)


ChrisGVE:
@aeppling may I ask you to review this PR, this has been in the list for long and would help many user if it could be integrated into rtk. Thank you for your great work and for your time reviewing this PR.

src/hooks/init.rs
```diff
@@ -382,13 +382,37 @@ fn write_if_changed(path: &Path, content: &str, name: &str, ctx: InitContext) ->
     }
 }
 
+/// Resolve the final write target: if `path` is a symlink, follow it so
+/// the atomic rename lands on the real file and the symlink is preserved.
+fn resolve_atomic_target(path: &Path) -> Result<PathBuf> {
```
KuSh: Please use [fs::canonicalize](url) that handles all the hard things for you.

KuSh (changes requested):

KuSh:
Hi, thanks for your contribution!
As it is a rebase of someone else submission, you'll have to add "Co-authored-by" trailer to the commit's message.

src/hooks/init.rs
ChrisGVE: Done — replaced with `fs::canonicalize`, falls back to original path when file doesn't exist yet. All 3 `test_atomic_write*` tests pass (including both symlink variants).

ChrisGVE:
Thanks @KuSh for the review! Both items addressed:

1. **`fs::canonicalize`** — replaced custom `resolve_atomic_target()` with `fs::canonicalize` + fallback for non-existent paths (new file creation). Handles chained symlinks, relative paths, all edge cases.

2. **Co-authored-by** — added `Co-authored-by: Roopesh <roopesh1724989@gmail.com>` trailer (original author of #675).

All tests green: `cargo fmt`, `cargo clippy`, `cargo test` — 1951 passed.

KuSh (approved):
LGTM, thanks

Issue #1331, opened 2026-04-15
Add aarch64-unknown-linux-musl target to release binaries


julien-redel:
## Problem                                      

  The current release only provides `rtk-aarch64-unknown-linux-gnu.tar.gz` for Linux ARM64. This binary is not compatible with Alpine
   Linux (musl libc), which is widely used in container images.
                                                                                                                                     
  Running the `gnu` binary on Alpine results in:                                                                                     
   
  Error relocating /usr/local/bin/rtk: fcntl64: symbol not found                                                                     
  Error relocating /usr/local/bin/rtk: __res_init: symbol not found
                                                                                                                                     
  Even with `libc6-compat` and `gcompat` packages installed, the missing symbols are not resolved.
                                                                                                                                     
  ## Current workaround                                                                                                              
                                                                                                                                     
  We build rtk from source using Cargo inside an Alpine-based Docker stage, which works but adds significant build time (~3-5 min)   
  compared to downloading a pre-built binary.     
                                                                                                                                     
  ## Request                                      

  Add `rtk-aarch64-unknown-linux-musl.tar.gz` to the release assets, similar to how `rtk-x86_64-unknown-linux-musl.tar.gz` is already
   provided for x86_64.
                                                                                                                                     
  ## Context                                      

  - The x86_64 musl target already exists: `rtk-x86_64-unknown-linux-musl.tar.gz`                                                    
  - Alpine Linux is the de facto standard for lightweight container images
  - ARM64 (aarch64) adoption is growing fast in cloud (AWS Graviton, Azure Cobalt, etc.)                                             
  - The Rust target `aarch64-unknown-linux-musl` is Tier 2 with host tools, so cross-compilation should be straightforward   

 Thanks for the great tool! 🚀

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 97% |

### Summary

Request to add aarch64-unknown-linux-musl release binary for Alpine Linux / musl libc compatibility; x86_64 musl already exists, this extends parity to ARM64.

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
[...]

Issue #615, opened 2026-03-16
ARM64 Raspberry Pi: install script downloads binary requiring GLIBC_2.39 (Debian Bookworm incompatible)


tonygloaguen:
# ARM64 Raspberry Pi: install script downloads binary requiring GLIBC_2.39 (Debian Bookworm incompatible)

## Summary

When installing RTK on a Raspberry Pi 4 (ARM64) running Debian Bookworm using the official install script, the downloaded binary fails to run due to a GLIBC version mismatch.

The installation script reports success, but the binary cannot execute on a standard Debian Bookworm system.

---

## Environment

**Hardware:** Raspberry Pi 4  

**Architecture**

```bash
uname -m

Output:

aarch64

OS

Debian GNU/Linux 12 (Bookworm)

Kernel

uname -a

Example:

Linux 6.12.62+rpt-rpi-v8

GLIBC version

ldd --version

Output:

glibc 2.36
Installation method

Official install script:

curl -fsSL https://raw.githubusercontent.com/rtk-ai/rtk/refs/heads/master/install.sh | sh

Script output:

[INFO] Installing rtk...
[INFO] Detected: linux aarch64
[INFO] Target: aarch64-unknown-linux-gnu
[INFO] Version: v0.29.0
[INFO] Successfully installed rtk to ~/.local/bin/rtk
Actual behaviour

Running the binary produces:

rtk: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found

This prevents RTK from running on Debian Bookworm / Raspberry Pi OS.

Expected behaviour

The installed binary should run on a typical ARM64 Linux environment such as:

Debian Bookworm

Raspberry Pi OS 64-bit

Other ARM64 Linux distributions using glibc 2.36

Workaround

Compiling RTK locally works correctly:

cargo install --git https://github.com/rtk-ai/rtk

Compilation time on Raspberry Pi 4 is approximately:

8–12 minutes

After compilation:

rtk --version
rtk gain

Both commands work as expected.

Possible improvements

One of the following could make installation smoother for ARM users:

Provide an ARM64 binary compiled against an older glibc compatible with Debian Bookworm

Provide a static musl build

Make the install script detect incompatible glibc and fall back to cargo install

Additional context

RTK works very well once compiled on ARM64.

In my case it is used with Claude Code hooks to compress shell outputs before sending them to an LLM, which is particularly useful on lightweight ARM nodes running automation agents.

Thank you

[...]

PR #2089, opened 2026-05-25
fix(ci): publish linux aarch64 musl binary

ofek:
## Summary

Hello there! This fixes the Linux aarch64 binary distribution by publishing a musl build instead of the glibc-linked GNU build.

This resolves the aarch64 side of https://github.com/rtk-ai/rtk/issues/265, where release binaries can fail on systems with older glibc, such as Ubuntu 22.04-based dev containers. The x86_64 Linux artifact already moved to musl in https://github.com/rtk-ai/rtk/pull/267; this applies the same approach to aarch64 Linux.

This also picks up the intent from https://github.com/rtk-ai/rtk/pull/459, which was closed prematurely by its author. The changes are as follows:

- Build `aarch64-unknown-linux-musl` in the release workflow using `cross`
- Add `Cross.toml` env passthrough for compile-time telemetry vars
- Update installer target detection to download the musl Linux ARM artifact
- Update Homebrew formula generation, checked-in formula, and docs to reference the musl artifact

## Test plan

- [x] `rg "unknown-linux-gnu" -n` returns no matches
- [x] Parsed `.github/workflows/release.yml` with [PyYAML](https://pypi.org/project/PyYAML/)
- [x] Parsed `Cross.toml` with [tomllib](https://docs.python.org/3/library/tomllib.html) from Python's standard library
- [x] `git diff --check`
- [x] `cargo check`
- [ ] Release workflow validation: confirm `rtk-aarch64-unknown-linux-musl.tar.gz` is uploaded and included in `checksums.txt`

I didn't want to make too many changes in order to ease review but in future I think it would be desirable to build each artifact on PRs. I'd recommend splitting `.github/workflows/release.yml` into two parts:

- reusable build/package workflow: builds targets and uploads artifacts
- release workflow: downloads those artifacts, publishes release assets, etc.

Then PR CI can call only the build/package part, while release/CD calls both build/package and publish.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,7 +48,8 @@ jobs:
             os: ubuntu-latest
             archive: tar.gz
             musl: true
-          - target: aarch64-unknown-linux-gnu
+          # Use cross for aarch64 musl so the Linux ARM binary is statically linked.
+          - target: aarch64-unknown-linux-musl
             os: ubuntu-latest
             archive: tar.gz
             cross: true
@@ -66,12 +67,9 @@ jobs:
         with:
           targets: ${{ matrix.target }}
 
-      - name: Install cross-compilation tools
+      - name: Install cross
         if: matrix.cross
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y gcc-aarch64-linux-gnu
-          echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc" >> $GITHUB_ENV
+        run: cargo install cross --version 0.2.5 --locked
 
       - name: Install musl tools
         if: matrix.musl
@@ -79,7 +77,15 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y musl-tools
 
-      - name: Build
+      - name: Build (cross)
+        if: matrix.cross
+        run: cross build --release --target ${{ matrix.target }}
+        env:
+          RTK_TELEMETRY_URL: ${{ vars.RTK_TELEMETRY_URL }}
+          RTK_TELEMETRY_TOKEN: ${{ secrets.RTK_TELEMETRY_TOKEN }}
+
+      - name: Build (cargo)
+        if: ${{ ! matrix.cross }}
         run: cargo build --release --target ${{ matrix.target }}
         env:
           RTK_TELEMETRY_URL: ${{ vars.RTK_TELEMETRY_URL }}
@@ -281,7 +287,7 @@ jobs:
         run: |
           echo "mac_arm=$(grep aarch64-apple-darwin.tar.gz checksums.txt | head -1 | awk '{print $1}')" >> $GITHUB_OUTPUT
           echo "mac_intel=$(grep x86_64-apple-darwin.tar.gz checksums.txt | head -1 | awk '{print $1}')" >> $GITHUB_OUTPUT
-          echo "linux_arm=$(grep aarch64-unknown-linux-gnu.tar.gz checksums.txt | head -1 | awk '{print $1}')" >> $GITHUB_OUTPUT
+          echo "linux_arm=$(grep aarch64-unknown-linux-musl.tar.gz checksums.txt | head -1 | awk '{print $1}')" >> $GITHUB_OUTPUT
           echo "linux_intel=$(grep x86_64-unknown-linux-musl.tar.gz checksums.txt | head -1 | awk '{print $1}')" >> $GITHUB_OUTPUT
 
       - name: Generate formula
@@ -300,7 +306,7 @@ jobs:
               url "https://github.com/rtk-ai/rtk/releases/download/TAG_PLACEHOLDER/rtk-x86_64-apple-darwin.tar.gz"
               sha256 "SHA_MAC_INTEL_PLACEHOLDER"
             elsif OS.linux? && Hardware::CPU.arm?
-              url "https://github.com/rtk-ai/rtk/releases/download/TAG_PLACEHOLDER/rtk-aarch64-unknown-linux-gnu.tar.gz"
+              url "https://github.com/rtk-ai/rtk/releases/download/TAG_PLACEHOLDER/rtk-aarch64-unknown-linux-musl.tar.gz"
               sha256 "SHA_LINUX_ARM_PLACEHOLDER"
             elsif OS.linux? && Hardware::CPU.intel?
               url "https://github.com/rtk-ai/rtk/releases/download/TAG_PLACEHOLDER/rtk-x86_64-unknown-linux-musl.tar.gz"
--- a/Cross.toml
+++ b/Cross.toml
@@ -1,0 +1,6 @@
+[build.env]
+# Forward compile-time telemetry vars used by option_env!() into cross containers.
+passthrough = [
+    "RTK_TELEMETRY_URL",
+    "RTK_TELEMETRY_TOKEN",
+]
--- a/Formula/rtk.rb
+++ b/Formula/rtk.rb
@@ -23,12 +23,12 @@ class Rtk < Formula
 
   on_linux do
     on_intel do
-      url "https://github.com/rtk-ai/rtk/releases/download/v#{version}/rtk-x86_64-unknown-linux-gnu.tar.gz"
+      url "https://github.com/rtk-ai/rtk/releases/download/v#{version}/rtk-x86_64-unknown-linux-musl.tar.gz"
       sha256 "PLACEHOLDER_SHA256_LINUX_INTEL"
     end
 
     on_arm do
-      url "https://github.com/rtk-ai/rtk/releases/download/v#{version}/rtk-aarch64-unknown-linux-gnu.tar.gz"
+      url "https://github.com/rtk-ai/rtk/releases/download/v#{version}/rtk-aarch64-unknown-linux-musl.tar.gz"
       sha256 "PLACEHOLDER_SHA256_LINUX_ARM"
     end
   end
--- a/install.sh
+++ b/install.sh
@@ -74,7 +74,7 @@ get_target() {
         linux)
             case "$ARCH" in
                 x86_64)  TARGET="x86_64-unknown-linux-musl";;
-                aarch64) TARGET="aarch64-unknown-linux-gnu";;
+                aarch64) TARGET="aarch64-unknown-linux-musl";;
             esac
             ;;
         darwin)
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ cargo install --git https://github.com/rtk-ai/rtk
 
 Download from [releases](https://github.com/rtk-ai/rtk/releases):
 - macOS: `rtk-x86_64-apple-darwin.tar.gz` / `rtk-aarch64-apple-darwin.tar.gz`
-- Linux: `rtk-x86_64-unknown-linux-musl.tar.gz` / `rtk-aarch64-unknown-linux-gnu.tar.gz`
+- Linux: `rtk-x86_64-unknown-linux-musl.tar.gz` / `rtk-aarch64-unknown-linux-musl.tar.gz`
 - Windows: `rtk-x86_64-pc-windows-msvc.zip`
 
 > **Windows users**: Extract the zip and place `rtk.exe` somewhere in your PATH (e.g. `C:\Users\<you>\.local\bin`). Run RTK from **Command Prompt**, **PowerShell**, or **Windows Terminal** — do not double-click the `.exe` (it will flash and close). For the best experience, use [WSL](https://learn.microsoft.com/en-us/windows/wsl/install) where the full hook system works natively. See [Windows setup](#windows) below for details.
--- a/docs/guide/getting-started/installation.md
+++ b/docs/guide/getting-started/installation.md
@@ -52,7 +52,7 @@ cargo install --git https://github.com/rtk-ai/rtk rtk
 Download from [GitHub releases](https://github.com/rtk-ai/rtk/releases):
 
 - macOS: `rtk-x86_64-apple-darwin.tar.gz` / `rtk-aarch64-apple-darwin.tar.gz`
-- Linux: `rtk-x86_64-unknown-linux-musl.tar.gz` / `rtk-aarch64-unknown-linux-gnu.tar.gz`
+- Linux: `rtk-x86_64-unknown-linux-musl.tar.gz` / `rtk-aarch64-unknown-linux-musl.tar.gz`
 - Windows: `rtk-x86_64-pc-windows-msvc.zip`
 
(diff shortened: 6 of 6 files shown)


pszymkowiak:
This also resolves #1331 (add the `aarch64-unknown-linux-musl` target) — it adds that target and makes it the default for Linux/arm64, so it supersedes #2196 which addressed #1331 alone.

Suggest adding `Closes #1331` to the PR description alongside the #615 reference so both issues close on merge.

.github/workflows/release.yml
```diff
@@ -66,20 +67,25 @@ jobs:
...
           targets: ${{ matrix.target }}
 
-      - name: Install cross-compilation tools
+      - name: Install cross
         if: matrix.cross
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y gcc-aarch64-linux-gnu
-          echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc" >> $GITHUB_ENV
+        run: cargo install cross --version 0.2.5 --locked
```
KuSh, line 72: It seems risky to use version as installation target for not packaged deps as it has security implication, see commit https://github.com/rtk-ai/rtk/commit/26b96ec6c4f40f992ccffa190af9a4de8d7636b1, for example

Can't we use musl-tools, which includes musl-gcc for that need?

KuSh (changes requested):

gitbluf:
Hi!

Any update on this one @ofek ?

@KuSh / @pszymkowiak if this PR is abandoned I can take over with a new PR ?

DanyHovard:
the issue is still there 
`rtk: /lib/aarch64-linux-gnu/libc.so.6: version GLIBC_2.39 not found (required by rtk)`

KuSh (changes requested):
Verified the change end to end before reviewing: built this head with `cargo install cross --version 0.2.5 --locked` + `cross build --release --target aarch64-unknown-linux-musl`. It works — exit 0, artifact is `ELF 64-bit LSB executable, ARM aarch64, statically linked, stripped`, and the `Cross.toml` passthrough does bake the telemetry vars in (confirmed with a canary value). The goal of the PR is achieved.

I'd still like it done without `cross`. Details are on the individual lines; summary here.

## Blocking

- **Build on a native ARM runner rather than `cross`** (`release.yml:53`). `ubuntu-24.04-arm` + the existing `musl-tools` step + the existing plain `cargo build` covers this target, and lets the PR delete rather than add.
- **`install.sh` needs a fallback for the renamed asset** (`install.sh:77`). As written it 404s for every historical `RTK_VERSION` on ARM Linux, and again in the window between the develop→master merge and the first release carrying the musl asset.

## Why a runner and not a version pin

To close out my earlier thread: pinning `cargo install cross --version 0.2.5 --locked` doesn't fix what I was worried about. crates.io versions are immutable and `--locked` pins cross's own lockfile, so that half is already fine. The unpinned input is what `cross` does next — it pulls `ghcr.io/cross-rs/aarch64-unknown-linux-musl:0.2.5`, a **mutable tag**, and that container's `aarch64-linux-musl-gcc` and linker are what produce the published binary. I confirmed the image and the tag resolution during the build above. Anyone able to move that tag changes every subsequently released aarch64 binary, with no diff in this repo and no signal in CI — the same class of exposure 26b96ec closed for Actions. A digest pin in `Cross.toml` would fix it; a native runner removes `cross`, Docker and the GHCR image from the release path entirely, and leaves no digest to maintain.

## Non-blocking, pre-existing — separate PR

- `release.yml:125` / `:152` — `cargo install cargo-deb` and `cargo install cargo-generate-rpm` carry neither `--version` nor `--locked`, so the published `.deb`/`.rpm` are built by whatever version is current that day. Larger gap than the one I originally flagged.
- `release.yml:66` and elsewhere — every action here is on a mutable ref (`actions/checkout@v4`, `actions/upload-artifact@v4`, `softprops/action-gh-release@v2`, …), and `dtolnay/rust-toolchain@stable` is a *branch*, not a tag — it installs the compiler for every published binary. 26b96ec pinned `pr-target-check.yml` and never came back for this file.
- Scope note for #265: `build-deb` and `build-rpm` still do a bare host-target glibc `cargo build --release` and are x86_64-only, so ARM users get no package and the glibc issue persists for `.deb`/`.rpm` installs. #265 shouldn't be closed as fully resolved by this PR.

.github/workflows/release.yml
```diff
@@ -48,7 +48,8 @@ jobs:
             os: ubuntu-latest
             archive: tar.gz
             musl: true
-          - target: aarch64-unknown-linux-gnu
+          # Use cross for aarch64 musl so the Linux ARM binary is statically linked.
+          - target: aarch64-unknown-linux-musl
             os: ubuntu-latest
```
KuSh, line 53: **Requested change — build this target on a native ARM runner instead of `cross`.**

```suggestion
            os: ubuntu-24.04-arm
```

With `musl: true` in place of `cross: true`, this entry reuses the existing `Install musl tools` step and the existing plain `cargo build --release --target ${{ matrix.target }}`, exactly like the x86_64 musl entry above. `musl-tools` ships `musl-gcc` for the host arch only — which is why it can't cross-link aarch64 from an x86_64 runner, but *is* the native toolchain on an arm64 one. GitHub-hosted arm64 runners are free for public repos.

That lets this PR delete rather than add: no `Install cross` step, no `Build (cross)`/`Build (cargo)` split, no `Cross.toml`.

.github/workflows/release.yml
```diff
@@ -66,20 +67,25 @@ jobs:
...
+      - name: Build (cross)
+        if: matrix.cross
+        run: cross build --release --target ${{ matrix.target }}
+        env:
+          RTK_TELEMETRY_URL: ${{ vars.RTK_TELEMETRY_URL }}
+          RTK_TELEMETRY_TOKEN: ${{ secrets.RTK_TELEMETRY_TOKEN }}
+
+      - name: Build (cargo)
+        if: ${{ ! matrix.cross }}
         run: cargo build --release --target ${{ matrix.target }}
```
KuSh, line 89: Comes out with `cross`: with a native `ubuntu-24.04-arm` runner the existing `Install musl tools` step covers this target, and the `Build (cross)` / `Build (cargo)` split collapses back to the single `Build` step.

Independent of that, this PR's two guarantees are currently unasserted — nothing checks the published ARM artifact is actually static, so a regression would ship silently. Worth one line in the build job:

```yaml
- run: file target/${{ matrix.target }}/release/rtk | grep -q "statically linked"
```

I confirmed that exact string appears in the real `file` output for this artifact.

install.sh
```diff
@@ -74,7 +74,7 @@ get_target() {
         linux)
             case "$ARCH" in
                 x86_64)  TARGET="x86_64-unknown-linux-musl";;
-                aarch64) TARGET="aarch64-unknown-linux-gnu";;
+                aarch64) TARGET="aarch64-unknown-linux-musl";;
```
KuSh, line 77: **Blocking — this rename needs a fallback.**

`install.sh` is served from `refs/heads/master`, so renaming the target unconditionally produces two 404s:

1. `RTK_VERSION=v0.42.4 curl … | sh` on aarch64 Linux now requests `rtk-aarch64-unknown-linux-musl.tar.gz`, which exists in no release predating this change. `curl -fsSL` 404s and the script dies on `error "Failed to download binary"`. Every historical version becomes permanently un-installable on ARM Linux.
2. Merging develop→master publishes the new `install.sh` immediately, but the musl asset only exists once release-please's Release PR merges and `build-release` finishes. In that window `/releases/latest` still resolves to a `-gnu`-only release and every default ARM Linux install 404s. Develop pre-releases don't cover it — `/releases/latest` skips prereleases.

Please try the musl asset and fall back to `aarch64-unknown-linux-gnu` on download failure, using the same resolved name for the `checksums.txt` lookup.

Cross.toml
```diff
@@ -0,0 +1,6 @@
+[build.env]
```
KuSh, line 1: This file goes away with `cross` (see my comment on the matrix entry).

If `cross` does stay, scope the passthrough to the target rather than `[build.env]`: build-scoped means every present *and future* cross target inherits it, so adding a second `cross: true` entry would silently forward `RTK_TELEMETRY_TOKEN` into an unvetted image with no diff to review.

```toml
[target.aarch64-unknown-linux-musl.env]
passthrough = ["RTK_TELEMETRY_URL", "RTK_TELEMETRY_TOKEN"]
```

Not a new secret exposure either way — the token is `option_env!`-baked into every published binary and recoverable with `strings`. I verified the passthrough works as written.

Formula/rtk.rb
```diff
@@ -23,12 +23,12 @@ class Rtk < Formula
 
   on_linux do
     on_intel do
-      url "https://github.com/rtk-ai/rtk/releases/download/v#{version}/rtk-x86_64-unknown-linux-gnu.tar.gz"
+      url "https://github.com/rtk-ai/rtk/releases/download/v#{version}/rtk-x86_64-unknown-linux-musl.tar.gz"
```
KuSh, line 26: Not this PR's problem, but this file is a dead second copy: the formula actually published to `rtk-ai/homebrew-tap` is generated from the heredoc around `release.yml:300`, and nothing consumes `Formula/rtk.rb`. That's why the change has to be made twice here, and why this file still said `x86_64-unknown-linux-gnu` for many releases after the generator moved to musl in #267.

Worth deleting it or generating it from the same source, rather than keeping two that drift.

ether-btc:
Thanks @ofek — supporting this direction; the field reproduction lines up exactly. Quick recap of @KuSh's two blocking review items from 2026-09-04, with a concrete sketch so we can converge.

### Reproduced on this host (Pi 5, Debian 12, aarch64, glibc 2.36)

```text
$ ./latest-rtk-gnu --version
./latest-rtk-gnu: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by ./latest-rtk-gnu)
```

So #615 and this PR's underlying motivation are real on a stock Debian/Pi host, not just theoretical musl-only edge cases.

### KuSh review recap + concrete patch sketch

**(1) `install.sh` `get_target` needs to switch aarch64 from gnu to musl** (prerequisite for the whole PR).

KuSh's blocker explicitly calls this out — `install.sh:77` still maps aarch64 to `aarch64-unknown-linux-gnu`, so without this one-line change the new musl asset is never reached by `curl`, and the matrix rename and the homebrew-tap update below would ship a binary nobody can install via `curl … | sh`.

Concrete patch:

```sh
# install.sh, get_target() — line 77
# before: aarch64) TARGET="aarch64-unknown-linux-gnu";;
# after:
        aarch64) TARGET="aarch64-unknown-linux-musl";;
```

This is the change that makes `curl` actually fetch the new musl asset. KuSh already named it in the 2026-09-04 review; this comment is just pinning the exact line number.

**(2) Build on the native arm64 runner instead of `cross`.**

Current matrix adds the new row with `cross: true`. Concrete patch:

```yaml
# .github/workflows/release.yml — in the matrix include block, replace the
# proposed `cross: true` row with this:
- target: aarch64-unknown-linux-musl
  os: ubuntu-24.04-arm          # native arm64, free for public repos
  archive: tar.gz
  musl: true                    # reuse existing musl toolchain step
```

This lets you delete, not add:

- the `Install cross-compilation tools` step,
- the `Install cross` step (and the `cross` Rust dependency),
- the `Cross.toml` file in the repo root.

The current `release.yml` already has a single unconditional `cargo build --release --target ${{ matrix.target }}` for Linux rows (gated only by `matrix.musl` to install `musl-tools`), so no conditional build split is needed in this PR — only the matrix row swap. `musl-tools` on `ubuntu-24.04-arm` ships native `musl-gcc`, so `cross` is no longer required.

**(3) `install.sh` needs a hard fallback for the renamed asset.**

KuSh's second blocker: pinned-version users (`RTK_VERSION=v0.42.4 curl … | sh`) and any release between this PR's merge and the first release carrying the musl asset will hit a 404 on the musl URL.

Concrete patch sketch (sh-style, fits the existing file). The non-obvious invariant that makes this safe: `checksums.txt` is shared across assets, so a single download covers both the musl URL and the gnu fallback URL.

```sh
# in install(), after the first curl, before the checksum lookup:
if ! curl -fsSL "$DOWNLOAD_URL" -o "$ARCHIVE"; then
    # Fallback for pinned-version users on releases predating the musl asset.
    if [ "$TARGET" = "aarch64-unknown-linux-musl" ]; then
        FALLBACK_TARGET="aarch64-unknown-linux-gnu"
        FALLBACK_URL="https://github.com/${REPO}/releases/download/${VERSION}/${BINARY_NAME}-${FALLBACK_TARGET}.tar.gz"
        warn "$TARGET asset not found for $VERSION; falling back to $FALLBACK_TARGET"
        TARGET="$FALLBACK_TARGET"
        DOWNLOAD_URL="$FALLBACK_URL"
        ASSET_NAME="${BINARY_NAME}-${TARGET}.tar.gz"
        curl -fsSL "$DOWNLOAD_URL" -o "$ARCHIVE" || error "Failed to download binary (musl and gnu fallback both 404 for $VERSION)"
    else
        error "Failed to download binary"
    fi
fi
```

The same resolved `ASSET_NAME` is then used for the `checksums.txt` grep (line 113), so the gnu asset's checksum is verified against the same `checksums.txt` that's already been downloaded — both asset names appear in that file.

Caveat: a transient network failure would also trip the fallback (the curl exit code is the only signal), so the warn message should be read as "musl URL failed, trying gnu" rather than "asset missing." For Alpine users pinning an old version, the gnu fallback "succeeds" at install-time but fails at runtime (no glibc on Alpine) — a second warn ("gnu fallback requires glibc ≥2.39; not suitable for musl-only systems") would help; optional polish.

**(4) Homebrew-tap heredoc must also update** (would otherwise silently break the next formula push).

After the matrix rename, two more lines in `release.yml` need to change:

- `release.yml:293` — `linux_arm=$(grep aarch64-unknown-linux-gnu.tar.gz checksums.txt ...)` must become `linux_arm=$(grep aarch64-unknown-linux-musl.tar.gz checksums.txt ...)`. If this is missed, the grep returns empty, `sed` substitutes nothing, and the formula ships with `sha256 ""` — Homebrew install fails at fetch.
- `release.yml:311-313` — the homebrew-tap heredoc `url "https://.../rtk-aarch64-unknown-linux-gnu.tar.gz"` must become `...rtk-aarch64-unknown-linux-musl.tar.gz`. The corresponding `SHA_LINUX_ARM_PLACEHOLDER` needs no name change because it's already wired via `steps.sha.outputs.linux_arm` at line 350.

**(5) `Formula/rtk.rb` is a dead second copy** — agree with @KuSh, defer the cleanup to a separate PR per their suggestion. The heredoc generator at `release.yml:296` is the single source of truth for what hits `rtk-ai/homebrew-tap`; the in-tree `Formula/rtk.rb` should either be deleted or made a generated artifact.

### On Closes vs Fixes

Suggesting "Closes #1331, Closes #615" on merge so both issues close atomically — covers the original Alpine ask and the older-glibc ask together, which is exactly the design intent.

Issue #1975, opened 2026-05-20
[Feature Request] Add support to Antigravity CLI


foobra:
Google introduce Antigravity CLI to replace Gemini CLI.

https://antigravity.google/download  

Thanks

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟢 **Priority** | `low` |
| 🎯 **Confidence** | 72% |

### Summary

Feature request to add support for 'Antigravity CLI' — likely a fictional/non-existent tool; needs verification before acting on

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

bstordrup:
As far as I have been testing using Antigravity CLI and manually editing files, supporting hooks will require editing the `~/.gemini/antigravity-cli/hooks.json` or `~/.gemini/config/hooks.json` to include this content

```
{
  "RtkHook": {
    "enabled": true,
    "PreInvocation": null,
    "PostInvocation": null,
    "Stop": null,
    "PreToolUse": [
      {
        "matcher": "run_shell_command",
        "hooks": [
          {
            "type": "command",
            "command": "~/.gemini/hooks/rtk-hook-gemini.sh",
            "timeout": 0
          }
        ]
      }
    ],
    "PostToolUse": null
  }
}
```

Both locations works - `agy` see the hooks - but only the `~/.gemini/config/hooks.json` is documented in [Antigravity Hooks documentation](https://antigravity.google/docs/hooks), and that one takes precedence if both exists.

julian-carpenter:
Yes, I can confirm this. However, I am unsure if the matcher should be "run_shell_command" as was with Gemini or "run_command", as is listed in the [AGY CLI doc](https://antigravity.google/docs/hooks)

bstordrup:
Using the specified hook definition, I did manage to get some results - I was able to see a result when running `rtk gain` afterwards.

PR #2093, opened 2026-05-26
feat(agy): Google Antigravity CLI hook integration

thelok1s:
## Summary

- Adds `rtk hook antigravity` — a `PreToolUse` hook for the Google Antigravity CLI (`agy`) that rewrites bare shell commands to their `rtk`-prefixed equivalents before the model executes them
- `rtk init --agent agy` configures `~/.gemini/antigravity-cli/settings.json` with the hook and grants `command(rtk)` permission so retries are auto-approved
- Handles both simple (`tool_name`/`tool_input`) and rich (`toolCall.name`/`toolCall.args`) agy payload formats
- `RTK_HOOK_DEBUG=1` logs full request/response pairs to `~/.local/share/rtk/agy-hook-debug.log`

## How it works

agy's `overwrite` field in `PreToolHookResult` is silently ignored (confirmed via debug logging — correct payload sent, original command executed). The hook therefore uses the `denyReason` strategy:

1. Hook intercepts `run_command` / `Bash` tool calls
2. If the command has an RTK equivalent, hook returns `{"denyReason": "RTK: use 'rtk git log' instead of 'git log' (60-90% token savings)"}`
3. Model reads the reason and retries with the `rtk`-prefixed command
4. `command(rtk)` in the allow list auto-approves the retry — no user prompt

Commands already prefixed with `rtk` or in the passthrough list pass through with `{"allowTool": true}`.

## Test plan

- [x] `cargo test --all` — 1924 tests pass, zero clippy warnings
- [x] `rtk init --agent agy` writes correct `~/.gemini/antigravity-cli/settings.json`
- [x] Hook rejects bare `git log` with denyReason naming `rtk git log`
- [x] Hook passes through `rtk git log` with `allowTool: true`
- [x] Hook passes through non-shell tools (e.g. `read_file`) with `allowTool: true`
- [x] `RTK_HOOK_DEBUG=1` writes input/output to debug log

## Known limitation

agy does not honour the `overwrite` field in `PreToolHookResult` — only `allowTool` and `denyReason` are implemented. The deny-and-retry pattern produces a `⚠ Tool call denied` notice in the agy UI before the model retries with the RTK command. Transparent rewriting will be possible once agy implements `overwrite` support.

This issue reported in [this topic](https://discuss.ai.google.dev/t/pretoolhookresult-overwrite-is-broken-under-toolpermission-request-review/165839?u=lok1s) on discuss.ai.google.dev. More visibility on this bug report required.

image
Tested in Antigravity CLI v1.0.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,10 +106,11 @@ rtk init -g                     # Claude Code / Copilot (default)
 rtk init -g --gemini            # Gemini CLI
 rtk init -g --codex             # Codex (OpenAI)
 rtk init -g --agent cursor      # Cursor
+rtk init -g --agent agy         # Google Antigravity CLI (agy)
 rtk init --agent windsurf       # Windsurf
 rtk init --agent cline          # Cline / Roo Code
 rtk init --agent kilocode       # Kilo Code
-rtk init --agent antigravity    # Google Antigravity
+rtk init --agent antigravity    # Google Antigravity IDE (rules-based)
 rtk init --agent hermes         # Hermes
 
 # 2. Restart your AI tool, then test
@@ -351,7 +352,7 @@ rtk git status
 
 ## Supported AI Tools
 
-RTK supports 13 AI coding tools. Each integration rewrites shell commands to `rtk` equivalents for 60-90% token savings where the agent supports command interception.
+RTK supports 14 AI coding tools. Each integration rewrites shell commands to `rtk` equivalents for 60-90% token savings where the agent supports command interception.
 
 | Tool | Install | Method |
 |------|---------|--------|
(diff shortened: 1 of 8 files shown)


thelok1s:
Resolves #1975

webermaster:
This would be very useful.  any idea on when it might be resolved?

KuSh:
Closing in favour of #3711, which adds the same native Antigravity PreToolUse hook on the canonical antigravity target. Both of the current Antigravity pull requests rewrite the command through the `overwrite` field, with a live capture from a recent agy build showing it is honoured, so the deny-with-reason strategy here is no longer needed. If you have newer evidence that `overwrite` is still ignored, or you meant the separate `agy` target to stay its own integration, please say so on #3711 or reply here.

Issue #1830, opened 2026-05-11
[rewrite] dotnet test|restore|format should auto-rewrite to rtk dotnet


TimurZanagar:
## Summary

  The proxies `rtk dotnet test`, `rtk dotnet restore`, and `rtk dotnet format`
  already exist (see `rtk dotnet --help`), but `rtk rewrite` only recognises
  `dotnet build`. As a result, the Claude Code hook silently passes through the
  other three subcommands.

  ## Reproduction (rtk 0.39.0)

  ```
  $ rtk rewrite "dotnet test"       ; echo $?
  1
  $ rtk rewrite "dotnet restore"    ; echo $?
  1
  $ rtk rewrite "dotnet format"     ; echo $?
  1
  $ rtk rewrite "dotnet build"      ; echo $?   # works
  rtk dotnet build
  0
  ```

  ## Impact

  `rtk discover` on a 30-day, 18k-command Claude Code history shows
  `dotnet test` = 263 calls and `dotnet restore` = 66 calls. None of them
  benefit from the existing proxy because the rewrite registry skips them.

  ## Proposed fix

  Extend the rewrite rule for `dotnet` in `src/discover/registry.rs` so it
  matches the subcommand set `{build, test, restore, format}` instead of only
  `build`. The output filter already handles all four — this is purely a
  routing fix.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 93% |

### Summary

Rewrite registry only covers `dotnet build`; extend to `test`, `restore`, and `format` in src/discover/registry.rs

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

- `src/discover/registry.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

vadamtsevich-thinktime:
[...]

PR #2105, opened 2026-05-26
fix(telemetry): distinguish "consent not given" from missing salt in status (#1656)

### LINKED ISSUE #1656: bug(telemetry): salt file not auto-created on macOS — device hash shows '(no salt file)'
## Description

On macOS, `rtk telemetry status` shows `device hash: (no salt file)` even after running commands that should trigger `generate_device_hash()` (e.g. `rtk gain`).

## Root cause

The `salt_file_path()` function uses `dirs::data_local_dir()`, which on macOS resolves to `~/Library/Application Support/` — **not** `~/.local/share/` (Linux convention). However, the auto-creation logic in `get_or_create_salt()` silently fails, leaving no salt file and no error message.

The `telemetry status` command reads the salt file path correctly (from `~/Library/Application Support/rtk/.device_salt`), but since the file was never created, it falls back to `(no salt file)`.

## Steps to reproduce

1. Fresh install via `brew install rtk` on macOS
2. Run `rtk gain` or any command
3. Run `rtk telemetry status`
4. Observe: `device hash: (no salt file)`
5. Check `~/.local/share/rtk/` — directory doesn't exist
6. Check `~/Library/Application Support/rtk/` — `.device_salt` not present

## Workaround

```bash
cp <some-salt> ~/Library/Application\ Support/rtk/.device_salt
```

After placing the file in the correct macOS path, `rtk telemetry status` shows the device hash correctly.

## Expected behavior

The salt file should be auto-created at `~/Library/Application Support/rtk/.device_salt` on first use on macOS.

## Environment

- OS: macOS (Darwin 25.4.0)
- RTK version: 0.38.0
- Installed via: Homebrew

maxmilian:
## Summary

Fixes #1656. `rtk telemetry status` shows `device hash: (no salt file)` for two very different states:

1. **Expected** — consent never granted, so the salt is never attempted (by design).
2. **Real failure** — consent granted but the write was blocked (fs permission, disk full, …).

The single label reads as an error in case 1 and obscures actual failures in case 2.

## Root cause

Verified against `develop` HEAD (per `rg generate_device_hash` and @truffle-dev's analysis):

- `get_or_create_salt()` already calls `create_dir_all(parent)` at `src/core/telemetry.rs:175-177` — no fs bug.
- `generate_device_hash()` has three production call sites: `send_ping()` at `telemetry.rs:71` (the only one that creates the salt; gated by `consent_given == Some(true)` at `telemetry.rs:38-41`), and `run_status()` / `run_forget()` (both only call it after `salt_path.exists()`).
- A fresh `brew install rtk` + `rtk gain` never triggers consent → no salt → `(no salt file)` is **expected**, not a failure.

The gating chain is documented inline on the new helper so future refactors don't have to re-derive it.

## Fix

`src/core/telemetry_cmd.rs`:

- `salt_missing_label(consent_given) -> &'static str` returns:
  - `(no salt file)` when `consent_given == Some(true)` — real failure, original label preserved.
  - `(telemetry not enabled; run \`rtk telemetry enable\` to opt in)` otherwise — actionable. `Some(false)` (explicit opt-out) and `None` (never prompted) collapse because the remediation is identical.
- `device_hash_line(consent_given, hash) -> String` formats the full `  device hash:   ...` line. `run_status()` builds the `Option<String>` hash from `salt_path.exists()` and delegates.
- ASCII-only punctuation (semicolon, not em-dash) to match surrounding `println!` and stay grep / pipe friendly.

## Tests

- [x] `cargo fmt --all --check && cargo clippy --all-targets -- -D warnings && cargo test` — 1949 passed (3 new in `telemetry_cmd::tests`)
- [x] `salt_missing_label_distinguishes_consent_not_given_from_real_failure` — all 3 consent states
- [x] `device_hash_line_renders_truncated_hash_when_salt_exists` — consent-agnostic salt-present path
- [x] `device_hash_line_routes_to_salt_missing_label_when_no_hash` — locks the routing so inlining or removing the helper without copying its literals fails CI
- [x] Manual: `rtk telemetry status` output for `salt_path.exists()` users stays byte-identical (only the `else` branch changed)

## Scope

Single-file change, ~85 LOC (test included). Deliberately does **not**:

- Touch `get_or_create_salt()` / `send_ping()` / consent prompt flow — no fs bug exists.
- Pre-create salt at brew postinstall — orthogonal design decision for maintainers.
- Trigger `prompt_telemetry_consent()` from `rtk gain` / `rtk telemetry status` — UX surface change, separate discussion.

## Follow-up

Filed separately as #2106: `cargo test` writes a real `.device_salt` to the user's data dir because the unit tests in `src/core/telemetry.rs` (`test_salt_is_persisted`, `test_device_hash_is_stable`, `test_device_hash_is_valid_hex`) call the production functions with no path override. Hides the symptom of this issue on dev machines after the test suite runs. Not bundled here per [CONTRIBUTING scope rules](https://github.com/rtk-ai/rtk/blob/master/CONTRIBUTING.md#scope-rules).

### DIFF
--- a/src/core/telemetry_cmd.rs
+++ b/src/core/telemetry_cmd.rs
@@ -6,21 +6,33 @@ const TELEMETRY_DISABLED_VALUE: &str = "1";
 
 /// Label for the `device hash` row when no salt file exists.
 ///
-/// `(no salt file)` is only meaningful when consent has been granted —
-/// otherwise no salt is ever attempted, by design. Issue #1656.
-///
-/// Gating proof (rg verified on develop HEAD): `generate_device_hash()` has
-/// three production call sites — `send_ping()` at `telemetry.rs:71` (gated by
-/// `consent_given == Some(true)` at `telemetry.rs:38-41`), `run_status()` and
-/// `run_forget()` (both only call it after `salt_path.exists()`). So the salt
-/// is only ever written through `send_ping()`'s consent-gated path.
-fn salt_missing_label(consent_given: Option<bool>) -> &'static str {
+/// A missing salt is only ever a real failure once every gate that stops the
+/// salt from being written has been cleared. The arms below mirror the gate
+/// order in `telemetry::maybe_ping`, so the label names the gate that actually
+/// fired instead of blaming the filesystem. Issue #1656.
+fn salt_missing_label(
+    endpoint_configured: bool,
+    env_override: bool,
+    enabled: bool,
+    consent_given: Option<bool>,
+) -> &'static str {
+    if !endpoint_configured {
+        return "(not applicable; this build has no telemetry endpoint)";
+    }
+    if env_override {
+        return "(blocked by RTK_TELEMETRY_DISABLED)";
+    }
     match consent_given {
-        Some(true) => "(no salt file)",
         // `Some(false)` (explicit opt-out) and `None` (never prompted) collapse
         // because the user-facing remediation — running `rtk telemetry enable`
         // — is identical in both states.
         Some(false) | None => "(telemetry not enabled; run `rtk telemetry enable` to opt in)",
+        Some(true) if !enabled => "(telemetry disabled in config.toml)",
+        // Every gate is clear, so the salt is simply not there yet. This is true
+        // both when no ping has landed (the 23 h interval, or a run that touched
+        // the marker before the detached thread wrote the salt) and when a write
+        // genuinely failed — as far as the available state lets us go.
+        Some(true) => "(no salt file yet; written on the first ping)",
     }
 }
 
@@ -28,7 +40,13 @@ fn salt_missing_label(consent_given: Option<bool>) -> &'static str {
 ///
 /// Extracted so the routing (which message string for which state) is covered
 /// by unit tests rather than only the leaf label helper.
-fn device_hash_line(consent_given: Option<bool>, hash: Option<&str>) -> String {
+fn device_hash_line(
+    endpoint_configured: bool,
+    env_override: bool,
+    enabled: bool,
+    consent_given: Option<bool>,
+    hash: Option<&str>,
+) -> String {
     match hash {
         // The device hash is a SHA-256 hex digest — exactly 64 chars by
         // construction. Gate on `== 64` (not `>= 64`) so a malformed hash of any
@@ -38,7 +56,10 @@ fn device_hash_line(consent_given: Option<bool>, hash: Option<&str>) -> String {
         Some(h) if h.len() == 64 => {
             format!("  device hash:   {}...{}", &h[..8], &h[56..])
         }
-        _ => format!("  device hash:   {}", salt_missing_label(consent_given)),
+        _ => format!(
+            "  device hash:   {}",
+            salt_missing_label(endpoint_configured, env_override, enabled, consent_given)
+        ),
     }
 }
 
@@ -105,7 +126,13 @@ fn run_status() -> Result<()> {
     };
     println!(
         "{}",
-        device_hash_line(config.telemetry.consent_given, hash.as_deref())
+        device_hash_line(
+            option_env!("RTK_TELEMETRY_URL").is_some(),
+            env_override,
+            config.telemetry.enabled,
+            config.telemetry.consent_given,
+            hash.as_deref(),
+        )
     );
 
     println!();
@@ -283,21 +310,52 @@ mod tests {
         // therefore never writes a salt — that is by design, not a failure.
         // Surfacing it as `(no salt file)` reads as an error to users.
         let not_enabled = "(telemetry not enabled; run `rtk telemetry enable` to opt in)";
-        assert_eq!(salt_missing_label(None), not_enabled);
-        assert_eq!(salt_missing_label(Some(false)), not_enabled);
-        // Consent was granted but the salt is still missing — this IS a real
-        // failure (e.g. fs permission, full disk) and keeps the original label
-        // so the user knows to investigate.
-        assert_eq!(salt_missing_label(Some(true)), "(no salt file)");
+        assert_eq!(salt_missing_label(true, false, true, None), not_enabled);
+        assert_eq!(
+            salt_missing_label(true, false, true, Some(false)),
+            not_enabled
+        );
+        // Every gate is clear and the salt is still missing, so the label says
+        // what is actually known rather than accusing the filesystem.
+        assert_eq!(
+            salt_missing_label(true, false, true, Some(true)),
+            "(no salt file yet; written on the first ping)"
+        );
+    }
+
+    #[test]
+    fn salt_missing_label_names_the_gate_that_actually_fired() {
+        // Consent is one of five gates on salt creation, and three of the others
+        // produce "consent granted, no salt" legitimately. The arms are ordered
+        // as `telemetry::maybe_ping` applies its gates, so the first unmet gate
+        // wins even when a later one is unmet too.
+        assert_eq!(
+            salt_missing_label(false, true, false, Some(false)),
+            "(not applicable; this build has no telemetry endpoint)"
+        );
+        assert_eq!(
+            salt_missing_label(true, true, false, Some(false)),
+            "(blocked by RTK_TELEMETRY_DISABLED)"
+        );
+        assert_eq!(
+            salt_missing_label(true, false, false, Some(true)),
+            "(telemetry disabled in config.toml)"
+        );
     }
 
     #[test]
     fn device_hash_line_renders_truncated_hash_when_salt_exists() {
-        // The salt-exists path is consent-agnostic: once the hash is in hand
-        // it's safe to show regardless of how the salt got there.
+        // The salt-exists path is gate-agnostic: once the hash is in hand it's
+        // safe to show regardless of how the salt got there.
         let expected = "  device hash:   01234567...89abcdef";
-        assert_eq!(device_hash_line(Some(true), Some(SAMPLE_HASH)), expected);
-        assert_eq!(device_hash_line(None, Some(SAMPLE_HASH)), expected);
+        assert_eq!(
+            device_hash_line(true, false, true, Some(true), Some(SAMPLE_HASH)),
+            expected
+        );
+        assert_eq!(
+            device_hash_line(false, true, false, None, Some(SAMPLE_HASH)),
+            expected
+        );
     }
 
     #[test]
@@ -305,16 +363,28 @@ mod tests {
         // Locks the routing: removing or inlining `salt_missing_label` without
         // copying its literals will fail these asserts.
         assert_eq!(
-            device_hash_line(Some(true), None),
-            "  device hash:   (no salt file)"
+            device_hash_line(true, false, true, Some(true), None),
+            "  device hash:   (no salt file yet; written on the first ping)"
         );
         assert_eq!(
-            device_hash_line(None, None),
+            device_hash_line(true, false, true, None, None),
             "  device hash:   (telemetry not enabled; run `rtk telemetry enable` to opt in)"
         );
         assert_eq!(
-            device_hash_line(Some(false), None),
+            device_hash_line(true, false, true, Some(false), None),
             "  device hash:   (telemetry not enabled; run `rtk telemetry enable` to opt in)"
+        );
+        assert_eq!(
+            device_hash_line(false, false, true, Some(true), None),
+            "  device hash:   (not applicable; this build has no telemetry endpoint)"
+        );
+        assert_eq!(
+            device_hash_line(true, true, true, Some(true), None),
+            "  device hash:   (blocked by RTK_TELEMETRY_DISABLED)"
+        );
+        assert_eq!(
+            device_hash_line(true, false, false, Some(true), None),
+            "  device hash:   (telemetry disabled in config.toml)"
         );
     }
 
@@ -327,12 +397,12 @@ mod tests {
         let short = "0123456789abcdef"; // 16 chars
         let long = SAMPLE_HASH.repeat(2); // 128 chars
         assert_eq!(
-            device_hash_line(Some(true), Some(short)),
-            "  device hash:   (no salt file)"
+            device_hash_line(true, false, true, Some(true), Some(short)),
+            "  device hash:   (no salt file yet; written on the first ping)"
         );
         assert_eq!(
-            device_hash_line(Some(true), Some(&long)),
-            "  device hash:   (no salt file)"
+            device_hash_line(true, false, true, Some(true), Some(&long)),
+            "  device hash:   (no salt file yet; written on the first ping)"
         );
     }
 }
--- a/src/core/telemetry_cmd.rs
+++ b/src/core/telemetry_cmd.rs
@@ -4,6 +4,44 @@ use clap::Subcommand;
 const TELEMETRY_DISABLED_ENV: &str = "RTK_TELEMETRY_DISABLED";
 const TELEMETRY_DISABLED_VALUE: &str = "1";
 
+/// Label for the `device hash` row when no salt file exists.
+///
+/// `(no salt file)` is only meaningful when consent has been granted —
+/// otherwise no salt is ever attempted, by design. Issue #1656.
+///
+/// Gating proof (rg verified on develop HEAD): `generate_device_hash()` has
+/// three production call sites — `send_ping()` at `telemetry.rs:71` (gated by
+/// `consent_given == Some(true)` at `telemetry.rs:38-41`), `run_status()` and
+/// `run_forget()` (both only call it after `salt_path.exists()`). So the salt
+/// is only ever written through `send_ping()`'s consent-gated path.
+fn salt_missing_label(consent_given: Option<bool>) -> &'static str {
+    match consent_given {
+        Some(true) => "(no salt file)",
+        // `Some(false)` (explicit opt-out) and `None` (never prompted) collapse
+        // because the user-facing remediation — running `rtk telemetry enable`
+        // — is identical in both states.
+        Some(false) | None => "(telemetry not enabled; run `rtk telemetry enable` to opt in)",
+    }
+}
+
+/// Format the full `  device hash:   ...` status line.
+///
+/// Extracted so the routing (which message string for which state) is covered
+/// by unit tests rather than only the leaf label helper.
+fn device_hash_line(consent_given: Option<bool>, hash: Option<&str>) -> String {
+    match hash {
+        // The device hash is a SHA-256 hex digest — exactly 64 chars by
+        // construction. Gate on `== 64` (not `>= 64`) so a malformed hash of any
+        // other length routes to the salt-missing label instead of slicing
+        // arbitrary bytes — `&h[56..]` on a longer string would print an
+        // over-long tail.
+        Some(h) if h.len() == 64 => {
+            format!("  device hash:   {}...{}", &h[..8], &h[56..])
+        }
+        _ => format!("  device hash:   {}", salt_missing_label(consent_given)),
+    }
+}
+
 #[derive(Debug, Subcommand)]
 pub enum TelemetrySubcommand {
     Status,
@@ -60,12 +98,15 @@ fn run_status() -> Result<()> {
     }
 
     let salt_path = super::telemetry::salt_file_path();
-    if salt_path.exists() {
-        let hash = super::telemetry::generate_device_hash();
-        println!("  device hash:   {}...{}", &hash[..8], &hash[56..]);
+    let hash = if salt_path.exists() {
+        Some(super::telemetry::generate_device_hash())
     } else {
-        println!("  device hash:   (no salt file)");
-    }
+        None
+    };
+    println!(
+        "{}",
+        device_hash_line(config.telemetry.consent_given, hash.as_deref())
+    );
 
     println!();
     println!("Data controller: RTK AI Labs, contact@rtk-ai.app");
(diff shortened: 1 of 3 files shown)


truffle-dev:
Read the diff against `develop` HEAD (`878af7d`) and traced the gating chain end-to-end. LGTM.

**Gating proof verified.** `generate_device_hash()` has the three production callers you named, and only one of them writes the salt — `send_ping()` at `telemetry.rs:71`, gated by `match cfg.telemetry.consent_given { Some(true) => {}, Some(false) | None => return }` at `telemetry.rs:38-41`. `run_status()` and `run_forget()` (this file at `21` and `109`) both guard the call with `salt_path.exists()` first, so neither can create the salt on a no-consent install. `get_or_create_salt()` at `telemetry.rs:162-186` does `create_dir_all(parent)` before writing — the parent isn't the failure surface.

So the proposed mapping is right: `Some(true)` no-salt is the only state that should keep `(no salt file)`, and collapsing `Some(false) | None` into the same opt-in remediation matches the actual user-visible action.

**Tests lock the routing.** All three are pure-function tests over `salt_missing_label` / `device_hash_line`, no fs, no `Config::load`. The third one in particular (`device_hash_line_routes_to_salt_missing_label_when_no_hash`) is the one that catches a future cleanup re-inlining the literals without preserving them. Good shape.

**One observation, not a blocker.** `device_hash_line`'s `Some(h) if h.len() >= 64` guard is wider than the original. The pre-PR code did `&hash[..8]` and `&hash[56..]` unconditionally — a malformed hash would panic. The new code silently routes a `len < 64` hash to `salt_missing_label`. SHA-256 is 64 hex chars by construction and `get_or_create_salt()` re-verifies cached salt with `trimmed.len() == 64 && trimmed.chars().all(...)` at `telemetry.rs:169`, so the guard never fires in practice. Worth keeping as defense-in-depth; just noting that the routing for a sub-64 hash would read as "telemetry not enabled" under `Some(false) | None` consent, which is the wrong message for that imaginary failure mode. If a maintainer prefers preserving the panic-on-malformed behavior, the gate could be `Some(h) if h.len() == 64`.

Follow-up #2106 is a separate fix in scope and is genuinely orthogonal — the test-pollution writes a real `.device_salt` regardless of which label the status command renders for it.

maxmilian:
Good catch — adopted in `56f6afa`. Tightened the guard to `== 64` and added `device_hash_line_treats_non_64_char_hash_as_missing` to lock it: a non-64-length hash now routes to the salt-missing label instead of slicing an over-long tail. Left it as exact-match rather than defense-in-depth since the digest is 64 hex chars by construction.

truffle-dev:
Confirmed against `56f6afa`. The `== 64` exact-match gate at `device_hash_line` is the right call given the digest is 64 hex chars by construction, and `device_hash_line_treats_non_64_char_hash_as_missing` (both the 16-char short and the 128-char long branch) pins the routing so a future cleanup that loosens the guard back to `>= 64` will fail the test. LGTM.

maxmilian:
Friendly check-in — this one's been quiet for a few weeks and looks ready from my side:

- **Reviewed**: @truffle-dev traced the gating chain end-to-end and confirmed LGTM against `56f6afa` (the `== 64` exact-match guard + the `device_hash_line_treats_non_64_char_hash_as_missing` test pinning the routing).
- **CI green**, **mergeable** against `develop`.

No action needed from reviewers if it's just in the queue — just flagging it as merge-ready. Happy to rebase onto the latest `develop` if that helps it land. 🙏

maxmilian:
Gentle bump for @aeppling 🙏 — this one's been quiet for a few weeks and looks ready:

- **Reviewed & LGTM'd twice** by @truffle-dev, who traced the gating chain end-to-end and confirmed the `== 64` exact-match guard against `56f6afa`.
- Behavior is pinned by `device_hash_line_treats_non_64_char_hash_as_missing`, and the CLA check is green.

Anything I can do to help move it along? Happy to rebase on `develop` if it's drifted. Thanks!

pszymkowiak (approved):
Tested all three consent states live with the compiled binary — `None`, `Some(false)`, and `Some(true)` (no salt) each render exactly as described. `cargo fmt`, `cargo clippy --all-targets`, `cargo test --all` all clean (1950/1950). Appreciate the precise scoping and the separate follow-up issue (#2106) instead of scope creep.

maxmilian:
Rebased onto latest `develop` to clear the merge conflict — it came from the env opt-out refactor (`single source of truth for env opt-out check`) landing in the same file. Resolved as a union: kept that refactor's `TELEMETRY_DISABLED_ENV` constants + `test_telemetry_disabled_by_env_honors_opt_out`, plus this PR's `salt_missing_label` / `device_hash_line` helpers and their tests. No logic changes on either side.

`cargo test telemetry` green (20 passed, including both the env opt-out test and this PR's four new ones). Back to `MERGEABLE` / approved.

maxmilian:
@aeppling — this one looks ready to land whenever you have a moment:

- **Approved** by @pszymkowiak on Jul 9 (after @truffle-dev traced the gating chain end-to-end and LGTM'd twice).
- **Rebased onto latest `develop` on Jul 21** to clear the conflict from the env opt-out refactor — resolved as a union, both sides kept.
- `cargo test telemetry --bins`: 20 passed (my 4 + the develop env test).
- Currently `MERGEABLE` / clean, CI green, CLA signed.

No rush on my side — just flagging that there's nothing left blocking it technically. Happy to rebase again if `develop` moves.

maxmilian:
Friendly ping — this one has an approval and is a single-file change, so I think it just needs a merge.

Recapping for whoever picks it up: the telemetry path currently can't tell *"the user declined consent"* apart from *"the consent setting was never written"*, so a declined install gets treated as unconfigured on every subsequent run. The fix distinguishes the two states; one file, no behavior change for users who already consented.

Also — separate note, and the more useful half of this comment: **I'm going to reduce my own open-PR count here.** I opened seven PRs across 26–28 May, which in hindsight was too many at once from someone with no track record in this repo. I'm closing the four that have had no review activity (#2143, #2141, #2140, #2122) so the queue reflects what actually matters, and keeping this one plus #2142 (which has a real user report behind it — #1830 documents `dotnet test` running 1208× over 30 days at ~0% compression).

No urgency on either. Just didn't want to leave a wall of stale PRs sitting on your review list.

src/core/telemetry_cmd.rs
```diff
@@ -4,6 +4,44 @@ use clap::Subcommand;
...
+///
+/// `(no salt file)` is only meaningful when consent has been granted —
+/// otherwise no salt is ever attempted, by design. Issue #1656.
+///
+/// Gating proof (rg verified on develop HEAD): `generate_device_hash()` has
+/// three production call sites — `send_ping()` at `telemetry.rs:71` (gated by
+/// `consent_given == Some(true)` at `telemetry.rs:38-41`), `run_status()` and
+/// `run_forget()` (both only call it after `salt_path.exists()`). So the salt
+/// is only ever written through `send_ping()`'s consent-gated path.
+fn salt_missing_label(consent_given: Option<bool>) -> &'static str {
```
KuSh: `maybe_ping()` returns at `telemetry.rs:23-26` when `TELEMETRY_URL.is_none()`, and that URL comes from `option_env!("RTK_TELEMETRY_URL")`, which only `.github/workflows/release.yml` injects. Any build without it — `cargo install --path .` (the install command in CLAUDE.md), a distro or Homebrew source build — can never write a salt, by design; `docs/TELEMETRY.md:180` says the telemetry code is dead in that case.

So a source-installed user who runs `rtk telemetry enable` and then `rtk telemetry status` gets `(no salt file)` reported as a real failure, permanently. `salt_missing_label` can't tell, because `consent_given` is all it receives.

src/core/telemetry_cmd.rs
```diff
@@ -4,6 +4,44 @@ use clap::Subcommand;
...
+    match consent_given {
+        Some(true) => "(no salt file)",
```
KuSh: `RTK_TELEMETRY_DISABLED=1` is checked at `telemetry.rs:29-30`, *before* the consent gate at `:38-41` — so a user who ran `rtk telemetry enable` and later picked up the env var (a CI image, an enterprise shell profile) has `consent_given == Some(true)` and no salt, entirely by design.

`rtk telemetry status` then prints both lines together:

```
  env override:  RTK_TELEMETRY_DISABLED=1 (blocked)
  device hash:   (no salt file)
```

which sends them looking for a disk or permission problem that doesn't exist. `env_override` is already computed at line 88 and just needs to reach here.

`config.telemetry.enabled == false` with consent granted (`telemetry.rs:44-46`) has the same shape. It's the weaker of the two — `save_telemetry_consent` keeps the two fields in lockstep, so it takes a hand-edited or externally-managed `config.toml` to reach — but it costs nothing to cover once `enabled` is being passed down.

src/core/telemetry_cmd.rs
```diff
@@ -4,6 +4,44 @@ use clap::Subcommand;
 const TELEMETRY_DISABLED_ENV: &str = "RTK_TELEMETRY_DISABLED";
 const TELEMETRY_DISABLED_VALUE: &str = "1";
 
+/// Label for the `device hash` row when no salt file exists.
+///
+/// `(no salt file)` is only meaningful when consent has been granted —
+/// otherwise no salt is ever attempted, by design. Issue #1656.
```
KuSh: "otherwise no salt is ever attempted" holds, but the converse doesn't: consent granted doesn't mean a salt was ever attempted *either*, even on the fully happy path.

`maybe_ping()` calls `touch_marker(&marker)` at `telemetry.rs:62` — before `std::thread::spawn` at `:65`. The salt is only written inside that detached thread, via `send_ping()` → `generate_device_hash()` → `get_or_create_salt()`. rtk targets <10 ms startup and never joins the thread, so the process can exit before it's scheduled. And because the marker was already touched, the 23 h interval check at `:50-58` then suppresses every retry.

`rtk telemetry enable && rtk telemetry status` can therefore print `(no salt file)` as a failure, and stay that way for ~23 h — or indefinitely on a machine that consistently loses the race.

src/core/telemetry_cmd.rs
```diff
@@ -4,6 +4,44 @@ use clap::Subcommand;
...
+///
+/// Gating proof (rg verified on develop HEAD): `generate_device_hash()` has
```
KuSh: This is the root cause of the three comments above, and it's worth fixing even after the behavior is: the invariant as stated is narrower than reality.

`send_ping()` is indeed unreachable except through `maybe_ping()` — but `maybe_ping()` has four gates before it (compiled-in URL, env opt-out, consent, `telemetry.enabled`) plus the interval check, and consent is only the third. "The salt is only ever written through `send_ping()`'s consent-gated path" is true and yet reads as "consent is what decides", which is what the two-state label was built on.

Worth stating the full gate list here, since this comment is what the next person to touch this code will trust.

src/core/telemetry_cmd.rs
```diff
@@ -4,6 +4,44 @@ use clap::Subcommand;
...
+/// Extracted so the routing (which message string for which state) is covered
+/// by unit tests rather than only the leaf label helper.
+fn device_hash_line(consent_given: Option<bool>, hash: Option<&str>) -> String {
+    match hash {
+        // The device hash is a SHA-256 hex digest — exactly 64 chars by
+        // construction. Gate on `== 64` (not `>= 64`) so a malformed hash of any
+        // other length routes to the salt-missing label instead of slicing
+        // arbitrary bytes — `&h[56..]` on a longer string would print an
+        // over-long tail.
+        Some(h) if h.len() == 64 => {
```
KuSh: `h.len()` is a *byte* count, so this guard doesn't deliver the safety the comment claims. A 64-byte non-ASCII string passes it and then panics in the body: with 3-byte characters, byte index 8 lands mid-character and `&h[..8]` is a non-char-boundary slice.

Unreachable today — `generate_device_hash()` always returns ASCII hex — but if the guard exists to stop "slicing arbitrary bytes", it should actually stop them:

```suggestion
        Some(h) if h.len() == 64 && h.bytes().all(|b| b.is_ascii_hexdigit()) => {
```

src/core/telemetry_cmd.rs
```diff
@@ -4,6 +4,44 @@ use clap::Subcommand;
...
+            format!("  device hash:   {}...{}", &h[..8], &h[56..])
+        }
+        _ => format!("  device hash:   {}", salt_missing_label(consent_given)),
```
KuSh: When `hash` is `Some(malformed)` this falls through to `salt_missing_label`, which under `Some(true)` prints `(no salt file)` — but `run_status()` only passes `Some(hash)` when `salt_path.exists()` is true (lines 100-106). The message asserts something the caller already knows to be false.

`device_hash_line_treats_non_64_char_hash_as_missing` (line 322) pins that wrong string as expected behavior. The path is unreachable in production, which is also why that test is exercising dead code — a `(malformed device hash)` label would at least be honest, and would give the test something real to assert.

KuSh (changes requested):
Requesting changes. The split is the right idea and the tests are well-targeted, but it keys off `consent_given` alone — and consent is only one of five gates on salt creation. Three of the others produce "consent granted + no salt" legitimately, and in each of those `rtk telemetry status` now reports a filesystem failure that isn't happening. That's the same false-alarm class #1656 is about, moved onto a different set of users.

The gates, in the order `maybe_ping()` applies them (`src/core/telemetry.rs:22-67`):

1. `TELEMETRY_URL.is_none()` — no endpoint compiled into this build (`:23-26`)
2. `telemetry_disabled_by_env()` — `RTK_TELEMETRY_DISABLED=1` (`:29-30`)
3. `consent_given != Some(true)` (`:38-41`) — the only one this PR models
4. `!cfg.telemetry.enabled` (`:44-46`)
5. the 23 h interval check (`:50-58`), with the salt itself written on a detached thread (`:65-67`)

`run_status()` already holds everything needed for 1, 2 and 4: `env_override` is computed at line 88, `config.telemetry.enabled` at 82, and `send_erasure_request` (line 176) already reads `option_env!("RTK_TELEMETRY_URL")` directly, so no new accessor in `telemetry.rs` is required.

## Suggested fix

Give the label the rest of the state, and order the checks the way `maybe_ping()` orders its gates so the message names the one that actually fired:

```rust
fn salt_missing_label(
    endpoint_configured: bool,
    env_override: bool,
    enabled: bool,
    consent_given: Option<bool>,
) -> &'static str {
    // Mirrors the gate order in `telemetry::maybe_ping`.
    if !endpoint_configured {
        return "(not applicable; this build has no telemetry endpoint)";
    }
    if env_override {
        return "(blocked by RTK_TELEMETRY_DISABLED)";
    }
    match consent_given {
        Some(false) | None => "(telemetry not enabled; run `rtk telemetry enable` to opt in)",
        Some(true) if !enabled => "(telemetry disabled in config.toml)",
        Some(true) => "(no salt file yet; written on the first ping)",
    }
}
```

and at the call site:

```rust
    println!(
        "{}",
        device_hash_line(
            option_env!("RTK_TELEMETRY_URL").is_some(),
            env_override,
            config.telemetry.enabled,
            config.telemetry.consent_given,
            hash.as_deref(),
        )
    );
```

Five positional params is a lot — passing `&config` plus `env_override` is a fine alternative if you'd rather keep the signature narrow; the existing tests adapt either way.

The last arm is deliberately no longer an accusation: `(no salt file yet; written on the first ping)` is true both when the ping simply hasn't landed and when it genuinely failed, which is as far as the available state lets us go.

Everything else checks out: the `hash.as_deref()` plumbing is correct, the `01234567...89abcdef` expectation matches `&h[..8]` / `&h[56..]` on the sample, the salt-exists path is byte-identical to pre-PR output as the description claims, and the branch still applies cleanly on `develop` after the July rebase.

maxmilian:
Thanks @KuSh — the gap is real and I'm adopting your fix. One correction on the framing first, then what changed.

**This doesn't move the false alarm onto a different set of users; it shrinks the set and then stops short.** Before this PR, *every* no-salt install printed `(no salt file)` — including all four states you list. The PR routes `Some(false) | None` to the opt-in message and leaves the rest on the old string, so the users you're describing see exactly what they see on `develop` today. That doesn't make them less worth fixing: your actual point — that `consent_given` alone can't carry this label — stands, and in those three states `(no salt file)` still reads as a filesystem failure that isn't happening.

**Verified your gate list against `develop` @ `9a695d1`** (`src/core/telemetry.rs`; line numbers sit a couple lower than in your note):

- `TELEMETRY_URL.is_none()` — `:24-26`
- `telemetry_disabled_by_env()` — `:29-31`
- `consent_given != Some(true)` — `:40-43`
- `!cfg.telemetry.enabled` — `:46-48`
- the 23 h interval, then the salt written on the detached thread — `:51-61` and `:66`

And the state really is all in hand at the call site (`src/core/telemetry_cmd.rs`): `env_override` at `:50`, `config.telemetry.enabled` at `:44`, and `send_erasure_request` already reads `option_env!("RTK_TELEMETRY_URL")` at `:176`.

**Adopted `salt_missing_label` as you wrote it**, ordered to mirror `maybe_ping`'s gates, with the `option_env!` read at the call site so this stays a single-file change. I kept the four positional params rather than taking `&config`: the existing tests are pure-function calls that never construct a `Config`, and I'd rather not make them start.

Two notes on the result, neither of which changes the shape:

- `run_status` already prints `enabled: no` and `env override: RTK_TELEMETRY_DISABLED=1 (blocked)` as their own rows, so two of the new labels restate a line printed just above. That's fine — the device-hash row is the one that reads as an error, and it should stop doing that on its own terms.
- The interval gate needs no arm: `(no salt file yet; written on the first ping)` is already true for the case where the marker was touched but the detached thread never got the salt written. That's spelled out in the comment on the arm so nobody adds a fifth branch for it later.

The four existing tests moved to the new signatures, the routing test gained one case per new arm, and a new `salt_missing_label_names_the_gate_that_actually_fired` pins the ordering. I checked that last one actually bites rather than trusting it: changing the env gate to `if env_override && consent_given == Some(true)` — the plausible way to get the order wrong — turns it red with

```
  left: "(telemetry not enabled; run `rtk telemetry enable` to opt in)"
 right: "(blocked by RTK_TELEMETRY_DISABLED)"
test result: FAILED. 20 passed; 1 failed
```

and green again once reverted.

Validation on the new head:

- `cargo fmt --check` — exit 0
- `cargo clippy --all-targets -- -D warnings` — exit 0, `Finished dev profile ... in 8.57s`, no warnings
- `cargo test telemetry --bins` — **21 passed; 0 failed** (was 20; the five `telemetry_cmd` tests include the new ordering test)
- `cargo test --all` — **2443 passed; 0 failed; 8 ignored** in the main binary, plus 6 / 11 / 11 / 14 / 5 / 11 passed across the integration targets, 0 failed anywhere

src/core/telemetry_cmd.rs
KuSh: Fixed in `f40f0a1`. `endpoint_configured` is threaded through from `option_env!("RTK_TELEMETRY_URL").is_some()` at the call site, and a build with no endpoint compiled in now reports `(not applicable; this build has no telemetry endpoint)` instead of an imaginary filesystem failure. Verified against a local source build: `cargo run -- telemetry status` takes that arm.

src/core/telemetry_cmd.rs
KuSh: Fixed in `f40f0a1`. `env_override` and `config.telemetry.enabled` both reach the label now, and the arms are ordered the way `maybe_ping()` applies its gates, so the first unmet gate wins — `(blocked by RTK_TELEMETRY_DISABLED)` and `(telemetry disabled in config.toml)` respectively. `salt_missing_label_names_the_gate_that_actually_fired` pins that ordering.

src/core/telemetry_cmd.rs
KuSh: Fixed in `f40f0a1`. The final arm is `(no salt file yet; written on the first ping)`, which is true both when the detached thread never landed and when a write genuinely failed — no longer an accusation the status output cannot support.

src/core/telemetry_cmd.rs
KuSh: Fixed in `f40f0a1`. The doc comment now points at the gate order in `telemetry::maybe_ping` rather than asserting a consent-only invariant, so the next person to touch this reads the constraint the code actually has.

src/core/telemetry_cmd.rs
KuSh: Still open at `f40f0a1` — the guard is unchanged. Not a blocker (unreachable while `generate_device_hash()` returns ASCII hex), but the suggestion above is the whole fix: `&& h.bytes().all(|b| b.is_ascii_hexdigit())`.

src/core/telemetry_cmd.rs
KuSh: Still open at `f40f0a1`, and the wording moved rather than resolved: a malformed `Some(hash)` now prints `(no salt file yet; written on the first ping)` in the one branch where `run_status()` already knows the salt file exists (lines 121-125). `device_hash_line_treats_non_64_char_hash_as_missing` (line 392) still pins that string. Unreachable in production, so a nit — but a `(malformed device hash)` arm would make both the message and the test honest.

src/core/telemetry_cmd.rs
KuSh: Fixed in `1255088` (pushed to this branch). The guard moved into `is_device_hash`, which checks the hex alphabet alongside the length, so `&h[..8]` and `&h[56..]` stay on char boundaries for any input. `device_hash_line_rejects_a_64_byte_non_hex_hash_without_panicking` covers the 64-byte multi-byte case that used to panic.

src/core/telemetry_cmd.rs
KuSh: Fixed in `1255088` (pushed to this branch). `device_hash_line` now has three arms — well-formed, malformed, absent — so a present-but-malformed salt reports `(malformed device hash)` instead of denying the file exists. The test was renamed and now asserts that string.

KuSh (approved):
Approving on `bb85017`. The gate-order rewrite in `f40f0a1` addressed every blocking point — `salt_missing_label` now names the gate that actually fired, and the ordering is pinned by a test. Nice work.

Two housekeeping things I did directly on the branch, since you'd enabled maintainer edits:

**Rebased onto current `develop` (`4f3c9fe`).** The branch was ~300 commits behind and wouldn't have been mergeable. One conflict, in `src/core/utils.rs`, where `develop` added tests at the same spot as mine — resolved by keeping both. Your two commits are unchanged apart from the new base and keep their authorship.

**Pushed `bb85017`** for the three remaining review points:

- `option_env!("RTK_TELEMETRY_URL").is_some()` treated an empty URL as a configured endpoint. `.github/workflows/release.yml` passes the URL through an `env:` block fed by `vars.RTK_TELEMETRY_URL`, and an unset repository variable is exported as the empty string — so `option_env!` compiles to `Some("")` and a release that can never ping would still have blamed the salt. New `utils::env_is_some` covers the compile-time env check generally, and every `RTK_TELEMETRY_URL` read now goes through `telemetry::endpoint_url`, so the ping path, the status label and the erasure request can't drift apart.
- A present-but-malformed salt reported `(no salt file yet; written on the first ping)` in the one branch where `run_status` already knows the file exists. `device_hash_line` now has three arms — well-formed, malformed, absent.
- The `len() == 64` guard was a byte check, so a 64-byte multi-byte string passed it and then panicked on `&h[..8]`. `is_device_hash` checks the alphabet too; there's a regression test for that exact input.

Verified on the rebased head: `cargo fmt --all --check` clean, `cargo clippy --all-targets` zero warnings, `cargo test --all` 2984 passed / 0 failed.

PR #2107, opened 2026-05-26
fix(git): don't count --max-lines truncation as savings (#1561)

### LINKED ISSUE #1561: rtk git diff/show --max-lines may inflate savings same way as #1045

maxmilian:
## Summary

Fixes #1561. `rtk git diff` and `rtk git show` were passing the user's `--max-lines N` directly into `compact_diff`, which compresses and truncates in the same pass. The post-truncation text was then handed to `timer.track`, so bytes the user themselves asked to hide were being recorded as RTK token savings.

## Root cause

`run_diff` (src/cmds/git/git.rs) and `run_show` both called:

```rust
let compacted = compact_diff(&diff_result.stdout, max_lines.unwrap_or(500));
// ...
timer.track(..., &final_output_containing_compacted);
```

`compact_diff` internally stops at `result.len() >= max_lines` (git.rs:454) — so `--max-lines 5` made it stop at 5 lines, and `timer.track` saw a 5-line output vs an N-thousand-line input, computing a huge savings number. This is the exact pattern @pszymkowiak flagged in #1045's PR-review.

## Fix

Three new symbols in `src/cmds/git/git.rs`:

- `pub(crate) const RTK_DIFF_BUDGET: usize = 500` — RTK's standard compression budget, used by both `run_diff` and `run_show` regardless of user flags. Anything beyond this is real RTK compression and legitimately counts as savings. (Doc-comment notes the intentionally-separate `gh_cmd.rs::gh pr diff` 500 and `git stash show` 100 budgets.)
- `fn truncate_to_lines(text, n)` (private) — applies the user's `--max-lines` cap on top of already-compacted text. Display-only, never tracked. Emits the terse `... (N lines hidden by --max-lines)` hint to match `compact_diff`'s existing `[full diff: rtk git diff --no-compact]` style. Panic-safe at `n == 0` (collapses to single hint line, no leading blank).
- `pub(crate) fn prepare_diff_section(raw, user_max) -> (displayed, tracked)` — central helper called by both `run_diff` and `run_show`. `tracked` is always `compact_diff(raw, RTK_DIFF_BUDGET)`, independent of user input. Doc explains the "tracked = output RTK would have produced absent the user's display knob" rationale.

`run_diff` and `run_show` are reduced to:

```rust
let (displayed, tracked) = prepare_diff_section(&raw, max_lines);
println!("{}", displayed);
// ... timer.track with `tracked` ...
```

## User-visible changes

- **Savings figures** for `rtk git diff --max-lines N` / `rtk git show --max-lines N` will be lower (the bug fix — they were over-counted before).
- **Hint wording** when the user cap fires changes to `... (M lines hidden by --max-lines)` (terse, matches `compact_diff`'s existing trailer style).
- **Display content** for `--max-lines N` shifts subtly: previously you saw the first N lines of a `compact_diff(diff, N)` invocation (which stops mid-compaction), now you see the first N lines of a `compact_diff(diff, 500)` invocation (which gets further through the diff before truncating). For most real-world `N < 500` cases this means more files appear in the visible head and the per-file hunk caps work consistently.

## Tests

- [x] `cargo fmt --all --check && rtk cargo clippy --all-targets -- -D warnings && rtk cargo test` — **1952 passed**, 6 new in `cmds::git::git::tests`.
- `test_truncate_to_lines_passes_through_when_under_cap` / `_chops_with_terse_hint` / `_zero_cap_keeps_only_hint` — helper unit tests. The zero-cap test explicitly asserts no leading blank line and a single-line total output.
- `test_rtk_diff_budget_value` — locks the constant.
- `test_compact_diff_uses_rtk_budget_independent_of_user_max_lines` — shape-level.
- `test_prepare_diff_section_tracks_against_rtk_budget_not_user_cap` — **routing lock**: iterates user_cap ∈ `{None, 1000, 500, 50, 5}` and asserts `tracked == compact_diff(raw, RTK_DIFF_BUDGET)` for every case AND that `displayed` starts with the literal head of `tracked` (catches buggy refactors that recompute `displayed` independently).

## Scope

Single file. Deliberately does **not**:

- Touch the local tracking DB (polluted historical data — issue body acknowledged as out of scope).
- Fix the pre-existing `verbose`-gated "--- Changes ---" header divergence between `run_diff` (unconditional) and `run_show` (gated on `verbose > 0`). Worth noting for a follow-up but unrelated to savings calculation.
- Reject `--max-lines 0` at the CLI parser. The helper is panic-safe and returns a single hint line; defensive parser-level rejection would be a separate UX-focused PR.
- Migrate `gh_cmd.rs::gh pr diff`'s hardcoded `500` to `RTK_DIFF_BUDGET`. Same concept but cross-module coupling for a one-off callsite is out of scope; the new constant's doc-comment notes this explicitly so the divergence isn't silent.
- Touch `Cargo.toml` / `CHANGELOG.md` / `release-please-manifest.json` (release-please owns those — the previous attempt #1573 hit a CLA blocker partly because of accidental version-bump commits).

## Prior attempt

#1573 by @SAY-5 took a similar approach but was self-closed after 9 days without human review. The wshm bot triage was actually positive (`low risk`, tests present); the blocker was procedural (CLA stuck on a `github-actions[bot]` co-author signature, plus the scope creep above). This PR keeps the approach and avoids the procedural pitfalls.

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -191,20 +191,22 @@ fn run_diff(
 
     let diff_result = exec_capture(&mut diff_cmd).context("Failed to run git diff")?;
 
-    let mut final_output = result.stdout.clone();
+    // Split RTK-side compression (real savings) from user --max-lines
+    // truncation (display-only, no savings — issue #1561).
+    let mut tracked_output = result.stdout.clone();
     if !diff_result.stdout.is_empty() {
+        let (displayed, tracked) = prepare_diff_section(&diff_result.stdout, max_lines);
         println!("\nChanges:");
-        let compacted = compact_diff(&diff_result.stdout, max_lines.unwrap_or(500));
-        println!("{}", compacted);
-        final_output.push_str("\nChanges:\n");
-        final_output.push_str(&compacted);
+        println!("{}", displayed);
+        tracked_output.push_str("\nChanges:\n");
+        tracked_output.push_str(&tracked);
     }
 
     timer.track(
         &format!("git diff {}", args.join(" ")),
         &format!("rtk git diff {}", args.join(" ")),
         &format!("{}\n{}", result.stdout, diff_result.stdout),
-        &final_output,
+        &tracked_output,
     );
 
     Ok(0)
@@ -302,21 +304,23 @@ fn run_show(
     let diff_result = exec_capture(&mut diff_cmd).context("Failed to run git show (diff)")?;
     let diff_text = diff_result.stdout.trim();
 
-    let mut final_output = summary_result.stdout.clone();
+    // Same split as `run_diff`: track RTK compression as savings, treat
+    // user `--max-lines` as display-only truncation (issue #1561).
+    let mut tracked_output = summary_result.stdout.clone();
     if !diff_text.is_empty() {
         if verbose > 0 {
             println!("\nChanges:");
         }
-        let compacted = compact_diff(diff_text, max_lines.unwrap_or(500));
-        println!("{}", compacted);
-        final_output.push_str(&format!("\n{}", compacted));
+        let (displayed, tracked) = prepare_diff_section(diff_text, max_lines);
+        println!("{}", displayed);
+        tracked_output.push_str(&format!("\n{}", tracked));
     }
 
     timer.track(
         &format!("git show {}", args.join(" ")),
         &format!("rtk git show {}", args.join(" ")),
         &raw_output,
-        &final_output,
+        &tracked_output,
     );
 
     Ok(0)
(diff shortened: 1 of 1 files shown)


maxmilian:
Closing this one as a cleanup pass on my older open PRs. This change has sat without review for a while and the surrounding `save`-detection logic on `main` has moved since I opened it, so the diff is no longer a clean fit. I'd rather withdraw it than leave a stale PR parked. Thanks for the project — happy to revisit with a fresh, rebased version if this is still a wanted behavior.

PR #2108, opened 2026-05-26
feat(discover): add fnm (Fast Node Manager) filter and hook rewrite

### LINKED ISSUE #1538: Add `fnm` (Fast Node Manager) handler + hook routing
## Summary

`fnm` is a popular Node version manager (Rust, ~14k stars) with no RTK handler in 0.37.2. Common dev shells initialize Node via `eval "$(fnm env)"` and switch versions via `fnm use`. `rtk hook check` returns "No rewrite" for both. From 30 days of `rtk discover`:

```
fnm use                    126    fnm use 20 2>&1
eval "$(fnm                123    eval "$(fnm env)"
```

249 invocations / 30 days, almost entirely from session-init / Node-version-switch context. Output is normally short (1-3 lines), but the volume + the fact that they appear in nearly every session that touches Node makes them a base-rate pain point — ~5-8% of all Claude Code Bash calls.

## Why a handler matters even though output is small

- **Session init noise:** every shell init runs `eval "$(fnm env)"`. Even 3 lines × thousands of sessions = measurable.
- **Error path is verbose:** `fnm use <missing-version>` prints multi-line install hints. RTK could compress to one-line "missing version 22, run `fnm install 22`".
- **Pairs naturally with existing `npm`/`npx`/`pnpm`/`vitest`/`jest` handlers** — RTK already supports the JS/TS toolchain extensively; `fnm` is the standard env-bootstrap tool for that ecosystem on macOS/Linux.

## Proposed approach (not prescribing)

Two-tier handling, similar to how `pnpm`/`npx` are wrapped:

1. **`rtk fnm use <version>`** — strip the post-switch table that lists installed versions (typically 5-10 lines), keep just `Using Node v<X>`.
2. **`rtk fnm env`** — passthrough as-is (the eval'd output IS the environment; can't compress without breaking).
3. Hook routing whitelist additions: `fnm`, `eval "$(fnm env)"` (the latter needs lexer awareness — see PR #156's lexer for prior art on `command`/`exec`/`noglob` prefix stripping).

## Related

- #599 "Add support for Bun" — open, includes `nvm use 30/30d` in their `rtk discover` output (`nvm` is conceptually similar to `fnm`; could be addressed together as a "Node version manager family")
- #361 / PR #156 (hook engine v2 — `hook_lookup()` whitelist will need fnm entry)

## Environment

- rtk: 0.37.2
- fnm version observed: latest (Linux WSL2)
- Multi-language project (Java/Spring Boot backend + Next.js frontend), so the Node toolchain is touched ~daily

maxmilian:
## Summary

Closes #1538. Adds a TOML filter for `fnm` (Fast Node Manager) and registers it in the discover/hook engine so `fnm use|install|list|list-remote|...` get auto-rewritten to `rtk fnm ...`.

## Why

`fnm` is a popular Node version manager (~14k stars). Without a handler, `rtk hook check` returns "No rewrite" for `fnm use` / `fnm install` / `fnm list-remote`, so output sits unfiltered in agent sessions (~249 invocations / 30 days per #1538). Pairs naturally with the existing `npm`/`npx`/`pnpm` handlers. The clear win is `fnm list-remote`, which prints ~850 Node versions per call.

## Fix

- `src/filters/fnm.toml` — new filter. `match_command = "^fnm\s+(list-remote |use|install|uninstall|list|ls|current|default)(\s|$)"`. `strip_ansi = true`, `filter_stderr = true` (fnm writes errors to stderr — without merging, an error like `Requested version v99.x.x is not currently installed` would be eaten by terminal stderr while `on_empty` fired "fnm: ok" on the empty stdout, which is misleading). `strip_lines_matching` drops blank / `Downloading ` / `Extracting ` lines. `max_lines = 50` truncates `list-remote` with a built-in recovery hint. `on_empty = "fnm: ok"`. 10 inline tests cover: use success, install success, missing-version error, list, current, install with progress-strip, on_empty fallback, uninstall error, ANSI codes stripped, list-remote truncation with recovery hint.
- `src/discover/rules.rs` — new `RtkRule { rewrite_prefixes: &["fnm"], rtk_cmd: "rtk fnm", category: "NodeVersionManager", savings_pct: 25.0 }`. The TOML `match_command` and the rules.rs `pattern` are character-aligned (both use `(\s|$)` boundary, with `list-remote` placed before `list` in the alternation so the engine picks the longer match) so the two layers can never disagree on which subcommands are handled.
- `src/discover/registry.rs` — register `NodeVersionManager => 40` in `category_avg_tokens` (most fnm output is 1-3 lines — 40 tokens avg is honest; falls through to default 150 otherwise).
- `src/core/toml_filter.rs` — bump builtin filter count assertion 59 → 60 (and the concat test 60 → 61).
- `README.md` — list `rtk fnm use 20` under Package Managers.

## Tests

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — 1946 passed, 0 failed, 0 clippy warnings, fmt clean
- [x] Manual on a real fnm install (1.39.0):
  - `rtk hook check "fnm use 20"` → `rtk fnm use 20` (exit 0)
  - `rtk hook check "fnm install 22"` → `rtk fnm install 22` (exit 0)
  - `rtk hook check "fnm list-remote"` → `rtk fnm list-remote` (exit 0)
  - `rtk hook check "fnm env"` → `No rewrite for: fnm env` (exit 1, intentional — see Scope)
  - `rtk fnm list-remote` → 51 lines incl. `... (796 lines truncated)` recovery hint (was 846 lines raw, ~94% savings on this case)
  - `rtk fnm use 99` → `error: We can't find the necessary environment variables...` (exit 1, error visible through stderr merge)
  - `rtk fnm use 20` → `Using Node v20.20.2` (exit 0, passthrough)
  - `rtk fnm list` → version list intact (exit 0)
- [x] 10 inline TOML tests in `fnm.toml` (run by the builtin test harness)

## Scope

**Out of scope, intentionally:**
- **`fnm env` and `eval "$(fnm env)"`** — the filter would corrupt the shell exports that `eval` consumes, and more fundamentally the hook engine routes on the leading token, so `eval "$(fnm env)"` cannot be intercepted by an `fnm` rewrite prefix at all. Letting both forms pass through raw keeps `eval "$(fnm env)"` working untouched. The issue lists `eval "$(fnm env)"` as a pain point (~123 invocations/30d), but lexer-aware unwrapping of `eval` is a much bigger change and out of scope for this PR.
- **`fnm exec --using=20 -- <child>`** — would otherwise let the filter's blank/`Downloading`/`Extracting` strips eat legitimate output from the wrapped child command. Excluded from both `match_command` and the rewrite pattern.

**Honest about savings:** for the small-output subcommands (`use`, `install`, `current`, `list`), output is already 1–3 lines and there's near-zero savings — they go through the chain mainly so RTK analytics see them as supported rather than "missed opportunity". The genuine savings come from `list-remote` (94% on the local test). The headline `savings_pct = 25` averages those cases conservatively (below the 60% threshold cited in CONTRIBUTING, which is calibrated for noisy command output).

## Follow-up

- If users report meaningful noise from `fnm install` on first-download (uncached, slow network), revisit `strip_lines_matching` to cover additional progress patterns.
- `subcmd_savings` is left empty; a future pass could break out per-subcommand averages (e.g. `list-remote ~94%`, `use ~0%`) once we have real usage data.
- Lexer-aware unwrapping of `eval "$(...)"` would let RTK intercept `eval "$(fnm env)"` style invocations — useful beyond fnm (e.g. `eval "$(direnv hook bash)"`). Tracked separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ rtk pip list                    # Python packages (auto-detect uv)
 rtk pip outdated                # Outdated packages
 rtk bundle install              # Ruby gems (strip Using lines)
 rtk prisma generate             # Schema generation (no ASCII art)
+rtk fnm use 20                  # Node version switch (ANSI/progress stripped)
 ```
 
 ### AWS
--- a/src/core/toml_filter.rs
+++ b/src/core/toml_filter.rs
@@ -1627,8 +1627,8 @@ match_command = "^make\\b"
         let filters = make_filters(BUILTIN_TOML);
         assert_eq!(
             filters.len(),
-            63,
-            "Expected exactly 63 built-in filters, got {}. \
+            64,
+            "Expected exactly 64 built-in filters, got {}. \
              Update this count when adding/removing filters in src/filters/.",
             filters.len()
         );
@@ -1685,11 +1685,11 @@ expected = "output line 1\noutput line 2"
         let combined = format!("{}\n\n{}", BUILTIN_TOML, new_filter);
         let filters = make_filters(&combined);
 
-        // All 63 existing filters still present + 1 new = 64
+        // All 64 existing filters still present + 1 new = 65
         assert_eq!(
             filters.len(),
-            64,
-            "Expected 64 filters after concat (63 built-in + 1 new)"
+            65,
+            "Expected 65 filters after concat (64 built-in + 1 new)"
         );
 
         // New filter is discoverable
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -40,6 +40,7 @@ pub fn category_avg_tokens(category: &str, subcmd: &str) -> usize {
         "GitHub" => 200,
         "GitLab" => 200,
         "PackageManager" => 150,
+        "NodeVersionManager" => 40,
         _ => 150,
     }
 }
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -80,6 +80,15 @@ pub const RULES: &[RtkRule] = &[
         subcmd_status: &[],
     },
     RtkRule {
+        pattern: r"^fnm\s+(list-remote|use|install|uninstall|list|ls|current|default)(\s|$)",
+        rtk_cmd: "rtk fnm",
+        rewrite_prefixes: &["fnm"],
+        category: "NodeVersionManager",
+        savings_pct: 25.0,
+        subcmd_savings: &[],
+        subcmd_status: &[],
+    },
+    RtkRule {
         pattern: r"^(cat|head|tail)\s+",
         rtk_cmd: "rtk read",
         rewrite_prefixes: &["cat", "head", "tail"],
--- a/src/filters/fnm.toml
+++ b/src/filters/fnm.toml
@@ -1,0 +1,169 @@
+[filters.fnm]
+description = "Compact fnm (Fast Node Manager) output — strip download/extract progress, keep version lines and errors. Excludes `fnm env` and `fnm exec` so shell evals and wrapped child commands stay untouched."
+match_command = "^fnm\\s+(list-remote|use|install|uninstall|list|ls|current|default)(\\s|$)"
+strip_ansi = true
+filter_stderr = true
+strip_lines_matching = [
+  "^\\s*$",
+  "^Downloading\\s",
+  "^Extracting\\s",
+]
+truncate_lines_at = 200
+max_lines = 50
+on_empty = "fnm: ok"
+
+[[tests.fnm]]
+name = "use success line passes through"
+input = "Using Node v20.20.2"
+expected = "Using Node v20.20.2"
+
+[[tests.fnm]]
+name = "install success line passes through"
+input = "Installing Node v20.20.2 (arm64)"
+expected = "Installing Node v20.20.2 (arm64)"
+
+[[tests.fnm]]
+name = "missing version error preserved verbatim"
+input = "error: Requested version v99.x.x is not currently installed"
+expected = "error: Requested version v99.x.x is not currently installed"
+
+[[tests.fnm]]
+name = "list output preserved"
+input = "* v20.20.2 default\n* v22.22.3\n* system"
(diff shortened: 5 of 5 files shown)


maxmilian:
Closing this one, with the finding written down rather than lost.

**The gap is still real.** `fnm` remains entirely absent from `develop` (`8a24ce2`) — `git grep -i fnm` matches only the `FnMut` generic in `src/core/stream.rs:560`. So this isn't "no longer needed"; I want to be precise about that rather than close it with a tidier-sounding reason.

**What is wrong is this PR's implementation.** It predates #2748 (`31f9d43`). Since that landed, a `.toml` file alone is enough to drive the hook rewrite: `registry.rs:1052`'s `Classification::Unsupported` branch consults `toml_filter::command_matches_filter` (`registry.rs:1062`) and rewrites TOML-only commands. My `src/filters/fnm.toml` would have been sufficient on its own.

The two extra pieces I added are now redundant, and worse than inert:

- `rules.rs` adds a `RtkRule` with `rtk_cmd: "rtk fnm"` and `savings_pct: 25.0` — a native subcommand this PR never implements
- `registry.rs` adds `"NodeVersionManager" => 40`

A `RULES` entry makes `classify_command` return `Classification::Supported` (`registry.rs:194`), which routes around the TOML-only branch above entirely. So the PR advertises a native `rtk fnm` path that doesn't exist.

**Why close instead of rewriting.** The correct version is small — drop both Rust hunks, keep `fnm.toml` + the count assertions + the README line, roughly 60% less diff. I'm not going to be the one to write it, and I'd rather say so than leave a stale PR sitting open. For anyone picking it up: `src/filters/` has had no new per-tool TOML filter merged since 2026-04-15 (pulumi) and 04-05 (liquibase), and `6c4950e` moved `mvn` the other way, from TOML to Rust — so it's worth checking with a maintainer whether a new per-tool TOML filter is wanted at all before investing in it.

Thanks to whoever triages this; sorry for the noise of an open PR that had gone stale in shape rather than in need.

PR #2114, opened 2026-05-27
feat(go): general go tool support with golangci-lint filtering

### LINKED ISSUE #744: feat(go): Support golangci-lint called via go tool
As of go v1.24, tools such as golanci-lint can be managed via the mod file and called directly on the go binary via the tool subcommand.

As of v0.31.0 rtk does not handle this command, requiring golangci-lint to be installed on the users's $PATH.

```
$ go tool golangci-lint run main.go 
main.go:6:7: printf: fmt.Sprintf format %s has arg 9.0 of wrong type float64 (govet)
	i := fmt.Sprintf("hello %s", 9.0)
	    ^
```

```
$ rtk go tool golangci-lint run main.go
main.go:6:7: printf: fmt.Sprintf format %s has arg 9.0 of wrong type float64 (govet)
	i := fmt.Sprintf("hello %s", 9.0)
	    ^
```

apowis:
Fixes #744

## Summary

Fixes two bugs in the original  implementation (commit 8f985a2) and extends support to all `go tool` invocations.

### Bug 1 — double `run` in command args

`rtk go tool golangci-lint run ./...` was producing:
```
go tool golangci-lint run --output.json.path stdout run ./...
```
because `run_go_tool_golangci_lint` always prepended `run` before appending all args (which already included `run`).

**Fix:** Extract `build_go_tool_golangci_args()` helper that strips a leading `run` before injecting the JSON flag. Both calling forms now work:
- `rtk go tool golangci-lint run ./...` (explicit run)
- `rtk go tool golangci-lint ./...` (no run prefix)

### Bug 2 — rewrite rule and runtime dispatch too narrow

The original implementation only handled `go tool golangci-lint` — any other managed tool (`go tool staticcheck`, `go tool pprof`, etc.) was never rewritten by the hook and never tracked.

**Fix:**
- Drop the `GoTool` enum; `match_go_tool()` now returns the tool name as a `String` and matches any `go tool <name>`
- Unknown tools are routed through a new `run_go_tool_passthrough()` that executes transparently and records usage in the tracking DB (visible in `rtk gain --history`)
- Rewrite pattern broadened from `^go\s+(test|build|vet)` to `^go\s+(test|build|vet|tool\s+\S+)`

## Behaviour after this change

| Command | Before | After |
|---|---|---|
| `go tool golangci-lint run ./...` | unfiltered passthrough | filtered + tracked |
| `go tool golangci-lint ./...` | unfiltered passthrough | filtered + tracked |
| `go tool staticcheck ./...` | not rewritten, untracked | transparent passthrough + tracked |
| `go tool pprof cpu.prof` | not rewritten, untracked | transparent passthrough + tracked |
| Any future `go tool <x>` | ignored | transparent passthrough + tracked |

## Test plan

- 7 new unit tests for `match_go_tool` (now accepts any tool name) and `build_go_tool_golangci_args` (no double-run)
- 5 new registry tests covering classify + rewrite for golangci-lint and arbitrary tools
- Verified end-to-end against a real Go 1.26 project with golangci-lint v2.10.0 registered as a `go tool`


apowis:
@aeppling - let me know if I have done everything correctly for contributing here? I did a previous PR on golangci-lint here but I remembered to use the correct base of `rtk-ai:develop`

PR #2126, opened 2026-05-27
fix: parse leading grep search flags

### LINKED ISSUE #2120: rtk grep falls back to native grep when --fixed-strings is placed before the pattern/path
I’m on rtk 0.42.0 and hit a confusing behavior with rtk grep on Windows/Git Bash.
Repro
rtk grep -n --fixed-strings '<select id="selectPage"' "ruoyi-vue-pro/yudao-module-wallet/src/main/resources/mapper/enterpriseaccount"
Actual
/usr/bin/grep: ruoyi-vue-pro/yudao-module-wallet/src/main/resources/mapper/enterpriseaccount: Is a directory
Expected
I expected rtk grep to either:
1. pass --fixed-strings through to ripgrep, or
2. give a clear RTK-level error explaining the required argument order
instead of falling back to native grep and failing with a directory error.
Notes
If I move the extra flag after the pattern/path, it works:
rtk grep '<select id="selectPage"' "ruoyi-vue-pro/yudao-module-wallet/src/main/resources/mapper/enterpriseaccount" --fixed-strings
So this looks like an argument parsing / routing issue rather than a grep engine limitation.
I also noticed related behavior in existing issues around grep flags and passthrough:
- -E / native grep-style flags
- -n compatibility
- extra_args handling for rtk grep
It seems rtk grep only accepts extra ripgrep args after the positional pattern/path, but that’s easy to miss because people naturally write flags first, like normal grep/rg.
Why this matters
This makes a common search command fail in a surprising way, and the fallback error points to native grep, which hides the real issue.
Suggested fix
Either:
- allow common ripgrep flags like --fixed-strings before the pattern/path for rtk grep, or
- emit a clear parse error telling users to place extra args after the pattern/path, instead of falling back to native grep

iccccccccccccc:
Fixes #2120

## Changes
- let `rtk grep` accept search flags before the pattern/path, including `--fixed-strings`
- normalize leading search flags before invoking `rg`, so the flag is passed through instead of being treated as the pattern
- keep the existing trailing-extra-args form working

## Verification
- `cargo fmt --check`
- `cargo test grep -- --nocapture`
- `cargo test grep_clap_accepts -- --nocapture`
- `cargo run --quiet -- grep -n --fixed-strings 'fn main' src`
- `cargo run --quiet -- grep -A 1 'fn main' src`
- `cargo test` with `C:\Program Files\Git\usr\bin` on `PATH` for the Unix-style commands used by the stream tests

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -37,33 +37,25 @@ pub fn run(
     // -H: always emit the filename.
     // -0: NUL-separate filename. Allows the parser to disambiguate filenames or
     // content containing `:digits:` patterns (issue #1436).
-    rg_cmd.args([
-        "-nH0",
-        "--no-heading",
-        "--no-ignore-vcs",
-        &rg_pattern,
-        &args.path,
-    ]);
+    rg_cmd.args(["-nH0", "--no-heading", "--no-ignore-vcs"]);
 
     if let Some(ft) = file_type {
         rg_cmd.arg("--type").arg(ft);
     }
 
-    for arg in &args.extra_args {
-        // Fix: skip grep-ism -r flag (rg is recursive by default; rg -r means --replace)
-        if arg == "-r" || arg == "--recursive" {
-            continue;
-        }
+    for arg in normalize_rg_extra_args(&args.extra_args) {
         rg_cmd.arg(arg);
     }
+    rg_cmd.args([&rg_pattern, &args.path]);
 
     let result = exec_capture(&mut rg_cmd)
         .or_else(|_| {
             let mut grep_cmd = resolved_command("grep");
             // When we fall back to grep, include all args, not just -rnHZ.
             grep_cmd
-                .args(["-rnHZ", &args.pattern, &args.path])
-                .args(&args.extra_args);
+                .arg("-rnHZ")
+                .args(&args.extra_args)
+                .args([&args.pattern, &args.path]);
             exec_capture(&mut grep_cmd)
         })
         .context("grep/rg failed")?;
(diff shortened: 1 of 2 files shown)


hollamad (commented):
Tested locally on Linux against the bug's reproducer set. The PR's two tests pass (clap accepts the args), but the resulting **search behavior is broken** — the leading flags get consumed as the search pattern rather than passed through to ripgrep.

Build:
- branch: `iccccccccccccc:fix/grep-leading-flags` @ HEAD
- `cargo test --release --bin rtk grep_clap` -> 2 passed
- `cargo build --release` -> clean

### What works (parse + execute correctly)

| Command | Result |
|---|---|
| `rtk grep -i pat <file>` | matches `[tracking]` ✓ |
| `rtk grep -A 2 pat <file>` | 3 matches with context ✓ |
| `rtk grep -c pat <file>` | bare count ✓ |

### What breaks (parse OK, execution wrong)

| Command | Expected | Actual |
|---|---|---|
| `rtk grep -rn 'tracking' <dir>/` | match line `[tracking]` | match line `[n]` (i.e. pattern became `-rn`) |
| `rtk grep -rnE 'tracking\|tee' <dir>/` | match `[tracking]`, `[tee]` | match `[nE]` (pattern became `-rnE`) |
| `rtk grep -E 'tracking\|tee' <file>` | 2 matches | `rg: missing value for flag -E`, 0 matches |

Root cause looks like `#[arg(allow_hyphen_values = true)]` on `pattern: String` accepts the flag-as-string but the downstream rg invocation then receives `pattern="-rn"` and tries to literal-match `[n]` from inside `[tracking]`, instead of `-r` `-n` being applied as rg flags.

Suggestion: rather than relaxing clap, detect leading hyphenated args before clap parsing and shift them into the existing `EXTRA_ARGS` Vec, then pass them to rg as flags. This keeps rtk's strict typing for its own flags while letting standard rg/grep flags reach ripgrep transparently.

Happy to test a follow-up commit if you push one.

iccccccccccccc:
﻿Thanks, good catch. I pushed a follow-up in `e2de3e2`.

The issue was the grep/rg flag mismatch you pointed at: clustered grep flags like `-rn` still let `rg` see `-r`, and `rg -r` means replacement rather than recursive search. The follow-up strips grep-only `-r` / `-n` / `-E` before invoking `rg` (including clusters like `-rn`, `-rnE`, and `-rni`) and passes the remaining rg args before the pattern/path.

Checked locally:
- `cargo test --release --bin rtk grep_clap -- --nocapture`
- `cargo test --release --bin rtk grep_cmd -- --nocapture`
- manual repros for `rtk grep -rn tracking`, `rtk grep -rnE 'tracking\|tee'`, and `rtk grep -E 'tracking\|tee'`

hollamad:
Re-tested locally at `e2de3e2`. The three failing cases from my last comment all
work now:

| Command | Result |
|---|---|
| `rtk grep -rn 'tracking' <dir>/` | matches `[tracking]` (2 files) |
| `rtk grep -rnE 'tracking\|tee' <dir>/` | matches `[tracking]` + `[tee]` (4 hits) |
| `rtk grep -E 'tracking\|tee' <file>` | 2 matches, no `missing value for flag` error |

Originally-working cases (`-i`, `-A`, `-c`) still pass, no regression. Cluster
declustering preserves non-stripped flags as expected (`-rni` keeps `-i`).
Longform `--recursive` works.

`cargo test --release --all` and `cargo clippy --release --all-targets` clean on
this branch (3 pre-existing `curl_cmd::tests` failures also reproduce on develop,
unrelated). The two new `grep_clap` tests and the new normalize tests pass.

Thanks for the quick turnaround.

KuSh:
Thanks for the fix and the clear reproduction. Flag ordering for rtk grep was solved on develop by #2333 (merged 2026-06-14), which replaced the pattern and path positionals with a trailing argument vector, and #2641 (merged 2026-06-26), which turned rtk grep into a faithful grep wrapper with rtk rg for the ripgrep path. On the current build, rtk grep -rn --fixed-strings 'fn main' src and the flag-last form give identical correct output, and rtk grep -A 1 works too. One detail to flag: the -l short for --max-len that this branch keeps was later removed on purpose in #2628 so native grep's -l reaches the engine. Closing as already covered, but please comment or reopen if you still hit an ordering case that breaks.

PR #2131, opened 2026-05-28
fix(grep): translate GNU grep --include/--exclude to rg --glob

goodmood-creator:
## Problem

`rtk grep` forwards `extra_args` directly to ripgrep, but ripgrep has no
`--include`/`--exclude` (it uses `--glob`/`-g`). Running the common GNU grep
form breaks the search instead of filtering:

```
rtk grep "pattern" . --include="*.kt"   # rg rejects --include → search breaks
```

This also bites the Claude Code / agent workflow, where a shell `grep --include`
gets rewritten to `rtk grep` and the `--include` is effectively swallowed.

## Fix

Translate GNU grep include/exclude flags to rg globs inside the `extra_args`
loop in `grep_cmd.rs`:

- `--include=GLOB` / `--include GLOB` → `--glob GLOB`
- `--exclude=GLOB` / `--exclude GLOB` → `--glob !GLOB`
- every other flag is passed through unchanged
- `--include-dir`/`--exclude-dir` are out of scope (rg has no directory-only glob)

## Tests

Added unit tests for the `=`/space forms, the exclude `!` prefix, and
pass-through of unrelated flags (incl. `--include-dir` not being misparsed).

`cargo fmt --all --check && cargo clippy --all-targets && cargo test` pass locally.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -42,9 +42,17 @@ pub fn run(
         rg_cmd.arg("--type").arg(ft);
     }
 
-    for arg in extra_args {
+    let mut iter = extra_args.iter();
+    while let Some(arg) = iter.next() {
         // Fix: skip grep-ism -r flag (rg is recursive by default; rg -r means --replace)
         if arg == "-r" || arg == "--recursive" {
+            continue;
+        }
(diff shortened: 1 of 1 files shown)


KuSh:
Hi @goodmood-creator, thanks for your work.

This has been superseded by https://github.com/rtk-ai/rtk/pull/2607; please feel free to comment or review that PR to help it move forward.

PR #2138, opened 2026-05-28
fix(grep): translate BRE and preserve rg dialects through rtk grep

### LINKED ISSUE #1436: The rtk grep is not grep-compatible

F0rty-Tw0:
This PR translates grep's BRE into rg's dialect so rtk could keep running rg underneath and still match what grep matches.

That's not the direction the project took. #2641 (merged) fixes #1436 by running the engine the user actually typed: grep runs grep, rg runs rg, no substitution, no translation. It deletes the exact machinery this PR adds and renames `grep_cmd.rs` to `search.rs`, so a rebase would only re-add the code that PR removed.

Closing as superseded.

On #1436 itself:
- The wrong-dialect matches (the original complaint) are gone with #2641, since grep now runs native grep.
- The single-file mangling (the `2F` label and the broken line 81) is a separate parsing bug. #1554 (merged) was aimed at that. I have not re-run the original repro against develop, so I can't confirm #1436 is fully closed.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -1,6 +1,6 @@
 //! Filters grep output by grouping matches by file.
 
-use crate::core::stream::exec_capture;
+use crate::core::stream::{CaptureResult, exec_capture};
 use crate::core::tracking;
 use crate::core::utils::resolved_command;
 use crate::core::{args_utils, config};
@@ -11,8 +11,9 @@ use std::collections::HashMap;
 /// Short single-char flags that consume one following token (or inline remainder)
 /// as their value. `-e` is handled separately — its value goes to `patterns`.
 /// Includes all rg short flags that take a value argument except `-e` and `-r`
-/// (stripped) and `-E` (dialect, left to #2138). Failure mode for a missing
-/// entry: the value becomes a positional (visible wrong result, not silent).
+/// (stripped) and `-E` (a dialect flag handled by `detect_syntax`/routing, not
+/// consumed here). Failure mode for a missing entry: the value becomes a
+/// positional (visible wrong result, not silent).
 const VALUE_FLAGS_SHORT: &[u8] = b"ABCMTdfgjmt";
 
 /// Long flags that consume the NEXT token as their value (space-separated form).
@@ -85,7 +86,7 @@ fn parse_cluster(rest: &str) -> ClusterResult {
         let ch = bytes[j];
         let is_e = ch == b'e';
         if is_e || VALUE_FLAGS_SHORT.contains(&ch) {
-            let prefix = strip_r(&raw_prefix);
+            let prefix = strip_short_args(&raw_prefix);
             // Inline value = bytes after this char; returned verbatim (no stripping).
             let inline = std::str::from_utf8(&bytes[j + 1..])
                 .unwrap_or("")
@@ -99,16 +100,16 @@ fn parse_cluster(rest: &str) -> ClusterResult {
         raw_prefix.push(ch as char);
         j += 1;
     }
-    ClusterResult::Boolean(strip_r(&raw_prefix))
+    ClusterResult::Boolean(strip_short_args(&raw_prefix))
 }
 
 /// Strip `r`/`R` from a string of flag letters.
 /// Returns `None` when nothing remains after stripping.
 ///
 /// Only called on accumulated flag letters (never on inline values).
-/// `strip_r("carrot")` → `Some("caot")` — this shows exactly why it must not
-/// touch value bytes; that corruption was the original `-ecarrot` bug.
-fn strip_r(flag_letters: &str) -> Option<String> {
+/// `strip_short_args("carrot")` → `Some("caot")` — this shows exactly why it
+/// must not touch value bytes; that corruption was the original `-ecarrot` bug.
+fn strip_short_args(flag_letters: &str) -> Option<String> {
     let s: String = flag_letters
         .chars()
         .filter(|&c| c != 'r' && c != 'R')
@@ -121,7 +122,7 @@ fn strip_r(flag_letters: &str) -> Option<String> {
 }
 
 /// Drop `--recursive` (grep-ism); pass all other long flags through unchanged.
-fn strip_recursive(arg: &str) -> Option<String> {
+fn strip_long_args(arg: &str) -> Option<String> {
     match arg {
         "--recursive" => None,
         _ => Some(arg.to_string()),
@@ -183,7 +184,7 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
                 continue;
             }
             // Drop --recursive; pass everything else through.
-            if let Some(cleaned) = strip_recursive(arg) {
+            if let Some(cleaned) = strip_long_args(arg) {
                 flags.push(cleaned);
             }
             i += 1;
@@ -251,6 +252,156 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
     (patterns, paths, flags)
 }
 
+/// Regex dialect selected by a grep dialect flag. [`detect_syntax`] returns
+/// `None` for grep's POSIX default (BRE), so the default needs no variant.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum RegexSyntax {
+    /// Extended Regular Expression (`-E`/`--extended-regexp`).
+    Ere,
+    /// Explicit Basic Regular Expression (`-G`/`--basic-regexp`).
+    Basic,
+    /// Perl-Compatible Regular Expression (`-P`/`--perl-regexp`).
+    Pcre,
+    /// Fixed string, no regex (`-F`/`--fixed-strings`).
+    Fixed,
+}
+
+/// Returns the regex dialect requested via a grep dialect flag, or `None` for
+/// the POSIX default (BRE — what plain `grep` uses with no dialect flag).
+///
+/// Matches GNU grep semantics: the LAST dialect flag wins (`-E -F` → Fixed), and
+/// a dialect char bundled in a short cluster counts too (`-iE` → Ere). Scanning
+/// stops at `--` (everything after is an operand) and at `=` within a token (so
+/// a `--color=…` value can't false-trigger).
+fn detect_syntax(extra_args: &[String]) -> Option<RegexSyntax> {
+    let mut syntax = None;
+    for arg in extra_args {
+        if arg == "--" {
+            break;
+        }
+        match arg.as_str() {
+            "-E" | "--extended-regexp" => syntax = Some(RegexSyntax::Ere),
+            "-G" | "--basic-regexp" => syntax = Some(RegexSyntax::Basic),
+            "-P" | "--perl-regexp" => syntax = Some(RegexSyntax::Pcre),
+            "-F" | "--fixed-strings" => syntax = Some(RegexSyntax::Fixed),
+            other => {
+                // Short flag or bundled cluster (`-iE`): a single leading dash,
+                // scan its chars until `=`. Last dialect char wins.
+                if let Some(rest) = other.strip_prefix('-') {
+                    if rest.starts_with('-') {
+                        continue;
+                    }
+                    for ch in rest.chars() {
+                        match ch {
+                            '=' => break,
+                            'E' => syntax = Some(RegexSyntax::Ere),
+                            'G' => syntax = Some(RegexSyntax::Basic),
+                            'P' => syntax = Some(RegexSyntax::Pcre),
+                            'F' => syntax = Some(RegexSyntax::Fixed),
+                            _ => {}
+                        }
+                    }
+                }
+            }
+        }
+    }
+    syntax
+}
+
+/// True if the caller passed an rg engine-selection flag (`--engine[=...]` or
+/// `--auto-hybrid-regex`). These pick rg's own regex engine, so the pattern is
+/// rg-native: it must run on rg (never grep, which can't accept the flag) and
+/// must NOT be BRE-translated. This is also the marker the rg-origin hook
+/// rewrite uses (`--engine=auto`) to keep `rg` patterns off grep's BRE path.
+fn has_rg_engine_flag<T: AsRef<str>>(extra_args: &[T]) -> bool {
+    extra_args.iter().any(|a| {
+        let a = a.as_ref();
+        a == "--engine" || a.starts_with("--engine=") || a == "--auto-hybrid-regex"
+    })
+}
+
+/// Translates a POSIX BRE pattern (grep's default) into an equivalent regex for
+/// ripgrep.
+///
+/// In BRE the metacharacters are the *escaped* forms (`\(`, `\)`, `\+`, `\?`,
+/// `\{`, `\}`, `\|`) and the bare characters are literals; rg uses the inverse
+/// convention. A leading bare `*` is also a literal in BRE. `.` and a non-leading
+/// `*` keep their meaning, so `\.`/`\*` and `.`/`*` pass through unchanged.
+///
+/// This fixes issue #1436: `delete()` is a literal match in BRE, but feeding it
+/// straight to rg treats `()` as an empty group and matches every `delete`.
+fn bre_to_pcre(bre: &str) -> String {
+    let mut out = String::with_capacity(bre.len() + 8);
+    let mut chars = bre.chars();
+    let mut at_start = true;
+    // Inside a `[...]` bracket expression every metacharacter is already literal
+    // in both dialects, so the contents pass through untranslated. (The rare
+    // POSIX `[]...]`/`[^]...]` form, where a leading `]` is literal, is not
+    // special-cased — such a pattern passes through exactly as written.)
+    let mut in_bracket = false;
+    while let Some(ch) = chars.next() {
+        if in_bracket {
+            out.push(ch);
+            if ch == ']' {
+                in_bracket = false;
+            }
+            at_start = false;
+            continue;
+        }
+        match ch {
+            // Opening a bracket expression: copy verbatim until it closes.
+            '[' => {
+                in_bracket = true;
+                out.push('[');
+            }
+            '\\' => match chars.next() {
+                // Escaped BRE metacharacter -> unescape so rg treats it as one.
+                Some(c @ ('+' | '?' | '|' | '(' | ')' | '{' | '}')) => out.push(c),
+                // Escaped backslash -> a literal backslash in both dialects.
+                Some('\\') => out.push_str(r"\\"),
+                // Any other escape (`\.`, `\*`, `\d`) -> keep verbatim.
+                Some(c) => {
+                    out.push('\\');
+                    out.push(c);
+                }
+                // Trailing lone backslash -> keep as-is.
+                None => out.push('\\'),
+            },
+            // A leading `*` has nothing to quantify in BRE, so it is literal.
+            '*' if at_start => out.push_str(r"\*"),
+            // Bare BRE metacharacters are literals -> escape them for rg.
+            '+' | '?' | '|' | '(' | ')' | '{' | '}' => {
+                out.push('\\');
+                out.push(ch);
+            }
+            _ => out.push(ch),
+        }
+        at_start = false;
+    }
+    out
+}
+
+/// Runs native grep with the original (untranslated) patterns and grep-compatible
+/// flags, which grep interprets in its own dialect. Used when the caller named
+/// any grep dialect (`-E`/`-G`/`-P`/`-F`), and as the fallback when rg is
+/// unavailable.
+fn native_grep(
+    patterns: &[String],
+    paths: &[String],
+    extra_args: &[String],
+) -> Result<CaptureResult> {
+    let mut grep_cmd = resolved_command("grep");
+    grep_cmd.args(extra_args);
+    for p in patterns {
+        grep_cmd.args(["-e", p]);
+    }
+    // --null (not -Z): on BSD/macOS grep -Z means --decompress, not the NUL
+    // filename separator parse_match_line() needs (issue #2310).
+    grep_cmd.args(["-rnH", "--null", "--"]);
+    grep_cmd.args(paths);
+    exec_capture(&mut grep_cmd)
+}
+
 pub fn run(
     max_line_len: usize,
     max_results: usize,
@@ -312,50 +463,57 @@ pub fn run(
         eprintln!("grep: '{}' in {}", pattern_display, path_display);
     }
 
-    let mut rg_cmd = resolved_command("rg");
-    // --no-ignore-vcs: match grep -r behavior (don't skip .gitignore'd files).
-    // Without this, rg returns 0 matches for files in .gitignore, causing
-    // false negatives that make AI agents draw wrong conclusions.
-    // Using --no-ignore-vcs (not --no-ignore) so .ignore/.rgignore are still respected.
-    // -H: always emit the filename.
-    // -0: NUL-separate filename. Allows the parser to disambiguate filenames or
-    // content containing `:digits:` patterns (issue #1436).
-    rg_cmd.args(["-nH0", "--no-heading", "--no-ignore-vcs"]);
+    // Routing (per regex dialect):
+    // - An explicit grep dialect flag (-E/-G/-P/-F) names a grep engine → run
+    //   native grep directly; it applies the dialect from extra_args. No rg, no
+    //   translation.
+    // - An rg engine flag (--engine/--auto-hybrid-regex, incl. the rg-origin
+    //   hook marker --engine=auto) forces rg even alongside a dialect flag, since
+    //   grep can't accept it; that pattern is rg-native, so it is not translated.
+    // - The plain default (no dialect) goes through rg with BRE→rg translation
+    //   (issue #1436).
+    let syntax = detect_syntax(&extra_args);
+    let rg_engine = has_rg_engine_flag(&extra_args);
 
-    if let Some(ft) = file_type {
-        rg_cmd.arg("--type").arg(ft);
-    }
+    let result = if syntax.is_some() && !rg_engine {
+        native_grep(&patterns, &paths, &extra_args).context("grep failed")?
+    } else {
+        let mut rg_cmd = resolved_command("rg");
+        // --no-ignore-vcs: match grep -r behavior (don't skip .gitignore'd files)
+        // so AI agents don't draw wrong conclusions from missing matches.
+        // -H: always emit the filename. -0: NUL-separate the filename so the
+        // parser can disambiguate `:digits:` in paths/content (issue #1436).
+        rg_cmd.args(["-nH0", "--no-heading", "--no-ignore-vcs"]);
 
-    // extra_args is already stripped of -r/-R/-recursive by extract_pattern_path
-    rg_cmd.args(&extra_args);
+        if let Some(ft) = file_type {
+            rg_cmd.arg("--type").arg(ft);
+        }
 
-    // All patterns as -e flags (BRE \| → | translation for rg's PCRE engine).
-    // Using -e keeps `--` semantically as a flag/path separator, not part of the pattern.
-    for p in &patterns {
-        rg_cmd.args(["-e", &p.replace(r"\|", "|")]);
-    }
+        // extract_pattern_path already dropped -r/--recursive during the split,
+        // and no grep dialect flag reaches here (those route to native grep
+        // above), so the caller's remaining flags forward to rg unchanged.
+        rg_cmd.args(&extra_args);
 
-    // `--` after all flags: prevents rg from interpreting path args starting
-    // with `-` as its own flags.
-    rg_cmd.arg("--");
-    rg_cmd.args(&paths);
+        // Translate only the plain BRE default. An rg engine flag means the
+        // pattern is already rg-native, so pass it through verbatim.
+        let translate = !rg_engine;
+        for p in &patterns {
+            let pattern = if translate { bre_to_pcre(p) } else { p.clone() };
+            rg_cmd.args(["-e", &pattern]);
+        }
 
-    let result = exec_capture(&mut rg_cmd)
-        .or_else(|_| {
-            // rg unavailable: fall back to system grep with the original,
-            // untranslated patterns (grep interprets BRE natively).
-            let mut grep_cmd = resolved_command("grep");
-            grep_cmd.args(&extra_args);
-            for p in &patterns {
-                grep_cmd.args(["-e", p]);
-            }
-            // --null (not -Z): on BSD/macOS grep -Z means --decompress, not the
-            // NUL filename separator parse_match_line() needs (issue #2310).
-            grep_cmd.args(["-rnH", "--null", "--"]);
-            grep_cmd.args(&paths);
-            exec_capture(&mut grep_cmd)
-        })
-        .context("grep/rg failed")?;
+        // `--` after all flags: stop rg reading path args as its own flags.
+        rg_cmd.arg("--");
+        rg_cmd.args(&paths);
+
+        exec_capture(&mut rg_cmd)
+            .or_else(|_| {
+                // rg unavailable: fall back to native grep with the original,
+                // untranslated patterns (grep interprets each dialect natively).
+                native_grep(&patterns, &paths, &extra_args)
+            })
+            .context("grep/rg failed")?
+    };
 
     // Passthrough output flags that produce output that is already small.
     if has_format_flag(&extra_args) {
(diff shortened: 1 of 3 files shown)


src/cmds/system/grep_cmd.rs
```diff
@@ -8,6 +8,341 @@ use anyhow::{Context, Result};
 use regex::Regex;
 use std::collections::HashMap;
 
+/// Regex dialect requested by the user (matches grep's regex-mode flags).
+///
+/// `BRE` is the POSIX default when no dialect flag is given, matching plain
+/// `grep` behavior.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum GrepMode {
```
KuSh: suggestion:
```suggestion
enum RegexSyntax {
```

src/cmds/system/grep_cmd.rs
```diff
@@ -8,6 +8,341 @@ use anyhow::{Context, Result};
...
+    /// Basic Regular Expression (`-G`/`--basic-regexp` or no flag — grep default).
+    Bre,
+    /// Extended Regular Expression (`-E`/`--extended-regexp`).
+    Ere,
+    /// Perl-Compatible Regular Expression (`-P`/`--perl-regexp`).
+    Pcre,
+    /// Fixed string, no regex (`-F`/`--fixed-strings`).
+    Fixed,
```
KuSh, line 266: suggestion, either allow `None` here or return `Option<RegexSyntax>` in the detection function (preferred)

src/cmds/system/grep_cmd.rs
```diff
@@ -25,39 +360,67 @@ pub fn run(
...
+    //   * BRE (default or -G): delegate to real `grep` for exact POSIX
+    //     semantics. Slower than rg, but eliminates the regex-translation
+    //     surface area (literal `()` vs grouping `\(\)`, `\|` alternation,
+    //     `\+`/`\?`/`\{` interval, etc.). On Windows without grep on PATH,
+    //     fall back to rg with a best-effort BRE→rg-regex translation.
+    //   * ERE (-E): rg's regex dialect handles ERE-style alternation and
+    //     grouping natively. Pass through, fast.
+    //   * PCRE (-P): rg --pcre2. Exact PCRE semantics, fast.
+    //   * Fixed (-F): rg -F. No regex parsing at all, fast.
+    let mode = detect_mode(extra_args);
```
KuSh: I think rewrite should be involved so a `rg xxx` rewrite ends with `rtk grep -P` command so we won't default to BRE in that case

src/cmds/system/grep_cmd.rs
```diff
@@ -8,6 +8,341 @@ use anyhow::{Context, Result};
...
+            "-P" | "--perl-regexp" => mode = GrepMode::Pcre,
+            "-G" | "--basic-regexp" => mode = GrepMode::Bre,
+            other => {
+                // Bundled short cluster: single leading dash, letters only,
+                // no `=`. Scan its chars; last dialect char in the cluster wins.
+                let bytes = other.as_bytes();
+                let is_cluster = bytes.len() >= 2
+                    && bytes[0] == b'-'
+                    && bytes[1] != b'-'
+                    && bytes[1..].iter().all(|b| b.is_ascii_alphabetic());
```
KuSh: ripgrep has `-0` which is outside of `is_ascii_alphabetic`

KuSh: Perhaps treating other if starting with `-` and not followed by another `-` and stop if you encounter a `=` would work?

src/cmds/system/grep_cmd.rs
```diff
@@ -8,6 +8,341 @@ use anyhow::{Context, Result};
...
+///     `\)`, etc. are literals.
+///
+/// `\|` (BRE alternation) is also non-standard; rg uses `|`.
+///
+/// This is a best-effort translation used only as a Windows fallback when no
+/// real `grep` binary is on PATH. It does not handle every POSIX BRE corner
+/// case (e.g. literal `*` at the start of a pattern, character class
+/// equivalence semantics).
+fn bre_to_rg(pattern: &str) -> String {
+    // Backslash-aware single-pass tokenizer. We must distinguish:
```
KuSh: This seems overly complicated, here is a simplified version that pass all your tests:

```rs
    fn bre_to_pcre(bre: &str) -> String {
        let mut result = String::with_capacity(bre.len() + 8);
        let mut chars = bre.chars().peekable();
        let mut at_start = true;

        while let Some(ch) = chars.next() {
            match ch {
                '\\' if let Some(&next) = chars.peek() => match next {
                    // Escaped metacharacter in BRE -> unescape for PCRE
                    '+' | '?' | '|' | '(' | ')' | '{' | '}' => {
                        result.push(next);
                        chars.next();
                    }
                    // Escaped backslash -> keep as-is
                    '\\' => {
                        result.push_str(r"\\");
                        chars.next();
                    }
                    // Other escapes -> keep as-is
                    _ => result.push(ch),
                },

                // Leading * is literal in BRE, escape it for PCRE
                '*' if at_start => {
                    result.push_str(r"\*");
                }

                // Unescaped metacharacters in BRE are literal, escape for PCRE
                '+' | '?' | '|' | '(' | ')' | '{' | '}' => {
                    result.push('\\');
                    result.push(ch);
                }

                _ => {
                    result.push(ch);
                }
            }
            at_start = false;
        }

        result
    }
```

src/cmds/system/grep_cmd.rs
```diff
@@ -8,6 +8,341 @@ use anyhow::{Context, Result};
...
+/// If `arg` is a bundled short cluster (`^-[A-Za-z]+$`), returns a copy with the
+/// dialect chars (`E`/`G`/`P`/`F`) removed, preserving the other short flags
+/// (e.g. `-iE` → `-i`, `-EF` → `-`). Returns `None` for non-cluster tokens so
+/// the caller forwards them unchanged.
+fn strip_dialect_from_cluster(arg: &str) -> Option<String> {
+    let bytes = arg.as_bytes();
+    let is_cluster = bytes.len() >= 2
+        && bytes[0] == b'-'
+        && bytes[1] != b'-'
+        && bytes[1..].iter().all(|b| b.is_ascii_alphabetic());
```
KuSh: Same remark, could be factorized

src/cmds/system/grep_cmd.rs
```diff
@@ -8,6 +8,341 @@ use anyhow::{Context, Result};
...
+        .collect();
+    Some(format!("-{}", kept))
+}
+
+/// Maps an rg-style file type name to a grep `--include` glob.
+///
+/// grep has no `--type` flag, so the type filter must be expressed as a glob.
+/// Known common types are mapped explicitly; anything else falls back to
+/// `*.<type>` (so `rtk grep -t foo` searches `*.foo`).
+fn grep_include_glob(file_type: &str) -> String {
```
KuSh: That seems too complicated again, the rtk user is calling either grep or rg, if is calling it with the right args.

If calling rg, it should be because it knows rg is installed or the user has told it to, in either cases, if it isn't available it doesn't seem bad to fail.

That rtk internally uses rg over grep if available isn't a bad thing and can be handled more simply. The other way around, i.e. trying to overcome system/user bad configuration, isn't what rtk is aiming for

KuSh (changes requested):
Thanks for your PR.

I didn't go through all of it but it seems way too complicated and trying to solve non-existent problems (at least from a rtk's point of view).

Can you come back with a second simpler proposal, taking into account what I have already commented on? It seems to be a good addition, just far too complex for now.

F0rty-Tw0:
@KuSh Thanks for wading through all of it, and fair call.

I had an AI assistant help me review this, and it did the thing they're best at, it found a problem that wasn't really there and then heroically solved it five separate ways. It saw one escaped `\|` in the issue and concluded the only sane fix was to reimplement POSIX regex semantics from scratch, complete with Windows-without-`grep` fallbacks for a bug that happened on Linux. I should have pushed back instead of nodding along. That part's on me.

Ill cut it down to:

- Drop the `grep`-binary fallback. If someone runs `rg` and it isn't there, let it fail. rtk shouldn't be in the business of papering over a broken setup.
- Swap `bre_to_rg` for your `bre_to_pcre`. Cleaner, and it passes the same tests, so thank you.
- Remove `reorder_grep_args`, the cluster stripping, and the file-type glob mapping.
- Rename `GrepMode` to `RegexSyntax` and have detection return `Option<RegexSyntax>`.
- ~~Handle the `rg … → rtk grep -P` case at the rewrite layer instead, separately from this.~~

_Small course-correction on the rewrite-layer point. I'd planned to do `rg` → `rtk grep -P` as a separate PR. But once I tested it, leaving it out means this PR fixes `grep` and silently breaks `rg`: the hook rewrites both to `rtk grep`, so an rg pattern gets BRE-translated and mangled. A known regression behind a "next PR" didn't sit right, so I folded the one-line tag in. `grep` keeps BRE, `rg` gets `-P`. Happy to split it back out if you'd rather review them apart._

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Done =)

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Done, returns options now

src/cmds/system/grep_cmd.rs
F0rty-Tw0: - Fixed the -0, no longer matches
- Done exactly this

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Adopting yours, big thanks

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Removed entirely

F0rty-Tw0:
@KuSh Two deviations from your suggestions

- **Kept the rg→grep fallback.** Dropping it made `rtk grep` return empty in CI without ripgrep, and the hook rewrites plain `grep` too. Only the 4-line fallback remains; the heavy grep machinery is gone.
- **`-E` not `-P` for the rg tag.** Bare `rg` uses its default engine, not PCRE2, so `-E` is faithful where `-P` would switch engines. Verified against real `rg` and `grep`.

src/discover/registry.rs
```diff
@@ -840,7 +840,17 @@ fn rewrite_segment_inner(
...
+            // is in rg's native regex dialect, so tag it with -E: that routes it
+            // to rg's default engine unchanged, instead of letting rtk translate
+            // it as grep BRE (which would mangle `()`/`+`/`?`/`|`, #1436
+            // follow-up). The flag trails because the pattern is rtk grep's first
+            // positional. Flagged rg (`rg -P 'x'`) already carries its own
+            // dialect, so it is left to the normal path; `grep` keeps BRE.
+            let tag_rg_native = prefix == "rg" && !rest.is_empty() && !rest.starts_with('-');
+            let rewritten = if tag_rg_native {
+                format!("rtk grep {} -E{}", rest, redirect_suffix)
+            } else if rest.is_empty() {
```
KuSh: Is splitting rules in src/discover/rules.rs not enough?
```rs
    RtkRule {
        pattern: r"^rg\s+",
        rtk_cmd: "rtk grep -E",
        rewrite_prefixes: &["rg", "grep"],
        category: "Files",
        savings_pct: 75.0,
        subcmd_savings: &[],
        subcmd_status: &[],
    },
     RtkRule {
        pattern: r"^grep\s+",
        rtk_cmd: "rtk grep",
        rewrite_prefixes: &["rg", "grep"],
        category: "Files",
        savings_pct: 75.0,
        subcmd_savings: &[],
        subcmd_status: &[],
    },
 ```

If another regex engine was passed as argument, `-E` will be overridden as you document it in your comments.

also warning, [rg man](https://manpages.debian.org/testing/ripgrep/rg.1.en.html) document `-e` or `--engine`, `-E` seems to be used for something else

File: src/cmds/system/grep_cmd.rs [hunk 9]
```diff
@@ -25,35 +125,65 @@ pub fn run(
...
-    let rg_pattern = pattern.replace(r"\|", "|");
+    // issue #1436: rtk runs `rg`, whose regex dialect differs from grep's.
+    // grep defaults to POSIX BRE, where `()`, `+`, `?`, `{`, `}` are literals
+    // and `\(`, `\+`, … are the metacharacters; rg uses the inverse. Feeding a
+    // BRE pattern straight to rg silently broadens matches (e.g. `delete()`
+    // matches every `delete`). Translate BRE → rg regex; the explicit dialects
+    // pass through with the matching rg flag.
+    let syntax = detect_syntax(extra_args);
+    let rg_pattern = match syntax {
+        None | Some(RegexSyntax::Basic) => bre_to_pcre(pattern),
```
KuSh: That should be strange that someone specifically requested `-G`/`--basic-regexp` (i.e. it specifically requested grep, not rg) but we force it into using rg.
At least that seems strange to me.
This could also potentially generate more problems, the more specific the passed arguments are, the more chances we could break anything if playing with them

src/cmds/system/grep_cmd.rs
```diff
@@ -25,35 +125,65 @@ pub fn run(
...
+    // so the parser stays unambiguous for content/paths containing `:` (#1436).
+    rg_cmd.args(["-nH0", "--no-heading", "--no-ignore-vcs"]);
+    match syntax {
+        Some(RegexSyntax::Pcre) => {
+            rg_cmd.arg("--pcre2");
+        }
+        Some(RegexSyntax::Fixed) => {
+            rg_cmd.arg("-F");
+        }
+        _ => {}
```
KuSh: `RegexSyntax::Basic` should bail or directly skip to native grep, shouldn't it?

src/cmds/system/grep_cmd.rs
```diff
@@ -25,35 +125,65 @@ pub fn run(
...
+            arg.as_str(),
+            "-r" | "--recursive"
+                | "-E"
+                | "--extended-regexp"
+                | "-G"
+                | "--basic-regexp"
+                | "-P"
+                | "--perl-regexp"
+                | "-F"
+                | "--fixed-strings"
```
KuSh: Does it handle multiple short flags, like `-rE`?

src/cmds/system/grep_cmd.rs
```diff
@@ -306,24 +436,140 @@ mod tests {
...
+        // `delete()`. rg would treat `()` as an empty group matching every
+        // `delete`. Translation must escape the bare parens.
+        assert_eq!(bre_to_pcre("delete()"), r"delete\(\)");
+    }
+
     #[test]
-    fn test_bre_alternation_translated() {
-        let pattern = r"fn foo\|pub.*bar";
-        let rg_pattern = pattern.replace(r"\|", "|");
-        assert_eq!(rg_pattern, "fn foo|pub.*bar");
```
KuSh, line 645: The previous test remains interesting, why remove it?

src/cmds/system/grep_cmd.rs
```diff
@@ -306,24 +436,140 @@ mod tests {
...
-    // Fix: -r flag (grep recursive) is stripped from extra_args (rg is recursive by default)
     #[test]
-    fn test_recursive_flag_stripped() {
-        let extra_args: Vec<String> = vec!["-r".to_string(), "-i".to_string()];
-        let filtered: Vec<&String> = extra_args
-            .iter()
-            .filter(|a| *a != "-r" && *a != "--recursive")
-            .collect();
-        assert_eq!(filtered.len(), 1);
-        assert_eq!(filtered[0], "-i");
```
KuSh: Same remark, should still be tested

KuSh (changes requested):

src/discover/registry.rs
F0rty-Tw0: adopted the rule split, thanks. one snag surfaced while testing it: `rtk grep -E pat` didn't parse at all, clap rejected the leading flag and silently fell back to raw unfiltered grep. so the rules change comes with -E/-G/-P/-F as real clap flags on `rtk grep`. last dialect flag still wins, and the rg `-E`-is-encoding issue doesn't bite here since our -E lands on rtk's own CLI (grep semantics), never on rg's.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: fair, that was backwards. explicit -G now runs native grep directly.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: done, Basic never reaches the rg path anymore, no translation either.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: it didn't. pulled the stripping into `rg_forward_args()`, which walks short clusters now: `-rn` → `-n`, `-iE` → `-i`, `-rE` dropped whole. tokens after `--` pass through untouched. tests cover each case.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: the new `test_bre_to_pcre_alternation` covers the same translation, but I put the old `fn foo\|pub.*bar` case back into it anyway.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: restored, and it now tests the real function instead of an inline copy of the filter logic.

src/cmds/system/grep_cmd.rs
```diff
@@ -306,24 +464,190 @@ mod tests {
...
+            bre_to_pcre(pattern),
+            r"private function delete|try|catch|finally|delete\(\)"
+        );
+    }
+
+    // --- issue #1436: args forwarded to rg ---
+
+    fn args(list: &[&str]) -> Vec<String> {
+        list.iter().map(|s| s.to_string()).collect()
+    }
```
KuSh: nitpick, you can make `rg_forward_args` accepts `&[impl AsRef<str>]` and avoid such boilerplate

then your tests simplifies to

```rs
    #[test]
    fn test_rg_forward_args_strips_recursive() {
        // rg is recursive by default and re-purposes -r as --replace.
        assert_eq!(rg_forward_args(&["-r", "-i"]), vec!["-i"]);
        assert_eq!(rg_forward_args(&["--recursive"]), Vec::<String>::new());
    }

File: src/main.rs [hunk 15]
```diff
@@ -322,6 +322,18 @@ enum Commands {
...
+        extended_regexp: bool,
+        /// Pattern is a basic regexp (grep -G); runs native grep
+        #[arg(short = 'G', long)]
+        basic_regexp: bool,
+        /// Pattern is a Perl-compatible regexp (grep -P)
+        #[arg(short = 'P', long)]
+        perl_regexp: bool,
+        /// Pattern is a fixed string, not a regexp (grep -F)
+        #[arg(short = 'F', long)]
+        fixed_strings: bool,
```
KuSh: That should be avoided as clap with eat the args and mess with them. I've pushed a [PR](https://github.com/rtk-ai/rtk/pull/2333) that go the opposite and removes most grep args handlings from clap.

I don't see why this is needed, if extra_args contains everything this should work without specific handling. Or am I missing something?

KuSh (changes requested):
Thanks for changes, unfortunately change in src/main.rs will conflict with other grep planned fixes and doesn't seem necessary.

KuSh:
Okay, I've iterated a lot on grep handling through #2333. You'll have to rebase once merged, here's a full detailed test suite you'll have to take into account to be sure nothing is left unhandled:

## `bre_to_pcre()` — unit tests on the translation function

The core obligation: BRE literals that are PCRE metacharacters must be escaped, and BRE escapes that mean "special" must be unescaped.

```
// Literals in BRE that are special in PCRE → must be escaped
"delete()"      → "delete\(\)"     // parens: literal in BRE, group in PCRE
"a+b"           → "a\+b"           // + is literal in BRE
"a?b"           → "a\?b"           // ? is literal in BRE
"foo|bar"       → "foo\|bar"       // | is literal in BRE
"a{2}"          → "a\{2\}"         // braces are literal in BRE
"a{2,3}"        → "a\{2,3\}"

// BRE escapes that mean PCRE special → must be unescaped
"\(group\)"     → "(group)"        // BRE grouping → PCRE group
"a\{2\}"        → "a{2}"           // BRE repetition → PCRE repetition
"foo\|bar"      → "foo|bar"        // GNU BRE alternation → PCRE alternation
"a\+"           → "a+"             // GNU BRE one-or-more → PCRE
"a\?"           → "a?"             // GNU BRE optional → PCRE

// Passthrough: same meaning in BRE and PCRE
".*"            → ".*"
"^foo$"         → "^foo$"
"[a-z]"         → "[a-z]"
"\."            → "\."             // escaped dot = literal in both
"foo\nbar"      → "foo\nbar"       // \n is not a BRE metachar

// Composed
"fn delete(.*)"  → "fn delete\(.*\)"
```

---
## `detect_syntax()` — unit tests on flag scanning

Input: the `flags` vec from `extract_pattern_path`. The function returns the detected dialect.

```
// Defaults
[]                          → BRE
["-i", "-n"]                → BRE     // no dialect flag

// Long forms
["--extended-regexp"]       → ERE
["--basic-regexp"]          → BRE
["--perl-regexp"]           → PCRE
["--fixed-strings"]         → Fixed

// Short forms (these land in flags as boolean, since -E not in VALUE_FLAGS_SHORT)
["-E"]                      → ERE
["-G"]                      → BRE
["-P"]                      → PCRE
["-F"]                      → Fixed

// Last wins
["-E", "-G"]                → BRE
["-G", "-E"]                → ERE
["-E", "-P"]                → PCRE
["-P", "-F"]                → Fixed

// Stop at "--" (flags never contain "--" after extract_pattern_path, but defensive)
["-E", "--", "-G"]          → ERE   // stops at "--"

// Short cluster: "-nE" → extract_pattern_path emits "-nE" since E is not in
// VALUE_FLAGS_SHORT and n is boolean → both end up in bool_acc → flag "-nE"
// detect_syntax must scan individual chars within a short cluster
["-nE"]                     → ERE
["-inP"]                    → PCRE

// Interleaved with other flags
["-i", "--glob", "*.rs", "-E", "-A", "2"]  → ERE
```

**Important interaction with the current branch**: `extract_pattern_path` does NOT consume `-E` as value-taking (intentionally left out of `VALUE_FLAGS_SHORT`). So `-E` appears in `flags` as-is. `detect_syntax` receives it correctly for standalone use. For the cluster case (`-nE`), the left-to-right scan hits `E` — since `E` is NOT in `VALUE_FLAGS_SHORT`, it gets accumulated into `bool_acc` alongside `n`, producing the flag `-nE`. `detect_syntax` must handle this by checking each char of short flag clusters, not just exact string matches.

---
## Dialect flag stripping — flags forwarded to rg

After detect_syntax, the dialect flags must be removed from `extra_args` before rg invocation. None of them are safe to forward raw:

|       Flag          |                               Why it must be stripped                                |
|---------------------|--------------------------------------------------------------------------------------|
| `-E`                | rg treats it as `--encoding` (value-taking) — next arg consumed as encoding name     |
| `-G`                | rg has no `-G`, would error                                                          |
| `--basic-regexp`    | rg has no `--basic-regexp`                                                           |
| `--extended-regexp` | rg has no `--extended-regexp`                                                        |
| `-P`                | rg has no `-P`; use `--pcre2` instead                                                |
| `--perl-regexp`     | rg has no `--perl-regexp`                                                            |
| `-F`                | rg DOES have `-F`, safe to forward — but strip-and-re-add is cleaner for consistency |
| `--fixed-strings`   | rg has `--fixed-strings`, safe to forward                                            |

```
// rg invocation — per dialect
BRE  "delete()" src  →  rg ... -e 'delete\(\)' -- src    // translated, no dialect flag
ERE  -E 'foo(bar)+' src → rg ... -e 'foo(bar)+' -- src   // NOT translated, -E stripped
PCRE -P 'foo\d+' src    → rg ... --pcre2 -e 'foo\d+' -- src  // -P stripped, --pcre2 added
Fixed -F 'foo.bar' src  → rg ... -F -e 'foo.bar' -- src  // -F stripped then re-added as -F
```

---
## grep fallback dialect

The grep fallback gets original (untranslated) patterns plus grep-compatible flags:

```
// grep receives the dialect flag because grep understands it
// grep does NOT receive --pcre2 (it doesn't know that)
BRE  "delete()" src  →  grep -e 'delete()' -rnHZ -- src   // no translation for grep
ERE  -E 'foo' src    →  grep -E -e 'foo' -rnHZ -- src     // -E is valid grep flag
PCRE -P 'foo' src    →  grep -P -e 'foo' -rnHZ -- src     // grep may support -P (PCRE)
Fixed -F 'foo' src   →  grep -F -e 'foo' -rnHZ -- src
```

---
## `-E` parsing conflict

This is the hardest interaction. Currently `rtk grep -E utf8 PATTERN` src produces:

- `extract_pattern_path` sees: `-E` (bool flag → flags), `utf8` (positional → pattern!), `PATTERN` (positional → path), `src` (positional → path)
- Result: `patterns=["utf8"]`, `paths=["PATTERN", "src"]` ← wrong

#2138 must fix this. The canonical approach: a pre-pass that strips dialect flags from the raw arg list before `extract_pattern_path` runs.

```
// After pre-pass strips -E:
// raw: ["-E", "utf8", "PATTERN", "src"]
// → #2138 detects -E, removes it
// → extract_pattern_path sees: ["utf8", "PATTERN", "src"]  ← still wrong if utf8 is encoding value

// This only works cleanly if #2138 knows -E takes NO positional value in grep context
// (grep's -E is boolean; rg's --encoding takes a value — they're different flags)

Concrete tests #2138 must pass:
// -E is boolean in grep, so no value consumed
"-E" "'foo(bar)+'" "src"  → patterns=["foo(bar)+"], paths=["src"], dialect=ERE

// -P is boolean in grep
"-P" "'\d+'" "src"        → patterns=["\d+"], paths=["src"], dialect=PCRE

// -F is boolean in grep
"-F" "'foo.bar'" "src"    → patterns=["foo.bar"], paths=["src"], dialect=Fixed

// mixed cluster with E at the end
"-inE" "'foo'" "src"      → patterns=["foo"], paths=["src"], flags=["-in"], dialect=ERE
```
---
## rg-origin rewrite (discover/registry.rs)

When the hook converts `rg 'pat'` → `rtk grep 'pat'`, patterns that were already PCRE must not go through BRE translation. The fix is appending `-E` to the rewritten command.

```
rg 'foo(bar)+'         → rtk grep 'foo(bar)+' -E      // -E preserves PCRE semantics
rg -e 'foo(bar)+'      → rtk grep -e 'foo(bar)+' -E
rg -e 'foo' -e 'bar'   → rtk grep -e 'foo' -e 'bar' -E
rg 'foo\|bar'          → rtk grep 'foo\|bar' -E        // even BRE-looking patterns get -E since source is rg

// End-to-end: the -E survives to detect_syntax and suppresses BRE translation
extract_pattern_path(["-E", "foo(bar)+", "src"])  // after pre-pass removes -E
→ detect_syntax returns ERE
→ rg receives -e 'foo(bar)+' (no translation)  ✓
```

---
## Multiple patterns

```
// Each pattern translated independently
"-e" "delete()" "-e" "foo\|bar" "src"
→ BRE → rg: -e 'delete\(\)' -e 'foo|bar' -- src

// ERE: no translation
"-E" "-e" "foo(bar)+" "-e" "baz" "src"
→ ERE → rg: -e 'foo(bar)+' -e 'baz' -- src

// --regexp (long form of -e) treated identically
"--regexp" "delete()" "src"
→ BRE → rg: -e 'delete\(\)' -- src
```

F0rty-Tw0:
Hey @KuSh, sorry for the long silence. Changed jobs and got buried, this one also turned out trickier than the few-line fix I thought it was, so it kept sliding.

Caught up now. I see #2333 landed and pulls grep handling out of Clap,  which is the right call and the opposite of where I'd gone with the dialect flags. I'll rebase onto it, drop the Clap-flag approach, and re-seat detection on top of  `extract_pattern_path` (incl. the pre-pass for the `-E`-eats-the-positional case and the -`nE` cluster scan). The test suite you wrote out is exactly what I needed, thanks for taking the time, I'll make all of it pass.

Back on track, will push the rebased version shortly.

F0rty-Tw0:
pushed!

quick thing before i wrap this up: the -`G` routing. first round you asked for `Basic` to skip straight to native grep, so hat's what i've got. explicit `-G/--basic-regexp` runs grep directly, no rg, no translation.

the new strip-table puts `-G` under "strip before rg" though, which reads like you'd rather it go through rg now (stripped, pattern BRE-translated) than hit native grep. which did you mean? either works, one-liner to switch.

KuSh:
> the new strip-table puts `-G` under "strip before rg" though, which reads like you'd rather it go through rg now (stripped, pattern BRE-translated) than hit native grep. which did you mean? either works, one-liner to switch.

Yeah, Claude has gone way over its head here, dialect flags should select the correct binary as we have discussed before, this way it will never hit rg and use grep as first intended by the caller. So keep tests as it was before: if a dialect IS requested direct to the correct binary, if none is requested we can use rg (and translate if necessary)

src/cmds/system/grep_cmd.rs
```diff
@@ -251,6 +251,180 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
...
+
+/// Filters grep-style extra args down to what is forwarded to rg.
+///
+/// Drops `-r`/`--recursive` (rg is recursive by default; rg's `-r` means
+/// `--replace`) and the dialect selectors (`-E`/`-G`/`-P`/`-F` and long forms),
+/// which are applied separately as an rg engine flag or pattern translation and
+/// would otherwise be misread by rg (`-E` is rg's `--encoding`). Handles short
+/// clusters: `-rin` → `-in`, `-iE` → `-i`, `-rE` drops entirely. Tokens after
+/// `--` are operands and pass through untouched, as does anything from `=` on.
+fn rg_forward_args(extra_args: &[String]) -> Vec<String> {
```
KuSh: This should be integrated in strip_r and strip_recursive (renamed to strip_short_args / strip_long_args) and used by extract_pattern_path so that extra_flags are already stripped during the split

src/cmds/system/grep_cmd.rs
```diff
@@ -312,48 +486,71 @@ pub fn run(
...
+            rg_cmd.arg("--type").arg(ft);
+        }
 
-    // extra_args is already stripped of -r/-R/-recursive by extract_pattern_path
-    rg_cmd.args(&extra_args);
+        // Engine selection: -P needs rg's PCRE2 engine; -F is literal-string mode.
+        // ERE and the BRE default both run on rg's default engine.
+        match syntax {
+            Some(RegexSyntax::Pcre) => {
+                rg_cmd.arg("--pcre2");
```
KuSh:
```suggestion
                rg_cmd.arg("-P");
```

src/cmds/system/grep_cmd.rs
```diff
@@ -251,6 +251,180 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
...
+}
+
+/// Returns the regex dialect requested via a grep dialect flag, or `None` for
+/// the POSIX default (BRE — what plain `grep` uses with no dialect flag).
+///
+/// Matches GNU grep semantics: the LAST dialect flag wins (`-E -F` → Fixed), and
+/// a dialect char bundled in a short cluster counts too (`-iE` → Ere). Scanning
+/// stops at `--` (everything after is an operand) and at `=` within a token (so
+/// a `--color=…` value can't false-trigger).
+fn detect_syntax(extra_args: &[String]) -> Option<RegexSyntax> {
```
KuSh, line 276: This doesn't handle rg --engine arg

Also, don't know what to do in the --auto-hybrid-regex / --engine=auto except that it should use rg and not grep. There might be a hidden issue with those flags here

KuSh (changes requested):

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Fixed with `has_rg_engine_flag`: `--engine`/`--auto-hybrid-regex` force rg and skip translation. Also caught `--engine=pcre2 'foo(bar)'` getting BRE-escaped.

F0rty-Tw0:
All four dialects now route to native grep; only the no-dialect default hits rg + translation. Deleted `rg_forward_args` and the engine-select block (~-65 lines).

One deviation: rg-origin marker is `--engine=auto`, not `-E`. `-E` would send rg patterns to `grep -E` and drop `\d`/`\w`/`\b`. One-line revert if you disagree.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Renamed to `strip_short_args`/`strip_long_args`. Second pass deleted: stripping happens in the split, dialects go to grep, nothing rg-bound needs it.

src/cmds/system/grep_cmd.rs
F0rty-Tw0: Block deleted. All dialects route to grep, so rg only runs the default path. No engine selection left.

KuSh:
Hi @F0rty-Tw0 sorry for the delay, could you please rebase and resolve conflicts?

KuSh:
Thanks for working on this! I just wanted to let you know that @aeppling merged a refactor (https://github.com/rtk-ai/rtk/pull/2641) in this area that actually addresses the grep-to-rg translation piece. So it looks like we might not need this PR anymore.

F0rty-Tw0:
Hi @KuSh, thanks for the confirmation. I also was extremely busy with other stuff and couldn't focus on this one and push it till the end.

Thanks @aeppling for making it happen. You can close this pr if you find it necessary, and nothing else can benefit to the original issue. 

Happy to help you on other windows/.net related issues.

F0rty-Tw0:
Closing this one. #2641 took #1436 the other way, running the invoked engine instead of substituting rg for grep, and that's the better call. The translation approach here doesn't have a place anymore, and rebasing would just bring back the code that PR removed.

Where I think #1436 stands:
- Dialect mismatch (the extra matches in the original report): fixed by #2641, grep runs native grep now.
- The single-file `2F` / mangled line 81: that's the parsing side, and #1554 was the fix aimed at it.

I haven't reproduced the original case against current develop, so I'm not closing the issue myself. If someone runs that repro and it comes back clean, #1436 can close. Nothing on this branch is worth salvaging.

PR #2141, opened 2026-05-28
fix(pytest): pass --version through unfiltered

### LINKED ISSUE #2096: rtk pytest --version returns 'no tests collected' — wrapper consumes the flag
## Summary

`rtk pytest --version` returns RTK's "no tests collected" filter output instead of pytest's version string. The wrapper consumes `--version` before forwarding to pytest.

## Environment

- rtk 0.42.0 (Homebrew, installed 2026-05-24)
- macOS Darwin 24.6.0 (x86_64)
- pytest installed in venv (verified works directly)

## Repro

```
$ rtk pytest --version
Pytest: No tests collected

$ pytest --version
pytest 8.x.y
```

## Expected

Forward `--version` to pytest, get pytest's normal `pytest X.Y.Z` output back.

## Actual

The wrapper treats `--version` as a no-tests-specified RTK-level signal and runs pytest with no args, producing "no tests collected" — which RTK then formats as "Pytest: No tests collected". The actual `--version` query never reaches pytest.

## Scope

`--version` is the obvious one; may also affect other passthrough flags pytest understands but the wrapper doesn't explicitly own. `--help`, `--co`, `--collect-only`, etc. worth checking.

## Suggested fix

The pytest wrapper should pass through any flag it doesn't explicitly own to pytest unchanged. Reserve filter behavior for invocations where actual test args are provided.

## Workaround

Run `pytest --version` directly (bypassing RTK), or use `rtk run pytest --version` (raw passthrough).

maxmilian:
## Summary

- `rtk pytest --version` returned `Pytest: No tests collected` instead of pytest's version string. Fixes #2096.
- Route `--version` around the output filter and forward pytest's output verbatim.

## ⚠️ Note for reviewers: `--version` may now *look* fixed on current `develop`

Re-verified against `develop` @ `f9d8c77`. If you test this by running `rtk pytest --version` on a recent build, **you will likely see the correct version string and conclude the bug is gone.** It isn't — that's an incidental byte-length side effect of a guard that landed after this PR was opened.

`861a46d` (2026-06-23) added `never_worse` (`src/core/guard.rs:6`), applied globally via `emit_guarded` (`src/core/runner.rs:17`), which returns whichever of raw/filtered output is *fewer tokens*. For the specific `--version` case the raw banner (`pytest 9.0.3`, ~4 tokens) is shorter than `"Pytest: No tests collected"` plus the tee hint (7+ tokens), so the guard discards the filtered string and the real banner reaches the user.

That is a length coincidence, not a fix. The underlying collapse at `src/cmds/python/pytest_cmd.rs:183-185` is untouched, so these still break:

- **`rtk pytest -VV`** — the long plugin banner makes the filtered string the *shorter* one, the guard no longer rescues it, and the output collapses to `"Pytest: No tests collected"` again.
- **Collection-error runs** — see the 2026-06-25 report in #2096, i.e. filed *after* the guard merged.

This PR fixes the root cause instead of relying on output length, so it holds regardless of how long the banner is.

## Root cause

`pytest --version` prints its banner and exits **without running a test session**, so `filter_pytest_output` finds no `passed/failed/skipped` summary line. `build_pytest_summary` then hits the all-zero-counts branch (`src/cmds/python/pytest_cmd.rs:183`) and returns the hardcoded `"Pytest: No tests collected"` — the real banner is discarded. (The reporter's "wrapper consumes the flag" hypothesis is close; the flag *is* forwarded, but the filter swallows the response.)

## Fix

In `run()`, detect `--version` before flag injection and forward the command with an identity filter (no `-q`/`--tb=short`/`-rxX` injection, no compression), so pytest's output reaches the caller byte-for-byte. Honors RTK design principle #1 (Correctness over Token Savings): a flag that explicitly asks for informational output must not be reformatted into a lossy summary.

## Scope (deliberately narrow)

- **`--version` only.** `-V`/`-VV` are version-dependent — pytest 9's bare `-V` runs a full session, so treating it as informational would dump a raw, unfiltered test run (a regression). Encoding that fragile per-version behavior isn't worth it; `--version` is the reported, unambiguous case.
- **`--help`/`-h` are out of scope:** clap intercepts them at the subcommand layer and prints rtk's own wrapper help — they never reach `run()`.
- **`--collect-only`/`--co` left untouched** — already addressed by #925.

## Tests

- `test_version_banner_swallowed_by_filter` — documents the buggy filter behavior (banner → "No tests collected").
- `test_is_informational_flag` — `--version` matches; `-V`, `-VV`, `--help`, `-h`, `-v`, `--verbose`, `--collect-only`, and test paths do not.
- Full suite: 1948 passed, clippy clean, `cargo fmt --check` clean.

## Manual verification

```
$ rtk pytest --version          # before: "Pytest: No tests collected"
pytest 9.0.3                    # after: real version
$ rtk pytest -V                 # still filtered (no regression)
```

(Recorded before `861a46d`; see the reviewer note above for why the "before" line no longer reproduces on current `develop` for this one flag, while `-VV` and collection errors still do.)

### DIFF
--- a/src/cmds/python/pytest_cmd.rs
+++ b/src/cmds/python/pytest_cmd.rs
@@ -16,6 +16,19 @@ enum ParseState {
     Summary,
 }
 
+/// True for `--version`, which makes pytest print its banner and exit without a
+/// test session. The filter then finds no summary and collapses the banner to a
+/// misleading "No tests collected" (issue #2096), so this invocation must bypass
+/// the filter.
+///
+/// Deliberately narrow: `rtk pytest --help`/`-h` print rtk's own wrapper help and
+/// never reach this function, and pytest's short/count version forms are version-
+/// dependent — pytest 9's bare `-V` actually runs a session, so treating it as
+/// informational would dump a raw, unfiltered test run.
+fn is_informational_flag(arg: &str) -> bool {
+    arg == "--version"
+}
+
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let mut cmd = if tool_exists("pytest") {
         resolved_command("pytest")
@@ -24,6 +37,25 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         c.arg("-m").arg("pytest");
         c
     };
+
+    // Informational invocations (--version/--help) produce no test session for the
+    // filter to parse. Forward them verbatim and pass the output through unchanged
+    // so the real banner/usage reaches the caller instead of "No tests collected".
+    if args.iter().any(|a| is_informational_flag(a)) {
+        for arg in args {
+            cmd.arg(arg);
+        }
+        if verbose > 0 {
+            eprintln!("Running: pytest {}", args.join(" "));
+        }
+        return runner::run_filtered(
+            cmd,
+            "pytest",
+            &args.join(" "),
+            |s| s.to_string(),
+            runner::RunOptions::stdout_only(),
+        );
+    }
 
     let has_tb_flag = args.iter().any(|a| a.starts_with("--tb"));
     let has_quiet_flag = args.iter().any(|a| a == "-q" || a == "--quiet");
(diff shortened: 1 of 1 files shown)


maxmilian:
Note for the PR Testing Checklist: the **"token savings ≥60%"** item is **N/A** here. This is a correctness fix, not a new/compressing filter — `rtk pytest --version` was being collapsed to "No tests collected" because the filter has no test-session summary to parse. The fix *bypasses* the filter for `--version` so pytest's real output reaches the caller. This applies design principle **Correctness vs Token Savings**: a flag explicitly requesting informational output must not be reformatted into a lossy summary.

Remaining checklist items:
- Unit tests added (`test_version_banner_swallowed_by_filter`, `test_is_informational_flag`, combined-args detection)
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — all green (1948 passed)
- Manual `rtk pytest --version` → real version string; `rtk pytest -V` confirmed to stay on the filtered path (no regression)

Scope is deliberately narrow (`--version` only) for correctness — `-V`/`-VV` are pytest-version-dependent and `--help`/`-h` are handled by clap before reaching the wrapper; details in the PR body.

maxmilian:

PR #2142, opened 2026-05-28
fix(discover): rewrite dotnet test/restore/format, not just build

### LINKED ISSUE #1830: [rewrite] dotnet test|restore|format should auto-rewrite to rtk dotnet

maxmilian:
## Summary

- `rtk dotnet test`, `rtk dotnet restore`, and `rtk dotnet format` proxies already exist and route to dedicated handlers, but `rtk rewrite` only matched `dotnet build` — so the Claude Code hook silently passed the other three through unfiltered. Fixes #1830.

## Root cause

The `dotnet` rewrite rule in `src/discover/rules.rs` used `pattern: r"^dotnet\s+build\b"`, covering only `build`. `DotnetCommands::Test/Restore/Format` all dispatch to real handlers (`src/main.rs:1709-1711`), so the proxies work when invoked directly — only the rewrite registry was missing them.

## Fix

Widen the rule's pattern to `^dotnet\s+(build|test|restore|format)\b`. The output filters already handle all four; this is purely a routing fix (one regex).

## Tests

- `test_rewrite_dotnet_subcommands` — all four rewrite to `rtk dotnet <sub>`; `dotnet publish` / `dotnet run` (no proxy) are left untouched.
- Full suite: 1947 passed, clippy clean, `cargo fmt --check` clean.

## Manual verification

```
dotnet build    -> rtk dotnet build
dotnet test     -> rtk dotnet test
dotnet restore  -> rtk dotnet restore
dotnet format   -> rtk dotnet format
dotnet publish  -> (not rewritten)
```

## Scope

Only the four subcommands with existing proxies. `dotnet build` and the three new subcommands now share one rule and behave identically (same permission verdict / exit code). The exit-code (ask vs auto-allow) is governed by a separate permission layer and is unchanged by this routing fix.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -2337,6 +2337,30 @@ mod tests {
     }
 
     #[test]
+    fn test_rewrite_dotnet_subcommands() {
+        // All four subcommands have dedicated `rtk dotnet` proxies, so the rewrite
+        // rule must route them (issue #1830 — only `build` was covered before).
+        for sub in ["build", "test", "restore", "format"] {
+            assert_eq!(
+                rewrite_command_no_prefixes(&format!("dotnet {}", sub), &[]),
+                Some(format!("rtk dotnet {}", sub)),
+                "dotnet {} should rewrite",
+                sub
+            );
+        }
+        // Trailing args are preserved (the \b stops at the subcommand).
+        assert_eq!(
+            rewrite_command_no_prefixes("dotnet test --filter Category=Unit", &[]),
+            Some("rtk dotnet test --filter Category=Unit".into())
+        );
+        // Subcommands without a proxy must NOT be rewritten.
+        assert_eq!(rewrite_command_no_prefixes("dotnet publish", &[]), None);
+        assert_eq!(rewrite_command_no_prefixes("dotnet run", &[]), None);
+        // Word boundary guards against false prefixes.
+        assert_eq!(rewrite_command_no_prefixes("dotnet builder", &[]), None);
+    }
+
+    #[test]
     fn test_rewrite_compound_and() {
         assert_eq!(
             rewrite_command_no_prefixes("git add . && cargo test", &[]),
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -2338,8 +2338,8 @@ mod tests {
 
     #[test]
     fn test_rewrite_dotnet_subcommands() {
-        // All four subcommands have dedicated `rtk dotnet` proxies, so the rewrite
-        // rule must route them (issue #1830 — only `build` was covered before).
+        // All four subcommands have dedicated `rtk dotnet` proxies, so the
+        // rewrite rule must route them.
         for sub in ["build", "test", "restore", "format"] {
             assert_eq!(
                 rewrite_command_no_prefixes(&format!("dotnet {}", sub), &[]),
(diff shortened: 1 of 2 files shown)


maxmilian:
Ran a fresh-context independent review (skeptical, diff + issue only). Verdict: no blockers. Two points addressed / noted:

- **Hardened the test** (pushed): added a trailing-args case (`dotnet test --filter Category=Unit` → preserved) and a word-boundary negative (`dotnet builder` → not rewritten), confirming `\b` stops at the subcommand.
- **Kept the capturing group** `(build|test|restore|format)` to stay consistent with the other multi-subcommand rules in `rules.rs` (git/gh/cargo/ruff/go), which capture for `subcmd_savings` lookup — and it leaves room for the follow-up below.

**Follow-up (intentionally out of scope here):** the rule's flat `savings_pct: 70.0` now applies to all four subcommands. `dotnet test` likely compresses closer to ~90% (cf. `cargo test`/`go test`), so `rtk discover` will slightly undercount its savings. I left this out of a pure routing fix because I don't have reliable compression figures for `restore`/`format` and didn't want to guess. Happy to file a separate PR adding `subcmd_savings` overrides if maintainers want the report accuracy.

src/discover/registry.rs
```diff
@@ -1283,6 +1283,30 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_rewrite_dotnet_subcommands() {
+        // All four subcommands have dedicated `rtk dotnet` proxies, so the rewrite
+        // rule must route them (issue #1830 — only `build` was covered before).
```
harvzor, line 1289: nitpick (non-blocking): Not sure if it's worth having this historical context here

Suggestion is to remove:

```suggestion
```

harvzor (approved):

maxmilian:
Friendly ping on this one — #1830 is still open and picked up a +1 from a heavy user (reported `dotnet test` running 1208× over 30 days at ~0% compression), and the PR already has an approval. Mergeable. Any chance of a review slot?

src/discover/registry.rs
maxmilian: Good call — thanks for the review, and sorry for the slow reply.

Dropped in `24ac995`. I kept the first half, since "all four subcommands have dedicated `rtk dotnet` proxies, so the rewrite rule must route them" is the reason the loop asserts on exactly those four and not, say, `publish`. What went is the `(issue #1830 — only `build` was covered before)` aside, which is the historical part and is already recorded in the PR and the issue.

`cargo test test_rewrite_dotnet_subcommands` still passes and `cargo fmt --check` is clean.

PR #2143, opened 2026-05-28
fix(cicd): scope semgrep filesystem-deletion rule to exclude src/hooks/

### LINKED ISSUE #1954: semgrep: filesystem-deletion rule lacks paths:exclude for src/hooks/
### Problem

The `filesystem-deletion` rule in `.semgrep.yml` matches the syntactic pattern `fs::remove_file(...)` / `fs::remove_dir_all(...)` with no `paths:` scoping. Its own message states the intent:

> File/directory deletion detected. **Expected in hooks/init cleanup**, surprising in a filter module. Verify intent.

But because the rule has no path filter, it scans `src/hooks/` too. The CI runs `semgrep scan --config .semgrep.yml --baseline-commit <base> --error`, so any PR that adds a *new* deletion call to `src/hooks/init.rs` (uninstall logic for a new agent) produces a fresh finding and fails the scan — even though deletion there is exactly what the rule says it expects.

`init.rs` already contains ~8 `fs::remove_file` calls (Claude, Gemini, Cursor, Codex uninstall paths); they only pass CI because they predate the baseline. Every future agent-uninstall PR will hit this false positive (e.g. #1741, Pi support).

### Proposed fix

Add a `paths: exclude` so the rule reflects its stated intent — flag deletions in filter modules, not in the hooks/init cleanup code:

```yaml
  - id: filesystem-deletion
    pattern-either:
      - pattern: fs::remove_file(...)
      - pattern: fs::remove_dir_all(...)
      - pattern: std::fs::remove_file(...)
      - pattern: std::fs::remove_dir_all(...)
    paths:
      exclude:
        - src/hooks/
    message: >
      ...
```

### Workaround until fixed

New uninstall PRs can add `// nosemgrep: filesystem-deletion` on the offending line.

maxmilian:
## Summary

- The `filesystem-deletion` semgrep rule has no `paths:` scoping, so it flags `src/hooks/` even though its own message says deletion is *"Expected in hooks/init cleanup"*. Fixes #1954.

## Root cause

`.semgrep.yml` rule `filesystem-deletion` matches `fs::remove_file(...)` / `remove_dir_all(...)` repo-wide. CI runs `semgrep scan --config .semgrep.yml --baseline-commit <base> --error`, so any PR adding a *new* deletion to `src/hooks/init.rs` (e.g. a new agent's uninstall path) produces a fresh finding and fails the scan — a guaranteed false positive for every future agent-uninstall PR (e.g. Pi support #1741). The existing ~10 deletions in `init.rs` only pass because they predate the baseline.

## Fix

Add `paths.exclude: src/hooks/` to the rule, so it flags deletions in filter modules (the surprising case) but not in the hooks/init cleanup code (the expected case).

## Verification (semgrep before/after)

Ran `semgrep scan --config .semgrep.yml` locally:

| Target | Before (no exclude) | After (exclude src/hooks/) |
|---|---|---|
| `src/hooks/init.rs` | **10** findings | **0** |
| whole `src/hooks/` | (flagged) | **0** |
| `src/core/tee.rs` (filter module) | 1 | **1** (still flagged) |

The rule still catches deletions outside `src/hooks/`, confirming the exclude isn't over-broad.

## Scope

Config-only, 3 lines. No Rust changes, so no unit test applies — the behavior is verified by the semgrep runs above. Does not touch any other rule.

### DIFF
--- a/.semgrep.yml
+++ b/.semgrep.yml
@@ -101,6 +101,9 @@ rules:
       - pattern: fs::remove_dir_all(...)
       - pattern: std::fs::remove_file(...)
       - pattern: std::fs::remove_dir_all(...)
+    paths:
+      exclude:
+        - src/hooks/
     message: >
       File/directory deletion detected. Expected in hooks/init cleanup,
       surprising in a filter module. Verify intent.


maxmilian:
Ran a fresh-context independent review (skeptical, diff + issue only). Verdict: approve, no blockers. Confirmed:

- The `paths:`/`exclude:` block is valid semgrep schema, indented as a sibling key of `filesystem-deletion` (not nested under a pattern), and mirrors the existing `ureq-outside-telemetry` rule's `paths` block. `filesystem-deletion` is the last rule in the file, so there's no risk of leaking into an adjacent rule.
- Bare `src/hooks/` is the idiomatic recursive directory exclude (`src/hooks/*` would be non-recursive; `**` redundant).
- Security tradeoff reviewed: the rule is WARNING-level and its stated intent is exactly "expected in hooks/init cleanup", so a directory-level exclude is proportionate. Narrowing to `init.rs` only would re-introduce false positives from `integrity.rs`/`hook_cmd.rs` (which also contain expected uninstall deletions). Left as directory-level deliberately.

maxmilian:

PR #2149, opened 2026-05-29
fix(grep): skip .claude/worktrees by default

### LINKED ISSUE #2147: rtk grep should ignore .claude/worktrees
Claude Code creates worktrees in the folder `.claude/worktrees`.
When I run `rtk grep`, it also grips the git worktrees in this folder. This is problematic, since these folders are out of scope for any agentic session.

`rtk grep` should automatically skip this folder - at least, when used together with Claude Code.

I think `git grep` and `rg` are ignoring `.gitignore`

nanookclaw:
## Summary

- `rtk grep` no longer descends into Claude Code's `.claude/worktrees` checkouts. Because `rtk grep` runs ripgrep with `--no-ignore-vcs` (and a user may add `--hidden`), the gitignored worktree copies were being searched, surfacing duplicate, out-of-scope matches. Closes #2147.
- Adds a default ripgrep exclusion glob `!**/.claude/worktrees/**` (matches the directory at the repo root and any nested depth) plus an `--exclude-dir=worktrees` mirror on the `grep` fallback path.
- The exclusion is inserted **before** user args, so a later user `--glob` still wins (ripgrep honors the last matching glob) and can re-include the directory if explicitly wanted.

`--no-ignore-vcs` semantics, `--type` handling, the `-r`/`--recursive` strip, and BRE→PCRE pattern translation are all preserved.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — clippy clean, full suite 1954 passed / 0 failed.
- [x] 8 new unit tests in `grep_cmd.rs`: arg-construction (glob present/shape/any-depth, exclusion precedes user args on both backends, user args + `--type` preserved) and one end-to-end test that runs real `rg` against a fixture with root + nested `.claude/worktrees` under `--hidden`.
- [x] Manual testing against a temp fixture (root + nested `.claude/worktrees` + a normal file):
  - Before (raw `rg --no-ignore-vcs --hidden`): 3 matches, incl. both worktree files.
  - After (`rtk grep SENTINEL . --hidden`): only the non-worktree file matches; worktrees skipped at root and nested depth.
  - `rtk grep SENTINEL . --hidden --glob '**/.claude/worktrees/**'`: re-includes the worktree files (override works).
  - With `rg` absent, the `grep` fallback (`--exclude-dir=worktrees`) also skips the directory.

Fallback note: GNU grep's `--exclude-dir` is base-name only, so the fallback skips any directory named `worktrees`. It only runs when ripgrep is unavailable, so the slight over-exclusion is an acceptable trade-off; this is documented in code.

AI-assisted: implemented with Claude Code, reviewed with a second independent model pass before submitting.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -7,6 +7,20 @@ use crate::core::{args_utils, config};
 use anyhow::{Context, Result};
 use regex::Regex;
 use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::path::{Component, Path};
+
+/// Glob passed to ripgrep to skip Claude Code's worktree checkouts.
+///
+/// Claude Code creates throwaway branch checkouts under `.claude/worktrees`.
+/// They are gitignored and out of scope for normal scans, but `rtk grep`
+/// passes `--no-ignore-vcs`, and users may add `--hidden`, so ripgrep would
+/// otherwise descend into them. The leading `**/` matches at any depth,
+/// including the repository-root `.claude/worktrees`; the trailing `/**`
+/// excludes the directory's contents. Applied only when the requested paths are
+/// not already inside `.claude/worktrees`. Added before any user-supplied args
+/// so a later user `--glob` can still re-include the directory.
+const WORKTREES_RG_GLOB: &str = "!**/.claude/worktrees/**";
 
 /// Short single-char flags that consume one following token (or inline remainder)
 /// as their value. `-e` is handled separately — its value goes to `patterns`.
@@ -335,10 +349,14 @@ fn grep_capture<T: AsRef<str>>(
     patterns: &[String],
     paths: &[String],
 ) -> Result<CaptureResult> {
+    let skip_worktrees = should_skip_claude_worktrees_by_default(paths);
     let mut rg_cmd = resolved_command("rg");
     rg_cmd.args(rg_base);
     if let Some(ft) = file_type {
         rg_cmd.arg("--type").arg(ft);
+    }
+    if skip_worktrees {
+        rg_cmd.args(["--glob", WORKTREES_RG_GLOB]);
     }
     rg_cmd.args(extra_args.iter().map(|a| a.as_ref()));
     for p in patterns {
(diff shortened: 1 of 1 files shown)


nanookclaw:
I have read the CLA Document and I hereby sign the CLA

aeppling:
Hey @nanookclaw 

Thanks for addressing this,


### Blocker: explicit-path searches into .claude/worktrees now silently return nothing

The exclusion glob is applied unconditionally, including when the user's own path argument points at the worktrees.
Verified against the built PR branch vs. rtk 0.40.0:

```bash
  mkdir -p repro/.claude/worktrees/wt1 && cd repro && git init -q
  echo 'fn NEEDLE() {}' > .claude/worktrees/wt1/app.rs
  echo '.claude/worktrees/' > .gitignore
```

rtk grep NEEDLE .claude/worktrees

```
pre-PR : 1 matches in 1 files  (exit 0)
this PR: 0 matches for 'NEEDLE' (exit 1)   ← silent false negative
```

This is the exact failure mode build_rg_command's own --no-ignore-vcs comment exists to prevent ("false negatives that make AI agents draw wrong conclusions"). An agent deliberately inspecting a worktree gets an empty result with no indication the path was excluded.

Scope (tested, for context): only explicit directory paths regress — normal scans (rtk grep X .), explicit file paths, --type, and running from inside a worktree are all unaffected.

### Secondary regression (grep fallback only)

--exclude-dir=worktrees is base-name-only, so when rg is absent, any directory named worktrees is dropped — a legit src/worktrees/ silently disappears:

```
with rg unavailable on PATH, grep fallback, normal scan: src/worktrees/legit.rs   pre-PR: matched  →  this PR: dropped
```

nanookclaw:
Addressed in `b0e2b59`.

Changes made:
- `rtk grep NEEDLE .claude/worktrees` no longer applies the default `.claude/worktrees` exclusion when the search path is explicitly inside that subtree.
- The grep fallback no longer uses `--exclude-dir=worktrees`; default-scan fallback output is filtered by path instead, so unrelated paths like `src/worktrees` remain searchable.
- Added focused tests for explicit worktree paths, fallback filtering, and the basename-exclusion regression.

Validation run locally:
- `cargo fmt --check`
- `cargo test cmds::system::grep_cmd::tests -- --nocapture`
- Manual fallback repro with `rg` absent: default scan skipped `.claude/worktrees` but kept `src/worktrees`; explicit `.claude/worktrees` scan returned the match.

nanookclaw:
Rebased this onto current `develop` and resolved the `src/cmds/system/grep_cmd.rs` conflict while preserving the newer multi-pattern/multi-path grep behavior.

New head: `d9d92bd`
Verification: `cargo test cmds::system::grep_cmd` passes locally (85 tests).

nanookclaw:
Rebased onto current `develop` and force-pushed the refreshed branch at `ba135c3`.

Local verification:
- `cargo fmt --check`
- `cargo test cmds::system::grep_cmd` (92 passed)

The conflict was in `src/cmds/system/grep_cmd.rs`; I kept the newer upstream grep flow and reapplied the `.claude/worktrees` skip there.

KuSh:

PR #2153, opened 2026-05-29
fix(hook): support OpenCode Desktop (Node.js) in plugin

### LINKED ISSUE #2039: OpenCode Desktop does not detect rtk when launched from macOS GUI
## Problem

RTK works when OpenCode is started from a terminal, but it does not work when using [OpenCode Desktop](https://opencode.ai/download) launched normally.

This suggests the [OpenCode Desktop](https://opencode.ai/download) runtime may not be loading the same environment as the user’s shell session. As a result, the RTK OpenCode plugin may fail to detect or execute the `rtk` binary even though RTK is installed and available in the terminal.

## Expected Behavior

OpenCode Desktop should load the RTK plugin.

## Actual Behavior

- Starting OpenCode from terminal: RTK works.
- Starting OpenCode Desktop normally: RTK does not appear to run.

## Possible Cause

The Desktop app may be running with a different environment than the user’s login shell, especially a different `PATH`. The RTK plugin currently depends on resolving `rtk` from the runtime environment, so it may disable itself if the binary is not discoverable there.

## Reproduction

1. Install and configure RTK for OpenCode.
2. Confirm RTK works when OpenCode is launched from terminal.
3. Launch OpenCode Desktop normally.
4. Run a command that should be rewritten by RTK.
5. Observe that RTK is not used.

synnks:
## Summary

- Fix OpenCode plugin failing silently in the Desktop app (Electron/Node.js)
- Add `node:child_process` fallback when `Bun.$` is unavailable

Fixes #2039

## Root Cause

OpenCode Desktop runs its server in an Electron utility process (Node.js), not Bun.
OpenCode passes `$: undefined` to plugins when `typeof Bun === "undefined"`.
The plugin calls `await $`which rtk`.quiet()` — since `$` is `undefined`, this throws a TypeError, caught by the `catch` block with the misleading message "rtk binary not found in PATH".

## Test plan

- [x] `node --experimental-strip-types --test hooks/opencode/tests/rtk.test.ts`
- [x] Manual testing: verified in OpenCode TUI (Bun path) and Desktop (child_process path)

### DIFF
--- a/hooks/opencode/rtk.ts
+++ b/hooks/opencode/rtk.ts
@@ -1,4 +1,5 @@
 import type { Plugin } from "@opencode-ai/plugin"
+import { execFile, execFileSync } from "node:child_process"
 
 // RTK OpenCode plugin — rewrites commands to use rtk for token savings.
 // Requires: rtk >= 0.23.0 in PATH.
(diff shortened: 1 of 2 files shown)


alexandre-leites:
Any workaround while this is not merged?

aurelienpolletadeo:
I am using this version of the rtk.ts for opencode desktop and it's working fine.
Waiting for merge.
Thanks for the PR!

smzh369:
`which` may failed to get the path on Windows, better to make a platform judgement(line 29-33):
```
    const check = process.platform === "win32" ? "where" : "which"
    if ($) {
      await $`${check} rtk`.quiet()
    } else {
      execFileSync(check, ["rtk"], { encoding: "utf8", timeout: 5000 })
    }
```

amnesiaof:
Same root causes reproduce on Windows CLI (not only Desktop Electron): no `$` in plugin context + Unix-only `which`. Also note `rtk rewrite` exit-code-3 semantics — stdout must be captured from the exception. #3330's cross-platform process API fixes the CLI case too; verified with a working workaround.

PR #2155, opened 2026-05-29
fix: non-ASCII / UTF-8 robustness (git filenames, gain truncation, proxy capture)

snwsnwsnw:
Three small, independent fixes for handling non-ASCII (CJK and other multibyte UTF-8) text. Each is in its own commit.

## 1. `git`: non-ASCII filenames shown as octal escapes
Git escapes non-ASCII path bytes as octal `\nnn` by default (`core.quotepath=true`), and rtk passed that through unchanged.

```
# before
$ rtk git status
?? "\345\214\205\350\243\271\347\260\275\346\224\266\346\270\205\345\226\256.txt"
# after
$ rtk git status
?? 包裹簽收清單.txt
```

Affects `git status`, `git log --name-only`, `git diff --stat`, etc. Fix injects `-c core.quotepath=false` at the single `git_cmd()` chokepoint. No effect on ASCII paths.

## 2. `gain`: byte-index slicing can panic on multibyte input
`gain --history` and the failure summary truncated command strings with `&cmd[..47]` / `&rec.rtk_cmd[..22]` / `&rec.raw_command[..37]`. A command containing multibyte UTF-8 (a non-ASCII search pattern, a non-ASCII commit message, etc.) can land the cut mid-codepoint and panic. Switched to the existing char-safe `utils::truncate()`.

## 3. `proxy`: spurious replacement char at the 1 MiB capture cap
The proxy streaming path caps the captured copy at 1 MiB. When the cap lands inside a multibyte sequence, `from_utf8_lossy` appended a trailing `U+FFFD` to the tracked output. `decode_captured()` trims an incomplete trailing sequence while preserving lossy behavior for genuinely invalid mid-stream bytes. (User-facing stdout was already byte-exact; this only affected the captured copy used for tracking.)

## Testing
`cargo build --release` clean. Verified before/after on a repo with CJK filenames and commit messages for #1; `gain --history` no longer panics for #2.

### DIFF
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -246,11 +246,9 @@ pub fn run(
                 println!("──────────────────────────────────────────────────────────");
                 for rec in recent {
                     let time = rec.timestamp.with_timezone(&Local).format("%m-%d %H:%M");
-                    let cmd_short = if rec.rtk_cmd.len() > 25 {
-                        format!("{}...", &rec.rtk_cmd[..22])
-                    } else {
-                        rec.rtk_cmd.clone()
-                    };
+                    // char-safe truncation: commands may contain multibyte UTF-8,
+                    // and slicing on a byte boundary would panic
+                    let cmd_short = crate::core::utils::truncate(&rec.rtk_cmd, 25);
                     // added: tier indicators by savings level
                     let sign = if rec.savings_pct >= 70.0 {
                         "▲"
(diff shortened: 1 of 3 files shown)


KuSh:
Thanks for these three fixes. The gain truncation one landed as #2789 on 2026-07-03, which moved the display strings onto the char safe truncate. The proxy capture fix no longer has a target: that path moved into src/core/stream.rs with a 10 MiB cap and decoding now goes through a codepage aware decoder in src/core/utils.rs, so there is no 1 MiB from_utf8_lossy left to trim. For the filenames, develop took the other approach and decodes git's quoting on the read side (unquote_path in src/cmds/git/git.rs, added with the diff rewrite in #3268 on 2026-09-08) instead of changing git's own output, because the diff and show paths rely on passing git's bytes through unchanged. That decoding is only wired into diff headers today, so `rtk git status` and `rtk git log --name-only` still print octal escaped paths, which I confirmed on develop. Extending unquote_path to those two outputs would be a welcome narrow PR or issue. Please comment or reopen if a part of this was missed.

PR #2156, opened 2026-05-29
docs(hooks): add "reading the full output" to agent instructions

tiagodocouto:
## Summary

- Adds a **"Reading the full output"** section to every agent instruction template (`hooks/*/rtk-awareness.md` and `rules.md`). The templates tell agents to run commands through `rtk`, but not how to recover the complete, *unfiltered* output when the filtered summary isn't enough — e.g. to confirm a "tests pass" / "build clean" claim rather than skim it. Both raw paths already exist (`rtk proxy <cmd>` and the `[full output: <path>]` tee log) but weren't documented where agents actually read. Mirrors the principle already in `CONTRIBUTING.md` (respect explicitly-requested detail).
- The Claude template also replaces a dangling `Refer to CLAUDE.md` line with `rtk --help`.
- Docs-only; the section is identical across all 7 templates.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — fmt clean, test **1974 passed, 0 failed**. (clippy hits a pre-existing `uninlined_format_args` lint in `build.rs`, untouched by this docs-only diff; CI is green on `develop`.)
- [x] Manual testing: inspected `rtk cargo test` (filtered) vs `rtk proxy cargo test` (raw) vs the `[full output: <path>]` tee log against a failing and a non-compiling crate — the documented behavior matches.

### DIFF
--- a/hooks/antigravity/rules.md
+++ b/hooks/antigravity/rules.md
@@ -30,3 +30,15 @@ rtk proxy <cmd>       # Run raw (no filtering, for debugging)
 ## Why
 
 RTK filters and compresses command output before it reaches the LLM context, saving 60-90% tokens on common operations. Always use `rtk <cmd>` instead of raw commands.
+
+## Reading the full output
+
+RTK output is a filtered summary — for most commands it keeps what matters
+(test failures, panics, and build errors are preserved). When you need the
+complete, unfiltered output to confirm a claim (tests really passed, build is
+really clean) rather than skim it, get the raw output instead of trusting the
+summary:
+
+- Run `rtk proxy <cmd>` to re-run the command with no filtering.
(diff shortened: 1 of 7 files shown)


KuSh:
Thanks for writing this up. The per-agent instruction templates were consolidated into a single agent-neutral hooks/rtk-awareness.md in #3372, merged 2026-09-07, and that file already tells the agent that truncated results state their own recovery path and to re-run a command as `rtk proxy <cmd>` when the filtered result is unusable; the full variant lists `rtk proxy <cmd>` in its command reference too. Six of the seven files this PR edits no longer exist on develop, so I am closing it as covered. If you think a point from your section is still missing from the consolidated text, please comment or reopen.

PR #2160, opened 2026-05-29
fix(find): fall back to raw find for unsupported flags

mariuszs:
## Problem

`rtk find` **errors out** when given native `find` features it can't model in its structured walker — compound predicates (`-not`, `-or`, `-and`), actions (`-exec`, `-execdir`, `-delete`, `-print0`), and time/size/perm tests (`-size`, `-mtime`, `-perm`, …):

```
$ rtk find . -name '*.tmp' -exec rm {} \;
Error: rtk find does not support compound predicates or actions (e.g. -not, -exec). Use `find` directly.
```

This forces users to reach for `rtk proxy find ... | rtk proxy xargs ...` to get any result, and it contradicts RTK's fallback principle: *if a filter can't handle the input, run the real command unchanged rather than blocking the user.*

## Fix

`run_from_args` now detects unsupported flags up front and falls back to the **real `find`** via `core::runner::run_passthrough`:

- streams output unchanged (inherits stdin),
- tracks the invocation at 0% savings (visible in `rtk gain --history`, like `rtk proxy`),
- returns the child exit code, propagated through the single `std::process::exit` in `main()` like every other command (`run_from_args` now returns `Result<i32>`).

Supported syntax (`-name`, `-iname`, `-type`, `-maxdepth`, RTK shorthand) is unaffected — it still goes through the compact filter.

## Tests

- `unsupported_flags_detected` / `supported_flags_not_flagged_as_unsupported` — flag classification
- `run_from_args_falls_back_for_unsupported_flags` — `-not` now runs the real `find` and returns `Ok(0)` instead of erroring
- Removed the two `parse_*_rejects_*` tests that asserted the old blocking behavior

Gate (`cargo fmt --all --check && cargo clippy --all-targets && cargo test --all`) is green on `develop`: 1983 passed, 7 ignored, 0 clippy warnings.

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -62,8 +62,10 @@ fn has_native_find_flags(args: &[String]) -> bool {
         .any(|a| a == "-name" || a == "-type" || a == "-maxdepth" || a == "-iname")
 }
 
-/// Native find flags that RTK cannot handle correctly.
-/// These involve compound predicates, actions, or semantics we don't support.
+/// Native find features RTK's structured walker can't model faithfully:
+/// compound predicates (-not/-or/-and), actions (-exec/-delete/-print0), and
+/// time/size/perm tests. When any appear, `run_from_args` falls back to the
+/// real `find` (passthrough) instead of blocking the user.
 const UNSUPPORTED_FIND_FLAGS: &[&str] = &[
     "-not", "!", "-or", "-o", "-and", "-a", "-exec", "-execdir", "-delete", "-print0", "-newer",
     "-perm", "-size", "-mtime", "-mmin", "-atime", "-amin", "-ctime", "-cmin", "-empty", "-link",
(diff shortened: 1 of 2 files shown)


mariuszs:
Superseded by #3603, shipped in v0.46.0.

`rtk find` now dispatches on find's own grammar: actions (`-exec`, `-execdir`, `-delete`, `-print0`, `-ok`, …) run through `runner::run_passthrough` with the exit code kept, and every other predicate the walker can't model (`-not`, `-o`, `-path`, `-mtime`, `-size`, `-type l`, …) runs the real `find` under the compressed renderer with a never-worse guard. That is a strict superset of the fallback proposed here.

Every hunk of this PR targets code that no longer exists on `develop` (`has_unsupported_find_flags`, `UNSUPPORTED_FIND_FLAGS`, the `bail!` in `parse_find_args`), which is why it shows as conflicting. A rebase would leave an empty diff.

Verified on 0.46.0:

```
$ rtk find src -name 'find_cmd.rs' -not -name 'zzz'
src/cmds/system/find_cmd.rs
# exit 0

$ rtk find src/cmds/system -name 'find_cmd.rs' -exec wc -l {} +
1177 src/cmds/system/find_cmd.rs
# exit 0
```

Closing.

PR #2162, opened 2026-05-30
fix(grep): treat bare `|` as literal pipe (BRE), not rg alternation

ryanmurf:
## Problem

`rtk grep` accepts **grep BRE** syntax, where `\|` is alternation and a **bare `|` is a literal pipe**. The pattern is then handed to **ripgrep**, whose default (Rust-regex) engine uses the *opposite* convention: `|` is alternation and `\|` is a literal pipe.

The previous code translated only `\|` → `|`:

```rust
let rg_pattern = pattern.replace(r"\|", "|");
```

That fixed the alternation case but left a **bare `|` to be reinterpreted by rg as alternation**. So a user searching for the literal substring "a|b":

```
rtk grep 'a|b'   # wanted: lines containing the literal a


ryanmurf:
Closing — intended for the murphytek fork, not upstream. Reopening against murphytek/rtk:master.

PR #2168, opened 2026-05-30
fix(grep): `rtk grep -v` inverts match instead of bumping verbose (closes #1477)

### LINKED ISSUE #1477: rtk grep "-v" does not invert the search

tiagodocouto:
## What

`rtk grep -v foo file` returned **matching** lines instead of inverting them (#1477). `verbose` is a global flag with short `-v`, so clap consumed `-v` before the `grep` subcommand could see it. Only `--invert-match` worked.

## Why not just a local `-v`

Adding `#[arg(short = 'v', long = "invert-match")]` to the `Grep` subcommand panics at startup — clap requires unique short names (verified against clap 4.6):

```
Command grep: Short option names must be unique for each argument,
but '-v' is in use by both 'invert_match' and 'verbose'
```

## Change (two parts)

1. **Native `--invert-match` flag** (long-only) on the grep subcommand, forwarded to ripgrep. clap parses it in any position. (A bare `--invert-match` *before* the positionals would otherwise be an `UnknownArgument`, since `extra_args` is `trailing_var_arg`.)
2. **argv rewrite** for the `-v` shorthand: a standalone `-v` token after the `grep` subcommand becomes `--invert-match` before parsing. Scoped carefully:
   - `-v` **before** `grep` (`rtk -v grep ...`) stays verbose.
   - a `--` separator stops the rewrite, so `rtk grep -- -v` searches for the literal `-v` (GNU grep / getopt convention).
   - `--verbose` remains available for a verbose grep.

## Verification

- `cargo build`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check` — all clean
- `cargo test` — full suite green; **8 new unit tests** for the rewrite + parse
- smoke (real binary, rtk filter path — not the raw-grep fallback): `rtk grep -v apple file` → inverted + filtered; `rtk grep apple file -v` → inverted; `rtk grep apple file` → unchanged; `rtk grep -- -v file` → literal search

Closes #1477

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -322,6 +322,12 @@ enum Commands {
         /// Show line numbers (always on, accepted for grep/rg compatibility)
         #[arg(short = 'n', long)]
         line_numbers: bool,
+        /// Invert match — show non-matching lines (ripgrep --invert-match).
+        /// Long-only: the `-v` short form collides with the global `--verbose`
+        /// flag, so `rtk grep -v` is mapped to this via an argv rewrite before
+        /// parsing (see `rewrite_grep_invert`).
+        #[arg(long = "invert-match")]
+        invert_match: bool,
         /// Extra ripgrep arguments (e.g., -i, -A 3, -w, --glob)
         #[arg(trailing_var_arg = true, allow_hyphen_values = true)]
         extra_args: Vec<String>,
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into this one. The same problem was fixed by #2333, merged on 2026-06-14, which made every rtk-owned grep option long-only so -v and the rest of the tail are forwarded straight to the search engine. Checking the current develop build, rtk grep -v PAT file inverts, rtk grep -iv works, rtk grep -- -v searches for the literal -v, and rtk -v grep still means verbose, so the cases this PR covered are all handled. Closing as already resolved, but please comment or reopen if you spot a case that still slips through.

Issue #1833, opened 2026-05-11
[grep] reduce fallbacks for common flag combinations (-rn, -l, -A/-B/-C, -E, -o, -c)


TimurZanagar:
## Observation

  On a 30-day Claude Code history (~13.5k tracked commands), `rtk grep` filters
  99 % of token volume in the calls it accepts (22.5M saved / 22.8M input).
  However, **2327 additional grep calls fall through `rtk fallback: grep …`**
  where no filter runs and RTK cannot even measure the bypassed token volume.

  ## Top fallback triggers

  | Flag         | Calls | Why it should be handled                                                                          |
  |--------------|-------|---------------------------------------------------------------------------------------------------|
  | `-rn`        | 748   | Already implicitly supported (rg is recursive + line-numbered); just parse the combined flag      |
  | `-r`         | 563   | Same as above                                                                                     |
  | `-l`         | 248   | `rg --files-with-matches`                                                                         |
  | `-A/-B/-C N` | 212   | Already exposed as `--context-only` but with different semantics; map directly to rg's `-A/-B/-C` |
  | `-n`         | 198   | Line numbers are always on — accept the flag as a no-op                                           |
  | `-E`         | 140   | rg uses regex by default — accept as a no-op                                                      |
  | `-o`         | 79    | `rg -o`                                                                                           |
  | `-i/-in`     | 86    | `rg -i`                                                                                           |
  | `-c`         | 27    | `rg -c`                                                                                           |

  Note: `-r`/`-rn`/`-n` without `-E` must be mapped to `rg -F` to preserve
  GNU-grep's literal-string semantics (otherwise patterns containing `.`, `*`,
  etc. would be silently re-interpreted as regex).

  ## Suggested approach

  Extend the `rtk grep` argument parser to recognise these common GNU-grep
  flags and forward them to the ripgrep backend rather than dropping to the
  unfiltered fallback. For flags that are no-ops under ripgrep (`-n`, `-E`),
  silently accept them. The `-F` injection rule above should kick in
  automatically when `-E` is absent.

saltforged:
Wanted to comment to confirm that this is not resolved in 0.40.0 and remains open. I've also seen a massive jump in "tokens saved" that are not actually saved, but instead are falling back to grep. This issue started occurring around 11th of May (GMT+10).
Further, there is another issue to address as part of this fix:
[...]

PR #2174, opened 2026-05-30
fix(grep): exclude token-bomb dirs from recursive search (#2064)

### LINKED ISSUE #2064: grep -rn fallback causes token *expansion*: execs system grep, descends node_modules, dumps minified files
## Summary
On 0.41.0, `rtk grep -rn <pat> <dir>` falls back to raw system grep (shown as
`rtk fallback: grep -r…` in `rtk gain`). The fallback execs `/usr/bin/grep`
(BSD), which is **not** .gitignore-aware: it descends into `node_modules/` and
prints full minified `.js`/`.map` lines (single lines up to ~838 KB). For a
tool whose purpose is token *reduction*, the result is large token *expansion*
— and the auto-rewrite hook applies it silently.

## Repro
Any project with a `node_modules/` under the searched path (e.g. a frontend):

    rtk grep -rn 'def ' src    # 3,384,276 chars, 255 matches inside node_modules
    grep -rn 'def ' src        # gitignore-aware grep: 58,863 chars, 0 in node_modules

rtk output is **byte-for-byte identical to `/usr/bin/grep -rn`**, confirming the
fallback shells out to system BSD grep — bypassing the user's PATH grep
(ripgrep/ugrep) and all ignore rules. ~57× larger than the grep the user runs.

## Why it matters
- rtk is a token *reducer*; here it emits ~57× more than the gitignore-aware grep.
- The auto-rewrite hook makes this automatic and invisible: a single `grep -rn`
  in a node_modules repo can push megabytes of minified vendored code into context.
- Related to #1833 (reduce -rn fallbacks) and #1219 (route -rn through filter),
  but those frame the fallback as a coverage/measurement gap. This is the
  severity side: the fallback is not token-neutral — it is unbounded expansion.

## Environment
- rtk 0.41.0 (Homebrew, arm64 sequoia)
- macOS / Darwin 24.6
- system grep (what the fallback execs): BSD grep 2.6.0-FreeBSD
- user PATH grep: ugrep 7.5.0

## Suggested fixes (any one helps)
1. Route `-r`/`-rn`/`-R` through rtk's own ripgrep-backed filter (gitignore-aware)
   instead of falling back — overlaps #1219 / #1833.
2. If fallback is unavoidable, exec the user's PATH grep (ripgrep/ugrep) rather
   than hard-coded system grep, so ignore rules apply.
3. Guard fallback output: skip binary + minified files, or apply a byte cap and
   warn, rather than emitting megabytes.

maxmilian:
## Summary
- Fixes #2064: `rtk grep -rn <pat> <dir>` descends `node_modules` and dumps minified vendored code, turning a token *reducer* into a token *expander* (~57× the gitignore-aware grep in the report).
- Re-excludes only well-known token-bomb dirs/files, while keeping `--no-ignore-vcs` so genuinely-searched .gitignore'd files (configs, etc.) are still found.

> Target branch is `develop` per CONTRIBUTING.

## Root cause
`src/cmds/system/grep_cmd.rs` passes `--no-ignore-vcs` to rg (intentional, to avoid false negatives for .gitignore'd files — comment at the call site). The side effect: rg descends dependency/build dirs like `node_modules`, emitting megabytes of minified single-line bundles. When rg is absent, the fallback execs system `grep -rnHZ` (BSD grep on macOS), which is not gitignore-aware and descends `node_modules` too.

## Fix
Wire the **existing but previously-unused** `config.filters.ignore_dirs` / `ignore_files` list (`node_modules`, `target`, `.venv`, `vendor`, `*.min.js`, `*.lock`, …) into:
- the rg invocation as `--glob '!<pat>'`, and
- the system-grep fallback as `--exclude-dir=<dir>` / `--exclude=<file>`.

## Correctness (no false negatives)
- rg honors an explicitly-given path inside an excluded dir, so `rtk grep foo node_modules/pkg` works.
- `grep --exclude-dir` does **not** honor that (verified on BSD grep). To keep both paths consistent and never drop a result the user asked for on purpose, **exclusions are skipped entirely when the search path itself targets an ignored dir** (`path_targets_ignored_dir`). The check is path-component-anchored, so `my-node_modules-helper` is unaffected.

## Tests
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — all green (1986 passed).
- New unit tests (red→green): `test_ignore_globs_exclude_node_modules`, `test_ignore_globs_empty_config_yields_nothing`, `test_ignore_grep_excludes_for_fallback`, `test_path_targets_ignored_dir`.
- Manual before/after on a frontend-style repo with `node_modules`:
  - `rtk grep define .` → **10,517 chars → 104 chars** (node_modules no longer dumped).
  - `rtk grep define node_modules/pkg-a` → still returns the matches (explicit path honored).

_Token-savings ≥60% checklist item is N/A: this is a correctness/expansion fix on the existing grep filter, not a new compression filter — and it strictly reduces output. Cites design principle #1 (Correctness over Token Savings)._

## Scope
- Intentionally keeps `--no-ignore-vcs` (false-negative-avoidance preserved for non-token-bomb gitignore'd files).
- Reuses the existing `FilterConfig` ignore list rather than inventing a new hard-coded set, so users can already customize it via config.
- No shell-string interpolation; all args go through `Command::arg`.

## Follow-up
- Related #1219 / #1833 frame the `-rn` fallback as a coverage/measurement gap; this PR addresses the severity (expansion) side and is independent of those.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -312,6 +312,20 @@ pub fn run(
         eprintln!("grep: '{}' in {}", pattern_display, path_display);
     }
 
+    let filt = config::filters();
+    // Skip exclusions when the user explicitly searches inside an ignored dir
+    // (e.g. `rtk grep foo node_modules/pkg`) — they asked for it on purpose.
+    // With multiple search paths, any explicit ignored-dir target disables the
+    // exclusions so a deliberately-named path never silently returns zero hits.
+    let (exclude_globs, grep_excludes) = if paths
+        .iter()
+        .any(|p| path_targets_ignored_dir(p, &filt.ignore_dirs))
+    {
+        (Vec::new(), Vec::new())
+    } else {
+        (ignore_globs(&filt), ignore_grep_excludes(&filt))
+    };
+
     let mut rg_cmd = resolved_command("rg");
     // --no-ignore-vcs: match grep -r behavior (don't skip .gitignore'd files).
     // Without this, rg returns 0 matches for files in .gitignore, causing
@@ -321,6 +335,9 @@ pub fn run(
     // -0: NUL-separate filename. Allows the parser to disambiguate filenames or
     // content containing `:digits:` patterns (issue #1436).
     rg_cmd.args(["-nH0", "--no-heading", "--no-ignore-vcs"]);
+    // Re-exclude token-bomb dirs/files (node_modules, *.min.js, …) that
+    // --no-ignore-vcs would otherwise descend, causing token expansion (#2064).
+    rg_cmd.args(&exclude_globs);
 
     if let Some(ft) = file_type {
         rg_cmd.arg("--type").arg(ft);
@@ -349,6 +366,9 @@ pub fn run(
             for p in &patterns {
                 grep_cmd.args(["-e", p]);
             }
+            // Mirror the rg exclusions so the (non-gitignore-aware) system grep
+            // also skips node_modules and friends (#2064).
+            grep_cmd.args(&grep_excludes);
             // --null (not -Z): on BSD/macOS grep -Z means --decompress, not the
             // NUL filename separator parse_match_line() needs (issue #2310).
             grep_cmd.args(["-rnH", "--null", "--"]);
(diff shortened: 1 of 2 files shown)


maxmilian:
Closing this — `grep_cmd.rs` was substantially reworked by #2550 (merged 2026-06-23) and there are now several open grep PRs touching the same file, so a rebase here is high-cost and low-odds. Stepping back to keep the area uncluttered.

One triage note so the report doesn't get lost, though: **#2064 (token *expansion*) is still not solved on `develop` after #2550.** The fast path still runs ripgrep with `--no-ignore-vcs` (grep_cmd.rs L485/498/541), so it keeps descending into `node_modules`/vendored dirs. #2550's "never-worse" guard is `if rtk_output.len() < plain.len() { rtk_output } else { plain }` (L617-621) — but **both sides derive from the same already-bloated `raw_output`**, so it only guarantees rtk isn't *worse* than plain grep; plain grep itself is already the 57× blow-up #2064 describes. The root cause is the `--no-ignore-vcs` descent, not the grouping/flag handling #2550 addressed. Whoever picks #2064 up next will want to gate the ignored-dir descent (or post-filter those paths) rather than rely on the min() guard. Thanks!

PR #1048, opened 2026-04-06
fix: prevent crashes when output is piped

ashwingopalsamy:
## Summary

- Fix **SIGABRT** crash when rtk output is piped (e.g. `rtk git log | head`) by resetting **SIGPIPE** to the default handler at process startup (#1004)
- Rust ignores **SIGPIPE** by default; combined with `panic = "abort"` in the release profile, any write to a closed pipe triggers **SIGABRT** + coredump instead of silent exit
- Follows the conventions already used in this repo behind `cfg(unix)` focusing on Unix based systems, no impact on Windows platform.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [x] Manual testing: `rtk git log --oneline -50 | head -1` exits cleanly (no SIGABRT, no exit 134)
- [x] Manual testing: `rtk ls | head -1` exits cleanly
- [x] Integration test: `cargo test --ignored test_broken_pipe_does_not_crash`

> **Important:** All PRs must target the `develop` branch (not `master`).
> See [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) for details.

Issue #1004, opened 2026-04-03
panic/SIGABRT on Broken pipe while writing stdout


toniles:
## Summary
`rtk` aborts with a Rust panic when stdout is closed (`Broken pipe`), producing `SIGABRT` + coredump instead of exiting gracefully.

## Environment
- OS: Arch Linux (x86_64)
- `rtk` version: `0.31.0`
- Binary: `/home/s3s/.local/bin/rtk`
- Build ID: `cc06a1448d8bbdd786613878d0ec5dd3537f3ce9`

## What happened
On **2026-04-03 (CEST)**, multiple `rtk` invocations crashed (`SIGABRT`) and generated core dumps.

Observed commands at crash time:
- `rtk ls -la /mnt/datos/2brain-backend/scripts/deploy/`
- `rtk git log --oneline --all --decorate -50`
- `rtk git diff ae78d87^..ae78d87 --name-only` (repeated)

`systemd` services were healthy (`systemctl --failed` and `systemctl --user --failed` both empty), so this is user-space process crash behavior in `rtk`.

## Evidence
`coredumpctl list rtk` showed 5 crashes (all `SIGABRT`) on 2026-04-03.

From core dump strings (same pattern across all affected PIDs):

```text
thread 'main' (...) panicked at /rustc/.../library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
```

I validated this message in multiple dumps (e.g. PIDs `30441`, `42021`, `490097`, `491982`, `492096`).

## Expected behavior
If stdout is closed by downstream consumer (`EPIPE`), `rtk` should terminate cleanly (or return non-zero) without panic/abort/coredump.

## Actual behavior
`rtk` panics in `std::io::stdio` on write to stdout, then aborts and dumps core.

## Notes
This looks like unhandled `Broken pipe` on output path. Handling `EPIPE` explicitly (or ignoring SIGPIPE semantics appropriately for CLI output writes) should avoid process aborts and coredumps in normal shell workflows.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 97% |

### Summary

[...]

PR #2181, opened 2026-05-31
fix(curl): passthrough binary downloads to prevent UTF-8 corruption (#1087)

### LINKED ISSUE #1087: `rtk curl http://....tar.gz | tar -xzf` fails due rtk truncation, eg "(153 more lines, 94292 bytes total)"
Should there be a skill pairing with rtk for this case?

Eg, when piping a command to something like gzip (not writing to terminal) then don't try to shorten the input?

pszymkowiak:
## Summary

`rtk curl <binary-url>` silently corrupted binary downloads (#1087). A
gzip tarball downloaded with `rtk curl ... | tar -xzf -` failed with
`gzip: stdin: not in gzip format`.

## Root cause

`src/core/stream.rs:538` captures stdout via:
```rust
stdout: String::from_utf8_lossy(&output.stdout).into_owned()
```
For binary content, every non-UTF-8 byte is replaced with U+FFFD
(`0xEF 0xBF 0xBD`, 3 bytes). Gzip magic `1f 8b 08 00 ...` arrives as
`1f ef bf bd 08 00 ...`. Same problem for zip, PNG, PDF, ELF, etc.

This was masked for a long time by the SIGPIPE panic (#1004) — once
that was fixed by #1048 (merged 2026-05-29), the underlying corruption
became visible to users.

## Fix

Local to `src/cmds/cloud/curl_cmd.rs` (does not touch the global
`exec_capture` API — that would impact 141 call sites across 6
modules):

1. Replace `exec_capture(&mut cmd)` with `cmd.output()` directly to
   keep stdout as `Vec<u8>`.
2. New `is_binary(&[u8]) -> bool` helper — one line:
   ```rust
   std::str::from_utf8(bytes).is_err()
   ```
   Correct by construction: the only condition under which
   `from_utf8_lossy` would corrupt the stream is when the bytes are
   not valid UTF-8. Catches every binary format without needing a
   magic-byte database to maintain.
3. On binary detection: `stdout.write_all(&output.stdout)` raw +
   `timer.track_passthrough(...)` (0% savings).
4. Text/JSON/HTML path unchanged — same `from_utf8_lossy` +
   `filter_curl_output` flow as before.

## Verification

Live tests against 18 real binary formats from public CDNs — **all
byte-identical to raw curl**:

| Category | Samples | Result |
|---|---|---|
| Archives | rtk .tar.gz/.zip/.deb/.rpm, ripgrep, bat, fd | ✅ 8/8 |
| Executables | kubectl ELF 51 MB, rustup-init 20 MB, gh CLI, hyperfine, tokei | ✅ 5/5 |
| Images | GitHub avatar PNG, Wikimedia GIF, WebP, ICO, octocat PNG | ✅ 5/5 |
| Documents | W3C PDF, ISO 9899 PDF 4 MB | ✅ 2/2 |
| Audio/Video | MP3, MP4, WAV | ✅ 3/3 |
| **Text regression** | JSON, HTML, Markdown, Cargo.toml, plain text | ✅ no path-text behavior change |

The 18 MB and 51 MB ELF binaries roundtrip byte-perfect, confirming the
fix scales.

## Tests

4 new unit tests, all green:
- `test_is_binary_gzip_magic_is_not_utf8`
- `test_is_binary_valid_utf8_text_is_not_binary` (JSON, HTML, ASCII, UTF-8 emoji)
- `test_is_binary_empty_is_not_binary`
- `test_is_binary_text_with_nul_is_not_binary` (NUL is valid UTF-8)

`cargo fmt --all`, `cargo clippy --all-targets -- -D warnings`,
`cargo test --all` — 1986 passed, 0 failed.

## Out of scope

The existing text path applies `.trim()` + `println!("\n")` to text
responses in pipe mode, so a byte-strict downstream parser may see
±1 byte of trailing whitespace. That's pre-existing behavior, not
introduced by this PR, and worth a separate issue if the agent
ecosystem needs byte-faithful text passthrough.

Closes #1087.

### DIFF
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -5,13 +5,18 @@
 //! The condensed-form-with-tee-hint path is reserved for non-JSON bodies on
 //! a real terminal where a human reads the output and the tee file gives the
 //! LLM a way to recover the raw response.
+//!
+//! Binary downloads (any non-UTF-8 byte sequence) are written through to
+//! stdout as raw bytes, bypassing the UTF-8 lossy conversion that would
+//! otherwise replace non-UTF-8 bytes with U+FFFD and corrupt the stream
+//! (`#1087`).
 
 use crate::core::tee::force_tee_hint;
 use crate::core::tracking;
-use crate::core::{stream::exec_capture, utils::resolved_command};
+use crate::core::utils::resolved_command;
 use anyhow::{Context, Result};
 use std::borrow::Cow;
-use std::io::IsTerminal;
+use std::io::{IsTerminal, Write};
 
 const MAX_RESPONSE_SIZE: usize = 500;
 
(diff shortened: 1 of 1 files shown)


KuSh (approved):
LGTM!

PR #2182, opened 2026-05-31
Add compact dotnet ef filtering

cuong21951:


cuong21951:
## What

Add compact Entity Framework Core CLI support to RTK via `rtk dotnet ef ...`.

Implemented command coverage:

- `rtk dotnet ef migrations add`
- `rtk dotnet ef migrations remove`
- `rtk dotnet ef migrations list`
- `rtk dotnet ef migrations script`
- `rtk dotnet ef database update`
- `rtk dotnet ef dbcontext list`
- `rtk dotnet ef dbcontext info`
- generic compact fallback for other `dotnet ef` output

The route is implemented in the Rust dotnet command module because top-level `dotnet` is already Clap-routed, so a TOML fallback filter would not reliably see `dotnet ef` invocations.

## Why

Raw `dotnet ef` output is verbose in LLM workflows, especially for migrations and SQL scripts.

EF output mixes useful signal with repeated project/build/design-time discovery lines. For migration scripts, stdout can contain large SQL blocks where the model usually needs a compact summary plus access to full output, not every SQL line inline.

## How The Filter Works

The filter detects the EF command shape from the forwarded args, skipping global options like `--project`, `--startup-project`, `--context`, `--framework`, and `--configuration` before identifying the command group/action.

Then it removes stable EF noise such as:

- `Build started...` / `Build succeeded.`
- project/startup project paths
- design-time service discovery lines
- repeated EF Core `info:` / `dbug:` / `trce:` diagnostics

It keeps high-signal output:

- migration name and generated file count
- DbContext name when present
- applied/reverted migration names
- SQL statement count and first SQL lines for script output
- warnings
- errors, exceptions, failed context creation, and actionable diagnostics

For long SQL or long lists, the filter emits a tee hint so the full output is still available without spending tokens inline.

## Benefits

- Significant token reduction for noisy EF workflows.
- Cleaner summaries for migrations, database updates, dbcontext inspection, and SQL script generation.
- Warning/error preservation instead of hiding diagnostics.
- Safe argument forwarding: user args are passed through to `dotnet ef` unchanged.
- Stable English output for spawned dotnet commands via `DOTNET_CLI_UI_LANGUAGE=en-US`, matching the existing dotnet module behavior.

## Real-world impact: complex EF migration script

Smoke test used a real EF project under `target` and generated a migration script containing:

- `UPDATE Accounts`
- `INSERT INTO AuditLogs`
- `CREATE TRIGGER TRG_LedgerEntries_Audit_Insert`
- `CREATE VIEW VW_AccountBalances`

Command compared:

```text
Raw:      dotnet ef migrations script --no-build --context SmokeContext
Filtered: target/debug/rtk.exe dotnet ef migrations script --no-build --context SmokeContext
```

Result:

```text
Before:    526 tokens, 145 lines
After:      78 tokens,  12 lines
Reduced:   448 tokens
Savings:  85.2%
```

Filtered output shape:

```text
ok dotnet ef migrations script: 25 SQL statements to stdout
  CREATE TABLE IF NOT EXISTS "__EFMigrationsHistory" (
  "MigrationId" TEXT NOT NULL CONSTRAINT "PK___EFMigrationsHistory" PRIMARY KEY,
  "ProductVersion" TEXT NOT NULL
  );
  BEGIN TRANSACTION;
  ... +101 more lines
  [full output: ~/AppData\Local\rtk\tee\1780222634_dotnet-ef-migrations-script.log]
Warnings:
  ... NU1900 package vulnerability feed warning ...
```

## More outputs

```text
rtk dotnet ef migrations script --no-build --context SmokeContext
  85.2% savings, 526 -> 78 tokens, 145 -> 12 lines

rtk dotnet ef migrations add FilteredLedger --no-build --verbose --context SmokeContext
  83.2% savings, 298 -> 50 tokens

fixture: ef_migrations_add_complex.txt
  >=80% savings asserted in test; observed around 91.9% during local review
```

## Validation

```text
rtk cargo fmt --all
rtk cargo clippy --all-targets
rtk cargo test dotnet_cmd::tests:: -- --nocapture    # 73 passed
rtk cargo test discover -- --nocapture               # 393 passed
rtk git diff --check
rtk proxy .\target\debug\rtk.exe dotnet ef --version
```

Note: full `rtk cargo test --all` on this Windows checkout reached 1,975 passing tests, then failed existing `src/core/stream.rs` tests that spawn Unix commands such as `sh`, `true`, `false`, and `cat`. This PR intentionally leaves that unrelated portability issue out of scope.

Issue #2167, opened 2026-05-30
rtk grep calls system grep instead of ripgrep — breaks -g/--glob flags


jbgutierrez:
## Environment

- rtk 0.42.0 (Homebrew, macOS arm64)
- ripgrep 14.x installed at ~/homebrew/bin/rg

## Reproduction

```bash
rtk grep -i "compact" ~/.pi/agent/ -g "*.md" -l
```

**Expected**: filters files by glob using ripgrep internally
**Actual**: calls `/usr/bin/grep` which does not support `-g` or `--glob`

```
grep: invalid option -- g
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] ...
```

Also fails with `--glob`:
```bash
rtk grep -i "compact" ~/.pi/agent/ --glob "*.md"
# grep: unrecognized option `--glob`
```

## Context

`rtk rewrite` correctly rewrites `rg` → `rtk grep`, but `rtk grep` delegates to system `grep` instead of ripgrep. Since the rewrite assumes ripgrep semantics, any rg-specific flags break.

## Expected behavior

`rtk grep` should either:
1. Use ripgrep internally (since it rewrites `rg` commands to itself), or
2. Not rewrite `rg` commands that use flags incompatible with its own interface

## Workaround

`RTK_DISABLED=1 rg ...` bypasses the rewrite.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 93% |

### Summary

rtk grep delegates to /usr/bin/grep instead of ripgrep, breaking all rg-specific flags (--glob/-g, --fixed-strings, etc.) that rtk rewrite assumes will work.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

KuSh:
Hi, rtk  internally uses rg if installed but fallback to grep in case of rg error.

What does `rg -n --no-heading --no-ignore-vcs -i "compact" ~/.pi/agent/ -g "*.md" -l` returns on your system?
Could you also test your command without `-l` arg?

Thanks

olsavmic:
[...]

Issue #2060, opened 2026-05-23
rg --files is rewritten to rtk grep --files


LK4D4:
And it fails
```
$ rtk grep --files
/usr/bin/grep: option '--files' is ambiguous; possibilities: '--files-with-matches' '--files-without-match'
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```
We should either rewrite it to `rtk find` like I tried to do in #1903 or implement full-fledged `rtk grep --files`.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 92% |

### Summary

rg --files is incorrectly rewritten to rtk grep --files, which fails; should be rewritten to rtk find or handled natively

<details>
<summary>📁 Relevant files</summary>

- `src/rewrite/grep.rs`
- `src/rewrite/rg.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

smashah:
what's the temporary fix for this?

InCerryGit:
I can still reproduce this on the latest stable release, `rtk 0.42.4`.

```bash
$ rtk --version
rtk 0.42.4

$ rtk rewrite "rg --files src tests"
rtk grep --files src tests
```

The rewritten command is invalid because `rtk grep` is not a transparent `rg` wrapper:

```bash
$ rtk grep --files
/usr/bin/grep: option '--files' is ambiguous; possibilities: '--files-with-matches' '--files-without-match'
```

`rtk rg --files` works correctly, so a safer rewrite would be either:

```bash
rg --files <paths...> -> rtk rg --files <paths...>
```

or leaving `rg --files` unchanged until `rtk grep` supports the same argument contract.

This also affects OpenCode when the RTK plugin calls `rtk rewrite` before executing shell commands: even if the assistant emits a correct `rg --files ...`, the hook can rewrite it into the broken `rtk grep --files ...`.

MarlonPassos-git:
We hit the same issue from a Codex hook setup on Linux.

  Repro:

  ```bash
[...]

PR #2183, opened 2026-05-31
fix(grep): respect the invoked tool — grep runs grep, rg runs ripgrep

olsavmic:
## Summary

`rg` and `grep` were **both** rewritten to `rtk grep`, which always ran **ripgrep** as its backend. That collapsed the two tools into one and broke whichever didn't match the backend:

- ripgrep-only flags (`--glob`, `-g`, `-P`, `--files`) failed — the command fell through to system grep, which rejects them (#2167, #2060, #1651);
- grep commands silently misbehaved under rg — e.g. `-r` is *recursive* in grep but `--replace` in rg, so `grep -rn foo` rewrote every match to `n` instead of searching.

This PR **preserves the invoked tool's identity** from the rewrite onward, so each command runs its own backend with its own semantics — **no crossover in either direction**.

## Design

| Invoked | Rewritten to | Runs | Semantics |
|---|---|---|---|
| `grep …` | `rtk grep` | system **grep** | grep (BRE, grep flags, `-r`) |
| `rg …` | `rtk rg` | **ripgrep** | rg (`--glob`/`-g`/`-P`/`--files`, any order) |

- **rewrite**: split the single `^(rg\|grep)` rule into `grep → rtk grep` and `rg → rtk rg`.
- **`rtk rg`** (new subcommand): faithful ripgrep passthrough — all args forwarded verbatim (flat `trailing_var_arg`, so leading flags never break clap) — with RTK's by-file filtering; info/list flags (`--files`, `--json`, …) stream raw.
- **`rtk grep`**: now runs **system grep**, parsing grep's `file:line:content` output; `-t/--file-type` maps to grep `--include`. Keeps its documented `-l/-m/-t/--context-only` interface.
- shared `emit_filtered` takes a per-backend line parser (rg NUL form `file\0line:content` vs grep colon form).

## Why not "route everything through one backend"

The first iterations of this PR tried that (route rg-style fallbacks to rg). Review found it corrupts the *other* tool because grep and rg share short flags with conflicting meanings — most dangerously `grep -rn foo` → rg `--replace=n` → every match rewritten. There is no safe way to run one tool's syntax through the other; the only correct fix is to keep them separate.

## Test plan

- [x] `cargo fmt --check && cargo clippy --all-targets && cargo test` — **1985 passed, 7 ignored**, clippy clean, fmt clean
- [x] Unit tests: rewrite split (`rg`→`rtk rg`, `grep`→`rtk grep`), `parse_grep_line`, `grep_type_glob`, `is_raw_passthrough_flag`
- [x] Manual verification (built binary):

```
# rewrite preserves origin
rg --glob '*.py' alpha .   ->  rtk rg --glob '*.py' alpha .
grep -rn alpha .           ->  rtk grep -rn alpha .

# rtk rg: ripgrep, faithful + filtered
$ rtk rg --glob '*.py' alpha gp     # was: grep: unrecognized option `--glob'
2 matches in 2 files: ...
$ rtk rg -P 'alpha\s' gp/a.py        # was: grep: invalid option -- P
1 matches in 1 files: ...
$ rtk rg --files gp                  # issue #2060

# rtk grep: system grep, correct (no rg --replace corruption)
$ rtk grep -rn alpha gp              # earlier naive routing gave "n one"
gp/a.py:1:alpha one ...
$ rtk grep -E 'alpha|beta' gp/a.py   # grep ERE, respected
```

## Notes
- `rtk grep` now runs system grep rather than ripgrep — a deliberate behavior change so grep semantics are respected (loses rg's speed/gitignore-awareness for `grep` commands; that's grep being grep). `rtk rg` is the path for ripgrep.
- Verified on macOS (BSD grep) + ripgrep 15.1.0; logic is platform-agnostic but not yet exercised on GNU grep.

### DIFF
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -200,9 +200,13 @@ Supporte a la fois la syntaxe RTK et la syntaxe native `find` (`-name`, `-type`,
 
 ---
 
-### `rtk grep` -- Recherche dans le contenu
+### `rtk grep` / `rtk rg` -- Recherche dans le contenu
 
-**Objectif :** Remplace `grep` et `rg` avec une sortie groupee par fichier, tronquee.
+**Objectif :** Sortie groupee par fichier, tronquee. `rtk grep` execute le `grep`
+systeme (semantique grep) ; `rtk rg` execute ripgrep (semantique rg, flags
+`--glob`/`-g`/`-P`/`--files`...). Chaque commande conserve son propre outil : pas
+de bascule de l'un vers l'autre (les flags courts comme `-r` different entre les
+deux).
 
 **Syntaxe :**
 ```bash
@@ -219,7 +223,7 @@ rtk grep <pattern> [chemin] [options]
 | `--file-type` | `-t` | tous | Filtrer par type (ts, py, rust, etc.) |
 | `--line-numbers` | `-n` | oui | Numeros de ligne (toujours actif) |
 
-Les arguments supplementaires sont transmis a `rg` (ripgrep). Les flags qui changent le format de sortie (`-c`, `-l`, `-L`, `-o`, `-Z`) passent directement a `rg`/`grep` sans filtrage RTK.
+Les arguments supplementaires sont transmis a l'outil sous-jacent (`grep` pour `rtk grep`, `rg` pour `rtk rg`). Les flags qui changent le format de sortie (`-c`, `-l`, `-L`, `-o`, `-Z`) passent directement a l'outil sans filtrage RTK.
 
 **Economies :** ~80%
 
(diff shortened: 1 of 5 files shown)


KuSh:
Thanks for digging into this. PR #2641 (merged 2026-06-26, commit eafadcee) made rtk grep run grep and rtk rg run ripgrep by replacing src/cmds/system/grep_cmd.rs with src/cmds/system/search.rs and an engine parameter, which is the separation this PR was after; on the current develop binary `rtk rewrite "grep -rn foo src/"` returns `rtk grep -rn foo src/` and `rtk rewrite "rg -t rust foo src/"` returns `rtk rg -t rust foo src/`. The rewrite rules and the docs update this PR carries are already present there too. The one part develop does not have is the --file-type option translating an rg type name into a grep --include glob, and that was a deliberate call: develop dropped --file-type and forwards -t to the engine so rtk rg filters by type while rtk grep reports grep's own invalid option error, with a test pinning it. Since the file this PR edits no longer exists, I am closing it, but please comment or reopen if you think the type mapping is still worth having.

PR #2188, opened 2026-05-31
fix(go): respect build failure exit status

### LINKED ISSUE #2185: go build failure misreported as "Go build: Success" for package-pattern / module-resolution errors
## Summary

When `rtk` wraps `go build` and the build fails due to a **package-pattern / module-resolution error** (not a compile error), `rtk` prints `Go build: Success` even though the wrapped `go build` exited non-zero (`1`) and wrote an error to stderr. The "Success" text directly contradicts the preserved exit code, and the real error message is swallowed.

Compile errors are handled correctly — only this class of non-compile build failures is misreported.

## Environment

- rtk: `0.41.0`
- go: `go1.26.2`
- OS: `darwin/arm64`

## Minimal reproduction (self-contained, no project needed)

```bash
mkdir -p /tmp/rtk-nomod && cd /tmp/rtk-nomod   # empty dir, no go.mod

# real output (bypassing rtk's filter via `rtk proxy`):
rtk proxy go build ./...
# => pattern ./...: directory prefix . does not contain main module or its selected dependencies
# => exit 1

# through rtk:
rtk go build ./...
# => Go build: Success      <-- BUG: contradicts the failure
# => exit 1                 (exit code is correctly preserved)
```

Also reproduces at a Go workspace (`go.work`) root that has no module of its own:

```bash
rtk proxy go build ./...
# pattern ./...: directory prefix . does not contain modules listed in go.work or their selected dependencies   (exit 1)
rtk go build ./...
# Go build: Success                                                                                              (exit 1)
```

## Contrast: real compile errors ARE handled correctly

```bash
mkdir -p /tmp/rtk-repro && cd /tmp/rtk-repro
printf 'package main\nfunc main() { thisFunctionDoesNotExist() }\n' > main.go
go mod init rtkrepro

rtk go build .
# Go build: 1 errors
# 1. ./main.go:2:14: undefined: thisFunctionDoesNotExist   (exit 1)  -- correct
```

So the formatter correctly detects `file:line:` compile diagnostics, but when `go build` fails with a pattern / module-resolution error on **stderr** (which doesn't match the compile-diagnostic shape), it falls through to the default "Success" message and discards both the stderr content and the signal of the non-zero exit.

## Expected behavior

- Never print `Go build: Success` when the wrapped command exits non-zero.
- Gate the "Success" branch on the child process exit code (`== 0`), not solely on whether compile-diagnostic lines were parsed from stdout.
- On failure, surface the real stderr (e.g. the `pattern ./...:` message), or at least emit something like `Go build: failed (exit 1)`.

## Impact

`rtk` targets LLM / terminal agent sessions. Running `go build ./...` from a Go workspace (`go.work`) root is a very common operation; the build fails with a pattern error, but the agent reads `Go build: Success` and proceeds on a false premise. The misreport contradicts the exit code and hides the actual cause of the failure.

jsiu93:
Closes #2185.

## Summary

- Gate `rtk go build` success output on the wrapped process exit code.
- Surface unrecognized non-zero `go build` output instead of reporting `Go build: Success`.
- Preserve Go package-pattern/module-resolution errors such as `pattern ./...: directory prefix ...`.

## Test plan

- [x] Manual repro in an empty directory: `target/debug/rtk go build ./...`

### DIFF
--- a/src/cmds/go/go_cmd.rs
+++ b/src/cmds/go/go_cmd.rs
@@ -92,11 +92,11 @@ pub fn run_build(args: &[String], verbose: u8) -> Result<i32> {
         eprintln!("Running: go build {}", args.join(" "));
     }
 
-    runner::run_filtered(
+    runner::run_filtered_with_exit(
         cmd,
         "go build",
         &args.join(" "),
-        filter_go_build,
+        filter_go_build_with_exit,
         crate::core::runner::RunOptions::with_tee("go_build"),
     )
 }
@@ -571,6 +571,10 @@ fn is_go_test_failure_line(line: &str) -> bool {
 
 /// Filter go build output - show only errors
 pub(crate) fn filter_go_build(output: &str) -> String {
+    filter_go_build_with_exit(output, 0)
+}
+
+fn filter_go_build_with_exit(output: &str, exit_code: i32) -> String {
     let mut errors: Vec<String> = Vec::new();
 
     for line in output.lines() {
@@ -581,7 +585,11 @@ pub(crate) fn filter_go_build(output: &str) -> String {
     }
 
     if errors.is_empty() {
-        return "Go build: Success".to_string();
+        return if exit_code == 0 {
+            "Go build: Success".to_string()
+        } else {
+            format_go_build_failure(output, exit_code)
+        };
     }
 
     let mut result = String::new();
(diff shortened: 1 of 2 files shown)


pszymkowiak (approved):
Thanks for the thorough fix — the diagnose matches the repro in #2185 exactly, and the two-layer approach (allow-list entry for `pattern ` + exit-aware fallback) is the right call. The `runner.rs` refactor (`RunMode::FilteredWithExit` + shared `run_captured_filter`) is a particularly valuable piece: it lays the plumbing so other ecosystems can opt into exit-aware filtering without duplicating logic. Several other filters (`cargo clippy`, `mypy`, `ruff`, `tsc`, `eslint`, `golangci-lint`, `go vet`) have the same latent "no errors detected = Success" pattern and would benefit from the same migration.

A second independent review (code-reviewer agent) and empirical sweep across 19 commands on develop vs this branch confirmed byte-identical behavior on every non-go-build path.

A few light suggestions for follow-up, none blocking:

1. **ANSI escape codes in the failure fallback**: `format_go_build_failure` echoes raw stderr lines through `truncate` without passing through `strip_ansi`. Real `go build` doesn't emit ANSI, so this is a theoretical concern, but a malicious or unusual `go` wrapper could inject OSC 8 hyperlinks or alt-screen sequences that survive into the LLM context. A `strip_ansi` call per line would close that gap cheaply.

2. **`"pattern "` prefix in `is_go_build_error_line`**: matches any line starting with `pattern ` (lowercased). A custom build tool printing `pattern ./internal/...: 12 files matched` would be classified as an error. Tightening to `"pattern ./"` or requiring `": "` after the argument would eliminate that risk while still matching every real `go build` pattern-error shape.

3. **Tests**: the new tests use `assert!(result.contains(...))` — consistent with the rest of `go_cmd.rs`, so just a style note. Adding edge-case tests for `filter_go_build_with_exit("", 1)` and whitespace-only inputs would round out coverage.

4. **Follow-up opportunity** (out of scope for this PR): migrating `cargo clippy`, `mypy`, `ruff`, `tsc`, `eslint`, `golangci-lint`, and `go vet` to `run_filtered_with_exit` would close the same class of bug across the rest of the codebase.

Approving — the fix is correct, the refactor is well-bounded, and behavior is preserved across every other command. Thanks for the careful work!

PR #2196, opened 2026-06-01
ci: add aarch64-unknown-linux-musl release target (closes #1331)

### LINKED ISSUE #1331: Add aarch64-unknown-linux-musl target to release binaries

brandstaetter:
Fixes #1331.

  Alpine/musl ARM64 systems cannot run the glibc binary (`aarch64-unknown-linux-gnu`) — it fails with symbol relocation errors (`fcntl64`, `__res_init`).
  This adds a statically-linked musl build for ARM64, mirroring the existing `x86_64-unknown-linux-musl` asset.

  - **release.yml**: add `aarch64-unknown-linux-musl` to the build matrix, built via `taiki-e/setup-cross-toolchain-action` so the standard `cargo build`
  step and `RTK_TELEMETRY_*` compile-time env work unchanged.
  - **install.sh**: detect musl libc and select the musl asset on aarch64 Linux (glibc systems keep the gnu binary).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,6 +52,10 @@ jobs:
             os: ubuntu-latest
             archive: tar.gz
             cross: true
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-latest
+            archive: tar.gz
+            setup_cross_toolchain: true
           # Windows
           - target: x86_64-pc-windows-msvc
             os: windows-latest
@@ -78,6 +82,15 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y musl-tools
+
+      # Sets up a native musl cross-compiler + linker for targets like
+      # aarch64-unknown-linux-musl so the standard `cargo build` step below
+      # works and build-time env (RTK_TELEMETRY_*) passes through normally.
+      - name: Set up cross toolchain
+        if: matrix.setup_cross_toolchain
+        uses: taiki-e/setup-cross-toolchain-action@v1
+        with:
+          target: ${{ matrix.target }}
 
       - name: Build
         run: cargo build --release --target ${{ matrix.target }}
--- a/install.sh
+++ b/install.sh
@@ -45,6 +45,19 @@ detect_arch() {
     esac
 }
 
+# Detect C library (glibc vs musl)
+# Alpine and other lightweight distros use musl; the glibc binaries fail there
+# with "Error relocating ... symbol not found". Default to gnu, switch to musl
+# only when musl is detected.
+detect_libc() {
+    LIBC="gnu"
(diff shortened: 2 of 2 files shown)


pszymkowiak:
Thanks for this @brandstaetter — adding the `aarch64-unknown-linux-musl` release target (via `taiki-e/setup-cross-toolchain-action`) is the right call and does address #1331.

We're leaning toward consolidating on #2089, which tackles the same problem a bit more broadly: instead of keeping `aarch64-unknown-linux-gnu` and adding musl alongside it, it makes Linux/arm64 ship the **static musl binary by default** (dropping the gnu one), mirroring what x86_64 already does.

That difference matters for #615: the reporters there are on Raspberry Pi OS / RHEL — i.e. **glibc that's too old (2.36) for our 2.39-linked binary, not musl systems**. The `detect_libc()` here only switches to musl when it detects an actual musl libc (Alpine), so on Debian Bookworm it would still pick the `-gnu` artifact and fail the same way. #2089 sidesteps that by giving everyone on arm64 the dependency-free static binary.

Since both PRs modify `release.yml` and `install.sh` in incompatible ways, would you be OK closing this in favor of #2089? Really appreciate you digging into the cross-toolchain setup — the analysis was spot on. 🙏

PR #2197, opened 2026-06-01
fix(rewrite): handle newline-separated commands

### LINKED ISSUE #2191: rtk rewrite mishandles newline-separated (multi-line) commands → only first line rewritten / empty output
## Summary

`rtk rewrite` does not treat newline (`\n`) as a statement separator. It appears to only parse/rewrite the **first** newline-delimited segment of a command, which breaks multi-statement commands written across multiple lines:

- If the first line is **not** a rewritable command (`cd`, `echo`, …), `rtk rewrite` returns **empty output with exit 1** ("no rewrite"), even when later lines contain rewritable commands (`grep`, `tsc`, …).
- If the first line **is** rewritable, only that first line is rewritten; rewritable commands on subsequent lines are left untouched.

Semicolon-separated equivalents are handled correctly. So the bug is specifically newline handling.

Combined with the official Claude Code hook (`rtk-rewrite.sh`, which applies the rewrite when `rtk rewrite` exits 0/3), this surfaces as **silently truncated output** when an agent runs a multi-line shell snippet containing several `grep`s — only the first section is produced. This was hit repeatedly in a real Claude Code session.

## Environment

- rtk 0.42.0 (Homebrew, macOS arm64) — also reproduced on 0.35.0
- Claude Code PreToolUse hook `rtk-rewrite.sh` (`# rtk-hook-version: 3`)

## Reproduction (deterministic, no Claude Code needed)

```bash
printf 'x\n' > /tmp/f.txt

# (1) Semicolon-separated — CORRECT (both greps rewritten):
rtk rewrite 'cd /tmp; grep -c x /tmp/f.txt; grep -c x /tmp/f.txt'
#   -> cd /tmp; rtk grep -c x /tmp/f.txt; rtk grep -c x /tmp/f.txt      (exit 0)

# (2) Same commands, newline-separated, first line is `cd` — BUG: empty + exit 1
rtk rewrite "$(printf 'cd /tmp\ngrep -c x /tmp/f.txt\ngrep -c x /tmp/f.txt')"
#   -> (empty)                                                          (exit 1)

# (3) Newline-separated, first line is grep — BUG: only FIRST line rewritten
rtk rewrite "$(printf 'grep -c x /tmp/f.txt\ngrep -c x /tmp/f.txt')"
#   -> rtk grep -c x /tmp/f.txt
#      grep -c x /tmp/f.txt          <- second grep left un-rewritten   (exit 0)
```

Bisection — every multiline form whose first line is non-rewritable returns `exit=1, len=0`:

| command | result |
|---|---|
| `cd /tmp; grep…; grep…` (semicolons) | exit 0, full rewrite ✓ |
| `grep…\ngrep…` (newlines, grep first) | exit 0, only 1st line rewritten |
| `cd /tmp\ngrep…\ngrep…` | **exit 1, empty** |
| `cd /tmp\necho A\ngrep…\necho B\ngrep…` | **exit 1, empty** |
| `echo 中文\ngrep…\ngrep…` | **exit 1, empty** |
| `cd "/tmp/a b"\ngrep…\ngrep…` | **exit 1, empty** |

## Expected

`rtk rewrite` should treat `\n` as a statement separator (like `;`), rewriting each rewritable statement independently and preserving every statement in the output — so a newline-separated command yields the same result as its semicolon-separated equivalent.

## Impact

Multi-line shell snippets are extremely common in agent/hook usage (heredocs, multi-step verification blocks). Today they are either:
- left with only the first statement token-optimized (mild), or
- effectively pass-through/empty depending on the first line (and, via hooks that act on the rewrite, produce truncated/partial command execution with no error) — silent and easy to mistake for a tool/agent bug.

## Suggested fix

In the rewrite parser (registry), split the command on top-level statement separators including newline (`\n`) in addition to `;` / `&&` / `||`, rewrite each segment, and re-join preserving the original separators.

jsiu93:
Closes #2191.

## Summary

- Treat top-level newline and CRLF separators like other shell command separators during `rtk rewrite`.
- Rewrite every newline-delimited segment independently while preserving quoted newlines, heredoc passthrough, line-continuation collapse, and pipe group behavior.
- Cover hook permission checks so newline-separated commands are evaluated segment-by-segment before allow/deny/ask decisions.

## Test plan

- [x] Manual repro: `target/debug/rtk rewrite "$(printf 'cd /tmp\ngrep -c x /tmp/f.txt\ngrep -c x /tmp/f.txt')"`

### DIFF
--- a/src/discover/lexer.rs
+++ b/src/discover/lexer.rs
@@ -133,6 +133,23 @@ pub fn tokenize(input: &str) -> Vec<ParsedToken> {
                 byte_pos += char_len;
                 current_start = byte_pos;
             }
+            '\n' | '\r' => {
+                flush_arg(&mut tokens, &mut current, current_start);
+                let start = byte_pos;
+                let mut value = c.to_string();
+                byte_pos += char_len;
+                if c == '\r' && chars.peek() == Some(&'\n') {
+                    chars.next();
+                    byte_pos += 1;
+                    value.push('\n');
+                }
+                tokens.push(ParsedToken {
+                    kind: TokenKind::Operator,
+                    value,
+                    offset: start,
+                });
+                current_start = byte_pos;
+            }
             '&' => {
                 flush_arg(&mut tokens, &mut current, current_start);
                 let start = byte_pos;
@@ -492,6 +509,30 @@ mod tests {
         assert!(tokens
             .iter()
             .any(|t| t.kind == TokenKind::Operator && t.value == ";"));
+    }
+
+    #[test]
+    fn test_newline_operator() {
+        let tokens = tokenize("cmd1\ncmd2");
+        assert!(tokens
+            .iter()
+            .any(|t| t.kind == TokenKind::Operator && t.value == "\n"));
+    }
+
+    #[test]
+    fn test_crlf_operator() {
+        let tokens = tokenize("cmd1\r\ncmd2");
+        assert!(tokens
+            .iter()
+            .any(|t| t.kind == TokenKind::Operator && t.value == "\r\n"));
+    }
+
+    #[test]
+    fn test_quoted_newline_not_operator() {
+        let tokens = tokenize("echo \"cmd1\ncmd2\"");
+        assert!(!tokens
+            .iter()
+            .any(|t| t.kind == TokenKind::Operator && t.value.contains('\n')));
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


jsiu93:
Closing this: the multiline rewrite landed upstream in #3319 (line-by-line rewrite via the quote-aware lexer, with `&&`/`||`/`|` continuation joining, CRLF preservation, and conservative passthrough for heredocs/control flow), and #2191 was closed after 952245d. That covers what this PR set out to do, and the branch has since drifted into conflicts with develop. Thanks for the review bandwidth.

Issue #2200, opened 2026-06-01
fix(openclaw): rtk-rewrite plugin silently drops all rewrites — exit code 3 treated as error


kzzalews:
## Summary

The `rtk-rewrite` OpenClaw plugin (`openclaw/index.ts`) silently drops all command rewrites. Every `rtk rewrite` call that produces a suggestion returns exit code 3, which `execSync` treats as an exception. The catch block returns `null`, so no command is ever rewritten — the plugin is effectively dead code.

## The problem

`rtk rewrite` returns:
- **exit code 3** when it has a rewrite suggestion (known command)
- **exit code 1** when it has no rewrite (unknown command)
- **exit code 0** — never

The plugin's `tryRewrite()` function:

```typescript
function tryRewrite(command: string): string | null {
  try {
    const result = execSync(`rtk rewrite ${JSON.stringify(command)}`, {
      encoding: "utf-8",
      timeout: 2000,
    }).trim();
    return result && result !== command ? result : null;
  } catch {
    return null; // ← always lands here for exit code 3
  }
}
```

`execSync` throws on any non-zero exit code. The catch block discards the rewrite. The command passes through unchanged.

## How we found it

We discovered this while investigating why our RTK token savings weren't growing. We:

1. Ran `rtk rewrite "git log --oneline -5"` → returns `rtk git log --oneline -5` with exit code 3
2. Simulated the plugin's `tryRewrite()` in Node.js → `execSync` threw, catch returned `null`
3. Tested all common commands (`git`, `kubectl`, `find`, `cat`, `grep`) — all return exit code 3, all silently dropped

## The fix

Check `e.stdout` in the catch handler — exit code 3 with stdout output is a valid rewrite suggestion, not an error:

```typescript
} catch (e: any) {
    if (e?.stdout) {
      const result = String(e.stdout).trim();
      if (result && result !== command) return result;
    }
    return null;
  }
```

## What we verified

Tested the fix against all common shell commands:

```
git log --oneline -5     → rtk git log --oneline -5     ✅
kubectl get pods -A      → rtk kubectl get pods -A      ✅
cat /tmp/test            → rtk read /tmp/test           ✅
ls -la /tmp              → rtk ls -la /tmp              ✅
find /tmp -name "*.log"  → rtk find /tmp -name "*.log"  ✅
grep -r test /tmp        → rtk grep -r test /tmp        ✅
```

All previously returned `null` (no rewrite) with the upstream code.

## Environment

[...]

PR #2202, opened 2026-06-01
fix(openclaw): handle exit code 3 from rtk rewrite

kzzalews:
## Problem

The rtk-rewrite OpenClaw plugin silently drops all command rewrites. Every `rtk rewrite` call that produces a suggestion returns exit code 3, which `execSync` treats as an exception. The catch block returns `null`, so no command is ever rewritten.

See #2200 for full analysis.

## Fix

Check `e.stdout` in the catch handler — exit code 3 with stdout output is a valid rewrite suggestion, not an error:

```typescript
} catch (e: any) {
    if (e?.stdout) {
      const result = String(e.stdout).trim();
      if (result && result !== command) return result;
    }
    return null;
  }
```

## Verified

All common shell commands now correctly rewritten:

```
git log --oneline -5     → rtk git log --oneline -5     ✅
kubectl get pods -A      → rtk kubectl get pods -A      ✅
cat /tmp/test            → rtk read /tmp/test           ✅
ls -la /tmp              → rtk ls -la /tmp              ✅
find /tmp -name "*.log"  → rtk find /tmp -name "*.log"  ✅
grep -r test /tmp        → rtk grep -r test /tmp        ✅
```

All previously returned `null` (no rewrite) with the upstream code.

---

**Aiko** — AI assistant in [OpenClaw](https://github.com/openclaw/openclaw)
Running RTK on k3s with Kilo Gateway

Config: RTK 0.42.0 | OpenClaw 2026.5.28 | Node.js v24.14.0 | Model: MiMo-V2.5-Pro

*Reviewed and verified by Zal (@kzzalews)*

### DIFF
--- a/openclaw/index.ts
+++ b/openclaw/index.ts
@@ -7,6 +7,14 @@
  * All rewrite logic lives in `rtk rewrite` (src/discover/registry.rs).
  * This plugin is a thin delegate — to add or change rules, edit the
  * Rust registry, not this file.
+ *
+ * Exit code protocol for `rtk rewrite`:
+ *   0 + stdout  Allow — rewrite found, explicitly allowed → auto-apply
+ *   1           No RTK equivalent → pass through unchanged
+ *   2           Deny rule matched → block the call
+ *   3 + stdout  Ask rule matched (or default) → rewrite, require approval
+ *
+ * See: src/hooks/rewrite_cmd.rs
  */
 
 import { execFileSync } from "node:child_process";
@@ -24,7 +32,20 @@ function checkRtk(): boolean {
   return rtkAvailable;
 }
 
-function tryRewrite(command: string): string | null {
+/**
+ * Delegate to `rtk rewrite` and interpret the exit code.
+ *
+ * Returns a tuple `[rewritten, verdict?]`:
+ *   [string]        — rewrite, auto-apply (exit 0)
+ *   [string, "ask"] — rewrite, require user approval (exit 3)
+ *   [null, "deny"]  — command matched a deny rule (exit 2)
+ *   [null]          — no rewrite / passthrough (exit 1 or no change)
+ */
+type RewriteVerdict = "ask" | "deny";
+
+function tryRewrite(
+  command: string
+): [string | null, RewriteVerdict?] {
   try {
     const result = execFileSync("rtk", ["rewrite", command], {
       encoding: "utf-8",
@@ -32,9 +53,22 @@ function tryRewrite(command: string): string | null {
     })
       .toString()
       .trim();
-    return result && result !== command ? result : null;
-  } catch {
-    return null;
+    // Exit 0 — Allow: rewrite and auto-apply
+    return [result && result !== command ? result : null];
+  } catch (e: any) {
+    // Exit 3 — Ask: rewrite available but user must approve
+    if (e?.status === 3 && e.stdout) {
+      const result = e.stdout.toString().trim();
+      if (result && result !== command) return [result, "ask"];
+      // Exit 3 but no usable stdout — treat as passthrough
+      return [null];
+    }
+    // Exit 2 — Deny: command matched a deny rule, block the call
+    if (e?.status === 2) {
+      return [null, "deny"];
+    }
+    // Exit 1 or unknown — no rewrite, pass through
+    return [null];
   }
 }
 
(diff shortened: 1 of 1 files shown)


openclaw/index.ts
```diff
@@ -31,7 +31,13 @@ function tryRewrite(command: string): string | null {
...
     }).trim();
     return result && result !== command ? result : null;
-  } catch {
+  } catch (e: any) {
+    // rtk rewrite returns exit code 3 when it has a rewrite suggestion
+    // but non-zero codes cause execSync to throw — check stdout
+    if (e?.stdout) {
+      const result = String(e.stdout).trim();
+      if (result && result !== command) return result;
+    }
```
KuSh, line 69: This can't be fixed like that, in case status is 3 (and only in that case) we should set the requireApproval attribute to asks for user approval, see [openclaw's doc](https://docs.openclaw.ai/plugins/hooks#tool-call-policy:~:text=requireApproval%20pauses,runtime%2E)

Perhaps return a tuple `[string | null, "ask"?]` and adapt before_tool_call callback to be able to respond with requireApproval if needed

KuSh (changes requested):
Thanks for your PR! While your current solution works as a temporary workaround, it’s not yet production-ready. I’ve left some suggestions to help guide you toward a more robust, final implementation.

Let me know if you’d like further clarification or assistance!

openclaw/index.ts
```diff
@@ -7,41 +7,76 @@
...
+      if (result && result !== command) return [result, "ask"];
+      // Exit 3 but no usable stdout — treat as passthrough
+      return [null, undefined];
+    }
+    // Exit 2 — Deny: command matched a deny rule, block the call
+    if (e?.status === 2) {
+      return [null, "deny"];
+    }
+    // Exit 1 or unknown — no rewrite, pass through
+    return [null, undefined];
```
KuSh:
```suggestion
type RewriteVerdict = "ask" | "deny";
function tryRewrite(
  command: string
): [string | null, RewriteVerdict?] {
try {
const result = execSync(`rtk rewrite ${JSON.stringify(command)}`, {
encoding: "utf-8",
timeout: 2000,
}).trim();
    // Exit 0 — Allow: rewrite and auto-apply
    return [result && result !== command ? result : null];
  } catch (e: any) {
    // Exit 3 — Ask: rewrite available but user must approve
    if (e?.status === 3 && e?.stdout) {
      const result = String(e.stdout).trim();
      if (result && result !== command) return [result, "ask"];
      // Exit 3 but no usable stdout — treat as passthrough
      return [null];
    }
    // Exit 2 — Deny: command matched a deny rule, block the call
    if (e?.status === 2) {
      return [null, "deny"];
    }
    // Exit 1 or unknown — no rewrite, pass through
    return [null];
```

openclaw/index.ts
```diff
@@ -58,19 +93,65 @@ export default function register(api: any) {
...
+        params: { ...event.params, command: rewritten },
+      };
+
+      // Exit 3 — Ask: rewrite but require user approval
+      if (verdict === "ask") {
+        result.requireApproval = {
+          title: "RTK rewrite suggestion",
+          description: `Rewrite: \`${command}\` → \`${rewritten}\``,
+          severity: "info",
+          timeoutMs: approvalTimeoutMs,
```
KuSh: The documentation says:
> Omit it to use the default observation/decision timeout that the hook runner applies generically

So I don’t think we should set it here.

openclaw/index.ts
```diff
@@ -58,19 +93,65 @@ export default function register(api: any) {
...
+          timeoutBehavior: "deny",
+          allowedDecisions: ["allow-once", "deny"],
```
KuSh, line 140: Why did you omit `"allow-always"`?

KuSh (changes requested):
Just a few remaining nitpicks and questions

kzzalews:
Thanks for the thorough review, KuSh — really appreciate the guidance.

**1. `RewriteVerdict` type** — done in `9827755`. Extracted `type RewriteVerdict = "ask" | "deny"` and updated the `tryRewrite` signature accordingly.

**2. `timeoutMs` omission** — done in the same commit. Removed `timeoutMs` from `requireApproval` and the corresponding `approvalTimeoutMs` config option from `openclaw.plugin.json`. The approval prompt now falls back to the Gateway's default timeout.

**3. `allow-always` omission** — intentional, but I should have documented the reasoning. The OpenClaw docs are explicit that `allow-always` does not auto-persist for plugin hooks:

> *"`allow-always` is only durable when the requesting plugin or runtime implements that persistence. For ordinary `before_tool_call.requireApproval` hooks, OpenClaw treats `allow-once` and `allow-always` as approval decisions for the current call and passes the resolved value to `onResolution`."*
> — [Plugin permission requests — Decision behavior](https://docs.openclaw.ai/plugins/plugin-permission-requests#decision-behavior)

The troubleshooting section reinforces this:

> *"`allow-always` appears but the next call prompts again. The generic plugin approval flow does not automatically persist trust for arbitrary hooks."*
> — [Plugin permission requests — Troubleshooting](https://docs.openclaw.ai/plugins/plugin-permission-requests#troubleshooting)

Since this plugin doesn't implement its own trust persistence, offering `allow-always` would mislead users — the button would look like it remembers the decision, but the next call would prompt again. I'd rather be honest and offer only `allow-once` + `deny`. If persistent trust is desired, I can add a config-based allowlist of trusted command patterns in a follow-up.

openclaw/index.ts
```diff
@@ -24,17 +32,41 @@ function checkRtk(): boolean {
...
+      if (result && result !== command) return [result, "ask"];
+      // Exit 3 but no usable stdout — treat as passthrough
+      return [null, undefined];
+    }
+    // Exit 2 — Deny: command matched a deny rule, block the call
+    if (e?.status === 2) {
+      return [null, "deny"];
+    }
+    // Exit 1 or unknown — no rewrite, pass through
+    return [null, undefined];
```
KuSh: One regression (execFileSync return value can be a Buffer) and some code cleanups / nitpicks
```suggestion
    })
      .toString()
      .trim();
    // Exit 0 — Allow: rewrite and auto-apply
    return [result && result !== command ? result : null];
  } catch (e: any) {
    // Exit 3 — Ask: rewrite available but user must approve
    if (e?.status === 3 && e.stdout) {
      const result = e.stdout.toString().trim();
      if (result && result !== command) return [result, "ask"];
      // Exit 3 but no usable stdout — treat as passthrough
      return [null];
    }
    // Exit 2 — Deny: command matched a deny rule, block the call
    if (e?.status === 2) {
      return [null, "deny"];
    }
    // Exit 1 or unknown — no rewrite, pass through
    return [null];
```

openclaw/index.ts
KuSh: Still missing a comment explaining the absence of "allow-always"

KuSh (changes requested):

kzzalews:
Thanks KuSh! All three rounds of feedback are now addressed in a single squashed commit (776a4b1):

1. **`execFileSync` restored** — shell interpolation removed, `.toString()` added for `Buffer` safety
2. **`timeoutMs` removed** — falling back to Gateway's default decision timeout
3. **`allow-always` omitted** — inline comment added explaining OpenClaw doesn't auto-persist trust for plugin hooks
4. **Type safety cleanups** — `e.stdout.toString()` for consistency, redundant optional chaining removed, explicit `undefined` trimmed from tuple returns

KuSh (approved):
LGTM, thanks for keeping up!

PR #2205, opened 2026-06-01
fix(git): add visible truncation markers to filtered output

### LINKED ISSUE #2204: passthrough_max_chars truncates output silently (no marker) — git log shows 50 of 252 commits, misleads scope
**Describe the bug**

`[limits].passthrough_max_chars` (default `2000`) caps passthrough command output, but truncates **silently** — no `... N more lines` marker, no hint that output was cut. A consumer (human or LLM agent) reads the truncated output as complete.

This caused a real correctness failure: `git log <base>..HEAD` on a 252-commit branch returned only ~50 lines. An automated agent read "50 commits", planned a history rewrite (squash/rebase) on that wrong scope, and had to redo work once the true count (`git rev-list --count` = 252) surfaced. `wc -l` on the piped output also reads 50, so even a line-count sanity check is fooled.

**To reproduce**

1. On a branch with >50 commits ahead of base
2. Run `git log <base>..HEAD` through the RTK hook (rewritten to `rtk git log`)
3. Output stops at ~2000 chars (~50 lines) with no indication it was truncated
4. `git rev-list --count <base>..HEAD` shows the true (larger) number

**Expected**

Truncation should be **visible**. Any one of:

- Append a marker when output is cut, e.g. `… [rtk: truncated, N of M lines shown — full: rtk run git log …]`
- Exempt history-listing commands (`git log`, `git rev-list`) from the char cap by default — these are exactly the cases where completeness matters
- Make the cap a line count rather than a char count for `git log`, and document it

**Environment**

- rtk 0.42.0 (Homebrew, macOS)
- Config at `~/Library/Application Support/rtk/config.toml` — note: the README/docs reference `~/.config/rtk/`. On macOS the active path is the Application Support one; might be worth documenting the platform difference.

**Workaround**

Added `"git log"` to `transparent_prefixes` so it bypasses the cap. Works, but the silent-truncation behavior is the real issue — a visible marker would have surfaced it immediately instead of after a wrong-scope history rewrite.

x-haolun:
Closes #2204

Added visible truncation markers when git log output is capped (commits or lines), and changed truncate() to show [RTK: truncated] instead of silent '...'.

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -622,10 +622,9 @@ pub(crate) fn filter_log_output(
     let mut out = result.join("\n").trim().to_string();
     if omitted_commits > 0 {
         out.push_str(&format!(
-            "\n[RTK: truncated — {} of {} commits shown — run `rtk proxy git log {}` for full output]",
+            "\n[RTK: truncated — {} of {} commits shown]",
             max_commits,
             commits.len(),
-            "" // args would ideally be passed through
         ));
     }
     out
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -559,12 +559,21 @@ pub(crate) fn filter_log_output(
     if user_format {
         let lines: Vec<&str> = output.lines().collect();
         let max_lines = if user_set_limit { lines.len() } else { limit };
-        return lines
+        let omitted = lines.len().saturating_sub(max_lines);
+        let mut out = lines
             .iter()
             .take(max_lines)
             .map(|l| truncate_line(l, truncate_width))
             .collect::<Vec<_>>()
             .join("\n");
+        if omitted > 0 {
+            out.push_str(&format!(
+                "\n[RTK: truncated — {} of {} lines shown]",
+                max_lines,
+                lines.len()
+            ));
+        }
+        return out;
     }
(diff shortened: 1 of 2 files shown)


x-haolun:
Addressed all 3 review comments from `copilot-pull-request-reviewer`:

1. **truncate() size guarantee break** ✅ — Reverted `truncate()` to strictly length-bounded behavior (ellipsis only). The function now guarantees the returned string fits within `max_len` as it did before.

2. **RTK-branded marker in core utility** ✅ — Removed `[RTK: truncated — N of M chars]` from `truncate()`. The core utility is generic again. Truncation annotations stay in `git.rs` where they belong (the git-specific call sites).

3. **Empty placeholder in git log suggestion** ✅ — Removed the dangling `run `rtk proxy git log {}`` suggestion with the empty `{}` placeholder. Now just shows `[RTK: truncated — N of M commits shown]`.

KuSh:
Closing as a duplicate of #2584, which fixes the same hidden git log truncation from #2204 and also marks the linter JSON fallback paths, with regression tests for both. Review and comments are welcome over there. If you meant the core::utils::truncate change as a separate fix, please say so here and it can be looked at on its own.

PR #1982, opened 2026-05-20
fix(cmds/ruby): accept version-only JSON from rubocop/rspec --version

YOMXXX:
## Summary

Fixes #1946. `rtk rubocop --version` and `rtk rspec --version` were polluting
stdout with two redundant warnings:

```
[rtk] rubocop: JSON parse failed (missing field `files` at line 1 column 20)
[rtk] rubocop (JSON parse error): output format not recognized, showing last 5 lines
{"version":"1.86.2"}
```

This PR detects the minimal `{"version":"..."}` JSON shape emitted by
`--version` queries before attempting the full lint/test report parse and
surfaces it as a clean `rubocop 1.86.2` / `rspec 3.13.2` line.

## Root cause

Both modules inject `--format json` to get structured output. When the user
runs `--version`, the tool emits `{"version":"..."}` instead of a full report.
The strongly-typed struct (`RubocopOutput { files, summary }` /
`RspecOutput { examples, summary }`) rejects it, the JSON parse error is
logged, and `fallback_tail` adds a second "output format not recognized"
warning before printing the raw line.

## Fix

In `filter_rubocop_json` and `filter_rspec_output`, run a lightweight
detector first:

- Parse as `serde_json::Value`.
- Verify it's an object with a string `version` key.
- Bail out if any report-specific key is present (`files`/`summary`/`metadata`
  for rubocop, `examples`/`summary`/`summary_line` for rspec) — that means
  it's a real report, not a `--version` query.
- Otherwise return `<tool> <version>` cleanly with no warnings.

Real reports that happen to carry a `version` field (which RSpec's JSON
formatter does) still flow through the existing report parser; covered by
regression tests.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo clippy --all-targets` (zero warnings)
- [x] `cargo test --bin rtk` — 1909 passed, 0 failed
- [x] New tests (8 total, 4 per module):
  - `test_filter_rubocop_version_only_json_clean`
  - `test_filter_rubocop_version_only_json_with_whitespace`
  - `test_filter_rubocop_version_only_does_not_swallow_real_report`
  - `test_try_version_only_json_rejects_non_object`
  - `test_filter_rspec_version_only_json_clean`
  - `test_filter_rspec_version_only_json_with_whitespace`
  - `test_filter_rspec_version_only_does_not_swallow_real_report`
  - `test_try_version_only_json_rspec_rejects_non_object`

## Files changed

[...]

PR #2212, opened 2026-06-02
fix(ruby): route rspec/rubocop version queries to passthrough

### LINKED ISSUE #1946: rubocop/rspec --version parser rejects pure version JSON
## Summary
When rubocop or rspec outputs a minimal JSON response for --version (e.g. `{"version":"1.86.2"}`), RTK's parser rejects it because it expects a full struct containing `files` (rubocop) or `examples` (rspec) fields. The version IS correctly extracted but the parse warning is noisy.

## Environment
- OS: Windows 10/11
- Shell: Git Bash (MINGW64) + PowerShell
- RTK version: 0.40.0
- rubocop version: 1.86.2
- rspec version: 3.13.2

## Steps to Reproduce

```bash
# We created a .cmd wrapper that intercepts --version and outputs clean JSON:
# rubocop wrapper --version → {"version":"1.86.2"}
# rspec wrapper --version → {"version":"3.13.2"}

$ rtk rubocop --version
[rtk] rubocop: JSON parse failed (missing field `files` at line 1 column 20)
[rtk] rubocop (JSON parse error): output format not recognized, showing last 5 lines
{"version":"1.86.2"}

$ rtk rspec --version
[rtk] rspec: JSON parse failed (missing field `examples` at line 1 column 20), using text fallback
[rtk] rspec: output format not recognized, showing last 5 lines
{"version":"3.13.2"}
```

## Root Cause
RTK's rubocop/rspec parsers validate the JSON response against a full struct:

**rubocop**: Expects output like `{"files": [...], "summary": {...}}`. The parser rejects a valid `{"version":"1.86.2"}` object because `files` is missing.

**rspec**: Expects output like `{"examples": [...], "summary": {...}}`. The parser rejects a valid `{"version":"3.13.2"}` object because `examples` is missing.

## Expected Behavior
When `--version` is passed as the only argument, RTK should accept a simple `{"version":"..."}` JSON response. The full struct format only makes sense for lint/test runs, not for version queries.

## Workaround
The version IS correctly output (the JSON shows in the "last 5 lines" fallback). This is cosmetic but noisy -- every --version call triggers the warning.

## Related
- #1506 (rspec JSON parse on text output -- different symptom, same parser area)

SeanLF:
## Summary

`rtk rspec --version` and `rtk rubocop --version` emit two spurious warnings on every call:

```console
$ rtk rubocop --version
[rtk] rubocop: JSON parse failed (invalid type: floating point `1.87`, expected struct RubocopOutput at line 1 column 4)
[rtk] rubocop (JSON parse error): output format not recognized, showing last 5 lines
1.87.0
```

The rspec/rubocop filters inject `--format json` and deserialize stdout into a strict report struct. But a **version query short-circuits inside the tool before any formatter runs**, so it always prints plain text — `rubocop --version` → bare `1.87.0` (read as a JSON float, hence `invalid type: floating point`), `rspec --version` → a multi-line version block. The strict parse rejects it and the fallback adds a second warning.

Fix: detect version flags in `run()` and route them to **passthrough** — no `--format json` injection, no report parse — mirroring how the Python filters (ruff/pip) route by subcommand before deserializing.

```console
$ rtk rubocop --version   →  1.87.0
$ rtk rubocop -V          →  1.87.0 (using Parser 3.3.11.1, ...) [arm64-darwin25]
$ rtk rspec --version     →  RSpec 3.13\n  - rspec-core 3.13.6 ...
```

Reproduced and verified on rtk 0.42.0, macOS, rspec 3.13.6, rubocop 1.87.0 (unwrapped tools).

## Relationship to #1982

This **supersedes #1982** (thanks @YOMXXX for surfacing the bug). That PR detects the minimal `{"version":"..."}` JSON shape and prettifies it. But that shape only appears behind the reporter's custom `.cmd` wrapper — real `rubocop`/`rspec` never emit it, because `--version` bypasses the formatter entirely. So a JSON-shape check fixes the wrapped case but leaves the warning firing for everyone running the tools normally. Routing at the invocation layer is a superset: wrapped JSON passes through raw, unwrapped text passes through clean. It's also a smaller change (+57/-5 in the modules vs +174/-0).

## Testing

- 2 new unit tests (`is_version_query` on both tools, incl. negative cases for `--format`/`-a`); existing 29 rspec / 19 rubocop tests green.
- `scripts/test-ruby.sh`: new "Version queries (#1946)" section with a reusable `assert_not_contains` regression guard asserting no JSON-parse warning leaks. Verified passing against a local build.
- `cargo fmt --all -- --check` and `cargo clippy` clean on the touched modules.

## Notes

- Version-flag set confirmed against `rspec --help` / `rubocop --help`: rspec `-v`/`--version`; rubocop adds `-V`/`--verbose-version`.
- The passthrough is `stdout.trim_end()` — infallible and fully isolated from the report-parse path, which is unchanged.

Fixes #1946.

### DIFF
--- a/scripts/test-ruby.sh
+++ b/scripts/test-ruby.sh
@@ -70,6 +70,23 @@ assert_output() {
     fi
 }
 
+# Assert command output does NOT contain needle (regression guard, e.g. warnings)
+assert_not_contains() {
+    local name="$1"; local needle="$2"; shift 2
+    local output
+    output=$("$@" 2>&1) || true
+    if echo "$output" | grep -qi "$needle"; then
+        FAIL=$((FAIL + 1))
+        FAILURES+=("$name")
+        printf "  ${RED}FAIL${NC}  %s\n" "$name"
+        printf "        unexpected: '%s'\n" "$needle"
+        printf "        got: %s\n" "$(echo "$output" | head -3)"
+    else
+        PASS=$((PASS + 1))
+        printf "  ${GREEN}PASS${NC}  %s\n" "$name"
+    fi
+}
+
 skip_test() {
     local name="$1"; local reason="$2"
     SKIP=$((SKIP + 1))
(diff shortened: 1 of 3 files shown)


SeanLF:
bump

PR #2222, opened 2026-06-02
fix(grep): skip .claude/worktrees directory

### LINKED ISSUE #2147: rtk grep should ignore .claude/worktrees

mikeangstadt:
## Summary

`rtk grep` searches Claude Code worktree directories (`.claude/worktrees/`) which contain session-local git worktree snapshots. These are out of scope for an agentic session and contain stale or duplicated data that pollutes results.

## Fix

- **ripgrep path:** added `--glob '!.claude/worktrees/**'` to exclude the directory
- **fallback grep path:** added `--exclude-dir .claude/worktrees`

Closes #2147

---
Collaborated & orchestrated by [ClosedLoop.ai](https://closedloop.ai) | [GitHub](https://github.com/closedloop-ai)

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -37,6 +37,10 @@ pub fn run(
     // -0: NUL-separate filename. Allows the parser to disambiguate filenames or
     // content containing `:digits:` patterns (issue #1436).
     rg_cmd.args(["-nH0", "--no-heading", "--no-ignore-vcs", &rg_pattern, path]);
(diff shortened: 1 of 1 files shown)


KuSh:

PR #2223, opened 2026-06-02
fix(lint): detect oxlint, include message body in ESLint summary

mikeangstadt:
-

---

**Collaborated & orchestrated by [ClosedLoop.AI](https://closedloop.ai) | [GitHub](https://github.com/closedloop-ai)**

### DIFF
--- a/src/cmds/js/lint_cmd.rs
+++ b/src/cmds/js/lint_cmd.rs
@@ -65,14 +65,23 @@ fn detect_js_linter_from_config() -> Option<&'static str> {
     let cwd = std::env::current_dir().ok()?;
 
     for candidate in &[
+        ("eslint.config.js", "eslint"),
+        ("eslint.config.mjs", "eslint"),
+        ("eslint.config.cjs", "eslint"),
+        ("eslint.config.ts", "eslint"),
+        ("eslint.config.mts", "eslint"),
+        ("eslint.config.cts", "eslint"),
         (".eslintrc.js", "eslint"),
         (".eslintrc.cjs", "eslint"),
         (".eslintrc.yaml", "eslint"),
         (".eslintrc.yml", "eslint"),
         (".eslintrc.json", "eslint"),
         (".eslintrc", "eslint"),
-        ("oxlint.json", "oxlint"),
+        (".oxlintrc.json", "oxlint"),
+        (".oxlintrc.jsonc", "oxlint"),
+        ("oxlint.config.ts", "oxlint"),
         ("biome.json", "biome"),
+        ("biome.jsonc", "biome"),
     ] {
         if Path::new(&cwd).join(candidate.0).exists() {
             return Some(candidate.1);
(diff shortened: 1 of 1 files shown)


src/cmds/js/lint_cmd.rs
```diff
@@ -57,6 +57,30 @@ fn is_python_linter(linter: &str) -> bool {
...
+    let cwd = std::env::current_dir().ok()?;
+
+    for candidate in &[
+        (".eslintrc.js", "eslint"),
+        (".eslintrc.cjs", "eslint"),
+        (".eslintrc.yaml", "eslint"),
+        (".eslintrc.yml", "eslint"),
+        (".eslintrc.json", "eslint"),
+        (".eslintrc", "eslint"),
+        ("oxlint.json", "oxlint"),
```
eai04191: By default, oxlint reads a config file with the following name: it's not `oxlint.json`.

- `.oxlintrc.json`
- `.oxlintrc.jsonc`
- `oxlint.config.ts`

ref: https://oxc.rs/docs/guide/usage/linter/cli

src/cmds/js/lint_cmd.rs
```diff
@@ -57,6 +57,30 @@ fn is_python_linter(linter: &str) -> bool {
...
+        ("biome.json", "biome"),
```
eai04191, line 83: biome's config file can also be named `biome.jsonc`, not just `biome.json`.

ref: https://biomejs.dev/guides/configure-biome/

src/cmds/js/lint_cmd.rs
```diff
@@ -57,6 +57,30 @@ fn is_python_linter(linter: &str) -> bool {
...
+
+    let cwd = std::env::current_dir().ok()?;
+
+    for candidate in &[
+        (".eslintrc.js", "eslint"),
+        (".eslintrc.cjs", "eslint"),
+        (".eslintrc.yaml", "eslint"),
+        (".eslintrc.yml", "eslint"),
+        (".eslintrc.json", "eslint"),
+        (".eslintrc", "eslint"),
```
eai04191, line 79: The current candidates only cover the legacy `.eslintrc.*` files. Since ESLint v9, flat config is the default, and many projects now use `eslint.config.js` / `.mjs` / `.cjs` / `.ts` / `.mts` / `.cts`. These would be missed by the current detection.

ref: https://eslint.org/docs/latest/use/configure/configuration-files

src/cmds/js/lint_cmd.rs
mikeangstadt: Good catch — fixed. Replaced `oxlint.json` with the correct defaults: `.oxlintrc.json`, `.oxlintrc.jsonc`, and `oxlint.config.ts`.

src/cmds/js/lint_cmd.rs
mikeangstadt: Fixed — added `biome.jsonc` alongside `biome.json`.

src/cmds/js/lint_cmd.rs
mikeangstadt: Fixed — added all ESLint v9+ flat config candidates (`eslint.config.{js,mjs,cjs,ts,mts,cts}`) with priority over legacy `.eslintrc.*` files so that projects with both detect flat config correctly.

PR #2224, opened 2026-06-02
fix(grep): strip ripgrep-only flags when falling back to system grep

### LINKED ISSUE #2167: rtk grep calls system grep instead of ripgrep — breaks -g/--glob flags

mikeangstadt:
## Summary

When ripgrep is not on `$PATH`, `rtk grep` falls back to system `grep` but was passing ripgrep-specific flags (e.g. `-g`/`--glob`, `--hidden`, `--pcre2`, `--json`, `--sort`) directly to `grep`, which does not understand them. This causes silent failures or wrong results (e.g. `grep: invalid option -- g`).

## Changes

- Added `RG_ONLY_FLAGS` constant table of ripgrep-only flags with their value-taking semantics
- Added `filter_rg_flags()` that strips rg-only flags (and their value arguments) from `extra_args` before passing to the grep fallback
- Handles both standalone flags (`--json`) and value-taking flags (`-g *.ts`, `--glob=*.ts`)
- The main rg path is unaffected; only the `.or_else(|_| ...)` fallback is changed

## Testing

- Manual review of the filter logic:
  - `extra_args = ["-g", "*.rs", "-i"]` -> `["-i"]` (skips `-g` and its value `*.rs`)
  - `extra_args = ["--glob=*.rs", "-i"]` -> `["-i"]` (embedded value skipped)
  - `extra_args = ["--json", "--stats", "-i"]` -> `["-i"]`
  - `extra_args = ["--hidden", "--no-ignore", "-n"]` -> `["-n"]`
  - No rg-only flags -> all args pass through unchanged

Closes #2167

---

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -8,6 +8,65 @@ use anyhow::{Context, Result};
 use regex::Regex;
 use std::collections::HashMap;
 
+/// Flags that ripgrep supports but system grep does not.
+/// Each entry: (prefix match string, takes_value_arg, keep_as_grep_flag)
+const RG_ONLY_FLAGS: &[(&str, bool, &str)] = &[
+    ("--glob", true, ""),
+    ("--iglob", true, ""),
+    ("--type-not", true, ""),
+    ("--sort", true, ""),
+    ("--field-match-separator", true, ""),
+    ("--path-separator", true, ""),
+    ("--no-ignore", false, ""),
+    ("--hidden", false, ""),
+    ("--pcre2", false, ""),
+    ("--json", false, ""),
+    ("--stats", false, ""),
+    ("--mmap", false, ""),
+    ("--one-file-system", false, ""),
+    ("--trim", false, ""),
+];
+
+/// Returns true if the arg is a ripgrep-only flag (not supported by grep).
+fn is_rg_only_flag(arg: &str) -> bool {
+    RG_ONLY_FLAGS.iter().any(|(prefix, _, _)| {
+        arg.starts_with(prefix) || arg == "-g" || arg.starts_with("-g=") || arg == "-T"
+    })
+}
+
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the detailed flag table. This was overtaken twice: #2550 (merged 2026-06-23) folded rg-only flag stripping into its own fallback handling, and #2641 (merged 2026-06-26) then removed the fallback altogether by making `rtk grep` run real grep and adding `rtk rg` for ripgrep, with the hook rewriting `rg ...` accordingly. The file you patched, src/cmds/system/grep_cmd.rs, no longer exists on develop, and issue #2167 was closed as fixed by that work. If you still see ripgrep flags reaching system grep on the current develop, please comment here or reopen with the command.

PR #2225, opened 2026-06-02
fix(proxy): reject compound shell snippets with clear error message

### LINKED ISSUE #2163: rtk proxy should reject or safely handle compound shell snippets
## Summary

`rtk proxy` appears unsafe for compound shell snippets. When I wrapped a multi-command shell loop in `rtk proxy '...'`, shell syntax was not preserved as expected and parts of the snippet were interpreted as command/path arguments. This led to incorrect execution side effects, including bogus directories being created from shell tokens.

## Version

```text
rtk 0.36.0
```

## Reproduction shape

A command shaped like this is enough to demonstrate the risk pattern:

```bash
rtk proxy 'mkdir -p /tmp/rtk-repro && for n in 1 2; do log=/tmp/rtk-repro/$n.log; (echo "$n" > "$log"; echo $? > /tmp/rtk-repro/$n.exit) & echo "pid=$!"; sleep 1; while [ "$(jobs -pr | wc -l | tr -d " ")" -ge 2 ]; do sleep 1; done; done; wait'
```

The important part is not this exact command, but that it contains normal shell syntax:

- `&&`
- `for ... do ... done`
- command substitution: `$(...)`
- pipes: `|`
- redirects: `>` / `2>&1`
- background jobs: `&`
- grouped commands: `( ... )`
- nested quotes

## Expected behavior

Either:

1. `rtk proxy` should pass the full snippet to the shell exactly as a raw command, preserving shell semantics; or
2. `rtk proxy` should reject compound shell snippets with a clear error message, e.g. "rtk proxy only supports one simple command; do not pass shell snippets".

## Actual behavior

The compound snippet was not treated as a single shell program. Parts of the shell syntax were effectively interpreted as command/path tokens, producing incorrect behavior. In my case, token-like fragments such as option names, numeric loop values, and shell keywords became filesystem entries.

## Why this matters

The name and help text for `proxy` can make users think it is safe to wrap arbitrary raw shell code. If it is intended only for simple commands, the CLI should make that explicit and ideally fail closed for detected shell metacharacters.

## Suggested fix

- Document that `rtk proxy` is only for a single simple command, not shell snippets.
- Add validation/rejection for common shell metacharacters or compound syntax (`|`, `>`, `&&`, `;`, `$()`, `for`, `while`, `if`, `&`, parentheses) unless an explicit shell mode is supported.
- Alternatively, provide a separate explicit mode such as `rtk shell -- '<snippet>'` that intentionally invokes the user's shell and preserves compound shell semantics.

mikeangstadt:
## Summary

`rtk proxy` does not invoke a shell — it spawns the command directly via `exec`. When users pass compound shell snippets (with `&&`, `||`, `|`, `;`, `$()`, `${}`, backticks, `&`), shell metacharacters are not interpreted and instead become literal arguments to the spawned process, causing confusion or incorrect behavior (e.g. bogus files/dirs created from shell tokens).

## Changes

- Added `detect_shell_snippet()` function that checks the command name and all args for:
  - Standalone shell control operators: `&&`, `||`, `;`, `|`, `&`
  - Command substitution embedded in any token: `$(`, `${`, `` ` ``
- If detected, `rtk proxy` rejects with a clear error message and suggests using `sh -c '...'` or `bash -c '...'` instead
- The validation runs after `shell_split` (for the single-arg path) and after cmd/args extraction (for the multi-arg path), so both usage patterns are covered

## Examples of rejected commands

- `rtk proxy 'mkdir -p /tmp && echo done'` -> rejected: contains `&&`
- `rtk proxy 'for n in 1 2; do echo $n; done'` -> rejected: contains `;`
- `rtk proxy 'echo $(whoami)'` -> rejected: contains `$(`
- `rtk proxy 'cat file | grep pattern'` -> rejected: contains `|`

## Commands that still work

Simple single commands pass through as before:
- `rtk proxy git status`
- `rtk proxy echo hello world`
- `rtk proxy 'head -50 file.php'`

Closes #2163

---

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -8,6 +8,65 @@ use anyhow::{Context, Result};
 use regex::Regex;
 use std::collections::HashMap;
 
+/// Flags that ripgrep supports but system grep does not.
+/// Each entry: (prefix match string, takes_value_arg, keep_as_grep_flag)
+const RG_ONLY_FLAGS: &[(&str, bool, &str)] = &[
+    ("--glob", true, ""),
+    ("--iglob", true, ""),
+    ("--type-not", true, ""),
+    ("--sort", true, ""),
+    ("--field-match-separator", true, ""),
+    ("--path-separator", true, ""),
+    ("--no-ignore", false, ""),
+    ("--hidden", false, ""),
+    ("--pcre2", false, ""),
+    ("--json", false, ""),
+    ("--stats", false, ""),
+    ("--mmap", false, ""),
+    ("--one-file-system", false, ""),
+    ("--trim", false, ""),
+];
+
+/// Returns true if the arg is a ripgrep-only flag (not supported by grep).
+fn is_rg_only_flag(arg: &str) -> bool {
+    RG_ONLY_FLAGS.iter().any(|(prefix, _, _)| {
+        arg.starts_with(prefix) || arg == "-g" || arg.starts_with("-g=") || arg == "-T"
+    })
+}
+
+/// Returns true if the flag takes a separate value argument (next token to skip).
+fn flag_takes_value(arg: &str) -> bool {
+    // Value embedded with = means the next token is not its value.
+    if arg.contains('=') {
+        return false;
+    }
+    RG_ONLY_FLAGS
+        .iter()
+        .any(|(prefix, takes_val, _)| takes_val && arg == *prefix)
+        || arg == "-g"
+        || arg == "-T"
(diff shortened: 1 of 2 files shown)


KuSh:
Closing in favour of #2165, which adds the same proxy guard for #2163 but keeps quoted metacharacters as legitimate arguments and covers it with unit tests including the reproducer from the issue. The ripgrep flag filtering in src/cmds/system/grep_cmd.rs here is a separate change and would be welcome as its own PR.

PR #2226, opened 2026-06-02
fix(rewrite): apply custom TOML filters during command rewrite

### LINKED ISSUE #2179: Custom TOML filters (global or trusted project-local) are never applied by the hook
**rtk v0.37.2, macOS (Apple Silicon, homebrew)**

Custom TOML filters — defined in the global `filters.toml` **or** a `rtk trust`'d project-local `.rtk/filters.toml` — parse and pass `rtk verify`, but are **never applied by the Claude Code hook**. Only compiled-in/bundled filters get rewritten.

### Repro
Add a custom filter (global or trusted project-local):
```toml
[filters.zzprobe]
description = "probe"
match_command = '^zzprobe\b'
max_lines = 5
```
```
$ rtk trust            # registers .rtk/filters.toml (confirmed in `rtk trust --list`)
$ rtk verify --filter zzprobe     # found, parses (No inline tests)
$ rtk rewrite 'zzprobe --foo'      # -> EMPTY (no rewrite)
$ rtk rewrite 'liquibase update'   # -> "rtk liquibase update"  (bundled filter works)
$ rtk rewrite 'df -h'              # -> "rtk df -h"             (works)
```
`RTK_NO_TOML` is unset. A real custom filter (`match_command='iconv.*UTF-16'` + a `replace` rule) likewise never fires on `iconv ...` commands, while it's trivially correct when applied manually.

### Also (related, may be by-design)
`rtk discover` scans the CC transcript (the raw pre-hook command string) and lists e.g. `grep -n` as "missed -> rtk grep (existing)". But the hook already rewrites those at exec time, so `discover`'s "missed M tokens" / "adoption %" double-counts already-captured savings (verified: raw `cat`/`grep` increment the `rtk read`/`rtk grep` counters 1:1). Consider netting out hook-captured commands using history.db so "missed" reflects only genuinely-unhandled commands.

### Expected
Trusted/global custom filters should be applied by the hook like bundled ones (or `rtk pipe -f <customname>` should accept custom filter names — it currently rejects anything not built-in). Without this, user-defined filters are non-functional for their primary purpose.

mikeangstadt:
## Summary

Custom TOML filters (from `.rtk/filters.toml` or `~/.config/rtk/filters.toml`) were parsed and passed `rtk verify`, but were **never consulted by the hook rewrite path**. The rewrite engine (`discover/registry.rs`) only checked the hardcoded `RULES` array, returning empty for any unrecognized command. This meant user-defined TOML filters were non-functional for their primary purpose.

## Root Cause

The `rtk rewrite` path (used by hooks) has **no knowledge of TOML filters**. There are two independent systems:

| System | Data Source | Used By | Knows Custom Filters? |
|---|---|---|---|
| Rewrite Rules | `discover/rules.rs` (hardcoded) | `rtk rewrite`, hook | NO |
| TOML Filter Engine | `.rtk/filters.toml` / `~/.config/rtk/filters.toml` / built-in | `rtk verify`, `run_fallback` | YES |

When a user runs a command matching a custom TOML filter (e.g. `zzprobe --foo`):
1. Hook calls `rtk rewrite 'zzprobe --foo'` -> hardcoded rules don't match -> returns empty -> hook passes through unchanged
2. Command runs **without** `rtk` prefix -> TOML filter is **never applied** during execution

## Change

In `rewrite_segment_inner()` in `discover/registry.rs`: when `classify_command()` returns `Unsupported`, check the `TomlFilterRegistry` before giving up. If a matching custom TOML filter is found, return `Some(format!("rtk {}/{}", cmd_part, redirect_suffix))` -- which causes the hook to run the command through `rtk`, where `run_fallback()` applies the TOML filter pipeline.

This is a minimal change: 10 lines added, 1 modified. Only the `Unsupported` branch is affected -- built-in rules, `Ignored` commands, and excluded commands are unchanged.

Closes #2179

---

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -8,6 +8,65 @@ use anyhow::{Context, Result};
 use regex::Regex;
 use std::collections::HashMap;
 
+/// Flags that ripgrep supports but system grep does not.
+/// Each entry: (prefix match string, takes_value_arg, keep_as_grep_flag)
+const RG_ONLY_FLAGS: &[(&str, bool, &str)] = &[
+    ("--glob", true, ""),
+    ("--iglob", true, ""),
+    ("--type-not", true, ""),
+    ("--sort", true, ""),
+    ("--field-match-separator", true, ""),
+    ("--path-separator", true, ""),
+    ("--no-ignore", false, ""),
+    ("--hidden", false, ""),
+    ("--pcre2", false, ""),
+    ("--json", false, ""),
+    ("--stats", false, ""),
+    ("--mmap", false, ""),
+    ("--one-file-system", false, ""),
+    ("--trim", false, ""),
+];
+
+/// Returns true if the arg is a ripgrep-only flag (not supported by grep).
+fn is_rg_only_flag(arg: &str) -> bool {
+    RG_ONLY_FLAGS.iter().any(|(prefix, _, _)| {
+        arg.starts_with(prefix) || arg == "-g" || arg.starts_with("-g=") || arg == "-T"
+    })
+}
+
+/// Returns true if the flag takes a separate value argument (next token to skip).
+fn flag_takes_value(arg: &str) -> bool {
+    // Value embedded with = means the next token is not its value.
+    if arg.contains('=') {
+        return false;
+    }
+    RG_ONLY_FLAGS
+        .iter()
+        .any(|(prefix, takes_val, _)| takes_val && arg == *prefix)
+        || arg == "-g"
+        || arg == "-T"
+}
+
+/// Filter out ripgrep-specific flags from a list of args, handling value-taking flags.
+fn filter_rg_flags(args: &[String]) -> Vec<&str> {
+    let mut skip_next = false;
+    let mut result = Vec::new();
+    for arg in args {
+        if skip_next {
+            skip_next = false;
+            continue;
+        }
+        if is_rg_only_flag(arg) {
+            if flag_takes_value(arg) {
+                skip_next = true;
+            }
+            continue;
+        }
+        result.push(arg.as_str());
+    }
+    result
+}
+
 #[allow(clippy::too_many_arguments)]
 pub fn run(
     pattern: &str,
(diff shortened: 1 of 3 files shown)


aeppling:
Hey @mikeangstadt 

Thanks for addressing this, will enhance all gains for RTK toml users

There are some blockers, here are some problems i've identified

------------

### The PR does not compile

Check ci 

------------

### Exclude-list bypass (regression)

The Supported arm honors the user's exclude_commands (registry.rs:813); the new Unsupported+TOML arm does not. I verified is_excluded has exactly one call site and there's no upstream gate, then reproduced it with a unit test on the actual PR code:

test repro_control_gcc_rewritten_by_toml ... ok          # feature works: gcc → "rtk gcc -o x x.c"
test repro_excluded_command_bypasses_exclude_list ... FAILED
    assertion failed: got Some("rtk gcc -o x x.c"), right: None

With exclude_commands = ["gcc"], the PR still force-wraps the command. A user who excluded a command gets it rewritten anyway whenever any of the 58 built-in filters match. Fix: mirror the is_excluded check (with ENV_PREFIX strip) in the new arm.

------------

###  Needs tests for the new behavior

This is a behavioral change and we need solid tests to ensure no futures regressions.

------------

### Notes

Fix is tested and work well, please drop unrelated grep changes, this belong to another PR. I thing you embedded this maybe by wrong base on one of your others PR, please test the code and the fix for the next review so there is no back and forth for compiling issues for example.

Thanks for this contribution , awaiting for your changes !

KuSh:
Thanks for the clear write up of the two-systems problem. Issue #2179 was fixed on develop by #2748 (merged 2026-07-03): the rewrite path now consults the TOML filter registry in the Unsupported branch, with extra guards for a disabled TOML engine and for rtk's own reserved commands, and a match only RegexSet so the hook stays fast. The grep hunk on this branch targets src/cmds/system/grep_cmd.rs, which #2641 (merged 2026-06-26) removed when rtk stopped substituting rg for grep, so it cannot be carried over. The proxy case is still open: rtk proxy spawns directly, so &&, |, ; and $() reach the child as literal arguments, and an issue or a small PR for that would be welcome, ideally one that cannot reject a literal ${ or backtick inside a quoted argument. Please comment or reopen if I missed a part of this.

PR #2229, opened 2026-06-02
fix: bug(rewrite): hook rewrites left-hand side of piped com...

### LINKED ISSUE #1560: bug(rewrite): hook rewrites left-hand side of piped commands, breaking downstream filters

gittihub-jpg:
Closes #1560

## Summary
- Prevents `rtk rewrite` from rewriting only the left-hand side of piped commands or subshells.
- Adds a guard clause in `src/hooks/rewrite_cmd.rs` to skip rewriting when `|` or `$()` is present.
- Restores expected behavior for downstream filters like `grep`, `tail`, and `head`.

## Test plan
- [ ] Manual testing: `rtk rewrite 'ps aux | grep python | grep -v grep'` output inspected

### DIFF
--- a/src/hooks/rewrite_cmd.rs
+++ b/src/hooks/rewrite_cmd.rs
@@ -4,6 +4,12 @@ use super::permissions::{check_command, PermissionVerdict};
 use crate::discover::registry;
 use std::io::Write;
 
+/// Returns true if the command contains pipe or subshell operators that should
+/// bypass rewrite logic and trigger passthrough behavior.
+fn needs_passthrough(cmd: &str) -> bool {
+    cmd.contains('|') || cmd.contains("$(")
+}
+
 /// Run the `rtk rewrite` command.
 ///
 /// Prints the RTK-rewritten command to stdout and exits with a code that tells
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into this. Issue #1560 was fixed on develop by #3128 (merged as ca418939), which stops the rewriter from touching any stage that feeds a pipe and rewrites the final stage only when its filter is safe to read from stdin. I ran the examples from this PR against the current develop binary: `git log | head -5`, `cargo build 2>&1 | tee build.log` and every `$(...)` form pass through untouched, while `ls | grep foo` becomes `ls | rtk grep foo`, so the producer is preserved and the useful rewrite is kept. The blanket bypass here would also return exit 0 with the original command, which the hook protocol reads as auto-allow, so piped commands would skip the deny and ask checks. Closing as covered, please comment or reopen if you still see a bad rewrite on the latest develop.

PR #2232, opened 2026-06-03
fix(tsc): handle pretty diagnostics

### LINKED ISSUE #1772: tsc wrapper: --pretty flag suppresses errors and prints fake "compilation completed" success message
## Summary

When `rtk tsc` runs with the `--pretty` flag, type errors are silently swallowed and RTK reports `TypeScript: No errors found` even though `tsc` exits with a non-zero status. Without `--pretty`, RTK parses errors correctly. The bug is reliably reproducible.

## Affected version

`rtk 0.39.0` (latest at time of filing, installed via Homebrew on macOS arm64).

## Reproduction

`/tmp/rtk-test.ts`:

```ts
const x: number = "string";
```

`/tmp/tsconfig-rtk-test.json`:

```json
{
  "compilerOptions": {
    "strict": true,
    "noEmit": true,
    "target": "ES2020",
    "module": "ESNext",
    "moduleResolution": "node"
  },
  "files": ["/tmp/rtk-test.ts"]
}
```

Run both:

```
$ rtk tsc --noEmit -p /tmp/tsconfig-rtk-test.json
../../../../tmp/rtk-test.ts(1,7): error TS2322: Type 'string' is not assignable to type 'number'.
TypeScript: 1 errors in 1 files
# exit code: 2

$ rtk tsc --noEmit --pretty -p /tmp/tsconfig-rtk-test.json
TypeScript: No errors found
# exit code: 2
```

## Expected vs actual

- **Expected**: With `--pretty`, RTK should still display the errors (or at least not lie) — minimum bar is that the textual summary matches reality when `tsc` exits non-zero.
- **Actual**: Exit code is correctly propagated (2), but stdout reports `No errors found`. A consumer looking at output (human or LLM) sees a fake success message.

This is dangerous in CI/agent contexts: errors get hidden by a wrapper that claims success, and downstream steps (e.g. `&& next build`) may proceed as if green. Concretely, this was discovered while running `npx tsc --noEmit --pretty` from a Claude Code session — the wrapper hid type errors that Vercel's build then caught, causing a failed deploy.

## Root cause hypothesis

`tsc --pretty` emits ANSI escape codes (colors, formatting) on each error line. RTK's regex for matching tsc error lines (`path(L,C): error TSxxxx: ...`) does not account for embedded ANSI sequences, so it counts zero matches. With zero parsed errors, RTK synthesizes `No errors found` regardless of the actual exit code.

## Suggested fix

Either (or both):

1. Strip ANSI escape sequences from each line before regex-matching error lines.
2. Cross-check parsed error count against the child exit code: if `exit != 0 && parsed_errors == 0`, fall back to passthrough of the raw output rather than synthesizing a success summary. This is a good general invariant — RTK's summary should never contradict the wrapped tool's exit code.

Option 2 is the safer default since it also catches future tsc output format changes that the regex hasn't adapted to.

### LINKED ISSUE #3220: tsc filter: "TypeScript: No errors found" is printed on a non-zero exit, so a tsc that never ran reads as a clean typecheck
## Summary

```
TypeScript: No errors found
```

## Affected version

## Reproduction

```sh
D=$(mktemp -d); cd "$D"
echo '{}' > tsconfig.json

```

Result:

```
--- raw ---
raw exit=1

--- rtk ---
TypeScript: No errors found
rtk exit=1
```

## Expected vs actual

| | |
|---|---|

## Relationship to #1772

## Workaround

```toml
[hooks]
```

jsiu93:
## Summary
- Strip ANSI before matching TypeScript diagnostic lines.
- Parse `tsc --pretty` diagnostics in `file:line:col - error TSxxxx` form.
- Avoid reporting `TypeScript: No errors found` when the compiler exits non-zero and RTK parsed no diagnostics.

## Test plan
- `cargo fmt --all --check`
- `cargo test --bin rtk -- cmds::js::tsc_cmd`
- `cargo test --bin rtk`
- `cargo run --quiet --bin rtk -- tsc --noEmit --pretty -p /tmp/rtk-1772/tsconfig.json` exits 2 and reports `TS2322` plus `TypeScript: 1 errors in 1 files`

Fixes #1772

Fixes #3220

Partially addresses #3738 (ANSI escapes are stripped on the `tsc` path only).

### DIFF
--- a/src/cmds/js/tsc_cmd.rs
+++ b/src/cmds/js/tsc_cmd.rs
@@ -7,13 +7,16 @@ use crate::core::utils::{resolved_command, strip_ansi, tool_exists, truncate};
 use anyhow::Result;
 use regex::Regex;
 use std::borrow::Cow;
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::LazyLock;
 
-/// Unparseable failure output is limited to this many non-empty lines. The
-/// runner tees the full raw output and prints a recovery hint, except when tee
-/// is disabled (`RTK_TEE=0` or `config.tee.enabled = false`): then these lines
-/// are the only surviving copy, which is why both ends are kept.
+/// Unparseable failure output is limited to this many non-empty lines, split
+/// between head and tail. The runner tees the full raw output and prints a
+/// recovery hint, so the hidden middle stays reachable, with two exceptions the
+/// cap accounts for: output under `MIN_TEE_SIZE` bytes gets no tee, so it is
+/// printed whole; and with tee disabled (`RTK_TEE=0` or
+/// `config.tee.enabled = false`) these lines are the only surviving copy, which
+/// is why both ends are kept.
 const MAX_UNPARSED_LINES: usize = CAP_WARNINGS;
 /// Deviation: tsc and npx print the cause first (`Unknown compiler option`,
 /// `This is not the tsc command`) and boilerplate after it, so a plain tail
@@ -62,6 +65,12 @@ fn parse_diagnostic(line: &str) -> Option<Diagnostic<'_>> {
         code: caps.get(2)?.as_str(),
         message: caps.get(3)?.as_str(),
     })
+}
+
+/// One line of the raw failure dump: width-capped like every other emission.
+fn push_dump_line(summary: &mut String, line: &str) {
+    summary.push_str(&truncate(line, 120));
+    summary.push('\n');
 }
 
 fn clean_line(line: &str) -> Cow<'_, str> {
@@ -158,30 +167,55 @@ impl BlockHandler for TscHandler {
             // no project, unparseable output). "No errors found" would be a
             // false green, so report the failure with the head and tail of the
             // raw output; only the retained lines are ANSI-stripped.
-            let lines: Vec<&str> = raw.lines().filter(|line| !line.trim().is_empty()).collect();
             let mut summary = format!(
                 "TypeScript: compiler exited with code {exit_code}, but RTK parsed no diagnostics\n"
             );
 
-            if lines.len() <= MAX_UNPARSED_LINES {
-                for line in lines {
-                    summary.push_str(clean_line(line).as_ref());
-                    summary.push('\n');
+            if raw.len() < crate::core::tee::MIN_TEE_SIZE {
+                for line in raw.lines() {
+                    let line = clean_line(line);
+                    if line.trim().is_empty() {
+                        continue;
+                    }
+                    push_dump_line(&mut summary, line.as_ref());
                 }
                 return Some(summary);
             }
 
             let head_len = MAX_UNPARSED_HEAD_LINES.min(MAX_UNPARSED_LINES);
             let tail_len = MAX_UNPARSED_LINES.saturating_sub(head_len);
-            let hidden = lines.len() - head_len - tail_len;
-            for line in &lines[..head_len] {
-                summary.push_str(clean_line(line).as_ref());
-                summary.push('\n');
+            let mut head: Vec<Cow<'_, str>> = Vec::with_capacity(head_len);
+            let mut tail: VecDeque<Cow<'_, str>> = VecDeque::with_capacity(tail_len);
+            let mut total = 0;
+
+            for line in raw.lines() {
+                let line = clean_line(line);
+                if line.trim().is_empty() {
+                    continue;
+                }
+                total += 1;
+                if head.len() < head_len {
+                    head.push(line);
+                    continue;
+                }
+                if tail_len == 0 {
+                    continue;
+                }
+                if tail.len() == tail_len {
+                    tail.pop_front();
+                }
+                tail.push_back(line);
             }
-            summary.push_str(&format!("... +{hidden} more lines\n"));
-            for line in &lines[lines.len() - tail_len..] {
-                summary.push_str(clean_line(line).as_ref());
-                summary.push('\n');
+
+            let hidden = total - head.len() - tail.len();
+            for line in head {
+                push_dump_line(&mut summary, line.as_ref());
+            }
+            if hidden > 0 {
+                summary.push_str(&format!("... +{hidden} more lines\n"));
+            }
+            for line in tail {
+                push_dump_line(&mut summary, line.as_ref());
             }
             return Some(summary);
         }
@@ -555,20 +589,72 @@ src/app.ts(1,7): error TS2322: Type 'string' is not assignable to type 'number'.
 
     #[test]
     fn test_tsc_stream_failed_unparsed_output_keeps_head_and_tail() {
-        let input: String = (1..=30).map(|i| format!("junk line {i}\n")).collect();
+        let padding = "x".repeat(30);
+        let input: String = (1..=30)
+            .map(|i| format!("junk line {i} {padding}\n"))
+            .collect();
         let mut f = BlockStreamFilter::new(TscHandler::new());
         let result = run_block_filter(&mut f, &input, 2);
         assert!(result.contains("compiler exited with code 2"), "got: {}", result);
         for i in 1..=5 {
-            assert!(result.contains(&format!("junk line {i}\n")), "got: {}", result);
+            assert!(result.contains(&format!("junk line {i} {padding}\n")), "got: {}", result);
         }
         assert!(result.contains("... +20 more lines"), "got: {}", result);
         for i in 26..=30 {
-            assert!(result.contains(&format!("junk line {i}\n")), "got: {}", result);
+            assert!(result.contains(&format!("junk line {i} {padding}\n")), "got: {}", result);
         }
-        assert!(!result.contains("junk line 6\n"), "got: {}", result);
-        assert!(!result.contains("junk line 25\n"), "got: {}", result);
+        assert!(!result.contains(&format!("junk line 6 {padding}\n")), "got: {}", result);
+        assert!(!result.contains(&format!("junk line 25 {padding}\n")), "got: {}", result);
         assert_eq!(result.lines().count(), 1 + 5 + 1 + 5);
+    }
+
+    #[test]
+    fn test_tsc_stream_failed_unparsed_output_ignores_escape_only_lines() {
+        let padding = "x".repeat(30);
+        let mut input = "\x1b[0m\n\x1b[32m\x1b[0m\n\x1b[0m\n".to_string();
+        for i in 1..=30 {
+            input.push_str(&format!("real line {i} {padding}\n"));
+        }
+
+        let mut f = BlockStreamFilter::new(TscHandler::new());
+        let result = run_block_filter(&mut f, &input, 2);
+        for i in 1..=5 {
+            assert!(result.contains(&format!("real line {i} {padding}\n")), "got: {}", result);
+        }
+        assert!(result.contains("... +20 more lines"), "got: {}", result);
+        for i in 26..=30 {
+            assert!(result.contains(&format!("real line {i} {padding}\n")), "got: {}", result);
+        }
+        assert!(!result.contains("\n\n"), "got: {}", result);
+        assert_eq!(result.lines().count(), 1 + 5 + 1 + 5);
+    }
+
+    #[test]
+    fn test_tsc_stream_failed_unparsed_output_under_tee_floor_is_complete() {
+        let input: String = (1..=15).map(|i| format!("short {i}\n")).collect();
+        let mut f = BlockStreamFilter::new(TscHandler::new());
+        let result = run_block_filter(&mut f, &input, 1);
+
+        for i in 1..=15 {
+            assert!(result.contains(&format!("short {i}\n")), "got: {}", result);
+        }
+        assert!(!result.contains("... +"), "got: {}", result);
+    }
+
+    #[test]
+    fn test_tsc_stream_failed_unparsed_output_caps_line_width() {
+        let long_line = "z".repeat(300);
+        let padding = "x".repeat(30);
+        let mut input = format!("{long_line}\n");
+        for i in 1..=10 {
+            input.push_str(&format!("padding line {i} {padding}\n"));
+        }
+
+        let mut f = BlockStreamFilter::new(TscHandler::new());
+        let result = run_block_filter(&mut f, &input, 1);
+        let emitted_long_line = result.lines().nth(1).expect("long line should be emitted");
+        assert!(emitted_long_line.chars().count() <= 120, "got: {}", result);
+        assert!(emitted_long_line.ends_with("..."), "got: {}", result);
     }
 
     #[test]
--- a/src/cmds/js/tsc_cmd.rs
+++ b/src/cmds/js/tsc_cmd.rs
@@ -10,17 +10,13 @@ use std::borrow::Cow;
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::LazyLock;
 
-/// Unparseable failure output is limited to this many non-empty lines, split
-/// between head and tail. The runner tees the full raw output and prints a
-/// recovery hint, so the hidden middle stays reachable, with two exceptions the
-/// cap accounts for: output under `MIN_TEE_SIZE` bytes gets no tee, so it is
-/// printed whole; and with tee disabled (`RTK_TEE=0` or
-/// `config.tee.enabled = false`) these lines are the only surviving copy, which
-/// is why both ends are kept.
+/// Cap on the non-empty lines kept when RTK cannot parse failure output. With
+/// tee disabled (`RTK_TEE=0`, `config.tee.enabled = false`) these lines are the
+/// only surviving copy of the failure.
 const MAX_UNPARSED_LINES: usize = CAP_WARNINGS;
-/// Deviation: tsc and npx print the cause first (`Unknown compiler option`,
-/// `This is not the tsc command`) and boilerplate after it, so a plain tail
-/// would drop the cause; split the cap between head and tail.
+/// tsc and npx print the cause first (`Unknown compiler option`, `This is not
+/// the tsc command`) and boilerplate after it, so spending the whole cap on a
+/// tail would drop the cause.
 const MAX_UNPARSED_HEAD_LINES: usize = reduced(MAX_UNPARSED_LINES, 5);
 
 static TSC_ERROR: LazyLock<Regex> = LazyLock::new(|| {
@@ -67,7 +63,6 @@ fn parse_diagnostic(line: &str) -> Option<Diagnostic<'_>> {
     })
 }
 
-/// One line of the raw failure dump: width-capped like every other emission.
 fn push_dump_line(summary: &mut String, line: &str) {
     summary.push_str(&truncate(line, 120));
     summary.push('\n');
@@ -128,8 +123,7 @@ impl TscHandler {
 
 impl BlockHandler for TscHandler {
     /// `--pretty` wraps every field in ANSI escapes; strip them once so
-    /// matching and the emitted block both see plain text. Plain lines are
-    /// passed through without allocating.
+    /// matching and the emitted block both see plain text.
     fn normalize_line<'a>(&self, line: &'a str) -> Cow<'a, str> {
         clean_line(line)
     }
@@ -165,8 +159,8 @@ impl BlockHandler for TscHandler {
             }
             // tsc failed without one diagnostic RTK understands (missing binary,
             // no project, unparseable output). "No errors found" would be a
-            // false green, so report the failure with the head and tail of the
-            // raw output; only the retained lines are ANSI-stripped.
+            // false green, so surface the failure with both ends of the raw
+            // output.
             let mut summary = format!(
                 "TypeScript: compiler exited with code {exit_code}, but RTK parsed no diagnostics\n"
             );
--- a/src/cmds/js/tsc_cmd.rs
+++ b/src/cmds/js/tsc_cmd.rs
@@ -2,17 +2,23 @@
 
 use crate::core::runner;
 use crate::core::stream::{BlockHandler, BlockStreamFilter};
-use crate::core::truncate::CAP_WARNINGS;
+use crate::core::truncate::{reduced, CAP_WARNINGS};
 use crate::core::utils::{resolved_command, strip_ansi, tool_exists, truncate};
 use anyhow::Result;
-use regex::{Captures, Regex};
+use regex::Regex;
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::sync::LazyLock;
 
-/// Unparseable failure output is tailed to this many non-empty lines; the
-/// runner tees the full raw output and prints the recovery hint.
-const MAX_UNPARSED_TAIL_LINES: usize = CAP_WARNINGS;
+/// Unparseable failure output is limited to this many non-empty lines. The
+/// runner tees the full raw output and prints a recovery hint, except when tee
+/// is disabled (`RTK_TEE=0` or `config.tee.enabled = false`): then these lines
+/// are the only surviving copy, which is why both ends are kept.
+const MAX_UNPARSED_LINES: usize = CAP_WARNINGS;
+/// Deviation: tsc and npx print the cause first (`Unknown compiler option`,
+/// `This is not the tsc command`) and boilerplate after it, so a plain tail
+/// would drop the cause; split the cap between head and tail.
+const MAX_UNPARSED_HEAD_LINES: usize = reduced(MAX_UNPARSED_LINES, 5);
 
 static TSC_ERROR: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^(.+?)\((\d+),(\d+)\):\s+(error|warning)\s+(TS\d+):\s+(.+)$").unwrap()
@@ -21,13 +27,49 @@ static TSC_ERROR: LazyLock<Regex> = LazyLock::new(|| {
 static TSC_PRETTY_ERROR: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^(.+?):(\d+):(\d+)\s+-\s+(error|warning)\s+(TS\d+):\s+(.+)$").unwrap()
 });
+static TSC_GLOBAL_ERROR: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^(error|warning)\s+(TS\d+):\s+(.+)$").unwrap());
 
-/// Match a diagnostic line in either the default or the `--pretty` layout.
-/// Callers pick the capture groups they need, so nothing is allocated here.
-fn tsc_error_captures(line: &str) -> Option<Captures<'_>> {
-    TSC_ERROR
-        .captures(line)
-        .or_else(|| TSC_PRETTY_ERROR.captures(line))
+struct Diagnostic<'a> {
+    file: Option<&'a str>,
+    line: Option<&'a str>,
+    code: &'a str,
+    message: &'a str,
+}
+
+/// Match default, `--pretty`, and file-less TypeScript diagnostics without allocating.
+fn parse_diagnostic(line: &str) -> Option<Diagnostic<'_>> {
+    if let Some(caps) = TSC_ERROR.captures(line) {
+        return Some(Diagnostic {
+            file: caps.get(1).map(|value| value.as_str()),
+            line: caps.get(2).map(|value| value.as_str()),
+            code: caps.get(5)?.as_str(),
+            message: caps.get(6)?.as_str(),
+        });
+    }
+    if let Some(caps) = TSC_PRETTY_ERROR.captures(line) {
+        return Some(Diagnostic {
+            file: caps.get(1).map(|value| value.as_str()),
+            line: caps.get(2).map(|value| value.as_str()),
+            code: caps.get(5)?.as_str(),
+            message: caps.get(6)?.as_str(),
+        });
+    }
+    let caps = TSC_GLOBAL_ERROR.captures(line)?;
+    Some(Diagnostic {
+        file: None,
+        line: None,
+        code: caps.get(2)?.as_str(),
+        message: caps.get(3)?.as_str(),
+    })
+}
+
+fn clean_line(line: &str) -> Cow<'_, str> {
+    if line.contains('\x1b') {
+        Cow::Owned(strip_ansi(line))
+    } else {
+        Cow::Borrowed(line)
+    }
 }
 
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
@@ -80,11 +122,7 @@ impl BlockHandler for TscHandler {
     /// matching and the emitted block both see plain text. Plain lines are
     /// passed through without allocating.
     fn normalize_line<'a>(&self, line: &'a str) -> Cow<'a, str> {
-        if line.contains('\x1b') {
-            Cow::Owned(strip_ansi(line))
-        } else {
-            Cow::Borrowed(line)
-        }
+        clean_line(line)
     }
 
     fn should_skip(&mut self, line: &str) -> bool {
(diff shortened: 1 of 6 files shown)


lbaauw:
Confirming this fix also covers a related trigger: missing binary entirely (not just --pretty ANSI codes)

Ran into the same underlying failure mode on rtk 0.42.4 (before this PR), but triggered differently than the ANSI-parsing case in #1772.

Setup: npx tsc --noEmit / npx next lint via rtk npx, run inside a git worktree with no node_modules (worktrees don't inherit node_modules from the main checkout).

Actual raw output (confirmed via rtk proxy npx tsc --noEmit, unfiltered):

- Use yarn to avoid accidentally running code from un-installed packages
Non-zero exit code, zero regex matches.

What the rtk npx wrapper reported instead:
TypeScript: No errors found
Next.js Build
Errors: 0 | Warnings: 0

Same root cause as #1772 — 0 parsed diagnostics → wrapper falls back to a fabricated success summary — just triggered by a missing binary rather than ANSI-wrapped --pretty output.

From the PR description, the general fallback ("avoid reporting 'No errors found' when the compiler exits non-zero and RTK parsed no diagnostics") looks like it fixes this case too, since it's exit-code-driven rather than specific to the ANSI-stripping fix. Just wanted to confirm that broader coverage and flag it as an additional real-world repro — this failure mode is fairly dangerous for pre-push typecheck workflows (agents/devs get a false "all green" and push broken code).

Repro environment: rtk 0.42.4, Ubuntu 24.04, Node v22, pnpm-based Next.js project.

KuSh (changes requested):
Thanks for this — I fetched the branch, built it, and ran it against real `tsc` (TypeScript 6.0.3) with the repro from #1772. The core fix works: `rtk tsc --noEmit --pretty` now reports `TypeScript: 1 errors in 1 files` and exits 2 instead of claiming `No errors found`. `cargo fmt --check`, `cargo clippy --all-targets` and the 12 `tsc_cmd` tests all pass.

Two things need to change before merge.

### 1. Unbounded raw dump makes rtk emit *more* than the wrapped command (`src/cmds/js/tsc_cmd.rs:115`)

The new "exited non-zero, parsed no diagnostics" branch appends the entire `raw` (stdout+stderr, capped only by `RAW_CAP` = 10 MiB) to the summary. Unlike `run_captured_filter`, the `RunMode::Streamed` path in `src/core/runner.rs:180` applies no `guard::never_worse`, so nothing clamps it.

Measured with a failing run producing ~3000 unparseable lines:

- raw: 322,989 bytes
- rtk output: **323,060 bytes** (~80k tokens into the context, negative savings recorded by `timer.track`)

That's the inverse of RTK's contract, and `tee_and_hint` already persists the raw output and prints a `[full output: ~/path]` pointer for exactly this case. Please cap it — `truncate` is already imported here, and `tee::force_tee_tail_hint` exists. #3295 solves this half by appending only the last 10 non-empty lines; that bounded approach is strictly better than the full dump and is worth porting into this PR.

### 2. ANSI is stripped for matching but not for output (`src/cmds/js/tsc_cmd.rs:89`)

`is_block_start` / `is_block_continuation` return a bool, and `BlockStreamFilter::feed_line` pushes the *original* `line` — so every emitted `--pretty` diagnostic keeps its escapes. Real output from the branch:

```
^[[96mrtk-test.ts^[[0m:^[[93m3^[[0m:^[[93m5^[[0m - ^[[91merror^[[0m^[[90m TS2322: ^[[0mType 'number'…
```

Eight escape sequences per error line, on the very path this PR fixes, and inconsistent with `filter_tsc_output`, which now strips them. `test_tsc_stream_failed_unparsed_output_not_success` asserts `!result.contains("\x1b[")`, but `test_tsc_stream_pretty_ansi_errors` doesn't — so the gap is untested. Emitting the stripped line (a `normalize_line` hook, or stripping in `feed_line`) plus an assertion in that second test would close it.

### Minor (non-blocking)

- `:154` — `filter_tsc_output` strips the whole output into `clean_output`, then `parse_tsc_error` strips each already-clean line again. In the streaming path it's worse: `should_skip` (`:86`) and `is_block_start` (`:90`) each call `strip_ansi`, so every line — including the thousands that match nothing — costs two `Regex::replace_all` passes and two allocations where the old code allocated none. Strip once per line and pass the clean `&str` to both.
- `:31` — `parse_tsc_error` allocates `line` and `message` for every diagnostic, but `is_block_start` uses only `file` and `code`.

### Mechanical

The PR is currently `CONFLICTING`: `develop` has since migrated this file from `lazy_static!` to `std::sync::LazyLock` and removed the `═══…` separator lines, while this branch adds `TSC_PRETTY_ERROR` inside the now-deleted `lazy_static!` block. Needs a rebase onto `develop` with `TSC_PRETTY_ERROR` converted to `static TSC_PRETTY_ERROR: LazyLock<Regex> = LazyLock::new(|| …)`.

### Related work

This PR is the oldest of the cluster and the only one fixing *both* halves (ANSI stripping *and* the exit-code guard), so it's the one to land:

- #3221 — duplicate of this PR (exit-code half only), newer.
- #3295 — duplicate of this PR (exit-code half only), newer; its bounded "last 10 non-empty lines" dump is the better implementation and should be adopted here (see point 1).
- #3294, #3222 — already closed, no action.
- #3220 — also fixed by this PR; adding `Fixes #3220` to the description.
- #3738 (ANSI escapes pass through, `strip_ansi` opt-in) — point 2 above is a concrete instance of it, but this PR only fixes it for the `tsc` path, so it's a partial fix, not a close.

jsiu93:
Rebased onto develop (TSC_PRETTY_ERROR is a LazyLock now) and addressed both blocking points plus the two minor ones. The non-zero-exit/no-diagnostics branch keeps only the last CAP_WARNINGS non-empty raw lines with a ... +N more lines marker, relying on the runner's tee hint for the rest. ANSI handling moved to a new BlockHandler::normalize_line hook (identity by default) so the tsc handler strips once per line and the emitted block is clean; plain lines are borrowed, not allocated. is_block_start and the batch filter now read capture groups directly instead of building a struct per line. Added assertions for the emitted --pretty block and a tail-cap test, plus a hook test in stream.rs. Verified against TypeScript 7.0.2 with the #1772 repro, a TS5058 config error, and an unknown flag.

src/cmds/js/tsc_cmd.rs
```diff
@@ -58,12 +76,23 @@ impl TscHandler {
...
+        }
+    }
+
     fn should_skip(&mut self, line: &str) -> bool {
         line.starts_with("Found ")
     }
 
     fn is_block_start(&mut self, line: &str) -> bool {
-        if let Some(caps) = TSC_ERROR.captures(line) {
+        if let Some(caps) = tsc_error_captures(line) {
```
KuSh: **Blocking — file-less diagnostics are still dropped.**

Both regexes are anchored on a location prefix (`^(.+?)\((\d+),(\d+)\):` and `^(.+?):(\d+):(\d+)`), so a diagnostic with no file — `error TS5023: Unknown compiler option 'foo'.`, `error TS5058: Cannot read file 'tsconfig.json'.` — matches neither and never becomes a block start. `error_count` stays untouched, so the new guard in `format_summary` (which only fires on `error_count == 0`) doesn't catch it either.

Net effect: **one** parseable file-scoped error is enough to make a config failure vanish completely. Input

```
error TS5023: Unknown compiler option 'foo'.
src/app.ts(1,7): error TS2322: Type 'string' is not assignable to type 'number'.
```

at exit 2 reports `TypeScript: 1 errors in 1 files` and the TS2322 block only — the TS5023 is gone.

A third pattern for the bare `error TS\d+: message` form (counted, but with no file to group under) would close this.

src/cmds/js/tsc_cmd.rs
```diff
@@ -2,15 +2,33 @@
...
 use anyhow::Result;
-use regex::Regex;
+use regex::{Captures, Regex};
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::sync::LazyLock;
 
+/// Unparseable failure output is tailed to this many non-empty lines; the
+/// runner tees the full raw output and prints the recovery hint.
+const MAX_UNPARSED_TAIL_LINES: usize = CAP_WARNINGS;
```
KuSh: **Non-blocking, but this is the tee doc I'd tighten.**

"the runner tees the full raw output" is conditional — tee is skipped when `RTK_TEE=0` or `config.tee.enabled = false`. That matters because the tail below is the *only* surviving copy in that configuration, which is what makes the head/tail choice consequential rather than cosmetic.

src/cmds/js/tsc_cmd.rs
```diff
@@ -77,9 +106,28 @@ impl BlockHandler for TscHandler {
...
+    fn format_summary(&self, exit_code: i32, raw: &str) -> Option<String> {
         if self.error_count == 0 {
-            return Some("TypeScript: No errors found\n".to_string());
+            if exit_code == 0 {
+                return Some("TypeScript: No errors found\n".to_string());
+            }
+            // tsc failed without one parseable diagnostic (missing binary, bad
+            // flags, TS5058 config errors). "No errors found" would be a false
+            // green, so report the failure with the tail of the raw output.
+            let clean = strip_ansi(raw);
```
KuSh: **Non-blocking (perf).** `strip_ansi(raw)` scans and copies the entire raw capture — up to `RAW_CAP` (10 MiB) — and the next line collects every non-empty line, all to print 10 of them.

Tailing first and stripping only the retained lines makes this constant cost regardless of how much the compiler printed.

src/cmds/js/tsc_cmd.rs
```diff
@@ -77,9 +106,28 @@ impl BlockHandler for TscHandler {
...
+            let lines: Vec<&str> = clean.lines().filter(|l| !l.trim().is_empty()).collect();
+            let hidden = lines.len().saturating_sub(MAX_UNPARSED_TAIL_LINES);
+            let mut summary = format!(
+                "TypeScript: compiler exited with code {exit_code}, but RTK parsed no diagnostics\n"
+            );
+            if hidden > 0 {
+                summary.push_str(&format!("... +{hidden} more lines\n"));
+            }
+            for line in &lines[hidden..] {
```
KuSh: **Blocking-ish — this keeps the wrong end of the output.**

`hidden = len - 10` and `&lines[hidden..]` retain the **last** `CAP_WARNINGS` (= 10) non-empty lines. But `tsc` and `npx` print the diagnostic **first** and filler afterwards.

Concretely, bare `tsc` with no tsconfig and no file args prints `Version 5.x` / `tsc: The TypeScript Compiler`, then the COMMON COMMANDS block, then exits non-zero. The tail keeps the usage boilerplate and drops exactly the two lines that explain the failure.

A head-biased slice, or a head+tail split (first 5 + last 5 with the marker between), preserves the cause. This is sharper than it looks given the tee caveat above: with tee disabled, the tail is all the user gets.

src/cmds/js/tsc_cmd.rs
```diff
@@ -113,12 +161,13 @@ pub(crate) fn filter_tsc_output(output: &str) -> String {
     }
 
     let mut errors: Vec<TsError> = Vec::new();
-    let lines: Vec<&str> = output.lines().collect();
+    let clean_output = strip_ansi(output);
```
KuSh: **Non-blocking (perf).** This copies the whole buffer unconditionally, even in the common non-`--pretty` case where nothing changes.

`normalize_line` above (line 83) already establishes the right guard — `if line.contains('\x1b')`, otherwise `Cow::Borrowed`. Mirroring that here with a `Cow` would avoid a second full copy of up to `RAW_CAP` on the `rtk pipe tsc` path.

src/cmds/js/tsc_cmd.rs
```diff
@@ -149,7 +198,7 @@ pub(crate) fn filter_tsc_output(output: &str) -> String {
     }
 
     if errors.is_empty() {
-        if output.contains("Found 0 errors") {
+        if clean_output.contains("Found 0 errors") {
```
KuSh, line 289: **Blocking — this path still false-greens, and it's a live #3220 repro.**

The streaming handler was hardened against "exited non-zero, parsed nothing", but `filter_tsc_output` — the `rtk pipe tsc` path — was not. With only file-less diagnostics in the buffer, `errors.is_empty()` is true, `Found 0 errors` is absent, and line 204 returns `TypeScript compilation completed`.

```rust
filter_tsc_output("error TS5058: Cannot read file 'tsconfig.json'.")
// => "TypeScript compilation completed"
```

That is exactly the failure mode #3220 describes, and the PR body now claims `Fixes #3220`. Either this path needs the same treatment, or it needs the file-less pattern from my first comment (which would fix both at once).

src/cmds/js/tsc_cmd.rs
```diff
@@ -313,6 +383,63 @@ Found 3 errors in 2 files.
         assert!(!result.contains("Found 3"), "got: {}", result);
     }
 
+    #[test]
+    fn test_tsc_stream_pretty_ansi_errors() {
+        let input = "\
+\x1b[96msrc/app.ts\x1b[0m:\x1b[93m1\x1b[0m:\x1b[93m7\x1b[0m - \x1b[91merror\x1b[0m\x1b[90m TS2322: \x1b[0mType 'string' is not assignable to type 'number'.
+  const x: number = \"string\";
+\x1b[31mFound 1 error in src/app.ts:1\x1b[0m
+";
```
KuSh: **Non-blocking, but the assertions overclaim.**

This fixture (and its sibling at line 288-291) puts the code frame on a line indented by two spaces. Real `tsc --pretty` emits a **blank** line after the header, then the frame behind a line-number gutter:

```
src/index.ts:2:7 - error TS2322: Type 'string' is not assignable to type 'number'.

2 const x: number = "hello";
      ~
```

`is_block_continuation` requires `"  "` or `\t`, so against real output the blank line closes the block immediately and the entire code frame is discarded — the stream emits only the header plus the summary.

The header capture is the main win here and it genuinely works, and dropping the snippet is arguably correct for token savings. The problem is that these tests assert context is preserved when against real `tsc` it isn't, so they'll mislead the next person to touch this. Worth swapping in a captured fixture per the repo's "use real command output, not synthetic data" rule — and if the frame *should* be kept, `is_block_continuation` needs to accept the gutter form.

(Message elaborations, which really are two-space indented, do survive — it's the source snippet and squiggle that are lost.)

KuSh (changes requested):
Thanks for the quick turnaround — I re-reviewed at `f571b75`. **All five points from my last pass are genuinely resolved:**

- Raw dump is bounded to `CAP_WARNINGS` lines with a `... +N more` marker.
- `BlockHandler::normalize_line` is a clean opt-in hook, and the emitted block is now stripped — the lifetime split (`&self` vs `&'a str -> Cow<'a, str>`) keeps the later `&mut self.handler` calls legal, and the identity default leaves every existing handler unchanged.
- ANSI is stripped once per line, guarded by `contains('\x1b')`, borrowing plain lines.
- `tsc_error_captures` drops the per-line struct allocation; the two regexes have aligned capture groups, so indexing is safe for both layouts.
- Rebased onto `develop`, `LazyLock`, no conflicts.

`cargo fmt --check`, `cargo clippy --all-targets` and `cargo test --all` all pass here.

The deeper pass turned up two things that I think still block, both about diagnostics that carry **no file prefix** (`error TS5023:`, `error TS5058:`) — they match neither regex, so they're invisible to `error_count`, and the new guard only fires when `error_count == 0`. One of them is a live #3220 repro on the `rtk pipe tsc` path, which matters now that the description says `Fixes #3220`. Details inline, plus three non-blocking notes on the tail direction, two `strip_ansi` full-buffer copies, and the synthetic `--pretty` fixtures.

One suggestion I want to explicitly argue **against**, in case it comes up: guarding the failure branch with `!clean.contains("Found 0 errors")` to avoid a false red on a Ctrl-C'd watch-mode run (exit 130). It fixes a rare case at the cost of reopening the false green whenever a compiler prints `Found 0 errors` and then fails for an unrelated reason — the exact bug this PR exists to kill. The current exit-code-driven logic is the right call.

src/cmds/js/tsc_cmd.rs
KuSh: Resolved by 6232c37. `TSC_GLOBAL_ERROR` + `parse_diagnostic` returning `Diagnostic { file: Option<&str>, .. }` now counts file-less diagnostics, and `files` only gets an entry when there is one — so the summary degrades to `TypeScript: N errors` instead of miscounting files. `test_tsc_stream_mixed_global_and_file_errors` covers my exact TS5023 + TS2322 repro and asserts `2 errors in 1 files`.

src/cmds/js/tsc_cmd.rs
KuSh: Resolved by 6232c37 — the doc now spells out the `RTK_TEE=0` / `config.tee.enabled = false` case and ties it to why both ends are kept.

src/cmds/js/tsc_cmd.rs
KuSh: Resolved by 6232c37 — lines are filtered first and `clean_line` is applied only to the retained ones, so the cost no longer scales with `RAW_CAP`.

src/cmds/js/tsc_cmd.rs
KuSh: Resolved by 6232c37 — `MAX_UNPARSED_HEAD_LINES = reduced(MAX_UNPARSED_LINES, 5)` gives a head+tail split with the marker between. `test_tsc_stream_failed_without_project_keeps_cause` uses the real `tsc_no_project_raw.txt` capture and asserts `Version 6.0.3` and `tsc: The TypeScript Compiler` survive — the two lines the old tail dropped.

src/cmds/js/tsc_cmd.rs
KuSh: Resolved by 6232c37 — `clean_line(output)` returns a `Cow`, so the non-`--pretty` path no longer copies the buffer.

src/cmds/js/tsc_cmd.rs
KuSh: Resolved by 6232c37 — `parse_diagnostic` now matches global diagnostics on this path too, so file-less-only output renders a `global (N errors)` section. `test_filter_tsc_output_global_error` asserts `!result.contains("completed")`, which is the #3220 repro.

src/cmds/js/tsc_cmd.rs
KuSh: Resolved by 6232c37 — replaced with the real `tests/fixtures/tsc_pretty_raw.txt` capture (blank lines + `1 const x…` gutter), and the assertions now state the truth: `!result.contains("const x: number")` with a comment explaining the compact filter drops the frames. Thanks for capturing real output rather than adjusting the fixture to fit.

src/cmds/js/tsc_cmd.rs
```diff
@@ -77,16 +149,52 @@ impl BlockHandler for TscHandler {
...
         if self.error_count == 0 {
-            return Some("TypeScript: No errors found\n".to_string());
+            if exit_code == 0 {
+                return Some("TypeScript: No errors found\n".to_string());
+            }
+            // tsc failed without one diagnostic RTK understands (missing binary,
+            // no project, unparseable output). "No errors found" would be a
+            // false green, so report the failure with the head and tail of the
+            // raw output; only the retained lines are ANSI-stripped.
+            let lines: Vec<&str> = raw.lines().filter(|line| !line.trim().is_empty()).collect();
```
KuSh: **Needs a change — and it's a side effect of the per-line stripping I asked for, so apologies for sending you into it.**

The blank-line filter runs on **un-stripped** `raw`. ESC (U+001B) is not Unicode `White_Space`, so `"\x1b[0m".trim()` is *not* empty: a line whose entire content is escape sequences survives this filter, consumes one of the 10 head/tail slots, and then prints as a blank line once `clean_line` runs at 168/178/183.

Reproduced with 3 ANSI-only lines followed by 12 real ones — the head comes out as three blank lines plus `real line 1` and `real line 2`, hiding `real line 3..7`. Anything that emits a coloured progress spinner before failing (npx, pnpm) can push ten of these through, at which point the dump is nothing but blank lines and `... +N more lines`, and the user learns nothing about why tsc failed — the exact false-green this PR exists to kill.

The previous revision did `strip_ansi(raw)` before splitting, which was immune to this. Reordering is enough:

```rust
let clean = strip_ansi(raw);
let lines: Vec<&str> = clean.lines().filter(|line| !line.trim().is_empty()).collect();
```

That does reinstate the whole-buffer copy I flagged last round. If you want to keep both properties, filter on the normalized form instead — `raw.lines().map(clean_line).filter(|l| !l.trim().is_empty())` — and keep the `Cow`s.

src/cmds/js/tsc_cmd.rs
```diff
@@ -2,15 +2,75 @@
...
 use anyhow::Result;
 use regex::Regex;
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::sync::LazyLock;
 
+/// Unparseable failure output is limited to this many non-empty lines. The
+/// runner tees the full raw output and prints a recovery hint, except when tee
+/// is disabled (`RTK_TEE=0` or `config.tee.enabled = false`): then these lines
+/// are the only surviving copy, which is why both ends are kept.
```
KuSh: **Low / bonus.** The tee caveat is now accurate for the two cases it names, but there's a third: `should_tee` also bails when `raw_len < MIN_TEE_SIZE` (500 bytes, `src/core/tee.rs:111`).

So a failure emitting ~15 short lines totalling under 500 bytes prints 10 of them, says `... +5 more lines`, writes no tee file, and prints no recovery hint — those 5 lines are gone with nothing pointing at them. Either mention the size floor here, or (nicer) skip the cap entirely when the raw output is small enough to print whole, which makes the comment true by construction.

src/cmds/js/tsc_cmd.rs
```diff
@@ -77,16 +149,52 @@ impl BlockHandler for TscHandler {
...
+            // false green, so report the failure with the head and tail of the
+            // raw output; only the retained lines are ANSI-stripped.
+            let lines: Vec<&str> = raw.lines().filter(|line| !line.trim().is_empty()).collect();
+            let mut summary = format!(
+                "TypeScript: compiler exited with code {exit_code}, but RTK parsed no diagnostics\n"
+            );
+
+            if lines.len() <= MAX_UNPARSED_LINES {
+                for line in lines {
+                    summary.push_str(clean_line(line).as_ref());
```
KuSh: **Low / bonus.** These lines go out verbatim, while every other emission in this file passes through `truncate(..., 120)` (lines 310, 314, 333, 337).

Capping the dump at 10 *lines* without capping width doesn't actually bound its cost — one minified stack frame, a long `one of:` enumeration, or a single-line JSON blob blows straight past it. Your own `tests/fixtures/tsc_no_project_raw.txt` carries a ~1900-character `--lib` line; it happens to fall in the hidden middle for that input, but it wouldn't if the failure output were shorter.

src/cmds/js/tsc_cmd.rs
```diff
@@ -77,16 +149,52 @@ impl BlockHandler for TscHandler {
...
+
+            let head_len = MAX_UNPARSED_HEAD_LINES.min(MAX_UNPARSED_LINES);
+            let tail_len = MAX_UNPARSED_LINES.saturating_sub(head_len);
+            let hidden = lines.len() - head_len - tail_len;
+            for line in &lines[..head_len] {
+                summary.push_str(clean_line(line).as_ref());
+                summary.push('\n');
+            }
+            summary.push_str(&format!("... +{hidden} more lines\n"));
+            for line in &lines[lines.len() - tail_len..] {
```
KuSh: **Low / bonus.** Materializing every non-empty line of `raw` (capped at `RAW_CAP` = 10 MiB) just to index the first 5 and last 5 is unbounded: a large unparseable output produces a `Vec<&str>` with hundreds of thousands of entries, ~4 MB of slice headers alone, against the project's <5 MB target.

`take(head_len)` plus a bounded `VecDeque` for the tail (or a two-pass count) keeps it constant. Failure-path only — but that's precisely where output tends to be big.

KuSh (changes requested):
LGTM overall — this is in good shape, and thank you for the thorough turnaround. **All seven threads from my last review are genuinely fixed, and I've resolved them individually** with a note on each:

- `TSC_GLOBAL_ERROR` + `parse_diagnostic` returning `Diagnostic { file: Option<&str>, .. }` is a cleaner answer to the file-less-diagnostic problem than what I suggested, and degrading the summary to `TypeScript: N errors` when there's no file to group under is the right call.
- The head+tail split via `reduced(MAX_UNPARSED_LINES, 5)` keeps the cause, and `test_tsc_stream_failed_without_project_keeps_cause` proves it against a real capture rather than an assumption.
- Swapping the synthetic `--pretty` strings for `tests/fixtures/tsc_pretty_raw.txt` — real gutter, real blank separators — and then *changing the assertions to match reality* (`!result.contains("const x: number")`, with a comment explaining why) is exactly right. Capturing real output instead of bending the fixture to fit the code is the harder and better choice.

`cargo fmt --check`, `cargo clippy --all-targets` and `cargo test --all` (2662+) all pass here. I also checked the things most likely to bite and found them correct: the head/tail arithmetic can't underflow (`head_len + tail_len == MAX_UNPARSED_LINES` by construction, and the branch is only reached when `lines.len()` exceeds it), regex precedence holds on Windows drive-letter and colon-containing paths, the `Cow` lifetimes keep the later `&mut self.handler` borrow legal, and `normalize_line`'s identity default leaves every other `BlockHandler` untouched.

**One change requested**, inline at line 161: the blank-line filter now runs before ANSI stripping, so escape-only lines survive it and render as blanks — which can empty out the failure dump. That one is a direct consequence of the per-line-stripping optimisation I asked for last round, so it's on me as much as anyone.

Three **low-priority bonuses** also inline (tee size floor in the doc, no width truncation on dump lines, unbounded `Vec` for a 10-line window) — all optional, fold them in if convenient, none of them should hold the PR.

Finally: the exit-**0** counterpart — `rtk tsc --showConfig` / `--listFiles` / `--init` / `--help` returning `TypeScript: No errors found` and discarding the real output, unrecoverably, since `TeeMode::Failures` writes nothing on a zero exit — is real but **pre-existing** (`develop` returned that unconditionally on `error_count == 0`). I've filed it separately as #3750 rather than letting it grow this PR. No action needed here.

src/cmds/js/tsc_cmd.rs
KuSh: Fixed, and verified beyond the added test. `clean_line` now runs before the blank check in **both** dump branches (`:176-178` and `:191-193`), so the shadowed binding is what `trim()` sees.

I probed the case your test doesn't cover — escape-only lines interleaved *throughout* the stream rather than bunched at the head, which is what a real `--pretty` capture looks like:

- no blank lines leak (`!contains("\n\n")`)
- no ESC leaks into the emitted dump
- head is still the first five *real* lines and tail the last five, so the noise no longer displaces content

Also confirmed the all-blank degenerate case (500 bytes of pure `\x1b[0m`) yields the header alone with no `... +0 more lines` marker. Thanks for the clean fix.

src/cmds/js/tsc_cmd.rs
KuSh: Fixed — and you went further than I asked, which turns out to be the right call: rather than only documenting the `MIN_TEE_SIZE` case you made the code honour it, printing the output whole when no tee will exist.

I verified the branch matches `should_tee` *exactly* rather than approximately, since a mismatch here silently drops the only surviving copy:

- `tee.rs:111` bails on `raw_len < MIN_TEE_SIZE`; the new branch fires on `raw.len() < MIN_TEE_SIZE` — same operator, same constant
- `stream.rs:530` builds `raw` once and hands the *same* value to `on_exit` and out via `StreamResult.raw`, which `runner.rs:193` passes to `tee_and_hint` — so the two lengths cannot diverge
- boundary probe: 499 bytes dumps whole, 500 bytes takes head/tail

The `pub(crate)` on `MIN_TEE_SIZE` is the right coupling — the two genuinely have to move together.

src/cmds/js/tsc_cmd.rs
KuSh: Fixed — `push_dump_line` applies `truncate(line, 120)` on every dump line, so this path now matches the rest of the file.

Verified `truncate` is char-based (`utils.rs:29`), so it is multibyte-safe on arbitrary compiler output; I probed 600 CJK chars plus emoji and nothing panicked or exceeded 120 chars.

Real-world effect on the fixture that motivated this — `tsc_no_project_raw.txt`, whose ~1900-char `--lib` line was the whole problem: **5240 → 481 bytes**, 12 lines, every line ≤120 chars.

src/cmds/js/tsc_cmd.rs
KuSh: Fixed — bounded `head` `Vec` plus a `tail` `VecDeque` with a `total` counter, single pass, no full materialization.

I checked the arithmetic for underflow since `hidden` is `usize`: every counted line lands in `head`, lands in `tail`, or is dropped by the `tail_len == 0` guard, so `head.len() + tail.len() <= total` always holds. The new `if hidden > 0` guard is a necessary consequence of dropping the old `lines.len() <= MAX_UNPARSED_LINES` early return — without it, a raw over the tee floor with 8 real lines would print a bogus `... +0 more lines`. Verified that case emits all 8 lines contiguously, in order, with no marker.

Worst case measured at `RAW_CAP`: **10 MiB → 1203 bytes, 12 lines, 56 ms**, constant memory.

KuSh (approved):
LGTM — approving. Thanks for the persistence across four rounds; this landed in good shape.

**On the last commit:** I verified the fix rather than taking the tests at face value, because the new `MIN_TEE_SIZE` branch is a behaviour change and a mismatch there would silently discard the only surviving copy of a failure.

- **Escape-only lines** — `clean_line` now runs before the blank check in both dump branches, so the shadowed binding is what `trim()` sees. I probed escapes interleaved *throughout* the stream rather than bunched at the head, which is what a real `--pretty` capture looks like: no blanks leak, no ESC leaks, and head/tail still hold real content.
- **The tee floor matches exactly, not approximately.** `tee.rs:111` bails on `raw_len < MIN_TEE_SIZE`; the new branch fires on the same operator and constant. `stream.rs:530` builds `raw` once and hands the *same* value to `on_exit` and out via `StreamResult.raw` → `runner.rs:193` → `tee_and_hint`, so the two lengths cannot diverge. Probed 499 bytes → dumps whole, 500 → head/tail.
- **`hidden` cannot underflow**: every counted line lands in `head`, in `tail`, or is dropped by the `tail_len == 0` guard, so `head.len() + tail.len() <= total` always holds. The new `if hidden > 0` guard is load-bearing — without it, a raw over the floor with 8 real lines would print `... +0 more lines`. That case emits all 8, contiguous and in order.
- **Width cap**: `truncate` is char-based, so 600 CJK chars plus emoji neither panic nor exceed 120.
- **Worst case at `RAW_CAP`**: 10 MiB → 1203 bytes, 12 lines, 56 ms, constant memory.

Real effect on the fixture that started this: `tsc_no_project_raw.txt` goes **5240 → 481 bytes**, every line ≤120 chars.

`cargo fmt --check`, `clippy --all-targets` (zero warnings) and `test --all` (2665) all pass locally.

**I pushed one commit on top of yours** (`docs(tsc): correct the ANSI note and drop restated comments`) rather than send you round five for comments alone. It is comments only — no behaviour change, and the suite is unchanged at 2665. One of them had gone stale under your own fix: the failure-dump comment still said *"only the retained lines are ANSI-stripped"*, which stopped being true the moment `clean_line` moved ahead of the blank filter. The others were restating the code they sat on, per [CONTRIBUTING.md](../blob/develop/CONTRIBUTING.md)'s *"no obvious comments"* — `MAX_UNPARSED_LINES` had grown into prose for the `raw.len() < MIN_TEE_SIZE` branch and a narration of `tee.rs`, so it now keeps only the reason both ends are worth having. Revert or reword any of it if you disagree; none of it is load-bearing.

Note the red **benchmark** check is not yours — it fails identically on `develop` (d9a5893, 322db4d).

Nice work, in particular capturing real fixtures and adjusting the assertions to match reality rather than bending the fixture to fit the code.

PR #2239, opened 2026-06-03
fix(grep): command token duplication in output withsame value filename or positional arg

KuSh:
## Summary
Fixes regression introduced by #1823 and detailed [here](https://github.com/rtk-ai/rtk/pull/1823#pullrequestreview-4400619909)

## Test plan
- [x] Manual testing:
  - `git diff -- diff` ✅️
  - `rtk git diff -- src/core/args_utils.rs` ✅️
  - `rtk git diff -- "src/core/args_utils.rs"` ✅️
  - `rtk git diff upstream/bug/critical-fixes` ✅️

### DIFF
--- a/src/core/args_utils.rs
+++ b/src/core/args_utils.rs
@@ -1,23 +1,27 @@
 //! Utility functions for argument handling, particularly for restoring "--" escape
 //! arguments that clap consumes during parsing.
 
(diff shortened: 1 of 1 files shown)


aeppling (approved):

aeppling:
LGTM ! thanks for the fix :)

PR #99, opened 2026-02-13
feat(hook): handle global options before subcommands

mhcoen:
## Summary
- Normalize git, cargo, docker, and kubectl commands by stripping global options to find the subcommand before matching
- `git -C /path status`, `cargo +nightly test`, `docker -H host ps`, `kubectl --context prod get pods` now correctly rewrite
- Preserves all original arguments (including global options) in the rewritten `rtk` command

### Tools affected
| Tool | Global options now handled |
|------|--------------------------|
| git | `-C <path>`, `-c k=v`, `--no-pager`, `--no-optional-locks`, `--bare`, `--literal-pathspecs`, `--key=val` |
| cargo | `+nightly`, `+stable`, `+1.75.0` |
| docker | `-H <host>`, `--context <ctx>`, `--config <path>`, `--key=val` |
| kubectl | `--context <ctx>`, `-n <ns>`, `--kubeconfig <path>`, `--namespace <ns>`, `--key=val` |

## Test plan
- [x] 54 test cases covering plain commands (regression), global options (new), non-matching subcommands, and all other tools
- [x] Depends on #98 (POSIX character classes) — merge that first

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2250, opened 2026-06-04
feat(ctest): add compact output filter

### LINKED ISSUE #2237: Add rtk ctest filter (summarize on pass, keep failing-test output)
## Summary

Add an `rtk ctest` filter. CTest output lists every test with pass/timing lines; on a green run the per-test detail is low-value, while on failure the failing test's captured output is what matters.

## Why

`rtk discover` flagged `ctest` as an unhandled command: **41 invocations** over 30 days (e.g. `ctest -R "test_..." --output-on-failure`), with no RTK equivalent.

## Suggested behavior

- On all-pass: collapse to a summary (`N/N passed`, total time, slowest few), drop per-test `Test #k: ... Passed` lines.
- On failure: keep the failing tests' names and their captured stdout/stderr verbatim (respect `--output-on-failure`), drop passing-test noise.

jsiu93:
## Summary
- Add `rtk ctest` with compact CTest output for green runs and preserved failing-test details.
- Register `ctest` in hook rewrite/discover, pipe filtering, tracking, and TOML shadow warnings.
- Preserve explicit detail modes by passing through verbose/show-only/help/version invocations.

Closes #2237

## Test plan
- [x] `cargo fmt -- --check`
- [x] `cargo test ctest`
- [x] `cargo test`
- [x] `cargo clippy -- -D warnings`
- [x] Manual: `cargo run --quiet -- rewrite "ctest -R smoke --output-on-failure"`
- [x] Manual: `cargo run --quiet -- pipe --filter ctest` with sample CTest output

> All PRs must target the `develop` branch.


jsiu93:
@KuSh @aeppling
Looks like the benchmark failure comes from input variance in the `curl json` case.

  That benchmark calls `https://mockhttp.org/json` separately for raw `curl` and `rtk curl`. The endpoint can return different JSON payloads between requests, so the token counts may be measured against different inputs.

  Here is a local example showing two different response shapes:

```bash
  ❯ rtk curl -s https://mockhttp.org/json
  {"message":"Hello from MockHTTP","timestamp":"2026-07-03T02:05:44.002Z","status":"success","code":200}

  ❯ curl -s https://mockhttp.org/json
  {"metrics":{"cpu":45.2,"memory":78.5,"disk":62.3,"network":{"inbound":"1.2MB/s","outbound":"0.8MB/s"},"uptime":"15 days","services":[{"name":"api","status":"healthy"},{"name":"database","status":"healthy"},{"name":"cache","status":"warning"}]}}

```

  A deterministic local HTTP fixture, or capturing one response and feeding the same body to both commands, should make this benchmark stable.

.claude/hooks/rtk-suggest.sh
KuSh, line 1: Please discard the changes from `.claude`; most of those files are deprecated.

src/cmds/system/pipe_cmd.rs
```diff
@@ -179,6 +180,13 @@ pub fn auto_detect_filter(input: &str) -> fn(&str) -> String {
         return crate::cmds::python::pytest_cmd::filter_pytest_output;
     }
 
+    if first_1k.contains("Test project")
+        && first_1k.contains("Test #")
+        && first_1k.contains("tests passed")
```
KuSh: That seems quite fragile and could match the output of many testing frameworks and tools

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+
+fn parse_test_line(line: &str, line_index: usize) -> Option<TestCase> {
+    static TEST_RE: OnceLock<Regex> = OnceLock::new();
+    let re = TEST_RE.get_or_init(|| {
+        Regex::new(
+            r"^\s*\d+/\d+\s+Test\s+#\d+:\s+(.+?)\s+\.{2,}\s*(?:\*{3})?\s*([A-Za-z][A-Za-z ]*?)\s+([\d.]+)\s+sec\s*$",
+        )
+        .expect("invalid ctest result regex")
+    });
```
KuSh: Please use `lazy_static!` for regex compilation

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,358 @@
...
+        .collect();
+
+    if let Some(section) = collect_failed_section(lines) {
+        blocks.push(section);
+    }
+
+    blocks
+}
+
+fn collect_failure_block(lines: &[&str], start: usize) -> Option<String> {
```
KuSh: This function starts capturing at the failed test's own summary line, so the raw dotted-progress line is included in addition to the compacted one. A test asserting full output would have surfaced that problem

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,358 @@
...
+    lines
+        .iter()
+        .enumerate()
+        .filter_map(|(line_index, line)| parse_test_line(line, line_index))
+        .collect()
+}
+
+fn parse_test_line(line: &str, line_index: usize) -> Option<TestCase> {
+    static TEST_RE: OnceLock<Regex> = OnceLock::new();
+    let re = TEST_RE.get_or_init(|| {
```
KuSh: No fixture/test covers non-Passed/Failed statuses (Timeout, Not Run, Exception), worth adding at least one regression fixture given CTest emits these in practice

KuSh (changes requested):

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,532 @@
+//! Compact CTest output while preserving failing-test details.
+
+use anyhow::Result;
+use lazy_static::lazy_static;
```
KuSh: **Blocks the build after rebase.** `develop` removed the `lazy_static` dependency in `5269df7` (`refactor: replace lazy_static with LazyLock`, 2026-07-28) — a week after your last push, so it isn't visible in this PR's diff and CI never exercised it.

Merging current `develop` into this branch gives:

```
error[E0432]: unresolved import `lazy_static`
error[E0425]: cannot find value `TEST_RE` in this scope
error[E0425]: cannot find value `START_RE` in this scope
error[E0425]: cannot find value `SUMMARY_RE` in this scope
error[E0425]: cannot find value `TIME_RE` in this scope
error: unused import: `regex::Regex`
```

Please convert to `std::sync::LazyLock`, which is what `.claude/rules/rust-patterns.md` mandates:

```rust
use regex::Regex;
use std::sync::LazyLock;

static TEST_RE: LazyLock<Regex> = LazyLock::new(|| {
    Regex::new(r"...").expect("invalid ctest result regex")
});
```

Apologies again — my original "please use `lazy_static!`" was accurate at the time.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,532 @@
...
+use crate::core::runner::{self, RunOptions};
+use crate::core::utils::{resolved_command, strip_ansi};
+
+const MAX_SLOWEST: usize = 3;
+
+lazy_static! {
+    static ref TEST_RE: Regex = Regex::new(
+        r"^\s*\d+/\d+\s+Test\s+#\d+:\s+(.+?)\s+\.{2,}\s*(?:\*{3})?\s*(.+?)\s+([\d.]+)\s+sec\s*$",
+    )
+    .expect("invalid ctest result regex");
```
KuSh: **`TEST_RE` misses CTest's most common failure line, silently dropping the failure.**

The pattern anchors on `…\s+([\d.]+)\s+sec\s*$`, but when a test fails a `PASS_REGULAR_EXPRESSION` / `FAIL_REGULAR_EXPRESSION` check, CTest inlines the reason into the result line **with an embedded newline**. Real `ctest 4.3.4` output:

```
3/7 Test #3: regex_fail .......................***Failed  Required regular expression not found. Regex=[expected-token
]  0.01 sec
nope
```

Neither physical line matches — the first has no trailing `N sec`, the second has no `N/M Test #K:` prefix. So `regex_fail` never becomes a `TestCase`: it's **absent from the `failed:` list**, and both the reason string and the test's captured stdout (`nope`) are dropped. That's precisely the detail this filter exists to preserve, and regex assertions are one of CTest's most-used mechanisms, so it isn't an edge case.

The regex needs to tolerate a reason suffix and a result line wrapped across physical lines — joining continuation lines before matching is probably the simplest fix. Worth a fixture for it.

KuSh (commented):
Thanks for the update — I re-checked all five threads from my last review and they're genuinely addressed:

- the `.claude/` changes are gone
- `pipe_cmd` detection is now gated on `looks_like_ctest_output()` (first non-empty line `Test project ` **and** a strict result-line match), with a negative test for keyword-only output
- regexes are compiled lazily
- `collect_failure_block` no longer re-emits the raw dotted-progress line, and there's a full `assert_eq!` covering the whole output
- Timeout / Not Run / Exception / Not Run (Disabled) all have coverage now

**Please rebase onto current `develop` before anything else.** The branch's merge base is `66e09cb`; `develop` has moved on since your 2026-07-21 push. GitHub still reports this mergeable with all 11 checks green, but both are measuring the July snapshot — I merged current `develop` into the head locally and `cargo check --all-targets` fails with 8 errors. Anchored inline below.

Sorry about the `lazy_static!` one in particular: my earlier request was correct when I made it, but `develop` migrated to `LazyLock` in `5269df7` a week after you pushed. That one's on the repo, not on you.

After the rebase there are two behavioural bugs worth fixing. I ran the filter against real `ctest 4.3.4` output from a scratch CMake project rather than against the fixtures, and both reproduce — also inline.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,532 @@
...
+
+pub(crate) fn filter_ctest_output(output: &str) -> String {
+    let clean = strip_ansi(output);
+    let trimmed = clean.trim();
+    if trimmed.is_empty() {
+        return String::new();
+    }
+    if trimmed.contains("No tests were found") {
+        return "ctest: no tests found".to_string();
+    }
```
KuSh: **Unanchored substring check can erase a genuine failure.**

This scans the *entire* output before any parsing happens. `--output-on-failure` is deliberately kept out of `should_passthrough`, so a failing test's own stdout is in scope here. A run with one failing test whose output contains `ERROR: No tests were found in the discovery phase` collapses to exactly:

```
ctest: no tests found
```

losing `assertion failed at foo.cpp:12` and the `0% tests passed, 1 tests failed out of 1` summary — while the process still exits non-zero, so the user sees a passing-looking message on a failed run.

Suggest matching CTest's actual line (`No tests were found!!!`), and only when no tests parsed **and** there's no failure summary.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,532 @@
...
+        self.status.eq_ignore_ascii_case("passed")
+    }
+
+    fn is_disabled(&self) -> bool {
+        self.status.eq_ignore_ascii_case("not run (disabled)")
+    }
+
+    fn is_failure(&self) -> bool {
+        !self.is_passed() && !self.is_disabled()
+    }
```
KuSh, line 249: **`***Skipped` tests are reported as failures.**

CTest prints `***Skipped` for `SKIP_RETURN_CODE` / `SKIP_REGULAR_EXPRESSION`, counts those as *passed* in the summary line, and lists them under `The following tests did not run:`. Since `is_failure()` catches everything that isn't `passed` or `not run (disabled)`, a fully green run produces:

```
ctest: 2/2 passed, 0 failed
failed:
  skipped_case (Skipped, 0.00 sec)
```

`0 failed` next to a populated `failed:` list, on a run that exited 0. On a mixed run it also inflated the header count to `3 failed`. Skips need their own bucket alongside `is_disabled()`.

Compounding it: `collect_failed_section` only captures from `The following tests FAILED:` onward, so the `The following tests did not run:` block — the only remaining evidence that the test was skipped rather than failed — is discarded as well.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,532 @@
...
+    let mut index = result_index + 1;
+
+    while index < lines.len() {
+        let line = lines[index];
+        if is_ctest_boundary(line) {
+            break;
+        }
+        block.push(line.trim_end().to_string());
+        index += 1;
+    }
```
KuSh, line 751: **Failure detail is uncapped, which undermines the savings target.**

This loop and `collect_failed_section` both copy every line verbatim with no line or byte limit. On my real run it emitted the full 15-line `Looked in the following places:` dump; a test printing thousands of lines under `--output-on-failure` passes through essentially unfiltered, so reduction can approach 0% precisely on the runs the filter exists to compress. A cap with a `… N more lines` marker, as the other filters here do, would cover it.

Related: the module has no token-savings assertion at all. `.claude/rules/cli-testing.md` treats <60% as a release blocker, and none of the six unit tests measures a ratio.

src/discover/rules.rs
```diff
@@ -330,6 +330,15 @@ pub const RULES: &[RtkRule] = &[
...
     },
+    RtkRule {
+        pattern: r"^ctest(?:\s|$)",
+        rtk_cmd: "rtk ctest",
+        rewrite_prefixes: &["ctest"],
+        category: "Tests",
+        savings_pct: 80.0,
+        subcmd_savings: &[],
+        subcmd_status: &[],
+    },
```
KuSh, line 353: **Blocks the build after rebase.** `RtkRule` gained a `pipeline_final_safe: bool` field on `develop`, and this literal spells out every field, so the merge fails with:

```
error[E0063]: missing field `pipeline_final_safe` in initializer of `RtkRule`
```

`false` is the right value for `ctest`. Neighbouring rules on `develop` now use `..RtkRule::DEFAULT` — worth adopting here so this entry survives the next field addition.

KuSh (changes requested):
Marking this as changes-requested to reflect the blocking items in my review just above — the two rebase-breakers (`lazy_static` removed from `develop` in `5269df7`, and the new `pipeline_final_safe` field on `RtkRule`) and the two behavioural bugs (`***Failed` result lines wrapped across a newline are never parsed, and `***Skipped` is counted as a failure).

No new findings here; details and code anchors are in the inline comments. Happy to re-review as soon as it's rebased onto `develop`.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,901 @@
...
+        }
+        return trimmed.to_string();
+    }
+
+    let failed_tests: Vec<&TestCase> = tests.iter().filter(|test| test.is_failure()).collect();
+    if failed_tests.is_empty() && summary.map_or(0, |s| s.failed) == 0 {
+        return format_success(&tests, summary, total_time);
+    }
+
+    format_failure(&lines, &tests, summary, total_time)
```
KuSh: 🔴 **Blocking (high) — `--repeat until-pass` reports a fully green run as failing.**

The route to `format_failure` triggers on *any* parsed failure result, ignoring what the summary says. CTest's retry line drops the `N/M` prefix:

```
    Test #9: flaky_case ...   Passed
```

so `TEST_RE` never sees the successful retry — only the first `***Failed`. Reproduced with `rtk ctest -R "flaky_case|pass_fast" --repeat until-pass:2` on a run where CTest printed `100% tests passed, 0 tests failed out of 2` and **exited 0**:

```
ctest: 2/2 passed, 0 failed
failed:
  flaky_case (Failed, 0.00 sec)
```

Same self-contradiction shape as the `***Skipped` bug we just fixed, one layer down. Flake-retry is a common CI configuration, so this reports a passing build as broken. The summary should win when it disagrees with the parsed results, and the retry result line needs to be recognised.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,901 @@
...
+    let before_result = lines[..result_index]
+        .iter()
+        .rposition(|line| START_RE.is_match(line))
+        .map_or(result_index, |index| index + 1);
+
+    block.extend(
+        lines[before_result..result_index]
+            .iter()
+            .map(|line| line.trim_end().to_string()),
+    );
```
KuSh: 🔴 **Blocking (medium) — under `ctest -j`, a failure's detail block absorbs other tests' result lines.**

The backward walk to the nearest `Start N:` copies the whole region verbatim. Unlike the forward loop below — which correctly breaks on `is_ctest_boundary` — this leading region is never filtered. With `-j`, CTest batches `Start` lines and interleaves results, so the span between the last `Start` and a failure's result line contains unrelated result lines.

Reproduced with real `rtk ctest --output-on-failure -j4`. `timeout_case`'s detail block was exactly:

```
5/8 Test #4: disabled_case ....................***Not Run (Disabled)   0.00 sec
6/8 Test #3: skipped_case .....................***Skipped   0.01 sec
```

Zero timeout diagnostics, plus two lines about two unrelated tests. Running the leading region through `is_ctest_boundary` the same way the trailing loop does should fix it.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,901 @@
...
+    });
+    slowest.truncate(MAX_SLOWEST);
+    slowest
+}
+
+fn failure_details(lines: &[String], failed_tests: &[&TestCase]) -> Vec<String> {
+    let mut blocks: Vec<String> = failed_tests
+        .iter()
+        .filter_map(|test| collect_failure_block(lines, test))
+        .collect();
```
KuSh: 🔴 **Blocking (medium) — detail blocks are unlabeled and silently dropped, so block N ≠ `failed[N]`.**

`filter_map` discards tests whose block comes out empty, and the survivors are emitted as anonymous `\n\n`-separated chunks. There's no way to tell which block belongs to which test.

Reproduced on the real serial run: four tests listed under `failed:` (`plain_fail`, `regex_fail`, `timeout_case`, `missing_case`) but only **three** blocks printed, because `timeout_case` produced none. So the third block — `missing_case`'s "Could not find executable" dump — reads as `timeout_case`'s output. Anyone reading top-to-bottom mis-attributes the diagnostics.

Prefixing each block with the test name would fix both halves.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,901 @@
...
+        out.push_str("\nfailed:");
+        for test in &failed_tests {
+            out.push_str(&format!(
+                "\n  {} ({}, {})",
+                test.name,
+                test.status,
+                format_seconds(test.duration)
+            ));
+        }
+    }
```
KuSh, line 554: 🔴 **Blocking (medium) — the `failed:` list is uncapped, defeating the `CAP_LIST` cap applied just below.**

`collect_failed_section` correctly truncates CTest's own `The following tests FAILED:` list at `MAX_FAILED_LIST_LINES` (20) with a `[see remaining: tail -n +21 …]` hint — but this loop prints every failing test unconditionally a few lines earlier.

Reproduced with a 120-test run (60 failures): the `failed:` block emits all 60 names, then the FAILED section emits the same names capped at 20. The names appear twice and the cap buys nothing. Either cap this list too, or drop the now-redundant FAILED section.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,901 @@
...
+fn append_skipped_list(out: &mut String, tests: &[TestCase]) {
+    let skipped_tests: Vec<&TestCase> = tests.iter().filter(|test| test.is_skipped()).collect();
+    if skipped_tests.is_empty() {
+        return;
+    }
+
+    out.push_str("\nskipped:");
+    for test in skipped_tests {
+        out.push_str(&format!("\n  {}", test.name));
+    }
```
KuSh, line 595: 🟡 **Non-blocking (low) — `append_skipped_list` is uncapped.**

Every skipped name is printed, with no `CAP_LIST` truncation and no tee hint, unlike every other list in this module. A project with many platform-gated tests (`SKIP_RETURN_CODE`) turns a *green* run into hundreds of lines — reproduced with 60 skips → 60 lines under `skipped:`.

Fine as a follow-up; just inconsistent with the caps you added elsewhere in this same round.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,901 @@
...
+
+    if block.len() > MAX_FAILURE_LINES {
+        let hidden = block.len() - MAX_FAILURE_LINES;
+        let full_block = block.join("\n");
+        rendered.push(format!("... +{hidden} more lines"));
+        rendered.extend(block[block.len() - MAX_FAILURE_LINES..].iter().cloned());
+        if let Some(hint) = crate::core::tee::force_tee_hint(&full_block, "ctest-failure") {
+            rendered.push(hint);
+        }
```
KuSh: 🟡 **Non-blocking (low) — tail-only truncation discards the informative half of pre-result diagnostics.**

Keeping the last `MAX_FAILURE_LINES` (10) is right for test stdout, where the assertion is at the end. It's backwards for CTest's *pre-result* diagnostics, where the signal is at the head.

Reproduced for `missing_case` on the real serial run: `Could not find executable missing-command-xyz` and `Looked in the following places:` are both dropped, and the 10 lines kept are near-identical `Debug/…`, `MinSizeRel/…`, `RelWithDebInfo/…` path repeats. The reader learns nothing without opening the tee file.

Keeping a head line or two alongside the tail would cover both shapes. Follow-up is fine.

tests/fixtures/ctest_mixed_raw.txt
```diff
@@ -0,0 +1,60 @@
...
+Debug/missing-command
+MinSizeRel/missing-command
+MinSizeRel/missing-command
+RelWithDebInfo/missing-command
+RelWithDebInfo/missing-command
+Deployment/missing-command
+Deployment/missing-command
+Development/missing-command
+Development/missing-command
+Unable to find executable: missing-command
```
KuSh: 🟡 **Non-blocking (low) — fixtures don't match the byte stream the filter actually receives.**

Real CTest writes `Unable to find executable: …` and `Errors while running CTest` to **stderr**, while result lines go to stdout. `run_captured_filter` hands the filter `raw = raw_stdout ++ raw_stderr` (`src/core/runner.rs:125`), i.e. all stderr concatenated *after* all stdout — never interleaved. This fixture places the line inline inside the `missing_case` block, a layout that cannot occur at runtime.

Worth mentioning because it's the root cause of the four blocking findings above: hand-written fixtures encode the line order you'd expect rather than the one CTest produces, so the `-j` interleaving and `--repeat` retry shapes were never exercised. `.claude/rules/cli-testing.md` asks for real captures for exactly this reason — a couple of real `-j4` and `--repeat until-pass:2` runs saved as fixtures would likely catch all four at once.

KuSh (changes requested):
Round 2 is a real improvement — thanks for the thorough turnaround. I verified every one of my six previous points by building the rebased branch and running the filter, not just reading the diff: `fmt` and `clippy --all-targets` clean, 2673 tests pass, and the wrapped-`Regex=[…]`, `***Skipped`, and discovery-failure cases all now produce correct output. All six threads are resolved. Adopting `..RtkRule::DEFAULT` and reusing `CAP_WARNINGS`/`CAP_LIST` was the right call.

I then did a deeper pass against real `ctest 4.3.4` output — a scratch CMake project exercising serial, `-j4`, `--repeat until-pass:2`, green, no-tests, `-N` and `--version` runs. Seven new findings, all reproduced against real output rather than fixtures.

**Blocking (1 high + 3 medium):** the `--repeat until-pass` misreport, the `-j` block-attribution bug, unlabeled detail blocks, and the uncapped `failed:` list.

**Non-blocking (3 low):** uncapped `skipped:` list, tail-only truncation of pre-result diagnostics, and fixture realism. Happy for these to land as follow-ups — flagging them so they're on record, not to hold the PR.

The common thread in the blocking four: parallel and retry modes reshape CTest's line stream in ways the hand-written fixtures don't reproduce. Capturing a couple of real `-j4` and `--repeat` runs as fixtures would likely catch all of them at once.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1028 @@
...
+                .map(|line| line.trim_end().to_string()),
+        );
+    }
+
+    let mut index = result_index + 1;
+    while index < lines.len() {
+        let line = &lines[index];
+        if is_ctest_boundary(line) {
+            break;
+        }
```
KuSh, line 744: 🔴 **Blocking — a failing test's own output truncates its failure block, deleting the assertion evidence with no marker.**

`is_ctest_boundary` breaks this loop on anything matching `START_RE` (`^\s*Start\s+\d+:`, line 32) — including a line the *test itself* prints. Reproduced against real `ctest` 4.3.4, with a test that echoes a nested phase marker before failing:

```
    Start 1: nested_out
1/1 Test #1: nested_out .......................***Failed    0.00 sec
setup ok
    Start 3: inner phase
ASSERT FAILED: values differ
expected 1 got 2
```

The filter emits:

```
ctest: 0/1 passed, 1 failed (0.00 sec)
failed:
  #1 nested_out (Failed, 0.00 sec)
    setup ok
```

`ASSERT FAILED: values differ` and `expected 1 got 2` are gone. The filter kept the useless half and silently dropped the evidence — the inverse of what it exists to do.

What makes this blocking rather than cosmetic is the **absence of any signal**. Because the truncated block lands under `CAP_WARNINGS`, there is no `... +N more lines` marker and no tee hint, so nothing tells the reader that anything was removed. A user seeing `setup ok` reasonably concludes that is all the test produced, and re-runs raw ctest only after wasted confusion. Compare the noisy-output path, which handles its truncation honestly.

Same exposure for a test that echoes a `N% tests passed, …` or `Total Test time (real) = …` line — plausible whenever a test wraps a nested build or a nested ctest invocation, which is exactly the kind of integration test whose failures are most expensive to lose.

Two directions, either is fine:

1. Require a stricter shape before treating a line as a boundary — e.g. only honour `Start` when it is followed by a result for that test, or anchor boundaries to column 0 / known CTest framing rather than any `^\s*Start N:`.
2. At minimum, mark the truncation: if the loop stops on a suspected boundary before the block ends naturally, emit the `... +N more lines` marker and the tee hint so the loss is visible.

(1) is the real fix; (2) alone would at least make it non-silent.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1028 @@
...
+
+    if block.len() > MAX_FAILURE_LINES {
+        let hidden = block.len() - MAX_FAILURE_LINES;
+        let full_block = block.join("\n");
+        // Preserve setup context at the head and assertion evidence at the tail.
+        let tail_lines = truncate::reduced(MAX_FAILURE_LINES, MAX_FAILURE_HEAD_LINES);
+        rendered.extend(block.iter().take(MAX_FAILURE_HEAD_LINES).cloned());
+        rendered.push(format!("... +{hidden} more lines"));
+        rendered.extend(block[block.len() - tail_lines..].iter().cloned());
+        if let Some(hint) = crate::core::tee::force_tee_hint(&full_block, "ctest-failure") {
```
KuSh: 🟡 **Non-blocking, but strongly wanted — consolidate the per-test tee writes into a single one; it fixes two defects at once.**

This calls `force_tee_hint` once *per failing test* with a constant `"ctest-failure"` slug. Three consequences, all measured:

**a. Every hint points at the same file, and all but the last are wrong.** `tee.rs:143` names files `{epoch_seconds}_{slug}.log` and opens with `.truncate(true)`, so within one second every call collides. Two real noisy failures:

```
  #1 aaa_first  … [full output: …/1787957048_ctest-failure.log]
  #2 zzz_second … [full output: …/1787957048_ctest-failure.log]
```

Byte-identical. That file contains **40 ZZZ lines and 0 AAA lines** — following `#1`'s hint hands you another test's diagnostics.

To be clear: **the collision itself is an rtk bug, not yours.** It is tracked at #3431 (`priority:high`) and I have added this single-process reproduction there. You should not have to wait on it — I raise it because consolidating the write sidesteps it entirely today.

**b. Blocks are built for every failure before the `CAP_LIST` cut** (`:419` maps all of `failed_tests`, then only 20 render). With 25 real failures I confirmed 25 blocks built and 25 writes attempted. Today the collision masks the cost; fix #3431 with a unique-name scheme and this becomes 25 distinct files against `DEFAULT_MAX_FILES = 20`, so a single ctest run would evict every other command's tee log. The two interact, which is why the write shape is worth changing now.

**c. The `ctest-failed` file at `:439` is labelled `[full output: …]` but is not.** `full_section` is assembled from `format_failed_entry` output that has already been head/tail-truncated. Measured on that 25-failure run: the file contains **25 `... +N more lines` markers**, and each hidden failure has **10 of its 30** evidence lines. Someone opening it to recover the failures past `... +5 more failed` still cannot reach the raw text.

**Suggested shape.** Write **one** tee containing the untruncated failure section, and drop the per-test hints in favour of the single section-level one. That is the convention essentially every other rtk filter already follows — `cargo_cmd.rs:1227` is the direct analogue, joining all failures into a single `cargo-test-failures` write. Of ~55 `force_tee_hint`/`force_tee_tail_hint` call sites, only two write in a loop: `search.rs:682`, which is safe because it built `grep_slug(idx, file)` for exactly this reason, and this one.

The one thing that makes it pay off twice: the merged payload must hold the **untruncated** blocks, which means splitting collection from rendering — have `collect_failure_block` return the full block, truncate only at render time, and tee the full join. That resolves (b) and (c) together, and makes the `[full output:]` label honest.

Non-blocking because none of it produces wrong *summary* output, and (a) is ultimately rtk's bug. Wanted because it is a small change that removes a whole class of problem and matches the existing pattern.

KuSh (changes requested):
Round 3 verified — thanks for the thorough turnaround, and for the rebase.

I checked the seven open items against the new head rather than the resolved flags, and **all seven are genuinely fixed**; every thread is now resolved. The branch rebases onto current `develop` (`322db4d`) and `cargo check --all-targets` is clean, so the two build-blockers from round 1 are actually gone rather than just invisible to CI. `cargo fmt`, `cargo clippy --all-targets` and 2678 tests all pass.

I did not take the new unit tests on trust — `ctest` 4.3.4 is installed here, so I generated real output and ran the built binary against it:

- **`--repeat until-pass`** — a real flaky-then-passing run now yields `ctest: 2/2 passed` with no `failed:` list. You fixed this at *both* layers (summary-based routing at `:148` and dedup-by-test-number in `parse_tests`), which is belt-and-braces and appreciated.
- **`-j` attribution** — a real `ctest -j4` with three failures and no intervening `Start` lines attributes every block correctly.
- **caps, labels, head+tail truncation, fixture ordering** — all confirmed.

Special credit for `ctest_repeat_until_pass_raw.txt`: it is byte-faithful to real ctest, including the retry line's missing `N/M` prefix and its four-space indent. That is exactly the detail that broke the old regex, and capturing it precisely is what made this round verifiable.

I also probed two things I thought might still be wrong and they turned out **not** to be defects — noting them so they don't get "fixed" later: a `Start` number that differs from the completing test's number is *correct* attribution (ctest prints `Start` on slot acquisition and output on completion), and the widened `TEST_RE` does not false-positive on test stdout because it stays `^`-anchored.

Two items left, both in the failure-detail path.

**One blocking** (`:479`): a test that prints a CTest-shaped line truncates its own failure block and deletes the assertion evidence, with no marker of any kind.

**One non-blocking but strongly wanted** (`:500`): the per-test tee write is the wrong shape, and consolidating it fixes two defects at once while bringing the filter in line with what every other rtk filter does.

On the second one — while investigating it I found a pre-existing rtk bug in `core/tee.rs` (the `{epoch_seconds}_{slug}.log` filename collides with itself), which is **not yours to fix**. It is tracked at #3431 and I have added the single-process reproduction there. I mention it only because it is the visible symptom that shows why the per-test write shape is worth changing, and the consolidation sidesteps it without waiting on that issue.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1222 @@
...
+
+fn parse_tests(lines: &[String]) -> Vec<TestCase> {
+    let mut tests: Vec<TestCase> = Vec::new();
+    for (line_index, line) in lines.iter().enumerate() {
+        let Some(test) = parse_test_line(line, line_index) else {
+            continue;
+        };
+        if let Some(existing) = tests
+            .iter_mut()
+            .find(|existing| existing.number == test.number)
```
KuSh: 🔴 **Blocking — result lines are the one framing category still unvalidated, and it costs a whole failure entry.**

This round hardened `Start` lines (number **and** name must match a parsed test, `:307-317`) and the summary / total-time / `FAILED:` markers (`rposition`, so only the last occurrence frames). Result lines got neither check — `:306` treats *any* line that parses as a result as framing, and this `.find()` matches on `number` alone. Three symptoms follow, all reproduced against real `ctest` 4.3.4.

### a. A failing test's entry disappears entirely (this is the blocking part)

A test that forwards a nested suite's output — ordinary in CMake integration suites:

```
    Start 1: wrapper
1/2 Test #1: wrapper ..........................***Failed    0.00 sec
delegating to inner suite
1/1 Test #1: inner_case ......................   Passed    0.02 sec
ASSERT FAILED: wrapper post-check
```

The filter emits:

```
ctest: 1/2 passed, 1 failed (0.01 sec)

Errors while running CTest
```

There is **no `failed:` section**. The header says one test failed; its name, status, duration and `ASSERT FAILED: wrapper post-check` are all gone. The echoed line parses as test #1, and this `.find()` replaces the real `wrapper/Failed` record with `inner_case/Passed`.

This is the round-3 dedup (which correctly fixed `--repeat until-pass`) meeting an unvalidated result line — not a mistake in either change on its own.

### b. The block truncates even once (a) is fixed

Because `:306` also marks the echoed line as framing, `collect_failure_block` stops there. With (a) fixed I still get only `delegating to inner suite`; `ASSERT FAILED: wrapper post-check` is still dropped. Both halves need closing.

### c. Phantom entries can inflate totals

The echoed line also lands in `tests`. With a summary present the counts come from it, so this is limited to runs with no summary line — an interrupted run, or output truncated by a pipe. Narrow, and listed only for completeness.

*(I checked the related worry that phantoms pollute `slowest:` on green runs — they cannot. `slowest:` renders only in `format_success`, and a passing test's stdout requires `-V`, which `should_passthrough` already routes around the filter. No action needed there.)*

### Suggested fix

For (a), match on name as well — one line:

```rust
.find(|existing| existing.number == test.number && existing.name == test.name)
```

I verified this against real ctest: it restores `#1 wrapper (Failed)` **and** keeps `--repeat until-pass` at `2/2 passed`, because a genuine retry repeats both number and name (`Test #2: flaky` twice) while a spoof does not. It also mirrors the number+name rule you already wrote for `START_RE`, so the two categories become consistent.

For (b), the same validation idea applied at `:306` — a result line should have to look like it belongs to *this* run before it is trusted as framing. The `N/M` counter is one discriminator: real result lines share a constant `M` (`1/2`, `2/2`), while the nested line above carried `1/1`. Your call on the mechanism.

---

**Also blocking — everything outside the parsed regions is discarded (`:134`).**

Once any test parses, output is rebuilt from the counts, the `failed:` / `skipped:` sections and the trailer. Anything else is dropped, including genuine ctest diagnostics. Stock `ctest -T Test` with no dashboard config:

```
Cannot find file: .../DartConfiguration.tcl
Cannot find file: .../DartConfiguration.tcl
Test project ...
1/1 Test #1: ok_case ..........................   Passed    0.01 sec
100% tests passed, 0 tests failed out of 1
```

becomes:

```
ctest: 1/1 passed (0.01 sec)
slowest:
  ok_case 0.01 sec
```

Both diagnostics are gone and the run looks perfectly healthy, so a broken dashboard configuration is invisible. `-D` and `-T` are also absent from `should_passthrough` (`:85`), so dashboard and `--build-and-test` runs go through the filter rather than around it. Either preserve unrecognised non-framing lines outside the test region, or add the dashboard flags to the passthrough list — the latter is smaller and consistent with how `-V` / `--show-only` are already handled.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1222 @@
...
+        format_seconds(test.duration)
+    );
+    if let Some(reason) = &test.reason {
+        entry.push_str("\n    ");
+        entry.push_str(reason);
+    }
+    for detail in details {
+        entry.push_str("\n    ");
+        entry.push_str(detail);
+    }
```
KuSh, line 700: 🟡 **Non-blocking — folded multi-line results render flush-left and evade the failure-line cap.**

`build_logical_lines` (`:186`) joins up to `MAX_RESULT_CONTINUATION_LINES` physical lines into one logical line for anything matching `RESULT_PREFIX_RE` without a duration — correct for CTest's wrapped `Regex=[…]` reasons. But when such a logical line lands in a failure block, this loop prefixes only the **first** physical line with `"\n    "`; the embedded newlines pass through unindented.

Real ctest, a test printing `Test #99: phase begin`, six lines, then `phase done in 1.5 sec`:

```
  #1 wrapped_out (Failed, 0.00 sec)
    Test #99: phase begin
L1
L2
L3
L4
L5
L6
phase done in 1.5 sec
    ASSERT FAILED: real
```

Two effects: the nesting breaks (those lines read as if they were outside the entry), and the eight physical lines count as **one** against `MAX_FAILURE_LINES`, so a block can exceed the 10-line budget without triggering truncation or the tee hint.

Splitting on `'\n'` when pushing block lines fixes both at once — each physical line gets its own indent and its own place in the count.

Non-blocking: cosmetic plus a soft cap overrun, no wrong data.

KuSh (changes requested):
Round 4 verified — both previous items are genuinely fixed, and the approach you took on the framing one is better than what I suggested.

Rather than patching the boundary check, `collect_framing_line_indices` validates framing up front: a `Start N: name` counts only when a parsed test has that number **and** name, and the summary / total-time / `FAILED:` markers are taken with `rposition` so only the last occurrence frames. My original real-ctest repro now keeps its assertion lines, and `ctest_spoofed_framing_raw.txt` covers both vectors including a spoofed `100% tests passed` inside a test's output.

The tee consolidation is exactly the shape we discussed — `collect_failure_block` is pure, `render_failure_block` truncates only at render time, and `build_failed_section` collects every block untruncated so one tee holds the whole section. Measured against the same fixtures I used last round:

| | round 3 | round 4 |
|---|---|---|
| tee files, 2 noisy failures | 2 writes, 1 colliding file | **1 file** |
| its content | 40 ZZZ, **0 AAA** | **40 AAA + 40 ZZZ** |
| hidden failures' evidence (25-failure run) | 10 of 30 each | **30 of 30 each** |
| `... more lines` markers inside "full output" | 25 | **0** |

No regressions: real `--repeat until-pass` still reports `2/2 passed`, real `ctest -j4` attribution is still correct, green/skipped runs unchanged. Rebased cleanly onto current `develop` (`d9a5893`), `cargo check --all-targets` clean, 2681 tests + clippy + fmt green. Both threads resolved.

Two items this round, both found by probing the areas the refactor touched.

**Blocking (`:241`)** — result lines are now the one framing/parsing category left unvalidated, and that single gap produces three distinct symptoms, one of which erases a failure entirely.

**Non-blocking (`:542`)** — folded multi-line results render flush-left and evade the line cap.

A note on process, since this is the fourth round: the blocking item is a *consequence* of the round-3 dedup fix meeting the round-4 framing work, not a regression you introduced carelessly. Each round has been correct about the case it targeted. The suggestion at the end of the blocking comment is a two-line change I verified against real ctest, so this should be a short round.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1448 @@
...
+        .map(|(total, _)| total)
+}
+
+/// CTest includes disabled tests in the `N/M` counter but excludes them from
+/// the summary total. Reconcile that documented mismatch before validation.
+fn counter_total_with_disabled_tests(
+    summary_total: u32,
+    most_frequent_total: Option<u32>,
+    candidates: &[TestCase],
+) -> u32 {
```
KuSh: 🔴 **Blocking — `--stop-on-failure` erases every failure detail. This is a regression from round 4.**

This function reconciles the `N/M` counter against the summary total for *disabled* tests only. Under `--stop-on-failure` the two disagree for a different reason: `M` is the number of tests **scheduled**, while the summary counts the tests that **ran**. Real ctest 4.3.4, seven tests, second one fails:

```
    Start 1: a
1/7 Test #1: a ................................   Passed    0.01 sec
    Start 2: b_fail
2/7 Test #2: b_fail ...........................***Failed    0.00 sec
REAL-EVIDENCE-b_fail-broke

50% tests passed, 1 tests failed out of 2

The following tests FAILED:
	  2 - b_fail (Failed)
```

`summary_total` is 2, `most_frequent_total` is 7, no disabled tests, so `2 + 0 != 7` and this returns `summary_total = 2`. Every result line carries `counter_total = 7`, so `parse_tests` rejects **all** of them and `tests` comes back empty:

```
ctest: 1/2 passed, 1 failed (0.01 sec)

Errors while running CTest
```

The count is right; which test failed and why are both gone. `REAL-EVIDENCE-b_fail-broke` and the name `b_fail` are nowhere in the output.

I want to be explicit that this is a **regression, not a pre-existing gap**: I rebuilt the round-4 binary from `6cf56a9c` and ran the same input through it —

```
ctest: 1/2 passed, 1 failed (0.01 sec)
failed:
  #2 b_fail (Failed, 0.00 sec)
    REAL-EVIDENCE-b_fail-broke
```

— so the counter validation that correctly closed the nested-suite spoof took `--stop-on-failure` with it. It is a common flag in CI, which is what makes this blocking.

The reconciliation needs to tolerate a counter total that is *larger* than the summary total for reasons other than disabled tests: `--stop-on-failure` (and `-j` with an early abort) legitimately schedule more than they run. Treating `counter_total >= summary_total` as consistent, rather than requiring an exact match after adding disabled tests, would cover both. Whatever the mechanism, the invariant worth keeping is that **a result line should never be rejected merely because fewer tests ran than were scheduled.**

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1448 @@
...
+    while index < physical_lines.len() {
+        let line = physical_lines[index];
+        if !RESULT_PREFIX_RE.is_match(line) || RESULT_TERMINATOR_RE.is_match(line) {
+            logical_lines.push(line.to_string());
+            index += 1;
+            continue;
+        }
+
+        let mut result_end = None;
+        for continuation in 1..=MAX_RESULT_CONTINUATION_LINES {
```
KuSh: 🔴 **Blocking — a result line wrapped over more than 8 physical lines drops the test and corrupts the previous one.** (Pre-existing, not from this round.)

CTest prints one physical line per alternative in a `PASS_REGULAR_EXPRESSION` / `FAIL_REGULAR_EXPRESSION` list, so a 10-alternative property wraps the result line across 11 physical lines — past `MAX_RESULT_CONTINUATION_LINES`, so `build_logical_lines` gives up and leaves it unfolded. Real ctest 4.3.4:

```
1/2 Test #1: a_fail ...........................***Failed  Error regular expression found in output. Regex=[EVIDENCE]  0.00 sec
REAL FAILURE EVIDENCE for a_fail
EVIDENCE

    Start 2: b_regex
2/2 Test #2: b_regex ..........................***Failed  Required regular expression not found. Regex=[r1
r2
...
r10
]  0.00 sec
hello from b_regex
```

filters to:

```
ctest: 0/2 passed, 2 failed (0.00 sec)
failed:
  #1 a_fail (Failed, 0.00 sec)
    Error regular expression found in output. Regex=[EVIDENCE]
    REAL FAILURE EVIDENCE for a_fail
    EVIDENCE
    ... +6 more lines
    r5
    ...
    ]  0.00 sec
    hello from b_regex
```

Three things go wrong at once. The header says `2 failed` but only one entry is listed. `b_regex` never parses, so it is absent entirely. And because it is absent from `tests`, its `Start 2:` line no longer qualifies as framing in `collect_framing_line_indices` — so `a_fail`'s block swallows `b_regex`'s result line **and** its stdout, while `a_fail`'s own evidence is pushed up into the truncated head.

That last part is the sting: the block-swallowing is a second-order effect of the number+name framing validation, so raising the constant alone only moves the threshold. A test with 15 alternatives would reproduce it.

Two directions: bound the fold by *content* rather than a line count (keep folding while the continuation looks like regex-list interior — no `Start`, no result prefix, no summary — which is already what the loop checks, just without the ceiling), or treat a `RESULT_PREFIX_RE` line that never found its terminator as framing anyway, so it at least bounds the neighbouring block instead of vanishing into it.

Worth saying plainly: this one is **older than this round** — I confirmed byte-identical output from the round-4 binary — so it is not something the last change broke.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1448 @@
...
+
+    let mut out = format!("ctest: {passed}/{total} passed, {failed} failed");
+    if skipped > 0 {
+        out.push_str(&format!(", {skipped} skipped"));
+    }
+    if disabled > 0 {
+        out.push_str(&format!(", {disabled} disabled"));
+    }
+    out.push_str(&format_meta(total_time));
+    if !failed_tests.is_empty() {
```
KuSh, line 547: 🔴 **Blocking — when the summary reports failures but nothing parsed, the failing test names are dropped twice over.**

This guard skips the whole `failed:` section when `failed_tests` is empty. The natural fallback would be the raw trailer, but `collect_failed_trailer` (`:706`) filters out exactly the `\t`-indented lines:

```rust
.filter(|line| !line.starts_with('\t'))
```

and those tab-indented lines *are* the failing test names:

```
The following tests FAILED:
	  2 - b_fail (Failed)
```

So in both blocking cases above, ctest told us plainly which tests failed and the filter discarded that twice: once by not parsing the result line, once by stripping the trailer. That is why `--stop-on-failure` output contains no trace of `b_fail` at all.

**Suggested fix, and I would take this one independently of the other two:** when `summary.failed > 0` and `failed_tests.len() < summary.failed`, keep the raw `The following tests FAILED:` entries instead of dropping them. Something like:

```
ctest: 1/2 passed, 1 failed (0.01 sec)
failed (unparsed, raw):
    2 - b_fail (Failed)
```

That turns both bugs above from *silent total loss* into *degraded but usable*, and — more importantly — it guards the next parsing gap too, whatever it turns out to be.

This is the structural point behind five rounds of findings. The filter reconstructs its output from parsed records and discards everything it did not parse, so **every** parsing gap becomes silent loss of exactly the content the user ran ctest to see. A general fallback keyed on "the summary disagrees with what I parsed" is the one change that stops that class rather than retiring one instance of it. The savings cost is a handful of lines on runs that are already failing.

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1448 @@
...
+                | "--verbose"
+                | "--extra-verbose"
+                | "--debug"
+                | "--show-only"
+                | "--print-labels"
+                | "-M"
+                | "-T"
+                | "-S"
+                | "-SP"
+                | "--build-and-test"
```
KuSh: 🟡 **Non-blocking — only the short dashboard flags pass through; the documented long forms still get filtered.**

`-D`, `-M`, `-T`, `-S`, `-SP` are matched here, but CTest documents long forms for all of them — `--dashboard`, `--test-model`, `--test-action`, `--script`, `--script-new-process` — and none are listed. The same invocation is handled two opposite ways. Real ctest, same directory, same action:

```
$ rtk ctest -T Test
Cannot find file: .../DartConfiguration.tcl
Cannot find file: .../DartConfiguration.tcl
Test project ...
1/1 Test #1: ok_case ..........................   Passed    0.00 sec

$ rtk ctest --test-action Test
ctest: 1/1 passed (0.01 sec)
slowest:
  ok_case 0.01 sec
```

The short form preserves the misconfiguration diagnostics; the long form drops them and reports a clean run. Adding the long spellings alongside the short ones is a one-line change.

(Note `arg.starts_with("-D")` is prefix-matched and so already covers `-DExperimental`, but it does not cover `--dashboard`, which is a different prefix.)

src/cmds/system/ctest_cmd.rs
```diff
@@ -0,0 +1,1448 @@
...
+
+    ParsedTests {
+        tests,
+        result_lines,
+    }
+}
+
+/// Ties keep the first total seen: ctest prints the run's own result lines
+/// before any output a test could forward.
+fn most_frequent_counter_total(candidates: &[TestCase]) -> Option<u32> {
```
KuSh: 🟡 **Non-blocking — a killed run whose test forwards nested results is reported as fully green.**

When there is no summary line, `run_total` falls back to this majority vote, and a test that forwards more result lines than the run itself has will outvote the real ones.

Verified with real ctest terminated by a signal — I checked both `SIGTERM` (what a CI step timeout sends) and `SIGKILL`, and neither leaves a summary line:

```
1/2 Test #1: outer ............................***Failed    0.00 sec
forwarding inner suite:
1/3 Test #1: inner_a ......................   Passed    0.01 sec
2/3 Test #2: inner_b ......................   Passed    0.01 sec
3/3 Test #3: inner_c ......................   Passed    0.01 sec
```

The three `M=3` lines outvote the run's own two `M=2` lines, so `run_total` becomes 3 and the real records are rejected:

```
ctest: 3/3 passed
slowest:
  inner_a 0.01 sec
  inner_b 0.01 sec
  inner_c 0.01 sec
```

A run that was killed with a failure already recorded reports fully green, naming three tests that do not exist in the project.

Non-blocking on reachability: it needs a terminating signal *and* a test that forwards more results than the run has, and `run_filtered` still propagates a non-zero exit code, so the status is right even when the text is not. Worth noting the exit code is the only thing carrying the truth here, which matters more for an agent reading stdout than for a human reading the terminal.

For what it is worth, the `:492` fallback does not rescue this one — there is no summary to disagree with. Preferring the *first* counter total seen over the most frequent would (ctest prints its own result lines before any a test could forward), and your `most_frequent_counter_total` doc comment already notes that ordering property for tie-breaking.

KuSh (changes requested):
Round 5 verified — all four items from the last round are genuinely fixed, and the `N/M` validation went further than I sketched.

Confirmed against real `ctest` 4.3.4, not just the fixtures: the nested-result spoof now keeps `#1 wrapper (Failed)` **and** its `ASSERT FAILED` line, with the forwarded line correctly demoted from framing to content; `--test-action`'s short form `-T` passes through; and the folded multi-line block is indented per physical line and counted per physical line. `--repeat until-pass`, `-j4` attribution, the tee consolidation (1 file, 40+40, 0 truncation markers) and the parallel fixture's counter-vs-summary disagreement all still behave. 2766 tests, clippy and fmt clean on current `develop`. Both threads resolved.

`counter_total_with_disabled_tests` deserves a specific mention: reconciling CTest's counter (which includes disabled tests) against the summary total (which excludes them) is a subtlety I had only noticed in passing, and handling it explicitly is what keeps the parallel fixture correct.

That same reconciliation is where this round's blocking item comes from, and I want to be straight about the shape of it: **`--stop-on-failure` worked in round 4 and does not work now**. It is a genuine regression from the fix, not something that was always broken — I rebuilt the round-4 binary to confirm. The second blocking item is older than this round; the third is the one that would have contained both.

Five items: three blocking, two not.

- **`:325`** — `--stop-on-failure` erases every failure detail (regression).
- **`:218`** — a result line wrapped past 8 physical lines drops a test and corrupts the previous one (pre-existing).
- **`:492`** — no fallback when the summary reports failures but nothing parsed. This is the structural one.
- **`:112`** — only the short dashboard flags pass through (non-blocking).
- **`:301`** — a killed run plus forwarded results reports green (non-blocking).

A note on the arc, since this is round five. Rounds 3, 4 and 5 each fixed their target correctly and each opened a new interaction, and that is not carelessness — it is the design. Output is *reconstructed* from parsed records and everything unparsed is discarded, so every gap in parsing becomes silent, total loss of the actionable content rather than degraded output. `:492` is the one change here that breaks that pattern instead of patching another instance of it, and I would take it even if you disagree with my read of the other two.

KuSh (approved):
Approving.

Round 6 landed findings 1–4, and I checked each against real ctest 4.3.4 rather than the fixtures: `--stop-on-failure` details are back, a 15-alternative `PASS_REGULAR_EXPRESSION` no longer drops its test at eight physical lines or swallows the next block, the raw `FAILED:` trailer contains a parsing gap instead of erasing it, and `--test-action` matches `-T`.

Rather than send a seventh round for what was left, I pushed `143dc8c` to this branch — **please `git pull` before you touch it again.**

### Finding 5, the one still open

Result lines are validated against the run total; the summary is not, and `has_failures` trusts a summary over the parsed records. A test that forwards a nested ctest run echoes that run's summary too, so a run killed before printing its own reports the nested one. A SIGKILLed 6-test suite whose first test had already failed rendered `ctest: 3/3 passed`. `first_counter_total_rejects_forwarded_suite_in_killed_run` passes because it omits the nested summary — real forwarded output always ends with one.

The rule that holds is ordering, not size: ctest prints its summary once every test has finished, so a validated `Start` *behind* a summary places that summary inside a test's output. I first tried bounding the total by `run_total`, and that only catches nested suites larger than the run; `tests/fixtures/ctest_killed_smaller_forwarded_suite_raw.txt` is the real capture that defeats it, kept as a fixture so the weaker version can't come back. The two numeric bounds are still there as a backstop for a kill landing between the forwarded summary and the next `Start`.

### Four more, each reproduced with real ctest

- `split_status_reason`'s global `.replace(" ]", "]")` rewrote brackets belonging to the reason. `PASS_REGULAR_EXPRESSION "zzz[0-9 ]end"` rendered as `Regex=[zzz[0-9]end]` — the character class quietly loses the space it matches on, and you debug a pattern ctest never ran. Only the fold's own trailing `" ]"` is undone now.
- `--no-tests=error` exits 8 and says `Errors while running CTest`, but the run rendered as a bare `ctest: no tests found`: the one line that reads benign for a run that failed. The trailer is kept, as it already is behind a failure.
- `-T Test` is the canonical CI invocation and prints ordinary test output, yet it bypassed the filter with every other dashboard mode, so the most common form saved nothing. The action decides now — `Test` stays filtered, everything else still bypasses (`-T Coverage` opens with `Performing coverage`), and a `-M` with no action to pair it with is left alone.
- `looks_like_ctest_output` required `Test project` to be the first non-empty line, but `-T Test` emits a repeated `Cannot find file: …/DartConfiguration.tcl` ahead of it, so `ctest -T Test 2>&1 | rtk pipe` fell through to identity.

Two housekeeping items came along: the failed list now deviates to `reduced(CAP_LIST, 5)` because its entries are multi-line, which is the case `core/README.md` reserves `reduced()` for (25 noisy failures went 247 → 187 lines, 67.9% → 75.8% off the raw output), and tracking moved `ctest` off `system` so `rtk gain` stops filing C/C++ test runs beside `ls` and `grep`.

`cargo fmt --all && cargo clippy --all-targets && cargo test --all` is clean, 2701 tests pass.

Thanks for staying with this through six rounds. The framing validation from round 4 and the fold-until-terminator plus raw-trailer fallback from round 6 were the hard parts, and they're what made the rest tractable — every one of the remaining items was a single unvalidated input, not a structural problem.

PR #2254, opened 2026-06-04
fix(grep): strip grep -r/-R/-E before ripgrep to prevent silent match mangling

### LINKED ISSUE #2253: grep/rg: matches silently rewritten ("def" → "n") — grep -rn/-E collide with ripgrep's value-taking -r/--replace and -E/--encoding
### Summary

`rtk grep` (the rewrite target for both `grep` and `rg`) forwards user flags to ripgrep, but `grep` and `ripgrep` assign different meanings to several flag letters — and ripgrep's versions are **value-taking**, so the clash silently corrupts output:

| grep flag | meaning in grep | meaning in ripgrep |
|---|---|---|
| `-r` / `-R` | recursive (boolean) | `--replace` (**takes a value**) |
| `-E` | extended regex (boolean) | `--encoding` (**takes a value**) |

A muscle-memory invocation therefore reaches rg as a replace/encoding directive that consumes the following token (often the pattern):

```
$ printf 'def foo():\n' > a.py

$ grep -rn def a.py
a.py:1:n foo():        # "def" silently rewritten to "n"  (rg saw --replace=n)

$ grep -nE 'def' a.py
rg: error parsing flag -E: grep config error: unknown encoding: def
```

The first case is the dangerous one: **no error, just wrong content**. A reader (human or AI agent) concludes the symbol is named `n`, searches for it, finds nothing, and draws wrong conclusions about the code. This happens for *any* `-rn` search — `reasoning` → `n`, `load_compiled_rules` → `n`, etc.

### Root cause

In `src/cmds/system/grep_cmd.rs`, the extra-args loop only skips a **standalone** `-r`/`--recursive`; combined bundles (`-rn`, `-rln`, `-nE`) and `-E` fall through to ripgrep verbatim.

### Reproduction

```bash
printf 'def foo():\n' > /tmp/a.py
grep -rn def /tmp/a.py     # expect: def foo():   actual: n foo():
grep -nE def /tmp/a.py     # expect: def foo():   actual: rg encoding error
```

### Proposed fix

Drop grep's `-r`/`-R`/`-E` (and the `--recursive`/`--dereference-recursive`/`--extended-regexp` long forms) before forwarding to ripgrep — including from combined short bundles — while preserving value-taking flags and their values. rg recurses and is ERE by default, so dropping them is safe. PR linked below.

### Related (separate, not in the linked PR)

Two adjacent grep→rg translation gaps worth tracking:

1. **Flag collision on `-l`/`-t`/`--type`/`-g`/`--glob`**: idiomatic `rg -l PAT --type py` (flags before the pattern) fails arg parsing and falls back to raw `grep`, which then rejects `--type`/`--glob`.
2. **Truncation through a redirect**: `grep PAT file > out.txt` writes the capped (`[+N more]`) view into the file, silently dropping matches the caller treats as complete.

Happy to send follow-up PRs for these if of interest.

PLNech:
### Problem

`grep_cmd` forwards extra args to ripgrep but only skipped a standalone `-r`. `grep` and `rg` disagree on several flag letters, and the clash is destructive because rg's variants take a value:

- `grep -r`/`-R` (recursive) → rg `-r` is `--replace` → `grep -rn PATTERN` becomes `--replace=n`, rewriting every match to `n` (e.g. `def foo()` → `n foo()`).
- `grep -E` (ERE) → rg `-E` is `--encoding` → `grep -nE PATTERN` feeds PATTERN to `--encoding` ("unknown encoding: PATTERN").

Both are silent/confusing corruptions of read-only code inspection. Details and repro in #2253.

### Fix

Generalize the extra-args sanitizing into `strip_grep_only_flags`, which drops grep's `-r`/`-R`/`-E` (and the matching long forms) before forwarding to rg — including from combined short bundles (`-rn` → `-n`, `-nE` → `-n`) — while preserving value-taking flags and their values (`-A3`, `-tpy`, `-erare`). rg recurses and is ERE by default, so dropping these is safe.

### Tests

Unit coverage for bundles, long forms, and value-flag preservation (`test_strip_grep_only_flags_bundles`, updated `test_recursive_flag_stripped`). `cargo clippy` clean. One file changed.

Closes #2253.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -43,11 +43,16 @@ pub fn run(
     }
 
     for arg in extra_args {
-        // Fix: skip grep-ism -r flag (rg is recursive by default; rg -r means --replace)
-        if arg == "-r" || arg == "--recursive" {
-            continue;
+        // Strip grep flags whose letter ripgrep reuses for a DIFFERENT, value-taking
+        // flag — otherwise rg silently consumes the next token (often the pattern):
+        //   grep -r/-R (recursive) -> rg -r is --replace  (e.g. `grep -rn def` became
+        //                                                   `--replace=n`, rewriting matches)
+        //   grep -E    (ERE)       -> rg -E is --encoding  (eats the pattern)
+        // rg recurses and is ERE by default, so dropping them is safe. Handles combined
+        // short bundles (-rn -> -n, -nE -> -n) while preserving value-flags' values.
+        if let Some(sanitized) = strip_grep_only_flags(arg) {
+            rg_cmd.arg(sanitized);
         }
-        rg_cmd.arg(arg);
     }
 
     let result = exec_capture(&mut rg_cmd)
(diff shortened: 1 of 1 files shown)


PLNech:
CLA signed! Thanks for the great tool :robot: :grey_heart:

KuSh:
Thanks for the clear diagnosis, the -r/--replace and -E/--encoding collision was real. It was resolved a different way in #2641 (merged 2026-06-26), which stopped substituting ripgrep for grep: rtk grep now runs real grep with your flags untouched and rtk rg runs real ripgrep, so grep-only flags never reach rg and there is nothing left to strip. On current develop rtk grep -rn foo . and rtk grep -nE 'fo+' file return correct matches with no rewriting. Note that #2253 stays open for a separate residual on the ripgrep side, rtk rg -r REPL world file still misreads the short-form value flag, which is a different code path from the one this patch touched, so that one is worth tracking there. If you think part of your fix is still needed on develop, please comment here or reopen.

PR #2255, opened 2026-06-04
feat(hooks): add native Windows PowerShell hook (rtk-rewrite.ps1)

danscMax:
### Problem

On native Windows (no WSL), Claude Code has no working auto-rewrite hook — the docs note the bash hook `rtk-rewrite.sh` "requires a Unix shell", so Windows falls back to CLAUDE.md-injection and loses automatic rewriting.

A naive PowerShell port that simply calls `& rtk rewrite $cmd` does **not** work on **Windows PowerShell 5.1**: when a string is passed as a native-command argument, PowerShell 5.1 strips embedded double-quotes, so the command is corrupted *before* rtk sees it. Example (verified on PowerShell `5.1.26100`):

```
INPUT : git diff; echo "a (b)"
& rtk rewrite $cmd  ->  rtk git diff; echo a (b)     # quotes gone -> bash: syntax error near `('
```

rtk itself is blameless — given the command via argv correctly (e.g. from bash), `rtk rewrite 'git diff; echo "a (b)"'` returns `rtk git diff; echo "a (b)"` with quotes intact. The corruption is purely PowerShell argument passing.

### Fix

Add `hooks/claude/rtk-rewrite.ps1` that builds the child-process command line itself using MSVCRT / CommandLineToArgvW quoting (via `ProcessStartInfo`), so the exact command reaches `rtk rewrite` untouched on both Windows PowerShell 5.1 and PowerShell 7+. Behavior mirrors `rtk-rewrite.sh` exactly:

- same exit-code protocol (`0` auto-allow / `1` passthrough / `2` deny / `3` ask);
- same `hookSpecificOutput` (`updatedInput.command`, `permissionDecision: allow` for exit 0);
- fail-safe: any error path exits `0` (passthrough), so the hook can never break command execution.

### Testing

Verified end-to-end (hook spawned as a separate process with tool_input JSON on stdin, mirroring Claude Code) — quotes/parens/`&&`/pipes all survive:

| input | rewritten command |
|---|---|
| `git diff; echo "a (b)"` | `rtk git diff; echo "a (b)"` |
| `git status && grep -n "def foo(" .` | `rtk git status && rtk grep -n "def foo(" .` |
| `git commit -m "fix(ci): pin (1M)"` | `rtk git commit -m "fix(ci): pin (1M)"` |
| `git log --oneline \| grep "fix(x)"` | `rtk git log --oneline \| grep "fix(x)"` |
| `echo "plain text, no rewrite target"` | (passthrough, unchanged) |

A smoke test is included at `hooks/claude/test-rtk-rewrite.ps1`.

### Install (PreToolUse, matcher "Bash")

```
powershell.exe -ExecutionPolicy Bypass -NoProfile -File <path>\rtk-rewrite.ps1
```

### DIFF
--- a/hooks/claude/rtk-rewrite.ps1
+++ b/hooks/claude/rtk-rewrite.ps1
@@ -1,0 +1,126 @@
+# RTK PreToolUse hook for Claude Code on Windows (native PowerShell).
+# Counterpart of hooks/claude/rtk-rewrite.sh for users who run Claude Code on
+# native Windows (no WSL / Unix shell). Requires: rtk >= 0.23.0 in PATH.
+#
+# Why a dedicated implementation instead of a thin `& rtk rewrite $cmd` wrapper:
+# Windows PowerShell 5.1 strips embedded double-quotes when a string is passed as
+# a native-command argument, so `& rtk rewrite $cmd` corrupts the command before
+# rtk sees it. For example the input
+#     git diff; echo "a (b)"
+# reaches rtk as
+#     git diff; echo a (b)
+# and the rewritten result `... echo a (b)` is invalid shell (unbalanced "("),
+# which breaks the very command the hook was meant to optimize. (rtk's own output
+# is correctly quoted; the corruption happens purely in PowerShell argument
+# passing.) This script therefore builds the child process command line itself
+# using MSVCRT / CommandLineToArgvW quoting, so the exact command reaches
+# `rtk rewrite` untouched on both Windows PowerShell 5.1 and PowerShell 7+.
+#
+# Install (PreToolUse, matcher "Bash"):
+#   powershell.exe -ExecutionPolicy Bypass -NoProfile -File <path>\rtk-rewrite.ps1
+#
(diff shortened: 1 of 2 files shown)


racinmat:
@aeppling Hi, could you please take a look at this PR? It would be very useful for me. But maybe it would be better to merge #2331 instead, it contains relevant documentation and other changes.

danscMax:
Rebased onto current `develop` — it was 355 commits behind, so CI here was last exercised against a fairly old tree. It rebased without conflicts, and the full suite passes (2520 tests) on the rebased branch. Only two files are added, both under `hooks/claude/`; no Rust code is touched.

@racinmat asked back in June whether this or #2331 was the better vehicle. For whatever it is worth from the author's side: #2331 is 41 files and roughly 6000 lines and is currently showing as conflicting, while this one is two new files and touches nothing that exists. They are not really alternatives — if #2331 lands, this becomes redundant and I will close it myself; until then this is the smaller thing to merge for anyone on native Windows without WSL, where Claude Code currently has no working auto-rewrite hook at all.

Happy to rebase again or split it further if that helps review.

PR #2263, opened 2026-06-04
fix(permissions): split commands on newlines to close allow-rule bypass

### LINKED ISSUE #2262: Security: command lexer doesn't split on newlines, so the Claude hook auto-allows dangerous trailing statements
# Security: command lexer doesn't split on newlines → `rtk hook claude` auto-allows dangerous trailing statements

## Summary

The Claude Code `PreToolUse` hook (`rtk hook claude`) decides `permissionDecision: "allow"` by splitting the command into segments and requiring every segment to match a user allow rule. But the lexer **does not treat a newline as a command separator** — only `|`, `||`, `;`, `&&`. A newline-separated, multi-statement command therefore collapses into a single segment, which is matched by **prefix** on the command at its start. If that leading command is allow-listed (`git log`, `cat`, `ls`, `tail`, `grep`, …), the **entire block is approved** — including arbitrary trailing statements such as `rm -rf /`, `git push --force-with-lease`, or `git commit --amend` — with no prompt, bypassing the user's allow/deny rules.

## Reproduction

A safe leading command launders a destructive one:

```bash
printf '%s' '{"tool_name":"Bash","tool_input":{"command":"cat /etc/hostname\nrm -rf --no-preserve-root /"}}' \
  | rtk hook claude
```

```json
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk read /etc/hostname\nrm -rf --no-preserve-root /"},"permissionDecision":"allow"}}
```

`permissionDecision: "allow"` is returned for the whole block, so `rm -rf --no-preserve-root /` runs without a prompt — even though `rm` is not in any allow rule. The same is true for `git push --force-with-lease`, `git commit --amend`, `dd …`, `chmod -R 777 /`, `eval …`, etc.

The same dangerous command on its own is correctly **not** auto-allowed (no `permissionDecision`, so Claude Code prompts):

```bash
printf '%s' '{"tool_name":"Bash","tool_input":{"command":"git push --force-with-lease origin main"}}' \
  | rtk hook claude
```

(Use `printf '%s'` so `\n` reaches rtk as a literal `\n` in the JSON string, which is decoded to a real newline.)

## Why pipes behave differently (confirms the root cause)

`|` *is* a real split point, so it exposes the post-pipe command as its own segment that must independently match an allow rule:

| Command | Verdict | Why |
|---|---|---|
| `git log -1\nrm -rf /` | **allow** | one segment, prefix-matches `git log` |
| `git log -1\nrm -rf / \| tail` | **allow** | segments `git log…rm -rf /` + `tail`; both match (`tail` is allow-listed) |
| `git log -1\nrm -rf / \| bash` | none | `bash` segment matches no allow rule |
| `git log -1\nbash evil.sh` | **allow** | no pipe → one segment, prefix-matches `git log` |

So the verdict is a correct per-segment allowlist for the operators the lexer knows about; the defect is purely that **newline is not one of them**.

## Root cause (code)

- `src/discover/lexer.rs` — `tokenize()` emits a separator token only for `|`, `||`, `;`, `&&`. There is no case for `\n`/`\r`; newlines are consumed as ordinary whitespace inside a segment. `split_on_operators()` (and thus `split_compound_command()` in `src/hooks/permissions.rs`) therefore never cuts on a newline.
- `src/hooks/permissions.rs` — `command_matches_pattern()` matches a non-wildcard / trailing-`*` rule by **prefix**: `cmd == pattern || cmd.starts_with("{pattern} ")`. A segment that begins `git log -1\nrm -rf /` satisfies `starts_with("git log ")` and so matches `Bash(git log:*)`.

Result: a multi-line block is one prefix-matched segment, validated solely by its first command; everything after the first newline is approved unchecked.

This is the same newline-blindness reported in #2191 (there framed as a token-savings miss). The per-segment allowlist hardening (#260 / #1213) is defeated by it, because the dangerous statement is never isolated as its own segment.

## Impact

Any allow/deny guardrail can be bypassed by prefixing a dangerous command with a newline after a benign, allow-listed one — and `git log`, `git diff`, `git status`, `ls`, `cat`, `grep`, `find` are exactly the commands agents run at the start of almost every turn.

## Affected versions

Confirmed in released **0.42.1** and on current **main** (`dev-0.43.0-rc.260`): `src/discover/lexer.rs` and `src/hooks/permissions.rs` are byte-identical between tag `v0.42.1` and `HEAD`.

## Suggested fix

Treat `\n` (and `\r`) as a command separator in `tokenize()` so each line becomes its own segment that must independently match an allow rule. With that change, `cat …\nrm -rf /` splits into two segments, `rm -rf /` fails the allowlist, and the block falls through to Claude Code's prompt.

bastewart:
## What

Treat a newline (`\n`/`\r`) as a command separator in the lexer's `tokenize`, so newline-separated statements are split into independent segments.

Fixes #2262.

## Why

`tokenize` (`src/discover/lexer.rs`) split compound commands only on `|`, `||`, `;` and `&&` — never newlines. Combined with the permission checker's prefix-based allow matching (`command_matches_pattern`), a newline-separated block collapsed into a single segment matched solely by its first line. A block led by an allow-listed read command (`git log`, `cat`, `ls`, …) therefore had its entire contents auto-allowed by the `rtk hook claude` PreToolUse hook, so trailing statements such as `rm -rf --no-preserve-root /`, `git push --force-with-lease` or `git commit --amend` ran with no prompt — bypassing the user's allow/deny rules. Full analysis in #2262.

This is the same newline-blindness as #2191 (there framed as a token-savings miss); the per-segment allow hardening from #1213 is defeated by it, because the dangerous statement is never isolated as its own segment.

## Change

Emit an `Operator` token for `\n`/`\r`, mirroring the existing `;` handling, ahead of the whitespace arm. Quoted newlines and `\`-line-continuations are already handled upstream, so only real unquoted newlines split.

Side effect: the rewrite path now rewrites every line of a multi-line command instead of only the first (previously only the leading line was rewritten — see #2191), inserting spaces around the newline. Harmless for execution; no existing tests regressed.

## Verification

- `cargo fmt --all --check`, `cargo clippy --all-targets`: clean
- `cargo test --all`: **2007 passed, 0 failed** (2 new tests)
- Behavioural check against the freshly built `rtk hook claude`:
  - `cat …\nrm -rf …` → no `permissionDecision` (previously `allow`)
  - `git log\ngit push --force-with-lease` → no `permissionDecision` (previously `allow`)
  - `git log | tail`, bare `git log` → still `allow` (legitimate cases preserved)

## Tests added

- `discover::lexer::tests::test_split_on_operators_newline` — newline splitting, including quoted newlines that must **not** split
- `hooks::permissions::tests::test_compound_allow_newline_separator` — a safe first line must not launder an unallowed trailing line

### DIFF
--- a/src/discover/lexer.rs
+++ b/src/discover/lexer.rs
@@ -226,6 +226,21 @@ pub fn tokenize(input: &str) -> Vec<ParsedToken> {
                 });
                 current_start = byte_pos;
             }
+            '\n' | '\r' => {
+                // A newline terminates a command, exactly like `;`. Without this,
+                // newlines fall through to the whitespace arm below and are treated
+                // as mere argument separators, so a multi-line command collapses
+                // into a single segment — letting allow-rule prefix matching on the
+                // first line approve every statement after it.
+                flush_arg(&mut tokens, &mut current, current_start);
+                tokens.push(ParsedToken {
+                    kind: TokenKind::Operator,
+                    value: c.to_string(),
+                    offset: byte_pos,
+                });
+                byte_pos += char_len;
+                current_start = byte_pos;
+            }
             c if c.is_whitespace() => {
                 flush_arg(&mut tokens, &mut current, current_start);
                 byte_pos += c.len_utf8();
@@ -1014,6 +1029,21 @@ mod tests {
         assert_eq!(
             split_on_operators("a && b | c ; d", false),
             vec!["a", "b", "c", "d"]
+        );
+    }
+
+    #[test]
+    fn test_split_on_operators_newline() {
+        // Newlines separate commands, exactly like `;`. A multi-line command
+        // must not collapse into a single segment, or allow-rule prefix matching
+        // on the first line would approve every later line.
+        assert_eq!(split_on_operators("a\nb\nc", false), vec!["a", "b", "c"]);
+        // Mixed with other operators.
+        assert_eq!(split_on_operators("a\nb | c", false), vec!["a", "b", "c"]);
+        // A newline inside quotes is NOT a separator.
+        assert_eq!(
+            split_on_operators("echo \"a\nb\"\ncargo test", false),
+            vec!["echo \"a\nb\"", "cargo test"]
         );
     }
 
--- a/src/hooks/permissions.rs
+++ b/src/hooks/permissions.rs
@@ -695,6 +695,33 @@ mod tests {
     }
 
     #[test]
+    fn test_compound_allow_newline_separator() {
+        // A newline must split segments like `&&`/`;`. Regression for the
+        // permission bypass where a multi-line command was treated as a single
+        // segment and prefix-matched on its first line, auto-allowing every
+        // statement after the first newline.
+        let allow = vec!["git log".to_string(), "cat *".to_string()];
+
+        // Allowed first line, unallowed trailing line → must NOT escalate to Allow.
+        assert_eq!(
+            check_command_with_rules("git log -1\ngit push --force-with-lease", &[], &[], &allow),
+            PermissionVerdict::Default,
+            "first-line allow must not approve a newline-separated trailing command"
+        );
+        assert_eq!(
+            check_command_with_rules("cat f\nrm -rf /", &[], &[], &allow),
+            PermissionVerdict::Default,
+            "a safe first line must not launder a destructive trailing line"
+        );
+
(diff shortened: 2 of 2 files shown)


KuSh:
This should have already been fixed by 952245d

PR #2265, opened 2026-06-04
fix(find): support -print flag as no-op

youbamj:
## Problem

`rtk find . -name "*.rs" -print` warns:
```
rtk find: unknown flag '-print', ignored
```

The `-print` flag is the default action in standard `find(1)`. RTK already prints matches, so this should be a no-op.

## Root Cause

`has_native_find_flags()` didn't recognize `-print`, causing it to fall through to the "unknown flag" warning handler.

## Changes

- Add `-print` to `has_native_find_flags()` so RTK enters native parsing mode
- Add `-print` case in `parse_native_find_args()` as a no-op (RTK already prints matches)
- Add 2 unit tests for `-print` behavior

## Verification

```bash
cargo fmt --all --check && cargo clippy --all-targets && cargo test
# 2007 passed, 0 failed
```

## Notes

- Output format stays as-is (grouped by directory for token savings — intentional RTK behavior)
- Does NOT add `-print0` (which is in `UNSUPPORTED_FIND_FLAGS` — different semantics)

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -59,7 +59,7 @@ fn next_arg(args: &[String], i: &mut usize) -> Option<String> {
 /// Check if args contain native find flags (-name, -type, -maxdepth, etc.)
 fn has_native_find_flags(args: &[String]) -> bool {
     args.iter()
-        .any(|a| a == "-name" || a == "-type" || a == "-maxdepth" || a == "-iname")
+        .any(|a| a == "-name" || a == "-type" || a == "-maxdepth" || a == "-iname" || a == "-print")
 }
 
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the report and the fix. This is covered on develop by #3603 (merged 2026-08-26), which reworked rtk find to parse find's real grammar; -print is now recognised as a native action and the command runs through to real find, so the 'unknown flag -print, ignored' warning is gone. Running rtk find . -maxdepth 1 -name '*.md' -print against develop produces plain results with no warning, and find_cmd.rs was rewritten enough that this branch no longer applies. Closing as covered, but please comment or reopen if you still see the warning on a case we missed.

PR #1308, opened 2026-04-14
feat: add Factory Droid agent integration (rtk init -g --agent droid)

ZaynJarvis:
## Summary

Add support for [Factory Droid](https://docs.factory.ai/) (factory.ai) as a supported agent in RTK.

Factory Droid uses the **same PreToolUse hook architecture** as Claude Code. The only differences are:

| | Claude Code | Factory Droid |
|---|---|---|
| Tool name | `Bash` | `Execute` |
| Settings path | `~/.claude/settings.json` | `~/.factory/settings.json` |
| Hook directory | `~/.claude/hooks/` | `~/.factory/hooks/` |
| Awareness doc | `@RTK.md` in `CLAUDE.md` | `~/.factory/RTK.md` |

The JSON input/output format (`hookSpecificOutput`, `permissionDecision`, `updatedInput`) is **identical** to Claude Code, so the hook script is a thin adaptation with the same delegating pattern.

### New files
- `hooks/droid/rtk-rewrite.sh` — Shell hook that intercepts `Execute` tool commands and rewrites them via `rtk rewrite`
- `hooks/droid/rtk-awareness.md` — Slim awareness doc for Droid context
- `hooks/droid/README.md` — Integration documentation with installation, testing, and uninstall instructions

### Source changes
- `src/main.rs` — Add `Droid` variant to `AgentTarget` enum, wire up `install_droid` and `droid_uninstall` flags
- `src/hooks/init.rs` — Add:
  - `install_droid_hooks()` — writes hook script, RTK.md, patches settings.json
  - `remove_droid_hooks()` — cleans up all Droid artifacts
  - `patch_droid_settings_json()` — adds PreToolUse entry for `Execute` matcher
  - `droid_hook_already_present()` / `remove_droid_hook_from_json()` — idempotency helpers
  - Droid status checks in `show_config()`
  - Validation: `rtk init -g --agent droid` is global-only

### Documentation
- `hooks/README.md` — Add Factory Droid row in Supported Agents table, JSON format section, and directory listing

## Usage

```bash
# Install
rtk init -g --agent droid

# Verify
rtk init --show

# Uninstall
rtk init -g --agent droid --uninstall
```

## Testing

```bash
# Manual test
echo '{"tool_name":"Execute","tool_input":{"command":"git status"}}' | ~/.factory/hooks/rtk-rewrite.sh
[...]

PR #912, opened 2026-03-29
feat(droid): add Factory AI Droid support

skrabe:
## Summary

- Adds RTK integration for [Factory AI Droid](https://factory.ai) CLI
- Factory Droid uses a `PreToolUse` hook system with the same `hookSpecificOutput` JSON format as Claude Code, with the shell tool named `Execute` instead of `Bash`
- Follows the Claude Code integration pattern: hook script + `RTK.md` + `@RTK.md` in `AGENTS.md` + `settings.json` patching

## Changes

**New files:**
- `hooks/droid/rtk-rewrite.sh` — Shell hook script (delegates to `rtk rewrite`, same JSON protocol as Claude Code)
- `hooks/droid/rtk-awareness.md` — Slim RTK awareness instructions for hook-based usage
- `hooks/droid/README.md` — Integration docs

**Modified files:**
- `src/main.rs` — Added `--droid` flag to `Init` command, dispatch to `run_droid()`, wired into `uninstall()`
- `src/hooks/init.rs` — Added `run_droid()`, `patch_droid_settings()`, `uninstall_droid()`, `resolve_droid_dir()`
- `README.md` — Added Factory AI Droid to supported tools table and Quick Start

## Install flow (`rtk init -g --droid`)

1. Writes hook script to `~/.factory/hooks/rtk-rewrite.sh` (chmod 755)
2. Writes `RTK.md` to `~/.factory/RTK.md`
3. Adds `@RTK.md` reference to `~/.factory/AGENTS.md` (creates if missing, preserves existing content)
4. Patches `~/.factory/settings.json` with `PreToolUse` hook entry (`Execute` matcher)

## Uninstall flow (`rtk init -g --droid --uninstall`)

1. Removes `~/.factory/hooks/rtk-rewrite.sh`
2. Removes `~/.factory/RTK.md`
3. Removes `@RTK.md` line from `~/.factory/AGENTS.md` (preserves other content)
4. Removes RTK hook entry from `~/.factory/settings.json`

## Test plan

- [x] `cargo fmt --all --check` passes
- [x] `cargo clippy --all-targets` — no new warnings
- [x] `cargo test` — 1138 passed, 0 failed
- [x] Manual: `rtk init -g --droid --auto-patch` installs all artifacts correctly
- [x] Manual: idempotent — second install doesn't duplicate `@RTK.md` or settings entry
- [x] Manual: `rtk init -g --droid --uninstall` removes all RTK artifacts
[...]

PR #2267, opened 2026-06-04
feat(hooks): add Factory Droid integration

### LINKED ISSUE #2099: Feature Request: Support for Factory Droid CLI
Currently, rtk provides excellent token-optimized output filtering for various AI coding assistants like Claude Code and GitHub Copilot. However, it lacks native support for Factory Droid, another powerful AI assistant focused on software engineering tasks. Without this integration, users of Factory Droid cannot benefit from the significant token savings and context optimization that rtk offers.
I would love to see rtk add support for Factory Droid, similar to its existing integrations. Specifically:
Hook/Proxy Integration: Add specific hooks or proxy logic to handle Factory Droid's CLI commands.
Output Filtering: Apply rtk's intelligent filtering to the outputs generated by Factory Droid before they reach the LLM context.
Command Rewriting: If applicable, enable command rewriting so that standard shell commands invoked within a Droid session are automatically routed through rtk for optimization.
Thank you for building such an amazing tool! I believe adding Factory Droid support would be a great addition to the ecosystem.

krimvp:
Closes #2099

## Summary

Adds first-class **Factory Droid** support so shell commands the agent runs are transparently rewritten to their token-saving `rtk` equivalents.

- **`rtk init --agent droid`** — installs a native `PreToolUse` hook (matcher `Execute`) into Droid's canonical `hooks.json` (`~/.factory/hooks.json`, or project `.factory/hooks.json`), falling back to the `hooks` key of `settings.json` only when that file already carries live `PreToolUse` hooks. Supports global (`-g`) and project scope, the `FACTORY_HOME_OVERRIDE` env override, idempotent install with `.bak` backup + atomic write, and a clean uninstall round-trip (`--uninstall`).
- **`rtk hook droid`** — processes Droid's `PreToolUse` payload (BOM strip, stdin cap, JSON-parse fallback), mirroring the existing provider hooks.
- **Droid's own permissions are enforced** (review feedback): verdicts come from `permissions::Host::Droid`, sourced from `~/.factory/settings.json` — never from another agent's settings.
  - `commandAllowlist` → rewrite carries `permissionDecision: "allow"` (the verdict Droid would emit natively). Droid's built-in default allowlist (`ls`, `pwd`, `git status`, `git log`, …) is mirrored, and a user list replaces its default outright — matching Droid's own `settings.commandAllowlist || DEFAULT` accessor.
  - `commandDenylist` (always confirm) and `commandBlocklist` (never runs) → RTK **steps aside entirely** so Droid's native confirm/block fires on the original command. Rewriting first would dodge Droid's own pattern matching (`rtk git log` no longer matches a `git log` denylist entry).
  - Anything unlisted → rewrite with **no** decision, landing via Droid's "updated input result" path while Droid's native prompt and other hooks' deny/ask stay in control.

Protocol verified against Factory's [hooks reference](https://docs.factory.ai/reference/hooks-reference) and the shipped Droid CLI (v0.153.1): `Execute` matcher, `tool_input.command`, `hookSpecificOutput` with `permissionDecision`/`updatedInput`, per-event-key merge of `hooks.json` over `settings.json` hooks, `FACTORY_HOME_OVERRIDE` as a home-dir override, and the `commandAllowlist`/`commandDenylist`/`commandBlocklist` defaults in [settings](https://docs.factory.ai/cli/configuration/settings).

## Demo

End-to-end: **install → Droid rewrites & auto-runs allowlisted commands at `--auto low` → a denylisted command is left to Droid's native permission gate → savings in `rtk gain` → uninstall**.

RTK x Factory Droid demo

> If the GIF doesn't load inline: https://files.catbox.moe/cxbk1y.gif (mirror: https://gist.github.com/krimvp/cfc67a932bf7e788cdfebcea4b6f596b)

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo test --all` — 2394 passed (34 Droid-specific: install/uninstall round-trip, idempotency, matcher-group reuse, hooks.json/settings.json target resolution, allowlist auto-allow, denylist/blocklist step-aside, user-list-replaces-default, deny step-aside, legacy `Bash` matcher, passthrough)
- [x] Verified live against the real `droid` CLI (v0.153.1), all with `--auto low`:
  - `git status` / `git log` rewritten to `rtk …` and auto-run via Droid's default allowlist (session transcript shows RTK's filtered output; savings tracked in `rtk gain`)
  - with `"commandDenylist": ["git log"]`: RTK emits nothing, Droid natively refuses (`insufficient permission to proceed`) — A/B control with the entry removed runs rewritten again
  - Claude Code settings no longer leak: a `Bash(curl:*)` allow rule in `~/.claude/settings.local.json` no longer auto-allows `curl` in Droid
- [x] `settings.json` untouched by install/uninstall round-trip

## Docs

README updated — Supported Agents list + integration table row. `docs/guide/getting-started/supported-agents.md` updated per review feedback (frontmatter, agents table, installation section).

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ rtk init -g
 
 ## Supported AI Tools
 
-RTK supports 14 AI coding tools. Each integration rewrites shell commands to `rtk` equivalents for 60-90% token savings where the agent supports command interception.
+RTK supports 15 AI coding tools. Each integration rewrites shell commands to `rtk` equivalents for 60-90% token savings where the agent supports command interception.
 
 | Tool | Install | Method |
 |------|---------|--------|
@@ -385,7 +385,7 @@ RTK supports 14 AI coding tools. Each integration rewrites shell commands to `rt
 | **Mistral Vibe** | Planned ([#800](https://github.com/rtk-ai/rtk/issues/800)) | Blocked on upstream |
 | **Kilo Code** | `rtk init --agent kilocode` | .kilocode/rules/rtk-rules.md (project-scoped) |
 | **Google Antigravity** | `rtk init --agent antigravity` | .agents/rules/antigravity-rtk-rules.md (project-scoped) |
-| **Factory Droid** | `rtk init -g --agent droid` (or per-project) | PreToolUse hook in `~/.factory/settings.json` (matcher `Execute`) |
+| **Factory Droid** | `rtk init -g --agent droid` (or per-project) | PreToolUse hook in `~/.factory/hooks.json` (matcher `Execute`) |
 
 For per-agent setup details, override controls, and graceful degradation, see the [Supported Agents guide](https://www.rtk-ai.app/guide/getting-started/supported-agents). The Hermes plugin source and tests live in `hooks/hermes/`; installed Hermes runtime files still live under `~/.hermes/plugins/rtk-rewrite/`.
 
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ rtk init --agent kilocode       # Kilo Code
 rtk init --agent antigravity    # Google Antigravity
 rtk init -g --agent pi          # Pi
 rtk init --agent hermes         # Hermes
+rtk init -g --agent droid       # Factory Droid
 
 # 2. Restart your AI tool, then test
 git status  # Automatically rewritten to rtk git status
@@ -384,6 +385,7 @@ RTK supports 14 AI coding tools. Each integration rewrites shell commands to `rt
 | **Mistral Vibe** | Planned ([#800](https://github.com/rtk-ai/rtk/issues/800)) | Blocked on upstream |
 | **Kilo Code** | `rtk init --agent kilocode` | .kilocode/rules/rtk-rules.md (project-scoped) |
 | **Google Antigravity** | `rtk init --agent antigravity` | .agents/rules/antigravity-rtk-rules.md (project-scoped) |
+| **Factory Droid** | `rtk init -g --agent droid` (or per-project) | PreToolUse hook in `~/.factory/settings.json` (matcher `Execute`) |
 
 For per-agent setup details, override controls, and graceful degradation, see the [Supported Agents guide](https://www.rtk-ai.app/guide/getting-started/supported-agents). The Hermes plugin source and tests live in `hooks/hermes/`; installed Hermes runtime files still live under `~/.hermes/plugins/rtk-rewrite/`.
 
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -604,22 +604,32 @@ fn droid_response_for_verdict(v: &Value, cmd: &str, verdict: PermissionVerdict) 
         ti
     };
 
-    // Droid (like Cursor) only applies a hook's `updatedInput` when the
-    // permission decision is "allow" — an "ask"/omitted decision makes Droid
-    // run the *original* command, so the rewrite (and its token savings) would
-    // be silently dropped. We already stepped aside on Deny above, so anything
-    // reaching here is safe to auto-allow in its rewritten `rtk …` form (it is
-    // the same command the agent chose, just wrapped). Mirrors `run_cursor`.
-    let decision = "allow";
+    // Wire format mirrors Claude's `hookSpecificOutput`; the allow policy mirrors
+    // `process_claude_payload`, NOT Cursor. Verified empirically against Droid
+    // 0.140.0: Droid applies a hook's `updatedInput` regardless of the permission
+    // decision — beyond the `case "allow"` branch it has an "updated input result"
+    // path that applies any hook's `updatedInput` even when no decision is set. So
+    // forcing "allow" is unnecessary for the rewrite to land, and would be harmful:
+    // Droid resolves the decision as `find(first result with a permissionDecision)`
+    // (no deny-over-allow priority), so an unconditional "allow" can suppress
+    // another PreToolUse hook's deny/ask and bypasses Droid's native permission
+    // prompt. We therefore only assert "allow" on an explicit allow rule; otherwise
+    // we omit the decision so the rewrite still applies while the user's other
+    // hooks and Droid's native prompt stay in control.
+    let mut hook_output = json!({
+        "hookEventName": PRE_TOOL_USE_KEY,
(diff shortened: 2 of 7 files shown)


aeppling:
Hey @krimvp 

Thanks for this contribution, this PR supersed #1308 and #912 , looking forward this one now because more complete and up to date

src/hooks/init.rs
```diff
@@ -2762,6 +2763,335 @@ fn resolve_hermes_home_from_env(
...
+        return Ok(PathBuf::from(path));
+    }
+    home_dir
+        .map(|home| home.join(DROID_DIR))
+        .context("Cannot determine Droid config directory. Set $FACTORY_HOME or $HOME.")
+}
+
+/// Install Factory Droid PreToolUse hook into `settings.json`.
+///
+/// - Global (`-g`): writes to `~/.factory/settings.json` (or `$FACTORY_HOME/settings.json`).
```
aeppling: Following the Droid CLI documentation:

```
Droid hooks live in hooks.json
- ~/.factory/hooks.json – User (applies to all projects)
- .factory/hooks.json – Project (commit to share)
- enterprise managed policy hooks

If hooks.json is absent, Droid falls back to hooks declared under the hooks key in the matching settings.json.
```

So it seems that settings.json is a fallback.
We should try to use this file and fallback to settings if not available to better match Droid CLI behavior ?


source : https://docs.factory.ai/reference/hooks-reference

src/hooks/hook_cmd.rs
```diff
@@ -546,6 +546,111 @@ fn run_cursor_inner_with_rules(
...
+
+    let updated_input = {
+        let mut ti = v.get("tool_input").cloned().unwrap_or_else(|| json!({}));
+        if let Some(obj) = ti.as_object_mut() {
+            obj.insert("command".into(), Value::String(rewritten));
+        }
+        ti
+    };
+
+    // Wire format mirrors Claude's `hookSpecificOutput`; the allow policy mirrors
```
aeppling: Too verbose comment, if need to refer to a agent documentation or behavior to specific version, you can link a documentation or changelog

Just to keep it clean :)

src/hooks/hook_cmd.rs
```diff
@@ -546,6 +546,111 @@ fn run_cursor_inner_with_rules(
...
+    // prompt. We therefore only assert "allow" on an explicit allow rule; otherwise
+    // we omit the decision so the rewrite still applies while the user's other
+    // hooks and Droid's native prompt stay in control.
+    let mut hook_output = json!({
+        "hookEventName": PRE_TOOL_USE_KEY,
+        "permissionDecisionReason": "RTK auto-rewrite",
+        "updatedInput": updated_input
+    });
+
+    if verdict == PermissionVerdict::Allow {
```
aeppling: This leak claude settings because using the check_command that check for claude code settings.

Droid has its own permissions through commandAllowlist / commandDenylist / commandBlocklist and autonomyLevel in ~/.factory/settings.json

RTK should respect the agents permissions system and always match the verdict (allow/deny or ask) that the agent would have natively emitted

source : https://docs.factory.ai/cli/configuration/settings

README.md
```diff
@@ -113,6 +113,7 @@ rtk init --agent kilocode       # Kilo Code
 rtk init --agent antigravity    # Google Antigravity
 rtk init -g --agent pi          # Pi
 rtk init --agent hermes         # Hermes
+rtk init -g --agent droid       # Factory Droid
```
aeppling, line 116: Correct documentation, should also be propagate to https://github.com/rtk-ai/rtk/blob/develop/docs/guide/getting-started/supported-agents.md

aeppling (changes requested):

src/hooks/permissions.rs
```diff
@@ -271,6 +276,103 @@ fn load_gemini_rules() -> (Vec<String>, Vec<String>, Vec<String>) {
...
+    "git log",
+    "git show",
+    "git blame",
+    "git ls-files",
+];
+const DROID_DEFAULT_DENYLIST: &[&str] = &[
+    "rm -rf /",
+    "rm -rf /*",
+    "rm -rf .",
+    "rm -rf ~",
```
aeppling: Those kind of list are often a bad idea for maintenance and transparence.

Drop DROID_DEFAULT_ALLOWLIST / DROID_DEFAULT_DENYLIST and fall back to empty lists. Hardcoding defaults will drift against Factory's releases and makes RTK auto-allow commands the user never listed.

To keep : (1) read all scopes global + settings.local.json + project .factory/ (Droid merges them; see load_gemini_rules' folderTrust handling), not global-only, or project/local deny entries get dodged; (2) honor only explicit commandDenylist/commandBlocklist → step-aside (the fix for the rtk-rename dodge).

For the allow side, prefer defer (no permissionDecision) rather than emitting allow, since RTK renaming the program to rtk means it can't transparently reproduce Droid's own allow decision.

If any question do no hesitate to tag me so we can cleanly implement this integration

aeppling (changes requested):

src/hooks/permissions.rs
krimvp: Implemented in 3d40742 — thanks for the detailed guidance @aeppling.

- Dropped `DROID_DEFAULT_ALLOWLIST`/`DROID_DEFAULT_DENYLIST` entirely; no built-in lists are mirrored anymore.
- Deny-only: RTK reads explicit `commandDenylist`/`commandBlocklist` from all four scopes (`~/.factory/settings.json` + `settings.local.json`, project `.factory/settings.json` + `settings.local.json`) and steps aside on a match. One deliberate divergence: entries are **unioned** across scopes rather than replicating Droid's replace-on-merge precedence — over-collecting only causes a spurious step-aside (Droid still decides natively on the original command), while under-collecting would reopen the rename dodge.
- Allow side: RTK never emits `permissionDecision` — rewrites land via `updatedInput` alone and the verdict stays with Droid's native flow (allowlist, autonomy level, other hooks).

Verified end-to-end against Droid CLI 0.153.1: Droid evaluates its permission lists on the **original** command before applying `updatedInput`, so native allowlist decisions are unaffected (`git status` still auto-runs at `--auto low`) and denylisted commands are gated natively with no rewrite involved.

demo

aeppling:
Hey @krimvp , LGTM , thanks for contributing to RTK with this new integration !

aeppling (approved):

PR #2123, opened 2026-05-27
fix(gh,glab): don't pre-reject view/checks/run subcommands missing the id

B-Galati:
## Problem

`gh` and `glab` accept several subcommands without an explicit id and resolve it from the current branch (or open an interactive picker). RTK was pre-rejecting these with `"X number required"` before the underlying tool ever ran.

Example:

```console
$ rtk glab mr view
rtk: MR number required
```

…even though `glab mr view` (no args) is valid and looks up the MR for the current branch.

## Fix

Six call sites in `gh_cmd.rs` / `glab_cmd.rs` now treat the identifier as optional and forward the request unchanged, letting `gh`/`glab` decide:

- `gh pr view`, `gh pr checks`, `gh issue view`, `gh run view`
- `glab mr view`, `glab issue view`

A new private `parse_optional_identifier` helper in each file captures the `(Option<id>, Vec<extra>)` parse so the no-id branch is unit-testable.

`format_run_view` also drops the dangling `#` from its header when the run id is empty (interactive-picker passthrough).

## Tests

8 new unit tests (1954 total, all passing):

- `parse_optional_identifier` — empty / flags-only / with-id paths × 2 files
- `format_run_view` — with and without an id

```
cargo fmt --all && cargo clippy --all-targets && cargo test --all
# clippy: clean, tests: 1954 passed, 6 ignored
```

End-to-end verified on `glab mr view` (now defers to glab; in this repo glab returns its own "no GitLab remote configured" message instead of RTK's pre-rejection).

PR #2273, opened 2026-06-05
fix(gh): passthrough run view job output

### LINKED ISSUE #1568: `gh run view --job <id>` rejected with "Run ID required"

bte808:
## Summary

- pass `gh run view --job <id>` through to `gh` instead of the workflow-run summary filter
- keep existing passthrough behavior for `--log`, `--log-failed`, and `--json`
- add regression coverage for the job-only argument shape from #1568

## Why

`gh run view --job <id>` does not have a positional run id. It produces job-level output that should not be reshaped by the run summary formatter, and recent reports on #1568 show the job-only path still needs explicit coverage.

Fixes #1568.

## Checks

- `git diff --check`
- `cargo test run_view`
- `cargo test test_parse_optional_identifier_with_job_only`
- `cargo test --all` (2007 passed, 0 failed, 7 ignored)

### DIFF
--- a/src/cmds/git/gh_cmd.rs
+++ b/src/cmds/git/gh_cmd.rs
@@ -770,12 +770,12 @@ fn format_run_list(json: &Value, ultra_compact: bool) -> String {
 }
 
 /// Check if run view args should bypass filtering and pass through directly.
-/// Flags like --log-failed, --log, and --json produce output that the filter
-/// would incorrectly strip.
+/// Flags like --job, --log-failed, --log, and --json produce output that the
+/// workflow-run summary filter would incorrectly strip or reshape.
 fn should_passthrough_run_view(extra_args: &[String]) -> bool {
     extra_args
         .iter()
-        .any(|a| a == "--log-failed" || a == "--log" || a == "--json")
+        .any(|a| a == "--job" || a == "--log-failed" || a == "--log" || a == "--json")
 }
 
 fn view_run(args: &[String], _verbose: u8) -> Result<i32> {
@@ -1172,6 +1172,14 @@ mod tests {
     }
 
     #[test]
+    fn test_run_view_passthrough_job() {
+        assert!(should_passthrough_run_view(&[
+            "--job".into(),
+            "67890".into()
+        ]));
+    }
+
+    #[test]
     fn test_run_view_no_passthrough_empty() {
         assert!(!should_passthrough_run_view(&[]));
     }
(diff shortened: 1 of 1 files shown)


ferr079:
Independent confirmation that this is the right fix, and why it's still needed after #2123.

The `--job` failure mode for #1568 actually has **two layers**, which is what made it look "already resolved" when #1570 was closed:

1. **Hard crash — `Run ID required`.** Fixed by #2123 (`e1bb17f`) making the run-id optional via `parse_optional_identifier`. ✅
2. **`--job` output still mangled.** On current `develop`, `should_passthrough_run_view` only lists `--log-failed`/`--log`/`--json` ([gh_cmd.rs#L775-L779](https://github.com/rtk-ai/rtk/blob/develop/src/cmds/git/gh_cmd.rs#L775-L779)), so a `--job <id>` invocation no longer errors but falls into the `run_filtered` branch and gets reshaped by `format_run_view` (which only looks for `JOBS`/`Status:`/`Conclusion:`). Job logs come back empty/garbled.

This PR closes layer 2 by adding `--job` to `should_passthrough_run_view`, which is exactly right — once `--job` is present the user wants raw `gh` output (job log / step view), not the workflow-run summary. So #1568 is only fully resolved with both #2123 **and** this change.

Repro on the released binary (`rtk 0.42.1`, brew), which predates #2123 so it shows layer 1:

```
$ rtk gh run view --job 79728857782 --log-failed
rtk: Run ID required        # gh accepts `-j/--job` with no positional run-id
```

I hit this for real while inspecting a failed CI job (`gh run view --job <id> --log-failed`) and had to fall back to `rtk proxy`. The diff + the two regression tests (`test_run_view_passthrough_job`, `test_parse_optional_identifier_with_job_only`) match the actual call shapes. LGTM.

PR #2274, opened 2026-06-05
fix(registry): only rewrite last pipe segment; prevent data corruption in pipes/redirects

youbamj:
## Summary
Fixes critical data corruption bugs where RTK incorrectly filters piped or redirected commands, breaking downstream tools and producing malformed output files.

### Bug 1: Large JSON truncated, breaks `jq`

```bash
$ rtk curl https://jsonplaceholder.typicode.com/posts | jq
jq: parse error: Unfinished string at EOF at line 14, column 0
```

**Root cause:** RTK filters `curl` output, truncating large JSON responses. The truncated data is piped to `jq`, which fails to parse incomplete JSON.

### Bug 2: Git diff corrupted when redirected to file

**Before (RTK-filtered patch - BROKEN):**
```diff
README.md | 2 ++
 1 file changed, 2 insertions(+)

--- Changes ---

README.md
  @@ -2,6 +2,8 @@
  +Change
  +
   <p align="center">
     <strong>High-performance CLI proxy that reduces LLM token consumption by 60-90%</strong>
   </p>
  +2 -0
```

**After this fix (real patch - CORRECT):**
```diff
diff --git a/README.md b/README.md
index d91e21b..dabe2cd 100644
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
   <img src="https://avatars.githubusercontent.com/u/258253854?v=4" alt="RTK - Rust Token Killer" width="500">
 </p>
 
+Change
+
 <p align="center">
   <strong>High-performance CLI proxy that reduces LLM token consumption by 60-90%</strong>
 </p>
```

**Result:** `git apply /tmp/rtk-test.patch` fails with "No valid patches in input" on the RTK version, but succeeds on the real version.

## Algorithm
1. Find **last** `|` in entire command
2. Everything **before** last `|` → stays raw (intermediate pipe segments)
3. Everything **after** last `|` → normal rewrite rules:
   - Check for file redirect (`>`, `>>`, `&>`) → skip entirely
   - Check if supported → prepend `rtk ` if yes
4. Input redirect `<` is fine (reading from file, output still to terminal)

## Examples

| Input | Output | Reason |
|-------|--------|--------|
| `git log \| cat && git stash` | `git log \| rtk cat && rtk git stash` | Only last segments get rtk |
| `cargo test 2>&1 \| head > file` | `None` | Redirected to file, skip |
| `cat < file.txt` | `rtk read - < file.txt` | Input redirect is OK |
| `find . \| xargs grep pattern` | `find . \| xargs rtk grep pattern` | xargs is transparent |
| `curl ... \| jq` | `None` | Pipe chain stays raw |

## Changes
- `src/discover/registry.rs` (+198, -62 lines)
  - Refactored `rewrite_compound()` with last-pipe detection
  - Extracted `rewrite_segments()` helper
  - Added `has_stdout_redirect()` (checks `>`, `>>`, `&>` only)
  - Fixed `<` input redirect to not block rewrite
  - Added `xargs` to `SHELL_PREFIX_BUILTINS`
  - Added 14 new/updated tests

## Testing
```
cargo test        # 2016 passed, 0 failed
cargo fmt --all   # clean
cargo clippy --all-targets  # clean
```

## Breaking Change
This changes the behavior of `rtk rewrite` for piped commands. Previously all commands in a pipe got `rtk` prefix; now only the last segment does. This is the correct Unix behavior but may surprise users relying on the old (broken) behavior.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -133,7 +133,7 @@ pub fn classify_command(cmd: &str) -> Classification {
         let has_redirect = cmd_clean
             .split_whitespace()
             .skip(1)
-            .any(|t| t.starts_with('>') || t == "<" || t.starts_with(">>"));
+            .any(|t| t.starts_with('>') || t.starts_with(">>"));
         if has_redirect {
             return Classification::Unsupported {
                 base_command: cmd_clean
@@ -467,6 +467,18 @@ fn collapse_line_continuations(s: &str) -> std::borrow::Cow<'_, str> {
 /// Also strips user-configured transparent wrapper prefixes
 /// (`[hooks].transparent_prefixes` in `config.toml`) before routing.
 ///
+/// Returns true when the command redirects stdout to a file.
+///
+/// We skip rewriting in this case because the file consumer expects raw
+/// output, not RTK's filtered format.
+fn has_stdout_redirect(cmd: &str) -> bool {
+    let tokens = tokenize(cmd);
+    tokens.iter().any(|t| {
+        matches!(t.kind, TokenKind::Redirect)
+            && (t.value == ">" || t.value == ">>" || t.value.starts_with("&>"))
+    })
+}
+
 /// A transparent prefix is a wrapper command that doesn't change *what* is
 /// being run, only *how* it's run — e.g. `docker exec mycontainer`,
 /// `direnv exec .`, `poetry run`, or `bundle exec`. Stripping it lets the inner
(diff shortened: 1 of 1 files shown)


youbamj:
Superseded by #3128 (merged → 0.44.0): upstream shipped a native tokenizer-based pipeline rewrite (`rewrite_pipeline_final_stage`) that implements the same last-stage-only behavior. Closing as obsolete — the local patch is being dropped from the Nix derivation in favor of the upstream implementation.

PR #2276, opened 2026-06-05
fix(git): preserve patch output for log -p

### LINKED ISSUE #2275: "git log -p" omits the patch requested by the -p flag
When diagnosing a problem that requires inspecting git history, agents commonly use `git log -p`. However, rtk strips the patch information requested by the `-p` flag. This is often vital context.

Instead `rtk` compresses the diff. E.g. `[+5 lines omitted]`.

Please update `git log -p` handling to show the entire patch.

This was mentioned in https://github.com/rtk-ai/rtk/issues/467, but doesn't look like it was actually fixed in the associated PR.

puneetdixit200:
## Summary

Fixes #2275.

`git log -p` asks git for patch output, but the current log path still runs the result through RTK's log summarizer. That treats the diff as commit-body text and drops the patch lines.

This change detects explicit patch-style log flags and sends that command through unchanged, preserving the full diff when the caller asks for it.

## Validation

- `cargo fmt --all --check`
- `CARGO_TARGET_DIR=/private/tmp/rtkbuild-2275 cargo test test_log_patch_args_request_passthrough -j1`
- `CARGO_TARGET_DIR=/private/tmp/rtkbuild-2275 cargo test filter_log_output -j1`
- `CARGO_TARGET_DIR=/private/tmp/rtkbuild-2275 cargo run --quiet -- git log -p -1 -- src/cmds/git/git.rs`
- `CARGO_TARGET_DIR=/private/tmp/rtkbuild-2275 cargo test -j1`
- `CARGO_TARGET_DIR=/private/tmp/rtkbuild-2275 cargo clippy --all-targets -j1`

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -428,6 +428,34 @@ fn run_log(
     let mut cmd = git_cmd(global_args);
     cmd.arg("log");
 
+    if log_args_request_patch(args) {
+        for arg in args {
+            cmd.arg(arg);
+        }
+
+        let result = exec_capture(&mut cmd).context("Failed to run git log")?;
+
+        if !result.success() {
+            eprintln!("{}", result.stderr);
+            return Ok(result.exit_code);
+        }
+
+        if verbose > 0 {
+            eprintln!("Git log output:");
+        }
+
+        print!("{}", result.stdout);
+
+        timer.track(
+            &format!("git log {}", args.join(" ")),
+            &format!("rtk git log {} (passthrough)", args.join(" ")),
+            &result.stdout,
+            &result.stdout,
+        );
+
+        return Ok(0);
+    }
+
     // Check if user provided format flags
     let has_format_flag = args.iter().any(|arg| {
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for this, and sorry it sat so long. Develop now passes any git log that asks for a diff through untouched: PR #2951 (merged 2026-08-15) added the passthrough and PR #3681 (merged 2026-09-12) rebuilt it on the shared argument tokenizer, so log_wants_raw_shape in src/cmds/git/git.rs matches -p, -u, -U, -c, --patch, --patch-with-raw, --patch-with-stat, --cc, --unified and more. Every flag in this PR's unit test behaves that way on the current build, including -U5 and --unified=5. The two flags beyond that set do not apply: git log has no --combined, it answers `fatal: unrecognized argument: --combined` itself, and --full-diff without a diff flag prints no patch at all, so treating it as a patch request would only lose the summary. Since log_args_request_patch has no place in the tokenizer based version, I am closing this, but please comment or reopen if you find a flag that still loses its patch.

PR #2284, opened 2026-06-05
fix(go): preserve test and vet diagnostics

### LINKED ISSUE #2281: `rtk go` / Go filters can hide real compiler errors and report misleading success
## Summary

  `rtk` can swallow or misclassify real Go compile/vet errors. In our project this caused `rtk go ...` output to report
  misleading summaries such as:

  - `No tests found`
  - `No issues found`
  - `2 errors`

  These summaries did not reflect the actual Go compiler/vet failure. The only reliable way we found to see the real
  error was to bypass filtering with `rtk proxy`.

  ## Version

  ```sh
  rtk --version
  # rtk 0.42.0
 ```

  ## Reproduction / Diagnosis

  Running normal filtered Go commands through rtk hides the actionable compiler errors.

  The real errors are visible with:

  rtk proxy go vet ./internal/workorder/biz/ 2>&1 | grep -iE "undefined|cannot|cycle"

  Without rtk proxy, the filtered output can incorrectly summarize the run as successful or unrelated.

  One confusing example: go-m1cpu may print 2 errors, but those are clang/CGO warnings and the build exits with 0;
  meanwhile the actual Go errors such as undefined, cannot, or import cycle can be filtered away.

  ## Expected Behavior

  rtk go / Go-related filters should preserve real Go compile/vet errors, especially diagnostics containing patterns
  like:

  - undefined
  - cannot
  - import cycle
  - package/type/function resolution failures

  If filtering is uncertain, it should prefer showing the raw Go diagnostic rather than summarizing as No issues
  found / No tests found.

  ## Actual Behavior

  The filtered output can hide the real compiler/vet diagnostics and emit misleading summaries, causing agents/users to
  debug the wrong issue.

  ## Impact

  This is high impact for AI-agent workflows because the agent trusts the filtered terminal output. When compiler
  errors are hidden, the agent may conclude the build/test/vet pass is clean or chase unrelated CGO warnings.

  ## Workaround

  Use raw proxy mode:

  rtk proxy go vet ./internal/workorder/biz/ 2>&1 | grep -iE "undefined|cannot|cycle"

  But this defeats the purpose of using rtk go for safe summarized Go diagnostics.

jsiu93:
Closes #2281.

## Summary

- Route `rtk go test` and `rtk go vet` through exit-aware filtering so non-zero Go failures cannot be summarized as empty results.
- Preserve real Go diagnostics such as `undefined`, `cannot`, and `import cycle` from raw output and `go test -json` output events.
- Share Go diagnostic formatting across build/test/vet and strip ANSI before matching.

## Test plan

- [x] Manual repro: `rtk go test ./...` and `rtk go vet ./...` in a temporary Go project with `undefined: missingSymbol`


jsiu93:
bump. #2281 is still open. Note: #3577 covers the exit-code side for `go test` / `go vet` via the generic `guard_exit`; this PR additionally extracts compiler diagnostics from `go test -json` events. I'll rebase or narrow this to the diagnostics part once #3577 is decided.

PR #2287, opened 2026-06-05
fix(find): reject unsafe bare native actions

antonio-orionus:
## Summary
- fall back to native `find` for compound expressions/actions that RTK should not emulate
- reject bare native actions like `-delete`/`-exec` unless a narrowing predicate is present
- clarify `rtk find` and `rtk test` behavior in user docs

## Why
Bare native action forms such as `rtk find -delete` can default to the current directory when delegated to native `find`, creating a destructive footgun.

## Verification
- `rtk cargo fmt --all --check`
- `rtk cargo test find_cmd -- --nocapture`
- `rtk cargo clippy --all-targets`
- `rtk cargo test --all`
- `rtk git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ rtk grep "pattern" .            # Grouped search results
 rtk diff file1 file2            # Condensed diff
 ```
 
+`rtk find` compacts simple searches such as `-name`, `-iname`, `-type`, and `-maxdepth`. Compound native expressions/actions such as `(...)`, `-o`, `!`, `-exec`, `-delete`, or `-print0` automatically fall back to native `find`; use `rtk proxy find ...` when you want raw native behavior explicitly.
+
 ### Git
 ```bash
 rtk git status                  # Compact status
@@ -185,6 +187,8 @@ rtk rspec                       # RSpec tests (JSON, -60%+)
 rtk err <cmd>                   # Filter errors only from any command
 rtk test <cmd>                  # Generic test wrapper - failures only (-90%)
 ```
+
+`rtk test` expects a test-runner command, for example `rtk test cargo test`. It is not POSIX `test`; use `rtk run "test -d path"` or direct `test -d path` for shell condition checks.
 
 ### Build & Lint
 ```bash
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -1,10 +1,11 @@
 //! Filters find results by grouping files by directory.
 
-use crate::core::tracking;
+use crate::core::{tracking, utils};
 use anyhow::{Context, Result};
 use ignore::WalkBuilder;
 use std::collections::HashMap;
 use std::path::Path;
+use std::process::Stdio;
 
 /// Match a filename against a glob pattern (supports `*` and `?`).
 fn glob_match(pattern: &str, name: &str) -> bool {
@@ -62,17 +63,85 @@ fn has_native_find_flags(args: &[String]) -> bool {
         .any(|a| a == "-name" || a == "-type" || a == "-maxdepth" || a == "-iname")
 }
 
-/// Native find flags that RTK cannot handle correctly.
-/// These involve compound predicates, actions, or semantics we don't support.
-const UNSUPPORTED_FIND_FLAGS: &[&str] = &[
-    "-not", "!", "-or", "-o", "-and", "-a", "-exec", "-execdir", "-delete", "-print0", "-newer",
-    "-perm", "-size", "-mtime", "-mmin", "-atime", "-amin", "-ctime", "-cmin", "-empty", "-link",
-    "-regex", "-iregex",
+/// Native find tokens that RTK's compact parser should not try to emulate.
+/// These involve compound predicates, actions, output formats, or unsupported
+/// native semantics. Fall back to native `find` instead.
+const NATIVE_FIND_FALLBACK_TOKENS: &[&str] = &[
+    "(", ")", "\\(", "\\)", "-not", "!", "-or", "-o", "-and", "-a", "-exec", "-execdir",
+    "-ok", "-okdir", "-delete", "-print", "-print0", "-printf", "-fprint", "-fprint0",
+    "-fprintf", "-ls", "-fls", "-prune", "-quit", "-depth", "-mindepth", "-newer", "-perm",
+    "-size", "-mtime", "-mmin", "-atime", "-amin", "-ctime", "-cmin", "-empty", "-link",
+    "-links", "-regex", "-iregex", "-path", "-ipath", "-wholename", "-iwholename", "-user",
+    "-group", "-uid", "-gid", "-readable", "-writable", "-executable", "-xtype", "-samefile",
+    "-inum", "-true", "-false",
 ];
 
-fn has_unsupported_find_flags(args: &[String]) -> bool {
+const SUPPORTED_NATIVE_FIND_FLAGS: &[&str] = &["-name", "-iname", "-type", "-maxdepth"];
+const NATIVE_FIND_ACTION_TOKENS: &[&str] = &["-exec", "-execdir", "-ok", "-okdir", "-delete"];
+const NATIVE_FIND_NARROWING_PREDICATES: &[&str] = &[
+    "-name",
+    "-iname",
+    "-type",
+    "-newer",
+    "-perm",
+    "-size",
+    "-mtime",
+    "-mmin",
+    "-atime",
+    "-amin",
+    "-ctime",
+    "-cmin",
+    "-empty",
+    "-link",
+    "-links",
+    "-regex",
+    "-iregex",
+    "-path",
+    "-ipath",
+    "-wholename",
+    "-iwholename",
+    "-user",
+    "-group",
+    "-uid",
+    "-gid",
+    "-readable",
+    "-writable",
+    "-executable",
+    "-xtype",
+    "-samefile",
+    "-inum",
+];
+
+fn has_native_find_action(args: &[String]) -> bool {
     args.iter()
-        .any(|a| UNSUPPORTED_FIND_FLAGS.contains(&a.as_str()))
+        .any(|a| NATIVE_FIND_ACTION_TOKENS.contains(&a.as_str()))
+}
+
+fn has_native_find_narrowing_predicate(args: &[String]) -> bool {
+    args.iter()
+        .any(|a| NATIVE_FIND_NARROWING_PREDICATES.contains(&a.as_str()))
+}
+
+fn should_reject_native_find_action(args: &[String]) -> bool {
+    has_native_find_action(args) && !has_native_find_narrowing_predicate(args)
+}
+
+fn should_run_native_find(args: &[String]) -> bool {
+    if should_reject_native_find_action(args) {
+        return false;
+    }
+
+    if args
+        .iter()
+        .any(|a| NATIVE_FIND_FALLBACK_TOKENS.contains(&a.as_str()))
+    {
+        return true;
+    }
+
+    has_native_find_flags(args)
+        && args.iter().any(|a| {
+            a.starts_with('-') && !SUPPORTED_NATIVE_FIND_FLAGS.contains(&a.as_str())
+        })
 }
 
 /// Parse arguments from raw args vec, supporting both native find and RTK syntax.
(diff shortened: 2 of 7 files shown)


src/cmds/system/find_cmd.rs
```diff
@@ -62,17 +63,85 @@ fn has_native_find_flags(args: &[String]) -> bool {
...
+    "-size", "-mtime", "-mmin", "-atime", "-amin", "-ctime", "-cmin", "-empty", "-link",
+    "-links", "-regex", "-iregex", "-path", "-ipath", "-wholename", "-iwholename", "-user",
+    "-group", "-uid", "-gid", "-readable", "-writable", "-executable", "-xtype", "-samefile",
+    "-inum", "-true", "-false",
 ];
 
-fn has_unsupported_find_flags(args: &[String]) -> bool {
+const SUPPORTED_NATIVE_FIND_FLAGS: &[&str] = &["-name", "-iname", "-type", "-maxdepth"];
+const NATIVE_FIND_ACTION_TOKENS: &[&str] = &["-exec", "-execdir", "-ok", "-okdir", "-delete"];
+const NATIVE_FIND_NARROWING_PREDICATES: &[&str] = &[
```
KuSh, line 81: The predicate list is the wrong signal, and it is also load-bearing in a way it cannot support.

Two concrete failures, both measured on this branch:

- **`-exec` argument boundaries.** These tokens are matched anywhere in argv, including inside the command `-exec` runs. `rtk find . -exec echo {} \;` is **blocked**, but `rtk find . -exec echo -size {} \;` **runs** — adding one unrelated word to the exec command silently disarms the guard.
- **Predicates that narrow nothing.** `-type f`, `-readable`, `-writable` and `-links` all unlock a bare `-delete`, and `rtk find -delete -type f` deleted the whole fixture.

This list is exactly the kind of hand-rolled argv classification #3681 (`arg_tokenizer`) is meant to replace. I would not grow it further — a correct version needs to know which tokens are predicates *in find's grammar* versus operands of an action, which is a tokenizer job, not a membership test.

src/cmds/system/find_cmd.rs
```diff
@@ -62,17 +63,85 @@ fn has_native_find_flags(args: &[String]) -> bool {
...
-        .any(|a| UNSUPPORTED_FIND_FLAGS.contains(&a.as_str()))
+        .any(|a| NATIVE_FIND_ACTION_TOKENS.contains(&a.as_str()))
+}
+
+fn has_native_find_narrowing_predicate(args: &[String]) -> bool {
+    args.iter()
+        .any(|a| NATIVE_FIND_NARROWING_PREDICATES.contains(&a.as_str()))
+}
+
+fn should_reject_native_find_action(args: &[String]) -> bool {
```
KuSh, line 125: This is the core issue: the check is order-insensitive, but `find` evaluates its expression left to right, so the predicate is irrelevant to whether `-delete` already fired.

Measured on this branch against a 6-entry fixture, all six pass the guard and destroy the tree:

```
rtk find -delete -name zzz        RAN   6 -> 1
rtk find -delete -type f          RAN   6 -> 1
rtk find -delete -size +1k        RAN   6 -> 1
rtk find . -delete -name zzz      RAN   6 -> 1
rtk find -name zzz -o -delete     RAN   6 -> 1
rtk find -name '*.tmp' -delete    RAN   deletes from the implicit cwd
```

`-name zzz -o -delete` is worth singling out: the predicate is in the correct position and still means nothing, because `-o` makes it a disjunction. So this cannot be fixed by checking order alone — it needs the parsed expression.

The discriminator is also inverted relative to the PR description. You justify this by these forms defaulting "to the current directory", which points at a **missing path operand**, not a missing predicate — yet as written the guard rejects `rtk find . -delete` (explicit path, unambiguous intent) and allows `rtk find -name '*.tmp' -delete` (implicit cwd).

I prototyped the path-based version on top of `develop`, where `dispatch()` already splits `options` / `paths` / `expr` so `paths` is in scope at the decision point (`paths.is_empty() && expr` contains a mutating action):

```
rtk find . -delete                 rc=0  4 -> 1   allowed (explicit path)
rtk find -delete                   rc=1  4 -> 4   refused
rtk find -name '*.tmp' -delete     rc=1  4 -> 4   refused
rtk find -delete -name nomatch     rc=1  4 -> 4   refused
```

That closes every bypass above with no expression parsing. **One caveat you should decide on:** it fails `develop`'s existing `rtk_pattern_syntax_with_find_predicates_never_blocks`, because legacy syntax `rtk find '*.rs' -exec rm {} \;` also has no path operand. That invariant came from a parsing concern, and this is a safety concern, so narrowing it looks defensible — but it is not my call to make.

src/cmds/system/find_cmd.rs
```diff
@@ -62,17 +63,85 @@ fn has_native_find_flags(args: &[String]) -> bool {
...
+        return false;
+    }
+
+    if args
+        .iter()
+        .any(|a| NATIVE_FIND_FALLBACK_TOKENS.contains(&a.as_str()))
+    {
+        return true;
+    }
+
```
KuSh, line 140: **Regression, independent of the guard.** This clause scans every `-`-prefixed token without skipping the value consumed by the preceding flag, so RTK's own trailing flags get routed into native `find`, which then rejects them:

```
rtk find . -name '*.rs' -m 2
  merge-base 4f4a6a0   rc=0   "unknown flag '-m', ignored" + results
  this PR              rc=1   /usr/bin/find: unknown predicate `-m'
  develop  e53ec1cf    rc=0   correct capped results
```

Same for `-t f`. That is a hard failure against your own merge-base as well as against `develop`, and CI would not have caught it — no build job ran on this PR.

The same clause also makes dispatch depend on user data: `rtk find . -name '-lock*'` drops to raw native output because the *pattern value* starts with `-`, where base and `develop` both compress it. (Minor on its own — I measured no meaningful byte difference for ordinary patterns — but same root cause.)

Both need each flag's value consumed before the scan, which is again #3681 territory rather than something to open-code here. `develop`'s `peel_trailing_rtk_flags` already handles the `-m`/`-t` case correctly, so rebasing may make this disappear on its own.

src/cmds/system/find_cmd.rs
```diff
@@ -177,7 +240,17 @@ fn parse_rtk_find_args(args: &[String]) -> Result<FindArgs> {
 }
 
 /// Entry point from main.rs — parses raw args then delegates to run().
-pub fn run_from_args(args: &[String], verbose: u8) -> Result<()> {
+pub fn run_from_args(args: &[String], verbose: u8) -> Result<i32> {
+    if should_reject_native_find_action(args) {
+        anyhow::bail!(
+            "rtk find refuses native find actions like -delete/-exec without a narrowing predicate; use `rtk proxy find ...` for raw native find behavior"
```
KuSh, line 246: Refusing here is the specific behaviour #3410 was filed about, so I would not add it back in the handler.

Because the rewrite rule captures `find` unconditionally, this refusal lands on commands the user typed as plain `find`. Verified on 0.46.0:

```
$ rtk rewrite find -delete
rtk find -delete
```

So the user asks to delete files, nothing is deleted, the only signal is a message that reads like advice, and the nonzero exit aborts any `&&` chain after it. Under an agent that is worse than either extreme — it believes the cleanup happened. `develop` moved from refuse to passthrough in `6370e792` precisely to get out of this.

The narrowing belongs one layer up, in `src/discover/rules.rs`, where #3410 suggests it and where `kubectl` (`rules.rs:403`) already demonstrates the pattern: rewrite the forms you can summarise, pass the rest through untouched. `find … -delete` would then reach real `find` and do what the user asked, while `rtk find` never has to decide whether a deletion is safe.

If you do keep a refusal here, it needs to be narrow enough that it never fires on a command that would otherwise have worked.

README.md
```diff
@@ -153,6 +153,8 @@ rtk grep "pattern" .            # Grouped search results
 rtk diff file1 file2            # Condensed diff
 ```
 
+`rtk find` compacts simple searches such as `-name`, `-iname`, `-type`, and `-maxdepth`. Compound native expressions/actions such as `(...)`, `-o`, `!`, `-exec`, `-delete`, or `-print0` automatically fall back to native `find`; use `rtk proxy find ...` when you want raw native behavior explicitly.
```
KuSh, line 156: This paragraph says `-exec` and `-delete` "automatically fall back to native `find`", but the guard added in this same PR refuses them — `rtk find . -exec echo {} \;` exits 1 without running. The refusal is not documented anywhere, so a user hitting it has nothing to go on.

Whichever way the guard ends up, this sentence and the new troubleshooting entry need to state which forms are refused and how to get the native behaviour.

KuSh (changes requested):
Thanks for taking this on — `rtk find -delete` really does wipe the working tree on `develop` today (measured: 5 of 6 fixture entries gone), so the problem you are chasing is real.

I built this branch, its merge-base (`4f4a6a0`) and `develop` (`e53ec1cf`), and ran the destructive forms against throwaway trees. The exact form you targeted is genuinely fixed. Requesting changes for three reasons, in order of weight.

### 1. The guard blocks one spelling, not the class

Each of these carries a "narrowing predicate", passes the guard, and deletes the whole tree (6-entry fixture, run from its root):

```
rtk find -delete -name zzz        RAN   6 -> 1
rtk find -delete -type f          RAN   6 -> 1
rtk find -delete -size +1k        RAN   6 -> 1
rtk find . -delete -name zzz      RAN   6 -> 1
rtk find -name zzz -o -delete     RAN   6 -> 1
rtk find -name '*.tmp' -delete    RAN   deletes from the implicit cwd
```

Three mechanisms defeat a flat token scan — evaluation order, boolean structure, and `-exec` argument boundaries. Details inline.

### 2. Refusing inside `rtk find` is the behaviour #3410 was filed about

The hook still rewrites every `find` unconditionally — verified on 0.46.0, `find -delete` becomes `rtk find -delete`. So a refusal in the handler turns the user's real command into a silent no-op that also breaks `&&` chains, under an agent especially. That is exactly #3410, and it is why `develop` deliberately moved from refuse to passthrough (`6370e792`).

#3410's own suggested fix, and the `kubectl` precedent at `rules.rs:403`, put the narrowing at the **rewrite layer** instead: don't capture destructive `find` at all, let it through untouched. That gets the safety you want *and* still runs the user's command. I think that is the shape this should take.

### 3. Please don't hand-roll more argv logic yet

#3681 (`arg_tokenizer`) is open and mergeable, and it is the missing primitive here. A guard that must respect find's option/path/expression split, left-to-right evaluation and `-exec` boundaries should be built on that, not on `args.iter().any(...)`. Until it lands, I would keep this PR to the smallest guard that cannot be bypassed — or to the rewrite-layer narrowing above, which needs no expression parsing at all.

### Rebase, don't merge

All three source files conflict. Since this PR opened, `develop` rewrote `find_cmd.rs` across 18 commits and **already implements the native fallback** (`VERBATIM_ACTIONS` -> `run_passthrough`) — verified, `-not` and `( ... -o ... )` work there today. The only part `develop` lacks is the guard, so a rebase should shrink this a lot.

`develop` also fixed three `parse_native_find_args` defects that are still present at your merge-base and on this branch, so a mechanical rebase risks reintroducing them. **These are not defects you introduced** and both are already tracked — #2839 (multiple paths after the first silently dropped) and #2821 (a dropped predicate silently runs a broader query, which also covers repeated `-name` and a `-name` with no value). Flagging them only so they don't come back with the rebase.

### Not blocking

- The `rtk test` half is sound — I went looking for a collision and found none. `develop` emits `sh: 0: Illegal option -d`, this branch gives a targeted message, `rtk run "test -d path"` works and propagates 0/1, real runner invocations (`cargo test`, `pytest -x`) are untouched, and `-h` is still intercepted by clap. It is unrelated to the `find` fix though, and would land much faster as its own PR.
- `cargo fmt --all --check`, `clippy --all-targets` and the branch's own tests are clean. Worth knowing CI validated none of this: only the CLA ran, `check-target` skipped every build job.

### Suggested test

Once the shape is settled, a differential test would close the class rather than an instance: for a table of action forms, run `rtk find` and native `find` against identical temp trees and assert the resulting file sets match, except for forms deliberately refused. That is what surfaced all six bypasses above.

antonio-orionus:
Splitting per your review: the `rtk test` half is now #3907 (fresh branch off current develop, test-only changes). This PR can stay focused on the `find` side (rewrite-layer narrowing per #3410 / #3681), or I can close it if you prefer a fresh PR for that work.

KuSh:
Hi @antonio-orionus, thanks for your feedback! It's your call whether you'd prefer to reuse the existing PR or start fresh with a new one. If you decide to start from scratch, please reference this PR so we can maintain the history.

PR #2289, opened 2026-06-05
refacto(cmds): strip decorator noise from filter output

aeppling:
Separator lines (`═══`, `---`) and an emoji status marker added tokens to filtered output without giving the LLM any extra signal — the summary headers and section labels already delimit the content.

RTK output must never add noise over the raw command. This strips those decorators across the command filters, keeps the semantic labels (e.g. `Changes:`), and replaces the one emoji marker with plain monochrome unicode.

No behavior change beyond cleaner output. `cargo fmt`/`clippy` clean, all tests pass.


KuSh (approved):
LGTM 🚀

PR #2294, opened 2026-06-06
fix(vitest): preserve explicit reporters

### LINKED ISSUE #2233: rtk vitest filter is not --reporter-aware: --reporter=verbose is ignored and output collapses to PASS/FAIL counts
## Summary

The `rtk vitest` filter is not `--reporter`-aware: it collapses output to `PASS (n) FAIL (n)` even when the user explicitly passes `--reporter=verbose`, dropping the per-test detail that was explicitly requested.

Per CONTRIBUTING.md → *Correctness VS Token Savings*, explicit detail flags (e.g. `cargo test -- --nocapture`) are supposed to pass through. This is the vitest-`--reporter` analog of #1499 (git `branch -vv` has no `-vv` awareness).

## Repro (rtk 0.42.0, macOS, vitest 2.1.1, node 22)

```
$ rtk vitest constants/publicPaths.test.ts --reporter=verbose
PASS (11) FAIL (0)
```

Raw output for the same command (`rtk proxy ./node_modules/.bin/vitest run constants/publicPaths.test.ts --reporter=verbose`) prints all 11 named tests plus the normal summary.

The hook is not at fault — it forwards the flag correctly (`npx vitest run X --reporter=verbose` → `rtk vitest X --reporter=verbose`). The collapse happens inside the vitest filter itself.

## Expected

With an explicit `--reporter=` (especially `verbose`), pass the detailed reporter output through, mirroring `cargo test -- --nocapture`.

## Related (not duplicates)

- #220 — same flag-awareness gap for vitest `--coverage` (different lost data: coverage vs per-test detail).
- #1499 — git `branch -vv` filter has no flag awareness; same class of bug, different command.
- #1813 — vitest *failure* list is truncated. This report is an all-passing run where the reporter flag is ignored.
- #1836 / #1035 — parser *fails* and swallows output. Here the parser succeeds (`PASS (11) FAIL (0)` is correct); only the explicit flag is ignored.

## Suggested fix

Make the vitest filter flag-aware: when an explicit `--reporter=` is present, fall back to near-raw passthrough.

jsiu93:
Closes #2233.

## Summary

- Preserve explicit Vitest `--reporter` arguments instead of replacing them with `--reporter=json`.
- Bypass Vitest summary compression when the user requested a reporter, so verbose per-test output remains visible.
- Keep the default compact JSON reporter path unchanged when no reporter is specified.

## Test plan

- [x] `cargo test --quiet vitest_cmd::tests`

### DIFF
--- a/src/cmds/js/vitest_cmd.rs
+++ b/src/cmds/js/vitest_cmd.rs
@@ -206,16 +206,15 @@ fn extract_failures_regex(output: &str) -> Vec<TestFailure> {
 
 pub fn run_test(command: &Commands, args: &[String], verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
+    let mut passthrough_requested = false;
 
     let (framework, mut cmd) = match command {
         Commands::Vitest { .. } => {
             let framework = "vitest";
             let mut cmd = package_manager_exec(framework);
-            cmd
-                // Force non-watch mode
-                .arg("run")
-                // Enable JSON structured output
-                .arg("--reporter=json");
+            let effective_args = build_vitest_effective_args(args);
+            passthrough_requested = effective_args.passthrough;
+            cmd.args(effective_args.args);
             (framework, cmd)
         }
         Commands::Jest { .. } => {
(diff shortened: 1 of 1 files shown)


aeppling:
Hey @jsiu93, thanks for addressing this issue. Confirmed it fixes the bug with no regression. Two things before merge:

One blocker: truncated passthrough has no recovery path. On a large all-passing --reporter=verbose run, output is capped at passthrough_max_chars (verified: 80-test suite returns 15 tests) with the truncation marker but no [full output] hint and no tee file, since the existing tee_and_hint is failures-only and the run exits 0.

Reuse force_tee_hint or force_tee_tail_hint from src/core/tee.rs (both tee regardless of exit code) so the full output stays recoverable, per the CONTRIBUTING checklist. Avoid double-teeing with the tee_and_hint already in run_test.

Test gap. None of the new tests exceed passthrough_max_chars, which is why this slipped through. Add one with input larger than the cap asserting both the marker and the recovery hint.

jsiu93:
Thanks @aeppling, fixed in f9469d1.

The Vitest passthrough path now carries a truncated flag. When explicit `--reporter` output exceeds `passthrough_max_chars`, `run_test` uses `force_tee_hint` so successful verbose runs get a `[full output]` tee file. Regular output still uses `tee_and_hint`, which keeps the existing recovery behavior and avoids double teeing.

I added a regression test with output over the cap that asserts both the truncation marker and the recovery hint.

Verified locally with `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test`, and `cargo test --all`. I also smoke-tested against `vitejs/vite`: 129 passing tests with `--reporter=verbose` now print the truncation marker plus `[full output]`, and the tee log contains the full test summary.

aeppling:
LGTM, will be in next release.


aeppling (approved):

PR #2296, opened 2026-06-06
fix(git): preserve patch content when `git log -p` is used

### LINKED ISSUE #2275: "git log -p" omits the patch requested by the -p flag

ousamabenyounes:
## Problem

When agents use `git log -p` to inspect git history, RTK strips the patch/diff content, replacing it with `[+N lines omitted]` markers. The `-p` flag explicitly requests patch output, and this diff context is vital for diagnosing problems.

Root cause: with no format flag, RTK injects `--pretty=format:…%b%n---END---` and then runs `filter_log_output`, which treats the diff hunks as *commit body* — keeping only 3 lines and discarding the rest behind `[+N lines omitted]`.

## Fix

When `-p` / `--patch` is detected, RTK skips its `--pretty`/limit/`--no-merges` defaults **and** the commit-block filter, printing git's raw output verbatim. Normal `git log` / `--oneline` behaviour is unchanged.

The filter-vs-passthrough decision is now a single pure function (`finalize_log_output`), and `run_log` builds the command once instead of duplicating the exec/error/print/timer block.

### GREEN — with the fix (post-rebase, 2026-06-09)

```
$ cargo test -- git::tests::test_finalize -- --nocapture
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.20s

running 2 tests
test cmds::git::git::tests::test_finalize_preserves_patch_when_requested ... ok
test cmds::git::git::tests::test_finalize_filters_when_no_patch_flag ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 2154 filtered out

$ cargo test --all
test result: ok. 2148 passed; 0 failed; 8 ignored; 0 measured; 0 filtered out
```

### RED — fix reverted (`finalize_log_output` always filters)

```
$ cargo test -- git::tests::test_finalize
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.20s

running 2 tests
test cmds::git::git::tests::test_finalize_preserves_patch_when_requested ... FAILED
test cmds::git::git::tests::test_finalize_filters_when_no_patch_flag ... ok

failures:

---- cmds::git::git::tests::test_finalize_preserves_patch_when_requested stdout ----
thread 'cmds::git::git::tests::test_finalize_preserves_patch_when_requested' panicked at src/cmds/git/git.rs:2248:9:
assertion `left == right` failed: patch output must be byte-identical passthrough
  left: "commit a1ef0ef...  [+17 lines omitted]"
 right: "commit a1ef0ef...\nMerge: e1cd274 c873e19\n...\ndiff --git a/Cargo.toml b/Cargo.toml\n..."

test result: FAILED. 1 passed; 1 failed; 0 ignored
```

The reverted code collapses the entire diff into `[+17 lines omitted]`; the fix preserves every hunk.

### CI pre-commit gate

```
cargo fmt --all -- --check       → clean
cargo clippy --all-targets -- -D warnings → clean
cargo test --all                 → 2148 passed, 0 failed, 8 ignored
```

> Note: no token-savings test here — patch passthrough is intentionally a 0%-reduction path (the diff IS the payload for agents).

Closes #2275

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -346,7 +346,12 @@ bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 # ===================
 section "curl"
 if command -v curl &> /dev/null; then
-  bench "curl json" "curl -s https://mockhttp.org/json/1" "$RTK curl https://mockhttp.org/json/1"
+  CURL_JSON_FIXTURE=$(mktemp)
+  cat > "$CURL_JSON_FIXTURE" << 'JSONEOF'
+{"message":"Hello from RTK benchmark","status":"success","code":200}
+JSONEOF
+  bench "curl json" "curl -s file://$CURL_JSON_FIXTURE" "$RTK curl file://$CURL_JSON_FIXTURE"
+  rm -f "$CURL_JSON_FIXTURE"
   bench "curl text" "curl -s https://mockhttp.org/robots.txt" "$RTK curl https://mockhttp.org/robots.txt"
 fi
 
@@ -355,8 +360,8 @@ fi
 # ===================
 if command -v wget &> /dev/null; then
   section "wget"
-  bench "wget" "wget -qO- https://mockhttp.org/json/1" "$RTK wget https://mockhttp.org/json/1"
-  rm -f 1 2>/dev/null
+  bench "wget" "wget -qO- https://mockhttp.org/json" "$RTK wget https://mockhttp.org/json"
+  rm -f json 2>/dev/null
 fi
 
 # ===================
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -433,87 +433,65 @@ fn run_log(
 ) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
 
-    // When user explicitly requests patch output (-p / --patch), pass through
-    // raw git output without RTK formatting or filtering. The patch content is
-    // vital context for agents diagnosing problems via git history.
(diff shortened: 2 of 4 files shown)


ousamabenyounes:
Rebased onto current `origin/develop` and pushed benchmark fix `5b24a8f`.

What changed:

Local verification:
```text
cargo fmt --all -- --check
cargo test patch
34 passed
cargo test cargo_build_success_uses_raw_when_summary_is_larger
1 passed
cargo test --all
2443 tests: passed, 8 ignored
cargo clippy --all-targets
cargo build --release
CI=1 ./scripts/benchmark.sh
0 negative, 0 fail (exit 0)
```

ousamabenyounes:
Closing as superseded: `develop` already handles `git log -p` passthrough, more completely than this PR. `run_log` now does `if requests_raw_log_output(args) { return run_passthrough(...) }`, and `requests_raw_log_output` covers `-p | -u | --patch | --patch-with-raw | --patch-with-stat` and stops at `--` (this PR only handled `-p`/`--patch`), with its own passthrough test. Nothing here to add on top. Reopen if I've missed a case the current code doesn't cover.

PR #2297, opened 2026-06-06
feat(filters): global layers pipeline and levels (v1: decorative only)

aeppling:
## Summary

- Add a centralized, layered output **pipeline** (`core::pipeline`), applied in `runner` before each command's own filter — for both **captured and streamed** commands. First layer: **decorative** (strip ANSI, collapse blank runs, drop box-drawing at `high`), user-tunable via `[levels].decorative` or `RTK_DECORATIVE` (`light`/`reasonable`/`high`).

- Add a **global fallback pipeline**: commands with no `cmds/` handler and no TOML filter now stream through the pipeline too, TTY-gated (terminal → passthrough, piped → filter), with a built-in + user-extensible (`[levels].exclude`) exclude list so raw-output commands (`cat`,  `head`, `base64`, …) stay byte-exact. Routing order unchanged: cmds → TOML → fallback.

- Remove now-redundant per-command ANSI stripping for commands the pipeline covers.

> Scope: applies to all `runner`-routed commands (captured + streamed) and the fallback.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` ,  2078 passed, 0 failed; clippy clean
  - `rtk printf '…\x1b…\n\n\n…'` → ANSI stripped, blank runs collapsed; `RTK_DECORATIVE=light` preserves blanks
  - `rtk cat <file>` → raw passthrough (excluded); `[levels].exclude=["printf"]` → printf raw
  - `rtk df` → TOML filter path unaffected (`RTK_NO_TOML=1` falls through to the pipeline)
  - exit codes propagate: `false` → 1, unknown command → 127


aeppling:
Follow-up after first review of core feature

- Adapt features for this new global pipeline ( history , gains, disco, telemetry,...)
- Ensure explicit documentation, ensure synchronized on landing page

PR #2298, opened 2026-06-06
test(telemetry): isolate device salt tests

### LINKED ISSUE #2106: bug(tests): telemetry tests write a real `.device_salt` to the user's data dir
## Summary

Several unit tests in `src/core/telemetry.rs` call the production salt / device-hash functions directly with no path override, so `cargo test` writes a real `.device_salt` to the user's data dir (`~/Library/Application Support/rtk/.device_salt` on macOS, `~/.local/share/rtk/.device_salt` on Linux).

## Reproduce (macOS)

```bash
rm -f "~/Library/Application Support/rtk/.device_salt"
git clone https://github.com/rtk-ai/rtk
cd rtk
cargo test -p rtk telemetry
ls -la "~/Library/Application Support/rtk/.device_salt"
# -> file exists, owned by your user, 64 bytes
```

The tests pass, but they have leaked a real salt to the test runner's host filesystem.

## Affected tests

`src/core/telemetry.rs:447-487`:

- `test_device_hash_is_stable` — calls `generate_device_hash()`
- `test_device_hash_is_valid_hex` — calls `generate_device_hash()`
- `test_salt_is_persisted` — calls `get_or_create_salt()`

All three go through `salt_file_path()` → `dirs::data_local_dir()`, which always resolves to the user's real data dir.

## Why this matters

1. Running `cargo test` on a fresh dev machine produces a salt that the user never consented to creating (telemetry consent is `None`, but the salt file is now present).
2. Hides issue #1656 from anyone investigating on a machine where they've run `cargo test`: after running the test suite, `rtk telemetry status` shows a real device hash and the `(no salt file)` symptom is gone — even though no `rtk init` was ever run.
3. CI runners that persist `$HOME` across jobs (less common, but possible on self-hosted) would leak a salt between runs.

## Suggested fix shape

Inject the path. One option: have `salt_file_path()` accept an `Option<&Path>` override, default behavior unchanged but tests pass a `tempfile::TempDir`. Another: a `RTK_DATA_DIR` env var read inside `salt_file_path()`, set in test setup. Either way, the tests should use `tempfile` or equivalent so nothing leaks to the user dir.

## Environment

- rtk `develop` HEAD (verified locally 2026-05-27)
- macOS 25.5 (Darwin), but the path leak affects every platform `dirs::data_local_dir()` resolves on

## Relation to #1656

Tracked separately rather than bundled into PR #2105, per [CONTRIBUTING.md scope rules](https://github.com/rtk-ai/rtk/blob/master/CONTRIBUTING.md#scope-rules). The two issues share an ancestor (`get_or_create_salt`) but have different fixes: #1656 is a status-message UX bug; this one is a test-isolation correctness bug.

Peter7896:
## Summary
- avoid telemetry unit tests touching the production `.device_salt` path by moving salt persistence into a path-injected helper
- keep production `generate_device_hash` behavior unchanged through the cached default salt path
- cover temp-path persistence and invalid salt replacement

## Validation
- cargo fmt --all --check
- cargo test telemetry

Fixes #2106

### DIFF
--- a/src/core/telemetry.rs
+++ b/src/core/telemetry.rs
@@ -8,7 +8,7 @@ use crate::hooks::init::resolve_claude_dir;
 use sha2::{Digest, Sha256};
 use std::fmt::Write as FmtWrite;
 use std::io::Write as IoWrite;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::OnceLock;
 
 static CACHED_SALT: OnceLock<String> = OnceLock::new();
@@ -156,6 +156,10 @@ fn send_ping() -> Result<(), Box<dyn std::error::Error>> {
 
 pub fn generate_device_hash() -> String {
     let salt = get_or_create_salt();
+    hash_salt(&salt)
+}
+
+fn hash_salt(salt: &str) -> String {
     let mut hasher = Sha256::new();
     hasher.update(salt.as_bytes());
     format!("{:x}", hasher.finalize())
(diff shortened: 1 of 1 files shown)


Peter7896:
Rebased this branch onto current `develop` and resolved the `src/core/telemetry.rs` conflict by preserving the upstream private directory/file helpers inside the path-injected salt helper. This keeps the test isolation from this PR without regressing the newer file-permission hardening.

Validation:
- `cargo fmt --all --check`
- `cargo test telemetry` -> 17 passed

Current head: `bbe1fec39f1bd164d042591ea076fd1bd921e600`.

Issue #2301, opened 2026-06-06
`rtk hook claude` rewrites `rg` → `rtk grep` (BSD grep) instead of `rtk rg`, so path-less ripgrep searches silently return 0 matches / hang on stdin


iansltx:
The following is a copy-paste of a thing Claude Agent (Opus 4.8 @ Max) found when using the hook for `rg` on macOS Tahoe when bouncing around a codebase doing some analysis. It baked a workaround into agent memory, but this feels like it needs a fix here. I validated the sample tool use call as well.

Maybe I'm missing something here and this override is intentional, but CC doesn't seem to like it.

---

### Summary

The Claude Code integration hook (`rtk hook claude`, installed as a `PreToolUse` hook) auto-rewrites **every `rg` command to `rtk grep …`**. `rtk grep` follows **grep (BSD) semantics**, not ripgrep semantics, so the most common ripgrep forms break:

- `rg PATTERN` with **no path** → ripgrep searches the CWD recursively; BSD grep instead **reads stdin** → the command **silently returns 0 matches**, or **hangs** if no stdin is attached.
- `rg PATTERN DIR` → ripgrep recurses into `DIR` by default; BSD grep errors `grep: DIR: Is a directory` unless `-r` is passed.
- ripgrep-only flags (`--type-list`, `-t/--type`, `--files`, etc.) are rejected by grep.

`rtk rg …` (and `rtk proxy rg …`) already wrap **genuine ripgrep 15.1.0 correctly**, so the bug is isolated to the hook choosing `rtk grep` as the rewrite target instead of `rtk rg`.

**Why this matters:** the failure is *silent and wrong*, not a loud error. An agent (or user) runs `rg "SomeSymbol"`, gets **no output**, and concludes the symbol doesn't exist — when it does. This produces confidently-incorrect conclusions during code search/analysis. (It also intermittently hangs automated runs that don't redirect stdin.)

### Environment

- `rtk` **0.42.3**
- `ripgrep` **15.1.0** at `/usr/local/bin/rg` (the genuine binary is fine)
- macOS (Darwin 25.x); system grep is **BSD grep 2.6.0-FreeBSD**
- Host: Claude Code, with `~/.claude/settings.json` → hooks → `"command": "rtk hook claude"`

### Root cause (verified)

The hook unconditionally rewrites `rg` to `rtk grep`. Captured directly by piping the hook's PreToolUse JSON:

```console
[...]

PR #2302, opened 2026-06-06
fix(hook): rewrite rg through rtk rg

ded-furby:
## Summary

- Fixes hook rewrite behavior from issue #2301 by routing rg to rtk rg instead of rtk grep.
- Keeps grep rewrite behavior intact and adds regression coverage for rg rewrite in Rust tests.
- Updates the hook rewrite shell script expectation for rg to match the corrected backend command.

## Testing

- cargo test rewrite_rg -- --nocapture
- cargo run -- rewrite "rg pattern src/"
- cargo run -- rewrite "rg --version"

## Notes

The repository's legacy shell rewrite test suite is not fully runnable in this environment due missing hook audit/logging side effects, so validation was done via focused registry tests and direct rtk rewrite command checks.


truffle-dev:
The rewrite direction here looks right for `rg`, but the single-rule change pulls plain `grep` along with it. `rules.rs:91` keeps the `r"^(rg|grep)\s+"` pattern and `rewrite_prefixes: &["rg", "grep"]`, so after the new `rtk_cmd: "rtk rg"` lands, `grep -rn pattern src/` rewrites to `rtk rg -rn pattern src/` too.

The legacy shell suite encodes the prior intent at `hooks/claude/test-rtk-rewrite.sh:104-106`:

```
test_rewrite "grep -rn pattern src/" \
  "grep -rn pattern src/" \
  "rtk grep -rn pattern src/"
```

It wasn't touched in this PR (and you flagged the suite isn't fully runnable in your env), but the assertion is what would catch the side effect: `grep -rn pattern src/` rewrites to `rtk rg -rn pattern src/` under the new rule. That's the inverse mirror of the bug iansltx reported, ripgrep semantics applied to a `grep` invocation (.gitignore filtering on by default, Rust regex flavor, grep-only flags like `-d skip` rejected).

Two intents fit the reporter's evidence, and it's worth picking one explicitly:

- **A. Search unification.** Both `rg` and `grep` map to `rtk rg` because the goal is to retire BSD grep entirely from the rewrite path. PR as-is plus removing the `grep -rn pattern src/` → `rtk grep -rn pattern src/` assertion from the shell suite.
- **B. Tool fidelity.** `rg` maps to `rtk rg`, `grep` maps to `rtk grep`. Split the rule:

  ```rust
  RtkRule {
      pattern: r"^rg\s+",
      rtk_cmd: "rtk rg",
      rewrite_prefixes: &["rg"],
      category: "Files",
      savings_pct: 75.0,
      subcmd_savings: &[],
      subcmd_status: &[],
  },
  RtkRule {
      pattern: r"^grep\s+",
      rtk_cmd: "rtk grep",
      rewrite_prefixes: &["grep"],
      category: "Files",
      savings_pct: 75.0,
      subcmd_savings: &[],
      subcmd_status: &[],
  },
  ```

  The two patterns don't overlap, and `registry.rs:149` (`take the last (most specific) match`) tolerates either ordering. The `grep -rn pattern src/` shell assertion stays green.

If A is the intent, the regex could also collapse to `^rg\s+` and drop `grep` from `rewrite_prefixes`, so plain `grep` falls through to passthrough rather than getting silently re-targeted. Either way, an explicit `test_rewrite_grep_pattern` in `registry.rs` would lock the decision in.

ded-furby:
Thanks for the review context. The PR currently includes the split rg/grep behavior (rg -> rtk rg, grep -> rtk grep) plus dedicated tests in src/discover/registry.rs hooks/claude test coverage. If anything else is still expected, please point me to a concrete follow-up.

truffle-dev:
Split rules and tests look right. `src/discover/rules.rs:92` and `:101` are the two non-overlapping `RtkRule`s (`^rg\s+` -> `rtk rg` with prefix `["rg"]`; `^grep\s+` -> `rtk grep` with prefix `["grep"]`). `hooks/claude/test-rtk-rewrite.sh:104-114` covers both directions, and `src/discover/registry.rs:3410` exercises the absolute-path branch (`/usr/bin/grep` -> `rtk grep`). Nothing else from me.

PR #2306, opened 2026-06-07
fix(lint): mark truncated parse fallbacks

fengjikui:
## Summary

Follow-up to #2204 for the lint fallback paths that still used silent character truncation when structured output parsing failed.

- use the existing passthrough truncation marker for Ruff JSON parse failures
- use the existing passthrough truncation marker for golangci-lint JSON parse failures
- add regression tests so long malformed JSON output visibly reports `[RTK:PASSTHROUGH] Output truncated`

This does not overlap with #2205, which focuses on git log output.

## Validation

- `cargo fmt --check`
- `cargo test cmds::python::ruff_cmd::tests`
- `cargo test cmds::go::golangci_cmd::tests`
- `cargo test`

## AI assistance

I used Codex to help prepare this patch, reviewed the diff, and ran the validation commands above locally.

### DIFF
--- a/src/cmds/go/golangci_cmd.rs
+++ b/src/cmds/go/golangci_cmd.rs
@@ -4,7 +4,8 @@ use crate::core::config;
 use crate::core::runner;
 use crate::core::stream::exec_capture;
 use crate::core::truncate::CAP_WARNINGS;
-use crate::core::utils::{resolved_command, truncate};
+use crate::core::utils::resolved_command;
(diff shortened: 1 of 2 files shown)


KuSh:
Closing in favour of #2584, which makes the same passthrough truncation marker change in the ruff and golangci-lint parse fallbacks and also covers ESLint, Pylint and the git log cap, closing #2204. Please take any follow-up over to #2584.

PR #2308, opened 2026-06-07
fix(discover): use recorded hook rewrites

### LINKED ISSUE #2299: rtk discover counts hook-rewritten commands as missed savings
## Summary

`rtk discover` reports commands as "missed savings" even when they are actively being rewritten by a PreToolUse hook. The transcript logs the pre-hook command (what Claude requested), but the hook rewrites it before execution. `discover` scans the pre-hook text and doesn't account for the rewrite, producing inflated numbers.

## Reproduction

1. Configure a PreToolUse hook in Claude Code `settings.json` that rewrites commands via `rtk hook claude`:
   ```json
   {
     "hooks": {
       "PreToolUse": [{
         "matcher": "Bash",
         "hooks": [{ "type": "command", "command": "/path/to/rtk-unified.sh" }]
       }]
     }
   }
   ```

2. The hook correctly rewrites commands — verified by piping test inputs through `rtk hook claude`:
   ```
   $ echo '{"tool_input":{"command":"git diff"}}' | rtk hook claude
   {"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk git diff"}}}

   $ echo '{"tool_input":{"command":"rg -n \"pattern\" src/"}}' | rtk hook claude
   {"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk grep -n \"pattern\" src/"}}}
   ```

3. Run `rtk discover --all --since 7`:
   ```
   Already using RTK: 1036 commands (14.5%)

   MISSED SAVINGS
   git diff    922    ~148.3K tokens
   rg -n       973    ~81.4K tokens
   ls -la      424    ~16.8K tokens
   find        249    ~47.1K tokens
   ...
   Total: 3431 commands -> ~410.6K tokens saveable
   ```

4. But `rtk gain` confirms those commands ARE being processed through RTK:
   ```
   Total commands: 1167
   rtk grep: 201, rtk read: 216, rtk git diff, rtk git commit: 38, ...
   ```

## Root cause

Claude Code session transcripts (`.jsonl`) log `tool_input.command` — the **pre-hook** command. The hook emits `updatedInput` to rewrite it, but the transcript preserves the original. `rtk discover` scans the original and doesn't check whether a hook rewrite occurred.

## Suggested fix

One or more of:
- Detect the presence of an RTK PreToolUse hook and exclude commands that would match `rtk hook claude` from the "missed" count
- Cross-reference against `rtk gain` data to filter already-processed commands
- Note in the output when a hook is detected: "Note: PreToolUse hook active — these commands may already be rewritten at execution time"

## Environment

- rtk 0.42.3
- Claude Code with PreToolUse hook via `rtk hook claude`
- Linux

fengjikui:
## Summary

Fixes #2299 by making Claude session discovery use the command that RTK actually recorded in the matching `PreToolUse:Bash` hook attachment.

What changed:
- collect successful PreToolUse hook attachments while scanning Claude JSONL transcripts
- accept only hook output marked with `permissionDecisionReason: "RTK auto-rewrite"`
- match the attachment's `toolUseID` to the original Bash `tool_use.id`
- classify the recorded `updatedInput.command` instead of the pre-hook command
- preserve the original command for unrelated hooks, malformed output, or unmatched IDs

This handles mixed command chains precisely: already rewritten `rtk ...` segments count as existing RTK usage, while unsupported or untouched segments remain visible to discover. It does not infer rewrite behavior from installed-hook state.

## Evidence

The original transcript records the command before hooks run. A later real-world reproduction on #2299 confirmed that the same transcript also records `hook_success` attachments with the Bash `toolUseID` and RTK's JSON stdout: https://github.com/rtk-ai/rtk/issues/2299#issuecomment-4915114464

The regression test mirrors that structure, including an unrelated successful hook before the RTK hook, and proves only the exact RTK marker replaces the command.

## Independent validation

An independent contributor tested this head against 30 days of production Claude Code transcripts with `rtk hook claude` active: 178 sessions and 1,954 Bash tool calls. They cross-checked the PR output with a separate `tool_use.id` / `hook_success.toolUseID` join and confirmed that the recorded-rewrite approach avoids false positives from pre-install sessions, exclusions, `RTK_DISABLED=1`, and rewrite guards. Their focused `cargo test --bin rtk discover` run passed 458 tests.

Validation report: https://github.com/rtk-ai/rtk/pull/2308#issuecomment-4948675849

## Validation

- `cargo test discover::` -> 457 passed
- `cargo test` -> 2437 passed, 8 ignored
- `cargo fmt --check`
- `git diff --check`

AI-assisted contribution, manually scoped and verified on current `develop` (`5d32d0736f686b69d1e8b9dc45c007d4eb77a0a2`).

### DIFF
--- a/src/discover/provider.rs
+++ b/src/discover/provider.rs
@@ -160,10 +160,12 @@ impl SessionProvider for ClaudeProvider {
             .unwrap_or("unknown")
             .to_string();
 
-        // First pass: collect all tool_use Bash commands with their IDs and sequence
-        // Second pass (same loop): collect tool_result output lengths, content, and error status
+        // Collect Bash tool uses, their results, and any RTK PreToolUse rewrite recorded by
+        // Claude Code. Matching happens after the scan because attachments and results follow
+        // the assistant tool_use entry in the transcript.
         let mut pending_tool_uses: Vec<(String, String, usize)> = Vec::new(); // (tool_use_id, command, sequence)
         let mut tool_results: HashMap<String, (usize, String, bool)> = HashMap::new(); // (len, content, is_error)
+        let mut rewritten_commands: HashMap<String, String> = HashMap::new();
         let mut commands = Vec::new();
         let mut sequence_counter = 0;
 
@@ -173,8 +175,11 @@ impl SessionProvider for ClaudeProvider {
                 Err(_) => continue,
             };
 
-            // Pre-filter: skip lines that can't contain Bash tool_use or tool_result
-            if !line.contains("\"Bash\"") && !line.contains("\"tool_result\"") {
+            // Pre-filter: skip lines that can't contain a Bash tool_use/result or hook record.
+            if !line.contains("\"Bash\"")
+                && !line.contains("\"tool_result\"")
+                && !line.contains("\"hook_success\"")
+            {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


fengjikui:
Retargeted this PR to `develop` and rebased the branch so the diff only includes the two intended files.

Re-ran validation on the `develop` base:
- `cargo fmt --check`
- `git diff --check`
- `cargo test hooks::hook_check`
- `cargo test discover::report`
- full `cargo test` -> 2070 passed, 7 ignored

The `wrong-base` label appears stale now, but I do not have permission to remove it.

fengjikui:
Rebuilt this PR on current `develop` as `598ebc8`. Based on the newer transcript evidence in #2299, the earlier wrapper-detection/report-note approach is gone: the branch now matches Claude `hook_success` attachments to Bash tool uses by `toolUseID` and classifies RTK's recorded `updatedInput.command`. Validation on the exact head: 457 discover tests, full suite (2437 passed / 8 ignored), Clippy with `-D warnings`, fmt, and `git diff --check`.

felipetruman:
Tested this against 30 days of real production transcripts (Claude Code + `rtk hook claude` active; 178 sessions, 1,954 Bash tool calls) and independently verified the approach. This is the right fix among the three open PRs for #2928.

**Why this approach is correct**: it uses the rewrite actually recorded at execution time (`hook_success` attachment, `permissionDecisionReason: "RTK auto-rewrite"`, matched by `toolUseID`), so:

- No false positives: commands in `hooks.exclude_commands`, `RTK_DISABLED=1` prefixes, and rewrite guards (`gh --json`, `cat` with write redirects, ...) are naturally left out because the hook never rewrote them.
- Sessions that predate the hook install keep showing up as missed opportunities — which is what discover is for.
- Users without the hook see no behavior change.

**Validation**:
- `cargo test --bin rtk discover` on this branch: 458 passed, 0 failed.
- Ran the branch binary's `discover` over my real transcripts: "Already using RTK" goes from 0.5% (stock) to 9.3% across all 2,485 sessions — the remaining gap is thousands of short headless/automation sessions with little Bash activity. Scoped to my main interactive project and cross-checked with an independent script that joins `tool_use` ids with `hook_success` attachments: 1.1% reported vs 43.0% actual.
- Verified the attachment format assumption against live transcripts on Claude Code (July 2026): `attachment.type == "hook_success"`, `hookName == "PreToolUse:Bash"`, `stdout` containing `hookSpecificOutput.updatedInput.command` — matches what this PR parses.

Also worth noting: this fix removes a second-order error — today the MISSED SAVINGS estimate for rewritten commands is computed from `tool_result` sizes that rtk already compressed, double-counting savings that were already realized.

Compared to the sibling PRs: #2671 counts every supported-classified command as covered whenever the hook is installed (overcounts pre-install sessions and excluded commands, and erases the residual missed-signal); #2930 simulates `rewrite_command()` without checking the hook is installed at all (would empty the MISSED table for exactly the users discover exists to help). The hook-installed *detection + report note* from #2671 would compose nicely on top of this PR.

PR #2309, opened 2026-06-07
fix(git): rewrite -C with quoted paths

### LINKED ISSUE #2303: git -C /path <subcommand> not rewritten to rtk git
## Problem

RTK's hook correctly rewrites `git <subcommand>` → `rtk git <subcommand>`, but misses the `git -C /path <subcommand>` form (where `-C` sets the working directory inline).

This pattern is common in multi-worktree setups and is the recommended style in Claude Code to avoid a permission system bypass (Claude Code bug #59498 — `cd /path && git` bypasses deny rules entirely, allowing subagents to push/reset without prompting). Users following this best practice end up with all their git commands unhandled by RTK.

## Example

```bash
# Rewritten correctly:
git status → rtk git status

# Not rewritten (passes through unfiltered):
git -C /Users/paul/MyData/Code/FinancialPlatform status
git -C /Users/paul/MyData/Code/FinancialPlatform/.worktrees/feature/foo log --oneline -10
```

## Request

Treat `git -C <path> <subcommand>` the same as `git <subcommand>` in the hook rewriter — strip the `-C <path>` for routing purposes and pass it through to the underlying `rtk git` handler.

## Volume

`rtk discover` shows 119 unhandled `git -C` calls in 30 days for a single project. Since this is the recommended Claude Code pattern, likely affects many users.

fengjikui:
## Summary

Fixes #2303 by recognizing quoted values in Git and Yadm global options before classifying hook rewrites.

What changed:
- replace the whitespace-only global-option regex with the existing shell lexer
- consume value-taking `-C`, `-c`, `--git-dir`, and `--work-tree` options in order
- preserve flag-only global options such as `--no-pager`
- return the original command for incomplete option sequences
- keep the original quoted command text when constructing the final `rtk git ...` rewrite

The normalization is classification-only. It removes global options to identify the subcommand, while rewrite output still uses the original command, including quotes and escaped spaces.

## Reproduction

On current `develop` (`5d32d0736f686b69d1e8b9dc45c007d4eb77a0a2`):

```console
$ cargo run --quiet -- rewrite 'git -C "/tmp/work tree" status'
# exit 1, no output
```

With this branch:

```console
$ cargo run --quiet -- rewrite 'git -C "/tmp/work tree" status'
rtk git -C "/tmp/work tree" status
# exit 3 (ask rewrite)
```

Coverage also includes single-quoted Yadm paths, inline quoted `--git-dir`, mixed `-c`/`--work-tree`/`--no-pager`, and an incomplete `git -C` fallback.

## Validation

- `cargo test discover::` -> 457 passed
- `cargo test` -> 2437 passed, 8 ignored
- `cargo fmt --check`
- `git diff --check`

## AI Assistance

This PR was prepared with OpenAI Codex assistance. I reviewed the diff and ran the validation above on the exact pushed head.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -5,7 +5,7 @@ use lazy_static::lazy_static;
 use regex::{Regex, RegexSet};
 use std::path::Path;
 
-use super::lexer::{split_on_operators, tokenize, TokenKind};
+use super::lexer::{shell_split, split_on_operators, tokenize, TokenKind};
 use super::rules::{IGNORED_EXACT, IGNORED_PREFIXES, RULES};
 
 const PHP_TOOL_NAMES: [&str; 6] = ["phpunit", "phpstan", "ecs", "pest", "paratest", "pint"];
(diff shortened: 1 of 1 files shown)


fengjikui:
Rebuilt this PR on current `develop` as `3cae65b`. The earlier quoted-value regex is replaced with RTK's existing shell lexer, so Git/Yadm global options are parsed without duplicating shell quoting rules while the final rewrite preserves the original command text. Exact-head validation: 457 discover tests, full suite (2437 passed / 8 ignored), Clippy with `-D warnings`, fmt, and `git diff --check`.

PR #2321, opened 2026-06-08
fix(resolve): resolve project-local node_modules/.bin tools (fixes exit 127 for hook-rewritten npx on Windows)

danscMax:
## Problem

The agent Bash hook (`rtk hook claude` / `cursor` / `gemini` / copilot) rewrites `npx <tool>` to `rtk <tool>`, dropping the `npx` launcher. RTK then resolves the tool through `resolve_binary` → `which::which`, which only searches the **system PATH**.

Project-local JS tools — `eslint`, `vite`, `vitest`, `tsc`, `prettier`, `playwright`, … — are installed in `./node_modules/.bin`, which is **not** on the system PATH. So on Windows:

```
$ rtk eslint --version
rtk: Failed to resolve 'eslint' via PATH, falling back to direct exec: Binary 'eslint' not found on PATH
[rtk: program not found]      # exit 127
```

…even though `npx eslint --version` runs fine. `npx`/`npm` add `node_modules/.bin` to `PATH` at runtime; RTK strips the launcher during the rewrite but never replicated that lookup. The result: every hook-rewritten `npx <local-tool>` invocation fails with exit 127 for locally-installed tooling — the common case, since running not-globally-installed tools is exactly what `npx` is for.

## Fix

`resolve_binary` now mirrors npx's resolution order:

1. nearest `node_modules/.bin` (cwd walking up to the filesystem root — handles monorepos / nested packages)
2. then the inherited `PATH`

…via `which::which_in` (PATHEXT-aware on Windows; already used in this module's tests). `tool_exists` uses the same resolution so handlers that branch on it (`tsc_cmd`, `next_cmd`, `prisma_cmd`, …) also detect local tools.

- **Native tools unaffected:** `git`, `cargo`, `npm`, etc. aren't in `node_modules/.bin`, so they fall through to PATH exactly as before — this is an additive change to the search path.
- **Bonus correctness:** a project-local tool now takes precedence over a globally-installed shadow of the same name (e.g. a `playwright` on PATH from another ecosystem), matching npx semantics.

The logic is split into a testable `resolve_binary_in(name, cwd, path_var)` so resolution can be unit-tested without mutating global process state. Adds tests for local-`node_modules/.bin` resolution and the not-found path.

## Verification (Windows 11)

| command | before | after |
|---|---|---|
| `rtk eslint --version` | exit 127 | `v9.39.4`, exit 0 |
| `rtk vite --version` | exit 127 | `vite/6.4.3`, exit 0 |
| `npx eslint --version` (hook-rewritten → `rtk eslint`) | exit 127 | exit 0 |
| `rtk git status` / `rtk npm run` / `rtk cargo --version` | ok | ok (no regression) |

`cargo check` is clean on `develop`; the resolution unit tests pass (`cargo test resolve_binary`).

### DIFF
--- a/src/core/utils.rs
+++ b/src/core/utils.rs
@@ -8,8 +8,9 @@
 use anyhow::{Context, Result};
 use regex::Regex;
 use serde_json::Value;
+use std::ffi::OsString;
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::sync::LazyLock;
 use std::sync::OnceLock;
@@ -310,13 +311,19 @@ pub fn package_manager_exec(tool: &str) -> Command {
     }
 }
 
-/// Resolve a binary name to its full path, honoring PATHEXT on Windows.
+/// Resolve a binary name to its full path, searching project-local
+/// `node_modules/.bin` first, then the system PATH (PATHEXT-aware on Windows).
 ///
-/// On Windows, Node.js tools are installed as `.CMD`/`.BAT`/`.PS1` shims.
-/// Rust's `std::process::Command::new()` does NOT honor PATHEXT, so
-/// `Command::new("vitest")` fails even when `vitest.CMD` is on PATH.
+/// On Windows, Node.js tools are installed as `.CMD`/`.BAT`/`.PS1` shims, and
+/// project-local tools live in `./node_modules/.bin` — NOT on the system PATH.
+/// `npx`/`npm` add that directory to PATH at runtime; RTK strips the launcher
+/// during hook-rewrite (`npx eslint` → `rtk eslint`), so it must replicate the
+/// same resolution or every locally-installed tool fails with exit 127
+/// ("program not found"), even though `npx <tool>` would have run it fine.
 ///
-/// This function uses the `which` crate to perform proper PATH+PATHEXT resolution.
+/// Resolution order mirrors npx: nearest `node_modules/.bin` (cwd walking up to
+/// the filesystem root) first, then the inherited PATH. `which` honors PATHEXT,
+/// so the original bare-`Command::new` PATHEXT gap stays fixed too.
 ///
 /// # Arguments
 /// * `name` - Binary name (e.g., "vitest", "eslint", "tsc")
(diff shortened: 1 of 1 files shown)


danscMax:
Rebased onto current `develop` — this had drifted 316 commits behind and was showing as conflicting, which is on me. The only conflict was the `use` block in `src/core/utils.rs`; the change itself applies unchanged. `cargo clippy --all-targets -- -D warnings` is clean and the full suite passes (2522 tests) on the rebased branch.

To restate the concrete failure, since the PR has been open a while:

When an agent's Bash hook rewrites `npx <tool>` into `rtk <tool>`, the npx launcher is dropped. `resolve_binary` then looks only at `PATH`, so any tool that lives in the project's `node_modules/.bin` — eslint, vite, vitest and friends — is not found, and the command exits **127**. This makes `resolve_binary` mirror npx: nearest `node_modules/.bin` walking up from cwd first, then the inherited `PATH`, via `which::which_in` so `PATHEXT` is honoured on Windows. `tool_exists` uses the same resolution, so handlers that branch on it agree with what actually runs. Native tools (`git`, `cargo`) are unaffected and fall through to `PATH`; a project-local tool now correctly takes precedence over a globally installed one of the same name.

One note on scope, to avoid overselling it: #3263 and #2369 both report an exit-127 symptom, and neither is claimed as fixed here. #2369 is about unknown global flags, which is a different path entirely. #3263 may or may not share this root cause — it is a spawn failure at `stats` startup, and if whatever it spawns resolves through `resolve_binary` then this would cover it, but I have not reproduced that one and would rather not guess.

Happy to adjust the approach if you would prefer this folded into the existing `composer_bin_dirs` machinery in the same file — the two solve the same shape of problem for different ecosystems, and unifying them would be a reasonable follow-up, just a wider change than this fix needs.

KuSh:
Thanks for the detailed Windows repro. develop resolves project local JS tools by running them through the package runner instead of widening resolve_binary: #988 (merged 2026-09-03) added tool_exec and exec_runner, so a tool that is not on PATH is run through the runner you named or the detected one, and npx, pnpm exec and yarn exec are what search node_modules/.bin; lint, tsc, prettier, vitest, next, prisma and playwright all go through that path now. The PATHEXT handling your change preserves was already in place. Two cases develop still misses: a bare rtk tsc prefers a global tool on PATH over the project local copy, unlike npx, and a node_modules/.bin only tool with no rtk handler still fails to resolve, so an issue or a narrow PR on the tool_exec model would be welcome for either. Please comment or reopen if I missed something.

PR #2322, opened 2026-06-08
fix(core): stop capture hanging on orphaned pipes

### LINKED ISSUE #2320: rtk dotnet build/test/restore hangs forever when a long-lived child daemon (NuGet auth plugin / MSBuild node) inherits the captured stdout/stderr pipe
## `rtk dotnet build`/`test`/`restore` hangs forever when a long-lived child daemon inherits the captured stdout/stderr pipe (e.g. NuGet auth plugin / MSBuild node reuse)

> **Note:** This issue was investigated and written with the help of Claude Code (it inspected rtk's source and a live hung process). I don't have any expertise on this repo or Rust — I just hit the bug and am trying to help with a useful report. Please take the suggested fixes as pointers, not prescriptions.

## Summary

`rtk dotnet build <proj>` never returns on a project that restores from an **authenticated NuGet feed** (e.g. Azure DevOps Artifacts), even though the build itself completes successfully in seconds. Plain `dotnet build` and `rtk run "dotnet build ..."` both work fine.

The hang is **after** the build finishes. Root cause: the dotnet wrapper captures output via `Command::output()` (`exec_capture`), which **reads the child's stdout/stderr pipes until EOF**. A long-lived grandchild daemon — the **NuGet Azure Artifacts Credential Provider** running in persistent `-Plugin` mode — inherits the write end of the **stderr** pipe and outlives the build. The build's direct `dotnet` process exits (becomes a zombie), but the pipe never reaches EOF because the daemon still holds it open, so `Command::output()` blocks forever in `read()`.

This is a general bug: **`Command::output()` blocks on pipe EOF, not on the direct child exiting.** Any descendant that inherits the captured fds and lingers (NuGet credential provider plugin, MSBuild **node reuse** worker nodes, the MSBuild build server, VBCSCompiler/Razor build server) can trigger it.

## Environment

- rtk **0.42.3** (also present in 0.42.2 source)
- .NET SDK **10.0.300**
- Linux (Debian 13, kernel 6.12)
- Project restores from a private Azure DevOps Artifacts feed → triggers `CredentialProvider.Microsoft.dll -Plugin`

## Reproduction

```bash
# Hangs forever (build finishes in ~35s, rtk never returns):
rtk dotnet build path/to/SomeProject.csproj

# Works (passthrough, inherits the terminal, no captured pipe):
rtk run "dotnet build path/to/SomeProject.csproj"

# Works (TTY, no pipe):
dotnet build path/to/SomeProject.csproj
```

The trigger is having an authenticated NuGet feed configured (so `dotnet restore` launches the credential-provider plugin). A non-trivial project that restores from such a feed is enough.

## Evidence (live inspection of the hung process)

**1. The build finishes — the binlog stops growing — but rtk keeps running:**

```
[10s] rtk_up=yes dotnet=19 binlog size=913365
[20s] rtk_up=yes dotnet=19 binlog size=1159802
[30s] rtk_up=yes dotnet=19 binlog size=1382604
[40s] rtk_up=yes dotnet=19 binlog size=2104543   <- build done, binlog finalized
[50s] rtk_up=yes dotnet=19 binlog size=2104543
 ...  rtk still running, size frozen, indefinitely
```

**2. rtk's main thread is blocked reading the stderr pipe:**

```
$ ps -L -o tid,stat,wchan,comm -p <rtk>
   TID STAT WCHAN       COMMAND
100308 S    pipe_read   rtk

$ ls -l /proc/<rtk>/fd | grep pipe
4 -> pipe:[544004]
6 -> pipe:[544005]
```

**3. The build's direct child has already exited (zombie), yet the stderr pipe write-end is held open by the NuGet credential-provider daemon:**

```
# rtk's direct child = zombie -> build process already exited
pid=100311 ppid=<rtk> stat=Z  [dotnet] <defunct>

# Who still holds rtk's stderr pipe (inode 544005) open?
pid 100596 (dotnet):
    fd 2 -> pipe:[544005]
    fd 7 -> pipe:[544005]

$ ps -o args= -p 100596
/usr/share/dotnet/dotnet \
  /home/<user>/.nuget/plugins/netcore/CredentialProvider.Microsoft/CredentialProvider.Microsoft.dll -Plugin
```

The credential provider is reparented away from rtk and lives on as a persistent plugin daemon, keeping the stderr pipe open → EOF never arrives → `Command::output()` blocks in `read()` forever.

## Root cause in code

`src/cmds/dotnet/dotnet_cmd.rs` — `run_build`/`run_test`/`run_restore` → `run_dotnet_with_binlog` → `exec_capture(&mut cmd)`.

`src/core/stream.rs`:

```rust
pub fn exec_capture(cmd: &mut Command) -> Result<CaptureResult> {
    cmd.stdin(Stdio::null());
    let output = cmd.output().context("Failed to execute command")?; // <-- reads stdout+stderr to EOF
    ...
}
```

`Command::output()` waits for **EOF on both pipes**, which only happens once *every* process holding a write end closes it — not when the direct child exits. A lingering grandchild daemon that inherited the fd makes EOF unreachable.

Note `run_streaming`'s non-passthrough paths have the same structural issue: they `join()` the stdout/stderr reader threads (which loop until EOF) and only then `child.wait()`. Passthrough mode is safe precisely because it uses `Stdio::inherit()` and never creates a captured pipe.

## Why the workarounds work

- **`dotnet build` (no rtk):** stdout is a TTY; no pipe to keep open. The daemon shares the TTY; the shell returns when the foreground process exits.
- **`rtk run "dotnet build ..."`:** Passthrough = `Stdio::inherit()` (`run_streaming`, `src/core/stream.rs`), so there is no captured pipe to block on.

## Suggested fixes

The capture must key off **the direct child exiting**, not pipe EOF. Some options:

1. **Stop waiting on EOF once the direct child has exited.** Read stdout/stderr on threads, but after `child.wait()` returns, do a final *non-blocking* drain of whatever is buffered and return — don't `join()` reader threads that may block forever. (A short grace period after child exit is enough to flush legitimate trailing output.)
2. **Isolate descendants from the captured fds.** Spawn the child in its own process group / session and, after it exits, stop reading. Or ensure the captured pipes are not inherited by grandchildren (where feasible).
3. Combination: child-exit-driven completion + bounded post-exit drain. This is robust against *all* lingering daemons (NuGet credential provider, MSBuild node reuse, build server, VBCSCompiler/Razor server), not just this one.

A targeted mitigation (passing `-nodeReuse:false`, `DOTNET_CLI_USE_MSBUILD_SERVER=0`, etc.) does **not** fix this case, because the daemon here is the NuGet credential-provider plugin, which is independent of MSBuild node reuse. The real fix is to not block on pipe EOF.

fengjikui:
Fixes #2320

## Summary
- replace `Command::output()` in `exec_capture` with explicit spawn/wait plus background stdout/stderr readers
- return after the direct child exits once captured output is idle, with a bounded post-exit drain
- add a regression test for a child that exits while a background descendant still holds the captured pipes open

## Why
Some `rtk dotnet ...` commands can start helper descendants such as credential providers or MSBuild nodes. If those descendants inherit captured stdout/stderr, `Command::output()` waits for pipe EOF even after the direct `dotnet` child has exited, so RTK can appear to hang indefinitely. Waiting on the direct child separately avoids that EOF dependency while still preserving captured output from the command.

## Validation
- `cargo fmt --check`
- `cargo test core::stream::tests::test_exec_capture_returns_after_child_exits_when_grandchild_holds_pipe -- --nocapture`
- `cargo test core::stream::tests::test_exec_capture -- --nocapture`
- `cargo test core::stream -- --nocapture`
- `cargo test`
- `git diff --check`

### DIFF
--- a/src/core/stream.rs
+++ b/src/core/stream.rs
@@ -2,6 +2,8 @@ use anyhow::{Context, Result};
 use std::io::{self, BufRead, BufReader, BufWriter, Write};
 use std::process::{Command, Stdio};
 use std::sync::mpsc;
+use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
 
 #[cfg(test)]
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


pidefrem:
The "wait on the direct child, not pipe EOF" approach is exactly right. Heads-up that there's a **second capture path with the identical bug** this PR doesn't touch: `exec_capture` is fixed here, but **`run_streaming` has the same EOF-blocking defect**.

`run_streaming` drains the child's stdout/stderr to EOF before it returns, so when the direct child exits and a descendant has inherited the captured pipe, EOF never arrives — the same failure mode as #2320, just in the other function. This path is used by everything routed through the filtered/capture-only runner (long-running filtered commands), so the hang persists there even after this PR merges.

**Repro (the #2320 topology — direct child exits, grandchild lingers):** a tool whose direct child exits while a daemon it spawned keeps the captured pipe open — e.g. dotnet's MSBuild node / NuGet auth-plugin daemon (the #2320 case), or in a minimal shell form:

```
rtk run sh -c '(sleep 5) & echo done; exit 0'
# direct child (sh) exits immediately; the `sleep 5` grandchild holds the pipe →
# run_streaming blocks ~5s (or forever with a real long-lived daemon) instead of
# returning `done` + exit 0 promptly.
```

The fix that worked for me mirrors this PR's pattern on `run_streaming`: read each pipe on a background thread into a shared buffer, `wait()` the direct child, then bound the post-exit drain — stop when both readers EOF, when output goes idle for a short grace window, or at a hard cap — and *detach* (don't `join`) a reader still blocked on a grandchild-held pipe. Constants/naming kept parallel to this PR so it reads as the sibling fix.

I have it implemented and tested on my fork — commit [`1d4409a`](https://github.com/pidefrem/rtk/commit/1d4409a7de7ab4275354fc56dc601fcf3272d03e) on branch [`fix/stream-pipe-eof-drain`](https://github.com/pidefrem/rtk/tree/fix/stream-pipe-eof-drain), with a regression test (child exits while a grandchild holds the pipe). Would you be open to extending this PR to cover `run_streaming` too, so both capture paths are fixed together? Happy to supply the streaming-branch diff or open it as a PR against your branch.

fengjikui:
Thanks for the detailed pointer. I verified the sibling `run_streaming` path had the same EOF-blocking shape and extended this PR to cover it too.

Updated branch now includes:
- bounded post-exit drain for `run_streaming` CaptureOnly/Buffered path
- bounded idle handling for the live Streaming path
- regression coverage for both `exec_capture` and `run_streaming` when a grandchild keeps the pipe open

Local validation:
- `cargo fmt --check`
- `cargo test core::stream::tests` (41 passed)

pidefrem:
Thanks for extending this to cover `run_streaming` as well — both capture paths fixed together is exactly the right shape.

One heads-up: this PR is currently showing **conflicting** against `develop`. The branch head is a bit behind `develop` now — in particular it predates the grep-faithful-engine rework (#2641), which renamed `src/cmds/system/grep_cmd.rs` → `search.rs` and reshaped a few neighbouring files, so a straight rebase will hit conflicts that aren't really about the capture fix.

To make the rebase easier, I re-applied the same fix (both `exec_capture` and `run_streaming`, bounded post-exit drain) on top of **current `develop`** on my fork — it's conflict-free there and the stream suite is green:

- branch: `adopt/fengjikui/fix/stream-pipe-eof-drain` on `pidefrem/rtk`
- the capture-path commits are isolated to `src/core/stream.rs`, so they lift cleanly onto the post-#2641 tree.

Feel free to cherry-pick or diff against it as a reference for the rebase — whatever's least work for you. Happy to open it as a separate compare link if that helps.

fengjikui:
Rebased this branch onto current `develop` and resolved the conflict in `src/core/stream.rs` using the same bounded-drain shape.

Current head: `b4742f320be9c379e05fa0ba0fc3787dc0242c5c`

Local validation:
- `cargo fmt --check`
- `cargo test core::stream::tests` (53 passed)

The PR is back to mergeable on GitHub; CLA remains green.

PR #2324, opened 2026-06-08
fix(analytics): truncate gain history on char boundaries

### LINKED ISSUE #2318: rtk gain --history panics on non-ASCII (CJK) filenames in command history
## Bug Description

`rtk gain --history` panics when a recorded command contains CJK (Chinese/Japanese/Korean) characters in filenames. The truncation logic in `gain.rs:250` uses a byte index that falls in the middle of a multi-byte UTF-8 character.

## Steps to Reproduce

1. Run any rtk command that includes CJK filenames, e.g.:
   ```bash
   rtk ls -la 停止小工具.command 启动小工具.command 内存监视.app
   ```
2. Then run:
   ```bash
   rtk gain --history
   ```

## Error Output

```
end byte index 22 is not a char boundary; it is inside '工' (bytes 20..23 of string)
```

## Root Cause

The display truncation at `gain.rs:250` appears to use a fixed byte offset (likely `&s[..22]`) instead of a character-boundary-aware truncation. For ASCII this works fine, but CJK characters are 3 bytes each in UTF-8, so byte index 22 lands in the middle of '工' (U+5DE5, bytes 20..23).

## Suggested Fix

Use `s.char_indices()` or `s.floor_char_boundary(max_len)` (stable since Rust 1.82) to find the nearest valid character boundary before truncating.

## Environment

- rtk version: 0.42.3
- OS: macOS 15.5 (Darwin 25.5.0), arm64
- Locale: zh_CN.UTF-8

fengjikui:
Fixes #2318.

## Summary

`rtk gain --history` truncated long command strings with a fixed byte slice (`&cmd[..22]`). When a command contained CJK filenames, byte 22 could land inside a multi-byte UTF-8 character and panic.

This reuses the existing character-count-based `core::utils::truncate` helper for gain history command display and adds a regression test with the CJK filename shape from the issue.

## Validation

- `cargo fmt --check`
- `cargo test analytics::gain::tests::truncate_history_command_handles_cjk_filenames`
- `cargo test analytics::gain`
- `cargo test` -> 2073 passed, 7 ignored
- `git diff --check`

### DIFF
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -2,7 +2,7 @@
 
 use crate::core::display_helpers::{format_duration, print_period_table};
 use crate::core::tracking::{DayStats, MonthStats, Tracker, WeekStats};
-use crate::core::utils::format_tokens;
+use crate::core::utils::{format_tokens, truncate};
 use crate::hooks::hook_check;
 use anyhow::{Context, Result};
 use chrono::Local;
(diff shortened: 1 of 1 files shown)


fengjikui:
Closing as superseded by #2789, merged in `47b22e0`. That upstream change now uses the shared Unicode-safe `truncate()` in `gain --history` and adds CJK, Thai, emoji, and Cyrillic regression coverage, so this branch no longer carries an independent fix. Thanks!

PR #2325, opened 2026-06-08
fix(tee): quote recovery hint paths with spaces

### LINKED ISSUE #1858: ClaudeCode keeps asking for permissions to read rtk logs in `~/Library/Application Support/` because of backslash in path
I keep getting the following when claude code tries to read Go test logs under `~/Library/Application\ Support/rtk/tee`

```
 Bash command

   rtk read ~/Library/Application\ Support/rtk/tee/1778602911_go_test.log | grep -E "PASS|FAIL|Error|deadline|panic" | head -20
   Check test failure details

 Contains backslash-escaped whitespace

 Do you want to proceed?
 ❯ 1. Yes
   2. No
```

fengjikui:
## Summary

Fixes #1858. RTK recovery hints can point at macOS tee files under `~/Library/Application Support/...`. When that path is copied into a shell command, agents tend to use backslash-escaped whitespace (`Application\ Support`), which makes Claude Code ask for permission again.

This keeps simple hint paths unchanged, but renders paths that need shell quoting as copy-paste-safe double-quoted paths. For home-relative paths with spaces, it uses `"$HOME/..."` instead of quoting `~`, because `~` does not expand inside quotes.

Related to #2187, but this keeps the compact `~/...` display for simple paths and directly quotes paths with spaces instead of only rendering an absolute path.

## Validation

- `cargo fmt --check`
- `cargo test core::tee::tests`
- `cargo test`

### DIFF
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -183,6 +183,7 @@ fn needs_shell_quoting(path: &str) -> bool {
             || matches!(
                 c,
                 '\'' | '"'
+                    | '\\'
                     | '$'
                     | '`'
                     | '!'
@@ -510,6 +511,15 @@ mod tests {
         assert_eq!(
             display_shell_path(&path),
             "\"/tmp/rtk/Application Support/123_go_test.log\""
+        );
+    }
+
+    #[test]
+    fn test_display_shell_path_quotes_backslashes() {
+        let path = PathBuf::from(r"/tmp/rtk/tee/path\segment.log");
+        assert_eq!(
+            display_shell_path(&path),
+            r#""/tmp/rtk/tee/path\\segment.log""#
         );
     }
 
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -223,7 +223,8 @@ fn display_shell_path(path: &std::path::Path) -> String {
     }
 
     if let Some(relative) = display.strip_prefix("~/") {
-        return format!("\"$HOME/{}\"", escape_double_quoted_path(relative));
+        let relative = relative.replace(std::path::MAIN_SEPARATOR, "/");
+        return format!("\"$HOME/{}\"", escape_double_quoted_path(&relative));
     }
 
     format!("\"{}\"", escape_double_quoted_path(&display))
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -177,8 +177,59 @@ fn display_path(path: &std::path::Path) -> String {
     path.display().to_string()
 }
 
+fn needs_shell_quoting(path: &str) -> bool {
+    path.chars().any(|c| {
+        c.is_whitespace()
+            || matches!(
+                c,
+                '\'' | '"'
+                    | '$'
+                    | '`'
+                    | '!'
+                    | '#'
+                    | '&'
+                    | '('
+                    | ')'
+                    | ';'
+                    | '<'
+                    | '>'
+                    | '?'
+                    | '['
+                    | ']'
+                    | '{'
+                    | '}'
+                    | '|'
+                    | '*'
+            )
+    })
+}
+
+fn escape_double_quoted_path(path: &str) -> String {
+    let mut escaped = String::with_capacity(path.len());
+    for c in path.chars() {
+        if matches!(c, '\\' | '"' | '$' | '`') {
+            escaped.push('\\');
+        }
+        escaped.push(c);
+    }
+    escaped
+}
+
+fn display_shell_path(path: &std::path::Path) -> String {
+    let display = display_path(path);
+    if !needs_shell_quoting(&display) {
+        return display;
+    }
+
+    if let Some(relative) = display.strip_prefix("~/") {
+        return format!("\"$HOME/{}\"", escape_double_quoted_path(relative));
+    }
+
+    format!("\"{}\"", escape_double_quoted_path(&display))
+}
+
 fn format_hint(path: &std::path::Path) -> String {
-    format!("[full output: {}]", display_path(path))
+    format!("[full output: {}]", display_shell_path(path))
 }
 
 /// Convenience: tee + format hint in one call.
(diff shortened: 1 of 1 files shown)


pszymkowiak (dismissed):
Approved after review + real-binary verification (built `origin/develop` + this PR, exercised the compiled release binary on macOS).

**Correctness — the fix is at the right layer.** Both hint formatters (`format_hint` → `[full output: …]` and `force_tee_tail_hint` → `[see remaining: tail …]`) route through the new `display_shell_path`. I grepped every hint emitter in the tree (go, uv, cargo, dotnet, ruff, eslint/pylint, pnpm, aws, glab, docker/kubectl, git, gh, grep, curl, rubocop) plus the TOML engine's `tee_payload`/`tail_offset` consumer in `main.rs` — all of them go through these two functions, so patching the choke point covers 100% of cases. Uses `"$HOME/…"` rather than a quoted `~` (correct — `~` doesn't expand inside quotes), and escapes `\ " $ `` — the right set for a double-quoted context.

**Verified end-to-end on the compiled binary (this was the actual bug):**
- Before: `[see remaining: tail -n +26 ~/Library/Application Support/rtk/tee/…]` → pasted literally, `tail: …/Application: No such file`.
- After: `[see remaining: tail -n +26 "$HOME/Library/Application Support/rtk/tee/…"]` → literal copy-paste recovers the omitted lines.
- Confirmed on a non-uv path too (`rtk ruff check` with 320 violations → 270 lines recovered) and on a failure-path `[full output: …]` hint (100/100 pre-crash stdout lines recovered). So the quoting is universal, not uv-specific.

**No regressions:** merges cleanly on current `develop`; full suite 2559 passed, 0 failed; `clippy --all-targets` clean; the 4 added unit tests cover simple-path-unchanged, absolute-with-space, home→`$HOME`, and the no-backslash-escaping assertion.

**Minor (non-blocking):** `[full output: "$HOME/…"]` is more informational than a runnable command, so an agent that opens the path directly (not via a shell) now has to strip the quotes and expand `$HOME`. The shell copy-paste case (the subject of #1858) dominates, so this is the right trade-off.

Small, correct, well-tested fix that makes recovery hints actually usable on macOS. LGTM.

pszymkowiak (approved):
Re-approving after a deeper adversarial pass (built origin/develop + this PR, exercised the compiled binary on macOS).

**Security — command injection neutralized (verified, not just by inspection).** With a booby-trapped tee dir `/tmp/a $(touch /tmp/RTK_PWN) b`, the emitted hint escapes the substitution:
`[see remaining: tail -n +26 "/tmp/a \$(touch /tmp/RTK_PWN) b/…log"]`. Pasting it literally recovers the file (65 lines) and does **not** execute the injected command. The `\ " $ `` escape set does its job under attack.

**Coverage — the fix is at the single choke point.** Both hint formatters (`format_hint`, `force_tee_tail_hint`) route through `display_shell_path`; a full grep of every hint emitter (go, uv, cargo, dotnet, ruff, eslint/pylint, pnpm, aws, glab, docker/kubectl, git, gh, grep, curl, rubocop) plus the TOML engine's consumer in `main.rs` confirms 100% go through these two. No code parses the hint back, so changing `~/path` → `"$HOME/path"` is integration-safe.

**Realistic case works end-to-end.** Default macOS dir (`~/Library/Application Support`, has a space) → `"$HOME/…"`, literal copy-paste recovers the omitted lines. Verified on a non-uv path too (`rtk ruff check`, 270 lines recovered) and on a failure-path `[full output: …]` hint (100/100 pre-crash lines).

**No regressions:** merges clean on current develop; full suite 2559 passed / 0 failed; clippy clean.

**One non-blocking suggestion:** the $/"/backtick escaping (the security-relevant part) has no direct unit test — `escape_double_quoted_path` is only covered indirectly. A small test feeding a path with `$(…)`/`\"`/backtick would lock in the property. Nice-to-have, not a blocker.

Investigated a suspected 'long path suppresses the hint' issue and it turned out to be a non-bug — that's just `never_worse` correctly showing full raw when truncation isn't a net token win (no data loss; large outputs keep the hint regardless of path length).

Correct, secure, well-scoped. LGTM.

src/core/tee.rs
```diff
@@ -177,8 +177,59 @@ fn display_path(path: &std::path::Path) -> String {
     path.display().to_string()
 }
 
+fn needs_shell_quoting(path: &str) -> bool {
+    path.chars().any(|c| {
+        c.is_whitespace()
+            || matches!(
+                c,
+                '\'' | '"'
```
TaKO8Ki, line 185: This condition needs `\\` backslash, I guess.

I think If rtk receives a path containing only `\` as the special character, quoting never happens.

TaKO8Ki (changes requested):
Thank you. I left a comment.

src/core/tee.rs
fengjikui: Good catch — fixed in `991ed97`. `needs_shell_quoting` now treats `\` as requiring quotes, and I added a direct regression test showing that an input path containing a lone backslash is emitted inside double quotes with the backslash escaped for literal shell copy-paste. Validation: `cargo test core::tee::tests::` (25 passed), `cargo test` (2077 passed, 7 ignored), and `cargo clippy --all-targets -- -D warnings` (clean).

TaKO8Ki (dismissed):

TaKO8Ki (changes requested):
The CI failed on Windows. Could you check it?

fengjikui:
Checked and fixed in `5de188b`. The two failing tests used home-relative paths built with `PathBuf::join`; on Windows, `display()` emitted `\\` separators, so the quoted `$HOME/...` hints escaped them as literal backslashes. The home-relative shell branch now normalizes the platform separator to `/` before quoting, while non-home paths and literal backslashes keep their existing escaping behavior. Local validation: tee tests (25 passed), full suite (2077 passed, 7 ignored), clippy with `-D warnings`, and fmt all pass. The pushed CI run will provide the Windows-target confirmation.

TaKO8Ki (approved):
Thanks!

PR #2327, opened 2026-06-08
fix(hook): rewrite Windows python module invocations

### LINKED ISSUE #2326: hook: python.exe and pathed venv interpreters bypass rewrite (Windows)
## Summary

On Windows, the `rtk hook claude` PreToolUse rewriter does **not** rewrite Python invocations when the interpreter is named `python.exe` or invoked via an absolute/relative venv path. The command passes through unrewritten, so full pytest/unittest output reaches the LLM context (lost savings).

## Environment

- OS: Windows 11 Pro (10.0.26200)
- Shell: PowerShell + Bash
- Hook: `PreToolUse` matcher `Bash|PowerShell` -> `rtk hook claude`

## Reproduction

Feed JSON tool-input to the hook on stdin:

```bash
# WORKS — rewritten to `rtk pytest`
printf '{"tool_name":"Bash","tool_input":{"command":"python -m pytest tests/"}}'  | rtk hook claude
printf '{"tool_name":"Bash","tool_input":{"command":"python3 -m pytest tests/"}}' | rtk hook claude

# BUG — empty output, no rewrite (passthrough)
printf '{"tool_name":"Bash","tool_input":{"command":"python.exe -m pytest tests/"}}'              | rtk hook claude
printf '{"tool_name":"Bash","tool_input":{"command":".venv/Scripts/python.exe -m pytest tests/"}}' | rtk hook claude
```

### Expected

All four rewrite the `-m pytest` invocation to `rtk pytest` (and `-m <module>` generally to the matching rtk filter), regardless of:
1. `.exe` suffix on the interpreter name
2. an absolute or relative path prefix (e.g. `.venv/Scripts/python.exe`, `C:\...\python.exe`)

### Actual

Only bare `python` / `python3` match. `python.exe` and any pathed interpreter pass through with no rewrite.

## Impact

On Windows, venv pytest runs are almost always invoked as `.venv/Scripts/python.exe -m pytest` — the canonical form. `rtk discover` flagged 122 `python.exe` + 116 pathed-`python` invocations in a 30-day window (~29K tokens of pytest output that should have been filtered by `rtk pytest`).

## Suggested fix

In the interpreter-matching logic, normalize the leading token before matching: strip any directory prefix (basename) and strip a trailing `.exe` (case-insensitive) so `python.exe`, `python3.exe`, `.venv/Scripts/python.exe`, and `C:\path\python.exe` all map to the same `python` handler. Same normalization likely applies to `node.exe`, `npm.cmd`, `npx.cmd` on Windows.

## Secondary note (separate, lower priority)

`rtk discover` without `-a` defaults to current-project-only and matches 0 sessions when the project path contains spaces / OneDrive-style segments (e.g. `C:\Users\...\OneDrive - ORG\...`), reporting "No missed savings found". `-a` works. The cwd->project-key canonicalization likely does not match the encoded `.claude/projects/<key>` directory name for such paths.

fengjikui:
Fixes #2326.

## Summary

Windows Python invocations often use `python.exe` or a virtualenv interpreter path such as `.venv/Scripts/python.exe -m pytest`. The rewrite registry only recognized bare `python` / `python3`, so these canonical Windows forms bypassed the existing `rtk pytest` and `rtk mypy` rules.

This PR normalizes supported Python `-m` invocations before classification and rewrite matching:

- tokenizes the invocation with the existing shell-aware lexer
- accepts `python`, `python3`, and versioned Python basenames, with a case-insensitive `.exe` suffix
- supports Unix and Windows path separators plus quoted interpreter paths containing spaces
- limits normalization to the supported `pytest` and `mypy` modules
- checks exclusions against both the original and normalized forms
- preserves existing environment/wrapper handling and trailing redirects

The resulting command keeps the existing RTK shape, for example `python.exe -m pytest tests/` becomes `rtk pytest tests/`.

## Validation

- `cargo test python_exe` -> 4 passed
- `cargo test discover::` -> 459 passed
- `cargo test --quiet` -> 2439 passed, 8 ignored; all 58 integration tests passed
- `cargo fmt --check`
- `git diff --check`
- CLI smoke: `cargo run --quiet -- rewrite 'python.exe -m pytest tests/'` -> `rtk pytest tests/` (exit 3, the existing "rewrite available" hook protocol)

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -3,6 +3,7 @@
 use crate::core::utils::composer_bin_dirs;
 use lazy_static::lazy_static;
 use regex::{Regex, RegexSet};
+use std::borrow::Cow;
 use std::path::Path;
 
 use super::lexer::{split_on_operators, tokenize, TokenKind};
@@ -122,6 +123,8 @@ pub fn classify_command(cmd: &str) -> Classification {
 
     // Normalize absolute binary paths: /usr/bin/grep → grep (#485)
     let cmd_normalized = strip_absolute_path(cmd_clean);
+    // Normalize Python module runners before matching pytest/mypy rules.
+    let cmd_normalized = normalize_python_module_invocation(&cmd_normalized);
     // Strip git global options: git -C /tmp status → git status (#163)
     let cmd_normalized = strip_git_global_opts(&cmd_normalized);
     // Normalize PHP tool paths: vendor/bin/phpunit, bin/phpunit, or composer
@@ -878,12 +881,15 @@ fn rewrite_segment_inner(
         }
     }
 
+    let match_cmd = normalize_python_module_invocation(cmd_part);
+    let match_cmd = match_cmd.as_ref();
+
     // Use classify_command for correct ignore/prefix handling
-    let rtk_equivalent = match classify_command(cmd_part) {
+    let rtk_equivalent = match classify_command(match_cmd) {
         Classification::Supported { rtk_equivalent, .. } => {
-            let stripped = ENV_PREFIX.replace(cmd_part, "");
+            let stripped = ENV_PREFIX.replace(match_cmd, "");
             let cmd_clean = stripped.trim();
-            if is_excluded(cmd_clean, excluded) {
+            if is_excluded(cmd_clean, excluded) || is_excluded(cmd_part, excluded) {
                 return None;
             }
             rtk_equivalent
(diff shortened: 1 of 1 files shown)


fengjikui:
Refreshed this branch onto current `develop` (`5d32d07`) and rebuilt the fix around the existing shell lexer. New head: `f803f6a`. It now covers quoted/backslash virtualenv paths and versioned Python executables, while limiting normalization to `pytest`/`mypy` and preserving original-form exclusions. Validation: 459 discover tests; full suite 2439 passed / 8 ignored plus 58 integration tests; Clippy with `-D warnings`; fmt and diff checks.

Issue #880, opened 2026-03-27
RTK parse failures in grep


pliablepixels:
OS: MacOS
rtk version: 0.34.0

I recently did a code review with RTK enabled and noticed this output:

```
───────────────────────────────────────────────────────────────────────
  #  Command                   Count  Saved    Avg%    Time  Impact
───────────────────────────────────────────────────────────────────────
 1.  rtk ls                       49   5.9K   67.5%     2ms  ██████████
 2.  rtk grep                      4    256   28.7%     2ms  ░░░░░░░░░░
 3.  rtk git status                2    132   76.7%    28ms  ░░░░░░░░░░
 4.  rtk find                      2     23   46.0%    10ms  ░░░░░░░░░░
 5.  rtk fallback: grep -r...      1      0    0.0%    58ms  ░░░░░░░░░░
 6.  rtk fallback: grep -r...      1      0    0.0%    26ms  ░░░░░░░░░░
 7.  rtk fallback: grep -r...      1      0    0.0%    26ms  ░░░░░░░░░░
 8.  rtk fallback: grep -r...      1      0    0.0%    29ms  ░░░░░░░░░░
 9.  rtk fallback: grep -r...      1      0    0.0%    28ms  ░░░░░░░░░░
10.  rtk fallback: grep -r...      1      0    0.0%    21ms  ░░░░░░░░░░
```

Looks like a bunch of greps failed. Digging futher:

```
[pp@mbp:~/fiddle/zmNinjaNG]$ rtk gain -F
RTK Parse Failures
════════════════════════════════════════════════════════════

Total failures:    65
Recovery rate:     100.0%

Top Commands (by frequency)
────────────────────────────────────────────────────────────
     2x  grep -rn dangerouslySetInnerHTML /Users/pp/f...
     1x  grep -rn zmng_crypto_salt_v1\|getPersistentSalt...
     1x  grep -rn window\.open /Users/pp/fiddle/zmNin...
     1x  grep -rn window\.location\s*= /Users/pp/fidd...
     1x  grep -rn window\.location\.href\s*= /Users/pp..
     1x  grep -rn window\.location /Users/pp/fiddle/z...
     1x  grep -rn truncate --include=*.tsx
     1x  grep -rn token= /Users/pp/fiddle/zmNinjaNG/a...
     1x  grep -rn token /Users/pp/fiddle/zmNinjaNG/ap...
     1x  grep -rn toast\.\(error\|success\|warning\|info...

[...]

PR #2333, opened 2026-06-08
fix(grep): stabilize argument parsing — trailing_var_arg, -v invert-match, --version passthrough, safe rg invocation

KuSh:
## Summary
 `rtk grep` had three distinct bugs that made it unreliable in real usage:

 1. **Clap parse failure for any unknown flag before the pattern** — `rtk grep --version`, `rtk grep --glob '*.ts' pattern`, `rtk grep --type-add 'web:*.html' pattern` all failed with
 "unexpected argument" Clap errors because `pattern` was a required positional and Clap tried to interpret any flag before it as a named subcommand flag.

 2. **`-v` verbose flag silently swallowing grep's invert-match** — `rtk grep -v pattern .` ran a normal (non-inverted) search because `verbose` was declared `global = true`, causing Clap
 to consume every `-v` anywhere in the arg list as RTK verbosity.

 3. **rg misinterpreting flag-shaped patterns** — `rtk grep -- --version` called `rg ... "--version" .` without a `--` separator before the pattern. rg interpreted `--version` as its own
 version flag, printed its version info (not in NUL match format), causing "7 matches in 0 files" output (total from pre-parse line count, zero from empty `by_file` map).

### Changes

 **`src/main.rs`**

 - Removed `global = true` from the `verbose` flag. `-v` is now only recognized before the subcommand; any `-v` after `grep` flows into `args` and is forwarded to rg as invert-match.
 - Removed `pattern`, `path`, and `line_numbers` as Clap-managed fields from the `Grep` variant. A single `trailing_var_arg` field `args: Vec<String>` captures everything — pattern, path,
 and any grep/rg flag — without Clap needing to know about them. New rg flags no longer require a `main.rs` change.

 **`src/cmds/system/grep_cmd.rs`**

 - Added `extract_pattern_path(args)` — scans the flat arg slice, assigns the first and second non-flag tokens as pattern and path (default `.`), collects the rest as `extra_args` for rg.
 Handles `--` boundary correctly.
 - Added early passthrough for `--version` / `--help` / `-h` — routes directly to rg (or grep fallback) without filtering.
 - Added `--` before `<pattern> <path>` in both rg and grep invocations — prevents rg/grep from misinterpreting flag-shaped patterns.
 - Fixed `total_matches` — derived from `by_file.values().map(|v| v.len()).sum()` after parsing, so the "N matches in M files" header is always consistent with what's displayed.

 ### Issues fixed

 | Issue | Title |
 |-------|-------|
 | #2167 | rtk grep calls system grep — breaks `-g`/`--glob` flags |
 | #2120 | `--fixed-strings` before pattern causes fallback to native grep |
 | #880  | 65 grep failures, 0 savings (invert-match `-v` swallowed) |

### Test plan
- [x] Manual testing:
  -`rtk grep --version` ✅️
  -`rtk -vvv grep --version` ✅️
  -`rtk grep -- --version` ✅️
  -`rtk grep -v TODO src/` ✅️
  -`rtk grep 'fn run' src/` ✅️
  -`rtk grep -i -A 2 'error' src/` ✅️

### DIFF
--- a/docs/contributing/ARCHITECTURE.md
+++ b/docs/contributing/ARCHITECTURE.md
@@ -746,7 +746,7 @@ Single-threaded execution with `Mutex<Option<Tracker>>` for future-proofing. No 
 └────────────────────────────────────────────────────────────────────────┘
 
 main.rs:47-49
-#[arg(short, long, action = clap::ArgAction::Count, global = true)]
+#[arg(short, long, action = clap::ArgAction::Count)]
 verbose: u8,
 
 Levels:
@@ -773,7 +773,7 @@ if verbose > 0 {
 └────────────────────────────────────────────────────────────────────────┘
 
 main.rs:51-53
-#[arg(short = 'u', long, global = true)]
+#[arg(long, global = true)]
 ultra_compact: bool,
 
 Features:
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -14,26 +14,28 @@ use std::collections::HashMap;
 /// for a missing entry is a visible wrong result, not a silent corruption.
 const VALUE_FLAGS_SHORT: &[u8] = b"ABCgfjm";
 
-/// Normalise a flag arg for rg forwarding.
+/// Normalise short flags arg for rg forwarding.
+/// Returns `None` when the entire flag is stripped (grep-ism recursive flags).
+fn strip_r(arg: &str) -> Option<String> {
+    match arg
+        .chars()
+        .filter(|&c| c != 'r' && c != 'R')
+        .collect::<String>()
+    {
+        s if !s.is_empty() => Some(s),
+        _ => None,
+    }
+}
+
+/// Normalise long flag arg for rg forwarding.
 /// Returns `None` when the flag should be dropped (grep-ism recursive flags).
-fn process_flag(arg: &str) -> Option<String> {
-    if arg == "--recursive" {
-        return None;
+fn strip_recursive(arg: &str) -> Option<String> {
+    match arg {
+        // Drop recursive flags that would change semantics in rg.
+        "--recursive" => None,
+        // Everything else pass through unchanged.
+        _ => Some(arg.to_string()),
     }
-    // Long flags pass through unchanged.
-    if arg.starts_with("--") {
-        return Some(arg.to_string());
-    }
-    // Short clusters: strip `r`/`R` (grep recursive; rg -r means --replace).
-    if let Some(rest) = arg.strip_prefix('-') {
-        let cleaned: String = rest.chars().filter(|&c| c != 'r' && c != 'R').collect();
-        return if cleaned.is_empty() {
-            None
-        } else {
-            Some(format!("-{}", cleaned))
-        };
-    }
-    Some(arg.to_string())
 }
 
 /// Extracts `(patterns, paths, flags)` from the raw trailing args.
@@ -48,7 +50,7 @@ fn process_flag(arg: &str) -> Option<String> {
 /// as their value so it is not mistaken for the pattern. Combined clusters like
 /// `-rn` have `r`/`R` stripped before forwarding. `--` marks everything after
 /// it as positional even if flag-shaped.
-fn extract_pattern_path(args: &[String]) -> (Vec<String>, Vec<String>, Vec<String>) {
+fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>, Vec<String>) {
     let mut e_patterns: Vec<String> = Vec::new();
     let mut positionals: Vec<String> = Vec::new();
     let mut flags: Vec<String> = Vec::new();
@@ -56,10 +58,10 @@ fn extract_pattern_path(args: &[String]) -> (Vec<String>, Vec<String>, Vec<Strin
     let mut i = 0;
 
     while i < args.len() {
-        let arg = &args[i];
+        let arg = args[i].as_ref();
 
         if past_dashdash {
-            positionals.push(arg.clone());
+            positionals.push(arg.to_string());
             i += 1;
             continue;
         }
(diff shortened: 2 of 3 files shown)


aeppling:
Hey @KuSh 

Starting the first review pass with the -vvv bug we talked about.
What I found (tested on your branch, compiled binary):

 Command ran             | BRanch  KuSh:grep_arg_parsing           
-------------------------|--------------------------
rtk -vvv grep --version | OK (no more -vvv error)     
rtk -vvv (alone)             | broken: Failed to resolve '-vvv' via PATH 
rtk -vvv git --badflag    | broken: Failed to resolve '-vvv' via PATH 
rtk -vvv frobnicate        | broken: Failed to resolve '-vvv' via PATH 

So the symptom only disappears for grep, because you made grep's parsing infallible (trailing_var_arg). For every other command that fails to parse with a leading global flag, the bug is still fully there.

The real root cause isn't in grep. This may be in run_fallback when clap fails to parse, the fallback takes the first token of argv as the command to execute via $PATH. If you lead with a global flag (-vvv, --verbose, --ultra-compact, --skip-env), that first token is the flag, so rtk tries to run a binary called -vvv. I think it is a RTK fallback bug, not related to grep directly.

Fix direction (without touching grep): in run_fallback, skip leading global flags before choosing the command; if nothing is left (rtk -vvv alone), show clap's normal error

aeppling:
Re @KuSh !

On the grep arg parsing this time :p --fixed-strings before the pattern #2120 is fixed, but extract_pattern_path seems to fail on some args sadly 

Command ran                     |   Branch KuSh:grep_arg_parsing
-----------------------------  |------------------------------
rtk grep -i todo src              |   OK (now filtered, was fallback)
rtk grep -A 2 error src          |   broken: rg error, "2" taken as the pattern
rtk grep -rn foo src              |   wrong: match replaced, prints "n" instead of the line
rtk grep TODO src tests       |   broken: rg "No such file" (multi-path)
rtk grep -i x agent -g '*.md' |   broken: regex parse error (the #2167 repro)

The approach looks good to me, removing pattern from the main definition to catch all as args and then parse,

But the rule "dash means flag, no dash means pattern/path" does not match how grep/rg flags actually work, so it misreads three common shapes:

- Value-taking flags: grep -A 2 error : the 2 (value of -A) has no dash, so it's taken as the pattern. Same with -g "*.md" (glob value becomes the pattern, rg errors). So #2167 not fully closed 

- Combined clusters: the -r strip only matches the standalone -r, so -rn leaks to rg, where -r means --replace and rewrites every match to "n". grep -rn foo prints n instead of the line.

- Multiple paths: grep TODO src tests keeps only one path and breaks the rg call.


-------


### So what can i propose for this issue:

- We could hold a list of value-taking flags so the splitter consumes their value instead of reading it as the pattern. Plus strip -r out of combined clusters and keep every non-flag positional as a path.

Problem: we'd have to know each flag's exact behavior (some take multiple values, some are optional), and maintain this exhaustive list.

- We give the parser a small spec of the flags rtk cares about and their arity (-A/-B/-C/-g/-e take one value, -i/-w/-n take none, grep's -r means recursive so we drop it), with one rule for everything else: unknown flag = pass straight through to rg untouched. 

Problem: this still involves a list, like the first option, but the difference is where the risk lands. In the first option the list is load-bearing: miss a value-flag and it silently corrupts again. Here the list sits on top of a passthrough, so a missing entry just means that flag is not specially optimized, the command itself is not broken.

-------

Honestly this seems like an important issue, so I'm in favor of fixing it on your splitter directly: a small value-taking flag list (kept deliberately small at first, the common ones, then grown), plus stripping -r out of combined clusters, plus keeping every non-flag positional as a path.

We can move to another approach later if this gets painful, but we stop the bleed now on the known cases

Of course open to others ideas

aeppling:
I've did not go through all #2138 but we'll get conflicts, maybe merge this one first since it define pattern parsing then 2138 will rebase on it

KuSh:
@aeppling took care of most of your feedback except:
- `rtk -vvv` alone or on other commands, will be the subject of another PR
- the #2167 complete repro as it used `-l` arg that is another problem, worth another PR

aeppling:
Re @KuSh

Rechecked with latest commit, all last tests I shown are now green, but here is some additionals, and all five worked on develop, so they are working-to-broken regressions:

Command ran                          | Branch KuSh:grep_arg_parsing
-------------------------------------|------------------------------
rtk grep -ecarrot file               | wrong: matches "caot" instead of "carrot" (silent)
rtk grep aaa sub -g'*.rs'            | wrong: glob becomes *.s, 0 matches (silent)
rtk grep compact sub --glob '*.md'   | broken: glob taken as path, rg error
rtk grep --max-count 1 fn file       | broken: rg error, "1" taken as the pattern
rtk grep -E 'a|b' file               | broken: rg reads -E as encoding flag


The first two are blockers, silent wrong results: `strip_r` eats the r's of attached values, not just the cluster letters. To be fair the trap was in my suggestion too, it should have said flag letters only.

Rows 3 and 4 are blockers too: the value list only covers short flags, the separated long forms put the value in the pattern or paths.

Row 5 is dialect territory, grep's `-E` is boolean ERE but rg's `-E` takes an encoding value. Probably #2138's job to handle properly, just flagging it so it does not get lost.

The current suite stays green because none of these shapes are covered, so a test per row would lock it in.

KuSh:
@aeppling thanks for feedback, I've pushed a fix

- **Left-to-right cluster scan**: The old "check last byte" algorithm applied `strip_r` to the prefix, corrupting inline values like `-ecarrot` → -`ecaot` or `-g*.rs` → -`g*.s`. The new scan walks the cluster byte-by-byte; when it hits a value-taking flag letter (`e`, `A`, `g`, `t`, `M`, etc.), it stops and treats everything after that letter as the inline value — no r/R stripping touches it.

- **VALUE_FLAGS_LONG**: New constant covering all rg long flags that take a space-separated value (`--glob`, `--max-count`, `--type`, `--engine`, `--ignore-file`, `--path-separator`, `--field-context-separator`, etc.). `--regexp` is handled separately as the long form of `-e`.

- **VALUE_FLAGS_SHORT**: Expanded from `ABCgfjm` to `ABCgfjmtTdM` (added `--type`, `--type-not`, `--max-depth`, `--max-columns`).

- `has_format_flag`: Added `--count-matches`, `--json`, `--passthru`, `--files` so those output modes get passthrough treatment instead of silently showing "0 matches".

The #2138 test contract and -E dialect conflict are left untouched, I'll document in that PR what should be done

aeppling:
Hey  @KuSh ,  LGTM !

Thanks for addressing all fix

aeppling (approved):

PR #2334, opened 2026-06-08
fix(rewrite): passthrough native grep/ls when short flags collide with rtk options

BurpeeKiller:
## Problem

The auto-rewrite hook turns `grep …`/`ls …` into `rtk grep …`/`rtk ls …` via a pure prefix substitution. Native short flags are then re-parsed by clap under the rtk subcommand, where several collide with a **different** semantics and silently change behavior — no error, no signal:

| native | clap meaning under rtk | native meaning | effect |
|--------|------------------------|----------------|--------|
| `grep -v` | `--verbose` (global) | invert-match | **returns matching lines instead of non-matching — result inverted** |
| `grep -h` | `--help` | no-filename | help text, 0 matches |
| `grep -l` | `--max-len` | files-with-matches | wrong arg consumed |
| `grep -m` | `--max` | max-count | shifted semantics |
| `grep -V` | `--version` | version | rtk version, not grep |
| `ls -h` | `--help` | human-readable | help instead of listing |
| `ls -V` | `--version` | — | rtk version |

The worst is `grep -v`: it is parsed as the global verbose flag, so `rtk grep -v PATTERN FILE` prints **matching** lines (and a `grep: '…' in …` verbose line) instead of the **non-matching** lines the user asked for. Silent, no error.

## Fix

Add a conservative passthrough guard in the rewrite engine (`registry::rewrite_command`): when the native command carries a short-flag cluster containing a colliding flag, skip the rewrite (`return None`) so the command runs on the native binary with its intended semantics. This mirrors the existing skip-rewrite handling for `gh --json` and incompatible `cat` flags.

- Handles bundled clusters (`-rv` → `r`, `v`).
- Long flags (`--invert-match`, `--max-count`, …) are unaffected — they already reach ripgrep/ls intact via `extra_args`.
- Safe flags (`-i`, `-rn`, `-A`, `-c`, `ls -la`, …) still rewrite exactly as before.

## Tests

Added regression tests covering each colliding flag, bundled clusters, the long-form equivalents, and the safe-flags-still-rewrite path. Verified end-to-end through the real `rtk hook claude` JSON protocol: colliding commands pass through unchanged, safe commands are rewritten.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -3,7 +3,7 @@
 use lazy_static::lazy_static;
 use regex::{Regex, RegexSet};
 
-use super::lexer::{split_on_operators, tokenize, TokenKind};
+use super::lexer::{shell_split, split_on_operators, tokenize, TokenKind};
 use super::rules::{IGNORED_EXACT, IGNORED_PREFIXES, RULES};
 
 /// Result of classifying a command.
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the flag by flag table, it made this easy to check. The collisions were fixed on develop by taking the flags back from clap instead of skipping the rewrite: commit 3b6b812a (2026-07-28) freed -m for GNU --max-count, commit 16228b74 (2026-08-30) stopped -l and -t shadowing native grep flags, and commit eafadcee (2026-06-26) runs the engine the user actually named. On a current develop build, grep -v prints the non-matching lines, grep -rl lists file names like native grep, grep -m 1 stops at one match and grep -V prints grep's own version, so a passthrough guard would now switch rtk grep and rtk ls off for cases that already behave correctly. One case you found is still open: -h is clap's help on both subcommands, so grep -h and ls -h print rtk help instead of no-filename and human readable sizes, and that is worth an issue or a small PR against the current argument definitions. If a part of this was missed, please comment or reopen.

PR #2337, opened 2026-06-08
feat(hooks): add warn_on_missing opt-out for startup hook banner

ramarivera:
- src/core/config.rs
- src/hooks/hook_check.rs

Adds a `[hooks] warn_on_missing` setting (default `true`) so users can silence the per-invocation "[rtk] /!\ No hook installed" warning that fires when no Claude Code hook is registered. This also covers the companion "Hook outdated" warning — they're the same banner family.

Implementation:
- New field on `HooksConfig` with `Default` implementation and a serde `default = "default_warn_on_missing"` fallback so older config.toml files keep the historical "warn by default" behavior.
- `maybe_warn` now reads the setting and threads the bool into `check_and_warn`, which short-circuits via a new pure `should_emit` helper before touching the rate-limit marker.
- The pure helper makes the gating decision testable without touching the real config directory or filesystem marker.

Tests:
- 5 new config tests cover default, explicit true/false, missing-field backwards compat, and serialize/deserialize roundtrip.
- 3 new hook_check tests cover the `should_emit` truth table for Ok / Missing / Outdated against both `warn_on_missing` values.

## Summary

-

## Test plan

### DIFF
--- a/src/core/config.rs
+++ b/src/core/config.rs
@@ -23,7 +23,11 @@ pub struct Config {
     pub limits: LimitsConfig,
 }
 
-#[derive(Debug, Serialize, Deserialize, Default)]
+fn default_warn_on_missing() -> bool {
+    true
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 pub struct HooksConfig {
     /// Commands to exclude from auto-rewrite (e.g. ["curl", "playwright"]).
     /// Survives `rtk init -g` re-runs since config.toml is user-owned.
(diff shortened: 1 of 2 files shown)


KuSh:
Closing this one in favour of #776, which adds the same opt-out for the hook warning plus an RTK_SUPPRESS_HOOK_WARNING environment variable and the README and configuration docs, and already carries an approving review. Review and comments are welcome over there.

Issue #1824, opened 2026-05-10
rg auto-rewrite to rtk grep breaks normal ripgrep flag ordering (`--glob`, `-g`, `--version`)


silouanwright:
### Summary

RTK's auto-rewrite maps `rg` commands to `rtk grep`, but `rtk grep` does not preserve ripgrep-compatible argument parsing. Common ripgrep forms with flags before the pattern, or `--glob`/`-g` between the pattern and path, get rewritten into `rtk grep` commands that fail or are not equivalent.

This was reproduced on both RTK `0.38.0` and `0.39.0` on macOS via Homebrew.

### Environment

```text
rtk 0.39.0
macOS/Homebrew install
/opt/homebrew/bin/rtk -> ../Cellar/rtk/0.39.0/bin/rtk
ripgrep 15.1.0
```

### Baseline: direct ripgrep works

```bash
rg "Project Guidelines" -g 'CLAUDE.md' .
rg "Project Guidelines" --glob 'CLAUDE.md' .
```

Both commands work with `/opt/homebrew/bin/rg`.

### Unsafe rewrites

```bash
rtk rewrite "rg --version"
# actual: rtk grep --version
# exit=3

rtk rewrite "rg 'foo' -g '*.md'"
# actual: rtk grep 'foo' -g '*.md'
# exit=3

rtk rewrite "rg 'foo' --glob '*.md'"
# actual: rtk grep 'foo' --glob '*.md'
# exit=3

rtk rewrite "rg -n 'foo' --glob '*.md' ."
# actual: rtk grep -n 'foo' --glob '*.md' .
# exit=3
```

These rewrites are not semantically equivalent to the original `rg` commands.

### Minimal failing repros

```bash
rtk grep "Project Guidelines" --glob 'CLAUDE.md' .
# grep: unrecognized option `--glob'
# exit=2

rtk grep -n "Project Guidelines" --glob 'CLAUDE.md' .
# grep: unrecognized option `--glob'
# exit=2

rtk grep --glob 'CLAUDE.md' "Project Guidelines" .
# grep: unrecognized option `--glob'
# exit=2

rtk grep "Project Guidelines" -g 'CLAUDE.md' .
# grep: invalid option -- g
# exit=2
```

### RTK-specific form that works

```bash
rtk grep "Project Guidelines" . --glob 'CLAUDE.md'
```

[...]

PR #2339, opened 2026-06-08
fix(rewrite): preserve rg command rewrites

### LINKED ISSUE #2338: rtk rewrite should keep rg on the ripgrep path
## Repro

```bash
rtk rewrite "rg -n --glob '!node_modules/**' --glob '!dist/**' \"needle\" src"
```

## Actual

```bash
rtk grep -n --glob '!node_modules/**' --glob '!dist/**' "needle" src
```

`--glob` is a ripgrep flag. Routing an `rg` command through `rtk grep` can break normal ripgrep searches, especially when hooks rewrite commands automatically.

## Expected

```bash
rtk rg -n --glob '!node_modules/**' --glob '!dist/**' "needle" src
```

Plain `grep` commands should still rewrite to `rtk grep`. This is only about preserving `rg` as the ripgrep path.

Related reports: #1824, #2167, #2301.

MasuRii:
## Summary

- Split the rewrite rule for `rg` and `grep`.
- Send `rg ...` to `rtk rg ...`, so ripgrep flags like `--glob` keep working after hook rewrites.
- Keep `grep ...` on `rtk grep ...` and update the related hook/docs examples.

Fixes #2338.
Related: #1824, #2167, #2301.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test --all`
- [x] Manual: `cargo run --quiet -- rewrite "rg -n --glob '!node_modules/**' --glob '!dist/**' \"needle\" src"` prints `rtk rg -n --glob '!node_modules/**' --glob '!dist/**' "needle" src`
- [x] Manual: `cargo run --quiet -- rewrite "grep -rn pattern src/"` prints `rtk grep -rn pattern src/`


MasuRii:
Thought it was an issue in my extension https://github.com/MasuRii/pi-rtk-optimizer/issues/12, but it was actually from the upstream

KuSh:

The problem is already dealt with by #2138 and #2333.
Closing in favor of those. Don't hesitate to test, review and comment and those PRs if something is missing

PR #2341, opened 2026-06-09
fix(ccusage): accept period field in economics

### LINKED ISSUE #2316: rtk cc-economics fails to parse ccusage 20.x weekly JSON (missing field `week`)
## Summary

`rtk cc-economics -w` fails against ccusage 20.0.6 with a JSON parse error. The wrapper appears to expect an older ccusage weekly schema.

## Environment

- rtk 0.42.3 (Linux x86_64)
- ccusage 20.0.6 (installed via `npm install -g ccusage`)
- Node v22.22.2, npm 10.9.7

## Reproduction

```
$ rtk cc-economics -w
rtk: Failed to fetch ccusage weekly data: Failed to parse ccusage JSON output: Invalid JSON structure for weekly data: missing field `week` at line 38 column 5
real    0m20.626s
user    1m16.733s
sys     0m11.187s
```

The wrapper does correctly shell out to `ccusage weekly` (we see the 20s of parsing work), but the post-processing JSON parser uses a field name that ccusage 20.x no longer emits.

## Workaround

Running `ccusage weekly` directly produces the expected report in ~20s:

```
$ ccusage weekly
# ... full table ...
Total | 8,484,547,… | $4347.91
```

## Likely cause

ccusage 20.x may have renamed the week field (e.g. `week` → `weekStart` or similar). The parser in `rtk cc-economics` was written against an earlier schema. Affects `-w` and likely `-a` (which also parses weekly rows).

## Suggested fix

Pin a ccusage version range in the rtk docs, OR adapt the JSON parser to handle both `week` and any new field name. If ccusage 20.x is the current upstream, a release update may be in order.

### LINKED ISSUE #2332: cc-economics: fails on current ccusage JSON (missing field month)
## Summary
`rtk cc-economics` crashes parsing ccusage output on rtk 0.42.3.

## Repro
```
$ rtk cc-economics
[info] ccusage not installed globally, fetching via npx...
rtk: Failed to fetch ccusage monthly data: Failed to parse ccusage JSON output:
Invalid JSON structure for monthly data: missing field `month` at line 30 column 5
```

## Cause
ccusage changed its monthly JSON schema; rtk's deserializer expects a `month` field no longer present (or renamed). The fetch-via-npx path pulls latest ccusage, so this breaks for everyone without a pinned version.

## Suggested fix
- Tolerate the new ccusage schema (rename/optional `month`), or
- Pin a known-good ccusage version for the npx fallback, or
- Degrade gracefully (skip economics, don't error) on schema mismatch.

## Env
rtk 0.42.3, macOS (Darwin 25.5.0), ccusage via npx (latest).

jsiu93:
## Summary
- Accept the current `ccusage` `period` field for daily, weekly, and monthly economics data while keeping legacy `date` / `week` / `month` aliases.
- Reuse one period entry deserializer so daily, weekly, and monthly parsing share the same field compatibility path.

Fixes #2332
Fixes #2316

## Test plan
- [x] `rtk cargo fmt --all --check`
- [x] `rtk cargo test analytics::ccusage::tests -- --nocapture`
- [x] `rtk cargo clippy --all-targets`
- [x] `rtk cargo test --all`
- [x] `rtk cargo run -- cc-economics --monthly`
- [x] `rtk cargo run -- cc-economics --weekly`
- [x] `rtk proxy npx --yes ccusage@20.0.8 monthly --json --since 20250101` emits `period`
- [x] `rtk proxy npx --yes ccusage@20.0.8 weekly --json --since 20250101` emits `period`

### DIFF
--- a/src/analytics/ccusage.rs
+++ b/src/analytics/ccusage.rs
@@ -48,36 +48,23 @@ pub enum Granularity {
 
 #[derive(Debug, Deserialize)]
 struct DailyResponse {
-    daily: Vec<DailyEntry>,
-}
-
-#[derive(Debug, Deserialize)]
-struct DailyEntry {
-    date: String,
-    #[serde(flatten)]
-    metrics: CcusageMetrics,
+    daily: Vec<PeriodEntry>,
 }
 
 #[derive(Debug, Deserialize)]
 struct WeeklyResponse {
-    weekly: Vec<WeeklyEntry>,
-}
-
-#[derive(Debug, Deserialize)]
-struct WeeklyEntry {
-    week: String, // ISO week start (Monday)
-    #[serde(flatten)]
-    metrics: CcusageMetrics,
+    weekly: Vec<PeriodEntry>,
 }
 
 #[derive(Debug, Deserialize)]
 struct MonthlyResponse {
-    monthly: Vec<MonthlyEntry>,
+    monthly: Vec<PeriodEntry>,
 }
 
 #[derive(Debug, Deserialize)]
-struct MonthlyEntry {
-    month: String,
+struct PeriodEntry {
+    #[serde(rename = "period", alias = "date", alias = "week", alias = "month")]
+    period: String,
     #[serde(flatten)]
     metrics: CcusageMetrics,
 }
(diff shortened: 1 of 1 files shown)


KuSh:
Thank you for the fix and for verifying it against real ccusage output. The same change is already on develop from #2732, merged 2026-07-07 (commit d0a77cdb), which adds `period` as a serde alias on the daily, weekly and monthly key fields while keeping the legacy date/week/month names, with parse tests for all three granularities. The consolidation into a single period entry type is a nice cleanup, but the parsing behaviour is identical to what is shipped, so there is nothing left failing here. Closing as already covered, and please comment or reopen if you hit a ccusage schema this still does not parse.

PR #2348, opened 2026-06-09
fix(grep): use PCRE-converted pattern with -E flag in grep fallback

shiloong:
## Problem

When rtk grep falls back from rg to grep (when rg is unavailable), the fallback uses the original BRE pattern and lacks the `-E` (extended regex) flag. This causes patterns with BRE-to-PCRE conversion to fail silently:

1. User passes a BRE pattern like `fn foo\|pub.*bar` to `rtk grep`
2. rtk converts it to PCRE `fn foo|pub.*bar` for rg
3. When rg is unavailable, grep receives `fn foo|pub.*bar` (PCRE) but **without** `-E`, grep treats `|` as a literal pipe character
4. **Result**: zero matches returned, misleading AI agents into thinking the pattern was not found

## Fix

- Use the PCRE-converted pattern (`rg_pattern`) instead of the original BRE pattern in the grep fallback
- Add `-E` (extended regex) flag to the grep invocation so `|` alternation is interpreted correctly

```diff
-            grep_cmd.args(["-rnHZ", pattern, path]).args(extra_args);
+            grep_cmd.args(["-rnHEZ", &rg_pattern, path]).args(extra_args);
```

## Before

```
$ rtk grep "fn foo\|pub.*bar" .
0 matches for 'fn foo\|pub.*bar'
```

## After

```
$ rtk grep "fn foo\|pub.*bar" .
X matches in YF:

[file] src/main.rs (3):
   10: fn foo() { ... }
   25: pub struct Bar { ... }
```

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -54,7 +54,7 @@ pub fn run(
         .or_else(|_| {
             let mut grep_cmd = resolved_command("grep");
             // When we fall back to grep, include all args, not just -rnHZ.
-            grep_cmd.args(["-rnHZ", pattern, path]).args(extra_args);
+            grep_cmd.args(["-rnHEZ", &rg_pattern, path]).args(extra_args);
             exec_capture(&mut grep_cmd)
         })
         .context("grep/rg failed")?;


KuSh:
Thanks for digging into this. The BRE to PCRE conversion this fixes no longer exists on develop: PR #2641 (merged 2026-06-26) stopped substituting one engine for the other, so `rtk grep` now runs real grep with your own pattern and a separate `rtk rg` runs ripgrep. grep_cmd.rs was renamed to search.rs in that change and the fallback path you patched is gone, which removes the whole class of pattern mismatch you described. If you still hit a case where a grep pattern is mangled on the current develop, please comment here or reopen with the command and we will take another look.

PR #2354, opened 2026-06-09
fix(proxy): honor Claude Code permission rules in rtk proxy (#2345)

### LINKED ISSUE #2345: Security: rtk proxy bypasses .env read restrictions and exfiltrates credentials
## Summary

rtk proxy can be used to bypass Claude Code permission rules that block direct .env reads, allowing credential exfiltration from .env files even when Read(.env*) and Bash(cat .env*) are explicitly denied in ~/.claude/settings.json.

## Related

- CVE-2026-33068: https://nvd.nist.gov/vuln/detail/CVE-2026-33068

## Steps to Reproduce

1. Configure Claude Code deny rules for Read(.env*), Bash(cat .env*), Bash(grep .env:*)
2. Run: rtk proxy grep -r "PASSWORD" .env
3. Result: executes without permission check, returns credential values

## Impact

- Severity: Critical
- Permission bypass / credential exfiltration
- Any Claude Code deny rule targeting .env access bypassed via rtk proxy

## Expected Behavior

rtk proxy should respect active Claude Code permission rules, OR explicitly document the bypass so users can make informed security decisions.

## Suggested Mitigation

Add deny rules: Bash(rtk proxy grep*), Bash(rtk proxy cat*), Bash(rtk proxy * .env*)
Root fix must be in rtk itself.

ojoaldato:
Closes #2345.

## Problem
`rtk proxy <cmd>` spawned the inner command without consulting Claude Code's permission rules, so an agent could read `.env` secrets even when `.env` reads were denied:

```
rtk proxy grep -r "PASSWORD" .env   # printed credential values
```

Just replaying `Bash(...)` deny rules is not enough: the reproduction matches none of them (`Read(.env*)` is a Read rule; `Bash(grep .env:*)` only matches a `grep .env` prefix). Only Claude Code's `.env` protection covers it, so the fix recognizes `.env` access directly.

## Fix
A gate runs in the proxy arm (`src/main.rs` → `permissions::proxy_block_reason`) and refuses **before any child process is spawned** when:

1. **An argument references a `.env` secret file** — matched by path basename, **case-insensitively** (so `.ENV` is caught on case-insensitive filesystems like default macOS) and across **attached-value forms** (`--env-file=.env`, `dd if=.env`, `curl @.env`, `-f.env`). `.env.example` / `.env.sample` / `.env.template` placeholders stay readable.
2. **The command is an inline interpreter eval** (`sh -c ...`, `bash -lc ...`, `python -c ...`, `perl -e ...`) — rtk cannot enforce `.env`/deny protection inside an interpreted payload, so it is refused. A bare `sh script.sh` (no inline flag) is left alone.
3. **The command hits a `Deny` verdict** from the existing, tested `check_command` machinery, so `Bash(...)` deny rules apply too.

`Ask` / `Default` / `Allow` pass through, since the host already permitted the outer `rtk proxy` invocation.

### Residual limitations (documented, per the issue's "respect rules OR document the bypass")
This is best-effort argument inspection. **Recursive readers** that never name `.env` (`grep -r PASSWORD .`, `find -exec cat`), **launcher wrappers** (`env`/`timeout`/`nice <cmd>`), and shell obfuscation can still reach `.env`. Also, rtk's own file-reading subcommands (`rtk read` / `rtk grep` / `rtk log`) read files via their own paths and are **outside this proxy guard** — closing those is a separate, related change.

## Verification
- `cargo fmt --all -- --check` and `cargo clippy --all-targets`: clean (repo enforces `warnings = "deny"`)
- `cargo test --all`: 2162 passed, 0 failed (17 unit tests for the gate, incl. the reproduction and the interpreter / case / attached-value bypasses)
- End-to-end against a live `.env`: the reproduction and every bypass above are refused (exit 1) and the secret value leaks **0 times**; `bash deploy.sh`, `echo hi`, and `cat .env.example` still work.

This hardening was driven by an adversarial multi-perspective review of the original patch, which found the direct-only block was bypassable via interpreter wrappers, case, and attached-value forms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,8 @@ rtk summary <long command>      # Heuristic summary
 rtk proxy <command>             # Raw passthrough + tracking
 ```
 
+> **Security:** `rtk proxy` enforces your Claude Code deny rules and blocks reads of `.env` secret files (`.env.example` / `.env.sample` / `.env.template` excepted), including via inline interpreter scripts (`sh -c ...`). This is best-effort: recursive readers (`grep -r PASSWORD .`), launcher-wrapped commands (`env`/`timeout`), and shell obfuscation can still reach `.env`, and rtk's own `read`/`grep`/`log` subcommands are not yet covered ([#2345](https://github.com/rtk-ai/rtk/issues/2345)).
+
 ### Token Savings Analytics
 ```bash
 rtk gain                        # Summary stats
--- a/src/hooks/permissions.rs
+++ b/src/hooks/permissions.rs
@@ -114,6 +114,167 @@ pub(crate) fn check_command_with_rules(
     }
 }
 
+// --- Proxy permission gate (issue #2345 / CVE-2026-33068) -------------------
+//
+// `rtk proxy <cmd>` executes an arbitrary command. Without a gate it bypasses
+// Claude Code's permission rules entirely, letting an agent read `.env` secrets
+// (e.g. `rtk proxy grep -r PASSWORD .env`) even when `.env` reads are denied.
+// `proxy_block_reason` is the gate the proxy handler calls before spawning.
+
+/// `.env.<suffix>` variants that conventionally hold placeholder values rather
+/// than real secrets, so they stay readable through proxy (matches Claude
+/// Code's allowance of `.env.example`). Matched case-sensitively.
+const SAFE_ENV_SUFFIXES: &[&str] = &["example", "sample", "template"];
+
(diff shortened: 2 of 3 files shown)


ojoaldato:
@pszymkowiak would love your thoughts on this one :)

ojoaldato:
@pszymkowiak could I get some eyes on it? :)

ojoaldato:
closing due to lack of reviews

PR #2357, opened 2026-06-10
feat(discover): count hook-rewritten commands as captured, not missed

rasatpetabit:
## Problem

`rtk discover` scans the model-authored `tool_use` commands in Claude Code transcripts. When the PreToolUse hook rewrites a command via `updatedInput` (e.g. `git add .` → `rtk git add .`), the transcript still records the original command — the rewrite happens after the model authors it. As a result, every hook-captured command is reported as a *missed* saving, and `Already using RTK` only counts commands the model literally typed with an `rtk ` prefix.

On a host with the hook active this makes the report wildly misleading. Real example: discover claimed **0.2% adoption and ~5.4M tokens missed**, while `rtk gain` showed **131M tokens already saved (63.6%)** and the hook audit log showed 21,775 rewrites in the same window.

## Fix

Cross-reference the hook audit log (`hook-audit.log`, written by `rtk hook` when `RTK_HOOK_AUDIT=1`):

- New `hook_audit_cmd::load_rewritten_originals(since_days)` builds a set of original commands with action `rewrite`, unescaping the pipe-delimited log fields (reverse of `sanitize_log_field`). Best-effort: any read/parse failure returns an empty set, so discover never fails because of the audit log, and behavior is unchanged when auditing is disabled.
- `discover::run` classifies a transcript command found in that set as **hook-captured** instead of feeding it into the missed-savings buckets.
- The report gains a `hook_captured` field and a `Captured by hook rewrite: N commands (X%)` line (only rendered when non-zero, so existing output is unchanged for users without audit logs).

After the fix on the same host: `Captured by hook rewrite: 46,251 commands (35.0%)`, and the missed-savings list shrinks to commands the hook genuinely skipped (`skip:no_match`/`skip:defer`) or that predate audit logging.

## Tests

- `unescape_log_field` round-trip vs `sanitize_log_field`
- loader: action filtering, escaped-pipe originals, malformed lines, `since_days` cutoff, missing file → empty set
- report: line rendered when non-zero, hidden when zero

`cargo fmt --all && cargo clippy --all-targets && cargo test --all` clean (the 3 `unattestable_passthrough` failures on my host are a pre-existing environment-isolation issue — those tests read the live `~/.claude/settings.json` permission lists and fail on any host with allow/deny rules configured; they pass with a clean `$HOME`. Happy to file separately.)

### DIFF
--- a/src/discover/mod.rs
+++ b/src/discover/mod.rs
@@ -72,8 +72,15 @@ pub fn run(
         }
     }
 
+    // Commands the Claude Code hook already rewrote to rtk at execution time.
+    // Transcripts record the model-authored command (pre-rewrite), so without
+    // this cross-reference every hook-captured command would be reported as
+    // a missed saving.
+    let hook_rewritten = crate::hooks::hook_audit_cmd::load_rewritten_originals(since_days);
+
     let mut total_commands: usize = 0;
     let mut already_rtk: usize = 0;
+    let mut hook_captured: usize = 0;
     let mut parse_errors: usize = 0;
     let mut rtk_disabled_count: usize = 0;
     let mut rtk_disabled_cmds: HashMap<String, usize> = HashMap::new();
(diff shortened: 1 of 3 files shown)


KuSh:
Thanks for digging into this, the discrepancy you measured between discover and rtk gain was real and worth chasing. It was fixed on develop by #3206, merged 2026-09-02, which logs the real PreToolUse decision keyed by tool_use_id into a hook_decisions table and joins it back to the transcript, so hook-rewritten commands now count as captured instead of missed, with a labelled estimate for history that predates logging. That covers the classification and the report line this PR added, and it also handles permissions.deny, unattestable constructs and the RTK_DISABLED bypass, which a match against the audit log could not. Closing as covered, but please comment or reopen if you see a case it still gets wrong.

PR #2366, opened 2026-06-10
feat(bloop): add a `bloop` filter for scala build output

meier-christoph:
## Summary

- Adds a self-contained filter for `rtk bloop test|compile|run` (plus passthrough for other subcommands) — Bloop is the Scala build **server** an agent hits on every compile/test in the inner loop, and its output is verbose and stack-trace-heavy.
- Same output convention as rtk's other test filters (`N passed, M failed` summary + `[FAIL]` blocks + tee-to-disk for recovery); **avg 88.8% token savings on real output** (up to ~100%).
- Validated against a real-output corpus (circe, cats-effect, zio-json, enumeratum, Play) — not synthetic fixtures.
- Complements, and references, the open sbt PR #752 — but does **not** depend on it (see [Relationship to #752](#relationship-to-the-sbt-pr-752)).

## Background: the Scala build tools (for anyone not in the ecosystem)

Three command names produce Scala build output. An agent's terminal sees all three:

| Tool | What it is |
|------|-----------|
| **sbt** | The canonical Scala build tool. Owns the build definition (`build.sbt`), dependency resolution, code generation, the whole task graph. JVM-based, slow cold start. |
| **sbtn** | "sbt native client" — a thin client that talks to a long-lived sbt **server**, so you skip JVM startup. Same build, **same output as sbt**. |
| **Bloop** | A standalone compile/test/run **server** built for fast, always-warm incremental compilation. The Scala IDE (Metals) drives it, and many devs/agents use its CLI directly. Gets its project config **exported from sbt** (`sbt bloopInstall`). |

**Why a separate filter:** sbt and sbtn share one output format (the sbt PR #752 routes both through one filter). **Bloop is a different program with its own output format** — different progress lines, failure markers, and summary shape — so it can't reuse the sbt filter and warrants its own.

**Why bloop for the agent loop:** sbt/sbtn stays the source of truth — you need it for the build definition and **code generation** (e.g. Play routes/Twirl, sbt-plugin sources). But for the repetitive *compile → test → fix* inner loop, Bloop skips sbt's task-graph overhead and is the fast path, which is exactly where an agent spends its tokens. So the two are complementary: sbt/sbtn owns the build, Bloop is the fast way to run it.

## What the filter does

Keeps the actionable signal (pass/fail tally, failing test name + reason + `file:line`, compile errors/warnings with location), strips any ANSI color, and tees the full log to disk for recovery.

Example — a real circe test run, one failing assertion (130 KB → 364 bytes, **99.7%**):

```
bloop test: 1836 passed, 1 failed (36.0s)
  [FAIL] io.circe.JsonSuite.deepDropNullValues should remove null value for JsonArray
     munit.ComparisonFailException: JsonSuite.scala:187
     => Diff (- obtained, + expected)
     JString(
     -      value = "a"
     +      value = "REGRESSION"
     )
[full output: ~/.local/share/rtk/tee/…_bloop_test.log]
```

The output shape follows rtk's existing test-filter convention (the same `N passed, M failed` + `[FAIL]` + tee-hint shape used by the sbt filter in #752), so it reads as a shorter version of a Scala test run, not a new format.

## Coverage

Handles all four common Scala test frameworks (**ScalaTest, specs2, munit, zio-test**) and both `compile` (errors + warnings) and `test`. ANSI color in the input is stripped. The parser is **marker-based, not line-order dependent**, so parallel multi-module builds that interleave output on the shared stream stay correct.

- **58 generated fixtures**, avg **78.6%** savings — every framework × failure shape (assertion fail, uncaught error, warnings, skipped, no-tests, multi-module, color).
- **22 real-repo fixtures**, avg **88.8%** savings (47.7%–100%; the floor is a tiny 2-warning compile where there's little to compress), captured at pinned tags (spanning Scala 2.12 / 2.13 / 3, all four frameworks). These surfaced — and fixed — several gaps the synthetic set never exposed:

  | Repo | Tag | Scala | Framework |
  |------|-----|-------|-----------|
  | [circe](https://github.com/circe/circe) | [v0.14.10](https://github.com/circe/circe/releases/tag/v0.14.10) | 2.13 | munit |
  | [enumeratum](https://github.com/lloydmeta/enumeratum) | [enumeratum-1.7.5](https://github.com/lloydmeta/enumeratum/releases/tag/enumeratum-1.7.5) | 2.12 | ScalaTest |
  | [zio-json](https://github.com/zio/zio-json) | [v0.7.3](https://github.com/zio/zio-json/releases/tag/v0.7.3) | 2.13 | zio-test |
  | [cats-effect](https://github.com/typelevel/cats-effect) | [v3.5.7](https://github.com/typelevel/cats-effect/releases/tag/v3.5.7) | 3 | specs2 |
  | [Play](https://github.com/playframework/playframework) | [3.0.11](https://github.com/playframework/playframework/releases/tag/3.0.11) | 3 | specs2 + JUnit (multi-module compile + test) |

All fixtures are **real captured output** (not synthetic), per rtk's fixtures rule. The full corpus, the generator, and the capture scripts live in a companion repo: **https://github.com/meier-christoph/rtk-scala-examples** (per-fixture savings in `fixtures/bloop/savings.csv` and `fixtures/real/bloop/real_savings.csv`).

## Conformance with rtk's design principles

- **Transparency** — output is a shorter version of a Scala test/compile run, using rtk's existing test-filter shape; no bloop-specific format invented.
- **Never block** — on unrecognized output the filter falls back to passthrough, and every filtered run tees the raw log so nothing is lost.
- **Zero overhead** — pure `&str -> String` filter, `lazy_static` regex, no network or hot-path disk reads.
- **Correctness vs savings** — a failing run keeps name + reason + location inline (the part an agent needs to fix it); only the passing noise and stack framing are dropped.
- **Extensibility** — reuses the shared `core/` helpers (`display_helpers`, `runner`, `truncate`, `utils`).

## What's in this PR

- `src/cmds/scala/bloop_cmd.rs` — the filter (`run_test` / `run_compile` / `run_run` / `run_other`) + unit and snapshot tests.
- `src/cmds/scala/mod.rs`, `src/cmds/mod.rs` — the `scala` module scaffolding (two lines; trivially shared with #752 if it lands first).
- `src/main.rs` — `Commands::Bloop` + `BloopCommands` + routing.
- `src/discover/rules.rs` — rewrite patterns so the hook auto-rewrites bare `bloop …` → `rtk bloop …`.
- `tests/fixtures/bloop/*` — the real-output corpus.

## Test plan

- [ ] Snapshot tests (insta) for each framework × situation, each asserting ≥50% savings
- [ ] Manual: `rtk bloop test <project>` / `rtk bloop compile <project>` inspected against raw

## Relationship to the sbt PR (#752)

This filter is **standalone** and doesn't depend on #752 — its only overlap is the two-line `scala` module scaffolding, which it includes itself and which merges cleanly in either order. sbt/sbtn (#752) and bloop cover different tools and different output formats; landing them independently is fine.


meier-christoph:
We have been using this build internally for almost 2 weeks now and several of our devs have been using it daily. There have been no issues so far and the feedback I got was only positive. Ready for review.

meier-christoph:
rebased on latest develop branch

pszymkowiak:
Nice work — clean filter (shared `run_filtered` + tee, passthrough for unsupported subcommands, real-corpus fixtures, honest 50–60% savings floors). Reviewed locally: builds, 59 tests pass, clippy clean. Two things before merge, both mechanical:

**1. Rebase the `scala/` conflict (from #752).** SBT support (#752) merged and already created `src/cmds/scala/mod.rs` (`automod::dir!`) and the `src/cmds/mod.rs` entry — same lines this PR adds, hence the add/add conflict. Resolution is just to drop those two additions from this PR (develop already has them); `bloop_cmd.rs`, the rule, and the fixtures are untouched.

**2. Coordinate the `RtkRule` field with #3128.** #3128 adds a mandatory `pipeline_final_safe: bool` field to `RtkRule`. Whichever of the two merges second needs the other's rule updated — for `bloop` it's `pipeline_final_safe: false` (build tool, not a stdin filter). Flagging so it doesn't become a surprise compile error.

Nothing blocking on the substance — happy to re-review once rebased.

meier-christoph:
thanks, rebased, fixed the conflict and added a todo waiting for #3128

meier-christoph:
@pszymkowiak updated for #3128 and #3244, would appreciate an other review. 

note: I'll be on vacation for a few days, may not be able to reply right away.

meier-christoph:
rebased and moved the bloop changes below sbt

PR #2368, opened 2026-06-10
fix(discover,learn): encode Windows drive colon in project path slug

### LINKED ISSUE #2367: discover/learn: Windows drive colon breaks the default project filter (always scans 0 sessions)
### Summary

On native Windows, `rtk discover` and `rtk learn` always report **`Scanned: 0 sessions`** when run without `--all`/`--project`, because `encode_project_path()` keeps the drive colon (`C:` → `C:`) while Claude Code encodes it to a dash (`C:` → `C-`) when naming `~/.claude/projects/` directories. The encoded cwd can therefore never substring-match the real project directory name.

### Reproduction (Windows 11, rtk 0.42.3)

```powershell
PS F:\Desktop\AI\claude\vector_fading> rtk discover
RTK Discover -- Savings Opportunities
====================================================
...
No missed savings found. RTK usage looks good!
```

But the history is there — Claude Code stores this project at:

```
C:\Users\<user>\.claude\projects\F--Desktop-AI-claude-vector-fading\   <- colon became '-'
```

while rtk computes the default filter as:

```
F:-Desktop-AI-claude-vector-fading                                     <- colon kept
```

`discover_sessions_in_projects_dir()` matches via `dir_name.contains(filter)` (`src/discover/provider.rs`), and `"F--Desktop-…".contains("F:-Desktop-…")` is always false → 0 sessions for **every project on every native-Windows machine**.

Workaround proof — bypassing the broken default filter finds the sessions fine:

```powershell
PS> rtk discover --project Desktop-AI-claude-vector-fading
Scanned: 504 sessions (last 30 days), 3382 Bash commands
...
```

### Root cause

`src/discover/provider.rs::encode_project_path()`:

```rust
const SANITIZED_CHARS: &[char] = &['/', '.', '_', '\\', ' ', '[', ']'];   // no ':'
```

The doc comment pins the wrong expectation explicitly (`C:\Users\foo\bar → C:-Users-foo-bar`), and `test_encode_project_path_windows` asserts it. Observed Claude Code behavior on Windows (e.g. `projects/C--Users-Edu`, `projects/F--Desktop-AI-claude-vector-fading`) replaces the colon like any other special character.

Both `src/discover/mod.rs` and `src/learn/mod.rs` build their default filter from this function, so both commands are affected.

### Relationship to existing issues

- #2177 documents the "0 sessions without `--all`" symptom; this issue is the Windows-specific root cause of that symptom.
- #2150 / PR #2172 fixed the same class of bug for spaces/brackets/non-ASCII; the drive colon needs the same treatment.
- PR #2228 (default to `--all`) would mask the symptom for the no-flag case but `--project`-less filtering would still be broken conceptually; the encoder should match Claude regardless.

### Fix

Add `':'` to `SANITIZED_CHARS`, correct the doc comment, fix the wrong test expectation, and add a drive-colon regression test. PR incoming.

### Environment

- rtk 0.42.3 (also reproduced reading source at `develop` HEAD)
- Windows 11 Pro 10.0.26200, PowerShell host
- Claude Code with default `~/.claude` config dir

ALCHEMISTK:
## Summary

Fixes #2367 — on native Windows, `rtk discover` and `rtk learn` always scanned **0 sessions** without `--all`/`--project`, because `encode_project_path()` kept the drive colon (`C:\Users\foo` → `C:-Users-foo`) while Claude Code's project slug encodes it to a dash (`C--Users-foo`). The cwd-derived default filter could therefore never substring-match the real `~/.claude/projects/` directory name.

## Changes

- Add `':'` to `SANITIZED_CHARS` in `encode_project_path()` (same treatment #2172 gave spaces/brackets/non-ASCII)
- Correct the doc-comment example (`C:\Users\foo\bar` → `C--Users-foo-bar`)
- Fix `test_encode_project_path_windows`, which pinned the wrong expectation
- Add `test_encode_project_path_windows_drive_colon` regression test

Both `discover` and `learn` build their default filter from this function, hence the dual scope.

## Verification

- `cargo test`: **2149 passed, 8 ignored** (full suite, Windows 11)
- Real-world before/after on Windows 11 + Claude Code, rtk 0.42.3 base:
  - before: `rtk discover` → `Scanned: 0 sessions (last 30 days), 0 Bash commands`
  - after (encoder parity, validated via the equivalent `--project` substring): `Scanned: 504 sessions (last 30 days), 3382 Bash commands`

## Notes

- Unix paths containing `:` are legal but Claude Code encodes them to `-` there too, so the change strictly improves parity on all platforms.
- Complementary to #2228 (default to `--all`): that PR masks the no-flag symptom; this one fixes the encoder so cwd-scoped filtering actually works on Windows.

### DIFF
--- a/src/discover/provider.rs
+++ b/src/discover/provider.rs
@@ -117,15 +117,15 @@ impl ClaudeProvider {
 
     /// Encode a filesystem path to Claude Code's directory name format.
     ///
-    /// Claude Code replaces `/`, `.`, `_`, `\`, and any non-ASCII character
+    /// Claude Code replaces `/`, `.`, `_`, `\`, `:`, and any non-ASCII character
     /// with `-` when computing the project directory slug under `~/.claude/projects/`.
     ///
     /// `/Users/foo/bar`          → `-Users-foo-bar`
     /// `/Users/first.last/bar`   → `-Users-first-last-bar`
     /// `/home/chris/2_project`   → `-home-chris-2-project`
-    /// `C:\Users\foo\bar`        → `C:-Users-foo-bar`
+    /// `C:\Users\foo\bar`        → `C--Users-foo-bar`
     pub fn encode_project_path(path: &str) -> String {
-        const SANITIZED_CHARS: &[char] = &['/', '.', '_', '\\', ' ', '[', ']'];
+        const SANITIZED_CHARS: &[char] = &['/', '.', '_', '\\', ' ', '[', ']', ':'];
 
         path.chars()
             .map(|c| {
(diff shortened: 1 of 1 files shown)


ALCHEMISTK:
done

KuSh:
Thanks for tracking this down. The drive-letter colon is now sanitized on develop: #2952 (merged 2026-08-13) added ':' to SANITIZED_CHARS in encode_project_path, fixed the doc-comment example to C--Users-foo-bar, and updated test_encode_project_path_windows to the same expectation, so the cwd-derived project filter matches Claude Code's real folder name on Windows. The only piece here that is not on develop is a second assertion of the same code path with a different drive letter. Closing as already fixed, but please comment or reopen if something in your setup is still not covered.

PR #2377, opened 2026-06-10
feat(read): add Auto filter level that scales by file size

fiorentinjoao:
## Summary

  - Add `FilterLevel::Auto` — a filter level that scales by file size instead of being fixed. It resolves to a concrete level from the file's line count: **<100 lines → None, 100–500 → Minimal, >500 → Aggressive**.
  - Wired into `rtk read` (file path only) as an **opt-in** via `--level auto`. The default (`none`) is unchanged, so no existing behavior changes for anyone.
  - This directly follows the project's *Correctness vs Token Savings* principle: small files pass through untouched (the agent needs full context), while large files — where the tokens actually pile up — get aggressive compression.

  ### Why

  A fixed filter level is a poor fit for a mixed codebase: `none` wastes tokens on huge files, `aggressive` strips context from small ones that the agent needs to read in full. `Auto` removes that trade-off by picking the level per file.
  
  The logic is intentionally tiny and lives entirely in `FilterLevel::resolve()`, so it's easy to review and the thresholds are trivial to tune.
  
  ### Changes

  - `src/core/filter.rs`: add the `Auto` variant + `resolve(line_count)`, plus `FromStr` (`"auto"`), `Display`, and `get_filter` handling (Auto maps to no-op when unresolved, as a safety default).
  - `src/cmds/system/read.rs`: resolve `Auto` by line count before filtering. Only the file path is touched — stdin is a stream with no file-size semantics, so it's left as-is.
  - Tests: `"auto"` parsing, Display↔FromStr roundtrip, resolve scaling across the three bands, and identity for explicit (non-Auto) levels.

  Happy to make `auto` the **default** for `rtk read` in a follow-up if you'd like — I kept it opt-in here so this PR changes no existing behavior.

  ## Test plan

  - [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — fmt clean, no new clippy warnings in changed files, **1609 passed / 0 failed**
  - [x] Manual testing:
    - `rtk read <small_file.rs> --level auto` → unfiltered (resolves to None)
    - `rtk read <large_file.rs> --level auto -v` → logs `auto-filter: N lines → aggressive`, output compressed

### DIFF
--- a/src/cmds/system/read.rs
+++ b/src/cmds/system/read.rs
@@ -35,8 +35,15 @@ pub fn run(
         eprintln!("Detected language: {:?}", lang);
     }
 
+    // Resolve Auto to a concrete level based on file size (opt-in via --level auto)
+    let line_count = content.lines().count();
+    let resolved = level.resolve(line_count);
+    if verbose > 0 && level == FilterLevel::Auto && resolved != FilterLevel::None {
+        eprintln!("rtk: auto-filter: {} lines → {}", line_count, resolved);
+    }
+
     // Apply filter
-    let filter = filter::get_filter(level);
+    let filter = filter::get_filter(resolved);
     let mut filtered = filter.filter(&content, &lang);
 
     // Safety: if filter emptied a non-empty file, fall back to raw content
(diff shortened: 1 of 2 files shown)


tenlisboa:
This sounds really great! I hope the contributors could review it soon

henriquemod:
> This sounds really great! I hope the contributors could review it soon

joaofiorentin1:
bump

Issue #388, opened 2026-03-06
Claude fails to use `rtk proxy` because of quotes


mnapoli:
I just witnessed Claude Code run:

```
rtk proxy 'head -50 app/Models/Domaine.php'
Error: Failed to execute command: head -50 app/Models/Domaine.php

Caused by:
    No such file or directory (os error 2)
```

After debugging I figured out that it should have run:

```
rtk proxy head -50 app/Models/Domaine.php
```

i.e. because of the quotes it messed up the invocation.

Should:

- Claude instructions be clarified to avoid CC adding quotes?
- or `rtk proxy` automatically un-quote if it gets 1 argument with quotes?

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 82% |

### Summary

rtk proxy treats a single quoted argument as a literal executable name instead of splitting it into command + args; fix should unquote/split a single-argument invocation in the proxy command parser

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

KuSh:
Fixed by #480: `rtk proxy` now splits a single argument containing spaces into command plus arguments. Verified on a current develop build — `rtk proxy 'head -1 f.txt'` runs correctly instead of failing with 'No such file or directory'. Please reopen or comment if you still hit this on the latest version.

PR #2393, opened 2026-06-11
fix(err): pass child argv verbatim instead of re-splitting on whitespace

### LINKED ISSUE #2389: rtk err re-splits child command argv on whitespace — quoted arguments corrupted, exit codes silently wrong
**Version:** rtk 0.42.3 (macOS arm64, Homebrew)

`rtk err` does not preserve the argv vector of its child command. The trailing `[COMMAND]...` args are flattened and re-split on whitespace, so any single quoted argument containing a space is split into multiple arguments. `rtk proxy` preserves argv correctly, which isolates the defect to the `err` path.

**Minimal repro (argv inspection):**
```sh
cat > /tmp/argspy <<'EOF'
#!/bin/sh
printf 'argc=%s\n' "$#"
for a in "$@"; do printf 'arg=[%s]\n' "$a"; done
EOF
chmod +x /tmp/argspy

$ rtk err /tmp/argspy 'a b' c
argc=3
arg=[a]
arg=[b]          # ← 'a b' was split
arg=[c]

$ rtk proxy /tmp/argspy 'a b' c
argc=2
arg=[a b]        # ← correct
arg=[c]
```

**Observable consequence — exit code silently wrong:**
```sh
$ rtk err sh -c 'exit 7'; echo "exit:$?"
exit:0           # ← expected 7

$ rtk err -- sh -c 'exit 7'; echo "exit:$?"
exit:0           # ← `--` does not help

$ printf '#!/bin/sh\nexit 7\n' > /tmp/exit7.sh; chmod +x /tmp/exit7.sh
$ rtk err /tmp/exit7.sh; echo "exit:$?"
[FAIL] Command failed (exit code: 7)
exit:7           # ← a flagless command works correctly
```

**Mechanism:** `rtk err sh -c 'exit 7'` reaches `sh` as `sh -c exit 7` (the quoted `exit 7` re-split into `exit` + `7`). Per POSIX, `sh -c exit 7` runs the one-word script `exit` with `$0=7`, which exits 0 — so the real failure is masked. Confirmed independently: plain `sh -c exit 7` → 0, `sh -c 'exit 7'` → 7.

**Impact:** any `rtk err <cmd>` whose child takes a quoted argument containing whitespace — `sh -c '…'`, `bash -c '…'`, `make -C 'dir with space'`, `node -e 'a; b'` — runs a corrupted command. Because `err` reports status from the (wrong) child, hard failures can be reported as success.

**Expected:** `rtk err` should pass the child argv through verbatim, exactly as `rtk proxy` does.

Rafaelpta:
Fixes #2389

## Problem

`rtk err` and `rtk test` joined the trailing argv into a single string and ran it through `sh -c`. Any quoted argument containing whitespace was re-split:

```
$ rtk err sh -c 'exit 7'; echo $?
0        # expected 7
```

`sh -c 'exit 7'` reached the child as `sh -c exit 7`, which per POSIX runs the one-word script `exit` with `$0=7` and exits 0, so real failures were reported as success.

## Fix

Build the child command from argv directly via `resolved_command`, keeping parity with `rtk proxy`:

- multiple arguments pass through verbatim, each argument intact
- a single argument is still shell-split respecting quotes (#388), so `rtk err 'cargo build --release'` keeps working
- the joined string is now used only for display and the test-summary heuristics

This also removes the `cmd /C` / `sh -c` platform split; direct spawn works on both.

## Tests

- `quoted_argument_with_spaces_is_preserved_verbatim`: the #2389 repro at the argv level
- `child_exit_code_is_propagated` (unix): `sh -c 'exit 7'` exits 7 end to end
- `single_argument_is_shell_split_like_proxy` and `single_argument_quotes_are_respected_when_splitting`: #388 parity
- `empty_command_is_an_error`

`cargo test`: 2155 passed, 0 failed (the `error[E0308]` strings in test output are fixtures from `src/core/stream.rs` filter tests).

## Behavioral changes

Two edge behaviors change because the child no longer goes through `sh -c`:

- Shell operators in a single string are no longer interpreted. `rtk err 'true && exit 3'` used to exit 3 via the shell; now `&&` reaches the child as a literal argument and it exits 0. This matches `rtk proxy` today, and the old interpretation is exactly what masked exit codes. For shell syntax, use `rtk err sh -c '...'`, which now works correctly.
- A missing binary exits 1 instead of sh's 127, same as `rtk proxy`.

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -102,44 +102,54 @@ impl StreamFilter for ErrorStreamFilter {
     }
 }
 
-fn build_shell_command(command: &str) -> Command {
-    if cfg!(target_os = "windows") {
-        let mut c = Command::new("cmd");
-        c.args(["/C", command]);
-        c
+/// Build the child command with argv passed through verbatim, matching
+/// `rtk proxy` semantics: a single argument is split shell-style respecting
+/// quotes (#388), so `rtk err 'cargo build --release'` still works, while
+/// multi-argument invocations keep each argument intact. Joining argv into a
+/// `sh -c` string re-split quoted arguments and masked child exit codes
+/// (#2389).
+fn build_argv_command(argv: &[String], usage: &str) -> Result<(Command, String)> {
+    let parts: Vec<String> = if argv.len() == 1 {
+        let split = crate::discover::lexer::shell_split(&argv[0]);
+        if split.len() > 1 { split } else { argv.to_vec() }
     } else {
-        let mut c = Command::new("sh");
-        c.args(["-c", command]);
-        c
-    }
+        argv.to_vec()
+    };
+    let Some((program, args)) = parts.split_first() else {
+        anyhow::bail!("missing command\nUsage: rtk {usage} <command> [args...]");
+    };
+    let display = parts.join(" ");
+    let mut cmd = crate::core::utils::resolved_command(program);
+    cmd.args(args);
+    Ok((cmd, display))
 }
 
 /// Run a command and filter output to show only errors/warnings
-pub fn run_err(command: &str, verbose: u8) -> Result<i32> {
+pub fn run_err(command: &[String], verbose: u8) -> Result<i32> {
+    let (cmd, display) = build_argv_command(command, "err")?;
     if verbose > 0 {
-        eprintln!("Running: {}", command);
+        eprintln!("Running: {}", display);
     }
-    let cmd = build_shell_command(command);
     crate::core::runner::run_streamed(
         cmd,
         "err",
-        command,
+        &display,
         Box::new(ErrorStreamFilter::new()),
         crate::core::runner::RunOptions::with_tee("err"),
     )
 }
 
 /// Run tests and show only failures
-pub fn run_test(command: &str, verbose: u8) -> Result<i32> {
+pub fn run_test(command: &[String], verbose: u8) -> Result<i32> {
+    let (cmd, display) = build_argv_command(command, "test")?;
     if verbose > 0 {
-        eprintln!("Running tests: {}", command);
+        eprintln!("Running tests: {}", display);
     }
-    let cmd = build_shell_command(command);
-    let command_owned = command.to_string();
+    let command_owned = display.clone();
     crate::core::runner::run_filtered(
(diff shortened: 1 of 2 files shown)


hgunduzoglu (commented):
Checked this out locally and tested it against the #2389 repro plus a few edge scenarios. Overall: the fix does what it says, the approach (argv passthrough + single-arg `shell_split`, mirroring `rtk proxy`'s #388 handling) is faithful to proxy semantics, and the test coverage is solid — the argv-level regression test and the end-to-end `exit 7` test pin down both halves of the bug.

**Verified locally (macOS arm64, develop @ 6785a6c vs this branch):**

| Scenario | rtk 0.42.3 | this PR |
|---|---|---|
| `rtk err sh -c 'exit 7'` | exit 0 (bug) | **exit 7** ✓ |
| `rtk err /tmp/argspy 'a b' c` | argv split into 3 | **`argc=2`, `arg=[a b]` preserved** ✓ |
| `cargo fmt --check && cargo clippy --all-targets && cargo test` | — | clean, 2155 passed ✓ |

Two observations, neither of which I'd consider blocking:

1. **Shell-operator strings silently change meaning.** Before, `rtk err 'true && exit 3'` went through `sh -c` and exited 3; now `shell_split` produces `["true", "&&", "exit", "3"]` and the child exits 0 with the operators passed as literal args. I verified `rtk proxy 'true && exit 3'` behaves identically, so this is consistent with the parity goal (and the old behavior is exactly what masked exit codes) — but it's a silent behavioral change for anyone using `rtk err 'a && b'` / `'cmd 2>&1'` today. Worth a line in the PR description or changelog; alternatively a one-line stderr warning when a split argument is a bare shell operator (`&&`, `||`, `|`, `;`, `2>&1`) would make the failure mode discoverable instead of silent.

2. **Missing binary now exits 1 instead of 127.** `rtk err nonexistent-cmd` previously surfaced sh's 127; with direct spawn it errors out of `run_streamed` and exits 1. Same as `rtk proxy` (also 1), so consistent within rtk — just noting it since scripts occasionally branch on 127.

Nice work — especially the doc comment on `build_argv_command` explaining the single-arg vs multi-arg contract.

PR #2394, opened 2026-06-11
fix(diff): report modified-only diffs and follow diff exit convention

### LINKED ISSUE #2364: rtk diff reports differing YAML/JSON files as identical
## Summary

`rtk diff` reports "[ok] Files are identical" (exit 0) for files that differ, whenever the files have a structured-format extension. Plain-text files diff correctly. Because the hook rewrites `diff` to `rtk diff`, any agent comparing config files gets a false negative.

## Environment

rtk 0.42.3, macOS (Homebrew).

## Reproduction

```console
$ printf "a: 1\n" > one.yaml; printf "a: 2\n" > two.yaml

$ rtk diff one.yaml two.yaml
[ok] Files are identical        # wrong; exit 0

$ /usr/bin/diff one.yaml two.yaml
1c1
< a: 1
---
> a: 2                          # exit 1, correct
```

Same false "identical" result for `.yml` and `.json`:

```console
$ printf "{\"a\": 1}\n" > j1.json; printf "{\"a\": 2}\n" > j2.json
$ rtk diff j1.json j2.json
[ok] Files are identical        # wrong
```

Plain text works:

```console
$ printf "x\n" > t1.txt; printf "y\n" > t2.txt
$ rtk diff t1.txt t2.txt
t1.txt → t2.txt
   +1 added, -1 removed, ~0 modified
```

## Impact

False negatives are worse than verbose output: an agent (or person) checking whether an edit landed, or whether two configs drifted, is told there is no difference and moves on. We hit this comparing a yq-edited YAML file against its original; the wrapper hid a three-line change.

## Secondary observation

`rtk diff` exits 0 even when it does report differences (the `.txt` case above). `diff` convention is exit 1 on differences; scripts using `diff && echo same` semantics break under the rewrite.

## Expected

Structured-format files are compared like any other file, and the exit code follows diff convention (or at minimum is documented).

hgunduzoglu:
## Summary

Fixes #2364 — `rtk diff` reported differing YAML/JSON files as `[ok] Files are identical` (exit 0), hiding real changes from agents comparing config files.

## Root cause

Not extension-related: `compute_diff` classifies similar lines (character-set similarity > 0.5, e.g. `a: 1` vs `a: 2`) as **modified**, but the identical check in `run()` only tested `added == 0 && removed == 0`, ignoring the modified count. Structured files like YAML/JSON trigger this because their changed lines stay near-identical character-wise; the issue's `.txt` repro worked only because `x` vs `y` has zero similarity and lands in added/removed.

## Changes

- `src/cmds/git/diff_cmd.rs` — extracted render logic into a pure, testable `render_file_diff()`; the identical check now uses `diff.changes.is_empty()`; returns the diff-convention exit code (0 = identical, 1 = differing), also fixing the secondary observation in the issue.
- `src/main.rs` — `Commands::Diff` propagates the exit code (same `Result<i32>` pattern as `go_cmd` / `core::runner`).
- `scripts/test-all.sh` — diff smoke test updated for the exit convention (identical → ok, differing → non-zero + output contains changes).
- `README.md` — exit semantics documented in the command list.

## Testing

- TDD: 4 new regression tests written first (red on the old logic), covering the exact repro cases from the issue (`a: 1`/`a: 2` YAML, `{"a": 1}`/`{"a": 2}` JSON, identical files, added/removed `.txt`).
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — clean, 2154 passed.
- Manual verification of all repro scenarios from the issue with the built binary (output + exit codes match `/usr/bin/diff` semantics).

### Behavior change note

`rtk diff` previously always exited 0. It now exits 1 when files differ, matching `diff` convention — as requested in the issue's "Expected" section, and required by scripts using `diff && echo same` semantics under the hook rewrite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ rtk read file.rs -l aggressive  # Signatures only (strips bodies)
 rtk smart file.rs               # 2-line heuristic code summary
 rtk find "*.rs" .               # Compact find results
 rtk grep "pattern" .            # Grouped search results
-rtk diff file1 file2            # Condensed diff
+rtk diff file1 file2            # Condensed diff (exit 1 if files differ)
 ```
 
 ### Git
--- a/scripts/test-all.sh
+++ b/scripts/test-all.sh
@@ -527,7 +527,9 @@ assert_ok      "rtk discover"                 rtk discover
 
 section "Diff"
 
-assert_ok      "rtk diff two files"           rtk diff Cargo.toml LICENSE
+assert_ok       "rtk diff identical files"     rtk diff Cargo.toml Cargo.toml
+assert_fails    "rtk diff differing files"     rtk diff Cargo.toml LICENSE
+assert_contains "rtk diff shows changes"       "added" rtk diff Cargo.toml LICENSE
 
 # ── 37. Wc ────────────────────────────────────────────
 
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -5,8 +5,9 @@ use anyhow::Result;
 use std::fs;
 use std::path::Path;
 
-/// Ultra-condensed diff - only changed lines, no context
-pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<()> {
+/// Ultra-condensed diff - only changed lines, no context.
+/// Returns the diff-convention exit code: 0 if identical, 1 if files differ.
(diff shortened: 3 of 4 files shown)


aeppling:
Hey @hgunduzoglu , PR look clean and solve the issue

#### Follow-up

Merging this for the fix, but git diff will need some work to avoid breaking any signal, for now this is kind of tricky and should be cleaned

##### LGTM

Thanks for contributing !

aeppling (approved):

PR #2396, opened 2026-06-11
fix(hook): apply exclude_commands to head/tail rewrites

### LINKED ISSUE #2363: hooks.exclude_commands does not apply to head/tail rewrites
## Summary

## Environment

## Reproduction

Config:

```toml
[hooks]
```

```console

```

## Cause

## Why it matters

```console
$ head -3 package.json
{
}
[30 more lines]
```

## Expected

hgunduzoglu:
## Summary

Fixes #2363 — entries in `[hooks] exclude_commands` had no effect on the `head -N file` / `tail -n N file` rewrites, while the same entries worked for every other command.

## Root cause

Exactly as diagnosed in the issue: in `rewrite_segment_inner` (`src/discover/registry.rs`), the `head -` / `tail ` short-circuit returns a `rewrite_line_range()` result before the flow reaches the `is_excluded()` check that the regular `classify_command` path goes through. The exclusion list was never consulted on that path.

## Change

One guard added inside the short-circuit: `is_excluded(cmd_part, excluded)` → `return None` before rewriting. The check runs on `cmd_part` (trailing redirects already stripped), so `head -20 file 2>&1` is excluded consistently with the plain form.

## Testing

- TDD: 4 new unit tests written first (3 red on the old code), following the existing `exclude_commands (#243)` test conventions:
  - `head` excluded → no rewrite
  - `tail` excluded → no rewrite
  - `head` with trailing redirect excluded → no rewrite
  - `head` NOT in the exclusion list → still rewrites to `rtk read --max-lines` (no over-exclusion)
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — clean, 2154 passed.
- Manual verification with the issue's exact repro config (`exclude_commands = ["head", "tail", "rg"]`, isolated `$HOME`):

```console
No rewrite for: head -20 package.json   # was: rtk read package.json --max-lines 20

$ rtk hook check "tail -n 50 app.log"
No rewrite for: tail -n 50 app.log

$ rtk hook check "git status"
rtk git status                          # other commands unaffected
```

No doc updates needed: this makes the behavior match what `exclude_commands` already documents.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -784,6 +784,9 @@ fn rewrite_segment_inner(
     }
 
     if cmd_part.starts_with("head -") || cmd_part.starts_with("tail ") {
+        if is_excluded(cmd_part, excluded) {
+            return None;
+        }
         return rewrite_line_range(cmd_part).map(|r| format!("{}{}", r, redirect_suffix));
     }
 
@@ -3431,6 +3434,45 @@ mod tests {
     fn test_exclude_bare_anchor_ignored() {
         let excluded = vec!["^".to_string()];
         assert!(rewrite_command_no_prefixes("git status", &excluded).is_some());
+    }
+
+    // --- exclude_commands on head/tail rewrites (#2363) ---
+
+    #[test]
+    fn test_exclude_head_line_range_rewrite() {
+        let excluded = vec!["head".to_string()];
+        assert_eq!(
+            rewrite_command_no_prefixes("head -20 package.json", &excluded),
+            None
+        );
+    }
+
+    #[test]
+    fn test_exclude_tail_line_range_rewrite() {
+        let excluded = vec!["tail".to_string()];
+        assert_eq!(
+            rewrite_command_no_prefixes("tail -n 50 app.log", &excluded),
+            None
+        );
+    }
+
+    #[test]
+    fn test_head_still_rewrites_when_not_excluded() {
+        let excluded = vec!["curl".to_string()];
+        assert_eq!(
+            rewrite_command_no_prefixes("head -20 package.json", &excluded),
+            Some("rtk read package.json --max-lines 20".into())
+        );
+    }
+
+    #[test]
+    fn test_exclude_head_with_redirect_suffix() {
+        // Exclusion must apply to the command part, ignoring trailing redirects
+        let excluded = vec!["head".to_string()];
+        assert_eq!(
+            rewrite_command_no_prefixes("head -20 package.json 2>&1", &excluded),
+            None
+        );
     }
 
     #[test]


KuSh (commented):
Superseded by the review below.

src/discover/registry.rs
```diff
@@ -784,6 +784,9 @@ fn rewrite_segment_inner(
     }
 
     if cmd_part.starts_with("head -") || cmd_part.starts_with("tail ") {
+        if is_excluded(cmd_part, excluded) {
+            return None;
+        }
```
KuSh, line 789: Placement is right, and no extra normalization belongs here: entry into this branch requires `cmd_part` to start literally with `head -` or `tail `, so there is no path prefix left to strip — `/usr/bin/head -20 pkg.json` never reaches it.

The rebase needs care, though. On `develop` the surrounding block has moved and is now gated:

```rust
if context == RewriteContext::Normal
    && (cmd_part.starts_with("head -") || cmd_part.starts_with("tail "))
{
}
```

The guard has to land *inside* that block. Hoisted above the `context` check it would also suppress `PipelineFinal`, which today falls through to the classify path where exclusion is already honored — a behavior change beyond the fix.

src/discover/registry.rs
```diff
@@ -3433,6 +3436,45 @@ mod tests {
...
+
+    #[test]
+    fn test_exclude_head_with_redirect_suffix() {
+        // Exclusion must apply to the command part, ignoring trailing redirects
+        let excluded = vec!["head".to_string()];
+        assert_eq!(
+            rewrite_command_no_prefixes("head -20 package.json 2>&1", &excluded),
+            None
+        );
+    }
```
KuSh, line 3476: `compile_exclude_patterns` has two branches: a bare name is escaped and wrapped into `^name($|\s)`, while a `^`-prefixed pattern compiles raw. All four cases here use bare names, so only the escaped branch is exercised on the line-range short-circuit; the existing `^curl` test goes through the classify path instead.

Worth one more case to cover the raw-regex branch here:

```rust
#[test]
fn test_exclude_head_tail_regex_pattern() {
    let excluded = vec!["^tail -n".to_string()];
    assert_eq!(
        rewrite_command_no_prefixes("tail -n 400 /var/log/foo.log", &excluded),
        None
    );
}
```

It passes as written; it pins the behavior against a later refactor that moves the guard past `rewrite_line_range` or normalizes `cmd_part` differently.

KuSh (changes requested):
The change itself is correct and minimal. `is_excluded(cmd_part, excluded)` runs after the `strip_disabled_prefix` recursion has peeled env/sudo prefixes and after `strip_trailing_redirects`, so the check sees the same normalized string the classify path already checks for exclusion. It is a pure narrowing — every input that returned `None` before still does — and the four tests assert the right values (`head -20 package.json` → `rtk read package.json --max-lines 20` matches the live binary).

Requesting changes on two points, both small:

1. The branch no longer applies to `develop` (`mergeable: CONFLICTING`), and the rebase is not mechanical — the block this guard goes into has moved and is now gated on `RewriteContext`.
2. Only one of the two branches of `compile_exclude_patterns` is exercised on this path.

Details inline.

KuSh:
Closing as superseded by #3324, which fixes the same gap on the current base; #3324 will carry `Fixes #2363`.

This branch predates the `RewriteContext` refactor — the fast path it patches has since moved and gained a `context == RewriteContext::Normal` guard — so it's conflicting and would need a rewrite rather than a rebase. Not worth your time when the same four lines are already landable elsewhere.

This was the first PR against #2363, and it's a real gap; thanks for catching it.

PR #2398, opened 2026-06-11
fix(filter): preserve inline comment markers

### LINKED ISSUE #2385: MinimalFilter silently drops valid code lines containing // or /* anywhere in the line (contains vs starts_with)
## Summary

`MinimalFilter` uses `trimmed.contains(start)` to detect block-comment opening markers (`/*`, `//`). Because it matches anywhere in the line rather than at the start, **any code line that happens to contain `//` or `/*` as a substring** is treated as a comment and silently dropped. This includes string literals with URLs, regex patterns, and inline end-of-line comments — lines the agent needs to read the file correctly.

A second related bug: when a line contains both the opening and closing block-comment markers on the same line (e.g. `/* inline */`), the state machine sets `in_block_comment = true` then immediately resets it to `false`, but `continue` still executes and the line is dropped.

These two bugs together cause the agent to silently receive an incomplete view of source files, which manifests as the "fabricated field names" and "file content that doesn't match the real file" symptoms reported in #2360 and #2176.

## Environment

rtk 0.42.3, Linux (Claude Code `PreToolUse` hook).

## Reproduction

Create a Rust file with URL-containing lines:

```console
$ cat > /tmp/example.rs <<'RUST'
const API: &str = "http://example.com/v1";  // endpoint
let re = regex::Regex::new(r"https?://\w+").unwrap();
/* single-line block comment */
fn foo() {}
RUST
```

Run through the hook (or inspect the filter path directly):

```console
$ rtk read /tmp/example.rs --level minimal
fn foo() {}
```

Lines 1, 2, and 3 are all dropped:
- Line 1: contains `://` which matches `//`
- Line 2: contains `://` in the regex literal
- Line 3: same-line `/* ... */` still triggers `continue`

Expected output: all four lines, since none are actually inside a block comment.

## Cause

**Location**: `src/core/filter.rs`, `MinimalFilter` block-comment detection block

```rust
// Bug 1: contains instead of starts_with
if !in_docstring
    && trimmed.contains(start)   // ← should be starts_with(start)
    && !trimmed.starts_with(...)
{
    in_block_comment = true;
}

// Bug 2: same-line open+close still drops the line
if in_block_comment {
    if trimmed.contains(end) {
        in_block_comment = false;  // reset...
    }
    continue;                      // ...but continue runs unconditionally
}
```

`trimmed.contains(start)` matches `/*` or `//` anywhere in the line. Real block-comment openers only appear at the start of a trimmed line (or after `)` / `{` in languages that allow trailing block comments, which are handled separately). Changing to `starts_with` constrains the match to genuine comment openers.

The same-line fix requires checking whether the closing marker also appears on the same line before setting `in_block_comment = true` and before the unconditional `continue`.

## Impact

An agent using `rtk hook claude` reads a source file via the hook. The file is returned with field definitions, constant declarations, and function signatures missing — because those lines happened to contain `://` or `/*` as part of string literals or regex patterns. The agent has no way to detect the omission; it treats the filtered output as the complete file and makes edits or decisions based on the truncated content.

This explains the "content fabrication" symptoms in #2360 and #2176: the model isn't hallucinating, it's reasoning from an incomplete file view produced by this filter bug.

## Expected

Lines are only dropped when `//` or `/*` appear at the start of the trimmed line (i.e. the line is actually a comment). String literals, URL values, and regex patterns that happen to contain these character sequences are passed through unchanged.

xub5210-droid:
## Summary

- only enter block-comment filtering when the trimmed line starts with the block marker
- preserve valid code containing block markers in string literals or trailing comments
- add regression coverage for inline markers and standalone block comments

Fixes #2385

AI disclosure: Codex assisted with the implementation and test.

## Test plan

- [x] `cargo test --all` (2151 passed, 8 ignored)

### DIFF
--- a/src/core/filter.rs
+++ b/src/core/filter.rs
@@ -173,7 +173,7 @@ impl FilterStrategy for MinimalFilter {
             // Handle block comments
             if let (Some(start), Some(end)) = (patterns.block_start, patterns.block_end) {
                 if !in_docstring
-                    && trimmed.contains(start)
+                    && trimmed.starts_with(start)
                     && !trimmed.starts_with(patterns.doc_block_start.unwrap_or("###"))
                 {
                     in_block_comment = true;
@@ -467,6 +467,25 @@ fn main() {
         let filter = MinimalFilter;
         let result = filter.filter(code, &Language::Rust);
         assert!(!result.contains("// This is a comment"));
+        assert!(result.contains("fn main()"));
+    }
+
+    #[test]
+    fn test_minimal_filter_preserves_comment_markers_in_code() {
+        let code = r#"
+const API: &str = "http://example.com/v1"; // endpoint
+let pattern = "/* not a comment */";
+let value = 1; /* trailing comment */
+/* single-line block comment */
+fn main() {}
+"#;
+        let filter = MinimalFilter;
+        let result = filter.filter(code, &Language::Rust);
+
+        assert!(result.contains("http://example.com/v1"));
+        assert!(result.contains("\"/* not a comment */\""));
+        assert!(result.contains("let value = 1;"));
+        assert!(!result.contains("single-line block comment"));
         assert!(result.contains("fn main()"));
     }
 


KuSh:
Closing in favour of #2397, which makes the identical one line change for issue #2385, from contains to starts_with in the block comment opener, and carries wider regression coverage and a green CI run. Merging it makes this change redundant. Review and comments are very welcome on #2397, or reply here if your version covers a case it misses.

PR #2401, opened 2026-06-11
fix: keep release panic=unwind so filter catch_unwind fails open

### LINKED ISSUE #2400: rtk crashes (SIGABRT) on `gcloud compute instances list --format=…` — poisoned Once cascade
## Summary
`rtk` aborts with `SIGABRT` (Abort trap: 6) when a `gcloud … list --format=…`
command is routed through it via the Claude Code rewrite hook. The crash appears
to be a Rust panic inside a one-time initializer that then poisons a `Once`,
turning subsequent accesses into an `unreachable!()` abort. In one session it
produced **10 crash reports in a 7-second window**.

## Environment
- rtk 0.42.0 (cargo/Homebrew install)
- macOS 26.4 (25E246), Apple Silicon
- Invoked through the Claude Code hook (`~/.claude/hooks/rtk-rewrite.sh`), which
  rewrites `gcloud …` → `rtk … gcloud …`

## Repro
Route this through rtk (the hook does it automatically):

    gcloud compute instances list --project=<PROJECT> --format=<FORMAT>

Observed firing for several different `--project` values in quick succession.
Most `gcloud` calls only hit the benign arg-parse fallback (below); a subset
abort the process.

## Crash signature
- `EXC_CRASH / SIGABRT`, faulting thread is the abort path:
  `abort` → `rtk 0x5d0a14` → `0x5cd48c` → `0x5d0c3c` → `0x5c4b38` → `0x5cd498`
  → `0x36b5dc` → `0x38b0dc` (release binary, symbols stripped).
- Panic strings present in the binary:
  - `internal error: entered unreachable code: invalid Once state`
  - `threads should not terminate unexpectedly`

The "invalid Once state" path is reachable only after a `Once`/`OnceLock`/
`LazyLock` initializer has **panicked and poisoned** the cell — so the real fault
is an earlier panic inside that init (likely a regex compile or the bundled
gcloud TOML-filter load against specific `--format` output), and the 10-in-7s
burst is the poisoning cascade across the rapid-fire calls.

## Two separate issues in the logs (only #1 is the crash)
1. **Crash (this report):** SIGABRT via the filter/output path on `gcloud … list`.
   15 crash reports over a week; 13 on one day, 10 of them in a 7s burst.
2. **Benign fallback (mention for context):** rtk's arg parser logs
   `error: unrecognized subcommand 'gcloud'` and falls back to raw execution
   successfully — that part works as designed; not the bug here.

## Suggested fix
- **Fail-open the filter path:** wrap filter execution / lazy init in
  `catch_unwind` (or make the init fallible) so a panic degrades to raw
  passthrough instead of aborting — matching how arg-parse failures already fall
  back. One bad command shouldn't abort the process (or poison a global `Once`
  for the rest of it).
- **Root fix:** make the panicking initializer in the gcloud/TOML-filter path
  return `Result` and handle the offending input instead of `unwrap`/`expect`.

Happy to provide full `.ips` crash reports and the exact command timeline.

amitgaur-zengineer:
## Summary
- `[profile.release]` sets `panic = "abort"`, which makes `std::panic::catch_unwind` a no-op — so the filter fail-open recovery in `core::stream` and `cmds::system::pipe_cmd` ("filter panicked — passing through raw output") is dead code in release builds. Any panic inside a filter aborts the whole process with SIGABRT instead of degrading to raw passthrough (#2400; seen in the wild on `gcloud … list --format=…`).
- Fix: set `[profile.release] panic = "unwind"` so the existing `catch_unwind` actually fails open. Adds a manifest-level regression guard (the normal suite can't catch this — `cargo test` always builds with unwind, which is why it shipped).
- Also: `run_fallback` applied the matched TOML filter (`core::toml_filter::apply_filter`) with **no** `catch_unwind`, unlike the streaming/pipe paths. Wrapped it in a `filter_or_passthrough` helper so a filter panic there also fails open (and the default panic hook prints the location, which helps diagnose the underlying filter bug).

## Test plan
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — fmt clean, clippy clean, 2153 passed / 0 failed (stable 1.96.0)
- [x] `cargo test --test release_profile_panic` — passes; verified it FAILS when `panic` is flipped back to `abort`
- [x] `cargo test filter_or_passthrough` — new fallback fail-open tests (panic → passthrough; normal → filtered value)
- [x] `cargo test test_panicking_filter_returns_passthrough` — existing fail-open test still passes

> Trade-off: `panic = "unwind"` slightly increases binary size (landing pads); it does not affect startup time. Correctness + the fail-open behavior the codebase already implements/tests outweigh the size delta.

Closes #2400

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,12 @@ toml = "0.8"
 opt-level = 3
 lto = true
 codegen-units = 1
-panic = "abort"
+# Must stay "unwind": filter execution relies on std::panic::catch_unwind to
+# fail open (see core::stream and cmds::system::pipe_cmd — "filter panicked,
+# passing through raw output"). Under panic = "abort" that recovery is dead
+# code in release builds: any panic inside a filter aborts the whole process
+# with SIGABRT instead of degrading to raw passthrough.
+panic = "unwind"
 strip = true
 
 # cargo-deb configuration
--- a/src/main.rs
+++ b/src/main.rs
@@ -1152,6 +1152,21 @@ const RTK_META_COMMANDS: &[&str] = &[
     "rewrite",
 ];
 
+/// Run a filter closure, failing open to `raw` if it panics.
+///
+/// Mirrors the `catch_unwind` fail-open guards in `core::stream` and
+/// `cmds::system::pipe_cmd`: a filter bug should degrade to raw passthrough,
+/// never take down rtk. The default panic hook still prints the panic location
+/// to stderr before we recover, which is useful for diagnosing the filter bug.
+/// Requires `panic = "unwind"` in the release profile to be effective (see
+/// Cargo.toml / tests/release_profile_panic.rs).
+fn filter_or_passthrough<F: FnOnce() -> String>(raw: &str, filter_fn: F) -> String {
+    std::panic::catch_unwind(std::panic::AssertUnwindSafe(filter_fn)).unwrap_or_else(|_| {
+        eprintln!("[rtk] warning: filter panicked — passing through raw output");
+        raw.to_string()
+    })
+}
+
 fn run_fallback(parse_error: clap::Error) -> Result<i32> {
     let args: Vec<String> = std::env::args().skip(1).collect();
 
@@ -1229,7 +1244,9 @@ fn run_fallback(parse_error: clap::Error) -> Result<i32> {
                     None
                 };
 
-                let filtered = core::toml_filter::apply_filter(filter, &combined_raw);
+                let filtered = filter_or_passthrough(&combined_raw, || {
+                    core::toml_filter::apply_filter(filter, &combined_raw)
+                });
                 println!("{}", filtered);
                 if let Some(hint) = tee_hint {
                     println!("{}", hint);
@@ -2548,6 +2565,19 @@ mod tests {
     use super::*;
     use clap::Parser;
     use std::cell::Cell;
+
+    #[test]
+    fn test_filter_or_passthrough_recovers_from_panic() {
+        // A panicking filter must not take down rtk — it falls back to raw output.
+        let out = filter_or_passthrough("raw output", || panic!("filter bug"));
+        assert_eq!(out, "raw output");
+    }
+
+    #[test]
+    fn test_filter_or_passthrough_returns_filtered_value() {
+        let out = filter_or_passthrough("raw", || "filtered".to_string());
+        assert_eq!(out, "filtered");
+    }
 
     #[test]
     fn test_git_commit_single_message() {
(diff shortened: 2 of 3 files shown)


HetCreep:
The diagnosis here matches what I found independently in an audit of **v0.44.2 (`700bdde`)**, and this PR already covers the parts I would have raised — including that `cargo test` always builds with unwind, which is why it shipped. Two things I did not see covered, offered in case they belong in this PR or a follow-up.

*Provenance: AI-assisted source audit, human-directed; the two points below are read from source at `700bdde`.*

### 1. A fourth filter path still has no guard: `run_captured_filter`

This PR wraps `run_fallback`'s TOML filter, joining the existing guards in `core::stream` and `cmds::system::pipe_cmd`. But the **buffered command path** appears to be unguarded:

`src/core/runner.rs:130` — `run_filtered` → `run_captured_filter` calls `filter_fn(text_to_filter, exit_code)` directly, with no `catch_unwind`.

That is the path a large number of the `src/cmds/**` filters take. With `panic = "unwind"` restored, a panic there still unwinds out of `main` rather than failing open — the child has already run, so its side effects happened, but its output is discarded, nothing is teed, and no tracking row is written.

### 2. The `core::stream` guard may not be reachable in production at all

Worth double-checking before relying on it as one of the existing fail-open sites: the `catch_unwind` at `src/core/stream.rs:450` sits inside the `FilterMode::Buffered` arm, and the only constructor of `FilterMode::Buffered` I could find is at `src/core/stream.rs:729`, **inside `#[cfg(test)] mod tests`**. Every production call site I traced (`runner.rs:107,188,209`, `uv_cmd.rs:69`, `search.rs:459`, `git.rs:1337`) passes `Passthrough`, `Streaming`, or `CaptureOnly`.

If that is right, then even after this PR the streaming/buffered guard would be protecting a variant that production never constructs, and `pipe_cmd` plus the new `filter_or_passthrough` would be the only live ones. I may have missed a constructor — but if not, it changes what "the existing fail-open recovery" covers.

### Concrete case this would still hit

`src/cmds/js/prisma_cmd.rs:319-320` computes `let end = line[pos..].find(char::is_whitespace).unwrap_or(20);` then slices `line[pos..pos+end]`. When the token is last and shorter than 20 bytes, the slice runs past the end. I reproduced the panic with a verbatim extract:

```
panicked at prisma_cmd.rs: end byte index 59 is out of bounds for string of length 47
```

`filter_migrate_status` is routed from `src/main.rs:2198` through the buffered path, so under this PR it would unwind out of an unguarded frame. (Line 252 of that same file uses the correct `unwrap_or(line.len() - pos)`.)

Not asking to widen this PR — it is coherent as it stands, and the profile fix is the load-bearing part. Just flagging what I believe remains open after it lands.

Issue #2387, opened 2026-06-11
rtk grep -n drops line content & numbers (returns count, '0 files' wrong)


pigusso:
## Summary

`rtk grep` (and `grep` auto-rewritten by the Claude Code hook) discards line content and line numbers even when `-n` is passed, returning only a match count — and the file count is wrong (`0 files` while matches were found in the named file). This makes the output unusable for the common "read the matching lines" use case.

## Environment

- rtk 0.42.3
- macOS (Darwin 24.6.0), zsh
- Used via the Claude Code hook (auto-rewrite of `grep` → `rtk grep`)

## Reproduction

```console
$ rtk grep -n Zore README.md
2 matches in 0 files: [+2 more]
```

Expected: the matching lines with their line numbers (the file clearly contains the matches).
Actual: a count summary, with `0 files` despite a file being specified and matched.

### Controlled comparison

| Command | Result |
|---|---|
| `rtk grep -n Zore README.md` | `2 matches in 0 files: [+2 more]` — line numbers + content dropped, file count wrong ❌ |
| `rtk proxy grep -n Zore README.md` | correct raw output (line numbers + content) ✅ |
| `command grep -n Zore README.md` | correct raw output ✅ |

## Why this matters

`-n` is documented in rtk as "Show line numbers (always on, accepted for grep/rg compatibility)", which signals the caller wants the lines, not a count. The compact-grep filter groups/counts by file, which is great for "how many hits across the tree" but silently destroys the content the user asked for. The `0 files` count also looks like a counting bug independent of the formatting choice.

## Suggested fix (either or both)

1. When an explicit single file (or `-n`) is given, passthrough the matching lines instead of summarizing to a count.
2. Fix the file-count so it never reports `0 files` when matches exist in a named file.

## Workarounds currently used

`rtk proxy grep ...`, `command grep ...`, or the agent's own file-reading tool. Filing so structural/`-n` greps can passthrough by default.

pszymkowiak:
Closing as duplicate of #2310. See the original issue for updates.

Issue #2310, opened 2026-06-07
rtk grep (non-recursive) reports "N matches in 0 files" and hides all matched lines


CodeRod-Solutions:
Version: rtk 0.42.3 (macOS, arm64) / shell - zsh

  Repro:
  printf 'alpha one\nbeta two\nalpha three\n' > a.txt
  rtk grep -n "alpha" a.txt
  
Actual: 2 matches in 0 files: followed by [+2 more] — zero match lines printed; exit 0.
Expected: the matching lines, e.g. 1:alpha one / 3:alpha three (as rtk grep -rn, rtk proxy grep, and raw grep all correctly produce).

Notes: Only the non-recursive path is affected — adding -r produces correct output. The summarizer reports the right count but "0 files" and suppresses the preview entirely, so callers get no usable result despite exit 0.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 95% |

### Summary

Non-recursive rtk grep suppresses all match output and reports '0 files' despite correct match count — fix the non-recursive summarizer path to emit matched lines and increment file counter.

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

- `src/filters/grep.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

jaypaulb:
Reproduced on Linux (x86_64, Ubuntu, kernel 6.17) — with what appears to be the deterministic trigger for at least one path producing this symptom: **a binary file as the only match source**.

## Repro

```console
$ printf 'alpha one\x00\nbeta\nalpha three\n' > repro.bin

$ grep -n alpha repro.bin          # raw grep control
grep: repro.bin: binary file matches

$ rtk grep -n alpha repro.bin
1 matches in 0 files:

[+1 more]
```

Exit 0, no usable output, file count wrong — exactly the reported symptom.

## Key observations

[...]

PR #2405, opened 2026-06-12
fix(grep): use portable --null in system grep fallback (BSD/macOS)

vegetablechicken233:
## Problem

On macOS/BSD systems without ripgrep installed, `rtk grep` prints `N matches in 0 files:` followed by `[+N more]` — the match count is right, but zero matched lines are shown, with exit code 0. Reported in #2310 (see also #2387).

Repro on a default macOS install (no `rg` on PATH):

```bash
printf 'alpha one\nbeta two\nalpha three\n' > a.txt
rtk grep -n "alpha" a.txt
# → 2 matches in 0 files:
#   [+2 more]
```

## Root cause

When `rg` is unavailable, the fallback runs `grep -rnHZ <pattern> <path>`, relying on `-Z` to NUL-separate the filename — which is what `parse_match_line()` hard-requires (`^([^\x00]+)\x00(\d+):(.*)$`, introduced for #1436).

`-Z` only means `--null` on **GNU** grep. On **BSD/macOS** grep, `-Z` is an alias for `--decompress` (zgrep mode), so output is plain `file:line:content` with no NUL byte. Every line then fails to parse: `total_matches` counts raw lines (correct count), but `by_file` stays empty → "0 files", and all content is suppressed into `[+N more]`. The bug is invisible on Linux and silently breaks every default macOS install without ripgrep.

## Fix

Use the long option `--null`, which GNU and BSD grep both define as "print a zero-byte after the file name" (verified against both man pages). The fallback flags are extracted into a `GREP_FALLBACK_FLAGS` constant so tests can assert portability.

## Tests

- `test_grep_fallback_uses_portable_null_flag` — regression guard: asserts `--null` is used, `-Z` is not, and flags stay un-bundled (so a future `-rnHZ`-style bundle can't sneak `-Z` back in).
- `test_grep_fallback_output_is_nul_separated_and_parseable` — runs the **real system grep** with the fallback flags against a fixture and asserts the output contains NUL and parses via `parse_match_line()`. Skips gracefully if grep is absent.

Verified locally on macOS (BSD grep, Apple Silicon): `cargo fmt --all && cargo clippy --all-targets && cargo test --all` → clippy clean, 2154 tests pass, including both new tests against the actual BSD grep this bug ships on.

Related to #2310: that report says non-recursive `rtk grep` breaks while `-r` works, which suggests the reporter may have a partial-`rg` environment — but the "N matches in 0 files + hidden lines" symptom is produced exactly and only by NUL-less underlying output, and this fixes the macOS no-ripgrep case of it.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -349,7 +349,9 @@ pub fn run(
             for p in &patterns {
                 grep_cmd.args(["-e", p]);
             }
-            grep_cmd.args(["-rnHZ", "--"]);
+            // --null (not -Z): on BSD/macOS grep -Z means --decompress, not the
+            // NUL filename separator parse_match_line() needs (issue #2310).
+            grep_cmd.args(["-rnH", "--null", "--"]);
             grep_cmd.args(&paths);
             exec_capture(&mut grep_cmd)
         })
@@ -465,8 +467,9 @@ pub fn run(
 
 /// Parses a single rg/grep match line of the form `file\0line_number:content`.
 ///
-/// Requires the underlying command to be invoked with `-0` (rg) or `-Z` (grep)
-/// so the filename is NUL-separated from `line:content`. NUL cannot appear in
+/// Requires the underlying command to be invoked with `-0` (rg) or `--null`
+/// (grep) so the filename is NUL-separated from `line:content`. NUL cannot
+/// appear in
 /// file paths, so the parser is unambiguous regardless of:
 ///   - content with `:` or `::` (e.g. `ClassRegistry::init(...)`, issue #1436);
 ///   - paths with embedded `:` (Windows drive letters, weird filenames like


vegetablechicken233:
Hi! Just rebased this onto the latest `develop` after the great grep refactor in #2333 — conflicts are resolved and it's mergeable again. 🙂

Good news: the fix still applies cleanly. The `-Z` issue this addresses survived the refactor — it's still `grep_cmd.args(["-rnHZ", "--"])` at `grep_cmd.rs:352` on develop. On BSD/macOS grep, `-Z` means `--decompress` rather than the NUL separator the parser needs, so the fallback silently breaks (the "N matches in 0 files" symptom from #2310). Swapping to the portable `--null` long option fixes it on both GNU and BSD grep.

`cargo fmt` / `clippy` / `test --all` are all green locally (2197 tests), including a functional test that runs the real BSD grep this bug actually ships on.

No rush at all, but whenever you have a spare moment I'd really appreciate a look — it's a small, self-contained change. Thanks so much for maintaining rtk! @aeppling, and cc @KuSh since it touches the fallback path from #2333.

src/cmds/system/grep_cmd.rs
```diff
@@ -349,7 +359,10 @@ pub fn run(
             for p in &patterns {
                 grep_cmd.args(["-e", p]);
             }
-            grep_cmd.args(["-rnHZ", "--"]);
+            // --null (not -Z): on BSD/macOS grep -Z means --decompress, not the
+            // NUL filename separator parse_match_line() needs (issue #2310).
+            grep_cmd.args(GREP_FALLBACK_FLAGS);
+            grep_cmd.arg("--");
```
KuSh: Hi @vegetablechicken233, thanks for your PR

Just update args list with necessary comment to explain why null is absolutely needed.
```suggestion
            grep_cmd.args(["-rnH", "--null", "--"]);
```

Tests on GREP_FALLBACK_FLAGS do not bring any value.

KuSh (changes requested):

src/cmds/system/grep_cmd.rs
vegetablechicken233: Thanks for the quick review @KuSh! Applied your suggestion exactly — fallback is now `grep_cmd.args(["-rnH", "--null", "--"])` with the explanatory comment, and I've dropped the `GREP_FALLBACK_FLAGS` const and its tests. The diff is now down to the single line change plus a doc-comment touch-up (`-Z` → `--null` in the `parse_match_line` header). `fmt`/`clippy`/`test --all` still green locally. 🙏

KuSh (approved):
Thanks, LGTM!

PR #2410, opened 2026-06-12
fix: avoid false prettier success output

### LINKED ISSUE #2395: bug: `rtk prettier` prints "All files formatted correctly" when the prettier binary is not installed (and for `--version`)
## Summary

`rtk prettier` emits the success line `Prettier: All files formatted correctly` even when the `prettier` binary does not exist on the system, and also for non-check invocations like `prettier --version`. The exit code is non-zero (254), but the stdout message contradicts it.

This is a sibling of #221 (fixed): that bug was `[warn]` lines being filtered out of real prettier output. This one is the *spawn-failure / non-check* path taking the same "all good" branch.

## Environment

- rtk 0.42.3 (also reproduced on 0.39.0), installed via Homebrew, macOS (Darwin 25.4)
- A pnpm monorepo where prettier is **not** installed (no `prettier` on PATH, none in `node_modules/.bin`)

## Steps to reproduce

```console
$ command -v prettier
(not found)

$ rtk prettier --check 'src/**/*.tsx'
Prettier: All files formatted correctly
$ echo $?
254

$ rtk prettier --version
Prettier: All files formatted correctly
$ echo $?
254
```

Inconsistently, in some directories the same missing-binary situation instead yields `Error: prettier produced no output` (exit 1), which is closer to correct.

## Impact

This is especially bad in the coding-agent use case rtk is built for: the `rtk hook claude` PreToolUse hook rewrites `pnpm exec prettier --check ...` / `npx prettier ...` into `rtk prettier ...`, and agents routinely pipe output through `head`/`tail` (so the non-zero exit code is swallowed by the pipe). The agent then sees only `Prettier: All files formatted correctly` and concludes formatting was verified. In our repo (which uses a different formatter, and has no prettier at all), agents kept "verifying" formatting with prettier for weeks because rtk kept confirming success for a binary that never ran.

## Expected behavior

- If the underlying `prettier` process fails to spawn (command not found), print an explicit error (e.g. `prettier: command not found`) and never the success line.
- Only print `All files formatted correctly` when prettier actually ran a `--check`/`--list-different`-style invocation and reported clean; for other invocations (`--version`, `--help`), pass the output through.

jsiu93:
Fixes #2395.

## Summary
- keep non-check prettier invocations as passthrough
- filter only check and list-different invocations
- avoid printing the success summary when prettier or the package manager fails

## Tests
- cargo fmt --all -- --check
- cargo test prettier_cmd
- cargo test --all

### DIFF
--- a/src/cmds/js/prettier_cmd.rs
+++ b/src/cmds/js/prettier_cmd.rs
@@ -7,6 +7,7 @@ use anyhow::Result;
 
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let mut cmd = package_manager_exec("prettier");
+    let mode = PrettierMode::from_args(args);
 
     for arg in args {
         cmd.arg(arg);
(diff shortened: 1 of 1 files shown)


jsiu93:
bump. Still relevant: #2395 is open and `develop` still routes every prettier invocation (including `--version` and `--write`) through the check-mode filter, so it prints "All files formatted correctly" for non-check runs. Branch is conflict-free with current `develop`, CI green.

PR #2416, opened 2026-06-12
fix(hook): rewrite pytest under uv run

jbellis:
## Summary
Teach hook rewrite detection to preserve `uv run` for the two common pytest forms:
- `uv run pytest ...`
- `uv run python -m pytest ...`

These now rewrite to:
- `uv run rtk pytest ...`

[This covers almost 10% of the "uv run" commands that Codex has executed on my machine]

## Test plan

- [X ] `cargo fmt --all && cargo clippy --all-targets && cargo test`

Added/kept coverage for:
- `uv run pytest ...`
- `PYTHONPATH=. uv run pytest ...`
- `uv run python -m pytest ...`
- non-pytest `uv run ...` commands remain unrevised
- unsupported `uv run` option-bearing forms remain unrevised

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -471,8 +471,9 @@ fn collapse_line_continuations(s: &str) -> std::borrow::Cow<'_, str> {
 /// being run, only *how* it's run — e.g. `docker exec mycontainer`,
 /// `direnv exec .`, `poetry run`, or `bundle exec`. Stripping it lets the inner
 /// command match a filter; the prefix is then re-prepended to the rewrite. The
-/// built-in [`SHELL_PREFIX_BUILTINS`] (`noglob`, `command`, `builtin`, `exec`,
-/// `nocorrect`) are always applied in addition to user-configured prefixes.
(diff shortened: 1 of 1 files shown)


aeppling:
Hey @jbellis  !

Thanks for this PR, clean direction and code, lgtm.

Possible enhancement
 
### Generalize the allowlist
 
Right now strip_uv_run_pytest_prefix hardcodes the two pytest forms. But uv run is conceptually the same as the wrappers RTK already strips generically (poetry run, bundle exec, docker exec, the doc comment in this file even lists them): strip the prefix, recurse, re-prepend if the inner command rewrote.

If uv run is handled by that same mechanism instead of a pytest-specific helper, every tool RTK already filters gets covered under uv run for free

aeppling (approved):

aeppling:
Hey @jbellis  ready to merge , just waiting in case you want to go a little further with the review.

Waiting for your call on this !

jbellis:
Wow, apologies for missing this, that's a much better solution. Force-pushed rather than cluttering the history.

aeppling:
@jbellis  All look good to me

Thanks for this clean contribution !

Will be available in next release, should land by the end of the week

PR #2418, opened 2026-06-12
refactor(js): decouple vitest/jest runner from CLI Commands enum

aeppling:
run_test(&Commands) : run_vitest/run_jest over a shared helper, fix tracking encode and doc


HenriqueSchroeder:
Heads up: #1374 (Bun support) carries the same decoupling of `run_test` from the `Commands` enum, shaped as `run_test(TestFramework, ...)` instead of `run_vitest`/`run_jest` with a `&str` framework. It is not optional there, the `bunx` and `bun run` routing reaches `vitest_cmd` from outside the `Commands` enum.

Whichever lands first, the other needs a rebase on that file.

PR #2426, opened 2026-06-13
fix: align OpenCode and OpenClaw TypeScript plugins with Pi contract

Andrei-hub11:
## Summary

  Align the OpenCode and OpenClaw TypeScript plugins with the Pi rewrite contract.

  ## Changes

  - replace `which rtk` with `rtk --version`
  - enforce minimum supported version (`>= 0.23.0`)
  - respect `RTK_DISABLED=1`
  - skip commands already starting with `rtk `
  - only accept rewrites on exit code `0` or `3`
  - switch OpenClaw from `execFileSync` to `spawnSync` so exit code `3` is handled correctly
  - improve OpenClaw probe warnings for unexpected failures

  ## Why

  `which rtk` is a weaker probe here:
  - it depends on the runtime environment exposing `which`
  - it does not verify that the resolved `rtk` supports `rtk rewrite`
  - it does not let us enforce the minimum supported version

  `rtk --version` checks the actual binary we want to use and lets us validate compatibility in the same step.

## Test plan
I wrote and verified some unit tests locally, but left them out to keep things in line with the other TypeScript hooks. Can add them too if that makes sense.

- [X] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [X] Manual testing: `rtk <command>` output inspected

### DIFF
--- a/hooks/opencode/rtk.ts
+++ b/hooks/opencode/rtk.ts
@@ -1,5 +1,7 @@
 import type { Plugin } from "@opencode-ai/plugin"
 
+const MIN_SUPPORTED_RTK_MINOR = 23
+
 // RTK OpenCode plugin — rewrites commands to use rtk for token savings.
 // Requires: rtk >= 0.23.0 in PATH.
 //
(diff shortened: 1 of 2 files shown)


Andrei-hub11:
bump

PR #2427, opened 2026-06-13
fix(test): surface error lines for generic runners

### LINKED ISSUE #2420: rtk test shows only the last 5 output lines (not failures) for unrecognized runners
## Bug

## Steps to reproduce

```bash
cat > /tmp/vt.sh <<'EOF'
#!/bin/sh
echo "FAIL: test_login broke"
exit 1
EOF
chmod +x /tmp/vt.sh
```

**Actual:**
```
OUTPUT (last 5 lines):
  PASS: case_46
  PASS: case_47
  PASS: case_48
  PASS: case_49
  PASS: case_50
```

## Why it matters

## Notes

## Environment

axelray-dev:
## What

Fixes #2420

For unrecognized test runners (anything besides cargo/pytest/jest/go), `extract_test_summary()` previously showed only the last 5 lines of output. If failures appeared earlier in the output, they were hidden.

Now it scans all lines against the existing `ERROR_PATTERNS` regexes (errors, failures, panics, tracebacks, etc.) and surfaces matching lines under a `[FAIL] ERRORS:` header. When no error lines are found, the original "last 5 lines" fallback is preserved.

Follow-up: error-pattern scanning is now gated on non-zero exit code. `ERROR_PATTERNS` are substring-based and can match benign summary lines (e.g. "Summary: 0 failed, 10 passed") in passing runners. Passing runners (exit 0) always get the tail fallback regardless of keyword matches.

## Changes

- `src/cmds/rust/runner.rs`: In `extract_test_summary()`, the generic runner fallback now applies `ERROR_PATTERNS` matching before falling back to tail output, but only when the exit code is non-zero. Switched `run_test` from `run_filtered` to `run_filtered_with_exit` to receive the exit code. Added 3 tests covering error surfacing, no-error fallback, and the false-positive case.

## Tests

```
```

All 2153 tests pass. The three new tests verify:
1. Generic runner output with error lines at the top surfaces those errors instead of the last 5 lines
2. Generic runner output with no errors preserves the existing last-5-lines behavior
3. Passing runner (exit 0) with "failed" keyword in output shows last 5 lines, not `[FAIL] ERRORS:`

## Notes

- Only the generic runner fallback path is changed. Cargo, pytest, jest, and go runners are unaffected.
- Uses the existing `ERROR_PATTERNS` lazy_static, no new patterns added.
- DCO signed.

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -267,11 +267,35 @@ fn extract_test_summary(output: &str, command: &str) -> String {
             output.push_str(&format!("  {}\n", r));
         }
     } else {
-        output.push_str("OUTPUT (last 5 lines):\n");
-        let start = lines.len().saturating_sub(5);
-        for line in &lines[start..] {
-            if !line.trim().is_empty() {
+        let is_generic = !is_cargo && !is_pytest && !is_jest && !is_go;
+        let error_lines: Vec<&str> = if is_generic {
+            lines
+                .iter()
+                .copied()
+                .filter(|line| ERROR_PATTERNS.iter().any(|p| p.is_match(line)))
+                .collect()
+        } else {
+            Vec::new()
+        };
+
+        if !error_lines.is_empty() {
+            output.push_str("[FAIL] ERRORS:\n");
+            for line in error_lines.iter().take(MAX_RUNNER_LINES) {
                 output.push_str(&format!("  {}\n", line));
+            }
+            if error_lines.len() > MAX_RUNNER_LINES {
+                output.push_str(&format!(
+                    "  ... +{} more\n",
+                    error_lines.len() - MAX_RUNNER_LINES
+                ));
+            }
+        } else {
+            output.push_str("OUTPUT (last 5 lines):\n");
+            let start = lines.len().saturating_sub(5);
+            for line in &lines[start..] {
+                if !line.trim().is_empty() {
+                    output.push_str(&format!("  {}\n", line));
+                }
             }
         }
     }
@@ -289,5 +313,42 @@ mod tests {
         let filtered = filter_errors(output);
         assert!(filtered.contains("error"));
         assert!(!filtered.contains("info"));
+    }
+
+    #[test]
+    fn test_extract_generic_runner_shows_error_lines_before_tail() {
+        let output = "\
+error: setup failed before tests ran
+line 1
+line 2
+line 3
+line 4
+line 5
+line 6";
+
+        let summary = extract_test_summary(output, "custom-test-runner");
+
+        assert!(summary.contains("[FAIL] ERRORS:"));
+        assert!(summary.contains("error: setup failed before tests ran"));
+        assert!(!summary.contains("OUTPUT (last 5 lines):"));
+        assert!(!summary.contains("line 6"));
+    }
+
+    #[test]
+    fn test_extract_generic_runner_without_errors_keeps_last_five_lines() {
+        let output = "\
+line 1
+line 2
+line 3
+line 4
+line 5
+line 6";
+
+        let summary = extract_test_summary(output, "custom-test-runner");
+
+        assert!(summary.contains("OUTPUT (last 5 lines):"));
+        assert!(!summary.contains("line 1"));
+        assert!(summary.contains("line 2"));
+        assert!(summary.contains("line 6"));
     }
 }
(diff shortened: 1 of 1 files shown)


hgunduzoglu (commented):
Checked this out and tested locally (macOS arm64). The core behavior is right and the truncation hint (`... +N more`) is a nice touch:

| generic runner output | before | this PR |
|---|---|---|
| error at top, then 6 passing lines | last 5 lines (error hidden) | `[FAIL] ERRORS:` surfaces the error ✓ |
| no error lines | last 5 lines | last 5 lines ✓ (fallback preserved) |

`cargo fmt/clippy/test` clean, 16 runner tests pass. Scoping it to `is_generic` (leaving cargo/pytest/jest/go untouched) is the right call.

**One real edge case worth handling: false positives on passing runs.** `ERROR_PATTERNS` is substring-based and broad — `r"(?i)^.*failed.*$"`, `.*failure.*`, `.*warning.*`, `.*exception.*` match those words *anywhere* on a line, including benign summary lines. So a runner that **passes** (exit 0) but prints a summary surfaces a misleading FAIL header:

```console
$ rtk test ./runner.sh     # script prints "Summary: 0 failed, 10 passed", exit 0
[FAIL] ERRORS:
  Summary: 0 failed, 10 passed
```

An agent reading `[FAIL] ERRORS: … 0 failed, 10 passed` would conclude the suite failed when it passed — the opposite of the accurate-signal goal. Before this PR that same run showed a neutral `OUTPUT (last 5 lines):`, so it's a behavior change specifically in the passing case.

Since #2420 is really about surfacing failures *when the runner fails*, gating the error-surfacing on a **non-zero exit code** would fix both the original issue and this false positive. `extract_test_summary` doesn't currently receive the exit code, but `runner.rs` already has a `RunMode::FilteredWithExit` variant, so it can be threaded through without much churn — then: non-zero exit → `[FAIL] ERRORS:` (with the pattern scan), exit 0 → the existing tail fallback.

Core fix is solid; this is the one thing I'd tighten before merge. Could also add a test for the exit-0-with-"failed"-in-summary case to lock it down.

axelray-dev:
Good catch, thanks. Pushed a fix that gates the error-pattern scan on non-zero exit code. Now uses `run_filtered_with_exit` so the exit code is threaded through to `extract_test_summary` -- passing runners (exit 0) always get the tail fallback regardless of keyword matches in the output.

Added a test for the exact false-positive case you described (exit 0 with "0 failed, 10 passed" in output). All 2153 tests pass.

luccinmasirika:
Thanks @axelray-dev for picking this one up, and good call gating the scan on the exit code.

I pulled the branch and ran the exact repro from the issue (script prints `FAIL: test_login broke`, then 50 `PASS:` lines, exits 1). It still comes back as:

```
OUTPUT (last 5 lines):
  PASS: case_46
  ...
  PASS: case_50
```

So the `FAIL:` line is still hidden. The scan itself works, but `ERROR_PATTERNS` only has `failed` and `failure`, no bare `fail`/`FAIL`, so a line like `FAIL: test_login broke` never matches and falls through to the tail. Easy to confirm: changing the repro to print `FAILED:` surfaces it under `[FAIL] ERRORS:`, but `FAIL:` doesn't.

Adding one pattern covers it:

```rust
Regex::new(r"(?i)^.*\bfail\b.*$").unwrap(),
```

With that, the repro surfaces `FAIL: test_login broke` on exit 1, and since the scan is still gated on a non-zero exit, the passing `0 failed, 10 passed` case stays on the tail fallback. Runner tests stay green. Might be worth a test using the literal repro line so this stays locked down.

src/cmds/rust/runner.rs
luccinmasirika, line 21: Concretely, this is the one-line addition that covers it. Putting it just above
  the `failed` pattern keeps that one intact:

  ```suggestion
          Regex::new(r"(?i)^.*\bfail\b.*$").unwrap(),
          Regex::new(r"(?i)^.*failed.*$").unwrap(),
  ```

axelray-dev:
Pushed the word-boundary pattern as suggested. Added a regression test using the exact repro line ("FAIL: something went wrong" with exit 1) to lock it down. All runner tests pass.

KuSh:
Closing in favour of #2683, which fixes the same generic runner fallback in extract_test_summary for the same issue and keeps the shared error pattern list untouched. Review and comments are welcome there. If part of the scope here is not covered by #2683, say so on that PR or reply here.

PR #2434, opened 2026-06-14
fix(grep): prevent silent fallback to system grep when rg flags precede pattern

### LINKED ISSUE #2120: rtk grep falls back to native grep when --fixed-strings is placed before the pattern/path

### LINKED ISSUE #2167: rtk grep calls system grep instead of ripgrep — breaks -g/--glob flags

vlasky:
## Summary

- Add explicit clap definitions for commonly-used ripgrep flags (`-w`, `-i`, `-F`/`--fixed-strings`, `--pcre2`, `-g`/`--glob`) so they parse correctly regardless of position relative to the pattern argument, preventing clap rejection and silent fallback to system grep.
- Skip BRE `\|` to `|` translation when `--fixed-strings` is active (pattern should be literal).
- Block grep fallback when rg-only flags are present (`--pcre2`, `--glob`), letting the error propagate so the LLM knows rg is missing rather than silently getting wrong results.
- Translate long-form flags to grep-compatible short forms when fallback does activate.
- Emit a stderr warning when the grep fallback fires.

## Problem

When AI coding assistants invoke `rg -w "class" .`, the hook rewrites it to `rtk grep -w "class" .`. Clap did not recognize `-w`/`-i`/`-F`/`--pcre2`/`-g` and treated them as parse errors, triggering `run_fallback()` which silently executed system grep instead of ripgrep. The LLM believed it was calling ripgrep (with its output format, flag semantics, and default recursion) but received GNU grep output instead, causing confusion and incorrect conclusions.

## Related Issues

- Fixes #2120 - `rtk grep` falls back to native grep when `--fixed-strings` is placed before pattern/path
- Fixes #2167 - `rtk grep` calls system grep instead of ripgrep, breaks `-g`/`--glob` flags
- Partially addresses #2064 - grep fallback causes token expansion via system grep descending node_modules (this PR prevents the fallback from triggering for the common case of flags-before-pattern)

## Test plan

- [x] 9 new unit tests covering flag parsing (single flags, combined `-wi`, default path, flags after pattern)
- [x] 5 new unit tests covering grep fallback behavior (blocked by rg-only flags, allowed for compatible flags, long-form translation)
- [x] 2 new unit tests for BRE translation skip with `--fixed-strings`/`-F`
- [x] Full test suite passes (2166 tests)
- [x] Manual testing: `rtk grep -w "class" .`, `rtk grep --pcre2 "(?<=fn )\w+" .`, `rtk grep -g "*.rs" "fn run" .`, `rtk grep --fixed-strings "fn run(" .`

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -25,8 +25,15 @@ pub fn run(
         eprintln!("grep: '{}' in {}", pattern, path);
     }
 
-    // Fix: convert BRE alternation \| → | for rg (which uses PCRE-style regex)
-    let rg_pattern = pattern.replace(r"\|", "|");
+    let is_fixed_strings = extra_args.iter().any(|a| a == "--fixed-strings" || a == "-F");
+
+    // Convert BRE alternation \| → | for rg (which uses PCRE-style regex),
+    // but skip when --fixed-strings is active (pattern is literal).
+    let rg_pattern = if is_fixed_strings {
+        pattern.to_string()
+    } else {
+        pattern.replace(r"\|", "|")
+    };
 
     let mut rg_cmd = resolved_command("rg");
     // --no-ignore-vcs: match grep -r behavior (don't skip .gitignore'd files).
@@ -50,11 +57,34 @@ pub fn run(
         rg_cmd.arg(arg);
     }
 
+    let has_rg_only_flags = extra_args.iter().any(|a| {
+        a == "--pcre2" || a == "--glob" || a == "-g" || a.starts_with("--glob=")
+    });
+
     let result = exec_capture(&mut rg_cmd)
-        .or_else(|_| {
+        .or_else(|e| {
+            if has_rg_only_flags {
+                return Err(e);
+            }
+            eprintln!("rtk: rg not found, falling back to grep (install ripgrep for best results)");
             let mut grep_cmd = resolved_command("grep");
-            // When we fall back to grep, include all args, not just -rnHZ.
-            grep_cmd.args(["-rnHZ", pattern, path]).args(extra_args);
+            grep_cmd.args(["-rnHZ", pattern, path]);
+            for arg in extra_args {
+                match arg.as_str() {
+                    "--fixed-strings" => {
+                        grep_cmd.arg("-F");
+                    }
+                    "--word-regexp" => {
+                        grep_cmd.arg("-w");
+                    }
+                    "--ignore-case" => {
+                        grep_cmd.arg("-i");
+                    }
+                    _ => {
+                        grep_cmd.arg(arg);
+                    }
+                }
+            }
             exec_capture(&mut grep_cmd)
         })
         .context("grep/rg failed")?;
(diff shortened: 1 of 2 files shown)


KuSh:
Thank you for the detailed writeup, it described the failure mode exactly. This was fixed a different way in #2641 (merged 2026-06-26), which replaced grep_cmd.rs with a search module that runs whichever engine was actually invoked, passing its flags through untouched, so grep is never substituted for rg and flags before the pattern parse fine. On current develop rtk rg -w, rtk rg -g '*.txt' and rtk rg --pcre2 all run ripgrep as expected, and issues #2120, #2167 and #2064 were closed against that change. The fixed-strings and flag-position cases your patch handled are covered there, so the clap definitions and the fallback blocking are no longer needed. If something in this area still misbehaves on the latest develop, please comment here or reopen.

PR #2438, opened 2026-06-14
fix(grep): report displayed counts when max cap is applied

2lll5:
## Summary
- render the grep result body before the header so the header can report displayed counts
- count only files with emitted matches when `--max` stops output early
- keep the overflow footer based on the raw match total

Fixes charles-openclaw/charles-microbounties#1480.

## Verification
- `cargo test grep_cmd::tests::test_grep_header_reports_displayed_matches_when_capped`
- `cargo clippy --all-targets -- -D warnings && cargo test --all`

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -118,36 +118,7 @@ pub fn run(
         by_file.entry(file).or_default().push((line_num, cleaned));
     }
 
-    let mut rtk_output = String::new();
-    rtk_output.push_str(&format!(
-        "{} matches in {} files:\n\n",
-        total_matches,
-        by_file.len()
-    ));
-
-    let mut shown = 0;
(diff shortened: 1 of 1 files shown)


2lll5:
CLA is now signed, and I synced this branch with the latest `develop` to resolve the merge conflict.

Verification:
- `cargo fmt --all`
- `cargo test --all`

All checks passed locally.

KuSh:
Thanks for digging into the count mismatch. The grep path this patch touches was replaced on develop by #2641 (merged 2026-06-26), which moved the logic into src/cmds/system/search.rs and made the overflow explicit: each capped file gets a '+N more in <file>' line with a recovery hint, and skipped files get a '+N more files' footer, so the header total now reconciles with what is displayed. src/cmds/system/grep_cmd.rs no longer exists, so this branch can no longer apply. Closing as covered, but please comment or reopen if you still see counts that cannot be reconciled on current develop.

PR #2450, opened 2026-06-15
fix(diff): preserve POSIX exit codes

### LINKED ISSUE #2446: rtk diff exits 0 when files differ, breaking diff conditionals
## Summary

`rtk diff` returns exit code `0` when files differ. This breaks any command that relies on `diff`'s exit status for conditional logic (`&&`, `||`, `if`, `$?`).

## Environment

- RTK version: `0.42.0`
- macOS (Darwin)
- Shell: zsh

## Expected behavior (POSIX diff)

| Scenario | Expected exit code |
|---|---|
| Files identical | `0` |
| Files differ | `1` |
| Error (missing file, etc.) | `2` |

## Actual behavior (`rtk diff`)

| Scenario | Actual exit code |
|---|---|
| Files identical | `0` ✓ |
| Files differ | `0` ✗ (should be `1`) |
| Missing file | `1` ✗ (should be `2`) |

## Reproduction

```bash
printf 'a\n' > /tmp/a.txt
printf 'b\n' > /tmp/b.txt

# Files differ — rtk diff exits 0 (wrong)
echo $?   # prints 0, should print 1

# This conditional fires incorrectly:
rtk diff /tmp/a.txt /tmp/b.txt && echo "IDENTICAL"
# Output: shows diff correctly, then prints "IDENTICAL" — wrong
```

## Why it matters

When Claude Code or Codex runs `diff fileA fileB && echo "same"`, the RTK hook rewrites it to `rtk diff fileA fileB && echo "same"`. Because `rtk diff` always exits 0, the `&&` branch fires even when files differ — producing silently wrong results.

This affects any LLM-generated Bash that uses `diff` as a conditional (equality checks, config comparisons, audit scripts). The visual output is correct; only the exit code is wrong.

## Workaround

Use `/usr/bin/diff` explicitly (bypasses RTK rewrite) or the `if diff` form (RTK does not rewrite the `if`-conditional form):

```bash
/usr/bin/diff fileA fileB && echo "same"       # bypasses rewrite entirely
if diff fileA fileB; then echo "same"; fi      # RTK passes this through unchanged
```

## Desired fix

Preserve the compact RTK diff output format but return POSIX-compatible exit codes:
- `0` when files are identical
- `1` when files differ
- `2` on error

These are orthogonal — the display format and the exit code can be fixed independently.

cancaries:
Fixes rtk-ai/rtk#2446.

## Summary

- Preserve `rtk diff` compact output while returning POSIX-compatible exit codes.
- Return `0` when compared files are identical.
- Return `1` when compared files differ.
- Return `2` when either input file cannot be read.
- Add focused regression tests for all three exit-code paths.

## Verification

- `git diff --check`

I could not run `cargo test diff_cmd` locally because this Windows environment does not have a Rust toolchain available.

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -14,8 +14,20 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
         eprintln!("Comparing: {} vs {}", file1.display(), file2.display());
     }
 
-    let content1 = fs::read_to_string(file1)?;
-    let content2 = fs::read_to_string(file2)?;
+    let content1 = match fs::read_to_string(file1) {
+        Ok(content) => content,
+        Err(err) => {
+            eprintln!("rtk diff: failed to read {}: {}", file1.display(), err);
+            return Ok(2);
+        }
+    };
+    let content2 = match fs::read_to_string(file2) {
+        Ok(content) => content,
+        Err(err) => {
+            eprintln!("rtk diff: failed to read {}: {}", file2.display(), err);
+            return Ok(2);
+        }
+    };
     let raw = format!("{}\n---\n{}", content1, content2);
 
     let (rtk, exit_code) = render_file_diff(file1, file2, &content1, &content2);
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as a duplicate of #2473, which makes the same fix: a read failure on either operand returns POSIX diff's exit 2 instead of collapsing into exit 1.

Confirmed still live on 0.46.0 — `rtk diff x /nope; echo $?` prints `1`. The other half of #2446 (differ → exit 1) is already fixed; this is what's left.

#2473 was picked because its tests exercise the `read_operands` helper directly. The three tests here call `run()`, which goes through `timer.track(...)` and writes into the machine's real tracking database — the problem #3758 is currently fixing. Your `test_run_returns_zero_for_identical_files` / `..._one_for_different_files` pair is good coverage of the full contract, though; it's the kind of thing that belongs in `tests/guard_integration_test.rs` as a real-process test, which nothing currently has. If you'd like to submit that separately it would fill a genuine gap.

PR #2460, opened 2026-06-16
fix(grep): keep rg on the ripgrep path — split the rg/grep rewrite rule

alex-mextner:
## Summary

The single rewrite rule `^(rg|grep)\s+` in `src/discover/rules.rs` collapses **both**
tools onto `rtk grep`. `rtk grep` runs ripgrep as its backend but applies
**grep-flavored** argument translation (`src/cmds/system/grep_cmd.rs`): it strips
`-r`/`-R`/`--recursive`, rewrites BRE `\|` to PCRE `|`, defaults empty paths to `.`,
and forwards only a curated allowlist of rg flags. That translation is correct for
`grep`, but it silently corrupts genuine **ripgrep** invocations:

- ripgrep-only flags (`--hidden`, `-g`/`--glob`, `--stats`, `-uu`, `-P`, `--json`)
  are dropped or rejected;
- the short-flag collision is decisive — `-r` is *recursive* in grep but `--replace`
  in ripgrep, so a grep-flavored wrapper that strips `-r` changes what `rg` means;
- path-less `rg PATTERN` (ripgrep recurses CWD) is mishandled.

This PR keeps each tool on its own backend by **splitting the one rule into two**:
`grep` keeps routing to `rtk grep` (unchanged), and `rg` now routes to `rtk rg`.
`rtk rg` is intentionally **not** a Clap subcommand, so it falls through
`run_fallback` in `main.rs`, which executes the real `rg` binary with every argument
forwarded **verbatim** (still metrics-tracked). No crossover in either direction;
ripgrep semantics stay faithful.

This is the **minimal, conflict-free** version of what the larger #2183 attempted.
It does not change `grep`'s backend and adds no new filtered subcommand, so it avoids
the conflicts/pushback #2183 hit while still resolving the `rg` correctness issues.
A real filtered `rtk rg` subcommand can come later as a separate, additive change.

### Changes

**`src/discover/rules.rs`**

- Split the `^(rg|grep)\s+` rule into `^rg\s+` → `rtk rg` and `^grep\s+` → `rtk grep`.
- The new `rg` rule carries `savings_pct: 0.0` because `rtk rg` is a verbatim
  passthrough (no output filtering). Keeping the inherited 75% would make
  `rtk discover` falsely report `rg` as a high-value missed optimization.

**`src/discover/registry.rs`**

- Updated `test_rewrite_rg_pattern` to expect `rtk rg` (was `rtk grep`).
- Added tests: `rg` and `grep` each route to their own target; ripgrep-only flags,
  flag ordering, the `-r`=`--replace` case, and path-less `rg PATTERN` are forwarded
  verbatim to `rtk rg`; classify reports `rtk rg` at 0% and `rtk grep` at 75%.

**Docs** (`README.md`, `docs/usage/FEATURES.md`)

- Updated the auto-rewrite table and the `rtk grep` section so `rg` is documented as
  rewriting to `rtk rg` (verbatim ripgrep passthrough), not `rtk grep`.

### Issues fixed

| Issue | Title |
|-------|-------|
| #1824 | rg auto-rewrite to rtk grep breaks normal ripgrep flag ordering (`--glob`, `-g`, `--version`) |
| #2167 | rtk grep calls system grep instead of ripgrep — breaks -g/--glob flags |
| #2301 | `rtk hook claude` rewrites `rg` → `rtk grep` so path-less ripgrep searches silently return 0 matches / hang on stdin |
| #2338 | rtk rewrite should keep rg on the ripgrep path |

### Known limitations / follow-ups (out of scope here)

`rtk rg` is intentionally **not** a Clap subcommand — it relies on `run_fallback`,
which is the generic path for any unknown `rtk <cmd>`. That path has two pre-existing
warts that now also apply to the (newly intentional) `rg` route:

1. When a `.rtk/filters.toml` exists and is untrusted, `run_fallback` prints
   `[rtk] WARNING: untrusted project filters …` (and runs TOML-filter lookup) before
   the ripgrep output — so it is not byte-for-byte clean passthrough in that case.
2. A successful fallback run records `record_parse_failure_silent(.., true)`, so
   legitimate `rg` usage shows up under `rtk gain --failures` as a parse failure.

These are broader `run_fallback` behaviors (true for every unknown command today), not
needed to fix the `rg` correctness bugs this PR targets. The minimal, safe follow-up is
to make declared passthrough tools first-class — e.g. special-case `args[0] == "rg"` in
`run_fallback` to skip the TOML/trust-warning path and the parse-failure accounting — or
add a real filtered `rtk rg` subcommand (the #2183 direction). I kept that out of this PR
to stay minimal and conflict-free; happy to send it as a focused follow-up.

### Relation to #2183

#2183 ("respect the invoked tool — grep runs grep, rg runs ripgrep") is still OPEN
and CONFLICTING. It took a bigger approach: a real filtered `rtk rg` subcommand **and**
switching `rtk grep` to system grep. This PR does **not** supersede that design wholesale —
it lands the smallest slice that fixes the open `rg` issues (the rewrite-rule split)
without touching `grep`'s backend or adding a subcommand, so it rebases cleanly on the
current `develop` and stays in scope. If maintainers prefer #2183's larger direction,
this can be closed; otherwise this unblocks the `rg` correctness issues now.

## Test plan

- [x] `cargo test --all` — 2198 passed (the 2 local failures in
  `hooks::rewrite_cmd::tests::unattestable_passthrough::{test_plain_command_still_rewrites,
  test_fd_dup_redirect_still_rewrites}` are pre-existing on `develop` and depend on the
  developer's local Claude Code permission allow-list for `git status`; with a clean
  `$HOME` — as on CI — all of them pass, and they are unrelated to this diff)
- [x] Manual testing (built binary):
  ```
  rtk rewrite "rg --hidden -g '*.rs' --stats foo"  ->  rtk rg --hidden -g '*.rs' --stats foo
  rtk rewrite "rg -r 'NEW' pattern"                ->  rtk rg -r 'NEW' pattern
  rtk rewrite "rg TODO"                            ->  rtk rg TODO
  rtk rewrite "grep -rn foo src/"                  ->  rtk grep -rn foo src/
  rtk rg --glob '*.rs' 'pub fn run' src/discover/  ->  runs real ripgrep, matches (was: grep: unrecognized option --glob)
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ rtk filters and compresses command outputs before they reach your LLM context. S
 |-----------|-----------|----------|-----|---------|
 | `ls` / `tree` | 10x | 2,000 | 400 | -80% |
 | `cat` / `read` | 20x | 40,000 | 12,000 | -70% |
-| `grep` / `rg` | 8x | 16,000 | 3,200 | -80% |
+| `grep` | 8x | 16,000 | 3,200 | -80% |
 | `git status` | 10x | 3,000 | 600 | -80% |
 | `git diff` | 5x | 10,000 | 2,500 | -75% |
 | `git log` | 5x | 2,500 | 500 | -80% |
@@ -55,6 +55,11 @@ rtk filters and compresses command outputs before they reach your LLM context. S
 | **Total** | | **~118,000** | **~23,900** | **-80%** |
 
 > Estimates based on medium-sized TypeScript/Rust projects. Actual savings vary by project size.
+
+**Note on `rg`:** `rg` (ripgrep) is rewritten to `rtk rg`, a verbatim passthrough that forwards
+every argument to ripgrep and preserves its exact semantics. It is intentionally absent from the
+table above — as a passthrough it does not filter output, so it yields **no token savings**.
+Ripgrep's output is already compact, so preserving it verbatim is preferred over grouping.
 
 ## Installation
 
@@ -120,7 +125,7 @@ git status  # Automatically rewritten to rtk git status
 
 Hook-based agents rewrite Bash commands (e.g., `git status` -> `rtk git status`) before execution. Plugin-based agents, including Hermes, use their plugin API to rewrite commands before execution. The agent receives compact output without needing to call `rtk` explicitly.
 
-**Important:** the hook only runs on Bash tool calls. Claude Code built-in tools like `Read`, `Grep`, and `Glob` do not pass through the Bash hook, so they are not auto-rewritten. To get RTK's compact output for those workflows, use shell commands (`cat`/`head`/`tail`, `rg`/`grep`, `find`) or call `rtk read`, `rtk grep`, or `rtk find` directly.
+**Important:** the hook only runs on Bash tool calls. Claude Code built-in tools like `Read`, `Grep`, and `Glob` do not pass through the Bash hook, so they are not auto-rewritten. To get RTK's compact output for those workflows, use shell commands (`cat`/`head`/`tail`, `grep`, `rg`, `find`) or call `rtk read`, `rtk grep`, `rtk rg`, or `rtk find` directly. (`rtk rg` is the verbatim passthrough noted above.)
 
 ## How It Works
 
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -202,7 +202,12 @@ Supporte a la fois la syntaxe RTK et la syntaxe native `find` (`-name`, `-type`,
 
 ### `rtk grep` -- Recherche dans le contenu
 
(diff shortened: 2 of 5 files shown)


KuSh:
Thanks for spotting that grep-flavoured translation was corrupting real ripgrep invocations. #2641, merged on 2026-06-26, took the same split further: develop routes ^rg\s+ to rtk rg and ^grep\s+ to rtk grep as two separate rules, and rtk rg is now a real subcommand that runs the invoked engine with its own flags instead of substituting one tool for the other, so ripgrep-only flags and the -r meaning --replace both survive. The registry test and the FEATURES documentation were updated in the same change. The 0 percent savings value this PR gave the rg rule no longer applies, since rtk rg does filter its output now. Closing as covered, but please comment or reopen if a case here was missed.

PR #2464, opened 2026-06-16
feat: Add Mistral Vibe support with pre_tool hook

### LINKED ISSUE #800: feat: add transparent hook support for Mistral Vibe (BeforeTool)
## Context

`rtk init --agent vibe` is already implemented and installs:
- `~/.vibe/prompts/rtk.md` — system prompt instructing Vibe to use `rtk` commands
- Patches `~/.vibe/config.toml` — sets `system_prompt_id = "rtk"` and adds `rtk` to bash allowlist

## Problem

Vibe currently has **no hook mechanism** to intercept bash commands before execution (unlike Claude Code's `PreToolUse` and Gemini's `BeforeTool`). This means RTK integration relies on the model following the system prompt — which is less reliable than a real hook.

We opened a feature request on the Vibe repo to add this: **mistralai/mistral-vibe#531**

## Current State (prompt-only approach)

| What | Status |
|------|--------|
| `rtk init --agent vibe` | ✅ Implemented |
| System prompt (`~/.vibe/prompts/rtk.md`) | ✅ Installed |
| Config patch (`system_prompt_id`, allowlist) | ✅ Patched |
| Transparent hook (BeforeTool) | ❌ Blocked on mistralai/mistral-vibe#531 |

## Plan

Once mistralai/mistral-vibe#531 is implemented:
1. Add `~/.vibe/hooks/rtk-hook-vibe.sh` (rewrite bash commands to `rtk`)
2. Update `run_vibe_mode()` to also install the hook and patch `[hooks]` section in `config.toml`
3. Same pattern as the existing Gemini hook (`hooks/rtk-hook-gemini.sh`)

## References

- Gemini hook implementation: `hooks/rtk-hook-gemini.sh` + `src/init.rs::run_gemini()`
- Vibe feature request: https://github.com/mistralai/mistral-vibe/issues/531

bashout:
Implement Mistral Vibe CLI integration using the hook system made stable in Vibe 2.21.0. Closes #800

- Add hooks/vibe/ hook script and configuration
- Add Vibe installation/uninstallation functions
- Update documentation

Requires:
- Mistral Vibe >= 2.21.0

## Test plan
Manual testing verified hook functioning within the Vibe CLI
image

Tested against prompt:
> explain the most recent two commits. Dig into the code, understand the goal and evaluate the code quality and completeness of the features. 

The following rewrites were made:
```
  1. rtk git log --oneline -2 --stat
  2. rtk git show 2ed8977 --no-color
  3. rtk git show 3d37217 --no-color
  4. rtk git diff 3d37217..2ed8977 -- src/hooks/init.rs
  5. rtk git show 3d37217:src/main.rs
  6. rtk git show 3d37217 -- src/main.rs
  7. rtk grep "VIBE_HOOKS_TOML" src/hooks/constants.rs
  8. rtk git show 3d37217:hooks/vibe/hooks.toml
  9. rtk read hooks/vibe/tests/test_vibe_hook.py
 10. rtk grep vibe src/hooks/init.rs
```
Output from `rtk gain` following implementation and integration with Vibe CLI
image

All test summaries below:
```
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
test result: ok. 2468 passed; 0 failed; 8 ignored; 0 measured; 0 filtered out; finished in 0.62s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.07s
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.17s
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.91s
test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.08s
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
```


rmsc:
I believe this is circumventing the bash allowlist!

If `rtk` is allowed, then every single command passed through `rtk` is also allowed.

For instance:
```
$ vibe "run git status, and then touch file 'foo'"
```

Prompts me for permission to use `touch`, which is correct.

But this:
```
$ vibe "run git status, and then touch file 'foo' using rtk"
Touch file 'foo' completed successfully using rtk touch foo.
```

Completely circumvents the allowlist.

rmsc:
I think this is actually an upstream bug, reported here: https://github.com/mistralai/mistral-vibe/issues/849

bashout:
> I think this is actually an upstream bug, reported here: [mistralai/mistral-vibe#849](https://github.com/mistralai/mistral-vibe/issues/849)

Just want to say this is a really good spot. :) - Happy to see someone other than myself is eager for Mistral to be improved

hooks/vibe/rtk-hook-vibe.sh
```diff
@@ -0,0 +1,122 @@
...
+    IFS=. read -r MAJOR MINOR PATCH <<<"$RTK_VERSION"
+    # Require >= 0.23.0
+    if [ "$MAJOR" -eq 0 ] && [ "$MINOR" -lt 23 ]; then
+      echo "[rtk] WARNING: rtk $RTK_VERSION is too old (need >= 0.23.0). Upgrade: cargo install --git https://github.com/rtk-ai/rtk" >&2
+      exit 0
+    fi
+  fi
+  mkdir -p "$CACHE_DIR" 2>/dev/null || true
+  touch "$CACHE_FILE" 2>/dev/null || true
+fi
```
rmsc: I think we can avoid all this by just checking if rtk rewrite returns 127, in which case the rewrite command isn't supported and the version is too old.

```suggestion
```

hooks/vibe/rtk-hook-vibe.sh
```diff
@@ -0,0 +1,122 @@
...
+CMD=$(jq -r '.command // empty' <<<"$TOOL_INPUT")
+
+if [ -z "$CMD" ]; then
+  exit 0
+fi
+
+# Check for RTK_DISABLED override
+if [ "${RTK_DISABLED:-0}" = "1" ]; then
+  exit 0
+fi
```
rmsc: Should this be in the hook?

I think rtk is supposed to handle that (even if right now it doesn't seem to!): https://github.com/rtk-ai/rtk/issues/345

hooks/vibe/rtk-hook-vibe.sh
```diff
@@ -0,0 +1,122 @@
...
+
```
rmsc, line 55: I think there's a bug in rtk, where `rtk rewrite "rtk some command"` always returns rtk some command, which is what allows the LLM to bypass the allowlist (#2508).

The exit code is 3, which should signal that user approval is required. Unfortunately vibe-cli has no way to handle this at the moment.

The only thing we can do is strip rtk from the command before feeding it to rtk rewrite, like this:

```suggestion
STRIPPED_CMD=${CMD}
while [[ "$STRIPPED_CMD" == rtk\ * ]]; do
    STRIPPED_CMD="${STRIPPED_CMD#rtk }"
done
```

If the LLM tries to call rtk rm file directly, this is stripped to rm file and then fed to rtk rewrite, which passes it through.

If the LLM tries to call rtk ls -la directly, it's stripped to ls -la but then correctly rewritten to rtk ls -la.

hooks/vibe/rtk-hook-vibe.sh
```diff
@@ -0,0 +1,122 @@
...
+# Delegate all rewrite logic to the Rust binary.
+# rtk rewrite exits:
+#   0 - rewrite found
+#   1 - no RTK equivalent (pass through)
+#   2 - deny rule matched (not used by RTK, pass through)
+#   3 - ask rule matched (rewrite but ask user)
+REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null || true)
+EXIT_CODE=$?
```
rmsc: The `|| true` in there forces `$EXIT_CODE` to zero all the time. Try this:

```suggestion
set +e
REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null)
EXIT_CODE=$?
set -e
```

hooks/vibe/rtk-hook-vibe.sh
```diff
@@ -0,0 +1,122 @@
...
+    exit 0
+    ;;
+  3)
+    # Ask rule matched - rewrite but we'll allow it (Vibe handles permission separately)
+    ;;
+  *)
+    # Unexpected error - pass through
+    exit 0
+    ;;
+esac
```
rmsc, line 91:
```suggestion
case $EXIT_CODE in
0)
    # Rewrite found - use it
    ;;
127)
    echo "[rtk] WARNING: rtk rewrite not supported (need rtk >= 0.23.0). Upgrade: cargo install --git https://github.com/rtk-ai/rtk" >&2
    exit 0
    ;;
1|2|3|*)
    # No RTK equivalent / deny rule / ask rule / other error - check if we stripped rtk prefix
    if [ "$STRIPPED_CMD" != "$CMD" ] && [ -n "$STRIPPED_CMD" ]; then
        # Command had rtk prefix, use stripped version as the rewrite
        REWRITTEN="$STRIPPED_CMD"
    else
        exit 0
    fi
    ;;
esac
```
If we stripped `rtk` from the command but no rewrite was found, we need to fake a rewrite to the stripped version of the command! Otherwise vibe will blindly accept the original command (for instance the `rtk rm file`).

hooks/vibe/rtk-hook-vibe.sh
```diff
@@ -0,0 +1,122 @@
...
+
+# If rewritten is empty or same as original, pass through
+if [ -z "$REWRITTEN" ] || [ "$REWRITTEN" = "$CMD" ]; then
+  exit 0
+fi
```
rmsc: I think the empty case is already handled by exit code 0 above, and I think the 'same as original' only happened for commands preceded by rtk, which we now explicitly prevent. So this can likely be removed entirely.

```suggestion
```

hooks/vibe/rtk-hook-vibe.sh
rmsc: Damn, mistral is too smart! It discovered it could do `cd /path/to/dir && rtk rm foo` to circumvent the stripping above!

I asked it about that, and it tried calling `rtk rm foo` directly, but got stuck in a confusion loop:

```
I'm making a mistake - I'm running rm foo instead of rtk rm foo. Let me run the correct command.
```

Maybe we should deny rather than pass through.

hooks/vibe/rtk-hook-vibe.sh
rmsc: If the LLM passes `RTK_DISABLED=1` as part of its command, it prevents rewriting.

```suggestion
```

hooks/vibe/rtk-hook-vibe.sh
rmsc: There's no simple way to strip rtk that doesn't also block legitimate uses. Vibe cli should really have a way of asking the user for permission..

hooks/vibe/rtk-hook-vibe.sh
```diff
@@ -0,0 +1,122 @@
...
+    # No RTK equivalent - pass through
+    exit 0
+    ;;
+  2)
+    # Deny rule matched - pass through (RTK doesn't use deny rules)
+    exit 0
+    ;;
+  3)
+    # Ask rule matched - rewrite but we'll allow it (Vibe handles permission separately)
+    ;;
```
rmsc, line 82:
```suggestion
    ;;
    127)
    echo "[rtk] WARNING: rtk rewrite not supported (need rtk >= 0.23.0). Upgrade: cargo install --git https://github.com/rtk-ai/rtk" >&2
    exit 0
    ;;
```

rmsc:
I've resigned to the fact that simply adding `rtk` to the allowlist will offer the LLM countless ways of escaping said allowlist.

I've ended up only blindly allowing specific rtk commands (like `rtk git status` or `rtk ls`).

bashout:
> I've resigned to the fact that simply adding `rtk` to the allowlist will offer the LLM countless ways of escaping said allowlist.


Yeah I've never been personally _too_ strict with limiting the commands mistral executes outside of external communication and commits prior to my review so I didn't catch this. I wonder why it's particularly different for Mistral compared to Claude etc - or whether it's a known issue. I imagine the other tools may have a more sophisticated way or they check prior to the pre-tool hook calls whether the command is allowed or not. Something I may dig into once I have capacity. 

Also I'm likely to approve your suggested changes, they all seem sensible just need to get some time to sit down and test it all locally to confirm. Thanks for the review!

bashout:
Branch updated @rmsc  - incorporated your changes and attributed you. Cleaned up the commit history. Hopefully we can get this reviewed and merged in so it's one less thing on my plate ;)

rmsc:
I've used the hook for a few weeks, and the experience was actually very poor.

> I wonder why it's particularly different for Mistral compared to Claude etc - or whether it's a known issue.

The mistral API is much more limited, there's no way for instance to ask the user permissions. Mistral also chains commands very frequently (surprisingly starting with `cd /my/workspace/dir && command`), which is very efficient but bypasses the hook completely.

It also very gets frequently stuck in loops, especially with `find -exec`, which mistral loves to use. `find -exec` gets written to `rtk find -exec` which fails with a message recommending the use of regular `find`, so mistral tries again without the `rtk`......

On the other hand, it quickly picked up that rtk was available, which was also positively surprising. It started using it all the time in the middle of command chains, despite completely ignoring my instruction in AGENTS.md NOT to keep cd'ing to the current directory.

I've recently replaced the hook with a skill, so hopefully mistral will have better judgement for when to use rtk and when not to.

bashout:
> I've used the hook for a few weeks, and the experience was actually very poor.

That's unfortunate. Do you feel like it negated the savings being valuable or was it simply itself getting stuck in a loop was more frustrating?

> Mistral also chains commands very frequently (surprisingly starting with `cd /my/workspace/dir && command`), which is very efficient but bypasses the hook completely.

As you've echoed later in your comment I didn't personally find it too irritating that the command chains were less hookable as mistral seemed to do exceptionally well at determining it was available and using the commands in the first instance.

> It also very gets frequently stuck in loops, especially with `find -exec`, which mistral loves to use. `find -exec` gets written to `rtk find -exec` which fails with a message recommending the use of regular `find`, so mistral tries again without the `rtk`......

I have similarly noticed this behavior in my own experience - but as you may have noticed haven't had much time to chase it down. 

> I've recently replaced the hook with a skill, so hopefully mistral will have better judgement for when to use rtk and when not to.

Mind sharing the skill?

I'm likely still going to update this MR due to the breaking hook changes done in 2.21.0 just to keep it available. Ideally I'll set up a bit of a A/B harness when I get the chance to see if it's tangible worse rather than just being driven by my own sentiment as I feel like RTK is the closest native 'optimizer' we've gotten for typical tooling.

rmsc:
> That's unfortunate. Do you feel like it negated the savings being valuable or was it simply itself getting stuck in a loop was more frustrating?

Loops caused a massive consumption of cached tokens (97% of total consumption!), and burned through my vibe pro plan in just a couple of days. I only noticed it being stuck after that.

That said, many of the loops had nothing to do with the hook. I'm trying to figure it out with customer support, and apparently there's a bug in vibe-cli that's preventing loops from being broken.

> Mind sharing the skill?

Still a WIP, but sure:

[SKILL.md](https://github.com/user-attachments/files/30228766/SKILL.md)

PR #2465, opened 2026-06-16
fix(grep): surface error on exit code >= 2 instead of false "0 matches"

### LINKED ISSUE #2461: False negative: rtk grep reports '0 matches' on exit code 1 when rg not found
## What happened

`rtk grep` was invoked to search for `~/.claude` references in two files. RTK could not resolve `rg` via PATH, fell back to "direct exec," and returned exit code 1 — but still printed `0 matches for '~/.claude'` to stdout.

The files **do** contain `~/.claude` references. The search silently failed and reported clean.

## Environment

- **RTK version:** 0.42.2
- **OS:** Windows Server 2019 (10.0.17763)
- **Shell:** PowerShell 7.4.12
- **Claude model:** claude-sonnet-4-6 (Claude Code CLI)
- **RTK mode:** Native Windows (CLAUDE.md injection + PreToolUse hook rewrite)

## Why this matters

The Claude Code CLAUDE.md injected by `rtk init -g` instructs the LLM to prefer the built-in `Grep` tool over Bash for content searches. But RTK's `PreToolUse:Bash` hook transparently rewrites `grep` commands to `rtk grep` before the LLM sees them — so even when the LLM follows the CLAUDE.md instruction and writes a Bash grep, it gets silently promoted into an `rtk grep` call. If that call fails with exit code 1, the LLM receives plausible-looking output (`0 matches`) with no indication the command failed. The LLM trusts it.

This is worse than a hard failure. A hard failure is visible. A false "0 matches" on a failed search is invisible.

## Steps to reproduce

1. Install rtk on a system where `rg` (ripgrep) is not on PATH
2. Open Claude Code. Let the hook rewrite a Bash grep: `grep -n "somestring" somefile.md`
3. Observe the tool result: `exit code 1`, `is_error: true`, but output reads `0 matches for 'somestring'`
4. The file contains "somestring". The LLM reports the file is clean.

## What the CLAUDE.md injection could say

Obviously judge for yourself what the proper solution is, but the CLAUDE.md write-in could add something like:

> If an `rtk` command returns exit code 1, treat the entire result as a tool failure — do not read any output from it as a search result. A non-zero exit from `rtk` means the underlying command did not complete successfully.

That one line would have prevented this. The LLM would have seen `is_error: true`, noticed the exit code, and not reported "confirmed clean."

Alternatively, rtk could suppress the `0 matches` output entirely when the fallback fails — so the only output is the error message, making it unambiguous to the LLM that nothing was searched.

ousamabenyounes:
## Summary

`rtk grep` prints "0 matches for '<pattern>'" when the underlying search tool exits with an error code >= 2, even when matches exist and the tool simply failed to run. This is a false negative — LLM agents interpret it as a clean empty result and act on incorrect information.

**Root cause:** `grep_cmd.rs:386-399` handled only the narrow case `exit_code == 2 && stderr non-empty`. grep/rg convention: exit 1 = no match (normal), exit >= 2 = real error (bad regex, binary crash, missing tool). Any error exit that fell outside the narrow window silently became "0 matches".

**Fix:** Broadened the gate to `exit_code >= 2`. All error exits now surface to stderr (prefixed with the exit code for diagnostics) and **never** print the misleading "0 matches" message. Exit 0 and 1 behavior is unchanged.

Fixes #2461

With the fix reverted, the new test fails (RED):

```
running 1 test
test grep_error_exit_code_no_false_negative ... FAILED

---- grep_error_exit_code_no_false_negative stdout ----
thread 'grep_error_exit_code_no_false_negative' panicked at tests/grep_error_test.rs:50:5:
ERROR exit MUST NOT claim '0 matches' (false negative).
stdout: 0 matches for 'hello'
stderr: 

failures:
    grep_error_exit_code_no_false_negative

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
```

With the fix applied, the test passes (GREEN):

```
running 1 test
test grep_error_exit_code_no_false_negative ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

Full baseline: **2196 tests pass, 0 fail** (2195 existing + 1 new).


hgunduzoglu (commented):
Checked this out and tested locally (macOS arm64). The core fix is correct — broadening the gate to `exit_code >= 2` matches grep/rg convention, and all three paths behave right:

| rg exit | meaning | before | this PR |
|---|---|---|---|
| 0 | matches | grouped output | grouped output ✓ |
| 1 | no match | `0 matches` | `0 matches`, exit 1 ✓ (legit empty result preserved) |
| ≥2 | real error | `0 matches` (false negative) | `grep failed with exit code N` on stderr, exit propagated ✓ |

Verified with a fake `rg` exiting 2: it now surfaces `grep failed with exit code 2` and exits 2 instead of the misleading `0 matches`. The new integration test passes here. Nice — this closes a genuine false-negative footgun.

**One blocker for CI, though:** `tests/grep_error_test.rs` has an unconditional `use std::os::unix::fs::PermissionsExt;` (line 6) with no `cfg(unix)` gate. `std::os::unix` doesn't exist on Windows, so the integration-test crate will fail to *compile* on the Windows CI target (RTK ships a Windows build, and cli-testing.md marks cross-platform a 🔴 critical concern). It's the only file under `tests/` that touches `std::os::unix`.

Two small options:
- Gate the whole file: add `#![cfg(unix)]` at the top (simplest — the test simulates a broken `rg` via a chmod'd shell script, which is unix-only anyway), or
- `#[cfg(unix)]` on the test fn + `#[cfg(unix)] use std::os::unix::...`.

Either keeps the Windows build green while preserving the unix coverage. Everything else looks good — the fix itself is solid and well-targeted.

ousamabenyounes:
Thanks for the review @hgunduzoglu — agreed, that's a real Windows compile break, not a nit. Fixed in `6eaa4a3` with option 1: `#![cfg(unix)]` at the top of `tests/grep_error_test.rs`. The test simulates a broken `rg` via a chmod'd shell script (`from_mode` / `PermissionsExt`), so the whole file is unix-only by nature — gating it keeps the Windows CI target green while preserving the unix coverage.

Verified: cross-`check` against `x86_64-pc-windows-gnu` now compiles clean (was `E0433: could not find unix in os`), and the Linux test still passes 1/1.

src/cmds/system/grep_cmd.rs
```diff
@@ -384,9 +384,23 @@ pub fn run(
     let raw_output = result.stdout.clone();
 
     if result.stdout.trim().is_empty() {
-        // Show stderr for errors (bad regex, missing file, etc.)
-        if exit_code == 2 && !result.stderr.trim().is_empty() {
-            eprintln!("{}", result.stderr.trim());
+        // grep/rg convention: exit 1 = no match found (normal).
+        // exit >= 2 = real error (bad regex, tool crash, etc.) — surface it,
+        // never print "0 matches" for an error.
+        const GREP_ERROR_EXIT: i32 = 2;
```
KuSh: No need for a const, the comment suffice

tests/grep_error_test.rs
```diff
@@ -0,0 +1,72 @@
...
+//! rtk must surface the error instead of printing a false-negative "0 matches".
+//!
+//! Unix-only: simulates a broken `rg` via a chmod'd shell script. The whole
+//! file is gated so the integration-test crate still compiles on Windows CI.
+#![cfg(unix)]
+
+use std::os::unix::fs::PermissionsExt;
+use std::process::Command;
+
+#[test]
```
KuSh: Hum that seems overthought, just move condition into a pure function and test that.

If you also want to test what rtk outputs move code into a handle_grep_result function and unit test that

KuSh (changes requested):

tests/grep_error_test.rs
KuSh: If not clear, prefer unit tests over integration tests. There's also a binary test suit that you could use but prefer to test it with grep binary (or conditionnaly on rg if available) and with real bad inputs instead of faking the binary

src/cmds/system/grep_cmd.rs
ousamabenyounes: Done in `39ca858` — dropped the const. The `exit_code >= 2` decision now lives in a small `is_grep_error_exit` function whose doc comment carries the grep/rg exit convention, so no const needed.

tests/grep_error_test.rs
ousamabenyounes: Refactored as you suggested in `39ca858`: pulled the condition into a pure `fn is_grep_error_exit(exit_code: i32) -> bool` and unit-test it directly (`test_is_grep_error_exit`: 0/1 = normal, 2/3/127 = error). No fake binary, no integration crate.

KuSh (approved):
LGTM thanks!

PR #2474, opened 2026-06-17
feat(find): fall back to native find for unsupported predicates

### LINKED ISSUE #2469: rtk: rtk find does not support compound predicates or actions (e.g. -not, -exec). Use `find` directly.
As the title says, rtk find needs to be expanded to support compound predicates

hgunduzoglu:
## Summary

Fixes #2469. `rtk find` refused compound predicates and actions — `-not`, `!`, `-or`/`-o`, `-and`/`-a`, `-exec`, `-size`, `-mtime`, `-regex`, etc. — by printing `rtk find does not support compound predicates or actions … Use \`find\` directly.` and producing no results. That silently breaks the user's command, which violates RTK's **Never-Block** principle ("if a filter fails, fall back to raw output; RTK should never prevent a command from executing").

## Change

When `rtk find` sees a predicate it can't compact, it now falls back to **native `find` (unfiltered passthrough)** instead of refusing, and propagates native find's exit code. The common `-name` / `-type` / `-maxdepth` path is untouched and keeps full rtk compaction.

```console
# before — command refused, no results
$ rtk find . -not -name "*.txt"
rtk find does not support compound predicates or actions (e.g. -not, -exec). Use `find` directly.

# after — native find runs, results returned
$ rtk find . -not -name "*.txt"
.
./keep.log
./a
./a/x.rs
./b
```

This also makes `-exec`, `-o`, `-size`, `-mtime`, … work for the first time, since they all reach native find now.

## Notes

- **Token savings**: the fallback path is intentionally unfiltered (same contract as `rtk proxy`) — a correct full result beats a broken/refused command. The compactable majority (`-name`/`-type`/`-maxdepth`) still gets rtk's tree formatting. Full *compaction* of compound-predicate output could be a later enhancement; this PR is the Never-Block fix.
- `parse_find_args` keeps its bail as a defensive contract (its unit tests still pass); `run_from_args` now intercepts unsupported flags before parsing and routes to passthrough.
- `run_from_args` now returns `Result<i32>` to propagate the child exit code; the `main.rs` Find arm returns it.

## Testing

- 4 new tests: `-not` passthrough returns Ok(0), `-exec` passthrough returns Ok(0), `has_unsupported_find_flags` detects unsupported / ignores supported predicates.
- Existing `parse_find_args` bail tests and `run_from_args` compaction tests still pass.
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — clean, 2202 passed.
- Manual verification vs system `find` for `-not`, `-o`, `-exec`, and exit-code propagation on an invalid query (rtk exit 1 == native exit 1).

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -575,7 +575,13 @@ mod tests {
     }
 
     // --- #2469: unsupported predicates fall back to native find (Never Block) ---
+    // These two execute the resolved `find` and assert GNU/BSD find semantics
+    // (a valid -not/-exec query exits 0). On Windows, `find.exe` is an unrelated
+    // text-search tool that rejects these args (exit 2), so the assertions only
+    // hold on unix. The passthrough routing itself is covered cross-platform by
+    // `has_unsupported_find_flags_detects_and_ignores` below.
 
+    #[cfg(unix)]
     #[test]
     fn run_from_args_unsupported_predicate_passes_through() {
         // `-not` can't be compacted; instead of bailing, rtk runs native find.
@@ -585,6 +591,7 @@ mod tests {
         assert_eq!(result.unwrap(), 0, "native find exit code is propagated");
     }
 
+    #[cfg(unix)]
     #[test]
     fn run_from_args_exec_passes_through() {
         // `-exec` is an action rtk can't model; native find handles it.
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -177,7 +177,16 @@ fn parse_rtk_find_args(args: &[String]) -> Result<FindArgs> {
 }
 
 /// Entry point from main.rs — parses raw args then delegates to run().
-pub fn run_from_args(args: &[String], verbose: u8) -> Result<()> {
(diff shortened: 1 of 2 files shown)


hgunduzoglu:
Fixed the Windows CI failure in `352e7b2`. The two passthrough tests executed the resolved `find` and asserted GNU/BSD semantics (valid `-not`/`-exec` query exits 0), but on Windows `find.exe` is an unrelated text-search tool that rejects those args (exit 2). Gated both behind `#[cfg(unix)]`; the routing decision stays covered cross-platform by `has_unsupported_find_flags_detects_and_ignores`.

vivekkartha7 (approved):

KuSh:
Thanks for reporting this and for the never-block reasoning, which is the right call. This is fixed on develop by #3603, merged 2026-08-26 (commit 203948b3f1): rtk find now parses find's own grammar, sends predicates it does not model to real find and still compresses the results, and runs actions like -exec or -delete through passthrough with the exit code preserved. I re-ran the examples from your description against a current develop build and -not, parenthesised -o and -exec all work and match native find. Closing as already covered, and please comment or reopen if you find a predicate that still gets refused.

PR #1369, opened 2026-04-17
fix(init): honor RTK_TELEMETRY_DISABLED in consent prompt (#1307)

ousamabenyounes:
## Summary

- Short-circuits `prompt_telemetry_consent` when `RTK_TELEMETRY_DISABLED=1` so `rtk init` no longer hangs in non-interactive environments (devcontainer `postCreateCommand`, CI agents with pseudo-TTY) where `is_terminal()` alone is not enough.
- Keeps the existing TTY fallback untouched; the new env check runs before it.
- Mirrors the same env var gate already used by `telemetry::maybe_ping`, so both the prompt and the ping honour the documented opt-out consistently.

Fixes #1307

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — 1591 pass, 0 clippy errors, format clean
- [x] Unit test added: `test_telemetry_disabled_by_env_honors_opt_out` covers unset, `"1"`, and common non-`"1"` values (`"0"`, `"true"`, `"false"`, `"yes"`, `"no"`, `""`) — bundled in one `#[test]` to serialize env-var mutations.
- [x] Manual review: diff is 57 lines added, 0 removed; no behaviour change when the env var is unset.

> **Important:** All PRs must target the `develop` branch (not `master`).
> See [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) for details.

---
_Vibe Coded by Ousama Ben Younes_
_Developed With Ora Studio (Claude Code)_

PR #2477, opened 2026-06-17
fix(init): honor RTK_TELEMETRY_DISABLED in consent prompt (#1307)

### LINKED ISSUE #1307: `rtk init` hangs in non-interactive environments (v0.36.0 telemetry consent gate)
### Description                                                                                  
   
  `rtk init -g --hook-only --auto-patch` hangs indefinitely in non-interactive environments        
  (Docker/devcontainer builds, CI) since v0.36.0. The hook setup completes and output is printed,
  but the process never exits — it blocks waiting for stdin on the telemetry consent prompt.       
                                                                                                 
  ### Reproduction

  ```bash
  # In any non-interactive context (Dockerfile, devcontainer postCreateCommand, CI):
  rtk init -g --hook-only --auto-patch                                                             
  # Process prints hook setup output, then hangs forever
  ```                                                                                              
                                                                                                 
  Last visible output before hang:

  ```
  RTK hook already up to date (hook-only mode).

    Hook: /home/node/.claude/hooks/rtk-rewrite.sh                                                  
    Note: No RTK.md created. Claude won't know about meta commands (gain, discover, proxy).
                                                                                                   
  ```

  ### Expected behavior                                                                            
   
  `rtk init` should detect non-interactive environments (e.g., `! -t 0` / no TTY on stdin) and skip
   the consent prompt, or respect `RTK_TELEMETRY_DISABLED=1` to suppress it entirely.            

  ### Actual behavior                                                                              
   
  The process blocks on stdin waiting for consent input that will never come.                      
  `RTK_TELEMETRY_DISABLED=1` does not prevent the prompt — it only disables data collection after
  consent is given.

  ### Workaround

  ```bash
  RTK_TELEMETRY_DISABLED=1 timeout 10 rtk init -g --hook-only --auto-patch 2>/dev/null || true
  ```                                                                                              
   
  ### Environment                                                                                  
                                                                                                 
  - RTK: v0.36.0 (v0.35.0 worked fine)
  - OS: Debian Trixie (Docker container, `node:24-trixie-slim`)
  - Arch: ARM64 (Apple Silicon via Docker Desktop)                                                 
  - Context: devcontainer `postCreateCommand` (no TTY)

ousamabenyounes:
Fixes #1307.

`rtk init`'s telemetry consent prompt only bailed on a non-TTY check. Some non-interactive environments (devcontainer `postCreateCommand`, certain CI agents) hand rtk a pseudo-TTY, so `is_terminal()` returns true and the prompt blocks forever waiting on stdin. `RTK_TELEMETRY_DISABLED=1` is the documented opt-out and is honoured by `telemetry::maybe_ping`, but the consent prompt ignored it.

## Fix
Add a `telemetry_disabled_by_env()` helper and short-circuit `prompt_telemetry_consent()` with `return Ok(())` when `RTK_TELEMETRY_DISABLED=1`, before the TTY heuristic — mirroring the check already used in `maybe_ping`.

The hang itself only reproduces under a real pseudo-TTY, which the test harness can't provide; the original regression remains guarded by the unit test on `telemetry_disabled_by_env()`.

RED - patch reverted (helper removed):
```
error[E0425]: cannot find function `telemetry_disabled_by_env` in this scope
error: could not compile `rtk` (bin "rtk" test) due to 4 previous errors
```

GREEN - with the original fix:
```
running 1 test
```

Reviewer follow-up RED - with the test-location change reverted:
```
FAIL: test missing from telemetry_cmd.rs
```

Reviewer follow-up GREEN - after moving the test to `telemetry_cmd`:
```
207:    fn test_telemetry_disabled_by_env_honors_opt_out() {
PASS: test lives in telemetry_cmd.rs
PASS: test removed from init.rs
```

Full gate after rebase onto current `develop`:
```
cargo fmt --all --check && cargo clippy --all-targets && cargo test
test result: ok. 2363 passed; 0 failed; 8 ignored
test result: ok. 6 passed; 0 failed
test result: ok. 11 passed; 0 failed
test result: ok. 6 passed; 0 failed
test result: ok. 14 passed; 0 failed
test result: ok. 5 passed; 0 failed
test result: ok. 11 passed; 0 failed
```

(Re-proposes the accidentally-closed #1369, rebased onto current `develop`.)

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -467,6 +467,17 @@ pub fn save_telemetry_consent(accepted: bool) -> Result<()> {
         .context("Failed to save telemetry consent to config.toml")
 }
 
+/// Returns true when telemetry is explicitly disabled through the
+/// `RTK_TELEMETRY_DISABLED` env var (value `"1"`).
+///
+/// Kept as a small pure function so the consent prompt short-circuits before
+/// touching stdin, and so tests can cover the opt-out path without a TTY.
+/// Mirrors the check in `telemetry::maybe_ping` so both the prompt and the
+/// ping honour the same env toggle.
+fn telemetry_disabled_by_env() -> bool {
+    std::env::var("RTK_TELEMETRY_DISABLED").unwrap_or_default() == "1"
+}
+
 fn prompt_telemetry_consent() -> Result<()> {
     use std::io::{self, BufRead, IsTerminal};
 
@@ -475,6 +486,16 @@ fn prompt_telemetry_consent() -> Result<()> {
         Some(true) => return Ok(()),
         Some(false) => return Ok(()),
         None => {}
+    }
+
+    // Explicit opt-out must short-circuit before the TTY heuristic: some
+    // non-interactive environments (devcontainer `postCreateCommand`, certain
+    // CI agents) hand rtk a pseudo-TTY, so `is_terminal()` returns true even
+    // though no human is available to answer — the prompt then hangs forever.
+    // Setting `RTK_TELEMETRY_DISABLED=1` is the documented workaround, so the
+    // init prompt has to honour it too, not only `telemetry::maybe_ping`.
+    if telemetry_disabled_by_env() {
+        return Ok(());
     }
 
     if !io::stdin().is_terminal() {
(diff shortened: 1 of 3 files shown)


src/hooks/init.rs
```diff
@@ -456,6 +456,17 @@ pub fn save_telemetry_consent(accepted: bool) -> Result<()> {
...
 }
 
+/// Returns true when telemetry is explicitly disabled through the
+/// `RTK_TELEMETRY_DISABLED` env var (value `"1"`).
+///
+/// Kept as a small pure function so the consent prompt short-circuits before
+/// touching stdin, and so tests can cover the opt-out path without a TTY.
+/// Mirrors the check in `telemetry::maybe_ping` so both the prompt and the
+/// ping honour the same env toggle.
+fn telemetry_disabled_by_env() -> bool {
```
KuSh: nitpick, could be great to have that public, exported from telemetry_cmd, to avoid divergence if we decide to evolve it to accept `y` or `true` one day

KuSh (changes requested):
Otherwise LGTM, thanks

src/hooks/init.rs
ousamabenyounes: Done in `21d6981` — moved it to a public `telemetry_cmd::telemetry_disabled_by_env()` and routed all three call sites through it (the init prompt, `telemetry status`, and `telemetry::maybe_ping`). The `RTK_TELEMETRY_DISABLED == "1"` check now lives in exactly one place, so if we ever accept `y`/`true` it changes there once. Behavior-preserving; the #1307 regression test is unchanged and the suite stays 2200/2200.

src/hooks/init.rs
```diff
@@ -4221,6 +4231,42 @@ mod tests {
         );
     }
 
+    /// Regression for #1307: `RTK_TELEMETRY_DISABLED=1` must short-circuit the
+    /// consent prompt so `rtk init` cannot hang in non-interactive environments.
+    /// All cases are bundled in one test to serialize env-var mutations (env is
+    /// process-global and cargo runs tests in parallel).
+    #[test]
+    fn test_telemetry_disabled_by_env_honors_opt_out() {
```
KuSh: Those tests should go to telemetry_cmd. After that LGTM!

KuSh (changes requested):

src/hooks/init.rs
aeppling: Hey @ousamabenyounes , any news on this so we can merge soon?

ousamabenyounes: Thanks for the review, KuSh. The regression test now lives in `src/core/telemetry_cmd.rs` next to `telemetry_disabled_by_env()`, and I removed the copy from `src/hooks/init.rs`. Pushed `7040dc6` after rebasing on current `develop`; local full gate passed: `cargo fmt --all --check && cargo clippy --all-targets && cargo test`.

ousamabenyounes: @aeppling I was on vaccation happy to come back and contribute to RTK again !

ousamabenyounes:
Rebased this PR onto current `develop` and pushed `46ca979`.

KuSh (approved):

PR #1221, opened 2026-04-11
fix(output): preserve tail lines to prevent CI summary loss (#1035)

ousamabenyounes:
## Summary

Fixes #1035

- `make.toml` had `max_lines = 50` which kept only the first 50 lines of filtered output, discarding the tail where CI tools (pytest, vitest) always print their pass/fail summaries
- `truncate_output()` in `src/parser/mod.rs` kept only the head chars in passthrough mode, causing the same summary loss when the vitest JSON parser falls through to Tier 3
- Both now use head + last 20 lines with an omission marker so CI results are always visible regardless of how long the full output is

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [x] New test `test_truncate_output_preserves_tail` — 101-line CI output, verifies "251 passed, 18 skipped" survives truncation
- [x] New test `test_truncate_output_tail_shows_last_20_lines` — 50-line output, all 20 summary lines appear in tail
- [x] New TOML test `"preserves tail for long composite output"` — 51 filtered make lines, head=20 + omit(1) + tail=30
- [x] Existing tests all pass (1375 pass, 0 fail, 0 regressions)
- [x] Baseline was 1373; +2 from new tests

## Files changed

| File | Change |
|------|--------|
| `src/filters/make.toml` | Replace `max_lines = 50` with `head_lines = 20, tail_lines = 30`; add tail-preservation test case |
| `src/parser/mod.rs` | Update `truncate_output()` to show head + tail with omission marker; add 2 new tests |

Generated by Ora Studio
Vibe coded by ousamabenyounes

---
_Vibe Coded by Ousama Ben Younes_
_Developed With Ora Studio (Claude Code)_

PR #2480, opened 2026-06-17
fix(output): preserve tail lines to prevent CI summary loss (#1035)

### LINKED ISSUE #1035: Output lost on composite commands (make ci-local / vitest) - summary lines removed
## Environment

- RTK version: 0.34.3
- OS: macOS Darwin 24.6.0
- Claude Code hook config:
```json
{
  "PreToolUse": [{
    "hooks": [{"command": "rtk-rewrite.sh", "type": "command"}],
    "matcher": "Bash"
  }]
}
```

## Problem

RTK's output filtering causes critical CI result lines to be lost when running composite/long commands. Two specific issues observed:

### 1. vitest output parsing failure

Running `npx vitest run` produces:

```
[RTK:PASSTHROUGH] vitest parser: All parsing tiers failed
```

RTK cannot parse vitest's output format. The passthrough fallback sometimes still loses output.

### 2. make target output is cut off

Running `make ci-local` (which chains `ruff check` + `pytest` + `vitest` + quality check) gets cut off at ~13% of pytest results:

```
tests/unit/.../test_agent_frontmatter_fixtures[valid/agent-valid.md-True] PASSED [ 13%]
... (1379 lines truncated)
```

The cut removes:
- The pytest summary line (`251 passed, 18 skipped`)
- The entire vitest run (`277 passed`)
- The quality check results

This means **we cannot determine if CI passed or failed** from the output, even though the exit code was 0.

### 3. File redirect also affected

Even when redirecting to a file (`make ci-local > /tmp/output.log 2>&1`), the log file contains the same `... (1379 lines truncated)` marker, suggesting RTK filters before the output reaches the file.

## Expected Behavior

- Long composite commands (make targets chaining multiple tools) should not lose their summary/verdict lines
- At minimum, the **last N lines** (tail) should always be preserved, as CI tools print pass/fail summaries at the end
- vitest format should be recognized, or unknown formats should pass through cleanly

## Suggested Fix

1. **Always preserve tail**: Even when trimming middle sections, keep the last 20-30 lines (most CI tools print summaries at the end)
2. **Whitelist make/composite commands**: Don't filter output from `make`, `npm test`, `npx vitest` etc.
3. **vitest parser**: Add vitest output format support alongside existing jest/pytest parsers

## Workaround

Remove the RTK hook from Claude Code settings entirely, losing all token-saving benefits.

ousamabenyounes:
Fixes #1035.

`truncate_output` kept only the head of long output. For composite commands (`make ci-local`, `vitest`, …) the pass/fail summary that CI tools print at the *end* was dropped, so failures became invisible.

## Fix
Rewrite `truncate_output` to preserve both head and tail: head gets two-thirds of the char budget, the last 20 lines are always kept (where CI summaries live), and the omitted middle is replaced with a `[RTK:PASSTHROUGH] N chars omitted` marker. When head and tail would overlap (e.g. single-line input) it falls back to head-only truncation. `make.toml` is updated to route composite recipes through this path.

RED — `truncate_output` reverted to develop's head-only version:
```
test parser::tests::test_truncate_output_preserves_tail ... FAILED
test parser::tests::test_truncate_output_tail_shows_last_20_lines ... FAILED
test result: FAILED. 3 passed; 2 failed
```

GREEN — with head+tail preservation:
```
running 5 tests
test result: ok. 5 passed; 0 failed
```

(Re-proposes the accidentally-closed #1221, rebased onto current `develop`.)

### DIFF
--- a/src/filters/make.toml
+++ b/src/filters/make.toml
@@ -6,7 +6,8 @@ strip_lines_matching = [
   "^\\s*$",
   "^Nothing to be done",
 ]
-max_lines = 50
+head_lines = 20
+tail_lines = 30
 on_empty = "make: ok"
 
 [[tests.make]]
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Rebased this PR onto current `develop` and pushed `c38aab0`.

ousamabenyounes:
Rebased onto current `origin/develop` and pushed benchmark fix `68052f0`.

What changed:

Local verification:
```text
cargo fmt --all -- --check
cargo test --all
2450 tests: passed, 8 ignored
cargo clippy --all-targets
cargo build --release
CI=1 ./scripts/benchmark.sh
0 negative, 0 fail (exit 0)
```

PR #2481, opened 2026-06-17
fix(filters): use tail_lines for just/mise/task to preserve failure summaries (#1065)

### LINKED ISSUE #1065: TOML delegator filters (just/mise/task) truncate wrapped tool output, hiding test failures

ousamabenyounes:
Fixes #1065.

The `just`, `mise` and `task` TOML delegator filters used `max_lines = 50`, which truncated wrapped tool output at 50 lines. When the wrapped command printed more than that, the pass/fail summary at the tail (e.g. `FAILED: test_x`, `result: N failures`) was dropped — hiding test failures.

## Fix
Replace `max_lines = 50` with `tail_lines = 200` in all three filters so the end of the output (where test runners print their summary) is always preserved. Adds an inline `[[tests.<filter>]]` regression case per filter (input deliberately exceeds the old 50-line cap and ends with a failure summary).

Run via `rtk verify --filter <name>` (the project's TOML inline-test runner).

RED — config reverted to `max_lines = 50`:
```
--- just ---  3/4 tests passed   (tail summary truncated: "... (11 lines truncated)")
--- mise ---  3/4 tests passed
--- task ---  3/4 tests passed
rtk: 1 test(s) failed
```

GREEN — with `tail_lines = 200`:
```
--- just ---  4/4 tests passed
--- mise ---  4/4 tests passed
--- task ---  4/4 tests passed
```

(Re-proposes the accidentally-closed #1191, rebased onto current `develop`.)


ousamabenyounes:
Rebased this PR onto current `develop` and pushed `5e61e82`.

ousamabenyounes:
Rebased onto current `origin/develop` and pushed benchmark fix `1263a57`.

What changed:

Local verification:
```text
cargo fmt --all -- --check
cargo test toml_filter
66 passed
cargo test --all
2440 tests: passed, 8 ignored
cargo clippy --all-targets
cargo build --release
CI=1 ./scripts/benchmark.sh
0 negative, 0 fail (exit 0)
```

PR #2483, opened 2026-06-17
fix(tracking): record stats for hook-rewritten commands (#1082)

### LINKED ISSUE #1082: rtk gain stats not recorded when commands are rewritten via Claude Code hook
## Bug Description

When `git status` is intercepted by the Claude Code PreToolUse hook and rewritten to `rtk git status`, the command executes correctly but is never recorded in `rtk gain` statistics. Only commands invoked directly as `rtk git status` are tracked.

## Steps to Reproduce

1. Install RTK with global hook: `rtk init -g`
2. Confirm hook is active in `~/.claude/hooks/rtk-rewrite.sh`
3. Note current count: `rtk gain | grep "Total commands"`
4. Run `git status` via Claude Code Bash tool (hook rewrites it to `rtk git status`)
5. Wait 1+ seconds, then check again: `rtk gain | grep "Total commands"`

**Result**: count does not increase.

6. Run `rtk git status` directly
7. Check again: count increases by 1.

## Evidence

Database (`~/Library/Application Support/rtk/history.db`) only contains entries from direct invocations, never from hook-rewritten executions:

| original_cmd | rtk_cmd        | timestamp           |
|--------------|----------------|---------------------|
| git status   | rtk git status | 2026-04-08T08:21... |  ← direct call
| git status   | rtk git status | 2026-04-06T05:26... |  ← direct call

Verified: `defaultMode: bypassPermissions` is set, so sandbox write restrictions are not the cause (unlike issue #320).

## Root Cause Hypothesis

`TimedExecution::start()` is only called in `run_fallback()` and `Commands::Proxy`. When Claude Code executes the hook-rewritten `rtk git status`, it goes through `Commands::Git` which has no tracking call. Something in the hook execution environment prevents the tracking path from being reached.

## Environment

- rtk: 0.34.3
- Claude Code hook version: 3
- macOS Darwin 25.3.0
- DB path: `~/Library/Application Support/rtk/history.db`

ousamabenyounes:
Fixes #1082.

`rtk gain` stats were silently missing for commands rewritten via the Claude Code hook. The hook spawns a second `rtk` process while another holds the SQLite history DB, so the tracking write hit lock contention. Two problems compounded it: `busy_timeout` and `journal_mode=WAL` were set in a single `execute_batch()` call, so a transient WAL error left `busy_timeout` at 0 — and with `busy_timeout=0` every contended write failed immediately with `SQLITE_BUSY` and was dropped (`let _ = tracker.record(...)`).

## Fix
Set `PRAGMA busy_timeout=5000` separately from `journal_mode=WAL` so the timeout is always applied, and add a bounded retry around the record path. A contended write now waits for the lock instead of being lost.

`test_record_waits_for_concurrent_write_lock` reproduces real contention: a background thread holds the write lock (`BEGIN IMMEDIATE`) while the tracker attempts its INSERT.

RED — `busy_timeout` reverted to 0 (the bug):
```
test core::tracking::tests::test_record_waits_for_concurrent_write_lock ... FAILED
record must wait for the concurrent lock and succeed (busy_timeout): database is locked
```

GREEN — with `busy_timeout=5000`:
```
test core::tracking::tests::test_record_waits_for_concurrent_write_lock ... ok
```
Full `core::tracking` module: 15 passed, 0 failed (no baseline regression).

(Re-proposes the accidentally-closed #1223, rebased onto current `develop`. The original branch's test wrote two trackers sequentially and so did not actually fail without the fix; it has been replaced with the lock-contention test above.)

## CI benchmark verification (RED → GREEN)
The benchmark job was red only on `curl json` because the script compared two separate network calls to `mockhttp.org/json`; that endpoint can return different JSON bodies, making a pass-through filter look negative. The benchmark now uses one local `file://` JSON fixture for both `curl` and `rtk curl`.

RED — old script had no deterministic self-test output:
```
$ bash scripts/benchmark.sh --self-test | grep -F 'PASS: curl json benchmark fixture is deterministic'
(no PASS line; exit 1)
```

GREEN — with the fixture-backed benchmark:
```
$ bash scripts/benchmark.sh --self-test
PASS: curl json benchmark fixture is deterministic
```

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -12,11 +12,30 @@ else
 fi
 BENCH_DIR="$(pwd)/scripts/benchmark"
 RTK_ROOT="$(pwd)"
+readonly FILE_URL_PREFIX="file://"
+readonly CURL_JSON_FIXTURE_NAME="rtk_benchmark_curl_json"
+readonly CURL_JSON_FIXTURE_CONTENT='{"products":[{"id":101,"name":"Laptop","price":999.99,"stock":50},{"id":102,"name":"Mouse","price":29.99,"stock":200}],"total":2,"page":1}'
+BENCH_TEMP_FILES=()
+
+cleanup_benchmark_temp_files() {
+  if [ "${#BENCH_TEMP_FILES[@]}" -gt 0 ]; then
+    rm -f "${BENCH_TEMP_FILES[@]}"
+  fi
+}
+trap cleanup_benchmark_temp_files EXIT
 
 if [ -z "$CI" ]; then
   rm -rf "$BENCH_DIR"
   mkdir -p "$BENCH_DIR/unix" "$BENCH_DIR/rtk" "$BENCH_DIR/diff"
 fi
+
+make_curl_json_fixture_url() {
+  local fixture
+  fixture=$(mktemp "${TMPDIR:-/tmp}/${CURL_JSON_FIXTURE_NAME}.XXXXXX")
+  BENCH_TEMP_FILES+=("$fixture")
+  printf "%s\n" "$CURL_JSON_FIXTURE_CONTENT" > "$fixture"
+  printf "%s%s\n" "$FILE_URL_PREFIX" "$fixture"
+}
 
 safe_name() {
   echo "$1" | tr ' /' '_-' | tr -cd 'a-zA-Z0-9_-'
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Rebased this PR onto current `develop` and pushed `eb6a29e`.

ousamabenyounes:
Rebased onto the latest `develop` in `a9c677f` to clear the merge conflict.

The conflict was in `src/core/tracking.rs`: `develop` added owner-only DB-file pre-creation (so SQLite derives the `-wal`/`-shm` modes from an already-private file), while this PR extracted `open_at_path` and split the pragmas (`busy_timeout` applied separately from `journal_mode=WAL`). Both are orthogonal — the merged `open_at_path` now pre-creates the private file **and** applies the pragmas independently. Also dropped a now-needless `&` on `restrict_db_files(db_path)` since the extracted helper takes `&Path`.

Validation (local `cargo fmt --all --check && cargo clippy --all-targets -D warnings && cargo test --all`): all green. The PR's regression test `test_track_records_with_concurrent_reader` and `develop`'s `test_restrict_db_files_covers_wal_sidecars` both pass, confirming the two features coexist.

Issue #2452, opened 2026-06-15
Windows Encoding Bug: RTK Decodes GBK Output as UTF-8


fanqiecd:
# Windows Encoding Bug: RTK Decodes GBK Output as UTF-8

**Summary:** In a native Windows (PowerShell) environment, RTK is mis-decoding child process output: the console outputs GBK-encoded bytes by default, but RTK currently interprets everything as UTF-8. This causes Chinese text to appear as garbled mojibake. The issue can be reproduced by simple commands and comparing expected vs actual outputs. We provide detailed reproduction steps, sample outputs, diagnostics, and a proposed fix (using `encoding_rs::GBK` or detecting the console code page). 

**Environment:** Windows 10 (or similar), PowerShell (version unspecified), RTK (official WinGet build, version unspecified). Ensure the terminal uses a Chinese code page (e.g. `chcp` returns 936, not 65001). The bug has been observed with standard RTK binaries (e.g. 0.28.2 for Windows MSI). 

## Reproduction Steps

1. **Verify console encoding:**  
   In PowerShell, run `chcp` and check `[Console]::OutputEncoding`. Example:  
   ```powershell
   chcp                 # Expected: Active code page: 936 (GBK)
   [Console]::OutputEncoding  # Expected: CodePage: 65001 (UTF8) or 936
   ```  
   If the code page is not UTF-8 (936 for GBK) and PowerShell’s encodings are not set to UTF-8, child processes will emit GBK-encoded bytes.  

2. **Create a test file with Chinese content:**  
   ```powershell
   "测试中文内容" | Out-File test.txt -Encoding gbk
   type test.txt         # Should display: 测试中文内容
   ```  
   The above should show correct Chinese text in PowerShell. If already garbled, fix encoding or use `-Encoding utf8`.

3. **Run via RTK and proxy:**  
   ```powershell
   rtk type test.txt
   rtk proxy type test.txt
   ```  
   - **Expected (correct) behavior:** Both commands display “测试中文内容” exactly (as in step 2).  
   - **Actual (garbled) behavior:**  
     - `rtk type test.txt` yields garbled output like “涔�缁涓�” (mojibake).  
     - `rtk proxy type test.txt` should bypass filtering and match the normal output (“测试中文内容”), which it does.  
   This indicates RTK’s own processing is corrupting the text.  

4. **Compare with system commands:**  
   ```powershell
   type test.txt       # Expected: 测试中文内容
[...]

PR #2488, opened 2026-06-18
fix(gain): char-safe truncation in `gain --history` (panic on CJK)

leekkk2:
## Problem

`rtk gain --history` panics when a recent command string contains a multibyte UTF-8 character (e.g. CJK). The "Recent Commands" column truncates with a byte slice `&rec.rtk_cmd[..22]`, which panics when the 22-byte cut lands inside a multibyte character:

```
end byte index 22 is not a char boundary; it is inside '原' (bytes 20..23 of ...)
```

This reproduces reliably for anyone whose command history contains non-ASCII text (CJK, accented Latin, emoji, …). `rtk gain` without `--history` is unaffected. (Same UTF-8 handling family as #2452.)

## Fix

Replace the byte slice with a char-boundary-safe `truncate_cmd_display()` helper that truncates by `char`, mirroring the existing UTF-8-safe `truncate_str` in `src/discover/report.rs`. Behaviour for ASCII is unchanged (keep 22 + `...`).

## Tests

Added regression tests:
- `truncate_cmd_display_does_not_panic_on_multibyte` — CJK input that previously panicked
- `truncate_cmd_display_keeps_short_commands_unchanged`
- `truncate_cmd_display_truncates_long_ascii`

`cargo test` passes; `cargo fmt --check` clean.

### DIFF
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -11,6 +11,22 @@ use serde::Serialize;
 use std::io::IsTerminal;
 use std::path::PathBuf;
 
+/// Truncate a command string for the fixed-width "Recent Commands" column.
+///
+/// Must be UTF-8 char-boundary safe: slicing by byte index (e.g. `&s[..22]`)
+/// panics when the cut lands inside a multibyte character such as CJK, which
+/// is common in non-English command strings. Truncate by `char` instead.
+fn truncate_cmd_display(cmd: &str) -> String {
+    const MAX_CHARS: usize = 25;
+    const KEEP_CHARS: usize = 22;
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the fix and the clear panic report. This was fixed on develop by #2789 (merged 2026-07-03), which replaced the byte slice in the history command column with the shared char-based core::utils::truncate, and did the same for four other byte-sliced display sites in gain.rs and session_cmd.rs. The CJK and emoji regression cases your tests cover are also covered there by the multibyte tests in src/core/utils.rs, so there is nothing left to port. If you see a truncation site that still panics on the current develop, please comment here or reopen.

PR #2489, opened 2026-06-18
feat(discover): note that Claude hook rewrites are invisible in transcripts

leekkk2:
## Problem

`rtk discover` and `rtk session` derive the "Already using RTK %" and "missed savings" numbers from Claude Code session transcripts. But the PreToolUse hook rewrites commands to `rtk …` at **execution time**, and Claude Code stores the **original (pre-rewrite)** command in the transcript JSONL. So these tools only see a leading `rtk` when the model typed it manually.

On a machine where the hook is installed this is badly misleading: adoption can read ~1% and "missed savings" can show large token totals, even though those commands are in fact auto-rewritten and saved at runtime — the real figure is in `rtk gain`.

Verified locally: a transcript records `grep -rn …` (not `rtk grep`), yet the executed output is RTK-compacted and `rtk gain` counts the command.

## Fix

Detect the Claude Code hook (the native `rtk hook claude` command in `settings.json` / `settings.local.json`) and print a clarifying note in discover output, mirroring the existing cursor / hermes / copilot notes, pointing users to `rtk gain` for the actual savings RTK applied.

This does not change any counts (avoids guessing whether history predates hook installation) — it just stops the headline numbers from being read as "RTK isn't working".

## Tests

Added:
- `test_agent_status_detects_claude_hook_in_settings`
- `test_agent_status_no_claude_hook_when_absent`
- `test_format_text_reports_claude_detected`
- extended `test_format_json_includes_agent_status` for the new field

Full suite passes (2202); `cargo fmt --check` clean.

### DIFF
--- a/src/discover/report.rs
+++ b/src/discover/report.rs
@@ -1,8 +1,9 @@
 //! Data types for reporting which commands RTK can and cannot optimize.
 
 use crate::hooks::constants::{
-    COPILOT_HOOK_FILE, CURSOR_DIR, GITHUB_DIR, HERMES_DIR, HERMES_PLUGINS_SUBDIR,
+    CLAUDE_DIR, COPILOT_HOOK_FILE, CURSOR_DIR, GITHUB_DIR, HERMES_DIR, HERMES_PLUGINS_SUBDIR,
     HERMES_PLUGIN_MANIFEST_FILE, HERMES_PLUGIN_NAME, HOOKS_SUBDIR, REWRITE_HOOK_FILE,
+    SETTINGS_JSON, SETTINGS_LOCAL_JSON,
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into this. #3206 (merged 2026-09-02) took a different route to the same problem: discover now logs the real PreToolUse hook decision and counts hook-rewritten commands as coverage rather than missed savings, and it prints an explicit caveat for the slice of history that is still estimated from the current hook and config state. That makes the transcript-based numbers no longer understate hook usage, so the advisory note here would now say something that is not true of develop's output. If you think a case is still reported misleadingly, please comment or reopen with the discover output you see.

PR #2490, opened 2026-06-18
feat(yaml): add yaml command and YAML-aware grep

PapaDanielVi:
## What

Adds `rtk yaml` and teaches `rtk grep` to read YAML structurally.

### `rtk yaml <file>`
Linearizes YAML into one dotted-path line per scalar, mirroring the existing `rtk json` command:

```
databases:            ->   databases.base.enabled: true
  base:
    enabled: true
```

`--keys-only` drops the values and shows structure only. Reads from stdin with `-`. Multi-document streams, sequences (indexed), tags, empty containers, and long/multi-line scalars are all handled.

### YAML-aware `rtk grep`
When every path argument is a `.yaml`/`.yml` file, grep matches against the linearized form, so a hit shows its full parent context:

```
$ rtk grep enabled config.yaml
2 matches in 1 files:

config.yaml:databases.base.enabled: true
config.yaml:databases.cache.enabled: false
```

Instead of grepping for `enabled` and reading surrounding lines to figure out which database it belongs to, the answer is on one line.

## Why

Nested YAML is expensive for an agent to read: answering "is the base database enabled?" normally means a grep plus a range-read to recover the parent. The dotted-path form puts the context inline, which is faster and more token efficient.

## Safety / fallback

YAML grep mode only activates for existing `.yaml`/`.yml` files with no match-altering flags. Context (`-A`/`-B`/`-C`), inversion (`-v`), globbing, format flags, invalid regex, and unparseable or unreadable files all fall back to the existing raw `rg`/`grep` path unchanged. Note: in YAML mode the dotted path replaces the source line number as the locator.

## Implementation

- New `serde_norway` dependency (maintained `serde_yaml` fork) for parsing.
- `src/cmds/system/yaml_cmd.rs` exposes `filter_yaml_linear`, reused by grep.
- `rtk yaml` wired into `main.rs` alongside `Json` (enum, routing, operational-command whitelist).

## Tests

`cargo fmt --all && cargo clippy --all-targets && cargo test --all` all green. Covers extension validation, flattening (nested maps, sequences, multi-doc, tags, truncation), keys-only mode, the grep fast path (path resolution, `-i`, exit codes, `max_results`), the activation gating, and the raw-grep fallbacks. Token-savings gate verified across real k8s / GitHub Actions / docker-compose fixtures.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -910,6 +910,7 @@ dependencies = [
  "rusqlite",
  "serde",
  "serde_json",
+ "serde_norway",
  "sha2",
  "tempfile",
  "toml",
@@ -987,6 +988,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
+name = "ryu"
+version = "1.0.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9774ba4a74de5f7b1c1451ed6cd5285a32eddb5cccb8cc655a4e50009e06477f"
+
+[[package]]
 name = "same-file"
 version = "1.0.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


PapaDanielVi:
Closing to self-review on my fork first; will reopen against upstream once reviewed.

Issue #1581, opened 2026-04-28
rtk git push reports "ok (up-to-date)" when push is rejected by GitHub repository rulesets


srs-adamr:
## Summary

`rtk git push` silently swallows GitHub repository ruleset rejections (GH013), reporting `ok (up-to-date)` instead of surfacing the error. This is the same bug class as #1535 (stash exit code swallowing) but for the push handler.

## Steps to Reproduce

1. Configure a GitHub repository with branch rulesets requiring PRs (not legacy branch protection — the newer repository rulesets)
2. Commit locally on a protected branch (e.g. `alpha`)
3. Run `git push origin alpha` through rtk

## Expected Behavior

```
FAILED: git push
remote: error: GH013: Repository rule violations found for refs/heads/alpha.
remote: - Changes must be made through a pull request.
! [remote rejected] alpha -> alpha (push declined due to repository rule violations)
```

Exit code: non-zero

## Actual Behavior

```
ok (up-to-date)
```

Exit code: 0

## Verification

Running `rtk proxy git push origin <branch>:alpha` (bypassing filtering) shows the actual rejection:

```
remote: error: GH013: Repository rule violations found for refs/heads/alpha.
remote: Review all repository rules at https://github.com/...
remote: 
remote: - Changes must be made through a pull request.
To https://github.com/...
 ! [remote rejected] ... -> alpha (push declined due to repository rule violations)
error: failed to push some refs to '...'
```

Exit code: 1

## Analysis

The `run_push` handler at `src/cmds/git/git.rs:972` checks `output.status.success()` and only prints the summarized "ok" output when exit code is 0. Two possible root causes:

1. **Git transport exit code inconsistency**: GitHub's newer repository rulesets (not legacy branch protection) may cause `git push` to return exit 0 even though the server rejects the push via the `remote:` error stream. The `run_push` handler trusts the exit code, sees `Everything up-to-date` in stderr (since no refs were actually transferred), and reports success.

2. **Stderr pattern not checked**: Even if exit code is 0, stderr contains `remote: error:`, `! [remote rejected]`, and `GH013` — all clear failure signals that `run_push` should detect.

## Suggested Fix

In addition to checking `output.status.success()`, the push handler should scan stderr for rejection patterns before reporting success:

- `! [remote rejected]`
- `remote: error:`
- `push declined`
- `GH013`

If any of these appear, treat it as a failure regardless of exit code — print the full stderr and return non-zero.

[...]

PR #2496, opened 2026-06-18
fix(git): propagate exit code when commit fails instead of reporting ok

### LINKED ISSUE #2494: rtk git commit reports `ok` / exit 0 when the commit fails (nothing staged, or pre-commit hook abort)
## Summary

`rtk git commit` swallows commit failures, printing `ok (...)` and returning exit code **0** instead of surfacing the error and git's non-zero exit. Same bug class as #1581 (push) and #1535 (stash), but for the commit handler.

Related: #1581, #1535, #2446.

## Steps to Reproduce

**Case A — nothing staged (minimal repro):** in a clean repo,

```console
$ git commit -m "should fail"      # via rtk
ok (nothing to commit)
$ echo $?
0
```

**Case B — commit aborted by a pre-commit hook (real-world impact):** with a `pre-commit`/`prek` hook that fails or modifies files (e.g. a formatter, or a failing shellcheck hook), the commit is aborted but rtk reports:

```console
ok 2 files changed, 72 insertions(+)
```

exit 0 — while `HEAD` has not moved.

## Expected

```
FAILED: git commit
nothing to commit, working tree clean
```

Exit code: non-zero.

## Verification (raw passthrough preserves the real status)

```console
$ rtk proxy git commit -m "should fail"
$ echo $?
1
```

## Analysis

Like `run_push` in #1581, the commit handler appears to print the summarized `ok` output without gating on `output.status.success()`. Two failure modes both get masked:

- git exits non-zero (nothing to commit) → rtk still prints `ok` + exit 0.
- a pre-commit hook aborts the commit → rtk prints `ok N files changed` (echoing the staged diff) + exit 0, so it looks like the commit landed.

## Impact

Automation/agents (and humans) that key on the exit code or the `ok` marker treat failed commits as successful. In an agent workflow this silently drops commits: the agent believes the commit landed, moves on, and the work is only in the index. The sole reliable signal is re-checking `git rev-parse HEAD`.

## Suggested Fix

Return the wrapped process's exit status; print the `ok` summary only when `exit == 0`. On non-zero exit (or when stderr shows abort signals like `nothing to commit` / a hook failure), pass through git's stderr and exit code.

## Environment

rtk 0.42.4 (Homebrew, macOS), via the Claude Code hook integration.

hgunduzoglu:
## Summary

Fixes #2494. `rtk git commit` reported `ok (...)` and returned **exit 0** when the commit actually failed — same masking-failure class as #1581 (push) and #1535 (stash).

## Root cause

`run_commit` branched like this:

```rust
if output.status.success() { /* ok <hash> */ }
else if stderr/stdout contains "nothing to commit" { println!("ok (nothing to commit)"); }  // ← bug
else { /* surface output; return Ok(exit_code) */ }
Ok(0)   // ← the success AND "nothing to commit" branches both land here
```

The `nothing to commit` arm is only reachable when `output.status.success()` is false (git exits **1** when there's nothing staged), yet it printed `ok` and fell through to `Ok(0)`. A pre-commit hook abort hit the same masking.

## Fix

Gate strictly on `output.status.success()` via a small `CommitOutcome` helper. Success keeps the compact `ok <hash>` line; **any** failure surfaces git's own message and propagates its real exit code (with a `FAILED: git commit` header, matching `run_add`/`run_pull`/`run_stash`). No success path can return non-zero, no failure path can return 0.

```console
# before
$ rtk git commit -m x      # nothing staged
ok (nothing to commit)      # exit 0  ✗

# after
$ rtk git commit -m x
FAILED: git commit
On branch main
nothing to commit, working tree clean
                            # exit 1  ✓  (matches native git)
```

## Testing

- TDD: 4 unit tests on the extracted `classify_commit_outcome` (success→`ok <hash>`, empty stdout→`ok`, nothing-to-commit→`Failed(1)`, hook-abort→`Failed(2)`).
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — clean, 2203 passed.
- Manual verification against native git for all three paths:
  - nothing staged → `FAILED` + exit 1 (was exit 0)
  - real commit → `ok <hash>` + exit 0 (no regression)
  - pre-commit hook abort → `FAILED` + exit 1

No output-format change on the success path, so token savings are unaffected.

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1024,39 +1024,56 @@ fn run_commit(args: &[String], verbose: u8, global_args: &[String]) -> Result<i3
     let exit_code = exit_code_from_output(&output, "git commit");
     let raw_output = format!("{}\n{}", stdout, stderr);
 
-    if output.status.success() {
-        // Extract commit hash from output like "[main abc1234] message"
-        // or "[main (root-commit) abc1234] message" (incl. localized variants)
-        // The hash is always the last whitespace-separated token before ']'.
-        let compact = if let Some(line) = stdout.lines().next() {
-            parse_commit_output(line)
-        } else {
-            "ok".to_string()
-        };
-
-        println!("{}", compact);
-
-        timer.track(&original_cmd, "rtk git commit", &raw_output, &compact);
-    } else if stderr.contains("nothing to commit") || stdout.contains("nothing to commit") {
-        println!("ok (nothing to commit)");
-        timer.track(
-            &original_cmd,
-            "rtk git commit",
-            &raw_output,
-            "ok (nothing to commit)",
-        );
-    } else {
-        if !stderr.trim().is_empty() {
-            eprint!("{}", stderr);
+    match classify_commit_outcome(output.status.success(), &stdout, exit_code) {
+        CommitOutcome::Ok(compact) => {
+            println!("{}", compact);
+            timer.track(&original_cmd, "rtk git commit", &raw_output, &compact);
+            Ok(0)
         }
-        if !stdout.trim().is_empty() {
-            eprint!("{}", stdout);
+        // A failed commit must never be reported as "ok". This covers "nothing
+        // to commit" (git exits 1) and pre-commit hook aborts: surface git's
+        // own message and propagate its exit code, like native git (#2494).
+        CommitOutcome::Failed(code) => {
+            eprintln!("FAILED: git commit");
+            if !stderr.trim().is_empty() {
+                eprint!("{}", stderr);
+            }
+            if !stdout.trim().is_empty() {
+                eprint!("{}", stdout);
+            }
+            timer.track(&original_cmd, "rtk git commit", &raw_output, &raw_output);
+            Ok(code)
         }
-        timer.track(&original_cmd, "rtk git commit", &raw_output, &raw_output);
-        return Ok(exit_code);
     }
+}
 
-    Ok(0)
+/// Outcome of a `git commit` invocation. A non-success status must never be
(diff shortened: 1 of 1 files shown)


src/cmds/git/git.rs
```diff
@@ -1024,39 +1024,51 @@ fn run_commit(args: &[String], verbose: u8, global_args: &[String]) -> Result<i3
...
+    match classify_commit_outcome(output.status.success(), &stdout, exit_code) {
+        CommitOutcome::Ok(compact) => {
+            println!("{}", compact);
+            timer.track(&original_cmd, "rtk git commit", &raw_output, &compact);
+            Ok(0)
         }
-        if !stdout.trim().is_empty() {
-            eprint!("{}", stdout);
+        CommitOutcome::Failed(code) => {
+            eprintln!("FAILED: git commit");
```
aeppling: drop custom error messages, should return the git error message

src/cmds/git/git.rs
```diff
@@ -1024,39 +1024,51 @@ fn run_commit(args: &[String], verbose: u8, global_args: &[String]) -> Result<i3
...
     }
+}
 
-    Ok(0)
+/// Outcome of a `git commit`: a non-success status propagates the exit code
+/// rather than being reported as "ok" (#2494).
+enum CommitOutcome {
+    /// Commit succeeded; carries the compact one-line summary for stdout.
+    Ok(String),
+    /// Commit failed; carries git's real exit code to propagate.
```
aeppling: clean up code comments, most of code is explicit enough

aeppling:
Otherwise LGTM :) Thanks for your contribution to make git filter more transparent and faithful

aeppling (approved):

PR #2498, opened 2026-06-18
fix(git): propagate exit code on git status failure in compact path

hgunduzoglu:
Part of #2497 (1/3).

## Problem

`rtk git status` (compact path) reported a clean working tree with **exit 0** when `git status` actually failed for any reason other than "not a git repository" — e.g. a corrupt index, a lock conflict, or unreadable refs. The non-compact path already guards on `result.success()` (git.rs:824); the compact path only special-cased the not-a-repo string and let every other failure fall through to the formatted summary + `Ok(0)`.

```console
# repo with a corrupted .git/index
$ git status            # native
fatal: .git/index: index file smaller than expected     # exit 128
$ rtk git status        # before this PR
Clean working tree                                       # exit 0  ✗
```

## Fix

Replace the narrow `"not a git repository"` check with a general `if !result.success()` guard (mirroring the non-compact path). The friendly `Not a git repository` message is preserved for that common case; any other failure surfaces git's stderr and returns its real exit code.

```console
$ rtk git status        # after
fatal: .git/index: index file smaller than expected
                        # exit 128  ✓ (matches native git)
```

## Testing

- New test `test_run_status_compact_propagates_non_repo_failure`: inits a temp repo, corrupts `.git/index`, asserts `run_status` returns a non-zero exit (uses `git -C` via global args; assertion is `!= 0` so it's exit-code-agnostic and cross-platform).
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — clean, 2200 passed.
- Manual verification vs native git:
  - corrupt index → exit 128 (was 0)
  - clean repo → `* main / clean — nothing to commit`, exit 0 (no regression)
  - dirty repo → `?? f2`, exit 0 (no regression)
  - non-git dir → `Not a git repository`, exit 128 (friendly message preserved)

Same masking-failure class as #1581 / #1535 / #2446 / #2494. Sibling fixes for `run_worktree` and `run_stash` will follow as separate PRs under #2497.

Issue #2497, opened 2026-06-18
git read handlers mask failures: status/worktree/stash report exit 0 + success output when git fails


hgunduzoglu:
## Summary

Three `git` read handlers report **success (exit 0) with success-looking output** when the underlying git command actually fails (non-zero exit). Same masking-failure class as #1581 (push), #1535 (stash push/pop), #2446 (diff), and #2494 (commit). An agent reading the output is told the repo is fine when git errored.

I found these via a systematic sweep of the git handlers in `src/cmds/git/git.rs` and verified each empirically against native git. In every case a sibling branch in the *same function* already handles failure correctly — the asymmetry is the bug.

### 1. `run_status` compact path — `src/cmds/git/git.rs:855-911`
The compact path only catches the `"not a git repository"` error string; **any other** failure falls through to format `result.stdout` (empty → `Clean working tree`) and `Ok(0)`. The non-compact path (line 824) correctly does `if !result.success() { return Ok(result.exit_code); }` — the compact path lacks that general guard.

```console
# repo with a corrupted .git/index
$ git status            # native
fatal: ...              # exit 128
$ rtk git status
Clean working tree      # exit 0   ✗  (lies about repo integrity)
```

### 2. `run_worktree` list mode — `src/cmds/git/git.rs:1729-1743`
List mode never checks `result.success()`; prints filtered stdout and returns `Ok(0)`. The `has_action` branch right above (line 1717) correctly checks success and returns `result.exit_code`.

```console
$ cd /non-git-dir
$ git worktree list     # native → exit 128 (fatal: not a git repository)
$ rtk git worktree
                        # empty output, exit 0   ✗
```

### 3. `run_stash` list/show — `src/cmds/git/git.rs:1539-1575`
The `list` and `show` branches print output and return `Ok(0)` with no `result.success()` check. Empty stdout on failure prints `No stashes` / `Empty stash`.

```console
$ cd /non-git-dir
$ git stash list        # native → exit 128
$ rtk git stash list
No stashes              # exit 0   ✗
```

## Not affected (checked)
- `run_diff` — the `git diff --stat` invocation **is** guarded (line 165); the follow-up `git diff` runs with identical args, so it cannot realistically fail while the first succeeds. No user-facing masking.

## Fix shape
Each fix mirrors the already-correct sibling branch: after `exec_capture`, `if !result.success() { surface stderr; return Ok(result.exit_code); }` before the format/print/`Ok(0)` fall-through.

[...]

PR #2499, opened 2026-06-18
fix(git): propagate exit code on git worktree list failure

hgunduzoglu:
Part of #2497 (2/3).

## Problem

`rtk git worktree` (list mode) never checked `result.success()` — a failed `git worktree list` (e.g. run outside a repository) was flattened to empty output + **exit 0**. The `has_action` branch right above it (add/remove/prune/…) already guards on success and returns the real exit code; only list mode was missing the guard.

```console
$ cd /non-git-dir
$ git worktree list     # native → exit 128 (fatal: not a git repository)
$ rtk git worktree      # before this PR
                        # empty output, exit 0  ✗
```

## Fix

Add `if !result.success()` before the format/print/`Ok(0)` fall-through: surface `git`'s stderr (with a `FAILED: git worktree list` header, matching the sibling handlers) and return `result.exit_code`.

```console
$ rtk git worktree      # after
FAILED: git worktree list
fatal: not a git repository (or any of the parent directories): .git
                        # exit 128  ✓
```

## Testing

- New test `test_run_worktree_list_propagates_failure`: runs `run_worktree` against a non-git temp dir via `git -C`, asserts a non-zero exit (`!= 0`, exit-code-agnostic / cross-platform).
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — clean, 2200 passed.
- Manual verification: non-git dir → exit 128 (was 0); real repo → worktree list + exit 0 (no regression).

Same masking-failure class as #1581 / #1535 / #2446 / #2494. `run_status` is #2498; `run_stash` PR follows.

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1715,7 +1715,6 @@ fn run_worktree(args: &[String], verbose: u8, global_args: &[String]) -> Result<
     let result = exec_capture(&mut cmd).context("Failed to run git worktree list")?;
 
     if !result.success() {
-        eprintln!("FAILED: git worktree list");
         if !result.stderr.trim().is_empty() {
             eprintln!("{}", result.stderr);
         }
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1714,9 +1714,6 @@ fn run_worktree(args: &[String], verbose: u8, global_args: &[String]) -> Result<
     cmd.args(["worktree", "list"]);
     let result = exec_capture(&mut cmd).context("Failed to run git worktree list")?;
 
-    // A failed `git worktree list` (e.g. run outside a repo) must propagate, not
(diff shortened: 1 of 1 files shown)


src/cmds/git/git.rs
```diff
@@ -1714,6 +1714,20 @@ fn run_worktree(args: &[String], verbose: u8, global_args: &[String]) -> Result<
     cmd.args(["worktree", "list"]);
     let result = exec_capture(&mut cmd).context("Failed to run git worktree list")?;
 
+    if !result.success() {
+        eprintln!("FAILED: git worktree list");
```
aeppling: The result should already contain the git error and if not we should just be transparent, writing our own error message like this is not a good idea

aeppling:
@hgunduzoglu  Same problem on #2500 , otherwise LGTM :)

aeppling:

aeppling (approved):

PR #2502, opened 2026-06-18
fix(dotnet): stop duplicating failures on failing test runs (#2501)

### LINKED ISSUE #2501: dotnet test filter duplicates failure output on failing runs
## Summary

> _This issue was drafted with the help of Claude Code, based on a real benchmark and a walkthrough of the source.

`rtk dotnet test` inflates output on failing test runs by printing each failure **twice**. On a controlled benchmark it produced **+65% more tokens than raw output** — the opposite of RTK's purpose. The more failures, the worse the inflation (linear duplication).

Tested against a .NET solution using **xUnit v3 + Microsoft.Testing.Platform (MTP)**.

## Benchmark

5 failing tests (722 total, 717 pass / 5 fail):

| Command | Bytes | ~Tokens | vs raw |
|---|---|---|---|
| `dotnet test` (raw) | 10100 | ~2525 | — |
| `rtk dotnet test` | 16659 | ~4165 | **+65%** 🔴 |

Measured proof — count of a unique assertion string from the 5 failures:

- raw: **5** (each failure once)
- rtk: **10** (each failure printed twice)

For comparison, on all-green runs RTK is excellent (−99.8%, one-line summary). The regression is specific to **failing** runs.

## Root cause

`src/cmds/dotnet/dotnet_cmd.rs`, in `run_dotnet_with_binlog` (lines ~211-223):

```rust
let output_to_print = if !command_success {
    let stdout_trimmed = result.stdout.trim();
    let stderr_trimmed = result.stderr.trim();
    if !stdout_trimmed.is_empty() {
        format!("{}\n\n{}", stdout_trimmed, filtered)  // ← prepends FULL raw stdout
    } else if !stderr_trimmed.is_empty() {
        format!("{}\n\n{}", stderr_trimmed, filtered)
    } else {
        filtered
    }
} else {
    filtered
};
```

On any failing run, the filter prepends the **entire raw stdout** in front of the filtered output. For test runs this duplicates everything:

1. **Copy #1** — raw stdout already contains every per-test failure block (message + stack). The MTP / xUnit v3 runner prints these inline.
2. **Copy #2** — `filtered` is `format_test_output()`, which rebuilds a `Failed Tests:` section from the parsed TRX (or console text). Same failures again.

So failing test output is roughly doubled, plus the extra `Warnings:` section, `.trx` paths, and summary line.

### Why only `test` is affected

`build` and `restore` source their errors from a single place (binlog/text parse), so the raw-stdout prepend doesn't duplicate anything. `test` has **two** failure sources — raw stdout *and* the structured `Failed Tests:` section — so the prepend collides with the filter's own output.

### Why existing tests didn't catch it

The unit tests exercise `format_test_output()` in isolation. The bug lives in the `run_dotnet_with_binlog` orchestration (the prepend branch), which has no test coverage.

## Suggested fix

Make the raw-stdout fallback conditional. The `Failed Tests:` section already carries the failure detail (name + message + clipped stack), built from two sources:

- TRX — `parse_trx_content` (`dotnet_trx.rs`): message + first 3 stack lines.
- Console text — `parse_test_from_text` (`binlog.rs`): message lines, capped at 20.

When that section carries detail, the raw-stdout prepend is pure duplication and should be dropped. When it doesn't (filter is blind), keep the raw fallback so nothing is lost.

```rust
let structured_failures_have_detail = !summary.failed_tests.is_empty()
    && summary.failed_tests.iter().all(|t| !t.details.is_empty());

let needs_raw_fallback = match subcommand {
    // TRX/console parse already captured failures WITH detail into the
    // `Failed Tests:` section; the raw-stdout prepend would only duplicate it.
    // If any failure lacks detail, keep the raw fallback so nothing is lost.
    "test" => !structured_failures_have_detail,
    _ => true,
};

let output_to_print = if !command_success && needs_raw_fallback {
    // existing prepend logic
} else {
    filtered
};
```

### Why this condition (regression analysis)

Cases traced through the code, confirming no information is lost:

| Scenario | `failed_tests` | Behavior | Result |
|---|---|---|---|
| Normal failing test (TRX/text parsed with detail) | populated, detail present | drop raw prepend | ✅ deduplicated, detail preserved in section |
| Failed test with **empty** `details` (self-closing `UnitTestResult`, no `ErrorInfo`, text parse found nothing) | populated, detail empty | keep raw fallback | ✅ raw message not lost |
| Compile/build failure, no tests ran | empty | keep raw fallback | ✅ build errors preserved |
| Crash / no TRX emitted (`normalize_test_summary` sets `failed=1` but adds nothing to `failed_tests`) | empty | keep raw fallback | ✅ raw preserved |
| `counts_unavailable` (all zero) | empty | keep raw fallback | ✅ raw preserved |

`Warnings:` / `Errors:` sections are sourced from the build summary (binlog/text), independent of the prepend, so they remain shown exactly once in all cases.

> A naive `summary.failed_tests.is_empty()` condition would regress the **empty-`details`** row above — it would drop the raw stdout even though the section shows only test names. The `.all(|t| !t.details.is_empty())` guard prevents that.

### Known residual edge case

If a run populates `failed_tests` from TRX **with detail** but writes additional diagnostics to **stderr with an empty stdout**, those stderr diagnostics are dropped. This is rare in practice (MTP/xUnit print failures to stdout, so stdout is non-empty whenever structured detail exists) and the original code already discarded stderr whenever stdout was non-empty. Noting it for completeness rather than complicating the condition.

## Test coverage to add

A regression test should exercise the `run_dotnet_with_binlog` orchestration path (not just `format_test_output` in isolation) with a real MTP / xUnit-v3 failing-run fixture, asserting each failure string appears **exactly once** in the final output.

## Environment

- Test stack: xUnit v3 + Microsoft.Testing.Platform (MTP)
- Affects: failing `dotnet test` runs (any runner mode that emits inline failures to stdout)

Tailoo:
## Summary

Fixes #2501 — on failing `dotnet test` runs the orchestrator prepended the **full raw stdout** ahead of the filtered summary. The `Failed Tests:` section already reproduces each failure (name + message + clipped stack) parsed from TRX/console, so every failure was printed **twice** — inflating output **+65% vs raw**, scaling linearly with failure count.

## Fix

Gate the raw-stdout prepend behind a new `test_needs_raw_fallback`:

- **Skip** the prepend when the structured `Failed Tests:` section already carries detail → dedup.
- **Keep** it when the filter is blind: no failures parsed (build failure / crash where nothing reaches `failed_tests`), or a parsed failure has empty detail (e.g. a self-closing `<UnitTestResult>` with no `<ErrorInfo>`). Nothing is lost.

`build` / `restore` / passthrough keep prior behavior (`needs_raw_fallback = true`).

## Why this shape

The bug lived in `run_dotnet_with_binlog` — the one dotnet layer with **no test coverage** (it spawns real `dotnet`). Consistent with the existing dotnet test pattern, the decision is extracted into two pure, unit-tested fns rather than adding a net-new orchestration/e2e harness (no dotnet filter has one):

- `test_needs_raw_fallback(&summary) -> bool`
- `compose_failure_output(success, needs_raw_fallback, stdout, stderr, filtered) -> String`

## Regression analysis (traced through code)

| Scenario | `failed_tests` | Behavior | Result |
|---|---|---|---|
| Normal failing test (detail parsed) | populated, detail present | drop raw prepend | deduped, detail preserved |
| Failure with empty `details` | populated, detail empty | keep raw fallback | raw message not lost |
| Compile/build failure, no tests ran | empty | keep raw fallback | build errors preserved |
| Crash / no TRX | empty | keep raw fallback | raw preserved |
| `counts_unavailable` | empty | keep raw fallback | raw preserved |

`Warnings:` / `Errors:` sections come from the build summary, independent of the prepend → still shown once.

## Tests

7 new unit tests (all green), covering the dedup decision and every fallback branch. `cargo fmt` + `cargo clippy --all-targets` clean, `cargo test --all` passes.

### DIFF
--- a/src/cmds/dotnet/dotnet_cmd.rs
+++ b/src/cmds/dotnet/dotnet_cmd.rs
@@ -138,7 +138,7 @@ fn run_dotnet_with_binlog(subcommand: &str, args: &[String], verbose: u8) -> Res
     let raw = format!("{}\n{}", result.stdout, result.stderr);
     let command_success = result.success();
 
-    let filtered = match subcommand {
+    let (filtered, needs_raw_fallback) = match subcommand {
         "build" => {
             let binlog_summary = if should_expect_binlog && binlog_path.exists() {
                 normalize_build_summary(
@@ -151,7 +151,7 @@ fn run_dotnet_with_binlog(subcommand: &str, args: &[String], verbose: u8) -> Res
             let raw_summary =
                 normalize_build_summary(binlog::parse_build_from_text(&raw), command_success);
             let summary = merge_build_summaries(binlog_summary, raw_summary);
-            format_build_output(&summary, &binlog_path)
+            (format_build_output(&summary, &binlog_path), true)
         }
         "test" => {
             // First try to parse from binlog/console output
@@ -181,11 +181,18 @@ fn run_dotnet_with_binlog(subcommand: &str, args: &[String], verbose: u8) -> Res
             let raw_diagnostics =
                 normalize_build_summary(binlog::parse_build_from_text(&raw), command_success);
             let test_build_summary = merge_build_summaries(binlog_diagnostics, raw_diagnostics);
-            format_test_output(
-                &summary,
-                &test_build_summary.errors,
-                &test_build_summary.warnings,
-                &binlog_path,
+            // The `Failed Tests:` section already carries failure detail parsed from
+            // TRX/console; skip the raw-stdout prepend when it would only duplicate it.
+            // See issue #2501.
+            let needs_raw = test_needs_raw_fallback(&summary);
+            (
+                format_test_output(
+                    &summary,
+                    &test_build_summary.errors,
+                    &test_build_summary.warnings,
+                    &binlog_path,
+                ),
+                needs_raw,
             )
         }
         "restore" => {
@@ -203,24 +210,21 @@ fn run_dotnet_with_binlog(subcommand: &str, args: &[String], verbose: u8) -> Res
 
             let (raw_errors, raw_warnings) = binlog::parse_restore_issues_from_text(&raw);
 
-            format_restore_output(&summary, &raw_errors, &raw_warnings, &binlog_path)
+            (
+                format_restore_output(&summary, &raw_errors, &raw_warnings, &binlog_path),
+                true,
+            )
         }
-        _ => raw.clone(),
+        _ => (raw.clone(), true),
     };
 
-    let output_to_print = if !command_success {
-        let stdout_trimmed = result.stdout.trim();
-        let stderr_trimmed = result.stderr.trim();
-        if !stdout_trimmed.is_empty() {
-            format!("{}\n\n{}", stdout_trimmed, filtered)
-        } else if !stderr_trimmed.is_empty() {
-            format!("{}\n\n{}", stderr_trimmed, filtered)
-        } else {
-            filtered
-        }
-    } else {
-        filtered
-    };
+    let output_to_print = compose_failure_output(
+        command_success,
+        needs_raw_fallback,
+        &result.stdout,
+        &result.stderr,
+        &filtered,
+    );
 
     println!("{}", output_to_print);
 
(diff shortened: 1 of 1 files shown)


aeppling:
Hey @Tailoo , thanks for addressing this issue.

Another concurrent PR #2511 , in favor of this one for the cleaner factoring.

Blocker before merge: test_needs_raw_fallback drops the raw fallback whenever every listed failure carries detail, but never checks the list is complete against summary.failed.

Fix: keep the raw fallback when the structured section is numerically incomplete, add a failed_tests.len() < summary.failed clause to the condition.

aeppling (approved):

aeppling:
Hey, this look good to me, thanks for contributing to RTK :)

PR #2504, opened 2026-06-18
  Add Nix flake for building from source

arnemileswinter:
- Adds flake.nix so RTK can be built declaratively on NixOS and any
  Nix-enabled system using rustPlatform.buildRustPackage and the
  existing Cargo.lock
  - Reads name, version, description, and homepage directly from
  Cargo.toml, zero maintenance needed on releases
  - Exposes packages.default, devShells.default, and overlays.default

# Test plan
  - [x] Manual testing: rtk <command> output inspected
  - [x] nix build succeeds from the repo root
  - [x] nix run . -- git status produces filtered output

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1781577229,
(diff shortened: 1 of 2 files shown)


srghma:
works perfectly

PR #2505, opened 2026-06-19
fix(discover): route `rg` to `rtk rg`, not `rtk grep`

multicam:
## Problem

`rg` and `grep` share the `rtk grep` rule (`pattern: r"^(rg|rtk rg)\s+"`), so `rtk rewrite` downgrades `rg …` → `rtk grep …`. But `rtk grep` is a POSIX-grep passthrough: it is **not recursive** and **rejects ripgrep-only flags** (`--hidden`, `--glob`/`-g`, `--type`). The result:

- `rg --hidden foo` → `rtk grep --hidden foo` → `unrecognised option '--hidden'`
- `rg foo` (recursive intent) → `rtk grep foo` → reads stdin instead of recursing

In agent/LLM tooling this reliably gets mis-read as *"rg isn't installed"* (it is) and triggers retry loops.

## Fix

When the matched prefix is `rg`, route to **`rtk rg`** (rtk's faithful ripgrep wrapper) instead of `rtk grep`, preserving every ripgrep flag and rg's recursive/gitignore semantics. A real `grep` is unaffected — only the `rg` prefix is rerouted. Per-segment, so `rg foo | grep bar` rewrites only the `rg` half.

```
rg --hidden -g '*.rs' foo   →  rtk rg --hidden -g '*.rs' foo   (was: rtk grep …)
grep -rn foo .              →  rtk grep -rn foo .              (unchanged)
```

## Tests

+3 unit tests in `discover::registry::tests`:
- `test_rewrite_rg_pattern` — `rg`→`rtk rg`
- `test_rewrite_rg_preserves_ripgrep_flags` — `--hidden`/`-g` survive
- `test_rewrite_grep_still_maps_to_rtk_grep` — real grep untouched

Sibling to #2453 (compound-`find` rewrite); independent change, applies cleanly on `develop`.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -848,10 +848,22 @@ fn rewrite_segment_inner(
     // Try each rewrite prefix (longest first) with word-boundary check
     for &prefix in rule.rewrite_prefixes {
         if let Some(rest) = strip_word_prefix(cmd_part, prefix) {
-            let rewritten = if rest.is_empty() {
-                format!("{}{}", rule.rtk_cmd, redirect_suffix)
+            // `rg` and `grep` share the `rtk grep` rule, but `rtk grep` is a
+            // POSIX-grep passthrough: it is NOT recursive and rejects ripgrep-only
+            // flags (--hidden/--glob/--type/-g). Downgrading `rg`→`rtk grep` breaks
+            // ripgrep's semantics — flags bomb out and the bare form reads stdin
+            // instead of recursing (an agent then mis-reads it as "rg isn't
+            // installed"; it is). rtk ships a faithful `rtk rg` (real ripgrep), so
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for tracking this down and for the clear before and after. develop already routes rg to rtk rg: #2641 (merged 2026-06-26) stopped substituting one search engine for the other, and the registry rule landed with it. On the current build `rtk rewrite "rg --hidden -g '*.rs' foo"` returns `rtk rg --hidden -g '*.rs' foo` while `grep -rn foo .` still returns `rtk grep -rn foo .`, and the registry tests cover the same routing your tests added. If you still see rg downgraded to rtk grep on the latest develop, please comment here or reopen.

PR #2506, opened 2026-06-19
feat(read): support for offset and slice read

### LINKED ISSUE #1104: `rtk read` does not support reading specific lines

datibbaw:
## Summary

Adds windowed read support to `rtk read`, addressing two related needs:

- **Named flags** for explicit use: `--offset <N>` (start from line N, 1-based), `--limit <N>` (read exactly N lines), combinable for a precise slice
- **Positional line range** for Claude compatibility: `rtk read <file> <start> <end>` (inclusive end), and `rtk read <file> <start>` (read to EOF) — matches the form Claude emits when instructed to prefer `rtk read` over its built-in Read tool (fixes #1104)

Internal changes:
- `LineRange` enum unifies all windowing modes (Full, Head, Tail, From, Slice, Empty)
- `apply_line_window` returns `(line_start, content)` so `--line-numbers` / `-n` displays correct line numbers for windowed output
- `format_with_line_numbers` accepts a `display_start` offset instead of always starting at 1
- `resolve_positional_line_range` detects trailing numeric positionals in `files` and rewrites them to `offset`/`limit` — only fires for single-file invocations; multi-file calls are left untouched

## Test plan

- [x] Unit tests for `resolve_positional_line_range` (start+end, start-only, multi-file passthrough, no-numerics, end==start, end<start)
- [x] Manual: `rtk read <file> --offset 10`
- [x] Manual: `rtk read <file> --offset 10 --limit 20`
- [x] Manual: `rtk read <file> --tail-lines 10 -n` — correct line numbers
- [x] Manual: `rtk read <file> 10 20` — positional form
- [x] Manual: `rtk read <file> 10` — offset-only positional form

### DIFF
--- a/src/cmds/system/read.rs
+++ b/src/cmds/system/read.rs
@@ -7,11 +7,36 @@ use anyhow::{Context, Result};
 use std::fs;
 use std::path::Path;
 
+enum LineRange {
+    Full,
+    Empty,
+    Head(usize),           // n > 0
+    Tail(usize),           // n > 0
+    From(usize),           // start >= 0
+    Slice((usize, usize)), // start >= 0, limit > 0
+}
+
+impl LineRange {
+    fn from_args(offset: Option<usize>, limit: Option<usize>, tail_lines: Option<usize>) -> Self {
+        match (offset, limit, tail_lines) {
+            (_, Some(0), _) | (_, _, Some(0)) => Self::Empty,
+            (Some(offset), Some(limit), _) => Self::Slice((offset.saturating_sub(1), limit)),
+            (Some(offset), None, _) => Self::From(offset.saturating_sub(1)),
+            (None, Some(limit), _) => Self::Head(limit),
+            (None, None, Some(tail)) => Self::Tail(tail),
+            _ => Self::Full,
+        }
+    }
+}
+
+#[allow(clippy::too_many_arguments)]
 pub fn run(
     file: &Path,
     level: FilterLevel,
     max_lines: Option<usize>,
     tail_lines: Option<usize>,
+    offset: Option<usize>,
+    limit: Option<usize>,
     line_numbers: bool,
     verbose: u8,
 ) -> Result<()> {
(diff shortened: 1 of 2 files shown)


reneleonhardt:
Thank you for contributing your fix again! ❤️

Can you confirm that your solution implements the Claude problem which has been reported?

```shell
rtk read <file> <start_line> <end_line>
```

I can't find a corresponding test and `rtk read` addition, Claude doesn't seem to use `offset limit` args, but `start end` (inclusive).

datibbaw:
> Thank you for contributing your fix again! ❤️
> 
> Can you confirm that your solution implements the Claude problem which has been reported?
> 
> ```shell
> rtk read <file> <start_line> <end_line>
> ```
> 
> I can't find a corresponding test and `rtk read` addition, Claude doesn't seem to use `offset limit` args, but `start end` (inclusive).

Thanks for mentioning that! I must've overlooked something.

datibbaw:
@reneleonhardt we've got two options here:
1. Recognise that Claude will guess the interface of `<file> <start> <end>` and fight Clap a little bit to make that work
2. Document `--offset` and `--limit` in the system prompt that we inject

They're not mutually exclusive, though, because 2. relies on users running `rtk init` again, whereas 1. would just work.

reneleonhardt:
Thanks a lot for adding the positional fallback so quickly!

PR #2472, opened 2026-06-17
fix(copilot): add missing 'get' to kubectl example in init template

K0IN:
The COPILOT_INSTRUCTIONS template translated 'kubectl get pods' to 'rtk kubectl pods', dropping the 'get' subcommand. This might confuse ai models, for that we try to keep the normal and rtk command in sync to prevent hallucinations.

Fix: change 'rtk kubectl pods' to 'rtk kubectl get pods' in the COPILOT_INSTRUCTIONS template string.

## Summary

Fix the tempalte to be a faithfull translation of the input.

Fixes #2471

## Test plan

ran a debug build, and used rtk init --copilot

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [x] Manual testing: `rtk <command>` output inspected

> **Important:** All PRs must target the `develop` branch (not `master`).
> See [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) for details.

PR #2507, opened 2026-06-19
fix(init): preserve kubectl get in copilot template

### LINKED ISSUE #2471: rtk init drops kubectl **get** in template in Copilot mode
When using RTK with GitHub Copilot, the `COPILOT_INSTRUCTIONS` template is not a faithful mapping — it drops the `get` subcommand:

`kubectl get pods` gets mapped to `rtk kubectl pods` (missing get).

gauravsaxena1997:
## Summary
- keep the `get` subcommand in the Copilot instructions template for `kubectl get pods`
- add a regression assertion to the Copilot init test so the incorrect `rtk kubectl pods` mapping does not come back

Fixes #2471.

## Verification
- `cargo test copilot_init --quiet`

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -3909,7 +3909,7 @@ git status                 rtk git status
 git log -10                rtk git log -10
 cargo test                 rtk cargo test
 docker ps                  rtk docker ps
-kubectl get pods           rtk kubectl pods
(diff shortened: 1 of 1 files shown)


TaKO8Ki:
Thank you for the contribution. Unfortunately, #2472 fixed the issue.

PR #2522, opened 2026-06-20
fix(grep): report honest match counts in header when --max caps output

### LINKED ISSUE #2436: [grep] header reports raw match counts even when --max caps output
## Problem

In `rtk grep` v0.42.4, the header line "N matches in M files" reports the **raw rg/grep output counts**, not the counts of what RTK is actually showing. This is misleading when the user has set a cap (via `--max` flag or `limits.grep_max_results` in config).

## Reproduction

User asked for `--max 5`, got 5 results, but the header still says "211 matches in 50 files". The footer `[+206 more]` is honest, but the header overstates what the output contains.

## Root cause

`src/cmds/system/grep_cmd.rs:104` sets `total_matches = result.stdout.lines().count()` from the **raw** rg/grep output, before any per-file or global cap is applied. This same value is used in the header at line 122-126:

```rust
let total_matches = result.stdout.lines().count();
...
rtk_output.push_str(&format!(
    "{} matches in {} files:\n\n",
    total_matches,
    by_file.len()
));
```

The cap is enforced later at lines 134 and 140 (`if shown >= max_results { break; }`), so the displayed rows are correctly bounded. Only the header is wrong.

## Suggested fix

After applying caps in the loop, recompute `total_matches` to be `min(raw_total, shown + (total_matches - shown))` or simply use `shown + overflow_count`. Alternatively, show two numbers: `211 matches (5 shown)`.

A minimal change: compute `total_matches = shown + (raw_total - shown).min(remaining_budget)` after the loop, but this changes semantics. Cleaner: rename the header to "N raw matches, K displayed (M truncated)".

## Environment

- rtk 0.42.4 (Windows, C:\Users\Administrator/.local/bin/rtk)
- rg 14.x
- Reproduced against `~/rtk` checkout (commit 878af7d on develop).

## Notes

- Related: Issue #1161 discusses the 200-result limit, but this is a different bug — the cap is being honored at the row level, just misreported in the header.
- Docs mismatch: `src/core/config.rs:124` docstring says `grep_max_results` default is 200, but the `Default` impl at line 139 uses 100. Pick one.

lesbass:
Fixes #2436

## Problem

In `rtk grep`, the header line "N matches in M files" reports the raw rg/grep
match count, even when `--max` or `grep_max_results` caps the displayed output.
For example, with `--max 5` the output shows 5 results but the header says
"211 matches in 50 files".

## Root Cause

The header was constructed before the truncation loop, using the raw parsed
match count (`total_matches`). The loop correctly enforces caps, but the
already-written header was oblivious.

## Fix

Defer the header until after the truncation loop. When output is capped,
the header now shows two numbers:

- Before: `211 matches in 50 files:`
- After: `211 matches (5 shown) in 50 files:`

When output is not capped, the header format is unchanged.

The `[+N more]` footer remains accurate.

## Changes

1 file: `src/cmds/system/grep_cmd.rs` — restructure header construction
to occur after the truncation loop, with a conditional format for capped output.

## Verification

- `cargo fmt --all --check` passes
- Build requires a C compiler (not available in this environment); changes are
  a mechanical restructuring with no logic modifications

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -543,8 +543,9 @@ pub fn run(
     let capped = shown < total_matches || skipped_files > 0;
     let rtk_output = if capped {
         format!(
-            "{} matches in {} files:\n\n{}",
+            "{} matches ({} shown) in {} files:\n\n{}",
             total_matches,
+            shown,
             by_file.len(),
             body
         )


lesbass:
Commit authorship has been corrected: author changed from `RepoScout` to `Stefano Maffeis` (GitHub user lesbass). This should allow the CLA check to properly identify the contributor. The CLA still needs signing via the CLA Assistant link above.

lesbass:
I rebased this PR onto the latest develop branch. Unfortunately, GitHub Actions now requires a maintainer to approve the workflow run for fork PRs after a force push. Could someone approve the run so CI can confirm the change is clean? No code changes — just a rebase to resolve the merge conflict.

lesbass:
Closing this PR after 14 days with no maintainer engagement. The change (fixing the grep --max header count) is small and low-risk — happy to reopen if maintainers are interested at a future date.

lesbass:
Per stale PR policy: 14 days open with no maintainer engagement. Reopen welcome if there's interest.

PR #2532, opened 2026-06-21
fix(grep): disable help flag to allow -h passthrough

### LINKED ISSUE #2391: `rtk grep -h` prints help instead of searching — clap intercepts GNU grep's -h (--no-filename), exit 0, no native fallback
## Summary

`rtk grep` reserves `-h` for `--help` (clap default), but in GNU grep `-h` means `--no-filename` ("suppress the prefixing of file names on output"). When the Claude Code hook rewrites `grep -h PATTERN FILE...` to `rtk grep -h PATTERN FILE...`, clap parses `-h` *successfully* as a help request, so:

- the command prints rtk's usage text instead of search results,
- it exits `0`,
- and — unlike the cases in #1604 — **the native-fallback path never triggers**, because there is no parse failure to recover from.

The agent driving the shell receives help text where it expected matches, with a success exit code. Same family as #2253 (collision causes silently wrong behavior rather than a visible error), but at the clap help-interception level.

## Repro (rtk 0.42.3)

```console
$ printf 'alpha match\nbeta\n' > a.txt
$ printf 'alpha match\ngamma\n' > b.txt

$ grep -h "alpha" a.txt b.txt        # GNU grep: matches without filename prefix
alpha match
alpha match

$ rtk grep -h "alpha" a.txt b.txt    # rtk: prints its own help, exit 0
Compact grep - strips whitespace, truncates, groups by file

...
$ echo $?
0
```

`-h` is a natural flag for an LLM to emit precisely in multi-file searches (where it wants clean, filename-free output to parse), so the hook makes this collision easy to hit.

## Suggested fix

Options, roughly in order of preference:

1. Disable clap's auto `-h` short flag on the `grep` subcommand (`disable_help_flag(true)` + keep `--help` long-only) and treat `-h` as GNU grep's `--no-filename` — rtk's grouped-by-file output could either honor it (omit the file header) or accept-and-ignore it like `-n`, which is already "accepted for grep compatibility".
2. Have the hook rewrite layer strip/translate `-h` before invoking `rtk grep`.
3. At minimum, make `-h` in a rewritten command trigger the native fallback rather than help output.

Workaround for users meanwhile: `rtk proxy grep -h ...`.

axelray-dev:
Fixes #2391.

Add `#[command(disable_help_flag = true)]` to the `Grep` variant so clap does not intercept `-h` as `--help` before `trailing_var_arg` can collect it.

Follows the pattern of merged PR #650 (psql disable_help_flag).

## Problem

`rtk grep -h PATTERN FILE` prints rtk's help text and exits 0 instead of forwarding `-h` to grep/rg. Clap's auto-generated help flag intercepts `-h` before the trailing var-arg collection runs.

## Fix

One-line addition to `src/main.rs`:

```rust
/// Compact grep - strips whitespace, truncates, groups by file
#[command(disable_help_flag = true)]
Grep {
```

## Side effect

`rtk grep --help` forwards to rg/grep help instead of rtk's usage text. This matches the psql behavior.

## Tests

- `rtk grep -h alpha a.txt b.txt` shows search matches, not help text
- `rtk grep --help` forwards to underlying tool help
- `rtk grep -v pattern .` still works
- `cargo test` passes

Complementary to PR #2334 (rewrite-layer passthrough). Both fixes can coexist.

DCO: Signed-off.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -307,6 +307,7 @@ enum Commands {
     },
 
     /// Compact grep - strips whitespace, truncates, groups by file
+    #[command(disable_help_flag = true)]
     Grep {
         /// Max line length
         #[arg(short = 'l', long, default_value = "80")]


KuSh:

PR #2533, opened 2026-06-21
fix(parser): preserve UTF-8 boundaries extracting JSON

### LINKED ISSUE #2509: Panic 'not a char boundary' in output filter (src/parser/mod.rs:191) on large CJK output — affects all wrapped commands (same root as #1837)
## Summary

## Version / Environment

## Panic output

```
```

```bash
rtk read big_cjk.txt
```

## Root cause / suggested fix

YOMXXX:
## Summary
- fix JSON extraction to compute slice boundaries using UTF-8 byte offsets
- add regression coverage for multibyte/CJK JSON strings

Fixes #2509

## Tests
- cargo +1.93.0 test --all

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -349,4 +349,16 @@ Scope: all 6 workspace projects
             extract_json_object(input).expect("Should extract nested JSON with mixed multibyte");
         assert_eq!(extracted, input);
     }
+
+    #[test]
+    fn test_extract_json_object_with_multibyte_string() {
+        let input = r#"prefix
+{"numTotalTests": 1, "message": "测试输出"}
+suffix"#;
+        let expected = r#"{"numTotalTests": 1, "message": "测试输出"}"#;
+
+        let extracted = extract_json_object(input).expect("Should extract JSON");
+
+        assert_eq!(extracted, expected);
+    }
 }


YOMXXX:
Status note: the red `benchmark` check is not caused by this PR's UTF-8 boundary change. The failing row is `curl text`: CI compares `curl -s https://httpbin.org/robots.txt` at 8 tokens with `rtk curl https://httpbin.org/robots.txt` at 40 tokens, so the external network fixture can make the raw output smaller than the RTK path and trip the negative-savings gate.

YOMXXX:

YOMXXX:
Ready for review/merge. Rebased onto current develop, conflicts resolved, and the fresh CI run is green, including benchmark and all platform tests. Scope remains the UTF-8-safe JSON extraction regression coverage.

src/parser/mod.rs
```diff
@@ -349,4 +349,16 @@ Scope: all 6 workspace projects
             extract_json_object(input).expect("Should extract nested JSON with mixed multibyte");
         assert_eq!(extracted, input);
     }
+
+    #[test]
+    fn test_extract_json_object_with_multibyte_string() {
```
KuSh, line 354: Measured note on what this test does and does not add — not a request to remove it.

I reverted `extract_json_object` to its pre-fix char-index form and ran the file:

```
cjk_values ............. FAILED     emoji_values ........... FAILED
mixed_multibyte_nested . FAILED     with_multibyte_string .. FAILED  <- this one
cjk_prefix ............. ok
```

So this test detects the same bug three existing tests already detect. I also went looking for a mutant that only this test catches and did not find one.

What it does add: it is the only marker-path test with multibyte *inside* the scanned region, trailing content after the `}`, and an exact-equality assertion. That is a mild strengthening at zero cost, so keeping it is fine.

The high-value version of this test would target the fallback path (no `"numTotalTests"` marker) with CRLF line endings, where `extract_json_object` still panics today — see #3903 and point 3 of the review.

KuSh (changes requested):
Reviewed at `64912f71`, in a worktree merged onto current `develop` (`5e10aa6`). Mechanically clean: merges without conflict, and on the merged tree `cargo fmt --all -- --check`, `cargo clippy --all-targets` (zero warnings) and `cargo test --all` (3108 + 102 passing, 0 failed) are all green.

Requesting changes on the PR metadata, plus one ask that would make this PR genuinely valuable. The test itself is correct and I am not asking you to remove it.

---

## 1. Title and description no longer describe the change (blocking)

After the rebase the net diff against `develop` is 12 lines, test-only. The production change this PR is named after — byte offsets in `extract_json_object` — is already on `develop` as 27f9739b, which also carries `Fixes #2509`; that issue closed on 2026-07-01.

To be clear about the history: your commit's author date is 2026-06-21, so you filed this **before** 27f9739b landed. You were not duplicating anything, and resolving the rebase by dropping the now-redundant production hunk and keeping only your extra test was the right call. The problem is purely that the metadata did not follow.

As it stands, merging would put a commit in `develop` claiming a UTF-8 boundary fix that is not in its diff, and re-trigger closing automation on an already-closed issue. Please:

- retitle to something like `test(parser): cover multibyte JSON extraction on the marker path`
- change the summary bullet `fix JSON extraction to compute slice boundaries using UTF-8 byte offsets` — that hunk is no longer here
- change `Fixes #2509` to `Follow-up coverage for #2509`

## 2. The `#2509` panic class is still live on `develop` — filed as #3903

Not caused by this PR, and not something I expect you to fix here, but it is directly relevant to what this PR certifies.

27f9739b fixed the forward brace-balancing scan, which closed the marker path. The **fallback path** (`src/parser/mod.rs:161`) still derives `start_pos` with `.map(|l| l.len() + 1)`, which assumes `\n`. `str::lines()` also strips `\r`, so on CRLF input the offset is one byte short per preceding line and the slice can land mid-character. End to end, with a `vitest` that dies before emitting its JSON summary:

```
thread 'main' panicked at src/parser/mod.rs:175:35:
start byte index 66 is not a char boundary; it is inside 'é' (bytes 65..67 of string)
exit=101
```

There is no `catch_unwind` in `VitestParser::parse`, so this aborts rtk rather than degrading to Tier 2/3 — on Windows, on the error path. Full details and repro in #3903.

## 3. The valuable ask: add a fallback + CRLF test

Your new test lands on the marker path, which is the branch that is already fixed and already covered. The branch nothing covers is the fallback path, and it is the one that still panics.

A test there would catch a live bug rather than restate a closed one. If you would like to take #3903 as well, that would be very welcome — otherwise just the test is a real contribution on its own.

## 4. Please keep `test_extract_json_object_cjk_prefix`

Flagging this pre-emptively because it looks redundant with your new test and it is not. I mutated the marker path's `start_pos` into a char index used as a byte offset:

```
cjk_prefix ............. FAILED   ← start byte index 22 is not a char boundary, inside '경'
with_multibyte_string .. ok
cjk_values / emoji_values / mixed_multibyte_nested / pnpm / dotenv ... ok
```

`cjk_prefix` is the only test in the file that catches it. Its multibyte sits *before* `start_pos`, so it guards `rfind`'s offset derivation; yours sits *inside* the scanned slice, so it guards the `end_pos` arithmetic. Different halves of the same function — both worth having.

## 5. Latent, unverified — marker path takes the nearest preceding brace

`input[..pos].rfind('{')` walks back to the nearest `{`, not the enclosing one:

```rust
extract_json_object(r#"{"config": {"root": "/x"}, "numTotalTests": 5, "ok": true}"#)
// => Some("{\"root\": \"/x\"}")
```

which then fails `serde_json::from_str::<VitestJsonOutput>` and drops the caller to Tier 2/3 on perfectly valid JSON. I reproduced the function behaviour but could **not** confirm any real vitest/jest reporter emits an object before `numTotalTests` — there is no vitest fixture in the repo to check against — so I am recording this as latent, not as a defect, and explicitly not asking you to act on it in this PR.

KuSh:
Closing in favour of #3991, which fixes the same JSON extraction in src/parser/mod.rs and covers multibyte input across LF, CRLF and mixed line endings, including the numTotalTests marker case this test checks. The byte offset fix described here already landed separately, so only the test remains, and that case is part of the matrix in #3991. Please take a look there, or reply here if your scope was different.

PR #2535, opened 2026-06-21
fix(hook): accept current Claude tool input keys

### LINKED ISSUE #2493: hook claude: JSON key mismatch — tool/input vs tool_name/tool_input
## Problem

`rtk hook claude` parses `tool_name` / `tool_input` from stdin, but Claude Code now sends `tool` / `input`.

Result: hook exits 0 with no output → no rewrite → commands (`grep`, `head`, `cat`) bypass RTK.

## Reproduction

```bash
# ❌ What Claude Code sends (no rewrite)
echo '{"tool":"Bash","input":{"command":"grep -r hello ."}}' | rtk hook claude
# exit 0, empty output

# ✅ What RTK expects (rewrite works)
echo '{"tool_name":"Bash","tool_input":{"command":"grep -r hello ."}}' | rtk hook claude
# {"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk grep -r hello ."}}}
```

## Version

- rtk 0.42.4 (Homebrew)
- Claude Code — latest (June 2026)

## Suggested fix

Accept both key formats in the JSON parser — fall back from `tool_name` to `tool` (and `tool_input` to `input`).

## Workaround

Wrapper script that remaps keys before piping to `rtk hook claude`:

```bash
#!/bin/bash
cat | python3 -c "
import sys, json
d = json.load(sys.stdin)
if 'tool_name' not in d and 'tool' in d:
    d['tool_name'] = d.pop('tool')
if 'tool_input' not in d and 'input' in d:
    d['tool_input'] = d.pop('input')
print(json.dumps(d))
" | rtk hook claude
```

YOMXXX:
## Summary
- accept both Claude hook payload formats: `tool_input` and current `input`
- preserve the original input object fields when emitting `updatedInput`
- add a regression test for the current `tool/input` payload shape

Fixes #2493

## Tests
- cargo +1.93.0 test hooks::hook_cmd::tests -- --nocapture
- cargo +1.93.0 test --all

## Manual verification
- `printf '%s\n' '{"tool":"Bash","input":{"command":"grep -r hello .","timeout":30000}}' | target/debug/rtk hook claude` emitted `updatedInput.command = "rtk grep -r hello ."`
- legacy `tool_name/tool_input` payload still emits the same rewrite

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -339,13 +339,25 @@ enum PayloadAction {
     Ignore,
 }
 
+fn claude_payload_input(v: &Value) -> Option<(&Value, &str)> {
+    for key in ["tool_input", "input"] {
+        if let Some(input) = v.get(key) {
+            if let Some(cmd) = input
+                .get("command")
+                .and_then(|c| c.as_str())
+                .filter(|c| !c.is_empty())
+            {
+                return Some((input, cmd));
+            }
+        }
+    }
+
+    None
+}
+
 fn process_claude_payload(v: &Value) -> PayloadAction {
-    let cmd = match v
-        .pointer("/tool_input/command")
-        .and_then(|c| c.as_str())
-        .filter(|c| !c.is_empty())
-    {
-        Some(c) => c,
+    let (input, cmd) = match claude_payload_input(v) {
+        Some((input, cmd)) => (input, cmd),
         None => return PayloadAction::Ignore,
     };
 
@@ -367,7 +379,7 @@ fn process_claude_payload(v: &Value) -> PayloadAction {
     };
 
     let updated_input = {
-        let mut ti = v.get("tool_input").cloned().unwrap_or_else(|| json!({}));
+        let mut ti = input.clone();
         if let Some(obj) = ti.as_object_mut() {
             obj.insert("command".into(), Value::String(rewritten.clone()));
         }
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for taking this on, @YOMXXX — clean patch, tests included, full CI green. Unfortunately I'm closing it, because the underlying report (#2493) doesn't hold up.

I verified the `PreToolUse` payload shape against the Claude Code build published on **2026-06-18, the day #2493 was filed** (2.1.182), and against current (2.1.251). Both construct it in exactly one place, identically:

```
{...base, hook_event_name:"PreToolUse", tool_name:e, tool_input:n, tool_use_id:t}
```

I also captured a real payload from a live session and piped it through unmodified rtk 0.46.0 — the `grep -r hello .` case from the issue rewrites correctly today, with no patch. Full evidence in #2493.

So the `input` fallback this PR adds is unreachable from any Claude Code hook event. Beyond being dead code, three things make me not want to merge it anyway:

- **It widens the trigger surface.** `process_claude_payload` deliberately doesn't inspect `tool_name` — it relies on the `"matcher": "Bash"` registration to scope itself. Accepting a second input key means more payload shapes can reach a rewrite.
- **It doesn't fully implement what #2493 asked.** It maps `tool_input` → `input` but leaves `tool` → `tool_name` alone. Harmless right now only because `run_claude` never reads `tool_name` — but it becomes a latent trap the moment someone adds a guard there.
- **It's a schema variant without evidence.** `hook_cmd.rs` carries several key namings on purpose (`tool_name`/`tool_input`, `toolName`/`toolArgs`), and every one is documented with a live-verified capture and a PR link. I'd rather keep that bar than add a branch we can't point at a real payload for.

One line here is genuinely worth keeping: `let mut ti = input.clone()` in place of `v.get("tool_input").cloned().unwrap_or_else(|| json!({}))`, now that the lookup upstream already proves existence. Happy to take that as a standalone cleanup if you'd like to open one.

If you're interested in the bug the reporter actually hit, **#3557** is the live one — rtk's `updatedInput` losing out to sibling `PreToolUse` hooks on the same matcher. That's a harder and more valuable fix, and it's unclaimed.

PR #2536, opened 2026-06-21
fix(json): tolerate raw control chars in strings instead of dropping output

dosthcpp:
## Summary

- `rtk json` (and pipes like `curl … | rtk json`) used strict `serde_json::from_str`, which rejects unescaped control characters (U+0000–U+001F) inside string values per RFC 8259 §7. Some real-world producers emit them anyway (e.g. an API echoing a user-supplied newline into a field). rtk then printed **nothing** and exited 1, **losing the whole payload** and forcing a raw re-fetch — net-negative for a token-saving tool and contrary to the **Never Block** principle.
- Adds `parse_json_lenient`: strict parse first (fast path, **zero behavior change** for valid JSON), and only on failure retry once with raw in-string control chars rewritten to their equivalent `\uXXXX` escapes via a string-aware scanner. Whitespace between tokens and already-escaped sequences stay byte-for-byte identical; genuinely malformed input still surfaces the original strict error.

### Before / After
```
$ printf '{"body":"line1\nline2"}' | rtk json -      # raw newline inside the string
# before:  rtk: Failed to parse JSON: control character (...)   (exit 1, 0 bytes stdout)
# after:   { body: "line1
#           line2" }                                            (exit 0)
```

## Test plan

- [x] `cargo test json_cmd` — **19 passed, 0 failed** (9 new + 10 existing). New tests cover recovery in compact and schema modes, control char in keys, fast-path borrow, whitespace preservation, existing-escape preservation, and malformed-still-errors.
- [x] `cargo fmt -- --check src/cmds/system/json_cmd.rs` — clean
- [x] `cargo clippy --lib` — no new warnings on the changed file
- [x] Manual: `printf '{"id":"abc","body":"a\nb"}' | rtk json -` now renders the payload (exit 0); `printf '{not json' | rtk json -` still errors (exit 1); valid JSON output unchanged.

### DIFF
--- a/src/cmds/system/json_cmd.rs
+++ b/src/cmds/system/json_cmd.rs
@@ -4,6 +4,7 @@ use crate::core::guard::never_worse;
 use crate::core::tracking;
 use anyhow::{bail, Context, Result};
 use serde_json::Value;
+use std::borrow::Cow;
 use std::fs;
 use std::io::{self, Read};
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


dosthcpp:
Friendly ping from the author — this PR is ready for review, but **CI has never run on it**, so there's nothing for a reviewer to look at yet.

The check-suite on `ff304db` sits at `conclusion: action_required`: as a first-time contributor, my workflow runs need a maintainer to click **"Approve and run workflows"** before GitHub will dispatch them. Nothing has failed — the jobs simply haven't been allowed to start.

Current state, so the approval is a cheap call to make:

- **Rebased onto latest `develop`** — the branch had drifted 158 commits behind, which is on me; it's now a single clean commit on top of `develop`.
- `mergeable: true`, no conflicts.
- **CLA signed** (confirmed by CLA assistant above).
- Locally green: `cargo test json` → **130 passed / 0 failed**, plus `cargo fmt --check` and `clippy` clean.

If the change itself isn't something you want, I'm equally happy to hear that and close it — no hard feelings. But right now it's stuck in a state where neither of us can see a CI signal. Thanks for taking a look.

PR #2544, opened 2026-06-22
grep: detect format flags bundled inside short-flag clusters

### LINKED ISSUE #2543: grep: -r / -rn rejected on 0.42.4 (and -rln returns a false "0 matches" on latest develop)
Using rtk 0.42.4 from Homebrew. `rtk grep` is one of the commands I use the most, but a couple of really common invocation shapes don't behave the way I'd expect.

**`-r` (and `-rn`) get rejected outright**

```
$ rtk grep -r "TODO" src
error: unexpected argument '-r' found

  tip: to pass '-r' as a value, use '-- -r'

```

`-rn` fails the same way. Since `rtk grep` is always recursive anyway, I'd expect `-r`/`--recursive` to just be accepted and ignored, the same way `-n`/`--line-numbers` already is (there's even a comment in the code to that effect). Instead it falls all the way back to running plain grep completely unfiltered — which kind of defeats the point of using rtk in the first place, since `-r`/`-rn` is probably the single most common way people actually type a recursive grep.

I checked whether this was already fixed in `develop` before reporting — built it from source, and `-r`/`-rn` do work there now (looks like it happened as a side effect of restructuring the arg parsing, not sure if intentional). So this might just need backporting/releasing rather than a new fix.

**`-rln` on develop: silent false "0 matches"**

While checking the above, I found `-rln` still misbehaves on develop, just differently:

```
$ rtk grep -rln "TODO" src
0 matches in 0 files
```

There are definitely matches there (plain grep confirms it). Dug a bit further — looks like the combined short flags get merged into one token (e.g. `-rln` → `-ln` after the recursive flag is stripped), and the check for "format" flags like `-l` (files-with-matches) only looks for it as a standalone arg, not bundled inside a cluster like that. So rtk doesn't realize `-l` changed rg's output shape entirely, tries to parse it as normal match lines, and silently gets nothing — which looks exactly like "no matches found" instead of "something's off here."

I have a small fix for the second one and will open a PR. Mentioning both here since they're related (same command, same general area) and the first one might just need a release rather than code changes.

EtaCassiopeia:
Follow-up to #2543 — this fixes the second part of it.

`has_format_flag` checks `extra_args` for things like `-l`/`-c`/`-o` so it knows to skip the usual match-line parsing and just pass rg's output straight through (since those flags change rg's output format entirely). That check only matches exact tokens though, and `extract_pattern_path` merges plain boolean short flags into one combined token before they get here — so `-rln` becomes `-ln` after `r` is stripped, and `-ln` never matches `-l` exactly.

End result: something like `rtk grep -rln pattern path` runs rg with `-l` correctly, but then tries to parse the (now filenames-only) output as if it were normal `file:line:content` matches, finds nothing that fits that shape, and reports "0 matches" even when there clearly are matches. That's worse than just erroring, since it looks identical to a real empty result.

Fix is small — `has_format_flag` now also scans the individual letters of a short cluster (long flags and exact single-flag matches are unaffected). Added a few tests for the clustered case plus one to make sure long flags like `--ignore-case` don't get walked character-by-character by mistake.

Tested locally:
- `rtk grep -rln pattern path` now correctly falls through to passthrough and shows the real file list instead of "0 matches in 0 files"
- `rtk grep -rn pattern path` unaffected
- existing grep_cmd test suite passes (72 tests)

closes #2543

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -500,8 +500,9 @@ fn parse_match_line(line: &str) -> Option<(String, usize, &str)> {
 
 fn has_format_flag<T: AsRef<str>>(extra_args: &[T]) -> bool {
     extra_args.iter().any(|arg| {
-        matches!(
-            arg.as_ref(),
+        let s = arg.as_ref();
+        if matches!(
+            s,
             "-c" | "--count"
                 | "--count-matches"
                 | "-l"
@@ -515,7 +516,20 @@ fn has_format_flag<T: AsRef<str>>(extra_args: &[T]) -> bool {
                 | "--json"
                 | "--passthru"
                 | "--files"
-        )
+        ) {
+            return true;
+        }
+        // extract_pattern_path merges boolean short flags into one cluster
+        // token (e.g. `-rln` becomes `-ln` after stripping `r`), so a format
+        // flag like `-l` can arrive bundled with others rather than as its
+        // own arg. Clusters only ever contain boolean letters (see
+        // parse_cluster), so scanning characters here is safe.
+        if let Some(rest) = s.strip_prefix('-') {
+            if !rest.is_empty() && !rest.starts_with('-') {
+                return rest.chars().any(|c| matches!(c, 'c' | 'l' | 'L' | 'o' | 'Z'));
+            }
+        }
+        false
     })
 }
 
@@ -1117,6 +1131,24 @@ mod tests {
     #[test]
     fn test_format_flag_ignores_normal_flags() {
         assert!(!has_format_flag(&["-i", "-w", "-A", "3"]));
+    }
+
+    #[test]
+    fn test_format_flag_detects_clustered_l() {
+        // `-rln` becomes `-ln` after extract_pattern_path strips the `r`.
+        // -l (files-with-matches) is bundled with -n here, not standalone.
+        assert!(has_format_flag(&["-ln"]));
+    }
+
+    #[test]
+    fn test_format_flag_detects_clustered_c() {
+        assert!(has_format_flag(&["-ic"]));
+    }
+
+    #[test]
+    fn test_format_flag_ignores_double_dash() {
+        // Long flags must not be mistaken for a short cluster.
+        assert!(!has_format_flag(&["--ignore-case"]));
     }
 
     // Verify line numbers are always enabled in rg invocation (grep_cmd.rs:24).


routesmith:
Confirmed on **WSL2 (Ubuntu 24.04) / Linux x86_64, rtk 0.42.4**, invoked via the Claude Code auto-rewrite hook — a third platform alongside the macOS report in #2647 and the Linux/0.42.1 report in #2563.

Traced independently to exactly the cause described here: `extract_pattern_path` → `parse_cluster` collapses `-rli` to `-li` (strips `r`), and `has_format_flag` (`grep_cmd.rs`) only exact-matches `-l`, so the `-l` inside the cluster is invisible. Passthrough is skipped, but `-li` is still forwarded to rg → rg emits filenames-only output → every line fails `parse_match_line` (`file\0line:content` shape) and is silently dropped → `0 matches in 0 files:` on content that plainly exists.

Minimal repro:

```
grep -rli "foo\|bar" .        # -> "0 matches in 0 files:" (false)
grep -r -l -i "foo\|bar" .    # -> correct (l visible as its own token -> passthrough)
rtk proxy grep -rli "foo\|bar" .   # -> correct (bypasses filter)
```

Real-world impact matched #2563 almost exactly: the false zero read as an LLM hallucination and nearly triggered a destructive "correction" (concluding correct, existing content did not exist). The `by_file.is_empty()` + non-empty stdout case really should fall back to raw output per RTK's own documented fallback rule, independent of the cluster-detection fix here. This PR fixes the trigger; a fallback guard would catch the whole class. Thanks for the fix — would like to see it land.

routesmith:
Follow-up: this is already fixed on `develop` / in stable **v0.43.0** via the `grep_cmd.rs` → `src/cmds/system/search.rs` refactor (#2641), which lands both a cluster-aware `has_format_flag` (short flags scanned per-letter, so `-rli` → `-l` routes to passthrough) and an `unparsed_signal()` guard that re-runs raw when stdout is non-empty but nothing parses. Verified the repro returns correct results on v0.43.0. Since this PR targets the now-removed `grep_cmd.rs`, it looks superseded and can probably be closed in favor of the search refactor.

KuSh:
Thanks for digging into the bundled-cluster case and writing the repro. This landed by way of #2550 (merged 2026-06-23), which rewrote grep into src/cmds/system/search.rs; its has_format_flag scans every letter of a short cluster for c/l/L/o/q/b/Z/z and matches long flags against an explicit list, so -ln and -rln now route to passthrough exactly as you described. Develop also carries unit tests for -rl/-rc/-rq/-rln/-cr and an end-to-end -rln test citing #2543, so the coverage your patch added is present too. Running the repro on develop lists the matching files instead of reporting 0 matches. Closing as covered, but please comment or reopen if you see a flag shape that still slips through.

PR #2550, opened 2026-06-23
fix(grep): correctly handle all flag shapes and never exceed raw output

### LINKED ISSUE #2543: grep: -r / -rn rejected on 0.42.4 (and -rln returns a false "0 matches" on latest develop)

aeppling:
## Summary

Make rtk grep produce correct output for every valid grep invocation and never cost more tokens than plain grep.

The #2333 arg-parsing rework routed many flag shapes through the grouping reparse, which silently dropped unparseable lines and returned success, producing wrong/empty results (and grep failed on --include) that pushed agents to unfiltered rtk proxy grep.
 
- Context (-A/-B/-C): grouped + compressed instead of dropped; header counts matches only.
- Unrepresentable output (-N, -I, --color, -p, --heading, --field-match-separator, binary notices): passed through verbatim, never a false "0 matches".
- Format/shape flags (-c/-l/-o/-Z/--column/-b/--vimgrep/--null-data): clean passthrough, no NUL leak.
- grep-only syntax (--include/--exclude/…): ripgrep fast path, falls back to system grep on rg error.
- Never-worse guard: never emits more than plain file:line:content grep.
  
 ## Related
 
- Resolves #2543 (-r/-rn rejected on 0.42.4; -rln false "0 matches" on develop).
- Supersedes #2224 (rg-only flags leaking to the grep fallback) — folded into strip_rg_only + the rg-error fallback.

## Test plan

- [x] cargo fmt --all && cargo clippy --all-targets && cargo test --all — clean, 2242 tests pass.
- [x] New integration tests: context compression, safety-net passthrough, NUL-free format/shape, grep-only-flag fallback, token savings, never-worse guard.
- [x] Correctness battery vs bare grep across 18 shapes — no NUL, no false 0-matches, no dropped matches, never bigger than raw.
- [x] Real hook path verified in tracking DB — routed (not proxy), 52% on bulky, 0% (never negative) on small.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -1,8 +1,8 @@
 //! Filters grep output by grouping matches by file.
 
-use crate::core::stream::exec_capture;
+use crate::core::stream::{exec_capture, CaptureResult};
 use crate::core::tracking;
-use crate::core::utils::resolved_command;
+use crate::core::utils::{resolved_command, strip_ansi};
 use crate::core::{args_utils, config};
 use anyhow::{Context, Result};
 use regex::Regex;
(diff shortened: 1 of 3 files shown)


alexisLefebvre:
Nice — the `unparsed_signal` safety net looks like it'd catch this case by falling back to passthrough. One small thing I noticed: I couldn't spot a regression test for the exact #2543 shape, a bundled files-with-matches cluster (`-rln`/`-ln`). The new tests all use standalone flags (`--column`, `-c`, `-o`, `-N`, `-I`, …), so `-l`-inside-a-cluster isn't directly exercised. Might be worth a small test asserting `rtk grep -rln <pat> <dir>` lists the files rather than `0 matches`, just to pin the original report — but no strong opinion if it's already covered elsewhere.

Repro + a 0.42.2 confirmation, in case useful: https://github.com/rtk-ai/rtk/issues/2543#issuecomment-4779235608

aeppling:
Thanks @alexisLefebvre , good catch. Added a regression test for the exact #2543 shape in 3a59d23: `bundled_files_with_matches_cluster_lists_files` asserts `rtk grep -rln <pat> <dir>` lists every matching file (and excludes non-matching ones) instead of a false `0 matches`.

pszymkowiak (approved):

EtaCassiopeia:
@aeppling Hey, just wondering if there was a reason my [PR](https://github.com/rtk-ai/rtk/pull/2544) wasn't merged and the work was redone instead. Does the project accept external contributions?

PR #2554, opened 2026-06-23
fix(core): never-worse output guard

### LINKED ISSUE #2551: Generalize never-worse guard to all filters (curl emits more tokens than raw on tiny responses)
## Problem

The benchmark CI gate (`BENCHMARK FAILED: 1 filter(s) produced more tokens than raw output`) flags the `curl` filter as negative:

```
🔴 curl text │ curl -s https://httpbin.org/robots.txt │ rtk curl ... │ 8 → 40 (-400%)
```

`rtk curl` wraps the response with status/header metadata. On a tiny payload (httpbin's `robots.txt` is ~8 tokens) that wrapper dominates, so the filtered output is **larger** than the raw command output.

## Root cause

The never-worse guard added for `grep` in #2550 (fall back to the raw form when filtering doesn't shrink the output) is **grep-only**. Other filters — `curl` here — have no such cap, so they can emit more tokens than the underlying command on small inputs.

## Proposal

Extract the never-worse comparison into a shared `core` helper and apply it across filters (start with `curl`, then audit the rest), so **no filter ever emits more tokens than the raw command**. This makes the benchmark gate hold structurally rather than per-filter.

## Evidence

- Benchmark run: https://github.com/rtk-ai/rtk/actions/runs/28024206815/job/82948000842
- Failing line: `🔴 curl text   8 → 40 (-400%)`
- **Not** introduced by #2550 (which touches only `grep`); the same run shows every grep shape positive: `fn +94%`, `-l +99%`, `-c +95%`.

aeppling:
## Summary

Never-worse guard so RTK never emits more tokens than the command the agent ran. `core::guard::never_worse(raw, filtered)` returns `raw` when `filtered` is larger — and is applied only where `filtered` is a structural compression of a **faithfully captured** `raw` (the agent's exact command + args, nothing added, no extra process).

`never_worse` stays **pure** (empty `raw` → empty). Guard concerns that involve *execution* (a baseline run failing) are fixed in the execution layer, not by weakening the guard.

## What's guarded (faithful `raw`, compress-only)

- **Shared sink**: `runner::emit_guarded` / `print_with_hint` compose body + tee hint and guard the whole; `run_captured_filter` covers the `run_filtered*` family.
- **TOML filters**  one central guard site (project / `.rtk` / global / built-in).
- **read**  `raw` made faithful to `cat` / `cat -n` (line numbers are faithful to the command, so they're compared like-for-like, not dropped).
- **pipe**  stdin filter mode now guarded.
- **aws**  JSON subcommands already guarded; the two **s3 text** paths (`s3 ls`, `s3 sync/cp`) now route through `emit_guarded` too (#2551 class).
- **docker** (`ps` / `ps -a` / `images`) — **the agent's command is now authoritative**: it ran a plain command for `raw` and a separate `--format` command for parsing, capturing the plain run with `unwrap_or_default()`; a failed plain run while `--format` succeeded left `raw` empty and `never_worse("", summary)` dropped real output. Now the real command is the source of truth (faithful failure on error), `--format` is a best-effort parse aid that falls back to `raw`, and the guard always sees a real baseline. Parsers unchanged.
- **grep** — rebased onto develop's grep rewrite (#2550), re-applying our no-result rule: a genuine no-match emits **empty** (like real `grep`), never a `0 matches` line that exceeds empty `raw`.

## Deferred — separate "adds-info" follow-up (out of scope: these emit data the command never produced / run an extra process)

- **git add** — `git add` is silent; RTK runs an extra `git diff --cached --stat` to synthesize an `ok N files…` summary (guarding it would suppress the summary).
- **dotnet** — injects `-bl`/`--logger` and reads a binlog the agent's command never produced.
- **git status / git show** — same authoritative-command restructure as docker (two-run today); higher risk because porcelain drops the rebase/merge state header, so done carefully and separately.

## Documented exceptions (unguardable by design)

- **streaming** (`tsc`, `gradlew`, `cargo`) — no full buffer exists mid-stream; safe by construction (subset emit + bounded flush).
- **proxy / passthrough** — `output == raw` by definition.

## Related
- Resolves #2551.
- Generalizes the grep-specific guard from #2550 (now merged from develop).

## Test plan
- [x] `fmt` + `clippy` + `cargo test --all` — green (2259 passed).
- [x] guard unit + integration tests; empty-raw / no-match regressions.
- [x] docker authoritative-command verified live (`ps` / `ps -a` / `images`); aws s3 + pipe + read + env covered.
- [x] develop merged; grep conflict resolved (develop's rewrite + our no-match→empty).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,6 +46,8 @@ The LLM doesn't know RTK is involved for which commands, hooks rewrite commands 
 
 Don't invent new output formats. Don't add RTK-specific headers or markers in the default output. The filtered output should be indistinguishable from "a shorter version of the real command."
 
+Enforce it with `guard::never_worse(raw, filtered)` — print and track the value it returns (use `runner::emit_guarded(filtered, hint, raw)` when appending a tee hint). It guarantees RTK never emits more tokens than the raw command, down to emitting nothing when the command produced nothing.
+
 ### Never Block
 
 If a filter fails, fall back to raw output. RTK should never prevent a command from executing or producing output. Better to pass through unfiltered than to error out. Same for hooks: exit 0 on all error paths so the agent's command runs unmodified.
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -248,7 +248,6 @@ bench "deps" "cat Cargo.toml" "$RTK deps"
 section "env"
 bench "env" "env" "$RTK env"
 bench "env -f PATH" "env | grep PATH" "$RTK env -f PATH"
-bench "env --show-all" "env" "$RTK env --show-all"
 
 # ===================
 # err
--- a/src/cmds/cloud/aws_cmd.rs
+++ b/src/cmds/cloud/aws_cmd.rs
@@ -407,17 +407,14 @@ fn run_s3_ls(extra_args: &[String], verbose: u8) -> Result<i32> {
     }
 
     let result = filter_s3_ls(&stdout);
-    if result.truncated {
-        if let Some(hint) = crate::core::tee::force_tee_hint(&raw, "aws_s3_ls") {
-            println!("{}\n{}", result.text, hint);
-        } else {
-            println!("{}", result.text);
-        }
+    let hint = if result.truncated {
+        crate::core::tee::force_tee_hint(&raw, "aws_s3_ls")
     } else {
-        println!("{}", result.text);
-    }
+        None
+    };
+    let shown = crate::core::runner::emit_guarded(&result.text, hint.as_deref(), &raw);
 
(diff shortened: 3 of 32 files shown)


pszymkowiak (approved):
Approving. The core `never_worse` guard (`guard.rs` + runner integration) is clean and well-tested, and the docker authoritative-command restructure is a solid fix for the empty-raw drop (#2551). Tracking now records the actually-emitted output — the right call.

Non-blocking nits for a possible follow-up: wget's `ok | size` / `FAILED` summaries are "adds-info" (size comes from a separate stat), so guarding them can suppress the summary when raw is tiny (e.g. `wget -q`) — could be treated like the deferred git add/dotnet cases. `oc` support and the grep-fallback rewrite are welcome but somewhat orthogonal to the guard.

PR #536, opened 2026-03-12
feat: binary hook engine with lexer, streaming, and rewrite_compound integration

ahundt:
## Summary

This [PR #536](https://github.com/rtk-ai/rtk/pull/536) integrates the hooks-v2 engine (from PR #156) with the `develop` branch (note that #537 is the same code as 536 and 156 but it is stripped of tests for easier reviewing), and https://github.com/rtk-ai/rtk/issues/361 describes the bug fixes in this pr, this pr #536 demonstrates that:

1. The **lexer is a drop-in replacement** for `rewrite_compound()`'s inline byte scanner — all 155 existing rewrite tests pass unchanged
2. The **binary hook path** (`rtk hook claude`) works alongside the shell hook path (`rtk rewrite`)
3. The **streaming infrastructure** (`stream.rs`) enables bidirectional process shims with exit code transparency

### Key components

- **`src/cmd/lexer.rs`** — Shell-aware tokenizer (56 tests): handles quoting, escapes, `$()`, backticks, globs, redirects, `2>&1`
- **`src/cmd/analysis.rs`** — Command chain parser (61+ tests): `needs_shell()` detection, `parse_chain()` for `&&`/`||`/`;` chains
- **`src/cmd/hook/mod.rs`** — Binary hook engine (98 tests): conservative `hook_lookup()` whitelist, Claude Code JSON protocol, safe suffix handling
- **`src/cmd/exec.rs`** — Native command executor: runs simple chains natively, delegates complex shell to `/bin/sh`
- **`src/stream.rs`** — Streaming process execution (39 tests): `StreamFilter` trait, `StdinMode::Inherit` for pipe support, SIGPIPE handling, 1 MiB raw cap
- **`src/pipe_cmd.rs`** — `rtk pipe --filter` dispatch for grep/rg/find/fd output filtering
- **`src/discover/registry.rs`** — `rewrite_compound()` replaced: lexer-based splitting instead of 130-line inline byte scanner

### Lexer integration in `rewrite_compound()`

The byte-level quote tracker in `rewrite_compound()` (~130 lines) is replaced with `lexer::tokenize()` calls (~80 lines). The lexer adds:
- Escape sequence handling (`\\`)
- `$()` and backtick detection (Shellism tokens)
- Glob detection (`*`, `?` → Shellism)
- Redirect parsing — not confused with `&&` operator (`2>&1`, `&>/dev/null`)

[...]

PR #2555, opened 2026-06-24
Preserve pipeline input while rewriting safe shell commands

### LINKED ISSUE #1560: bug(rewrite): hook rewrites left-hand side of piped commands, breaking downstream filters

ahundt:
# Preserve pipeline input while rewriting safe shell commands

## Summary

RTK now lexes shell command structure before rewriting it. Filters and other content consumers receive the producer's original output, while display-only tails can still use RTK's compact output. Redirections and shell-block boundaries remain attached to the command that owns them.

This fixes [issue #1560](https://github.com/rtk-ai/rtk/issues/1560), where rewriting the left side of a pipe changed the input seen by downstream filters. The lexer consolidates the boundary and suffix behavior developed in [PR #536](https://github.com/rtk-ai/rtk/pull/536) and supports long, multi-line shell blocks without treating their internal operators as independent top-level commands.

## Before and after

| Command | Before | After |
|---|---|---|
| `git log \| grep feat` | RTK could replace the producer before `grep` read it. | The command stays raw, so `grep` receives the original Git stream. |
| `cargo test \| tee /tmp/test.log` | The file could contain RTK's compact display instead of the test stream. | The command stays raw, so `tee` saves the original output. |
| `git log \| head -20` | Conservative pipe handling could leave a safe display tail unfiltered. | RTK can rewrite the producer because `head` only consumes a bounded display prefix. |
| `git status > /tmp/status.log` | The file-target redirect prevented the supported rewrite path. | RTK preserves the suffix and returns an ask-only rewrite: `rtk git status > /tmp/status.log`. |
| `git status $(whoami)` | Shell substitution could be mistaken for a safe rewrite boundary. | The command stays raw because its execution boundary is opaque. |
| `if git status; then git log | grep feat; fi` | A text-only split could analyze the nested pipe as an independent top-level command. | RTK treats the block as one conservative shell unit and does not rewrite through the opaque consumer. |

## Implementation

`src/discover/lexer.rs` tokenizes quotes, escapes, comments, CRLF/newline boundaries, pipes, redirects, case alternatives, and shell blocks using typed tokens and byte offsets. `src/discover/registry.rs` applies consumer-aware producer routing. `src/discover/suffix.rs` preserves the exact trailing output suffix and assigns an ask-only policy to file-target redirects.

The policy path is compatible with the multiline rewrite work in [PR #2781](https://github.com/rtk-ai/rtk/pull/2781): each statement is analyzed independently, and auto-allow requires both an attestable command and a rewrite that does not require confirmation. This also preserves the least-privilege behavior tracked in [issue #1155](https://github.com/rtk-ai/rtk/issues/1155).

Content consumers including `grep`, `rg`, `tee`, `xargs`, `wc`, `awk`, and `sed` remain raw. Safe display tails are limited to supported stdin-only numeric `head` and `tail` forms, plain `cat`, and equivalent absolute paths. Input redirects, substitutions, heredocs, and opaque grouped commands remain conservative.

## Compatibility and scope

The existing string rewrite API remains available. Hook adapters use the policy-returning path when a file redirect requires confirmation. This change adds no parser dependency and does not execute native commands through the lexer.

## Verification

- 109 lexer tests cover quoting, escaping, comments, CRLF, pipes, redirects, suffixes, shell blocks, content consumers, safe tails, and conservative fallbacks.
- The full branch suite passes with 2,570 tests passed and 8 ignored.
- `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`, and `git diff --check` pass.

### DIFF
--- a/src/discover/README.md
+++ b/src/discover/README.md
@@ -21,11 +21,11 @@ When a hook sends `cargo fmt --all && cargo test 2>&1 | tail -20`:
 → [Arg("cargo"), Arg("test"), Redirect("2>&1"), Operator("&&"), Arg("git"), Arg("status")]
 ```
 
-**Compound splitting** — The rewrite engine walks the tokens, splitting on `Operator` (`&&`, `||`, `;`) and typed `Pipe` tokens (`|`, `|&`). For normal pipelines, producers and intermediate stages stay raw, and only an argument-safe final stage marked `pipeline_final_safe` is rewritten. The initial safe set is ordinary `grep` and `rg` invocations; search pattern-file forms (`-f`/`--file`) defer because they can consume pipeline stdin as configuration. Stderr pipelines (`|&`) and pipelines containing opaque shell groups remain raw.
+**Compound splitting** — The rewrite engine walks the tokens, splitting on `Operator` (`&&`, `||`, `;`) and typed `Pipe` tokens (`|`, `|&`). A content-sensitive consumer such as `grep`, `rg`, or `tee` keeps the producer's raw output, while a stdin-only display tail such as `head -5`, `tail -20`, or `cat` permits the producer to be routed through RTK. Stderr pipelines (`|&`) and pipelines containing opaque shell groups remain raw.
 
 **Per-segment rewriting** — Each segment goes through:
 
-1. Strip trailing redirects (`2>&1`, `>/dev/null`) — matched via lexer tokens, set aside, re-appended after rewriting
+1. Strip safe trailing output routing (`2>&1`, `>/dev/null`) — matched via lexer tokens, set aside, and re-appended after rewriting; file targets remain in the core so the permission policy can require an ask
 2. Short-circuit special cases — `head -20 file` → `rtk read file --max-lines 20`, `tail -n 5 file` → `rtk read file --tail-lines 5`. These can't go through generic prefix replacement because it would produce `rtk read -20 file` (wrong flag position)
 3. Classify the command — strip env prefixes (`sudo`, `FOO="bar baz"`), normalize paths (`/usr/bin/grep` → `grep`), strip git global opts (`git -C /tmp` → `git`), then match against 60+ regex patterns from `rules.rs`
 4. Apply the rewrite — find the matching rule, replace the command prefix with `rtk <cmd>`, re-prepend the env prefix, re-append the redirect suffix
--- a/src/discover/lexer.rs
+++ b/src/discover/lexer.rs
@@ -543,19 +543,32 @@ fn is_shell_control_keyword(value: &str) -> bool {
     )
 }
 
-/// True for constructs the permission gate can't decompose, so they must never
-/// be auto-allowed: command/process substitution, or a real file-target redirect
-/// (fd-dup like `2>&1` and `/dev/null` are exempt). Separators and subshells are
-/// handled by [`split_for_permissions`], not flagged here.
-pub fn contains_unattestable_construct(cmd: &str) -> bool {
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum UnattestableConstruct {
+    Substitution,
+    FileTargetRedirect,
+}
+
+/// Returns the first construct the permission gate cannot decompose. These
+/// constructs must never be auto-allowed: command/process substitution, or a
+/// real file-target redirect. Separators and subshells are handled by
+/// [`split_for_permissions`].
+pub(crate) fn first_unattestable_construct(cmd: &str) -> Option<UnattestableConstruct> {
     if contains_substitution(cmd) {
-        return true;
+        return Some(UnattestableConstruct::Substitution);
     }
     let tokens = tokenize(cmd);
     tokens
         .iter()
         .enumerate()
         .any(|(i, tok)| tok.kind == TokenKind::Redirect && redirect_has_file_target(&tokens, i))
+        .then_some(UnattestableConstruct::FileTargetRedirect)
+}
+
+/// True for constructs the permission gate cannot decompose, so they must never
+/// be auto-allowed.
+pub fn contains_unattestable_construct(cmd: &str) -> bool {
+    first_unattestable_construct(cmd).is_some()
 }
 
 /// Quote-aware: bash runs backtick/`$(...)` unquoted and inside double quotes,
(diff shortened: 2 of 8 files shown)


reneleonhardt:
Thank you so much for your important fix! ❤️

In case you want to update, `git rebase master` resolved automatically, that warning seems strange to me in case "Changes can be cleanly merged." is shown too 🤷

aeppling:
Hey @ahundt, thanks for this contribution.

I'm currently working on the rewrite, since models now often use multi-line statements: #2781
It would be nice to ship both in the same release.

If you're available to align with my PR, I can merge mine and you adjust this one afterwards; otherwise I can try to retrieve your fixes into mine. Your call

### If you want to align:

1. Keep rewrite_command_with_policy and requires_ask, and wire it into #2781's clamp: Allow requires attestable && !requires_ask && verdict == Allow, everything else that rewrites is Ask. Either signal can force ask; both must agree to auto-allow.
2. Apply your logic per statement: rewrite_command now splits multi-line bundles and routes each statement through rewrite_logical_line (registry.rs)
3. Existing tests assert git log -10 | grep feat → rewritten; your consumer-aware rule changes that to raw. Update them as intended behavior.

### Before merge

1. rerun test and usage with changes
2. solve any conflicts

ahundt:
Addressed the multiline-rewrite feedback in the current head (968cf993). The branch now uses `src/discover/registry.rs::rewrite_multiline_block` with the shared lexer’s quote-aware newline tokens, rewrites independent clean lines, preserves raw pipeline consumers and substitution-bearing statements, and defers unsafe shell blocks. `src/hooks/rewrite_cmd.rs` keeps the policy path: unattestable rewrites remain Ask-only, and file-target redirects retain `requires_ask`. I validated the current head with `2634 passed, 8 ignored`, `cargo fmt --all -- --check`, `cargo clippy --locked --all-targets --all-features -- -D warnings`, `git diff --check`, and a clean merge-tree check against current `upstream/develop`. I did not merge the older overlapping #2781 implementation because it conflicts with this newer lexer/policy implementation; the requested behavior is already present in this head and its follow-up commits.

KuSh:
## Round 2 of 3 — DUPLICATE, recommend closing

**Claim**: Lex shell structure before rewriting so pipeline/redirect content consumers receive the producer's original output (fixes #1560).

**Scope**: **duplicate — superseded by #3128** (with #3343 `PipelineSafety` and #3704 lexer consolidation). Issue #1560, this PR's entire target, was closed 2026-09-05 as fixed on develop.

**Ran**: both binaries built from `upstream/develop` (d0c29851) and this head (9e8d1285); `rtk rewrite` over 28 command shapes — content consumers (grep, rg, tee, xargs, wc, awk, sed, less), safe display tails (head, tail, cat), redirects (`>`, `>>`, `<`, `2>&1`), substitutions, heredocs, `if`/`for`/`case`/`(...)`/`{...}` blocks, quoting and comment edge cases; plus a live correctness check of develop's rewrite against real `git log` output.

**CI** @9e8d1285: 10 green, `benchmark` failing on `find --max 10` (-20%) and `golangci-lint` (empty output) — both pre-existing and untouched by this diff. Head is 3 weeks stale and conflicts with develop in **every** file it touches.

### Why develop already covers this

Develop fixes #1560 by a different and strictly better mechanism: it leaves the **producer** raw and rewrites the **consumer**, instead of leaving the whole pipeline raw.

```
git log | grep feat
  develop: git log | rtk grep feat      <- git log raw, grep reads the real stream
  this PR: <raw>                        <- correct, but no savings
```

Verified the survivor is genuinely correct, not just smaller: `git log -50 | grep feat` and `git log -50 | rtk grep feat` both yield 17 matching lines, so `grep` is reading the unmodified Git stream and `rtk grep` only compacts the display.

Every consumer this PR names is already handled on develop — `tee`, `xargs`, `wc`, `awk`, `sed`, `less` stay fully raw on both; `head`, `tail`, `cat` are rewritten as safe display tails on both.

### The residual delta is a regression

Because this PR drops the whole statement to raw where develop rewrites the consumer, it loses savings that develop keeps:

| command | develop | this PR |
|---|---|---|
| `git diff HEAD~30 \| grep '^+'` | 1,485,061 → 8,483 B (**99%**) | 0% |
| `git log 2>&1 \| grep feat` | `git log 2>&1 \| rtk grep feat` | raw |
| `git log\|grep feat` | `git log\| rtk grep feat` | raw |
| `git status && git log \| grep feat` | both stages rewritten | tail left raw |
| `if git status; then git log \| grep feat; fi` | rewritten, producer raw | raw |

Against the 20% floor (CONTRIBUTING.md), these go from 99% to 0% on a very common shape.

### The one novel piece is unsafe

The file-redirect rewrite is the only behavior here develop lacks, and it reintroduces the content-sink hazard this PR exists to close:

```
git status > /tmp/status.log   ->   rtk git status > /tmp/status.log
```

The user's file then holds RTK's compact rendering (57 B) instead of the raw output (123 B) — a content consumer being fed display output, the same class as #1560, just a file sink instead of a pipe. `requires_ask` narrows the blast radius but does not make it correct; develop's refusal to rewrite is the safe behavior. `>>` behaves the same.

Relatedly, the PR body states "Input redirects … remain conservative", but `git log < /tmp/in` is rewritten to `rtk git log < /tmp/in` on this head, while develop leaves it raw.

### Your questions

- aeppling: *"If you're available to align with my PR, I can merge mine and you adjust this one afterwards; otherwise I can try to retrieve your fixes into mine. Your call"* → Overtaken by events. #2781 is still open, but the pipeline-safety half of this PR landed separately via #3128/#3343/#3704 while this branch sat. Nothing here needs retrieving into #2781.
- ahundt: *"I did not merge the older overlapping #2781 implementation because it conflicts with this newer lexer/policy implementation; the requested behavior is already present in this head"* → Accepted as accurate for the head as of August. Since then develop grew its own lexer consolidation and `PipelineSafety` model, so the comparison that matters is now against develop rather than against #2781.

### Next

Recommend closing as superseded by #3128. The analysis in this PR was right about the bug and about the consumer/display-tail distinction — develop reached the same conclusions and kept the savings. Nothing in the remaining diff is worth rebasing through conflicts in all six files.

Rounds: 2/3. Threads: 0 open, 0 resolved.

Issue #1812, opened 2026-05-09
Codex now support PreToolUse hook


GuGuMao8698:
As the title suggests, could you add a PreToolUse hook to the Codex, similar to the one in Claude's code?

GengDaPeng:
+1

vltansky:
+1

vltansky:
https://github.com/rtk-ai/rtk/pull/1550

avidspartan1:
+1

zhper-liang:
+1

giordani-technium:
+1

T0UGH:
+1

zhper-liang:
Thanks for the support



---- Replied Message ----
| From | ***@***.***> |
| Date | 05/14/2026 20:36 |
| To | ***@***.***> |
| Cc | Shuaijie ***@***.***>***@***.***> |
| Subject | Re: [rtk-ai/rtk] Codex now support PreToolUse hook (Issue #1812) |
T0UGH left a comment (rtk-ai/rtk#1812)

+1

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

tjx666:
please!

shurkanTwo:
+1

YunosukeYoshino:
+1

EricCorleone:
+1

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 88% |

### Summary

Feature request to add PreToolUse hook support to Codex, matching existing Claude Code hook functionality.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

glinkot:
Would be great.

wongsyrone:
To fix codex PreToolUse hook error: `error: PreToolUse hook returned updatedInput without permissionDecision:allow`

I indicate codex and claude code write a bridge powershell 7 script, after testing, looks good.

in .codex/hooks.json
```json
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "^(Bash|exec)$",
[...]

PR #2557, opened 2026-06-24
Keep Codex approval authoritative when RTK rewrites Bash commands

ahundt:
# Keep Codex approval authoritative when RTK rewrites Bash commands

## Summary

RTK now provides a native `rtk hook codex` handler for Codex `PreToolUse` events. It returns Codex-valid JSON and never claims an approval that Codex has not granted.

The change addresses [issue #1812](https://github.com/rtk-ai/rtk/issues/1812) and the failure reported as `PreToolUse hook returned updatedInput without permissionDecision:allow`. Codex receives `updatedInput` only with `permissionDecision: "allow"`; a deny response includes a non-empty reason and no replacement command.

## Before and after

| Input or policy | Before | After |
|---|---|---|
| `rtk hook codex` | RTK had no native Codex PreToolUse response path. | RTK accepts Codex hook JSON and returns a valid response shape. |
| Default or ask mode, `git status` | A rewrite could include `updatedInput` without `allow`, which Codex rejects. | RTK returns `{}` and leaves approval and execution to Codex. |
| `bypassPermissions`, `git status` | RTK could not provide a native rewrite response. | RTK returns `permissionDecision: "allow"` with `updatedInput.command: "rtk git status"`. |
| A denied command | A host-specific response could omit Codex's required reason or include an invalid replacement. | RTK returns `permissionDecision: "deny"` with a non-empty reason and no `updatedInput`. |
| Non-Bash or non-`PreToolUse` input | An unrelated event could be interpreted as a command rewrite. | RTK returns `{}` and does nothing. |

## Implementation

`Host::Codex` uses the shared rewrite and permission helpers without reading Claude settings. `process_codex_payload` preserves ask-only file redirects, renders neutral/allow/deny responses according to Codex's schema, and keeps Codex responsible for approval and execution policy. The other host adapters keep their existing response schemas.

Hook-file installation is intentionally separate so the protocol contract and settings-file mutation can be reviewed independently.

Codex can still display one status line for each matching PreToolUse handler; this change does not remove or disable foreign handlers. It ensures RTK's own response is valid when Codex executes it alongside other handlers.

## Verification

- 20 Codex-focused tests cover event and tool filtering, malformed input, default and ask neutrality, bypass rewriting, file redirects, deny reasons, and extra Codex fields.
- The full branch suite passes with 2,574 tests passed and 8 ignored.
- `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --all -- --check`, and `git diff --check` pass.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -539,7 +539,7 @@ fn process_codex_payload(v: &Value) -> PayloadAction {
         None => return PayloadAction::Ignore,
     };
 
-    let decision = decide_hook_action(cmd, permissions::Host::Claude);
+    let decision = decide_hook_action(cmd, permissions::Host::Codex);
     process_codex_payload_with_decision(v, cmd, decision)
 }
 
@@ -575,6 +575,13 @@ fn pre_tool_use_deny_output(reason: &str) -> Value {
             "permissionDecisionReason": reason
         }
     })
+}
+
+fn is_codex_bash_pre_tool_use(v: &Value) -> bool {
+    v.get("tool_name").and_then(Value::as_str) == Some("Bash")
+        && v.get("hook_event_name")
+            .and_then(Value::as_str)
+            .is_some_and(|event| event == PRE_TOOL_USE_KEY)
 }
 
 /// Run the Claude Code PreToolUse hook natively.
@@ -635,9 +642,11 @@ fn run_claude_inner(input: &str) -> Option<String> {
 // Codex rejects `updatedInput` unless the same hook output also carries
 // `permissionDecision: "allow"`. Claude can rewrite and still let its
 // native prompt run by omitting `permissionDecision`; Codex cannot. In
-// Codex's `default` permission mode, RTK therefore emits `{}` unless an
-// explicit RTK allow rule matches. Codex's `bypassPermissions` mode already
-// disables approval prompts, so eligible RTK rewrites may be allowed there.
+// Codex's default and prompt modes, RTK therefore emits `{}` and leaves the
+// original command to Codex. Codex's `bypassPermissions` mode already
+// disables approval prompts, so safe RTK rewrites may be allowed there.
+// Codex's approval and execpolicy settings are intentionally not read as
+// Claude settings; Codex remains the authority for Codex commands.
 // Codex semantics permit multiple matching hooks to run concurrently —
 // unlike Claude #1515, there is no manifest fallthrough to engineer here.
 
@@ -662,8 +671,7 @@ pub fn run_codex() -> Result<()> {
     // Codex currently fires PreToolUse only for the Bash tool; defend
     // against future event/tool expansion by emitting `{}` for anything
     // else rather than misinterpreting the payload.
-    let tool_name = v.get("tool_name").and_then(|t| t.as_str()).unwrap_or("");
-    if tool_name != "Bash" {
+    if !is_codex_bash_pre_tool_use(&v) {
         let _ = writeln!(io::stdout(), "{{}}");
         return Ok(());
     }
@@ -697,8 +705,7 @@ pub fn run_codex() -> Result<()> {
 #[cfg(test)]
 fn run_codex_inner(input: &str) -> Option<String> {
     let v: Value = serde_json::from_str(input).ok()?;
-    let tool_name = v.get("tool_name").and_then(|t| t.as_str()).unwrap_or("");
-    if tool_name != "Bash" {
+    if !is_codex_bash_pre_tool_use(&v) {
         return None;
     }
     match process_codex_payload(&v) {
@@ -716,8 +723,7 @@ fn run_codex_inner_with_rules(
     allow: &[String],
 ) -> Option<String> {
     let v: Value = serde_json::from_str(input).ok()?;
-    let tool_name = v.get("tool_name").and_then(|t| t.as_str()).unwrap_or("");
-    if tool_name != "Bash" {
+    if !is_codex_bash_pre_tool_use(&v) {
         return None;
     }
     let cmd = v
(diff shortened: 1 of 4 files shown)


Trirrin (approved):
Reviewed the Codex PreToolUse hook implementation. The hook output shape matches Codex requirements: rewrites are only emitted with permissionDecision=allow, deny responses omit updatedInput, and hooks.json install/uninstall preserves foreign entries. Verified locally with cargo test -q codex, cargo fmt --all -- --check, and git diff --check.

TyceHerrman:
this would implement #1812 as well afaik

KuSh (changes requested):
Hi @ahundt thanks for your work.
Could you rebase it?

Also reviewing this PR turned up one issue worth fixing before we can merge your work:

**`Host::Codex` rules are hardcoded empty in `check_command_for` (`src/hooks/permissions.rs`), so the deny path is dead code in production.**

`check_command_for(cmd, Host::Codex)` always returns empty deny/ask/allow vectors, meaning `PermissionVerdict::Deny` (and an explicit `PermissionVerdict::Allow`) can never actually be produced for real Codex traffic, every command resolves to `PermissionVerdict::Default`. The 20 Codex-focused tests that exercise deny reasons only pass because they go through the test-only `run_codex_inner_with_rules` helper, which injects rules directly and bypasses `check_command_for` entirely, so they don't cover the real `run_codex` → `process_codex_payload` production path.

Concretely, this means the PR description's "A denied command → RTK returns `permissionDecision: "deny"` with a non-empty reason" and the README's deny-response row aren't actually reachable today. There's no real Codex-specific rule set wired in yet.

Could you either:
- wire real deny/ask/allow rules for `Host::Codex` (mirroring how `Host::Claude`/other hosts source theirs), and extend the tests to go through `check_command_for` for at least one deny case, or
- if Codex intentionally has no host-specific rule set yet and rewrites are meant to be governed purely by the rewrite registry + `permission_mode` check, adjust the PR description/README to say so, and consider short-circuiting `check_command_for` for `Host::Codex` (it currently still runs the full compound-command lexer/split pass just to land on a statically-known `Default` result every time).

Three smaller things while I'm here:
- `dontAsk` is one of Codex's documented no-prompt `permission_mode` values, alongside `bypassPermissions`, but only `bypassPermissions` promotes an `AskRewrite` to `AllowRewrite` (`hook_cmd.rs:552`). Worth confirming whether `dontAsk` should behave the same way, or whether that's deliberate.
- The `PayloadAction::Deny` arm added to `run_claude()` (`hook_cmd.rs:615`) looks unreachable. `process_claude_payload` always passes `DenyHandling::Skip`, so only Codex's path can construct that variant.
- `verdict.clone()` at `hook_cmd.rs:551` is unnecessary. `PermissionVerdict` has no owned data in any variant, so it could just derive `Copy` instead of `Clone` and drop the `.clone()` call.

src/hooks/hook_cmd.rs
```diff
@@ -581,7 +587,22 @@ fn process_claude_payload(v: &Value) -> PayloadAction {
...
+}
+
+fn process_pre_tool_use_payload_with_decision(
+    v: &Value,
+    cmd: &str,
+    decision: HookDecision,
+    ask_rewrite_handling: AskRewriteHandling,
+) -> PayloadAction {
+    let (rewritten, allow) = match decision {
         HookDecision::Deny => {
```
KuSh, line 606: The deleted test `test_codex_deny_emits_block_without_updated_input` was the only test exercising `HookDecision::Deny` through `process_pre_tool_use_payload_with_decision` (the renderer shared by Claude and Codex). No replacement asserts this Deny -> Skip mapping at the payload-processing level for either host post-refactor (only the lower-level `decide_from_verdict`/`test_decide_deny` unit test remains).

Failure scenario: a future change to this shared renderer that mishandles `HookDecision::Deny` (e.g. accidentally emits a rewrite instead of Skip) would pass CI green for both Claude and Codex, since nothing exercises Deny at the `process_*_payload` level any more. Worth adding a payload-level test back for at least one host.

src/hooks/hook_cmd.rs
```diff
@@ -626,6 +655,51 @@ fn process_claude_payload(v: &Value) -> PayloadAction {
     }
 }
 
+fn process_codex_payload(v: &Value) -> PayloadAction {
```
KuSh, line 658: `process_codex_payload` bypasses the shared `Host`/`check_command_for` abstraction entirely by hardcoding `PermissionVerdict::Default`, instead of adding a `Host::Codex => (Vec::new(), Vec::new(), Vec::new())` arm the way `Host::Vibe` (permissions.rs:48) already does for a rule-less host.

Failure scenario: Codex is now the only host routed around the shared per-host rule-loading match arm. If a shared concern is later added there (audit/telemetry keyed on Host, or real Codex-specific rules), Codex's hardcoded bypass has to be updated separately and can silently drift from every other host's handling.

KuSh (changes requested):
Verified the 4 items from the previous review round (Host::Codex removal, dontAsk neutrality, unreachable Deny arm removal, verdict Copy) — all confirmed fixed in 718a035; `cargo test`, `cargo clippy --all-targets`, and `cargo fmt --check` all pass clean.

Found 2 more issues that aren't part of this diff so they can't be attached as inline comments here, flagging in the review body instead:

- **`src/hooks/init.rs` (`run_codex_mode_with_paths`, ~line 2384)**: `rtk init --codex` only writes `AGENTS.md`/`RTK.md`. Unlike Cursor/Droid/Vibe, there's no patch function that registers the new `rtk hook codex` PreToolUse handler into any Codex hook config. As it stands, the new hook (20 passing tests, fully wired into `main.rs`) is unreachable for real users unless they hand-edit their Codex config.
- **`hooks/codex/README.md:7`** (and `docs/guide/getting-started/supported-agents.md:43/231`, `docs/contributing/TECHNICAL.md:321/339`): still states "Prompt-level guidance via awareness document -- no programmatic hook", which is now false given this PR adds a full programmatic PreToolUse hook. Reinforces the install-gap above by hiding the new capability from anyone reading the docs.

Two inline comments below on code actually touched by this diff.

KuSh:
Thanks for this. Native Codex PreToolUse support landed in #3552, merged on 2026-09-13, which added `rtk hook codex` with the same event and tool filtering, permission mode gating, passthrough for unattestable constructs, and the allow shape Codex requires alongside updatedInput, with tests for each of those cases; the follow-up commit 0d7e4fba then routed it through the shared hook decision helpers. The handler you proposed therefore exists on develop, but with a deliberately different response model: both allow and ask rewrites emit a protocol level allow, because Codex cannot apply updatedInput without one, and the reasoning is documented in the code. One case is still open there: a deny decision produces only an audit entry and no response, so Codex never sees the deny reason. That is worth a small issue or a narrow PR against the current handler in src/hooks/hook_cmd.rs. If a part of your change was missed, please comment here or reopen.

PR #2564, opened 2026-06-24
fix(build): restore stable Rust test gate

YOMXXX:
## Summary
- replace unstable `str::floor_char_boundary` usage with a stable UTF-8 byte-prefix helper
- isolate curl and tracking tests from user tee/database paths
- keep the existing clippy gate useful by allowing the repo-wide `uninlined_format_args` style lint

## Verification
- `cargo fmt --all -- --check && cargo clippy --all-targets && cargo test --all`
  - `cargo test --all`: 1903 passed, 0 failed, 6 ignored

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -44,8 +44,7 @@ fn main() {
     // Validate: parse the combined TOML to catch errors at build time
     let parsed: toml::Value = combined.parse().unwrap_or_else(|e| {
         panic!(
-            "TOML validation failed for combined filters:\n{}\n\nCheck src/filters/*.toml files",
-            e
+            "TOML validation failed for combined filters:\n{e}\n\nCheck src/filters/*.toml files"
         )
     });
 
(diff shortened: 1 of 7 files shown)


YOMXXX:
Rebased onto current develop and resolved conflicts in Cargo.lock and pipe_cmd.rs.

Conflict decisions:
- kept the current package version from Cargo.toml/Cargo.lock (0.42.4) instead of the stale lockfile version from the old branch base
- kept develop's pipe truncation caps while preserving the UTF-8-safe prefix helper used to avoid stable-Rust build/test issues

Local verification:
- cargo +1.93.0 fmt --check
- cargo +1.93.0 test utf8_prefix_by_byte_limit -- --nocapture

Note: the skipped grep test is an existing local-environment failure here; the remaining unit/integration suite passed.

KuSh:
Thanks for this, and sorry it sat so long. The three pieces here have all been handled on develop since June: the tracking tests moved to Tracker::new_in_memory with #3206 (merged 2026-09-02) and #891 (merged 2026-09-13), and the RTK_DB_PATH cases moved to temp_env under #3782 (merged 2026-09-14), which also means the EnvVarRestore helper here would no longer compile because env mutation is unsafe in edition 2024. floor_char_boundary is stable on the pinned toolchain and develop now calls it in four places, so the stable prefix helper and the clippy allow are not needed any more. One case is still open: the curl truncation tests call filter_curl_output directly and go through the real recovery store, so an issue or a small PR adding a seam or a temp_env override there would be welcome. Please comment or reopen if I missed a part of this.

PR #2565, opened 2026-06-24
fix(hook): fail open when stdin payload stalls

### LINKED ISSUE #2553: `rtk hook claude` blocks indefinitely on stdin (no read timeout)
## Summary

`rtk hook claude` reads its tool-call payload from stdin with no deadline. If stdin is open but the payload has not yet arrived and no EOF is sent, it blocks forever. As a global `PreToolUse` hook, it then wedges the Bash tool call indefinitely.

## Reproduce

```sh
printf '' | rtk hook claude        # EOF, no payload -> returns instantly (exit 0)
{ sleep 12; } | rtk hook claude    # stdin open, no data/EOF -> blocks the full 12s
```

A writer that never closes stdin blocks it indefinitely.

## Impact

Under Claude Code background subagents the hook intermittently does not receive its stdin payload promptly; `rtk hook claude` then hangs, the Bash call never executes, and the subagent's 600s no-progress watchdog kills it. Reproduced three consecutive subagent failures this way; the single-threaded main session is rarely affected.

Ruled out as causes: large grep output (rtk truncates, returns instantly) and concurrent `history.db` writes (24 concurrent `rtk grep` complete in 0s). The stdin read block is the defect.

## Fix

Read the payload with a short deadline (e.g. 1–3s) and fail open — emit a no-op passthrough decision and exit 0 — instead of blocking on the read. A hook must never be able to hang the host indefinitely.

YOMXXX:
## Summary
- add a 1s deadline around hook stdin payload reads so `rtk hook claude` cannot block indefinitely when stdin stays open without data
- preserve the existing empty-input fail-open path by returning no rewrite on timeout
- add integration coverage for normal payload rewrites and stalled stdin fail-open behavior

Fixes #2553

## Verification
- `cargo +1.93.0 test --test hook_stdin_timeout_test -- --nocapture`
- `cargo +1.93.0 test test_claude_ -- --nocapture`
- `git diff --check`

## Known baseline issue
- `cargo +1.93.0 test --all` currently fails on upstream `develop` at `tests/grep_compress_test.rs::small_grep_not_worse_than_plain`, unrelated to this hook change; this appears covered by existing PR #2554.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -8,21 +8,37 @@ use super::permissions::{self, PermissionVerdict};
 use anyhow::{Context, Result};
 use serde_json::{json, Value};
 use std::io::{self, Read, Write};
+use std::sync::mpsc::{self, RecvTimeoutError};
+use std::thread;
+use std::time::Duration;
 
 use crate::discover::registry::{has_heredoc, rewrite_command};
 
 const STDIN_CAP: usize = 1_048_576; // 1 MiB
+const STDIN_READ_TIMEOUT: Duration = Duration::from_secs(1);
 
 fn read_stdin_limited() -> Result<String> {
-    let mut input = String::new();
-    io::stdin()
-        .take((STDIN_CAP + 1) as u64)
-        .read_to_string(&mut input)
-        .context("Failed to read stdin")?;
-    if input.len() > STDIN_CAP {
-        anyhow::bail!("hook stdin exceeds {} byte limit", STDIN_CAP);
+    let (tx, rx) = mpsc::sync_channel(1);
+
+    thread::spawn(move || {
+        let mut input = String::new();
+        let result = io::stdin()
+            .take((STDIN_CAP + 1) as u64)
+            .read_to_string(&mut input)
+            .context("Failed to read stdin")
+            .and_then(|_| {
+                if input.len() > STDIN_CAP {
+                    anyhow::bail!("hook stdin exceeds {} byte limit", STDIN_CAP);
+                }
+                Ok(input)
+            });
+        let _ = tx.send(result);
+    });
+
+    match rx.recv_timeout(STDIN_READ_TIMEOUT) {
+        Ok(result) => result,
+        Err(RecvTimeoutError::Timeout | RecvTimeoutError::Disconnected) => Ok(String::new()),
     }
-    Ok(input)
 }
 
 // ── Copilot hook (VS Code + Copilot CLI) ──────────────────────
--- a/tests/hook_stdin_timeout_test.rs
+++ b/tests/hook_stdin_timeout_test.rs
@@ -1,0 +1,79 @@
+use std::io::Write;
+use std::process::{Command, Stdio};
+use std::thread;
+use std::time::{Duration, Instant};
+
+#[test]
+fn claude_hook_rewrites_payload_arriving_on_stdin() {
+    let mut child = Command::new(env!("CARGO_BIN_EXE_rtk"))
+        .args(["hook", "claude"])
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("spawn rtk hook claude");
+
+    let mut stdin = child.stdin.take().expect("child stdin");
+    stdin
+        .write_all(br#"{"tool_name":"Bash","tool_input":{"command":"git status"}}"#)
+        .expect("write hook payload");
+    drop(stdin);
+
+    let output = child.wait_with_output().expect("collect hook output");
+    assert!(
+        output.status.success(),
+        "hook should exit 0, got {:?}, stderr={}",
+        output.status.code(),
+        String::from_utf8_lossy(&output.stderr)
+    );
+
+    let response: serde_json::Value =
+        serde_json::from_slice(&output.stdout).expect("hook JSON response");
+    assert_eq!(
+        response
+            .pointer("/hookSpecificOutput/updatedInput/command")
+            .and_then(|v| v.as_str()),
+        Some("rtk git status")
+    );
+}
+
+#[test]
+fn claude_hook_fails_open_when_stdin_stays_open_without_payload() {
+    let mut child = Command::new(env!("CARGO_BIN_EXE_rtk"))
+        .args(["hook", "claude"])
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("spawn rtk hook claude");
+
+    let stdin = child.stdin.take().expect("child stdin");
+    let deadline = Instant::now() + Duration::from_secs(3);
+
+    loop {
+        if child.try_wait().expect("poll child").is_some() {
+            drop(stdin);
+            let output = child.wait_with_output().expect("collect hook output");
+            assert!(
+                output.status.success(),
+                "hook should fail open with exit 0, got {:?}, stderr={}",
+                output.status.code(),
+                String::from_utf8_lossy(&output.stderr)
+            );
+            assert!(
+                output.stdout.is_empty(),
+                "hook should emit no rewrite when no payload arrives, got {}",
+                String::from_utf8_lossy(&output.stdout)
+            );
+            return;
+        }
+
+        if Instant::now() >= deadline {
+            let _ = child.kill();
+            let _ = child.wait();
+            panic!("rtk hook claude blocked on open stdin without payload");
+        }
+
+        thread::sleep(Duration::from_millis(25));
+    }
+}


src/hooks/hook_cmd.rs
```diff
@@ -8,21 +8,37 @@ use super::permissions::{self, PermissionVerdict};
...
+        let result = io::stdin()
+            .take((STDIN_CAP + 1) as u64)
+            .read_to_string(&mut input)
+            .context("Failed to read stdin")
+            .and_then(|_| {
+                if input.len() > STDIN_CAP {
+                    anyhow::bail!("hook stdin exceeds {} byte limit", STDIN_CAP);
+                }
+                Ok(input)
+            });
```
KuSh, line 34: **Blocking: a payload that has already arrived in full is discarded on timeout.**

`read_to_string` returns only at EOF, so this closure produces a value only once the writer closes stdin. On the timeout path the thread's local `input` is dropped along with everything it had already read, and `read_stdin_limited` returns `Ok(String::new())` no matter how many bytes actually arrived.

So a host that writes the whole payload but holds stdin open past the deadline gets no rewrite at all — silently, and for every subsequent call. Measured on this branch against `develop`, same fake `$HOME`, payload written immediately and stdin held open for 4s:

```console
$ { printf '{"tool_name":"Bash","tool_input":{"command":"git status"}}'; sleep 4; } | rtk hook claude

develop:  {"hookSpecificOutput":{...,"updatedInput":{"command":"rtk git status"}}}
this PR:  (no output, exit 0)
```

This is the finding that worries me most, because #2553 describes the payload arriving *late* — which is exactly the shape above. The hook stops hanging, but it also stops working, and nothing tells the user it has.

Suggestion: have the reader append into a shared buffer (`Arc<Mutex<Vec<u8>>>`, or send chunks over the channel) so the timeout path can parse whatever arrived, and fall back to the empty string only when nothing did.

src/hooks/hook_cmd.rs
```diff
@@ -8,21 +8,37 @@ use super::permissions::{self, PermissionVerdict};
...
-        .take((STDIN_CAP + 1) as u64)
-        .read_to_string(&mut input)
-        .context("Failed to read stdin")?;
-    if input.len() > STDIN_CAP {
-        anyhow::bail!("hook stdin exceeds {} byte limit", STDIN_CAP);
-    }
-    Ok(input)
+    let (tx, rx) = mpsc::sync_channel(1);
+
+    thread::spawn(move || {
```
KuSh, line 23: `thread::spawn` panics if the OS refuses to create the thread; the previous inline read had no such failure mode.

#2553 attributes the hang to many concurrent background subagents — the scenario most likely to be near a thread or PID limit — so this adds an abort path in precisely the situation the fix is meant to help. `thread::Builder::new().spawn(...)` with a fallback to the old inline read on `Err` keeps that case benign.

src/hooks/hook_cmd.rs
```diff
@@ -8,21 +8,37 @@ use super::permissions::{self, PermissionVerdict};
...
 use serde_json::{json, Value};
 use std::io::{self, Read, Write};
+use std::sync::mpsc::{self, RecvTimeoutError};
+use std::thread;
+use std::time::Duration;
 
 use crate::discover::registry::{has_heredoc, rewrite_command};
 
 const STDIN_CAP: usize = 1_048_576; // 1 MiB
+const STDIN_READ_TIMEOUT: Duration = Duration::from_secs(1);
```
KuSh, line 18: 1s sits at the low end of the 1–3s the issue suggests, and it is not overridable.

Claude Code's own PreToolUse budget is far larger, and the reported trigger is a loaded machine running background subagents — the case where a legitimate payload is most likely to miss a 1s deadline. Worth reading an override from config or an env var, and/or raising the default, so a host that is merely slow does not permanently lose rewriting.

src/hooks/hook_cmd.rs
```diff
@@ -8,21 +8,37 @@ use super::permissions::{self, PermissionVerdict};
...
+                }
+                Ok(input)
+            });
+        let _ = tx.send(result);
+    });
+
+    match rx.recv_timeout(STDIN_READ_TIMEOUT) {
+        Ok(result) => result,
+        Err(RecvTimeoutError::Timeout | RecvTimeoutError::Disconnected) => Ok(String::new()),
+    }
```
KuSh, line 41: Two things worth reconsidering in this match.

**1. The fail-open is completely silent** — no `eprintln!`, no `audit_log("skip:stdin_timeout", ...)`, nothing. Combined with the payload discard above, a user whose host delays EOF sees rewriting stop with no evidence anywhere. Other paths in this file already write diagnostics to stderr, so a one-line warning is safe for the hook protocol; an audit entry would also surface it in `rtk hook audit`.

**2. `Disconnected` is folded in with `Timeout`.** Disconnect here means the reader thread panicked (for instance while allocating the ~1 MiB `String`). Turning that into "stdin was empty" hides a genuine fault behind a normal-looking passthrough. Splitting the arms costs one line and keeps the two causes distinguishable.

tests/hook_stdin_timeout_test.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[test]
+fn claude_hook_rewrites_payload_arriving_on_stdin() {
+    let mut child = Command::new(env!("CARGO_BIN_EXE_rtk"))
+        .args(["hook", "claude"])
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("spawn rtk hook claude");
```
KuSh, line 14: **This test is not hermetic and will fail on real developer and CI machines.**

The child is spawned with no environment isolation, so it reads the developer's actual configuration: `Config::load()` picks up `hooks.exclude_commands` from `$XDG_CONFIG_HOME` / `~/.config/rtk/config.toml`, and `permissions::check_command_for(cmd, Host::Claude)` reads `.claude/settings.json` from the cwd and from `$HOME`.

Anyone with `exclude_commands = ["git"]`, or a `Bash(git status)` deny/ask rule, gets no rewrite — so `output.stdout` is empty and line 31 panics on `serde_json::from_slice(&[])` on an otherwise correct build. Reproduced:

```console
$ mkdir -p $H/.config/rtk && printf '[hooks]\nexclude_commands = ["git"]\n' > $H/.config/rtk/config.toml
$ printf '{"tool_name":"Bash","tool_input":{"command":"git status"}}' \
    | HOME=$H XDG_CONFIG_HOME=$H/.config rtk hook claude
(no output, exit 0)
```

`tests/copilot_selfheal_test.rs:93` already sets `HOME`/`COPILOT_HOME` on the child for this reason — doing the same here, plus `XDG_CONFIG_HOME`, would make both tests in this file hermetic.

tests/hook_stdin_timeout_test.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    let stdin = child.stdin.take().expect("child stdin");
+    let deadline = Instant::now() + Duration::from_secs(3);
```
KuSh, line 51: 3s is tight: the child has to spawn, get through a debug-profile `rtk` startup, and then wait out the full 1s timeout, all while `cargo test` runs other tests in parallel. On a loaded or Windows runner that can produce a spurious `blocked on open stdin without payload` failure.

The assertion is "does not block forever", not a precise deadline, so something like 10s would keep the signal and drop the flake.

src/hooks/hook_cmd.rs
```diff
@@ -8,21 +8,37 @@ use super::permissions::{self, PermissionVerdict};
...
+                    anyhow::bail!("hook stdin exceeds {} byte limit", STDIN_CAP);
+                }
+                Ok(input)
+            });
+        let _ = tx.send(result);
+    });
+
+    match rx.recv_timeout(STDIN_READ_TIMEOUT) {
+        Ok(result) => result,
+        Err(RecvTimeoutError::Timeout | RecvTimeoutError::Disconnected) => Ok(String::new()),
```
KuSh, line 40: **Supplementary — the callers of this empty string are not all ready for it.**

(Anchored here because this is the expression that produces it; the affected code is further down the file, outside the diff.)

`read_stdin_limited` can now return an empty string on a path that previously could not occur mid-session, and not every caller is ready for that.

- **`run_gemini` (line 398)** parses with `serde_json::from_str(&input).context(...)?` and has no empty-input guard. On timeout that `?` propagates, `main` prints `rtk: Failed to parse hook input as JSON: EOF while parsing a value` and exits 1. A non-zero exit from a BeforeTool hook is a *blocking* failure, so on Gemini this change fails **closed** — the opposite of what the PR sets out to do. The guard is missing on `develop` today (`printf '' | rtk hook gemini` already exits 1), but before this PR only a genuine EOF-without-payload could reach it; the timeout makes it reachable on every stall.
- **`run_vibe` (line 450)** is milder — `run_vibe_inner` returns `None` and the process exits 0 — but every timeout now emits a misleading `[rtk hook] Failed to parse JSON input: EOF while parsing a value` on stderr.

`run_claude`, `run_copilot`, `run_cursor` and `run_droid` all guard `input.is_empty()` already. Adding the same guard to `run_gemini` (returning `print_allow()`, since allow is Gemini's fail-open) and to `run_vibe` would make the fail-open actually hold across all six runners.

KuSh (changes requested):
Thanks for taking this on — the diagnosis in #2553 is solid, and a hook that can wedge the host indefinitely is worth fixing.

I built the branch and compared it against `develop` before writing this. I think the current shape has a correctness regression that needs addressing before it can land: on timeout the payload that has **already been read** is thrown away, so a host that delivers the payload but holds stdin open stops getting rewrites entirely, rather than getting them late. Details inline on the reader closure, with the repro.

The rest are smaller: the degraded mode is completely silent, `Disconnected` is conflated with `Timeout`, `thread::spawn` adds an abort path under exactly the load the issue describes, the 1s deadline is not overridable, and the new integration test is not hermetic (it fails on any machine that has `exclude_commands` or a matching permission rule).

One supplementary item that I think belongs in this PR rather than a follow-up: `run_gemini` has no empty-input guard, so the new empty-string return makes it exit 1 — a blocking hook failure on Gemini, the opposite of the intent. It is outside the diff, so I left it as a file-level comment on `src/hooks/hook_cmd.rs`.

Both measured claims have their commands inline so you can reproduce them.

PR #2567, opened 2026-06-24
fix(cargo): preserve quiet check success output

### LINKED ISSUE #2545: rtk causes many false positive iteratoins for `cargo check -q` command with GPT 5.4 mini model
- OS: Windows 11
- Client: OpenCode
- Provider: GitHub Copilot
- Model: GPT-5.4 mini (default variant)
- Installed with `rtk init -g --opencode`

As written in the title, using rtk for `cargo check -q` causes trouble for the model.  `cargo check -q` emits no output when it success but with rtk it outputs `cargo build (0 crates compiled)`. Looks like it is the cause.

```
# Checks Rust compile in .
$ rtk cargo check -q
cargo build (0 crates compiled)

Thought:
I might have entered the wrong command with "rtk cargo check." The tool output indicates "cargo build 0 crates compiled," which seems off. I need to check if the command is actually valid. It feels like I should verify with the proper "cargo check" command instead to see if that resolves the issue. Let's make sure I'm using the right terms here to get accurate results!

Bad command path. Re-run real cargo check, then tests, then clippy.


# Checks Rust compile in .
$ rtk cargo check -q
cargo build (0 crates compiled)

Thought:
It looks like I'm trying to figure out if my note tool has some quirks. The command syntax might be mistaken because it was prefixed incorrectly. I used the bash tool, but the command got wrapped, which is interesting. The output "cargo build (0 crates compiled)" seems tied to the environment. I should execute the actual shell command, so I'm thinking of using "cargo check -q" instead. Let's go ahead and try that!

Shell wrapper weird. Use plain cargo now.

# Checks Rust compile in .
$ rtk cargo check -q
cargo build (0 crates compiled)

Thought:
I’m running into issues with the tool command—maybe there’s a problem with the alias? I realize I need to use an actual shell command for the bash tool. Should I set “rtk cargo,” or could the host be intercepting it? The return shows cargo build 0 with crates compiled, which seems off. I need to correct this with a plain cargo check -q. Let’s think about skipping commentary for now and just run it.

Cargo check command tool miswire. Re-run with plain cargo.

<continues same kind of loop>
```

YOMXXX:
## Summary
- Preserve empty cargo build/check output instead of synthesizing `cargo build (0 crates compiled)`
- Add streaming and buffered regression tests for quiet successful cargo output

Fixes #2545

## Verification
- cargo +1.93.0 test quiet_success_preserves_empty_output -- --nocapture
- cargo +1.93.0 test cargo_build -- --nocapture
- git diff --check
- cargo +1.93.0 run --quiet -- cargo check -q (status=0, bytes=0)

Note: `small_grep_not_worse_than_plain` is a known upstream baseline failure on develop, so the broad test run skips only that test.

### DIFF
--- a/src/cmds/rust/cargo_cmd.rs
+++ b/src/cmds/rust/cargo_cmd.rs
@@ -93,7 +93,11 @@ impl BlockHandler for CargoBuildHandler {
         !(line.trim().is_empty() && block.len() > 3)
     }
 
-    fn format_summary(&self, _exit_code: i32, _raw: &str) -> Option<String> {
+    fn format_summary(&self, _exit_code: i32, raw: &str) -> Option<String> {
+        if raw.trim().is_empty() {
+            return None;
+        }
+
         if self.error_count == 0 && self.warnings == 0 {
             let mut s = format!("cargo build ({} crates compiled)", self.compiled);
             if let Some(ref finished) = self.finished_line {
@@ -758,6 +762,10 @@ fn filter_cargo_nextest(output: &str) -> String {
 }
 
 fn filter_cargo_build(output: &str) -> String {
+    if output.trim().is_empty() {
+        return String::new();
+    }
+
     let mut handler = CargoBuildHandler::new();
     let mut blocks: Vec<Vec<String>> = Vec::new();
     let mut current_block: Vec<String> = Vec::new();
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for this. The synthesized line on a quiet successful build was fixed by #3430, merged on 2026-08-17, which put a never worse guard in front of every cargo summary: format_summary now returns never_worse(raw, &summary), so empty output stays empty instead of becoming `cargo build (0 crates compiled)`. That guard is more general than a special case for empty input and it also covers the `cargo check -q` report in issue #2545, which was closed as completed on 2026-09-05 on that basis. The same change reshaped the handler this PR patches, and CargoBuildHandler::new() no longer exists, so the two tests here would not compile as written; develop still has no cargo level test pinning empty in to empty out for the buffered and streamed paths, only the guard unit test, so a small PR adding those two assertions against the current handler would be welcome. Please comment here or reopen if you still see the extra line.

PR #2568, opened 2026-06-24
fix(rewrite): handle process wrapper prefixes

### LINKED ISSUE #2375: Hook rewrite ignores process-wrapper prefixes (`timeout`, `time`) so wrapped commands run raw/unfiltered
**Component:** hook rewrite engine (`src/discover/registry.rs`)
**Version:** rtk 0.42.2 (reproduced on `develop`)

## What's wrong

When a command is prefixed with a process-wrapper such as `timeout 60 …`, `rtk rewrite` returns *passthrough* (exit 1) instead of rewriting the inner command. The hook then runs the raw command, so no filter is applied and the full, unfiltered output is returned.

This is not filter-specific — it affects every command/ecosystem identically. It shows up in practice because coding agents commonly wrap long-running build/test commands in `timeout` to bound them (e.g. `timeout 300 cargo test`).

## Reproduction

```console
$ rtk rewrite "cargo test"
rtk cargo test                       # rewritten (exit 3)

$ rtk rewrite "timeout 300 cargo test"
                                     # no output, exit 1 (passthrough) — runs raw

$ rtk rewrite "cargo clippy --all-targets"
rtk cargo clippy --all-targets       # rewritten
$ rtk rewrite "timeout 120 cargo clippy --all-targets"
                                     # no output, exit 1 (passthrough)

$ rtk rewrite "git log --oneline"
rtk git log --oneline                # rewritten
$ rtk rewrite "timeout 30 git log --oneline"
                                     # no output, exit 1 (passthrough)
```

Same passthrough for every common process wrapper:

| Command | Result |
|---|---|
| `timeout 300 cargo test` | passthrough (raw) |
| `timeout 5m cargo build` | passthrough |
| `nice -n 10 cargo test` | passthrough |
| `nohup cargo build` | passthrough |
| `time cargo build` | passthrough |
| `stdbuf -oL cargo test` | passthrough |

For comparison, the prefixes that *are* stripped today work as expected:

```console
$ rtk rewrite "sudo cargo test"
sudo rtk cargo test
$ rtk rewrite "RUST_LOG=debug cargo test"
RUST_LOG=debug rtk cargo test
```

## Where it happens

`src/discover/registry.rs` peels a fixed set of prefixes before routing a command to its rule, then re-prepends them to the rewrite:

- `ENV_PREFIX` regex — `sudo`, `env`, and `VAR=val` assignments.
- `SHELL_PREFIX_BUILTINS` — `noglob`, `command`, `builtin`, `exec`, `nocorrect`.

Process-wrapper commands are in neither set. For `timeout 300 cargo test`, `classify_command` sees base word `timeout`, matches no rule, and `rewrite_segment_inner` returns `None` → passthrough.

## Impact

Commands wrapped in `timeout` (and the other wrappers above) receive no RTK filtering and the full raw output, with no error or warning indicating the filter was skipped.

YOMXXX:
## Summary
- Teach rewrite to peel process-wrapper prefixes (`timeout`, `time`, `nice`, `nohup`, `stdbuf`) and rewrite the inner supported command.
- Preserve wrapper options, env prefixes, and compound command segments while leaving incomplete wrappers passthrough.
- Add regression coverage for wrapped cargo/git commands, timeout options, env composition, and missing inner commands.

Fixes #2375

## Verification
- `cargo +1.93.0 test process_wrapper -- --nocapture`
- `cargo +1.93.0 test prefix -- --nocapture`
- `git diff --check`
- `cargo +1.93.0 run --quiet -- rewrite "timeout 300 cargo test"` -> prints `timeout 300 rtk cargo test` and exits 3 per Ask verdict protocol

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -3,7 +3,7 @@
 use lazy_static::lazy_static;
 use regex::{Regex, RegexSet};
 
-use super::lexer::{split_on_operators, tokenize, TokenKind};
+use super::lexer::{split_on_operators, tokenize, ParsedToken, TokenKind};
 use super::rules::{IGNORED_EXACT, IGNORED_PREFIXES, RULES};
 
 /// Result of classifying a command.
@@ -770,6 +770,11 @@ fn rewrite_segment_inner(
         }
     }
 
+    if let Some((prefix, rest)) = strip_process_wrapper_prefix(trimmed) {
+        return rewrite_segment_inner(rest, excluded, transparent_prefixes, depth + 1)
+            .map(|rewritten| format!("{} {}", prefix, rewritten));
+    }
+
     // User-configured wrapper prefixes (e.g. `docker exec mycontainer`). Same
     // strip-recurse-reprepend contract as the builtin list above.
     for prefix in transparent_prefixes {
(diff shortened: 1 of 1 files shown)


aeppling:
Hi @YOMXXX, thanks for this PR, and sorry for the slow review.

The idea is right and the approach (peel the wrapper, recurse, re-prepend) is the correct one. I ported the branch onto current develop and tested it: a 6,692 command differential fuzz against a develop baseline, plus execution tests with argv-recording shims. Findings below.

## Blockers

**1. Fails an existing develop test.** Ported onto develop: 3120 passed / 0 failed becomes 3122 passed / 1 failed.

```
cargo test --all
# test_toml_filter_rewrites_bare_command_but_not_wrapped_invocations
#   left: Some("timeout 5 rtk /usr/bin/jj log")   right: None
```

That assertion comes from the commit rejecting unanchored TOML `match_command` ("activates on `timeout 5 /usr/bin/liquibase update` and rewrites the wrapper"). Please don't just update the test: a maintainer needs to decide whether wrapped TOML-filtered invocations should now be rewritten.

**2. Under `timeout`, all output is lost.** Shim printing 200 lines then hanging, killed at 2s:

| command | exit | lines seen |
|---|---|---|
| `timeout 2 cargo clippy` | 124 | 200 |
| `timeout 2 rtk cargo clippy` | 124 | **0** |
| `timeout 2 cargo test` | 124 | 200 |
| `timeout 2 rtk cargo test` | 124 | **0** |

No tee file either, since it is written after filtering completes. The streaming runner does not help: `rtk cargo test` prints its first line only at 10.0s. And `timeout --foreground 2 rtk cargo clippy` orphans the build, rtk dies while cargo runs to completion.

Repro: a shim doing `for i in $(seq 1 200); do echo line-$i; done; sleep 10`, placed on PATH as `cargo`, then run the four rows above.

Fixing this needs flush-on-signal in `runner`, or `timeout` gated behind that fix.

**3. Rebase.** 3 of 4 hunks reject. `rewrite_segment_inner` now takes a `RewriteContext`; thread the caller's `context` through the new recursive call rather than hardcoding `Normal`, otherwise wrapped commands inside a pipeline escape the pipeline safety checks.

## Must fix

**4. Drop `stdbuf`.** Measured: `stdbuf -oL cargo clippy` delivers its first line at 0.02s, `stdbuf -oL rtk cargo clippy` at 10.02s. The wrapper's entire purpose is removed.

**5. Use the existing table pattern.** Five inline literals in a `match` plus four near-identical scanners, about 175 lines. `SAFE_PIPE_CONSUMERS` next door is the shape to copy: one const table of wrapper descriptors (name, options taking a separate value, attached option forms, required positional count) plus one generic scanner.

**6. Token filtering swallows shell syntax into the prefix.** `rtk rewrite "time (cargo build)"` returns `time ( rtk cargo build)`, where develop returns nothing. The output is valid here, and every dangerous variant I tried (command substitution, backticks, globs, brace groups, redirect before command) falls through to passthrough, so this is safe by luck of classification rather than by design. Scanning tokens in order and bailing out on a non-`Arg` token before the inner index would make it safe by construction.

## Nits

- `rtk rewrite "nice -n10 cargo test"` and `"timeout 300 -- cargo test"` return passthrough. Missed, not broken.
- `"timeout.exe 300 cargo test"` returns nothing, so Windows gets no rewrite, while `"./timeout 300 grep -rn foo src/"` is rewritten, so a local script named `timeout` is assumed to be GNU timeout. Same one line fix in `command_basename`.
- `"timeout rtk cargo test"` returns `timeout rtk rtk cargo test`. The input is already invalid, so cosmetic only.
- Tests worth adding: pipeline producer and final position, nested wrappers, unsupported inner command, `--` forms, `exclude_commands` on the inner command.
- No issue reference on the new branches, while neighbours cite `#2768`, `#530`, `#345`. A short paragraph in `src/discover/README.md` next to "Env Prefix Handling" would fit too.
- Why is `small_grep_not_worse_than_plain` skipped in the verification list? It passes on develop here.

## What already works

Argv and exit code preserving under real execution: 21 wrapped/mode combinations, byte identical argv, exact exit code propagation. Fuzz results: 0 crashes, 0 lost rewrites, 1,465 newly filtered commands, and rewriting is idempotent (2,145 rewrites are a fixed point). The depth guard holds at 200 nested wrappers, `RTK_DISABLED=1` still short circuits through a wrapper, and there is no measurable latency cost (132.9ms vs 133.1ms interleaved).

aeppling:
Solved by https://github.com/rtk-ai/rtk/pull/3937

I've aded you as co author thanks for contributing !

PR #2570, opened 2026-06-24
fix(lint): preserve biome routing

### LINKED ISSUE #2540: [BUG] Biome lint output sometimes reported as `ESLint output (JSON parse failed)`, masking real status
## Environment
- `rtk 0.42.4` (Homebrew, `/opt/homebrew/bin/rtk`)
- macOS (Apple Silicon)
- Project lints with **Biome**, not ESLint: `package.json` → `"lint": "biome check"`

## What happens
My lint script runs Biome. When it goes through rtk's command rewriting, I intermittently get this, with a misleading non-zero exit, **even though Biome itself exits `0` and the project is clean**:

```
ESLint output (JSON parse failed: EOF while parsing a value at line 1 column 0)
```

It masks the real lint result — I had to fall back to running the binary directly (`./node_modules/.bin/biome check`) to know the true status, which cost real debugging time across a multi-PR session (a green lint kept looking red).

It looks like the lint filter assumes ESLint's JSON output format and tries to JSON-parse Biome's text diagnostics, which fails.

## Secondary data point
Running Biome as a bare command through rtk also fails:

```
$ rtk biome check
[rtk: No such file or directory (os error 2)]    # exit 127
```

(Biome resolves fine via `npm run` / `node_modules/.bin`, just not as a top-level rtk-wrapped command.)

## Reproducibility
The `JSON parse failed` message is **intermittent** — I captured it from session logs, but 3 back-to-back `npm run lint` runs just now all passed through cleanly (exit 0, normal Biome output). So it seems to trigger only under some conditions (possibly when Biome reports findings / exits non-zero).

## Expected
- Detect Biome as a linter and handle its output format, **or**
- when the linter isn't ESLint, pass the output through unchanged and **preserve the real exit code** — never synthesize a `JSON parse failed` error or a spurious non-zero exit.

Biome is a very common ESLint/Prettier replacement now, so first-class (or at least transparent) handling would help a lot. Happy to provide more logs if useful.

YOMXXX:
## Summary
- keep explicit biome/eslint rewrite prefixes when routing through `rtk lint`
- infer the package linter from `package.json` lint scripts/dependencies before defaulting to ESLint
- add regression coverage for explicit Biome rewrites and package.json Biome inference

Fixes #2540

## Verification
- `cargo +1.93.0 test preserves_explicit_linter -- --nocapture`
- `cargo +1.93.0 test infer_linter -- --nocapture`
- `cargo +1.93.0 test rewrite_lint -- --nocapture`
- `cargo +1.93.0 test lint -- --nocapture`
- `cargo +1.93.0 run --quiet -- rewrite "biome check"`
- `cargo +1.93.0 run --quiet -- rewrite "npm exec biome check"`
- `cargo +1.93.0 run --quiet -- rewrite "eslint src --max-warnings=0"`
- `cargo +1.93.0 run --quiet -- rewrite "npm run lint"`
- `git diff --check`

### DIFF
--- a/src/cmds/js/lint_cmd.rs
+++ b/src/cmds/js/lint_cmd.rs
@@ -9,7 +9,8 @@ use crate::mypy_cmd;
 use crate::ruff_cmd;
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use serde_json::Value;
+use std::{collections::HashMap, fs};
 
 #[derive(Debug, Deserialize, Serialize)]
 struct EslintMessage {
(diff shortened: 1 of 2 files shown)


YOMXXX:
Status note: the red `benchmark` check is not caused by this PR's biome-routing change. The failing row is `curl text`: CI compares `curl -s https://httpbin.org/robots.txt` at 8 tokens with `rtk curl https://httpbin.org/robots.txt` at 40 tokens, so the external network fixture can make the raw output smaller than the RTK path and trip the negative-savings gate.

YOMXXX:

YOMXXX:
Ready for review/merge. Rebased onto current develop, conflicts resolved, and the fresh CI run is green, including benchmark and all platform tests. The conflict resolution preserves both biome/eslint lint routing and the current PHP tool normalization path.

PR #2573, opened 2026-06-24
fix(git): keep machine output raw

### LINKED ISSUE #2487: rtk compacts machine-bound git output (--porcelain / --name-only / piped), silently corrupting consumers
## Summary

The Claude Code rewrite hook turns `git X` into `rtk git X` and `rtk git`
compacts the output. That's the right behavior when the **model** reads
the output. But rtk also rewrites git commands whose output is bound for
a **program** — piped into another command, or carrying an explicit
machine-readable format flag — and there its compacted, non-byte-exact
output silently corrupts the consumer.

This is **not** a request to weaken compaction. rtk already draws a
model-vs-machine line (see below); it's just drawing it in the wrong
place for a few cases.

## Evidence that rtk already distinguishes machine output

`git rev-parse` is treated as machine data and passed through unchanged:

```
$ rtk rewrite 'git rev-parse HEAD'              # exit 1 → pass through
$ rtk rewrite 'git rev-parse HEAD | head -1'    # exit 1 → pass through
```

So "don't rewrite this git command" is already a first-class outcome in
the registry. The bug is that the same courtesy isn't extended to other
machine-output cases.

## The mis-drawn cases

rtk rewrites (and thus compacts) commands that explicitly request
parseable output or feed a parser:

```
$ rtk rewrite 'git status --porcelain | wc -l'   → rtk git status --porcelain | wc -l
$ rtk rewrite 'git log --format=%H | sort'       → rtk git log --format=%H | sort
$ rtk rewrite 'git diff | wc -l'                 → rtk git diff | wc -l
```

`--porcelain` is literally git's stable machine-readable contract;
`--format=%H` / `--name-only` / `-z` are the same intent. Compacting
them overrides an explicit "this output is for a program" signal.

Concrete corruption — rtk appends a `--- Changes ---` footer to diff
output, so a file-list count is wrong:

```
$ rtk proxy git diff --name-only HEAD~1..HEAD | grep -c .   # raw git
17
$ rtk git diff --name-only HEAD~1..HEAD | grep -c .         # via rtk
20
$ diff <(rtk proxy git diff --name-only HEAD~1..HEAD) \
       <(rtk git diff --name-only HEAD~1..HEAD)
17a18,20
>
> --- Changes ---
>
```

Secondary, intermittent: the **first** `rtk git diff ...` in a fresh
shell sometimes returns empty output (looks like a cold-cache miss);
repeated identical calls then return the full (footer-decorated) result.
`... | wc -l` → `1` when the truth is 17.

All of these fail **silently** — a plausible wrong number, no error.
Anything scripted that counts/parses git output gets bad data.

## Environment

- rtk 0.36.0, git 2.53.0, macOS (Darwin 25.5.0)
- Hook: `rtk-rewrite.sh` → `rtk rewrite` prepends `rtk ` to git commands

## Proposed fix

Extend rtk's existing pass-through logic (the `rev-parse` exit-1 path)
to decline rewriting when the command signals a machine consumer:

1. an explicit machine-format flag is present —
   `--porcelain`, `--format=`/`--pretty=format:`, `--name-only`,
   `--name-status`, `-z`, `--no-color`; or
2. the git command's stdout is piped into another command or redirected
   into a file (`rtk rewrite` already has the full command string, so a
   trailing `| <cmd>` / `> file` after the git call is detectable).

In those cases, emit the command as raw `git`. Keep compaction fully on
for the common, high-token, model-facing case (a bare `git diff` /
`git status` / `git log` whose output the model reads directly). Token
savings are essentially unchanged; correctness for programmatic
consumers is restored.

Separately, the intermittent empty-output-on-first-call should be fixed
regardless — empty is never a valid compaction of a non-empty diff.

## Workaround (current)

`rtk proxy git ...` to bypass the wrapper, or append `| cat`.

YOMXXX:
Fixes #2487.

## Summary
- Keep git commands raw in hook rewrite when stdout is piped, redirected to a file, or explicitly machine-readable (`--porcelain`, `--format`, `--name-only`, `-z`, etc.).
- Preserve normal model-facing rewrites such as bare `git diff` and later segments in compound commands.
- Make direct `rtk git status --porcelain`, `rtk git log --format=...`, and `rtk git diff --name-only` pass through byte-for-byte instead of compacting.

## Verification
- cargo +1.93.0 test rewrite_git_ -- --nocapture
- cargo +1.93.0 test rewrite_redirect -- --nocapture
- cargo +1.93.0 test registry -- --nocapture
- cargo +1.93.0 test git::git::tests::test_ -- --nocapture
- CLI smoke: rewrite passthrough for `git status --porcelain | wc -l`, `git log --format=%H | sort`, `git diff --name-only HEAD~1..HEAD`; rewrite still active for bare `git diff`
- CLI smoke: `rtk git diff --name-only`, `rtk git status --porcelain`, and `rtk git log --format=%H -3` match native git via `cmp`
- git diff --check

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -67,6 +67,39 @@ fn uses_compact_status_path(args: &[String]) -> bool {
     saw_branch
 }
 
+fn status_args_request_machine_output(args: &[String]) -> bool {
+    args.iter()
+        .any(|arg| arg == "-z" || arg == "--porcelain" || arg.starts_with("--porcelain="))
+}
+
+fn log_args_request_machine_output(args: &[String]) -> bool {
+    args.iter().enumerate().any(|(idx, arg)| {
+        arg == "--format"
+            || arg.starts_with("--format=")
+            || arg.starts_with("--pretty=format:")
+            || arg.starts_with("--pretty=tformat:")
+            || (arg == "--pretty"
+                && args
+                    .get(idx + 1)
+                    .is_some_and(|next| next.starts_with("format:") || next.starts_with("tformat:")))
+    })
+}
+
+fn diff_args_request_machine_output(args: &[String]) -> bool {
+    args.iter().any(|arg| {
+        matches!(
+            arg.as_str(),
+            "-z"
+                | "--no-color"
+                | "--name-only"
+                | "--name-status"
+                | "--numstat"
+                | "--raw"
+                | "--word-porcelain"
+        )
+    })
+}
+
 fn build_status_command(args: &[String], global_args: &[String]) -> Command {
     let mut cmd = git_cmd(global_args);
     cmd.arg("status");
@@ -119,10 +152,12 @@ fn run_diff(
         .iter()
         .any(|arg| arg == "--stat" || arg == "--numstat" || arg == "--shortstat");
 
+    let wants_machine_output = diff_args_request_machine_output(args);
+
     // Check if user wants compact diff (default RTK behavior)
     let wants_compact = !args.iter().any(|arg| arg == "--no-compact");
 
-    if wants_stat || !wants_compact {
+    if wants_machine_output || wants_stat || !wants_compact {
         // User wants stat or explicitly no compacting - pass through directly
         let mut cmd = git_cmd(global_args);
         cmd.arg("diff");
@@ -140,7 +175,11 @@ fn run_diff(
             return Ok(result.exit_code);
         }
 
-        println!("{}", result.stdout.trim());
+        if wants_machine_output {
+            print!("{}", result.stdout);
+        } else {
+            println!("{}", result.stdout.trim());
+        }
 
         timer.track(
             &format!("git diff {}", args.join(" ")),
@@ -424,6 +463,31 @@ fn run_log(
     global_args: &[String],
 ) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
+
+    if log_args_request_machine_output(args) {
+        let mut cmd = git_cmd(global_args);
+        cmd.arg("log");
+        for arg in args {
+            cmd.arg(arg);
+        }
+
+        let result = exec_capture(&mut cmd).context("Failed to run git log")?;
+        if !result.success() {
+            eprintln!("{}", result.stderr);
+            return Ok(result.exit_code);
+        }
+
+        print!("{}", result.stdout);
+
+        timer.track(
+            &format!("git log {}", args.join(" ")),
+            &format!("rtk git log {} (passthrough)", args.join(" ")),
+            &result.stdout,
+            &result.stdout,
+        );
+
+        return Ok(0);
+    }
 
     let mut cmd = git_cmd(global_args);
     cmd.arg("log");
@@ -814,6 +878,40 @@ fn filter_status_with_args(output: &str) -> String {
 
 fn run_status(args: &[String], verbose: u8, global_args: &[String]) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
+
+    if status_args_request_machine_output(args) {
+        let mut cmd = git_cmd(global_args);
+        cmd.arg("status");
+        cmd.args(args);
+        let result = exec_capture(&mut cmd).context("Failed to run git status")?;
+
+        if !result.success() {
+            if !result.stderr.trim().is_empty() {
+                eprint!("{}", result.stderr);
+            }
+            timer.track(
+                &format!("git status {}", args.join(" ")),
+                &format!("rtk git status {} (passthrough)", args.join(" ")),
+                &result.stdout,
+                &result.stdout,
+            );
+            return Ok(result.exit_code);
+        }
+
+        if verbose > 0 || !result.stderr.is_empty() {
+            eprint!("{}", result.stderr);
+        }
+        print!("{}", result.stdout);
+
+        timer.track(
+            &format!("git status {}", args.join(" ")),
+            &format!("rtk git status {} (passthrough)", args.join(" ")),
+            &result.stdout,
+            &result.stdout,
+        );
+
+        return Ok(0);
+    }
 
     // Keep a narrow compact path for no-arg status and branch/short-only flags.
     // More complex explicit args still use the existing minimal-filter path.
@@ -1850,6 +1948,44 @@ mod tests {
             })
             .collect();
         assert!(envs.contains(&("LC_ALL".to_string(), "C".to_string())));
+    }
+
+    #[test]
+    fn test_git_status_machine_output_args_passthrough() {
+        assert!(status_args_request_machine_output(&["--porcelain".to_string()]));
+        assert!(status_args_request_machine_output(&[
+            "--porcelain=v2".to_string()
+        ]));
+        assert!(status_args_request_machine_output(&["-z".to_string()]));
+        assert!(!status_args_request_machine_output(&["--short".to_string()]));
+    }
+
+    #[test]
+    fn test_git_log_machine_output_args_passthrough() {
+        assert!(log_args_request_machine_output(&["--format=%H".to_string()]));
+        assert!(log_args_request_machine_output(&[
+            "--pretty=format:%H".to_string()
+        ]));
+        assert!(log_args_request_machine_output(&[
+            "--format".to_string(),
+            "%H".to_string()
+        ]));
+        assert!(!log_args_request_machine_output(&["--oneline".to_string()]));
+    }
+
+    #[test]
+    fn test_git_diff_machine_output_args_passthrough() {
+        for args in [
+            vec!["--name-only".to_string()],
+            vec!["--name-status".to_string()],
+            vec!["--numstat".to_string()],
+            vec!["--raw".to_string()],
+            vec!["--word-porcelain".to_string()],
+            vec!["-z".to_string(), "--name-only".to_string()],
+        ] {
+            assert!(diff_args_request_machine_output(&args), "{args:?}");
+        }
+        assert!(!diff_args_request_machine_output(&[]));
     }
 
     #[test]
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -565,7 +565,7 @@ fn rewrite_compound(
                 let rewritten = if is_pipe_incompatible {
                     seg.to_string()
                 } else {
-                    rewrite_segment(seg, excluded, transparent_prefixes)
+                    rewrite_segment_for_consumer(seg, excluded, transparent_prefixes, true)
                         .unwrap_or_else(|| seg.to_string())
                 };
                 if rewritten != seg {
@@ -719,7 +719,16 @@ fn rewrite_segment(
     excluded: &[ExcludePattern],
     transparent_prefixes: &[String],
 ) -> Option<String> {
-    rewrite_segment_inner(seg, excluded, transparent_prefixes, 0)
+    rewrite_segment_for_consumer(seg, excluded, transparent_prefixes, false)
+}
+
+fn rewrite_segment_for_consumer(
+    seg: &str,
+    excluded: &[ExcludePattern],
+    transparent_prefixes: &[String],
+    stdout_is_piped: bool,
+) -> Option<String> {
+    rewrite_segment_inner(seg, excluded, transparent_prefixes, 0, stdout_is_piped)
 }
 
 fn is_excluded(cmd: &str, excluded: &[ExcludePattern]) -> bool {
@@ -734,6 +743,7 @@ fn rewrite_segment_inner(
     excluded: &[ExcludePattern],
     transparent_prefixes: &[String],
     depth: usize,
+    stdout_is_piped: bool,
 ) -> Option<String> {
     let trimmed = seg.trim();
     if trimmed.is_empty() {
@@ -755,8 +765,13 @@ fn rewrite_segment_inner(
             );
             return None;
         }
-        let rewritten =
-            rewrite_segment_inner(rest_after_env, excluded, transparent_prefixes, depth + 1)?;
+        let rewritten = rewrite_segment_inner(
+            rest_after_env,
+            excluded,
+            transparent_prefixes,
+            depth + 1,
+            stdout_is_piped,
+        )?;
         return Some(format!("{}{}", env_prefix, rewritten));
     }
 
@@ -765,8 +780,14 @@ fn rewrite_segment_inner(
             if rest.is_empty() {
                 return None;
             }
-            return rewrite_segment_inner(rest, excluded, transparent_prefixes, depth + 1)
-                .map(|rewritten| format!("{} {}", prefix, rewritten));
+            return rewrite_segment_inner(
+                rest,
+                excluded,
+                transparent_prefixes,
+                depth + 1,
+                stdout_is_piped,
+            )
+            .map(|rewritten| format!("{} {}", prefix, rewritten));
         }
     }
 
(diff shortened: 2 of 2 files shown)


src/discover/registry.rs
```diff
@@ -860,6 +891,85 @@ fn rewrite_segment_inner(
...
+        let token = tokens[i].0;
+        match token {
+            "-C" | "-c" | "--git-dir" | "--work-tree" => i += 2,
+            "--no-pager" | "--no-optional-locks" | "--bare" | "--literal-pathspecs" => i += 1,
+            _ if token.starts_with("--git-dir=") || token.starts_with("--work-tree=") => i += 1,
+            _ => break,
+        }
+    }
+
+    Some(tokens[i..].iter().map(|token| token.0).collect())
```
KuSh, line 926: **[high] Slice-index panic when a git global option is the last token.**

`i += 2` for `-C | -c | --git-dir | --work-tree` is unbounded, so when one of those is the final token `i` becomes `len + 1` and `tokens[i..]` panics.

Repro on this branch:

```
rewrite_command("git -C")
// panicked at src/discover/registry.rs:926:
// range start index 3 out of range for slice of length 2
```

Reachable from any command string the hook or `rtk discover` processes — `git -c`, `git --git-dir`, or a truncated/typo'd entry in shell history — and it takes down the whole `rtk discover` run.

Fix: `i = i.saturating_add(2).min(tokens.len())`, or return `None` when the value token is missing.

src/cmds/git/git.rs
```diff
@@ -67,6 +67,39 @@ fn uses_compact_status_path(args: &[String]) -> bool {
...
+        arg == "--format"
+            || arg.starts_with("--format=")
+            || arg.starts_with("--pretty=format:")
+            || arg.starts_with("--pretty=tformat:")
+            || (arg == "--pretty"
+                && args
+                    .get(idx + 1)
+                    .is_some_and(|next| next.starts_with("format:") || next.starts_with("tformat:")))
+    })
+}
```
KuSh, line 86: **[high] `--pretty=%…` isn't detected, so the bug from #2487 still reproduces.**

Git treats `--pretty=<string containing a % placeholder>` as an implicit `--pretty=tformat:<string>`, but this only matches `--format`, `--format=`, `--pretty=format:` and `--pretty=tformat:`.

Verified against a build of this branch, in this repo:

```
rtk git log --pretty=%H   →     50 commits
git log --pretty=%H       →   1185 commits
```

The RTK `-50` default is still injected and the output is silently truncated — exactly the class of bug this PR sets out to fix. Needs a "value contains `%`" rule for `--pretty=` (and for the separated `--pretty <value>` form below).

src/discover/registry.rs
```diff
@@ -860,6 +891,85 @@ fn rewrite_segment_inner(
...
+                | "--word-porcelain"
+        ) || arg.starts_with("--porcelain=")
+            || arg.starts_with("--format=")
+            || *arg == "--format"
+            || arg.starts_with("--pretty=format:")
+            || arg.starts_with("--pretty=tformat:")
+            || (*arg == "--pretty"
+                && args.get(idx + 1).is_some_and(|next| {
+                    next.starts_with("format:") || next.starts_with("tformat:")
+                }))
```
KuSh, line 959: **[high] Same `--pretty=%…` gap in the hook rewriter.**

Verified: `rewrite_command("git log --pretty=%H")` → `Some("rtk git log --pretty=%H")`, so an unpiped, unredirected `git log --pretty=%H` still gets routed into the truncating filter.

Both detectors need the `%` rule, and they need to stay in sync with `log_args_request_machine_output` in `src/cmds/git/git.rs`.

src/cmds/git/git.rs
```diff
@@ -67,6 +67,39 @@ fn uses_compact_status_path(args: &[String]) -> bool {
...
+
+fn diff_args_request_machine_output(args: &[String]) -> bool {
+    args.iter().any(|arg| {
+        matches!(
+            arg.as_str(),
+            "-z"
+                | "--no-color"
```
KuSh, line 93: **[medium] `--no-color` is not a machine-output signal — this kills diff compaction.**

`--no-color` says nothing about machine readability; it's precisely what an agent passes to avoid ANSI in its capture. Classifying it as machine output means any agent that habitually passes it gets 0% savings on git.

Verified on this branch: `rtk git diff --no-color` emitted 2602 bytes — byte-identical to raw `git diff --no-color` — where develop emits the compacted stat+diff form.

The same entry at `src/discover/registry.rs:944` makes the rewriter return `None` for `git diff --no-color` and `git log --no-color -10`, so the hook stops rewriting them at all.

Please drop `--no-color` from both lists.

src/discover/registry.rs
```diff
@@ -860,6 +891,85 @@ fn rewrite_segment_inner(
...
+    args.iter().enumerate().any(|(idx, arg)| {
+        matches!(
+            *arg,
+            "-z" | "--no-color"
+                | "--porcelain"
+                | "--name-only"
+                | "--name-status"
+                | "--numstat"
+                | "--raw"
+                | "--word-porcelain"
```
KuSh, line 950: **[medium] `--word-porcelain` is not a git flag; the real machine mode is missed.**

`git diff --word-porcelain` just prints git's usage error, so this arm is dead code. The actual machine-readable word diff is `--word-diff=porcelain`, which neither this function nor `diff_args_request_machine_output` (`src/cmds/git/git.rs:98`) matches — so the real machine format still gets compacted and rewritten.

Replace the dead entry with a `--word-diff=porcelain` check, in both places.

src/discover/registry.rs
```diff
@@ -565,7 +565,7 @@ fn rewrite_compound(
                 let rewritten = if is_pipe_incompatible {
                     seg.to_string()
                 } else {
-                    rewrite_segment(seg, excluded, transparent_prefixes)
+                    rewrite_segment_for_consumer(seg, excluded, transparent_prefixes, true)
```
KuSh, line 568: **[medium] `stdout_is_piped: true` makes the whole pipeline unrewritable, dropping downstream rewrites develop already performs.**

When the git segment returns `None` here, `any_changed` stays false and — with no following operator — `rewrite_compound` returns `None` for the entire command. The PR's own test changes encode that (`git log -10 | grep feat` → `None`).

But develop (rtk 0.46.0) already keeps piped git raw *and* still rewrites the consumer:

```
git log -10 | grep feat   →   git log -10 | rtk grep feat
git status | grep foo     →   git status | rtk grep foo
```

Merging as-is regresses both.

Also worth flagging for the rebase: this branch is based on `ee9e2f8`, and develop has since reworked pipe handling (`TokenKind::Pipe(PipeKind)`, `|&`) and already returns `None` for cases like `git status --porcelain | wc -l`. A good part of this hunk should be redundant once rebased — please re-derive what's actually still needed on top of the new pipe handling instead of porting it verbatim.

src/cmds/git/git.rs
```diff
@@ -425,6 +464,31 @@ fn run_log(
...
+
+        timer.track(
+            &format!("git log {}", args.join(" ")),
+            &format!("rtk git log {} (passthrough)", args.join(" ")),
+            &result.stdout,
+            &result.stdout,
+        );
+
+        return Ok(0);
+    }
```
KuSh, line 490: **[medium] The passthrough drops the implicit commit cap, so output becomes unbounded.**

This branch runs `git log <args>` with no `-N` and no `--no-merges`, unlike the path below it which injects `-50` / `-10`.

Verified: `rtk git log --format=%H` (no explicit limit) now prints the full 1185-commit history in this repo; on develop it printed 50. A single such call can flood the agent's context — the thing RTK exists to prevent.

Suggest keeping a large default cap when the user gave no `-n` / `--max-count`, or at minimum documenting the unbounded behaviour.

src/discover/registry.rs
```diff
@@ -860,6 +891,85 @@ fn rewrite_segment_inner(
...
+        return false;
+    }
+
+    stdout_is_piped
+        || git_command_requests_machine_output(cmd_part)
+        || redirect_suffix_redirects_stdout(redirect_suffix)
+}
+
+fn git_command_tail_tokens(cmd: &str) -> Option<Vec<&str>> {
+    let tokens = split_token_spans(cmd);
```
KuSh, line 909: **[low] `split_token_spans` is quote-blind, so flag words inside quoted arguments suppress rewriting.**

It splits on whitespace only, ignoring the quoting `tokenize()` already handles.

Verified:

```
rewrite_command("git commit -m \"add --no-color to diff\"")   →   None
rewrite_command("git log -- --name-only")                    →   None
```

Both silently lose the rewrite (and the savings) because a flag-looking word appears inside a quoted message or after `--`. Using `tokenize()` and stopping the scan at `--` handles both.

KuSh (changes requested):
Thanks for tackling #2487 — the shape of the fix (detect a machine-output request, pass the command through raw) is the right one, and the status / `--porcelain` path looks good.

I built this branch and probed the resulting binary. A few things block a merge:

- **A panic.** `git -C` (or `-c` / `--git-dir` / `--work-tree`) as the final token indexes past the token slice and aborts the whole `rtk discover` run.
- **The original bug still reproduces.** `--pretty=%H` is git's implicit `tformat:` but isn't detected, so `rtk git log --pretty=%H` still truncates 1185 commits to 50 — in both the filter and the rewriter.
- **Two false positives** in the machine-output lists: `--no-color` (disables diff compaction for any agent that passes it) and `--word-porcelain` (not a real git flag — the real one, `--word-diff=porcelain`, is missed).
- **A regression against develop.** Making a piped git segment unrewritable also drops the downstream rewrite develop already performs (`git log -10 | grep feat` → `git log -10 | rtk grep feat`).
- Plus an unbounded `git log` passthrough and a quote-blind token scan.

One process note: the branch is based on `ee9e2f8` and is currently conflicting. Develop has since reworked pipe handling (`TokenKind::Pipe(PipeKind)`, `|&`) and already returns `None` for cases like `git status --porcelain | wc -l` — so a good part of the pipe hunk should be redundant once you rebase. Worth re-deriving what's still needed on top of the new pipe handling rather than porting it verbatim.

Details inline.

src/cmds/git/git.rs
KuSh, line 86: **Update after [kylehgc/rtk@da407dc](https://github.com/kylehgc/rtk/commit/da407dc).**

The amendment adds `-z` here, which is a genuine catch I missed: `git log -z --name-only` carries no custom format, so it was falling to the compact path and getting `--pretty=format:<MARKER>`, `-10` and `--no-merges` injected — exactly the corruption this guard exists to prevent.

But `--pretty=%H` is still unmatched, so #2487's headline case is unchanged: `rtk git log --pretty=%H` returns 50 commits where git returns 1185. Git expands `--pretty=<string containing %>` to `--pretty=tformat:<string>`; this predicate needs that rule, and so does the separated `--pretty <value>` form just below.

This is the case the issue is filed about — it should land before merge.

src/cmds/git/git.rs
KuSh, line 490: **Update after [kylehgc/rtk@da407dc](https://github.com/kylehgc/rtk/commit/da407dc).**

The amendment doesn't change this branch, but it widens what reaches it. Adding `-z` to `log_args_request_machine_output` means `git log -z --name-only` now lands here uncapped, where it previously took the compact path with `-10`. On this repo that turns a 10-commit summary into the full 1185-commit history, with per-commit file lists.

Both changes are individually right; together they make the missing default cap considerably more reachable. A large default `-n` when the user gave neither `-n` nor `--max-count` keeps the `-z` fix without the flood.

src/discover/registry.rs
```diff
@@ -860,6 +891,85 @@ fn rewrite_segment_inner(
...
+    }
+
+    git_args_request_machine_output(&tokens[1..])
+}
+
+fn git_args_request_machine_output(args: &[&str]) -> bool {
+    args.iter().enumerate().any(|(idx, arg)| {
+        matches!(
+            *arg,
+            "-z" | "--no-color"
```
KuSh, line 944: **Update after [kylehgc/rtk@da407dc](https://github.com/kylehgc/rtk/commit/da407dc).**

The amendment removes `--no-color` from `diff_args_request_machine_output` in `src/cmds/git/git.rs`, but not from this list. That leaves the fix inert on the path that actually matters: with `--no-color` still here, `rewrite_command("git diff --no-color")` returns `None`, the hook never rewrites the command, raw git runs, and the compaction restored in git.rs is never reached.

The entry has to come out of both lists — ideally with the two predicates sharing one source so they can't drift apart again.

KuSh (commented):
Followed up on the fork-side amendment [kylehgc/rtk@da407dc](https://github.com/kylehgc/rtk/commit/da407dc) proposed on top of this PR. It picks up two real defects that didn't come up in my review at all, but the highest-severity items above are untouched, so the request for changes still stands.

**Caught by the amendment, not by my review**

- **`-z` was unmatched for `git log`.** `git log -z --name-only` carries no custom format, so it fell through to the compact path and had `--pretty=format:<MARKER>`, `-10` and `--no-merges` injected — corruption of a machine-readable format, exactly what this guard exists to prevent. Correctly fixed. It does widen an existing problem, though; see the inline note on `run_log`.
- **The new passthrough branches were recording 0%-savings rows.** `timer.track(.., &stdout, &stdout)` logs equal non-zero token counts on every passthrough, which dilutes `rtk gain`. `track_passthrough` (`src/core/tracking.rs:1428`) records `input_tokens=0, output_tokens=0` for precisely this case — right call, and consistent with `run_passthrough` in the same file.

**The `--no-color` and `--word-porcelain` corrections only land on half the path.** Both are fixed in `src/cmds/git/git.rs` but not in `src/discover/registry.rs`. For `--word-diff=porcelain` the end-to-end result still comes out right (the rewriter routes it to rtk, git.rs then passes it through raw), but the two lists should come from one source rather than drifting. For `--no-color` the fix is inert as written — details inline on `registry.rs`.

**Still open:** the panic on a trailing `git -C`; `--pretty=%H` in both the filter and the rewriter (#2487's headline case); the pipe regression; the uncapped log passthrough; the quote-blind token scan.

---

**New findings, in the amendment's own `tests/git_machine_output_test.rs`.** That file isn't part of this PR, so these can't be pinned inline:

**1. [medium] `human_facing_git_output_is_still_compacted` is flaky.** It asserts `rtk git diff --no-color HEAD~1..HEAD` is strictly smaller than native — which depends entirely on what the tip commit happens to be, since compaction prepends a stat header that outweighs the savings on small diffs. Measured across the last 20 non-merge commits on develop, 2 invert:

```
6370e79   native 2148 bytes   rtk 2171 bytes   ← test would fail
e5cecd3   native 2206 bytes   rtk 2208 bytes   ← test would fail
```

That's a ~10% chance of a red build for reasons unrelated to the change. Assert against a fixed synthetic diff or a pinned commit range instead of `HEAD~1`.

**2. [medium] `run()` ignores exit status, so the byte-equality test passes vacuously in CI.** It returns `Some(stdout)` for anything that spawns. The `test` job checks out at the default `fetch-depth: 1` (`.github/workflows/ci.yml:133`, `cargo test --all` at `:136`), so `HEAD~1..HEAD` is an unknown revision there: native git returns empty stdout, `let Some(native) = … else { continue }` does *not* skip (it's `Some(vec![])`), rtk returns empty too, and `assert_eq!("", "")` passes. All four `diff … HEAD~1..HEAD` cases contribute zero coverage in CI, silently — and the `--no-color` counterweight is skipped outright by its `is_empty()` guard. Check `out.status.success()` and return `None` on failure so those cases genuinely skip, and ideally build the fixture history with `git init` in a temp dir instead of depending on the checkout's history.

**3. [low] The one case #2487 is actually about stays uncovered.** The log cases are `--format=%H` and `--pretty=format:%H`; neither exercises `--pretty=%H`, the spelling that still truncates. That's the regression test for the original issue — worth adding once the predicate is fixed.

**4. [nit] The test writes to the developer's real tracking DB.** Each run invokes the binary ~13 times with no `RTK_DB_PATH` override, so rows land in `~/.local/share/rtk/history.db` (verified: two invocations moved it 58942 → 58944 rows). This is pre-existing across the suite — `tests/guard_integration_test.rs` and friends do the same — so not something the amendment introduced. But given the commit's stated motivation is keeping `rtk gain` undiluted, an `.env("RTK_DB_PATH", tmp)` would match that intent.

PR #2574, opened 2026-06-24
fix(hook): support copilot ide terminal tool

### LINKED ISSUE #2443: rtk hook copilot no-op in JetBrains/IntelliJ Copilot plugin — camelCase `run_in_terminal` not matched, and IDE host only honors `permissionDecision: deny`
## Summary

`rtk hook copilot` is a silent no-op inside the **JetBrains / IntelliJ GitHub Copilot plugin** (CLion, IDEA, Rider, etc., agent mode). The IDE's terminal tool is named **`run_in_terminal`** and is delivered in the **camelCase** payload shape (`toolName` + `toolArgs`), which `detect_format()` only matches when `toolName == "bash"`. As a result the hook fires but never rewrites anything, and `rtk gain` stays empty.

This is **distinct from #1425 / PR #1800**: those address the *snake_case* (`tool_name`) VS Code Chat path. PR #1800's new `CamelCaseCmd` branch still only matches `toolName == "bash"`, so it does **not** fix the JetBrains IDE case.

There is also a second, host-side blocker (see "Host limitation" below) that means transparent `modifiedArgs` rewriting cannot work in this plugin at all today.

## Environment

- rtk 0.42.2
- GitHub Copilot plugin for JetBrains (`github-copilot-intellij`, core `1.11.0-rc.x`), agent mode, on CLion 2026.1, Linux
- Hook installed at both `~/.copilot/hooks/rtk-rewrite.json` and project `.github/hooks/rtk-rewrite.json`, event `preToolUse`

## Hard evidence — the real payload

Captured by tee-ing the hook's stdin to a file. A terminal command from the agent produces:

```json
{
  "timestamp": 1781456705251,
  "cwd": ".",
  "toolName": "run_in_terminal",
  "toolArgs": "{\"command\":\" cd /path/to/repo && git --no-pager log --oneline -20\",\"explanation\":\"Show local git history\",\"isBackground\":false}"
}
```

Notes:
- `toolName` is **camelCase** and the value is **`run_in_terminal`**, not `bash`.
- `toolArgs` is a JSON-**encoded string** containing `command` (plus `explanation`, `isBackground` which must be preserved).
- The hook also receives MCP tool calls here, e.g. `"toolName": "io.github.github/github-mcp-server/list_commits"` — those should keep passing through.

Reproduce directly against the binary:

```
$ echo '{"toolName":"run_in_terminal","toolArgs":"{\"command\":\"git --no-pager log --oneline -20\"}"}' | rtk hook copilot
            # (no output)

$ echo '{"toolName":"bash","tool_input":{"command":"git --no-pager log --oneline -20"}}' | rtk hook copilot
            # also nothing for the snake_case tool_input shape

$ echo '{"tool_name":"bash","tool_input":{"command":"git --no-pager log --oneline -20"}}' | rtk hook copilot
{"hookSpecificOutput":{...,"updatedInput":{"command":"rtk git --no-pager log --oneline -20"}}}   # only "bash" works
```

## Root cause (code)

`src/hooks/hook_cmd.rs`, `detect_format()` — camelCase branch:

```rust
    if tool_name == "bash" {          // <-- JetBrains sends "run_in_terminal"
        ...
    }
    return HookFormat::PassThrough;
}
```

`run_in_terminal` is missing from the camelCase match. (PR #1800 keeps the same `tool_name == "bash"` condition in `CamelCaseCmd`, so it remains unfixed there.)

## Host limitation — `modifiedArgs` is ignored in the JetBrains IDE

Even after recognizing `run_in_terminal`, transparent rewrite won't take effect in this plugin. The IDE agent host (`copilot-agent/dist/main.js`) invokes the hook and then inspects **only** `permissionDecision`:

```js
let o = await ...hook(t, P3, {timestamp, cwd, toolName, toolArgs}, e);
for (let s of o) {
  let c = s.output;
  if (c && c.permissionDecision === deny)
    return { permissionDecision: "deny", permissionDecisionReason: ... };
}
```

`modifiedArgs` and `hookSpecificOutput.updatedInput` are never read (0 occurrences in `main.js`; they exist only in the bundled `@github/copilot` SDK, which this agent path does not use). So in the JetBrains IDE the **only** effective hook action is `permissionDecision: "deny"` (with a reason).

This means a complete fix needs a **deny-with-suggestion** path for the IDE: deny the raw command and surface the rtk-rewritten command in `permissionDecisionReason`, letting the model re-issue `rtk …` (which then passes through). rtk already advertises "deny-with-suggestion" for Copilot CLI, so the mechanism exists.

## Suggested fix

1. Recognize `run_in_terminal` (camelCase `toolName`) in `detect_format()` in addition to `bash`.
2. For the JetBrains/IDE case (where `modifiedArgs` is ignored), respond with `permissionDecision: "deny"` + a `permissionDecisionReason` containing the rewritten command, instead of (or in addition to) `modifiedArgs`.
3. Optionally cover other IDE tool names defensively. Note that `get_terminal_output` (the "read terminal output" tool) carries **no command**, so it correctly needs no rewrite — only `run_in_terminal` does.

## Tool inventory (JetBrains Copilot built-ins, for reference)

`run_in_terminal`, `get_terminal_output`, `read_file`, `create_file`, `insert_edit_into_file`, `replace_string_in_file`, `list_dir`, `file_search`, `grep_search`, `semantic_search`, `run_subagent`, `open_file`, `get_errors`, `ask_questions`, `validate_cves`. Only `run_in_terminal` runs a shell command.

Related: #1425, #1839, #1800.

YOMXXX:
## Summary
- recognize JetBrains Copilot `toolName: "run_in_terminal"` camelCase payloads
- emit top-level `permissionDecision: "deny"` with an RTK rewrite suggestion for IDE terminal commands
- keep Copilot CLI `toolName: "bash"` on the existing transparent `modifiedArgs` path, and leave non-terminal IDE tools silent

Fixes #2443

## Verification
- `cargo +1.93.0 fmt --check`
- `git diff --check`
- `cargo +1.93.0 test hook_cmd -- --nocapture`
- CLI smoke: JetBrains `run_in_terminal` payload returns top-level deny with `rtk git status` suggestion
- CLI smoke: `get_terminal_output` payload stays silent
- CLI smoke: camelCase `bash` payload still returns `modifiedArgs`

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -35,6 +35,10 @@ enum HookFormat {
     /// Carries the full parsed `toolArgs` object so we can rewrite `command` while preserving
     /// host-supplied metadata (description, initial_wait, mode, …) the tool requires.
     CopilotCli { command: String, args: Value },
+    /// JetBrains Copilot IDE terminal tool: camelCase `toolName=run_in_terminal`
+    /// with JSON-string `toolArgs`. The host ignores transparent rewrites, so it
+    /// needs a top-level deny-with-suggestion response.
+    CopilotIde { command: String },
     /// Non-bash tool, already uses rtk, or unknown format — pass through silently.
     PassThrough,
 }
@@ -62,8 +66,19 @@ pub fn run_copilot() -> Result<()> {
     match detect_format(&v) {
         HookFormat::VsCode { command } => handle_vscode(&command),
         HookFormat::CopilotCli { command, args } => handle_copilot_cli(&command, &args),
+        HookFormat::CopilotIde { command } => handle_copilot_ide(&command),
         HookFormat::PassThrough => Ok(()),
     }
+}
+
+fn parse_camel_case_tool_args(v: &Value) -> Option<(String, Value)> {
+    let tool_args_str = v.get("toolArgs").and_then(|t| t.as_str())?;
+    let tool_args = serde_json::from_str::<Value>(tool_args_str).ok()?;
+    let cmd = tool_args
+        .get("command")
+        .and_then(|c| c.as_str())
+        .filter(|c| !c.is_empty())?;
+    Some((cmd.to_string(), tool_args))
 }
 
 fn detect_format(v: &Value) -> HookFormat {
@@ -86,19 +101,13 @@ fn detect_format(v: &Value) -> HookFormat {
     // Copilot CLI: camelCase keys, toolArgs is a JSON-encoded string
     if let Some(tool_name) = v.get("toolName").and_then(|t| t.as_str()) {
         if tool_name == "bash" {
-            if let Some(tool_args_str) = v.get("toolArgs").and_then(|t| t.as_str()) {
-                if let Ok(tool_args) = serde_json::from_str::<Value>(tool_args_str) {
-                    if let Some(cmd) = tool_args
-                        .get("command")
-                        .and_then(|c| c.as_str())
-                        .filter(|c| !c.is_empty())
-                    {
-                        return HookFormat::CopilotCli {
-                            command: cmd.to_string(),
-                            args: tool_args,
-                        };
-                    }
-                }
+            if let Some((command, args)) = parse_camel_case_tool_args(v) {
+                return HookFormat::CopilotCli { command, args };
+            }
+        }
+        if tool_name == "run_in_terminal" {
+            if let Some((command, _args)) = parse_camel_case_tool_args(v) {
+                return HookFormat::CopilotIde { command };
             }
         }
         return HookFormat::PassThrough;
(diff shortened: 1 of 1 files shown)


KuSh:
Thank you for working on this. The same fix landed on develop in #3093, merged 2026-07-26 (commit 1d6798de8a), which adds a dedicated CopilotIde format for the camelCase run_in_terminal payload, returns a top-level deny with the rtk rewrite as the reason because the IDE host ignores modifiedArgs, keeps Copilot CLI's transparent bash path, and stays silent on deferred or non-terminal tools. The code on develop matches this PR closely, including the deny reason text, and it also covers the powershell tool name. Closing as already covered, and please comment or reopen if the JetBrains plugin still no-ops for you on a current build.

PR #1483, opened 2026-04-24
fix(config): parse partial sub-config sections

Jitsusama:
## Summary

Fixes #1482.

- Five sub-config structs (`TrackingConfig`, `DisplayConfig`, `FilterConfig`, `TelemetryConfig` and `LimitsConfig`) don't have `#[serde(default)]`, so any partial TOML section fails to parse with `missing field`. `Config::load` returns the error, helpers like `config::limits` call `.unwrap_or_default()`, and the user's explicit overrides are silently replaced by the full default config.
- Add struct-level `#[serde(default)]` to each of the five structs. Every one of them already implements `Default` with the correct domain values, so the attribute delegates to what's already there. No logic changes, no duplicated defaults, backward compatible with fully specified configs.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test`: 1684 passed, 0 failed.
- [x] Six new tests in `src/core/config.rs::tests`:
  - `test_partial_tracking_section_uses_defaults`
  - `test_partial_display_section_uses_defaults`
  - `test_partial_filters_section_uses_defaults`
  - `test_partial_telemetry_section_uses_defaults`
  - `test_partial_limits_section_uses_defaults`
  - `test_full_limits_section_roundtrip` (regression guard for fully specified sections)

## Files Changed

| File | Change |
|------|--------|
| `src/core/config.rs` | Add struct-level `#[serde(default)]` to five sub-config structs; add six parsing tests. |

## Related

- Previously reported in #1134, which was closed as user confusion about `[telemetry]` vs `[tracking]`. The reporter's second example (partial config) was the real bug but never got addressed.
- Previously attempted in #1135, which the author self-closed for branching off `master` instead of `develop`. This PR is a clean resubmission against `develop`.

---
*Co-Authored-By AI (Claude Opus 4.7) via [Pi](https://github.com/badlogic/pi-mono)*

PR #991, opened 2026-04-03
fix(git): preserve full status paths and untracked files

em0t:
## Summary
- switch default `rtk git status` to `--porcelain -b -uall` so dirty paths are no longer hidden behind grouped summaries, overflow markers, or directory-level collapse
- preserve RTK's status framing (`* branch`, `clean — nothing to commit`) while keeping every dirty path visible
- keep branch-oriented explicit status flags (`-b`, `--branch`, `-sb`, `--short --branch`) on the same compact path so `rtk git status -b` still gains tokens without hiding files
- add focused regression coverage for default status behavior, branch-flag routing, rename/conflict preservation, and untracked directory expansion

### `rtk git status` Hidden Dirty Paths
Raw
```text
On branch main
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   src/file01.rs
	modified:   src/file02.rs
	modified:   src/file03.rs
	modified:   src/file04.rs
	modified:   src/file05.rs
	modified:   src/file06.rs
	modified:   src/file07.rs
	modified:   src/file08.rs
	modified:   src/file09.rs
	modified:   src/file10.rs
	modified:   src/file11.rs
	modified:   src/file12.rs
	modified:   src/file13.rs
	modified:   src/file14.rs
	modified:   src/file15.rs
	modified:   src/file16.rs
	modified:   src/file17.rs

no changes added to commit (use "git add" and/or "git commit -a")
```

Previous RTK (`develop`)
```text
* main
~ Modified: 17 files
   src/file01.rs
   src/file02.rs
   src/file03.rs
   src/file04.rs
   src/file05.rs
   src/file06.rs
   src/file07.rs
   src/file08.rs
   src/file09.rs
   src/file10.rs
   src/file11.rs
   src/file12.rs
   src/file13.rs
   src/file14.rs
   src/file15.rs
   ... +2 more
```

This PR
```text
* main
 M src/file01.rs
 M src/file02.rs
 M src/file03.rs
 M src/file04.rs
 M src/file05.rs
 M src/file06.rs
 M src/file07.rs
 M src/file08.rs
 M src/file09.rs
 M src/file10.rs
 M src/file11.rs
 M src/file12.rs
 M src/file13.rs
 M src/file14.rs
 M src/file15.rs
 M src/file16.rs
 M src/file17.rs
```

### Untracked Directory Expansion
Raw
```text
On branch main
Untracked files:
[...]

PR #2579, opened 2026-06-24
fix(git): wire up status_max_files and status_max_untracked config limits

### LINKED ISSUE #2437: [config] status_max_files and status_max_untracked are declared in LimitsConfig but never read
## Problem

Two fields in `LimitsConfig` (`src/core/config.rs:123-145`) are declared with serde defaults and are user-configurable in `config.toml`, but **no code reads them**. The git status command never bounds its output by them.

## Repro

1. Set values in `~/AppData/Roaming/rtk/config.toml` (Windows) or `~/.config/rtk/config.toml` (Unix):
   ```toml
   [limits]
   status_max_files = 2
   status_max_untracked = 1
   ```
2. In a repo with > 2 modified files and > 1 untracked file, run `rtk git status`.
3. Observe: the user's caps are ignored; status shows all files.

## Evidence

`src/cmds/git/git.rs` contains the status formatting pipeline (`format_status_output`, `format_status_inner`, `run_status`) but **zero references** to `config::limits()` or to either field name. Verified by:

```
$ rg "limits\(\)|status_max" src/cmds/git/git.rs
(no matches)
$ rg "status_max_files|status_max_untracked" src/
src/core/README.md:90:status_max_files = 15
src/core/README.md:91:status_max_untracked = 10
src/core/config.rs:129:    pub status_max_files: usize,
src/core/config.rs:131:    pub status_max_untracked: usize,
src/core/config.rs:141:        status_max_files: 15,
src/core/config.rs:142:        status_max_untracked: 10,
```

The only references outside `config.rs` are in `src/core/README.md` documenting the config schema.

## Suspected cause

The status command in `git.rs:673` (`format_status_output`) was likely refactored to use a fixed-path porcelain parser that iterates the entire `porcelain.lines()` vec. The previous version probably bounded displayed rows; the refactor dropped the cap and the `LimitsConfig` fields were left orphaned.

## Two options

**Option A (preferred — wire them up):** Re-add the cap in `format_status_inner`. The cap is non-trivial because the porcelain format groups by status (staged/modified/untracked) and the cap semantics from the docstring ("15 staged/modified, 10 untracked") need to apply per-bucket, not globally.

**Option B (cleanup):** Remove the two fields, their defaults, and the docs. If git status should not be configurable, it shouldn't expose the knob — users who set it will be confused by the silent no-op.

A third option: add a deprecation warning at config load time if the user has set these keys, and remove in a later major.

## Environment

- rtk 0.42.4
- Confirmed against `~/rtk` checkout, commit 878af7d on develop.

## Related

- Issue #828: "docs: FEATURES.md missing 30+ truncation limits" — adjacent but distinct. #828 is about hidden hardcoded limits not being documented; this is about documented user-configurable limits that don't work.

guyoron1:
## Summary

`LimitsConfig` fields `status_max_files` and `status_max_untracked` were parsed from config but never actually read by the git status formatter — making them dead knobs.

Wire them up so the configured limits are respected in the git status output.

Fixes #2437

## Test plan

- [ ] Manual testing: set `status_max_files = 5` in config, run `rtk git status` in a repo with many changed files, verify truncation at 5
- [ ] Manual testing: set `status_max_untracked = 3` in config, verify untracked files are capped at 3

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1,6 +1,7 @@
 //! Filters git output — log, status, diff, and more — keeping just the essential info.
 
 use crate::core::args_utils;
+use crate::core::config;
 use crate::core::guard::never_worse;
 use crate::core::runner::{self, RunOptions};
 use crate::core::stream::{
@@ -632,14 +633,31 @@ fn truncate_line(line: &str, width: usize) -> String {
 }
 
 pub(crate) fn format_status_output(porcelain: &str) -> String {
-    format_status_inner(porcelain, None)
+    let limits = config::limits();
+    format_status_inner(
+        porcelain,
+        None,
+        limits.status_max_files,
+        limits.status_max_untracked,
+    )
 }
 
 pub(crate) fn format_status_output_detached(porcelain: &str, detached_ref: &str) -> String {
-    format_status_inner(porcelain, Some(detached_ref))
+    let limits = config::limits();
+    format_status_inner(
+        porcelain,
+        Some(detached_ref),
+        limits.status_max_files,
+        limits.status_max_untracked,
+    )
 }
 
-fn format_status_inner(porcelain: &str, detached: Option<&str>) -> String {
+fn format_status_inner(
+    porcelain: &str,
+    detached: Option<&str>,
+    max_files: usize,
+    max_untracked: usize,
+) -> String {
     let lines: Vec<&str> = porcelain
         .lines()
         .filter(|line| !line.trim().is_empty())
@@ -661,8 +679,32 @@ fn format_status_inner(porcelain: &str, detached: Option<&str>) -> String {
         }
     }
 
+    let mut tracked_count = 0usize;
+    let mut tracked_hidden = 0usize;
+    let mut untracked_count = 0usize;
+    let mut untracked_hidden = 0usize;
+
     for line in lines.iter().skip(1) {
-        output.push((*line).to_string());
+        if line.starts_with("??") {
+            untracked_count += 1;
+            if untracked_count <= max_untracked {
+                output.push((*line).to_string());
+            } else {
+                untracked_hidden += 1;
+            }
+        } else {
+            tracked_count += 1;
+            if tracked_count <= max_files {
+                output.push((*line).to_string());
+            } else {
+                tracked_hidden += 1;
+            }
+        }
+    }
+
+    if tracked_hidden > 0 || untracked_hidden > 0 {
+        let total = tracked_hidden + untracked_hidden;
+        output.push(format!("[+{} hidden, use git status]", total));
     }
 
     if lines.len() == 1 && lines[0].starts_with("##") {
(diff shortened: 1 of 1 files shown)


mysticgohan1:
@KuSh rebased on current develop (`git.rs` is `git_cmd.rs` now) and reworked to fit today's code. You closed #2459 in favour of this PR, so one design point for you:

Wiring the two limits with the documented defaults (15 tracked / 10 untracked) means `rtk git status` truncates for everyone, which reverses part of the "status completeness" work in 6ebde6d2 and 7753e487. I kept the documented defaults, but the alternatives are cheap: default both to unlimited (opt-in caps), or take option B from #2437 and delete the dead knobs. Which do you prefer?

What changed since the earlier push: entries past a cap are `... +N more` with the usual recall hint (the old "use git status" marker would be rewritten straight back to `rtk git status` by the hook), conflicted paths never count against a cap, and `LimitsConfig` gets `#[serde(default)]`, because a `[limits]` section with only the two status keys (the repro in #2437) failed to parse and silently dropped the whole config.

Checked on a 25-modified + 15-untracked repo: develop lists 41 lines; this lists 5, and `rtk recall <hash>` returns exactly the other 37. There are unit tests plus integration tests through the real binary.

guyoron1:
@KuSh rebased on current develop (`git.rs` is `git_cmd.rs` now) and reworked to fit today's code. You closed #2459 in favour of this PR, so one design point for you:

Wiring the two limits with the documented defaults (15 tracked / 10 untracked) means `rtk git status` truncates for everyone, which reverses part of the "status completeness" work in 6ebde6d2 and 7753e487. I kept the documented defaults, but the alternatives are cheap: default both to unlimited (opt-in caps), or take option B from #2437 and delete the dead knobs. Which do you prefer?

What changed since the earlier push: entries past a cap are `... +N more` with the usual recall hint (the old "use git status" marker would be rewritten straight back to `rtk git status` by the hook), conflicted paths never count against a cap, and `LimitsConfig` gets `#[serde(default)]`, because a `[limits]` section with only the two status keys (the repro in #2437) failed to parse and silently dropped the whole config.

Checked on a 25-modified + 15-untracked repo: develop lists 41 lines; this lists 5, and `rtk recall <hash>` returns exactly the other 37. There are unit tests plus integration tests through the real binary.

src/cmds/git/git_cmd.rs
```diff
@@ -2198,14 +2199,56 @@ fn truncate_line(line: &str, width: usize) -> String {
 }
 
 pub(crate) fn format_status_output(porcelain: &str) -> String {
-    format_status_inner(porcelain, None)
+    let limits = config::limits();
+    format_status_inner(
+        porcelain,
+        None,
+        limits.status_max_files,
```
KuSh, line 2206: **Blocks merge (1).** With the documented defaults, this line caps every stock `rtk git status` at 15 tracked / 10 untracked: a repo with 25 modified + 16 untracked files goes from all 41 entries on develop to 25 plus `... +16 more`, and with recall off the hidden paths cannot be recovered from rtk. That reverses #991. Making the caps `Option<usize>`, unset by default, clears it; details in the review body.

KuSh (changes requested):
**Claim**: `[limits] status_max_files` / `status_max_untracked` now bound the compact `rtk git status` listing (the rest stays recoverable through recall, conflicts never counted), and a partial `[limits]` section no longer drops the whole config.
**Scope**: accept if narrowed to "the two caps apply **when set**; unset by default, so `rtk git status` stays complete (#991)", plus the `#[serde(default)]` fix as is. (round 1, frozen)
**Ran**: `rtk rewrite "git status"` → `rtk git status` on real repos (25 modified + 15 untracked + an untracked dir; a 20-file merge with 3 conflicts; detached HEAD), develop merge-base 0c279f54 vs 3f0776a4 vs the merge into current develop, each binary built separately. Cases: no config, `[limits]` 2/1, 0/0, 100/100; `-sb`, `-b`, `--short --branch`, `-s`, `--porcelain`; `rtk pipe -f git-status`; `git -C <other>`; `RTK_RECALL=0`; sqlite and tee recovery modes, each hint followed through `rtk rewrite`. Invariants: under the hook | over-cap + accounting (shown + recalled = `git status --porcelain`, both modes) | non-default target (`-C`) | every decision (tracked / untracked / unmerged bucket, cap hit / exact / zero). Savings vs raw `git status` (964 B): develop 56%, this PR with defaults 67%, caps 2/1 91% — all above the 20% floor (CONTRIBUTING.md).
**Matrix**: never run before — executed this round
**CI** @3f0776a4: all green (ubuntu, macos, windows, clippy, fmt, benchmark)

1. `src/core/config.rs:238` — **the documented defaults now apply to everyone**, so a stock install hides dirty paths that develop lists. Evidence, no config.toml, 25 modified + 16 untracked: develop lists all 41; this PR lists 25 and prints `... +16 more [+16 hidden: rtk recall …]`. With `RTK_RECALL=0` or `[retriever] mode = "disabled"` the note is just `... +16 more` and the paths cannot be recovered from rtk at all. This is exactly what #991 removed (its test is kept here, but now run with caps of 100). The store write also costs time on every such status: `hyperfine` 6.5 ms (develop) → 9.8 ms (this PR, defaults), and 6.6 ms for both when nothing is hidden. See "Your questions" for the change that clears this.

- The listing tests that call `format_status_output` now read the developer's own `~/.config/rtk/config.toml`: with `status_max_files = 0` set locally, `test_format_status_output_{thai_filename,mixed_changes,preserves_rename_and_conflict_lines,preserves_nested_untracked_paths}` fail. A clean CI runner never sees it. Routing those seven callers through a small helper that calls `format_status_inner` with `usize::MAX` caps keeps them independent of local config.
- `rtk pipe -f git-status` fed `--porcelain=v2` output puts `u UU …` into the tracked bucket, so under a cap a v2 conflict can be hidden (still recallable). The formatter only speaks v1, and nothing in rtk sends it v2, so this is only a note.
- When `never_worse` falls back to raw (short headerless input through `rtk pipe`), the recall row or tee file has already been written for a hint that is never printed. Harmless apart from one orphan row.

### Filed as follow-ups
- #3431 (existing) — tee files are named `{epoch_seconds}_{slug}.log`, so two capped statuses in the same second share one file and the first hint returns the second repo's entries. Reproduced with `rtk git status` in repo A, then `rtk git -C B status`: A's `tail -n +3 …_git-status.log` printed `B_f3`…`B_f6`. Pre-existing for every tee caller (`find`, `ls`, `search`, …). This PR adds one caller, reached only with caps set **and** tee mode, so it's absorbed there (comment with the repro added) rather than held here.
- Partial `[tracking]` / `[display]` / `[filters]` sections still fail to parse and drop the whole config, `[limits]` included — pre-existing, covered by open #1483. Both PRs add the same `#[serde(default)]` line to `LimitsConfig`; whichever lands second rebases trivially.

- Sqlite recall returns exactly the hidden entries; shown + recalled = `git status --porcelain`, with no duplicates (38 hidden at 2/1, 16 at defaults). Tee mode's `tail -n +N <file>` is left native by `rtk rewrite` and returns the hidden tail.
- A real merge with 20 staged files and 3 `UU` entries under caps 2/1: all three conflicts shown, only staged/untracked entries counted.
- `-sb`, `-b`, `--short --branch` take the capped path; `-s` and `--porcelain` pass through untouched. `git -C <other> status`, detached HEAD and `rtk pipe -f git-status` apply the same caps.
- `#[serde(default)]` on `LimitsConfig`: on develop, `[limits]` with only the two status keys makes `rtk config` fail (`missing field … in 'limits'`) and drops the config; here it parses and keeps the other defaults.
- `cargo fmt --check`, `cargo clippy --all-targets` (0 warnings), `cargo test --all` (3936 passed) on this branch merged into current develop.

- "`config::limits()` re-parses config.toml on the hot path": measured, develop 6.6 ms vs this PR 6.6 ms with caps set high enough that nothing is hidden. The time difference in the blocker comes from the store write, not the parse.
- "One combined `+N more` count across both buckets hides which kind was cut": wording preference; recall shows it.
- "`TEST_ENV_LOCK.lock().unwrap()` in the new helper breaks the test rule": it's the established idiom in the existing `tee.rs` and `retriever.rs` tests.

### Your questions
- "I kept the documented defaults, but the alternatives are cheap: default both to unlimited (opt-in caps), or take option B from #2437 and delete the dead knobs. Which do you prefer?" → **Opt-in: unset by default.** Option B would throw away a feature this PR makes work, and the documented defaults undo #991 for everyone (blocker 1). One constraint I checked: `usize::MAX` as the "unlimited" default breaks `rtk config` and `rtk config --create` (`rtk: out-of-range value for u64 type`, since TOML integers are i64), so the fields need to become `Option<usize>` with `skip_serializing_if = "Option::is_none"`. The docs that name the defaults (`src/core/README.md`, `src/cmds/git/README.md`, `docs/contributing/CODING_PRACTICES.md`) change with them. I checked that shape locally: it passes fmt, clippy and `cargo test --all`, and on the same repo gives: no config → all 41 entries, byte-identical to develop; caps 2/1 → capped with a working recall; `status_max_files = 0` alone → honoured; `rtk config` / `--create` serialize and round-trip.

### Next
Make the two caps opt-in (`Option<usize>`, unset by default). After that this merges.

PR #2581, opened 2026-06-24
fix(stream): collapse terminal redraw controls

### LINKED ISSUE #2230: fix(filtering): remove terminal output overwritten by control flow (\r and \b)
This is a more general case of https://github.com/rtk-ai/rtk/issues/624.

It's not uncommon for tools (like ollama, npm, curl, bazel, ninja, cmake...) to emit self-overwriting progress messages like like for example `⠋\b⠙\b⠹\b⠸\b⠼\b⠴\b⠦\b⠧\b⠇\b⠏\b` or `Downloading... 0%\rDownloading... 1%\rDownloading... 2%\rDownloading... 3%\r`.  I can't really think of any situation where one would want an LLM to see output intermediate outputs that would be only transiently visible to a human.

`\r` and `\b` are the most common cases to handle, but there's also ansi sequences that e.g. ncurses uses, like

 - `\e[A` (moves up one line)
 - `\e[{N}A`  (e.g., `\e[3A` moves up 3 lines)
 - `\e[B` (move down)
 - `\e[C` (move forward)
 - `\e[D` (more backward)
 - `\e[{row};{col}H` (e.g., `\e[1;1H` moves to the top-left corner)

There's currently functionality to [_strip out_ ansi sequences](https://github.com/rtk-ai/rtk/blob/4c278d18c87d4ee71cb5cd28181d4fb84fb167b7/src/core/utils.rs#L48-L53), which is good for color etc, but that's not necessarily the desired behavior here.  Implementing this comprehensively would probably end up being equivalent to writing a terminal emulator, particularly with the absolute positioning sequences.  That might be tricky to implement, but there are many common cases which only update part or all of the most recent line, which could be handled more easily.

This should be applied _before_ the output is separated into lines (especially if one wishes to eventually deal with multi-line edits).  For one thing, `\r\n` is a completely different animal from `\r` without the `\n` (which should also probably be normalized, in general, for the benefit of Windows users who otherwise might find toml matchers failing to detect lines that look like they should be detected).  Also perhaps obviously would need to happen before `strip_ansi` gets applied

YOMXXX:
## Summary
- collapse common terminal redraw controls before captured output reaches filters
- handle carriage-return progress rewrites and backspace overwrites for `exec_capture` and `run_streaming` capture/filter paths
- preserve normal CRLF line endings while avoiding transient progress/spinner states in model-facing output

Fixes #2230

## Verification
- `cargo +1.93.0 fmt --check`
- `git diff --check`
- `cargo +1.93.0 test core::stream::tests -- --nocapture`

### DIFF
--- a/src/core/stream.rs
+++ b/src/core/stream.rs
@@ -241,6 +241,47 @@ pub fn status_to_exit_code(status: std::process::ExitStatus) -> i32 {
     1
 }
 
+/// Collapse common single-line terminal redraw controls before filters parse output.
+fn collapse_terminal_control(text: &str) -> String {
+    let mut visible = String::new();
+    let mut line: Vec<char> = Vec::new();
+    let mut cursor = 0usize;
+    let mut chars = text.chars().peekable();
+
+    while let Some(ch) = chars.next() {
+        match ch {
+            '\r' if chars.peek() == Some(&'\n') => {
+                chars.next();
+                visible.extend(line.drain(..));
+                visible.push('\n');
+                cursor = 0;
+            }
+            '\r' => {
+                cursor = 0;
+            }
+            '\n' => {
+                visible.extend(line.drain(..));
+                visible.push('\n');
+                cursor = 0;
+            }
+            '\u{8}' => {
+                cursor = cursor.saturating_sub(1);
+            }
+            ch => {
+                if cursor < line.len() {
+                    line[cursor] = ch;
+                } else {
+                    line.push(ch);
+                }
+                cursor += 1;
+            }
+        }
+    }
+
+    visible.extend(line);
+    visible
+}
+
 // ISSUE #897: ChildGuard RAII prevents zombie processes that caused kernel panic
 pub const RAW_CAP: usize = 10_485_760; // 10 MiB
 
@@ -366,6 +407,7 @@ pub fn run_streaming(
                     StreamLine::Stderr(l) => (l, true),
                     StreamLine::Stdout(l) => (l, false),
                 };
+                let line = collapse_terminal_control(&line);
                 if is_stderr {
                     if !capped_err {
                         if raw_stderr.len() + line.len() < RAW_CAP {
@@ -447,6 +489,7 @@ pub fn run_streaming(
                             );
                         }
                     }
+                    raw_stdout = collapse_terminal_control(&raw_stdout);
                     filtered = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
                         filter_fn(&raw_stdout)
                     }))
@@ -472,15 +515,16 @@ pub fn run_streaming(
                             );
                         }
                     }
+                    raw_stdout = collapse_terminal_control(&raw_stdout);
                     filtered = raw_stdout.clone();
                 }
             }
         }
 
-        raw_stderr = stderr_thread.join().unwrap_or_else(|e| {
+        raw_stderr = collapse_terminal_control(&stderr_thread.join().unwrap_or_else(|e| {
             eprintln!("[rtk] warning: stderr reader thread panicked: {:?}", e);
             String::new()
-        });
+        }));
     }
     if let Some(t) = stdin_thread {
         t.join().ok();
@@ -535,8 +579,8 @@ pub fn exec_capture(cmd: &mut Command) -> Result<CaptureResult> {
     cmd.stdin(Stdio::null());
     let output = cmd.output().context("Failed to execute command")?;
     Ok(CaptureResult {
-        stdout: String::from_utf8_lossy(&output.stdout).into_owned(),
-        stderr: String::from_utf8_lossy(&output.stderr).into_owned(),
+        stdout: collapse_terminal_control(&String::from_utf8_lossy(&output.stdout)),
+        stderr: collapse_terminal_control(&String::from_utf8_lossy(&output.stderr)),
         exit_code: status_to_exit_code(output.status),
     })
 }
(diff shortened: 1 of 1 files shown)


YOMXXX:
Status note: the red `benchmark` check is not caused by this PR's terminal-control change. The failing row is `curl text`: CI compares `curl -s https://httpbin.org/robots.txt` at 8 tokens with `rtk curl https://httpbin.org/robots.txt` at 40 tokens, so the external network fixture can make the raw output smaller than the RTK path and trip the negative-savings gate.

pidefrem:
Thanks for this — I picked it up while working on the same capture path and it's the right approach (collapse to the final rendered line, on both `exec_capture` and `run_streaming`). One correctness issue I hit while testing, with a fix, in case it's useful:

**`collapse_terminal_control` has no CSI awareness, so `ESC[K` (erase-in-line) is overlaid as literal characters.** That's the sequence programs emit for a *shrinking* redraw — `\r` to column 0, then erase a longer previous frame, then write the shorter one. Today it both leaks the escape bytes and leaves the stale tail:

```
input : "Downloading 100%\r\x1b[KDone\n"
#2581 : "\x1b[KDoneding 100%\n"   // escape leaked + "loading 100%" tail survives
want  : "Done\n"
```

Root cause is that escape bytes and printable bytes go through different write paths. The fix that worked for me: parse a CSI sequence (`ESC [ … final`) as a unit, honor `ESC[K` (params 0/1/2) against the cursor, and overlay every *other* CSI (SGR colour, cursor motion) through the **same** write path as printable text — so colour still passes through untouched and a later `\r` redraw overwrites prior escape bytes cell-for-cell instead of shifting them.

A couple of cases I'd suggest adding as regression tests (all pass with the fix):
- `"Downloading 100%\r\x1b[KDone\n"` → `"Done\n"` (erase-to-EOL clears tail, no leak)
- `"abcdef\rXY\x1b[K\n"` → `"XY\n"` (mid-line erase)
- `"\x1b[32mPASS\x1b[0m\n"` → unchanged (SGR colour must pass through — `strip_ansi`'s job)
- `"Downloading 100%\rDone\n"` → `"Doneloading 100%\n"` (bare `\r`, *no* erase, stays faithful to what a real terminal renders — this one should NOT change)

One observation while validating, not a criticism: rtk captures over a pipe, and I measured that tools suppress `\r` progress redraws when stdout isn't a TTY. Piped (the capture path), `pip3 download`, `npm`, `brew`, and `cargo build --color=always` all emit **zero `\r`** (cargo keeps colour, drops the redraw); only a program that ignores `isatty()` produced any. So the `\r`-collapse mostly pays off on a PTY path / tools that force progress, whereas the `ESC[K` and `\b` handling is what reliably reaches the filters under piped capture. Doesn't change your approach at all — `\r`/`\b`/`ESC[K` as one cursor model is right; just context in case you're weighting where it helps.

I have the fix implemented and tested on my fork — commit [`1f32130`](https://github.com/pidefrem/rtk/commit/1f3213014f297efd0e55fd5a45facf34075dc8df) on branch [`fix/stream-crlf-line-splitting`](https://github.com/pidefrem/rtk/tree/fix/stream-crlf-line-splitting) (which builds on your `collapse_terminal_control`). Happy to open it as a PR against your branch, or you can take the diff directly — whichever you prefer.

YOMXXX:
Fixed in `7fb1339` (`fix(stream): honor erase-line terminal controls`).

What changed:
- `collapse_terminal_control` now parses CSI sequences as a unit.
- `ESC[K` / `ESC[0K` erases from the cursor to EOL, so shrinking redraws no longer leak `\x1b[K` or preserve stale tails.
- `ESC[1K` and `ESC[2K` are handled in the same erase-line path.
- Non-erase CSI sequences, including SGR color like `ESC[32m`, still pass through the same overlay path unchanged.
- Bare `\r` behavior is unchanged, so `Downloading 100%\rDone\n` still renders as `Doneloading 100%\n` unless an erase-line sequence is present.

Regression coverage added for:
- `Downloading 100%\rESC[KDone\n` -> `Done\n`
- `abcdef\rXYESC[K\n` -> `XY\n`
- SGR preservation
- bare CR tail preservation

Local verification:
- `cargo +1.93.0 fmt --check`
- `git diff --check`
- `cargo +1.93.0 test collapse_terminal_control -- --nocapture`
- `cargo +1.93.0 test core::stream::tests -- --nocapture`

HetCreep:
The CSI-as-a-unit rewrite in `7fb1339` is relevant to a defect I ran into elsewhere in the tree during an audit of **v0.44.2 (`700bdde`)**, and I think this PR quietly builds the primitive that fixes it.

*Provenance: AI-assisted source audit, human-directed. Source read at `700bdde`; the two regexes below are quoted verbatim, but I did not run either stripper against a live spinner.*

### `strip_ansi` cannot match a CSI private-parameter sequence

`src/core/utils.rs:52-57`:

```rust
static ANSI_RE: LazyLock<Regex> =
    LazyLock::new(|| Regex::new(r"\x1b\[[0-9;]*[a-zA-Z]").unwrap());
```

The parameter class is `[0-9;]*`, which excludes `?`. So `\x1b[?25l` and `\x1b[?25h` — the cursor hide/show pair emitted by npm, cargo and gradle spinners — do not match, and the whole escape survives into the filtered output. This function has ~67 call sites across `src/`.

That is the same class of input this PR is handling; the difference is that `collapse_terminal_control` now parses the sequence structurally, while `strip_ansi` is still pattern-matching a subset of CSI.

### The PHP copy fails on the same input, differently

`src/cmds/php/utils.rs`:

```rust
pub fn strip_ansi_and_controls(input: &str) -> String {
    let no_ansi = ANSI_RE.replace_all(input, "");
    CONTROL_RE.replace_all(&no_ansi, "").to_string()
}
```

Same `ANSI_RE`, so `\x1b[?25l` survives it too — but `CONTROL_RE`'s range `[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]` covers `\x1B`, so the second pass removes the ESC byte and leaves a literal `[?25l` in the text. Worth confirming against the actual constant, but if that reading is right, the two strippers produce different wrong answers for the same bytes: one leaves the full escape, the other leaves visible garbage. Callers are `artisan_cmd.rs` and `ecs_cmd.rs`.

### Suggestion

If `collapse_terminal_control` ends up as a general CSI-aware scanner, having `strip_ansi` delegate to it (or share the parameter-parsing step) would remove the divergence and cover private-parameter sequences in one place. Widening the regex to `\x1b\[[0-9;?]*[a-zA-Z]` would be the one-character-ish version, though it still leaves two implementations to keep in sync.

Not asking to expand this PR — it is scoped to the capture path and the CSI rewrite already addresses the harder half. Flagging because the parser you just wrote is the thing the rest of the tree is missing.

PR #2584, opened 2026-06-24
fix(filters): mark truncated fallback output

### LINKED ISSUE #2204: passthrough_max_chars truncates output silently (no marker) — git log shows 50 of 252 commits, misleads scope

YOMXXX:
## Summary
- add visible recovery hints when `rtk git log` reaches RTK-injected default caps
- use parser passthrough truncation markers for ruff, ESLint, Pylint, and golangci-lint JSON parse fallbacks
- add regression tests for hidden truncation in git log caps and parse-fail fallback paths

Fixes #2204

## Verification
- `cargo +1.93.0 fmt --check`
- `git diff --check`
- `cargo +1.93.0 test parse_failure_marks_truncation -- --nocapture`
- `cargo +1.93.0 test filter_log_output -- --nocapture`

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -568,24 +568,27 @@ pub(crate) fn filter_log_output(
     if user_format {
         let lines: Vec<&str> = output.lines().collect();
         let max_lines = if user_set_limit { lines.len() } else { limit };
-        return lines
+        let mut result = lines
             .iter()
             .take(max_lines)
             .map(|l| truncate_line(l, truncate_width))
-            .collect::<Vec<_>>()
-            .join("\n");
+            .collect::<Vec<_>>();
+        if !user_set_limit && lines.len() >= max_lines && !result.is_empty() {
+            result.push(log_limit_marker(max_lines));
+        }
+        return result.join("\n");
     }
 
     // RTK injected format: split output into commit blocks separated by ---END---
-    let commits: Vec<&str> = output.split("---END---").collect();
+    let commits: Vec<&str> = output
+        .split("---END---")
+        .map(str::trim)
+        .filter(|block| !block.is_empty())
+        .collect();
     let max_commits = if user_set_limit { commits.len() } else { limit };
 
     let mut result = Vec::new();
     for block in commits.iter().take(max_commits) {
-        let block = block.trim();
-        if block.is_empty() {
-            continue;
-        }
         let mut lines = block.lines();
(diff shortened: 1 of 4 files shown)


YOMXXX:
Status note: the red `benchmark` check is not caused by this PR's truncation-marker change. The failing row is `curl text`: CI compares `curl -s https://httpbin.org/robots.txt` at 8 tokens with `rtk curl https://httpbin.org/robots.txt` at 40 tokens, so the external network fixture can make the raw output smaller than the RTK path and trip the negative-savings gate.

YOMXXX:

YOMXXX:
Ready for review/merge. Rebased onto current develop and the fresh CI run is green, including benchmark and all platform tests. Scope remains the visible truncation markers for capped fallback/filter output.

PR #2588, opened 2026-06-24
fix(init): replace stale legacy hook registrations

### LINKED ISSUE #1962: rtk-rewrite.sh: No such file or directory
PreToolUse:Bash hook error      ⎿  Failed with non-blocking status code: /bin/sh:
                                                                     /Users/name/.claude/hooks/rtk-rewrite.sh: No such file
                                                                      or directory

 ✘  ~  rtk init -g

RTK hook registered (global).

  Command:   rtk hook claude
  RTK.md:    /Users/name/.claude/RTK.md (10 lines)

 ~  rtk gain

Total commands:    1936
Input tokens:      10.1M
Output tokens:     1.3M
Tokens saved:      8.8M (86.8%)
Total exec time:   218m33s (avg 6.8s)
Efficiency meter: █████████████████████░░░ 86.8%

[warn] No hook installed — run `rtk init -g` for automatic token savings

YOMXXX:
## Summary
- treat legacy `rtk-rewrite.sh` settings entries as stale when installing the native `rtk hook claude` command
- remove legacy settings entries even when the old hook script file is already missing
- add regression coverage for missing legacy hook registrations and native idempotency detection

Fixes #1962

## Verification
- `cargo +1.93.0 fmt --check`
- `git diff --check`
- `cargo +1.93.0 test legacy -- --nocapture`
- `cargo +1.93.0 test hooks::init -- --nocapture`
- manual temp-HOME smoke for `rtk init -g --auto-patch` replacing a missing `rtk-rewrite.sh` registration

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -1094,9 +1094,11 @@ fn insert_hook_entry(root: &mut serde_json::Value, hook_command: &str) -> Result
     Ok(())
 }
 
-/// Check if RTK hook is already present in settings.json
-/// Matches on legacy rtk-rewrite.sh path OR new `rtk hook claude` command
+/// Check if the requested RTK hook is already present in settings.json.
+/// Legacy script entries do not satisfy the native `rtk hook claude` command;
+/// they must be migrated instead of blocking the new hook registration.
 fn hook_already_present(root: &serde_json::Value, hook_command: &str) -> bool {
+    let installing_legacy_hook = hook_command.contains(REWRITE_HOOK_FILE);
     let pre_tool_use_array = match root
         .get("hooks")
         .and_then(|h| h.get(PRE_TOOL_USE_KEY))
@@ -1112,7 +1114,9 @@ fn hook_already_present(root: &serde_json::Value, hook_command: &str) -> bool {
         .flatten()
         .filter_map(|hook| hook.get("command")?.as_str())
         .any(|cmd| {
-            cmd == hook_command || cmd == CLAUDE_HOOK_COMMAND || cmd.contains(REWRITE_HOOK_FILE)
+            cmd == hook_command
+                || (installing_legacy_hook
+                    && (cmd == CLAUDE_HOOK_COMMAND || cmd.contains(REWRITE_HOOK_FILE)))
         })
 }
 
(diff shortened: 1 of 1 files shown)


mattbuot:
Flagging a behaviour change in this diff before it lands, from reproducing the #1962 state against current `develop`.

On `develop` today, `rtk init -g` with a legacy install splits into two outcomes, decided by the `dominated_by_legacy` predicate in `remove_legacy_hook_entries_from_json` (`init.rs:1356`):

- rtk's legacy entry alone in the `Bash` matcher group → script deleted **and** entry removed; `settings.json` ends as `"PreToolUse": []`.
- another hook sharing that group → script deleted, entry **kept** → the dangling reference reported in #1962. Verified: with a co-resident `/opt/acme/audit-hook.sh`, `settings.json` comes back byte-identical and the script is gone.

This diff hoists `remove_legacy_settings_entries` out of the deletion-success branch, so the entry is stripped unconditionally — including when `fs::remove_file` failed (the `Err` arm now only logs) and when no legacy script exists at all.

That does remove the dangling reference, but by removing the registration rather than repairing it, and the replacement write is not guaranteed: `patch_settings_json_command` under the default `PatchMode::Ask` calls `prompt_user_consent`, which returns `Ok(false)` on non-TTY stdin (`init.rs:445`) and yields `PatchResult::Declined`. The `Declined` arm at `init.rs:1211` prints nothing, and the `RTK hook registered (global).` banner at `init.rs:1179` has already printed at step 4, before `settings.json` is touched at step 5.

So for the shared-group case this trades a visible failure for a silent one — Claude Code raises a hook error on a dangling script, and raises nothing at all when there is no entry. I reproduced the non-interactive run in an isolated `HOME` and it exits 0 with the success banner in both shapes.

The ordering PR #3570 adopted for Cursor avoids this: patch the config first, delete the legacy artefacts only after that write succeeds. Applying the same order here would let this PR close #1962 without the trade.

Happy to send a rebased version on top of current `develop` with that ordering and regression coverage for both the declined-prompt path and the co-resident-hook case — this PR is `mergeable_state: dirty` and not rebaseable in place. Full credit to @YOMXXX for the original.

PR #2591, opened 2026-06-24
fix(grep): show binary-only match lines

### LINKED ISSUE #2310: rtk grep (non-recursive) reports "N matches in 0 files" and hides all matched lines

YOMXXX:
## Summary

- parse binary-only grep/rg match lines instead of dropping them
- count the searched file when `rg` emits an unprefixed `binary file matches ...` line
- handle prefixed rg and BSD grep binary-match formats in the same parser
- add an integration regression test for a single binary file match

## Issue

Fixes #2310.

## Verification

- `cargo +1.93.0 test --all -- --skip small_grep_not_worse_than_plain` before changes: passed
- `cargo +1.93.0 test --test grep_binary_test -- --nocapture` before fix: failed with `1 matches in 0 files` and `[+1 more]`
- `cargo +1.93.0 test --test grep_binary_test -- --nocapture`: passed
- `cargo +1.93.0 test grep_cmd::tests::test_parse -- --nocapture`: passed
- `cargo +1.93.0 test grep_cmd::tests -- --nocapture`: passed
- `cargo +1.93.0 fmt --check`: passed
- `git diff --check`: passed
- `cargo +1.93.0 clippy --all-targets`: passed
- `cargo +1.93.0 test --all -- --skip small_grep_not_worse_than_plain`: passed

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -578,6 +578,10 @@ pub fn run(
 /// The `bool` in the tuple is `true` for match lines (`:` separator) and
 /// `false` for context lines (`-` separator, emitted by -A/-B/-C).
 fn parse_match_line(line: &str) -> Option<(String, usize, bool, &str)> {
+    if let Some((file, content)) = parse_binary_match_line(line) {
+        return Some((file, 0, true, content));
+    }
+
     lazy_static::lazy_static! {
         static ref MATCH_LINE_RE: Regex = Regex::new(r"^([^\x00]+)\x00(\d+)([:-])(.*)$").unwrap();
     }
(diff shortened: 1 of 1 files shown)


YOMXXX:
Rebased onto current develop and resolved the grep parser conflict by keeping the NUL-delimited parser as the default path, with a narrow special case for binary-match notices so binary-only results are counted and rendered instead of falling through to passthrough.

Local verification:
- cargo +1.93.0 fmt --check
- git diff --check
- cargo +1.93.0 test grep_single_binary_match_reports_file --test grep_binary_test -- --nocapture
- cargo +1.93.0 test grep -- --skip small_grep_not_worse_than_plain --nocapture

Note: the skipped test is an existing local-environment failure here; the targeted binary regression and the rest of grep-focused coverage passed.

YOMXXX:
Follow-up after the first rebase CI: the new integration test depended on runner-specific grep/rg binary-file behavior. Ubuntu returned `0 matches` while Windows emitted a raw `Binary file ... matches` line, so that test was not portable.

I replaced it with platform-independent unit coverage for the parser/safety-net path, including the Windows-style `Binary file C:\...\repro.bin matches` notice from CI. Re-ran locally:
- cargo +1.93.0 fmt --check
- git diff --check
- cargo +1.93.0 test parse_binary_match -- --nocapture
- cargo +1.93.0 test windows_grep_binary_match -- --nocapture
- cargo +1.93.0 test unparsed_signal_windows_binary_notice_parse_ok -- --nocapture
- cargo +1.93.0 test grep -- --skip small_grep_not_worse_than_plain --nocapture

PR #2593, opened 2026-06-24
fix(pnpm): preserve install failure output

### LINKED ISSUE #2023: `pnpm install` failure is erased, causing loop in OpenCode
```
% rtk pnpm install                   
%
% pnpm install && echo OK || echo ERR
Scope: all 3 workspace projects
 ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/@rdfjs%2Fserializer-nquads: Not Found - 404

This error happened while installing a direct dependency of /XYZ/packages/server

@rdfjs/serializer-nquads is not in the npm registry, or you have no permission to fetch it.

No authorization header was set for the request.
../..                                    | Progress: resolved 13, reused 13, downloaded 0, added 0
ERR 
```

The install failure happens because the package `@rdfjs/serializer-nquads` doesn't exist.

In OpenCode with Kimi K2.6 this caused a loop for me (in subagent):
```
# Install types package dependencies in packages/types
$ rtk pnpm install
(no output)

Thought: I need to stop making this mistake. I've been prepending "rtk" to my bash commands repeatedly. This is clearly a persistent error. I need to write the correct command now. The command should be just "pnpm install" without "rtk".

# Install types package dependencies in packages/types
$ rtk pnpm install
(no output)

Thought: I need to stop making this mistake. I've been prepending "rtk" to my bash commands repeatedly. This is clearly a persistent error. I need to write the correct command now. The command should be just "pnpm install" without "rtk".# Install types package dependencies in packages/types

$ rtk pnpm install
(no output)

Thought: I need to stop making this mistake. I've been prepending "rtk" to my bash commands repeatedly. This is clearly a persistent error. I need to write the correct command now. The command should be just "pnpm install" without "rtk".

...
```

YOMXXX:
## Summary
- print captured stdout as well as stderr when `pnpm install` exits non-zero
- keep the existing success-path filtering unchanged
- add a fake-pnpm regression test for install failures that report errors on stdout

Fixes #2023.

## Verification
- RED: `cargo +1.93.0 test --test pnpm_install_failure_test -- --nocapture` failed before the fix with `pnpm stdout error must be preserved, got: ""`
- `cargo +1.93.0 test --test pnpm_install_failure_test -- --nocapture`
- `cargo +1.93.0 test cmds::js::pnpm_cmd::tests -- --nocapture`
- `cargo +1.93.0 fmt --check`
- `git diff --check`

### DIFF
--- a/src/cmds/js/pnpm_cmd.rs
+++ b/src/cmds/js/pnpm_cmd.rs
@@ -483,6 +483,7 @@ fn run_install(args: &[String], verbose: u8) -> Result<i32> {
     let result = exec_capture(&mut cmd).context("Failed to run pnpm install")?;
 
     if !result.success() {
+        print!("{}", result.stdout);
         eprint!("{}", result.stderr);
         return Ok(result.exit_code);
     }
--- a/tests/pnpm_install_failure_test.rs
+++ b/tests/pnpm_install_failure_test.rs
@@ -1,0 +1,86 @@
+use std::env;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+
+fn write_fake_pnpm(dir: &Path) -> PathBuf {
+    #[cfg(windows)]
+    {
+        let path = dir.join("pnpm.cmd");
+        fs::write(
+            &path,
+            "@echo off\r\n\
+             if \"%1\"==\"install\" (\r\n\
+             echo Scope: all 3 workspace projects\r\n\
+             echo ERR_PNPM_FETCH_404 GET https://registry.npmjs.org/@missing/pkg: Not Found - 404\r\n\
+             echo This error happened while installing a direct dependency of /tmp/app\r\n\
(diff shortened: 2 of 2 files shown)


src/cmds/js/pnpm_cmd.rs
```diff
@@ -483,6 +483,7 @@ fn run_install(args: &[String], verbose: u8) -> Result<i32> {
     let result = exec_capture(&mut cmd).context("Failed to run pnpm install")?;
 
     if !result.success() {
+        print!("{}", result.stdout);
```
KuSh, line 486: This should only be needed if stderr is ever empty, which is what the issue describes. Also, you should filter it with `filter_pnpm_install` instead of printing it raw.

src/cmds/js/pnpm_cmd.rs
KuSh, line 381: This has the exact same bug you're fixing. Would you mind fixing it as well?

KuSh (changes requested):
Hi @YOMXXX, this sounds like a good fix, but lacks some guards

PR #2595, opened 2026-06-24
fix(rewrite): stop rewritting gradle to gradlew

### LINKED ISSUE #2374: gradle command getting rewritten to gradlew
Invocations of `gradle` are getting rewritten to `gradlew`.  Please invoke `gradle` rather than `gradlew` when `gradle` is specified as mise is being used to manage gradle versions, not the wrapper script.

guyoron1:
## Summary

The rewrite rule treats `gradle` and `gradlew` as interchangeable, but they're different binaries. Users managing Gradle via mise/sdkman get their `gradle` commands silently rewritten to the project wrapper (`gradlew`), which may be a different version or not exist.

Removed `gradle` from the rewrite pattern so only wrapper variants (`./gradlew`, `gradlew`, `gradlew.bat`) are rewritten.

Fixes #2374

## Test plan

- [ ] Manual testing: `gradle build` should no longer be rewritten to `gradlew build`
- [ ] Manual testing: `./gradlew build` and `gradlew build` should still rewrite correctly

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -4314,13 +4314,10 @@ mod tests {
     }
 
     #[test]
-    fn test_classify_gradle() {
+    fn test_classify_gradle_not_rewritten() {
         assert!(matches!(
             classify_command("gradle build"),
-            Classification::Supported {
-                rtk_equivalent: "rtk gradlew",
-                ..
-            }
+            Classification::Unsupported { .. }
         ));
     }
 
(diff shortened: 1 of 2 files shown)


KuSh:
Closing as a duplicate of #3551, which fixes the same problem of gradle being silently rewritten to the wrapper. #3551 keeps `gradle build` filtered by giving system Gradle its own `rtk gradle` route instead of dropping the rewrite, and it comes with tests and docs for both routes. Review and comments are very welcome over there, and if you had a case in mind that #3551 does not cover, say so here and we will reopen.

PR #2596, opened 2026-06-24
fix(ci): make curl benchmark deterministic

YOMXXX:
## Summary
- Replace the benchmark curl/httpbin text and JSON calls with local `file://` fixtures.
- Avoids comparing two independent external network responses in the benchmark job.
- Keeps the benchmark exercising the `rtk curl` path while removing the flaky `curl text` negative-token failure seen across open PRs.

## Verification
- bash -n scripts/benchmark.sh
- git diff --check
- cargo +1.93.0 build --release
- Manual `curl -s file://...` vs `target/release/rtk curl file://...` check for both JSON and text fixtures

Note: a full local `CI=1 scripts/benchmark.sh` run still fails in this worktree because local warm-cache commands like `cargo build` produce tiny raw output and become existing unrelated negative cases. The curl rows specifically now report equal token counts instead of a negative result.

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -346,8 +346,18 @@ bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 # ===================
 section "curl"
 if command -v curl &> /dev/null; then
-  bench "curl json" "curl -s https://mockhttp.org/json" "$RTK curl https://mockhttp.org/json"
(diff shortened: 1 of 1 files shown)


YOMXXX:
Ready for review/merge. Rebased onto current develop and the fresh CI run is green, including benchmark. This removes the flaky external curl benchmark fixture by using local file:// fixtures, which should unblock the related benchmark-red PRs.

KuSh:

Issue #2589, opened 2026-06-24
Bug: `rtk find` stops working after `git init`


Saerdna:
## Description

`rtk find` stops finding files after the directory is initialized with `git init`. Before git init, it works correctly.

## Steps to Reproduce

1. In a non-git directory, create a file and run `rtk find`:
   ```bash
   # No .git directory exists
   touch message.csv
   rtk find ./ -name "message.csv"
   # ✅ Returns: ./message.csv
   ```

2. Initialize git, then run the same command:
   ```bash
   git init
   rtk find ./ -name "message.csv"
   # ❌ Returns: 0 for 'message.csv' (empty)
   ```

3. The raw `find` command still works:
   ```bash
   find ./ -name "message.csv"
   # ✅ Returns: ./message.csv
   ```

4. `rtk proxy` (bypasses filtering) also works:
   ```bash
   rtk proxy find ./ -name "message.csv"
   # ✅ Returns: ./message.csv
   ```

## Expected Behavior

`rtk find` should return the same results as `find` regardless of git status.

## Actual Behavior

- **Before `git init`**: `rtk find` works correctly
- **After `git init`**: `rtk find` returns no results

## Root Cause

rtk's output filtering logic incorrectly suppresses `find` command results when the working directory is under git version control.

## Environment

- rtk version: 0.42.4
- OS: macOS (arm64)
- Shell: zsh

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 97% |

### Summary

rtk find silently returns empty results in git repos due to incorrect output filtering; PR #2598 already open by reporter.

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

- `src/parser/mod.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

PR #2598, opened 2026-06-25
fix: rtk find returns empty results after git init (#2589)

Saerdna:
## Summary

  - Fix `rtk find` returning empty results after `git init` (#2589)
  - Disable gitignore filtering (`git_ignore`/`git_global`/`git_exclude`) in `WalkBuilder` so `rtk find` matches native `find` behavior
  - Add regression test covering find in both non-git and git-initialized directories

  ## Test plan

  - [x] `cargo clippy --all-targets && cargo test` — 30/30 find_cmd tests pass, clippy clean
  - [x] Manual testing: `rtk find . -name "message.csv"` returns correct results in git-initialized directory with global gitignore `*.csv` rule

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -217,9 +217,9 @@ pub fn run(
     let mut builder = WalkBuilder::new(path);
     builder
         .hidden(!search_hidden) // skip hidden files/dirs unless pattern targets dotfiles
-        .git_ignore(true) // respect .gitignore
-        .git_global(true)
-        .git_exclude(true);
+        .git_ignore(false) // do not respect .gitignore — match native find (#2589)
+        .git_global(false)
+        .git_exclude(false);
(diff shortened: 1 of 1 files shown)


Saerdna:
Requesting review when convenient. This is a small behavior fix for `rtk find` to match native `find`, with a regression test included.

KuSh:
Thanks for the regression test and for chasing this down. The underlying problem was fixed by #3854, merged on 2026-09-03, which keeps gitignore filtering but discloses hidden and ignored matches instead of dropping them silently, and develop already carries an equivalent test (hidden_and_ignored_matches_are_collected_for_disclosure in src/cmds/system/find_cmd.rs). Running the reporter's case on current develop, rtk find ./ -name message.csv in a repo that ignores *.csv now reports the filtered match and the recovery log rather than an empty result, so this PR's test is covered. Closing as already handled, but please comment or reopen if you see a case it still misses.

PR #2599, opened 2026-06-25
fix(tee): harden recovery-file perms against world-read and symlink attack

wernerkasselman-au:
## Summary

- Tee recovery files (the raw output of failed commands, which can include secrets such as auth headers, env dumps, or `--debug` traces) were written **world-readable `0644`** via `std::fs::write` into a `0755` dir. This is on **by default** (`TeeConfig` `enabled=true`, `mode=Failures`).
- Now `create_dir_private` locks the tee dir to `0700`, and `write_file_private` writes `0600` (Unix). `write_file_private` also uses `create_new(true)` (**O_EXCL**), so a symlink pre-planted at the predictable `{epoch}_{slug}.log` name is refused rather than followed or clobbered. That closes a symlink TOCTOU under an `RTK_TEE_DIR` override that was world-writable before the chmod.
- `write_tee_file` retries on `AlreadyExists` with a `{epoch}_{slug}.{n}.log` suffix (bounded `n<100`, still `create_new`), so legitimate same-second collisions still tee without ever following a symlink.
- All changes are `cfg(unix)`-gated; non-Unix behavior is unchanged.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test`, run on Rust **1.94.1** (the default host toolchain is 1.90, below the declared MSRV `1.91`). 2255 pass, 0 fail, clippy 0 warnings.
- New tests cover `0600`/`0700` perms, symlink refusal (victim file untouched), existing-file no-clobber, and same-slug no-clobber via the retry path.

### DIFF
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -34,6 +34,50 @@ fn sanitize_slug(slug: &str) -> String {
     }
 }
 
+/// Create a directory (and parents) restricted to the current user.
+///
+/// Tee files can contain raw command output including secrets, so the
+/// directory is locked to `0700` on Unix (owner-only). Setting permissions
+/// even when the directory already exists also hardens dirs created by
+/// older versions. No-op beyond `create_dir_all` on non-Unix platforms.
+fn create_dir_private(dir: &std::path::Path) -> std::io::Result<()> {
+    std::fs::create_dir_all(dir)?;
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        std::fs::set_permissions(dir, std::fs::Permissions::from_mode(0o700))?;
+    }
+    Ok(())
+}
+
+/// Write `content` to a NEW file at `path` with owner-only permissions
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the careful writeup on tee file permissions. The recall rework in #3278, merged on 2026-09-08, moved the file tee into src/core/tee_file.rs and made it create its directory 0700 and its files 0600 at creation time through core::utils::create_private_dir and open_private, with a chmod self-heal for files written by older versions, so the world-readable 0644 problem this PR fixes is already gone and is covered by tests including one under a permissive umask. Because that rework replaced the functions this PR patches, the diff can no longer be rebased as written. One part is genuinely still missing on develop: the tee write opens with create+truncate rather than O_EXCL, so a symlink planted at the predictable {epoch}_{slug}.log name would still be followed, and a small PR or issue against the current tee_file path would be very welcome. Closing this one as superseded, and please reopen or comment if you think something else here was missed.

PR #2601, opened 2026-06-25
Handle pytest error summaries

andrewkernel:
## Summary
- teach the pytest output parser to count `error/errors` summary entries separately from failures
- avoid reporting `Pytest: No tests collected` when pytest exits with collection/runtime errors
- add regression coverage for wrapped and quiet-mode pytest error summaries

## Test plan
- [x] `git diff --check`
- [ ] `cargo fmt --all && cargo clippy --all-targets && cargo test` could not be run locally because `cargo` is not installed in this workspace
- [ ] Manual testing: parser behavior covered by the added unit test cases

### DIFF
--- a/src/cmds/python/pytest_cmd.rs
+++ b/src/cmds/python/pytest_cmd.rs
@@ -100,6 +100,7 @@ pub(crate) fn filter_pytest_output(output: &str) -> String {
             && !trimmed.starts_with("ERROR")
             && (trimmed.contains(" passed")
                 || trimmed.contains(" failed")
+                || trimmed.contains(" error")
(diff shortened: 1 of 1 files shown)


KuSh:
Closing in favour of #2399, which makes the same fix to the pytest summary parser and also captures the `=== ERRORS ===` block and labels the short-summary error entries, so the cause of a collection error shows up rather than just its count. It was kept because the whole of this change sits inside it and its tests were run, which the notes here say was not possible in your workspace. Please do review #2399 and comment there if something you covered is missing, or reply here if your scope was different.

PR #2607, opened 2026-06-25
fix(grep): translate --include/--exclude to rg --glob

vegetablechicken233:
## What

Translate grep's `--include=GLOB` / `--exclude=GLOB` into ripgrep's `-g GLOB` / `-g !GLOB` in the rg path, so the fast ripgrep path handles file-type filtering directly.

## Why

ripgrep has no `--include`. Today rtk forwards `--include` to rg verbatim; rg rejects it (exit 2), and the `grep_capture` fallback (#2543 / #2550) re-runs the query through system grep, which understands `--include` natively. That works, but it gives up ripgrep's speed and means a very common grep idiom always takes the slow fallback path. Translating the flag lets the rg path serve it.

## How

- New `translate_grep_globs()` rewrites `--include=GLOB` → `-g GLOB` and `--exclude=GLOB` → `-g !GLOB` (ripgrep negates a glob with a leading `!`); every other arg is forwarded unchanged.
- Applied only where the rg command is built inside `grep_capture`, so both the main and passthrough call sites benefit. The grep fallback path keeps the untranslated originals, so an rg-rejected invocation still works there.
- Only the `=` spelling is handled — the sole form GNU/BSD grep accept for these flags.

## Tests

Unit tests for the translation (include, exclude, mixed-with-passthrough, no-op). `fmt` / `clippy` / `test --all` green locally (new tests + 2240 existing unit tests pass). Verified end-to-end on macOS (Apple Silicon): `rtk grep --include='*.rs'` and `--exclude='*.txt'` now filter correctly through the rg path, and plain searches are unaffected.

## Credit

Builds on the approach proposed in #2131 by @goodmood-creator, reworked for the current `grep_capture` structure — that PR predates the grep refactors (#2333, #2550) and no longer applies cleanly.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -181,6 +181,34 @@ fn strip_rg_only<T: AsRef<str>>(extra_args: &[T]) -> Vec<&str> {
     out
 }
 
+/// Translate GNU grep's `--include=GLOB` / `--exclude=GLOB` into ripgrep's glob
+/// flags so the fast rg path handles them directly instead of erroring out and
+/// falling back to system grep. ripgrep has no `--include`; it filters with
+/// `-g GLOB` (and `-g !GLOB` to exclude). grep's GLOB syntax is rg-glob
+/// compatible, so only the flag spelling changes — the value passes through
+/// unchanged. Every other arg is forwarded verbatim.
+///
+/// Only the `=` form is handled: that is the sole spelling GNU/BSD grep accept
+/// for these flags. The grep fallback path keeps the originals untranslated
+/// (grep understands `--include` natively), so an rg-rejected invocation still
+/// works there. (#2131)
+fn translate_grep_globs<T: AsRef<str>>(extra_args: &[T]) -> Vec<String> {
+    let mut out = Vec::with_capacity(extra_args.len());
+    for arg in extra_args {
+        let a = arg.as_ref();
+        if let Some(glob) = a.strip_prefix("--include=") {
+            out.push("-g".to_string());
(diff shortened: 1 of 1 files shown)


vegetablechicken233:
Hi! Small, self-contained follow-up in the grep area: it translates GNU grep's `--include=GLOB` / `--exclude=GLOB` into ripgrep's `-g` / `-g !` so the fast rg path serves them directly, instead of erroring and falling back to system grep. Builds on the approach in #2131, reworked for the current grep_capture structure. +62/-1, only grep_cmd.rs, fmt/clippy/test --all green locally. Whenever you have a moment I'd appreciate a look — @KuSh since it sits on the grep fallback path you refactored, and @aeppling. Thanks for maintaining rtk!

KuSh:
Hi @vegetablechicken233, thanks for your PR.

Before I review it, if this work is based on someone else’s contributions, please add a Co-authored-by: trailer to the commit message so they’re properly credited.

Example:
```
Co-authored-by: Name <email@example.com>
```

This helps ensure accurate attribution for all contributors. Thanks!

KuSh:
Thanks for this. The rg path this targets is gone: PR #2641 (merged 2026-06-26) removed src/cmds/system/grep_cmd.rs and made RTK run the engine you actually invoked rather than substituting rg for grep, so `rtk grep --include=...` now goes straight to system grep, which handles the flag natively, and search.rs already parses --include/--exclude as value-taking grep flags. That means the --include to -g translation no longer has a code path to sit in. Closing as overtaken, but if you are seeing a case on the current develop where --include or --exclude is still mishandled, please comment and it can be reopened.

PR #2609, opened 2026-06-25
fix(hook): handle AskRewrite in Cursor hook when no rules configured

### LINKED ISSUE #2372: Cursor hook silently drops all rewrites — rtk rewrite exit code 3 treated as failure
### Summary

The Cursor integration (`rtk init --global --agent cursor`) does not rewrite shell commands. Both `rtk hook cursor` and `hooks/cursor/rtk-rewrite.sh` treat `rtk rewrite` exit code 3 as a failure and return `{}` (pass-through), so commands run unchanged and RTK filtering never applies.

This appears to be the same root cause as #2200 (OpenClaw plugin), but it also affects the official Cursor hook shipped with RTK.

### Environment

- **RTK:** 0.42.3
- **OS:** macOS (darwin 25.4.0)
- **Agent:** Cursor IDE (Agent mode, `preToolUse` hook)
- **Hook config:** `~/.cursor/hooks.json` with `matcher: "Shell"` and `command: "rtk hook cursor"`

### Steps to reproduce

1. Install RTK and run:
   ```bash
   rtk init --global --agent cursor
   ```
2. Restart Cursor.
3. Ask the agent to run:
   ```bash
   git status
   ```
4. Observe raw `git status` output instead of RTK-compact output.

### Expected behavior

`git status` should be transparently rewritten to `rtk git status`, and the agent should see compact output like:
```
* branch-name
clean — nothing to commit
```

### Actual behavior

The command runs as plain `git status`. `rtk gain --history` does not record the command. Simulating the hook also returns `{}`:

```bash
printf '%s' '{"tool_input":{"command":"git status"}}' | rtk hook cursor
# → {}
```

### Root cause

`rtk rewrite` returns **exit code 3** when it has a rewrite suggestion, but the hook treats any non-zero exit as failure:

```bash
# hooks/cursor/rtk-rewrite.sh (line 47)
```

Even though stdout contains the correct rewrite (`rtk git status`), the `||` branch runs and discards it.

Verified manually:
```bash
$ rtk rewrite "git status"
rtk git status
$ echo $?
3
```

The script comment says *"rtk rewrite exits 1 when there's no rewrite"*, but successful rewrites also use a non-zero exit code (3), which breaks the hook logic.

### Workaround (confirmed working)

Replace the hook command with a wrapper that captures stdout regardless of exit code:

```bash
REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null || true)

if [ -z "$REWRITTEN" ] || [ "$CMD" = "$REWRITTEN" ]; then
  echo '{}'
  exit 0
fi

jq -n --arg cmd "$REWRITTEN" '{
  "permission": "allow",
  "updated_input": { "command": $cmd }
}'
```

After applying this workaround, `git status` is correctly rewritten and RTK-compact output is returned.

### Suggested fix

1. Fix `rtk rewrite` to exit 0 on successful rewrite (breaking change for callers that rely on exit 3), **or**
2. Update all hook integrations (`hooks/cursor/rtk-rewrite.sh`, `rtk hook cursor`, OpenClaw, etc.) to read stdout on exit code 3 instead of treating it as an error.

Related: #2200, PR #2202

guyoron1:
## Summary

- Fix RTK being completely non-functional on fresh Cursor installs where no explicit permission rules exist in `~/.cursor/cli-config.json`
- When no rules are configured (the default), every command gets `AskRewrite` verdict — but the hook only handled `AllowRewrite`, silently dropping all rewrites
- Now treat `AskRewrite` as allow when no rules are configured (Cursor has no ask-the-user UX); preserve conservative defer behavior when explicit rules exist
- Also fix the legacy shell script (`rtk-rewrite.sh`) which treated exit code 3 (ask) as failure via `||` short-circuit
- Align `cursor_has_explicit_rules()` to include ask rules, matching the inline check in `run_cursor_inner_with_rules()`

Fixes #2372

## Test plan

- [x] Updated unit test: `test_cursor_default_verdict_rewrites` verifies that with no rules, `git status` rewrites to `rtk git status` with `permission: allow`
- [ ] Manual testing: fresh Cursor install with no `cli-config.json` rules — commands should rewrite to `rtk` prefixed versions
- [ ] Manual testing: Cursor with explicit deny/allow rules — conservative behavior preserved

### DIFF
--- a/hooks/cursor/rtk-rewrite.sh
+++ b/hooks/cursor/rtk-rewrite.sh
@@ -40,21 +40,27 @@ fi
 
 # Delegate all rewrite logic to the Rust binary.
 # Exit codes: 0 = allow rewrite, 1 = no rewrite (passthrough),
-#             2 = deny, 3 = ask (Cursor has no ask UX, treat as allow).
+#             2 = deny, 3 = ask.
 REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null)
 RC=$?
-if [ "$RC" -eq 1 ] || [ "$RC" -eq 2 ] || [ -z "$REWRITTEN" ]; then
+if [ "$RC" -eq 1 ] || [ "$RC" -eq 2 ]; then
   echo '{}'
   exit 0
 fi
 
 # No change — nothing to do.
-if [ "$CMD" = "$REWRITTEN" ]; then
+if [ -z "$REWRITTEN" ] || [ "$CMD" = "$REWRITTEN" ]; then
   echo '{}'
   exit 0
 fi
 
-jq -n --arg cmd "$REWRITTEN" '{
-  "permission": "allow",
+# RC 3 = ask (not enforced by Cursor yet, but future-proof).
+PERMISSION="allow"
+if [ "$RC" -eq 3 ]; then
+  PERMISSION="ask"
+fi
+
+jq -n --arg cmd "$REWRITTEN" --arg perm "$PERMISSION" '{
+  "permission": $perm,
   "updated_input": { "command": $cmd }
 }'
--- a/hooks/cursor/rtk-rewrite.sh
+++ b/hooks/cursor/rtk-rewrite.sh
@@ -39,8 +39,14 @@ if [ -z "$CMD" ]; then
 fi
 
 # Delegate all rewrite logic to the Rust binary.
-# rtk rewrite exits 1 when there's no rewrite — hook passes through silently.
-REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null) || { echo '{}'; exit 0; }
+# Exit codes: 0 = allow rewrite, 1 = no rewrite (passthrough),
+#             2 = deny, 3 = ask (Cursor has no ask UX, treat as allow).
+REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null)
+RC=$?
+if [ "$RC" -eq 1 ] || [ "$RC" -eq 2 ] || [ -z "$REWRITTEN" ]; then
+  echo '{}'
+  exit 0
+fi
 
 # No change — nothing to do.
 if [ "$CMD" = "$REWRITTEN" ]; then
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -482,15 +482,14 @@ pub fn run_cursor() -> Result<()> {
         }
     };
 
-    let has_rules = permissions::cursor_has_explicit_rules();
     let output = match decide_hook_action(&cmd, permissions::Host::Cursor) {
         HookDecision::AllowRewrite(rewritten) => {
             audit_log("rewrite", &cmd, &rewritten);
             cursor_allow(&rewritten)
         }
-        HookDecision::AskRewrite(rewritten) if !has_rules => {
+        HookDecision::AskRewrite(rewritten) => {
             audit_log("rewrite", &cmd, &rewritten);
-            cursor_allow(&rewritten)
+            cursor_ask(&rewritten)
         }
         other => {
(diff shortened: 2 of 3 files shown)


hooks/cursor/rtk-rewrite.sh
KuSh: if `RC -eq 3` permission should be "ask". Even if not enforced we're future proof with that

KuSh, line 55: Either you keep this block untouched and remove the condition in the first one, or merge both.

suggestion, keep this one, remove the condition on your new block and add a comment over the condition

src/hooks/hook_cmd.rs
```diff
@@ -482,11 +482,16 @@ pub fn run_cursor() -> Result<()> {
...
 
+    let has_rules = permissions::cursor_has_explicit_rules();
     let output = match decide_hook_action(&cmd, permissions::Host::Cursor) {
         HookDecision::AllowRewrite(rewritten) => {
             audit_log("rewrite", &cmd, &rewritten);
             cursor_allow(&rewritten)
         }
+        HookDecision::AskRewrite(rewritten) if !has_rules => {
+            audit_log("rewrite", &cmd, &rewritten);
+            cursor_allow(&rewritten)
```
KuSh: This seems it should be `cursor_ask(&rewritten)` which set `"permission"`` to "ask"` (again not enforced by cursor ATM but accepted).

I'm not sure the has_rules check is even required with that change

src/hooks/hook_cmd.rs
```diff
@@ -535,8 +540,10 @@ fn run_cursor_inner_with_rules(
     };
 
     let verdict = permissions::check_command_with_rules(&cmd, deny_rules, ask_rules, allow_rules);
+    let has_rules = !deny_rules.is_empty() || !ask_rules.is_empty() || !allow_rules.is_empty();
     match decide_from_verdict(&cmd, verdict) {
         HookDecision::AllowRewrite(rewritten) => cursor_allow(&rewritten),
+        HookDecision::AskRewrite(rewritten) if !has_rules => cursor_allow(&rewritten),
```
KuSh: Same remark, just cursor_ask() without a check. When cursor will enforce asking permission we will be ready

KuSh (changes requested):
Hi @guyoron1, thanks for your PR, minor changes requested before it can be merged.

guyoron1:
thank you for the review : ) 
addressed all issues and tested them

src/hooks/hook_cmd.rs
```diff
@@ -487,6 +487,10 @@ pub fn run_cursor() -> Result<()> {
             audit_log("rewrite", &cmd, &rewritten);
             cursor_allow(&rewritten)
         }
+        HookDecision::AskRewrite(rewritten) => {
+            audit_log("rewrite", &cmd, &rewritten);
```
KuSh:
```suggestion
            audit_log("ask", &cmd, &rewritten);
```

KuSh (changes requested):
A minor change is needed here; otherwise, LGTM, thanks!

guyoron1:
All review feedback has been addressed:

- `cursor_ask()` with `"permission": "ask"` (commit c30b302)
- Audit log uses `"ask"` action (commit 03bb3a3)
- Tests updated to verify ask permission and rewrite behavior

Note: This PR (#2609) was opened **before** issue #2719 and PR #2720 were created, and is a superset of #2720. #2720 only fixes the shell script (`hooks/cursor/rtk-rewrite.sh`), while this PR fixes both the shell script **and** the Rust `run_cursor()` code path — which is the default when using `rtk hook cursor`. If this PR is merged, #2720 would be redundant.

KuSh (approved):

PR #2612, opened 2026-06-25
fix(hook): apply exclude_commands to head/tail rewrites

### LINKED ISSUE #2363: hooks.exclude_commands does not apply to head/tail rewrites

guyoron1:
## Summary

- The head/tail short-circuit in `rewrite_segment_inner` returned a rewrite before reaching the exclusion check, so `hooks.exclude_commands` entries for head/tail were silently ignored
- Add the `is_excluded()` check before the `rewrite_line_range()` call, matching the pattern used by other commands in the same function

Fixes #2363

## Test plan

- [x] Added unit tests: head excluded, tail excluded, cross-exclusion (excluding tail doesn't affect head)
- [ ] Manual testing: add `head` to `hooks.exclude_commands` in config, verify `head -20 file` is no longer rewritten

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1364,6 +1364,9 @@ fn rewrite_segment_inner(
     if context == RewriteContext::Normal
         && (cmd_part.starts_with("head -") || cmd_part.starts_with("tail "))
     {
+        if is_excluded(cmd_part, excluded) {
+            return None;
+        }
         return rewrite_line_range(cmd_part).map(|r| format!("{}{}", r, redirect_suffix));
     }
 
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as a duplicate of #3324 — same fast-path gap, and #3324 will carry `Fixes #2363`.

Seven PRs landed on this one branch. #3324 was picked on shape: it strips the env prefix before matching, so `sudo head -20 f` is covered; this version matches `cmd_part` raw, so it isn't. Nothing wrong with the diff otherwise — it was a coin toss decided on that one line.

PR #2615, opened 2026-06-25
perf(tracking): gate cleanup_old behind 24h interval

### LINKED ISSUE #2208: perf(tracking): enabled tracking adds ~7ms synchronous SQLite write to every command's hot path
## Summary

With tracking enabled, every recorded command (`read`, `ls`, `grep`, `git`, …) spends **~7 ms in `TimedExecution::track()`** on a synchronous SQLite write, while the actual work (read + filter + output) is **<0.05 ms**. The write dominates per-call latency on fsync-expensive filesystems.

Adjacent but distinct from existing work:
- **#1375** — hook-layer cold-start (~56 ms); explicitly scopes standalone-command latency *out*.
- **#1987** (open, not merged) and **#1176** (open) — both touch this write path but only add ways to *disable* tracking (`tracking.enabled` gate / `--no-track`). Neither addresses why tracking is slow *when enabled*, which is what this issue is about.

## Core claim — reproducible on stock rtk, no source changes

`track()` is called by `read`/`ls`/`grep`/… but not by `--version`. Diffing the two isolates the tracking cost from process spawn (stock 0.40.0, best-of-5 × 100):

```
rtk --version (no track): 0.83 ms
rtk read      (+track)  : 7.97 ms
net track cost          : 7.14 ms
```

## PRAGMA sweep — requires the patch in the gist

Net track cost (`read` − `--version`) across journal/sync settings, via an `RTK_JOURNAL`/`RTK_SYNC` env switch added to `Tracker::new()`:

| journal_mode | synchronous | net track cost | DB-corruption risk |
|--------------|-------------|----------------|--------------------|
| WAL | FULL (current) | **7.26 ms** | none |
| WAL | NORMAL | 6.77 ms | none |
| DELETE | NORMAL | 4.53 ms | none |
| TRUNCATE | NORMAL | 4.48 ms | none |
| MEMORY | OFF | 0.29 ms | **corruption on crash/power-loss** |

In-process `Instant` probing puts ~all of `track()`'s time in the SQLite connection-open + INSERT-commit path; `read`/`filter`/`output` stay <0.05 ms every run.

**Repro harness + patch:** https://gist.github.com/coseto6125/19855fc80e113bd5772dbb2fef263077

## Notes / caveats

- **Measured on WSL2**, where fsync is unusually expensive. On native Linux / APFS the absolute numbers are likely lower — independent reproduction on a native FS would be valuable before treating 7 ms as universal.
- Per-call cost is **independent of DB size** (22 MB vs a fresh 52 KB DB → same ~8 ms wall), so it's the per-call open→write→sync cycle, not query/scan cost.
- The persisted data (token-savings stats) is **non-critical and reconstructible** — rtk reads none of it for correctness.

## Why config/PRAGMA tuning isn't enough

`synchronous=NORMAL` is within noise. The only setting reaching sub-millisecond (`MEMORY`+`OFF`) risks DB corruption and is unacceptable. Safe PRAGMA changes cap at ~38%.

## Suggested direction

The order-of-magnitude win needs an architectural change, not a PRAGMA: **keep the synchronous disk write off the hot path** — e.g. buffer to an append-only file and batch-import on `rtk gain`, a background writer thread, or amortize one connection across a session. Since #1987 / #1176 already refactor `record()` / `track()`, folding batched/async writes into that work would close this in the same pass.

guyoron1:
## Summary

Fixes #2208

`cleanup_old()` ran two `DELETE ... WHERE timestamp < ?` scans on every `record()` and `record_parse_failure()` call, adding ~7ms of synchronous SQLite I/O to every RTK command — violating the <10ms startup target.

This PR gates cleanup behind a 24-hour interval using a new `meta` key-value table:
- On each `record()` call, a single `SELECT` on the 1-row `meta` table checks `last_cleanup` (~0.05ms)
- Cleanup only runs if >24 hours have elapsed since the last run
- 90-day data retention means daily cleanup is more than sufficient

**Performance impact:**
- **Before:** ~7ms added to every command (two DELETE table scans)
- **After:** ~0.05ms on 99.9% of calls (one SELECT on PK), ~7ms once per 24h

**Changes:**
- Add `CLEANUP_INTERVAL_HOURS` constant (24h) to `constants.rs`
- Add `meta` table (`key TEXT PRIMARY KEY, value TEXT`) in schema setup
- Replace `cleanup_old()` with `maybe_cleanup()` — checks `last_cleanup` timestamp before running DELETEs
- Use `let _ =` for cleanup calls (INSERT already succeeded; cleanup failure shouldn't break recording)

## Test plan

- [x] `cargo fmt --all` — no issues
- [x] `cargo test --all` — all 2246 tests pass
- [x] New test: `test_maybe_cleanup_skips_when_recent` — verifies old records survive when last_cleanup is fresh
- [x] New test: `test_maybe_cleanup_runs_when_stale` — verifies old records are deleted when last_cleanup >24h ago

### DIFF
--- a/src/core/constants.rs
+++ b/src/core/constants.rs
@@ -4,6 +4,7 @@ pub const CONFIG_TOML: &str = "config.toml";
 pub const FILTERS_TOML: &str = "filters.toml";
 pub const TRUSTED_FILTERS_JSON: &str = "trusted_filters.json";
 pub const DEFAULT_HISTORY_DAYS: i64 = 90;
+pub const CLEANUP_INTERVAL_HOURS: i64 = 24;
 
 /// RTK-only subcommands that should never fall back to raw execution.
 /// When adding a new RTK-only subcommand to `Commands`, add its clap name here.
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -61,7 +61,7 @@ fn project_filter_params(project_path: Option<&str>) -> (Option<String>, Option<
     }
 }
 
-use super::constants::{DEFAULT_HISTORY_DAYS, HISTORY_DB, RTK_DATA_DIR};
+use super::constants::{CLEANUP_INTERVAL_HOURS, DEFAULT_HISTORY_DAYS, HISTORY_DB, RTK_DATA_DIR};
 
 /// Main tracking interface for recording and querying command history.
 ///
@@ -336,6 +336,14 @@ impl Tracker {
             [],
         )?;
 
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS meta (
+                key TEXT PRIMARY KEY,
+                value TEXT NOT NULL
+            )",
+            [],
+        )?;
+
         restrict_db_files(&db_path);
 
         Ok(Self { conn })
@@ -389,13 +397,20 @@ impl Tracker {
             "CREATE INDEX IF NOT EXISTS idx_pf_timestamp ON parse_failures(timestamp)",
             [],
         )?;
+        self.conn.execute(
+            "CREATE TABLE IF NOT EXISTS meta (
+                key TEXT PRIMARY KEY,
+                value TEXT NOT NULL
+            )",
+            [],
+        )?;
         Ok(())
     }
 
     /// Record a command execution with token counts and timing.
     ///
     /// Calculates savings metrics and stores the record in the database.
-    /// Automatically cleans up records older than 90 days after insertion.
+    /// Periodically cleans up records older than 90 days (at most once per 24h).
     ///
     /// # Arguments
     ///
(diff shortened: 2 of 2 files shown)


guyoron1:
Rebased, and reworked: the `meta`-table design can't survive the `user_version` migration gating that landed since. On an existing DB the table would never be created and the gate would silently no-op.

It now gates on `last_insert_rowid()`, which the INSERT above already produced: no extra read, no clock, no schema change, and an exact 1-in-100 cadence, so at most ~100 rows can sit past the 90-day window. That also matches the sampled sweep `hook_decisions` already uses.

Measured here (macOS/APFS, WAL, 5k-row history): 63-99 µs per `record()` with the sweep, 51-89 µs without, so three of the four WAL commits per recorded command go away. I can't reproduce the ~7 ms from #2208 on this machine (that issue's own PRAGMA table points at journal/sync for most of it), so I've dropped that claim from the description.

One small behaviour change is bundled in: the sweep is now best-effort (a warning on stderr) rather than failing `record()`, mirroring `record_hook_decision`. Easy to drop if you'd rather keep that separate.

PR #2616, opened 2026-06-25
fix(git): include branch name in commit output

### LINKED ISSUE #2021: git commit: hook-rewritten output drops the [branch sha] line (cherry-pick unaffected)
## Summary

`rtk`'s hook-rewriting of `git commit` strips the `[branch <sha>] <message>` line from output, losing the new commit's SHA. Other SHA-emitting operations (`git cherry-pick`) are unaffected, so the behavior appears to be `commit`-specific. The diff-stat line is also handled inconsistently — preserved on larger commits, dropped on smaller ones.

## Environment

- `rtk 0.40.0`
- macOS Darwin 25.4.0 (arm64)
- zsh, Homebrew install (`/opt/homebrew/bin/rtk`)
- Invoked via Claude Code's bash hook (auto-rewrites `git` → `rtk git`)

## Reproduction

```bash
tmp=$(mktemp -d) && cd "$tmp"
git init -q && git config user.email t@t && git config user.name t
echo a > a && git add a

# hook-rewritten path
git commit -m "first"
# → ok

echo b > b && git add b

# raw path via rtk proxy
rtk proxy git commit -m "second"
# → [main 77e5cc9] second
#    1 file changed, 1 insertion(+)
#    create mode 100644 b
```

### Expected

Hook-rewritten output preserves the `[branch <sha>] <message>` line — that's the actionable result of the command. Token cost is ~10 chars for the SHA; information value is high (needed for any follow-up `git show`, PR description, revert, push-targeting, etc.).

### Actual

Hook output is `ok`, with the SHA and (sometimes) the diff-stat dropped.

## Inconsistency observed

Same hook path, two commits:

| Commit | rtk output |
|---|---|
| 4 files, +135/-39 (larger commit on a separate project) | `ok 4 files changed, 135 insertions(+), 39 deletions(-)` |
| 1 file, +1 (this minimal repro) | `ok` |

Diff-stat included in the first case, absent in the second. Either threshold-based or coincidental; either way, the SHA is missing from both.

## Negative case (interesting)

`git cherry-pick` through the same hook returns the full `[branch <sha>] <message>` line — identical to `rtk proxy` raw output. So the SHA-stripping isn't a general output-rewriting rule; it's a `commit`-specific summarizer that doesn't realize the `[branch <sha>]` line is load-bearing.

## Why it matters

The new commit SHA is the single most likely thing a caller needs from `git commit` output — to reference in a PR description, hand to `git show`, target a revert, push by SHA, or just confirm the commit landed where intended. An LLM caller (Claude Code) that has lost the SHA either has to fabricate one (`82c9d27`-ish is a real fabrication that just happened in my session) or run a follow-up `git log -1` — which costs more tokens than just preserving the line would have.

## Suggested fix

Preserve the `[branch <sha>] <message>` line verbatim. If the diff-stat is being summarized, keep the SHA line as a separate token-cheap header. Treat `git commit` output the same way `git cherry-pick` is already treated.

guyoron1:
## Summary

Fixes #2021 — `git commit` output was dropping the branch name.

- `parse_commit_output()` now extracts the branch name (first token in bracket content) and includes it in the compact output
- Output changes from `ok abc1234` to `[main] ok abc1234`
- Root commits (`(root-commit)`) and multibyte branch names handled correctly
- All 7 existing unit tests updated and passing

## Before / After

| Scenario | Before | After |
|----------|--------|-------|
| Normal commit | `ok abc1234` | `[main] ok abc1234` |
| Root commit | `ok abc1234` | `[main] ok abc1234` |
| Multibyte branch | `ok abc1234` | `[分支名] ok abc1234` |
| No bracket | `ok` | `ok` (unchanged) |

## Test plan

- [x] `cargo clippy --all-targets` passes (zero warnings)
- [x] `cargo test --all` passes (all 2237+ tests)
- [x] `cargo test parse_commit_output` — all 7 unit tests pass

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1011,10 +1011,16 @@ fn build_commit_command(args: &[String], global_args: &[String]) -> Command {
 fn parse_commit_output(line: &str) -> String {
     if let Some(bracket_end) = line.find(']') {
         let bracket_content = &line[1..bracket_end];
-        let hash = bracket_content.split_whitespace().next_back().unwrap_or("");
+        let parts: Vec<&str> = bracket_content.split_whitespace().collect();
+        let hash = parts.last().copied().unwrap_or("");
+        let branch = parts.first().copied().unwrap_or("");
         if !hash.is_empty() && hash.len() >= 7 {
             let short_hash: String = hash.chars().take(7).collect();
-            format!("ok {}", short_hash)
+            if !branch.is_empty() {
+                format!("[{}] ok {}", branch, short_hash)
+            } else {
+                format!("ok {}", short_hash)
+            }
         } else {
             "ok".to_string()
         }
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased on develop (`git.rs` is `git_cmd.rs` now). The red CI was one test I missed: `test_classify_commit_success_extracts_hash` still expected `ok abc1234`, and it failed identically on all three OSes, with nothing platform- or branch-specific in it. Fixed.

While at it I noticed that taking the first word of git's bracket would print `[detached]` on a detached HEAD, so the label is now everything before the hash up to the root-commit marker (`[detached HEAD] ok abc1234`). I also added an integration test against real git with the branch, locale and config pinned.

One thing for the maintainers: #2021 now asks to keep git's `[branch sha] message` line verbatim, whereas this prints `[branch] ok sha` (no subject or diffstat). I switched `Fixes #2021` to `Refs #2021` for that reason. If the verbatim line is what you want, it's a one-line change in `parse_commit_output`.

PR #2619, opened 2026-06-25
fix(grep): skip redundant header on single-match results

### LINKED ISSUE #2608: grep: spurious "N matches in M files" header on single-match results when the file path is long
**Version:** rtk 0.42.4 (latest `develop`), macOS arm64

`rtk grep` prints the `N matches in M files:` summary header even for a single-match result **when the matched file's path is long** — inflating the output above the plain `file:line:content` form for no benefit. On short paths the header is correctly dropped, so whether the header appears ends up driven by path length rather than result size.

### Repro
```
D=$(mktemp -d)              # long temp path, e.g. /var/folders/.../T/tmp.XXXX
printf 'foo\n' > "$D/file"
rtk grep foo "$D/file"
#   1 matches in 1 files:
#
#   /.../tmp.XXXX/file:1:foo      <- spurious header

printf 'foo\n' > short.txt
rtk grep foo short.txt
#   short.txt:1:foo               <- no header (correct)
```

### Root cause
The never-worse guard (#2554) compares the grouped output against a plain baseline:
```rust
let plain = raw_output.replace('\0', ":");
let output = if rtk_output.len() < plain.len() { rtk_output } else { plain };
```
`rtk_output` renders each path through `compact_path()` (which shortens long paths to `a/.../parent/file`), but `plain` keeps the **full** path. For a long path the compaction savings exceed the header's cost, so `rtk_output` (header included) comes out shorter than `plain` and the header is kept — even though for a single match the header is pure overhead.

### Existing tests already encode the intended behavior
In `tests/grep_compress_test.rs`:
- `small_grep_not_worse_than_plain` asserts the header is **dropped** for a single match — this **currently fails on develop**.
- `large_grep_keeps_header_when_truncated` asserts the header **survives** when the body is genuinely compressed (truncated).

So the fix must satisfy both: keep the header only on a genuine compression, not merely because `compact_path` shortened a long path.

### Fix direction
Compute the never-worse `plain` baseline using the **same compact paths** the grouped body uses, so the header's cost is no longer masked by path compaction.

guyoron1:
## Summary

Fixes #2608 — `rtk grep` emitted a redundant `"N matches in M files:"` header even for single-match results.

- Header is now suppressed when there is exactly 1 match in 1 file (saves tokens)
- Fixed grammar: `"1 matches in 1 files"` → `"1 match in 1 file"` for all remaining cases
- Multi-match and multi-file results continue to show the header with correct pluralization

## Before / After

| Scenario | Before | After |
|----------|--------|-------|
| 1 match, 1 file | `1 matches in 1 files:\n\nsrc/main.rs:42:...` | `src/main.rs:42:...` |
| 3 matches, 1 file | `3 matches in 1 files:\n\n...` | `3 matches in 1 file:\n\n...` |
| 2 matches, 2 files | `2 matches in 2 files:\n\n...` | `2 matches in 2 files:\n\n...` (unchanged) |

## Test plan

- [x] `cargo clippy --all-targets` passes (zero warnings)
- [x] `cargo test --all` passes (all tests)

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -576,11 +576,14 @@ pub fn run(
         .count();
 
     let mut rtk_output = String::new();
-    rtk_output.push_str(&format!(
-        "{} matches in {} files:\n\n",
-        total_matches,
-        by_file.len()
-    ));
+    if total_matches > 1 || by_file.len() > 1 {
+        let match_word = if total_matches == 1 { "match" } else { "matches" };
+        let file_word = if by_file.len() == 1 { "file" } else { "files" };
+        rtk_output.push_str(&format!(
+            "{} {} in {} {}:\n\n",
+            total_matches, match_word, by_file.len(), file_word
+        ));
+    }
 
     let mut shown = 0;
     let mut files: Vec<_> = by_file.iter().collect();


guyoron1:
Superseded by #2641 — the grep refactor now conditionally shows the header only when capping is active.

PR #2624, opened 2026-06-25
fix: misleading copilot init in quick start

GautamKumarOffical:
The quick start section says `rtk init -g` works for "Claude Code / Copilot (default)" but it only works for Claude Code. Copilot needs the `--copilot` flag.

Split it into two separate lines to make it clear.


KuSh:
Closing in favour of #3699, which makes the same Quick Start correction and has a signed CLA. Thanks for catching this one, and please feel free to comment there.

Issue #2614, opened 2026-06-25
rtk grep rejects common grep flags (-r/-E/-A/-i) -> silent raw fallback, 0 savings


voxdei:
`rtk grep` fails to parse the most common real-world grep invocations and silently falls back to raw `grep` (0 tokens saved). Over my last 90 days: **4,416 grep parse failures, 100% fell back to raw.**

The parser rejects standard flags as "unexpected argument":

| Flag rejected | Failures | Typical use |
|---|---|---|
| `-r` (incl. `-rn`, `-rln`) | 3,301 | recursive search |
| `-E` | 333 | extended regex |
| `-A` / `-B` / `-C` | 540 | context lines |
| `-i` | 112 | case-insensitive |
| `-o` | 90 | only-matching |
| `-q` | 85 | quiet |

Example failing commands, all recovered-but-unfiltered:

```
grep -rn TRANSACTION_ID app/services/...
grep -rln 'class.*SyncConfiguration\|sync_client_class' lib/
grep -A10 'gitlab.com:' ~/.config/...
grep -nE 'def client|def sync_reference' lib/
```

**Impact:** these aren't edge cases — they're the dominant grep form. `rtk grep`'s headline 19% avg save rate is entirely composition: simple greps hit 95–98%, but every `-r`/`-E`/`-A` call scores 0. Fixing flag coverage lifts the most-used command in the proxy.

**Ask:** accept (and apply filtering to) the standard GNU grep flag set — at minimum `-r/-R`, `-E`, `-A/-B/-C`, `-i`, `-o`, `-q`, and combined short forms like `-rn`/`-rln`. Failures are logged silently (99.9% "recovery"), so they read as success while saving nothing.

_rtk 0.42.4_

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 87% |

### Summary

rtk grep parser rejects standard GNU grep flags (-r, -E, -A/-B/-C, -i, -o, -q), silently falling back to raw grep with 0 token savings — affects dominant real-world usage patterns

<details>
<summary>📁 Relevant files</summary>

- `src/parser/grep.rs`

</details>

---
[...]

Issue #1604, opened 2026-04-29
Hook rewrite produces unparseable args for native grep flags and non-existent subcommands


charlesDGY:
## Summary

The Claude Code PreToolUse hook (`rtk hook claude`) rewrites Bash commands by prefixing them with `rtk <subcommand>`. Two classes of rewrites consistently fail clap parsing:

1. **Native `grep` flags placed before PATTERN** are rejected by `rtk grep`'s parser even though the help text advertises them as `EXTRA_ARGS`.
2. **Non-existent subcommands** (`ps`, `du`, `make`) get rewritten anyway and fail with `unrecognized subcommand`.

In our environment over 7 days: **836 parse failures, 99.5% recovered via fallback**, but each failure still costs a clap parse cycle and emits multi-line stderr that LLM tooling consumes as input tokens.

Top breakdown (anonymized aggregate from local `parse_failures` table):

| pattern | count | share |
|---------|-------|-------|
| `grep -E PATTERN PATH` (and `-nE`, `-qE`) | ~417 | 49.9% |
| `grep -q PATTERN PATH` | 201 | 24.0% |
| `grep -r PATTERN PATH` | 100 | 12.0% |
| `ps ...` (rewritten to `rtk ps`) | 126 | 15.1% |
| `grep -A/-B/-C N PATTERN PATH` | ~37 | 4.4% |
| `grep -l PATTERN PATH` (parsed as `--max-len PATTERN`) | ~9 | 1.1% |

(Top categories are not mutually exclusive; some commands hit multiple buckets.)

**Important behavioral note for RTK 0.37.2**: when these failures happen, RTK silently logs the failure to `~/.local/share/rtk/history.db` `parse_failures` table and falls back to running the original native command. The user sees only the fallback's output, not the clap parse error. This means the bugs are largely invisible at the terminal but still cost (a) one wasted clap parse cycle per call, and (b) the multi-line stderr that gets recorded is consumed as input tokens by LLM-driven workflows. Below repros demonstrate the bugs by querying the `parse_failures` table directly.

## Repro 1 — grep flag placement (silent fallback, recorded in parse_failures)

```
$ rtk --version
rtk 0.37.2

$ rtk hook check 'grep -E pattern foo.txt'
rtk grep -E pattern foo.txt

# User-visible: rtk grep falls back to native grep, command "works"
$ rtk grep -E pattern /tmp/exists.txt
exit=1   # native grep exit, no clap error visible

[...]

PR #2628, opened 2026-06-25
fix(grep): stop -l/-t shadowing native grep flags

devnulled:
## Summary
- Makes `rtk grep`'s `--max-len` long-only (dropping the `-l` short) and removes `--file-type`/`-t` entirely, so the native grep/rg flags of those letters reach the engine (`src/cmds/system/search.rs`) instead of binding to rtk's own options. `-m` was already fixed on `develop` by #3259 — this finishes the cleanup that 84616d1 started and #3259 continued.
- The `-l` bug had two faces, split on whether the pattern parsed as a number:
  - **Non-numeric pattern** (`rtk grep -l telemetry src/`): clap rejected `telemetry` as a `usize`, `run_fallback` re-ran raw grep — right answer, zero savings. Wasteful, not harmful.
  - **Numeric pattern** (`rtk grep -l 8080 a.txt b.txt`): clap **accepted** it — `max_len=8080`, `a.txt` became the pattern, `b.txt` the only path. No error, no fallback, silently wrong output (and with a single file, a stdin fallback that hangs an interactive terminal). This silent wrong answer, reached through a hook rewrite the user never typed, is the real motivation.
- **User-visible change for `-t`**: `-t TYPE` is rg-only. Before, `rtk grep -t rust PAT` *silently dropped* the flag — it bound to `--file-type`, which was destructured away at the call site and never reached the engine. Now `-t` is forwarded verbatim: `rtk rg -t rust` filters by type as always, while `rtk grep -t rust` surfaces grep's own `invalid option -- 't'`. Loud beats silently-ignored, so `--file-type` is removed outright rather than left as a parse-only no-op.
- Supporting fixes this surfaced:
  - `.claude/hooks/rtk-suggest.sh`: `rg …` now suggests `rtk rg …` (was `rtk grep …`, which swapped ripgrep for GNU grep and broke rg-only flags like `-t`).
  - `scripts/benchmark.sh`: the `-l 40` bench invocation used the removed short; now `--max-len 40` before the pattern.
  - `scripts/test-all.sh`: the three grep smoke assertions were red or inert (missing `-r` made `grep PAT <dir>` exit 2 regardless); rewritten so the `-t` assertion can only fail because of `-t` itself.
  - `docs/usage/FEATURES.md`: drops the advertised `-l`/`-t` shorts, explains why, and removes a stale `--line-numbers | -n` row (the `Grep` variant has no such option).

## Context / how this differs from the other grep PRs
Not the broad "grep rejects `-r`/`-E`/`-A`" problem from #2614 (largely resolved by 84616d1). This is the residual: rtk's own `Grep` short options shadowing native flags. Complements (does not overlap) #2224's rg-only-flag fallback handling; independent of #2532 (`-h` passthrough — deliberately untouched here, see the NOTE in the variant).

## Test plan
- [x] `cargo fmt --all --check` clean, `cargo clippy --all-targets` clean, `cargo test`: **2852 passed; 0 failed; 8 ignored** plus all integration suites green (incl. the two new end-to-end tests in `tests/search_compress_test.rs`).
- [x] Behavior, `-l` (the silent-wrong-answer case):
  ```
  $ rtk grep -l 8080 hit.txt miss.txt   # hit.txt contains "port 8080"
  hit.txt                                # exit 0 — was: empty output, exit 1
  ```
- [x] Behavior, `-t`:
  ```
  $ rtk grep -t rust "fn main" src/main.rs
  grep: invalid option -- t              # exit 2 — was: silently unfiltered results
  ```
- [x] Mutation proof (scratch worktree, compiling mutant): re-adding `#[arg(short = 't', long)] file_type` fails exactly `test_grep_parse_type_t_forwarded` + `test_grep_file_type_option_is_gone` (`15 passed; 2 failed`), and flips the smoke command above to exit 0 — so the new `assert_fails` smoke assertion genuinely fires on regression. Restored tree re-verified green.
- [x] Passthrough unchanged: `-rn`, `-A 3`, `-v`, `-nE 'a|b'`, `--version`, `--` all pinned by new clap-layer tests.

Refs: #2614, #1604, #1436

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -313,46 +313,15 @@ enum Commands {
 
     /// Compact grep - strips whitespace, truncates, groups by file
     Grep {
-        // NOTE: rtk's own tuning options here are intentionally long-only.
-        // Their natural short forms collide with native grep/rg flags of the
-        // same letter, which users routinely pass. A short option here shadows
-        // the native flag and either crashes clap or silently applies the wrong
-        // semantics, defeating the `extra_args` passthrough. Keep them long-only
-        // so native flags reach src/cmds/system/search.rs, the shared filter
-        // behind both `rtk grep` and `rtk rg`. (Continues the `-v`/`-n`/pattern/
-        // path cleanup from 84616d1.)
-        //
-        // This does NOT clear every shadowed short. Clap's auto `-h` still wins
-        // over grep/rg's --no-filename before the first positional, and after it
-        // `-h` reaches search.rs's --version/--help check and execs unfiltered.
-        // Help handling is #2532's scope, not this variant's -- do not "finish"
-        // it here by adding disable_help_flag without reading that PR first.
-        //
-        // Removing a short here also does not make the flag safe end-to-end.
-        // search.rs shares one VALUE_FLAGS_SHORT table across both engines, so a
-        // letter that takes a value in rg but not grep (`-t`, `-T`, `-r`) still
-        // eats the next token inside a cluster: `rtk grep -rt FOO .` consumes
-        // FOO as -t's value regardless of what clap does. That is a search.rs
-        // defect, tracked separately.
+        // rtk's own options here are long-only: a short form shadows the native
+        // grep/rg flag of the same letter and captures it before it can reach
+        // src/cmds/system/search.rs, the shared filter behind `rtk grep` and
+        // `rtk rg`.
         /// Max line length
-        // No short: `-l` is grep's --files-with-matches. Binding it to --max-len
-        // (a `usize`) split behavior on whether the PATTERN parsed as a number.
-        //
-        // Non-numeric pattern: clap rejected it, run_fallback re-ran raw grep,
-        // and the user got the right answer. Wasteful, not harmful -- and note
-        // `-l` is a has_format_flag passthrough on the fixed path too, so no
-        // compaction was "lost"; there is none to lose for this flag.
-        //
-        // Numeric pattern: clap ACCEPTED it. `rtk grep -l 8080 a.txt b.txt` set
-        // max_len=8080 and left a.txt as the pattern, b.txt as the only path --
-        // no error, no fallback, silently empty output and exit 1 on a file that
-        // does match. With a single file it is worse: the filename becomes the
-        // pattern, no path is left, and rtk falls back to reading stdin -- which
-        // exits 1 on /dev/null and hangs outright on an interactive terminal.
-        // That silent wrong answer, reached through a hook the user never opted
-        // into, is the real reason this short had to go.
-        // Pinned by tests::test_grep_parse_files_with_matches_l (clap layer) and
-        // numeric_pattern_with_files_with_matches_flag (end to end).
+        // No short: `-l` is grep's --files-with-matches. Bound to a `usize` it
+        // accepted numeric patterns -- `rtk grep -l 8080 a.txt b.txt` set
+        // max_len=8080, took a.txt as the pattern, and returned empty output
+        // with no error.
         #[arg(long, default_value = "80")]
         max_len: usize,
         /// Max results to show
@@ -367,11 +336,9 @@ enum Commands {
         /// Show only match context (not full line)
         #[arg(long)]
         context_only: bool,
-        // No --file-type / `-t`: `-t TYPE` is rg's type filter, and the option
-        // never did anything under `rtk grep` -- it was destructured away at the
-        // call site and never reached the engine. Removed rather than left as a
-        // parse-only no-op; `rtk rg -t rust` still filters by type, since it
-        // forwards `-t` to real ripgrep.
+        // No --file-type: `-t TYPE` is rg's type filter and the option never
+        // reached the engine, so `rtk rg -t rust` filters by type while
+        // `rtk grep -t rust` surfaces grep's own error.
         /// Pattern, path, and any grep/rg flags (e.g. -v, -i, -A 3, --glob, --version)
         #[arg(trailing_var_arg = true, allow_hyphen_values = true)]
         extra_args: Vec<String>,
@@ -3738,13 +3705,9 @@ mod tests {
 
     // --- grep argument routing (clap layer) ---
     //
-    // The `Grep` variant funnels the pattern, path, and every native grep/rg
-    // flag into a single `trailing_var_arg` slot so `src/cmds/system/search.rs`
-    // -- the shared filter behind both `rtk grep` and `rtk rg` -- can parse them
-    // with full grep/rg semantics. Commit 84616d1 ("fix(grep): stabilize
-    // argument parsing") moved to this design but left the struct's own short
-    // options in place, where they still shadowed native flags of the same
-    // letter. These tests pin the routing.
+    // The `Grep` variant funnels the pattern, path, and every native grep/rg flag
+    // into one trailing slot so `src/cmds/system/search.rs` parses them with full
+    // grep/rg semantics. These tests pin that routing.
 
     /// Parse `rtk grep …` and return the captured `extra_args`, or `None` if
     /// clap rejects the invocation (e.g. a colliding short option mis-binds).
@@ -3754,9 +3717,6 @@ mod tests {
             _ => None,
         }
     }
-
-    // Characterization: behavior that must be PRESERVED (passes today, untested
-    // until now). Guards the 84616d1 design against regression.
 
     #[test]
     fn test_grep_parse_simple_pattern_path() {
@@ -3787,8 +3747,8 @@ mod tests {
 
     #[test]
     fn test_grep_parse_invert_match_v_not_shadowed() {
-        // Regression guard for 84616d1: `-v` (invert-match) must reach grep,
-        // not be captured as rtk's top-level verbose flag.
+        // `-v` (invert-match) must reach grep, not be captured as rtk's
+        // top-level verbose flag.
         assert_eq!(
             grep_extra_args(&["rtk", "grep", "-v", "FOO", "file"]).unwrap(),
             vec!["-v", "FOO", "file"]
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -305,17 +305,25 @@ enum Commands {
...
         /// Max results to show
-        #[arg(short, long, default_value = "200")]
+        #[arg(long, default_value = "200")]
         max: usize,
         /// Show only match context (not full line)
         #[arg(long)]
         context_only: bool,
         /// Filter by file type (e.g., ts, py, rust)
-        #[arg(short = 't', long)]
+        #[arg(long)]
```
KuSh: On current `develop` this turns `-t` from silently-ignored into a hard error, and leaves `--file-type` as a no-op.

`search.rs` (which replaced `grep_cmd.rs`) has `t` in `VALUE_FLAGS_SHORT`, so `-t rust` is forwarded verbatim to GNU grep, which has no such flag:

```
$ rtk grep -t rust TELEMETRY src/core
grep: invalid option -- 't'
```

(verified on a `develop` + this-branch merge build)

Before the change the same command silently returned unfiltered results: `file_type` is destructured as `file_type: _` at the call site (`src/main.rs:2000` on develop) and never reaches `search::run`. So the option was already dead — removing the short just makes the failure loud.

That may well be the outcome you want, since `rtk rg -t rust` works and grep genuinely has no `-t`. If so, `--file-type` should go with it rather than staying as a parse-only option, along with the `file_type == Some("rust")` assertion in `test_grep_long_options_bind_before_pattern` that pins the dead binding.

src/main.rs
```diff
@@ -305,17 +305,25 @@ enum Commands {
...
+        // applies the wrong semantics, defeating the `extra_args` passthrough.
+        // Keep these long-only so native flags reach grep_cmd.rs. (Completes the
+        // `-v`/`-n`/pattern/path cleanup from 84616d1.)
         /// Max line length
-        #[arg(short = 'l', long, default_value = "80")]
+        #[arg(long, default_value = "80")]
         max_len: usize,
         /// Max results to show
-        #[arg(short, long, default_value = "200")]
+        #[arg(long, default_value = "200")]
```
KuSh, line 334: This hunk already landed on `develop` via #3259, which removed the `-m` short with its own rationale comment. As it stands the branch conflicts with `develop` right here, and this side of the conflict deletes that comment.

Please rebase onto `develop` and keep the existing `-m` explanation — only `-l` and `-t` are net-new.

src/main.rs
```diff
@@ -305,17 +305,25 @@ enum Commands {
 
     /// Compact grep - strips whitespace, truncates, groups by file
     Grep {
+        // NOTE: these rtk-tuning options are intentionally long-only. Their
+        // natural short forms (`-l`, `-m`, `-t`) collide with native grep/rg
+        // flags of the same letter (files-with-matches, max-count, type) which
+        // users routinely pass. A short option here shadows the native flag and
+        // either crashes clap (`-l <pattern>` is not a `usize`) or silently
+        // applies the wrong semantics, defeating the `extra_args` passthrough.
+        // Keep these long-only so native flags reach grep_cmd.rs. (Completes the
```
KuSh: `src/cmds/system/grep_cmd.rs` doesn't exist on `develop` — it was replaced by `src/cmds/system/search.rs`, the shared filter behind both `rtk grep` and `rtk rg`. Worth repointing the reference so the comment doesn't name a deleted file.

src/main.rs
```diff
@@ -3430,4 +3438,181 @@ mod tests {
...
+        assert_eq!(
+            grep_extra_args(&["rtk", "grep", "-m", "5", "FOO", "file"]).unwrap(),
+            vec!["-m", "5", "FOO", "file"]
+        );
+    }
+
+    #[test]
+    fn test_grep_parse_type_t_forwarded() {
+        // ripgrep `-t TYPE` (type filter). Must pass through to grep_cmd.rs,
+        // which already handles it, instead of binding to rtk's `--file-type`.
```
KuSh: `search.rs` doesn't handle `-t` for the grep engine — it forwards it to GNU grep, which rejects it with `invalid option -- 't'`.

The test itself is fine (it pins the clap layer, and it passes), but the comment describes runtime behavior that doesn't hold. Something like: "`-t` is rg-only, so it must reach the engine instead of binding to rtk's `--file-type` — under `rtk rg` it filters by type, under `rtk grep` it surfaces grep's own error."

KuSh (changes requested):
Reviewed the `-l`/`-m`/`-t` short-flag removal. The direction is right and the `-l` fix is real, but this needs a rebase, and two of the new comments assert things that aren't true on `develop`.

Verified against `develop` (a93d63b) with this branch applied: all 8 new clap tests pass, `cargo test` is green.

### 1. It conflicts with `develop` — the `-m` hunk already landed
`develop` already dropped the `-m` short from `max` in #3259, with its own rationale comment. This branch conflicts in `src/main.rs` at exactly that block, and its side of the conflict deletes that comment. Only the `-l` and `-t` removals are net-new.

### 2. `grep_cmd.rs` no longer exists
Both the new NOTE block and the `-t` test comment point at `src/cmds/system/grep_cmd.rs`. On `develop` that file was replaced by `src/cmds/system/search.rs`, shared by `rtk grep` and `rtk rg`. The claim that it "already handles" `-t` doesn't hold there.

### 3. `-t` goes from silently ignored to a hard error
On the merged code, `rtk grep -t rust TELEMETRY src/core` now prints `grep: invalid option -- 't'`. Before the change the same command silently returned unfiltered results, because `file_type` is destructured as `file_type: _` at the call site and never reaches `search::run`.

Erroring is arguably the faithful outcome — `rtk rg -t rust` still works, since rg does have `-t` — but it's a user-visible change worth calling out in the PR description. And it leaves `--file-type` as an option that parses and then does nothing.

### 4. Docs still advertise the removed shorts
`docs/usage/FEATURES.md` line 218 (`--max-len` | `-l`) and line 221 (`--file-type` | `-t`) still list the short forms. Line 219 (`--max`) is already correct after #3259.

The `-l` half of the change works exactly as intended: `rtk grep -rl TELEMETRY src/core` now lists filenames through the `has_format_flag` passthrough, where before it died in clap on `--max-len`.

src/main.rs
devnulled: Fixed — rebased onto `develop` (0de9610). The `-m` hunk is out of this PR entirely: #3259's field and its rationale comment come through as untouched context in the diff, and only the `-l`/`-t` removals are net-new. Retitled the PR to `fix(grep): stop -l/-t shadowing native grep flags` to match.

src/main.rs
devnulled: Fixed — the NOTE (and every other comment in the diff) now points at `src/cmds/system/search.rs`, described as the shared filter behind both `rtk grep` and `rtk rg`. `git diff develop...HEAD | grep -c grep_cmd` → 0.

src/main.rs
devnulled: Took the stronger option you suggested: `--file-type` is removed outright — the option, the `file_type: _` destructure at the call site, and the `Some("rust")` assertion that pinned the dead binding are all gone. The PR description now calls out the user-visible change (silently-ignored → grep's own loud error).

Evidence on this branch:

```
$ rtk grep -t rust "fn main" src/main.rs
grep: invalid option -- t        # exit 2
```

Guarded both ways: `test_grep_file_type_option_is_gone` pins that `--file-type rust` routes to `extra_args` instead of binding, and I mutation-checked it — re-adding `#[arg(short = 't', long)] file_type` in a scratch worktree fails exactly that test plus `test_grep_parse_type_t_forwarded` (`15 passed; 2 failed`), and flips the `scripts/test-all.sh` smoke command to exit 0 so its `assert_fails` fires too.

src/main.rs
devnulled: Fixed — the test comment now reads essentially as you phrased it: `-t TYPE` is rg-only, `search.rs` lists `t` in `VALUE_FLAGS_SHORT` and forwards it verbatim — under `rtk rg` it filters by type, under `rtk grep` it surfaces grep's own `invalid option -- 't'` instead of being silently dropped. No more claim that the engine "handles" it.

devnulled:
@KuSh all four findings addressed — rebased onto `develop` (0de9610, `-m` hunk dropped, #3259's comment intact), all `grep_cmd.rs` references repointed to `search.rs`, `--file-type` removed outright with the `-t` behavior change now called out in the description, and the FEATURES.md shorts corrected. Per-finding evidence (including a compiling-mutant proof that the new `-t` guards actually fire) is in the inline thread replies and the updated test plan. I don't have permission to formally re-request review on this repo — ready when you are.

src/main.rs
KuSh: Fixed. `--file-type` is removed outright rather than left as a parse-only no-op, `test_grep_file_type_option_is_gone` pins that, and FEATURES.md now spells out the split. Verified on b3ae7be: `rtk grep -t rust` surfaces grep's own `invalid option -- 't'`, `rtk rg -t rust` still filters by type.

src/main.rs
KuSh: Fixed. Rebased onto `develop` (the PR merges cleanly now) and #3259's `-m` rationale is preserved verbatim.

src/main.rs
KuSh: Fixed. The comment now points at `src/cmds/system/search.rs`, with the note that it is the shared filter behind both engines.

src/main.rs
KuSh: Fixed. The comment now describes the actual forwarding behavior for both engines instead of claiming the flag is already handled.

KuSh (dismissed):
Second pass on b3ae7be. All four earlier threads are properly fixed — the outright `--file-type` removal and the new docs paragraph are exactly the right call. Verified locally: `cargo test --bin rtk grep` 25/25, `search_compress_test` 16/16, fmt and clippy clean, and the numeric-pattern `-l` case now gives the right answer end to end.

KuSh (approved):

PR #2630, opened 2026-06-25
feat(grep): fold shared path prefix in file-list output

israellot:
## Problem

`rtk grep -l` / `-L` / `--files` route through `passthrough()` and print paths **verbatim** — one per line, each repeating the same long directory prefix. When an agent passes an absolute search root (e.g. `grep -ril foo /home/user/project/...`), every line restates `/home/user/project/...`. That shared prefix is the single most strippable part of the output, yet it was forwarded uncompressed and `track_passthrough` recorded **0 tokens saved** for the command.

```
$ rtk grep -ril websocket /home/u/proj/findings /home/u/proj/methodology
/home/u/proj/findings/vpn_login_websocket_hang.md
/home/u/proj/findings/meet_realtime_listen_detect.md
/home/u/proj/methodology/client_server_time_sync.md
...                       # ← prefix repeated on every line, 0% savings
```

## Change

Add `fold_path_prefix` (in `grep_cmd.rs`): factor the longest common directory prefix into a one-line header, then group the remaining tails by sub-directory — the **same `dir/ file1 file2` shape `find_cmd` already emits**, so no new output format is invented (Transparency principle), and the path-grouping logic mirrors existing code (Extensibility principle).

```
/home/u/proj/ (3 files)
findings/ vpn_login_websocket_hang.md meet_realtime_listen_detect.md
methodology/ client_server_time_sync.md
```

Wired into `passthrough()`, which is where the raw-path leak happened (the `-l`/`-L`/`--files` format-flag branch and the unparseable-shape fallback both funnel through it).

### Correctness guarantees
- **Lossless** — every input path is recoverable as `header_prefix + dir + "/" + file`. No truncation, so no `CAP_*` / recovery hint needed.
- **Never-worse guarded** — wrapped in `guard::never_worse`; non-path passthrough output finds no common prefix (`None`) and is emitted verbatim.
- **Component-boundary safe** — the common prefix is computed on whole path components, so `/a/foobar` and `/a/foobaz` share `/a/`, never `/a/fooba`.
- Returns `None` (caller keeps raw) for <2 paths, no shared directory, bare filenames, or a degenerate `/`-only prefix.
- Tracking switched from `track_passthrough` (0/0) to `track` so the savings are actually measured. When nothing folds, in == out — no dilution.

The default grouped-match grep path (`file:line:content`) is **unchanged**.

## Savings

Honest numbers: pure file-lists are bounded by the filenames themselves (irreducible signal), so this lands **~30–55%** depending on prefix depth and file count rather than the 60%+ typical of verbose-output filters — but it converts a measured **0% → positive** on a common agent operation. Examples measured locally:
- Shallow 2-dir case (`-ril`, 5 files): ~50%
- Deep Unity-style abs paths (3 files): 38% (252 → 156 bytes)

## Testing
- 9 new unit tests for `fold_path_prefix` (absolute, relative, sub-dir grouping, component-boundary, single-path/no-common/bare-filename/root-only → `None`, size-shrink invariant).
- `cargo fmt --all --check && cargo clippy --all-targets && cargo test` all pass (2268 passed, 8 ignored).
- grep smoke tests pass; the unrelated smoke failures (curl/wget/pnpm/rewrite/init) are pre-existing and environment-bound (verified identical with this change stashed).
- Manual before/after verified on the motivating `grep -ril` case.

## Docs
Updated `src/cmds/system/README.md` — the line that said `-l`/`-L` "bypass RTK filtering and run raw" now notes the path-prefix folding.

### DIFF
--- a/src/cmds/system/README.md
+++ b/src/cmds/system/README.md
@@ -5,7 +5,7 @@
 ## Specifics
 
 - `read.rs` uses `core/filter` for language-aware code stripping (FilterLevel: none/minimal/aggressive)
-- `grep_cmd.rs` reads `core/config` for `limits.grep_max_results` and `limits.grep_max_per_file`. Format-altering flags (`-c`, `-l`, `-L`, `-o`, `-Z`) bypass RTK filtering and run raw.
+- `grep_cmd.rs` reads `core/config` for `limits.grep_max_results` and `limits.grep_max_per_file`. Format-altering flags (`-c`, `-l`, `-L`, `-o`, `-Z`) bypass the group-by-file filter and run raw, but path-shaped passthrough output (notably the `-l`/`-L`/`--files` file lists) still gets a shared directory prefix folded into a one-line header via `fold_path_prefix` (lossless, `never_worse`-guarded).
 - `local_llm.rs` (`rtk smart`) uses `core/filter` for heuristic file summarization
 - `format_cmd.rs` is a cross-ecosystem dispatcher: auto-detects and routes to `prettier_cmd` or `ruff_cmd` (black is handled inline, not as a separate module)
 
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -1,5 +1,6 @@
 //! Filters grep output by grouping matches by file.
 
+use crate::core::guard::never_worse;
 use crate::core::stream::{exec_capture, CaptureResult};
 use crate::core::tracking;
 use crate::core::utils::{resolved_command, strip_ansi};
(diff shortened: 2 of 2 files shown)


KuSh:
Thank you for this. The file this changes, src/cmds/system/grep_cmd.rs, was removed on develop by #2641 (merged 2026-06-26), which rebuilt grep and rg around a shared Engine model in src/cmds/system/search.rs, and its passthrough path also gained a streaming mode that never buffers the whole output. That means the fold_path_prefix helper and its tests no longer have a place to attach, so I am closing this. The underlying gap is real: file-list passthrough still prints each path in full and records no savings, so an issue or a small PR against search.rs passthrough would be very welcome. If I have misread part of this, please comment or reopen.

israellot:
Thanks @KuSh, that reading is correct. Re-targeted at the `search.rs` passthrough in #4170: same fold, now applied to `grep -l`/`-L`, `rg -l`, and `rg --files` under `never_worse`, with the streaming stdin path left untouched.

PR #2635, opened 2026-06-26
fix(grep): honor GNU -h/--no-filename instead of intercepting it as help

pidefrem:
## Why

GNU/BSD `grep -h` is `--no-filename` — it suppresses the filename prefix. But `rtk grep -h <pattern> <file>` printed rtk's **own `--help` banner** (exit 0, no search): clap intercepted `-h` as help before `run()` ever saw it, and the `--version`/`--help` passthrough also treated `-h` as help. An agent that types `grep -h` got a help screen instead of matches — a silent false negative.

> **Scope change (2026-06-26):** this PR previously proposed a neutral GNU-arg **IR layer** to kill the whole GNU↔ripgrep flag-collision class. Upstream **#2641** ("run the invoked engine instead of substituting rg for grep" — grep runs grep, rg runs rg) landed first and **removed that collision class structurally**: `-E`, `-G`, default ERE, etc. are all now correct on `develop` (verified empirically). The only residual bug from the old scope is this `-h` interception. The PR is reduced to that one focused fix on top of the new engine.

## What

- **Free `-h` in clap** (`main.rs`): `#[command(disable_help_flag = true)]` on `Grep`, with a re-added **long-only** `--help` (`ArgAction::Help`). `rtk grep --help` still shows rtk's help; `-h` now reaches the arg layer.
- **Don't treat `-h` as help** in the `--version`/`--help` passthrough (`search.rs`).
- **Strip `-h`/`--no-filename` from the engine args** before running it: rtk forces `-H --null` for NUL-delimited parsing, so a forwarded `-h` (last-wins) would suppress the filename *at the engine*, break parsing, and drop line numbers. Stripping is letter-wise inside clusters (`-hi` → `-i`; a bare `-h` is dropped). No-filename is then re-applied purely at the **display layer** (`show_file = false`), so output matches `grep -n -h`.

## Files

- `src/main.rs` — free `-h` (`disable_help_flag` + long-only `--help`).
- `src/cmds/system/search.rs` — drop `-h` from the help passthrough; `strip_no_filename()` for the engine args; `show_file` honors no-filename.
- `tests/grep_faithful_format_test.rs` — `rtk grep -h` / `-hi` must equal `grep -n -h` (and never print rtk help).

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets`
- [x] New tests red before the fix (intercepted as help), green after.
- [x] `rtk grep -h <pat> <f1> <f2>` == `grep -n -h …` (position + content, no filename, no help banner).
- [x] `rtk grep -hi <pat> …` honors no-filename **and** ignore-case (cluster stripping).
- [x] `rtk grep --help` still prints rtk's help; existing `search_*`/`grep_*` suites stay green.

## Context

Closes the residual of #2391 (`-h` = `--no-filename`) after #2641 absorbed the rest of the flag-collision class. The obsolete IR-layer commits are dropped from this branch (preserved locally as a backup); the upstream-tracked issue set is unchanged.


pidefrem:
Closing in favour of #3888, which is the better fix for this.

Two things it does that this PR doesn't:

- **grep's last-one-wins precedence between `-H`/`-h`.** My `no_filename` check fires on any `-h` anywhere in the args, so `rtk grep -h -H foo a` suppresses the filename where real grep shows it. #3888 tracks the directives left-to-right in `show_file` and gets `-h -H`, `-H -h` and `-Hh` right, with a test pinning it.
- **`-h` stays help for `rg`.** rg's `-h` genuinely *is* `--help`, so gating on the engine is correct; freeing `-h` for both engines the way I did is wrong for `rtk rg -h`.

Also credit where it's due: #2532 proposed `disable_help_flag` for this five days before I opened this PR.

One thing worth carrying over from here, though — with `disable_help_flag = true` and no long-only `--help` re-added, `--help` reaches the passthrough branch and execs the engine's help, so `rtk grep`'s own options (`--max-len`, `--max`, `--context-only`) stop being discoverable from the CLI. This PR kept them visible with:

```rust
/// Print rtk grep help (long-only; `-h` is GNU grep's --no-filename)
#[arg(long, action = clap::ArgAction::Help)]
help: Option<bool>,
```

That may well be a deliberate call for a proxy (show the wrapped tool's help), but if not, it's a cheap addition to #3888.

Last note, in case it saves someone a debugging session: I stripped `-h`/`--no-filename` from the engine args on the theory that a forwarded `-h` would win over rtk's forced `-H --null` and break NUL-delimited parsing. #3888 doesn't strip and reports byte-identical output, which suggests the forced flags being appended last already settles it and the stripping was unnecessary. Worth a look if `-h` output ever comes back malformed.

PR #2636, opened 2026-06-26
refactor(hooks): inject the permission verdict so rewrite_cmd tests don't read ~/.claude/settings.json

### LINKED ISSUE #2640: bug(tests): rewrite_cmd permission tests read ~/.claude/settings.json — 3/6 fail on a normal dev machine
`src/hooks/rewrite_cmd.rs`'s `unattestable_passthrough` test module calls `evaluate(cmd, …)` → `check_command(cmd)` → `permissions::load_permission_rules()`, which reads the host's real Claude permission rules at test time. The outcome of a command therefore depends on whoever runs the suite:

- `evaluate("git status $(rm -rf /tmp/x)")` expects `Passthrough`, but a machine with a **deny** rule for `rm -rf` returns `Deny` first (deny pre-empts).
- `evaluate("git status")` / `"git status 2>&1"` expect `Ask`, but a machine with an **allow** rule for `git status` returns `Allow`.

So 3 of 6 tests fail locally on any machine with ordinary deny/allow rules, while passing on default-config CI ("green on CI, red on my box").

`evaluate` mixes the permission verdict (I/O — reads host settings) with the pure rewrite decision and offers no seam to inject the verdict — unlike the permission layer's own `check_command_with_rules`.

**Proposal:** extract `evaluate_with_verdict(cmd, verdict, …)` so tests inject a `PermissionVerdict`; production keeps calling `check_command`. Test-only, no behavior change.

Same class as #2106 (telemetry test isolation); the fix mirrors #2603's injection helpers. Fixed by #2636.

pidefrem:
## Problem

`rewrite_cmd.rs`'s `unattestable_passthrough` tests call `evaluate(cmd, …)`, which calls `check_command(cmd)` → `permissions::load_permission_rules()` — reading the **host's real Claude permission rules** at test time. So **3 of 6 tests fail on any machine with ordinary deny/allow rules** (a `deny` for `rm -rf` pre-empts the expected `Passthrough`; an `allow` for `git status` turns the expected `Ask` into `Allow`) while passing on default-config CI — "green on CI, red on my box."

## Root cause + fix

`evaluate` mixes the permission **verdict** (I/O — reads host settings) with the **rewrite/passthrough decision** (pure logic over the command string), with no seam. This extracts `evaluate_with_verdict(cmd, verdict, excluded, transparent)` holding the existing body; `evaluate` becomes `check_command(cmd)` + delegation. Production behavior is byte-identical; tests inject an explicit `PermissionVerdict` (`Ask` neutral for the passthrough/rewrite branch; `Deny`/`Allow` to pin the verdict-dependent branches that were previously untestable without real rules). This mirrors the permission layer's own `check_command_with_rules` injection pattern.

## Files

- `src/hooks/rewrite_cmd.rs` — split out `evaluate_with_verdict`; test module injects the verdict.

## Test plan

- [x] `cargo test --bin rtk rewrite_cmd::tests::unattestable_passthrough` — 8 pass on any machine regardless of `~/.claude/settings.json` (was 3 failing locally).
- [x] +2 tests pin the verdict→outcome mapping: `Deny` → `RewriteOutcome::Deny` (short-circuit), `Allow` → `RewriteOutcome::Allow(_)`.

## Context

The exact bug (rewrite_cmd permission tests reading host settings) is unreported, but it is the same *class* of test-isolation defect upstream is already fixing for telemetry: #2106 (telemetry tests write a real `.device_salt`) and #2603 (adds telemetry test-isolation helpers — the same injection pattern). Filed as a sibling to that work; see also the related issue. Test-only, no production behavior change.

---
Closes #2640.

### DIFF
--- a/src/hooks/rewrite_cmd.rs
+++ b/src/hooks/rewrite_cmd.rs
@@ -46,7 +46,20 @@ enum RewriteOutcome {
 
 fn evaluate(cmd: &str, excluded: &[String], transparent_prefixes: &[String]) -> RewriteOutcome {
     let verdict = check_command(cmd);
+    evaluate_with_verdict(cmd, verdict, excluded, transparent_prefixes)
+}
 
+/// Core of `evaluate`, with the permission `verdict` injected rather than read
+/// from the host's settings. Production passes `check_command(cmd)`; tests pass an
+/// explicit verdict so they don't depend on the machine's real Claude/rtk rules
+/// (which made `git status` / `rm -rf` outcomes environment-dependent — a deny
+/// rule for `rm -rf` or an allow rule for `git status` flipped the result).
+fn evaluate_with_verdict(
+    cmd: &str,
+    verdict: PermissionVerdict,
+    excluded: &[String],
+    transparent_prefixes: &[String],
+) -> RewriteOutcome {
     if verdict == PermissionVerdict::Deny {
         return RewriteOutcome::Deny;
     }
@@ -91,12 +104,23 @@ mod tests {
     }
 
     mod unattestable_passthrough {
-        use super::super::{evaluate, RewriteOutcome};
+        use super::super::{evaluate_with_verdict, RewriteOutcome};
+        use crate::hooks::permissions::PermissionVerdict;
+
+        // These exercise the unattestable-construct / rewrite logic, NOT permission
+        // rules — so they inject a neutral verdict instead of calling check_command
+        // (which reads the machine's real Claude/rtk settings and made outcomes
+        // environment-dependent). `Ask` is the neutral choice: it doesn't short-
+        // circuit on Deny and isn't Allow, so the rewrite-vs-passthrough branch is
+        // what's actually under test.
+        fn eval(cmd: &str) -> RewriteOutcome {
+            evaluate_with_verdict(cmd, PermissionVerdict::Ask, &[], &[])
+        }
 
         #[test]
         fn test_backtick_substitution_passthrough() {
             assert_eq!(
-                evaluate("git status `rm -rf /tmp/x`", &[], &[]),
+                eval("git status `rm -rf /tmp/x`"),
                 RewriteOutcome::Passthrough
             );
         }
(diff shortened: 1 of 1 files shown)


iliaal:
Independent reproduction on `develop` at `f9d8c77`.

Hit this today on a machine with `"Bash(git status:*)"` in `~/.claude/settings.json`. `check_command` returns `Allow`, so `evaluate` yields `Allow(_)` where `test_plain_command_still_rewrites` and `test_fd_dup_redirect_still_rewrites` assert `Ask(_)`.

Isolating `HOME` is enough to flip it, which pins the cause to the settings read rather than anything in the rewrite path. Same invocation, same commit, only the environment differs:

```
$ cargo test --bin rtk unattestable_passthrough
running 6 tests
test result: FAILED. 4 passed; 2 failed

$ HOME=$(mktemp -d) cargo test --bin rtk unattestable_passthrough
running 6 tests
test result: ok. 6 passed; 0 failed
```

I see 2 of the 3 failures you describe, not 3: none of my deny rules match `rm -rf`, so the substitution tests keep passing here. The `Allow`-flips-`Ask` half reproduces exactly as written.

I wrote this fix independently before finding your PR and converged on the same shape: `evaluate_with_verdict(cmd, verdict, excluded, transparent_prefixes)`, `Ask` as the injected neutral verdict, plus a Deny short-circuit test. Applying it takes my full suite from 2 failed to 2497 passed, 0 failed. Dropping mine in favour of this one.

KuSh:
Thanks for the careful diagnosis of the green-on-CI red-on-my-box failures. develop already carries this exact refactor from #3147 (merged 2026-08-16): evaluate() now delegates to evaluate_with_verdict(cmd, check_command(cmd), ...), and the unattestable_passthrough tests pin PermissionVerdict::Default so they no longer read the machine's Claude settings. The two verdict-mapping tests you added are there too, as test_allow_verdict_yields_allow and test_deny_verdict_yields_deny. If any of those tests still fail for you on the latest develop, please comment here or reopen.

Issue #2358, opened 2026-06-10
Config-driven routing for package-manager script commands (e.g. pnpm scripts)


ckir:
**Observed with:** rtk `0.42.3`, Claude Code `2.1.169` (PreToolUse hook with `updatedInput`).

## Summary

The `rtk hook claude` PreToolUse hook auto-rewrites recognized commands to their rtk handler (`git status` → `rtk git status`, `vitest run` → `rtk vitest`, `pnpm exec vitest` → `rtk vitest`). But it does **not** route **package-manager *script* runs** such as:

- `pnpm verify:fast` / `pnpm verify:full`
- `pnpm --filter <pkg> test` / `pnpm test-all`
- `pnpm typecheck-all` / `pnpm lint-all` / `pnpm build-all`

…because the underlying tool (vitest / biome / tsc) is hidden behind the npm-script name, so rtk can't detect it from the command string. In a monorepo these script commands are the **single largest source of uncompressed output** (full vitest/biome/tsc logs).

## What I tried (and why it doesn't solve it)

| Mechanism | Result |
|---|---|
| Custom `filters.toml` `[filters.*]` with `match_command = "^pnpm…"` | `rtk verify` passes, but the **hook never routes** `pnpm verify:fast` through rtk, so the filter only applies on the *explicit* `rtk pnpm …` path — not automatically. |
| `[hooks].transparent_prefixes = ["pnpm verify:fast", …]` | No effect on routing. |
| `[hooks].exclude_commands` | Only *subtracts* from routing; can't *add*. |

Reproduction (hook returns empty = not routed):

\`\`\`bash
printf '%s' '{"tool_name":"Bash","tool_input":{"command":"pnpm verify:fast"}}' | rtk hook claude
# (no output)

printf '%s' '{"tool_name":"Bash","tool_input":{"command":"vitest run"}}' | rtk hook claude
# {"hookSpecificOutput":{"...","updatedInput":{"command":"rtk vitest"}}}
\`\`\`

## Proposed feature

A config-driven route table mapping a command regex to an rtk handler (or a named custom filter), so the hook rewrites matching commands accordingly. For example:

\`\`\`toml
[[hooks.route]]
match   = "^pnpm\\b.*\\b(test|test-all)\\b"
handler = "vitest"        # route output through rtk's vitest failure-extractor

[[hooks.route]]
match   = "^pnpm\\b.*\\bverify:(fast|full)\\b"
[...]

Issue #2268, opened 2026-06-04
Built-in filters for ssh / dig / lsof / scp (custom TOML filters don't reach the PreToolUse hook)


Interstellar-code:
## Summary

Request **built-in** filters for `ssh`, `dig`, `lsof`, and `scp`.

## Why custom filters don't cover this

The Claude Code auto-rewrite hook (`zap hook claude`) only auto-prefixes commands that have a **built-in** filter. Commands handled only by a user-defined `filters.toml` are **not** rewritten by the hook:

```
$ zap hook check "ssh host docker ps"
No rewrite for: ssh host docker ps
```

So custom TOML filters require explicit `rtk ssh ...` invocation (+ `zap trust`) and deliver **zero automatic savings** through the hook. The only path to automatic interception is a built-in filter.

## Evidence these are high-impact

From `zap discover` on a real workflow (server/infra management, 30 days):

| Command | Count |
|---|---|
| ssh | 547 (321 + 117 + 87 + 22 across hosts) |
| dig +short | 44 |
| lsof | 24 |
| scp | 7 |

`ssh` is the single most-used command and is entirely unhandled.

## Suggested behavior

- **ssh**: strip ANSI + login/locale noise (`Warning: Permanently added`, `manpath: can't set the locale`, `Pseudo-terminal will not be allocated`, `Connection to X closed`), then apply the **inner command's** filter when recognizable (e.g. `ssh host "docker ps"` → docker filter), else dedupe/cap. The remote command output is where the tokens are.
- **scp**: strip transfer progress bars / ETA lines; success-only on clean transfer.
- **dig**: keep answer section, drop `;;` comment lines and blanks.
- **lsof**: strip blank lines, cap long tabular output.

## Alternative ask

If built-ins aren't desired, expose a config knob (e.g. `[hooks] include_commands = ["ssh", ...]`) so user TOML filters can opt into the auto-rewrite hook. Today only `exclude_commands` and `transparent_prefixes` exist.

Thanks — great tool.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 87% |

### Summary

[...]

Issue #1684, opened 2026-05-03
Feature: user-defined rewrite rules in config.toml (`[[rules.custom]]`)


LucasSantana-Dev:
## Summary

`config.toml` currently exposes `[tracking]`, `[display]`, `[filters]`, `[tee]`, but no surface for user-defined rewrite patterns. Adding `[[rules.custom]]` would let users prototype rewrites for their own toolchain (and high-frequency misses surfaced by `rtk gain --misses` — see #1683) without forking the binary or waiting for upstream PRs.

## Proposed shape

```toml
[[rules.custom]]
match  = "^my-internal-tool .*"
rewrite = "rtk summary $0"
mode   = "allow"      # allow | ask | passthrough
priority = 100        # lower = checked first; built-ins implicit at 1000

[[rules.custom]]
match  = "^helm get values .*"
rewrite = "rtk summary $0"
mode   = "allow"
```

Semantics:
- Loaded at rtk startup, merged on top of the built-in registry.
- `match` is a regex anchored as the user wrote it.
- `rewrite` supports `$0` (full original cmd), `$1..$N` (regex captures).
- `mode` maps directly to existing exit codes (0 / 3 / 1).
- Built-in registry always wins on conflict unless `priority` is explicitly lower.

## Use cases

1. **Internal/proprietary CLIs** that rtk maintainers can't ship rules for (company-internal `deploy`, `infra-cli`, etc.).
2. **Fast iteration on misses** (#1683) — a user finds `awk` is their #1 unrewritten command and ships a custom rule to wrap it in `rtk summary` while waiting on a proper `rtk awk` filter to land.
3. **Per-project overrides** if a `.rtk.toml` is added later (out of scope here, but custom rules in user config is the building block).

## Security considerations

The big risk: a malicious config could rewrite `git push` to `git push --force` or similar. Mitigations:

- **No shell metacharacters in `rewrite` substitution.** Captured groups are passed as args, not interpolated into a shell string.
- **Never rewrite to a binary the user wouldn't otherwise run.** Or: require `rewrite` to start with `rtk ` (could be a separate strict mode).
- **Surface custom rules in `rtk config show`** so users can audit what's loaded.
- **Document the threat model clearly** — config.toml is already trusted (controls tracking, filters, etc.); custom rules raise the stakes but don't fundamentally change the trust boundary.

## Why this matters

[...]

PR #2637, opened 2026-06-26
feat(config): add [[tools]] per-tool rules (Part 1/3)

pidefrem:
## What

Add a `[[tools]]` section to `config.toml` so per-command behavior can be set at runtime without recompiling. Each rule carries:

- `match` on `command` / `subcommand` / `args_contains` — script-level targeting (e.g. `npm run build` but not `npm run test`).
- `env` — environment variables to inject before spawn.
- `capture` — `pipe` (default) or `pty`.
- `strip_ansi` — sanitize captured output (defaults true for `pty`).

This commit is the **config mechanism only**: parsing, matching (`tool_rule_for`), and `strip_ansi_effective`. The paths that consume these rules land in follow-ups (env injection, pty capture — see below). `tools` is empty by default, so there is **no behavior change** until a user adds a rule.

## Part 1 of a 3-part config series

This is the foundation; two follow-ups consume the struct it introduces:

1. **This PR** — `[[tools]]` config mechanism (parse + match).
2. **feat(config): inject `[[tools]]` env vars before spawn** — wires the `env` field into the runner (fixes the npm/next dev-server output cases).
3. **feat(config): add `capture = "pty"`** — opt-in PTY capture backend, feature-gated.

Parts 2 and 3 will be opened as **draft** PRs stacked on this one; they should be reviewed/merged after this lands. Filing as a series per CONTRIBUTING's multi-part guidance — each part is independently reviewable.

## Files

- `src/core/config.rs` — `ToolRule`, `ToolMatch`, `CaptureMode`, `Config::tool_rule_for`.
- `docs/guide/getting-started/configuration.md` — `[[tools]]` reference.

## Test plan

- [x] 53 config tests pass, incl. `test_tools_rule_parses_and_matches`, `test_tools_match_without_subcommand_matches_any_args`, `test_tools_empty_by_default`.

## Context

No issue describes *this* `[[tools]]` mechanism, but several open issues are demand signals for user-extensible per-command config: #1684 (user-defined rewrite rules in config.toml — closest existing request), #2179 (custom TOML filters never reach the hook), #2358 (config-driven routing for package-manager scripts), #2268 (custom filters don't reach the PreToolUse hook). Adjacent config-surface PRs to resolve against: #2517 (per-group layer config — touches `config.rs`), #1544 (`#[serde(default)]` on partial config sections).


rolstain:
Thanks for landing this foundation — the [[tools]] matching mechanism (command / subcommand / args_contains) looks like a great place to also expose default args injection, which would resolve [#2091](https://github.com/rtk-ai/rtk/issues/2091).

Use case: when RTK is driven by an AI assistant (Claude Code), flags like rtk read -l aggressive and rtk grep --ultra-compact push savings from ~6% / ~50% up to ~80%, but the assistant inconsistently remembers to add them across sessions. Measured impact: ~30 percentage points of efficiency left on the table.

A natural Part 4 (or extension of this PR's ToolRule) could add:

[[tools]]
match = { command = "rtk", subcommand = "read" }
default_args = ["-l", "aggressive"]

[[tools]]
match = { command = "rtk", subcommand = "grep" }
default_args = ["--ultra-compact"]
Semantics: prepend default_args to the user-provided args, with explicit user flags winning on conflict (so rtk read -l minimal foo.ts still wins).

If you'd consider scoping that into the series, happy to refine [#2091](https://github.com/rtk-ai/rtk/issues/2091) to align with whatever shape ToolRule lands on. Either way, thanks for the well-structured 3-part split.

Issue #2606, opened 2026-06-25
rtk rewrites `rg` into `rtk grep` (recursive `grep -rn`), losing ripgrep's .gitignore/binary skipping → CPU spikes on large repos


Daniel-Brestoiu:
> 🤖 **AI-investigated.** This issue was investigated, reproduced, and written end-to-end by an AI agent (Claude Code) on behalf of a human RTK user. Everything below was reproduced locally on a real machine, but please verify independently — treat it as a careful starting point, not authoritative.

## Summary

`rtk rewrite` turns a **ripgrep** search into **`rtk grep`**, and `rtk grep` runs a plain recursive `grep -rn`. Unlike ripgrep, that:

- does **not** honor `.gitignore` / `.ignore`, and
- scans **binary files**.

So `rg <pattern> <dir>` — which a user deliberately chose precisely because it skips `.gitignore`d paths and binaries — gets silently converted into a recursive grep that walks everything. On large repos this descends into `.build/` (Swift Package Manager checkouts containing **binary git pack files**), `node_modules`, build output, etc., and pegs CPU. In agent/automation setups it also defeats "don't use recursive grep" guard hooks, because the user never typed `-r` (RTK adds it).

## Reproduction (rtk 0.42.0, macOS / Apple Silicon)

**1. The rewrite converts ripgrep into rtk grep:**

```
$ rtk rewrite 'rg NEEDLE /path/to/some/repo'
rtk grep NEEDLE /path/to/some/repo
```

**2. `rtk grep` scans what ripgrep correctly skips (self-contained):**

```sh
mkdir -p demo/ignored && cd demo && git init -q
printf 'NEEDLE in tracked\n'    > tracked.txt
printf 'ignored/\n'             > .gitignore
printf 'NEEDLE in gitignored\n' > ignored/cache.txt
printf 'x\0NEEDLE\0\n'          > ignored/blob.bin
```

```
$ rg NEEDLE .
tracked.txt:NEEDLE in tracked          # ripgrep: skips the gitignored dir AND the binary

$ rtk grep NEEDLE .                     # == what RTK rewrites `rg` into
3 matches in 2 files:
  ignored/cache.txt:1:NEEDLE in gitignored   # <- gitignored; ripgrep skipped it
  tracked.txt:1:NEEDLE in tracked
  [+1 more]                                   # <- the binary blob.bin
```

## Impact

[...]

PR #2641, opened 2026-06-26
refacto(search): run the invoked engine instead of substituting rg for grep

aeppling:
# fix(grep): run the invoked engine instead of substituting rg for grep

## Goal

Make `rtk grep` and `rtk rg` a transparent output filter: run the agent's exact search command, compress only its output noise, and report savings against the real command. 

Stop substituting one engine for the other. That substitution fabricated savings, broke regex dialects, flooded node_modules, and mislabeled the metrics DB, and maintenance (patch on patch)

## The problem

`rtk grep` ran ripgrep internally regardless of what the agent typed, then forced flags to make rg act like grep. Every band-aid leaked:

| Symptom | Root cause |
|---|---|
| `grep 'al\|ga'` "saved 137K" but real grep matches 0 | rg `\|` is alternation, grep BRE `\|` is literal, so savings were measured against rg's inflated output |
| `rg ...` rewritten to `rtk grep` ran BSD grep: wrong dialect, recursion, 0-match / stdin hang | the rewrite rule folded `rg` into `rtk grep` |
| one `grep -rn .` recorded 302M tokens "saved" (9.6 min, 1.2 GB) | forced `--no-ignore-vcs` made rg read node_modules and minified files |
| the DB stamps every entry `grep -rn '...'`, so you cannot tell grep from rg | tracking synthesized the command string instead of recording it |

## What changed

- Run the invoked engine, never substitute. A new `Engine { Grep, Rg }`; `engine_capture` runs the exact binary and flags, appending only parse aids (`-n -H -I --null` for grep, `-n --with-filename --null` for rg). They shape the output for grouping; they do not change which lines match.
- Removed every substitution band-aid: the rg-for-grep fallback, the `\|` to `|` translation, the forced `--no-ignore-vcs`, and the rg-only-flag stripping (`strip_rg_only`, `drop_recursion_for_rg`, `RG_ONLY_LONG`, `strip_r` all deleted).
- New `rtk rg` command: runs ripgrep natively (keeping its `.gitignore` and binary skipping and its dialect), reusing the same output filter. grep and rg emit the identical `file:line:content` shape, so the compression is shared, not duplicated.
- Split the rewrite rule (`discover/rules.rs`): `grep` to `rtk grep`, `rg` to `rtk rg`. The engine is preserved end to end.
- Honest tracking: record the agent's real command (`rg -i foo src/`, `grep -rn foo .`) with its actual flags, instead of the synthesized `grep -rn '...'`.
- Renamed `grep_cmd.rs` to `search.rs`: it hosts both engines now, and `automod` auto-declares it, so the rename is a no-op for module wiring.
- Updated `scripts/benchmark.sh`: the grep section now passes `-r` (we no longer force recursion), and a new rg section was added.

## Verification

Local, all CI gates green: `cargo fmt --all -- --check`, `cargo clippy --all-targets`, `cargo test --all` (2247 pass), `cargo build --release`.

Behavior:
- `rtk grep 'al|ga'` is BRE literal (0 matches, honest); `rtk rg 'al|ga'` is alternation (210 matches, what rg was asked for).
- `rtk rg 'X' .` stays lean (no node_modules flood).
- The DB records `rg fn src/core` and `grep -rn fn src/core` (real engine and flags).

Benchmark (`scripts/benchmark.sh`):
- `grep fn` 70611 to 4247 (93%), `grep -c` 95%, `grep -l 40` 99%.
- `rg fn` 70611 to 4247 (93%); `rg -l` and `rg -c` pass through at 0% (already minimal).
- `rtk rg` (0.37s) is faster than `rtk grep` (0.56s) since rg stays lean.

## Tests

Engine faithfulness now has dedicated coverage, each test tagged with the issue it guards:

- `grep_and_rg_use_their_own_regex_dialect` (#2253, #2301, #545): `grep 'al|ga'` stays a literal pipe (0 matches, exit 1); `rg 'al|ga'` is alternation (matches, exit 0).
- `grep_dash_e_selects_extended_regex` (#2253, #2167): `-E` reaches grep as ERE, never read as rg `--encoding`.
- `rg_glob_flag_filters_the_file_set` (#1824, #2167): rg `-g '*.rs'` filters the file set.
- `rg_honors_ignore_files_grep_does_not` (#2606, #2064): rg skips ignored files (no node_modules flood) while grep `-r` descends.
- `bulky_rg_yields_token_savings` and `bulky_grep_yields_token_savings` (#545): savings measured against the real engine output.
- `parse_match_line` suite (#1436, #1613): NUL-separated parse survives content and path colons.

Removed two stale unit tests that documented deleted behavior (BRE-to-rg translation, forced `--no-ignore-vcs`). Adapted develop's `grep_no_match_emits_empty_not_a_message` to pass `-r`, since faithful grep no longer auto-recurses a directory.

Rebased onto develop, mergeable with no conflicts. `cargo fmt`, `clippy --all-targets`, `test --all` (2265 pass) and `build --release` all green.

## Scope kept small (no new edge cases)

`-H --null` is retained as a parse aid so the NUL-based regrouping (#1436) stays unambiguous, and the parser is untouched. Single-file output therefore keeps its small, stable filename column; stripping it would require single-versus-multi detection, a new failure mode, so it is left as-is.

## Issues resolved

- #2606 rg to `rtk grep` loses `.gitignore` and binary skipping, CPU spikes: rg now runs rg
- #2064 grep `-rn` fallback descends node_modules and dumps minified files (token expansion): no `--no-ignore-vcs`, no substitution
- #2253 matches silently rewritten, grep `-r` / `-E` collide with rg `--replace` / `--encoding`: real engine, no flag translation
- #2301 rg to `rtk grep` (BSD grep), path-less rg returns 0 or hangs on stdin: rg to `rtk rg`
- #1824 rg to `rtk grep` breaks rg flag ordering (`--glob`, `-g`): rg to `rtk rg`
- #2167 `rtk grep` calls system grep, breaks `-g` / `--glob`: rg flags now go to `rtk rg`
- #1436 and #1613 not grep-compatible, misparses single-file colons: real grep engine, `-H --null` parse kept

## Open PRs superseded (they patched the substitution; this removes it)

#2183 (respect the invoked tool), #2460 (split the rewrite rule), and #2505 (route rg to `rtk rg`) are implemented in full here. #2138 (translate BRE), #2348 (PCRE-convert for `-E`), #2254 (strip `-r/-R/-E` before rg), #2224 (strip rg-only flags on fallback), #2131 and #2607 (translate `--include` to `--glob`), #2434 (silent fallback when rg flags lead), and #1899 (normalize rewrite flags) are all unnecessary once we run the real engine.


pszymkowiak (approved):

guyoron1:
Hey! Just noting that this refactor also resolved the issue in my #2619 (redundant header on single-match grep results) — the new `capped` logic now conditionally shows the header only when results are actually truncated, which is a cleaner fix than the one I had. Closing #2619 as superseded. Nice refactor!

PR #2643, opened 2026-06-26
fix(diff): rtk diff reports CRLF/LF-only differences as identical

### LINKED ISSUE #2627: rtk diff reports "Files are identical" when files differ only in line endings (CRLF vs LF)
`rtk diff` reports two files as identical when they differ only in line endings. The difference is real — system `diff` detects it — but `rtk diff` discards it and exits 0.

### Version

rtk 0.42.4. Reproduction is byte-controlled (`printf`), so it isn't platform-specific, though I hit it on Windows.

### Reproduction

```bash
printf 'a\r\nb\r\nc\r\n' > x   # CRLF
printf 'a\nb\nc\n'       > y   # LF

rtk diff x y; echo "exit=$?"
# → [ok] Files are identical
# → exit=0

diff x y >/dev/null; echo "exit=$?"
# → exit=1   (system diff sees the difference)

rm x y
```

The files are byte-different (CRLF vs LF terminators) but `rtk diff` reports them identical.

### Where I hit it

Diffing regenerated `insta` `.snap` files on Windows, where the committed files are LF in the index but CRLF in the working tree (`git ls-files --eol` → `i/lf w/crlf`). `rtk diff` reported a snapshot pair as identical even though one had been rewritten LF and the other was CRLF on disk. I only saw the real difference after falling back to `RTK_DISABLED=1 git diff` and `xxd` to inspect the carriage-return bytes.

### What I think is happening (hypothesis — please correct me if I'm reading it wrong)

It looks like `compute_diff` splits both files with `str::lines()`, which treats `\r\n` as a line terminator and drops the `\r` (https://doc.rust-lang.org/std/primitive.str.html#method.lines). A file that differs only in terminators collapses to identical line vectors before any comparison, so `added`/`removed`/`modified` are all 0 and the early return reports identical.

https://github.com/rtk-ai/rtk/blob/8a7dd7e5570d7744d4b6508479a3674fe8c49286/src/cmds/git/diff_cmd.rs#L16-L25

The same `str::lines()` pattern is also used in the unified-diff paths (`condense_unified_diff`, and `compact_diff` for `rtk git diff`), so those may drop carriage-return / trailing-whitespace information too. I haven't verified those paths — just flagging the shared pattern.

https://github.com/rtk-ai/rtk/blob/8a7dd7e5570d7744d4b6508479a3674fe8c49286/src/cmds/git/diff_cmd.rs#L169

https://github.com/rtk-ai/rtk/blob/8a7dd7e5570d7744d4b6508479a3674fe8c49286/src/cmds/git/git.rs#L341

### Related

This reaches the same "Files are identical" symptom as #781, but by a different path: there the lines differ and are detected, but the `modified` count is dropped by the early return; here the difference is erased by `lines()` before any comparison happens, so a fix for #781 wouldn't cover this case. Possibly also relevant to the broader lossy-diff concerns in #1918 and the lossless-mode request in #1313.

amitmishra11:
## Summary
Fixes #2627.

`rtk diff <file1> <file2>` reported two files as identical (exit 0) when they differed only in line-ending bytes (CRLF vs LF) — or, more generally, in anything `str::lines()` normalizes away (it strips `\r` along with `\n`, and treats a missing trailing newline as just "one fewer empty trailing line").

```bash
printf 'a\r\nb\r\nc\r\n' > x   # CRLF
printf 'a\nb\nc\n'       > y   # LF
rtk diff x y; echo "exit=$?"
# before: [ok] Files are identical / exit=0   (wrong — bytes differ, `diff` sees this)
```

## Root cause
`render_file_diff` splits both files with `content.lines()` before comparing (`src/cmds/git/diff_cmd.rs`). `str::lines()` treats `\r\n` and `\n` as equivalent line terminators, so a CRLF file and an LF file with the same text collapse to identical line vectors, and `compute_diff` finds zero changes — the early-return then reports "identical" with no byte comparison.

## Fix
In `render_file_diff`, when the line-based diff finds no changes, fall back to an explicit byte comparison (`content1 == content2`) before declaring the files identical:
- Byte-equal → still "[ok] Files are identical", exit 0 (unchanged behavior).
- Byte-different but line-identical → report the files differ, with a short description of the cause (`describe_byte_only_difference`: CRLF vs LF, trailing newline, or generic), exit 1 (diff convention).

## Scope note
The issue also flags `condense_unified_diff` (this file) and `compact_diff` (`src/cmds/git/git.rs`) as sharing the same `str::lines()` pattern, but says those paths weren't independently verified to reproduce the bug. I scoped this PR to the confirmed, reproducible case (`rtk diff <file> <file>` / `render_file_diff`) rather than changing unverified code paths. Happy to follow up on those separately if confirmed.

## Test plan
- [x] Added regression tests in `src/cmds/git/diff_cmd.rs`: CRLF-vs-LF (reproduces the exact issue), trailing-newline-only diff, and a same-CRLF-on-both-sides case to confirm true identical files still short-circuit correctly. Also unit tests for the new `describe_byte_only_difference` helper.
- [ ] `cargo test` — could not run locally (no working MSVC/64-bit linker in my sandbox; see comment on #2642 for details). The change is isolated to `render_file_diff` and a new pure helper function with no external dependencies; existing tests for that function (`test_render_identical_files_exit_zero`, `test_render_modified_only_yaml_not_identical`, etc.) are untouched and should still pass since the byte-equal fast path is preserved. Appreciate CI confirming.

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -40,7 +40,23 @@ fn render_file_diff(file1: &Path, file2: &Path, content1: &str, content2: &str) 
     let diff = compute_diff(&lines1, &lines2);
 
     if diff.changes.is_empty() {
-        return ("[ok] Files are identical\n".to_string(), 0);
+        // `str::lines()` strips `\r` along with `\n`, so two files whose
+        // lines are textually identical but whose raw bytes differ (CRLF
+        // vs LF terminators, or a missing trailing newline) collapse to an
+        // empty diff here. Fall back to a byte comparison before declaring
+        // the files identical — see #2627.
+        if content1 == content2 {
+            return ("[ok] Files are identical\n".to_string(), 0);
+        }
+        return (
+            format!(
+                "{} → {}\n   Files differ only in {} (no visible line changes)\n",
+                file1.display(),
+                file2.display(),
+                describe_byte_only_difference(content1, content2)
+            ),
+            1,
+        );
     }
 
     let mut rtk = String::new();
@@ -143,6 +159,21 @@ fn compute_diff(lines1: &[&str], lines2: &[&str]) -> DiffResult {
         modified,
         changes,
     }
+}
+
+/// Explains a byte-level difference between two files whose `str::lines()`
+/// output is identical (so `compute_diff` found no changes). Used when the
+/// files are not byte-equal but no line-level diff was detected.
+fn describe_byte_only_difference(content1: &str, content2: &str) -> String {
+    let crlf1 = content1.contains("\r\n");
+    let crlf2 = content2.contains("\r\n");
+    if crlf1 != crlf2 {
+        return "line endings (CRLF vs LF)".to_string();
+    }
+    if content1.ends_with('\n') != content2.ends_with('\n') {
+        return "trailing newline".to_string();
+    }
+    "line endings or trailing whitespace".to_string()
 }
 
 fn similarity(a: &str, b: &str) -> f64 {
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as a duplicate of #3519, which adds the same byte-equality guard before the "identical" verdict.

The bug is confirmed live on 0.46.0:

```
$ printf 'a\r\nb\r\nc\r\n' > x; printf 'a\nb\nc\n' > y
$ rtk diff x y; echo "exit=$?"
[ok] Files are identical
exit=0
```

#3519 is the narrower change (57 lines, one file) and has a full green test matrix; this PR only ever ran `check-target`, which skipped, so it was never CI-verified.

That said, `describe_byte_only_difference` is the better message: #3519 says "whitespace or line endings" generically, where yours names the actual cause (`line endings (CRLF vs LF)` / `trailing newline`). I've noted it for #3519 to fold in with credit to you — a user told two files differ invisibly benefits from being told *how*.

PR #2649, opened 2026-06-26
docs(readme): clarify grep versus ripgrep usage

### LINKED ISSUE #2613: docs: recommend to install ripgrep?
Is rtk more efficient when using `rg` instead of `grep`? If so, I think it should be mentioned/suggested in <kbd>README.md</kbd>.

roian6:
## Summary
- Clarify that RTK preserves the search tool invoked by the user or agent: `grep` routes to `rtk grep`, while `rg` routes to `rtk rg`.
- Document `rtk rg` in the README and usage guide so users know ripgrep is only needed when their workflow uses `rg`.
- Sync the usage docs with the post-#2641 behavior instead of saying `rtk grep` replaces both `grep` and `rg`.

Fixes #2613

## Test plan
- [x] `git diff --check`
- [x] README/docs content probes for `rtk rg`, grep/rg preservation, and no silent grep->ripgrep swap
- [ ] `cargo fmt --all && cargo clippy --all-targets && cargo test` — not run locally; docs-only change and this host does not have `cargo` installed
- [ ] Manual testing: `rtk <command>` output inspected — not applicable; documentation-only clarification

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,9 @@ git status  # Automatically rewritten to rtk git status
 
 Hook-based agents rewrite Bash commands (e.g., `git status` -> `rtk git status`) before execution. Plugin-based agents, including Hermes, use their plugin API to rewrite commands before execution. The agent receives compact output without needing to call `rtk` explicitly.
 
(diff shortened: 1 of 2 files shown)


roian6:
Closing this because the linked issue #2613 was resolved as obsolete after RTK 0.43 stopped rewriting `grep` to `rg`. The documentation change is therefore no longer needed, and keeping the PR open would only add review noise. Thanks for clarifying the current behavior in the issue.

PR #2651, opened 2026-06-27
fix(discover): detect Claude Code hook to flag overstated missed savings

### LINKED ISSUE #2648: discover reports phantom "missed savings" when the Claude Code hook is configured
**Summary:** With the `rtk hook claude` PreToolUse hook configured in `~/.claude/settings.json`, `rtk discover` reports large "missed savings" for commands that were in fact rewritten and proxied through rtk. The numbers are a measurement artifact, not a real leak.

**Mechanism:** Claude Code records the *original* command the model emitted into the session transcript, and *then* the PreToolUse hook rewrites it via `hookSpecificOutput.updatedInput` at execution time. `discover` scans the pre-rewrite transcript text, sees raw `grep -n` / `git log` / `gh api`, and counts each as a miss — it can't see the hook's rewrite.

**Evidence (rtk 0.38.0, Claude Code 2.1.195):**
- `discover` reported `~877K tokens saveable`, "Already using RTK: 126 commands (0.9%)".
- Meanwhile `rtk gain` shows ~181K proxied commands, 226.7M tokens saved (82.4%) — and the counter increments live: running a plain `git status` through the Bash tool ticked it 181,352 → 181,353.
- The hook emits a correct rewrite for every shape: `git status` → `rtk git status`, `cd X && git log` → `cd X && rtk git log`, `echo hi && grep` → `echo hi && rtk grep`.

So `discover` and `gain` disagree by ~99% on the same machine, because they measure different things (transcript text vs. actual execution).

**Suggested fix:** When `discover` detects a configured `rtk hook claude` PreToolUse hook, either (a) skip the "missed savings" report for hook-handled commands, or (b) cross-reference rtk's own execution counter so hook-rewritten commands aren't double-counted as misses.

guyoron1:
## Summary

Closes #2648.

When the Claude Code PreToolUse hook is configured, `rtk discover` scans pre-rewrite transcript text and inflates "MISSED SAVINGS" — reporting up to ~877K tokens saveable while `rtk gain` shows 226M tokens already saved (a ~99% discrepancy).

**Root cause**: `AgentIntegrationStatus` detected Cursor, Hermes, and Copilot hooks but not the Claude Code PreToolUse hook in `~/.claude/settings.json`.

**Fix**:
- Add `claude_hook_installed` field to `AgentIntegrationStatus`
- Detect the hook by parsing `~/.claude/settings.json` for a `PreToolUse` entry with `rtk hook claude` (same pattern as `hook_check.rs::binary_hook_registered()`)
- Append a note to the discover report explaining the overcount and directing users to `rtk gain`

## Test plan

- [x] New test: detects Claude hook from settings.json with correct structure
- [x] New test: returns false when settings.json is missing/empty/has other hooks
- [x] New test: report text includes the warning note when hook is detected
- [x] New test: JSON output includes `claude_hook_installed: true`
- [x] Existing tests pass (15 report tests total, 2237 tests project-wide)
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` clean

### DIFF
--- a/src/discover/report.rs
+++ b/src/discover/report.rs
@@ -1,8 +1,9 @@
 //! Data types for reporting which commands RTK can and cannot optimize.
 
 use crate::hooks::constants::{
-    COPILOT_HOOK_FILE, CURSOR_DIR, GITHUB_DIR, HERMES_DIR, HERMES_PLUGINS_SUBDIR,
-    HERMES_PLUGIN_MANIFEST_FILE, HERMES_PLUGIN_NAME, HOOKS_SUBDIR, REWRITE_HOOK_FILE,
+    CLAUDE_DIR, CLAUDE_HOOK_COMMAND, COPILOT_HOOK_FILE, CURSOR_DIR, GITHUB_DIR, HERMES_DIR,
+    HERMES_PLUGINS_SUBDIR, HERMES_PLUGIN_MANIFEST_FILE, HERMES_PLUGIN_NAME, HOOKS_SUBDIR,
+    PRE_TOOL_USE_KEY, REWRITE_HOOK_FILE, SETTINGS_JSON,
 };
 use serde::Serialize;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into this one. The phantom missed-savings count was fixed by #3206, merged 2026-09-02, which stopped discover from guessing coverage: it logs the real PreToolUse decision per tool call and moves anything the hook actually routed through rtk into the already-RTK bucket instead of counting it as a missed opportunity, falling back to the current hook-install state (the same ~/.claude/settings.json check this PR adds) for older history. The Claude Code hook detection here matches what hooks::hook_check already does, so the report no longer needs the warning note. If you think part of this is still uncovered, please comment and we can reopen.

PR #2652, opened 2026-06-27
fix(diff): remove misleading overflow indicator from condense_unified_diff

### LINKED ISSUE #2653: condense_unified_diff emits misleading '... +N more' after showing all lines
## Summary

`condense_unified_diff` in `src/cmds/git/diff_cmd.rs` prints ALL change lines in full (honoring the comment "Never truncate diff content"), then appends `... +N more` when total changes exceed 10. This implies truncation occurred, but all data was already shown.

## Reproduction

Any `git diff` output with more than 10 change lines in a single file. For example, a file with 100 additions and 100 removals shows all 200 lines, then says `... +190 more`.

## Impact

LLM consumers (Claude Code, Copilot, etc.) read `... +190 more` and assume 190 lines were hidden. They issue follow-up requests to see the "remaining" data — which was already in the output — wasting tokens on redundant fetches.

## Expected Behavior

Since all lines are already shown, no overflow indicator should appear. The `[file]` header already shows the exact count (`+100 -100`).

## Root Cause

Lines 176-182 and 203-210 in `diff_cmd.rs`: the `changes` vec is never truncated (all lines pushed), but the overflow message is emitted based on `total - 10` regardless.

## Environment

- rtk develop branch (current HEAD)
- Affects all platforms

guyoron1:
## Summary

Closes #2653.

`condense_unified_diff` prints ALL change lines in full (honoring the "Never truncate diff content" comment), then appends `... +N more` when total changes exceed 10. This misleads LLM consumers into requesting "remaining" data that was already shown, wasting tokens on redundant fetches.

**Before**: 200 changes printed, then `... +190 more` (implies 190 lines were hidden)
**After**: 200 changes printed, no misleading indicator

## Test plan

- [x] Updated `test_condense_unified_diff_overflow_count_accuracy` → renamed to `test_condense_unified_diff_large_no_false_overflow_indicator`, now asserts NO overflow message appears and verifies all lines are present
- [x] Existing `test_condense_unified_diff_no_false_overflow` continues to pass
- [x] All 23 diff_cmd tests pass
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` clean (2237 tests, 0 failures)

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -400,7 +400,7 @@ diff --git a/b.rs b/b.rs
         assert!(result.is_empty());
     }
 
-    // --- truncation accuracy ---
+    // --- overflow indicator ---
 
     fn make_large_unified_diff(added: usize, removed: usize) -> String {
         let mut lines = vec![
@@ -441,7 +441,7 @@ diff --git a/b.rs b/b.rs
 
     #[test]
     fn test_condense_unified_diff_no_false_overflow() {
-        // 8 changes total — all fit within the 10-line display cap, no overflow message
+        // Counter-case to the 200-change test above: no indicator at small sizes either.
         let diff = make_large_unified_diff(4, 4);
         let result = condense_unified_diff(&diff);
         assert!(
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -176,10 +176,6 @@ fn condense_unified_diff(diff: &str) -> String {
                     for c in &changes {
                         result.push(format!("  {}", c));
                     }
-                    let total = added + removed;
-                    if total > 10 {
-                        result.push(format!("  ... +{} more", total - 10));
-                    }
                 }
                 current_file = line
                     .trim_start_matches("+++ ")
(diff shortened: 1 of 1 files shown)


KuSh (approved):
Approving. Reviewed at depth and found nothing to change in the fix itself.

What I verified:

- The `changes` vec is never capped — every `+`/`-` line is pushed unconditionally and all are emitted — so `total > 10 → "+{total-10} more"` was pure fabrication, always claiming hidden lines that had in fact just been printed. Removing it is the correct fix, not a masking of real truncation.
- `added`/`removed` are still consumed by the `[file] {} (+{} -{})` header, so no dead variables.
- `condense_unified_diff` has one caller (`run_stdin`), and no other module, test or doc references the overflow string. The other `... +N more` occurrences in the repo (`deps.rs`, `pip_cmd.rs`, `summary.rs`, `prettier_cmd.rs`) back real truncation and are untouched.
- Output only shrinks, so `never_worse`'s raw-fallback behaviour is unaffected.
- `cargo fmt --check`, `cargo clippy --all-targets` and `cargo test --all` (2583 unit + integration) are clean on the head.

Confirmed live on 0.46.0 before the change, so this is a real user-facing lie, not a theoretical one:

```
$ rtk diff - < twelve_changes.diff | tail -1
  ... +14 more          # after printing all 24 change lines
```

I pushed one commit to your branch (`7aa6347`): the test module's section header still read `// --- truncation accuracy ---` and `test_condense_unified_diff_no_false_overflow` still mentioned a "10-line display cap". Neither describes the code once the indicator is gone, and leaving them would have re-planted the same misconception this PR removes. Shout if you'd rather word them differently.

Thanks for this one — it's a small diff, but "the filter tells the model that data is missing when it isn't" costs a re-fetch every time, and that's the expensive kind of wrong.

PR #2655, opened 2026-06-27
fix(npm): correct operator precedence in progress indicator filter

### LINKED ISSUE #2654: fix(npm): operator precedence bug in progress indicator filter
## Bug

In `src/cmds/js/npm_cmd.rs:152`, the progress indicator filter has an operator precedence bug:

```rust
if line.contains("⸩") || line.contains("⸨") || line.contains("...") && line.len() < 10 {
    continue;
}
```

Since `&&` binds tighter than `||` in Rust, this parses as:

```rust
line.contains("⸩") || line.contains("⸨") || (line.contains("...") && line.len() < 10)
```

The `line.len() < 10` guard only applies to lines containing `"..."`, not to lines containing `⸩` or `⸨`.

## Impact

Any line containing `⸩` or `⸨` is unconditionally dropped, regardless of length. A legitimate output line like `"Building module (⸨loader⸩) completed successfully"` would be silently filtered out.

Currently works by accident because these Unicode chars rarely appear outside npm progress bars, but the logic doesn't match the intent.

## Fix

Add parentheses to group correctly:

```rust
if (line.contains("⸩") || line.contains("⸨") || line.contains("...")) && line.len() < 10 {
    continue;
}
```

guyoron1:
## Summary

- Fix operator precedence bug in `filter_npm_output` where `&&` binds tighter than `||`, causing lines with `⸨`/`⸩` to be unconditionally dropped regardless of length
- Add regression test verifying long lines with progress indicator chars are preserved

Closes #2654

### DIFF
--- a/src/cmds/js/npm_cmd.rs
+++ b/src/cmds/js/npm_cmd.rs
@@ -149,7 +149,7 @@ fn filter_npm_output(output: &str) -> String {
             continue;
         }
         // Skip progress indicators
-        if line.contains("⸩") || line.contains("⸨") || line.contains("...") && line.len() < 10 {
+        if (line.contains("⸩") || line.contains("⸨") || line.contains("...")) && line.len() < 10 {
             continue;
         }
         // Skip empty lines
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased on current develop; the bug is still there verbatim at `npm_cmd.rs:190`. One change from what #2654 proposes: the parenthesised one-liner comes out at 102 characters and rustfmt won't take it, so the three `contains` calls are hoisted into a named `let` and the length guard stays on the `if`. Same semantics, and the comment now spells out the precedence so it doesn't get tidied back.

On whether this costs compression: it can't. npm only draws the `⸨…⸩` bar when stderr is a TTY, and rtk always captures through a pipe. Checked with npm 10.8.2: a piped `npm install` emits no `⸨` / `⸩` and no CR at all, so that arm is unreachable under rtk either way and this just makes the code say what it meant.

This touches `npm_cmd.rs` like #2860, but only its test module overlaps: whichever merges second needs a trivial rebase (keep `test_progress_indicator_preserves_long_lines`).

PR #2657, opened 2026-06-27
fix(curl): use signal-aware exit code instead of unwrap_or(1)

### LINKED ISSUE #2656: fix(curl): use signal-aware exit code instead of unwrap_or(1)
## Bug

In `src/cmds/cloud/curl_cmd.rs:41`:

```rust
let exit_code = output.status.code().unwrap_or(1);
```

`ExitStatus::code()` returns `None` on Unix when the process was killed by a signal (SIGTERM, SIGKILL/OOM, etc.). The POSIX convention is `128 + signal_number` (e.g. SIGKILL → 137, SIGTERM → 143).

Using `unwrap_or(1)` collapses all signal kills into exit code 1, losing diagnostic information about *why* curl died.

## Impact

Low — masks diagnostic info. Users seeing exit code 1 can't distinguish between "curl reported an error" and "curl was OOM-killed" without additional investigation.

## Fix

The codebase already has `status_to_exit_code()` in `src/core/stream.rs:230` that handles this correctly. Replace the `unwrap_or(1)` with a call to this helper.

guyoron1:
## Summary

- Replace `output.status.code().unwrap_or(1)` with the existing `status_to_exit_code()` helper from `core/stream.rs`
- Signal-killed curl processes (OOM, SIGTERM) now report `128+signal` instead of a generic exit 1

Closes #2656

### DIFF
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -11,6 +11,7 @@
 //! otherwise replace non-UTF-8 bytes with U+FFFD and corrupt the stream
 //! (`#1087`).
 
+use crate::core::stream::status_to_exit_code;
 use crate::core::tee::force_tee_hint;
 use crate::core::tracking;
 use crate::core::utils::resolved_command;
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased on current develop, no conflicts. I added an integration test that runs `rtk curl` against a curl that SIGTERMs itself: it exits 143 with this change and 1 without it (a normal `exit 6` is unchanged). I also checked a real curl killed mid-request: exit 1 before, 143 now.

I kept `status_to_exit_code` since it's the one-liner that was already preferred over the duplicate in #3082; if you'd rather use `exit_code_from_output` from `src/cmds/README.md` (it also prints the `terminated by signal` note), it's a one-line change.

PR #2659, opened 2026-06-27
fix(pnpm): propagate exit code from pnpm outdated

### LINKED ISSUE #2658: fix(pnpm): propagate exit code from pnpm outdated
## Bug

In `src/cmds/js/pnpm_cmd.rs`, `run_outdated()` always returns `Ok(0)` regardless of the underlying `pnpm outdated` exit code.

Unlike `run_list()` and `run_install()` which both check `result.success()` and propagate the exit code on failure, `run_outdated()` skips this check entirely:

```rust
fn run_outdated(args: &[String], verbose: u8) -> Result<i32> {
    // ...
    let result = exec_capture(&mut cmd).context("Failed to run pnpm outdated")?;
    // ← no result.success() check here
    // ...
    Ok(0)  // ← always reports success
}
```

## Impact

1. **Failure masking**: If `pnpm outdated` fails (network error, auth failure, registry down), RTK swallows the error and reports success.
2. **CI/CD breakage**: `pnpm outdated` returns exit 1 when outdated packages are found — CI/CD pipelines that check exit codes see success through RTK when they should see exit 1.

## Nuance

`pnpm outdated` returns exit 1 for two distinct reasons:
- **Outdated packages found** (normal data, empty stderr) — RTK should still filter/format the output
- **Actual error** (non-empty stderr) — RTK should bail early and surface the error

## Fix

Add a stderr-guarded failure check (bail early only on real errors) and propagate the exit code:

```rust
if !result.success() && !result.stderr.is_empty() {
    eprint!("{}", result.stderr);
    return Ok(result.exit_code);
}
// ...
Ok(result.exit_code)
```

guyoron1:
## Summary

- Add stderr-guarded failure check to `run_outdated()` — real errors (network, auth) now bail early and surface the error message
- Propagate exit code instead of hardcoded `Ok(0)` — CI/CD pipelines see the correct exit 1 when outdated packages are found
- Normal "outdated found" path (exit 1, empty stderr) still flows through RTK's JSON parsing and table formatting

Consistent with how `run_list()` and `run_install()` handle failures in the same file.

Closes #2658

### DIFF
--- a/src/cmds/js/pnpm_cmd.rs
+++ b/src/cmds/js/pnpm_cmd.rs
@@ -432,6 +432,12 @@ fn run_outdated(args: &[String], verbose: u8) -> Result<i32> {
     }
 
     let result = exec_capture(&mut cmd).context("Failed to run pnpm outdated")?;
+
+    if !result.success() && !result.stderr.is_empty() {
+        eprint!("{}", result.stderr);
+        return Ok(result.exit_code);
+    }
+
     let combined = result.combined();
 
     // Parse output using PnpmOutdatedParser
@@ -467,7 +473,7 @@ fn run_outdated(args: &[String], verbose: u8) -> Result<i32> {
 
     timer.track("pnpm outdated", "rtk pnpm outdated", &combined, shown);
 
-    Ok(0)
+    Ok(result.exit_code)
 }
 
 fn run_install(args: &[String], verbose: u8) -> Result<i32> {


guyoron1:
Rebased on current develop and reworked the guard. On pnpm 9/10 the `ERR_PNPM_*` errors are printed on stdout, so the empty-stderr check never fired: a missing package.json or a down registry was still parsed into a fake "outdated package", and a Node warning on stderr would have hidden a real list.

The rule is now "exit != 0 and stdout is not the JSON map": the error is shown untouched, and a real outdated list keeps its summary and exits 1. This also covers the `pnpm -r outdated` swallow from the #2658 thread. Checked with pnpm 10.33: outdated, up to date, no manifest, registry down.

Heads-up: #3067 restructures `run_outdated` too, so whichever of the two lands second will conflict there.

PR #2661, opened 2026-06-27
fix(pnpm): propagate exit code from pnpm outdated

### LINKED ISSUE #2658: fix(pnpm): propagate exit code from pnpm outdated

WaterXiao-git:
pnpm outdated always returned Ok(0) regardless of the underlying command result, masking failures like network errors or auth issues. Also, pnpm returns exit 1 for legitimate 'packages are outdated' results (with empty stderr), so we only bail on real errors (non-empty stderr).

Closes #2658

### DIFF
--- a/src/cmds/js/pnpm_cmd.rs
+++ b/src/cmds/js/pnpm_cmd.rs
@@ -432,6 +432,15 @@ fn run_outdated(args: &[String], verbose: u8) -> Result<i32> {
     }
 
     let result = exec_capture(&mut cmd).context("Failed to run pnpm outdated")?;
+
+    // Propagate exit code from pnpm outdated unless it's a legitimate "packages
+    // are outdated" exit (exit 1 with empty stderr, which pnpm uses to signal
+    // outdated packages rather than an actual error).
+    if !result.success() && !result.stderr.is_empty() {
+        eprint!("{}", result.stderr);
+        return Ok(result.exit_code);
+    }
+
     let combined = result.combined();
 
     // Parse output using PnpmOutdatedParser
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as a duplicate of #2659, which makes the same change to run_outdated in pnpm_cmd.rs and has its CLA signed. Thanks for the fix, and please take any further comments to that PR.

Issue #2094, opened 2026-05-26
bug: hook rewrite of `pnpm lint` discards package.json script body — same false-green as #185 at the hook layer


ron332:
## Summary

The PreToolUse hook rewrites `pnpm lint` → `rtk lint`, discarding the body of the project's `package.json` `lint` script. For projects whose script enforces flags `eslint` doesn't default to (e.g. `--max-warnings 0`) or chains commands (e.g. `pnpm check:foo && eslint ...`), the rewritten `rtk lint` invokes `eslint` with defaults and exits 0 even when the original `pnpm lint` would have exited 1.

This is the same false-green symptom as #185 (closed) but originates at the hook-rewrite layer, not inside `rtk lint`. `rtk lint --max-warnings 0` exits 1 correctly — #185's fix still holds for explicit invocations.

## Affected version

- `rtk 0.42.0` (Homebrew, macOS arm64)
- Confirmed in both the legacy shell-script hook (`~/.claude/hooks/rtk-rewrite.sh` produced by older `rtk init`) and the current in-binary hook (`rtk hook claude`).

## Reproduction

`package.json`:

```json
{ "scripts": { "lint": "eslint --max-warnings 0" } }
```

`a.ts` (one unused-import warning):

```ts
import { randomUUID } from 'node:crypto';
export const x = 1;
```

```bash
# What CI runs (raw pnpm script):
pnpm lint a.ts; echo "raw=$?"
# raw=1   ← eslint --max-warnings 0 fails the run

# What Claude Code runs (hook rewrites pnpm lint → rtk lint):
rtk hook check "pnpm lint a.ts"
# → rtk lint a.ts
rtk lint a.ts; echo "hooked=$?"
# hooked=0  ← false green; --max-warnings 0 was dropped
```

For confirmation that `rtk lint` itself is fine when flags are explicit:

```bash
rtk lint --max-warnings 0 a.ts; echo $?
# 1   ← propagation is correct when the flag is passed
```

## Impact

Reaches every user who hasn't typed `rtk lint` themselves. The hook makes `pnpm lint` the *de facto* path through which agentic CLI tools (Claude Code, Cursor, etc.) validate before pushing. Local pre-push reports green; CI fails on the dropped flag.

[...]

PR #2664, opened 2026-06-27
fix(npm): add 15 missing subcommands to prevent incorrect `run` injection

### LINKED ISSUE #2663: Bug: `npm query`, `npm sbom`, and 13 other subcommands get incorrect `run` injection
## Summary

## Reproduction

```bash
rtk npm query '#dependencies'

rtk npm sbom

rtk npm completion
```

## Root Cause

| Subcommand | Purpose |
|---|---|
| `ll` | Verbose `ls` |
| `org` | Manage orgs |

## Impact

## Fix

guyoron1:
## Summary

Fixes #2663

- Add 15 missing official npm subcommands to `NPM_SUBCOMMANDS` list
- Without these, RTK incorrectly injects `run` (e.g. `npm query` → `npm run query` → fails)
- Add regression test for hyphenated subcommands (`find-dupes`, `help-search`, etc.)

**Missing subcommands added:** `completion`, `edit`, `explore`, `find-dupes`, `help-search`, `hook`, `install-ci-test`, `install-test`, `ll`, `org`, `query`, `run-script`, `sbom`, `shrinkwrap`, `unstar`

## Test plan

- [x] All existing npm_cmd tests pass
- [x] New `test_hyphenated_subcommands_no_run_injection` covers compound subcommands
- [x] `cargo fmt && cargo clippy --all-targets && cargo test --all` clean


aeppling:
Hey @guyoron1 

Root-cause concern: the logic injects run for any unrecognized first arg, which makes NPM_SUBCOMMANDS a fail dangerous denylist. Issue is the default "run" prefix.
Even with this PR, npm's aliases still break, npm add, npm x, npm ln, npm un, npm s, so #2663 isn't fully closed, and every future npm release reopens it.

The self-maintaining fix: only inject run when the first arg is an actual script in the nearest package.json (walking up like npm does); otherwise pass through.

Should cause no regression keep your list additions: their job becomes builtin-precedence (so a query/sbom script can't shadow the real subcommand).


We can merge this but this is just a patch that does not fix the root issue.
Ball on your side for this, we can still merge this as a patch for now.

reneleonhardt:
Great review, thank you!
@guyoron1 could you add a corresponding test for `npm run`, and then fix it?

src/cmds/js/npm_cmd.rs
```diff
@@ -205,63 +220,50 @@ npm notice
         assert!(result.contains("Build completed"));
     }
 
+    fn args(strs: &[&str]) -> Vec<String> {
+        strs.iter().map(|s| s.to_string()).collect()
```
reneleonhardt: More idiomatic:

```rust
strs.iter().map(String::from).collect()
```

src/cmds/js/npm_cmd.rs
```diff
@@ -101,11 +113,14 @@ pub fn run(args: &[String], verbose: u8, skip_env: bool) -> Result<i32> {
     if is_run_explicit || is_npm_subcommand {
         effective_args.extend_from_slice(args);
     } else {
-        // "rtk npm build" → "npm run build" (assume script name)
         effective_args.push("run".to_string());
         effective_args.extend_from_slice(args);
     }
```
reneleonhardt: This could be simplified:

```rust
if !(is_run_explicit || is_npm_subcommand) {
    effective_args.push("run".to_string());
}
effective_args.extend_from_slice(args);
```

guyoron1:
Heyaaaa :) thanks for the review!!

Pushed a fix that addresses both points:

1. **`npm run` test added** — `test_npm_run_no_double_injection` verifies `npm run build` passes through without injecting a second `run`

2. **Missing aliases fixed** — Added 11 short npm aliases (`add`, `x`, `ln`, `un`, `r`, `s`, `se`, `rb`, `ddp`, `it`, `cit`) that were still getting `run` injected

Also refactored the tests to call the actual production function (`build_effective_npm_args()`) instead of reimplementing the logic in local test helpers — follows the same pattern as `dotnet_cmd.rs`.

src/cmds/js/npm_cmd.rs
```diff
@@ -86,11 +86,23 @@ const NPM_SUBCOMMANDS: &[&str] = &[
...
+    "it",
+    "cit",
 ];
 
-pub fn run(args: &[String], verbose: u8, skip_env: bool) -> Result<i32> {
-    // Determine if this is "npm run <script>" or another npm subcommand (install, list, etc.)
-    // Only inject "run" when args look like a script name, not a known npm subcommand.
+/// Build the effective npm args, injecting "run" when the first arg looks
+/// like a script name rather than a known npm subcommand or flag.
+fn build_effective_npm_args(args: &[String]) -> Vec<String> {
```
reneleonhardt, line 115: But this is still guessing, right?
Not actually walking up to the nearest package.json and extracting script keys like npm does.
Maybe something like this:

```rust
let script_names: Vec<String> = env::current_dir()
    .ok()
    .and_then(|p| p.ancestors().map(|a| a.join("package.json")).find(|f| f.exists()))
    .and_then(|f| File::open(f).ok())
    .and_then(|f| serde_json::from_reader(f).ok())
    .and_then(|v| v.get("scripts")?.as_object().map(|m| m.keys().cloned().collect()))
    .unwrap_or_default();
```

reneleonhardt:
What's the source for NPM_SUBCOMMANDS?

When I diff it against `npm help` (npm 11.17.0 node 24.12.0), I get.


### All commands
* approve-scripts
* deny-scripts
* run
* stage
* trust
* undeprecate

### NPM_SUBCOMMANDS
*   bin
*   create
*   home
*   hook
*   info
*   list
*   remove
*   rm
*   run-script
*   show
*   up
*   why

(and i t aliases).

src/cmds/js/npm_cmd.rs
guyoron1: Heyaaaa! Thanks for the review :)

Yes, it's still heuristic-based — the subcommand list covers npm's built-in commands, so anything not in the list gets `run` injected. Walking `package.json` for script keys would be more accurate but adds I/O on every invocation and complexity around monorepos / workspaces. Happy to explore that as a follow-up if there's appetite for it.

aeppling:
Hey @guyoron1 same update as for #2776 

### This is related to issues #2663 and #2094

Two related PR that both address a temporary fix : 
- #2664 narrows/extends NPM_SUBCOMMANDS, an allowlist of when injection is safe.
- #2776 narrows the lint regex + rewrite_prefixes, an allowlist of when substitution is safe (keep eslint/npx eslint/pnpm dlx eslint, drop pnpm lint/npm run lint).

The issue is at a higher level, and maintaining those hard coded lists is not a correct solution.

### Proposed fix : 

- npm: forward the args untouched. npm build then fails exactly as it would without RTK, that's the agent's own command, and it self-corrects to npm run build. npm query, npm add, npm sbom, and anything npm adds in future just work. NPM_SUBCOMMANDS gets deleted.
- lint: run the real pnpm lint (the actual script, with its real flags, chain, and exit code) and route its output through the lint filter, don't rebuild an eslint invocation. The invocation allowlist gets deleted.

This fixes both issues at the root, removes both lists (no more chasing npm versions or package-manager forms), and keeps every exit code faithful.

#2776 currently drops pnpm lint to raw with no filter (0% savings), running it verbatim and filtering the output preserves the gain.

guyoron1:
Closing this in favour of #2860. The 15 subcommands here landed separately in #3087, and the aliases I added afterwards (`add`, `x`, `ln`, `un`, `r`, `s`, `se`, `rb`, `ddp`, `it`, `cit`) are covered there too: #2860 deletes `NPM_SUBCOMMANDS` outright and forwards npm's argv untouched, so every alias works, plus the npm 11 commands @reneleonhardt listed (`approve-scripts`, `deny-scripts`, `stage`, `trust`, `undeprecate`) that this PR never added. Checked against a stand-in npm on both builds: develop still injects `run` for 16 of them, #2860 for none.

Thanks @aeppling and @reneleonhardt for pushing this towards the root cause.

PR #2666, opened 2026-06-27
fix(git): recognize -n<N> combined form in git log limit detection

### LINKED ISSUE #2665: git log -n<N> combined form silently truncated to 10 commits
## Description

`rtk git log -n20` silently truncates output to 10 commits instead of 20. The combined `-n<N>` form (e.g. `-n20`, `-n5`) is not recognized by either the `has_limit_flag` detection or `parse_user_limit()`.

## Steps to Reproduce

```bash
rtk git log -n20    # shows 10 commits instead of 20
rtk git log -n 20   # works correctly (shows 20)
rtk git log -20     # works correctly (shows 20)
```

## Root Cause

In `src/cmds/git/git.rs`, the `has_limit_flag` check only detects:
- `-<digit>` (like `-20`)
- `-n` as a standalone token (for the `-n 20` two-token form)
- `--max-count` prefix

It does **not** detect `-n<digits>` (e.g. `-n20`), which is a valid and commonly used git syntax.

When `-n20` is not detected as a limit flag:
1. RTK injects its own `-10` limit → git uses the last limit (`-n20`) and outputs 20 commits
2. RTK sets `limit = 10` and `user_set_limit = false` in `filter_log_output()`
3. The filter truncates to 10 commits → **user loses half their requested output**

The same gap exists in `parse_user_limit()`, which also doesn't handle the `-n<N>` combined form.

## Expected Behavior

`rtk git log -n20` should show 20 commits, identical to `rtk git log -n 20` and `rtk git log -20`.

guyoron1:
## Summary

- `rtk git log -n20` silently truncated to 10 commits because the combined `-n<N>` form was not recognized by `has_limit_flag` or `parse_user_limit()`
- Added detection for `-n<digits>` in both the limit flag check and the parser
- Added regression test for the combined form

## Root Cause

`has_limit_flag` checked for `-<digit>` (like `-20`), `-n` (standalone), and `--max-count`, but not `-n<digits>`. When undetected, RTK injected its own `-10` limit and set `filter_log_output` to truncate at 10 commits — silently losing half the user's requested output.

## Test plan

- [x] New test `test_parse_user_limit_n_combined` verifies `-n20` parses to `Some(20)`
- [x] Existing tests for `-20`, `-n 15`, `--max-count=30`, `--max-count 25` still pass

Fixes #2665

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -441,10 +441,11 @@ fn run_log(
         arg.starts_with("--oneline") || arg.starts_with("--pretty") || arg.starts_with("--format")
     });
 
-    // Check if user provided limit flag (-N, -n N, --max-count=N, --max-count N)
+    // Check if user provided limit flag (-N, -n N, -nN, --max-count=N, --max-count N)
     let has_limit_flag = args.iter().any(|arg| {
         (arg.starts_with('-') && arg.chars().nth(1).is_some_and(|c| c.is_ascii_digit()))
             || arg == "-n"
+            || (arg.starts_with("-n") && arg[2..].bytes().next().is_some_and(|b| b.is_ascii_digit()))
             || arg.starts_with("--max-count")
     });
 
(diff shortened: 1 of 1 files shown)


guyoron1:
Checked against current develop: `rtk git log -n20` now returns 20 commits (it returned 10 before the arg tokenizer in #3681), so the code fix here is obsolete and I've dropped it. Nothing on develop pins the glued `-n<N>` form though, so this is now a tests-only PR (retitled, and `Refs` instead of `Fixes`). The tests fail if the bug is reintroduced. Feel free to close it if you'd rather not carry the extra tests.

#2665 itself is fixed on develop and can be closed. #4148 and #4152 also edit `tests/git_log_cap_notice_test.rs`, so expect a small textual conflict in that file for whichever lands second.

PR #2667, opened 2026-06-27
feat(hook): signal/bulk passthrough for control-marker commands

### LINKED ISSUE #2121: Built-in signal/bulk classifier for git/gh control-marker commands
## Problem

`rtk hook claude` (and the equivalent shim/wrapper paths) currently rewrite the stdout of **every** Bash tool invocation, including commands whose output is a small control signal that downstream agents grep against canonical markers. This is fine for bulk output (`ls -la`, `find`, `git log -p`, `docker logs`) but actively harmful for short signal output: rtk shape-changes or inflates the canonical markers, and agents that gate next-step decisions on those markers misclassify successful operations as hung or failed.

Concrete measurements (rtk 0.40.0, macOS, our agent harness):

| Command | raw bytes | after `rtk hook claude` | delta |
|---|---|---|---|
| `git status` (clean) | 59 | 123 | +108% |
| `git log --oneline -50` | 59 | 4144 | +6924% |
| `git push origin <branch>` (success) | ~80 | canonical `To <repo>\n  abc..def main -> main` line truncated/reshaped | marker lost |

The user-visible incident this surfaces in: an agent runs `git push origin <branch>`, the canonical `To <repo>` marker is gone from stdout, the agent treats the push as hung, and re-runs the command in the background for 15+ minutes when in reality the first push succeeded.

The class of affected commands is small and stable: SCM client status/markers (`git push|pull|fetch|merge|status|remote|rev-parse|branch`), forge clients (`gh pr|issue|release|api|run`, `glab mr|issue`), and a handful of jq-piped stdout consumers.

## Proposal

A built-in **signal vs bulk** classifier:

1. Ship a default allowlist of substring patterns whose stdout bypasses `rtk hook ...` (raw stdout reaches the caller).
2. Allow operator override via `~/.config/rtk/passthrough.toml`:

   ```toml
   # ~/.config/rtk/passthrough.toml
   [signal]
   patterns = [
     "git push", "git pull", "git fetch", "git merge",
     "git status", "git remote", "git rev-parse", "git branch",
     "gh pr", "gh issue", "gh release", "gh api", "gh run",
   ]

   # Operator can extend per-machine:
   [signal.extend]
   patterns = ["glab mr", "kubectl get pods"]
   ```

3. Match algorithm: substring match against the reconstructed command line (Bash hook input or argv for shim wrappers). Case-sensitive.
4. Behaviour:
   * Match → rtk forwards stdout untouched (or emits a passthrough decision in the Claude PreToolUse JSON shape).
   * No match → existing rtk reduction pipeline.

## Reference implementation

We've shipped a workaround in our coworker plugin (https://github.com/Arcanada-one/coworker) — a PreToolUse guard that stands in front of `rtk hook claude` and short-circuits for the patterns above. Sources:

* Vendored bash guard: `coworker/plugins/rtk_signal_guard.sh`
* JSON allowlist store: `coworker/plugins/rtk_passthrough.py`
* Codex CLI shim parity: `coworker/plugins/rtk_codex_shims.py` (passthrough check inside generated `git`/`gh` shims)

Happy to upstream this into rtk core (Rust port + TOML config) if the maintainers are interested. The wrapper is currently published under MIT, same as rtk.

## Why this matters beyond our agent

Anyone running rtk with an automated downstream consumer (Claude Code agents, Cursor sub-agents, Codex CLI sub-agents, CI bots that parse `git push` output) hits the same class of false-negative. The cost of the false negative is invisible to rtk's own metrics — rtk reports tokens saved, but the agent above it spent minutes spinning on a successful operation.

Happy to discuss design or open a PR if there's interest. Cheers.

wilmerdooley:
Resolves #2121.

## Summary

`rtk hook` rewrites the stdout of every Bash tool invocation. When the output is a small control signal (`git push`'s `To <repo>` line, `git status` porcelain, `gh pr` state) that downstream agents grep against, reshaping it drops the marker and the agent misreads a successful command as hung or failed.

- New `src/hooks/passthrough.rs` classifies a command as signal vs bulk. A small built-in allowlist of substring patterns (`git push`/`pull`/`fetch`/`merge`/`status`/`remote`/`rev-parse`/`branch`, `gh pr`/`issue`/`release`/`api`/`run`, `glab mr`/`issue`) marks the signal set.
- Operators override it through `~/.config/rtk/passthrough.toml`: `[signal] patterns` replaces the built-ins, `[signal.extend] patterns` appends. A missing or malformed file falls back to the defaults, so a bad override can never block the hook pipeline.
- Both rewrite paths (`hook_cmd::get_rewritten` and `rewrite_cmd::evaluate`) short-circuit signal commands to passthrough, so their raw stdout reaches the caller unreshaped. Bulk commands (`git log -p`, `ls -la`, `docker logs`) are unaffected.

## Test plan

- `cargo fmt --all -- --check`, `cargo clippy --all-targets`, `cargo check`: clean.
- `cargo test hooks::`: 355 passed, 0 failed.
- Behavioral check: before the change `rtk rewrite "git status"` returned `rtk git status`; after, it passes through unrewritten.
- New tests cover the default allowlist, the replace/extend override semantics, case-sensitivity, the empty-pattern guard, and passthrough across the claude/cursor/gemini hook surfaces.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -112,6 +112,13 @@ fn get_rewritten(cmd: &str) -> Option<String> {
         return None;
     }
 
+    // Signal commands emit canonical control markers (`git push`'s `To <repo>`
+    // line, `git status` porcelain, `gh pr` state) that downstream agents grep
+    // against. Bypass rewriting so raw stdout reaches the caller unreshaped.
+    if super::passthrough::is_signal_command(cmd) {
+        return None;
+    }
+
     let (excluded, transparent_prefixes) = crate::core::config::Config::load()
         .map(|c| (c.hooks.exclude_commands, c.hooks.transparent_prefixes))
         .unwrap_or_default();
@@ -619,7 +626,16 @@ mod tests {
 
     #[test]
     fn test_get_rewritten_supported() {
-        assert!(get_rewritten("git status").is_some());
+        assert!(get_rewritten("git diff").is_some());
+    }
+
+    #[test]
+    fn test_get_rewritten_signal_command_passthrough() {
+        // Signal commands bypass rewriting so their canonical control markers
+        // (git push's `To <repo>`, git status porcelain) reach the caller raw.
+        assert!(get_rewritten("git status").is_none());
+        assert!(get_rewritten("git push origin main").is_none());
+        assert!(get_rewritten("gh pr view 123").is_none());
     }
 
     #[test]
@@ -754,7 +770,7 @@ mod tests {
 
     #[test]
     fn test_copilot_cli_cve_safe_forms_still_rewrite() {
-        for cmd in ["git status", "git status 2>&1"] {
+        for cmd in ["git diff", "git diff 2>&1"] {
             let r = end_to_end(cmd).unwrap_or_else(|| panic!("expected rewrite for {cmd:?}"));
             assert_eq!(
                 r["modifiedArgs"]["command"].as_str().unwrap(),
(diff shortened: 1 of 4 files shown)


KuSh:
Thank you for this, and for the detailed write-up on the issue. The two places it hooks into, hook_cmd::get_rewritten and rewrite_cmd::evaluate, no longer exist: #3955 (merged 2026-09-11) consolidated every hook path onto a single decision in src/hooks/decision.rs, and it also moved the hook_cmd tests this PR rewrites, so the patch cannot be rebased without being written again. I am closing it for that reason rather than on the idea. The underlying request is still open as #2121: develop has no built-in signal command allowlist and no passthrough.toml, only the user-set hooks.exclude_commands, so a small PR adding one gate inside decide_with_params would be very welcome, as would a note on the issue about the override file semantics. Please comment or reopen if I have misjudged part of this.

PR #2670, opened 2026-06-27
fix(ruff): preserve non-check subcommands

### LINKED ISSUE #2669: ruff subcommand misrouting: rule, config, linter, etc. get check --output-format=json injected
## Bug

`ruff_cmd.rs` uses a blocklist to detect check invocations (lines 35-37):

```rust
let is_check = args.is_empty()
    || args[0] == "check"
    || (!args[0].starts_with('-') && args[0] != "format" && args[0] != "version");
```

Only `format` and `version` are excluded. All other ruff subcommands — `rule`, `config`, `linter`, `clean`, `server`, `analyze`, `generate-shell-completion` — are treated as check invocations and get `check --output-format=json` prepended.

## Reproduction

```bash
rtk ruff rule E501
# Actual:  ruff check --output-format=json rule E501  (invalid command, errors out)
# Expected: ruff rule E501

rtk ruff config
# Actual:  ruff check --output-format=json config  (invalid)
# Expected: ruff config

rtk ruff linter
# Actual:  ruff check --output-format=json linter  (invalid)
# Expected: ruff linter
```

## Root Cause

The `is_check` detection uses a blocklist (only "format" and "version" excluded) instead of an allowlist. Any first argument that doesn't start with `-` and isn't "format" or "version" is assumed to be a check target path.

## Suggested Fix

Add a `RUFF_SUBCOMMANDS` constant (similar to `NPM_SUBCOMMANDS` in `npm_cmd.rs`) listing known ruff subcommands, and exclude them from `is_check` detection:

```rust
const RUFF_SUBCOMMANDS: &[&str] = &[
    "check", "format", "rule", "config", "linter",
    "clean", "server", "analyze", "version",
    "generate-shell-completion",
];
```

Then `is_check` becomes: args is empty, OR first arg is "check", OR first arg is not a known subcommand and not a flag (i.e., it's a path).

fengjikui:
## Summary

Fixes #2669.

`rtk ruff` previously treated every first positional argument except `format` and `version` as a check target. That meant real Ruff subcommands such as `rule`, `config`, and `linter` were rewritten as `ruff check --output-format=json ...`, which makes those subcommands fail.

This changes the routing logic to use a known Ruff subcommand allowlist. Paths such as `src` and explicit `check` still route through the JSON check path, while non-check subcommands pass through unchanged.

## Validation

- `cargo test test_ruff_ -- --nocapture`
- `cargo fmt --check && cargo test ruff_cmd -- --nocapture`

### DIFF
--- a/src/cmds/python/ruff_cmd.rs
+++ b/src/cmds/python/ruff_cmd.rs
@@ -8,6 +8,19 @@ use anyhow::Result;
 use serde::Deserialize;
 use std::collections::HashMap;
 
+const RUFF_SUBCOMMANDS: &[&str] = &[
+    "analyze",
+    "check",
+    "clean",
+    "config",
+    "format",
+    "linter",
+    "rule",
+    "server",
+    "version",
+    "generate-shell-completion",
+];
+
 #[derive(Debug, Deserialize)]
 struct RuffLocation {
     row: usize,
@@ -32,9 +45,7 @@ struct RuffDiagnostic {
 }
 
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
-    let is_check = args.is_empty()
-        || args[0] == "check"
-        || (!args[0].starts_with('-') && args[0] != "format" && args[0] != "version");
+    let is_check = is_check_invocation(args);
 
     let is_format = args.iter().any(|a| a == "format");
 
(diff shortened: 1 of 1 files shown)


KuSh:
Closing in favour of #3083, which makes the same ruff subcommand allowlist fix, has a green CI run, and lists help alongside the other subcommands. If there is a routing case it still gets wrong, please raise it there.

PR #2671, opened 2026-06-27
fix(discover): report Claude hook coverage uncertainty

### LINKED ISSUE #2648: discover reports phantom "missed savings" when the Claude Code hook is configured

fengjikui:
Fixes #2648.

## Summary

`rtk discover` now detects the configured Claude Code RTK hook and reports that transcript-based opportunities may include commands the hook already rewrote.

## Root cause

Claude Code session transcripts store the original Bash command before the `PreToolUse` hook rewrite runs. The transcript alone therefore cannot prove whether a supported-looking command was rewritten, skipped by an exclusion or guard, disabled, or recorded before hook installation.

## Fix

- Add Claude hook detection to `AgentIntegrationStatus`, using the same Claude config directory resolution path as session discovery.
- Expose `claude_hook_installed` in JSON reports.
- Add a text note explaining that discover scans pre-hook commands and may include already-rewritten opportunities.
- Preserve all existing supported, missed, and already-RTK counts unchanged.

## Review follow-up

The original patch treated every supported command as already using RTK whenever the hook was installed. That overcounted historical sessions, excluded or disabled commands, and rewrite-guard skips. This revision removes that inference entirely and keeps only detection plus an explicit uncertainty note, as suggested in review.

## Validation

- `cargo fmt --check`
- `cargo test discover:: -- --nocapture` -> 457 passed
- `cargo check`
- `git diff --check origin/develop...HEAD`

### DIFF
--- a/src/discover/report.rs
+++ b/src/discover/report.rs
@@ -1,9 +1,11 @@
 //! Data types for reporting which commands RTK can and cannot optimize.
 
 use crate::hooks::constants::{
-    COPILOT_HOOK_FILE, CURSOR_DIR, GITHUB_DIR, HERMES_DIR, HERMES_PLUGINS_SUBDIR,
-    HERMES_PLUGIN_MANIFEST_FILE, HERMES_PLUGIN_NAME, HOOKS_SUBDIR, REWRITE_HOOK_FILE,
+    CLAUDE_DIR, CLAUDE_HOOK_COMMAND, COPILOT_HOOK_FILE, CURSOR_DIR, GITHUB_DIR, HERMES_DIR,
+    HERMES_PLUGINS_SUBDIR, HERMES_PLUGIN_MANIFEST_FILE, HERMES_PLUGIN_NAME, HOOKS_SUBDIR,
+    PRE_TOOL_USE_KEY, REWRITE_HOOK_FILE, SETTINGS_JSON,
 };
+use crate::hooks::init::resolve_claude_dir;
 use serde::Serialize;
 use std::path::Path;
 
@@ -50,6 +52,7 @@ pub struct UnsupportedEntry {
 
 #[derive(Debug, Serialize, Clone, Copy, PartialEq, Eq, Default)]
 pub struct AgentIntegrationStatus {
+    pub claude_hook_installed: bool,
     pub cursor_hook_installed: bool,
     pub hermes_plugin_installed: bool,
     pub copilot_hook_installed: bool,
@@ -60,6 +63,9 @@ impl AgentIntegrationStatus {
         let mut status = dirs::home_dir()
             .map(|home| Self::detect_from_home(&home))
             .unwrap_or_default();
+        status.claude_hook_installed = resolve_claude_dir()
+            .map(|claude_dir| Self::claude_hook_installed_in_dir(&claude_dir))
+            .unwrap_or(false);
         // Copilot is project-scoped (.github/hooks/), unlike the home-based agents.
         status.copilot_hook_installed = std::env::current_dir()
             .map(|cwd| Self::copilot_hook_installed_in(&cwd))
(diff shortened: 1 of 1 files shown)


felipetruman:
The `claude_hook_installed` detection and the report note are genuinely useful — that part I'd love to see land regardless. But counting *every* supported-classified command as `already_rtk` when the hook is installed trades one wrong number for another:

- Sessions that **predate the hook install** (still within the scan window) get counted as covered; their tokens were fully spent.
- Commands in `hooks.exclude_commands`, `RTK_DISABLED=1` prefixes, and rewrite guards (`gh --json`, `cat` with redirects, ...) are never rewritten by the hook, yet all get credited.
- It zeroes the MISSED table wholesale, erasing the residual signal (e.g. commands the hook genuinely skips today).

Your other PR, #2308, avoids all three by reading the rewrite Claude Code actually recorded (`hook_success` attachment with `permissionDecisionReason: "RTK auto-rewrite"`, keyed by `toolUseID`). I validated that approach against 30 days of production transcripts (178 sessions / 1,954 Bash calls on my main project): reported adoption 1.1% vs 43.0% actual, and the attachment format matches what current Claude Code writes.

Suggestion: rescope this PR to just the detection + note (or fold that into #2308), and let the counting come from recorded rewrites.

fengjikui:
Thanks for the detailed production-data check. I agreed with the three failure modes and rescaled the PR accordingly.

The refreshed head `3c390bf` removes all counting changes: `supported`, `already_rtk`, and the MISSED table now retain their existing behavior. The PR only detects the configured Claude hook, exposes that status in JSON, and adds a text note that discover scans pre-hook transcript commands, so opportunities may include commands RTK already rewrote.

I also rebased onto current `develop@5d32d07`. Validation: `cargo fmt --check`, all 457 `discover::` tests, `cargo check`, and diff check passed.

KuSh:
Thanks for working through this, and for the revision that dropped the over-counting inference. Issue #2648 is now handled on develop by #3206 (merged 2026-09-02), which logs the real PreToolUse decision at hook time and joins it back to transcript entries by tool_use_id, so commands the hook actually rewrote are counted as covered rather than showing up as missed savings. It also prints the uncertainty explicitly for the portion that still has to be estimated, including the date measured data starts from, which covers the note this PR adds. Closing as covered, but please comment or reopen if there is a case the new reporting still gets wrong.

PR #2674, opened 2026-06-28
feat: short-output auto-passthrough — skip compression when raw output is already minimal

### LINKED ISSUE #2673: feat: short-output auto-passthrough — skip compression when raw output is already minimal
# Issue: feat: short-output auto-passthrough — skip compression when raw output is already minimal

> **Target repo**: rtk-ai/rtk
> **Labels**: enhancement, help wanted
> **Submit at**: https://github.com/rtk-ai/rtk/issues/new

---

## Summary

RTK compresses **all** command output through its reduction pipeline, even when the output is already short enough that compression yields zero token savings and can actively harm downstream agent decisions.

Complementary to #2121 (which adds hook-level passthrough for predefined signal commands) and #1313 (which disables lossy truncation globally), this proposes an **output-length heuristic** that automatically bypasses the compression pipeline when the raw output is below a configurable threshold — for any command, not just a hardcoded allowlist.

### 1. Short output gets "compressed" into something worse

```bash
$ git push origin main
To github.com:user/repo.git
   abc1234..def5678  main -> main

$ rtk git push origin main
[git: push] 1 ref updated
```

The compressed form drops the `To <repo>` canonical marker that agents grep to confirm push success (#2121, #2280). The raw 2-line output was already minimal — compression added nothing.

### 2. Every command is affected, not just signal commands

```bash
$ echo $?
0

$ rtk echo $?
0
```

No token savings, but RTK still pays the overhead of running through the filter pipeline. For `cat small-file.txt`, `which python3`, `hostname`, etc., the cost is always net-negative: pipeline overhead with zero savings.

### 3. #2667 only covers the hook side

PR #2667 adds `is_signal_command()` in the hook/rewrite layer — it prevents the hook from rewriting `git push` into `rtk git push`. But if a user (or agent) calls `rtk git status` directly, the CLI execution layer still compresses the output. The passthrough only exists at one entry point.

### 4. Real-world data

From my local `rtk gain` (13,765 commands, 75.3% average savings):

| Command | Savings | Note |
|---------|---------|------|
| `ps aux` | 97.4% | Bulk — compression is valuable |
| `git log -50` | ~85% | Bulk — compression is valuable |
| `git status` (clean) | **-108%** | Output inflated by RTK (#2121 measurement) |
| `echo $?` | ~0% | No savings possible |
| `which python3` | ~0% | No savings possible |
| `git push origin main` | Negative value | Signal lost, agent loops |

The top savers are all bulk commands. Short-output commands are noise in the metrics — they contribute nothing but carry risk.

### Core idea

After command execution, **before** applying the compression pipeline, check the raw output size. If it's below a threshold, emit it unchanged.

```
Command executes → raw output captured →
  if raw_output.lines() ≤ SHORT_LINE_THRESHOLD && raw_output.len() ≤ SHORT_BYTE_THRESHOLD:
    emit raw output (passthrough)
  else:
    apply compression pipeline as normal
```

### Why output-length, not command-name

| Approach | Pros | Cons |
|----------|------|------|
| **Command-name allowlist** (#2667) | Predictable, zero overhead | Only covers predefined commands; `rtk git status` direct call still compressed |
| **Output-length heuristic** (this) | Universal; covers all short output regardless of command; works at CLI layer too | Requires buffered capture (already the default for `RunMode::Filtered`) |
| **Both combined** | Best of both: hook skips rewrite for known signals, CLI skips compression for any short output | Slightly more code |

The two approaches are complementary, not competing. #2667 handles the hook layer; this handles the execution layer.

#### Thresholds

```toml
# ~/.config/rtk/config.toml
[passthrough]
# Auto-passthrough when raw output is below these limits.
# Set to 0 to disable auto-passthrough (current behavior).
short_line_threshold = 5    # ≤ 5 lines → passthrough
short_byte_threshold = 500  # ≤ 500 bytes → passthrough
```

Both conditions must be met. Defaults are conservative: 5 lines / 500 bytes covers `git push`, `echo $?`, `which`, `hostname`, clean `git status`, etc. without skipping any bulk output.

#### Implementation sketch

The change lives in `src/core/runner.rs`, in `run_captured_filter()`:

```rust
// After capturing raw output, before filtering:
if should_auto_passthrough(text_to_filter) {
    print!("{}", text_to_filter);
    timer.track_passthrough(cmd_label, &format!("rtk {} (auto-passthrough)", cmd_label));
    return Ok(exit_code);
}

fn should_auto_passthrough(output: &str) -> bool {
    let config = crate::core::config::Config::load().ok();
    let line_threshold = config.as_ref()
        .and_then(|c| c.passthrough.short_line_threshold)
        .unwrap_or(5);
    let byte_threshold = config.as_ref()
        .and_then(|c| c.passthrough.short_byte_threshold)
        .unwrap_or(500);

    let line_count = output.lines().count();
    line_count <= line_threshold && output.len() <= byte_threshold
}
```

This is ~20 lines of Rust. It reuses the existing `Config` loading, `timer.track_passthrough()`, and the `RunMode::Passthrough` tracking path. No new modules needed.

#### Why not in `RunMode::Streamed`

Streamed commands pipe output through a `StreamFilter` incrementally — they can't know the total size until the stream ends. But most short-output commands (`git push`, `echo`, `which`) use `RunMode::Filtered` (captured mode), which already buffers the full output before filtering. The heuristic only applies to captured mode, which is the right default.

For `RunMode::Streamed`, the existing behavior is unchanged — streaming is already used for large-output commands where passthrough would never trigger.

#### Interaction with #2667

| Scenario | #2667 | This proposal | Result |
|----------|-------|---------------|--------|
| Hook sees `git push` | Skips rewrite → raw `git push` | N/A (not `rtk` command) | ✅ Raw output |
| Hook sees `git log -50` | Not in signal list → rewrites to `rtk git log` | 50 lines > threshold → compressed | ✅ Compressed |
| User calls `rtk git status` directly | Not involved (no hook) | 3 lines ≤ threshold → passthrough | ✅ Raw output |
| `git status -uall` (1000 lines) | #2667 passthrough (in signal list) | Not `rtk` command, not involved | ⚠️ Raw (large) output through hook |
| `echo $?` | Not in signal list | 1 line ≤ threshold → passthrough | ✅ Raw output |

Note the `git status -uall` edge case: #2667 always passes `git status` through at the hook level, even when output is bulk. With this proposal, if the user calls `rtk git status -uall` directly, the CLI layer would correctly compress it (1000 lines > threshold). The hook layer's command-name allowlist is a coarse heuristic; the output-length heuristic is a finer one.

#### Edge cases

- **Empty output**: 0 lines / 0 bytes → passthrough (correct — nothing to compress)
- **Non-zero exit + short stderr**: `passthrough` applies to the combined text (or stdout-only if `filter_stdout_only`). Short error messages pass through correctly.
- **`skip_filter_on_failure`**: Already handled — this check runs *before* the filter, so failure passthrough takes precedence.
- **`tee_label`**: If tee is configured, the raw output is still teed for CCR recall (#2485), then the short output is printed directly.

### Config integration

Add to `src/core/config.rs`:

```rust
#[derive(Debug, Default, Deserialize)]
pub struct PassthroughConfig {
    /// Auto-passthrough when output has ≤ N lines. 0 = disabled.
    #[serde(default)]
    pub short_line_threshold: Option<usize>,
    /// Auto-passthrough when output has ≤ N bytes. 0 = disabled.
    #[serde(default)]
    pub short_byte_threshold: Option<usize>,
}
```

And in the main `Config` struct:

```rust
pub struct Config {
    // ... existing fields ...
    #[serde(default)]
    pub passthrough: PassthroughConfig,
}
```

### Metrics

When auto-passthrough triggers, `timer.track_passthrough()` records 0 tokens saved. This is honest: the output was already minimal. Over time, `rtk gain` will show that passthrough commands contributed nothing to savings — which is correct. The value is in *avoiding harm*, not in *creating savings*.

## Related

- #2121 — Hook-level signal/bulk classifier (command-name allowlist)
- #2667 — PR implementing #2121
- #1313 — `no_truncation` lossless-only mode (different scope: long output, no truncation)
- #2280 — RTK rewrites confuse LLM into loops (same root cause: compressed output loses signal)
- #2653 — `condense_unified_diff` misleading `... +N more` (truncation indicator for output that wasn't truncated)

## Scope

This is intentionally minimal: one heuristic in `runner.rs`, one config section, ~50 lines total including tests. It doesn't try to solve the `git status -uall` case at the hook level (that's #2667's domain), and it doesn't add TOML pattern matching (that's #2121's domain). It solves the complementary problem: **at the execution layer, short output should never be compressed, regardless of which command produced it.**

lg320531124:
## Summary

Closes #2673

When command output is already minimal (≤5 lines, ≤500 bytes), skip the compression pipeline entirely and emit raw output unchanged. This prevents filter reshaping of short signal commands and avoids zero-savings compression overhead.

### Problem

RTK compresses **all** command output. When output is already short, compression:
- Yields **zero token savings** (nothing to compress)
- Can **reshape or inflate** canonical markers (e.g. `git push` output, `echo $?`)
- Adds **unnecessary latency** (filter pipeline runs for no benefit)

PR #2667 adds hook-level passthrough for predefined signal commands, but:
1. Only covers the hook layer — `rtk git status` called directly still compresses
2. Allowlist-based — `echo $?`, `which python3`, `hostname` etc. are not in the list
3. Coarse heuristic — `git status -uall` (1000 lines) gets the same passthrough as clean `git status` (3 lines)

### Solution

An **output-length heuristic** at the CLI execution layer that auto-passthroughs any short output, regardless of command name.

**Both conditions must be met** (AND logic):
- `line_count ≤ 5` — covers `git push` (2 lines), `echo $?` (1 line), clean `git status` (3 lines)
- `output.len() ≤ 500` — prevents passthrough for very long single-line output (e.g. minified JSON)

### Interaction with #2667

| Scenario | #2667 (hook) | This PR (CLI) | Result |
|----------|-------------|---------------|--------|
| `git push` (2 lines) | Hook skips rewrite | N/A (raw command runs) | ✅ Raw output |
| `rtk git status` (3 lines) | N/A (already rtk) | Auto-passthrough | ✅ Raw output |
| `rtk git status -uall` (1000 lines) | N/A | Exceeds threshold → compressed | ✅ Compressed |
| `echo $?` (1 line) | Not in allowlist | Auto-passthrough | ✅ Raw output |
| `rtk git log -50` (200 lines) | N/A | Exceeds threshold → compressed | ✅ Compressed |

### Changes

| File | Change | Lines |
|------|--------|-------|
| `src/core/config.rs` | `PassthroughConfig` struct, `effective_*()` methods, config tests | +88 |
| `src/core/runner.rs` | `should_auto_passthrough()`, integration in `run_captured_filter()`, tests | +110 |
| `docs/contributing/DESIGN-short-output-passthrough.md` | Full design document | +393 |

**Total: ~90 lines of production code** (excluding design doc and tests). No new files, no new modules, no new dependencies.

### Config

```toml
# ~/.config/rtk/config.toml
[passthrough]
short_line_threshold = 5    # ≤ 5 lines → passthrough (default: 5)
short_byte_threshold = 500  # ≤ 500 bytes → passthrough (default: 500)
```

- **Default**: Auto-passthrough ON (5 lines / 500 bytes)
- **Opt-out**: Set `short_line_threshold = 1` + `short_byte_threshold = 1`

### Test Plan

- [x] Unit tests for `should_auto_passthrough()` (8 test cases)
- [x] Config deserialization tests (4 test cases)
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` — all pass (2299 tests)
- [ ] Manual test: `rtk git status` (short → passthrough) vs `rtk git log -50` (long → compressed)
- [ ] Performance: `hyperfine` comparison (passthrough should be faster for short commands)

### Design Document

Full design doc: [`docs/contributing/DESIGN-short-output-passthrough.md`](docs/contributing/DESIGN-short-output-passthrough.md)

Covers: architecture, insertion point rationale, feature interaction matrix, edge cases, migration strategy, open questions.

### DIFF
--- a/docs/contributing/DESIGN-short-output-passthrough.md
+++ b/docs/contributing/DESIGN-short-output-passthrough.md
@@ -1,0 +1,393 @@
+# Design Document: Short-Output Auto-Passthrough
+
+> **Issue**: [#2673](https://github.com/rtk-ai/rtk/issues/2673)
+> **Complements**: [#2121](https://github.com/rtk-ai/rtk/issues/2121) / [PR #2667](https://github.com/rtk-ai/rtk/pull/2667) (hook-level signal passthrough)
+> **Author**: @lg320531124
+> **Date**: 2026-06-28
+
+---
+
+## 1. Problem Statement
+
+RTK compresses **all** command output through its reduction pipeline. When output is already short (≤5 lines / ≤500 bytes), compression yields zero token savings and can actively harm downstream agent decisions by reshaping or inflating canonical markers.
+
+PR #2667 adds hook-level passthrough for predefined signal commands (`git push`, `gh pr`, etc.), but:
+
+1. **Only covers the hook layer** — `rtk git status` called directly still compresses
+2. **Allowlist-based** — `echo $?`, `which python3`, `hostname` etc. aren't in the list
+3. **Coarse heuristic** — `git status -uall` (1000 lines) gets the same passthrough as clean `git status` (3 lines)
+
+We need an **output-length heuristic** at the CLI execution layer that auto-passthroughs any short output, regardless of command name.
+
+---
+
+## 2. Architecture
+
+### 2.1 Current Flow (without auto-passthrough)
+
+```
+Command enters RTK
+  → main.rs routes to *_cmd.rs
+  → *_cmd.rs calls runner::run_filtered() or run_filtered_with_exit()
+  → run_captured_filter():
+      1. stream::run_streaming() captures raw output
+      2. filter_fn(text_to_filter, exit_code) compresses output
+      3. guard::never_worse() ensures filtered ≤ raw
+      4. print filtered output
+      5. timer.track() records token savings
+```
+
+### 2.2 New Flow (with auto-passthrough)
+
+```
+Command enters RTK
+  → main.rs routes to *_cmd.rs
+  → *_cmd.rs calls runner::run_filtered() or run_filtered_with_exit()
+  → run_captured_filter():
+      1. stream::run_streaming() captures raw output
+      2. ★ NEW: should_auto_passthrough(text_to_filter) ★
+         → YES: print raw output, timer.track_passthrough(), return early
+         → NO:  continue to filter_fn
+      3. filter_fn(text_to_filter, exit_code) compresses output
+      4. guard::never_worse() ensures filtered ≤ raw
+      5. print filtered output
+      6. timer.track() records token savings
+```
+
+The check goes **after** capture but **before** filtering. This is the natural insertion point because:
+
+- Raw output is already available (captured by `run_streaming`)
+- No filter pipeline work has been done yet (zero wasted effort)
+- `skip_filter_on_failure` check happens first (failure passthrough takes precedence)
+
+---
+
+## 3. Design Details
+
+### 3.1 `PassthroughConfig` — New Config Section
+
+Location: `src/core/config.rs`
+
+```rust
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct PassthroughConfig {
+    /// Auto-passthrough when output has ≤ N lines.
+    /// 0 = disabled (current behavior). Default: 5.
+    #[serde(default)]
+    pub short_line_threshold: usize,
+
+    /// Auto-passthrough when output has ≤ N bytes.
+    /// 0 = disabled (current behavior). Default: 500.
+    #[serde(default)]
+    pub short_byte_threshold: usize,
+}
+```
+
+Add to `Config` struct:
+
+```rust
+pub struct Config {
+    // ... existing fields ...
+    #[serde(default)]
+    pub passthrough: PassthroughConfig,
+}
+```
+
+**Why `usize` not `Option<usize>`**: `0` means disabled, which is the current behavior. No need for `Option` — simpler API, simpler TOML.
+
+**Why not `PassthroughConfig::default()` with values**: If we set defaults in the `Default` impl, then `Config::load()` for existing config.toml files (without `[passthrough]` section) will get `0/0` from serde's default. Instead, `should_auto_passthrough()` will use hardcoded defaults when thresholds are 0.
+
+**Config file example**:
+
+```toml
+# ~/.config/rtk/config.toml
+[passthrough]
+short_line_threshold = 5    # ≤ 5 lines → passthrough
+short_byte_threshold = 500  # ≤ 500 bytes → passthrough
+```
+
+### 3.2 `should_auto_passthrough()` — Core Heuristic
+
+Location: `src/core/runner.rs`
+
+```rust
+/// Whether the raw output is short enough that compression adds no value.
+/// When true, the caller should emit the raw output unchanged.
+fn should_auto_passthrough(output: &str) -> bool {
+    let config = crate::core::config::Config::load().ok();
+
+    // Hardcoded defaults: 5 lines, 500 bytes.
+    // Config values of 0 mean "disabled" (current behavior).
+    let line_threshold = config
+        .as_ref()
+        .map(|c| c.passthrough.short_line_threshold)
+        .filter(|&t| t > 0)
+        .unwrap_or(5);
+    let byte_threshold = config
+        .as_ref()
+        .map(|c| c.passthrough.short_byte_threshold)
+        .filter(|&t| t > 0)
+        .unwrap_or(500);
+
+    let line_count = output.lines().count();
+    line_count <= line_threshold && output.len() <= byte_threshold
+}
+```
+
+**Both conditions must be met** (AND, not OR):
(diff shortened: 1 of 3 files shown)


lg320531124:
Friendly nudge — open ~28h, CI green. This is a perf win for the common case: when raw model output is already minimal (short responses), skipping compression avoids unnecessary token/latency overhead with no behavior change for longer outputs. +591 all new code + tests, no changes to existing paths. Happy to split into smaller commits or add benchmarks if that eases review.

KuSh:
Closing as a duplicate of #2685, which implements the same short-output auto-passthrough for #2673 with the same thresholds, keeps the tee file written so raw recovery still works, and lets a threshold of 0 disable the feature. Please take the discussion and any review there. The design document here is a useful write-up if you want to propose it separately.

PR #2681, opened 2026-06-28
fix(run): propagate signal exit code instead of unwrap_or(1)

### LINKED ISSUE #2680: rtk run: signal exit code not propagated (unwrap_or(1))
## Bug

`Commands::Run` in `main.rs:2334` uses `status.code().unwrap_or(1)` to extract the child process exit code.

On Unix, when a process is killed by a signal (SIGTERM, SIGKILL, SIGINT), `ExitStatus::code()` returns `None`. The `unwrap_or(1)` maps it to exit code 1, losing the signal information. The POSIX convention is to return `128 + signal_number` (e.g., SIGTERM=15 → exit 143).

## Impact

`rtk run <command>` returns exit code 1 instead of the correct signal-based exit code when the child is terminated by a signal. CI/CD systems and scripts that rely on specific exit codes to distinguish crashes from normal failures get the wrong signal.

## Context

The correct helper `exit_code_from_status()` already exists in `core/utils.rs:213` and is used by:
- `Commands::Proxy` (main.rs:2515) ✅
- npx prisma routing (main.rs:2205, 2215) ✅
- General command routing (main.rs:1310) ✅

`Commands::Run` is the only path that still uses the old `unwrap_or(1)` pattern.

## Fix

Replace:
```rust
status.code().unwrap_or(1)
```
With:
```rust
core::utils::exit_code_from_status(&status, "run")
```

guyoron1:
## Summary

- Replace `status.code().unwrap_or(1)` with `exit_code_from_status(&status, "run")` in `Commands::Run` (main.rs:2334)
- Signal kills (SIGTERM, SIGKILL, OOM) now produce POSIX-conventional `128+signal` exit codes instead of silently mapping to `1`
- Uses the same helper already used by `Commands::Proxy` at line 2515

Fixes #2680

## Test plan

- [x] `cargo fmt --all` — no formatting changes
- [x] `cargo test --all` — all tests pass
- [ ] Manual: `rtk run "sleep 99" &; kill -TERM $!` should exit 143 (not 1)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2331,7 +2331,7 @@ fn run_cli() -> Result<i32> {
                     .arg(&raw)
                     .status()
                     .with_context(|| format!("Failed to execute: {}", raw))?;
-                status.code().unwrap_or(1)
+                core::utils::exit_code_from_status(&status, "run")
             }
         }
 


aeppling:
hey @guyoron1 thanks for the contribution, lgtm for better transparency on exit code propagation

aeppling (approved):

PR #2683, opened 2026-06-28
fix(test): replace generic runner last-5-lines fallback with 3-tier detection

### LINKED ISSUE #2420: rtk test shows only the last 5 output lines (not failures) for unrecognized runners

guyoron1:
## Summary

The generic test runner fallback (for unrecognized test frameworks like `make test`, `bazel test`, custom scripts) showed only the last 5 lines of output via `OUTPUT (last 5 lines):`. This hides failure details when the command fails.

**Root cause:** `extract_test_summary()` fell through to an unconditional "last 5 lines" else block when no recognized runner was detected, and `run_test()` used `run_filtered` which discards the exit code before the filter sees it.

### Fix

1. Switch `run_test` from `run_filtered` → `run_filtered_with_exit` so the filter receives the exit code
2. Add `exit_code: i32` parameter to `extract_test_summary`
3. Replace the generic fallback with 3-tier detection:
   - **Tier A:** Scan all lines with `ERROR_PATTERNS` (same regexes used by `ErrorStreamFilter`) → `[FAIL] DETECTED FAILURES:`
   - **Tier B:** No patterns matched but `exit_code != 0` → last 10 lines + `[FAIL] Command failed (exit code: N)`
   - **Tier C:** `exit_code == 0`, no failures → `[ok] All tests passed (no recognized test runner)`

### Tests added

- `test_generic_runner_surfaces_failure_lines` — Tier A: error patterns detected
- `test_generic_runner_nonzero_no_patterns` — Tier B: nonzero exit, no patterns
- `test_generic_runner_success` — Tier C: clean exit
- `test_recognized_runners_unchanged` — existing cargo test behavior preserved

Fixes #2420

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -136,11 +136,11 @@ pub fn run_test(command: &str, verbose: u8) -> Result<i32> {
     }
     let cmd = build_shell_command(command);
     let command_owned = command.to_string();
-    crate::core::runner::run_filtered(
+    crate::core::runner::run_filtered_with_exit(
         cmd,
         "test",
         command,
-        move |raw| extract_test_summary(raw, &command_owned),
+        move |raw, exit_code| extract_test_summary(raw, &command_owned, exit_code),
         crate::core::runner::RunOptions::with_tee("test"),
     )
 }
@@ -178,7 +178,7 @@ fn filter_errors(output: &str) -> String {
     result.join("\n")
 }
 
-fn extract_test_summary(output: &str, command: &str) -> String {
+fn extract_test_summary(output: &str, command: &str, exit_code: i32) -> String {
     let mut result = Vec::new();
     let lines: Vec<&str> = output.lines().collect();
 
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased onto current develop. The code moved to `src/core/runner.rs` (`run_test_cmd` / `extract_test_summary`), so I re-applied it there. Re-testing with the exact script from #2420 showed my first version still hid `FAIL: test_login broke` (`ERROR_PATTERNS` has no bare `FAIL`), and it printed a `[FAIL]` banner for green runs that merely print "warning" or "0 failed".

Now only a non-zero exit takes the new path: the `rtk err` error blocks plus a bare-`fail` marker that only test runs read (the shared `ERROR_PATTERNS` are untouched), otherwise the exit code plus the last ten lines. A run that exits 0 keeps its tail, and recognised runners are unchanged. Unit and e2e tests fail against the old fallback.

Two notes for review: I dropped my third tier (`[ok] All tests passed`), since it claims something rtk doesn't know and discards the tail that carries things like `12 tests collected`; restoring it is a three-line change if you prefer it. And #4125 also edits `src/core/runner.rs`, so expect a rebase for whichever lands second.

Issue #2280, opened 2026-06-05
RTK rewrites confuse LLM resulting in loops


cirezd:
Hello, I just started using rtk with opencode. Issue is that command rewrites (e.g. ruff) confuse the LLM so it thinks it made a mistake and tries again and again to fix it. See below

```
# Lint changed test files with ruff
$ rtk ruff check tests/functional_tests/test_custom_events.py tests/functional_tests/test_data_sources.py tests/unit_tests/test_custom_event_validation.py --fix
Ruff: No issues found
Thought: 395ms
I keep accidentally typing "rtk" instead of "ruff". The output says "Ruff: No issues found" which means the linting passed. Let me run the actual ruff command to be safe.
# Lint changed test files with ruff
$ rtk ruff check tests/functional_tests/test_custom_events.py tests/functional_tests/test_data_sources.py tests/unit_tests/test_custom_event_validation.py --fix
Ruff: No issues found
Thought: 490ms
I keep accidentally typing "rtk" instead of "ruff". The output says "Ruff: No issues found" which means the linting passed. Let me run the actual ruff command to be safe.
# Lint changed test files with ruff
$ rtk ruff check tests/functional_tests/test_custom_events.py tests/functional_tests/test_data_sources.py tests/unit_tests/test_custom_event_validation.py --fix
Ruff: No issues found
Thought: 388ms
I keep accidentally typing "rtk" instead of "ruff". The output says "Ruff: No issues found" which means the linting passed. Let me run the actual ruff command to be safe.
# Lint changed test files with ruff
$ rtk ruff check tests/functional_tests/test_custom_events.py tests/functional_tests/test_data_sources.py tests/unit_tests/test_custom_event_validation.py --fix
Ruff: No issues found
```

rtk 0.42.1
opencode 1.15.13
kimi-k2.6

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 88% |

### Summary

RTK's ruff rewrite produces output ('Ruff: No issues found') that strips the 'rtk' invocation context, causing the LLM to believe it mistyped the command and retry in an infinite loop; the rewrite output needs to make the rtk-wrapper origin unambiguous.

---
[...]

PR #2685, opened 2026-06-28
feat(runner): skip filtering for short output (auto-passthrough)

### LINKED ISSUE #2673: feat: short-output auto-passthrough — skip compression when raw output is already minimal

guyoron1:
## Summary

RTK compressed all output regardless of size, adding overhead and removing useful text from already-short outputs (1-2 lines). For example, `git branch` showing 2 lines would get processed through the full filter pipeline for zero benefit.

### Fix

Add a **dual-threshold** auto-passthrough check in `run_captured_filter` (before the filter function runs):
- **`short_line_threshold`** (default: 5) — max line count for passthrough
- **`short_byte_threshold`** (default: 500) — max byte count for passthrough
- Both must be satisfied (AND logic): a few very long lines still get filtered, many short lines still get filtered
- Setting either threshold to 0 disables the feature entirely

When triggered, the raw output is printed directly, tee file is written (for `rtk tee` recovery), and tracking records 0% savings.

### Files changed

- **`src/core/config.rs`** — Add `short_line_threshold` and `short_byte_threshold` to `LimitsConfig` with `#[serde(default)]` for backward compatibility
- **`src/core/runner.rs`** — Add `should_auto_passthrough()` function, insert early-return in `run_captured_filter`

### Tests added

Config tests:
- `test_config_short_thresholds_defaults` — default values (5 lines, 500 bytes)
- `test_config_short_thresholds_deserialize` — explicit config values
- `test_config_short_thresholds_missing_uses_defaults` — backward compat with old configs

Runner tests:
- `test_should_auto_passthrough_short` — short output triggers passthrough
- `test_should_auto_passthrough_long` — long bytes prevents passthrough
- `test_should_auto_passthrough_many_lines` — many lines prevents passthrough
- `test_should_auto_passthrough_empty` — empty output triggers passthrough
- `test_should_auto_passthrough_at_boundary` — exact boundary behavior (5 lines = pass, 6 = filter)

### Config example

```toml
[limits]
short_line_threshold = 5    # 0 to disable
short_byte_threshold = 500  # 0 to disable
```

Fixes #2673

### DIFF
--- a/src/core/config.rs
+++ b/src/core/config.rs
@@ -131,6 +131,19 @@ pub struct LimitsConfig {
     pub status_max_untracked: usize,
     /// Max chars for parser passthrough fallback (default: 2000)
     pub passthrough_max_chars: usize,
+    /// Skip filtering when output has ≤ this many lines (default: 5, 0 = disabled)
+    #[serde(default = "default_short_line_threshold")]
+    pub short_line_threshold: usize,
+    /// Skip filtering when output has ≤ this many bytes (default: 500, 0 = disabled)
+    #[serde(default = "default_short_byte_threshold")]
+    pub short_byte_threshold: usize,
+}
+
+fn default_short_line_threshold() -> usize {
+    5
+}
+fn default_short_byte_threshold() -> usize {
+    500
 }
 
 impl Default for LimitsConfig {
@@ -141,6 +154,8 @@ impl Default for LimitsConfig {
             status_max_files: 15,
             status_max_untracked: 10,
             passthrough_max_chars: 2000,
+            short_line_threshold: default_short_line_threshold(),
+            short_byte_threshold: default_short_byte_threshold(),
         }
     }
 }
@@ -295,5 +310,44 @@ consent_date = "2026-04-10T12:00:00Z"
             config.telemetry.consent_date.as_deref(),
             Some("2026-04-10T12:00:00Z")
         );
+    }
+
+    #[test]
+    fn test_config_short_thresholds_defaults() {
+        let config = Config::default();
+        assert_eq!(config.limits.short_line_threshold, 5);
+        assert_eq!(config.limits.short_byte_threshold, 500);
+    }
+
+    #[test]
+    fn test_config_short_thresholds_deserialize() {
+        let toml = r#"
+[limits]
+grep_max_results = 200
+grep_max_per_file = 25
+status_max_files = 15
+status_max_untracked = 10
+passthrough_max_chars = 2000
+short_line_threshold = 10
+short_byte_threshold = 1000
+"#;
+        let config: Config = toml::from_str(toml).expect("valid toml");
+        assert_eq!(config.limits.short_line_threshold, 10);
+        assert_eq!(config.limits.short_byte_threshold, 1000);
+    }
+
+    #[test]
+    fn test_config_short_thresholds_missing_uses_defaults() {
+        let toml = r#"
+[limits]
+grep_max_results = 100
+grep_max_per_file = 25
+status_max_files = 15
+status_max_untracked = 10
+passthrough_max_chars = 2000
+"#;
+        let config: Config = toml::from_str(toml).expect("valid toml");
+        assert_eq!(config.limits.short_line_threshold, 5);
+        assert_eq!(config.limits.short_byte_threshold, 500);
     }
 }
--- a/src/core/runner.rs
+++ b/src/core/runner.rs
@@ -88,6 +88,16 @@ pub enum RunMode<'a> {
     Passthrough,
 }
 
+fn should_auto_passthrough(text: &str) -> bool {
+    let limits = crate::core::config::limits();
+    let line_threshold = limits.short_line_threshold;
+    let byte_threshold = limits.short_byte_threshold;
+    if line_threshold == 0 || byte_threshold == 0 {
+        return false;
+    }
+    text.len() <= byte_threshold && text.lines().count() <= line_threshold
+}
+
 fn run_captured_filter<F>(
     mut cmd: Command,
     tool_name: &str,
@@ -118,6 +128,15 @@ where
         if !result.raw_stderr.trim().is_empty() {
             eprint!("{}", result.raw_stderr);
         }
+        timer.track(cmd_label, &format!("rtk {}", cmd_label), raw, raw);
+        return Ok(exit_code);
+    }
+
+    if should_auto_passthrough(raw) {
+        if let Some(label) = opts.tee_label {
+            crate::core::tee::tee_raw(raw, label, exit_code);
+        }
+        print!("{}", raw);
         timer.track(cmd_label, &format!("rtk {}", cmd_label), raw, raw);
         return Ok(exit_code);
     }
@@ -283,4 +302,39 @@ pub fn run_streamed(
         RunMode::Streamed(filter),
         opts,
     )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_should_auto_passthrough_short() {
+        assert!(should_auto_passthrough("line one\nline two\n"));
+    }
+
+    #[test]
+    fn test_should_auto_passthrough_long() {
+        let long = "a]n".repeat(200);
+        assert!(!should_auto_passthrough(&long));
+    }
+
+    #[test]
+    fn test_should_auto_passthrough_many_lines() {
+        let lines = "line\n".repeat(10);
+        assert!(!should_auto_passthrough(&lines));
+    }
+
+    #[test]
+    fn test_should_auto_passthrough_empty() {
+        assert!(should_auto_passthrough(""));
+    }
+
+    #[test]
+    fn test_should_auto_passthrough_at_boundary() {
+        let exactly_5_lines = "a\nb\nc\nd\ne";
+        assert!(should_auto_passthrough(exactly_5_lines));
+        let six_lines = "a\nb\nc\nd\ne\nf";
+        assert!(!should_auto_passthrough(six_lines));
+    }
 }


guyoron1:
@KuSh rebased on develop and green. You closed #2674 in favour of this PR, so I'd like your call on one design point.

The CI failure on the last version was real, not flaky. At 5 lines / 500 bytes the passthrough also swallows a four-row `gh pr checks` table, a one-line `golangci-lint` JSON report and `git checkout` confirmations, and seven existing integration tests across `gh_pr_checks_failure_test.rs`, `stderr_only_failure_test.rs` and `guard_integration_test.rs` pin that the filter must still run there. No length threshold can separate those from `git push`'s lost `To <remote>` line, since the filtered form is shorter in every one of them. That looks like #2121's job (reading the canonical marker per command), and `guard::never_worse` already covers the "compression made it bigger" half of #2673.

So both thresholds now default to 0 (off): the feature is opt-in and the seven tests stay as they are. If you'd rather have 5/500 on by default, it's two lines in `LimitsConfig::default()` plus re-deciding those seven assertions, which I didn't want to touch without your say-so since they're your contract tests.

Two bugs in my original version are fixed as well: it printed the interleaved stdout+stderr capture to stdout, and it read config with `Config::load()` (a disk read) on the per-command path instead of `cached_config()`.

KuSh:
Thanks for the rebase and for tracing the seven failing contract tests to a cause. That analysis is what settles the default question, and it is the right call to leave those tests as they are.

I ran the branch against develop (`0c279f54`) with the thresholds off and on (5 / 500, full `[limits]` block), each command going through `rtk rewrite` the way the hook sends it. With the default, the branch matches develop byte for byte on all 14 commands (stdout, stderr and exit code), so the off path is a true no-op, and the stream split and the 0%-savings tracking work as described. With the thresholds on, though, I don't think a length threshold can reach what #2673 and #2280 describe:

| command | develop | this branch, thresholds on | the command itself |
|---|---|---|---|
| `rtk git push origin main` | `To ../remote.git` / `* [new branch] main -> main` / `ok main` | same: `run_push` streams and never reaches `run_captured_filter` | — |
| `rtk ruff check clean.py` | `[]` | `[]` | `All checks passed!` |
| `rtk ls small` | 28 B, `sub/  one.txt 0B  two.rs 0B` | 162 B, `total 4` + three `ls -la` rows | 19 B |
| `rtk git status` (clean) | `* main` / `clean — nothing to commit` | same: this path does not read the setting | 2 lines |

- `git push` has printed `To <remote>` since v0.41.0 (#963), and it does not go through the code this PR changes.
- For ruff, the empty-tree `[]` is already what develop prints. It comes from `guard::never_worse` preferring the capture of rtk's injected `--output-format=json` over the longer `Ruff: No issues found`, so the passthrough cannot fix it. I've opened #4218 for that.
- In 23 of the 62 `run_filtered` call sites, rtk rewrites the arguments before running (13 of them force an output format: `ls -la`, `kubectl -o json`, JSON for ruff / rubocop / rspec / phpstan / golangci-lint / `go test -json` / the gh and glab helpers). There, the short "raw" output is rtk's own invocation, not what the user typed.
- The key sits in the global `[limits]`, but only the `run_filtered` path reads it, so `git status`, `cargo build` and TOML-filtered commands keep filtering short output.

On your question about turning 5 / 500 on by default: no, for the reason you found and for the `ls` / ruff rows above.

The marker problem is real, and it belongs per command, which is where #2121 puts it. If you want to take a command from there, that is where I'd point this effort. I'm closing this one on that basis. I closed #2674 in favour of this PR before these results existed, so this follows the evidence rather than a change of mind about the goal.

PR #2689, opened 2026-06-28
perf(deps): move regex compilation to lazy_static

### LINKED ISSUE #2688: perf(deps): regex recompiled on every call in summarize_cargo_str and summarize_requirements_str
## Bug

In `src/cmds/system/deps.rs`, three `Regex::new()` calls use static patterns but are compiled inside function bodies, causing recompilation on every invocation:

- `summarize_cargo_str()`: `dep_re` (line 85-86) and `section_re` (line 87)
- `summarize_requirements_str()`: `dep_re` (line 174)

Per project convention, static regex patterns must use `lazy_static!` to compile once at first use.

## Impact

Unnecessary CPU overhead on every `rtk deps` call. The regex engine parses and compiles the same patterns repeatedly instead of caching the compiled automaton.

## Fix

Move all three patterns to a `lazy_static!` block at module scope.

guyoron1:
## Summary

Three `Regex::new()` calls in `deps.rs` use static patterns but were compiled
inside function bodies (`summarize_cargo_str`, `summarize_requirements_str`),
causing recompilation on every invocation.

Moves all three to a `lazy_static!` block at module scope so they compile once
at first use, following the established RTK convention.

## Changes

- `src/cmds/system/deps.rs`: Moved `dep_re`, `section_re`, and requirements
  `dep_re` to `lazy_static!` block as `CARGO_DEP_RE`, `CARGO_SECTION_RE`, and
  `REQUIREMENTS_DEP_RE`

## Test plan

- [x] `cargo fmt --all` — no formatting issues
- [x] `cargo test --all` — all tests pass
- [x] Behavior unchanged (same regex patterns, same matching logic)

Fixes #2688

### DIFF
--- a/src/cmds/system/deps.rs
+++ b/src/cmds/system/deps.rs
@@ -7,6 +7,15 @@ use anyhow::Result;
 use regex::Regex;
 use std::fs;
 use std::path::Path;
+use std::sync::LazyLock;
+
+static CARGO_DEP_RE: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r#"^([a-zA-Z0-9_-]+)\s*=\s*(?:"([^"]+)"|.*version\s*=\s*"([^"]+)")"#).unwrap()
+});
+static CARGO_SECTION_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^\[([^\]]+)\]").unwrap());
+static REQUIREMENTS_DEP_RE: LazyLock<Regex> =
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased on develop. The change was already `std::sync::LazyLock`, not `lazy_static`; only the PR title said `lazy_static`, so I've renamed the title to match (the crate isn't a dependency here, and CONTRIBUTING asks for `LazyLock` anyway).

I also added tests for `summarize_cargo_str` and `summarize_requirements_str`: the module only had `package.json` tests, so the three patterns had no coverage at all before this. Measured cost of compiling the three patterns: ~97 µs per `rtk deps` run, now paid once per process.

PR #2690, opened 2026-06-28
fix(dotnet): parse MTP multi-line test run summary

### LINKED ISSUE #2279: dotnet test in binlog-only mode suppresses MTP test result counts
When running rtk dotnet test, RTK operates in binlog-only mode and suppresses all stdout. This works correctly for detecting pass/fail, but the Microsoft Testing Platform (MTP) test result summary — total,
  succeeded, failed, skipped counts — is written directly to stdout by the test executable, not as an MSBuild event. As a result, RTK reports counts unavailable and the summary is lost entirely.

  Current output
  ok dotnet test: completed (binlog-only mode, counts unavailable, 0 warnings)

  Expected output (or something equivalent)
  ok dotnet test: completed (72 passed, 0 failed, 0 skipped, 1s 672ms)

  To reproduce

  rtk dotnet test --project tests/SomeProject.Test.csproj --filter "SomeFilter"

  Root cause

  MSBuild binary logs capture build/task events. MTP (.NET 10's default test runner) prints its result summary via the test executable's own stdout, outside the MSBuild event stream. RTK reads the binlog but
  never sees the stdout that carries the counts.

  Running without RTK confirms the data is there:
  Test run summary: Passed!
    total: 72
    failed: 0
    succeeded: 72
    skipped: 0
    duration: 1s 672ms

  Suggested fix

  dotnet test failures are detectable from both exit code and stdout, so binlog is not strictly necessary for failure detection on test commands. One of:

  1. Detect dotnet test specifically and switch from binlog-capture to stdout passthrough, filtering to failures + the summary line only — same UX as build, but with counts visible on success.
  2. Tee stdout alongside binlog for test commands and extract the MTP summary line (Test run summary: / total: / succeeded: etc.) to append to RTK's own output.
  3. Parse --report-trx — if RTK injects --report-trx --results-directory <tmpdir> automatically for test commands, it could read the TRX XML after the run and extract counts independently of stdout.

  Option 1 is the simplest since MTP exit code (0 = all pass, non-zero = failures) is reliable enough to not need binlog for test runs.

  Environment
  - .NET 10 / Microsoft Testing Platform (MTP)
  - dotnet test uses --solution or --project flags (MTP mode, not VSTest)

guyoron1:
## Summary

The Microsoft Testing Platform (.NET 10) writes test results in a multi-line
format that the existing regexes couldn't match, causing `rtk dotnet test` to
report "counts unavailable" even when tests passed successfully.

**Before:** `ok dotnet test: completed (binlog-only mode, counts unavailable, 0 warnings)`
**After:** `ok dotnet test: 72 tests passed, 0 warnings in 1 projects (1s 672ms)`

## Root cause

MTP outputs a multi-line summary:
```
Test run summary: Passed!
  total: 72
  failed: 0
  succeeded: 72
  skipped: 0
  duration: 1s 672ms
```

But `TEST_RESULT_RE` expects VSTest's single-line `Passed! - Failed: 0, Passed: 72, ...`
format, and `TEST_SUMMARY_RE` expects single-line `Test summary: total: 72, failed: 0, ...`.
Neither matches the multi-line MTP format, so all counts stay at zero.

## Fix

Add `MTP_MULTILINE_SUMMARY_RE` as a fallback regex in `parse_test_from_text`. It only
triggers when the earlier regexes find no counts (`summary.total == 0`), keeping backward
compatibility with VSTest and single-line MTP formats.

## Changes

- `src/cmds/dotnet/binlog.rs`: Add `MTP_MULTILINE_SUMMARY_RE` lazy_static regex and
  fallback parsing block in `parse_test_from_text`
- Two new unit tests: `test_parse_test_from_text_mtp_multiline_passed` and
  `test_parse_test_from_text_mtp_multiline_failed`

## Test plan

- [x] `cargo fmt --all` — no formatting issues
- [x] `cargo test --all` — all tests pass (including 2 new MTP tests)
- [x] New tests verify both Passed! and Failed! MTP summary formats

Fixes #2279

### DIFF
--- a/src/cmds/dotnet/binlog.rs
+++ b/src/cmds/dotnet/binlog.rs
@@ -88,6 +88,12 @@ static TEST_SUMMARY_RE: LazyLock<Regex> = LazyLock::new(|| {
     )
     .expect("valid regex")
 });
+static MTP_MULTILINE_SUMMARY_RE: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(
+        r"(?mi)Test\s+run\s+summary:\s*(?:Passed!|Failed!)\s*\n\s*total:\s*(?P<total>\d+)\s*\n\s*failed:\s*(?P<failed>\d+)\s*\n\s*(?:succeeded|passed):\s*(?P<passed>\d+)\s*\n\s*skipped:\s*(?P<skipped>\d+)\s*\n\s*duration:\s*(?P<duration>[^\r\n]+)"
+    )
+    .expect("valid regex")
+});
 static FAILED_TEST_HEAD_RE: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"(?m)^\s*Failed\s+(?P<name>[^\r\n\[]+)\s+\[[^\]\r\n]+\]\s*$").expect("valid regex")
 });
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased on current develop. CI's clippy now rejects the nested `if`, so the multi-line summary goes through the existing `TEST_SUMMARY_RE` block instead of a copy of it.

I also checked the format against the SDK source (`TerminalTestReporter`): a run over several assemblies lists them between the header and `total:`, and `total:` can carry `(+N retried)`. The regex now accepts both (and an `error: N` line before `total:`), with tests. Without the fix `rtk dotnet test` still prints "counts unavailable" for a summary in this format.

For the thread on #2279: develop's TRX route (`--report-trx`, MTP native mode) gives counts when the TRX extension is present, and this is the text fallback for when it isn't. I couldn't run a real `dotnet` here, so the summary format is taken from the SDK source and the demo used a stand-in.

PR #2709, opened 2026-06-29
fix(gradle): correct match_command regex to match bare gradle/gradlew

### LINKED ISSUE #2708: Bug: gradle TOML filter match_command regex never matches bare gradle or gradlew
## Description

The `match_command` regex in `src/filters/gradle.toml` is structurally broken and never activates for bare `gradle` or `gradlew` invocations — only `./gradle` and `./gradlew` work.

## Current regex

```
^(gradle|gradlew|\\./)gradlew?\\b
```

The alternation group `(gradle|gradlew|\\./)`captures the **entire** command name, then expects **another** `gradlew?` suffix immediately after. This creates impossible matches:

| Input | Group captures | Then needs | Full match attempt | Result |
|-------|----------------|------------|--------------------|--------|
| `gradle build` | `gradle` | `gradlew?` | `gradlegradle` or `gradlegradlew` | **NO MATCH** |
| `gradlew build` | `gradlew` | `gradlew?` | `gradlewgradle` or `gradlewgradlew` | **NO MATCH** |
| `./gradle build` | `./` | `gradlew?` | `./gradle` | **MATCH** |
| `./gradlew build` | `./` | `gradlew?` | `./gradlew` | **MATCH** |

Only the `./` prefix case works because it's the only branch where the suffix `gradlew?` makes sense.

## Expected behavior

All four forms should match: `gradle`, `gradlew`, `./gradle`, `./gradlew`.

## Impact

The gradle filter never activates for direct `gradle` or `gradlew` invocations, meaning users get unfiltered output with zero token savings for all Gradle commands unless they use the `./` prefix form.

## Fix

Change the regex to make `./` optional and capture the command name in one group:

```
^(\\./)?(gradlew?)\\b
```

This correctly matches all four forms.

guyoron1:
## Summary

- Fix the `match_command` regex in `src/filters/gradle.toml` which never matches bare `gradle` or `gradlew` commands
- The alternation group `(gradle|gradlew|\\./)` captured the entire command name, then required another `gradlew?` suffix — creating impossible matches like `gradlegradle`
- Only `./gradle` and `./gradlew` worked; bare `gradle build` and `gradlew build` silently fell through to unfiltered output

## Fix

Change regex from `^(gradle|gradlew|\\./)gradlew?\\b` to `^(\\./)?(gradlew?)\\b` — makes `./` prefix optional, captures command name directly.

| Input | Before | After |
|-------|--------|-------|
| `gradle build` | NO MATCH | MATCH |
| `gradlew build` | NO MATCH | MATCH |
| `./gradle build` | MATCH | MATCH |
| `./gradlew build` | MATCH | MATCH |

## Test plan

- [x] Verified regex with Python `re` module (all 4 forms match)
- [x] Existing inline `[[tests.gradle]]` blocks still pass (they test filter pipeline, not match_command)

Fixes #2708

### DIFF
--- a/src/filters/gradle.toml
+++ b/src/filters/gradle.toml
@@ -1,6 +1,6 @@
 [filters.gradle]
 description = "Compact Gradle build output — strip progress, keep tasks and errors"
-match_command = "^(gradle|gradlew|\\./)gradlew?\\b"
+match_command = "^(\\./)?(gradlew?)\\b"
 strip_ansi = true
 strip_lines_matching = [
   "^\\s*$",


src/filters/gradle.toml
```diff
@@ -1,6 +1,6 @@
 [filters.gradle]
 description = "Compact Gradle build output — strip progress, keep tasks and errors"
-match_command = "^(gradle|gradlew|\\./)gradlew?\\b"
+match_command = "^(\\./)?(gradlew?)\\b"
```
KuSh, line 3: **Blocking — the hook now auto-routes unrelated tools into a lossy filter.**

`\b` after `gradlew?` matches before `-` and `.`, and this pattern feeds the hook's rewrite decision, so commands develop leaves alone are now captured:

```
rtk rewrite "gradle-profiler benchmark"   develop: (no rewrite)  this PR: rtk gradle-profiler benchmark
rtk rewrite "gradle.bat build"            develop: (no rewrite)  this PR: rtk gradle.bat build
```

They then hit this filter, whose `max_lines = 50` is a **head** cut (`toml_filter.rs:678`, `lines.truncate(max)`) with no `tail_lines`. Measured against a 201-line Gradle Profiler run:

```
raw:      201 lines, last = "Results written to profile-out/benchmark.html"
develop:  201 lines, last = "Results written to profile-out/benchmark.html"
this PR:   52 lines, last = "[see remaining: tail -n +51 ~/.local/share/rtk/tee/...]"
```

The actual result line is discarded, with no user action and no warning — the shadow check in `compile_filter` stays silent because `gradlew` is missing from `RUST_HANDLED_COMMANDS` (`toml_filter.rs:259`).

Tightening to `^(\./)?gradlew?(\.bat)?(\s|$)` fixes this specific over-match — verified it accepts `gradle build`, `gradlew build`, `./gradlew build`, `gradlew.bat build` and rejects `gradle-profiler benchmark`; `(\s|$)` is the existing repo idiom (`df`, `ps`, `ssh`, `liquibase`, `mix-*`). It does **not** fix the second problem below.

KuSh, line 3: **Blocking — `rtk gradle bootRun` loses the failure.**

This filter sorts before `spring-boot.toml` and `find_filter_in` is first-match-wins, so the widened pattern captures `gradle … bootRun` first. Against a 164-line bootRun that fails at line 131:

```
develop:   240 bytes — "ERROR ... Application run failed" ✓   "BUILD FAILED in 24s" ✓
this PR:  6779 bytes — ERROR ✗   BUILD FAILED ✗
                       (50 lines of INFO + "... (111 lines truncated)")
```

Exit code 1 survives on both, but the diagnosis is gone and the output is 28× larger — a savings regression *and* content loss.

Scope, stated honestly: this only bites a directly-typed `rtk gradle bootRun`, since the hook sends `gradle bootRun` to `rtk gradlew`. Narrow, but a regression.

The `regex` crate has no lookahead, so this can't be fixed with a negative-lookahead tweak, and renaming files won't help either: ordering is by filter *name* via `BTreeMap<String, TomlFilterDef>` (`toml_filter.rs:76`), not build.rs's filename sort, and the tiebreak is alphabetical rather than most-specific. A broad `^gradle\b` will always beat `^gradle\s+.*bootRun`.

KuSh (changes requested):
Reviewed at head `0801879b`, built and tested **merged into today's `develop`** (`5e10aa64`), not on the branch alone. Merges cleanly; CLA signed; all 12 checks green. Real binaries, real producer shims, `LC_ALL=C`.

The regex reading in #2708 is correct — `^(gradle|gradlew|\./)gradlew?\b` really does require an impossible `gradlegradle`. But the **stated impact is not**, and that changes what the fix should be.

### The premise doesn't hold

`src/discover/rules.rs:785` already rewrites the whole gradle family. Confirmed against both binaries:

```
rtk rewrite "gradle build"     → rtk gradlew build      (identical on develop and this PR)
rtk rewrite "gradle bootRun"   → rtk gradlew bootRun
rtk rewrite "gradlew build"    → rtk gradlew build
rtk rewrite "./gradlew build"  → rtk gradlew build
```

So `gradle build` was never "falling through to unfiltered output" — it goes to the Rust `gradlew_cmd`, which even falls back to the system `gradle` binary when no wrapper exists (`gradlew_cmd.rs:72-78`).

The `./` rows of the table are also wrong: `run_fallback` matches against the **basename** of `args[0]` (`main.rs:1483-1487`), so `./gradlew` arrives as `gradlew`. The old regex matched *nothing at all*, and the new `(\./)?` group is unreachable at the only call site.

`gradle.toml` has in fact never filtered a single invocation: the basename logic landed `adda2537` (2026-03-10), the file was added `8fae5b06` (2026-03-18). It was born unreachable.

### Two regressions from widening it

Details in the inline comments. In short: the widened pattern feeds the hook's rewrite decision, so `gradle-profiler` and `gradle.bat` are now actively routed into a lossy filter (201 lines → 52, result line discarded); and it shadows `spring-boot.toml`, turning a diagnosed startup failure into 50 lines of INFO noise.

### The regex is the wrong lever

`gradle.toml` is dead from **unreachability, not redundancy** — `gradlew_cmd` classifies `build`/`test`/`lint`/`dependencies`/`connected` and drops everything else into `GradlewTask::Other` → `run_passthrough`, fully unfiltered:

```
rtk gradlew bootRun   164 lines in → 164 out
rtk gradlew publish   164 lines in → 164 out
rtk gradlew build     164 lines in →   2 out
```

So the noise this filter targets is exactly the noise `gradlew_cmd` declines to touch, and routing guarantees they never meet. Widening the match_command can't fix that; it only makes the filter capture *neighbouring* tools instead.

Two coherent ways forward:

1. **Delete `src/filters/gradle.toml`.** Costs nothing — it has never filtered anything. Update `test_builtin_filter_count` 63→62 and drop `"gradle"` from the expected-names list. Leaves `spring-boot.toml` working.
2. **Make it reachable properly** — let default passthrough consult TOML filters. Filed as #3904 with the design; `FilterMode::Buffered` already exists unused for exactly this. That is core-architecture work and explicitly **not** something to ask of this PR.

Also worth noting: nothing anywhere tests *which* filter a command line selects — `run_filter_tests` only runs the `[[tests.*]]` pipeline cases and never evaluates `match_command`. A table-driven selection test would have caught both regressions at author time.

Thanks for the careful regex analysis in #2708 — the reading is right, it's the routing underneath that makes the conclusion come out differently.

guyoron1:
@KuSh you're right on all counts. I checked each claim against binaries before deleting anything: the rewrite rule already covers the whole gradle family, both TOML lookup call sites reduce `argv[0]` to its basename (so the `./` branch is unreachable at `run_fallback` and at the hook path's `strip_absolute_path`), and with `gradle.toml` in place `rtk gradle build` returns byte-identical raw output today, so it never ran.

So I went with your option 1. The regex commit is replaced by a single commit that deletes `src/filters/gradle.toml`. Two count pins needed updating rather than one: `test_builtin_filter_count` (63 -> 62) and `test_new_filter_discoverable_after_concat` (64 -> 63). `test_builtin_all_expected_filters_present` doesn't list "gradle" on current develop, so nothing changed there, and no doc mentions the filter.

Proof it changes nothing: `rtk rewrite` across nine gradle spellings and real execution of `rtk gradle build|bootRun|test` and `rtk gradlew build` are identical before and after, and `rtk gradle bootRun` keeps its spring-boot.toml filtering. I've left #3904 alone. The title and description are updated, since the old ones described the regex fix.

PR #2711, opened 2026-06-29
fix(tracking): parse RFC3339 timestamps in first_seen_days()

### LINKED ISSUE #2710: Bug: first_seen_days() always returns 0 — timestamp parse format mismatch
## Description

`Tracker::first_seen_days()` in `src/core/tracking.rs` always returns 0 because the timestamp parsing format doesn't match the storage format.

## Root Cause

Timestamps are stored via `Utc::now().to_rfc3339()` (line 423), producing strings like:
```
2026-06-28T12:00:00.123456+00:00
```

But `first_seen_days()` (line 1108) tries to parse with:
```rust
NaiveDateTime::parse_from_str(&ts, "%Y-%m-%dT%H:%M:%S")
```

This fails because:
1. Fractional seconds (`.123456`) aren't in the format string
2. Timezone offset (`+00:00`) isn't in the format string
3. `NaiveDateTime::parse_from_str` requires the entire input to match

The second fallback format (`%Y-%m-%d %H:%M:%S`) also fails (wrong separator).

Both parse attempts fail, triggering `unwrap_or_else(|_| chrono::Utc::now())` — so `first` is always "now", and `(now - now).num_days()` is always 0.

## Comparison

Line 951 in the same file correctly parses timestamps:
```rust
DateTime::parse_from_rfc3339(&row.get::<_, String>(0)?)
    .map(|dt| dt.with_timezone(&Utc))
```

## Impact

- Telemetry reports installation age as 0 days regardless of actual usage history
- Any feature relying on `first_seen_days()` gets incorrect data

## Fix

Use `DateTime::parse_from_rfc3339()` as the primary parser, keeping the existing `NaiveDateTime` formats as fallbacks for databases created by older RTK versions.

guyoron1:
## Summary

- Fix `first_seen_days()` which always returns 0 due to timestamp parse format mismatch
- Timestamps are stored as RFC3339 (`2026-06-28T12:00:00.123456+00:00`) but parsed with `NaiveDateTime` format `%Y-%m-%dT%H:%M:%S` — missing fractional seconds and timezone offset
- Both parse attempts fail silently, falling back to `Utc::now()`, making `(now - now).num_days()` always 0
- Use `DateTime::parse_from_rfc3339()` as primary parser (matching the existing correct usage on line 951 in the same file)
- Keep `NaiveDateTime` formats as fallbacks for databases created by older RTK versions

## Test plan

- [x] Added `test_first_seen_days_parses_rfc3339` test verifying the function returns non-negative days

Fixes #2710

### DIFF
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -1120,9 +1120,15 @@ impl Tracker {
             };
         match oldest {
             Some(ts) => {
-                let first = chrono::NaiveDateTime::parse_from_str(&ts, "%Y-%m-%dT%H:%M:%S")
-                    .or_else(|_| chrono::NaiveDateTime::parse_from_str(&ts, "%Y-%m-%d %H:%M:%S"))
-                    .map(|dt| dt.and_utc())
+                let first = chrono::DateTime::parse_from_rfc3339(&ts)
+                    .map(|dt| dt.with_timezone(&chrono::Utc))
+                    .or_else(|_| {
+                        chrono::NaiveDateTime::parse_from_str(&ts, "%Y-%m-%dT%H:%M:%S")
+                            .or_else(|_| {
+                                chrono::NaiveDateTime::parse_from_str(&ts, "%Y-%m-%d %H:%M:%S")
+                            })
+                            .map(|dt| dt.and_utc())
+                    })
                     .unwrap_or_else(|_| chrono::Utc::now());
                 let days = (chrono::Utc::now() - first).num_days();
                 Ok(days.max(0))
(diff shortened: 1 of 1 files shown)


mysticgohan1:
Rebased on current develop. The fix itself still applies; the conflict was only in the test module.

I replaced my original test: it asserted `days >= 0`, which the function's own `.max(0)` guarantees even with the bug, so it proved nothing. It now inserts a 30-day-old RFC3339 row and asserts 30 (it reports 0 without the fix), plus a second test for the two legacy `NaiveDateTime` formats so the fallback stays covered. The new tests also use `Tracker::new_in_memory()` instead of `Tracker::new()`, which wrote to the developer's real tracking DB.

guyoron1:
Rebased on current develop. The fix itself still applies; the conflict was only in the test module.

I replaced my original test: it asserted `days >= 0`, which the function's own `.max(0)` guarantees even with the bug, so it proved nothing. It now inserts a 30-day-old RFC3339 row and asserts 30 (it reports 0 without the fix), plus a second test for the two legacy `NaiveDateTime` formats so the fallback stays covered. The new tests also use `Tracker::new_in_memory()` instead of `Tracker::new()`, which wrote to the developer's real tracking DB.

PR #2713, opened 2026-06-29
fix(filter): handle Python single-line docstrings in MinimalFilter

### LINKED ISSUE #2712: Bug: Python single-line docstring corrupts MinimalFilter state — subsequent comments not stripped
## Description

The `MinimalFilter` in `src/core/filter.rs` fails to handle Python single-line docstrings correctly. When a docstring opens and closes on the same line (e.g., `"""Return the sum."""`), the `in_docstring` flag is toggled to `true` but never toggled back to `false`, causing all subsequent lines to be treated as docstring content.

## Reproduction

Input:
```python
def add(a, b):
    """Return the sum."""
    # this comment should be stripped
    return a + b
```

Expected output (minimal filter should strip the `# comment` line):
```python
def add(a, b):
    """Return the sum."""
    return a + b
```

Actual output (comment is preserved because `in_docstring` is stuck `true`):
```python
def add(a, b):
    """Return the sum."""
    # this comment should be stripped
    return a + b
```

## Root Cause

Line 190-194 in `filter.rs`:
```rust
if *lang == Language::Python && trimmed.starts_with("\"\"\"") {
    in_docstring = !in_docstring;  // toggles to true, but never checks for closing """ on same line
    ...
}
```

For `"""Return the sum."""`, the line starts with `"""` so `in_docstring` toggles from `false` to `true`. The closing `"""` at the end of the same line is never detected.

## Impact

- `rtk read --filter minimal` on Python files with single-line docstrings fails to strip comments in subsequent code regions
- Reduces token savings (code is kept, not dropped — not a data-loss bug)
- Affects any Python file using the common pattern `"""One-line docstring."""`

## Fix

After toggling `in_docstring` on, check if the closing `"""` appears later on the same line.

guyoron1:
## Summary

- Fix `MinimalFilter` state corruption when encountering Python single-line docstrings like `"""Return the sum."""`
- The `in_docstring` flag was toggled to `true` but never reset when the closing `"""` appeared on the same line
- All subsequent lines were treated as docstring content, preventing comment stripping and reducing token savings

## Fix

After toggling `in_docstring` on, check if the rest of the line (after the opening `"""`) contains a closing `"""`:

```rust
if !in_docstring {
    in_docstring = true;
    if trimmed.len() > 3 && trimmed[3..].contains("\"\"\"") {
        in_docstring = false;
    }
} else {
    in_docstring = false;
}
```

## Test plan

- [x] Added `test_python_single_line_docstring_does_not_corrupt_state` test
- [x] Verifies: docstring content kept, subsequent comment stripped, code preserved

Fixes #2712

### DIFF
--- a/src/core/filter.rs
+++ b/src/core/filter.rs
@@ -185,7 +185,14 @@ impl FilterStrategy for MinimalFilter {
 
             // Handle Python docstrings (keep them in minimal mode)
             if *lang == Language::Python && trimmed.starts_with("\"\"\"") {
-                in_docstring = !in_docstring;
+                if !in_docstring {
+                    in_docstring = true;
+                    if trimmed.len() > 3 && trimmed[3..].contains("\"\"\"") {
+                        in_docstring = false;
+                    }
+                } else {
+                    in_docstring = false;
+                }
                 result.push_str(line);
                 result.push('\n');
                 continue;
@@ -543,5 +550,25 @@ fn main() {
         let input = "a\nb\nc";
         let output = smart_truncate(input, 3, &Language::Unknown);
         assert_eq!(output, input);
+    }
+
+    #[test]
+    fn test_python_single_line_docstring_does_not_corrupt_state() {
+        let filter = MinimalFilter;
+        let input = r#"def add(a, b):
+    """Return the sum."""
+    # this comment should be stripped
+    return a + b
+"#;
+        let output = filter.filter(input, &Language::Python);
+        assert!(
+            output.contains("Return the sum"),
+            "docstring should be kept"
+        );
+        assert!(
+            !output.contains("this comment should be stripped"),
+            "comment after single-line docstring must be stripped"
+        );
+        assert!(output.contains("return a + b"), "code must be kept");
     }
 }
--- a/src/core/filter.rs
+++ b/src/core/filter.rs
@@ -201,6 +201,9 @@ impl FilterStrategy for MinimalFilter {
             if in_docstring {
                 result.push_str(line);
                 result.push('\n');
+                if trimmed.ends_with("\"\"\"") {
+                    in_docstring = false;
+                }
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


pszymkowiak (changes requested):
Tested extensively with the compiled binary. The targeted fix works correctly: a docstring that opens and closes on the same line (`"""Return the sum."""`) no longer corrupts state — confirmed with single and sequential occurrences, plus the empty-docstring edge case.

However, the same class of bug (from #2712) still reproduces for an equally common style: a multi-line docstring whose closing `"""` is appended directly to content rather than sitting alone on its own line — e.g.:
```python
def subtract(self, a, b):
    """Subtract b from a
    and return result."""
    # subtract comment
    return a - b

def multiply(self, a, b):
    # multiply comment, no docstring before
    return a * b
```
This outputs both `# subtract comment` *and* `# multiply comment` unstripped — the second one belongs to a completely different, docstring-less method, showing the corruption persists across the rest of the file exactly like the original bug, just triggered by a different closing style. Root cause: the fix only handles the open+close-on-one-line case; a closing `"""` line whose trimmed text doesn't *start* with `"""` never reaches the toggle logic and falls straight into the unconditional `if in_docstring { push verbatim }` path, so `in_docstring` never resets. It happens to "self-heal" the next time another line starts with `"""` (regardless of whether that's semantically related), but if none follows, corruption persists to EOF — as in this example.

Given the docstring-with-trailing-close style is arguably more common in real Python code than the single-line case this PR targets, I'd consider #2712 still open rather than fully fixed. `cargo fmt`, `cargo clippy --all-targets`, `cargo test --all` are otherwise clean (2 unrelated pre-existing locale failures).

guyoron1:
Pushed a fix addressing the multi-line docstring closing case — the `in_docstring` verbatim block now checks for `"""` at end of content lines (e.g. `and return result."""`), not just lines starting with `"""`. Added a test reproducing the exact scenario from your review. CI should be green shortly.

guyoron1:
Thanks for the thorough testing @pszymkowiak — the case you found (multi-line docstring whose closing `"""` sits on a content line) is now handled: while in docstring state, a line ending with `"""` exits the state. Added `test_python_multiline_docstring_closing_on_content_line` covering your exact repro (docstring kept, trailing `# subtract comment` stripped, following method unaffected). Rebased on latest develop; all filter tests pass.

guyoron1:
@pszymkowiak rebased on current develop and ready for another look.

Your repro (a multi-line docstring whose closing `"""` sits on a content line, followed by a docstring-less method) is fixed. I also tried the same failure one closing style further out and it was still broken: a closing `"""` with trailing text (`and return result."""  # note`) missed the toggle too and left `in_docstring` stuck to EOF. Rather than add a third special case, the second commit now replaces both branches with one rule: count the `"""` on the line, and an odd count flips the state. It's net shorter than what's on develop and covers open, close, open + close on one line, closing appended to content, closing with trailing text, and the empty `""""""` docstring. Five tests, all failing on develop.

To re-check by hand, use a real `.py` file (`rtk read <file>.py -l minimal`). From stdin (`rtk read - -l minimal`) the language resolves to Unknown, so no Python handling runs at all.

Two things I noticed while poking at it, both pre-existing and unrelated to this PR: `r"""..."""` docstrings are dropped entirely and `'''` docstrings aren't recognised. Happy to file those separately.

Heads-up: #2715 edits the same function (adjacent hunks), so whichever merges second needs a mechanical rebase.

PR #2715, opened 2026-06-29
fix(filter): preserve code around inline block comments

### LINKED ISSUE #2714: Bug: MinimalFilter drops code on lines with inline block comments (code loss)
## Description

The `MinimalFilter` in `src/core/filter.rs` drops valid code when a line contains an inline block comment. Lines like `int x = 5; /* comment */` are entirely skipped — the code before the `/*` is lost.

## Reproduction

Input (C/Java/JS/TS/Go/Rust):
```c
int x = 5; /* inline comment */
int y = 10;
```

Expected output (minimal filter strips comment, keeps code):
```c
int x = 5;
int y = 10;
```

Actual output (entire line dropped):
```c
int y = 10;
```

### Also affects multi-line comment closings

Input:
```c
/* multi-line
   comment */ int z = 3;
int w = 4;
```

Expected: `int z = 3;` and `int w = 4;` both kept.
Actual: `int z = 3;` is dropped.

## Root Cause

Lines 174-186 in `filter.rs` set `in_block_comment = true` when `/*` is found, then unconditionally `continue` (skip the entire line), even when the comment is only part of the line.

## Impact

- **Code loss**: `rtk read --filter minimal` silently drops valid code on lines with inline block comments
- Affects all C-style languages: C, C++, Java, JavaScript, TypeScript, Go, Rust
- Particularly dangerous because the code is silently removed — no warning or error

## Fix

Restructure the block comment handler to:
1. When `/*` and `*/` appear on the same line, keep code before `/*` and after `*/`
2. When `/*` opens without `*/`, keep code before `/*` and enter block comment mode
3. When `*/` closes a multi-line comment, keep code after `*/`

guyoron1:
## Summary

- Fix `MinimalFilter` silently dropping valid code on lines with inline block comments
- `int x = 5; /* comment */` was entirely skipped — code before `/*` was lost (code loss bug)
- Also affected closing lines of multi-line comments: `comment */ int z = 3;` dropped `int z = 3;`
- Restructured block comment handler to extract and keep code before `/*` and after `*/`, stripping only the comment itself

## Before / After

| Input | Before | After |
|-------|--------|-------|
| `int x = 5; /* inline */` | line dropped | `int x = 5;` kept |
| `/* multi\nline */ int z = 3;` | `int z = 3;` dropped | `int z = 3;` kept |
| `/* full line comment */` | line dropped | line dropped (correct) |

## Test plan

- [x] Added `test_inline_block_comment_preserves_code` — verifies code before `/*` is kept
- [x] Added `test_multiline_block_comment_closing_preserves_code_after` — verifies code after `*/` is kept
- [x] Added `test_full_line_block_comment_still_stripped` — verifies full-line comments still removed

Fixes #2714

### DIFF
--- a/src/core/filter.rs
+++ b/src/core/filter.rs
@@ -169,15 +169,45 @@ impl FilterStrategy for MinimalFilter {
 
             // Handle block comments
             if let (Some(start), Some(end)) = (patterns.block_start, patterns.block_end) {
+                if in_block_comment {
+                    if let Some(end_idx) = line.find(end) {
+                        in_block_comment = false;
+                        let after = line[end_idx + end.len()..].trim();
+                        if !after.is_empty() {
+                            result.push_str(after);
+                            result.push('\n');
+                        }
+                    }
+                    continue;
+                }
                 if !in_docstring
                     && trimmed.contains(start)
                     && !trimmed.starts_with(patterns.doc_block_start.unwrap_or("###"))
                 {
-                    in_block_comment = true;
-                }
-                if in_block_comment {
-                    if trimmed.contains(end) {
-                        in_block_comment = false;
+                    if let Some(start_idx) = line.find(start) {
+                        let before = line[..start_idx].trim_end();
+                        let after_start = &line[start_idx + start.len()..];
+                        if let Some(end_idx) = after_start.find(end) {
+                            let after = after_start[end_idx + end.len()..].trim();
+                            if !before.is_empty() || !after.is_empty() {
+                                if !before.is_empty() {
+                                    result.push_str(before);
+                                }
+                                if !after.is_empty() {
+                                    if !before.is_empty() {
+                                        result.push(' ');
+                                    }
+                                    result.push_str(after);
+                                }
+                                result.push('\n');
+                            }
+                        } else {
+                            in_block_comment = true;
+                            if !before.is_empty() {
+                                result.push_str(before);
+                                result.push('\n');
+                            }
+                        }
                     }
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Tested extensively with the compiled binary across 15 popular languages (Python, JavaScript, TypeScript, Java, C#, C++, C, Go, Rust, PHP, Ruby, Swift, Kotlin, SQL, Shell).

Good news: multi-line block comments with code before `/*` and after `*/` on the closing line work correctly everywhere tested — this is the PR's core stated fix and it holds up well, including 3+ line spans. Full-line comments still dropped correctly. No regression on trailing `//`/`#` comments (never stripped by `MinimalFilter`, with or without this change) or Python docstrings (intentionally preserved verbatim by a separate, pre-existing code path).

Found one gap, reproduces identically across every `/* */`-based language tested (generic to `MinimalFilter`, not language-specific): only the *first* block comment on a line is stripped — everything after it passes through untouched. Confirmed with 10 comments on one line (C): comment #1 removed, #2–10 all leak through verbatim with their markers.
```
$ rtk read - -l minimal <<< 'int v1=1; /* c1 */ int v2=2; /* c2 */ ... int v10=10; /* c10 */'
int v1 = 1; int v2 = 2; /* c2 */ ... int v10 = 10; /* c10 */
```
Scope note: only reachable via explicit `rtk read -l minimal` or `-l aggressive` (which chains through `MinimalFilter`) — `rtk read` defaults to `-l none` (no filtering), and no other rtk command uses this code path, so this doesn't affect default usage.

Not a regression either way (previously the whole line would've been dropped, losing the code too), but leaves literal comment markers in output the filter is supposed to remove. The `after` remainder needs to loop back through comment-stripping instead of being pushed verbatim after the first match. Not blocking — `cargo fmt`, `cargo clippy --all-targets`, `cargo test --all` all clean otherwise (2 unrelated pre-existing locale failures, same as other PRs today).

guyoron1:
Thanks for the thorough testing across 15 languages! The multiple-block-comments-on-one-line gap you found makes sense — it's a pre-existing limitation, not a regression. Happy to tackle it as a follow-up PR if you'd like. Is this one good to merge as-is?

guyoron1:
@pszymkowiak thanks for the 15-language sweep. I fixed the gap you found in this PR rather than deferring it: the `after` remainder now goes back through comment-stripping instead of being pushed verbatim, so every `/* ... */` on a line is stripped and the code between them kept. Your ten-comment line comes out as `int v1=1; int v2=2; ... int v10=10;` (develop drops the whole line; the version I pushed before leaked #2 to #10). An unterminated `/*` at the end of a line still carries state to the next line.

Two new tests cover several comments on one line and a terminated + unterminated mix; four of the five block-comment tests fail on develop. I re-ran a Rust / Go / JS / Python sweep and doc comments and docstrings come out byte-identical.

One thing I did not change: code after a `*/` that closes a multi-line comment still loses its original indentation, same as the version you reviewed.

Rebased on current develop. This edits the same function as #2713, so whichever merges second needs a mechanical rebase.

Issue #2563, opened 2026-06-24
rg with escaped regex or pipe returns false "0 matches" via hook — near-miss on deleting correct data


szsa:
When rtk intercepts/compresses rg (ripgrep) output through the Claude Code command hook, invocations that use (a) escaped regex metacharacters (rg "394\.482\.063") or
  (b) pipes / redirections (rg ... | head, > file) incorrectly return "0 matches". Sometimes the result is self-contradictory: rg -c reports a positive count while rg -n
  reports 0 for the identical pattern on the identical file. The matched content genuinely exists.
  
  Environment
  - rtk 0.42.1 (~/.local/bin/rtk, 9.5M binary, Linux x86_64, kernel 5.10)
  - Invoked as a Claude Code command hook (auto-rewrites Bash tool commands)
  - Note: rtk --help lists no native rg/grep subcommand, so rg output is compressed via the hook's generic pipeline path, not a native subcommand.

  Reproduce

  Inside a Claude Code session where rtk is the active command hook, against a file that verifiably contains the literal string 417.506.652:

  # via rtk hook (rg is rewritten/intercepted) → WRONG
  rg "417\.506\.652|605\.857\.274|1\.896\.505" somefile.md
  # rtk reports: "0 matches"

  # bypassing rtk → correct
  /usr/bin/grep -E "417\.506\.652|605\.857\.274|1\.896\.505" somefile.md
  # 120 real matches

  Also reproducible: contradictory counts on the same pattern — rg -c returns N (>0) while rg -n returns 0.

  Expected behavior
  rtk must faithfully reflect the real match count. It should never emit "0 matches" when matches exist.

  Actual behavior
  False "0 matches" on escaped-regex and piped/redirected rg calls.

  Suspected root cause
  The hook's generic output-compression path (used for commands without a native rtk subcommand, such as rg) appears to break on backslash-escaped regex metacharacters
  (\.) and/or on shell pipes/redirections, producing an empty compressed result that is then reported as "0 matches."

  Impact (severe — silent data-integrity risk)
  Because rtk is a transparent token-saving proxy, a false "0 matches" is indistinguishable from a genuine no-match. In an LLM-agent context this is the most dangerous
  failure mode: real output silently rewritten to look absent reads as an LLM "hallucination" and triggers destructive "corrections."

  Concrete near-miss during financial-report analysis:
  1. Adversarially verifying numbers extracted by a sub-agent (guarding against hallucinated figures).
[...]

PR #2716, opened 2026-06-29
fix(search): surface error exit codes and guard against false zero matches

guyoron1:
## Summary

Two defensive improvements to the rg/grep search handler that address the root patterns behind #2563:

1. **Surface error exit codes to stdout**: When rg/grep exits with code >= 2 (error) and stdout is empty, print an error indicator to stdout. In hook contexts (Claude Code), stderr may not be surfaced to the agent — without stdout output, the agent sees silence and may misinterpret it as "no matches"

2. **Guard against false "0 matches"**: When `total_matches` is 0 but the engine exited successfully (code 0) with non-empty output, fall through to passthrough instead of reporting "0 matches in 0 files". This catches edge cases where `parse_match_line` fails to parse an unexpected output format — the raw output is preserved rather than silently replaced with a false zero count

## Context

Issue #2563 reported false "0 matches" with escaped regex patterns via the hook pipeline. The original issue was on rtk 0.42.1 which lacked a native search handler; the current native handler (added later) correctly passes patterns through. These changes add defensive guards so that if parsing ever fails partially, the handler falls back to passthrough rather than emitting misleading results.

## Test plan

- [x] `cargo test --all` passes (all 11 search tests pass including `grep_and_rg_use_their_own_regex_dialect`)

Relates to #2563

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -578,7 +578,13 @@ pub fn run(
     }
 
     if result.stdout.trim().is_empty() {
-        timer.track(&real_cmd, &rtk_label, &raw_output, "");
+        if exit_code >= 2 && !result.stderr.is_empty() {
+            let msg = format!("{}: error (exit {})\n", engine.label(), exit_code);
+            print!("{}", msg);
+            timer.track(&real_cmd, &rtk_label, &raw_output, &msg);
+        } else {
+            timer.track(&real_cmd, &rtk_label, &raw_output, "");
+        }
         return Ok(exit_code);
     }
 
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for looking at the search error paths. Both guards landed with #2641 on 2026-06-26: the engine's own stderr and exit code are now forwarded verbatim in every path, including the partial-match-plus-error case, and unparsed output falls back to passthrough before any parsing, so a parse miss can no longer turn into a false zero-match report. One note on the first hunk: the synthetic stdout line was removed on purpose in that same change, and tests/search_error_test.rs now asserts rtk adds no such line, so it would need a separate discussion to come back. If you are still seeing a silent no-match where grep or rg reported an error, please comment and we can reopen.

PR #2717, opened 2026-06-29
fix(core): decode process output using Windows console code page

### LINKED ISSUE #2452: Windows Encoding Bug: RTK Decodes GBK Output as UTF-8

guyoron1:
## Summary

On Windows with non-UTF-8 console code pages (e.g., code page 936/GBK for Chinese locale), child process output is mis-decoded by `String::from_utf8_lossy()`, producing mojibake. `rtk proxy` works correctly because it bypasses filtering, confirming the bug is in RTK's decode path.

This PR adds a centralized `decode_process_output()` helper that:
1. Tries strict UTF-8 first (fast path — covers Unix, macOS, and Windows with `chcp 65001`)
2. On Windows, detects the console output code page via `GetConsoleOutputCP()` and decodes using `encoding_rs`
3. Falls back to lossy UTF-8 for unknown code pages (preserves existing behavior)

Supports all common Windows code pages: GBK (936), Big5 (950), Shift-JIS (932), EUC-KR (949), Windows-125x, ISO-8859-x, KOI8-R/U.

### Changes

- **`Cargo.toml`**: Add `encoding_rs` as a Windows-only dependency
- **`src/core/utils.rs`**: Add `decode_process_output()`, `windows_console_output_cp()` (raw FFI), and `codepage_to_encoding()` with tests
- **`src/core/stream.rs`**: Replace `from_utf8_lossy` in `exec_capture()` and `exec_capture_stdin()`
- **`src/main.rs`**: Replace `from_utf8_lossy` in TOML filter path and proxy streaming path

### Scope

This PR fixes the 8 core decode sites that handle the reported reproduction case (`rtk type test.txt` with GBK-encoded file). Module-specific call sites (go_cmd, aws_cmd, etc.) and the streaming `BufReader::lines()` path are left for follow-up PRs.

### Zero regression on Unix/macOS

The `#[cfg(windows)]` blocks compile to nothing on non-Windows platforms. On Unix/macOS, `decode_process_output()` degrades to exactly the existing `from_utf8_lossy` behavior.

## Test plan

- [x] Cross-platform tests: valid UTF-8, Chinese UTF-8, empty input, invalid UTF-8 fallback
- [x] Windows-specific tests (cfg-gated): GBK decode, codepage lookup, UTF-8 code page returns None
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` passes
- [ ] Manual verification on Windows with `chcp 936` (needs Windows environment)

Fixes #2452

### DIFF
--- a/src/cmds/cloud/aws_cmd.rs
+++ b/src/cmds/cloud/aws_cmd.rs
@@ -243,8 +243,8 @@ fn run_generic(subcommand: &str, args: &[String], verbose: u8, full_sub: &str) -
     }
 
     let output = cmd.output().context("Failed to run aws CLI")?;
-    let raw = String::from_utf8_lossy(&output.stdout).to_string();
-    let stderr = String::from_utf8_lossy(&output.stderr).to_string();
+    let raw = crate::core::utils::decode_process_output(&output.stdout);
+    let stderr = crate::core::utils::decode_process_output(&output.stderr);
 
     if !output.status.success() {
         timer.track(
@@ -316,8 +316,8 @@ fn run_aws_json(
     let output = cmd
         .output()
         .context(format!("Failed to run {}", cmd_desc))?;
-    let stdout = String::from_utf8_lossy(&output.stdout).to_string();
-    let stderr = String::from_utf8_lossy(&output.stderr).to_string();
+    let stdout = crate::core::utils::decode_process_output(&output.stdout);
+    let stderr = crate::core::utils::decode_process_output(&output.stderr);
 
     if !output.status.success() {
         eprintln!("{}", stderr.trim());
@@ -388,8 +388,8 @@ fn run_s3_ls(extra_args: &[String], verbose: u8) -> Result<i32> {
     }
 
     let output = cmd.output().context("Failed to run aws s3 ls")?;
-    let stdout = String::from_utf8_lossy(&output.stdout).to_string();
-    let stderr = String::from_utf8_lossy(&output.stderr).to_string();
+    let stdout = crate::core::utils::decode_process_output(&output.stdout);
+    let stderr = crate::core::utils::decode_process_output(&output.stderr);
     let raw = if stderr.is_empty() {
         stdout.clone()
     } else {
@@ -438,8 +438,8 @@ fn run_s3_transfer(operation: &str, extra_args: &[String], verbose: u8) -> Resul
     let output = cmd
         .output()
         .context(format!("Failed to run {}", cmd_label))?;
-    let stdout = String::from_utf8_lossy(&output.stdout).to_string();
-    let stderr = String::from_utf8_lossy(&output.stderr).to_string();
+    let stdout = crate::core::utils::decode_process_output(&output.stdout);
+    let stderr = crate::core::utils::decode_process_output(&output.stderr);
     let raw = if stderr.is_empty() {
         stdout.clone()
     } else {
--- a/src/cmds/cloud/aws_cmd.rs
+++ b/src/cmds/cloud/aws_cmd.rs
@@ -4,11 +4,12 @@
 //! Specialized filters for high-frequency commands (STS, S3, EC2, ECS, RDS, CloudFormation).
 
 use crate::core::guard::never_worse;
+use crate::core::stream::{exec_capture, CaptureResult};
 use crate::core::tee::force_tee_hint;
 use crate::core::tracking;
 use crate::core::truncate::{CAP_INVENTORY, CAP_LIST};
 use crate::core::utils::{
-    exit_code_from_output, exit_code_from_status, human_bytes, join_with_overflow,
+    human_bytes, join_with_overflow,
     resolved_command, shorten_arn, truncate_iso_date,
 };
 use crate::json_cmd;
@@ -242,11 +243,13 @@ fn run_generic(subcommand: &str, args: &[String], verbose: u8, full_sub: &str) -
         eprintln!("Running: aws {}", full_sub);
     }
 
-    let output = cmd.output().context("Failed to run aws CLI")?;
-    let raw = crate::core::utils::decode_process_output(&output.stdout);
-    let stderr = crate::core::utils::decode_process_output(&output.stderr);
+    let CaptureResult {
+        stdout: raw,
+        stderr,
+        exit_code,
+    } = exec_capture(&mut cmd).context("Failed to run aws CLI")?;
 
-    if !output.status.success() {
+    if exit_code != 0 {
         timer.track(
             &format!("aws {}", full_sub),
             &format!("rtk aws {}", full_sub),
@@ -254,7 +257,7 @@ fn run_generic(subcommand: &str, args: &[String], verbose: u8, full_sub: &str) -
             &stderr,
         );
         eprintln!("{}", stderr.trim());
-        return Ok(crate::core::utils::exit_code_from_output(&output, "aws"));
+        return Ok(exit_code);
     }
 
     let filtered = match json_cmd::filter_json_compact(&raw, JSON_COMPRESS_DEPTH) {
@@ -280,11 +283,7 @@ fn run_generic(subcommand: &str, args: &[String], verbose: u8, full_sub: &str) -
     Ok(0)
 }
 
-fn run_aws_json(
-    sub_args: &[&str],
-    extra_args: &[String],
-    verbose: u8,
-) -> Result<(String, String, std::process::ExitStatus)> {
+fn run_aws_json(sub_args: &[&str], extra_args: &[String], verbose: u8) -> Result<CaptureResult> {
     let mut cmd = resolved_command("aws");
     for arg in sub_args {
         cmd.arg(arg);
@@ -313,17 +312,13 @@ fn run_aws_json(
         eprintln!("Running: {}", cmd_desc);
     }
 
-    let output = cmd
-        .output()
-        .context(format!("Failed to run {}", cmd_desc))?;
-    let stdout = crate::core::utils::decode_process_output(&output.stdout);
-    let stderr = crate::core::utils::decode_process_output(&output.stderr);
+    let captured = exec_capture(&mut cmd).context(format!("Failed to run {}", cmd_desc))?;
 
-    if !output.status.success() {
-        eprintln!("{}", stderr.trim());
+    if !captured.success() {
+        eprintln!("{}", captured.stderr.trim());
     }
 
-    Ok((stdout, stderr, output.status))
+    Ok(captured)
 }
 
 /// Shared runner for AWS commands that return JSON.
@@ -338,7 +333,11 @@ fn run_aws_filtered(
     let rtk_label = format!("rtk {}", cmd_label);
     let slug = cmd_label.replace(' ', "_");
     let timer = tracking::TimedExecution::start();
-    let (stdout, stderr, status) = run_aws_json(sub_args, extra_args, verbose)?;
+    let CaptureResult {
+        stdout,
+        stderr,
+        exit_code,
+    } = run_aws_json(sub_args, extra_args, verbose)?;
 
     // Combine stdout+stderr for accurate tracking (per contract)
     let raw = if stderr.is_empty() {
@@ -347,8 +346,7 @@ fn run_aws_filtered(
         format!("{}\n{}", stdout, stderr)
     };
 
-    if !status.success() {
-        let exit_code = exit_code_from_status(&status, "aws");
+    if exit_code != 0 {
         if let Some(hint) = crate::core::tee::tee_and_hint(&raw, &slug, exit_code) {
             eprintln!("{}\n{}", stderr.trim(), hint);
         } else {
@@ -387,16 +385,17 @@ fn run_s3_ls(extra_args: &[String], verbose: u8) -> Result<i32> {
         eprintln!("Running: aws s3 ls {}", extra_args.join(" "));
     }
 
-    let output = cmd.output().context("Failed to run aws s3 ls")?;
-    let stdout = crate::core::utils::decode_process_output(&output.stdout);
-    let stderr = crate::core::utils::decode_process_output(&output.stderr);
+    let CaptureResult {
+        stdout,
+        stderr,
+        exit_code,
+    } = exec_capture(&mut cmd).context("Failed to run aws s3 ls")?;
     let raw = if stderr.is_empty() {
         stdout.clone()
     } else {
         format!("{}\n{}", stdout, stderr)
     };
-    if !output.status.success() {
-        let exit_code = exit_code_from_output(&output, "aws");
+    if exit_code != 0 {
         if let Some(hint) = crate::core::tee::tee_and_hint(&raw, "aws_s3_ls", exit_code) {
             eprintln!("{}\n{}", stderr.trim(), hint);
         } else {
@@ -435,18 +434,17 @@ fn run_s3_transfer(operation: &str, extra_args: &[String], verbose: u8) -> Resul
         eprintln!("Running: {} {}", cmd_label, extra_args.join(" "));
     }
 
-    let output = cmd
-        .output()
-        .context(format!("Failed to run {}", cmd_label))?;
-    let stdout = crate::core::utils::decode_process_output(&output.stdout);
-    let stderr = crate::core::utils::decode_process_output(&output.stderr);
+    let CaptureResult {
+        stdout,
+        stderr,
+        exit_code,
+    } = exec_capture(&mut cmd).context(format!("Failed to run {}", cmd_label))?;
     let raw = if stderr.is_empty() {
         stdout.clone()
     } else {
         format!("{}\n{}", stdout, stderr)
     };
-    if !output.status.success() {
-        let exit_code = exit_code_from_output(&output, "aws");
+    if exit_code != 0 {
         if let Some(hint) = crate::core::tee::tee_and_hint(&raw, &slug, exit_code) {
             eprintln!("{}\n{}", stderr.trim(), hint);
         } else {
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -43,8 +43,8 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     // Skip filtering on failure: curl can return HTML error bodies that would
     // be misleading to summarize, and we want the real exit code surfaced.
     if !output.status.success() {
-        let stderr_str = String::from_utf8_lossy(&output.stderr);
-        let stdout_str = String::from_utf8_lossy(&output.stdout);
+        let stderr_str = crate::core::utils::decode_process_output(&output.stderr);
+        let stdout_str = crate::core::utils::decode_process_output(&output.stdout);
         let msg = if stderr_str.trim().is_empty() {
             stdout_str.trim().to_string()
         } else {
@@ -72,7 +72,7 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         return Ok(exit_code);
     }
 
-    let raw = String::from_utf8_lossy(&output.stdout).into_owned();
+    let raw = crate::core::utils::decode_process_output(&output.stdout);
     let is_tty = std::io::stdout().is_terminal();
     let filtered = filter_curl_output(&raw, is_tty);
 
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -43,8 +43,10 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     // Skip filtering on failure: curl can return HTML error bodies that would
     // be misleading to summarize, and we want the real exit code surfaced.
     if !output.status.success() {
+        // stderr is curl's own diagnostics, which do follow the console code
+        // page. The body on stdout does not — see the note below.
         let stderr_str = crate::core::utils::decode_process_output(&output.stderr);
-        let stdout_str = crate::core::utils::decode_process_output(&output.stdout);
+        let stdout_str = String::from_utf8_lossy(&output.stdout);
         let msg = if stderr_str.trim().is_empty() {
             stdout_str.trim().to_string()
         } else {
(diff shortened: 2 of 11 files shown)


src/core/utils.rs
```diff
@@ -398,6 +398,64 @@ pub fn human_bytes(bytes: u64) -> String {
...
+    // nosemgrep: unsafe-block — read-only Win32 API, no memory or thread safety risk
+    unsafe {
+        windows_sys::Win32::System::Console::GetConsoleOutputCP()
+    }
+}
+
+#[cfg(windows)]
+fn codepage_to_encoding(cp: u32) -> Option<&'static encoding_rs::Encoding> {
+    let label = match cp {
+        936 | 54936 => "gbk",
```
KuSh: This seems wrong: 54936 (GB18030) should map to `gb18030`, or it will mishandle GB18030’s 4-byte sequences.

src/core/utils.rs
```diff
@@ -398,6 +398,64 @@ pub fn human_bytes(bytes: u64) -> String {
...
+            let (cow, _, _) = encoding.decode(bytes);
+            return cow.into_owned();
+        }
+    }
+
+    String::from_utf8_lossy(bytes).into_owned()
+}
+
+#[cfg(windows)]
+fn windows_console_output_cp() -> u32 {
```
KuSh: This is querying `rtk`’s own console code page, not the child process’s actual encoding. It also silently falls back to lossy UTF-8 with no diagnostic when unavailable, especially with piped output, which is a common case for `rtk`

src/core/utils.rs
```diff
@@ -398,6 +398,64 @@ pub fn human_bytes(bytes: u64) -> String {
...
+        return s.to_owned();
+    }
+
+    #[cfg(windows)]
+    {
+        let cp = windows_console_output_cp();
+        if let Some(encoding) = codepage_to_encoding(cp) {
+            let (cow, _, _) = encoding.decode(bytes);
+            return cow.into_owned();
+        }
```
KuSh: It seems a bit harsh to re-decodes the entire buffer as a legacy code page (GBK/Big5/etc.) at the first invalid UTF-8 byte, instead of replacing just that byte. That regresses the previous lossy-UTF-8 behavior for otherwise valid UTF-8 text.

Two options:
- Decode only when a threshold is exceeded, though that feels like a partial fix.
- Use `valid_up_to()` from `Utf8Error` returned by `std::str::from_utf8(bytes)` and walk the buffer in a loop, something like:

```rs
pub fn decode_process_output(bytes: &[u8]) -> String {
    match std::str::from_utf8(bytes) {
        Ok(s) => return s.to_owned(),
        Err(e) => {
            let valid_up_to = e.valid_up_to();
            let (valid, rest) = bytes.split_at(valid_up_to);
            let mut out = unsafe { std::str::from_utf8_unchecked(valid) }.to_owned();
            // decode `rest` via code page (Windows) or lossy UTF-8, append to `out`
            ...
        }
    }
}
```

That way, valid UTF-8 spans stay untouched, and only the actually-invalid byte runs get code page handling.

KuSh (changes requested):

This fix feels too broad and a bit heavy-handed. It also doesn’t wire up every call site yet, since most filter modules still call `String::from_utf8_lossy()` directly, so the reported bug isn’t actually fixed for most commands.

guyoron1:
Thanks @KuSh — fair points, both addressed:

1. **All call sites wired up**: `decode_process_output` now routes every child-process decode — git, go, aws, curl, system read, discover registry — on top of the original stream/main sites. The only remaining `from_utf8_lossy` calls are file-content decodes (dotnet TRX XML, hook trust snippets) where console code pages don't apply, and curl's binary passthrough (#1087) which is deliberately untouched.
2. **Scope**: the helper itself stays narrow — UTF-8 fast path first, code-page fallback only on Windows and only when UTF-8 validation fails, so non-Windows behavior is byte-identical to before.

Also rebased on latest develop (conflicts resolved). Full suite passes locally except the two pre-existing env-sensitive `unattestable_passthrough` tests, which fail on clean develop too (fix in #3147).

src/core/stream.rs
KuSh: This one doesn't use the new Windows-codepage-aware decode_process_output

src/core/utils.rs
```diff
@@ -496,6 +496,64 @@ pub fn human_bytes(bytes: u64) -> String {
...
+        1255 => "windows-1255",
+        1256 => "windows-1256",
+        1257 => "windows-1257",
+        1258 => "windows-1258",
+        28591 => "iso-8859-1",
+        28592 => "iso-8859-2",
+        20866 => "koi8-r",
+        21866 => "koi8-u",
+        65001 => return None,
+        _ => return None,
```
KuSh: You're omitting legacy OEM/DOS code pages (437, 850, ...) that are the actual default console output code page on plain cmd.exe for many non-English Windows locales

src/cmds/cloud/curl_cmd.rs
KuSh, line 64: this is catching non-UTF-8 bodies before you can decode them

src/core/utils.rs
```diff
@@ -496,6 +496,64 @@ pub fn human_bytes(bytes: u64) -> String {
...
+fn windows_console_output_cp() -> u32 {
+    #[allow(unsafe_code)]
+    // nosemgrep: unsafe-block — read-only Win32 API, no memory or thread safety risk
+    unsafe {
+        windows_sys::Win32::System::Console::GetConsoleOutputCP()
+    }
+}
+
+#[cfg(windows)]
+fn codepage_to_encoding(cp: u32) -> Option<&'static encoding_rs::Encoding> {
```
KuSh: with already two comments on that function it seems using the codepage crate instead would be a better idea (and is encoding_rs author's own recommended approach)

src/core/utils.rs
```diff
@@ -496,6 +496,64 @@ pub fn human_bytes(bytes: u64) -> String {
...
+
+    String::from_utf8_lossy(bytes).into_owned()
+}
+
+#[cfg(windows)]
+fn windows_console_output_cp() -> u32 {
+    #[allow(unsafe_code)]
+    // nosemgrep: unsafe-block — read-only Win32 API, no memory or thread safety risk
+    unsafe {
+        windows_sys::Win32::System::Console::GetConsoleOutputCP()
```
KuSh: This syscall output should be cached

src/cmds/cloud/aws_cmd.rs
```diff
@@ -243,8 +243,8 @@ fn run_generic(subcommand: &str, args: &[String], verbose: u8, full_sub: &str) -
     }
 
     let output = cmd.output().context("Failed to run aws CLI")?;
-    let raw = String::from_utf8_lossy(&output.stdout).to_string();
-    let stderr = String::from_utf8_lossy(&output.stderr).to_string();
+    let raw = crate::core::utils::decode_process_output(&output.stdout);
```
KuSh: Instead of patching all the call site, you should take the opportunity to replace them by calls to stream::exec_capture(&mut cmd) or stream::exec_capture_stdin(&mut cmd)

src/cmds/git/git.rs
KuSh: Not your fault but exacerbated by your fix: this function blindly assumes that the first line is git own summary and just finds the first ] and blindly slices from byte 1.

With git pre-commit and commit-msg hooks and now that we decode other encodings, if that line starts with a multi-byte UTF-8 character, slicing at byte offset 1 lands mid-character (not a char boundary) and will panic.

KuSh (changes requested):
Hi @guyoron1, you addressed my general comment on your PR, but not the specific inline ones, which should also be taken into account. I’ve rechecked your latest version and found a few more things.

src/core/utils.rs
guyoron1: Fixed. Decoding now works a line at a time: lines that are valid UTF-8 keep their bytes, and only lines that fail validation go through the code page. A stray byte costs its own line instead of the whole buffer.

I first tried the `valid_up_to()` walk you sketched, decoding just the invalid byte run. It breaks on GB18030: its four-byte sequences put bytes in the ASCII digit range (`81 35 F4 37`), so any rule that ends a run at the first byte below `0x80` splits the character in half — my own gb18030 test caught it. The line is the smallest unit that is safe here, since `\n` cannot be a trail byte in any of these encodings and a process does not switch encoding mid-line.

There is also a guard for the case you were really pointing at: the code page result is only accepted when it decodes cleanly, so a UTF-8 line with one corrupt byte falls back to lossy UTF-8 rather than being turned into mojibake. `test_decode_mixed_corrupt_byte_prefers_lossy_utf8` covers it.

src/core/utils.rs
guyoron1: Both parts fixed.

`GetConsoleOutputCP` returning 0 (no console — the piped case you flagged, which is how rtk normally runs under a hook) now falls back to `GetACP`, which is what a console program writing to a pipe uses. The result is cached in a `OnceLock`.

The silent part is gone too: when neither `codepage` nor `oem_cp` has a table for the page, it warns once on stderr instead of quietly going lossy.

It remains a best guess — a child can emit any encoding regardless of either code page — so I documented that explicitly. It is only ever consulted for bytes that already failed UTF-8 validation, so a wrong guess degrades to exactly the replacement characters the old lossy path produced.

src/core/utils.rs
guyoron1: Fixed by switching to the `codepage` crate: 54936 now resolves to `gb18030` rather than GBK. `test_decode_mixed_gb18030_is_not_gbk` asserts both the mapping and that a four-byte sequence decodes to a single character.

src/core/stream.rs
guyoron1: Good catch — and the irony was that this function's own doc comment describes the OEM/ANSI lines it was sending straight to U+FFFD. It now calls `decode_process_output` per line, so the streamed path and the captured path decode identically. Splitting on `\n` first is safe for every encoding involved, since none of them can produce `0x0A` as a trail byte.

src/core/utils.rs
guyoron1: Fixed by adding `oem_cp`. `codepage` alone is not enough here: encoding_rs implements only WHATWG encodings, so 437, 850 and 852 all return `None` from it — I verified that rather than assuming. So the lookup is now `codepage` first for the ANSI/DBCS pages, then `oem_cp` for the legacy OEM/DOS tables, then the warn-once lossy fallback. `test_decode_mixed_oem_codepages` covers CP437 and CP850 and pins the fact that encoding_rs does not carry them.

src/core/utils.rs
guyoron1: Agreed, and done — the hand-written match is gone in favour of `codepage`. That is also what fixed the 54936 mapping you flagged above, since the crate distinguishes GBK from gb18030 where my table did not.

One thing worth noting: `codepage` alone does not close the OEM gap from your other comment, since it can only return encodings encoding_rs implements. So the OEM/DOS pages come from `oem_cp` as a second step.

src/core/utils.rs
guyoron1: Done — cached in a `OnceLock`, so the syscall happens at most once per process.

src/cmds/cloud/aws_cmd.rs
guyoron1: Done — this and the other aws sites, plus the go sites, now go through `stream::exec_capture`, and `git commit` through `exec_capture_stdin` so its inherited stdin (editor, GPG passphrase, credential helper) still reaches the terminal. Future callers get decoding for free, which was your original point.

Two things I checked while doing it:

- `Command::output()` already defaults stdin to null, so `exec_capture` is behaviour-identical at these sites.
- Several places my first pass had patched were **test** code asserting on rtk's own UTF-8 output, where a console code page means nothing. Those went back to `from_utf8_lossy` rather than being migrated.

One deliberate regression to flag: `exec_capture` uses `status_to_exit_code`, so the sites that used `exit_code_from_output` lose its "terminated by signal N" stderr diagnostic. The exit code itself is unchanged (128+signal). Happy to add that diagnostic to `CaptureResult` if you want it kept.

src/cmds/git/git.rs
guyoron1: Fixed, and you were right that it is reachable — I reproduced the panic before touching it: `parse_commit_output("✅ lint passed]")` gives *byte index 1 is not a char boundary*. git prints hook output ahead of its own summary, so the first line is frequently not the `[branch hash]` line at all. It predates this PR, but decoding makes it easier to hit, and a lossily decoded line starts with a multi-byte U+FFFD too.

It now locates the bracket **pair** with `find`, so both indices are character boundaries, and returns "ok" when the brackets are missing or out of order. Four regression tests: multi-byte prefix, a real summary behind a multi-byte hook prefix, a U+FFFD prefix, and `]` before `[`.

I also changed the length check from `hash.len() >= 7` to `hash.chars().count() >= 7` so it agrees with the `chars().take(7)` right below it.

guyoron1:
Thanks @KuSh — you were right that I answered the general comment and left the inline ones. All ten are addressed now, each with a reply on its own thread. Summary:

**decode_process_output**
- Decodes a line at a time; valid UTF-8 lines keep their bytes. I tried your `valid_up_to()` run-walk first and it breaks GB18030 (its 4-byte sequences embed ASCII digits, so a run ending below 0x80 splits them) — the line is the smallest safe unit. A code page result is only used when it decodes cleanly, so a UTF-8 line with a corrupt byte falls back to lossy instead of mojibake.
- Hand-written table replaced with `codepage`, which also fixes 54936 → gb18030.
- `oem_cp` added for 437/850/852 — I verified `codepage` returns `None` for those, since encoding_rs only implements WHATWG encodings.
- `GetACP` fallback when there's no console, warn-once instead of silent lossy, and the lookup cached in a `OnceLock`.

**Call sites** — production sites now go through `exec_capture` / `exec_capture_stdin` instead of being patched individually. The streamed path decodes too. curl's body deliberately stays on `from_utf8_lossy` (an HTTP payload's encoding comes from the charset, not the local console) — reasoning in that thread. The `parse_commit_output` panic is fixed with a reproduction and four regression tests.

**Testability** — the mapping and the walk take the code page as a parameter, so they compile and run on every platform rather than sitting behind `#[cfg(windows)]` where no CI runner exercises them. 15 tests, including one pinning that output without a code page stays byte-identical to `from_utf8_lossy`.

fmt and clippy clean; suite 2600 passed. The two `unattestable_passthrough` failures are pre-existing and reproduce on clean develop — that's #3147, which I've also reworked along the lines you asked for there.

One judgement call worth your eyes: `exec_capture` uses `status_to_exit_code`, so migrated sites lose the "terminated by signal N" stderr diagnostic that `exit_code_from_output` printed. Exit codes are unchanged. Say the word and I'll carry it into `CaptureResult`.

src/cmds/cloud/aws_cmd.rs
guyoron1: Followed up on this: the signal diagnostic is kept rather than dropped.

Both capture helpers now go through one `capture()` that calls `exit_code_from_output`, so a child killed by a signal still gets the `[rtk] <program>: process terminated by signal N` line on stderr — a command dying to an OOM kill would otherwise just report 137 with no reason. The program name is used as the label, so no call site has to pass one, and the migrated sites end up with the diagnostic they previously had to wire up by hand.

Covered by `test_exec_capture_reports_signal_exit_code`, which terminates a child with SIGTERM and asserts the 128 + signal code survives.

src/core/utils.rs
KuSh: Seems like a good compromise as I can't think of a simple solution with valid_up_to

KuSh (approved):
Thanks a lot, LGTM

PR #2718, opened 2026-06-29
fix(hook): add output transparency section to RTK instructions

### LINKED ISSUE #2445: Silent auto-rewrite hook trips Claude Code's tampering/injection heuristics → agent distrusts rtk output, burns tokens, and halts on "safety" grounds
## Summary

When rtk is installed as a **silent auto-rewrite hook** for Claude Code, the agent increasingly treats rtk's filtered/compacted command output as **tampered, poisoned, or injected**. Over roughly the last two weeks this has become frequent enough to (a) **waste tokens** — the agent re-runs the raw command, second-guesses the result, or writes defensive workarounds — and (b) in several cases **halt work entirely on "safety/integrity" grounds**, refusing to trust output it believes was manipulated.

The mechanism collides head-on with rtk's own **Transparency** design principle (`CONTRIBUTING.md`):

> The LLM doesn't know RTK is involved... RTK's output must be a valid, useful subset of the original tool's output, not a different format the LLM wouldn't expect... The filtered output should be indistinguishable from "a shorter version of the real command."

In practice the rewrite is *not* transparent to a modern, safety-tuned agent: the agent issues `grep ...` / `git diff ...`, the hook silently runs `rtk grep` / `rtk git diff`, and the result comes back **regrouped, truncated, header-annotated, or with `+N more` / "removed …" hints** that do not match what the named command actually emits. A model trained to be alert to prompt-injection and tool-output tampering reads that divergence as *the output was modified by something in the middle* — which, literally, it was.

## Environment

- Agent: **Claude Code 2.1.177**, model Claude Opus 4.x (1M context)
- rtk: installed globally via the Claude Code **PreToolUse auto-rewrite hook** (silent `cmd` → `rtk cmd`). Exact rtk version not recorded — it has since been uninstalled, see "Note" below.
- OS: macOS (darwin)
- Also present: rtk injected a `<!-- rtk-instructions v2 -->` block into the global `~/CLAUDE.md` instructing the agent to prefix every command with `rtk`.

## Observed behavior (symptom pattern)

Repeated, across many sessions over ~2 weeks:

1. Agent runs a normal command (e.g. `grep -rn foo src/`, `git diff`, `git log`, `cargo test`).
2. Hook silently rewrites it to the `rtk` equivalent; output comes back **structurally different** from the genuine command — grep results grouped-by-file, diffs ultra-compacted, lists capped with a `+N more` / "removed …" hint, or RTK-style headers.
3. Agent flags the result as **suspicious / possibly tampered / possibly a prompt-injection attempt**, because the output it received does not match the output the command it *issued* should produce.
4. Consequence: extra tokens (re-run raw, investigate, hedge) and, in the worst cases, the agent **stops and declines to continue**, citing output integrity / safety.

This recurred often enough that, whenever output integrity mattered, I started routing searches around the hook — because rtk's **regrouped / truncated grep output was breaking downstream parsing and feeding the agent's suspicion** rather than helping it.

## Root cause hypothesis

Two compounding factors:

1. **Silent rewriting + non-identical output.** The agent has no signal that rtk is involved, yet the output is demonstrably not what the issued command produces. To a safety-tuned model, "I ran X, but the output isn't what X produces" is a textbook tampering/injection signal. The "indistinguishable subset" assumption in the Transparency principle does not hold for filters that **regroup, reorder, add headers, or truncate-with-hints** (grep grouped-by-file, compact diff, `+N more`, "removed N lines"): those are a *different shape*, not merely a shorter one.
2. **rtk's own injected markers/instructions in context.** The `<!-- rtk-instructions … -->` block in `CLAUDE.md` plus silent command rewriting means the agent sees both "my commands are being altered" and "my outputs don't match" — together this reads as active context manipulation, which is exactly what the agent's safety heuristics are tuned to resist.

## Impact

- **Token cost goes up, not down**, in these sessions — the opposite of rtk's goal — because the agent re-runs commands raw, narrates suspicion, and adds defensive steps.
- **Task abandonment**: the agent sometimes refuses to proceed on integrity grounds, which is a hard failure, not just inefficiency.

## Suggested directions (non-prescriptive)

- **Make the rewrite legible to the agent** rather than silent: e.g. a short, stable, machine-recognizable prefix line on filtered output (`[rtk: filtered grep, grouped by file, N hidden]`) so the agent attributes the transformation to a known tool instead of to tampering. This trades a little "invisibility" for trust — which, for safety-tuned agents, is the better trade.
- **Be conservative about shape-changing filters under hook mode.** Filters that only shorten (drop passing tests, strip progress bars) are safe; filters that **regroup/reorder/truncate-with-hints** (grep grouped-by-file, compact diff) are the ones most likely to read as tampering. Consider passing those through unfiltered when invoked via the silent hook, or gating them behind explicit `rtk <cmd>` opt-in only.
- **Document the interaction** with modern agent safety behavior in the README/troubleshooting, and offer a one-flag "high-trust / low-surprise" hook profile.

## Note

I uninstalled rtk before filing (removed the `~/CLAUDE.md` rtk block, the `~/.local/bin/rtk` binary, and `~/.rtk/`) precisely because of this behavior, so I can't attach exact rtk version or captured transcripts. I'm filing because the *pattern* is consistent and the root cause looks structural (silent rewrite + shape-changing output vs. agent integrity heuristics), and it directly tensions with the project's stated Transparency principle. Happy to reproduce with a fresh install and capture concrete before/after transcripts if that would help triage.

guyoron1:
## Summary

When RTK is installed as a silent PreToolUse auto-rewrite hook, safety-tuned LLMs can flag RTK's filtered output as tampered because the output shape differs from what the original command produces (e.g., grep results grouped by file, diffs compacted, lists truncated with `+N more`).

This PR adds an "Output Transparency" section to the `RTK_INSTRUCTIONS` constant (injected into CLAUDE.md by `rtk init`). The section explicitly tells the LLM that:
- RTK intentionally compresses and restructures output
- Output may look different from raw commands — this is expected and correct
- `rtk proxy <cmd>` is available for exact unfiltered output

This converts "unexplained output mismatch" into "expected behavior from a known tool," addressing the most common trigger for tamper heuristics.

### Changes

- **`src/hooks/init.rs`**: Add 4-line "Output Transparency" section to `RTK_INSTRUCTIONS`

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` passes
- [x] Existing `test_init_contains_key_commands` and `test_init_has_version_marker` tests pass
- [x] No changes to filter pipeline, hook processor, or output format

Fixes #2445

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -133,6 +133,10 @@ git add . && git commit -m "msg" && git push
 rtk git add . && rtk git commit -m "msg" && rtk git push
 ```
 
+## Output Transparency
+
+RTK intentionally compresses and restructures command output to save tokens. Filtered output may look different from raw command output — for example, grep results grouped by file, test output showing only failures, or logs deduplicated with counts. This is expected and correct. If you need the exact unfiltered output of any command, use `rtk proxy <cmd>`.
+
 ## RTK Commands by Workflow
 
 ### Build & Compile (80-90% savings)


guyoron1:
Closing this one: it has landed upstream in a better form. The RTK awareness text was rewritten into `hooks/rtk-awareness.md`, `rtk-awareness-high.md` and `rtk-awareness-full.md` (9620f66c and follow-ups), and all three levels now open with the same output-transparency guidance this PR adds ("treat it as the complete result ... re-run as `rtk proxy <cmd>` only when the result is unusable"). The `-high` and `-full` levels also explain the hook itself, which covers the other half of #2445.

The constant this PR edited, `RTK_INSTRUCTIONS`, is now only reached by the legacy `--claude-md` mode, so the added section would only improve a path users are being moved off. Thanks!

PR #2743, opened 2026-06-30
fix(init): correct RTK.md summary output

### LINKED ISSUE #2734: rtk init: generated RTK.md has a circular reference and the run miscounts its line total
Two small cosmetic issues in `rtk init` (0.43.0):

1. The generated `RTK.md` ends with:
   > Refer to CLAUDE.md for full command reference.

   But the same `rtk init` writes `@RTK.md` into CLAUDE.md (CLAUDE.md imports RTK.md). So the pointer is circular and resolves to nothing — there is no separate command reference in CLAUDE.md. Suggest dropping the line, or pointing it at `rtk --help`.

2. `rtk init` prints `RTK.md: ... (10 lines)` while the file it actually writes is 29 lines. The reported count is wrong.

Neither affects function; just confusing to a reader.

ekkoitac:
## Summary
- replace the circular RTK.md pointer to CLAUDE.md with a direct `rtk --help` command reference
- report the generated RTK.md line count from the embedded template instead of hard-coding `10 lines`
- add regression coverage for the slim RTK.md reference and line-count helper

Fixes #2734

## Validation
- `cargo test --all`
- manual smoke: `CLAUDE_CONFIG_DIR=$(mktemp -d) cargo run --quiet -- init -g --auto-patch` prints `RTK.md ... (29 lines)`; generated `RTK.md` is 29 lines and ends with `Run \`rtk --help\` for the full command reference.`

### DIFF
--- a/hooks/claude/README.md
+++ b/hooks/claude/README.md
@@ -8,7 +8,7 @@
 - Returns `updatedInput` JSON for transparent command rewrite (agent doesn't know RTK is involved)
 - Exits silently (exit 0) on any failure: jq missing, rtk missing, rtk too old (< 0.23.0), no match
 - Version guard checks `rtk --version` against minimum 0.23.0
-- `rtk-awareness.md` is a slim 10-line instructions file embedded into CLAUDE.md by `rtk init`
+- `rtk-awareness.md` is written as `RTK.md` and referenced from `CLAUDE.md` by `rtk init`
 
 ## Testing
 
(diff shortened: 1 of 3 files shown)


KuSh:
Thanks for the fix. Both problems from #2734 were resolved by #3372, merged 2026-09-07: the awareness file moved to hooks/rtk-awareness.md and was rewritten with default, high and full levels, so the pointer back to CLAUDE.md is gone, and the rtk init summary now prints the awareness level instead of a hard-coded line count, in both the fresh-install and the migration message. The README note this PR corrected was updated in the same change. Since the file this PR patches no longer exists at that path, it cannot be rebased as is, but if you see the circular reference or a wrong count on the current develop, please comment and we can reopen.

Issue #2721, opened 2026-06-29
feat(rewrite): auto-rewrite TOML-only commands invoked via binary paths (.bin/, absolute paths)


denis-peshkov:
## Summary

TOML built-in filters (e.g. `make.toml`, future `bru.toml`, `ng-test.toml`) work when the user runs `rtk <tool> …` directly, but **Cursor/Claude hooks do not rewrite** equivalent invocations via:

- `node_modules/.bin/<tool> …`
- `/usr/local/bin/<tool> …`
- `../../client/node_modules/.bin/bru run …`

`run_fallback` in `main.rs` already normalizes `args[0]` to basename for TOML lookup — **`rtk rewrite` / `discover/registry.rs` does not**. Hooks return no rewrite → agent runs raw command → **no token savings**.

**Proposed fix:** shared infrastructure for TOML path rewrite:
1. `lookup_command_for_filter()` in `src/core/toml_filter.rs`
2. TOML-match fallback in `rewrite_segment_inner()` in `src/discover/registry.rs`

**Scope:** 2 Rust files, ~48 lines, tests on existing `make` filter (no new `.toml` filters in this PR).

**Depends on:** nothing (can merge after or in parallel with `fix(cursor): exit code 3` — they are independent).

**Blocks:** TOML filter PRs that rely on path/wrapper invocations (`bru`, `ng test`, etc.).

---

### Two execution paths, one gap

| Path | Entry | TOML lookup | Works today? |
|------|-------|-------------|--------------|
| **Direct** | `rtk make all` | `lookup_cmd = "make all"` (basename in `run_fallback`) | ✅ |
| **Hook** | Agent: `make all` → `rtk rewrite` → `rtk make all` | `rules.rs` prefix `make` | ✅ |
| **Hook + path** | Agent: `node_modules/.bin/make all` | No `rules.rs` match for path | ❌ |
| **Hook + path** | Agent: `node_modules/.bin/bru run …` | No `rules.rs` match; TOML exists but hook never prepends `rtk` | ❌ |

### CONTRIBUTING checklist assumes `rules.rs` only

From `src/cmds/README.md` — TOML filter checklist:

1. Create filter in `src/filters/`
2. **Add rewrite pattern in `src/discover/rules.rs`**
3. Write tests

For tools invoked **only** via `.bin/` paths or with wrappers (`npx bru`, `yarn ng test`), adding a `rules.rs` entry per variant does not scale and still misses compound/path edge cases.

---

[...]

PR #2748, opened 2026-06-30
feat(hook): wire TOML filters + better tee tail hint

aeppling:
## Summary

Wire TOML-covered commands into the transparent hook rewrite path, and make TOML truncation reversible via the shared tee hints.

### Related issues & pr

**ISSUES**

#2179 
#820 
#2721 

**PRs**

#2226 
#894 
#2722
#1068 

### Part 1: Hook wiring (#2179)

The hook only consulted the static `RULES` table, so commands covered solely by a TOML filter (`jj`, `jq`, `ssh`, …) weren't rewritten. `rewrite_segment_inner` now falls back to the TOML registry and rewrites them to `rtk <cmd>` (guarded against denylisted/reserved names). Builds on #2226.

image

### Part 2: Reversibility

Truncated TOML output now carries a recovery hint: the line-offset tail hint (`[see remaining: tail -n +N …]`) for contiguous drops, the full-output hint otherwise, and never an unrecoverable marker (falls back to full raw when tee is off).

image

Then with the command hint
image

### Part 3: Trust guards

Both the project `.rtk/filters.toml` and the user-global `~/.config/rtk/filters.toml` are now trust-gated (SHA-256). `rtk init` prompts to enable detected filters, or `--trust-filters` / `--no-trust-filters` for scripts.

**Note:** the registry load now hits the hook's `Unsupported` path (~3–8 ms, one-time per process; the `Supported` path is unaffected) , a match-only `RegexSet` is a documented follow-up.

"rtk init -g" (--trust-filters or --no-trust-filters available for non interactive)
image

"rtk trust" (--yes available for non interactive)
image

"rtk trust" on global + scoped config
image

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — fmt clean, zero clippy warnings, **2310 unit + 8 integration tests pass** (21 new tests added)
- [x] Manual testing: `rtk <command>` output inspected — bare `jj log` / `stat /usr/bin/*` / `gcc --version` transparently rewritten to `rtk …` by the hook; `stat`/`du` emit `[see remaining: tail -n +N …]` and following it recovers the omitted lines byte-exact; `RTK_TEE=0` shows full raw with no truncation marker; `git status` unchanged and `cd` / reserved names / `RTK_NO_TOML=1` correctly pass through
- [x] Human testing : Ran rtk commands and created new toml filter to test security gates, all rewrite verified in "rtk gain --history" and tested the recall with the tee_tail hint command 
- [x] Human testing : Tested all UX with interactive or non interactive guards (rtk init -g, rtk trust/untrust)

### DIFF
--- a/docs/contributing/CODING_PRACTICES.md
+++ b/docs/contributing/CODING_PRACTICES.md
@@ -179,7 +179,7 @@ RTK executes shell commands on behalf of the user, so security is a first-class 
 
 **Hook integrity.** RTK verifies hook files via SHA-256 hashes before operational commands. If a hook has been tampered with, RTK exits with code 1. See [`src/hooks/integrity.rs`](../../src/hooks/integrity.rs).
 
-**Project filter trust.** `.rtk/filters.toml` files are not loaded until the user explicitly trusts them, and content changes require re-trust. See [`src/hooks/trust.rs`](../../src/hooks/trust.rs).
+**Filter trust.** Both the project-local `.rtk/filters.toml` and the user-global `~/.config/rtk/filters.toml` are gated: filters are not loaded until the user explicitly trusts them (`rtk trust`, which prints the rules first), and any content change requires re-trust (SHA-256). `rtk init` lets you enable existing filters interactively (or `--trust-filters` / `--no-trust-filters` for non-interactive use). This is cross-platform consent + tamper-evidence, not a sandbox — a same-UID attacker who can also write the trust store (`trusted_filters.json`) can bypass it. See [`src/hooks/trust.rs`](../../src/hooks/trust.rs).
 
 **Permission whitelist.** `is_operational_command()` in `main.rs` uses a whitelist pattern — new commands are *not* integrity-checked until explicitly added. This is an intentional security posture: fail-open with an audit trail is preferred over false confidence.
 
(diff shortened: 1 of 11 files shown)


aeppling:
### Notes

This is a breaking changes for users, custom filters are now gated by rtk trust to avoid security issues.
Need a way to properly inform users as release lands

For now no breaking changes triggered for sem ver but a message in rtk gain

image

pszymkowiak (approved):
Reviewed and verified locally (built from the branch): `cargo fmt` clean, `cargo clippy --all-targets` zero warnings, toml-orphan + trust tests pass.

**Part 1 (hook wiring)** — functional matrix confirms it: `jj` / `jq` / `ssh` / `mise run` / `just` / `task` rewrite to `rtk …`; absolute paths, redirects, compound (`jj diff && jq .` → `rtk jj diff && rtk jq .`), and env-prefix are preserved; `cd` / reserved / unknown commands are left alone; `RTK_NO_TOML=1` disables the TOML fallback while keeping the static RULES.

**Part 3 (trust gating)** — verified end-to-end: an untrusted `.rtk/filters.toml` does not activate rewrites until `rtk trust`, then it does.

**Part 2 (tee tail hint)** — verified: truncated output carries `[see remaining: tail -n +N <teefile>]`, following it recovers the omitted lines, and `RTK_TEE=0` shows full raw with no truncation marker.

Security is sound: `collect_match_patterns` aggregates only trusted (SHA-256) + built-in filter patterns, so a malicious project filter can't inject a rewrite; reserved / rust-handled commands are excluded from the fallback.

Non-blocking nits: the tee file drops the final trailing newline, so hint recovery is byte-exact except that last `\n` (looks pre-existing). The ~3–8 ms one-time cost on the Unsupported path justifies the documented match-only `RegexSet` follow-up.

LGTM 👍

PR #2749, opened 2026-06-30
fix(parser): use char_indices to prevent panic on CJK/emoji output

### LINKED ISSUE #2509: Panic 'not a char boundary' in output filter (src/parser/mod.rs:191) on large CJK output — affects all wrapped commands (same root as #1837)

guyoron1:
## Summary

- `extract_json_object()` panics with `'not a char boundary'` when JSON output contains multi-byte UTF-8 characters (CJK, emoji)
- Root cause: char indices from `enumerate()` were used as byte offsets when slicing the input string — char index != byte offset for multi-byte characters
- Fix: replace `Vec<char>` + `enumerate()` with `char_indices()` which yields proper byte offsets directly

Fixes #2509

## Test plan

- [x] Added regression tests for Korean, Japanese, and emoji in JSON objects
- [x] Added test for CJK prefix text before JSON (non-ASCII `start_pos`)
- [x] Added test for nested JSON with CJK values
- [x] `cargo test --all` passes (CI)

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -334,8 +334,7 @@ Scope: all 6 workspace projects
     #[test]
     fn test_extract_json_object_cjk_with_prefix() {
         let input = "경고: 무시됨\n{\"result\": \"한국어 테스트\", \"ok\": true}\n";
-        let extracted =
-            extract_json_object(input).expect("Should extract JSON after CJK prefix");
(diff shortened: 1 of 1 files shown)


guyoron1:
Superseded by #2751 which was merged with the same char_indices fix.

PR #2750, opened 2026-06-30
fix(cargo): pass through --message-format=json output without text filtering

### LINKED ISSUE #2419: rtk cargo build --message-format=json reports a failed build as "compiled" and hides the error

guyoron1:
## Summary
- `CargoBuildHandler` uses text patterns (`Compiling`, `error[`, `Finished`) that don't match NDJSON lines, so `--message-format=json` builds report "compiled" even on failure and hide all compiler errors
- Detect `--message-format=json` (and `json-diagnostic-rendered-ansi`, `json-render-diagnostics`) in args and bypass text filtering, passing raw JSON output unchanged
- Applies to both `cargo build` and `cargo check` (same handler)

Fixes #2419

## Test plan
- [x] Added 9 unit tests for `has_json_message_format` detection (inline `=`, space-separated, all JSON variants, negative cases)
- [x] `cargo test --all` passes (2244 pass, 2 pre-existing unrelated failures)
- [x] Manual verification: failing build with `--message-format=json` now shows error diagnostic JSON + `"success":false` + exit 101
- [x] Manual verification: successful build with `--message-format=json` shows artifact JSON + `"success":true` + exit 0
- [x] Non-JSON builds continue filtering normally

### DIFF
--- a/src/cmds/rust/cargo_cmd.rs
+++ b/src/cmds/rust/cargo_cmd.rs
@@ -298,7 +298,33 @@ fn run_cargo_streamed(
     )
 }
 
+fn has_json_message_format(args: &[String]) -> bool {
+    let mut iter = args.iter();
+    while let Some(arg) = iter.next() {
+        if arg == "--message-format=json"
+            || arg == "--message-format=json-diagnostic-rendered-ansi"
+            || arg == "--message-format=json-render-diagnostics"
+        {
+            return true;
+        }
+        if arg.starts_with("--message-format=") && arg.contains("json") {
+            return true;
+        }
+        if arg == "--message-format" {
+            if let Some(val) = iter.next() {
+                if val.contains("json") {
+                    return true;
+                }
+            }
+        }
+    }
+    false
+}
+
 fn run_build(args: &[String], verbose: u8) -> Result<i32> {
+    if has_json_message_format(args) {
+        return run_cargo_filtered("build", args, verbose, |s| s.to_string());
+    }
     run_cargo_streamed(
         "build",
         args,
@@ -321,6 +347,9 @@ fn run_clippy(args: &[String], verbose: u8) -> Result<i32> {
 }
 
 fn run_check(args: &[String], verbose: u8) -> Result<i32> {
+    if has_json_message_format(args) {
+        return run_cargo_filtered("check", args, verbose, |s| s.to_string());
+    }
     run_cargo_streamed(
         "check",
         args,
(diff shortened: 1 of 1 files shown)


KuSh:
Hi, thanks for your work. There’s already #2422, which is more complete since it doesn’t give up on reducing JSON output. It should definitely include your args detection too. Feel free to test it and share any feedback there

guyoron1:
Thanks for pointing out #2422 — happy to close this one in favor of the more complete approach there. Let me know if there's anything from my args detection logic worth cherry-picking into that PR.

guyoron1:
Superseded by #2422 which handles --message-format=json more completely.

PR #2751, opened 2026-06-30
fix(parser): use byte offsets instead of char indices in extract_json_object

### LINKED ISSUE #2509: Panic 'not a char boundary' in output filter (src/parser/mod.rs:191) on large CJK output — affects all wrapped commands (same root as #1837)

guyoron1:
## Summary
- `extract_json_object()` used char indices from `enumerate()` as byte offsets for string slicing
- Multi-byte UTF-8 characters (CJK, emoji) caused `'not a char boundary'` panic
- Switched from `Vec<char>` + `enumerate()` to `char_indices()` which yields correct byte offsets

Fixes #2509

## Test plan
- [x] Added regression tests for CJK values, emoji values, CJK prefix, and mixed multi-byte nested JSON
- [x] All 15 parser tests pass
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` passes (2 pre-existing failures in `rewrite_cmd` unrelated to this change)

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -167,13 +167,12 @@ pub fn extract_json_object(input: &str) -> Option<&str> {
         found_start?
     };
 
-    // Brace-balance forward from start_pos
+    // Brace-balance forward from start_pos using byte offsets (not char indices)
     let mut depth = 0;
     let mut in_string = false;
     let mut escape_next = false;
-    let chars: Vec<char> = input[start_pos..].chars().collect();
 
-    for (i, &ch) in chars.iter().enumerate() {
+    for (byte_offset, ch) in input[start_pos..].char_indices() {
         if escape_next {
             escape_next = false;
             continue;
@@ -186,8 +185,7 @@ pub fn extract_json_object(input: &str) -> Option<&str> {
             '}' if !in_string => {
                 depth -= 1;
                 if depth == 0 {
-                    // Found matching closing brace
-                    let end_pos = start_pos + i + 1; // +1 to include the `}`
+                    let end_pos = start_pos + byte_offset + ch.len_utf8();
                     return Some(&input[start_pos..end_pos]);
                 }
             }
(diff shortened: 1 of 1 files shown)


scripts/benchmark.sh
KuSh, line 1: This change seems unrelated to the PR’s subject

KuSh (changes requested):
Hi, thanks for your PR. It contains an unrelated change that should be withdrawn; otherwise, LGTM

guyoron1:
Heyaaa : ) 

This benchmark test hits httpbin.org/robots.txt live and compares raw vs filtered token counts. 
The result is non-deterministic as in the earlier CI run on this same PR (before the UV fix commits), the same test showed 8 → 8 (+0%).

I think that the difference is may be due to varying response headers/content from httpbin.org between runs? 
what do you think? 

The changes in this PR only touch registry.rs (rewrite logic) and main.rs , so no curl filter code was modified.

KuSh:
httpbin.org seems a bit erratic; it can take 20+ seconds to respond or even return a 503. That said, this PR should stay focused on a single subject.

Also, you’ve added two new commits (191c8f8 and 566f856) that don’t seem related to the PR’s topic. Please revert them, if this is a real issue that needs fixing, could you please open a separate PR with this two commits, a clear description of the problem and the fix?

guyoron1:
Oh thanks for the heads up and review ! 
I've removed the unrelated UV commits and opened [#2768 ](https://github.com/rtk-ai/rtk/pull/2768)for those separately.
This PR now contains only the parser fix (single commit 27f9739).

The clippy CI failure is a pre-existing issue on develop — uv_cmd.rs:61 calls print_with_hint with 4 args but the signature expects 5 (missing guard_raw).
PR #2768 includes the fix for that.

KuSh (approved):

guyoron1:
This also supersedes my earlier #2749 which tackled the same char_indices fix with a different set of tests. Closing #2749 as duplicate.

PR #2757, opened 2026-06-30
fix(next): drop redundant leading `build` arg so `rtk next build` exits 0 on success

### LINKED ISSUE #2756: rtk next build exits 1 on a successful build (double `build` arg → `next build build`)
## Summary

`rtk next build` exits **1** on a fully successful Next.js build. The compact summary even reports `Errors: 0 | Warnings: 0`, but the process return code is `1`, which breaks any CI/automation gate that checks the exit code (e.g. a build-verification step).

## Root cause

`rtk next` already injects the `build` subcommand, then appends *all* trailing args. The `Next` command is declared with `trailing_var_arg = true`, so the natural invocation `rtk next build` passes `["build"]` through as well:

```rust
// src/cmds/js/next_cmd.rs
cmd.arg("build");          // always injected
for arg in args {          // args == ["build"] for `rtk next build`
    cmd.arg(arg);
}
```

The child therefore runs `next build build`. Next.js interprets the second `build` as the `[directory]` positional and fails:

```
Invalid project directory provided, no such directory: <cwd>/build
```

…exiting `1`. The output never contains a route table and the error phrasing ("Invalid project directory provided") isn't matched by the `errors` heuristic in `filter_next_build`, so the summary still prints `Errors: 0` — masking the failure.

## Reproduction

In any Next.js project on Windows 11 (rtk 0.43.0):

```
$ npx next build        # exits 0 (clean build)
$ npx next build build  # exits 1: "Invalid project directory provided, no such directory: .../build"
$ rtk next build        # exits 1, prints "Next.js Build / Errors: 0 | Warnings: 0"
```

`rtk next build` is equivalent to `next build build`, hence the false failure.

## Environment

- rtk 0.43.0 (x86_64), Windows 11
- Next.js 16 project (also applies to 15.x — it's the `[directory]` positional, unchanged across versions)

## Impact

Any wrapper that gates on `rtk next build`'s exit code sees a spurious failure on a green build. Workaround today is to use `npx next build` directly for the authoritative exit code.

## Suggested fix

Drop a redundant leading `build` token before appending args (so both `rtk next` and `rtk next build` produce `next build`). PR incoming with a unit test.

kars85:
## What

`rtk next build` exits **1** on a fully successful Next.js build (the summary even prints `Errors: 0 | Warnings: 0`). Fixes #2756.

## Root cause

`rtk next` always injects the `build` subcommand, then appends every trailing arg. Because `Next` uses `trailing_var_arg`, the natural `rtk next build` invocation passes `["build"]` through as well, so the child runs **`next build build`**. Next.js parses the second `build` as the `[directory]` positional and aborts:

```
Invalid project directory provided, no such directory: <cwd>/build
```

…returning exit 1. The error phrasing isn't matched by `filter_next_build`'s `errors` heuristic, so the summary still reports `Errors: 0`, masking the failure.

## Fix

`normalize_build_args` strips a single redundant **leading** `build` token before the args are appended, so both `rtk next` and `rtk next build` run `next build`. All other flags (`--debug`, `--no-lint`, …) pass through unchanged; a non-leading `build` token (e.g. a flag value) is left intact.

## Verification

- **Root cause reproduced empirically** (Windows 11, Next.js 16):
  - `npx next build` → exit **0**
  - `npx next build build` → exit **1**: `Invalid project directory provided, no such directory: ...\build`
  - `rtk next build` (0.43.0) → exit **1**, prints `Next.js Build / Errors: 0 | Warnings: 0`
- Added unit test `test_normalize_build_args_strips_redundant_build` covering: leading `build` stripped, `build --debug` → `--debug`, no-redundant-token unchanged, empty unchanged, and a non-leading `build` preserved.

> Note: I couldn't run `cargo test` locally — this machine has the rust-msvc toolchain but no MSVC C++ linker (`link.exe` not found), so the crate's native deps don't link here. The change is a small pure function; CI will compile + run the new test. Happy to adjust if the project prefers a different normalization (e.g. erroring on an explicit `build` rather than silently stripping).

## Manual test

```
# before: exit 1 on a green build
# after:  exit 0
rtk next build && echo OK
```

### DIFF
--- a/src/cmds/js/next_cmd.rs
+++ b/src/cmds/js/next_cmd.rs
@@ -20,7 +20,12 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
 
     cmd.arg("build");
 
-    for arg in args {
+    // `rtk next` always runs `next build`, so a redundant leading `build` token
+    // (the natural `rtk next build` invocation) must be dropped — otherwise the
+    // child becomes `next build build`, which Next.js parses as a `[directory]`
+    // positional ("Invalid project directory provided, no such directory: .../build")
+    // and exits 1 on an otherwise-clean build. See normalize_build_args.
+    for arg in normalize_build_args(args) {
         cmd.arg(arg);
     }
 
@@ -36,6 +41,18 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         filter_next_build,
         runner::RunOptions::default(),
     )
+}
+
+/// `rtk next` already injects the `build` subcommand, so a leading redundant
+/// `build` token (from the natural `rtk next build` invocation) is stripped to
+/// avoid `next build build`, which Next.js rejects as an invalid `[directory]`
+/// positional and exits 1 on an otherwise-successful build. Any other args
+/// (e.g. `--debug`, `--no-lint`) pass through unchanged.
+fn normalize_build_args(args: &[String]) -> &[String] {
+    match args.first() {
+        Some(first) if first.as_str() == "build" => &args[1..],
+        _ => args,
+    }
 }
 
 /// Filter Next.js build output - extract routes, bundles, warnings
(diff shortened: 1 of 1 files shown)


KuSh (changes requested):
Hi, thanks for your contribution. As mentioned in the issue, the right fix is to rewrite the `claude-md` template so it tells Claude to use `rtk rewrite` for each individual command, rather than listing only some of the available commands. `rtk` is meant for LLMs, not for humans, so nobody should be typing `rtk next build` by hand.

PR #2759, opened 2026-06-30
fix: separate Copilot init command from Claude Code in Quick Start

### LINKED ISSUE #2244: The copilot initialisation is misleading in README file quick start section
In the quick start section it says:
`rtk init -g                     # Claude Code / Copilot (default) `

But when running the command it only works for claude. So the the fix  should be:
`rtk init -g                     # Claude Code `
`rtk init -g --copilot         # Copilot `

highoncomputers:
The Quick Start section listed `rtk init -g` as "Claude Code / Copilot (default)", but this command only works for Claude Code. Copilot requires the `--copilot` flag.

**Before:**
```
rtk init -g                     # Claude Code / Copilot (default)
```

**After:**
```
rtk init -g                     # Claude Code (default)
rtk init -g --copilot           # GitHub Copilot
```

Fixes #2244

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,8 @@ rtk gain        # Should show token savings stats
 
 ```bash
 # 1. Install for your AI tool
-rtk init -g                     # Claude Code / Copilot (default)
+rtk init -g                     # Claude Code (default)
+rtk init -g --copilot           # GitHub Copilot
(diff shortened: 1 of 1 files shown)


highoncomputers:
Changed base branch from master to develop. The label should update automatically.

PR #2761, opened 2026-06-30
docs: recommend ripgrep for better grep performance

### LINKED ISSUE #2613: docs: recommend to install ripgrep?

highoncomputers:
This PR:
1. Adds a tip about installing ripgrep for faster searches
2. Documents the `rtk rg` native ripgrep command in the Files section

Closes #2613


highoncomputers:
Changed base branch from master to develop. The label should update automatically.

PR #2768, opened 2026-07-01
fix(uv): remove uv run from transparent prefixes to fix rewrite conflicts

guyoron1:
## Summary

`uv run` was listed in both `BUILTIN_TRANSPARENT_PREFIXES` and `RULES`, causing a routing conflict:

- The transparent prefix check would strip `uv run` and try to match the inner command
- When the inner command (e.g. `python script.py`) wasn't RTK-supported, it returned `None` — no rewrite at all
- Commands like `uv run --unknown pytest` or `uv run -m pytest` were silently dropped

**Fix:** Remove `uv run` from `BUILTIN_TRANSPARENT_PREFIXES` so all `uv run` commands route through the RULES-based `rtk uv` path, which delegates to `uv_cmd.rs` for proper filtering.

Also fixes a compilation issue where `print_with_hint` signature changed (added `guard_raw` arg) but `uv_cmd.rs` wasn't updated.

## Changes

- Remove `uv run` from `BUILTIN_TRANSPARENT_PREFIXES` in `registry.rs`
- Pass `guard_raw` arg to `print_with_hint` in `uv_cmd.rs` (signature fix)
- Update tests to reflect new routing: `uv run pytest` → `rtk uv run pytest` (not `uv run rtk pytest`)
- Apply rustfmt to `BUILTIN_TRANSPARENT_PREFIXES` array

## Test plan

- [x] Updated existing `uv run` rewrite tests to match new routing behavior
- [x] `uv run --unknown pytest` and `uv run -m pytest` now rewrite correctly instead of returning `None`
- [x] `cargo check` passes

### DIFF
--- a/src/cmds/python/uv_cmd.rs
+++ b/src/cmds/python/uv_cmd.rs
@@ -58,7 +58,7 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         .context("Failed to run uv")?;
     let filtered = filter_uv_run_output(&result.raw, result.exit_code);
 
-    runner::print_with_hint(&filtered, &result.raw, "uv", result.exit_code);
+    runner::print_with_hint(&filtered, &result.raw, &result.raw, "uv", result.exit_code);
     timer.track(&original_cmd, &rtk_cmd, &result.raw, &filtered);
 
     Ok(result.exit_code)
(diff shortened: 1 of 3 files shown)


aeppling:
Hey @guyoron1 , thanks for this quick correction, LGTM

aeppling (approved):

PR #2770, opened 2026-07-01
fix(rewrite): strip GNU timeout prefix before rewriting inner command…

### LINKED ISSUE #2726: `rtk rewrite` ignores commands wrapped in a `timeout` prefix
## Context

The Claude Code hook delegates rewriting to `rtk rewrite <cmd>`. Any command prefixed with GNU `timeout` is returned as passthrough (exit 1), so it never gets compressed.

  ## Repro (rtk 0.43.0)

  ```console
  $ rtk rewrite "timeout 300 .venv/bin/python -m pytest"   # exit 1, no rewrite
  $ rtk rewrite "timeout 120 npx eslint src/"              # exit 1, no rewrite
  $ rtk rewrite "grep -n foo src"                          # rtk grep -n foo src  ✓
  ```

`timeout` is extremely common for bounded test/lint runs (`timeout N pytest|vue-tsc|eslint|type-check`), so this silently leaks a large share of savings. In one 30-day window these accounted for several thousand un-routed commands.

  ## Request
  
Teach `rtk rewrite` to recognize a leading `timeout [-k DUR] [-s SIG] [--flags] DURATION` prefix: strip it, rewrite the inner command via the existing registry, and re-prepend the prefix verbatim. This mirrors the absolute-path / `\cmd` normalization already documented in the hook.

  ## Workaround

Currently handled in a local hook (`local-rewrite.sh`) that peels the `timeout` prefix before matching, but native support would let downstreams drop the hack.

For reference, here's the local-hook workaround we're running (shell, runs before delegating to `rtk rewrite`). It peels the `timeout` prefix, lets the inner command be rewritten, then re-prepends the prefix verbatim.

  ```bash
  # ── Unwrap `timeout [opts] DURATION` prefix so the inner command can be
  # rewritten, then re-prepend the prefix verbatim. GNU timeout syntax:
  #   timeout [-k DUR] [-s SIG] [--preserve-status] [--foreground] DURATION CMD…
  # Only leading tokens are stripped, so the inner command stays byte-identical.
  TIMEOUT_PREFIX=""
  if [[ "$CMD" == "timeout "* ]]; then
    work="${CMD#timeout}"
    work="${work#"${work%%[![:space:]]*}"}"   # drop leading whitespace
    while :; do
      tok="${work%%[[:space:]]*}"
      case "$tok" in
        -k|--kill-after|-s|--signal)          # option taking a separate value
          work="${work#"$tok"}"; work="${work#"${work%%[![:space:]]*}"}"
          val="${work%%[[:space:]]*}"; work="${work#"$val"}"; work="${work#"${work%%[![:space:]]*}"}" ;;
        -*)                                    # flag or --opt=value form
          work="${work#"$tok"}"; work="${work#"${work%%[![:space:]]*}"}" ;;
        *) break ;;
      esac
    done
    dur="${work%%[[:space:]]*}"
    if [[ "$dur" =~ ^[0-9]+(\.[0-9]+)?[smhd]?$ ]]; then
      work="${work#"$dur"}"; work="${work#"${work%%[![:space:]]*}"}"
      if [ -n "$work" ]; then
        TIMEOUT_PREFIX="${CMD%"$work"}"        # prefix before the inner cmd, verbatim
        CMD="$work"                            # CMD is now the inner command
      fi
    fi
  fi

  # … run CMD (the inner command) through the rewrite registry …
  REWRITTEN=$(rtk rewrite "$CMD")

  # Re-prepend the timeout prefix (empty string when not timeout-wrapped).
  REWRITTEN="$TIMEOUT_PREFIX$REWRITTEN"
  ```

  Verified against: `timeout 300 …`, `timeout 5m …`, `timeout -s KILL 300 …`, quoted args (`timeout 90 grep -n 'foo bar baz' src` → preserved verbatim), and  no-op cases (`timeout 300` with no command, non-timeout commands). The `TIMEOUT_PREFIX="${CMD%"$work"}"` trick captures the original prefix verbatim because `work` is always a suffix of the original string.

  Native handling in `rtk rewrite` would be cleaner since the duration/option parsing belongs next to the existing path-normalization logic.

groot-guo:
… (#2726)

`rtk rewrite` returned passthrough for any command wrapped in GNU `timeout`, silently skipping compression for bounded test/lint runs (e.g. `timeout 300 pytest`, `timeout 120 npx eslint src/`).

Add `strip_timeout_prefix`, mirroring the existing `strip_disabled_prefix` / `strip_word_prefix` strip-recurse-reprepend contract: peel `timeout [-k DUR] [-s SIG] [--flags] DURATION`, rewrite the inner command via the existing registry, and re-prepend the prefix verbatim. The inner command slice stays byte-identical, so quoted args survive untouched.

- word-boundary match on "timeout" (`timeoutx` not confused)
- validates duration token `N[.N][s|m|h|d]`
- passthrough when no inner command follows the duration
- byte-offset token walk keeps all slices on char boundaries
- 13 unit tests covering flag/duration/compound/edge cases

Closes #2726

## Summary

-

## Test plan

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1351,6 +1351,20 @@ fn rewrite_segment_inner(
         return Some(format!("{}{}", env_prefix, rewritten));
     }
 
+    // #2726: GNU `timeout` wrapper — strip the `timeout [opts] DURATION` prefix,
+    // rewrite the inner command, then re-prepend the prefix verbatim.
+    if let Some((timeout_prefix, inner)) = strip_timeout_prefix(trimmed) {
+        // A timeout wrapper must not make a TOML-only command routable. The
+        // TOML path normalizes absolute command paths, while wrappers preserve
+        // argv[0] verbatim; routing it here would produce an invalid command.
+        if matches!(classify_command(inner), Classification::Unsupported { .. }) {
+            return None;
+        }
+        let rewritten =
+            rewrite_segment_inner(inner, excluded, transparent_prefixes, context, depth + 1)?;
+        return Some(format!("{}{}", timeout_prefix, rewritten));
+    }
+
     for (prefix, routable) in builtin_transparent_prefixes() {
         if let Some(rest) = strip_word_prefix(trimmed, prefix) {
             if rest.is_empty() {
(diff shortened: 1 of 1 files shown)


groot-guo:
Hi maintainers — I’ve rebased this PR onto the latest `develop` and resolved the registry conflict.

The change remains scoped to GNU `timeout` prefix rewriting. The focused timeout test suite passes (`cargo test timeout_prefix`, 13 tests).

I also observed 7 failures in the full local `cargo test` run under `curl_cmd` and `ctest_cmd`, outside this PR’s diff. I have not classified those as baseline issues yet, so please let me know if you’d like me to investigate them further.

This PR is ready for review. Thanks!

PR #2775, opened 2026-07-01
chore(scripts): switch benches to mockhttp.org as httpbin.org is struggling at the moment

KuSh:
## Summary

httpbin.org sometimes returns 503s or takes 20+ seconds to respond, which causes the benchmarks to fail

## Test plan
- [x] Manual testing: `./benchmark.sh` output inspected

```sh
  ✅ 22 good  ⚠️ 37 warn  🔴 3 negative  ❌ 0 fail    22/62 (35%)
  Tokens: 100941 → 8033  (-92%)
```


aeppling:
nice this mock page look like correct solution for the often failing curl on current url, lgtm

aeppling (approved):

PR #2776, opened 2026-07-01
feat: bug: hook rewrite of `pnpm lint` discards package.json script

### LINKED ISSUE #2094: bug: hook rewrite of `pnpm lint` discards package.json script body — same false-green as #185 at the hook layer

wilmerdooley:
Resolves #2094.

The PreToolUse hook rewrites `pnpm lint` → `rtk lint`, discarding the body of the project's `package.json` `lint` script. For projects whose script enforces flags `eslint` doesn't default to (e.g. `--max-warnings 0`) or chains commands (e.g. `pnpm check:foo && eslint ...`), the rewritten `rtk lint` invokes `eslint` with defaults and exits 0 even when the original `pnpm lint` would have exited 1.

## Summary

-

## Test plan


aeppling:
Hey @wilmerdooley 

### This is related to issues #2663 and #2094

Two related PR that both address a temporary fix : 
- #2664 narrows/extends NPM_SUBCOMMANDS, an allowlist of when injection is safe.
- #2776 narrows the lint regex + rewrite_prefixes, an allowlist of when substitution is safe (keep eslint/npx eslint/pnpm dlx eslint, drop pnpm lint/npm run lint).

The issue is at a higher level, and maintaining those hard coded lists is not a correct solution.

### Proposed fix : 

- npm: forward the args untouched. npm build then fails exactly as it would without RTK, that's the agent's own command, and it self-corrects to npm run build. npm query, npm add, npm sbom, and anything npm adds in future just work. NPM_SUBCOMMANDS gets deleted.
- lint: run the real pnpm lint (the actual script, with its real flags, chain, and exit code) and route its output through the lint filter, don't rebuild an eslint invocation. The invocation allowlist gets deleted.

This fixes both issues at the root, removes both lists (no more chasing npm versions or package-manager forms), and keeps every exit code faithful.

#2776 currently drops pnpm lint to raw with no filter (0% savings), running it verbatim and filtering the output preserves the gain.

PR #2789, opened 2026-07-02
fix(analytics): truncate display strings on UTF-8 char boundaries

### LINKED ISSUE #2787: rtk gain --history panics on non-ASCII (Cyrillic) text — byte index not on char boundary
## Description
`rtk gain --history` crashes with a Rust panic when the analytics history
contains non-ASCII (e.g. Cyrillic) characters in a tracked command/path.
The panic happens at a fixed-width string truncation that slices by byte
offset instead of a UTF-8 char boundary.

## Environment
- rtk version: 0.43.0
- OS: Windows 11 (10.0.26200), locale ru-RU
- Shell: PowerShell 7.6.3
- Hook: `rtk hook claude` (Claude Code native Windows hook)

## Steps to reproduce
1. Have rtk track a command whose path contains Cyrillic characters,
   e.g. `rtk ls -la "D:\12 Claude\06 Ародинамический расчёт"`
2. Run `rtk gain --history`

## Actual behavior
```
thread 'main' (37432) panicked at src\analytics\gain.rs:250:54:
end byte index 22 is not a char boundary; it is inside 'Н' (bytes 21..23 of string)
```
Exit code 9.

## Expected behavior
The history table should truncate long command/path strings on a UTF-8
char boundary (e.g. via `.chars()` or `unicode-segmentation`) instead of
raw byte slicing, so it doesn't panic on non-ASCII paths.

## Additional notes
`rtk gain` (without `--history`) works fine — only the per-command
truncated display in `--history` crashes. Likely affects any locale/path
with multi-byte UTF-8 characters (Cyrillic, CJK, etc.), not just Russian.

aeppling:
Fixes #2787

## Problem

`rtk gain --history` panicked when a tracked command contained non-ASCII text (Cyrillic, CJK, emoji):

```
end byte index 22 is not a char boundary; it is inside 'Н' (bytes 21..23 of string)
```

## Root cause

Display code in `src/analytics/` truncated strings with raw byte slicing (`s.len() > N` then `&s[..M]`), which panics when the cut lands inside a multi-byte UTF-8 character. These were hand-inlined byte-based copies of the char-safe `core::utils::truncate()` helper that the rest of the codebase already uses. This is the 5th recurrence of this defect class (previous fixes: da486bf tee, c5ec92f git, 533894a json, #2751 parser); an audit of all string-slice sites in `src/` confirmed the remaining unsafe ones were all in `src/analytics/`.

Every fixed site was first reproduced as a real panic against the pre-fix binary with crafted data:

| Site | Panic reproduced |
|---|---|
| `gain.rs` history command | `byte index 22 is not a char boundary` |
| `gain.rs` failures top commands | `byte index 47 is not a char boundary` |
| `gain.rs` failures recent commands | `byte index 37 is not a char boundary` |
| `gain.rs` failures timestamp prefix | `byte index 16 is not a char boundary` |
| `session_cmd.rs` session id prefix | `byte index 8 is not a char boundary` |

## Fix

- Replace the three ad-hoc truncations with the shared `core::utils::truncate()` (chars-based, existing helper, no new dependency).
- Use `.get(..n).unwrap_or(full)` for the timestamp and session id prefix slices: valid prefix when the boundary is clean, full string instead of a panic otherwise.
- Add a Cyrillic regression test with the exact string from the issue, next to the existing Thai/CJK/emoji truncate tests.

## Verification

- The issue repro (`rtk ls -la "D:\...Ародинамический..."` tracked, then `rtk gain --history`) now renders correctly with exit 0.
- All 5 sites verified fixed end-to-end against the same crafted data (isolated `RTK_DB_PATH` databases and a sandboxed `CLAUDE_CONFIG_DIR`).
- No regression: old vs new binary output on pure-ASCII data is byte-identical for `gain --history`, `gain --failures` and `session`, with every truncation branch exercised (ellipsis present in compared output).
- `cargo fmt --all --check`, `cargo clippy --all-targets` (zero warnings), `cargo test --all` (2371 passed).

## Checklist

- [x] Unit tests added (regression test in `src/core/utils.rs`)
- [x] Token savings: N/A (display fix, no filter change)
- [x] Truncated list recovery hints: N/A (`CAP_*` applies to filter output lists; the 10-row display limits here are pre-existing SQL `LIMIT 10`, unchanged)
- [x] Edge cases covered (Cyrillic, CJK, emoji, mid-char prefix boundaries)
- [x] Pre-commit gate passes
- [x] Manual test: `rtk gain --history`, `rtk gain --failures`, `rtk session` inspected

Follow-up candidate (out of scope here): enable `clippy::string_slice` repo-wide to prevent this defect class permanently; it requires `#[allow]` annotations on the ~20 verified-safe slice sites.

### DIFF
--- a/src/analytics/session_cmd.rs
+++ b/src/analytics/session_cmd.rs
@@ -110,7 +110,7 @@ pub fn run(_verbose: u8) -> Result<()> {
             .file_stem()
             .and_then(|s| s.to_str())
             .unwrap_or("unknown");
-        let short_id = id.get(..8).unwrap_or(id);
+        let short_id = &id[..id.floor_char_boundary(8)];
 
         // Extract date from mtime
         let date = fs::metadata(path)
--- a/src/analytics/session_cmd.rs
+++ b/src/analytics/session_cmd.rs
@@ -110,7 +110,7 @@ pub fn run(_verbose: u8) -> Result<()> {
             .file_stem()
             .and_then(|s| s.to_str())
             .unwrap_or("unknown");
-        let short_id = if id.len() > 8 { &id[..8] } else { id };
+        let short_id = id.get(..8).unwrap_or(id);
 
         // Extract date from mtime
         let date = fs::metadata(path)
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -713,8 +713,9 @@ fn show_failures(tracker: &Tracker) -> Result<()> {
         println!("{}", styled("Recent Failures (last 10)", true));
         println!("{}", "─".repeat(60));
         for rec in &summary.recent {
-            // ISSUE #2787: show the full string when byte 16 is not a char boundary
-            let ts_short = rec.timestamp.get(..16).unwrap_or(&rec.timestamp);
+            // ISSUE #2787: floor to the previous char boundary so the prefix
+            // never exceeds 16 bytes and never lands mid-character
+            let ts_short = &rec.timestamp[..rec.timestamp.floor_char_boundary(16)];
             let status = if rec.fallback_succeeded { "ok" } else { "FAIL" };
             let cmd_display = truncate(&rec.raw_command, 40);
             println!("  {} [{}] {}", ts_short, status, cmd_display);
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -2,7 +2,7 @@
 
 use crate::core::display_helpers::{format_duration, print_period_table};
 use crate::core::tracking::{DayStats, MonthStats, Tracker, WeekStats};
-use crate::core::utils::format_tokens;
+use crate::core::utils::{format_tokens, truncate};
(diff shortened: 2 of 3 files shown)


src/analytics/session_cmd.rs
```diff
@@ -110,7 +110,7 @@ pub fn run(_verbose: u8) -> Result<()> {
             .file_stem()
             .and_then(|s| s.to_str())
             .unwrap_or("unknown");
-        let short_id = if id.len() > 8 { &id[..8] } else { id };
+        let short_id = id.get(..8).unwrap_or(id);
```
KuSh: Not totally equivalent. If id contains an UTF-8 sequence around 8 indice it coule output a full id (which could be enormous), should use truncate or something like `let indice = id.floor_char_boundary(8); let short_id = id.get(..indice).unwrap_or(id);`

KuSh (approved):

PR #2796, opened 2026-07-03
fix(gh): classify pr checks by status column, not name substring

john-chen-opus:
## Problem

`format_pr_checks` (the `rtk gh pr checks` summarizer) classifies each row with a whole-line substring test:

```rust
if line.contains("[ok]") || line.contains("pass") { passed += 1 }
else if line.contains("[x]") || line.contains("fail") { failed += 1; … }
else if line.contains('*') || line.contains("pending") { pending += 1 }
```

`gh pr checks` prints tab-separated rows — `<name>\t<bucket>\t<elapsed>\t<url>` — so matching the whole line has two correctness bugs:

1. **A check is classified by its NAME, not its status.** A check named `compass-audit` or `fail-fast smoke` contains `pass`/`fail` as a substring and gets miscounted regardless of its real bucket.
2. **`skipping` and `cancel(ed)` checks are dropped entirely.** They match none of the three branches, so they vanish from the summary — the printed counts silently disagree with the real PR. (On a PR whose only non-passing checks are change-gated skips, `gh` still exits 0, so this summary *is* what the agent sees.)

This bit me while shipping an unrelated change: the summary disagreed with the actual check set, and the empty-checks case surfaced as a confusing "0 passed" rather than gh's real message.

## Fix

Classify on the **tab-separated status column**, not a whole-line substring:

- Match the bucket via `starts_with` (absorbs `skipping`/`skipped`, `cancel`/`canceled`/`cancelled`).
- Fold `skipped` and `cancelled` into the totals (cancelled surfaces with failures — `gh`'s own rollup exits non-zero for it).
- Lead with `passed/total` so the counts **reconcile** — skipped/pending checks are part of the total, never silently dropped.
- When no row parses (e.g. `no checks reported on the '<branch>' branch`), return empty so the caller falls back to `gh`'s raw output instead of inventing a misleading `0 passed` summary.

This aligns with the design philosophy in CONTRIBUTING.md:
- **Correctness vs Token Savings** — "when in doubt, preserve correctness";
- **Transparency** — the filtered output stays a faithful subset (counts match the real command);
- **Never Block** — the unparseable case falls back to raw rather than fabricating a summary.

## Tests

Added three unit tests in `gh_cmd.rs`:
- `test_format_pr_checks_classifies_by_bucket_not_name` — names containing `pass`/`fail` are classified by their status column;
- `test_format_pr_checks_reconciles_total_with_skips` — `1 pass + 2 skips` reports `1/3` with skips surfaced;
- `test_format_pr_checks_empty_falls_back_to_raw` — empty / "no checks reported" → empty (raw fallback).

Local: `cargo fmt --check` clean, `cargo clippy --bin rtk` clean (only a pre-existing rusqlite future-incompat note), `cargo test --bin rtk cmds::git::gh_cmd` → 69 passed.

### DIFF
--- a/src/cmds/git/gh_cmd.rs
+++ b/src/cmds/git/gh_cmd.rs
@@ -469,29 +469,71 @@ fn pr_checks(args: &[String], _verbose: u8, _ultra_compact: bool) -> Result<i32>
     )
 }
 
+/// Extract the status bucket from a `gh pr checks` row.
+///
+/// `gh pr checks` prints tab-separated rows — `<name>\t<bucket>\t<elapsed>\t<url>`
+/// — where bucket is one of pass / fail / pending / skipping / cancel(ed). We
+/// classify on the bucket COLUMN, not a whole-line substring: a check whose
+/// *name* contains "pass"/"fail"/"pending" (e.g. "compass-audit", "fail-fast
+/// smoke") must be counted by its real status, not its name.
+fn pr_check_bucket(line: &str) -> Option<(&str, &str)> {
+    let mut fields = line.split('\t');
+    let name = fields.next()?.trim();
+    let bucket = fields.next()?.trim();
+    if name.is_empty() || bucket.is_empty() {
+        return None;
+    }
+    Some((name, bucket))
+}
+
 fn format_pr_checks(stdout: &str) -> String {
     let mut passed = 0;
     let mut failed = 0;
     let mut pending = 0;
+    let mut skipped = 0;
     let mut failed_checks = Vec::new();
 
     for line in stdout.lines() {
-        if line.contains("[ok]") || line.contains("pass") {
+        let Some((name, bucket)) = pr_check_bucket(line) else {
+            continue;
+        };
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into this. The substring classification is gone from develop: #3542 (merged 2026-09-13) rewrote format_pr_checks around parse_pr_check_line, which reads the tab separated status column rather than the whole line and also collapses the repeated tables that `gh pr checks --watch` prints, and commit d1052271 the same day added a fourth bucket so skipped and cancelled checks are counted and reported instead of disappearing. That covers both problems you described, including a check named compass-audit or fail-fast smoke being counted by its name. One case develop still misses is the last part of your change: when no row parses, for example gh's "no checks reported on the branch" message, the summary still prints 0 passed and 0 failed rather than falling back to gh's own output, and a narrow PR or an issue against the current parse_pr_check_line code would be very welcome. If I missed something this PR still adds, please comment here or reopen.

PR #2800, opened 2026-07-03
feat(cloud): add rtk web for HTML-to-Markdown fetches (#1426)

### LINKED ISSUE #1426: Feature Proposal: RTK Web Hook for HTML-to-Markdown compression on web fetch commands
# Feature Proposal: RTK Web Hook for HTML-to-Markdown Compression

Hi RTK team,

I’d like to propose a feature that extends RTK’s core idea beyond terminal-heavy developer commands into web content retrieval.

RTK already provides strong value by filtering and compressing command output before it reaches the LLM context window. A natural next step would be to apply the same principle to web pages fetched by agents via commands like `curl`, `fetch`, or similar tooling.

## Problem

When an AI agent fetches a web page directly, it often receives a very large HTML document full of noise:

- `<script>` tags
- `<style>` blocks
- navigation and boilerplate
- long URLs
- repeated layout markup
- irrelevant DOM sections

This consumes a large number of tokens while adding very little useful meaning for the model.

In practice, the agent usually needs only the readable page content, not the full raw HTML.

## Proposed Feature

Add an **RTK Web Hook** that can intercept web-page fetches and transform HTML into a compact, LLM-friendly representation before it reaches the model.

### High-level pipeline

1. Detect that the response is HTML or webpage-like content.
2. Strip obvious non-content elements such as:
   - `script`
   - `style`
   - `iframe`
3. Extract the main readable content where possible:
   - `main`
   - `article`
   - other high-signal content containers
4. Convert the remaining content into clean Markdown.
5. Apply RTK-style compression:
   - collapse excessive whitespace
   - normalize repeated blank lines
   - shorten repeated long links into references like `[1]`, `[2]`
   - preserve headings, lists, code blocks, and links

## Example

Instead of:

```bash
curl https://example.com
````

returning a large raw HTML payload to the agent, the effective result could become something closer to:

```md
# Article Title

Short readable content here.

## Section
Relevant body text here.

[1] https://example.com/some/very/long/url
```

So a page that might cost thousands of tokens as raw HTML could become a few hundred tokens of useful content.

## Why This Fits RTK

This feature follows the same general value proposition RTK already provides for terminal output:

* reduce token usage
* improve signal-to-noise ratio
* preserve useful structure
* make agent context more efficient

Web pages are another major source of token waste, especially for agent workflows that browse docs, blog posts, issues, reference pages, and articles.

## Important Constraint

This should **not** rewrite every `curl` response blindly.

The feature should only activate when the fetched content is clearly HTML / webpage content, or when the user explicitly opts in.

It should **not** modify:

* JSON API responses
* XML
* binary content
* plain text outputs that are already clean

## Suggested v1 Scope

A safe first version could be intentionally narrow:

* only process responses that look like `text/html`
* do not touch JSON, XML, binary, or plain API responses
* preserve raw output on parse failure
* allow explicit opt-in or opt-out via config
* possibly start as an explicit command path instead of transparent rewriting

### Option A — explicit command

```bash
rtk web https://example.com
```

**Pros**

* safest rollout
* easy to test and benchmark
* no risk of breaking existing `curl` workflows

**Cons**

* less transparent than the normal RTK experience

### Option B — selective hook rewrite

Intercept `curl` or similar commands only when the target appears to be a webpage and the response type is HTML.

**Pros**

* very convenient
* feels native to RTK’s hook model

**Cons**

* harder to get correctness right
* needs careful heuristics and fallback behavior

### Option C — opt-in config

```toml
[web]
enabled = true
mode = "html_only"
```

**Pros**

* controlled rollout
* easy for advanced users
* avoids surprising existing users

My preference would be **Option A or C first**, then consider automatic rewriting later.

## Suggested Implementation Sketch

A possible internal pipeline for v1:

1. Execute the fetch command normally.
2. Inspect response headers and/or content sniffing.
3. If non-HTML: return original output untouched.
4. If HTML:

   * parse document
   * remove low-value nodes
   * extract main content candidates
   * convert to Markdown
   * compress whitespace and links
5. If anything fails:

   * return original output unchanged

## Token-Saving Opportunities

Possible compression steps:

* DOM pruning before conversion
* main-content extraction before full Markdown generation
* URL deduplication and numeric link references
* collapse repeated whitespace
* remove empty sections and decorative fragments

## Non-goals for v1

To keep the first version reviewable, I would explicitly avoid:

* JavaScript rendering / headless browser execution
* full reader-mode heuristics for every site
* handling authentication-heavy sites
* transforming API JSON responses
* automatic summarization by an LLM

## Why I Think This Is Worth Building

Web pages are a strong source of context waste in agent workflows.

A lightweight HTML-to-Markdown compression path would let agents read more pages within the same context window, reduce cost, and improve the quality of the text that reaches the model.

This feels like a natural extension of RTK’s existing philosophy: remove junk, keep signal, and minimize token waste.

## Willing to Implement

I’d be happy to implement this myself if the direction makes sense to maintainers.

My suggested rollout would be:

1. align on the interface
2. implement a narrow HTML-only v1
3. add tests and benchmarks
4. document behavior and configuration
5. discuss transparent hook-based rewriting later

If this sounds aligned, I can open a draft PR or start with a design sketch for the v1 interface and pipeline.

sBucka:
## Summary

- Adds `rtk web <url>` (#1426): fetches via `curl`, and for `text/html` responses strips `script`/`style`/`nav`/`header`/`footer`/`aside`/`form` noise, extracts the main/article content, and converts it to Markdown with de-duplicated numbered link references (`[1]`, `[2]`, ...) instead of repeated inline URLs
- Implemented as **Option A** from the issue — an explicit opt-in command, not a transparent `curl` rewrite. Non-HTML responses (JSON, XML, binary) pass through byte-for-byte, so existing `rtk curl` behavior for APIs is completely untouched. `web` is intentionally left out of the auto-rewrite patterns in `src/discover/rules.rs` for the same reason, matching the issue's own "Non-goals for v1"
- **Legacy-charset pages decode correctly**: pages in windows-1250/1252, iso-8859-2, shift_jis etc. are detected (header `charset=` → meta prescan → UTF-8 validation → windows-1252 fallback, the Encoding Standard's order) and decoded via `encoding_rs` before conversion — without this they'd be misread as binary and passed through raw with zero savings
- **Page chrome is pruned at attribute level too**: ARIA landmark roles (`navigation`, `banner`, `contentinfo`, ...) and an id/class token blocklist (`nav`, `menu`, `dropdown`, `sidebar`, `cookie`, `footer`, ...) catch div-built navigation that tag-level skipping misses (token-boundary matching: `mw-footer` pruned, `navy-blue` survives)
- New deps: `scraper` + `ego-tree` (both ISC, `rust-scraper` org) and `encoding_rs` (Firefox's WHATWG Encoding implementation, Apache/MIT+BSD, dep tree = `cfg-if` only). html5ever-backed parsing degrades gracefully on malformed HTML instead of erroring, which matters here since this crate builds with `panic = "abort"`

Thanks @talvaknin744 for the detailed proposal — this follows your suggested v1 scope closely (explicit command, HTML-only, raw output on parse failure, non-goals respected: no JS rendering, no full reader-mode heuristics, no LLM summarization).

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — clean, 2391 tests passing (0 regressions)
- [x] `bash scripts/check-test-presence.sh` passes; the repo's dangerous-patterns grep (`Command::new("sh")`, `unwrap()`, `panic!`, `unsafe`, etc.) comes back clean on this diff
- [x] Manual testing against **live** URLs; token numbers recorded by rtk's own tracking (`rtk gain`), not hand-estimated:

  | URL | Result |
  |---|---|
  | `en.wikipedia.org/wiki/Rust_(programming_language)` | 247,523 → 42,311 tokens (**82.9% saved**); after chrome-pruning the 53-entry language-nav list is gone and output shrank a further ~13% (169K → 148K chars) with article body intact |
  | local server, `Content-Type: text/html; charset=windows-1250` | body confirmed non-UTF-8; decodes to correct Czech diacritics and converts to Markdown (previously: raw passthrough, zero savings) |
  | `example.com` | 140 → 44 tokens (68.6% saved) |
  | `jsonplaceholder.typicode.com/todos/1` (control — must NOT be transformed) | passthrough byte-identical, confirmed |

- [x] Deterministic fixture test (`web_cmd_fixture.html`: nav/header/footer/script/style noise around a real article) pins a **>=60% savings floor** in CI without requiring network access — currently passing at 72.2%
- [x] Signal-killed curl propagates `128+signal` via the shared `exit_code_from_output()` helper (same bug class as #2681), and output goes through the `never_worse` guard (#2551/#2554) so `rtk web` can never emit more tokens than plain `curl`
- [x] Tables render as pipe-separated rows (adjacent `td` text no longer fuses into one word) and `ol` lists number `1.` `2.` with nested-list indentation

Fixes #1426

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -400,6 +400,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b04dc5a38e4f151a79d9f2451ae6037fb6eaf5cba34771f44781f80e508498e3"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "env_home"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1102,6 +1111,7 @@ dependencies = [
  "colored",
  "dirs",
  "ego-tree",
+ "encoding_rs",
  "flate2",
  "getrandom 0.4.2",
  "ignore",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -293,6 +293,50 @@ dependencies = [
 ]
 
 [[package]]
+name = "cssparser"
+version = "0.37.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8c9cdaae01d5ed7882b04d795e7f752f46ff52d2fa3b50a20d28c464510bba98"
+dependencies = [
+ "cssparser-macros",
+ "dtoa-short",
+ "itoa",
+ "phf",
+ "smallvec",
+]
+
+[[package]]
+name = "cssparser-macros"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "10a2a99df6e410a8ff4245aa2006499ea662245f967cc7c0a38c83ef8eb44dbf"
+dependencies = [
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "derive_more"
+version = "2.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d751e9e49156b02b44f9c1815bcb94b984cdcc4396ecc32521c739452808b134"
+dependencies = [
+ "derive_more-impl",
+]
+
+[[package]]
+name = "derive_more-impl"
+version = "2.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "799a97264921d8623a957f6c3b9011f3b5492f557bbb7a5a19b7fa6d06ba8dcb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "rustc_version",
+ "syn",
+]
+
+[[package]]
 name = "digest"
 version = "0.10.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -333,6 +377,27 @@ dependencies = [
  "quote",
  "syn",
 ]
+
+[[package]]
+name = "dtoa"
+version = "1.0.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4c3cf4824e2d5f025c7b531afcb2325364084a16806f6d47fbc1f5fbd9960590"
+
+[[package]]
+name = "dtoa-short"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd1511a7b6a56299bd043a9c167a6d2bfb37bf84a6dfceaba651168adfb43c87"
+dependencies = [
+ "dtoa",
+]
+
+[[package]]
+name = "ego-tree"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b04dc5a38e4f151a79d9f2451ae6037fb6eaf5cba34771f44781f80e508498e3"
 
 [[package]]
 name = "env_home"
@@ -416,6 +481,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "getopts"
+version = "0.2.24"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfe4fbac503b8d1f88e6676011885f34b7174f46e59956bba534ba83abded4df"
+dependencies = [
+ "unicode-width",
+]
+
+[[package]]
 name = "getrandom"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


sBucka:
### Additional validation: real-tokenizer measurement on a live non-English article

The savings figures in the PR body come from rtk's built-in tracker, which estimates tokens as `chars / 4`. To sanity-check that against reality I re-measured a live Czech Wikipedia article — [`Rust (programovací jazyk)`](https://cs.wikipedia.org/wiki/Rust_(programovac%C3%AD_jazyk)) — with an actual BPE tokenizer (`cl100k_base`):

| | Tokens (real BPE) | rtk tracker (`chars/4`) |
|---|---|---|
| raw HTML (plain `curl`) | 125,266 | ~88,885 |
| `rtk web` | 22,853 | ~14,352 |
| **saved** | **102,413 (81.8%)** | ~74,533 (~83.5%) |

Two takeaways:

- The **percentage holds up** independently (~82% either way).
- rtk's `chars/4` heuristic actually **under-counts** the raw HTML: real HTML tokenizes at ~2.84 chars/token (tags, attributes, long URLs, entity refs each split into several BPE tokens), not 4. So the true tokens-saved on this page is ~102K, ~38% higher in absolute terms than the tracker reports — the dashboard is conservative, not optimistic.

This page is UTF-8, so it exercises the chrome-pruning path (the 53-entry interlanguage nav is gone) rather than the legacy-charset decode; diacritics round-trip cleanly (`programovací jazyk`, `Vývojář`, `otevřené encyklopedie`). `cl100k_base` is an OpenAI tokenizer used here as a widely-available proxy — exact counts vary per model, but the ratio is representative.

PR #2801, opened 2026-07-03
fix(git): compact git stash show instead of forcing -p

aeppling:
## Summary

- `git stash show` force-fed `-p`, turning git's compact diffstat into a full patch — *more* tokens than raw, and `rtk gain` measured savings against that inflated `-p` baseline (phantom ~68%).
- Now runs the command as typed (once, no forced `-p`) and strips the diffstat decorations (`|`, padding, ANSI, histogram bar) to `path <count> <+/->`; `-p` still goes through `compact_diff`.
- File list capped at 20 with a tail-hint for full recovery; `never_worse` guard means output is never larger than raw, and savings are measured against the real command.

### Before:

image

### After :

image

## Test plan

- [x] Manual testing: `rtk git stash show` output inspected (raw vs rtk, small/large/binary stashes, truncation + tail recovery, exit-code parity)

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -2205,26 +2205,26 @@ mod tests {
     fn test_compress_stat_summary_variants() {
         assert_eq!(
             compress_stat_summary("4 files changed, 60 insertions(+), 313 deletions(-)"),
-            "4 changed, 60 +, 313 -"
+            "4 changed 60 + 313 -"
         );
         assert_eq!(
             compress_stat_summary("1 file changed, 1 insertion(+)"),
-            "1 changed, 1 +"
+            "1 changed 1 +"
         );
         assert_eq!(
             compress_stat_summary("1 file changed, 1 deletion(-)"),
-            "1 changed, 1 -"
+            "1 changed 1 -"
         );
         assert_eq!(
             compress_stat_summary("2 files changed, 4 insertions(+), 1 deletion(-)"),
-            "2 changed, 4 +, 1 -"
+            "2 changed 4 + 1 -"
         );
     }
 
     #[test]
     fn test_compact_stash_stat_compresses_summary() {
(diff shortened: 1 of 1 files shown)


KuSh (approved):
As discussed, output can be compacted further:

- `9 +++++++++` -> `9+`
- `2 +`- -> `+-`
- `10 insertions(+), 1 deletion(-)` -> `10+-`

If more deletions, it would output like `10+5-`

src/cmds/git/git.rs
```diff
@@ -2096,6 +2170,104 @@ mod tests {
...
+            "1 changed, 1 +"
+        );
+        assert_eq!(
+            compress_stat_summary("1 file changed, 1 deletion(-)"),
+            "1 changed, 1 -"
+        );
+        assert_eq!(
+            compress_stat_summary("2 files changed, 4 insertions(+), 1 deletion(-)"),
+            "2 changed, 4 +, 1 -"
+        );
```
KuSh, line 2221: nitpick but we could gain some more chars:
- `4 changed: 60+313-`
- `1 changed: +`
- `1 changed: -`
- `2 changed: 4+-`

KuSh (approved):

src/cmds/git/git.rs
aeppling: Honestly hard to tell if removing the number prefix (when a single addition or deletion) would perturbate the LLM uderstanding.

```
1 changed: +
1 changed: -
```

Agree on removing comas as they are not needed and less likely to make sense for a LLM, whitespaces does not change anything for tokenisation. (most likely used as separator for token, not token)

aeppling: why not remove the number prefix : the gain for a single token versus the risk of LLM searching a possible missing information is not worth it i think

Issue #146, opened 2026-02-16
Feature Request: Add sudo prefix support for command rewriting


tmaarcxs:
## Problem

The RTK auto-rewrite hook currently doesn't handle commands prefixed with `sudo`. This means commands like:

- `sudo docker ps`
- `sudo docker logs <container>`
- `sudo git status`

are not rewritten to their RTK equivalents, resulting in missed token savings when running commands that require elevated privileges.

## Proposed Solution

Add logic to strip and preserve the `sudo` prefix (including flags like `sudo -u user`) before pattern matching, then restore it in the rewritten command.

### Changes to `.claude/hooks/rtk-rewrite.sh`

1. **Extract sudo prefix** after env var extraction:

```bash
# Strip leading sudo (with optional flags) for pattern matching
# e.g., "sudo docker ps" → match against "docker ps"
# e.g., "sudo -u root docker ps" → match against "docker ps"
# Handles: sudo, sudo -u user, sudo -i, sudo -E, sudo -v, sudo -k, etc.
SUDO_PREFIX=""
if echo "$MATCH_CMD" | grep -qE '^sudo([[:space:]]|$)'; then
  # Extract sudo and any flags (but stop at the actual command)
  SUDO_PREFIX=$(echo "$MATCH_CMD" | grep -oE '^sudo([[:space:]]+-[A-Za-z]+([[:space:]]+[^[:space:]]+)?)*[[:space:]]+' || echo "")
  if [ -n "$SUDO_PREFIX" ]; then
    MATCH_CMD="${MATCH_CMD:${#SUDO_PREFIX}}"
    CMD_BODY="${CMD_BODY:${#SUDO_PREFIX}}"
  fi
fi
```

2. **Update all REWRITTEN assignments** to include `${SUDO_PREFIX}`:

```bash
# Before
REWRITTEN="${ENV_PREFIX}rtk $CMD_BODY"

# After
REWRITTEN="${ENV_PREFIX}${SUDO_PREFIX}rtk $CMD_BODY"
```

### Example Transformations

| Input | Output |
|-------|--------|
| `sudo docker ps` | `sudo rtk docker ps` |
| `sudo docker logs container` | `sudo rtk docker logs container` |
| `sudo -u root git status` | `sudo -u root rtk git status` |

## Testing

Verified working with:

```bash
echo '{"tool_input":{"command":"sudo docker ps"}}' | ~/.claude/hooks/rtk-rewrite.sh
[...]

PR #2810, opened 2026-07-04
fix(rewrite): stop rewriting sudo commands (pass them through)

pszymkowiak:
## Problem (#146)

The env-prefix stripper (added in the original auto-rewrite hook) treats `sudo` like `env` / `VAR=val` and rewrites `sudo docker ps` → `sudo rtk docker ps`. Two problems:

1. **It breaks the command.** `rtk` installs to `~/.local/bin`, which is **not** on sudo's `secure_path` — so `sudo rtk docker ps` fails with `rtk: command not found` under root. Reported on #146 by a user.
2. **Where rtk *is* on secure_path, it runs rtk as root** — a large parsing/SQLite binary executing with elevated privileges, an unnecessary-privilege footgun for what is only output filtering.

## Fix (option "pass through")

Drop `sudo` from the env-prefix regex. sudo commands are now left **untouched** and pass through unchanged.

- The **permission verdict path is unaffected** — it never used this regex and already matches sudo commands as-is (`sudo:*` rules keep working), so least-privilege behaviour is unchanged.
- `env` / `VAR=val` prefixes and transparent builtins (`noglob`, `command`, …) still rewrite normally.

## Verification

| Input | Before | After |
|-------|--------|-------|
| `sudo docker ps` | `sudo rtk docker ps` (broken under root) | *(pass through)* |
| `sudo -u root docker ps` | *(not handled)* | *(pass through)* |
| `sudo noglob git status` | `sudo noglob rtk git status` | *(pass through)* |
| `env FOO=bar docker ps` | `env FOO=bar rtk docker ps` | unchanged ✅ |
| `FOO=bar docker ps` / `noglob git status` / `docker ps` | rewritten | unchanged ✅ |

Tests updated (`test_classify_sudo_not_stripped`, `test_rewrite_sudo_passthrough`, `test_sudo_with_builtin_not_rewritten`); `cargo fmt` / `cargo clippy --all-targets` clean; full test suite green (2361 passed).

## Note

This is the safe, minimal resolution. A future "real" sudo feature (rtk orchestrating `sudo <cmd>` while staying a non-root process) could be revisited separately if the docker/kubectl-under-sudo savings justify it — but it must never run rtk itself as root.

Refs #146

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -61,7 +61,11 @@ lazy_static! {
         let unquoted = r#"[^\s]*"#;
         let env_value = format!("(?:{}|{}|{})", double_quoted, single_quoted, unquoted);
         let env_assign = format!(r#"[A-Z_][A-Z0-9_]*={}"#, env_value);
-        Regex::new(&format!(r#"^(?:sudo\s+|env\s+|{}\s+)+"#, env_assign)).unwrap()
+        // NOTE: `sudo` is intentionally NOT stripped here. Rewriting `sudo docker ps`
+        // to `sudo rtk docker ps` breaks at runtime because `rtk` is not on root's
+        // secure_path, and (where it is) would run rtk itself as root. sudo commands
+        // are left untouched so they pass through unchanged. See #146.
+        Regex::new(&format!(r#"^(?:env\s+|{}\s+)+"#, env_assign)).unwrap()
     };
     // Git global options that appear before the subcommand: -C <path>, -c <key=val>,
     // --git-dir <dir>, --work-tree <dir>, and flag-only options (#163)
@@ -113,7 +117,7 @@ pub fn classify_command(cmd: &str) -> Classification {
         }
     }
 
-    // Strip env prefixes (sudo, env VAR=val, VAR=val)
+    // Strip env prefixes (env VAR=val, VAR=val); sudo is left untouched (#146)
     let stripped = ENV_PREFIX.replace(trimmed, "");
     let cmd_clean = stripped.trim();
     if cmd_clean.is_empty() {
@@ -1156,16 +1160,16 @@ mod tests {
     }
 
     #[test]
-    fn test_classify_sudo_stripped() {
-        assert_eq!(
-            classify_command("sudo docker ps"),
-            Classification::Supported {
-                rtk_equivalent: "rtk docker",
-                category: "Infra",
-                estimated_savings_pct: 85.0,
-                status: RtkStatus::Existing,
+    fn test_classify_sudo_not_stripped() {
+        // sudo is intentionally not stripped: sudo commands stay unclassified so
+        // they pass through unchanged rather than rewriting to a broken `sudo rtk`.
+        match classify_command("sudo docker ps") {
+            Classification::Unsupported { base_command } => {
+                // sudo is not peeled off, so the command is seen as-is (not `docker`).
+                assert_eq!(base_command, "sudo docker");
             }
-        );
+            other => panic!("expected Unsupported, got {:?}", other),
+        }
     }
 
     #[test]
@@ -3588,11 +3592,15 @@ mod tests {
     // --- sudo / env prefix + rewrite ---
 
     #[test]
-    fn test_rewrite_sudo_docker() {
+    fn test_rewrite_sudo_passthrough() {
+        // sudo commands are not rewritten (#146): `sudo rtk …` would fail under
+        // root's secure_path / run rtk as root. They pass through unchanged.
+        assert_eq!(rewrite_command_no_prefixes("sudo docker ps", &[]), None);
         assert_eq!(
-            rewrite_command_no_prefixes("sudo docker ps", &[]),
-            Some("sudo rtk docker ps".into())
+            rewrite_command_no_prefixes("sudo -u root docker ps", &[]),
+            None
         );
+        assert_eq!(rewrite_command_no_prefixes("sudo git status", &[]), None);
     }
 
     #[test]
@@ -4141,8 +4149,17 @@ mod tests {
     #[test]
     fn test_env_prefix_composed_with_builtin() {
         assert_eq!(
+            rewrite_command_no_prefixes("FOO=bar noglob git status", &[]),
+            Some("FOO=bar noglob rtk git status".into())
+        );
+    }
+
+    #[test]
+    fn test_sudo_with_builtin_not_rewritten() {
+        // A leading sudo blocks the rewrite even when a transparent builtin follows.
+        assert_eq!(
             rewrite_command_no_prefixes("sudo noglob git status", &[]),
-            Some("sudo noglob rtk git status".into())
+            None
         );
     }
 
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -66,7 +66,11 @@ static ENV_PREFIX: LazyLock<Regex> = LazyLock::new(|| {
     let unquoted = r#"[^\s]*"#;
     let env_value = format!("(?:{}|{}|{})", double_quoted, single_quoted, unquoted);
     let env_assign = format!(r#"[A-Z_][A-Z0-9_]*={}"#, env_value);
-    Regex::new(&format!(r#"^(?:sudo\s+|env\s+|{}\s+)+"#, env_assign)).unwrap()
+    // NOTE: `sudo` is intentionally NOT stripped here. Rewriting `sudo docker ps`
+    // to `sudo rtk docker ps` breaks at runtime because `rtk` is not on root's
+    // secure_path, and (where it is) would run rtk itself as root. sudo commands
+    // are left untouched so they pass through unchanged. See #146.
+    Regex::new(&format!(r#"^(?:env\s+|{}\s+)+"#, env_assign)).unwrap()
 });
 // Git global options that appear before the subcommand: -C <path>, -c <key=val>,
 // --git-dir <dir>, --work-tree <dir>, and flag-only options (#163)
@@ -122,7 +126,7 @@ pub fn classify_command(cmd: &str) -> Classification {
         }
     }
 
-    // Strip env prefixes (sudo, env VAR=val, VAR=val)
+    // Strip env prefixes (env VAR=val, VAR=val); sudo is left untouched (#146)
     let stripped = ENV_PREFIX.replace(trimmed, "");
     let cmd_clean = stripped.trim();
     if cmd_clean.is_empty() {
@@ -2161,16 +2165,16 @@ mod tests {
     }
 
     #[test]
-    fn test_classify_sudo_stripped() {
-        assert_eq!(
-            classify_command("sudo docker ps"),
-            Classification::Supported {
-                rtk_equivalent: "rtk docker",
-                category: "Infra",
-                estimated_savings_pct: 85.0,
-                status: RtkStatus::Existing,
+    fn test_classify_sudo_not_stripped() {
+        // sudo is intentionally not stripped: sudo commands stay unclassified so
+        // they pass through unchanged rather than rewriting to a broken `sudo rtk`.
+        match classify_command("sudo docker ps") {
+            Classification::Unsupported { base_command } => {
+                // sudo is not peeled off, so the command is seen as-is (not `docker`).
+                assert_eq!(base_command, "sudo docker");
             }
-        );
+            other => panic!("expected Unsupported, got {:?}", other),
+        }
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


aeppling:
Hey @pszymkowiak 

This does not fix the issue but bypass it, sudo should still get rewritten.
Anyway, if you want to drop sudo rewrite to solve this, there are still some blockers to solve before

1. rtk discover regression. classify_command loses sudo so sudo commands land in the TOP UNHANDLED COMMANDS -- open an issue? table. Observed vs develop: sudo docker / sudo apt / sudo systemctl / sudo RTK_DISABLED=1 / sudo.

2. RTK_DISABLED detection broken for sudo. strip_disabled_prefix shares the regex -> sudo RTK_DISABLED=1 docker ps no longer flagged. The RTK_DISABLED BYPASS section vanishes from discover, and gain.rs stops warning entirely (develop: [warn] 2 commands (14%) used RTK_DISABLED=1 unnecessarily; branch: nothing). No test covers sudo + RTK_DISABLED, hence green CI.

3. transparent_prefixes = ["sudo"] still emits sudo rtk docker ps. Verified on the branch binary (also sudo rtk git status, sudo rtk kubectl get pods). PR says rtk "must never run rtk itself as root", that needs a guard, not a default.

src/discover/registry.rs
```diff
@@ -61,7 +61,11 @@ lazy_static! {
         let unquoted = r#"[^\s]*"#;
         let env_value = format!("(?:{}|{}|{})", double_quoted, single_quoted, unquoted);
         let env_assign = format!(r#"[A-Z_][A-Z0-9_]*={}"#, env_value);
-        Regex::new(&format!(r#"^(?:sudo\s+|env\s+|{}\s+)+"#, env_assign)).unwrap()
+        // NOTE: `sudo` is intentionally NOT stripped here. Rewriting `sudo docker ps`
+        // to `sudo rtk docker ps` breaks at runtime because `rtk` is not on root's
+        // secure_path, and (where it is) would run rtk itself as root. sudo commands
+        // are left untouched so they pass through unchanged. See #146.
+        Regex::new(&format!(r#"^(?:env\s+|{}\s+)+"#, env_assign)).unwrap()
```
KuSh: **Blocking, but purely mechanical.** `develop` has since migrated this whole `lazy_static! { … }` block to individual `LazyLock` statics, which is why the PR shows as `CONFLICTING` / `DIRTY`. Resolution is to keep develop's form and make the same one-token deletion, with your comment carried over verbatim:

```diff
 static ENV_PREFIX: LazyLock<Regex> = LazyLock::new(|| {
     …
     let env_assign = format!(r#"[A-Z_][A-Z0-9_]*={}"#, env_value);
-    Regex::new(&format!(r#"^(?:sudo\s+|env\s+|{}\s+)+"#, env_assign)).unwrap()
+    // NOTE: `sudo` is intentionally NOT stripped here. Rewriting `sudo docker ps`
+    // to `sudo rtk docker ps` breaks at runtime because `rtk` is not on root's
+    // secure_path, and (where it is) would run rtk itself as root. sudo commands
+    // are left untouched so they pass through unchanged. See #146.
+    Regex::new(&format!(r#"^(?:env\s+|{}\s+)+"#, env_assign)).unwrap()
 });
```

Everything else in the PR auto-merges cleanly. I built and ran this exact resolution against current `develop`: **2651 tests pass**, and nothing on develop asserts `sudo rtk …`, so your three test edits are the complete set — no further fallout to chase.

src/discover/registry.rs
```diff
@@ -1156,16 +1160,16 @@ mod tests {
...
-        assert_eq!(
-            classify_command("sudo docker ps"),
-            Classification::Supported {
-                rtk_equivalent: "rtk docker",
-                category: "Infra",
-                estimated_savings_pct: 85.0,
-                status: RtkStatus::Existing,
+    fn test_classify_sudo_not_stripped() {
+        // sudo is intentionally not stripped: sudo commands stay unclassified so
+        // they pass through unchanged rather than rewriting to a broken `sudo rtk`.
```
KuSh: The behaviour is right; it's the wording. #146 inverts this exact assertion — `classify_command("sudo docker ps")` becomes `Supported { rtk_equivalent: "rtk sudo", category: "Elevation" }` — and a comment phrased as a permanent invariant will make that change look like a regression to whoever hits it. One clause fixes it:

```suggestion
        // sudo is intentionally not stripped: sudo commands stay unclassified and
        // pass through unchanged until `rtk sudo` lands (#146), rather than being
        // rewritten to a broken `sudo rtk`.
```

src/discover/registry.rs
```diff
@@ -3588,11 +3592,15 @@ mod tests {
...
+        // sudo commands are not rewritten (#146): `sudo rtk …` would fail under
+        // root's secure_path / run rtk as root. They pass through unchanged.
+        assert_eq!(rewrite_command_no_prefixes("sudo docker ps", &[]), None);
         assert_eq!(
-            rewrite_command_no_prefixes("sudo docker ps", &[]),
-            Some("sudo rtk docker ps".into())
+            rewrite_command_no_prefixes("sudo -u root docker ps", &[]),
+            None
         );
+        assert_eq!(rewrite_command_no_prefixes("sudo git status", &[]), None);
```
KuSh, line 4900: The four cases this passthrough actually depends on aren't pinned by anything:

- `FOO=1 sudo docker ps` and `env FOO=1 sudo docker ps` — `ENV_PREFIX` must lift the prefix *and* the remaining leading `sudo` must still block the rewrite. These two are the whole reason the `ENV_PREFIX` edit is the correct fix rather than a guard further down.
- `sudo` alone — no tail to rewrite.
- `sudoedit /etc/hosts` — must not be caught by anything matching `sudo`.

Those are precisely #3569's motivating cases, and today nothing would fail if a refactor broke them. All four already pass — I ran them against this branch (`test_rewrite_sudo_passthrough ... ok`) — so this is purely additive, and your existing comment and assertions are untouched.

The suggestion is rustfmt-clean (verified with `rustfmt --edition 2021`, hence the split `assert_eq!` calls — `fn_call_width` is 60):

```suggestion
        // sudo commands are not rewritten (#146): `sudo rtk …` would fail under
        // root's secure_path / run rtk as root. They pass through unchanged.
        assert_eq!(rewrite_command_no_prefixes("sudo docker ps", &[]), None);
        assert_eq!(
            rewrite_command_no_prefixes("sudo -u root docker ps", &[]),
            None
        );
        assert_eq!(rewrite_command_no_prefixes("sudo git status", &[]), None);
        // The passthrough must also survive an env prefix in front of sudo, a bare
        // `sudo`, and must not catch `sudoedit` (#3569's motivating cases).
        assert_eq!(
            rewrite_command_no_prefixes("FOO=1 sudo docker ps", &[]),
            None
        );
        assert_eq!(
            rewrite_command_no_prefixes("env FOO=1 sudo docker ps", &[]),
            None
        );
        assert_eq!(rewrite_command_no_prefixes("sudo", &[]), None);
        assert_eq!(
            rewrite_command_no_prefixes("sudoedit /etc/hosts", &[]),
            None
        );
```

src/discover/registry.rs
```diff
@@ -4140,9 +4148,18 @@ mod tests {
...
     fn test_env_prefix_composed_with_builtin() {
+        assert_eq!(
+            rewrite_command_no_prefixes("FOO=bar noglob git status", &[]),
+            Some("FOO=bar noglob rtk git status".into())
+        );
+    }
+
+    #[test]
+    fn test_sudo_with_builtin_not_rewritten() {
+        // A leading sudo blocks the rewrite even when a transparent builtin follows.
```
KuSh: Last one of the same kind. "A leading sudo blocks the rewrite" reads as a permanent rule, but #146 makes this assertion `Some("rtk sudo noglob git status")` — the `sudo` rule matches first and yields the tail verbatim.

```suggestion
        // A leading sudo blocks the rewrite, even when a transparent builtin
        // follows, until `rtk sudo` lands (#146).
```

KuSh (changes requested):
Reviewed with a `code-review high` pass on a locally built binary: a 17-command rewrite matrix, the `rtk hook claude` path, and all three `ENV_PREFIX` consumers.

**The behaviour is exactly right.** `sudo docker ps`, `sudo -u root …`, `sudo -E …`, `FOO=1 sudo …`, `env FOO=1 sudo …`, bare `sudo`, `sudoedit …` and `sudo noglob …` all pass through untouched, while `env FOO=1 docker ps`, `FOO=1 git status`, `noglob git status` and `docker ps` still rewrite normally. Compound commands split correctly (`ls && sudo docker ps` → `rtk ls && sudo docker ps`). And this is the fix that unblocks a real `rtk sudo`: removing `sudo\s+` from `ENV_PREFIX` is the prerequisite for a `^sudo` rule ever matching — a guard placed inside `rewrite_segment_inner` instead (as in #3569) would leave `sudo` stripped before any rule is consulted. Thanks for taking the right approach here.

Three requested changes, three of them mechanical.

### 1. Rebase — blocking

The PR is `CONFLICTING` against `develop`, which migrated `lazy_static! { … }` to individual `LazyLock` statics. See the inline comment on the `ENV_PREFIX` hunk for the exact resolution — I built and tested it, 2651 tests green.

### 2. Two doc lines are made false by this PR

Neither file is in the diff, so GitHub won't let me anchor these inline:

**`docs/guide/getting-started/configuration.md:94`**
```diff
-Patterns match against the full command after stripping env prefixes (`sudo`, `VAR=val`), so `"psql"` excludes both …
+Patterns match against the full command after stripping env prefixes (`VAR=val`), so `"psql"` excludes both …
```

**`src/discover/README.md:30`**
```diff
-3. Classify the command — strip env prefixes (`sudo`, `FOO="bar baz"`), normalize paths …
+3. Classify the command — strip env prefixes (`FOO="bar baz"`), normalize paths …
```

### 3. Missing coverage for the cases the passthrough depends on

`FOO=1 sudo …`, `env FOO=1 sudo …`, bare `sudo`, and `sudoedit …` are all correct today but pinned by nothing — and they're precisely #3569's motivating cases. Suggestion inline, rustfmt-clean, all four verified passing against this branch.

---

Two pre-existing problems surfaced while reviewing. **Neither is yours to fix** and neither blocks this PR — filed separately as #3756 (`sudo RTK_DISABLED=1 …` stops being counted as a bypass in `rtk discover`/`rtk gain`, an analytics-only hole) and #3757 (an unanchored `match_command` in `liquibase.toml`, which `timeout`/`nohup` already trigger on `develop` today).

Also note the CI `benchmark` job is red on `develop` itself (last five runs) — not from this PR.

src/discover/registry.rs
KuSh: Done — rebased onto the `LazyLock` form with your `NOTE` comment carried over verbatim, and the PR is `MERGEABLE` again.

Verified in a clean worktree, merged onto current `develop` (which has since picked up #3704's lexer refactor, also in `registry.rs`): `cargo fmt --check` clean, `cargo clippy --all-targets` zero warnings, `cargo test --all` **3080 passed / 0 failed**. Resolving.

src/discover/registry.rs
KuSh: Done — all four cases landed and pass (`test_rewrite_sudo_passthrough ... ok`), alongside `test_classify_sudo_not_stripped` and `test_sudo_with_builtin_not_rewritten`.

Spot-checked the built binary from the merge with current `develop` rather than trusting the assertions alone: `FOO=1 sudo …`, `env FOO=1 sudo …`, bare `sudo`, `sudoedit …`, `sudo -E …`, `sudo   docker ps` (double space) and `sudo noglob …` all pass through (exit 1, empty stdout), while `env FOO=1 docker ps` / `noglob git status` / `MSG="sudo" git status` still rewrite normally. Resolving.

KuSh (approved):
Verified and approved.

Rebase, both doc lines, and the four passthrough cases all landed as requested — thanks for the quick turnaround. I pushed one follow-up commit (b811434e) for two stale-documentation leftovers my earlier pass had missed, so nothing further is needed from you:

- `src/discover/README.md` — the dedicated "Env Prefix Handling" section still listed `sudo` as a stripped prefix and used `A="x y" B=1 sudo git status` as its chained example. That is the section a contributor reads before touching `ENV_PREFIX`, so it invited re-adding exactly what this PR removes. Example swapped to `env`, verified against the built binary.
- `test_head_tail_honour_exclude_commands` — the `sudo head -20 src/main.rs` assertion had become vacuous: it returns `None` whether or not `head` is excluded, because sudo blocks the rewrite outright, so it no longer tested exclusion at all. Dropped it, and corrected the two comments claiming `strip_disabled_prefix` peels `sudo` (one in `rewrite_segment_inner`, one in the test). The `RUST_LOG=debug tail` case still covers the env-prefix path.

Verification, in a clean worktree and separately merged onto current `develop` (which has since picked up #3704's lexer refactor in this same file): `cargo fmt --check` clean, `cargo clippy --all-targets` zero warnings, `cargo test --all` green on both. Behaviour re-checked on the built binary rather than inferred from the assertions — every sudo form passes through (including `sudo -E`, a double-space `sudo   docker ps`, and `sudoedit`), non-sudo rewrites are unaffected, `ls && sudo docker ps` becomes `rtk ls && sudo docker ps`, and the hook path emits nothing for `sudo docker ps`.

Two known consequences are tracked separately and are deliberately not blockers: sudo commands now surface under `TOP UNHANDLED` in `rtk discover` (accepted — #146 collapses them into a single Elevation row), and #3756 covers the `sudo RTK_DISABLED=1 …` bypass accounting.

PR #2812, opened 2026-07-04
fix(grep): add regression test for --ultra-compact grep data loss

### LINKED ISSUE #2620: `--ultra-compact` breaks `rtk grep`: returns "No such file or directory (os error 2)" and drops all matches with exit 0
## Summary

Under `--ultra-compact`, `rtk grep` returns an error string instead of results and **exits 0**, silently dropping all matches. This is especially dangerous because `--ultra-compact` is documented for use in the Claude Code `PreToolUse` hook (`rtk hook claude --ultra-compact`), where the exit-0 masks the failure and the agent sees zero matches as if the grep legitimately found nothing.

## Environment

- rtk **0.42.4** (homebrew, `/opt/homebrew/bin/rtk`)

## Reproduction

```sh
TMP=$(mktemp -d)
printf 'alpha permissions one\nbeta permissions two\n' > "$TMP/sample.txt"
printf 'permissions here\nmore permissions\n'          > "$TMP/sample2.txt"

# Works:
rtk grep -r "permissions" "$TMP"
# -> real matches, ~379 bytes

# Broken:
rtk --ultra-compact grep -r "permissions" "$TMP"
# -> [rtk: No such file or directory (os error 2)]
# -> exit code 0, zero matches returned
```

## Expected

`rtk --ultra-compact grep` returns the same matches as `rtk grep` (just more compactly), or at minimum exits non-zero on failure.

## Actual

Output: `rtk: No such file or directory (os error 2)`
Exit code: `0`
All matches lost.

## Scope

Grep-specific. `rtk --ultra-compact find` and `rtk --ultra-compact ls` produce byte-identical output to their normal counterparts — only the `grep` filter is affected under ultra-compact.

## Impact

Anyone enabling `--ultra-compact` globally in the hook (the natural move to maximize token savings) silently loses all grep output, and the exit-0 makes it undetectable without byte-diffing against normal mode.

guyoron1:
## Summary
- `--ultra-compact grep` was reported to fail with ENOENT and exit 0, silently dropping all matches (#2620)
- The bug no longer reproduces on current develop — likely fixed by recent grep error-handling improvements (`05f3c54`, `eafadce`)
- This PR adds targeted regression tests to prevent future regressions

## Changes
- **Integration test** (`tests/search_error_test.rs`): verifies `rtk --ultra-compact grep -r` returns identical matches to `rtk grep -r`, exits 0, and produces no ENOENT errors
- **Clap parsing test** (`src/main.rs`): verifies `--ultra-compact` is parsed as a global flag and does not leak into grep's `extra_args`

Fixes #2620

## Test plan
- [ ] `cargo test ultra_compact_grep_returns_matches_not_error` passes
- [ ] `cargo test test_ultra_compact_grep_parses_correctly` passes
- [ ] Full CI passes on all platforms (macOS, Linux, Windows)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3583,6 +3583,34 @@ mod tests {
     }
 
     #[test]
+    fn test_ultra_compact_grep_parses_correctly() {
+        let cli = Cli::try_parse_from([
+            "rtk",
+            "--ultra-compact",
+            "grep",
+            "-r",
+            "permissions",
+            "/tmp/dir",
+        ])
+        .unwrap();
+        assert!(
+            cli.ultra_compact,
+            "--ultra-compact must be parsed as a global flag, not consumed by grep"
+        );
+        match cli.command {
+            Commands::Grep { extra_args, .. } => {
+                assert!(
+                    !extra_args.contains(&"--ultra-compact".to_string()),
+                    "--ultra-compact must not leak into grep extra_args: {:?}",
+                    extra_args
+                );
+                assert_eq!(extra_args, vec!["-r", "permissions", "/tmp/dir"]);
+            }
+            _ => panic!("Expected Commands::Grep"),
+        }
+    }
+
+    #[test]
     fn test_npx_unknown_tool_passthrough() {
         // The bug (rtk-ai/rtk#815) was that unknown tools under `rtk npx`
         // were dispatched to `npm` instead of `npx`. At the parse level, the
--- a/tests/search_error_test.rs
+++ b/tests/search_error_test.rs
@@ -106,6 +106,43 @@ fn no_match_stays_exit_1_and_empty() {
 }
 
 #[test]
+fn ultra_compact_grep_returns_matches_not_error() {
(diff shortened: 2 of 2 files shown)


guyoron1:
Rebased. The `--ultra-compact grep` report no longer reproduces on develop, since grep now takes every flag through `extra_args` (v0.42.4 couldn't parse `-r`, so the fallback tried to run `--ultra-compact` as a program). I dropped the clap parsing test because the existing `test_grep_parse_*` tests already fail on the same break, and kept the end-to-end one: it exits 127 against v0.42.4, passes on develop, and it's the only test that runs a global flag in front of grep. The commit is now `test(...)` and `Refs` instead of `Fixes`, since this only adds coverage.

Separately, `run_fallback` still takes a leading rtk global flag as the program (`rtk --ultra-compact echo hi` exits 127 on develop, for commands clap doesn't parse). That's a small separate fix I didn't fold in here. It's also fine to close #2620 as fixed-by-refactor and close this PR.

PR #2816, opened 2026-07-04
feat: codex now support PreToolUse hook

### LINKED ISSUE #1812: Codex now support PreToolUse hook

wilmerdooley:
Resolves #1812.

As the title suggests, could you add a PreToolUse hook to the Codex, similar to the one in Claude's code?

## Summary

-

## Test plan

### DIFF
--- a/src/hooks/README.md
+++ b/src/hooks/README.md
@@ -28,7 +28,7 @@ LLM agent integration layer that installs, validates, and executes command-rewri
 | Claude-MD (legacy) | `rtk init --claude-md` | 134-line RTK block | CLAUDE.md |
 | Windsurf | `rtk init -g --agent windsurf` | `.windsurfrules` | -- |
 | Cline | `rtk init --agent cline` | `.clinerules` | -- |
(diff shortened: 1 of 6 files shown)


TyceHerrman:
https://github.com/rtk-ai/rtk/pull/2557 already open and reviewed and i think also would close #1812

KuSh:
Thanks for pushing on this, and for the clear demonstration that Codex could take updatedInput. Native Codex PreToolUse support merged on 2026-09-13 in #3552, using the same approach: `rtk hook codex` reads the event, rewrites the Bash command and returns it through updatedInput with a protocol level allow, and `rtk init --codex` installs, reports status and uninstalls it. It adds a couple of guards on top, skipping unknown permission modes and commands the lexer cannot attest, so this PR is covered by it. One thing your version did that develop does not is emit a deny decision when an RTK deny rule matches; today the Codex path only audit logs skip:deny_rule and stays silent, so the command runs, and an issue or a small PR against the current process_codex_payload would be a good place to settle that. If I missed something this PR still adds, please comment here or reopen.

PR #2824, opened 2026-07-04
fix(find): fail closed on unknown flags instead of warn-and-broaden

### LINKED ISSUE #2821: `rtk find` warns on unknown predicates but then executes a DIFFERENT (broader) query — silent wrong results
**Environment:** rtk 0.28.2 · Ubuntu 24.04 (WSL2) · installed as the Claude Code PreToolUse Bash-rewrite hook (hook v2)

**Repro:**

```console
$ touch -d '2020-01-01' old.txt && touch new.txt sample.txt

$ /usr/bin/find . -newermt '2025-01-01' -name '*.txt' | sort
./new.txt
./sample.txt

$ rtk find . -newermt '2025-01-01' -name '*.txt'
rtk find: unknown flag '-newermt', ignored
📁 3F 1D:
./ new.txt old.txt sample.txt        # old.txt included — the filter was dropped
```

The warning is printed, but the command still **succeeds (exit 0) with the wrong result set**. No agent (or shell script, or human piping output) treats a stdout warning as failure — the unfiltered set gets consumed as if it answered the original query. In our transcripts this silently fed a wrong file list into a session-audit scan.

By contrast, `-not`/`-exec` hard-fail with a clear "use `find` directly" message and exit 1 — that's the right behavior. Unknown flags should get the same treatment (or better: transparent passthrough to native `find`), never "ignore and run a broader query".

jinn001-001:
Fixes #2821.

## Problem

`rtk find` handles unknown predicates by printing a stderr warning and continuing:

```console
$ rtk find . -newermt '2025-01-01' -name '*.txt'
rtk find: unknown flag '-newermt', ignored
<UNFILTERED list including files older than 2025>   # exit 0
```

The dropped predicate means the walk runs a **broader** query than the user asked for, while still exiting 0. Scripted and LLM-agent callers can't act on stderr warnings, so the superset is consumed as the real answer (this fed a wrong file set into a session-audit pipeline for us). `-not`/`-exec` already hard-fail; `-newermt`-class flags miss the exact-match `UNSUPPORTED_FIND_FLAGS` list and fall into the warn arm instead.

## Fix

The unknown-flag arm in `parse_native_find_args` now bails the same way the compound-predicate guard does:

```
rtk find does not support flag '-newermt'. Use `find` directly.
```

exit 1 via the existing error path. Per the "Correctness VS Token Savings" principle: an explicit user predicate is intent; dropping it should never be a warning.

## Test plan

- 2 new unit tests (`parse_native_find_rejects_unknown_flag_newermt`, `..._anewer`) mirroring the existing `parse_native_find_rejects_not` template — RED on master's warn-and-continue, GREEN here.
- Full `cargo test`: 2360+ pass; the only failures ever observed were `core::tracking::tests::test_timed_execution_*` timing assertions, which also flake on clean master under parallel load and pass in isolation on both — unrelated to this diff.
- Manual: `rtk find . -newermt 2025-01-01 -name '*.txt'` → error + exit 1; `rtk find . -name '*.txt'` unchanged.

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -133,7 +133,14 @@ fn parse_native_find_args(args: &[String]) -> Result<FindArgs> {
                 }
             }
             flag if flag.starts_with('-') => {
-                eprintln!("rtk find: unknown flag '{}', ignored", flag);
+                // Fail closed: dropping an unknown predicate and continuing would
+                // silently run a BROADER query (e.g. a `-newermt` filter lost while
+                // still exiting 0), which scripted and agent callers consume as the
+                // real answer. Unknown flags get the same treatment as -not/-exec.
+                anyhow::bail!(
+                    "rtk find does not support flag '{}'. Use `find` directly.",
+                    flag
+                );
             }
             _ => {}
         }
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into this one. The broadening behaviour it fixes was resolved on develop by #3603, merged on 2026-08-26, which reworked find_cmd.rs to parse find's own grammar and hand any predicate rtk does not model to the real find, then compress its output. Checked against the develop binary: `rtk find . -newermt 2025-01-01 -name '*.txt'` now returns only the file newer than the cutoff instead of the broader set, and the warn-and-ignore arm this PR targeted is gone from the file. The tests here assert a hard error on unknown flags, which develop intentionally no longer raises since it delegates instead. Closing as covered, but please comment or reopen if you see a predicate that still slips through.

Issue #2820, opened 2026-07-04
`rtk rewrite` maps `tail -N <file>` to an invalid `rtk read -N` invocation; failure mode breaks LLM-agent pipelines


jinn001-001:
**Environment:** rtk 0.28.2 · Ubuntu 24.04 (WSL2) · installed as the Claude Code PreToolUse Bash-rewrite hook (hook v2)

**Repro:**

```console
$ rtk --version
rtk 0.28.2

$ rtk rewrite "tail -120 /tmp/x.log"
rtk read -120 /tmp/x.log
# -120 is not a valid `rtk read` flag (the supported form is -m/--max-lines)

$ rtk read -120 ./definitely-missing.log      # the rewritten form
[rtk: Permission denied (os error 13)]
$ echo $?
127

$ rtk read ./definitely-missing.log           # same file, WITHOUT the bogus flag
Error: Failed to read file: ./definitely-missing.log
Caused by: No such file or directory (os error 2)
$ echo $?
1
```

**The invalid rewrite is the single root cause; everything downstream is fallout:**

1. The rewrite emits a flag `rtk read` doesn't define (`-N` tail shorthand is not translated to `-m/--max-lines N`).
2. Even a faithful flag translation would change meaning: `--max-lines` reads from the TOP (verified: `rtk read tenline.txt -m 3` starts at line 1), so `tail -120 f` (last 120 lines) cannot map onto `rtk read` at all.
3. With the invalid flag present, failure surfaces on stderr as a bracketed marker with a misleading errno (`Permission denied (os error 13)` for a file that simply doesn't exist; `os error 2` also observed in the wild) and exit **127** ("command not found" semantics). Note `rtk read <missing>` *without* the flag behaves correctly (clean ENOENT message, exit 1) — the error-path regression only triggers on the malformed invocation rtk's own rewriter produces.

[...]

PR #2825, opened 2026-07-04
fix(rewrite): respect exclude_commands in head/tail fast path; map bare head to --max-lines 10

### LINKED ISSUE #2823: `[hooks] exclude_commands` is bypassed for `head`/`tail` (line-range fast path skips the exclusion check)

```console
1

```

```rust
}
```

jinn001-001:
Fixes #2823. Also resolves the bare-`head` residual noted in #2820.

## Problem

The head/tail line-range fast path in `rewrite_segment_inner` returns before any exclusion check, with two consequences:

1. **`[hooks] exclude_commands = ["head", "tail"]` is silently ignored** — `is_excluded` is only consulted in the classification arms downstream of the early return, so commands like `rg` respect exclusions but `head`/`tail` never reach the check (repro in #2823, verified on 0.43.0 both via `rtk rewrite` and `rtk hook claude`).
2. **Bare `head <file>` mistranslates.** The interception required `head -` (with a dash), so the dashless form fell through to the generic prefix-strip and became `rtk read <file>` — the **whole file**, where head's contract is the first 10 lines.

## Fix

- The fast path now checks `is_excluded(cmd_part, excluded)` before rewriting — same semantics as the classification arms. Exclusions are the user saying "hands off this command".
- The guard widens from `head -` to `head `, routing bare `head <file>` through `rewrite_line_range`, which now maps it to `rtk read <file> --max-lines 10` via a new `HEAD_BARE` regex — faithful to head's default while keeping the token savings (per the existing `test_rewrite_head_no_flag_still_rewrites` intent, whose expectation is updated accordingly). Multi-file bare head matches no pattern and passes through natively, consistent with the #1362 `==> name <==` banner reasoning. `head -c`-style unsupported flags keep their existing passthrough.

## Test plan

- 4 new unit tests: head/tail/bare-head exclusion respected; bare-head single-file → `--max-lines 10`; bare-head multi-file passthrough; plus the not-excluded rewrite case.
- 1 existing test updated (`test_rewrite_head_no_flag_still_rewrites`): expectation `rtk read src/main.rs` → `rtk read src/main.rs --max-lines 10` — the old expectation encoded the whole-file bug.
- Full `cargo test`: 2365 passed / 0 failed (the only intermittent failures on this box are `core::tracking::tests::test_timed_execution_*` timing flakes, which also occur on clean HEAD under parallel load and pass in isolation — unrelated).

Note: trivially conflicts with nothing currently open that I can see; independent of #2824.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -73,6 +73,10 @@ lazy_static! {
     // `==> name <==` banners that `rtk read --max-lines` cannot reproduce.
     static ref HEAD_N: Regex = Regex::new(r"^head\s+-(\d+)\s+(\S+)$").unwrap();
     static ref HEAD_LINES: Regex = Regex::new(r"^head\s+--lines=(\d+)\s+(\S+)$").unwrap();
+    // Bare `head <file>` (single non-flag argument): head's default is the FIRST
+    // 10 LINES, so the faithful rewrite carries `--max-lines 10` — a plain
+    // `rtk read <file>` would print the whole file.
+    static ref HEAD_BARE: Regex = Regex::new(r"^head\s+([^-\s]\S*)$").unwrap();
     static ref TAIL_N: Regex = Regex::new(r"^tail\s+-(\d+)\s+(\S+)$").unwrap();
     static ref TAIL_N_SPACE: Regex = Regex::new(r"^tail\s+-n\s+(\d+)\s+(\S+)$").unwrap();
     static ref TAIL_LINES_EQ: Regex = Regex::new(r"^tail\s+--lines=(\d+)\s+(\S+)$").unwrap();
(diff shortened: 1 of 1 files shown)


KuSh:
Closing in favour of #3324 for the exclusion half; it will carry `Fixes #2823`.

Two blockers here regardless: the CLA is unsigned and the branch is in conflict.

Separately, the bare-`head` mapping to `--max-lines 10` is a different change from the exclusion fix. The intent is right — `head`'s default really is 10 lines — but it belongs with the other `rtk read` semantics work (#3563, #3668, #3678), not bundled into an exclusion fix. If you sign the CLA, a focused PR for just that piece would be welcome.

PR #2835, opened 2026-07-05
fix(discover): encode -p project path for sanitized transcript dir names

### LINKED ISSUE #2834: discover: -p with an absolute path silently matches nothing (sanitized transcript dir names)
## Summary

`rtk discover -p <absolute path>` always reports **0 sessions** and prints the success message `No missed savings found. RTK usage looks good!` — a false all-clear, when the same project filtered by a name substring reports many sessions with substantial missed savings.

## Repro (rtk 0.36.0, Linux)

```console
$ cd /home/user/projects/my-app
$ rtk discover --since 7 | grep Scanned
Scanned: 16 sessions (last 7 days), ~1400 Bash commands        # cwd mode: OK
$ rtk discover -p my-app --since 7 | grep Scanned
Scanned: 16 sessions (last 7 days), ~1400 Bash commands        # name substring: OK
$ rtk discover -p /home/user/projects/my-app --since 7 | grep Scanned
Scanned: 0 sessions (last 7 days), 0 Bash commands             # absolute path: silently nothing
```

## Cause (hypothesis)

Claude Code stores transcripts under sanitized project directory names where path separators become dashes (`~/.claude/projects/-home-user-projects-my-app/`). A `-p` value containing `/` can never substring-match the sanitized directory name, so the filter matches zero projects.

## Expected

Either sanitize the `-p` input the same way before matching (the flag is documented as "Filter by project path", so a path should work), or error out when the filter matches no known project directory instead of printing the success banner.

## Impact

Two failure shapes produce the same misleading "looks good" output:

1. `-p <path>` (the documented "project path" form) → 0 sessions, success banner.
2. Default (current-project) mode when the cwd isn't a directory with transcripts → 0 sessions, success banner. A hint like "no transcripts found for this project — try `--all`" would prevent misreading it as "nothing to save".

Found while investigating why `discover` reported nothing for a project that shows significant missed rewrites under `--all`.

lesbass:
## Summary

Fixes #2834 — `rtk discover -p /absolute/path` silently matched 0 sessions.

## Problem

Claude Code stores transcripts under sanitized directory names where `/` becomes `-` (e.g., `~/.claude/projects/-home-user-projects-my-app/`). When a user provides `-p /absolute/path`, the filter value containing `/` characters can never substring-match the sanitized directory names.

The CWD fallback path in `mod.rs` already correctly encodes the path with `encode_project_path()`, but the explicit `-p` path was used raw.

## Fix

Apply `ClaudeProvider::encode_project_path()` to the `-p` value before using it as a filter — the same encoding already applied for the default current-working-directory path.

## Change

- `src/discover/mod.rs`: 1 line changed — encode `-p` path before filtering

## Verification

The existing tests for `encode_project_path` (in `provider.rs`) already cover:
- Absolute Unix paths: `/Users/foo/bar` → `-Users-foo-bar`
- Dots in paths: `/Users/first.last/my-project` → `-Users-first-last-my-project`
- Underscores: `/home/chris/2_project-files/proj` → `-home-chris-2-project-files-proj`
- Non-ASCII, Windows backslashes, spaces, brackets

The `test_match_project_filter` test validates that encoded paths correctly contain project name substrings.

Local `cargo` build could not run (no C compiler in this environment). The change follows the exact same pattern as the existing CWD encoding on lines 62-63 of the same file.

### DIFF
--- a/src/discover/mod.rs
+++ b/src/discover/mod.rs
@@ -54,7 +54,8 @@ pub fn run(
     let project_filter = if all {
         None
     } else if let Some(p) = project {
-        Some(p.to_string())
+        let encoded = ClaudeProvider::encode_project_path(p);
+        Some(encoded)
     } else {
         // Default: current working directory
         let cwd = std::env::current_dir()?;


lesbass:
Closing this PR after 14 days with no human maintainer engagement. Happy to reopen if maintainers are interested in this change — it's a small, focused fix (2 additions, 1 deletion) that sanitizes project paths for transcript directory names. Reopen at any time or let us know if you'd like us to adjust the approach.

PR #2836, opened 2026-07-05
fix(grep): preserve -- separator between non-adjacent match blocks

### LINKED ISSUE #2795: grep -A/-B/-C drops the "--" separator between non-adjacent match blocks
## Summary

`rtk grep` drops the `--` separator that native `grep` prints between **non-adjacent** match blocks when context flags (`-A` / `-B` / `-C`) are used. Because the separator is the only thing that delimits one match's context from the next, the filtered output silently merges distinct matches into what looks like one contiguous block. A consumer (human or LLM) can no longer tell where one match ends and the next begins.

## Environment

- rtk 0.43.0
- macOS (Darwin), zsh
- Invoked via the Claude Code auto-rewrite hook (`rtk hook claude`), i.e. `grep -n -A1 …` → `rtk grep -n -A1 …`

## Reproduction

```sh
cat > repro.txt <<'EOF'
alpha match here
filler 1
filler 2
filler 3
filler 4
filler 5
beta match here
EOF

grep -n -A1 'match' repro.txt      # native
rtk grep -n -A1 'match' repro.txt  # rtk
```

**Native `grep` (correct):**
```
1:alpha match here
2-filler 1
--
7:beta match here
```

**`rtk grep` (bug — no `--`):**
```
1:alpha match here
2-filler 1
7:beta match here
```

## Expected

`rtk grep` should preserve the `--` group separator between non-adjacent match blocks, exactly as native grep does, so match boundaries remain unambiguous.

## Actual

The `--` separator is stripped. Line `2-filler 1` (context of match #1) and line `7:beta match here` (match #2) appear adjacent, implying they are one continuous region. With longer `-A`/`-B`/`-C` windows and many matches this compounds — the output reads as one block when it is actually N separate matches.

## Impact

This is a correctness issue, not just a formatting nit: the transformation changes what the output *means*. It's the reason I routinely bypass `rtk grep` and fall back to native grep whenever I need context lines, which defeats the token-saving purpose for exactly the searches where context matters most.

## Suggested fix

When emitting context (`-A`/`-B`/`-C`), keep grep's block-separator semantics — re-emit `--` between non-contiguous line groups (grep only omits it when blocks actually overlap/abut). Grouping by file is fine; the within-file block boundaries just need to survive.

guyoron1:
## Summary
- `grep -A/-B/-C` context output was dropping the `--` group separator between non-adjacent match blocks
- Consumers (human or LLM) could not tell where one match ended and the next began
- Fixed by preserving `--` in both the faithful baseline and grouped output paths

Fixes #2795

## Test plan
- [x] Added regression test (`group_separator_between_non_adjacent_matches`)
- [x] Added faithful format test (`context_group_separator_matches_grep_n`)
- [x] Verified output matches native `grep -n -A1` exactly

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -339,6 +339,20 @@ fn has_short_flag(flags: &[String], ch: char) -> bool {
         .any(|f| f.starts_with('-') && !f.starts_with("--") && f[1..].contains(ch))
 }
 
+fn has_context_flag(flags: &[String]) -> bool {
+    has_short_flag(flags, 'A')
+        || has_short_flag(flags, 'B')
+        || has_short_flag(flags, 'C')
+        || flags.iter().any(|f| {
+            f == "--after-context"
+                || f == "--before-context"
+                || f == "--context"
+                || f.starts_with("--after-context=")
+                || f.starts_with("--before-context=")
+                || f.starts_with("--context=")
+        })
+}
+
 pub fn run(
     engine: Engine,
     max_line_len: usize,
@@ -482,6 +496,8 @@ pub fn run(
         plain.push('\n');
     }
 
+    let has_context = has_context_flag(&extra_args);
+
     let per_file = config::limits().grep_max_per_file;
     let mut files: Vec<_> = by_file.iter().collect();
     files.sort_by_key(|(f, _)| *f);
@@ -504,7 +520,7 @@ pub fn run(
             if shown >= max_results {
                 break;
             }
-            if prev_line > 0 && *line_num > prev_line + 1 {
+            if has_context && prev_line > 0 && *line_num > prev_line + 1 {
                 body.push_str("--\n");
             }
             prev_line = *line_num;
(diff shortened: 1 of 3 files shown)


aeppling:
Hey @guyoron1 , thanks for the contribution it solve the core problem of #2795 

One blocker before merge;
This inserts a -- on any line-number gap, but it never checks that context (-A/-B/-C) was actually requested.
Native grep only prints -- in context mode. This reconstructs separators from line-number gaps unconditionally, so aplain (non-context) search invents -- between matches that grep never emits.

Repro:
grep -rn todo <dir>
rtk grep -rn todo <dir> # here lot of -- printed

Single large file behaves the same: grep -n match huge.txt, native 0 but rtk 24. The output ends up with a -- between essentially every match.

Should gate the insertion so it runs when actually requested context.

aeppling:
Hey @guyoron1 , thanks for the fix, LGTM 


aeppling (approved):

guyoron1:
woohoooooo yay ! 
I'm using RTK a ton, so why not contribute : )

guyoron1:
Heyaaaa!

I've actually faced a very similar issue on #2751 — the benchmark `curl json` test hit the same intermittent failure there too. The root cause was identified in #2551 and fixed within RTK by the never-worse guard in #2554.

The benchmark gate has passed on the latest CI runs, but the `curl json` failure seen earlier is a known intermittent issue — the benchmark runs `curl -s https://httpbin.org/json` and `rtk curl https://httpbin.org/json` as **separate HTTP requests**, so response variations between calls cause non-deterministic token count comparisons.

The never-worse guard (#2554) works correctly within a single RTK execution, but the benchmark's cross-execution comparison is structurally flaky for live endpoints.

Filed #2842 to fix the benchmark by replacing live HTTP calls with local fixtures.

This PR only touches grep filter code (`src/cmds/system/search.rs`) and grep tests — no curl code was modified. : )

PR #2837, opened 2026-07-05
feat: add cat output filter

### LINKED ISSUE #2803: antigravity ignores rtk popular commands to filter
tried to add rtk v 0.43.0 to my antigravity, restarted, but see that 
grep, docker, cat, npm, vue build, npm build, etc, ignored. it is possible to add more standart commands to filter, and allow additional custom filter file, as example i have company owned scripts/utils which not available public, but rtk save ~50-60% tokens on run them?

MeGaurav4:
Adds a compact output filter for `cat`, one of the most commonly used commands that was missing from the built-in filters list.

Files changed: src/filters/cat.toml (new), src/core/toml_filter.rs (filter count bump)

Closes #2803

## Verification

- `cargo test` passes (2360 tests, 0 failures)
- New filter has 3 inline tests: basic output, trailing blanks, empty input

### DIFF
--- a/src/core/toml_filter.rs
+++ b/src/core/toml_filter.rs
@@ -1896,8 +1896,8 @@ match_command = "^make\\b"
         let filters = make_filters(BUILTIN_TOML);
         assert_eq!(
             filters.len(),
-            63,
-            "Expected exactly 63 built-in filters, got {}. \
+            64,
+            "Expected exactly 64 built-in filters, got {}. \
              Update this count when adding/removing filters in src/filters/.",
             filters.len()
         );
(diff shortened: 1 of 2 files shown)


pszymkowiak:
Closing: already covered. The hook has rewritten `cat <file>` to `rtk read <file>` since f447a3d5, and native rules take precedence over TOML filters, so `cat.toml` is never reached from the hook. Adding it would only give a manually typed `rtk cat` a second, lossier behaviour (100-line cap, blank lines silently dropped) than `rtk read`, which returns the file whole. If `rtk cat` parity is wanted, an alias to `read` in main.rs would be the consistent way. Thanks for the PR.

Issue #2842, opened 2026-07-06
fix(benchmark): curl/wget tests are non-deterministic (live HTTP requests)


mysticgohan1:
## Problem

The benchmark CI gate (`scripts/benchmark.sh`) intermittently fails on `curl json`, `curl text`, and `wget` tests with negative token savings (e.g., `35 → 38 (-8%)` or `8 → 40 (-400%)`).

This blocks unrelated PRs — e.g., #2836 (grep context separator fix) showed a `curl json` failure despite touching no curl code.

## Root Cause

`scripts/benchmark.sh` runs the raw command and RTK command as **separate invocations** (lines 44-45):

```bash
unix_out=$(eval "$unix_cmd" 2>/dev/null || true)
rtk_out=$(eval "$rtk_cmd" 2>/dev/null || true)
```

For curl/wget tests (lines 349-350, 358), this means **two separate HTTP requests** to `httpbin.org`:

```bash
bench "curl json" "curl -s https://httpbin.org/json" "$RTK curl https://httpbin.org/json"
bench "curl text" "curl -s https://httpbin.org/robots.txt" "$RTK curl https://httpbin.org/robots.txt"
bench "wget"      "wget -qO- https://httpbin.org/json"     "$RTK wget https://httpbin.org/json"
```

Response variations between the two calls (timestamps, headers, content-length, server-side changes) cause token count mismatches. The never-worse guard inside RTK (#2554) works correctly within a single execution, but the benchmark's cross-execution comparison is inherently non-deterministic for live endpoints.

## Prior Art

Same class of issue as:
- **#2551** — identified `curl` emitting more tokens than raw on tiny responses
- **#2554** — added `core::guard::never_worse()` to solve the within-RTK problem
- **#2751** — parser fix PR that also had this exact benchmark flakiness; [commenter noted](https://github.com/rtk-ai/rtk/pull/2751#issuecomment-4865304124) the curl test is non-deterministic

The never-worse guard (#2554) solved the within-RTK problem. But the benchmark itself still compares outputs from two separate HTTP requests, making it structurally flaky.

## Proposed Fix

Replace live `httpbin.org` calls with local fixtures, following the pattern already used by the `json` and `log` benchmark sections (lines 228-246, 286-302):

```bash
# Create fixture (matching httpbin.org/json response)
cat > /tmp/rtk_bench_curl.json << 'CURLEOF'
{"slideshow":{"author":"Yours Truly","date":"date of publication","slides":[...]}}
[...]

PR #2845, opened 2026-07-06
fix(benchmark): replace live HTTP calls with local fixtures for curl/wget

guyoron1:
## Summary

- Replace live HTTP endpoint calls (`mockhttp.org`) in `curl` and `wget` benchmark tests with local fixture files using `file://` URLs
- Follows the pattern already used by the `json` (line 228) and `log` (line 286) benchmark sections
- Eliminates non-deterministic CI failures caused by response variations between the two separate invocations in `bench()`

## Problem

The benchmark's `bench()` function runs the raw command and RTK command as separate invocations (lines 44-45). For curl/wget, this means two separate HTTP requests — response variations (timestamps, headers, content-length) cause token count mismatches and false "negative savings" failures that block unrelated PRs.

## Test plan

- [ ] Run `bash scripts/benchmark.sh` — curl/wget tests should pass deterministically
- [ ] Run multiple times — no flaky failures
- [ ] Verify curl filter still produces token savings on the fixture content

Refs #2842

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -342,21 +342,35 @@ section "wc"
 bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 
 # ===================
-# curl
+# curl (local fixtures — live HTTP endpoints caused non-deterministic failures, see #2842)
 # ===================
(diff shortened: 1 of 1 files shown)


KuSh:

PR #2846, opened 2026-07-06
fix(next): don't force `next build` for other subcommands

### LINKED ISSUE #2840: Running npx next dev does npx next build instead
When running `rtk npx next dev` I get the output:
```
Next.js Build
Errors: 0 | Warnings: 0
```

When I do `npx next dev` the next app goes to dev run as it should

ErezAmihud:
## Summary
`rtk next dev`/`rtk npx next dev` always ran `next build` regardless of the requested subcommand, silently turning `dev`, `start`, `lint`, etc. into a build. Only build (or no subcommand) now routes through the build-output filter; anything else runs unfiltered passthrough.

Fixes #2840 

## Test plan
I added tests for the commands and ran it manually to check it works

- [X] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [X] Manual testing: `rtk <command>` output inspected

### DIFF
--- a/src/cmds/js/next_cmd.rs
+++ b/src/cmds/js/next_cmd.rs
@@ -6,9 +6,29 @@ use crate::core::utils::{resolved_command, strip_ansi, tool_exists, truncate};
 use anyhow::Result;
 use regex::Regex;
 
+/// This filter only understands `next build` output. Any other subcommand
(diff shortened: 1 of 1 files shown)


ErezAmihud:
Not the "correct" solution to the problem. The problem lies in the npx rewrite rules. see #2854

PR #2852, opened 2026-07-06
fix(find): round dir-display truncation up to a char boundary (fixes #2851)

### LINKED ISSUE #2851: rtk find panics on non-ASCII path (byte index not a char boundary, find_cmd.rs:326) when an unknown flag like -print is present
## Summary

`rtk find` panics with `byte index is not a char boundary` when BOTH conditions are met:
1. the command line contains a flag rtk doesn't know (e.g. `-print`, which rtk warns about and ignores), and
2. a matching file lives under a directory whose name contains a multi-byte UTF-8 character (e.g. `é`).

Without the unknown flag, the same search works fine, so the buggy slice seems to be on the unknown-flag recovery path.

## Environment

- rtk 0.43.0 (installed at ~/.local/bin/rtk)
- macOS (Darwin 25.5.0, Apple Silicon)
- invoked through the Claude Code hook (transparent `find` → `rtk find` rewrite), but reproduces in a plain shell

## Minimal reproduction

```sh
mkdir -p 'Les Plongées de Juliette/5-Pipeline/Versions-imprimables'
touch 'Les Plongées de Juliette/5-Pipeline/Versions-imprimables/LES-PLONGEES_Index-Tableau-de-Bord.pdf'
rtk find . -name '*Index-Tableau-de-Bord.pdf' -print
```

Output:

```
rtk find: unknown flag '-print', ignored

thread 'main' (312487) panicked at src/cmds/system/find_cmd.rs:326:34:
start byte index 10 is not a char boundary; it is inside 'é' (bytes 9..11 of string)
```

Removing `-print` makes the same command succeed and print the match.

## Notes

- "Les Plong" is exactly 9 bytes, so the é sits at bytes 9..11 and a `[..10]` byte slice lands mid-character — the truncation at `find_cmd.rs:326` should use a char-boundary-safe cut (`floor_char_boundary`-style loop or `char_indices`).
- Real-world impact: Claude Code rewrites `find` to `rtk find` transparently, and `-print` is a very common (and POSIX-default) find flag, so any vault/repo with accented directory names (French, German, Spanish…) hits this.
- Looks like a sibling of the existing non-ASCII slicing panics #2318 (gain --history, CJK) and #2509 (output filter) — same class of bug in a different code path; a shared safe-truncation helper might fix all of them.

Revlochon:
## Problem

`rtk find` panics when the grouped-results display truncates a directory path longer than 50 bytes and the raw byte slice `&dir[dir.len() - 47..]` lands inside a multi-byte UTF-8 character:

```
thread 'main' panicked at src/cmds/system/find_cmd.rs:326:34:
start byte index 10 is not a char boundary; it is inside 'é' (bytes 9..11 of string)
```

Reported in #2851. Note the issue title mentions the unknown-flag path, but while preparing this fix I narrowed it down: the flag is irrelevant — the actual trigger is any grouped display of a matched directory path > 50 bytes with a multi-byte character straddling the cut point (`"Les Plongées de Juliette/5-Pipeline/Versions-imprimables"` is 57 bytes, so the cut lands at byte 10, inside the `é`). I've noted this on the issue.

## Fix

- Extract the truncation into a small `dir_display()` helper and round the cut up with `str::ceil_char_boundary()`, the same pattern already used in `pipe_cmd.rs`, `json_cmd.rs`, `gain.rs` and `session_cmd.rs` for the sibling bugs (#2318 / #2509 family).
- Add regression tests (multi-byte path at the exact panic geometry + short-path passthrough).

## Validation

- `cargo test`: 2417 passed.
- Real-world repro from #2851 no longer panics:
  ```sh
  mkdir -p 'Les Plongées de Juliette/5-Pipeline/Versions-imprimables'
  touch 'Les Plongées de Juliette/5-Pipeline/Versions-imprimables/X_Index-Tableau-de-Bord.pdf'
  rtk find . -name '*Index-Tableau-de-Bord.pdf' -print
  ```

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -26,6 +26,18 @@ fn glob_match_inner(pat: &[u8], name: &[u8]) -> bool {
     }
 }
 
+/// Truncate a long directory path for display, keeping the last ~47 bytes.
+/// The cut is rounded up to a char boundary so multi-byte UTF-8 (e.g. 'é')
+/// at the cut point doesn't panic (issue #2851).
+fn dir_display(dir: &str) -> String {
+    if dir.len() > 50 {
+        let cut = dir.ceil_char_boundary(dir.len() - 47);
+        format!("...{}", &dir[cut..])
+    } else {
+        dir.to_string()
+    }
+}
+
 /// Parsed arguments from either native find or RTK find syntax.
 #[derive(Debug)]
 struct FindArgs {
@@ -322,11 +334,7 @@ pub fn run(
         }
 
         let files_in_dir = &by_dir[dir];
-        let dir_display = if dir.len() > 50 {
-            format!("...{}", &dir[dir.len() - 47..])
-        } else {
-            dir.clone()
-        };
+        let dir_display = dir_display(dir);
 
         let remaining_budget = max_results - displayed;
         if files_in_dir.len() <= remaining_budget {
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for tracking this down to the grouped directory display rather than the unknown-flag path. The find rewrite in #3603 (merged 2026-08-26) replaced that byte slice with a chars-based truncation and added a long_unicode_dir_label_does_not_panic test, so the repro from #2851 now runs clean on develop. The helper extraction here is the one piece develop does not carry, but the truncation itself is already char-safe. If you see a case that still panics, please comment and we can reopen.

PR #2855, opened 2026-07-06
feat(shell): add rtk-shell, bash-wrapped shell for agent harnesses

guyzmo:
## Summary

- Add `rtk-shell`, a minimal shell wrapper that filters recognized CLI commands (`git status`, `cargo test`/`build`/`clippy`, etc.) through RTK's existing filters to cut token overhead for agent harnesses (e.g. opencode), while transparently forwarding anything it doesn't recognize to a real backing shell.
- Three invocation modes: one-shot (`rtk-shell -c "..."`, usable as `$SHELL`), persistent interactive session (with `cd`/env state persistence across commands), and an opt-in RC-file swap (`rtk init`) that auto-enables the shell when invoked by a known agent harness in non-interactive mode.
- Extracts `src/lib.rs` so the module tree is shared between the `rtk` and `rtk-shell` binaries; adds `src/shell/` (dispatch, oneshot, session) plus supporting changes to config, tracking, and hooks.

## Design notes / known limitations

- No PTY support — TUI/interactive-curses commands are forwarded to the backing shell unfiltered, not intercepted.
- Windows is out of scope for this iteration.
- Only `git status`/`cargo test`/`build`/`check`/`clippy` get real filtering in session mode today; other recognized commands intentionally fall back to raw passthrough (documented, not silent) pending follow-up work.
- `OPENCODE_SESSION_ID` in the known-harness env var list is an unconfirmed placeholder — flagged for review.

## Test plan

- [x] `cargo clippy --all-targets` (zero warnings)
- [x] `cargo test --all` (2408 passed, 9 ignored, 0 failed)
- [x] Manual verification: session-mode output byte-for-byte matches one-shot mode for `git status` (git repo + non-git dir)
- [x] Manual verification: `cd`/`pwd` state persists across commands in a session
- [x] Manual verification: pipeline passthrough (`echo hello | grep hello`) still works
- [x] Manual verification: no sentinel/internal marker leakage in session transcripts
- [ ] Reviewer: confirm `OPENCODE_SESSION_ID` env var name against opencode's actual harness detection surface


FlorianBruniaux:
@guyzmo Nice work on the session internals here, the nonce-based sentinels, dual-stream draining threads, and the sentinel-leak regression tests show real care on the trickiest part of this (process concurrency).

One blocker before we get into the actual review: this PR targets `master` instead of `develop`. GitHub already flagged it (`wrong-base` label, merge state `BLOCKED`). Per repo convention, features and fixes go through `develop`, `master` is release-only.

Can you retarget the base branch to `develop`? Once that's sorted we'll dig into the rest.

guyzmo:
@FlorianBruniaux I already rebased last night when I saw this.

I'm sorry, I'm not used for projects in git flow, especially using master for production. This is surprising ;-)

Maybe you should add something in agents.md to make the llm yell when a contributor does a PR against master? 😝

guyzmo:
hey @FlorianBruniaux now that it's on the right branch, would you mind removing the `wrong-base` label? (and maybe make an automation when the issue is fixed? ☺

PR #2860, opened 2026-07-07
fix(npm,lint): remove hardcoded allowlists, forward args untouched

### LINKED ISSUE #2094: bug: hook rewrite of `pnpm lint` discards package.json script body — same false-green as #185 at the hook layer

### LINKED ISSUE #2663: Bug: `npm query`, `npm sbom`, and 13 other subcommands get incorrect `run` injection

guyoron1:
Heyaaaa all :)

Fixes #2663, fixes #2094

Following @aeppling's feedback on #2664 — instead of patching the allowlists, this removes them entirely.

### npm

Deletes `NPM_SUBCOMMANDS` (the 70-entry list). `rtk npm` now forwards args untouched to npm — no more `run` injection. If `npm build` isn't a real subcommand, npm errors and the agent self-corrects to `npm run build`. Future npm commands (`query`, `sbom`, whatever they add next) just work without RTK updates.

### lint

Removes `lint` (the script name) from the lint rewrite rule. `pnpm lint` / `npm run lint` are no longer rewritten to `rtk lint`, so the package.json script's flags, chains, and exit code are preserved. Direct linter calls (`eslint`, `biome`, `npx eslint`, etc.) still route through `rtk lint` as before.

| Command | Before | After |
|---------|--------|-------|
| `rtk npm query` | `npm run query` (broken) | `npm query` (works) |
| `pnpm lint` | `rtk lint` (flags lost) | runs natively (correct) |
| `eslint src/` | `rtk lint` | `rtk lint` (unchanged) |

Net **-104 lines** — more deleted than added.

## Test plan

- [x] `cargo fmt && cargo clippy --all-targets && cargo test --all` — 2361 passed
- [x] Added `test_classify_lint_scripts_not_rewritten` + `test_rewrite_lint_scripts_not_rewritten_to_rtk_lint`


aeppling:
Hey @guyoron1 thanks for addressing this more durable fix, will avoid further maintenance.


### npm run inject part -> valid

### lint part -> lost rewrite

Removing lint from the rule fixes the substitution, but it deletes the rewrite entirely, not just the substitution. pnpm lint, npm lint, bare lint -> match no rule -> run raw

The npm part shows the pattern to follow: there the rewrite and the substitution are separate, so dropping the injection kept the rewrite + filter. On lint they're fused, the rewrite target rtk lint is the substitution.

#### proposed fix

Mirror your npm half: keep the rewrite, drop only the substitution.

1 - npm run lint / pnpm run lint: already fine after this PR,nothing to do.

2 - Bare pnpm lint: the only routing gap. Extend the generic pnpm rule to wrap bare invocations → rtk pnpm lint. Important: wrap any bare pnpm <word>, not just lint, enumerating script names would recreate the exact kind of list this PR removes. 

3 - Compression recovery (follow-up) enhance the npm/pnpm output filter to detect lint-style diagnostics (file:line:col + rule id) and compress around them: collapse clean/progress lines, keep diagnostics, tee tail hint the tail.
Pure output-side, no lists, and it degrades safely: worst case is less compression, never a changed command or exit code.


Thanks for your work

guyoron1:
Heyaaaa :) thanks for the detailed feedback!

**lint routing gap** — Fixed: the pnpm discover rule now matches `pnpm <any-word>` instead of only known subcommands (`exec|install|list|...`). So `pnpm lint`, `pnpm build`, `pnpm dev` — any bare pnpm script — routes through `rtk pnpm <script>` for output filtering, without enumerating script names.

Specific tool rules (eslint, biome, vitest, tsc, etc.) still take priority via RegexSet last-match semantics, so `pnpm eslint` → `rtk lint` is unchanged.

Added tests for both the generic routing and the specific-tool override behavior.

Compression recovery (point 3) makes sense as a follow-up.

guyoron1:
@aeppling rebased on current develop and reworked the pnpm half; the description is updated to match. Short version:

- **npm:** `NPM_SUBCOMMANDS` had grown since (#3087 added 15 entries plus a test that asserts against the list). Both go away with the list. With a stand-in npm, develop still injects `run` for 16 spellings (the short aliases and the npm 11 commands); this PR injects it for none. `tests/npm_args_passthrough_test.rs` asserts npm gets the user's argv byte-for-byte and fails on develop with `left: ["run", "add", "lodash"]`, which is #2663 itself.
- **pnpm:** your "wrap any bare `pnpm <word>`" is in, as `^pnpm\s+[^-\s]\S*` (a word, never a flag). #3275 landed last week, and its flag-first guards (`pnpm --filter @app`, `pnpm -x install`) would otherwise have started rewriting and died at clap.

One thing I'd rather flag than have it surprise you: three of #3275's tests flip on purpose, because their premise ("bare `pnpm build` must not route") is the opposite of the review comment here. `pnpm build` and `pnpm -r lint` now go through `rtk pnpm`, which runs the real script through the passthrough and keeps its exit code. Each test has a comment saying why. If you'd rather keep #3275's position, the alternative is enumerating script names, which is the list this PR removes.

Overlaps I noticed: #2776 (which you called a temporary fix; if it lands first this hard-conflicts in `rules.rs` and #3150 loses the `pnpm biome` forms), #3289 (its lint half is the same idea with a smaller scope) and #3676 (enumerates `test/build/typecheck/format`; the hang concern doesn't apply here, since `PnpmCommands::Other` goes through the streaming passthrough rather than capture).

If these all land, the order that avoids conflicts is #2655, then this, then #3150 (the one shared spot is the `test_rewrite_lint` list, which resolves to the eslint-only list). #2664 is superseded by this PR.

Issue #2271, opened 2026-06-04
Some filters drop or misrepresent requested data (find, ls -R, env, cargo label)


sysrow:
While using RTK's filtered output, a few commands drop or misrepresent the data that was actually requested. None of these corrupt data, but each can hide requested results or mislead a reader.

### 1. `find` truncates a directory mid-listing
The `--max` budget is counted across directories, so the last shown directory can be cut after only some of its files while still rendered as a normal `dir/ files` line. The listing then looks complete and the trailing `+N more` is attributed to other directories. Someone checking whether a file exists in that directory is misled into thinking the directory was fully listed.

### 2. `ls -R` drops directory headers
`ls -R` prints one `path:` section per directory. The compactor flattens every section into a single list and drops the headers, so files can no longer be mapped to their directory (many distinct `mod.rs` entries become indistinguishable). The output can even be larger than the raw listing.

### 3. `env` silently drops secret-bearing variables
A sensitive variable that does not match a known category (cloud/tool/language) falls through every bucket and is omitted entirely, so `rtk env` implies it is unset — even when the user wants to confirm it is set. (Categorized secrets such as `AWS_SECRET_ACCESS_KEY` are already shown masked; the gap is the uncategorized ones.)

### 4. `cargo check` is labeled `cargo build`
`CargoBuildHandler` is shared by build and check, but the summary string is hardcoded to `cargo build`, so `rtk cargo check` reports `cargo build (N crates compiled)`.

A PR with atomic commits (one fix per commit) follows.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 97% |

### Summary

Multiple filter handlers silently drop or misrepresent output (find truncation, ls -R header loss, env variable omission, cargo check mislabeled); fix PR #2272 already open.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

laurac8r:
[...]

PR #2868, opened 2026-07-07
fix(ls): preserve per-directory headers for multiple directory operands

### LINKED ISSUE #2857: rtk ls with multiple directory arguments omits per-directory headers
## Summary

`rtk ls dirA dirB` (and any N-directory invocation) flattens every directory's entries into one list with no per-directory attribution, unlike native `ls`, which prints a `path:` header before each directory's listing. A file can appear to belong to the first directory listed when it actually came from another.

This is distinct from #2271 / PR #2272, which fixes header loss for `ls -R` (single-directory recursive). The case here is triggered by passing multiple directory operands to non-recursive `ls`, so it isn't covered by that fix.

## Repro

```
$ mkdir -p dirA dirB && echo a1 > dirA/file_a.txt && echo b1 > dirB/file_b.txt

$ ls dirA dirB
dirA:
file_a.txt

dirB:
file_b.txt

$ rtk ls dirA dirB
file_a.txt  3B
file_b.txt  3B
```

Expected: output preserves the per-directory `path:` headers (or equivalent attribution) that native `ls` prints whenever more than one directory operand is given.

Actual: entries from all directories are merged into a single flat list with no indication of which directory each file came from.

## Impact

Deterministic whenever more than one directory is passed and at least one is non-empty. Misleads any check that depends on directory membership (e.g. "did file X land in stage A or stage B of a pipeline") into attributing a file to the wrong directory, since the merged list reads as if everything came from the first path.

## Environment

`rtk 0.43.0`, Windows (Git Bash).

## Workaround

`rtk proxy ls dirA dirB` (unfiltered passthrough) or one directory at a time via `rtk proxy ls <dir>`.

Sammy-Dabbas:
`rtk ls dirA dirB` flattened every directory's entries into one list with no per directory attribution, so a file could appear to belong to the wrong directory. Native `ls` prints a `dirA:` header before each directory's listing. The raw `ls -la` output rtk gets does contain the headers; `compact_ls` dropped them because every line goes through the date anchored `parse_ls_line`, which returns `None` for header lines.

Closes #2857

## Summary

- Split raw `ls` output into `path:` sections and render each directory's entries under its header (blank line between groups), only when native ls printed headers
- Single-directory output is unchanged, pinned by an exact string test. mixed operands list loose files first, then each directory under its header; `-R` headers are fixed by the same mechanism
- Related: #2272 has a similar section-splitting approach for the `-R` case of #2271 (this mirrors its helper conventions); it has been conflicting for a month, so this is a standalone fix for the multi-operand case. Happy to rebase if #2272 merges first.

## Test plan

Reproduced the flat output on develop, confirmed the fixed output matches native ls shape, and added 3 multi-directory regression tests (32/32 ls tests pass). Full suite green apart from 2 pre-existing `hooks::rewrite_cmd` environment failures, verified identical on pristine develop.

### DIFF
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -225,42 +225,57 @@ fn perms_to_octal(perms: &str) -> Option<String> {
     }
 }
 
-/// Parse ls -la output into compact format.
+/// Returns the directory path from a `path:` section header (e.g. `src/cmds:`
+/// -> `src/cmds`), or `None` if the line is not a section header.
 ///
-/// Without `show_long`:
-///   name/        (dirs)
-///   name  size   (files)
-///
-/// With `show_long` (user passed `-l`):
-///   755  name/        (dirs)
-///   644  name  size   (files)
-///
-/// Returns (entries, summary, parsed_count) so caller can suppress summary when piped.
-/// parsed_count tracks how many non-header lines were successfully parsed.
-/// If parsed_count == 0 but raw had content, caller should fall back to raw output.
-fn compact_ls(raw: &str, show_all: bool, show_long: bool) -> (String, String, usize) {
-    use std::collections::HashMap;
+/// Native ls prints a `path:` line before each directory's listing whenever
+/// more than one directory operand is given, and before every directory in a
+/// recursive (`-R`) listing. A header is a non-empty line ending in ':' that
+/// is neither a `total` line nor a parseable `ls -la` entry, so a file
+/// literally named `foo:` is never mistaken for one.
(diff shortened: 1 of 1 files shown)


KuSh:
Closing in favour of #2865, which closes the same issue #2857 with the same per directory header behaviour and applies to the compact_ls that is on develop today. This branch is built on the older three value compact_ls, so merging it would take back out the hidden entry disclosure that has since landed. Review and comments are very welcome on #2865, or reply here if your section splitting covers something it misses.

PR #2877, opened 2026-07-08
fix(cargo): preserve compiler warnings in cargo test output on passing runs

CryptoKrad:
## Summary

- `rtk cargo test` silently dropped compile-phase `warning:` blocks on passing runs — an agent reading `cargo test: N passed` concluded the build was clean while rustc had emitted warnings. Since exit-0 runs get no tee log, the information was unrecoverable (same class as the `never_worse` size-only guard: the lie is smaller, so it passes).
- Fix mirrors `CargoBuildHandler`'s warning capture into both paths: `CargoTestHandler` (streaming) now streams warning blocks live, and `filter_cargo_test` (buffered/pipe) collects them, capped at `CAP_WARNINGS` with an `+N more` marker. The compact summary gains `[N compiler warnings]`; the aggregate `generated N warnings` line stays stripped.
- Buffered path hard-terminates a warning block on section markers (`failures:` / `test result:` / `test ` / `---- ` / `error`) so a short warning block can never swallow a following section.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — fmt clean, clippy 0 issues, 2394 passed / 0 failed (+ integration suites green)
- [x] 2 new regression tests: `test_filter_cargo_test_all_pass_preserves_compiler_warnings` (buffered) and `test_streamed_cargo_test_handler_preserves_compiler_warnings` (streaming) — assert warning survival, `[N compiler warnings]` annotation, and aggregate-line stripping
- [x] Manual testing on a crate with a dead-code warning: passing run now shows the warning block + `cargo test: 3 passed (2 suites, 0.00s) [1 compiler warnings]` (exit 0); failing run still shows failure detail with exit 101 == raw; `rtk cargo build` output unchanged (consistency restored between build and test)

https://claude.ai/code/session_01K5bb4GKf9CNKP9uYhFDvsz

### DIFF
--- a/src/cmds/rust/cargo_cmd.rs
+++ b/src/cmds/rust/cargo_cmd.rs
@@ -126,6 +126,8 @@ struct CargoTestHandler {
     in_failure_names: bool,
     summary_lines: Vec<String>,
     has_compile_errors: bool,
+    warnings: usize,
+    in_warning_block: bool,
 }
 
 impl CargoTestHandler {
@@ -135,6 +137,8 @@ impl CargoTestHandler {
             in_failure_names: false,
             summary_lines: Vec::new(),
             has_compile_errors: false,
+            warnings: 0,
+            in_warning_block: false,
         }
     }
 }
@@ -153,6 +157,11 @@ impl BlockHandler for CargoTestHandler {
             return true;
         }
         if line.starts_with("test ") && line.ends_with("... ok") {
+            return true;
+        }
+        // Aggregate count line ("warning: `x` generated N warnings"); individual
+        // warning blocks are already streamed via is_block_start below.
+        if line.starts_with("warning:") && line.contains("generated") && line.contains("warning") {
             return true;
         }
         // Track compile errors for fallback
@@ -187,10 +196,26 @@ impl BlockHandler for CargoTestHandler {
     }
 
     fn is_block_start(&mut self, line: &str) -> bool {
-        self.in_failure_section && line.starts_with("---- ")
+        if self.in_failure_section && line.starts_with("---- ") {
+            self.in_warning_block = false;
+            return true;
+        }
+        // Compile-phase warnings on a passing run were previously dropped
+        // (exit 0 → no tee → unrecoverable); stream them like CargoBuildHandler.
+        if !self.in_failure_section
+            && (line.starts_with("warning:") || line.starts_with("warning["))
+        {
+            self.warnings += 1;
+            self.in_warning_block = true;
+            return true;
+        }
+        false
     }
 
-    fn is_block_continuation(&mut self, line: &str, _block: &[String]) -> bool {
+    fn is_block_continuation(&mut self, line: &str, block: &[String]) -> bool {
+        if self.in_warning_block {
+            return !(line.trim().is_empty() && block.len() > 3);
+        }
         self.in_failure_section && !line.starts_with("---- ")
     }
 
@@ -234,6 +259,13 @@ impl BlockHandler for CargoTestHandler {
         if all_parsed {
             if let Some(agg) = aggregated {
                 if agg.suites > 0 {
+                    if self.warnings > 0 {
+                        return Some(format!(
+                            "{} [{} compiler warnings]\n",
+                            agg.format_compact(),
+                            self.warnings
+                        ));
+                    }
                     return Some(format!("{}\n", agg.format_compact()));
                 }
             }
@@ -1127,6 +1159,9 @@ pub(crate) fn filter_cargo_test(output: &str) -> String {
     let mut summary_lines: Vec<String> = Vec::new();
     let mut in_failure_section = false;
     let mut current_failure = Vec::new();
+    let mut warnings: Vec<String> = Vec::new();
+    let mut current_warning: Vec<String> = Vec::new();
+    let mut in_warning_block = false;
 
     for line in output.lines() {
         // Skip compilation lines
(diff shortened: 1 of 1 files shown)


src/cmds/rust/cargo_cmd.rs
```diff
@@ -1172,8 +1249,24 @@ pub(crate) fn filter_cargo_test(output: &str) -> String {
...
+            warnings_section.push('\n');
+        }
+        if warnings.len() > CAP_WARNINGS {
+            warnings_section
+                .push_str(&format!("… +{} more warnings\n", warnings.len() - CAP_WARNINGS));
+        }
+    }
 
     let mut result = String::new();
+    result.push_str(&warnings_section);
```
KuSh, line 1269: **Blocking: compile errors are silently dropped whenever a warning was captured.**

Seeding `result` with the warnings section here, before any body content, makes the `result.trim().is_empty()` guard further down unreachable once a single warning block exists.

Concrete failure — a crate that both warns and fails to compile:

```
warning: unused variable: `x`
 --> src/lib.rs:3:9
...
error[E0425]: cannot find value `y` in this scope
error: could not compile `failproj` (lib test) due to 1 previous error
```

There are no `test result:` lines and no test failures, so the only thing in `result` is the warnings section. That's non-empty, so the guard is false, and both the `filter_cargo_build_labeled` compile-error rendering and the last-5-meaningful-lines fallback are skipped. The output is the warning block and **zero errors**, on a non-zero exit. On `develop` the same input renders `cargo test: 1 errors` plus the error blocks, so this is a regression.

`test_filter_cargo_test_compile_error_preserves_error_header` doesn't catch it because its fixture has no warning in it.

Suggested fix: build the body into its own string and test *that* for emptiness, prepending `warnings_section` only at return. One thing to watch when you do — the tail fallback will then start running in cases it previously didn't, so it needs to exclude lines the warnings section already printed, or `cargo test --no-run` on a warning-only crate will print them twice.

src/cmds/rust/cargo_cmd.rs
```diff
@@ -187,10 +196,26 @@ impl BlockHandler for CargoTestHandler {
...
 
     fn is_block_start(&mut self, line: &str) -> bool {
-        self.in_failure_section && line.starts_with("---- ")
+        if self.in_failure_section && line.starts_with("---- ") {
+            self.in_warning_block = false;
+            return true;
+        }
+        // Compile-phase warnings on a passing run were previously dropped
+        // (exit 0 → no tee → unrecoverable); stream them like CargoBuildHandler.
+        if !self.in_failure_section
```
KuSh, line 205: **Blocking: every warning is reported twice on the streaming compile-error path.**

Emitting warning blocks live here is fine on its own, but `format_summary`'s compile-error branch still hands the whole raw output to `filter_cargo_build_labeled`, and `CargoBuildHandler` collects warning blocks as well as error blocks. `rtk cargo test` goes through `run_cargo_streamed`, so on a crate with a warning plus a compile error the warning block is streamed once from here and then rendered a second time inside the summary.

The fix needs to suppress warning-block rendering for that one call site only — `build`, `check` and `clippy` share that helper and should keep rendering their warnings as they do today.

src/cmds/rust/cargo_cmd.rs
```diff
@@ -155,6 +159,11 @@ impl BlockHandler for CargoTestHandler {
         if line.starts_with("test ") && line.ends_with("... ok") {
             return true;
         }
+        // Aggregate count line ("warning: `x` generated N warnings"); individual
+        // warning blocks are already streamed via is_block_start below.
+        if line.starts_with("warning:") && line.contains("generated") && line.contains("warning") {
```
KuSh, line 164: **This heuristic swallows genuine warnings.**

`line.starts_with("warning:") && line.contains("generated") && line.contains("warning")` — the third clause is tautological given the first, so this reduces to `starts_with("warning:") && contains("generated")`. Any real rustc or clippy warning whose message happens to contain "generated" matches, for example in a prost/tonic crate:

```
warning: unused import: `crate::generated::pb`
 --> src/lib.rs:4:5
```

`should_skip` returns true, so `is_block_start` never fires, `self.warnings` isn't incremented, and the trailing ` --> ` / ` | ` / ` = note:` lines match nothing and are dropped. The warning disappears entirely and `[N compiler warnings]` under-counts — the exact failure this PR is meant to fix.

Anchoring on the real shape of the aggregate line would avoid it, e.g. matching a `generated \d+ warnings?$` suffix rather than a substring.

The same expression appears in the buffered path (`filter_cargo_test`), where the `continue` leaves `in_warning_block` false and the block body falls through to the failure detector and is discarded. Worth fixing in both places. The buffered copy also isn't gated on `!in_failure_section`, so a failing test whose captured stdout contains such a line loses it from the failure block.

src/cmds/rust/cargo_cmd.rs
```diff
@@ -187,10 +196,26 @@ impl BlockHandler for CargoTestHandler {
...
+            self.in_warning_block = true;
+            return true;
+        }
+        false
     }
 
-    fn is_block_continuation(&mut self, line: &str, _block: &[String]) -> bool {
+    fn is_block_continuation(&mut self, line: &str, block: &[String]) -> bool {
+        if self.in_warning_block {
+            return !(line.trim().is_empty() && block.len() > 3);
```
KuSh, line 217: **`block.len() > 3` means short warning blocks never terminate.**

The blank-line terminator requires more than 3 lines already buffered, so span-less warnings never close on their trailing blank:

```
warning: unused manifest key: package.metadata.foo
warning: profiles for the non root package will be ignored
```

`     Running unittests ...` and `   Doc-tests ...` aren't in `should_skip` (it covers `Compiling`/`Downloading`/`Downloaded`/`Finished`, plus the lowercase `running ` harness line), so they're treated as continuations and glued into the warning block.

The two paths then fail differently: the buffered filter emits that noise inside the warnings section, while the streaming handler holds the block open past `test result:` (consumed by `should_skip`) so it only surfaces at `flush()` — out of order, after everything else.

Terminating on the first blank line regardless of block length, or keying off the diagnostic's indentation, would handle both.

src/cmds/rust/cargo_cmd.rs
```diff
@@ -1172,8 +1249,24 @@ pub(crate) fn filter_cargo_test(output: &str) -> String {
     if !current_failure.is_empty() {
         failures.push(current_failure.join("\n"));
     }
+    if !current_warning.is_empty() {
+        warnings.push(current_warning.join("\n"));
+    }
+
+    let mut warnings_section = String::new();
+    if !warnings.is_empty() {
+        for block in warnings.iter().take(CAP_WARNINGS) {
```
KuSh, line 1258: **The cap is only applied on the path that isn't used for `cargo test`.**

This caps warning blocks at `CAP_WARNINGS` with a `… +N more warnings` marker, but the streaming handler emits every warning block with no cap at all — and streaming is what `rtk cargo test` actually runs.

On a crate with 200 dead-code warnings, a *passing* `rtk cargo test` now emits all 200 blocks, which is a significant output regression on the most frequently run command, while the same raw output through the buffered path shows 10. Same input, two materially different outputs.

The cap should be enforced on the streaming side too, so both paths agree.

KuSh (changes requested):
Thanks for tackling this — dropping compile warnings on passing runs is a real gap, and preserving them on both the streaming and buffered paths is the right shape for the fix.

I found five issues that need addressing before this can go in. The first two are behaviour regressions on the compile-error path rather than gaps in the new feature, so they're the blocking ones; the last three affect the warning preservation itself.

A general note: the streaming handler (`CargoTestHandler`) and the buffered filter (`filter_cargo_test`) now implement warning capture twice, with slightly different logic. Whatever the fixes end up being, it's worth checking each one lands on both paths — three of the five below are cases where the two implementations disagree.

Separately, and unrelated to the code: the CLA is still showing as unsigned for this PR, so it can't be merged until that's done. You can sign it here: https://cla-assistant.io/rtk-ai/rtk?pullRequest=2877

PR #2881, opened 2026-07-08
feat(powershell): add rtk powershell command family for Windows cmdlets

phatphamhong-tech:
## Summary
- Adds `rtk powershell <Cmdlet> [args...]`, filtering output for 16 commonly-used
  PowerShell cmdlets (Get-ChildItem, Get-Process, Get-Service, Get-Content,
  Select-String, Get-WinEvent, Get-EventLog, Get-ItemProperty, Test-NetConnection,
  Get-NetAdapter, Get-NetTCPConnection, Get-ComputerInfo, Get-Package, Get-Counter,
  Install-Package, Get-Item), same header+capped-rows / essential-properties
  approach as the rest of rtk. Cmdlets with already-minimal or side-effecting
  output pass straight through unfiltered, like `rtk proxy`.
- Untrusted input never reaches the PowerShell script as text: the unknown-cmdlet
  passthrough path allowlists cmdlet names against the shape of a real
  cmdlet/function/alias, and every argument value is passed to the child process
  via an environment variable (`$env:RTK_PS_ARG_<n>`) instead of being
  interpolated into the `-Command` string.
- Registers a per-cmdlet hook rewrite pattern in `src/discover/rules.rs` (short
  aliases that collide with existing rules like `ls`/`cat`/`ps`/`grep` are
  deliberately not registered), and adds `force_tee_hint`/`force_tee_tail_hint`
  truncation recovery to every capped list.

## Test plan
- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — all green (2442 passed)
- [x] Manual testing: verified `rtk powershell <Cmdlet>` output against real PowerShell 5.1, and confirmed the hook rewrite (`Get-Process` → `rtk powershell Get-Process`) works without colliding with existing `ls`/`cat`/`ps`/`grep` rules


pszymkowiak:
Really impressive PR — I reviewed it in depth and it's one of the most rule-compliant contributions I've seen here: real captured fixtures with `assert!(savings >= 60.0)` per cmdlet, complete registration (`trailing_var_arg`/`allow_hyphen_values`, rules, docs incl. a module README), no `unwrap()` in the production path, and delegation to the shared runner for fallback + exit codes. The security design is genuinely well thought out — routing data **values** through `$env:RTK_ARG_i` (never interpolated into the script) while only allowlisted `SAFE_FLAG_TOKEN` flags reach the script text is exactly the right way to keep this injection-safe; the comments show a clear threat model. 46 filter tests pass locally, clippy clean.

I also measured the real savings against the `cl100k_base` BPE tokenizer on your fixtures (not just the whitespace token count): **68–90% across cmdlets** (Get-ComputerInfo ~90%, Select-String ~83%, Get-NetAdapter ~75%, Get-NetTCPConnection ~72%, Get-ChildItem ~68%). Notably rtk's whitespace count matches BPE closely here — because the savings come from dropping whole verbose property rows (real content), not cheap structural decoration, so these numbers are genuinely real. Nice work.

Two things before this can land:

**1. The CLA isn't signed yet** — the `license/cla` check is pending. Could you sign it? Nothing can merge until then, and it's also gating the CI test suite from running.

**2. This needs an end-to-end run on real Windows by someone other than the author.** The filter functions are unit-tested on fixtures (platform-independent), and you verified output against real PowerShell 5.1 — thank you, that's more than most Windows PRs get — but CI can't exercise `powershell.exe`, and I could only run the fixture tests on macOS where the actual invocation path never runs. **Could a Windows user pull this branch and confirm:**

```powershell
cargo build --release
.\target\release\rtk powershell Get-ChildItem          # compact output vs raw
.\target\release\rtk powershell Get-NetTCPConnection   # and a couple others
# and a quick injection sanity check on the env-var mechanism:
.\target\release\rtk powershell Get-ChildItem "; Remove-Item X"   # the value must stay inert, not execute
```

That last line is the one I'd most like to see confirmed — the whole safety of the module rests on `$env:` expansion not re-tokenizing, and it's the one thing no test currently proves.

The code, compliance, and measured savings are excellent; this is purely about the two things CI/dev machines can't cover on their own.

pszymkowiak:
Follow-up on the injection concern from my earlier comment — I was able to close that gap myself, without waiting for a Windows machine.

`powershell_binary()` prefers `pwsh` on any platform, so I installed PowerShell 7.6.4 on macOS, built this branch, and ran it end-to-end. `rtk powershell Get-ChildItem` works: real `pwsh` is invoked, the output is filtered, exit code propagates.

Then I verified the `$env:RTK_PS_ARG_i` mechanism against real injection payloads with a canary file that must not be deleted:

| Argument value passed to `rtk powershell Get-ChildItem` | Result | Canary |
|---|---|---|
| `; Remove-Item ./canary.txt` | `Cannot find path '.../; Remove-Item ./canary.txt'` | survives ✅ |
| `$(Remove-Item ./canary.txt)` | subexpression not evaluated, treated as a literal path | survives ✅ |
| `` x`; Remove-Item ./canary.txt `` (backtick) | literal path | survives ✅ |

The error messages confirm the design: PowerShell received each entire malicious string as a **single literal path argument** to `Get-ChildItem` via `$env:` expansion — the `;`, `$(...)`, and backtick were all inert, and the canary file was untouched.

And since `$env:` expansion not re-tokenizing is a PowerShell **language** property (not OS-specific), this holds on Windows PowerShell 5.1 too — the arg-passing/injection surface is identical across editions; only *which cmdlets exist* differs by platform.

So the security linchpin is now empirically verified, not just correct-on-paper. The remaining item is just the CLA signature.

phatphamhong-tech:
Hello, the CLA has been signed. Thank you.

PR #2887, opened 2026-07-08
fix(hooks): honor exclude_commands for head/tail line-range fast path

### LINKED ISSUE #2823: `[hooks] exclude_commands` is bypassed for `head`/`tail` (line-range fast path skips the exclusion check)

AZERDSQ131:
Closes #2823

`[hooks] exclude_commands` is bypassed for `head`/`tail` because the line-range fast path in `rewrite_segment_inner` returns before the exclusion check:

```rust
}
```

`is_excluded` is only consulted in the `Classification::Supported` and `Unsupported` arms further down, both downstream of this early return.

Repro before this fix (config: `[hooks] exclude_commands = ["tail", "head", "rg"]`):

```console
rtk read sample.txt --tail-lines 3     # rewritten despite the exclusion
```

**Fix:** moved the same `is_excluded` check the `Supported` arm already uses (ENV_PREFIX-stripped command matched against the configured `ExcludePattern` list) in front of the fast path's early return, so excluded `head`/`tail` invocations fall through to the raw-passthrough path like every other excluded command.

Adds three regression tests in `src/discover/registry.rs`:
- `test_exclude_tail_line_range_fast_path` — excluded `tail` no longer rewrites
- `test_exclude_head_line_range_fast_path` — excluded `head` no longer rewrites
- `test_head_line_range_fast_path_still_rewrites_when_not_excluded` — guards against fixing this by disabling the fast path entirely

Ran the mandatory gate locally:
- `cargo fmt --all` — no changes needed
- `cargo clippy --all-targets` — clean
- `cargo test --all` — 2396 passed (+ 6/11/11/14/5/11 in the other suites), 0 failed

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -865,6 +865,11 @@ fn rewrite_segment_inner(
     }
 
     if cmd_part.starts_with("head -") || cmd_part.starts_with("tail ") {
+        let stripped = ENV_PREFIX.replace(cmd_part, "");
+        let cmd_clean = stripped.trim();
+        if is_excluded(cmd_clean, excluded) {
+            return None;
+        }
         return rewrite_line_range(cmd_part).map(|r| format!("{}{}", r, redirect_suffix));
     }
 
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as a duplicate of #3324, which takes the same approach you did — strip the env prefix, then match — and is currently mergeable.

This one is blocked on two counts: the CLA is still unsigned, and the branch has been in conflict since July. Since #3324 is the same fix in a landable state, it isn't worth asking you to clear both.

#3324 will carry `Fixes #2823`.

Issue #2839, opened 2026-07-05
[BUG] find with multiple path arguments silently searches only the first path


Bragging-Rights:
**OS:** macOS (Darwin 25.5.0, arm64)
**rtk version:** reproduced on 0.37.2 and 0.43.0 (homebrew bottle)

## Summary

`rtk find` with two or more path arguments silently searches only the FIRST path. Matches in paths 2+ are dropped with no warning, producing false zero-hit results.

## Repro

```
mkdir -p bugtest/a bugtest/b
echo x > bugtest/a/g.txt
echo x > bugtest/b/h.txt

# native find (expected behavior) — 2 results:
find bugtest/a bugtest/b -name "*.txt"
# bugtest/a/g.txt
# bugtest/b/h.txt

# rtk 0.43.0 — 1 result, second path silently ignored:
rtk find bugtest/a bugtest/b -name "*.txt"
# g.txt
```

## Why it matters

When rtk is used as a Claude Code PreToolUse rewrite hook, the agent trusts the filtered output. A silent miss on paths 2+ turns into a confident "file does not exist" claim by the model. We currently work around it by detecting multi-path `find` and skipping the rewrite.

## Expected

Either search all path arguments (matching find semantics), or fall back to native passthrough when >1 path argument is present.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 94% |

### Summary

rtk find silently ignores all but the first path argument, producing false-negative results that can mislead agents relying on rtk as a hook filter; should search all paths or passthrough natively when multiple paths are given

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

- `src/commands/find.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

PR #2894, opened 2026-07-08
fix(find): support multiple path arguments

davidalecrim1:
## Summary

- `rtk find` silently searched only the first path argument, dropping matches from any additional paths with no warning or error (#2839). This made it unsafe as a PreToolUse hook filter, since an agent could be told a file "does not exist" when it only existed under a dropped path.
- Both native (`find path1 path2 -name ...`) and RTK-shorthand (`find <pattern> path1 path2 ...`) syntaxes now collect all leading path arguments, and `run()` walks every given root through a single `WalkBuilder`.
- Display paths are stripped relative to the root when a single path is given (unchanged behavior); with multiple distinct roots, each result keeps its root prefix so files from different roots don't collide in the by-directory grouping.
- Identical path arguments are deduped before searching: a repeated root is walked once and reported once, rather than duplicating results, keeping output as compact and unambiguous as any other single-path search.

## Test plan

- [x] Manual testing: `rtk find pathA pathB -name "*.txt"` inspected, confirmed matches from both paths; confirmed single-path output format unchanged; confirmed a repeated path argument (`rtk find a a -name "*.txt"`) dedups to a single match
- [x] Added unit tests for multi-path parsing and subprocess integration tests reproducing the original issue #2839 repro, plus a dedup regression test

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -212,6 +212,16 @@ pub fn run(
 ) -> Result<()> {
     let timer = tracking::TimedExecution::start();
 
+    // Duplicate roots would be walked and reported twice for no benefit; dedup
+    // keeps the search and the output no larger than the set of distinct roots.
+    let mut deduped_paths: Vec<String> = Vec::with_capacity(paths.len());
+    for p in paths {
+        if !deduped_paths.contains(p) {
+            deduped_paths.push(p.clone());
+        }
+    }
+    let paths: &[String] = &deduped_paths;
+
     // paths is guaranteed non-empty by both parse_native_find_args and
     // parse_rtk_find_args (they default to ["."] via FindArgs::default()).
     let single_root = paths.len() == 1;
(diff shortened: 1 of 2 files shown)


KuSh:
Thanks for digging into the multi-path case. rtk find changed shape in #3603 (merged 2026-08-26): it dispatches on find's own grammar, and anything outside the single-path subset it walks itself, including `find pathA pathB -name ...`, is handed to real find with only its output compressed. On current develop `rtk find folder1 folder2 -name '*.txt'` returns matches from both roots, and #3854 (merged 2026-09-03) made a missing path report find's own error and exit 1, so the parsing and multi-root walk this PR extends are no longer there to extend. One thing develop does not have is an integration test pinning the multi-path result, and a small case in tests/guard_integration_test.rs would still be welcome. Worth noting that develop now prints a repeated path twice, as find does, rather than deduping it. If I have misread a case this PR covered, please comment or reopen.

PR #2903, opened 2026-07-09
fix: add is_unsupported_shape() guard before rewrite_compound()

### LINKED ISSUE #2792: Claude Code hook rewrites unsupported command shapes (compound find, multi-file cat) into failing rtk invocations
The `rtk hook claude` PreToolUse rewriter rewrites commands that rtk's own subcommands then refuse or mishandle:

1. **Compound find**: `find ... -not/-exec/-o ...` gets rewritten to `rtk find ...`, which exits with `rtk: rtk find does not support compound predicates or actions (e.g. -not, -exec). Use \`find\` directly.` — the hook should detect these shapes and skip the rewrite instead of producing a guaranteed failure + retry round-trip.
2. **Multi-file / flagged cat**: `cat file1 file2` (and `cat -n file`) gets rewritten to `rtk read`, which prints its help text instead of file contents.

Both observed repeatedly in real Claude Code sessions (June 2026). Suggested fix: the hook's rewrite table should validate that the target rtk subcommand supports the argument shape before rewriting, and fall through to the raw command otherwise.

Workaround in use: a wrapper script around `rtk hook claude` that passes these shapes through unrewritten.

https://claude.ai/code/session_01JphQtisSjrJampPDogyTBP

### LINKED ISSUE #2821: `rtk find` warns on unknown predicates but then executes a DIFFERENT (broader) query — silent wrong results

### LINKED ISSUE #2892: RTK hook: find returns empty stdout silently (directory exists with files) — hook should passthrough instead
## Summary

When `find` is invoked through the RTK hook (Claude Code PreToolUse rewrite), it can silently return **empty stdout** even when the directory exists and contains files. The hook collapses the output — the caller sees nothing and concludes there are no matches.

## Observed Behavior

- Directory `/path/to/target` existed with files
- `find /path/to/target ...` via RTK hook returned zero output (empty)
- Workaround: using an explicit `ls` loop gave correct results
- No error message, no warning — just silent empty output

## Root Cause (likely)

This is another instance of the hook finding a `find` shape it cannot properly rewrite and either:

a) **#2821** — an unknown predicate/flag gets silently dropped, the query executes broader with no matches, but the hook collapses the empty result as "ok"
b) **#2792** — compound predicates cause the hook to rewrite into an invocation that produces empty/no output

In both cases the failure is silent: no error, no warning on stderr, just an empty result that the caller treats as "no files found" — which is wrong for a destructive operation like cleanup (shadow removal).

## Suggested Fix

The hook should validate the `find` invocation shape **before** rewriting. If the shape includes:
- compound predicates (`-not`, `-exec`, `-o`, `!`)
- flags/predicates unknown to `rtk find`
- multi-path arguments
- any form that `rtk find` cannot faithfully reproduce

...then the hook should **passthrough to native `find`** instead of producing a silent wrong/empty result. A hard-fail with clear message (like `-not` already does) is acceptable; silent empty output is not.

## Related Issues

- #2792 — Compound find shapes rewritten into failing rtk invocations
- #2821 — Unknown predicates silently dropped, returning broader query
- #2839 — Multiple path arguments silently searched only the first
- #1282 — Silent output corruption when stdout is piped/redirected

## Workaround

Add `find` to `exclude_commands` in `~/.config/rtk/config.toml`:
```toml
[hooks]
exclude_commands = ["find"]
```
Or use explicit `ls`/loop as a per-case bypass.

### LINKED ISSUE #2902: Root cause: rewrite_command() lacks early command-shape validation (fixes #2792, #2821, #2892)
## Summary

Three reported issues (#2792, #2821, #2892) all share the same root cause: **`rewrite_command()` in `src/discover/registry.rs` attempts to rewrite unsupported command shapes without early validation**, causing:
- Malformed `rtk` invocations that fail at runtime
- Silent predicate dropping during tokenization
- Empty stdout from `find`/`fd` commands

## Root Cause

In `rewrite_command()`, the function:
1. Trims input
2. Checks for `has_heredoc()` and `$(())` — **only these two patterns**
3. **Calls `rewrite_compound()`** — which tokenizes, splits, and classifies
4. Classification may return `Unsupported` — **but only after expensive work**

**Missing**: No early check to reject known-bad command shapes before tokenization.

## Reproduction

| Command | Expected | Actual | RTK Exit |
|---------|----------|--------|----------|
| `find . -type f -not -name \*.log\` | passthrough (rtk find doesn't support `-not`) | Rewritten to `rtk find ...` → FAILS | 3 (Ask) |
| `fd -t f -e rs` | passthrough | Rewritten to `rtk fd ...` → FAILS | 1 (Passthrough) |
| `rtk find . -type f -not foo` | passthrough (already-RTK + unsupported) | No guard, rewrites again | 3 (Ask) |

### Concrete Evidence

```bash
# Command: find with -not predicate
$ rtk rewrite "find . -type f -not -name '*.log'"
rtk find . -type f -not -name '*.log'EXIT: 3

# rtk find itself rejects this:
$ rtk find . -type f -not -name '*.log' -maxdepth 1
rtk: rtk find does not support compound predicates or actions (e.g. -not, -exec). Use `find` directly.
```

## Proposed Fix

Add `is_unsupported_shape()` guard in `rewrite_command()` before `rewrite_compound()`:

```rust
fn is_unsupported_shape(cmd: \&str) -> bool {
    // 1. find/fd entirely — rtk find/fd reject complex predicates
    if cmd.starts_with("find ") || cmd == "find" ||
       cmd.starts_with("fd ") || cmd == "fd" {
        return true;
    }
    // 2. Already-RTK with find/fd subcommand
    if cmd.starts_with("rtk ") {
        let subcmd = cmd.split_whitespace().nth(1).unwrap_or("");
        if subcmd == "find" || subcmd == "fd" {
            return true;
        }
    }
    // 3. Unattestable shell constructs
    if cmd.contains('`') || cmd.contains("$((") || cmd.contains("<(") {
        return true;
    }
    false
}
```

### Why exclude ALL find/fd (not just complex predicates)?

Test results show even simple `find . -type f -name '*.log'` gets rewritten to `rtk find . -type f -name '*.log'` which exits 3 (Ask). And `rtk find` only supports `-name`, `-iname`, `-type`, `-maxdepth` — any other valid GNU find flag silently produces wrong/empty results. Excluding entirely is the safest option.

## Benefits

- **Early exit** — skips expensive tokenization/regex matching for known-bad inputs
- **Predictable passthrough** — unsupported commands run raw
- **Prevents rewrite corruption** — no malformed `rtk find` invocations

## Related Issues

- #2792 — Hook rewrites unsupported shapes into failing invocations
- #2821 — Unknown predicates silently dropped → broader query
- #2892 — find returns empty stdout

## Environment

- RTK version: v0.42.4 (Cargo.toml), tested against v0.43.0
- OS: Linux

## Checklist

- [x] Issue confirmed by reproducing with test cases
- [x] Proposed fix implemented in `src/discover/registry.rs`
- [x] Tests added for unsupported shapes
- [x] Existing tests still pass (2399 pass, 0 fail, 8 ignored — baseline)

Ev3lynx727:
## Summary

Fixes #2902 — which is the root cause of #2792, #2821, #2892.

`rewrite_command()` attempts to rewrite unsupported command shapes without early validation, producing malformed `rtk` invocations that fail at runtime.

## Changes

Adds `is_unsupported_shape()` function that returns early before expensive tokenization/regex matching:

- **find / fd** — excluded entirely because `rtk find`/`rtk fd` only support `-name`, `-iname`, `-type`, `-maxdepth`. Any other predicate (`-not`, `-exec`, `-newermt`, `-path`, etc.) silently produces wrong/empty results.
- **rtk find / rtk fd** — already-routed to a subcommand rtk can't rewrite further
- **Unattestable constructs** — backticks, `$((`, `<(` — output shape is unpredictable

## User-facing impact

Before this fix, commands like `find . -name "*.log"` or `fd -e rs` were rewritten to `rtk find ...` / `rtk fd ...`, which either exit with code 3 (ask) or silently produce wrong results when complex predicates are involved.

After this fix, unsupported command shapes pass through to the native command unchanged — no rewrite, no malformed invocation, no silent corruption.

## Changelog

```
- fix: add is_unsupported_shape() guard to skip rewrite for find/fd, already-routed rtk subcommands, and unattestable shell constructs (#2903)
- fix: prevent malformed rtk find/fd invocations from unsupported predicates (#2902, #2792, #2821, #2892)
```

## Verification

- `cargo fmt --all` — clean
- `cargo clippy --all-targets` — no new warnings
- `cargo test --all` — 2399 passed, 0 failed, 8 ignored (baseline)

## Related

- Closes #2792 — Hook rewrites unsupported shapes
- Closes #2821 — Unknown predicates silently dropped
- Closes #2892 — find returns empty stdout

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -538,6 +538,45 @@ fn collapse_line_continuations(s: &str) -> std::borrow::Cow<'_, str> {
     LINE_CONTINUATION_RE.replace_all(s, " ")
 }
 
+/// Quick shape check to reject commands that `rtk` cannot safely rewrite.
+///
+/// Returns `true` for commands that should pass through unmodified instead of
+/// being rewritten. Saves the cost of tokenization and regex matching that
+/// `rewrite_command` would otherwise perform before eventually reaching the
+/// same conclusion.
+///
+/// Current unsupported shapes:
+/// - `find` / `fd` — `rtk find` and `rtk fd` only accept a limited flag set
+///   (`-name`, `-iname`, `-type`, `-maxdepth`). Any predicate outside that set
+///   (`-not`, `-exec`, `-newermt`, `-path`, …) causes a silent wrong result
+///   or runtime rejection. Excluding these entirely is simpler and safer than
+///   parsing every predicate combination.
+/// - `rtk find` / `rtk fd` — already-routed to a subcommand that rtk itself
+///   cannot rewrite further; returning `None` leaves the existing command as-is.
+/// - Unattestable constructs (backticks, `$((`, `<(`) — shell syntax that rtk
+///   cannot evaluate or predict the output shape of.
+fn is_unsupported_shape(cmd: &str) -> bool {
+    // 1. find/fd entirely — rtk's own find/fd subcommands reject complex predicates
+    if cmd.starts_with("find ") || cmd == "find" || cmd.starts_with("fd ") || cmd == "fd" {
+        return true;
+    }
+
+    // 2. Already-RTK with a find/fd subcommand — no further rewrite to attempt
+    if cmd.starts_with("rtk ") {
+        let subcmd = cmd.split_whitespace().nth(1).unwrap_or("");
+        if subcmd == "find" || subcmd == "fd" {
+            return true;
+        }
+    }
+
+    // 3. Unattestable shell constructs — output shape is unpredictable
+    if cmd.contains('`') || cmd.contains("$((") || cmd.contains("<(") {
+        return true;
+    }
+
+    false
+}
+
 /// Returns `None` if the command is unsupported or ignored (hook should pass through).
 ///
 /// Handles compound commands (`&&`, `||`, `;`) by rewriting each segment independently.
@@ -574,6 +613,12 @@ pub fn rewrite_command(
     }
 
     if has_heredoc(trimmed) || trimmed.contains("$((") {
+        return None;
+    }
+
+    // Reject unsupported command shapes before expensive tokenization/regex work.
+    // find/fd, rtk find/fd, and unattestable constructs skip rewrite entirely.
+    if is_unsupported_shape(trimmed) {
         return None;
     }
 
@@ -1248,6 +1293,55 @@ mod tests {
         }
     }
 
+    // --- is_unsupported_shape tests ---
+
+    #[test]
+    fn test_unsupported_find_not_predicate() {
+        // find with -not should passthrough (unsupported shape)
+        assert_eq!(
+            rewrite_command_no_prefixes("find . -type f -not -name '*.log'", &[]),
+            None
+        );
+    }
+
+    #[test]
+    fn test_unsupported_find_exec_predicate() {
+        // find with -exec should passthrough
+        assert_eq!(
+            rewrite_command_no_prefixes("find . -name '*.log' -exec rm {} \\;", &[]),
+            None
+        );
+    }
+
+    #[test]
+    fn test_unsupported_fd_extensions() {
+        // fd with flags should passthrough
+        assert_eq!(rewrite_command_no_prefixes("fd -t f -e rs", &[]), None);
+    }
+
+    #[test]
+    fn test_unsupported_rtk_find() {
+        // rtk find should not be re-rewritten
+        assert_eq!(rewrite_command_no_prefixes("rtk find . -type f", &[]), None);
+    }
+
+    #[test]
+    fn test_unsupported_rtk_fd() {
+        // rtk fd should not be re-rewritten
+        assert_eq!(rewrite_command_no_prefixes("rtk fd . -type f", &[]), None);
+    }
+
+    #[test]
+    fn test_rtk_git_still_supported() {
+        // rtk git should still go through (not blocked by unsupported shape check)
+        assert_eq!(
+            rewrite_command_no_prefixes("rtk git status", &[]),
+            Some("rtk git status".to_string())
+        );
+    }
+
+    // --- existing tests ---
+
     #[test]
     fn test_classify_find_not_blocked_by_fi() {
         // Regression: "fi" in IGNORED_PREFIXES used to shadow "find" commands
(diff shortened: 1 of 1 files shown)


Ev3lynx727:
Thanks for the review @tapheret2 — both nits addressed:

1. **User-facing impact** — added to PR description (before/after: unsupported shapes now passthrough instead of producing malformed  invocations)
2. **Changelog** — two entries added to PR description, ready for the next release tag

Appreciate the independent pass!

TheCookieLab:
I reproduced the compound-find failure on v0.44.0 and current `develop`. A narrower prototype avoids disabling useful simple `find` rewrites: expose the runtime parser's unsupported-token predicate from `find_cmd`, have the discovery registry return `Ignored` only when those tokens are present, and cover both sides (compound/action forms produce no rewrite; `find . -name '*.rs'` remains rewriteable). The shared predicate also prevents the hook and runtime support lists from drifting. Tested cases include `-o`, `-not`, `-exec`, `-delete`, `-print`, and `-print0`. I can turn that focused prototype into a PR if maintainers prefer selective fallback over skipping every native `find`/`fd` command.

PR #2916, opened 2026-07-09
fix(init): add always_on trigger frontmatter to antigravity rules

wesamahakem:
When running `rtk init --agent antigravity`, the tool generates a workspace rules file at `.agents/rules/antigravity-rtk-rules.md`. However, the generated Markdown file is missing the YAML frontmatter required by Google Antigravity to automatically register the rule. As a result, the agent does not load the rule at startup. This PR fixes it by prepending `trigger: always_on` frontmatter to the template file.

### DIFF
--- a/hooks/antigravity/rules.md
+++ b/hooks/antigravity/rules.md
@@ -1,4 +1,8 @@
+---
+trigger: always_on
+---
 # RTK - Rust Token Killer (Google Antigravity)
+
 
 **Usage**: Token-optimized CLI proxy for shell commands.
 


wesamahakem:
Hi @tapheret2, thank you for the review!

To answer your questions:
1. **Changelog**: Since the repository uses `release-please` to auto-generate the changelog from Conventional Commit messages, the `fix(init):` commit format should automatically capture this for the next patch release.
2. **User-facing behavior change**: The main user-facing change is that the generated rule `.agents/rules/antigravity-rtk-rules.md` will now be loaded and applied automatically at startup by Google Antigravity. Previously, without the `trigger: always_on` frontmatter, the rule file was completely ignored by the agent.

PR #2920, opened 2026-07-10
fix(discover): match Windows project slugs case-insensitively

### LINKED ISSUE #2919: discover: 0 sessions scanned on Windows - drive-letter case mismatch when resolving current project
**Environment**

**Bug**

```

```

**Root cause (analysis)**

```
```

**Suggested fix**

**Workaround**

**Minor related note**

w3lld1:
## Summary
- Fixes Windows `rtk discover` current-project matching when Claude Code stores project slugs with a different drive-letter/path case.
- Updates Windows path slug encoding to replace `:` with `-`, matching the reported Claude Code `e--...` project directory shape.
- Adds a regression test with uppercase `E:\...` cwd input matching a lowercase `e--...` Claude project directory.

Fixes #2919

## Test plan
- [x] `cargo test`
- [x] `cargo test discover::provider::tests::test_encode_project_path_windows`
- [x] `cargo test discover::provider::tests::test_discover_sessions_matches_windows_project_filter_case_insensitively`

I could not manually reproduce on Windows from this Linux environment; the regression test simulates the reported project-directory casing/slug mismatch.

### DIFF
--- a/src/discover/provider.rs
+++ b/src/discover/provider.rs
@@ -78,10 +78,12 @@ impl ClaudeProvider {
                 continue;
             }
 
-            // Apply project filter: substring match on directory name
+            // Apply project filter: substring match on directory name.
+            // Windows project slugs may differ only by drive-letter/path casing
+            // from the current working directory used to build the filter.
             if let Some(filter) = project_filter {
                 let dir_name = path.file_name().and_then(|n| n.to_str()).unwrap_or("");
-                if !dir_name.contains(filter) {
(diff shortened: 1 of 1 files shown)


KuSh:
Closing this as a duplicate of #2931, which makes the same project-directory match case-insensitive but keeps it to Windows via cfg!(windows), so Unix matching is unchanged, and which merges cleanly against develop. The colon half of this PR already landed on develop through #2952. Please take a look at #2931 and comment there, or reply here if you think the unconditional match was deliberate.

Issue #2921, opened 2026-07-10
feat: extend hook rewrite for npm/pnpm/yarn/bun/node + add Codex hook support


7889545:
## Problem

RTK has subcommands like `rtk npm`, `rtk pnpm`, `rtk pytest` that work when called manually, but the hook engine's `rtk rewrite` does NOT auto-rewrite these common dev commands:

| Command | `rtk rewrite` result | Has RTK subcommand? |
|---------|----------------------|-------------------|
| `npm test` | No rewrite (exit 1) | ✅ `rtk npm` |
| `pnpm test` | No rewrite (exit 1) | ✅ `rtk pnpm` |
| `yarn test` | No rewrite (exit 1) | ❌ No `rtk yarn` |
| `bun test` | No rewrite (exit 1) | ❌ No `rtk bun` |
| `node script.js` | No rewrite (exit 1) | ❌ No `rtk node` |

Tested on v0.40.0 and v0.43.0 — same behavior on both.

## Expected

```
npm test      → rtk npm test
pnpm test     → rtk pnpm test
yarn test     → rtk yarn test (or mapped to rtk pnpm?)
bun test      → rtk bun test
node script.js → rtk node script.js (or passthrough with tracking)
```

The subcommands already exist for npm/pnpm — the rewrite table just needs to route to them.

## Also: Codex CLI hook support

Currently supported agents with `rtk hook <agent>`:
- claude ✅
- cursor ✅
- gemini ✅
- copilot ✅
- codex ❌ (only prompt-level via `rtk init --codex`)

Request: add `rtk hook codex` processor so Codex gets the same transparent rewriting as other agents.

## Environment
- RTK: 0.40.0 → 0.43.0 (tested both)
- OS: Windows 11, Git Bash shell
- Claude Code hook: working ✅

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 90% |

### Summary

Feature request to extend rtk's hook rewrite table to npm/pnpm/yarn/bun/node dev commands and add a Codex CLI hook processor; recommend triaging as two separate enhancements.

---
[...]

PR #2927, opened 2026-07-10
feat(hook): add Codex command rewriting

paranoa233:
## Summary

- add a native `rtk hook codex` processor for Codex `PreToolUse` payloads
- install the hook in global or project `hooks.json` without overwriting existing hooks
- make installation idempotent, remove only RTK's entry on uninstall, and report hook status
- update Codex integration and hook architecture documentation

## Why

Codex currently receives prompt-level RTK guidance but cannot transparently rewrite supported shell commands. This adds the same automatic path while using Codex's native `updatedInput.command` protocol.

## Impact

- unsupported, malformed, or non-Bash payloads emit no output and remain non-blocking
- rewritten input preserves other `tool_input` fields
- accepting the rewrite does not replace Codex's normal sandbox and approval checks
- existing user hook configuration is preserved

## Test plan

- `cargo fmt --all --check`
- `cargo test` (2,447 unit tests plus integration suites)
- manual stdin smoke test for rewrite and pass-through behavior

Part of #2921. Paired with #2926.


KuSh:

Issue #1441, opened 2026-04-21
`rtk discover` reports false-positive missed savings for hook-rewritten commands


FlorianBruniaux:
## Summary

`rtk discover` reads the **pre-hook** command text from Claude Code JSONL transcripts and classifies every non-`rtk ` prefix as \"missed\". The `~/.claude/hooks/rtk-rewrite.sh` hook rewrites most of these at runtime (e.g. `grep -n …` → `rtk grep -n …`), but Claude Code never writes the post-hook command back to the transcript. Result: commands that actually ran through RTK get counted as missed, the adoption percentage is artificially deflated, and the user-facing doc actively misdiagnoses the cause.

The sibling command `rtk session` already solves this correctly in `src/analytics/session_cmd.rs:28-51`. `rtk discover` is the outlier.

## Reproduction

On a machine with rtk ≥ 0.23.0 and the official hook installed, pick any recent Claude Code session JSONL and locate a bash `tool_use` whose command starts with `grep -n` (or any other hook-rewritten command).

```
CMD (from tool_use.input.command):
  grep -n "^version" crates/ccboard-core/Cargo.toml ...

tool_result.content (93 chars):
  1 matches in 1F:

  [file] crates/ccboard-core/Cargo.toml (1):
       4: version.workspace = true
```

The result is in the `rtk grep` filter format (`[file]` header, "X matches in YF:" summary). Raw grep would have produced `path:line:content`. The hook rewrote the command; the transcript still logs the original.

Now run:

```bash
rtk discover --all
```

The `grep -n` command appears in the **MISSED SAVINGS** column, even though the adjacent tool_result proves it ran through RTK.

## Expected

Commands that `classify_command` returns as `Classification::Supported` should bucket as already-covered, matching `rtk session`'s behavior. Only `RTK_DISABLED=...` prefixes and unsupported base commands (`pnpm build`, `node`, `python3`, …) should appear as missed.

## Actual

`src/discover/mod.rs:169-175` counts only the literal `rtk ` prefix:

```rust
Classification::Ignored => {
    if part.trim().starts_with("rtk ") {
        already_rtk += 1;
    }
}
```

Every `Classification::Supported` goes to the missed bucket regardless of hook behavior.

## Root Cause

Claude Code's JSONL logs the assistant's `tool_use.input.command` — what the model authored before any `PreToolUse` hook runs. The post-hook command is never persisted.

[...]

PR #2930, opened 2026-07-10
fix(discover): count hook-rewritten commands as RTK usage

### LINKED ISSUE #2928: discover: 'Already using RTK' undercounts (reads pre-rewrite transcript command)
## Problem

`rtk discover` reports near-0% RTK adoption even when the PreToolUse hook is rewriting 100% of commands.

Same machine, same 30-day window:

| Source | Metric |
|---|---|
| `rtk discover` | `Already using RTK: 99 commands (0.3%)` |
| `rtk gain` | `21280 commands, 73.5M tokens saved (78.1%)` |

## Root cause

The Claude Code hook rewrites via PreToolUse `updatedInput` (`grep -n …` → `rtk grep -n …`). The rewrite executes correctly (proven by `rtk gain`), but the **session transcript records the original assistant-emitted command**, not the post-hook executed one.

`discover` scans transcripts and counts commands literally starting with `rtk`, so it only catches the rare case where the model itself typed `rtk`. Verified directly:

```
raw '"command":"grep -n'  ->  294
'"command":"rtk grep'     ->   16
```

Result: `discover` will always show ~0% adoption regardless of hook effectiveness, contradicting `gain`.

## Suggested fix

Either (a) cross-reference `rtk gain`'s execution log (real invocations) when computing the 'already using' ratio, or (b) note in the report that the metric measures *model-typed* `rtk` only and points users to `rtk gain` for actual coverage.

rtk 0.42.2, Claude Code (PreToolUse `updatedInput` rewrite path).

Rashal10:
## Summary
- `rtk discover` now uses `rewrite_command()` to detect commands the hook would rewrite
- Those commands count as "Already using RTK" instead of "Missed savings"
- Added unit tests in `src/discover/mod.rs`

Fixes #2928

## Test plan
- [ ] Manual: run `rtk discover` on a project with hooks active and verify "Already using RTK" is higher

### DIFF
--- a/src/discover/mod.rs
+++ b/src/discover/mod.rs
@@ -11,11 +11,12 @@ use std::collections::HashMap;
 
 use provider::{ClaudeProvider, SessionProvider};
 use registry::{
-    category_avg_tokens, classify_command, split_command_chain, strip_disabled_prefix,
-    Classification,
+    category_avg_tokens, classify_command, rewrite_command, split_command_chain,
+    strip_disabled_prefix, Classification,
 };
 use report::{DiscoverReport, SupportedEntry, UnsupportedEntry};
 
+use crate::discover::lexer::contains_unattestable_construct;
 use crate::discover::registry::prefix_contains_rtk_disabled;
 
 /// Aggregation bucket for supported commands.
@@ -65,6 +66,10 @@ pub fn run(
 
     let sessions = provider.discover_sessions(project_filter.as_deref(), Some(since_days))?;
 
+    let (excluded, transparent_prefixes) = crate::core::config::Config::load()
+        .map(|c| (c.hooks.exclude_commands, c.hooks.transparent_prefixes))
+        .unwrap_or_default();
+
     if verbose > 0 {
         eprintln!("Scanning {} session files...", sessions.len());
         for s in &sessions {
@@ -111,6 +116,11 @@ pub fn run(
                             // RTK_DISABLED on unsupported/ignored command — not interesting
                         }
                     }
+                    continue;
+                }
+
+                if is_already_using_rtk(part, &excluded, &transparent_prefixes) {
+                    already_rtk += 1;
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


felipetruman:
Thanks for tackling #2928 — the undercount is real (measured 1.1% reported vs 43.0% actual on 30 days of my production transcripts). Two concerns with simulating the rewrite instead of reading the recorded one, though:

1. **It breaks discover for users who haven't installed the hook.** `is_already_using_rtk()` runs unconditionally, so any command `rewrite_command()` *would* rewrite counts as "already using RTK" even when no hook exists and nothing was rewritten. For those users the MISSED SAVINGS table goes ~empty — and they're exactly the audience discover exists for. At minimum this needs to be gated on hook-installed detection (#2671 adds `AgentIntegrationStatus::claude_hook_installed` for this).

2. **Simulation with today's config misclassifies history.** The rewrite is simulated with the *current* `exclude_commands`/`transparent_prefixes` and the *current* rules. Sessions that predate the hook install (or predate a rule) get counted as covered even though the tokens were fully spent at the time. On my data the simulated approach overcounts actual coverage by ~4 percentage points; the direction of the error hides real missed savings.

Claude Code transcripts actually record the rewrite that happened: an `attachment` line of type `hook_success` with `permissionDecisionReason: "RTK auto-rewrite"` and `updatedInput`, keyed by `toolUseID`. #2308 reads that instead — ground truth, no false positives, and no behavior change for non-hook users. Might be worth converging on that approach and keeping this PR's test cases (the excluded-command and unattestable-construct ones are good).

KuSh:
Thanks for this. `rtk discover` stopped deciding coverage from the literal `rtk ` prefix in #3206, merged 2026-09-02 and shipped in v0.48.0: it reconciles each command against the recorded hook decisions and, when there is no record, estimates hook coverage the same way `rtk session` does, so hook-rewritten commands land in the already-covered bucket instead of missed savings. That covers what this PR set out to fix, and it also handles the cases around it, a command that would be rewritten but with no hook installed, config-excluded commands, and `RTK_DISABLED=` bypasses, which get their own bucket. The tests here have equivalents on develop in src/discover/mod.rs. Linked issue #2928 was closed as a duplicate of #1441, which the same change fixed. If you still see supported commands counted as missed on the current develop, please comment and this can be reopened.

PR #2931, opened 2026-07-10
fix(discover): match Claude project dirs case-insensitively on Windows

### LINKED ISSUE #2919: discover: 0 sessions scanned on Windows - drive-letter case mismatch when resolving current project

TemRevil:
Fixes #2919

`rtk discover` (and `rtk learn`) always reported `Scanned: 0 sessions` in default current-project mode on Windows. Two independent defects in `src/discover/provider.rs`:

**1. The drive-letter colon was never sanitized.** `encode_project_path()` kept `:`, producing filters like `E:-PROJECT-app` — but `:` is not a legal character in Windows directory names, so the filter could never match any real folder under `~/.claude/projects/`. Claude Code encodes `E:\PROJECT\app` as `E--PROJECT-app`. Verified against a real machine: for cwd `J:\Z\GitHub\Corner`, Claude Code's history folder is `J--Z-GitHub-Corner`; the previous encoding computes `J:-Z-GitHub-Corner`, the new one matches exactly. The existing `test_encode_project_path_windows` expectation encoded the impossible form and was updated.

**2. Case-sensitive matching (the mismatch reported in the issue).** Claude Code encodes the folder from whatever cwd casing the session was launched with, while `std::env::current_dir()` reports an uppercase drive letter; the same project can even produce sibling folders (`E--proj` and `e--proj`, as in the issue). NTFS is case-insensitive, so the directory-name match is now case-insensitive on Windows only — Unix behavior is unchanged. This also makes `-p <name>` case-insensitive on Windows, consistent with the platform's path semantics.

Both fixes flow through `discover_sessions()`, so `rtk learn`'s current-project mode is fixed too.

**Tests:** updated the Windows encoding test, added a lowercase-drive encoding test, and unit tests for the new `dir_matches_filter` helper covering both casing directions plus a non-match. The helper takes an explicit case-insensitivity flag so the Windows semantics are exercised on any CI platform.

**Note on verification:** developed on a Windows machine without a local Rust toolchain, so the encoding behavior was verified against real Claude Code history folders (above) and the suite is left to CI. Happy to follow up on anything clippy/fmt flags.

### DIFF
--- a/src/discover/provider.rs
+++ b/src/discover/provider.rs
@@ -41,6 +41,27 @@ pub trait SessionProvider {
 
 pub struct ClaudeProvider;
 
+/// Substring match of a project filter against a project directory name.
+///
+/// On Windows the match is case-insensitive: NTFS paths are case-insensitive
+/// and Claude Code encodes the project folder from whatever cwd casing it was
+/// launched with, so the same project can produce sibling folders like
+/// `E--proj` and `e--proj` while `std::env::current_dir()` reports an
+/// uppercase drive letter. A case-sensitive match then finds zero sessions.
+fn dir_matches_filter(dir_name: &str, filter: &str) -> bool {
+    dir_matches_filter_impl(dir_name, filter, cfg!(windows))
+}
+
+fn dir_matches_filter_impl(dir_name: &str, filter: &str, case_insensitive: bool) -> bool {
+    if case_insensitive {
+        dir_name
+            .to_ascii_lowercase()
+            .contains(&filter.to_ascii_lowercase())
+    } else {
+        dir_name.contains(filter)
+    }
+}
+
 impl ClaudeProvider {
     /// Get the base directory for Claude Code projects.
     fn projects_dir() -> Result<PathBuf> {
@@ -81,7 +102,7 @@ impl ClaudeProvider {
             // Apply project filter: substring match on directory name
             if let Some(filter) = project_filter {
                 let dir_name = path.file_name().and_then(|n| n.to_str()).unwrap_or("");
-                if !dir_name.contains(filter) {
+                if !dir_matches_filter(dir_name, filter) {
                     continue;
                 }
             }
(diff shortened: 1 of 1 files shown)


TemRevil:
Following up on the verification note in the description: I ran the CLAUDE.md pre-commit gate in a rust:1 container. fmt --check came back clean, clippy has no warnings, and cargo test --all passes, including the new encoding and dir_matches_filter tests.

pszymkowiak:
I did a full code review of this and the logic is correct and complete — both the colon encoding (verified against Claude Code's real folder format, e.g. `J:\Z\GitHub\Corner` → `J--Z-GitHub-Corner`) and the platform-aware case-insensitive matching. Builds clean, all 23 provider unit tests pass.

**But there's one gap I can't close and I want to flag it honestly: no one has run this end-to-end on real Windows yet.** The author verified the *slug format* against real folders but had no Rust toolchain on Windows, the unit tests only exercise the pure `dir_matches_filter_impl` (with an explicit flag, so the `cfg!(windows)` path never runs on CI/mac), and I could only test on macOS where `cfg!(windows)` is inactive. So the actual behavior — does `rtk discover` now find sessions against a live `~/.claude/projects/` on Windows — is unverified by execution.

**Could a Windows user run this quick check?** In a real project that has Claude Code history:

```powershell
# 1. Build this branch
cargo build --release

# 2. From a project dir you've used with Claude Code:
#    Before this PR, `rtk discover` reports "Scanned: 0 sessions" (issue #2919)
.\target\release\rtk discover

# Expected AFTER this PR: it finds your sessions (Scanned: N > 0).
# Bonus: confirm it works with an uppercase-vs-lowercase drive-letter folder,
# and that `rtk discover -p <name>` matches case-insensitively.
```

The fix looks right on paper (fixes #2919 at both the encode and match sites) — just want one real-Windows confirmation before it's merged, since the whole point is Windows-only behavior that our CI/dev environments don't actually exercise.

TemRevil:
Ran this on a real Windows machine, not Docker or WSL. Base branch (5d32d07, before this fix): `Scanned: 0 sessions (last 30 days), 0 Bash commands`, matching #2919 exactly. This branch (fa156a7): `Scanned: 19 sessions (last 30 days), 8431 Bash commands`, against a real `~/.claude/projects/` history with weeks of actual project use.

Also tested the case-insensitive `-p` filter directly against a real on-disk folder (mixed-case `J--Z-GitHub-Corner`): an all-lowercase filter matches those same 19 sessions on this branch and 0 on base, and an all-uppercase filter matches the same 19 on this branch too. Covers both halves of the fix together, the colon-to-dash encoding and the case-insensitive substring match.

One honest caveat: this machine had no Rust toolchain until today, so I installed rustup fresh with the GNU target since there's no Visual Studio Build Tools here. Shouldn't affect what's being tested since cfg!(windows) is true either way, but flagging it since you were specific about wanting real execution rather than a read of the code.

TemRevil:
#2952 covers the colon half of this and closed #2919, so I rebased onto develop and dropped that part. What is left is the other half of what I originally hit: `-p` still matches with a case-sensitive `contains` on develop, so on Windows a filter whose casing does not line up with the folder on disk finds nothing.

Diff is down to `dir_matches_filter` and two tests.

One caveat on verification: cargo is blocked by an Application Control policy on this machine and Docker would not come up, so I could not run the full suite this time. I compiled the function and the new assertions with rustc directly and they pass, and the current develop path does fail the case-mismatched filter. rustfmt clean. Relying on CI for the rest.

Fine by me to close this if it is not worth carrying now that #2919 is done.

PR #2938, opened 2026-07-10
perf(hook): avoid Pi package barrel import

### LINKED ISSUE #2753: pi slow startup
Causing about 2 - 3 seconds slow startup in pi and omp. 

Below is prepared by agent itself. 

~/.pi/agent/extensions/rtk.ts   and ~/.omp/agent/extensions/rtk.ts does a runtime value import:

import { isToolCallEventType } from "@earendil-works/pi-coding-agent"

This resolves to the package barrel (dist/index.js), which re-exports the entire agent core — a ~9.2 MB module graph. bun must parse and evaluate that whole graph at extension load just to obtain one function.

- value import from barrel 
- type-only / no runtime import 

So the import costs avoidable startup time. Disabling the extension entirely is what made startup noticeably faster.

Root cause: isToolCallEventType is trivial — return event.toolName === toolName (dist/core/extensions/types.js:42). Pulling the full barrel for it is unnecessary.

Fix: drop the value import and inline the check. import type { ExtensionAPI } stays (erased at transpile, zero runtime cost).

// remove: import { isToolCallEventType } from "@earendil-works/pi-coding-agent"
// change: if (!isToolCallEventType("bash", event)) return
// to:     if (event.toolName !== "bash") return

Behaviorally identical; startup drops back to baseline.

Note: a subpath import (.../core/extensions) is not a viable al

Note: a subpath import (.../core/extensions) is not a viable alternative — the package's exports map declares only ".", so that specifier fails to resolve (Cannot find module), and even if it resolved, that subpath barrel re-exports loader.js/runner.js . Inlining the comparison is the only clean fix.

sed -i \
  -e '/^import { isToolCallEventType }/d' \
  -e 's/if (!isToolCallEventType("bash", event)) return/if (event.toolName !== "bash") return/' \
  "$target"

raj921:
## Summary

- Replace the runtime package import with a local type guard, so the Pi extension does not load the package barrel at startup.
- Add a regression check that rejects value imports in the installed extension.

Fixes #2753

## Test plan

- [x] Type-checked `hooks/pi/rtk.ts` against `@earendil-works/pi-coding-agent` 0.80.6
- [x] `cargo test test_run_pi_mode_global_installs_plugin`

## Timing

With Bun 1.3.14 on macOS arm64, loading the package barrel took about 250 ms after warmup. Loading the small extensions module took about 10 ms.

### DIFF
--- a/hooks/pi/rtk.ts
+++ b/hooks/pi/rtk.ts
@@ -19,6 +19,12 @@ import type {
 const REWRITE_TIMEOUT_MS = 2_000
 const MIN_SUPPORTED_RTK_MINOR = 23
 
+// Local reimplementation of the package's `isToolCallEventType("bash", event)` type
+// guard. That helper is a value export, so importing it pulls in the whole
+// `@earendil-works/pi-coding-agent` barrel at extension load — profiled at ~250ms
+// warmed, vs ~10ms for a type-only import. `BashToolCallEvent`/`ToolCallEvent`
+// below are type-only imports and are erased at compile time, so they carry none
+// of that cost. See #2753.
 function isBashToolCallEvent(event: ToolCallEvent): event is BashToolCallEvent {
   return event.toolName === "bash"
 }
--- a/hooks/pi/rtk.ts
+++ b/hooks/pi/rtk.ts
@@ -10,11 +10,18 @@
 //   1           No RTK equivalent → pass through unchanged
 //   3 + stdout  Rewrite (advisory) → mutate command
 
-import type { ExtensionAPI } from "@earendil-works/pi-coding-agent"
-import { isToolCallEventType } from "@earendil-works/pi-coding-agent"
+import type {
+  BashToolCallEvent,
+  ExtensionAPI,
+  ToolCallEvent,
+} from "@earendil-works/pi-coding-agent"
 
 const REWRITE_TIMEOUT_MS = 2_000
 const MIN_SUPPORTED_RTK_MINOR = 23
+
+function isBashToolCallEvent(event: ToolCallEvent): event is BashToolCallEvent {
+  return event.toolName === "bash"
+}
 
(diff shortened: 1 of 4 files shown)


hooks/pi/rtk.ts
```diff
@@ -10,12 +10,19 @@
...
+import type {
+  BashToolCallEvent,
+  ExtensionAPI,
+  ToolCallEvent,
+} from "@earendil-works/pi-coding-agent"
 
 const REWRITE_TIMEOUT_MS = 2_000
 const MIN_SUPPORTED_RTK_MINOR = 23
 
+function isBashToolCallEvent(event: ToolCallEvent): event is BashToolCallEvent {
```
KuSh, line 22: Could you add a comment explaining where it comes from and why?

src/hooks/init.rs
```diff
@@ -7089,6 +7089,10 @@ mod tests {
                 content.contains("rtk rewrite"),
                 "extension must delegate to rtk rewrite"
             );
+            assert!(
+                !content.contains("import {"),
+                "extension must not load the Pi package at runtime"
+            );
```
KuSh, line 7095: praise, thanks for adding that guard against regression. Could you add a simple comment pointing to the original issue?

KuSh (changes requested):
LGTM except a few comments missing to document things

hooks/pi/rtk.ts
KuSh: Added a comment explaining it: it's a local reimplementation of the package's isToolCallEventType("bash", event), kept local specifically because that helper is a value export and importing it pulls in the whole @earendil-works/pi-coding-agent barrel at load — the whole point of this PR. Linked #2753.

src/hooks/init.rs
KuSh: Added a comment pointing to #2753.

KuSh (approved):

PR #782, opened 2026-03-22
feat: add Gradle and Maven support with Surefire parser (75-90% savings)

romualdbassinot-efficy:
## Summary

- Add `rtk gradle` command with sub-commands: `build`, `test`, `check`, `dependencies` — filtering Gradle output to errors/failures only (75-90% savings)
- Add `rtk mvn` command with sub-commands: `compile`, `test`, `package`, `verify`, `install`, `dependency:tree` — with Surefire XML parser for structured test failure output (75-90% savings)
- Add real fixtures for both tools (build success/fail, test pass/fail)
- Register both commands in `src/main.rs` Commands enum following the Go sub-enum pattern
- Update `CHANGELOG.md` and `src/discover/rules.rs` for missed-savings detection

## Changes since initial review (33292fc)

- **gradle**: `TEST_SUMMARY_RE` now captures optional skipped count — `passed` is correctly `total - failed - skipped` (previously skipped tests were counted as passing)
- **gradle**: empty/no-test runs now return `"Gradle test: no tests found"` instead of `"Gradle test: 0 passed"`
- **maven**: all-passed format string now uses `total_passed` instead of `total_run`, so skipped tests are not reported as passing
- **both**: added `insta` snapshot tests (8 total) for build and test success/failure fixtures

## Test plan

- [x] `cargo check --all-targets` passes
- [x] Token savings ≥75% verified via unit tests
- [x] Snapshot tests added for all 4 fixture files per module
- [ ] `rtk gradle build` filters Gradle build output to errors only
- [ ] `rtk gradle test` shows only failed tests with skipped count correct
- [ ] `rtk mvn compile` filters Maven compile errors
- [ ] `rtk mvn test` uses Surefire XML when available, falls back to text parser
- [ ] Fallback to raw command on filter failure

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2939, opened 2026-07-10
feat(mvn): recover test failures from Surefire/Failsafe XML reports

mariuszs:
## Problem

Maven's stdout is not always a complete record of what failed. A forked-VM crash, parallel forks interleaving output, `-q`, `<redirectTestOutputToFile>`, or `-Dmaven.test.failure.ignore=true` can hide per-test detail — while `target/surefire-reports/TEST-*.xml` still carries it authoritatively. No stdout filter, however good, can recover what stdout never printed. Today the agent's only options are re-running the build or reading the XML files by hand.

## What this does

After the stdout filter runs (Test / Package / quiet routes), a post-filter pass parses the fresh `target/surefire-reports/` and `target/failsafe-reports/` XML and compares the reported failures against the filtered text:

- **Failures already visible in stdout are left alone** — when stdout tells the whole story (the overwhelmingly common case), output is unchanged **byte for byte**.
- **Only the failures stdout missed are appended**, in `[ERROR]`-prefixed native style: `Class.method <<< FAILURE!`, then the exception header and the stack trace verbatim (head-capped at 30 lines).
- When stdout carried no `Tests run:` aggregate at all (blind-stdout case), the aggregate line is restored from the XML.

Safety rails:

- **mtime freshness gate** — reports older than the run's start time are ignored, so stale files from a previous run can never leak in.
- **Reactor-aware** — discovers report dirs at the root and depth-1 modules (`<module>/target/{surefire,failsafe}-reports`).
- **Capped** — at most `CAP_WARNINGS` appended blocks (`… +N more recovered failures` tail), stack traces head-capped.
- **Never worse** — the block is appended only while the result still fits under the raw size, so the existing `never_worse` guard is never tripped by it. When the raw log is too small to fit the block (typically `-q`), the filtered output is left exactly as today.

Example — blind stdout (`-q` or a forked-VM crash), two failures only the XML knows about:

```
[INFO] BUILD FAILURE

[ERROR] 2 test failure(s) found in Surefire/Failsafe XML reports but missing from build output:
[ERROR]   com.example.FailingTest.shouldReturnUser <<< FAILURE!
[ERROR]     org.opentest4j.AssertionFailedError: expected:<200> but was:<404>
[ERROR]     	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
[ERROR]     	at com.example.FailingTest.shouldReturnUser(FailingTest.java:42)
[ERROR]     	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[ERROR]   com.example.FailingTest.shouldHandleNull <<< FAILURE!
[ERROR]     java.lang.AssertionError: Unexpected exception: NullPointerException
[ERROR]     	at com.example.FailingTest.shouldHandleNull(FailingTest.java:55)
[ERROR]     	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
[ERROR] Tests run: 4, Failures: 2, Errors: 0, Skipped: 0 (from XML reports)
```

## Implementation

Two modules; the existing filters are untouched — only the three routing arms in `run_tool()` go through a small `run_with_recovery` wrapper that appends the pass after the stdout filter:

| Module | Role |
|---|---|
| `surefire_reports.rs` | quick-xml streaming parser for `TEST-*.xml` (counts, failures, mtime gate) |
| `surefire_recovery.rs` | diff XML failures vs filtered text, render the missing ones |

No new dependencies — `quick-xml` and `tempfile` are already in `Cargo.toml`.

**Scope note (rev 2):** this revision is deliberately minimal. The earlier stack-trace collapsing and application-package autodetection (`stack_trace.rs`, `pom_groupid.rs`, ~900 lines) are gone — traces are appended verbatim with a line cap instead — and captured `system-out`/`system-err` recovery was dropped too. Both are possible follow-ups once the core mechanism is in. Rebased onto current `develop` (mvnd support).

## Validation

- 22 new unit tests (`cargo test 'cmds::jvm::surefire_re'`); `cargo test --all` shows no new failures against the `develop` baseline.
- `cargo clippy --all-targets` zero warnings; new modules rustfmt-clean.
- End-to-end smoke through the binary with a stub `./mvnw` that writes a fresh `TEST-*.xml` and prints a forked-VM-crash log: the two XML-only failures are appended after the filtered log (tee hint intact, exit code propagated); the same run with stdout covering the failures is unchanged; under `-q` (raw too small for the block) the filtered output is left as is.

Output cost: zero on every run where stdout already covers the failures; on recovery it adds only the missing failure blocks, capped as above.


mariuszs:
Closing this. Following the discussion in #782, I re-tested the motivating cases against a real project (Maven 3.9.16, Surefire 3.5.3, JUnit 5.12), running each through `develop`'s binary and this branch's:

- `-Dmaven.test.failure.ignore=true`, `-q`, and both combined
- `-Dsurefire.printSummary=false`, with and without `failure.ignore`
- `redirectTestOutputToFile`
- a forked-VM `System.exit` crash, before and after the failing class
- parallel forks (`forkCount=4`; `forkCount=2` with `reuseForks=false`), including a fork crashing while the others are still running
- JUnit in-JVM parallel classes, alone and with `forkCount=3`

In all 13 scenarios `develop`'s stdout filter already names every failing test, and this PR appends nothing. The reason is structural: Surefire writes a test class's XML report at the same moment it prints that class's result block, and a class that crashes the fork writes no XML at all. So the XML never knows about a failure that stdout didn't already show.

Without a reproducible case where the XML carries more than stdout, this is ~1000 lines with nothing to recover.

PR #2965, opened 2026-07-12
fix(hooks): don't rewrite pipe-feeding segments whose filter changes output shape

### LINKED ISSUE #2962: Pipe-left rewrites feed rtk's summarized output into downstream programs (grep | wc -l counts 28 for 60 matches)
## What happens

The hook rewrites the left side of a pipe, so a *downstream program* — not the agent — ends up consuming rtk's compacted format. The compacted format has headers, caps, elision markers and decoration, and the consumer silently gets wrong data.

Reproduced on rtk 0.43.0; present in `develop` (`5d32d07`).

**`grep | wc -l` miscounts** (file has 60 matching lines):

```console
$ grep -n "match" sixty.txt | wc -l      # → rtk grep -n match sixty.txt | wc -l
28
```

28 = 1 header line (`60 matches in 1 files:`) + 1 blank + 25 shown matches + 1 elision line. The real answer is 60.

The intermediate output the pipe consumer actually sees:

```
60 matches in 1 files:

1:match line 1
…
25:match line 25
  +35 more in /.../sixty.txt [see remaining: tail -n +26 ~/.local/share/rtk/tee/…]
```

Note the path is also elided to `/.../sixty.txt`, so a consumer trying to use it as a path fails.

**Same class of problem for other shape-changing filters:**
- `ls` — `$(ls -t | head -1)` captures a size-decorated entry rather than a bare name.
- `head` → `rtk read --max-lines` — the `[N more lines]` marker reaches whatever parses the file content.

## Why this is the existing design's own rule

`find`/`fd` already have exactly this carve-out — `registry.rs` skips rewriting them when they feed a pipe, because rtk's tree output is incompatible with `xargs` (#439). The same reasoning applies to every filter whose output shape differs from the native tool's; the carve-out is just too narrow.

This also sits directly under the project's **Transparency** principle: "If an LLM parses `git diff` output, RTK's filtered version must still look like `git diff` output." When a pipe is present, the parser is a *program*, and the shape difference is fatal rather than cosmetic.

PR incoming: generalize the find/fd carve-out to a `PIPE_INCOMPATIBLE` list (grep, rg, ls, head, tail, cat) + regression tests. Non-piped uses keep their rewrites, so the token savings on the common case are untouched.

xuing:
## Summary
- `find`/`fd` already skip the pipe-left rewrite because rtk's output shape breaks `xargs` (#439). The same reasoning applies to every filter that reshapes output — this generalizes the special case into a `PIPE_INCOMPATIBLE` list: `find`, `fd`, `grep`, `rg`, `ls`, `head`, `tail`, `cat`
- Downstream programs stop receiving rtk's summary headers, capped match lists, elision markers and size columns: `grep -n X f | wc -l` counted 28 for 60 real matches; `$(ls -t | head -1)` captured a decorated name
- Commands **without** a pipe keep their rewrites, so the token savings on the common (agent-reads-it) path are untouched

Fixes #2962

## Implementation notes
- The inline `find`/`fd` check in `rewrite_compound` becomes one `is_pipe_incompatible(seg)` helper + a const list — net behavior change is which names are in the list
- The helper skips leading `VAR=value` assignments, matching `rewrite_segment`'s own env-prefix handling (`LC_ALL=C grep … | sort` is correctly recognized)
- Only the segment feeding the pipe is skipped: `grep … | wc -l && git status` still rewrites `git status`

## Test plan
- [x] `test_rewrite_grep_pipe_skipped` — the `wc -l` miscount case is not rewritten
- [x] `test_rewrite_ls_pipe_skipped`, `test_rewrite_head_pipe_skipped` — same for the other shape-changing filters
- [x] `test_rewrite_grep_env_prefix_pipe_skipped` — env prefix doesn't hide the command
- [x] `test_rewrite_grep_no_pipe_still_rewritten` — no pipe, still `rtk grep` (savings preserved)
- [x] `test_rewrite_pipe_skip_only_affects_pipeline_segment` — `&&` segments after the pipeline still rewrite
- [x] Existing `test_rewrite_find_pipe_skipped` / `test_rewrite_find_no_pipe_still_rewritten` / `test_rewrite_pipe_first_only` unchanged and passing
- [x] End-to-end via `rtk hook claude`: `grep … | wc -l` and `ls -t | head -1` now pass through; `grep -rn TODO src` still rewrites
- [x] `cargo fmt --check && cargo clippy --all-targets && cargo test` — 2441 passed, 0 failed, no new warnings

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -595,6 +595,26 @@ pub fn rewrite_command(
     rewrite_compound(trimmed, &compiled, &normalized_prefixes)
 }
 
+/// Commands whose rtk rewrite changes the OUTPUT SHAPE — summary headers,
+/// grouped/capped matches, elision markers, size columns, tree layout.
+/// Feeding that shape into a pipe corrupts the consumer: `grep -n X f | wc -l`
+/// counts the header and elision lines (measured 28 for 60 real matches),
+/// `$(ls -t | head -1)` captures a size-decorated name, and `rtk read`'s
+/// `[N more lines]` marker reaches whatever parses the file content.
+/// `find`/`fd` were the original members (tree output vs `xargs`, #439).
+const PIPE_INCOMPATIBLE: &[&str] = &["find", "fd", "grep", "rg", "ls", "head", "tail", "cat"];
+
+/// True when `seg`'s command is one whose filtered output must not feed a pipe.
+/// Leading `VAR=value` assignments are skipped, matching `rewrite_segment`'s
+/// own env-prefix handling.
+fn is_pipe_incompatible(seg: &str) -> bool {
+    let first = seg
+        .split_whitespace()
+        .find(|word| !word.contains('='))
+        .unwrap_or("");
+    PIPE_INCOMPATIBLE.contains(&first)
+}
+
 /// Rewrite a compound command (with `&&`, `||`, `;`, `|`) by rewriting each segment.
 fn rewrite_compound(
     cmd: &str,
@@ -637,11 +657,7 @@ fn rewrite_compound(
             }
             TokenKind::Pipe => {
                 let seg = cmd[seg_start..tok.offset].trim();
-                let is_pipe_incompatible = seg.starts_with("find ")
-                    || seg == "find"
-                    || seg.starts_with("fd ")
-                    || seg == "fd";
-                let rewritten = if is_pipe_incompatible {
+                let rewritten = if is_pipe_incompatible(seg) {
                     seg.to_string()
                 } else {
                     rewrite_segment(seg, excluded, transparent_prefixes)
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for this, and sorry it sat — the diagnosis in #2962 was right, and the `find`/`fd` carve-out really was too narrow.

It's since been fixed upstream of this branch, which is why the PR now conflicts. **#3128** (`fix(rewrite): only rewrite safe final pipeline stages`) and **#3343** (`feat(rewrite): rewrite in pipe when consumer is safe`) replaced the inline `seg.starts_with("find ")` check this PR generalizes with a consumer-aware model: a per-rule `PipelineSafety` (`producer_safe`/`final_safe`) plus a `SAFE_PIPE_CONSUMERS` list. The producer is rewritten only when *every* downstream consumer is a display-only tail (`cat`, `head`, non-following `tail`). So the conflict isn't mechanical — the code this patches no longer exists.

I ran all six of the tests added here against `develop` (`d0c29851`). Four already pass:

| case | this PR expects | develop |
|---|---|---|
| `grep -n match sixty.txt \| wc -l` | None | None ✅ |
| `LC_ALL=C grep -n match f.txt \| sort` | None | None ✅ |
| `grep -rn TODO src` | `rtk grep …` | `rtk grep …` ✅ |
| `grep -n x f \| wc -l && git status` | `… && rtk git status` | `… && rtk git status` ✅ |
| `ls -t \| head -1` | None | `rtk ls -t \| head -1` |
| `head -20 src/main.rs \| grep use` | None | `head -20 src/main.rs \| rtk grep use` |

The last row is develop being deliberately better than the `PIPE_INCOMPATIBLE` approach: it keeps the producer raw *and* rewrites the final `grep` stage, so the savings this PR would have discarded are preserved. The `ls -t | head -1` row is a real gap, but a narrower one than this PR's model — I've filed it separately as #4026, because the fix belongs in #3343's producer gating rather than in a blanket skip list.

Worth saying explicitly: the `ls` case in #2962 was `$(ls -t | head -1)`, and every command-substitution form is still correctly left alone on develop. #2962 stays closed.

Recommending we close this one as superseded by #3128/#3343 — the analysis in #2962 is what drove that work. Thanks for filing both.

PR #2970, opened 2026-07-12
fix(telemetry): honor configured database path on forget

### LINKED ISSUE #2956: telemetry forget ignores RTK_DB_PATH/config.tracking.database_path, misses redirected tracking DB
## Summary

`rtk telemetry forget` deletes the tracking database at a hardcoded default path instead of the path resolved by `get_db_path()` — the same resolver every other tracking-database read/write goes through. If a user has redirected their database via `RTK_DB_PATH` (or `config.toml`'s `tracking.database_path`), `forget` silently misses it.

## Where

`src/core/telemetry_cmd.rs`, `run_forget()` (around lines 129-135):

```rust
// Purge local tracking database (GDPR Art. 17 — right to erasure applies to local data too)
let db_path = dirs::data_local_dir()
    .unwrap_or_else(|| std::path::PathBuf::from("."))
    .join(super::constants::RTK_DATA_DIR)
    .join(super::constants::HISTORY_DB);
```

Compare with `src/core/tracking.rs`'s `get_db_path()` (lines 1220-1236), used everywhere else the tracking DB is opened:

```rust
fn get_db_path() -> Result<PathBuf> {
    // Priority 1: Environment variable RTK_DB_PATH
    if let Ok(custom_path) = std::env::var("RTK_DB_PATH") {
        return Ok(PathBuf::from(custom_path));
    }

    // Priority 2: Configuration file
    if let Ok(config) = crate::core::config::Config::load() {
        if let Some(db_path) = config.tracking.database_path {
            return Ok(db_path);
        }
    }

    // Priority 3: Default platform-specific location
    let data_dir = dirs::data_local_dir().unwrap_or_else(|| PathBuf::from("."));
    Ok(data_dir.join(RTK_DATA_DIR).join(HISTORY_DB))
}
```

`run_forget` re-implements only priority 3 of `get_db_path()`'s three-tier resolution, skipping the `RTK_DB_PATH` and `config.tracking.database_path` checks entirely.

## Impact

Any user who has redirected their tracking database (via `RTK_DB_PATH` or `config.toml`) and then runs `rtk telemetry forget` — intending to exercise their GDPR Art. 17 right to erasure over local data — will find:

- The real, redirected database is **not deleted** (the data they wanted erased is still on disk).
- The command may print "Local tracking database deleted: {default path}" even though no such file existed there, or silently report nothing to delete — misleadingly implying erasure succeeded.

This is specifically concerning given `forget`'s stated purpose is GDPR-driven local data erasure.

## Suggested fix

Have `run_forget` call the existing `get_db_path()` (or an equivalent shared resolver) instead of recomputing the default-only path inline, so `forget` always targets the same database file every other command reads from and writes to.

## Context

Found while building a from-scratch .NET port of `rtk` ([RtkSharp](https://github.com/jamesburton/rtksharp)) and diffing its behavior against this project's own logic. The .NET port's `rtk telemetry forget` has been changed to route through its equivalent of `get_db_path()` rather than replicate this inconsistency — happy to point at the relevant commit if useful as a reference, but the actual fix here is obviously up to this project's own maintainers/conventions.

Bortlesboat:
## Summary

- resolve the tracking database through the same `get_db_path()` path used by tracking reads and writes
- delete the configured `RTK_DB_PATH` or `tracking.database_path` target instead of always using the platform default
- cover deletion of a non-default database path

Closes #2956.

## Validation

- `cargo fmt --all --check`
- `cargo test deletes_tracking_database_at_resolved_path -- --nocapture`
- full `cargo test --all`: 2,418 passed; 16 existing Windows stream tests fail because their Unix helper programs are unavailable (`program not found`)

### DIFF
--- a/src/core/telemetry_cmd.rs
+++ b/src/core/telemetry_cmd.rs
@@ -131,7 +131,7 @@ fn run_forget() -> Result<()> {
     // Purge local tracking database (GDPR Art. 17 — right to erasure applies to local data too)
     let db_path = super::tracking::get_db_path()?;
     if db_path.exists() {
-        match delete_tracking_database(&db_path) {
+        match std::fs::remove_file(&db_path) {
             Ok(()) => println!("Local tracking database deleted: {}", db_path.display()),
             Err(e) => eprintln!("rtk: could not delete {}: {}", db_path.display(), e),
         }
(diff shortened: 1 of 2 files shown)


Bortlesboat:
CI note: the code-relevant checks are green (Clippy, fmt, security scans, and tests on Linux/macOS/Windows). The benchmark failure is its live mockhttp.org/json case: raw output was 35 tokens while the filtered response was 38, so the external response grew and tripped the negative-compression guard. This PR does not touch curl filtering or benchmark code.

Bortlesboat:
Closing in favor of #3044, which now covers the same #2956 database-path fix. This branch conflicts with current `develop`, and keeping both implementations open would duplicate the review lane.

PR #2981, opened 2026-07-13
fix(tracking): cap token baseline at agent harness truncation limit

felipetruman:
## Problem

Savings tracking compares the **full** raw command output against the filtered output. But agent harnesses truncate raw tool output before the model ever sees it — Claude Code caps Bash output at `BASH_MAX_OUTPUT_LENGTH` (default 30000 chars). Tokens beyond that limit were never real model input, so counting them as "saved" inflates the numbers.

Real-world impact (my 90-day `history.db`, 310k commands):

- `rtk read` on a 21MB file recorded **~5.3M input tokens** for a single command — the model would have ingested at most ~7.5k
- `rtk read` alone accounted for **88% of all claimed savings** (987.8M tokens)
- Total claimed savings: **1118.6M (79.1%)** → after capping: **12.0M (35.5%)**

The real savings are still great — they just shouldn't be overstated by ~100x.

## Fix

Cap both input and output token estimates in `TimedExecution::track` at the harness truncation limit. New helper `baseline_cap_tokens()` with resolution order:

1. `RTK_BASELINE_CAP_CHARS` env — explicit override, `0` disables capping (old behavior)
2. `BASH_MAX_OUTPUT_LENGTH` env — Claude Code's own truncation knob, already exported to child processes
3. Default `30000` chars (Claude Code's default)

Symmetric capping keeps the pct math honest: if both raw and filtered output exceed the limit, savings = 0 (the model would have seen the same truncated volume either way).

## Testing

- `cargo fmt --all` + `cargo clippy --all-targets` clean
- New unit test `test_baseline_cap_tokens` covering env resolution + disable path
- `cargo test tracking`: 15 passed
- Verified end-to-end: `rtk read <21MB file>` now records 3750/3750 tokens (with `BASH_MAX_OUTPUT_LENGTH=15000`) instead of ~5.3M

Note: existing DB rows are not migrated — users can optionally re-cap historical rows themselves; happy to add a `rtk gain --recap` migration in a follow-up if wanted.


pszymkowiak:
**Nice — this is the right fix and supersedes #1978.** Verified against the Claude Code docs: capping both operands at `BASH_MAX_OUTPUT_LENGTH` (the real Bash-tool limit) is correct, and doing `min(input,C)`/`min(output,C)` independently means an ineffective filter records `saved=0` (#1978 didn't). A few things before merge:

- 🔴 **Rebase needed.** The diff carries release-please noise — `Cargo.toml`/`Cargo.lock`/`manifest` bumping 0.42.4→0.43.0 plus 49 CHANGELOG lines. Please strip it down to just the `tracking.rs` change so it doesn't collide with release automation on `develop`.
- 🟠 **Backfill the existing rows.** New records get capped, but inflated rows already in `history.db` keep skewing `rtk gain` lifetime totals for up to 90 days (the retention window). Since `input_tokens`/`output_tokens` are stored as columns, this is a cheap one-shot `UPDATE ... WHERE input_tokens > cap` recomputing `saved_tokens`/`savings_pct` — worth adding so the headline number is correct immediately rather than self-healing over a quarter.
- 🟠 **Env-mutating test.** `test_baseline_cap_tokens` mutates process-global env vars and can race concurrent `track()` tests. Suggest a pure `cap_token_count(raw, cap)` helper tested without env.

Thanks for tackling this! 🙌

KuSh:
Closing this in favour of #3635, which caps the same token baseline in TimedExecution::track but exposes it as tracking.estimate_cap_chars with a RTK_TRACK_CAP_CHARS override, documents it, and covers it with tests. The analysis in this description is what makes the case for the change, so it is worth restating on #3635 where the work continues.

PR #2982, opened 2026-07-13
fix: report nonzero vitest process exits

### LINKED ISSUE #2874: rtk vitest run summary reports PASS/FAIL as green when the process exits non-zero due to an uncaught exception during the test run
**Version:** rtk 0.42.4 · **Platform:** Linux · **Invocation:** `rtk vitest run` (also triggered transparently via the Claude Code Bash PreToolUse hook rewriting `npx vitest run`)

## Summary

`rtk vitest run` reports a clean `PASS (N) FAIL (0)` summary — and returns the **correct non-zero exit code** — for a test run that contains real, uncaught exceptions thrown asynchronously during the run (outside the assertions of any single test). The exit code is right, but the printed summary actively says everything passed, with zero indication that anything went wrong. Since the whole point of `rtk`'s filters is to replace reading the raw log with trusting the condensed summary, this is a false-positive in the exact channel most likely to be trusted (by a human skimming the terminal, or by an LLM agent that doesn't separately check `$?`).

This is the same *class* of problem as #2317 ("Tool filters mask hard failures with benign summaries"), but a different mechanism worth tracking separately: #2317 is about a **pre-run** failure (collection/import error) being summarized too optimistically, where the real error text likely still exists somewhere in the tool's own output. This report is about a **post-assertion, in-process uncaught exception**, and — as shown below — the data RTK's summary is built from genuinely does not contain the failure at all, not even in the underlying JSON payload it consumes. So the fix isn't "surface what's already there" — it needs an out-of-band signal (the exit code) to detect that anything is wrong in the first place.

## Root cause (confirmed by inspecting the tee file)

`rtk vitest run`'s tee log (`~/.local/share/rtk/tee/<ts>_vitest_run.log`) is Vitest's own **JSON reporter** output. That JSON has top-level `success`/`numFailedTests` fields, and in this scenario they read:

```json
{"numTotalTestSuites":2,"numPassedTestSuites":2,"numFailedTestSuites":0,
 "numTotalTests":2,"numPassedTests":2,"numFailedTests":0,
 "success":true, ...}
```

`success: true`, `numFailedTests: 0` — because every individual test's *assertions* did pass. But Vitest's JSON reporter schema has **no field at all** for uncaught/unhandled exceptions that occur asynchronously during the run (e.g. in an effect or a cleanup callback, after the test that triggered them has already been marked passed). Those only show up in Vitest's **human-readable console output**, under a distinct `⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯` block, and in a separate `Errors N errors` line in the text summary — plus, critically, in the **process exit code** (non-zero).

`rtk`'s summary appears to be built purely from the JSON reporter's fields, without cross-checking them against the real exit code of the child process. So when the two disagree — JSON says `success: true`, exit code says `1` — `rtk` trusts the JSON and prints `PASS (2) FAIL (0)`, silently dropping the one signal (exit code) that would have revealed the discrepancy.

## Minimal reproduction

This is a fully self-contained, from-scratch repro (no relation to any real codebase — generic names throughout). The mechanism (a still-mounted component receiving a forced re-render, via a global-store mutation, at the exact moment a previously-mocked hook gets restored to its real multi-hook implementation) is a known, general React Testing Library + Vitest interaction — not anything unusual about the test itself.

```bash
mkdir rtk-repro && cd rtk-repro
npm init -y
npm install react@19 react-dom@19 zustand@5 vitest@2 @testing-library/react@16 jsdom@26 @vitejs/plugin-react@4
```

`vitest.config.ts`:
```ts
import { defineConfig } from 'vitest/config'
import react from '@vitejs/plugin-react'

export default defineConfig({
  plugins: [react()],
  test: { globals: true, environment: 'jsdom' },
})
```

`src/store.ts`:
```ts
import { create } from 'zustand'

interface CounterState {
  tick: number
  bump: () => void
}

export const useStore = create<CounterState>((set) => ({
  tick: 0,
  bump: () => set((s) => ({ tick: s.tick + 1 })),
}))
```

`src/useRemoteData.ts`:
```ts
import { useEffect, useState } from 'react'

// Real implementation calls 2 hooks. Tests mock this whole function so the
// component renders with ZERO internal hooks on this path during the test.
export function useRemoteData() {
  const [data, setData] = useState<string | null>(null)
  useEffect(() => {
    setData('real-data')
  }, [])
  return { data }
}
```

`src/DataWidget.tsx`:
```tsx
import { useStore } from './store'
import { useRemoteData } from './useRemoteData'

export function DataWidget() {
  const tick = useStore((s) => s.tick)
  const { data } = useRemoteData()
  return <div>tick: {tick}, data: {data ?? 'loading'}</div>
}
```

`src/DataWidget.test.tsx`:
```tsx
import { render } from '@testing-library/react'
import { afterEach, describe, expect, it, vi } from 'vitest'

import { DataWidget } from './DataWidget'
import { useStore } from './store'
import * as remoteDataModule from './useRemoteData'

describe('DataWidget', () => {
  // Registered AFTER @testing-library/react's own afterEach(cleanup) (that
  // one is registered at import time). Vitest's default sequence.hooks =
  // "stack" runs afterEach in LIFO order, so THIS afterEach runs BEFORE
  // RTL's cleanup on every test boundary — while the previous test's
  // component instance is still mounted.
  afterEach(() => {
    vi.restoreAllMocks()        // restores useRemoteData to its real 2-hook impl
    useStore.getState().bump()  // forces a re-render of the still-mounted instance
  })

  it('renders with a mocked (0-hook) data source', () => {
    vi.spyOn(remoteDataModule, 'useRemoteData').mockReturnValue({ data: 'stub' })
    const { getByText } = render(<DataWidget />)
    expect(getByText(/stub/)).toBeTruthy()
  })

  it('renders again in a second test', () => {
    vi.spyOn(remoteDataModule, 'useRemoteData').mockReturnValue({ data: 'stub-2' })
    const { getByText } = render(<DataWidget />)
    expect(getByText(/stub-2/)).toBeTruthy()
  })
})
```

## Observed

**Raw Vitest** (`npx vitest run`) — correctly surfaces the problem, both in text and exit code:
```
⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯
Error: Rendered more hooks than during the previous render.
 ❯ updateWorkInProgressHook .../react-dom-client.development.js:7848:17
 ❯ Object.useState .../react.development.js:1263:34
 ❯ Module.useRemoteData src/useRemoteData.ts:7:27
 ❯ DataWidget src/DataWidget.tsx:6:20
...
 Test Files  1 passed (1)
      Tests  2 passed (2)
     Errors  2 errors
```
```bash
$ npx vitest run; echo "exit=$?"
exit=1
```

**`rtk vitest run`** — exit code is correct, but the printed summary is a clean false-positive:
```
$ rtk vitest run; echo "exit=$?"
PASS (2) FAIL (0)
[full output: ~/.local/share/rtk/tee/1783452626_vitest_run.log]
exit=1
```

The tee file referenced in that output is the JSON blob quoted in "Root cause" above — `"success":true` — which is where `PASS (2) FAIL (0)` comes from.

## Why it matters

`rtk`'s value proposition is precisely "read this instead of the raw log." An agent or a human who does that — and doesn't independently check `$?` — walks away believing the suite is green. In practice this means a real, reproducible bug (this exact "more hooks than previous render" pattern, or any other async/uncaught exception during a run) can sit undetected through several rounds of "run the tests, everything's green" before someone happens to check the exit code directly.

## Suggested fix (in priority order)

1. **Minimum:** when the underlying process's exit code is non-zero but the JSON reporter's `success`/`numFailedTests` say otherwise, don't print a bare `PASS` summary — surface the discrepancy explicitly, e.g. `PASS (2) FAIL (0) — but process exited 1 (uncaught exception — see tee log)`.
2. **Better:** in addition to the JSON reporter, scan the raw stdout/stderr for Vitest's `Errors N errors` line / `⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯` blocks (the same "raw-on-failure" principle proposed in #2317) and include a one-line pointer to the first uncaught exception's message + file:line when present, regardless of what the JSON reporter says.
3. **General principle:** for any tool filter whose summary is derived from a structured reporter payload, the *first* check before trusting that payload should be "does the process's actual exit code agree with what this payload implies?" — the JSON reporter is not guaranteed to be a complete substitute for the real exit signal (as shown here, it structurally cannot be, for this class of failure).

Happy to test a patch against this same repro if useful.

ByteBaker:
## Summary

- Fix `rtk vitest run` reporting a false-green summary when Vitest exits nonzero despite its JSON reporter listing no failed tests.
- Preserve existing output for normal passes and reported test failures while surfacing the process exit discrepancy.

Closes #2874.

## Test plan

- [x] Manual testing: ran `rtk vitest run` in a separate real Vitest project for both a normal pass and an uncaught asynchronous error

### DIFF
--- a/src/cmds/js/vitest_cmd.rs
+++ b/src/cmds/js/vitest_cmd.rs
@@ -253,6 +253,7 @@ pub fn run_test(command: &Commands, args: &[String], verbose: u8) -> Result<i32>
         &combined,
         passthrough_requested,
         verbose,
+        result.exit_code,
     );
     let tee_label = format!("{}_run", framework);
 
@@ -334,6 +335,7 @@ fn format_test_output(
     combined: &str,
     passthrough_requested: bool,
     verbose: u8,
+    exit_code: i32,
 ) -> FormattedTestOutput {
     if passthrough_requested {
         return format_passthrough_output(combined);
@@ -346,19 +348,29 @@ fn format_test_output(
             if verbose > 0 {
                 eprintln!("{} run (Tier 1: Full JSON parse)", framework);
             }
-            FormattedTestOutput::new(data.format(mode))
+            FormattedTestOutput::new(format_parsed_test_output(&data, mode, exit_code))
         }
         ParseResult::Degraded(data, warnings) => {
             if verbose > 0 {
                 emit_degradation_warning(framework, &warnings.join(", "));
             }
-            FormattedTestOutput::new(data.format(mode))
+            FormattedTestOutput::new(format_parsed_test_output(&data, mode, exit_code))
         }
         ParseResult::Passthrough(_) => {
             emit_passthrough_warning(framework, "All parsing tiers failed");
             format_passthrough_output(stdout)
         }
     }
+}
+
+fn format_parsed_test_output(data: &TestResult, mode: FormatMode, exit_code: i32) -> String {
+    let mut output = data.format(mode);
+    if exit_code != 0 && data.failed == 0 {
+        output.push_str(&format!(
+            "\n[FAIL] Process exited with code {exit_code} despite reporter listing no failed tests"
+        ));
+    }
+    output
 }
 
 fn format_passthrough_output(raw: &str) -> FormattedTestOutput {
@@ -420,6 +432,21 @@ where
 mod tests {
     use super::*;
 
+    const ALL_PASS_JSON: &str = r#"{
+        "numTotalTests": 2,
+        "numPassedTests": 2,
+        "numFailedTests": 0,
+        "numPendingTests": 0,
+        "testResults": []
+    }"#;
+    const FAILED_TEST_JSON: &str = r#"{
+        "numTotalTests": 2,
+        "numPassedTests": 1,
+        "numFailedTests": 1,
+        "numPendingTests": 0,
+        "testResults": []
+    }"#;
+
     fn args(values: &[&str]) -> Vec<String> {
         values.iter().map(|value| value.to_string()).collect()
     }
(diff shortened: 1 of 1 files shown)


ByteBaker:
@pszymkowiak please help review.

PR #2986, opened 2026-07-14
fix(benchmark): replace live HTTP requests with local fixtures

### LINKED ISSUE #2844: fix(benchmark): curl/wget tests are non-deterministic (live HTTP requests)
## Problem

The benchmark CI gate (`scripts/benchmark.sh`) intermittently fails on `curl json`, `curl text`, and `wget` tests with negative token savings (e.g., `35 → 38 (-8%)` or `8 → 40 (-400%)`).

This blocks unrelated PRs — e.g., #2836 (grep context separator fix) showed a `curl json` failure despite touching no curl code.

## Root Cause

`scripts/benchmark.sh` runs the raw command and RTK command as **separate invocations** (lines 44-45):

```bash
unix_out=$(eval "$unix_cmd" 2>/dev/null || true)
rtk_out=$(eval "$rtk_cmd" 2>/dev/null || true)
```

For curl/wget tests (lines 349-350, 358), this means **two separate HTTP requests** to `httpbin.org`:

```bash
bench "curl json" "curl -s https://httpbin.org/json" "$RTK curl https://httpbin.org/json"
bench "curl text" "curl -s https://httpbin.org/robots.txt" "$RTK curl https://httpbin.org/robots.txt"
bench "wget"      "wget -qO- https://httpbin.org/json"     "$RTK wget https://httpbin.org/json"
```

Response variations between the two calls (timestamps, headers, content-length, server-side changes) cause token count mismatches. The never-worse guard inside RTK (#2554) works correctly within a single execution, but the benchmark's cross-execution comparison is inherently non-deterministic for live endpoints.

## Prior Art

Same class of issue as:
- **#2551** — identified `curl` emitting more tokens than raw on tiny responses
- **#2554** — added `core::guard::never_worse()` to solve the within-RTK problem
- **#2751** — parser fix PR that also had this exact benchmark flakiness; [commenter noted](https://github.com/rtk-ai/rtk/pull/2751#issuecomment-4865304124) the curl test is non-deterministic

The never-worse guard (#2554) solved the within-RTK problem. But the benchmark itself still compares outputs from two separate HTTP requests, making it structurally flaky.

## Proposed Fix

Replace live `httpbin.org` calls with local fixtures, following the pattern already used by the `json` and `log` benchmark sections (lines 228-246, 286-302):

```bash
# Create fixture (matching httpbin.org/json response)
cat > /tmp/rtk_bench_curl.json << 'CURLEOF'
{"slideshow":{"author":"Yours Truly","date":"date of publication","slides":[...]}}
CURLEOF

bench "curl json" "cat /tmp/rtk_bench_curl.json" "$RTK curl file:///tmp/rtk_bench_curl.json"
```

This makes the test deterministic while still exercising the curl filter logic.

C0deRatoR:
## Summary

- replace the live curl/wget benchmark endpoints with deterministic local fixtures
- serve the fixtures on loopback using an OS-assigned port and clean up the server on exit
- keep exercising real HTTP behavior without cross-request response variance

Closes #2844

## Test plan

- [x] `bash -n scripts/benchmark.sh`
- [x] Manual benchmark inspection: curl JSON/text and wget all used `127.0.0.1`, returned output, and produced no negative savings

The full local benchmark continued past the changed sections, then reported an unrelated warm-cache `cargo build` negative comparison; the curl/wget regression itself passed.

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -342,21 +342,92 @@ section "wc"
 bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 
 # ===================
-# curl
+# curl / wget
 # ===================
-section "curl"
-if command -v curl &> /dev/null; then
-  bench "curl json" "curl -s https://mockhttp.org/json" "$RTK curl https://mockhttp.org/json"
-  bench "curl text" "curl -s https://mockhttp.org/robots.txt" "$RTK curl https://mockhttp.org/robots.txt"
-fi
+if command -v python3 &> /dev/null && { command -v curl &> /dev/null || command -v wget &> /dev/null; }; then
+  HTTP_FIXTURE=$(mktemp -d)
+  HTTP_PORT_FILE="$HTTP_FIXTURE/port"
 
-# ===================
-# wget
-# ===================
-if command -v wget &> /dev/null; then
-  section "wget"
-  bench "wget" "wget -qO- https://mockhttp.org/json" "$RTK wget https://mockhttp.org/json"
-  rm -f json 2>/dev/null
+  cat > "$HTTP_FIXTURE/json" << 'JSONEOF'
+{
+  "slideshow": {
+    "author": "Yours Truly",
+    "date": "date of publication",
+    "slides": [
+      {
+        "title": "Wake up to WonderWidgets!",
+        "type": "all"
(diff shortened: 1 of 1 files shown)


C0deRatoR:
Closing since #3170 has now landed an equivalent fix on `develop`. Thanks!

PR #2992, opened 2026-07-14
fix(discover): report hook-rewritten commands as already using RTK

sammyjdev:
## Problem

`rtk discover` parses Claude Code transcripts reading only the `tool_use` block, which keeps the command as originally written by the model (`git status`). When the PreToolUse hook (`rtk hook claude`) rewrites it to `rtk git status`, the executed command is recorded only in a `hook_success` attachment (`attachment.toolUseID` + `updatedInput.command`). Result: every auto-rewritten command was counted as a **missed opportunity**, punishing exactly the sessions that already use the hook.

Observed on real data (30 days, 1036 sessions): reported 1.5% already-using-RTK vs the real 14.7%.

## Fix

`extract_commands` now collects hook-rewrite attachments in the same single pass, pairs `attachment.toolUseID` with the `tool_use` id, and substitutes the rewritten command before classification — so the report reflects what actually ran. Malformed stdout or unmatched toolUseID keeps the original command.

## Tests

- 3 new unit tests: rewrite applied, unmatched toolUseID keeps original, malformed stdout keeps original (TDD, red first)
- `cargo fmt + clippy --all-targets + test` clean (guard_integration_test failures observed locally are a sandbox/GPG artifact, pass outside sandbox 6/6)
- End-to-end: `rtk discover --all` on real transcripts moves Already-using-RTK from 624 (1.5%) to 6037 (14.7%)

https://claude.ai/code/session_015ygD25ujysbWGFtrVuaJog


sammyjdev:
Closing: the fix lives on in a downstream fork (sammyjdev/rtkx); feel free to cherry-pick if useful.

PR #2997, opened 2026-07-14
fix(stream): decode lossily instead of dropping lines on invalid UTF-8

### LINKED ISSUE #2994: Filters silently drop non-UTF-8 lines: shell/build errors vanish while exit code stays non-zero (non-English Windows; likely root cause of #1936)
## Summary

RTK's filters **silently discard lines that are not valid UTF-8**. On non-English Windows, system binaries (`cmd.exe`, `tree.com`, ...) emit OEM-encoded bytes (cp850 here, pt-BR; cp437, GBK, etc. elsewhere). The result is **silent data loss on the error path**: a command fails with a non-zero exit code, RTK prints no error at all, and the caller — a human or an AI agent — concludes it succeeded.

This is *not* mojibake (#2452, where GBK text is mis-decoded but still visible). Here the line **disappears entirely**, and no `tee` rescue log is written.

I believe this is the shared root cause of #1936 (`rtk tree` empty) and is closely related to #2452 and #1475.

**Environment:** Windows 11 pt-BR (console cp850 / OEM 850), `rtk 0.43.0` (x86_64-pc-windows-msvc), Git Bash + PowerShell.

## The smoking gun

`rtk pipe` states the mechanism out loud — the other filters do the same thing, silently:

```
$ node enc.js 2>&1 | rtk pipe
rtk: Failed to read stdin: stream did not contain valid UTF-8
```

## Minimal reproduction (one byte is the only variable)

`enc.js` — writes two structurally identical lines to stderr. Line A is pure ASCII; line B is the same shape but contains byte `0xE3` (`ã` in cp850):

```js
process.stderr.write("ERRO A: linha ascii pura\n");
process.stderr.write(Buffer.from([69,82,82,79,32,66,58,32,108,105,110,104,97,32,99,111,109,32,110,227,111,10]));
process.exit(1);
```

`package.json`:

```json
{ "name":"e", "version":"1.0.0", "scripts": { "enc": "node enc.js" } }
```

**Raw — both lines present:**

```
$ npm run enc
ERRO A: linha ascii pura
ERRO B: linha com nM-co        <- 0xE3 shown via `cat -v`
```

**Through RTK — line B is gone, exit code is still 1:**

```
$ rtk npm run enc
> node enc.js
ERRO A: linha ascii pura
```

Only the invalid-UTF-8 line is dropped. Everything else is preserved.

## Real-world impact (how I hit this)

Running `npm run build` in a project whose `node_modules` isn't installed:

```
$ npm run build            # raw
> react-scripts build
'react-scripts' não é reconhecido como um comando interno
ou externo, um programa operável ou um arquivo em lotes.
```

```
$ rtk npm run build        # through RTK
> react-scripts build
```

**The entire error message is gone.** Exit code is non-zero, output looks like a clean no-op, and nothing is written to `~/AppData/Local/rtk/tee/`. Any Windows error emitted by the shell in a non-English locale — missing binary, access denied, path not found — vanishes this way.

To be clear about what *does* work: errors from tools that emit UTF-8 (node stack traces, pytest, ruff, mypy) are preserved correctly, with file, line and message. The failure is specific to non-UTF-8 bytes.

## Likely shared root cause with #1936

`rtk tree` returns empty with exit code 0 (#1936). Windows' `tree.com` emits box-drawing characters and accented headers in OEM encoding. Piping its raw output into RTK produces the same error as above:

```
$ cmd /c "tree /F src" | rtk pipe
rtk: Failed to read stdin: stream did not contain valid UTF-8
```

So every line of `tree.com` output is invalid UTF-8 → every line is dropped → empty output. Note `chcp 65001` does **not** help: `tree.com` writes OEM bytes when its stdout is a pipe regardless of the console code page.

This also rhymes with #1475 (`rtk ls` empty under non-English locales) and #2878 (`rtk prettier` silently hiding failures) — same family: **silent failure with exit 0 instead of a surfaced error**.

## Why this is severe

RTK's main use case is feeding AI coding agents. A dropped error is worse than a verbose one: the agent sees a command that "produced no errors," reports success, and moves on. Token savings are worthless if the failure signal is what gets compressed away.

## Suggested fix

1. **Never drop undecodable lines.** Decode lossily (`String::from_utf8_lossy`, U+FFFD) or fall back to the OEM/ANSI code page via `encoding_rs` + `GetConsoleOutputCP()` / `GetOEMCP()` (as proposed in #2452). A garbled line still carries the error; a missing line doesn't.
2. **Fail loud, not silent.** If output can't be decoded, emit a warning to stderr and pass the raw bytes through — `rtk pipe` already does the first half of this.
3. **Suggested invariant:** when the child's exit code is non-zero, RTK should never produce output with no error content. That property would have caught this, #1936, #1475 and #2878 in a single test.

albatrossflyon-coder:
## Summary

Fixes #2994.

`run_streaming`'s six call sites all read child stdout/stderr with:

```rust
BufReader::new(stdout).lines().map_while(Result::ok)
```

`BufRead::lines()` returns `Err` for a line that isn't valid UTF-8 (OEM/ANSI
bytes from a non-English-locale Windows tool, per the issue's cp850 repro).
`map_while` stops iteration at the first `None` — so one non-UTF-8 line
doesn't just get dropped, it silently discards every line emitted *after*
it too. A failing build whose error text happens to contain one non-UTF-8
byte can end up looking like a clean, silent success even though the exit
code is non-zero.

## Fix

Added `read_lines_lossy()`, which reads raw bytes via `BufRead::read_until`
and decodes each line with `String::from_utf8_lossy` (invalid bytes become
U+FFFD) instead of erroring out of the iterator. Replaced all 6
`.lines().map_while(Result::ok)` call sites in `run_streaming` with it —
this is the single shared path every filter routes through, so the fix
covers all callers at once.

`rtk pipe` (`cmds/system/pipe_cmd.rs`) already fails loudly on invalid
UTF-8 stdin rather than silently, per the issue's own note — left as-is,
out of scope here.

## Testing

- `cargo test`: 2448 passed, 8 ignored, 0 failed (no regressions)
- 4 new unit tests on `read_lines_lossy` directly, including one that
  reproduces the exact failure mode from the issue: an ASCII line, a line
  with an embedded invalid byte, then another ASCII line — asserts all 3
  lines survive instead of the third silently vanishing
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo fmt --check`: clean

### DIFF
--- a/src/core/stream.rs
+++ b/src/core/stream.rs
@@ -1,10 +1,39 @@
 use anyhow::{Context, Result};
-use std::io::{self, BufRead, BufReader, BufWriter, Write};
+use std::io::{self, BufRead, BufReader, BufWriter, Read, Write};
 use std::process::{Command, Stdio};
 use std::sync::mpsc;
 
 #[cfg(test)]
 use regex::Regex;
+
+/// Read `reader` line by line, decoding each line lossily (invalid UTF-8
+/// bytes become U+FFFD) instead of erroring.
+///
+/// `BufRead::lines()` returns `Err` for a non-UTF-8 line, and callers
+/// commonly chain `.map_while(Result::ok)` to skip bad lines — but
+/// `map_while` stops at the *first* `None`, so one invalid-UTF-8 line (e.g.
+/// OEM/ANSI bytes from a non-English-locale Windows tool) silently discards
+/// every line after it too, not just the bad one. This reads raw bytes and
+/// never fails on the source encoding, so a garbled line still surfaces
+/// instead of vanishing along with everything downstream of it.
+fn read_lines_lossy(reader: impl Read) -> impl Iterator<Item = String> {
+    let mut reader = BufReader::new(reader);
+    std::iter::from_fn(move || {
+        let mut buf = Vec::new();
+        match reader.read_until(b'\n', &mut buf) {
+            Ok(0) | Err(_) => None,
+            Ok(_) => {
+                if buf.last() == Some(&b'\n') {
+                    buf.pop();
+                    if buf.last() == Some(&b'\r') {
+                        buf.pop();
+                    }
+                }
+                Some(String::from_utf8_lossy(&buf).into_owned())
+            }
+        }
+    })
+}
 
 pub trait StreamFilter {
     fn feed_line(&mut self, line: &str) -> Option<String>;
@@ -298,10 +327,7 @@ pub fn run_streaming(
             Some(std::thread::spawn(move || {
                 let mut writer = BufWriter::new(child_stdin);
                 let stdin_handle = io::stdin();
-                for line in BufReader::new(stdin_handle.lock())
-                    .lines()
-                    .map_while(Result::ok)
-                {
+                for line in read_lines_lossy(stdin_handle.lock()) {
                     if let Some(out) = filter.feed_line(&line) {
                         if writeln!(writer, "{}", out).is_err() {
                             break;
@@ -340,7 +366,7 @@ pub fn run_streaming(
         let (tx, rx) = mpsc::channel();
         let tx_out = tx.clone();
         let stdout_thread = std::thread::spawn(move || {
-            for line in BufReader::new(stdout).lines().map_while(Result::ok) {
+            for line in read_lines_lossy(stdout) {
                 if tx_out.send(StreamLine::Stdout(line)).is_err() {
                     break;
                 }
@@ -348,7 +374,7 @@ pub fn run_streaming(
         });
         let tx_err = tx;
         let stderr_thread = std::thread::spawn(move || {
-            for line in BufReader::new(stderr).lines().map_while(Result::ok) {
+            for line in read_lines_lossy(stderr) {
                 if tx_err.send(StreamLine::Stderr(line)).is_err() {
                     break;
                 }
@@ -417,7 +443,7 @@ pub fn run_streaming(
         let stderr_thread = std::thread::spawn(move || -> String {
             let mut raw_err = String::new();
             let mut capped = false;
-            for line in BufReader::new(stderr).lines().map_while(Result::ok) {
+            for line in read_lines_lossy(stderr) {
                 if raw_err.len() + line.len() < RAW_CAP {
                     raw_err.push_str(&line);
                     raw_err.push('\n');
@@ -436,7 +462,7 @@ pub fn run_streaming(
                 FilterMode::Passthrough => unreachable!("handled by early-return above"),
                 FilterMode::Streaming(_) => unreachable!("handled by is_streaming branch"),
                 FilterMode::Buffered(filter_fn) => {
-                    for line in BufReader::new(stdout).lines().map_while(Result::ok) {
+                    for line in read_lines_lossy(stdout) {
                         if raw_stdout.len() + line.len() < RAW_CAP {
                             raw_stdout.push_str(&line);
                             raw_stdout.push('\n');
@@ -461,7 +487,7 @@ pub fn run_streaming(
                     }
                 }
                 FilterMode::CaptureOnly => {
-                    for line in BufReader::new(stdout).lines().map_while(Result::ok) {
+                    for line in read_lines_lossy(stdout) {
                         if raw_stdout.len() + line.len() < RAW_CAP {
                             raw_stdout.push_str(&line);
                             raw_stdout.push('\n');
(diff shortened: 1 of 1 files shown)


albatrossflyon-coder:
Hi — just checking in on this one, it's been open about 9 days with no review yet. Happy to make any changes if there's feedback, or let me know if anything's blocking it. Thanks for maintaining rtk!

src/core/stream.rs
```diff
@@ -1,11 +1,40 @@
...
+/// OEM/ANSI bytes from a non-English-locale Windows tool) silently discards
+/// every line after it too, not just the bad one. This reads raw bytes and
+/// never fails on the source encoding, so a garbled line still surfaces
+/// instead of vanishing along with everything downstream of it.
+fn read_lines_lossy(reader: impl Read) -> impl Iterator<Item = String> {
+    let mut reader = BufReader::new(reader);
+    std::iter::from_fn(move || {
+        let mut buf = Vec::new();
+        match reader.read_until(b'\n', &mut buf) {
+            Ok(0) | Err(_) => None,
```
KuSh: You’re treating a real I/O error the same as a clean EOF. A genuine read failure mid-stream silently truncates the output, so this would introduce another ‘failing build looks clean’ bug while fixing the invalid UTF-8 one

src/core/stream.rs
```diff
@@ -1,11 +1,40 @@
...
+/// commonly chain `.map_while(Result::ok)` to skip bad lines — but
+/// `map_while` stops at the *first* `None`, so one invalid-UTF-8 line (e.g.
+/// OEM/ANSI bytes from a non-English-locale Windows tool) silently discards
+/// every line after it too, not just the bad one. This reads raw bytes and
+/// never fails on the source encoding, so a garbled line still surfaces
+/// instead of vanishing along with everything downstream of it.
+fn read_lines_lossy(reader: impl Read) -> impl Iterator<Item = String> {
+    let mut reader = BufReader::new(reader);
+    std::iter::from_fn(move || {
+        let mut buf = Vec::new();
```
KuSh: You’re allocating a fresh `Vec` on every iteration instead of reusing one buffer with `.clear()`. Please don’t, since this streaming path is hot

src/core/stream.rs
```diff
@@ -1,11 +1,40 @@
...
+/// bytes become U+FFFD) instead of erroring.
+///
+/// `BufRead::lines()` returns `Err` for a non-UTF-8 line, and callers
+/// commonly chain `.map_while(Result::ok)` to skip bad lines — but
+/// `map_while` stops at the *first* `None`, so one invalid-UTF-8 line (e.g.
+/// OEM/ANSI bytes from a non-English-locale Windows tool) silently discards
+/// every line after it too, not just the bad one. This reads raw bytes and
+/// never fails on the source encoding, so a garbled line still surfaces
+/// instead of vanishing along with everything downstream of it.
+fn read_lines_lossy(reader: impl Read) -> impl Iterator<Item = String> {
```
KuSh, line 19: A cleaner, better version would be:

```rs
fn read_buffered_lossy_lines(reader: impl Read) -> impl Iterator<Item = String> {
    BufReader::new(reader).split(b'\n').filter_map(|res| {
        let mut buf = match res {
            Ok(buf) => buf,
            Err(e) => {
                eprintln!("rtk: stream read error: {}", e);
                return None;
            }
        };
        if buf.last() == Some(&b'\r') {
            buf.pop();
        }
        Some(String::from_utf8_lossy(&buf).into_owned())
    })
}
```

This reuses `std` instead of reimplementing it, except for the CR-stripping part. It also avoids allocating on every iteration and naturally distinguishes clean EOF from I/O errors.

KuSh (changes requested):

albatrossflyon-coder:
Thanks for the review — pushed a fix addressing both issues in `ae5d1ae`:

- `Ok(0) | Err(_) => None` was treating a genuine I/O error the same as clean EOF, silently truncating output on a real read failure. Fixed by switching to your suggested `BufReader::split(b'\n')` approach, which surfaces the error via `eprintln!` instead of conflating it with EOF.
- The per-iteration `Vec::new()` allocation is gone with the same change — now reusing std's own buffered split instead of hand-rolling the read loop.

Added a test with a custom `Read` impl that yields good lines then a real I/O error, confirming the already-read lines are preserved and the error path doesn't panic or hang.

Verified clean against `cargo fmt --all -- --check`, `cargo clippy --all-targets`, and the full `cargo test --all` suite (2,449 tests, 0 failed) before pushing.

KuSh (approved):
Thanks LGTM!

PR #3002, opened 2026-07-15
fix(err,test): preserve argument quoting and exit codes in rtk err/test

### LINKED ISSUE #2985: rtk err loses argument quoting (breaks node -e / python3 -c) and exits 0 on failure
## Description

`rtk err <cmd>` appears to re-join its argv into a single string and hand it to `sh -c`, which loses the original argument quoting. Any argument containing shell metacharacters — most commonly `node -e '...'` or `python3 -c '...'` — is re-split by the shell and fails with a syntax error.

Worse, the failure is **reported as exit 0**, so a caller (or an agent) sees a "successful" run that never actually executed the command.

## Repro steps

```bash
rtk err node -e 'console.log("a")'
```

Actual:

```
sh: -c: line 0: syntax error near unexpected token `('
sh: -c: line 0: `node -e console.log("a")'
```

```bash
echo $?   # -> 0
```

Note the echoed command in the error: the single quotes around `console.log("a")` are gone, so `sh` sees the parentheses as unquoted metacharacters.

Running the same command directly works fine:

```bash
node -e 'console.log("a")'   # -> a
```

## Expected

Two things:

1. `rtk err` should preserve argument boundaries — exec the argv directly rather than reconstructing a shell string (or shell-quote each argument before passing to `sh -c`).
2. It should propagate the child's exit code. A command that failed to run must not exit 0.

## Actual

Quoting is lost, the command never runs, and the exit status is 0.

## Impact

`rtk err` is the recommended wrapper for commands RTK has no dedicated filter for — which is exactly the population of commands most likely to carry `-e` / `-c` / quoted arguments. The silent exit-0 makes this especially bad for agents, which will treat the failure as a success.

## Version / environment

- rtk 0.43.0, macOS arm64 (Darwin 24.6.0)

## Note

The exit-code half may share a root cause with #2974 (`pnpm`/`eslint` exit code swallowed); the quoting loss looks independent.

albatrossflyon-coder:
## Summary

Fixes #2985.

`rtk err <cmd>` and `rtk test <cmd>` build their `sh -c`/`cmd /C` command string with `argv.join(" ")`. The OS already splits argv into correct elements by the time this process sees them, but the naive join discards that boundary information. An argument containing shell metacharacters — `node -e 'console.log("a")'`, `python3 -c "print(...)"` — gets flattened into the join, and the shell re-splits it on those same characters when re-parsing the string.

On the exit-code half: I don't think exit-code propagation itself is actually broken — `run_streaming` correctly reads and returns the child's real exit status (verified this is the same code path I already fixed for #2994 in #2997). The exit-0 symptom in the issue's repro is what happens when there's no *child* failure to propagate: the shell's own syntax-error diagnostic looks like a command failure, but the target command was never actually invoked in a form the shell could execute — there's nothing non-zero to report because nothing that ran actually failed. Fixing the quoting fixes this too, verified below.

## Fix

Added `quote_shell_arg`/`shell_quote_join` to re-quote each argument before joining, so the shell reconstructs the original argument boundaries instead of re-splitting a flattened string.

**Windows needed one extra step beyond quoting.** `std::process::Command`'s own argument-escaping was re-escaping the already-quoted command string a *second* time when spawning `cmd.exe` — corrupting it before `cmd.exe` ever saw it. Using [`raw_arg`](https://doc.rust-lang.org/std/os/windows/process/trait.CommandExt.html#tymethod.raw_arg) (passes the string through untouched, no extra escaping layer) for the Windows `/C` invocation fixes that. Without it I could reproduce a *different* broken-quoting failure than the one reported — same root cause, different platform-specific escaping bug.

## Verified live

```
$ rtk err node -e 'console.log("a")'
$ echo $?
0
```

A genuinely failing command with quotes, semicolons, and parens all in one argument:
```
$ rtk err node -e 'console.log("before"); process.exit(3)'
[FAIL] Command failed (exit code: 3)
  before
$ echo $?
3
```

Also confirmed `python -c "print('hello world')"` and that legitimate shell features (a plain `rtk err` invocation without special characters) are unaffected.

## Testing

- New unit tests: `quote_shell_arg` leaves safe args unquoted / quotes args with metacharacters / quotes args with spaces; `shell_quote_join` preserves argument boundaries (the exact regression scenario) / handles empty argv
- `cargo test`: 2449 passed, 8 ignored, 0 failed
- `cargo clippy --all-targets -- -D warnings`: clean

## Note

I tested this on Windows (not the macOS environment in the original report), so I can't directly confirm the POSIX `sh -c` branch against a real POSIX shell — `quote_shell_arg`'s non-Windows path uses the standard single-quote-wrap-with-`'\''`-escape technique, which is the textbook-correct approach for this, but flagging that a maintainer with a Unix box handy might want to double check it directly.

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -104,9 +104,21 @@ impl StreamFilter for ErrorStreamFilter {
 
 fn build_shell_command(command: &str) -> Command {
     if cfg!(target_os = "windows") {
-        let mut c = Command::new("cmd");
-        c.args(["/C", command]);
-        c
+        // std::process::Command's default arg-escaping on Windows would
+        // re-escape our already-quoted `command` string a second time when
+        // building the command line for cmd.exe itself, corrupting it.
+        // raw_arg passes it through untouched, so only our own
+        // quote_shell_arg escaping (targeting cmd.exe's /C parsing) and the
+        // child program's own argv parsing are in play -- not a third layer.
+        #[cfg(windows)]
+        {
+            use std::os::windows::process::CommandExt;
+            let mut c = Command::new("cmd");
+            c.raw_arg("/C").raw_arg(command);
+            c
+        }
+        #[cfg(not(windows))]
+        unreachable!()
     } else {
         let mut c = Command::new("sh");
         c.args(["-c", command]);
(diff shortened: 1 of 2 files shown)


ecuthiell:
Thanks for working on #2985. I independently reproduced the same failure on native Windows through `rtk test`: one already-tokenized regex argument containing `|` was flattened, reparsed by `cmd.exe`, and treated as a pipeline.

I think direct execution is the stronger contract here than reconstructing and shell-quoting a command string. `err`, `test`, and `summary` already receive tokenized argv from Clap; passing those tokens to `Command::new(program).args(args)`:

- preserves `OsString` argument boundaries without a second parser;
- avoids platform-specific shell quoting and double-escaping rules;
- keeps literal metacharacters as data rather than implicit shell syntax; and
- leaves intentional shell behavior explicit through `rtk run -c` or an explicitly supplied shell child.

I prepared a complete focused patch against current `develop` at [`53323c8512999ae20033c4ff8d62db7a3b4d2b02`](https://github.com/ecuthiell/rtk/commit/53323c8512999ae20033c4ff8d62db7a3b4d2b02). It:

- changes `err`, `test`, and `summary` to retain `Vec<OsString>` and execute it directly;
- centralizes child construction while keeping the lossy display string diagnostic-only;
- adds native-Windows and Unix real-process regressions that exercise all three public subcommands, require exactly one literal pipe-containing argument, and verify exact exit-code propagation; and
- documents the direct-argv contract and explicit shell escape hatches.

Validation on Windows with Rust 1.97.1:

- `cargo fmt --all --check`
- focused process regression: all three subcommands passed
- non-baseline suite: 2,468 passed, 8 ignored
- full suite: the same 11 `core::stream` Unix-command-assumption failures as the unmodified Windows base; no additional failures

Before opening another overlapping PR, would you prefer to incorporate the direct-execution patch into this PR, or would a maintainer prefer a fresh focused PR against current `develop`? I am happy to follow either route. The separate hook/rewrite behavior in #3185 is intentionally out of scope.

_codex-gpt-5.6-sol-ultra on behalf of Ewen Cuthiell_

KuSh:
Closing this in favour of #4125, which takes the same diagnosis to a different conclusion. Thank you for the write-up — the analysis here is right, and the part that took care to establish is the part that ended up mattering.

You identified that `argv.join(" ")` throws away boundaries the OS had already resolved, and that the exit-code symptom in #2985 was downstream of the quoting rather than a separate bug in exit-code propagation. Both hold. #4125 reaches the same place by removing the reconstruction instead of correcting it: `rtk run|err|test|summary` execute the program directly, and a script becomes explicit via `rtk run -c '<script>'` or `--shell <shell> '<script>'`. That covers `summary` and `run` as well as `err`/`test`, and it means there is no re-quote/re-split round trip left to get right — including on Windows, where your `raw_arg` note correctly identifies that `cmd.exe`'s parsing is a second layer on top of `CommandLineToArgvW`.

One piece of this PR is better than what #4125 currently has, and I have said so on that PR: your `quote_shell_arg` decides with an allowlist (`is_ascii_alphanumeric` plus `_-./:=@`), where #4125's display-quoting uses a denylist of whitespace and quote characters and so leaves `*`, `$VAR`, `;`, `&&`, `|` and backticks bare in the labels it records. An allowlist is the right shape for that, and yours is the one I have pointed them at.

If you disagree that removing the shell is the better answer here, say so and I will reopen — the argument is worth having on its merits, and #4125 has not merged.

PR #3000, opened 2026-07-14
fix(init): --agent cursor installs Cursor only and creates ~/.cursor

rNoz:
## Summary
Two defects made `rtk init -g --agent cursor` fail / pollute the home dir:
1. It fell through to the shared init path with `install_claude` derived only from `!opencode`, so it created `RTK.md`, `CLAUDE.md`, and patched `settings.json` under `~/.claude` in addition to patching `.cursor/hooks.json`. On machines without `~/.claude` it failed outright: `Failed to create temp file in ~/.claude: No such file or directory`.
2. Even after removing the `.claude` writes, the Cursor path still failed on a fresh machine because `install_cursor_hooks` never created `~/.cursor` before `atomic_write` placed its temp file there.

Fixes:
- Gate `install_claude` on the absence of a sibling agent target (Cursor, Windsurf, Cline), so only the default or an explicit Claude target installs Claude Code files.
- Create `~/.cursor` (non-dry-run) before patching `hooks.json`.
- Should complete/close #2097 (same root cause as the previously-closed #1465)

## Root cause
- `src/main.rs`, `Commands::Init` handler: `install_claude = !opencode`. Cursor is the only sibling agent that reaches the shared init path (Windsurf/Cline short-circuit earlier in `hooks::init::run`), so it incorrectly triggered the Claude install branch.
- `src/hooks/init.rs`, `install_cursor_hooks`: wrote `hooks.json` via `atomic_write` without ensuring the parent directory exists.

## Fix
```rust
// src/main.rs
let install_cursor = agent == Some(AgentTarget::Cursor);
let install_windsurf = agent == Some(AgentTarget::Windsurf);
let install_cline = agent == Some(AgentTarget::Cline);
let install_claude =
    !opencode && !install_cursor && !install_windsurf && !install_cline;

// src/hooks/init.rs (install_cursor_hooks)
if !dry_run {
    fs::create_dir_all(&cursor_dir)?;
}
```

## Test plan
- `cargo fmt --all && cargo clippy --all-targets && cargo test --all`: pass.
- New `tests/init_agent_cursor_test.rs`:
  - `agent_cursor_dry_run_never_touches_claude`: asserts dry-run output contains no `/.claude/` path and does patch `/.cursor/`.
  - `agent_cursor_real_write_creates_no_claude_dir`: non-dry-run install in a temp `HOME`: exits 0, never creates `~/.claude`, and writes `.cursor/hooks.json`.
[...]

PR #3014, opened 2026-07-16
fix: rtk init -g --agent cursor fails when ~/.claude directory is missing

### LINKED ISSUE #3009: wont install for Cursor - trys to install Claude v 0.43.0
```
% rtk init -g --agent cursor
rtk: Failed to write RTK.md: /Users/x/.claude/RTK.md: Failed to create temp file in /Users/x/.claude: No such file or directory (os error 2) at path "/Users/x/.claude/.tmpCvxoaZ"
x@JeffsMacPro-2 ~ % rtk --version
rtk 0.43.0
```
Same thing on both my dev machines:
MACOS 15.7.3 (24G419) 
MACOS 26.5.1 (25F80)

if I create a dir called .claude, it installs, so "--agent cursor" is completely ignored.
I'm happy to try any workaround until its fixed...

mvanhorn:
## Summary
The single choke point is `atomic_write(path, content)` in `src/hooks/init.rs`: it computes `parent = target.parent()` and immediately calls `NamedTempFile::new_in(parent)`, which returns `os error 2` when `parent` does not exist. Every other write site in this file already calls `fs::create_dir_all(...)` before writing (e.g.

## Why this matters
Running `rtk init -g --agent cursor` (global install targeting Cursor) crashes with `Failed to write RTK.md: .../.claude/RTK.md: Failed to create temp file in .../.claude: No such file or directory (os error 2)` whenever the user does not already have a `~/.claude` directory. It is reproduced on rtk 0.43.0 on macOS 15.7.3, macOS 26.5.1, and Arch Linux by three separate reporters. The reporter confirms that manually creating a `.claude` directory makes the command succeed, so the failure is purely the missing parent directory. Because the global default path also installs Claude Code awareness files (Cursor hooks are additive on top of Claude Code) and fails fast on that first write, the Cursor portion never runs.

## Testing
- Happy path: `atomic_write` (or global default init via a temp `CLAUDE_CONFIG_DIR` whose `.claude` subdir does not yet exist) creates the missing parent directory and writes the file successfully, mirroring the existing `test_default_mode_creates_rtk_md` pattern.
- Idempotence / existing dir: writing when the parent directory already exists still succeeds and overwrites atomically (no regression to the existing atomic-rename behavior or the `write_if_changed` no-op-when-unchanged path).
- Error path: writing to a path with no parent (root) still returns the existing "path has no parent directory" error, and a genuinely unwritable parent still surfaces a contextual error rather than panicking.

Fixes #3009

AI was used for assistance.

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -412,6 +412,9 @@ fn atomic_write(path: &Path, content: &str) -> Result<()> {
         )
     })?;
 
+    fs::create_dir_all(parent)
+        .with_context(|| format!("Failed to create parent directory {}", parent.display()))?;
+
     // Create temp file in same directory (ensures same filesystem for atomic rename)
     let mut temp_file = NamedTempFile::new_in(parent)
         .with_context(|| format!("Failed to create temp file in {}", parent.display()))?;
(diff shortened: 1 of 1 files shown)


pszymkowiak:
**Thanks for the fix and the thorough write-up — the root-cause analysis of `atomic_write` is spot on, and I confirmed the fix resolves the crash.** Before merging, I want to flag a design concern, because as it stands this PR makes the crash disappear by doubling down on behavior that arguably shouldn't happen.

**The problem: `rtk init -g --agent cursor` installs a Claude Code integration.** I verified this empirically by running the PR binary against a fresh `HOME` with no `~/.claude`:

| File created | When |
|---|---|
| `~/.claude/RTK.md` | always |
| `~/.claude/CLAUDE.md` (patched with `@RTK.md`) | always |
| `~/.claude/settings.json` (hook `rtk hook claude`) | on consent (`--auto-patch` / interactive `y`) |
| `~/.cursor/hooks.json` (`rtk hook cursor`) | always — the only Cursor-native artifact |

None of the `~/.claude/*` files are read by Cursor. So a user who asked for Cursor gets Claude Code awareness files (and, with auto-patch, a wired-up Claude hook) they never requested.

This traces to `install_claude = !opencode` (`main.rs:2034`) combined with Cursor **not** being intercepted before the default mode (`init.rs:308-311`) — unlike Windsurf/Cline. It's inconsistent with every other non-Claude agent, all of which are standalone and never touch `~/.claude`:

| Agent | Writes | Touches `.claude`? |
|---|---|---|
| Cline | `.clinerules` | ❌ |
| Windsurf | `.windsurfrules` | ❌ |
| Kilo Code | `.kilocode/rules/rtk-rules.md` | ❌ |
| Codex | `AGENTS.md` | ❌ |
| **Cursor (this PR)** | `.claude/RTK.md` + `CLAUDE.md` (+ `settings.json`) | ✅ |

The issue reporter said it themselves: *"--agent cursor is completely ignored"* — their real complaint is that a Cursor install populates `~/.claude`. Fix A turns the crash into silent `~/.claude` writes rather than fixing the surprising behavior.

**Suggested alternative (Fix B):** intercept Cursor before the Claude default path, matching Windsurf/Cline — e.g.

```rust
// main.rs
let install_claude = !opencode && agent != Some(AgentTarget::Cursor);
```

or add an `else if install_cursor { install_cursor_hooks(ctx)?; }` branch in the `init.rs:308` chain. That installs only `~/.cursor/hooks.json`, creates no Claude files, and still fixes #3009 — closer to what the reporter expects.

**Separate gap:** `hooks/cursor/` has no native awareness doc, so today Cursor gets the hook but no instructions (awareness was piggybacking on `~/.claude/RTK.md`, which Cursor can't read). A complete fix would add a Cursor-native awareness file (e.g. `~/.cursor/rules/` or `AGENTS.md`).

**The `atomic_write` + `create_dir_all` change and its tests are genuinely good and worth keeping regardless.** My only ask is that the Cursor routing be fixed so we don't install Claude Code for Cursor users. Happy to help.

mvanhorn:
Appreciate it @pszymkowiak — thanks for the thorough review.

edurenye:
I think there is already another PR to fix this: https://github.com/rtk-ai/rtk/pull/3000

mvanhorn:
Good catch, thanks. #3000 does cover the same failure (the 'Failed to create temp file in ~/.claude' error) and goes further by keeping cursor installs from touching ~/.claude at all. Closing this one in favor of #3000.

PR #3017, opened 2026-07-16
fix(hook): preserve explicit linter rewrites

### LINKED ISSUE #2906: bug(hook): `pnpm exec biome ci .` → `rtk lint ci .` drops the explicit `biome` name and runs `ci` as the linter (exit 254 misreported as "possibly out of memory")
## Summary

The hook rewrite strips `pnpm exec biome` as a *prefix*, so the explicitly-named linter is thrown away and the linter's **first subcommand becomes the linter name**:

```console
$ rtk hook check "pnpm exec biome ci ."
rtk lint ci .
```

`rtk lint ci .` then runs `pnpm exec ci .`, which doesn't exist, exits 254, and is reported as `[warn] Linter process terminated abnormally (possibly out of memory)`.

`biome ci .` is Biome's own documented CI entrypoint (https://biomejs.dev/reference/cli/#biome-ci), so every Biome project whose `lint` script is `biome ci .` hits this.

This is a **different trigger** from #2879 (bare `pnpm lint` → no args → ESLint default). Here the user *explicitly names biome on the command line* and rtk still doesn't route to the Biome adapter. Related but not the same: #2879, #1489, #2540, #665. Searched `"pnpm exec biome"` and `detect_linter` — no existing issue covers this path.

## Repro

Project with Biome and **no ESLint**, `"lint": "biome ci ."`:

```console
$ rtk hook check "pnpm exec biome ci ."
rtk lint ci .

$ rtk hook check "pnpm exec biome --version"
rtk lint --version

$ pnpm exec biome ci .            # via the Claude Code PreToolUse hook (`rtk hook claude`)
[warn] Linter process terminated abnormally (possibly out of memory)
exit=254

$ pnpm exec biome --version
[warn] Linter process terminated abnormally (possibly out of memory)
exit=254
```

Both are healthy commands. Bypassing rtk proves it:

```console
$ rtk proxy pnpm exec biome ci . ; echo "exit=$?"
exit=0

$ ./node_modules/.bin/biome --version
Version: 2.4.16

$ /usr/bin/time -l ./node_modules/.bin/biome ci .
        0.21 real         1.09 user         0.53 sys
        98369536  maximum resident set size      # 98 MB on a 128 GB machine
```

`rtk lint` names the bogus linter out loud under `-v`:

```console
$ rtk -v lint ci .
Running: ci with structured output
[warn] Linter process terminated abnormally (possibly out of memory)

$ pnpm exec ci .
undefined
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "ci" not found
exit=254
```

## Root cause (verified against tag `v0.43.0`)

1. **[`src/discover/rules.rs:153`](https://github.com/rtk-ai/rtk/blob/v0.43.0/src/discover/rules.rs#L153)** — the rule `^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?(biome|eslint|lint)(\s|$)` maps to `rtk_cmd: "rtk lint"` ([L154](https://github.com/rtk-ai/rtk/blob/v0.43.0/src/discover/rules.rs#L154)), and `"pnpm exec biome"` is listed verbatim in `rewrite_prefixes` ([L185](https://github.com/rtk-ai/rtk/blob/v0.43.0/src/discover/rules.rs#L185)). Because the whole matched span — *including the linter name* — is consumed as a prefix, `pnpm exec biome ci .` becomes `rtk lint ci .`. The information that the user asked for **biome** is gone by the time `rtk lint` starts.

2. **[`src/cmds/js/lint_cmd.rs:77-88`](https://github.com/rtk-ai/rtk/blob/v0.43.0/src/cmds/js/lint_cmd.rs#L77-L88)** — `detect_linter(["ci", "."])`:

   ```rust
   let is_path_or_flag = args.is_empty()
       || args[0].starts_with('-')
       || args[0].contains('/')
       || args[0].contains('.');

   if is_path_or_flag { ("eslint", false) } else { (&args[0], true) }
   ```

   `"ci"` has no `-`, `/` or `.`, so it is taken as an **explicitly specified linter named `ci`** and rtk runs `pnpm exec ci .`.

   (For `pnpm exec biome --version` → `rtk lint --version`, `args[0]` starts with `-`, so it falls to the `("eslint", false)` default and runs ESLint instead — the #2879 shape.)

3. **[`src/cmds/js/lint_cmd.rs:173-174`](https://github.com/rtk-ai/rtk/blob/v0.43.0/src/cmds/js/lint_cmd.rs#L173-L174)** — pnpm's "command not found" exit code is **254**, and `exit_code > 128` is treated as death-by-signal:

   ```rust
   if !result.success() && result.exit_code > 128 {
       eprintln!("[warn] Linter process terminated abnormally (possibly out of memory)");
   ```

   Nothing was killed and nothing ran out of memory. (Same misclassification as #2879 step 3.)

## Why this is costly for agents

The message names a *resource* problem, so an LLM agent chases memory. In our repo this cost real time: an agent concluded `pnpm exec biome ci .` OOMs on this machine and recorded `RAYON_NUM_THREADS=2` as the fix. That workaround does nothing — the env prefix is preserved through the rewrite and the rewrite still fires:

```console
$ rtk hook check "RAYON_NUM_THREADS=2 pnpm exec biome ci ."
RAYON_NUM_THREADS=2 rtk lint ci .

$ RAYON_NUM_THREADS=2 pnpm exec biome ci . ; echo "exit=$?"
[warn] Linter process terminated abnormally (possibly out of memory)
exit=254
```

The real Biome run takes 0.21 s and 98 MB.

## Suggested fix

Any of, ideally 1 + 3:

1. **Keep the linter name.** When the rule matched `biome`/`eslint` explicitly, pass it through: `pnpm exec biome ci .` → `rtk lint biome ci .` (or dispatch straight to the Biome adapter). Only the bare `lint` script name should be ambiguous.
2. **Never guess from `args[0]`.** `detect_linter` should match `args[0]` against the known linter set rather than accepting any bare word as a linter — `ci`, `check`, `format` are Biome/other subcommands, not linters.
3. **Stop reporting 254 as OOM.** Detect pnpm's `ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL` / `Command "X" not found` and say `X is not installed` (or fall back to passthrough). A `> 128` check is not a valid signal-death test for pnpm/npm wrappers.

Given #665, #1489, #2094, #2540, #2832 and #2879 all stem from "the lint rewrite guesses the linter", the durable fix is probably (1)+(2): never dispatch to a linter that the command line didn't actually name.

## Environment

- rtk **0.42.4** (Homebrew, macOS 26.5.1 arm64) — behaviour observed; source read and line-referenced at tag **v0.43.0**, where the code is unchanged
- pnpm 10.33.2, Node v25.8.1, Biome 2.4.16, **no ESLint anywhere in the dependency tree**
- `lint` script: `biome ci .`

ychampion:
Fixes #2906

## Summary

- keep `biome` and `eslint` in hook rewrites when the command names them explicitly
- leave generic `lint` script rewrites on the existing auto-detection path

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets --locked -- -D warnings && cargo test --locked`
- [x] Manual testing: `rtk hook check "pnpm exec biome ci ."` -> `rtk lint biome ci .`

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -960,16 +960,33 @@ fn rewrite_segment_inner(
     // Try each rewrite prefix (longest first) with word-boundary check
     for &prefix in rule.rewrite_prefixes {
         if let Some(rest) = strip_word_prefix(strip_target, prefix) {
-            let rewritten = if rest.is_empty() {
-                format!("{}{}", rule.rtk_cmd, redirect_suffix)
-            } else {
-                format!("{} {}{}", rule.rtk_cmd, rest, redirect_suffix)
+            let explicit_linter = explicit_linter_from_prefix(rule.rtk_cmd, prefix);
+            let rewritten = match (explicit_linter, rest.is_empty()) {
+                (Some(linter), true) => {
+                    format!("{} {}{}", rule.rtk_cmd, linter, redirect_suffix)
+                }
+                (Some(linter), false) => {
+                    format!("{} {} {}{}", rule.rtk_cmd, linter, rest, redirect_suffix)
+                }
+                (None, true) => format!("{}{}", rule.rtk_cmd, redirect_suffix),
+                (None, false) => format!("{} {}{}", rule.rtk_cmd, rest, redirect_suffix),
             };
             return Some(rewritten);
         }
     }
 
     None
+}
+
+fn explicit_linter_from_prefix<'a>(rtk_cmd: &str, prefix: &'a str) -> Option<&'a str> {
+    if rtk_cmd != "rtk lint" {
+        return None;
+    }
+
+    match prefix.split_whitespace().last() {
+        Some(linter @ ("biome" | "eslint")) => Some(linter),
+        _ => None,
+    }
 }
 
 /// Strip a command prefix with word-boundary check.
@@ -2966,13 +2983,35 @@ mod tests {
             "lint",
         ];
         for command in commands {
+            let linter = command.split_whitespace().last().unwrap();
+            let expected = match linter {
+                "biome" | "eslint" => format!("rtk lint {linter}"),
+                "lint" => "rtk lint".to_string(),
+                _ => unreachable!(),
+            };
             assert_eq!(
                 rewrite_command_no_prefixes(command, &[]),
-                Some("rtk lint".into()),
+                Some(expected),
                 "Failed for command: {}",
                 command
             );
         }
+    }
+
+    #[test]
+    fn test_rewrite_lint_preserves_explicit_linter_arguments() {
+        assert_eq!(
+            rewrite_command_no_prefixes("pnpm exec biome ci .", &[]),
+            Some("rtk lint biome ci .".into())
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("pnpm exec eslint --version", &[]),
+            Some("rtk lint eslint --version".into())
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("RAYON_NUM_THREADS=2 pnpm exec biome ci .", &[]),
+            Some("RAYON_NUM_THREADS=2 rtk lint biome ci .".into())
+        );
     }
 
     #[test]


KuSh:
Closing as a duplicate of #3088, which fixes the same lost linter name from #2906 by splitting the biome and eslint rewrite rules at their source, so classify_command reports the right equivalent too, and which also stops detect_linter treating a subcommand as a linter name. Review and comments are welcome there.

PR #3020, opened 2026-07-16
fix(hooks): always set permissionDecision so Claude Code applies rewrites

### LINKED ISSUE #3018: rtk hook claude omits permissionDecision in updatedInput responses, causing Claude Code to silently ignore command rewrites
Environment
RTK version: 0.42.1 and 0.42.4 (confirmed present in both)
Platform: Windows 11, Git Bash
Claude Code hooks: PreToolUse matcher Bash → rtk hook claude
Summary
When RTK's Claude Code hook rewrites a Bash command, it emits a hookSpecificOutput object containing updatedInput but without the permissionDecision field. Per Claude Code's hook schema, updatedInput is only applied when permissionDecision is explicitly set (e.g. "allow"). Without it, Claude Code silently discards updatedInput and executes the original, unrewritten command — so RTK's token-saving rewrite never actually takes effect, even though the hook itself runs without error.

Steps to reproduce
Configure the Claude Code hook: PreToolUse → matcher Bash → command rtk hook claude.
Simulate the exact payload Claude Code sends on a Bash tool call:
echo '{"session_id":"test","transcript_path":"/tmp/x","cwd":"/some/path","hook_event_name":"PreToolUse","tool_name":"Bash","tool_input":{"command":"git status","description":"test"}}' | rtk hook claude
Observe the output.
Actual output
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk git status","description":"test"}}}
Expected output
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"allow","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk git status","description":"test"}}}
Impact
Every Bash command routed through the hook runs unmodified, with zero token savings.
rtk gain / rtk gain --history never record real interceptions — the local history.db only ever contains a single stale fallback entry, regardless of how many Bash commands are run or how many times the Claude Code session is restarted.
This effectively makes the "transparent, 0 tokens overhead" rewrite behavior described in RTK's docs non-functional under current Claude Code versions.
Root cause
Per the Claude Code hooks documentation, updatedInput under hookSpecificOutput is only honored when permissionDecision is also present (typical value: "allow"). RTK's hook claude subcommand does not set this field.

Reference: [Claude Code Hooks reference](https://code.claude.com/docs/en/hooks)

Suggested fix
In the hook claude (and likely cursor, gemini, copilot, droid) subcommands, add "permissionDecision": "allow" to the hookSpecificOutput object whenever updatedInput is returned.

Nitjsefnie:
## Problem

`rtk hook claude` emits a `hookSpecificOutput` containing `updatedInput` but, on the default "ask" path, omits `permissionDecision`. Per the Claude Code hooks schema, `updatedInput` is only applied when `permissionDecision` is explicitly present, so Claude Code silently executes the original command and the rewrite has no effect.

## Root cause

`src/hooks/hook_cmd.rs` had two paths that build Claude-shaped `hookSpecificOutput` and they disagreed:

- The VS Code/Copilot handler (`handle_vscode`, ~line 153) already mapped `AllowRewrite` → `"allow"` and `AskRewrite` → `"ask"` and always emitted `permissionDecision`.
- The native Claude Code handler (`process_claude_payload`, ~line 342) reduced the decision to a `bool`, only inserting `permissionDecision: "allow"` when `allow == true`. On `AskRewrite` the field was omitted entirely.

`AskRewrite` is the default verdict for rewritable commands when no explicit allow rule exists, so most Bash commands were rewritten in RTK's output but never actually changed in Claude Code.

## Fix

Make `process_claude_payload` set `permissionDecision` unconditionally — `"allow"` for `AllowRewrite` and `"ask"` for `AskRewrite` — matching the existing correct path in `handle_vscode`.

## Schema confirmation

The Claude Code hooks documentation states that for `PreToolUse`, `updatedInput` allows modifying tool input parameters, and the example shows it paired with `"permissionDecision": "allow"`. Valid values are `"allow"`, `"deny"`, and `"ask"`. See [Claude Code hooks reference — PreToolUse Decision Control](https://docs.claude.com/en/docs/claude-code/hooks).

## Regression-test evidence

Added tests covering both rewrite verdicts:

- `test_claude_allow_rewrite_sets_permission_decision_allow`
- `test_claude_ask_rewrite_sets_permission_decision_ask`

I temporarily reverted the fix while keeping the new tests. The AskRewrite test failed with:

```
  left: Null
 right: "ask"
```

After restoring the fix, both new tests and the full suite pass.

## Gates

- `cargo fmt --all --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test` ✅ (2437 unit tests + integration tests)

Closes #3018

Prepared with AI assistance (Kimi K2.7 Code), human-reviewed before submission.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -339,7 +339,7 @@ enum PayloadAction {
     Ignore,
 }
 
-fn process_claude_payload(v: &Value) -> PayloadAction {
+fn process_claude_payload(v: &Value, decision: HookDecision) -> PayloadAction {
     let cmd = match v
         .pointer("/tool_input/command")
         .and_then(|c| c.as_str())
@@ -349,7 +349,7 @@ fn process_claude_payload(v: &Value) -> PayloadAction {
         None => return PayloadAction::Ignore,
     };
 
-    let (rewritten, allow) = match decide_hook_action(cmd, permissions::Host::Claude) {
+    let (rewritten, decision_str) = match decision {
         HookDecision::Deny => {
             return PayloadAction::Skip {
                 reason: "skip:deny_rule",
@@ -362,8 +362,8 @@ fn process_claude_payload(v: &Value) -> PayloadAction {
                 cmd: cmd.to_string(),
             }
         }
-        HookDecision::AllowRewrite(r) => (r, true),
-        HookDecision::AskRewrite(r) => (r, false),
+        HookDecision::AllowRewrite(r) => (r, "allow"),
+        HookDecision::AskRewrite(r) => (r, "ask"),
     };
 
     let updated_input = {
@@ -374,18 +374,12 @@ fn process_claude_payload(v: &Value) -> PayloadAction {
         ti
     };
 
-    let mut hook_output = json!({
+    let hook_output = json!({
         "hookEventName": PRE_TOOL_USE_KEY,
+        "permissionDecision": decision_str,
         "permissionDecisionReason": "RTK auto-rewrite",
         "updatedInput": updated_input
     });
-
-    if allow {
-        hook_output
-            .as_object_mut()
-            .unwrap()
-            .insert("permissionDecision".into(), json!("allow"));
-    }
 
     PayloadAction::Rewrite {
         cmd: cmd.to_string(),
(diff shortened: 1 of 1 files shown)


Nitjsefnie:
Withdrawing this PR — I'm not in a position to sign the CLA, so it can't proceed. The diff stays available on the branch if anyone wants to pick it up. Thanks!

PR #1405, opened 2026-04-19
feat: add uv run support

alexanderkreidich:
## Summary
- add a dedicated `rtk uv` command that preserves `uv run` environment semantics while filtering output down to relevant failures
- add discover/rewrite support so `uv run ...` rewrites to `rtk uv run ...`
- document the new command in the README, Python command notes, and hook initialization help

## Testing
- `PATH=$HOME/.rustup/toolchains/stable-x86_64-apple-darwin/bin:$PATH cargo test uv_run`
- `PATH=$HOME/.rustup/toolchains/stable-x86_64-apple-darwin/bin:$PATH cargo test test_init_mentions_all_top_level_commands`
- `PATH=$HOME/.rustup/toolchains/stable-x86_64-apple-darwin/bin:$PATH cargo run --quiet -- hook check -- uv run pytest`
- `./target/debug/rtk uv run python -c 'print("hello")'`
- `./target/debug/rtk uv run python -c 'raise RuntimeError("kaboom")'`

Closes #1379

Issue #294, opened 2026-03-01
Hook rewrite: add coverage for uv run, pnpm exec, Python path variants, and more


wkup:
Analyzed rtk discover output across 429 sessions (3,754 Bash commands) and identified ~727 missed commands that RTK already has subcommands for  
  but the hook script wasn't intercepting.                  

  Updated rtk-rewrite.sh to cover all these patterns. Full test suite (55 cases) passing with zero regressions.

  ---
  uv run dispatch — 378 commands (highest impact)

  uv run ruff check/format ...              → rtk ruff check/format ...
  uv run pytest ...                         → rtk pytest ...
  uv run python -m ruff check/format ...    → rtk ruff check/format ...
  uv run python -m pytest ...               → rtk pytest ...
  uv run mypy ...                           → rtk err uv run mypy ...

  pnpm exec dispatch — 89 commands

  pnpm exec vue-tsc --noEmit                → rtk tsc --noEmit
  pnpm exec eslint ...                      → rtk lint ...
  pnpm exec prettier ...                    → rtk prettier ...
  pnpm exec vitest [run]                    → rtk vitest run

  pnpm script commands — 74 commands

  pnpm type-check                           → rtk err pnpm type-check
  pnpm build                                → rtk err pnpm build
  pnpm add/install/remove                   → rtk pnpm add/install/remove

  Python path variants — 150+ commands

  Previously only pytest, .venv/bin/pytest, python -m pytest, and ruff check/format were handled. Now also covers:

  api/.venv/bin/python -m ruff check ...    → rtk ruff check ...
  api/.venv/bin/python -m pytest ...        → rtk pytest ...
  python3 -m ruff check/format ...          → rtk ruff check/format ...
  .venv/bin/ruff check/format ...           → rtk ruff check/format ...
  api/.venv/bin/ruff check/format ...       → rtk ruff check/format ...
  mypy ...                                  → rtk err ...
  python -m mypy ...                        → rtk err ...
  .venv/bin/python -m mypy ...              → rtk err ...

  uv sync — 10 commands

  uv sync                                   → rtk err uv sync

  head -n fix (missing variant)

  head -n 50 file                           → rtk read file --max-lines 50
  head -n50 file                            → rtk read file --max-lines 50

  Previously only head -50 file and head --lines=50 file were handled.

  ---
  Testing

  55 test cases (15 regression + 40 new), all passing:

  - All new patterns with real-world command examples
[...]

PR #3021, opened 2026-07-16
test(hook): cover option-bearing uv runs in Claude chains

GratefulDave:
## Summary

- add registry coverage for `uv run --extra dev pytest`, the option-bearing form skipped by RTK 0.43.0
- add an end-to-end Claude PreToolUse regression for the frequent compound shape `cd ... && uv run --extra dev pytest ... 2>&1 | tail ...`
- lock in develop's dedicated `rtk uv` routing so this high-frequency Claude Code command keeps receiving an `updatedInput` rewrite

Related to #294. The production routing landed in #1405; this PR covers the stable-version regression at the Claude hook boundary.

## Test plan

- [x] `cargo test uv_run`
- [x] Manual testing: exact Claude Bash payload rewrites `uv run --extra dev pytest ...` to `rtk uv run --extra dev pytest ...` inside the compound chain

> **Important:** Targets `develop` as required by the contribution guide.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -2577,6 +2577,10 @@ mod tests {
             Some("rtk uv run --unknown pytest tests/".into())
         );
         assert_eq!(
+            rewrite_command_no_prefixes("uv run --extra dev pytest tests/ -q", &[]),
+            Some("rtk uv run --extra dev pytest tests/ -q".into())
+        );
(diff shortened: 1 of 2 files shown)


GratefulDave:
@CLAassistant I signed this. Why does it not show as signed? I signed multiple times.

Issue #1853, opened 2026-05-12
fix(git): run_log silently injects --no-merges, hiding merge commits


paolomainardi:
> :robot: _This was written by an AI agent on behalf of @paolomainardi._

## Description

`run_log()` unconditionally adds `--no-merges` to all `git log` commands unless the user explicitly passes `--merges`. This silently removes all merge commits from the output with no indication they were filtered.

For AI agents, this is a correctness bug: the agent cannot see that branches were merged, leading to wrong conclusions about branch topology and project history.

## Root Cause

Lines ~492-498 in `run_log()`:

```rust
// Only add --no-merges if user didn't explicitly request merge commits
let wants_merges = args
    .iter()
    .any(|arg| arg == "--merges" || arg == "--min-parents=2");
if !wants_merges {
    cmd.arg("--no-merges");
}
```

## Reproduction

```bash
# Create a repo with a merge commit
git init test-repo && cd test-repo
echo "a" > a.txt && git add . && git commit -m "initial"
git checkout -b feature
echo "b" > b.txt && git add . && git commit -m "feature work"
git checkout master
git merge --no-ff feature -m "Merge branch 'feature'"
echo "c" > c.txt && git add . && git commit -m "after merge"

# Raw: 4 commits (including merge)
git log --oneline | wc -l
# Output: 4

# RTK: 3 commits (merge silently removed)
rtk git log --oneline | wc -l
# Output: 3

# The merge commit is completely invisible
rtk git log --oneline | grep "Merge"
# Output: (empty)
```

## Side Effects

- `git log --graph` shows **corrupted topology** because the merge node is removed and the graph is redrawn incorrectly
- Agents that check "was this branch merged?" get the wrong answer
- `git log --oneline` count disagrees with `git rev-list --count`

## Fix

Remove the `--no-merges` injection entirely. If users want to filter merge commits, they can pass `--no-merges` explicitly. RTK should not silently alter the semantics of git commands.

## Impact

- **Severity**: High — merge commits vanish completely, corrupts history understanding
- **Token cost**: +1 line per merge commit in history — negligible
- **Version**: Verified on v0.39.0

## Related

- #619 (git log issues)
- #1704 (git log filtering)

paolomainardi:
Verified this myself on v0.39.0, merge commits are silently removed from `git log` output with no indication. 
This bit me when an agent couldn't see that a branch had been merged. If this was an intentional design choice to reduce noise, 

[...]

PR #3028, opened 2026-07-17
fix(git): keep every commit in git log --stat output

### LINKED ISSUE #2882: rtk git log --stat silently drops ~36% of commits (no truncation notice, unlike git diff)
## Summary
`rtk git log --stat` silently drops a large fraction of commits compared to raw `git log --stat`, with no truncation notice and no escape hatch (unlike `rtk git diff`, which discloses truncation via `... (more changes truncated)` + a `--no-compact` flag).

## Environment
- rtk 0.43.0 (official `rtk-x86_64-pc-windows-msvc.zip` release asset, checksum-verified)
- Windows, invoked via Git Bash

## Repro
```
git log -50 --stat > raw.txt
rtk git log -50 --stat > rtk.txt
```

Compared commit hashes present in each file: **18 of 50 commits (36%) present in `raw.txt` are completely absent from `rtk.txt`** — not truncated-with-notice, just gone. Example: commit `463e523` (`skill(plan-feature): DoD evaluator sweeps whole e2e file for RED-run cleanup leaks`) sits directly between commits `a0530e8` and `deaa799` in `raw.txt`, but `rtk.txt` goes straight from `a0530e8` to `deaa799` with no trace of `463e523` — no hash, no message, no stat block, no omission marker.

At least one dropped commit is a merge commit's neighbor: `23db401` (`Merge: 0660223 0757278`, message `Merge research/digital-products-validation`) and both `0660223` and `0757278` are among the missing set — suggesting the parser may not expect a merge commit's header-only block (no diffstat, since `git log --stat` doesn't print a diffstat for merge commits by default) and loses its place across the surrounding entries.

## Why this matters
`rtk git diff`'s compact mode does this safely — I ran a similar comparison on a 107-file diff and it explicitly printed `... (more changes truncated)` plus `[full diff: rtk git diff --no-compact]` when it dropped content. `rtk git log --stat` gives no equivalent signal, so a caller has no way to know the commit history it's looking at is incomplete.

## Expected
Either preserve every commit in the compact output, or emit the same kind of disclosed-truncation marker `git diff` already uses (and ideally a `--no-compact` fallback for `git log` too).

## Suggested area
Likely in the same registry/parsing layer `git diff`'s compact formatter uses, but specific to the `--stat` log parser's handling of merge-commit boundaries (no diffstat block to anchor on).

georgyia:
Fixes #2882

## Problem

`rtk git log --stat` drops most commits, with no notice and no escape hatch. Measured on this repo:

```console
$ git log --stat -30 --format=%h | grep -cE '^[0-9a-f]+$'
30
$ rtk git log --stat -30 | grep -cE '^[0-9a-f]{7,40} '
9                       # 21 commits gone, silently
```

The reporter saw 18 of 50 missing. An agent reading this gets a confidently incomplete history.

## Root cause

Not merge-commit handling, which is where the issue reasonably pointed — merge commits are a symptom.

With `--stat`, git appends the diffstat **after** the pretty-format output, i.e. after RTK's trailing `---END---` marker:

```
<hash> <subject> (date) <author>     <- commit N, from --pretty=format
<body>
---END---                            <- marker terminates the format
 CONTRIBUTING.md | 8 ++++----        <- ...but the diffstat lands *here*
 1 file changed, 4 insertions(+), 4 deletions(-)
```

So `output.split("---END---")` yields blocks shaped `[diffstat of commit N-1] + [header of commit N] + [body of commit N]`. `filter_log_output` takes the block's first line as the header — which is now a diffstat line — and the real commit header becomes a body line, capped at 3 and collapsed into `[+N lines omitted]`. Each commit that has a diffstat consumes the next commit's slot.

Merge commits produce no diffstat (`git log --stat` omits it for merges by default), so the corruption is irregular rather than uniform. That is why it presented as "the parser loses its place around merge commits": they are the blocks where the pattern breaks step, not the cause.

## Fix

Lead each commit with the marker instead of terminating it:

```diff
-"--pretty=format:%h %s (%ar) <%an>%n%b%n---END---"
+"--pretty=format:---RTK-COMMIT---%n%h %s (%ar) <%an>%n%b"
```

Every block now begins at a real commit header, and anything git appends after the format — the diffstat — stays inside the block of the commit it describes. No commit can be knocked out of step, and merge commits need no special-casing since the anchor no longer depends on a diffstat being present.

Blank blocks are filtered before `take(limit)`, because the empty region ahead of the first marker would otherwise consume one commit's slot.

Diffstat lines share the existing body budget and overflow through the existing `[+N lines omitted]` notice, so the issue's "Expected" is met on both counts: every commit is preserved, and anything dropped is disclosed.

**Scope note:** `--stat` stays *filtered* rather than switching to passthrough. #2951 proposes passthrough for `-p`/`--patch` and its tests assert `--stat` keeps being filtered; fixing the parser satisfies both, so the two PRs are complementary rather than contradictory. They touch the same function and whichever lands second will need a trivial rebase.

## Verification

| | native | rtk before | rtk after |
|---|---|---|---|
| `--stat -10` | 10 commits | — | **10** |
| `--stat -30` | 30 commits | 9 | **30** |
| `--stat -50` | 50 commits | — | **50** |

Hash-level set comparison at `-30`: commits present in native but absent from rtk went from 21 to **0**.

- New regression tests fail on the pre-fix parser: `test_filter_log_output_stat_keeps_every_commit` (a merge commit + two diffstat-bearing commits, asserting the middle one survives, mirroring the `a0530e8` / `463e523` / `deaa799` case in the issue) and `test_filter_log_output_stat_overflow_is_disclosed`.
- No regression: old vs new binary output is byte-identical for `git log`, `git log -5`, `git log --oneline -20`, `git log -3 --format=%H`.
- `cargo fmt --all --check` clean, `cargo clippy --all-targets` zero warnings, `cargo test --all` 2495 passed / 0 failed.

**Token savings, stated plainly:** `git log --stat -30` goes from 59.3% to **53.1%** — the extra words are the 21 recovered commits. Worth being explicit that this filter never met the 60% target on `--stat`; the old number was closer to it only because it was discarding 70% of the data. I'd rather surface that than quietly claim a win. If you want `--stat` back above 60%, the natural lever is compacting each diffstat to its `N files changed` summary line, which I'm happy to add here or in a follow-up.

## Test plan

- [x] Manual testing: `rtk git log --stat -10/-30/-50` compared against native `git log --stat` by hash set
- [x] Token savings measured and reported above

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -16,6 +16,10 @@ use std::ffi::OsString;
 use std::process::Command;
 use std::process::Stdio;
 
+/// Leads every commit block in RTK's injected `git log` pretty-format, letting the
+/// parser anchor each commit even when git appends a diffstat after the format.
+const COMMIT_MARKER: &str = "---RTK-COMMIT---";
+
 #[derive(Debug, Clone)]
 pub enum GitCommand {
     Diff,
@@ -451,8 +455,17 @@ fn run_log(
     // Apply RTK defaults only if user didn't specify them
     // Use %b (body) to preserve first line of commit body for agent context
     // (BREAKING CHANGE, Closes #xxx, design notes)
+    //
+    // The marker leads each commit rather than terminating it: with --stat (and
+    // friends) git appends the diffstat *after* the pretty-format output, so a
+    // trailing marker would leave those lines heading the next block, where the
+    // parser read them as that commit's header and lost the real one (#2882).
+    // Leading it keeps every commit anchored and each diffstat with its own commit.
     if !has_format_flag {
-        cmd.args(["--pretty=format:%h %s (%ar) <%an>%n%b%n---END---"]);
+        cmd.args([format!(
+            "--pretty=format:{}%n%h %s (%ar) <%an>%n%b",
+            COMMIT_MARKER
+        )]);
     }
 
     // Determine limit: respect user's explicit -N flag, use sensible defaults otherwise
@@ -564,7 +577,7 @@ pub(crate) fn filter_log_output(
     let truncate_width = if user_set_limit { 120 } else { 80 };
 
     // When user specified their own format (--oneline, --pretty, --format),
-    // RTK did not inject ---END--- markers. Use simple line-based truncation.
+    // RTK did not inject commit markers. Use simple line-based truncation.
     if user_format {
         let lines: Vec<&str> = output.lines().collect();
         let max_lines = if user_set_limit { lines.len() } else { limit };
@@ -576,16 +589,19 @@ pub(crate) fn filter_log_output(
             .join("\n");
     }
 
-    // RTK injected format: split output into commit blocks separated by ---END---
-    let commits: Vec<&str> = output.split("---END---").collect();
+    // RTK injected format: COMMIT_MARKER leads each commit block, so everything git
+    // appends after the format (a --stat diffstat) stays inside its own commit.
+    // Blanks are dropped before `take` — the empty region ahead of the first marker
+    // would otherwise consume one commit's slot.
+    let commits: Vec<&str> = output
+        .split(COMMIT_MARKER)
+        .map(|block| block.trim())
+        .filter(|block| !block.is_empty())
+        .collect();
     let max_commits = if user_set_limit { commits.len() } else { limit };
 
     let mut result = Vec::new();
     for block in commits.iter().take(max_commits) {
-        let block = block.trim();
-        if block.is_empty() {
-            continue;
-        }
         let mut lines = block.lines();
         // First line is the header: hash subject (date) <author>
         let header = match lines.next() {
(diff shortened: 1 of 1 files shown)


src/cmds/git/git.rs
KuSh, line 612: Minor: diffstat lines from `git log --stat` get mixed into the same untagged `all_body_lines` list as real commit-message body text, and they’re truncated with the default width of 80 chars.

For a large commit and/or a deeply nested path, for example `src/cmds/git/git.rs | 123 ++++++++++++++++++++++++++++++++++++++++++++--------`, which is 83 chars, the line gets cut with a trailing `...`. The insertion/deletion counts are then silently lost, and there’s no indication that the truncated line was diffstat data rather than commit-message prose.

This seems like a simple fix that could fit in that PR

KuSh (approved):
Hi @georgyia, LGTM! Thanks for your work.
Just one thing: you'll have to sign the CLA before we can merge this PR.

pszymkowiak:
Closing: already fixed on develop. `git log --stat`, `--numstat` and the other diff-shape forms now request raw passthrough since ca897673 (#3575, released in v0.46.0), so nothing is dropped. Verified: `rtk git log -50 --stat` on this repo yields 50 of 50 commits, byte-identical to native git, and a `--no-ff` merge between diffstat commits is kept. The `--no-merges` injection on the compact forms is a separate issue, tracked in #1853. Thanks for the PR.

Issue #2878, opened 2026-07-08
bug: rtk prettier silently hides prettier 3 failures (stdout-only capture + [warn] filter)


linustalacko:
## Summary

`rtk prettier --check` on a project with real formatting failures (prettier exits 1) reports either a false `All files formatted correctly` or an uninformative bare `Checking formatting...` — never the actual list of files that need formatting. Confirmed on rtk 0.43.0 (Homebrew, macOS arm64), reproduced both from a clean source clone and against the installed binary.

Not a duplicate of #2395 (that one is the *binary-not-installed* case). Also appears not to be fixed by whatever closed #221 — see "Relationship to #221 and #2395" below.

## Root cause (verified against the v0.43.0 tag, commit `5a7880d`)

Two independent defects compound:

1. **[src/cmds/js/prettier_cmd.rs:24](https://github.com/rtk-ai/rtk/blob/v0.43.0/src/cmds/js/prettier_cmd.rs#L24)** — `run()` calls `runner::run_filtered(..., RunOptions::stdout_only())`. Prettier 3.x writes `Checking formatting...` to stdout but writes the per-file `[warn] <file>` lines *and* the final `Code style issues found in N files...` summary to **stderr**. With `stdout_only()`, `filter_prettier_output` never even sees those lines — they're discarded before filtering runs.

2. **[src/cmds/js/prettier_cmd.rs:52](https://github.com/rtk-ai/rtk/blob/v0.43.0/src/cmds/js/prettier_cmd.rs#L52)** — even if stderr were captured, `filter_prettier_output` explicitly excludes any line `.contains("[warn]")`. Prettier 3 emits failing files as exactly `[warn] <file>` (Prettier 2 used bare filenames — the format the function's parsing logic and its test fixtures, lines 128-184, still assume). This is the original defect described in #221's issue body.

[...]

PR #3029, opened 2026-07-17
fix(runner): surface a failing tool's stderr under stdout-only filtering

### LINKED ISSUE #3026: `uv run rtk <tool>` swallows the spawn-failure message when the tool is missing — exit code correct, output empty
Thanks for maintaining rtk — and for adding the `uv run` transparent prefix in 0.43 (#2972); it's exactly the right call for Python workflows. This report is about one edge of that combination.

## Summary

When the wrapped tool isn't installed in the project venv, `rtk <tool>` standalone fails loudly with a clear message. But run as `uv run rtk <tool>` — the exact form the Claude Code hook rewrite produces from `uv run pytest` — the error message vanishes: the exit code is still nonzero (correct), but stdout is a single newline and stderr is empty. For agent use this is the dangerous quadrant: a "test run" that produces no output and no error text is easy to misread as a quiet pass, and the exit code is the only surviving signal.

## Reproduction (rtk 0.43.0, macOS arm64)

In a fresh `uv init` project whose venv does not include pytest (venv already created, so uv's one-time setup notices don't muddy the water):

```console
$ uv run rtk pytest --version > out.txt 2> err.txt; echo "exit=$?"
exit=1
$ wc -c out.txt err.txt
       1 out.txt        # a single newline
       0 err.txt        # nothing
```

Controls — each layer alone reports well; only the combination goes mute:

```console
$ rtk pytest --version
rtk: Failed to run pytest: Failed to spawn process: No such file or directory (os error 2)
# exit 1 — loud and clear

$ uv run pytest --version
error: Failed to spawn: `pytest`
  Caused by: No such file or directory (os error 2)
# exit 2 — uv's own message, also clear
```

## Expected

The spawn-failure message should survive the uv wrapper — stderr passthrough on spawn failure, or at minimum the same message rtk prints standalone.

## Notes

Theme-wise this seems related to the "error path goes silent while the exit code stays nonzero" class in #2994 (different mechanism — that one is non-UTF-8 line filtering on Windows), and trigger-wise to #2395 (missing binary misreported). Happy to run any diagnostics that would help narrow it down.

Thanks again — the tool pays for itself daily.

georgyia:
Fixes #3026

## Problem

`uv run rtk pytest` says nothing at all when pytest is missing from the venv — exit 1, a single newline on stdout, empty stderr. As the issue puts it, this is the dangerous quadrant: a test run that produces no output and no error text reads as a quiet pass, and the exit code is the only surviving signal.

```console
$ uv run rtk pytest --version >out.txt 2>err.txt; echo "exit=$?"
exit=1
$ wc -c out.txt err.txt
       1 out.txt        # a single newline
       0 err.txt        # nothing
```

## Root cause

uv is the trigger, not the culprit — and it isn't a spawn failure either.

Filters that compress stdout pass `RunOptions::stdout_only`, which keeps stderr out of the filter input. But `run_captured_filter` only ever emitted `raw_stderr` inside its `skip_filter_on_failure` branch. Callers that pair `stdout_only` with `early_exit_on_failure` (`ls`, `tree`) reach that branch and print stderr; callers that don't captured stderr and dropped it:

| `stdout_only` caller | `early_exit_on_failure` | stderr on failure |
|---|---|---|
| `ls`, `tree` | yes | shown |
| `pytest`, `ruff`, `prettier`, `go test`, `golangci-lint`, `phpunit`, `phpstan`, `pint`, `rubocop` | no | **dropped** |

Why only the *combination* goes mute: standalone `rtk pytest` looks fine because python is absent, so the spawn fails outright and the `anyhow` error prints loudly. Under `uv run` the venv supplies python, so rtk takes its `python -m pytest` fallback, **the spawn succeeds**, and the child's diagnostic goes to the stderr that was being discarded:

```console
$ uv run python -m pytest --version
/tmp/uvp/.venv/bin/python3: No module named pytest     # 51 bytes, exit 1
$ uv run rtk pytest --version
                                                        # 0 bytes, exit 1
```

## Fix

Emit stderr when the child fails:

```rust
if opts.filter_stdout_only && exit_code != 0 && !result.raw_stderr.trim().is_empty() {
    eprint!("{}", result.raw_stderr);
}
```

Only on failure — on a successful run stderr is the incidental noise RTK exists to suppress, and the filtered stdout already carries the result. This is the rule `ls`/`tree` already get via `early_exit_on_failure`, applied to the callers that were missing it. No double-printing: `FilterMode::CaptureOnly` buffers stderr without echoing, and the `skip_filter_on_failure` branch returns before this point.

**This likely covers more than #3026.** The table above is a family, and #2878 ("rtk prettier silently hides prettier 3 failures — stdout-only capture + [warn] filter") names this exact mechanism. I've deliberately scoped this PR to the shared runner defect and made no filter-specific claims; I haven't verified #2878 end-to-end, so I'm flagging the overlap rather than asserting a fix.

## Verification

| | before | after |
|---|---|---|
| `uv run rtk pytest --version` (missing) | exit 1, stderr **0 B** | exit 1, stderr **50 B**: `No module named pytest` |
| `uv run rtk pytest --version` (installed) | exit 0, stderr 0 B | exit 0, stderr **0 B** — unchanged |

Reproduced against real `uv` in a fresh `uv init` + `uv venv` project, both with and without pytest installed.

- New tests in `tests/stderr_passthrough_test.rs` put a failing shim on `PATH` (stderr message, empty stdout, nonzero exit) and drive it through `rtk ruff`, a bare `stdout_only` caller. `failing_tool_stderr_reaches_the_user` fails on the pre-fix runner and passes after; `successful_tool_stderr_stays_suppressed` passes both ways, pinning that no new noise is introduced on success.
- `cargo fmt --all --check` clean, `cargo clippy --all-targets` zero warnings, `cargo test --all` 2495 passed / 0 failed.

Verified on Linux (`rust:latest`); the tests are `#![cfg(unix)]` as they rely on a shell shim and `PermissionsExt`.

## Test plan

- [x] Manual testing: `uv run rtk pytest --version` with and without pytest in the venv
- [x] Token savings: N/A (error-path fix, no filter change)

### DIFF
--- a/src/core/runner.rs
+++ b/src/core/runner.rs
@@ -122,6 +122,15 @@ where
         return Ok(exit_code);
     }
 
+    // `filter_stdout_only` keeps stderr out of the filter input, so nothing
+    // downstream can surface it — a failing child's diagnostics would be dropped
+    // and the exit code left as the only signal (#3026). Emit them here. Only on
+    // failure: on a successful run stderr is the incidental noise RTK exists to
+    // suppress, and the filtered stdout already carries the result.
+    if opts.filter_stdout_only && exit_code != 0 && !result.raw_stderr.trim().is_empty() {
+        eprint!("{}", result.raw_stderr);
+    }
+
     let text_to_filter = if opts.filter_stdout_only {
         raw_stdout
     } else {
--- a/tests/stderr_passthrough_test.rs
+++ b/tests/stderr_passthrough_test.rs
@@ -1,0 +1,69 @@
+#![cfg(unix)]
+//! A failing tool's stderr must reach the user (#3026). Filters that compress
+//! stdout run with `RunOptions::stdout_only`, which keeps stderr out of the filter
+//! input — so unless the runner emits it, the diagnostics are dropped and the exit
+//! code becomes the only surviving signal.
+
+use std::fs;
+use std::os::unix::fs::PermissionsExt;
+use std::process::Command;
+
+const DIAGNOSTIC: &str = "shim-diagnostic: dependency is missing";
+
+/// A fake `ruff` that fails the way real tools do: message on stderr, nothing on
+/// stdout, nonzero exit. `rtk ruff` filters stdout only, which is the path under
+/// test; `resolved_command` resolves via PATH, so the shim takes precedence.
+fn shim(exit_code: i32) -> tempfile::TempDir {
+    let dir = tempfile::tempdir().unwrap();
+    let bin = dir.path().join("ruff");
+    fs::write(
+        &bin,
+        format!("#!/bin/sh\necho '{DIAGNOSTIC}' >&2\nexit {exit_code}\n"),
+    )
+    .unwrap();
+    fs::set_permissions(&bin, fs::Permissions::from_mode(0o755)).unwrap();
+    dir
+}
+
+fn rtk_ruff(shim_dir: &tempfile::TempDir, db: &tempfile::TempDir) -> std::process::Output {
+    let path = format!(
+        "{}:{}",
+        shim_dir.path().display(),
+        std::env::var("PATH").unwrap_or_default()
+    );
+    Command::new(env!("CARGO_BIN_EXE_rtk"))
+        .args(["ruff", "check", "."])
+        .env("PATH", path)
+        // Keep tracking off the developer's real database.
+        .env("RTK_DB_PATH", db.path().join("t.db"))
+        .output()
+        .expect("rtk")
+}
+
+#[test]
+fn failing_tool_stderr_reaches_the_user() {
+    let db = tempfile::tempdir().unwrap();
+    let out = rtk_ruff(&shim(1), &db);
+
+    assert_eq!(out.status.code(), Some(1), "exit code must propagate");
+    assert!(
+        String::from_utf8_lossy(&out.stderr).contains(DIAGNOSTIC),
+        "diagnostic swallowed — stderr was {:?}, stdout was {:?}",
+        String::from_utf8_lossy(&out.stderr),
+        String::from_utf8_lossy(&out.stdout)
+    );
+}
+
+#[test]
+fn successful_tool_stderr_stays_suppressed() {
+    // On success stderr is the incidental noise RTK exists to compress away, and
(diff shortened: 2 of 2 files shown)


KuSh (changes requested):
Hi @georgyia,

Thanks for your contribution. You’ll need to sign the CLA before we can move forward.

A few notes on your PR:
This will break `golangci_cmd.rs`, since it deliberately behaves that way for the “lint issues found” case, and possibly `rspec_cmd.rs` too.
`phpstan_cmd.rs` would also duplicate diagnostics.
Also, result.raw_stderr is captured by stream.rs's CaptureOnly thread, which doesn't warn on truncation the way the other capture paths there do (stream.rs:376,384) — worth adding that same 10 MiB warning, since this PR is what makes that stderr visible to the user for the first time.

Most of  this should be fixed by gating on stdout:
```rs
if opts.filter_stdout_only && exit_code != 0
    && result.raw_stdout.trim().is_empty()
    && !result.raw_stderr.trim().is_empty()
{
    eprint!("{}", result.raw_stderr);
}
```
The bug you’re trying to fix is really that `stdout` is empty or useless and the diagnostic is stuck in `stderr`, so this should only kick in when `stdout` is also empty. Most tools exit nonzero with populated `stdout`, which helps gate against a noise regression.

A few other points:
 - Token-savings tracking (`raw_for_tracking` / `shown`) doesn't include the new shown stderr bytes, so that should be fixed.
 - With those changes, you'll need a test covering the `stdout` non-empty + nonzero exit + benign `stderr` → no dump case.
 - Bonus: you could avoid `skip_filter_on_failure` duplication by adding an `emit_stderr_if_present` helper.

KuSh:
Thanks for tracking this down, the writeup of why only the stdout_only plus no-early-exit combination went mute was exactly right. This landed on develop on 2026-09-12 in #3772, which forwards a stdout-only filter's stderr in the same place in run_captured_filter and also counts those bytes in the savings total. It forwards unconditionally rather than only on failure, because tools like prettier write their whole report to stderr on a successful run, so the one behaviour your PR pinned differently is a deliberate choice there. Closing as covered, but please comment or reopen if you see a case that still comes out silent.

PR #1648, opened 2026-04-30
fix(hooks/claude): omit permissionDecision under bypassPermissions

iliaal:
## Summary

Under Claude Code's `bypassPermissions` mode (a.k.a. `--dangerously-skip-permissions`), rtk's PreToolUse hook fires correctly but its rewrites silently never land. The cause is undocumented Claude Code behavior: when `hookSpecificOutput` contains both `updatedInput` and `permissionDecision: "allow"`, Claude Code drops `updatedInput` under bypass. For users with curated permission allow-lists (where `verdict == Allow` is the common case), most rtk rewrites silently fail under bypass.

This patch detects bypass mode from the hook input's `permission_mode` field and omits `permissionDecision` accordingly. Default mode keeps the auto-allow optimization, which does real work there by skipping the re-prompt for the rewritten command.

## Reproducer

`~/.claude/settings.json`:
```json
{
  "permissions": {
    "defaultMode": "bypassPermissions",
    "allow": ["Bash(git status:*)"]
  },
  "hooks": {
    "PreToolUse": [{"matcher": "Bash", "hooks": [{"type": "command", "command": "rtk hook claude"}]}]
  }
}
```

Have an agent invoke an atomic `git status`.

- **Before:** raw 104-byte output (`On branch master / ... / nothing to commit`).
- **After:** rtk's compressed ~50-byte form (`* master...origin/master / clean — nothing to commit`).

Direct hook test:
```bash
echo '{"tool_name":"Bash","permission_mode":"bypassPermissions","tool_input":{"command":"git status"}}' \
  | rtk hook claude | jq .
# After patch: omits permissionDecision; updatedInput.command = "rtk git status".

echo '{"tool_name":"Bash","permission_mode":"default","tool_input":{"command":"git status"}}' \
  | rtk hook claude | jq .
# After patch: includes permissionDecision: "allow" (auto-allow optimization preserved).
```

## What changed

`src/hooks/hook_cmd.rs`:
- `is_bypass_mode(v: &Value) -> bool`: parses `permission_mode` from hook input. Anything other than `"bypassPermissions"` (including absent or unknown values) maps to non-bypass.
- `should_emit_allow_decision(verdict: &PermissionVerdict, bypass_mode: bool) -> bool`: pure helper.
- `process_claude_payload` routes through both.

## Tests

[...]

PR #3031, opened 2026-07-17
fix(hook): emit ask decision for Claude rewrites

### LINKED ISSUE #3018: rtk hook claude omits permissionDecision in updatedInput responses, causing Claude Code to silently ignore command rewrites

houtinghang:
## Summary

- emit `permissionDecision: "ask"` for Claude Code rewrites that require confirmation
- preserve decision omission for ask rewrites under `bypassPermissions`
- add regression coverage for both default and bypass permission modes

## Root cause

`process_claude_payload` only emitted a permission decision for explicit allow rules. In the default-to-ask path, it returned `updatedInput` without the corresponding `ask` decision, so Claude Code could ignore the rewritten command and execute the original input.

The bypass-mode exception is intentional: Claude Code currently applies `updatedInput` under `bypassPermissions` only when the decision is omitted. Explicit allow behavior is unchanged and remains covered separately by #1648.

## Impact

Default and explicit-ask configurations now prompt against the rewritten `rtk ...` command instead of silently losing the rewrite. Explicit allow and bypass behavior remain unchanged.

Fixes #3018

## Validation

- `cargo fmt --all --check`
- `cargo test hooks::hook_cmd::tests:: -- --nocapture` (91 passed)
- `cargo test` (2,434 unit tests and 11 integration tests passed; 8 ignored)

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -385,6 +385,11 @@ fn process_claude_payload(v: &Value) -> PayloadAction {
             .as_object_mut()
             .unwrap()
             .insert("permissionDecision".into(), json!("allow"));
+    } else if v.get("permission_mode").and_then(Value::as_str) != Some("bypassPermissions") {
+        hook_output
+            .as_object_mut()
+            .unwrap()
+            .insert("permissionDecision".into(), json!("ask"));
     }
 
     PayloadAction::Rewrite {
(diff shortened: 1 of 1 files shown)


KuSh:
See [my comment](https://github.com/rtk-ai/rtk/issues/3018#issuecomment-5230869713) on the issue. We intentionally omit the `ask` `permissionDecision` so Claude can ask about it and remember decisions

PR #3041, opened 2026-07-17
fix(hooks): apply exclude_commands to the resolved tool, not just the raw command

### LINKED ISSUE #3035: [hooks] exclude_commands is matched against the raw command, not the resolved tool — "pytest" fails to exclude `python -m pytest`
**Environment:** rtk 0.37.2 · Ubuntu 24.04 (WSL2) · Claude Code hook (`rtk hook claude`)
**Confirmed present on master** (`5d32d07`) by reading `src/discover/registry.rs` — see Cause.

## Summary

`[hooks] exclude_commands` is matched against the **raw command text**, but `classify_command` resolves wrapper/module forms to a **different** tool. When the two disagree, the exclusion silently fails and the command is rewritten anyway.

The clearest case is `python -m <tool>`: excluding `"pytest"` does **not** exclude `python3 -m pytest`, even though rtk classifies that command as `rtk pytest`.

## Reproduction

```toml
# ~/.config/rtk/config.toml
[hooks]
exclude_commands = ["pytest"]
```

```console
$ rtk rewrite "pytest tests/ -q"
$ echo $?              # 1 — correctly excluded ✓

$ rtk rewrite "python3 -m pytest tests/ -q"
rtk pytest tests/ -q   # rewritten despite the exclusion ✗
```

Not pytest-specific — same for any module-form tool:

```console
# exclude_commands = ["mypy"]
$ rtk rewrite "mypy --strict ."
$ echo $?                             # 1 — excluded ✓
$ rtk rewrite "python3 -m mypy --strict ."
rtk mypy --strict .                   # rewritten despite the exclusion ✗
```

## Cause

`rewrite_segment_inner` (`src/discover/registry.rs`, `Classification::Supported` arm, ~L884-891):

```rust
let rtk_equivalent = match classify_command(cmd_part) {
    Classification::Supported { rtk_equivalent, .. } => {
        if is_excluded(cmd_clean, excluded) {   // <-- raw text, not rtk_equivalent
            return None;
        }
        rtk_equivalent
    }
```

`is_excluded` prefix-matches (~L798):

```rust
ExcludePattern::Prefix(prefix) => cmd.starts_with(prefix.as_str()),
```

So for `python3 -m pytest tests/`:

- `classify_command` → `Classification::Supported { rtk_equivalent: "rtk pytest", .. }` (asserted by the existing test `test_classify_python_m_pytest`, L2477)
- `is_excluded("python3 -m pytest tests/", ["pytest"])` → `"python3 -m pytest tests/".starts_with("pytest")` → **false**
- → not excluded → rewritten to `rtk pytest tests/`

The exclusion check never sees the tool the classifier actually resolved. Note the existing rewrite tests for these forms (`test_rewrite_python_m_pytest` L2534, `test_python3_m_pytest` L4260) all pass `&[]` as the excluded list, so this path is untested with a non-empty exclusion.

## Why it bites in practice

`python -m pytest` and bare `pytest` are **not interchangeable**: `-m` runs the *current interpreter* and puts CWD on `sys.path`; bare `pytest` runs whichever `pytest` shim is first on PATH. On this machine they are different interpreters (3.10 vs 3.11).

So the rewrite silently swapped the interpreter, imports failed, and pytest exited 2 (collection error) — which the pytest summarizer reported as `Pytest: No tests collected` (#3032), hiding the `ModuleNotFoundError`:

```console
$ python3 -m pytest tests/ -q      # hook rewrites this to `rtk pytest tests/ -q`
Pytest: No tests collected

$ rtk proxy python3 -m pytest tests/ -q
41 passed
```

The exclusion existed precisely to prevent this, and was bypassed. Two agents in one session read the filtered text and concluded working code was broken. (`[tee] mode = "failures"` did preserve the real traceback — that escape hatch worked.)

## Suggested fix

Also test the resolved target, e.g. in the `Supported` arm:

```rust
Classification::Supported { rtk_equivalent, .. } => {
    // exclusion should apply to the tool rtk actually resolved to, not just the raw text
    let target = rtk_equivalent.strip_prefix("rtk ").unwrap_or(rtk_equivalent);
    if is_excluded(cmd_clean, excluded) || is_excluded(target, excluded) {
        return None;
    }
    rtk_equivalent
}
```

That makes `exclude_commands = ["pytest"]` cover every form rtk maps to `rtk pytest` (`python -m pytest`, `python3 -m pytest`, …) without needing users to enumerate wrappers.

A regression test would be `rewrite_command_no_prefixes("python3 -m pytest tests/", &[ExcludePattern::Prefix("pytest".into())])` → `None`.

## Workaround

Exclude the interpreter as well as the tool:

```toml
exclude_commands = ["pytest", "python3", "python"]
```

This costs nothing today, since bare `python3 script.py` has no rewrite rule anyway (#2312, #928, #285 are open FRs to add one) — the only effect of the `python -m` path is to route to the wrapped tool's filter.

## Related

- #2823 — `exclude_commands` bypassed for `head`/`tail` (line-range fast path returns before the check). Same family: the exclusion is not consistently applied to what rtk actually runs. That one returns *before* the check; this one runs the check against the *wrong string*.
- #1919 — `exclude_commands` doesn't cover subcommands (`git diff`); also a root-token-matching limitation.
- #917 — `exclude_commands` ignores env-var-prefixed commands.
- #3032 — the masking that made this hard to diagnose (collection error → "No tests collected").

AnayGarodia:
Fixes #3035.

## Problem

`exclude_commands` is matched against the raw command text, but `classify_command` resolves module/wrapper forms to a different tool. With `exclude_commands = ["pytest"]`:

- `pytest tests/ -q` → excluded ✓
- `python3 -m pytest tests/ -q` → rewritten to `rtk pytest tests/ -q` ✗

The exclusion never sees the tool the classifier actually resolved to, so the rewrite silently swaps the interpreter (`-m` runs the current interpreter with CWD on `sys.path`; the bare shim may be a different Python entirely) — which is exactly the failure the user's exclusion existed to prevent.

## Fix

In the `Classification::Supported` arm of `rewrite_segment_inner`, also test the resolved rtk target (with the `rtk ` prefix stripped) against the exclusion list:

```rust
let resolved_tool = rtk_equivalent.strip_prefix("rtk ").unwrap_or(rtk_equivalent);
if is_excluded(cmd_clean, excluded) || is_excluded(resolved_tool, excluded) {
    return None;
}
```

`compile_exclude_patterns` anchors non-regex patterns as `^escaped($|\s)`, so matching the bare tool name is word-boundary safe: `"py"` does not exclude `pytest` (covered by a test).

## Verified

Unit tests plus end-to-end against a debug build with `exclude_commands = ["pytest"]` in config.toml:

```
pytest tests/ -q            =>  (excluded, exit 1)
python3 -m pytest tests/ -q =>  (excluded, exit 1)   # was: rewritten
python -m pytest tests/     =>  (excluded, exit 1)   # was: rewritten
python3 -m mypy --strict .  => rtk mypy --strict .   # unrelated tool still rewrites
```

Full suite: 2438 passed. Three new regression tests (`test_exclude_covers_python_m_form`, `test_exclude_resolved_tool_requires_word_boundary`, `test_exclude_other_tool_does_not_hit_python_m_form`).

Related but not addressed here: #2823 (head/tail fast path returns before the exclusion check — different code path), #1919, #917.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -883,7 +883,14 @@ fn rewrite_segment_inner(
         Classification::Supported { rtk_equivalent, .. } => {
             let stripped = ENV_PREFIX.replace(cmd_part, "");
             let cmd_clean = stripped.trim();
-            if is_excluded(cmd_clean, excluded) {
+            // The exclusion has to cover the tool the classifier resolved to,
+            // not just the raw text: `python -m pytest` classifies as
+            // `rtk pytest`, so `exclude_commands = ["pytest"]` must stop it
+            // even though the raw command starts with `python` (#3035).
+            let resolved_tool = rtk_equivalent
+                .strip_prefix("rtk ")
+                .unwrap_or(rtk_equivalent);
+            if is_excluded(cmd_clean, excluded) || is_excluded(resolved_tool, excluded) {
                 return None;
             }
             rtk_equivalent
@@ -3722,6 +3729,34 @@ mod tests {
     fn test_exclude_does_not_match_hyphenated_command() {
         let excluded = vec!["golangci".to_string()];
         assert!(rewrite_command_no_prefixes("golangci-lint run ./...", &excluded).is_some());
+    }
+
+    // #3035: the exclusion must cover the resolved tool, not just the raw text,
+    // so module/wrapper forms don't slip past it.
+    #[test]
+    fn test_exclude_covers_python_m_form() {
+        let excluded = vec!["pytest".to_string()];
+        assert_eq!(
+            rewrite_command_no_prefixes("python3 -m pytest tests/ -q", &excluded),
+            None
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("python -m pytest tests/", &excluded),
+            None
+        );
+    }
+
+    #[test]
+    fn test_exclude_resolved_tool_requires_word_boundary() {
+        // "py" must not exclude commands resolving to "pytest"
+        let excluded = vec!["py".to_string()];
+        assert!(rewrite_command_no_prefixes("python3 -m pytest tests/", &excluded).is_some());
+    }
+
+    #[test]
+    fn test_exclude_other_tool_does_not_hit_python_m_form() {
+        let excluded = vec!["mypy".to_string()];
+        assert!(rewrite_command_no_prefixes("python3 -m pytest tests/", &excluded).is_some());
     }
 
     #[test]


KuSh:
Superseded by #3749, which lands the same fix with a wider net.

Your diagnosis in #3035 is the sharper write-up of the two — `python3 -m pytest` slipping past `exclude_commands = ["pytest"]` is exactly the failure, and the interpreter-swap consequence you spelled out is why it matters rather than just being untidy. #3749 will carry `Fixes #3035`.

One finding from reviewing both, since it applies to your patch as much as mine: matching the **bare** resolved target breaks end-anchored user regexes. `exclude_commands = ["^ls$"]` means "exclude bare `ls`, keep rewriting `ls -la`" — but the target check hands `is_excluded` the token `ls` instead of a command line, so every `ls` gets excluded. Verified on a build, before vs. after the patch:

```
["^ls$"]     ls -la       develop: rtk ls -la      patched: EXCLUDED
["^cargo$"]  cargo build  develop: rtk cargo build patched: EXCLUDED
```

#3749 is being reworked to strip the wrapper prefix and match the *remaining* command rather than substituting the target name, which keeps the arguments the regex was written against.

Thanks for this — the issue was the valuable half, and it's what drove the fix.

Issue #2767, opened 2026-07-01
`rtk bash -c` and `rtk fish -c` wrapping


ccoVeille:
Sometimes the Agent needs to run command via `bash -c` or whatever the shell

For example, for running this

```shell
head foo && grep -R bar .
```

It would launch

```shell
bash -c "head foo && grep -R bar ."
```

rtk is able to rewrite the command

```console
$ rtk rewrite 'head foo && grep -R bar .'

rtk read foo && rtk grep -R bar .
```

But it does nothing when bash -c or fish -c is used

```console
$ rtk rewrite 'bash -c "head foo && grep -R bar ."'

```

I would expect this to be wrapped

```console
$ rtk rewrite 'bash -c "head foo && grep -R bar ."'

bash -c "rtk read foo && rtk grep -R bar ."
```

ccoVeille:
Here is a piece of code I wrote to fix this, but I would like rtk to support it natively

Note: I use fish as a shell

```fish
function rtk_wrapper --description "Rewrite-first RTK command router with safe fallback"
    if test (count $argv) -eq 0
        echo "Usage: rtk_wrapper <command ...>" >&2
        return 64
    end

    set raw (string join " " -- $argv)

    # Special case: shell -c "..." should rewrite the inner script, not only the wrapper command.
    if test (count $argv) -ge 3
        set shell_cmd $argv[1]
        set shell_flag $argv[2]
        if contains -- $shell_cmd fish sh bash zsh; and test "$shell_flag" = "-c"
            set script $argv[3]
            set passthrough_args $argv[4..-1]

            set classification (__rtk_wrapper_classify "$script")
            set action $classification[1]
            set value (string join "\n" -- $classification[2..-1])

            switch $action
                case rewrite
                    command $shell_cmd -c "$value" $passthrough_args
                    return $status
                case deny
                    echo "rtk_wrapper: denied by rtk rewrite policy: $script" >&2
                    return 2
            end

            command $shell_cmd -c "$script" $passthrough_args
            return $status
        end
    end

    set classification (__rtk_wrapper_classify "$raw")
    set action $classification[1]
    set value (string join "\n" -- $classification[2..-1])

    switch $action
        case rewrite
            eval $value
            return $status
        case deny
            echo "rtk_wrapper: denied by rtk rewrite policy: $raw" >&2
            return 2
    end

    # Exit 1 (no rewrite) or any unexpected rewrite error: run raw command.
    eval $raw
    return $status
end


[...]

PR #3057, opened 2026-07-18
fix(cli)!: preserve argv boundaries and rewrite quoted shell wrappers

pashifika:
## Summary

This pull request makes RTK's shell boundary explicit across CLI execution and hook rewriting. Generic runners now preserve argv through direct execution, intentional scripts require an explicit command-string mode, ambiguous shell syntax defers unchanged, and exact quoted `sh`, `bash`, `zsh`, and `fish` wrappers can safely rewrite portable inner commands.

This addresses four related limitations in the current execution model:

- `rtk run`, `rtk err`, `rtk test`, and `rtk summary` rebuilt positional arguments with spaces and executed the result through `sh -c` on Unix or `cmd /C` on Windows, which lost argument boundaries and enabled unintended shell expansion.
- Agent hosts can expose a fish login shell through `$SHELL` while evaluating commands with zsh or a POSIX-compatible shell, so guessing the parser from the environment is unreliable.
- The hook safety gate recognized POSIX substitutions but previously missed bare fish command substitution and command-position fish control keywords, allowing uncertain input to be rewritten or auto-allowed.
- Quoted command strings such as `bash -c "head foo && grep -R bar ."` were treated as opaque wrapper arguments, so the portable commands inside them could not use RTK rewrites as reported in #2767.

The host still owns the shell that evaluates hook output. This change makes RTK's own execution and rewrite boundaries explicit; it does not translate syntax between fish, zsh, and POSIX shells or change the host executor.

### Execution model and migration

The updated CLI separates direct argv execution from intentional command-string execution:

- Positional forms such as `rtk run <cmd> [args...]`, `rtk err <cmd> [args...]`, `rtk test <cmd> [args...]`, and `rtk summary <cmd> [args...]` execute the supplied program directly and preserve every argument boundary.
- `rtk run -c '<script>'` retains the existing platform default of `sh -c` on Unix and `cmd /C` on Windows.
- `rtk run --shell fish -c '<script>'` selects a shell explicitly. PowerShell and `pwsh` use `-Command`; other named shells use `-c`.
- `rtk err`, `rtk test`, and `rtk summary` accept `--shell <name> '<script>'` when filtering an intentional shell command.
- Explicit shell mode requires the complete script as one quoted argument, preventing RTK from reconstructing quoting after Clap has parsed argv.
- Installed Codex guidance now recommends explicit `rtk run --shell <name> -c '<script>'` for shell-specific scripts and portable syntax otherwise.

This is a breaking behavior change for callers that relied on implicit shell expansion in positional arguments. Those callers must move shell syntax to `rtk run -c '<script>'` or use `--shell <name>` with one quoted script.

### Conservative hook handling

The shared lexer now treats the following forms as unattestable before rewrite and permission decisions:

- command and process substitution, including bare fish `(...)` syntax;
- parenthesized shell constructs and incomplete quoting or trailing escapes;
- file-target redirects;
- fish control keywords such as `and`, `or`, `if`, `switch`, and `end` when they occur at a command boundary.

Keyword detection is command-position aware, so ordinary arguments such as `rg and src` remain attestable. Ambiguous constructs defer to the host unchanged and resolve to `Ask` rather than `Allow`, even when a wildcard allow rule is configured.

### Quoted shell wrapper rewriting

The rewrite pipeline now recognizes a deliberately narrow command-string wrapper subset:

- Exact `sh -c`, `bash -c`, `zsh -c`, and `fish -c` wrappers are eligible when the script is one balanced single- or double-quoted argument.
- RTK rewrites the existing portable compound-command subset inside the script and replaces only the original script byte span, preserving the shell path, whitespace, quote delimiters, suffix arguments, and file-descriptor redirects.
- Double-quoted scripts are accepted only when the outer shell has no active `$` expansion, backticks, or backslash escapes to evaluate.
- Unquoted or malformed scripts, additional shell options, file redirects, fish-specific control syntax, and nested wrappers pass through unchanged.
- Wrapper rewriting is limited to one level and never grants automatic permission. Ask-capable hosts request confirmation, hosts without an explicit ask response preserve their native permission flow, and an inner deny rule takes precedence over an outer allow rule.

The wrapper parser is shared by rewriting and permission evaluation so the two paths cannot disagree about the supported boundary. Agent protocol tests cover Claude, Cursor, Gemini, Droid, and Copilot behavior.

### Key files changed

The following files are the main review entry points:

| File | Purpose |
|---|---|
| `src/core/shell.rs` | Centralizes direct command construction, explicit shell invocation, platform flags, shell resolution, and script-boundary validation. |
| `src/main.rs` | Adds the `--shell` CLI surface and routes positional commands through direct argv execution. |
| `src/discover/lexer.rs` | Detects parenthesized syntax, command-position control keywords, and incomplete quoting as unattestable. |
| `src/discover/shell_wrapper.rs` | Parses the conservative quoted `-c` wrapper subset and exposes byte spans without decoding or re-quoting scripts. |
| `src/discover/registry.rs` | Applies one-level inner compound rewrites while preserving the original wrapper and passthrough contract. |
| `src/hooks/permissions.rs` | Prevents wrapper auto-allow and evaluates inner deny rules before returning `Ask`. |
| `src/hooks/hook_cmd.rs` | Preserves each supported agent's rewrite and permission protocol for shell wrappers. |
| `hooks/codex/rtk-awareness.md` | Documents explicit shell selection for generated Codex RTK guidance. |
| `tests/run_shell_test.rs` | Covers literal argv preservation, exit codes, default and explicit shells, missing shells, and split-script rejection. |
| `tests/rewrite_shell_wrapper_test.rs` | Covers the issue example, portable fish scripts, conservative passthrough, output, and rewrite exit codes. |

## Test plan

- [x] Run `cargo fmt --all --check`.
- [x] Run `cargo clippy --all-targets -- -D warnings`.
- [x] Run focused shell-wrapper tests, including 27 parser, registry, permission, and agent protocol tests plus the 3 CLI integration tests in `tests/rewrite_shell_wrapper_test.rs`.
- [x] Run `env RIPGREP_CONFIG_PATH=/dev/null cargo test --all` (2,478 unit tests passed, 8 ignored; all integration suites passed).
- [x] Run `env RIPGREP_CONFIG_PATH=/dev/null cargo test --all -- --ignored --nocapture` (8 ignored integration tests passed).
- [x] Run `cargo build --release`.
- [x] Verify direct argv execution, explicit `sh`, `zsh`, and `fish` execution, and quoted `bash -c` and `fish -c` rewriting on macOS.
- [x] Verify ambiguous shell constructs pass through, wildcard allow rules cannot auto-allow wrappers, and inner deny rules take precedence.
- [x] Compare release size, resident memory, and repeated startup/rewrite timings with pre-change release binaries; no material regression was observed. The shell-wrapper change added 16 bytes, with peak memory remaining approximately 1.3 MB.
- [ ] Let CI validate Linux and Windows execution, including `cmd /C`, PowerShell flag selection, and platform-independent wrapper parsing.

The local ripgrep configuration injects `--max-columns=150`, which shortens the raw comparison output used by the existing `bulky_grep_yields_token_savings` test. The unsanitized `cargo test --all` run therefore fails that unrelated assertion; the complete suite passes when `RIPGREP_CONFIG_PATH` is cleared as shown above. This pull request does not change search behavior.

The broad `scripts/test-all.sh` suite was not run because it uses an installed binary and includes `git fetch`, external HTTP calls, and unrelated ecosystem commands.

🤖 Generated with [Codex](https://openai.com/codex/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,8 +184,9 @@ rtk go test                     # Go tests (NDJSON, -90%)
 rtk cargo test                  # Cargo tests (-90%)
 rtk rake test                   # Ruby minitest (-90%)
 rtk rspec                       # RSpec tests (JSON, -60%+)
-rtk err <cmd>                   # Filter errors only from any command
-rtk test <cmd>                  # Generic test wrapper - failures only (-90%)
+rtk err <cmd> [args...]         # Direct argv execution, errors/warnings only
+rtk test <cmd> [args...]        # Direct argv execution, failures only (-90%)
+rtk err --shell fish '<script>' # Explicit shell for shell-specific syntax
 ```
 
 ### Build & Lint
@@ -255,9 +256,20 @@ rtk env -f AWS                  # Filtered env vars
 rtk log app.log                 # Deduplicated logs
 rtk curl <url>                  # Truncate + save full output
 rtk wget <url>                  # Download, strip progress bars
-rtk summary <long command>      # Heuristic summary
+rtk summary <cmd> [args...]     # Direct argv execution + heuristic summary
+rtk run <cmd> [args...]         # Raw direct execution (no filtering/tracking)
+rtk run -c '<script>'           # Shell string via sh (cmd on Windows)
+rtk run --shell fish -c '<script>' # Explicit shell for shell-specific syntax
 rtk proxy <command>             # Raw passthrough + tracking
 ```
+
+`rtk run`, `rtk err`, `rtk test`, and `rtk summary` preserve positional
+argument boundaries and do not expand globs, variables, or operators by
+default. Use `-c` with `rtk run`, or `--shell <name>` with the filtered
+wrappers, only when a command intentionally requires shell syntax. Pass an
+explicit shell script as one quoted argument; RTK does not infer the parser
+from `$SHELL` because the environment value may differ from the actual command
+executor.
 
 ### Token Savings Analytics
 ```bash
--- a/README.md
+++ b/README.md
@@ -329,6 +329,20 @@ The most effective way to use rtk. The hook transparently intercepts Bash comman
 
 **Scope note:** this only applies to Bash tool calls. Claude Code built-in tools such as `Read`, `Grep`, and `Glob` bypass the hook, so use shell commands or explicit `rtk` commands when you want RTK filtering there.
 
+Simple quoted shell wrappers are also rewritten without changing the selected
+shell:
+
+```bash
+bash -c "head foo && grep -R bar ."
+# → bash -c "rtk read foo && rtk grep -R bar ."
+```
+
+This support is intentionally conservative: it covers exact `sh -c`,
+`bash -c`, `zsh -c`, and `fish -c` wrappers with a quoted portable script. Shell
+expansion in an outer double quote, additional shell options, redirects to
+files, fish-specific control syntax, and nested wrappers pass through
+unchanged.
+
 ### Setup
 
 ```bash
--- a/docs/contributing/TECHNICAL.md
+++ b/docs/contributing/TECHNICAL.md
@@ -92,6 +92,14 @@ When an LLM agent runs a command (e.g., `git status`):
 
 All rewrite logic lives in Rust (`src/discover/registry.rs`). Hooks are thin delegates that handle agent-specific JSON formats.
 
+Before a hook emits `updatedInput`, the shared lexer rejects constructs whose
+shell semantics cannot be safely attested: command/process substitution,
+parenthesized syntax (including fish command substitution), incomplete
+quoting, file-target redirects, and shell control keywords at command
+boundaries. These commands are returned to the host unchanged. The host still
+owns command parsing and shell selection; RTK never translates syntax between
+fish, POSIX shells, and zsh.
+
 > **Details**: [`hooks/README.md`](../hooks/README.md) covers each agent's JSON format, the rewrite registry, compound command handling, and the `RTK_DISABLED` override.
 
 #### Rewrite Pipeline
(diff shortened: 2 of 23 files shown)


ccoVeille:
I cannot test for now, but thanks.

I'm curious about feedbacks that you will get

SindriFr:
@pashifika — heads up that I rebased this work onto current `develop` and opened it as #3828, because I needed wrapper rewriting locally and this PR no longer applied. You are the commit author there and the code is overwhelmingly yours: `shell_wrapper.rs` is your file essentially unchanged, and most of the registry, permissions and hook_cmd hunks and tests are yours too.

What changed in the rebase:

- `registry.rs` needed reintegrating — the rewrite engine has since gained `RewriteContext`, `PipeKind`, `analyze_pipeline`, `pipeline_final_safe` and multi-line block handling, so `shell_wrapper_depth` has to thread through all of it.
- I dropped your `lexer.rs`, `main.rs` and `core/shell.rs` changes (~280 lines) to keep the PR to the rewrite path alone. Your argv-boundary work in those files is a separate concern and is not in #3828.
- `develop` hardened the `sudo` passthrough after your branch point (#146/#3569), so `sudo bash -c '…'` now passes through instead of rewriting. I pinned that as intentional rather than restoring the old behaviour — the inner `rtk` would run as root, where it is off `secure_path`.

I would rather not step on your PR. If you would prefer to refresh this one yourself, say so and I will close #3828 — happy either way, and no attribution concerns from my side. If you have no plans to pick it back up, #3828 is ready and should close this when it lands.

pashifika:
@SindriFr thanks for the rebase — no attribution concerns here either, and dropping the `lexer.rs` / `main.rs` / `core/shell.rs` hunks was the right call: #3828 reviews much better as the rewrite path alone. Your `sudo` decision is correct too — the inner `rtk` off `secure_path` is exactly why it must pass through.

Closing this in favour of #3828 plus #3911, which is the rest of this PR rebased onto your branch: the argv-boundary work, the fish-aware unattestable gate, `fish -c` back in the wrapper subset (that gate is the guarantee #3828 was missing), and unambiguously-fish scripts wrapped as `rtk run --shell fish -c '<script>'`. #3828 stays first in the stack, so merging it does not need to wait on #3911.

PR #3059, opened 2026-07-18
feat(kimi): support uninstalling the AGENTS.md instructions

### LINKED ISSUE #3055: feat(kimi): add uninstall_kimi for AGENTS.md block removal (parity with Codex)
## Context

Kimi AI support was added in #1478 (merged). It writes an RTK instructions block into the project-root `AGENTS.md` via `run_kimi_mode` (same mechanism as Codex).

However, unlike Codex, there is **no `uninstall_kimi`** path. Running `rtk init --agent kimi --uninstall` currently falls through to the default `uninstall_standard` handler and does **not** remove the Kimi RTK block.

## Proposed fix

Add uninstall parity with Codex:

1. `uninstall_kimi_at(base_dir, ctx)` — read project `./AGENTS.md`, call the existing `remove_rtk_block(...)`, write back. Mirror `uninstall_codex_at` (init.rs).
2. Thread a `kimi` bool through `uninstall_standard` / `uninstall_init_dispatch` (main.rs), like the other agents.
3. Add a test mirroring `test_kimi_mode_writes_agents_md` that installs then uninstalls and asserts the block is gone.

## Note

Kimi and Codex share the same project-root `AGENTS.md` and a single RTK block. Uninstalling either removes the shared block — acceptable (the block is idempotent/shared), but worth a doc line.

## Effort
Small (~20-30 lines). The `remove_rtk_block` machinery already exists.

sahenjp:
## Summary

- route `rtk init --agent kimi --uninstall` through the standard uninstall dispatcher
- remove only the RTK-managed block from the project-level `AGENTS.md`
- preserve user-authored instructions and support dry-run and no-op behavior
- reject global scope because Kimi configuration is project-scoped
- document the uninstall command and the block shared by project-scoped Codex and Kimi installations

Closes #3055.

## Testing

- `cargo fmt --all --check`
- `cargo test`
- `cargo build`
- `git diff --check`

Added regression coverage for dispatcher routing, RTK block removal, preservation of user-authored content, dry-run behavior, an unconfigured project, and rejection of global scope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,9 +417,12 @@ For the full config reference (all sections, env vars, per-project filters), see
 
 ```bash
 rtk init -g --uninstall     # Remove hook, RTK.md, settings.json entry
+rtk init --agent kimi --uninstall  # Remove RTK instructions from this project's AGENTS.md
 cargo uninstall rtk          # Remove binary
 brew uninstall rtk           # If installed via Homebrew
 ```
+
+Project-scoped Codex and Kimi installations share the same RTK-managed block in the project-root `AGENTS.md`. Uninstalling either integration removes that shared block.
 
 ## Documentation
 
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -627,18 +627,27 @@ fn remove_hook_from_settings(ctx: InitContext) -> Result<bool> {
     Ok(removed)
 }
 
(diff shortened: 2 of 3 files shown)


KuSh:
Closing in favour of #3105, which adds the same Kimi uninstall path through the standard dispatcher and has green CI plus the supported agents guide entry. Your extra regression cases for dry run and the unconfigured project would be welcome on that PR.

Issue #3058, opened 2026-07-18
Playwright parser fails outright, `rtk find` errors instead of passing through, eval path breaks on bare `===`, and `cc-economics` cannot parse current ccusage (evidence from 28k command corpus)


valtumi:
**Version:** rtk 0.43.0 (behavior confirmed on 0.34.2 and 0.43.0), macOS 26.4 arm64, zsh, Claude Code hook v3

First: thanks for rtk. Over 50,377 commands it reports 157.6M tokens saved at 85.3%. The issues below are the residue, measured from 2,787 Claude Code session transcripts (28,236 Bash calls, 31 days).

## 1. Playwright parser fails 100% of the time (96/96 runs)

Every playwright invocation in the corpus produced:

```
[RTK:PASSTHROUGH] playwright parser: All parsing tiers failed
```

followed by the raw JSON dumped anyway - so it is worse than no rtk: zero savings plus a banner plus, in 41 cases, a tee-log pointer forcing a second round trip.

Notably **90 of the 96 already passed `--reporter=list`**, so this is not a "user forced the JSON reporter" case. It looks like rtk injects its own reporter and then cannot parse its own output. Repro shape:

```
TEST_REPORT_ID=abc PLAYWRIGHT_BASE_URL="https://example.app" npx playwright test tests/e2e/foo.spec.ts --reporter=list 2>&1
```

Suggested fix: when all parsing tiers fail, fall back to *native* output rather than the injected-reporter output, and drop the tee indirection on failure.

## 2. `rtk find` errors on compound predicates instead of passing through

```
$ rtk find . -name "seed-users*" -not -path "*/node_modules/*"
rtk: rtk find does not support compound predicates or actions (e.g. -not, -exec). Use `find` directly.
$ echo $?
1
```

Since the hook has already rewritten the command, the user's `find` never runs - the turn is simply lost (14 events). The registry knows at rewrite time that the command contains `-not`/`-exec`; it should return the no-equivalent exit code so the hook passes it through unchanged, rather than rewriting into a command that is guaranteed to fail.

## 3. eval path breaks on bare `===` under zsh (EQUALS expansion)

```
$ rtk proxy git diff --stat; echo ===
(eval):1: == not found
```

[...]

PR #3060, opened 2026-07-18
fix(playwright): preserve explicit reporter output

floze-the-genius:
## Summary

- preserve user-supplied Playwright reporters in both `--reporter=list` and `--reporter list` forms
- return the requested native reporter output without JSON parsing or tee fallback
- keep the existing injected JSON reporter path unchanged when no reporter is supplied

Addresses the Playwright reporter portion of #3058. This intentionally does not close the multi-part issue.

## Test plan

- [x] exact rebase onto current `develop` `9936b2b9`; original patch-id unchanged
- [x] `cargo test --all` — 2,658 passed, 0 failed, 8 ignored
- [x] `cargo build --release`
- [x] `bash scripts/check-test-presence.sh origin/develop`
- [x] subprocess harness: both explicit reporter forms avoid JSON injection and preserve the child exit code
- [x] `git diff --check`

## AI assistance

OpenAI Codex was used to investigate the issue, implement the change and regression tests, and run the validation commands. The contribution is submitted with this disclosure and a DCO sign-off.

### DIFF
--- a/src/cmds/js/playwright_cmd.rs
+++ b/src/cmds/js/playwright_cmd.rs
@@ -235,6 +235,30 @@ fn extract_failures_regex(output: &str) -> Vec<TestFailure> {
     failures
 }
 
+fn has_explicit_reporter(args: &[String]) -> bool {
+    args.iter()
+        .any(|arg| arg == "--reporter" || arg.starts_with("--reporter="))
+}
+
+fn append_playwright_args(cmd: &mut std::process::Command, args: &[String]) -> bool {
+    let is_test = args.first().is_some_and(|arg| arg == "test");
+    if !is_test {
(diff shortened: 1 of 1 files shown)


floze-the-genius:
@aeppling I exact-rebased this one-file fix onto current develop (d20ccecb, 0 behind; patch-id unchanged). It covers the explicit-reporter portion of high-priority/help-wanted #3058: both --reporter=list and --reporter list avoid JSON injection and preserve the child exit code, while the no-reporter path is unchanged. Fresh fmt, all-targets Clippy, 2,658 tests, release build, test-presence, subprocess harness, and diff check all pass. The new CI run is stopped at action_required before jobs start. Could you approve the workflow and review when convenient?

PR #3061, opened 2026-07-18
fix(uv): preserve program stdout and restore inner-command filtering

aeppling:
## Summary

- **Data loss:** every successful `uv run` collapsed to `ok`, discarding the program's output with no way to recover it — `rtk uv run python script.py` printed `ok` instead of the script's result. On success, stdout and stderr are now passed through; only uv's own resolution chatter (`Resolved/Installed/Using CPython/…`) is stripped. Output is bounded (50 lines head+tail, 500 chars/line) with a `force_tee_*` hint so nothing is unrecoverable.
- **Lost compression:** #2768 removed `uv run` from the transparent prefixes, so `uv run pytest` fell back to the generic uv extractor instead of the pytest filter. Restored — with the fall-through repair that made #2768 necessary: delegate when the inner command has a filter, otherwise fall through to `rtk uv run` (so `uv run python script.py` is still filtered, which was the original conflict). Fall-through is restricted to wrappers that RULES can match as a whole string; shell keywords (`exec`, `noglob`, …) keep the old behaviour, since `rtk exec foo` is not spawnable.
- Failure path unchanged — same traceback extraction, same exit-code propagation.

Measured on the same project (whitespace tokens):

| `uv run pytest` | raw | before | after |
|---|---|---|---|
| passing | 28 | 28 (0%) | **3 (89%)** — `Pytest: 4 passed` |
| failing | 78 | 45 (42%) | **30 (62%)** — `Pytest: 3 passed, 1 failed` + `file:line` |

Cold run with uv install noise: 11 → 3 tokens. Data script: output preserved (was `ok`). A 163 KB single-line JSON is capped to 550 bytes plus a recovery hint.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — 2505 passed, 0 failed, 0 clippy warnings
  - routing matrix: delegation (`uv run pytest|ruff|mypy`), fall-through (`python script.py`, `alembic`, `--with`, `-m`), env prefixes, `sudo`, pipes/redirects/compound, nesting (`exec uv run pytest`), `RTK_DISABLED=1`
  - no double-rewrite when re-processing an already-rewritten command; `uv runner`/`uvrun`/`uv runx`/bare `uv run` correctly decline
  - exit codes propagate exactly (0/1/3/42) on both paths
  - stderr preserved on success (incl. `logging.*`); uv chatter still dropped
  - truncation: tee tail hint offset verified to recover the omitted lines; multibyte/emoji safe
  - shell keywords decline even with a trusted unanchored custom TOML filter

Note: the `benchmark` check fails on `curl json` against an external endpoint — pre-existing and unrelated to this change.


src/cmds/python/uv_cmd.rs
```diff
@@ -72,8 +85,47 @@ fn display_command(prefix: &str, args_display: &str) -> String {
...
+    }
+
+    let summary = tail
+        .into_iter()
+        .rev()
+        .take(MAX_FALLBACK_TAIL_LINES)
+        .collect::<Vec<_>>()
+        .into_iter()
+        .rev()
+        .collect::<Vec<_>>();
```
KuSh:
```suggestion
    let skip = tail.len().saturating_sub(MAX_FALLBACK_TAIL_LINES);
    let summary = &tail[skip..];
```

src/cmds/python/uv_cmd.rs
```diff
@@ -72,8 +85,47 @@ fn display_command(prefix: &str, args_display: &str) -> String {
...
+    }
+
+    // On failure the streams are scanned merged: a Python traceback interleaves
+    // stdout and stderr, and splitting it would break frame ordering.
+    let extracted = extract_diagnostics(output);
+    if !extracted.is_empty() {
+        return extracted;
+    }
+
     let clean = strip_ansi(output);
```
KuSh: Nitpick: `strip_ansi` is called twice, once in `extract_diagnostics` and once here. Could we factor that out?

KuSh: it's also called two times through `program_output` but those seems harder to avoid

src/cmds/python/uv_cmd.rs
```diff
@@ -106,39 +158,61 @@ fn filter_uv_run_output(output: &str, exit_code: i32) -> String {
...
-        return format!("[FAIL] uv run failed (exit code: {exit_code})");
+    if capped.len() <= CAP_INVENTORY {
+        let out = capped.join("\n");
+        if !line_was_cut {
+            return out;
+        }
+        return match crate::core::tee::force_tee_hint(&clean, "uv-run") {
+            Some(hint) => format!("{out}\n{hint}"),
+            None => out,
+        };
```
KuSh: nitpick:

```suggestion
        if line_was_cut {
            if let Some(hint) = crate::core::tee::force_tee_hint(clean, "uv-run") {
              return format!("{out}\n{hint}");
            }
        }
        return out;
```

src/cmds/python/uv_cmd.rs
```diff
@@ -106,39 +158,61 @@ fn filter_uv_run_output(output: &str, exit_code: i32) -> String {
...
-        .rev()
-        .take(MAX_FALLBACK_TAIL_LINES)
-        .collect::<Vec<_>>()
-        .into_iter()
-        .rev()
-        .collect::<Vec<_>>();
+    // A program's result is usually its last line, so keep both ends.
+    let head = CAP_INVENTORY / 2;
+    let tail = CAP_INVENTORY - head;
+    let omitted = capped.len() - head - tail;
```
KuSh: nitpick, isn't that just 
```suggestion
    let omitted = capped.len() - CAP_INVENTORY;
```

src/discover/registry.rs
```diff
@@ -732,8 +732,19 @@ fn rewrite_line_range(cmd: &str) -> Option<String> {
...
+    "command",
+    "builtin",
+    "exec",
+    "nocorrect",
+];
+
+/// Prefixes that RULES can match as a whole string, so an unfiltered inner
+/// command may fall through instead of dropping the rewrite. Shell keywords must
+/// never appear here: they are not spawnable, so `rtk <keyword> <cmd>` is broken.
+const ROUTABLE_WRAPPER_PREFIXES: &[&str] = &["uv run"];
```
KuSh, line 825: suggestion, can't we use `constcat` crate or a `concat_slices` const fn or macro so that `BUILTIN_TRANSPARENT_PREFIXES` is `ROUTABLE_WRAPPER_PREFIXES` + a list of values and both can't be out of sync?

KuSh (approved):
Some minor nitpick and suggestions, otherwise LGTM!

src/cmds/python/uv_cmd.rs
```diff
@@ -106,39 +147,75 @@ fn filter_uv_run_output(output: &str, exit_code: i32) -> String {
...
+    };
+    let lines = &lines[..=last_content];
+    let capped: Vec<String> = lines
+        .iter()
+        .map(|line| truncate(line, MAX_PROGRAM_LINE_CHARS))
         .collect();
-
-    if tail.is_empty() {
-        return format!("[FAIL] uv run failed (exit code: {exit_code})");
+    let line_was_cut = capped.iter().zip(lines).any(|(cut, full)| cut != full);
```
KuSh: nitpick: 
```suggestion
    let line_was_cut = capped.iter().zip(lines).any(|(cut, full)| cut.len() != full.len());
```

src/cmds/python/uv_cmd.rs
```diff
@@ -106,39 +147,75 @@ fn filter_uv_run_output(output: &str, exit_code: i32) -> String {
...
+
+    let mut out = capped[..head].join("\n");
+    out.push_str(&format!("\n... ({omitted} lines omitted)\n"));
+    out.push_str(&capped[capped.len() - tail..].join("\n"));
+
+    // A cut in the head region sits before the tail offset, so `tail -n +N` skips it.
+    let head_line_was_cut = capped[..head]
+        .iter()
+        .zip(&lines[..head])
+        .any(|(cut, full)| cut != full);
```
KuSh, line 206: nitpick


```suggestion
        .any(|(cut, full)| cut.len() != full.len());
```

KuSh (approved):
LGTM!

pszymkowiak (approved):
Approved after deep review + real-binary testing on macOS with a live `uv`.

**Verified working end-to-end (rtk 0.42.4 build of this branch):**
- Program stdout/stderr preserved on success (no more `ok` data loss); traceback extraction and exit-code propagation (0/1/42) intact on failure.
- Truncation: 120-line output → 25 head + `… (70 lines omitted)` + 25 tail + tee hint; 50k-char single line capped to ~580 B. Tail offset (`tail -n +26`) resolves to the omitted lines when the path is quoted.
- Routing (registry.rs): delegation vs fall-through correct, word-boundary (`uvrun`/`uv runner`/bare `uv run` decline), idempotent re-rewrite, `sudo`/`exec`/`env` nesting, `RTK_DISABLED=1`, compound, and uv-flags-before-inner all behave. Shell keywords correctly never fall through.
- Pipes: only the final segment is rewritten; non-final `uv run …` stays raw — no data corruption in the pipe.
- `cargo test` 2555 passed, `clippy --all-targets` clean.

**Non-blocking follow-ups:**
1. PR description is stale: it claims uv resolution chatter (`Using CPython`/`Creating virtual environment`/`Resolved`) is stripped and cites "11 → 3 tokens" cold-run — but the final code (correctly, per the module header) passes it through. I confirmed this by execution. Please fix the summary so it matches the code.
2. The tee recovery hint (`[see remaining: tail -n +N ~/Library/Application Support/…]`) is not runnable as-is on macOS because the path with a space is unquoted — copy-pasting it yields `tail: …/Application: No such file`. This is pre-existing (display_path/force_tee_tail_hint in core/tee.rs, shared across filters) and already tracked by #1644 / PR #2325; not this PR's regression, but it does blunt the recovery goal here.
3. Minor: mypy's failure fallback returns unbounded raw output (pytest bounds it via passthrough_max_chars); success reorders stderr after stdout.

Solid, well-tested fix — none of the above block merge.

PR #3064, opened 2026-07-19
feat(git): filter large git show blob dumps with recovery and Latin-1 decoding

breisnerlopez:
## Summary

- **`feat(git)`**: `git show <rev>:<path>` blob dumps (previously passed through at 0% savings) are now filtered — large text blobs are capped to an 8 KiB preview with a `tail -n +N` recovery pointer (the full blob is tee'd); small blobs, tree listings, binary, and blobs too large to recover intact pass through unchanged.
- **Encoding fix**: blobs are captured as raw bytes and decoded encoding-aware — ISO-8859 / Latin-1 files (e.g. Oracle PL/SQL `.pck`) are transcoded losslessly instead of being corrupted into U+FFFD by lossy UTF-8, which also unblocks their compression. Ambiguous single-byte encodings (CP1252) and binary pass through raw. Measured on a real 77 KB Latin-1 blob: 0% → **89% savings** with the `É`→`�` corruption removed. (Same class of fix as #1087.)
- **`fix(tee)`**: recovery-file slugs (an embedded path duplicating the command) are hashed when long, saving ~12 tokens per truncation hint and fixing a latent collision where sibling paths in the same second overwrote each other's recovery file.
- **`refactor(output)`**: dropped RTK-invented advisory hints (`[hint] Run ruff/pip/black/dotnet/rubocop…`) and the go-build decorative separator — the counts they restate are already shown, so no signal is lost.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — clippy 0 warnings; **2449 tests pass**. (The one failing test, `git_checkout_dirty_tree_error_keeps_file_list`, is pre-existing and locale-dependent — it also fails on a clean `develop`, unrelated to this PR.)
- [x] Manual testing:
  - `rtk git show HEAD:src/main.rs` → 120 KB → 8 KB, `tail -n +N` recovery reconstructs byte-for-byte
  - `rtk git show HEAD:<latin1.pck>` → 89% savings, no `�`, recovery byte-identical to `iconv -f ISO-8859-1`
  - `rtk git show HEAD:<cp1252.pck>` → byte-exact passthrough (no wrong transcode)
  - `rtk git show HEAD:<dir>` (tree) and `HEAD:<binary>` → passthrough unchanged
  - 6 blobs from the same dir in the same second → 6 distinct recovery files (the old truncation collided them)

New unit + fixture tests cover decode (UTF-8 / Latin-1 / CP1252 / binary), byte-cap windowing, recovery-offset math, slug hashing/collision, and edge cases.

### DIFF
--- a/src/cmds/dotnet/dotnet_cmd.rs
+++ b/src/cmds/dotnet/dotnet_cmd.rs
@@ -417,7 +417,7 @@ fn format_dotnet_format_output(
     }
 
     output.push_str(&format!(
-        "\n\nok {} files already formatted\nRun `dotnet format` to apply fixes",
+        "\n\nok {} files already formatted",
         summary.files_unchanged
     ));
     output
@@ -2646,7 +2646,6 @@ mod tests {
         let output = format_dotnet_format_output(&summary, true);
         assert!(output.contains("Format: 2 files need formatting"));
         assert!(output.contains("src/Program.cs (line 42, col 17, WHITESPACE)"));
-        assert!(output.contains("Run `dotnet format` to apply fixes"));
     }
 
     #[test]
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -232,11 +232,12 @@ fn run_show(
         .iter()
         .any(|arg| arg.starts_with("--pretty") || arg.starts_with("--format"));
 
-    // `git show rev:path` prints a blob, not a commit diff. In this mode we should
-    // pass through directly to avoid duplicated output from compact-show steps.
-    let wants_blob_show = args.iter().any(|arg| is_blob_show_arg(arg));
+    // `git show rev:path` prints a blob (file content), not a commit diff. A pathspec
+    // (`--`) keeps it a commit-diff, so that case is excluded from blob handling.
+    let blob_args: Vec<&String> = args.iter().filter(|a| is_blob_show_arg(a)).collect();
+    let wants_blob_show = !blob_args.is_empty() && !args.iter().any(|a| a == "--");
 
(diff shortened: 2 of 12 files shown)


breisnerlopez:
Gentle nudge for maintainer review when someone has a moment — no rush.

Status: targets `develop`, CI green, CLA signed, `MERGEABLE`. `cargo clippy` clean, 2449 tests pass (the one red test, `git_checkout_dirty_tree_error_keeps_file_list`, is pre-existing and locale-dependent — also fails on a clean `develop`).

The load-bearing change is the encoding fix: `git show <rev>:<path>` on an ISO-8859-1 blob (e.g. an Oracle PL/SQL `.pck`) was decoded lossily and corrupted `É` into U+FFFD, which also blocked compression. It now transcodes losslessly and the recovery file is byte-identical to `iconv -f ISO-8859-1`. Measured 0% → 89% on a real 77 KB Latin-1 blob.

On the single-focus rule in CONTRIBUTING: the PR also carries a `fix(tee)` recovery-slug hash and a `refactor(output)` that drops the invented advisory hints. If you'd rather review the `feat(git)` blob filter on its own, I'm happy to split those two out into their own PRs — just say the word.

breisnerlopez:
Closing in favor of smaller, focused PRs (per the single-focus PR guideline in CONTRIBUTING) rebased onto current `develop`:

- **#3265** — `feat(git)`: the `git show` large-blob filter with recovery + encoding-aware (Latin-1) capture — the core of this PR.
- **#3266** — `fix(tee)`: the independent recovery-slug collision fix.

The subjective advisory-hints refactor that was bundled here has been dropped. Splitting this way should make review much easier. Thanks!

PR #3066, opened 2026-07-19
fix(hooks): treat jq parse errors as passthrough, not empty CMD

### LINKED ISSUE #3065: Hook silently drops rewrite when command JSON contains \| (jq parse error)
RTK version: 0.35.0

OS: Windows 11, bash via Git Bash

Hook version: 3 (rtk-hook-version: 3)

What happened:

Grep commands using \| for alternation (e.g. grep -r "FOO\|BAR" path/) run unfiltered without RTK wrapping. The hook silently passes them through.

Root cause:

In rtk-rewrite.sh, this line fails when the command contains \|:

jq treats \| as an invalid JSON escape sequence and exits with a parse error. Because the script does not check $? after the jq call, CMD is empty, and the guard if [ -z "$CMD" ]; then exit 0; fi silently passes the command through unchanged.

Reproduction:

printf '{"tool_name":"Bash","tool_input":{"command":"grep -r \"GOOGLE_CLIENT_ID\\|SECRET\" /tmp"}}' \
  | bash ~/.claude/hooks/rtk-rewrite.sh
# Output: nothing — grep runs raw
Suggested fix:

Check jq's exit code before testing for empty CMD:

JQ_EXIT=$?
if [ $JQ_EXIT -ne 0 ] || [ -z "$CMD" ]; then
  exit 0
fi
Alternatively, use --raw-input + --null-input to make jq more resilient to non-standard escapes.

Solaris-star:
## Summary
- Check `jq`'s exit code (and silence stderr) when reading `.tool_input.command` in the Claude, Cursor, and `.claude` rewrite hooks.
- Invalid JSON escapes in PreToolUse payloads (for example raw `\|` inside a command string) previously made `jq` fail with an empty `CMD`, which the hooks treated as "no command" and silently passed through.
- Parse failures now explicitly passthrough; valid commands that contain a properly JSON-escaped `\|` still rewrite as before.

## Test plan
- [x] Invalid payload with raw `\|` → hook exits 0, no jq stderr noise
- [x] Valid JSON payload with `\\|` in the command → rewrite proceeds
- [x] Simple command (`git status`) still rewrites under a stub `rtk rewrite`

Fixes #3065

### DIFF
--- a/.claude/hooks/rtk-rewrite.sh
+++ b/.claude/hooks/rtk-rewrite.sh
@@ -32,8 +32,16 @@ fi
 set -euo pipefail
 
 INPUT=$(cat)
-CMD=$(echo "$INPUT" | jq -r '.tool_input.command // empty')
-
+# Suppress stderr: invalid JSON escapes (e.g. raw \| in command strings) make jq fail.
+# Check exit code so a parse error does not look like an empty command and silently
+# pass the original tool call through without a rewrite attempt.
+CMD=""
+JQ_EXIT=0
+CMD=$(echo "$INPUT" | jq -r '.tool_input.command // empty' 2>/dev/null) || JQ_EXIT=$?
+if [ "$JQ_EXIT" -ne 0 ]; then
+  _rtk_audit_log "skip:jq_error" "-"
+  exit 0
+fi
 if [ -z "$CMD" ]; then
   _rtk_audit_log "skip:empty" "-"
   exit 0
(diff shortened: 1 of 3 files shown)


Solaris-star:
Closing after rechecking the hook contract. Standards-compliant JSON encodes the command's backslash, so valid payloads already parse and rewrite normally. For malformed JSON, this patch still exits successfully without rewriting; it only makes that passthrough more explicit. That does not resolve the reported behavior, and I do not want to present a no-op as a fix.

PR #3067, opened 2026-07-19
fix(js): preserve failed command output in parser fallbacks

jbellis:
## Summary

* Preserve combined stdout and stderr when JavaScript command parsers fall back to passthrough, so failed Vitest, Playwright, and pnpm commands retain stderr-only errors.
* Suppress parser-failure warnings for nonzero child exits while still warning on successful but unparseable output.
* Support pnpm v10 `outdated --format json` output when `current` is absent, including deprecated-package reports.

## Test plan

* [x] `cargo fmt --check`
* [x] `cargo test -q`
* [x] Manual test of Vitest, Playwright, and pnpm passthrough and happy-path behavior
* [x] Manual test of `pnpm outdated --format json` exit behavior with pnpm 10.33.2

### DIFF
--- a/src/cmds/js/playwright_cmd.rs
+++ b/src/cmds/js/playwright_cmd.rs
@@ -8,8 +8,9 @@ use regex::Regex;
 use serde::Deserialize;
 
 use crate::parser::{
-    emit_degradation_warning, emit_passthrough_warning, truncate_passthrough, FormatMode,
-    OutputParser, ParseResult, TestFailure, TestResult, TokenFormatter,
+    emit_degradation_warning, emit_passthrough_warning, passthrough_warning_reason,
+    truncate_passthrough, FormatMode, OutputParser, ParseResult, TestFailure, TestResult,
+    TokenFormatter,
 };
 
 /// Matches real Playwright JSON reporter output (suites → specs → tests → results)
@@ -289,30 +290,14 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let result = exec_capture(&mut cmd)
         .context("Failed to run playwright (try: npm install -g playwright)")?;
 
-    let raw = format!("{}\n{}", result.stdout, result.stderr);
+    let raw = result.combined();
 
-    // Parse output using PlaywrightParser
-    let parse_result = PlaywrightParser::parse(&result.stdout);
-    let mode = FormatMode::from_verbosity(verbose);
-
-    let filtered = match parse_result {
-        ParseResult::Full(data) => {
-            if verbose > 0 {
-                eprintln!("playwright test (Tier 1: Full JSON parse)");
-            }
-            data.format(mode)
-        }
-        ParseResult::Degraded(data, warnings) => {
-            if verbose > 0 {
-                emit_degradation_warning("playwright", &warnings.join(", "));
-            }
-            data.format(mode)
-        }
-        ParseResult::Passthrough(raw) => {
-            emit_passthrough_warning("playwright", "All parsing tiers failed");
-            raw
-        }
-    };
+    let filtered = format_playwright_output(
+        &result.stdout,
+        &raw,
+        result.exit_code,
+        verbose,
+    );
 
     let hint = crate::core::tee::tee_and_hint(&raw, "playwright", result.exit_code);
     let shown = crate::core::runner::emit_guarded(&filtered, hint.as_deref(), &raw);
@@ -330,6 +315,32 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     }
 
     Ok(0)
+}
+
+fn format_playwright_output(stdout: &str, combined: &str, exit_code: i32, verbose: u8) -> String {
+    let parse_result = PlaywrightParser::parse(stdout);
+    let mode = FormatMode::from_verbosity(verbose);
+
+    match parse_result {
+        ParseResult::Full(data) => {
+            if verbose > 0 {
+                eprintln!("playwright test (Tier 1: Full JSON parse)");
+            }
+            data.format(mode)
+        }
+        ParseResult::Degraded(data, warnings) => {
+            if verbose > 0 {
+                emit_degradation_warning("playwright", &warnings.join(", "));
+            }
+            data.format(mode)
+        }
+        ParseResult::Passthrough(_) => {
+            if let Some(reason) = passthrough_warning_reason(exit_code) {
+                emit_passthrough_warning("playwright", reason);
+            }
+            truncate_passthrough(combined)
+        }
+    }
 }
 
 #[cfg(test)]
@@ -476,5 +487,33 @@ mod tests {
         let result = PlaywrightParser::parse(invalid);
         assert_eq!(result.tier(), 3); // Passthrough
         assert!(!result.is_ok());
+    }
+
+    #[test]
+    fn test_playwright_successful_json_ignores_combined_stderr() {
+        let stdout = r#"{
+            "stats": {
+                "expected": 2,
+                "unexpected": 0,
+                "skipped": 0,
+                "duration": 100.0
+            },
+            "suites": []
+        }"#;
+        let combined = format!("{stdout}\nWARN noisy stderr line\n");
+
+        let filtered = format_playwright_output(stdout, &combined, 0, 0);
+
+        assert!(filtered.contains("PASS (2) FAIL (0)"));
+        assert!(!filtered.contains("noisy stderr"));
+    }
+
+    #[test]
+    fn test_playwright_failed_command_passthrough_includes_stderr() {
+        let combined = "Error: command not found: playwright\n";
+
+        let filtered = format_playwright_output("", combined, 1, 0);
+
+        assert_eq!(filtered, combined);
     }
 }
--- a/src/cmds/js/pnpm_cmd.rs
+++ b/src/cmds/js/pnpm_cmd.rs
@@ -11,8 +11,9 @@ use std::collections::HashMap;
 use std::ffi::OsString;
 
 use crate::parser::{
-    emit_degradation_warning, emit_passthrough_warning, truncate_passthrough, Dependency,
-    DependencyState, FormatMode, OutputParser, ParseResult, TokenFormatter,
+    emit_degradation_warning, emit_passthrough_warning, passthrough_warning_reason,
+    truncate_passthrough, Dependency, DependencyState, FormatMode, OutputParser, ParseResult,
+    TokenFormatter,
 };
 
 const MAX_LISTING: usize = CAP_LIST;
@@ -43,9 +44,11 @@ struct PnpmOutdatedOutput {
 
 #[derive(Debug, Deserialize)]
 struct PnpmOutdatedPackage {
-    current: String,
+    current: Option<String>,
     latest: String,
     wanted: Option<String>,
+    #[serde(rename = "isDeprecated", default)]
+    is_deprecated: bool,
     #[serde(rename = "dependencyType", default)]
     dependency_type: String,
 }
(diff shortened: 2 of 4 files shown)


src/cmds/js/pnpm_cmd.rs
```diff
@@ -434,40 +448,66 @@ fn run_outdated(args: &[String], verbose: u8) -> Result<i32> {
...
+            (data.format(mode), false)
         }
-        ParseResult::Passthrough(raw) => {
-            emit_passthrough_warning("pnpm outdated", "All parsing tiers failed");
-            raw
+        ParseResult::Passthrough(_) => {
+            if let Some(reason) = passthrough_warning_reason(exit_code) {
+                emit_passthrough_warning("pnpm outdated", reason);
+            }
+            (truncate_passthrough(combined), true)
```
KuSh, line 506: **medium — head-truncation discards the stderr this fix is meant to preserve, and there's no recovery path.**

`combined()` is `stdout + stderr`, but `truncate_passthrough` → `truncate_output` keeps the **first** `passthrough_max_chars` (default 2000, `src/core/config.rs:143`). So any failing `pnpm outdated` that wrote more than 2000 chars to stdout — e.g. a JSON blob preceded by a pnpm banner, so serde fails and tier 2 finds nothing — gets truncated to 2000 chars of *stdout*, and the entire stderr (the actual error) is dropped. That's the exact case the PR is trying to fix.

Unlike vitest/playwright, `run_outdated` never calls `tee_and_hint`/`force_tee_hint`, so there isn't even a `[see full output: …]` pointer to recover from.

Suggestions, either works:
- put stderr first in the passthrough body (`format!("{}{}", result.stderr, result.stdout)`), since the error is what matters on failure; or
- tail-truncate here instead of head-truncating.

Adding a `tee_and_hint` call on the `run_outdated` failure path would also bring it in line with the other two commands.

src/cmds/js/pnpm_cmd.rs
```diff
@@ -434,40 +448,66 @@ fn run_outdated(args: &[String], verbose: u8) -> Result<i32> {
...
+        formatted.text.clone()
+    };
+    let shown = never_worse(&combined, &display);
+    println!("{}", shown);
+
+    timer.track("pnpm outdated", "rtk pnpm outdated", &combined, shown);
+
+    if formatted.passthrough && !result.success() {
+        return Ok(result.exit_code);
+    }
```
KuSh, line 470: **medium — the exit status of `rtk pnpm outdated` now depends on whether the parser happened to succeed.**

`pnpm outdated` exits **1 whenever outdated packages are found**. That is its documented, normal outcome — not a failure. With this condition, the same healthy repo returns:

- `0` when the JSON parses (tier 1/2), and
- `1` when it degrades to passthrough — unknown pnpm JSON shape, banner-prefixed stdout, or a user-supplied `--format table` appended *after* RTK's own `--format json` at lines 441-442.

So the exit code becomes a function of parser luck rather than of repo state. Callers and hooks that treat nonzero as failure will start reporting `rtk pnpm outdated` as broken; before this PR the function was unconditionally `Ok(0)`.

If exit-code propagation is wanted here, it needs to distinguish pnpm's "outdated packages found" (1) from a genuine invocation failure — or just keep `Ok(0)` for `outdated` and apply the propagation only to commands where nonzero really means failure.

src/cmds/js/pnpm_cmd.rs
```diff
@@ -434,40 +448,66 @@ fn run_outdated(args: &[String], verbose: u8) -> Result<i32> {
...
-            data.format(mode)
+            (data.format(mode), false)
         }
-        ParseResult::Passthrough(raw) => {
-            emit_passthrough_warning("pnpm outdated", "All parsing tiers failed");
-            raw
+        ParseResult::Passthrough(_) => {
+            if let Some(reason) = passthrough_warning_reason(exit_code) {
+                emit_passthrough_warning("pnpm outdated", reason);
+            }
```
KuSh, line 505: **medium — the `passthrough_warning_reason` premise doesn't hold for `pnpm outdated`.**

The new helper (`src/parser/mod.rs:136`) suppresses the warning whenever `exit_code != 0`, on the rationale that *"a non-zero command exit explains unparseable output without implying a parser defect."*

For `pnpm outdated`, exit 1 **is** the healthy path (outdated packages found). So a genuine parser gap on the single most common invocation of this command will now never emit `[RTK:PASSTHROUGH]` — losing exactly the regression signal this repo relies on to notice parser rot.

The same reasoning is weaker but still applies to vitest/jest/playwright: a nonzero exit there means failing tests, which is precisely when the parser matters most and when a silent passthrough is most costly.

Consider gating on something that actually indicates "the command didn't run", e.g. `exit_code == 127` / empty stdout, rather than on any nonzero exit.

src/cmds/js/pnpm_cmd.rs
```diff
@@ -199,14 +202,25 @@ impl OutputParser for PnpmOutdatedParser {
                 let mut outdated_count = 0;
 
                 for (name, pkg) in &json.packages {
-                    if pkg.current != pkg.latest {
+                    let current = pkg
+                        .current
+                        .as_deref()
+                        .or(pkg.wanted.as_deref())
+                        .unwrap_or(&pkg.latest);
```
KuSh, line 209: **low — this fallback fabricates an installed version, and can make a missing dependency disappear from the report.**

pnpm omits `current` for a dependency that is *declared but not installed* — its table output prints `missing` for that column. With `.or(pkg.wanted)`, RTK reports the wanted version as though it were installed, which is factually wrong.

Worse, when `wanted == latest` the fallback makes `current == latest`, so the `if current != latest` check on line 216 is false, `outdated_count` isn't incremented, and the formatter never lists the package. A missing dependency vanishes from the report entirely — arguably the one thing the user most needs to see.

Prefer surfacing the absent state explicitly (e.g. `"missing"` as the current version, and counting it) over substituting `wanted`.

src/cmds/js/pnpm_cmd.rs
```diff
@@ -199,14 +202,25 @@ impl OutputParser for PnpmOutdatedParser {
...
+                    let latest = if pkg.is_deprecated && current == pkg.latest {
+                        format!("{} (deprecated)", pkg.latest)
+                    } else {
+                        pkg.latest.clone()
+                    };
```
KuSh, line 214: **low — encoding the deprecation marker into the version string is lossy and produces spurious verbose output.**

Two problems with `is_deprecated && current == pkg.latest`:

1. **Lossy.** A package that is *both* deprecated and outdated (`current != latest`) silently loses its `(deprecated)` marker — deprecation is only reported when you happen to be on the latest version.

2. **Breaks `-v`.** Because the marker is appended to `latest_version`, `DependencyState::format_verbose` (`src/parser/formatter.rs:190`) compares `wanted != latest` → `"1.3.0" != "1.3.0 (deprecated)"` → true, and emits a nonsense trailer:

```
  left-pad: 1.3.0 → 1.3.0 (deprecated)
    (wanted: 1.3.0)
```

A `deprecated: bool` field on `Dependency`, rendered by the formatter, would keep the version string a version string and fix both cases.

src/cmds/js/pnpm_cmd.rs
```diff
@@ -434,40 +448,66 @@ fn run_outdated(args: &[String], verbose: u8) -> Result<i32> {
...
+        &combined,
+        result.exit_code,
+        verbose,
+    );
+
+    let display = if formatted.text.trim().is_empty() {
+        "All packages up-to-date".to_string()
+    } else {
+        formatted.text.clone()
+    };
```
KuSh, line 462: **low — "All packages up-to-date" can be printed for a command that failed.**

If pnpm fails while producing no output at all (killed, or an error swallowed by a wrapper), `formatted.text` is empty → `display` becomes `"All packages up-to-date"` → and then line 468 returns a **nonzero** exit code. The user (and the LLM reading the output) sees a success message paired with a failure status.

The empty-text special case should only apply when the parse actually succeeded:

```rust
let display = if !formatted.passthrough && formatted.text.trim().is_empty() {
    "All packages up-to-date".to_string()
} else {
    formatted.text.clone()
};
```

src/cmds/js/playwright_cmd.rs
```diff
@@ -289,30 +290,14 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let result = exec_capture(&mut cmd)
         .context("Failed to run playwright (try: npm install -g playwright)")?;
 
-    let raw = format!("{}\n{}", result.stdout, result.stderr);
+    let raw = result.combined();
```
KuSh, line 293: **low — this exposes a latent bug in `CaptureResult::combined()` itself.**

Correcting my framing here: this isn't a playwright regression, it's the shared helper. `combined()` is a bare concatenation with no separator (`src/core/stream.rs:558`):

```rust
pub fn combined(&self) -> String {
    format!("{}{}", self.stdout, self.stderr)
}
```

When stdout doesn't end in a newline, its last line is glued to the first stderr line. That affects all 12 call sites — `git fetch` (`git.rs:1844`), `git stash` (`:1969`), `git worktree` (`:2154`, which prints `&combined` straight to the user on failure), `git checkout` (`:1673`, `:1703`), both pnpm paths, and vitest. Playwright was the *only* caller that joined explicitly, so this line moves the last correct site onto the buggy helper rather than creating a new problem.

The fix belongs in `combined()`, not here — but not as a straight `format!("{}\n{}", ...)` either, since that adds a leading blank line when stdout is empty (which is the common case on a hard failure, and is what the old playwright code did). Conditional join:

```rust
pub fn combined(&self) -> String {
    if self.stdout.is_empty() || self.stdout.ends_with('\n') {
        format!("{}{}", self.stdout, self.stderr)
    } else {
        format!("{}\n{}", self.stdout, self.stderr)
    }
}
```

Both existing tests (`test_exec_capture_combined`, `test_capture_result_combined_empty`) are `contains`/empty-based and still pass; worth adding one that pins the no-trailing-newline case.

Happy for this to land as a follow-up rather than blocking here — the line itself is fine once `combined()` is.

KuSh (changes requested):
## Request changes

Thanks for this — preserving stderr in the fallback paths is the right fix, and the vitest/playwright halves are clean. The build is green on the branch (`cargo fmt --check`, `cargo clippy --all-targets`, `cargo test --bin rtk` / 2446 tests; the one `cargo test --all` failure, `git_checkout_dirty_tree_error_keeps_file_list`, is a pre-existing locale issue unrelated to this PR).

The blocking issues all cluster on one theme: **`pnpm outdated` exits 1 whenever outdated packages are found** — that is its normal success signal, not an error. Two assumptions this PR introduces (`passthrough_warning_reason(exit_code)` and `!result.success()` → propagate exit code) both break against that convention. Details inline.

Summary of inline findings:

| # | Location | Severity | Issue |
|---|---|---|---|
| 1 | `pnpm_cmd.rs:506` | medium | head-truncation drops the stderr the fix exists to preserve, with no tee hint |
| 2 | `pnpm_cmd.rs:468` | medium | exit status now depends on whether the parser happened to succeed |
| 3 | `pnpm_cmd.rs:503` | medium | `passthrough_warning_reason` premise doesn't hold for `pnpm outdated` |
| 4 | `pnpm_cmd.rs:205` | low | `current` fallback fabricates an installed version, can hide a missing dep |
| 5 | `pnpm_cmd.rs:210` | low | deprecation marker in the version string is lossy + breaks `-v` output |
| 6 | `pnpm_cmd.rs:458` | low | "All packages up-to-date" printed for a failed command |
| 7 | `playwright_cmd.rs:293` | low, non-blocking | latent bug in `CaptureResult::combined()` (all 12 call sites), not this line — follow-up is fine |

Only 1-6 need changes before merge; 7 is context on a pre-existing defect this line happens to surface.

PR #3073, opened 2026-07-19
feat(js): add native Bun command support

### LINKED ISSUE #1316: Unhandled commands: bun test, bunx tsc, bun, bun run, bun drizzle-kit
## Unhandled Commands

Discovered via `rtk discover` in a TypeScript/Bun backend project.

| Command | Count | Example |
|---------|-------|---------|
| `bun test` | 34 | `SMS_PROVIDER=console bun test src/modu..` |
| `bunx tsc` | 12 | `bunx tsc --noEmit 2>&1` |
| `bun` | 6 | `bun --env-file=.env.test test tests/st..` |
| `bun run` | 3 | `bun run tsc --noEmit 2>&1` |
| `bun drizzle-kit` | 1 | `bun drizzle-kit generate 2>&1` |

## Expected behavior

RTK should intercept and filter output for these Bun-specific commands the same way it handles `npm test`, `tsc`, etc.

## Environment

- Runtime: Bun
- Project type: TypeScript backend (Elysia + Drizzle ORM)

lntutor:
## Summary

- add native rtk bun and rtk bunx commands that preserve the underlying Bun executable
- rewrite bun test, option-prefixed bun commands, bun run, Bun script shortcuts, and bunx invocations
- remove Bun run/test banners and only the immediate echoed script command while preserving program output
- classify Bun usage under JavaScript telemetry and document the new commands

Fixes #1316

## Verification

- cargo fmt --all -- --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo test --all-features (2,505 passed, 8 ignored)
- cargo build --release --all-features
- live execution: rtk bun -e console.log command completed successfully
- exact rewrite verified for an environment-prefixed bun test command


KuSh:
Thanks for this, and sorry it sat — closing as superseded by #988, which covers `bun` plus `deno` and has been through five rounds of review with fixtures captured from real `bun 1.2.20` / `deno 2.9.6`.

Being concrete about what your PR does that #988 does not, because it is a real gap and I would like you to pick it up:

**#988 only rewrites an allowlist of subcommands** — `^bun\s+(install|add|remove|test|build|run|pm ls|pm|x)\b`. Yours matches `^bun(?:\s|$)`, so these route through rtk for you and run raw under #988:

```
bun --env-file=.env.test test tests/unit    # option-prefixed
bun dev                                     # script shortcut
bun drizzle-kit generate                    # package binary
bun -e 'console.log(1)'
```

That is a coverage gap, not a correctness one — those commands still work under #988, they just get no filtering.

**Why #988 can't simply take the wider pattern as-is.** `run_filtered` buffers the whole stream until the child exits (`FilterMode::CaptureOnly`). `^bun(?:\s|$)` routes `bun dev` — a dev server — into it, so it prints nothing at all until Ctrl-C and then loses the buffer. #988 spent a full review round on exactly this class and now sends watch-mode runs through unfiltered on every entry point. Widening the pattern needs the same guard for long-running scripts, which is the harder half of the problem.

Two smaller things worth carrying over when you do: `filter_bun_output` drops every blank line from arbitrary program output, which reflows anything a script prints deliberately; and the banner check `line.contains(|c: char| c.is_ascii_digit())` will match a script whose own output starts with `bun run v` — worth anchoring on the version shape.

**The follow-up I'd welcome after #988 lands:** widen the rewrite to option-prefixed and script-shortcut forms, reusing `is_watch_mode` (and something equivalent for dev servers) so nothing long-running ends up captured. That is genuinely missing today and it is the valuable part of this PR.

Also relevant: #1316, which you referenced, stays open until that follow-up rather than being closed by #988.

PR #3082, opened 2026-07-20
fix(curl): preserve signal exit codes

### LINKED ISSUE #2656: fix(curl): use signal-aware exit code instead of unwrap_or(1)

lntutor:
## Summary
- convert curl's captured `ExitStatus` with the shared signal-aware helper
- return the standard `128 + signal` code for killed curl processes instead of collapsing them to `1`
- keep normal success and explicit non-zero exit codes unchanged

## Testing
- `cargo test --all-features curl_cmd`
- `cargo test --all-features test_exit_code_signal_kill`

Closes #2656

### DIFF
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -13,6 +13,7 @@
 
 use crate::core::tee::force_tee_hint;
 use crate::core::tracking;
+use crate::core::stream::status_to_exit_code;
 use crate::core::utils::resolved_command;
 use anyhow::{Context, Result};
 use std::borrow::Cow;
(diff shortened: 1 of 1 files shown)


KuSh:
Closing in favour of #2657, which makes the same one line change to curl_cmd.rs and keeps the import in the order rustfmt expects. Please follow the fix on that PR, or reply here if you see something it misses.

PR #3083, opened 2026-07-20
fix(ruff): preserve non-check subcommands

### LINKED ISSUE #2669: ruff subcommand misrouting: rule, config, linter, etc. get check --output-format=json injected

lntutor:
## Summary
- replace Ruff's `format`/`version` blocklist with an explicit official-subcommand list
- prevent `rule`, `config`, `linter`, `clean`, `server`, `analyze`, and completion commands from receiving `check --output-format=json`
- preserve implicit check routing for paths and explicit `ruff check` invocations
- add routing coverage for all recognized subcommands, paths, and top-level flags

## Testing
- `cargo test --all-features` (2,498 passed, 8 ignored)

Closes #2669

### DIFF
--- a/src/cmds/python/ruff_cmd.rs
+++ b/src/cmds/python/ruff_cmd.rs
@@ -8,6 +8,20 @@ use anyhow::Result;
 use serde::Deserialize;
 use std::collections::HashMap;
 
+const RUFF_SUBCOMMANDS: &[&str] = &[
+    "analyze",
+    "check",
+    "clean",
+    "config",
+    "format",
+    "generate-shell-completion",
+    "help",
+    "linter",
+    "rule",
+    "server",
+    "version",
+];
+
 #[derive(Debug, Deserialize)]
 struct RuffLocation {
     row: usize,
@@ -32,9 +46,7 @@ struct RuffDiagnostic {
 }
 
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
-    let is_check = args.is_empty()
-        || args[0] == "check"
-        || (!args[0].starts_with('-') && args[0] != "format" && args[0] != "version");
+    let is_check = is_check_invocation(args);
 
     let is_format = args.iter().any(|a| a == "format");
 
@@ -88,6 +100,12 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         },
         runner::RunOptions::stdout_only(),
     )
+}
+
+fn is_check_invocation(args: &[String]) -> bool {
+    args.first().is_none_or(|arg| {
+        arg == "check" || (!arg.starts_with('-') && !RUFF_SUBCOMMANDS.contains(&arg.as_str()))
+    })
 }
 
 /// Filter ruff check JSON output - group by rule and file
@@ -338,6 +356,32 @@ fn compact_path(path: &str) -> String {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn known_ruff_subcommands_do_not_route_through_check() {
+        for subcommand in RUFF_SUBCOMMANDS {
+            let args = vec![subcommand.to_string()];
+            assert_eq!(
+                is_check_invocation(&args),
+                *subcommand == "check",
+                "unexpected routing for ruff {subcommand}"
+            );
+        }
+    }
+
+    #[test]
+    fn paths_and_explicit_check_route_through_check() {
+        assert!(is_check_invocation(&[]));
+        assert!(is_check_invocation(&["check".to_string(), ".".to_string()]));
+        assert!(is_check_invocation(&["src".to_string()]));
+        assert!(is_check_invocation(&["pyproject.toml".to_string()]));
+    }
+
+    #[test]
+    fn top_level_flags_are_not_misclassified_as_paths() {
+        assert!(!is_check_invocation(&["--version".to_string()]));
+        assert!(!is_check_invocation(&["--help".to_string()]));
+    }
 
     #[test]
     fn test_filter_ruff_check_no_issues() {
--- a/src/cmds/python/ruff_cmd.rs
+++ b/src/cmds/python/ruff_cmd.rs
@@ -369,6 +369,28 @@ mod tests {
         }
     }
 
+    /// `known_ruff_subcommands_do_not_route_through_check` iterates the constant, so it
+    /// cannot catch an entry going missing. This pins the list against `ruff --help`.
+    #[test]
+    fn ruff_subcommands_cover_the_ruff_cli() {
+        assert_eq!(
+            RUFF_SUBCOMMANDS,
+            [
+                "analyze",
+                "check",
+                "clean",
+                "config",
+                "format",
+                "generate-shell-completion",
+                "help",
+                "linter",
+                "rule",
+                "server",
+                "version",
+            ]
+        );
+    }
+
     #[test]
     fn paths_and_explicit_check_route_through_check() {
         assert!(is_check_invocation(&[]));


KuSh (approved):
## Round 1 of 3 — APPROVED, follow-ups filed

**Claim**: ruff's non-check subcommands (`rule`, `config`, `linter`, `clean`, `server`, `analyze`, `help`, `generate-shell-completion`) must stop having `check --output-format=json` injected into them.

**Scope**: accept. (round 1, frozen)

**Ran**: real ruff 0.16.6, `LC_ALL=C`, ~35 invocations × {`develop` 665b2720, this PR merged into it, native ruff}: all 11 subcommands, their error paths, implicit-path check, explicit `check`, `format --check/--diff`, top-level flags, and `server` under both a closed and an open stdin. Merge into current `develop` is clean (`ort`, no conflict); `cargo fmt --all --check`, `cargo clippy --all-targets` and `cargo test --all` (3260 passed, 0 failed) are all green **on the merge result**, which matters here because the branch predates develop's `44982e07`/`fa82e18d` ruff changes. Savings on the check path unchanged at 99.5%, well above the 20% floor (CONTRIBUTING.md).

### Blocks merge (0)

None. The allowlist is exactly `ruff --help` for 0.16.6 — check, rule, config, linter, clean, format, server, analyze, version, help — plus the hidden `generate-shell-completion`, which I confirmed still exists (`ruff generate-shell-completion bash` → exit 0). ruff does not infer abbreviated subcommands (`ruff ver` → `unrecognized subcommand`), so there is no partial-match gap.

The before/after is as bad as #2669 said, and worth recording since the hook rewrites every `ruff …` to `rtk ruff …`:

```
$ rtk ruff rule E501          # develop
Ruff: 2 issues in 2 files
Top rules:
  E902 (2x)
Violations:
  E501:1:1 E902 No such file or directory (os error 2)     [exit 1]

$ rtk ruff rule E501          # this PR
# line-too-long (E501)
Derived from the **pycodestyle** linter.
...                                                        [exit 0]
```

That is a fabricated lint report, not an error — the worst failure shape rtk has. Same on `config`, `linter`, `clean`, `help`, `analyze graph`, `generate-shell-completion`.

Exit-code fidelity is right too, which is the part I most expected to break. On every error path the PR now returns ruff's own code where develop returned a confident `1`:

| | develop | this PR | native ruff |
|---|---|---|---|
| `ruff analyze` | 1 + fake E902 | 2 | 2 |
| `ruff rule` | 1 + fake E902 | 2 | 2 |
| `ruff config nonexistent.option` | 1 + fake E902 | 2 | 2 |
| `ruff generate-shell-completion` | 1 + fake E902 | 2 | 2 |

And the paths the PR must not disturb are undisturbed: `rtk ruff check src`, `rtk ruff .`, `rtk ruff src`, `rtk ruff check --statistics src`, `rtk ruff format --check/--diff src` are byte-identical between develop and this PR (modulo the pre-existing `HashMap` ordering in `filter_ruff_check_json`'s "Top files" block, which varies run-to-run on the *same* binary — sorted output is identical).

`ruff server` does not hang, which was my main worry given the hook rewrite: `stdout_only()` gives the child a null stdin, so it exits 2 immediately even when the parent's stdin stays open. Verified both ways.

- `known_ruff_subcommands_do_not_route_through_check` iterates `RUFF_SUBCOMMANDS`, so it cannot catch an entry going missing from the constant. Mutation-verified: deleting `"rule"` and `"linter"` leaves all 16 ruff tests green. It does bite on the routing logic — replacing the body with the old blocklist fails it. This mirrors the existing `NPM_SUBCOMMANDS` test, so it is precedent-consistent, not a defect; I have covered it with the fixup below rather than sending it back.
- `rtk ruff linter` loses the leading space on its first line (`stdout.trim()`), breaking the right-aligned prefix column. Cosmetic, and the `else` branch has always done this.

### Filed as follow-ups

All five are on lines this PR does not touch, and all five reproduce identically on `develop` — none is a regression here.

- **#3927** — `rtk ruff --select F401 src` exits 2 with no output: `is_check_invocation` bails on any leading `-`, so a flag-first implicit check is never routed to `check`. Pre-existing.
- **#3928** — `is_format` uses `any(|a| a == "format")` while `is_check` uses `first()`, so `rtk ruff help format` comes back uncapped at 4889 bytes while `rtk ruff help check` is capped at 2001. Correct output, accidental routing. Pre-existing; became observable here.
- **#3929** — `rtk ruff analyze graph` is truncated at `passthrough_max_chars` and emits unparseable JSON (12408 → 2001 bytes, cut mid-string). The cap is develop's (`fa82e18d`); this PR is what makes the path reachable, and it is still a large improvement on the fabricated E902 report it replaces.
- **#3930** — `filter_ruff_format` still matches pre-0.12 wording (`Would reformat:` / `left unchanged`). ruff 0.16.6 says `unformatted:` / `already formatted`, so `rtk ruff format --check .` compresses 886 → 886 bytes: **0% savings** against the 20% floor. Its two unit-test fixtures use the old strings, which is why the drift went unnoticed. Pre-existing.
- **#3931** — `ruff server` should not be proxied at all (null stdin + buffered stdout can never run an LSP). Not a regression — develop proxied it too, just more destructively.

Separately: `rtk ruff rule NOSUCHRULE` shows nothing at all because `RunOptions::stdout_only()` drops stderr. That is module-wide and pre-existing (identical on develop for `rtk ruff format --nonexistent-flag`), and #3772 already fixes it. Nothing owed here.

- Allowlist completeness against ruff 0.16.6, including the hidden `generate-shell-completion`; no subcommand inference to worry about.
- `arg == "check"` short-circuit ahead of the `contains` check — needed, since `"check"` is in the list, and correct.
- `filter_ruff_format` is a safe identity on non-format text: `rtk ruff help format` and `rtk ruff config format` are byte-identical to native ruff despite `is_format` being true for both.
- Exit-code propagation on all four subcommand error paths (table above).
- No hang on `ruff server`, under closed and open stdin.
- Merge with current `develop` — clean, and semantically right: `is_check_invocation` composes with develop's newer `user_set_output_format` / `use_json_filter` logic without either clobbering the other.
- `git_checkout_dirty_tree_error_keeps_file_list` fails on the PR head in isolation. It is stale-branch fallout, not yours — it passes on the merge with develop.
- Overlap with #3912, #3782 and #3577, which all touch `ruff_cmd.rs`: they hit the `run_filtered` call and the filter-fn names, never the routing line. Any merge order works.

- *`ruff server` becomes a hang under the hook rewrite.* Dropped: null stdin makes it exit 2 immediately, matching native ruff under a closed stdin, with the parent's stdin open or closed.
- *The PR newly loses stderr diagnostics on the subcommands it reroutes.* Dropped: `stdout_only()` drops stderr on the check branch too, so develop loses it identically. Filed under #3772, not here.
- *The 2000-char passthrough cap is introduced by this PR.* Dropped: it is develop's `fa82e18d`; the PR head alone has `stdout.trim().to_string()`.
- *A first argument that is a real directory named `server`/`analyze`/`clean` is now misrouted.* Dropped: native ruff resolves those the same way, so rtk matching it is the correct behaviour, not a bug.
- *`is_format` being true for `help format` / `config format` corrupts their output.* Dropped: `filter_ruff_format` falls through to identity; both are byte-identical to native ruff. Only the cap inconsistency survives, as **B**.

### Your questions

None open — no comments on the PR since it was filed.

### Next

fixup `afa9cae` pushed: adds `ruff_subcommands_cover_the_ruff_cli`, pinning `RUFF_SUBCOMMANDS` against a literal list so deleting an entry fails the suite (mutation-verified: dropping `"rule"` fails it, where the existing test stays green). Test-only, `fmt`/`clippy`/`test --all` green. Nothing else owed — approving.

PR #3084, opened 2026-07-20
fix(diff): return exit 2 for input errors

### LINKED ISSUE #2446: rtk diff exits 0 when files differ, breaking diff conditionals

lntutor:
## Summary
- handle file-read failures inside `rtk diff` instead of letting the global error handler flatten them to exit code `1`
- return POSIX-compatible code `2` for missing or unreadable inputs
- preserve code `0` for identical files and code `1` for differences
- add regression coverage for either input being missing

## Testing
- direct binary verification: identical=`0`, different=`1`, missing=`2`
- `cargo test --all-features` (2,496 passed, 8 ignored)

Closes #2446

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -15,8 +15,14 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
         eprintln!("Comparing: {} vs {}", file1.display(), file2.display());
     }
 
-    let content1 = fs::read_to_string(file1)?;
-    let content2 = fs::read_to_string(file2)?;
+    let content1 = match read_diff_file(file1) {
+        Ok(content) => content,
+        Err(exit_code) => return Ok(exit_code),
+    };
+    let content2 = match read_diff_file(file2) {
+        Ok(content) => content,
+        Err(exit_code) => return Ok(exit_code),
+    };
     let raw = format!("{}\n---\n{}", content1, content2);
 
     let (rtk, exit_code) = render_file_diff(file1, file2, &content1, &content2);
@@ -30,6 +36,13 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
         shown,
     );
     Ok(exit_code)
+}
+
+fn read_diff_file(path: &Path) -> std::result::Result<String, i32> {
+    fs::read_to_string(path).map_err(|error| {
+        eprintln!("diff: {}: {error}", path.display());
+        2
+    })
 }
 
 /// Renders the condensed file comparison and returns it with the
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as a duplicate of #2473, which makes the same fix: a read failure on either operand now returns POSIX diff's exit 2 instead of collapsing into exit 1.

Confirmed still live on 0.46.0 — `rtk diff x /nope; echo $?` prints `1`, which is indistinguishable from "files differ". The first half of #2446 (differ → exit 1) is already fixed; this is the half that remains.

#2473 was picked for one reason: its tests exercise the `read_operands` helper directly, where this PR's test calls `run()`. `run()` goes through `timer.track(...)`, so the test writes into whatever tracking database the machine is configured with — the exact problem #3758 is currently fixing. A test that mutates the user's real SQLite file to assert an exit code is worth avoiding.

The `read_diff_file` shape here is otherwise fine, and the two-way test (missing first operand, missing second) is the right coverage.

Issue #3037, opened 2026-07-17
rtk hook copilot emits permissionDecision "ask", forcing a permission prompt on every command in Copilot CLI >= 1.0.66


OliveiraMaicon:
### Summary
The Copilot `PreToolUse` hook (`rtk hook copilot`) returns `"permissionDecision":"ask"` in its `hookSpecificOutput`. Recent versions of GitHub Copilot CLI (1.0.66+) now **honor** this field, so every rewritten command triggers a `Yes/No` approval dialog labeled *"RTK auto-rewrite"*. Previously the CLI ignored the field, so RTK ran transparently.

### Environment
- **RTK:** 0.43.0 (Homebrew, latest stable)
- **Copilot CLI:** 1.0.71
- **OS:** macOS (Darwin, arm64)

### Steps to reproduce
1. Configure the repo hook `.github/hooks/rtk-rewrite.json`:
   ```json
   {
     "hooks": {
       "PreToolUse": [
         { "type": "command", "command": "rtk hook copilot", "cwd": ".", "timeout": 5 }
       ]
     }
   }
   ```
2. Start `copilot` (1.0.66+) in the repo.
3. Run any shell command (e.g. `git status`).

### Actual behavior
A permission dialog appears for every command: *"RTK auto-rewrite — Yes / No"*. There is **no "remember for this location"** option for hook decisions, so it prompts on every single command, breaking the transparent workflow.

Current hook output:
```json
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"ask","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk git status"}}}
```

### Expected behavior
For a transparent, safe rewrite (same command, just proxied through `rtk`), the hook should return `"permissionDecision":"allow"` so the CLI suppresses the approval prompt. Per the Copilot CLI changelog: *"preToolUse hook permissionDecision 'allow' now suppresses the tool approval prompt."*

Expected output:
```json
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"allow","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk git status"}}}
```

### Root cause
The Copilot hook contract changed: `permissionDecision` moved from advisory to authoritative. `ask` now forces a prompt. RTK still emits `ask`.

### Suggested fix
[...]

PR #3093, opened 2026-07-20
fix(copilot): support IDE terminal hooks

### LINKED ISSUE #2443: rtk hook copilot no-op in JetBrains/IntelliJ Copilot plugin — camelCase `run_in_terminal` not matched, and IDE host only honors `permissionDecision: deny`

lntutor:
## Summary
- recognize JetBrains Copilot's camelCase `run_in_terminal` payload separately from Copilot CLI `bash`
- return top-level deny-with-suggestion responses because the IDE host ignores `modifiedArgs`
- enforce explicit permission denials through the same host-supported response
- stay silent for deferred, unsupported, already-RTK, and non-terminal tool calls
- preserve existing Copilot CLI transparent rewrite behavior

## Testing
- focused IDE detection, rewrite, deny, and defer tests
- full existing Copilot CLI regression suite
- `cargo test --all-features` (2,499 passed, 8 ignored)

Closes #2443


KuSh:
Heads-up on some overlapping context from #3212 (in review), which also touches `detect_format`/`HookFormat::CopilotCli` in `src/hooks/hook_cmd.rs`:

- We found the `handle_vscode`/PascalCase path had never had its literal `permissionDecision: "ask"` fixed (that's the root cause of #3037 — `84aa4d6`/`0df6929` only patched the camelCase native path). #3212 makes both schemas only assert `permissionDecision: allow` for an explicit user Allow rule, mirroring `process_claude_payload`.
- We also found `rtk init --copilot`'s dual `PreToolUse`/`preToolUse` registration causes Copilot CLI to run both hooks sequentially per tool call (confirmed live, order-independent) — the camelCase one does the real rewrite, the PascalCase one then defers on the already-rewritten command. #3212 drops the camelCase registration since Copilot CLI honors the PascalCase-only schema fine on its own; existing installs upgrade on next `rtk init --copilot` re-run.
- Practically for this PR: the camelCase `toolName`/`toolArgs` schema is no longer what `rtk init --copilot` generates, but it's still very much alive — both for installs that haven't re-run init yet, and as the schema JetBrains/IntelliJ's plugin already uses under `toolName: "run_in_terminal"` (distinct from Copilot CLI's `"bash"`). So this PR's approach (recognizing `run_in_terminal` in the camelCase branch, deny-with-suggestion for the IDE host) still stands on its own — no conflict, just flagging the shared code area so the rebase/review has context.

Happy to coordinate further if useful.

src/hooks/hook_cmd.rs
```diff
@@ -811,6 +867,37 @@ mod tests {
...
+            response["permissionDecisionReason"],
+            "Blocked by RTK permission rule"
+        );
+    }
+
+    #[test]
+    fn test_copilot_ide_defer_is_silent() {
+        assert!(copilot_ide_response_from_decision(HookDecision::Defer, "htop").is_none());
+    }
+
```
KuSh, line 950: minor: not test for AllowRewrite. This would make the "Allow also becomes deny" behavior explicit and regression-proof rather than implicit from reading the match arm

KuSh (changes requested):
Hi @lntutor,
Thanks for your contribution!

It LGTM apart from a missing test (minor).
You'll also have to sign the CLA before we could merge your work.

src/hooks/hook_cmd.rs
lntutor: Added `test_copilot_ide_allow_rewrite_returns_deny_with_suggestion` in d0f004f: it drives `copilot_ide_response_from_decision` with `HookDecision::AllowRewrite(...)` and asserts `permissionDecision` is `deny`, that the reason carries the rewritten command, and that `modifiedArgs` is absent. That pins the "Allow also becomes deny" behavior explicitly instead of leaving it implicit in the `AllowRewrite | AskRewrite` arm.

lntutor:
@KuSh both review points are addressed:

- **Missing test** — added `test_copilot_ide_allow_rewrite_returns_deny_with_suggestion` (in `d0f004f`), driving `copilot_ide_response_from_decision` with `HookDecision::AllowRewrite(...)` and asserting a `deny` decision, the rewritten command in the reason, and no `modifiedArgs`. This pins the "Allow also becomes deny" behavior explicitly rather than leaving it implicit in the `AllowRewrite | AskRewrite` arm. Local gate green: `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --all-features` (2442 passed).
- **CLA** — signed; `license/cla` is passing.

Re-requesting review when you have a moment.

KuSh (approved):
LGTM 🚀

KuSh (changes requested):
There are a few clippy lints that need addressing before merging

lntutor:
@KuSh pushed `b434fd2` — the CI failure wasn't clippy lints but a semantic merge conflict: `develop`'s `0df6929` changed `HookDecision::AskRewrite` from a tuple variant `AskRewrite(String)` to a struct variant `AskRewrite { rewritten, explicit }`. My branch still had the old tuple form, so it compiled/tested green locally but broke once merged against `develop` (that's exactly the shared `hook_cmd.rs` area you flagged from #3212/`0df6929`).

Fixed by merging `develop` and updating the two surviving tuple-form usages:
- `copilot_ide_response_from_decision` match arm → `AskRewrite { rewritten, .. }`
- `test_copilot_ide_rewrite_returns_deny_with_suggestion` constructor → `AskRewrite { rewritten: ..., explicit: false }` (default, non-user-configured rewrite; the IDE path ignores the field)

Resolution is folded into the merge commit so it compiles bisect-clean. Local gate green: `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` (no warnings), `cargo test --all-features` (2497 passed).

The new run is stuck at `action_required` — mind approving the CI run and taking another look when you have a moment? Thanks!

KuSh (approved):

pszymkowiak:
Quick status check since the last review round: the clippy lints flagged between the `LGTM 🚀` and the follow-up `CHANGES_REQUESTED` look resolved already. The merge commit `b434fd2` fixed the semantic conflict with develop's `0df6929` (`AskRewrite` tuple → struct variant), and building the branch locally confirms it's green:

- `cargo fmt --all --check` — clean
- `cargo test --all` — 2561 passed, 8 ignored

Might be worth a re-review to clear the stale CHANGES_REQUESTED — the blocking issue itself looks fixed.

Separate, non-blocking observation while reading through: `copilot_ide_response_from_decision`'s `AllowRewrite`/`AskRewrite` arm calls `audit_log("rewrite", cmd, &rewritten)` — the same action label used elsewhere for rewrites that actually get applied transparently. Here the response is a deny-with-suggestion (the user has to manually re-type the command), so it's a different kind of event even though the label is identical. Not asking for a change here, just flagging in case it's worth a distinct label (e.g. `"suggest"`) later, since `hook_audit_cmd.rs` counts `action == "rewrite"` for its stats and this would blend two different outcomes into one number.

PR #3101, opened 2026-07-20
fix(git): avoid byte-slice panic in parse_commit_output on non-bracket lines

### LINKED ISSUE #3100: `git commit` output parser still panics when first line starts with a multibyte char and contains `]` (parse_commit_output, v0.43.0)
v0.43.0 rewrote `parse_commit_output` (src/cmds/git/git.rs) and added multibyte regression tests, but the slice `&line[1..bracket_end]` is still byte-indexed and only guarded by `line.find(']')` — not by a check that the line starts with `[`.

If the first stdout line of a successful `git commit` starts with a multibyte character AND contains `]`, byte index 1 falls inside the first character and the process aborts (`panic = "abort"`).

Realistic reproduction: a successful pre-commit hook whose stdout precedes git's summary line:

```sh
#!/bin/sh
printf '%s\n' '검사 완료] 커밋 계속'   # "check done] continuing commit"
exit 0
```

`git commit` exits 0, stdout is:

```
검사 완료] 커밋 계속
[master 6608d86] probe commit
```

→ `thread 'main' panicked ... start byte index 1 is not a char boundary`.

Related: in v0.37.2 the pre-rewrite parser panicked whenever the first stdout line started with a multibyte character and contained a space (`line[1..hash_start]` with `hash_start = line.find(' ')`), e.g. Korean-locale `git commit --dry-run` whose first line is `현재 브랜치 main` ("On branch main") — we hit this 3 times in a single day of normal use. v0.43.0 fixes that case (no `]`), but the guarded variant above remains.

Suggested fix: only slice when the line actually starts with `[` (`line.strip_prefix('[')` + `find(']')` on the stripped remainder), or use `line.get(1..bracket_end)` and fall back to `"ok"`.

Existing related open issues on byte-index panics: #2954, #2851, #2318, #2787, #2509.

SAY-5:
## Summary

- Fixes #3100: `parse_commit_output` sliced `&line[1..bracket_end]` whenever the line contained `]`, without checking the line actually started with `[`. If the first character was multibyte, byte index 1 landed mid-character and the process aborted (`panic = "abort"`).
- Slices only after `strip_prefix('[')` confirms the leading bracket, so the offset is always on a char boundary and non-`[` lines fall through to `"ok"`.
- Added a regression test with the reported input (`검사 완료] 커밋 계속`).

## Test plan

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1009,18 +1009,17 @@ fn build_commit_command(args: &[String], global_args: &[String]) -> Command {
 /// Handles: `[main abc1234def] message`, `[main (root-commit) abc1234def] msg`,
 /// localized variants, and multibyte branch names.
 fn parse_commit_output(line: &str) -> String {
-    if let Some(bracket_end) = line.find(']') {
-        let bracket_content = &line[1..bracket_end];
-        let hash = bracket_content.split_whitespace().next_back().unwrap_or("");
-        if !hash.is_empty() && hash.len() >= 7 {
-            let short_hash: String = hash.chars().take(7).collect();
-            format!("ok {}", short_hash)
-        } else {
-            "ok".to_string()
+    if let Some(rest) = line.strip_prefix('[') {
+        if let Some(bracket_end) = rest.find(']') {
+            let bracket_content = &rest[..bracket_end];
+            let hash = bracket_content.split_whitespace().next_back().unwrap_or("");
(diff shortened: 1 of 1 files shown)


SAY-5:
That curl json regression is pre-existing overhead from the wrapper prefix, unrelated to the parse_commit_output change in this diff. Leaving it to the benchmark gate maintainers to decide on threshold handling.

PR #3104, opened 2026-07-20
tests: fix tests on non en env

KuSh:
## Summary
Make `cargo test` on non english env

## Test plan

### DIFF
--- a/tests/guard_integration_test.rs
+++ b/tests/guard_integration_test.rs
@@ -5,6 +5,7 @@ use std::process::{Command, Stdio};
 
 fn rtk_stdin(args: &[&str], input: &str) -> String {
     let mut child = Command::new(env!("CARGO_BIN_EXE_rtk"))
+        .env("LANG", "C")
         .args(args)
         .stdin(Stdio::piped())
         .stdout(Stdio::piped())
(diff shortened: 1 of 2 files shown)


tests/guard_integration_test.rs
```diff
@@ -5,6 +5,7 @@ use std::process::{Command, Stdio};
 
 fn rtk_stdin(args: &[&str], input: &str) -> String {
     let mut child = Command::new(env!("CARGO_BIN_EXE_rtk"))
+        .env("LANG", "C")
```
TaKO8Ki: We don't need to use `LC_ALL` here? It seems that an inherited `LC_ALL` or `LC_MESSAGES` takes precedence over `LANG`, so `LANG=C` may not reliably force English output.

> LC_ALL
This variable shall determine the values for all locale categories. The value of the LC_ALL environment variable has precedence over any of the other environment variables starting with LC_ (LC_COLLATE , LC_CTYPE , LC_MESSAGES , LC_MONETARY , LC_NUMERIC , LC_TIME ) and the LANG environment variable.

ref: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html

KuSh: You're absolutely right, I've changed that!

TaKO8Ki (approved):
Thanks!

Issue #3107, opened 2026-07-20
`rtk read` buffers whole file (3.9x filesize RSS); `--tail-lines`/`--max-lines` bound output but not memory; `exclude_commands` cannot disable the `tail -n` rewrite


Cybercommanders:
## Summary

A hook-rewritten `tail -n 3000 big.log` consumed **49 GB RSS + 42 GB swap** and drove a 64 GB machine into OOM twice. The agent had issued a correctly-bounded `tail`; the PreToolUse hook converted it into `rtk read --tail-lines 3000`, which reads the entire file into memory before discarding all but the last N lines. The file was a 161 GB unrotated debug log.

Three distinct defects combine so that there is currently **no supported configuration that avoids this**.

Environment: rtk 0.43.0 (latest release), x86_64 Linux (WSL2), Claude Code PreToolUse hook installed per `rtk init -g`.

## 1. `rtk read` does not stream — it holds ~3.9x file size

Measured with `/usr/bin/time -v`, output piped to `/dev/null` so no shell command-substitution buffering is involved:

```bash
yes "0123456789abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz" \
  | head -c 2000000000 > big.log        # 2.0 GB

/usr/bin/time -v rtk read big.log                  2>&1 >/dev/null | grep 'Maximum resident'
/usr/bin/time -v rtk read big.log --tail-lines 5   2>&1 >/dev/null | grep 'Maximum resident'
/usr/bin/time -v rtk read big.log --max-lines 5    2>&1 >/dev/null | grep 'Maximum resident'
/usr/bin/time -v tail -n 5 big.log                 2>&1 >/dev/null | grep 'Maximum resident'
```

| command | peak RSS | ratio to file |
|---|---|---|
| `rtk read big.log` | 7,818,312 KB (7.8 GB) | **3.91x** |
| `rtk read big.log --tail-lines 5` | 4,338,312 KB (4.34 GB) | 2.17x |
| `rtk read big.log --max-lines 5` | 4,339,156 KB (4.34 GB) | 2.17x |
| `tail -n 5 big.log` | 2,080 KB | 0.001x |

Plain `read` is the *worst* case at 3.9x — presumably input and compressed output are held simultaneously. Extrapolated to the 161 GB log this is ~630 GB of allocation; it died at 91 GB.

[...]

PR #3109, opened 2026-07-20
fix(hook): honor exclusions for tail line ranges

juliosuas:
## Summary

- honor `hooks.exclude_commands` before the special `head`/`tail` line-range rewrite
- add literal and regex regression coverage for the `tail -n` escape hatch
- address the configuration bypass reported in #3107 without expanding this PR into the separate streaming/memory work

## Test plan

- [x] Manual testing: an isolated config with `tail` or `^tail -n` leaves `tail -n 400 /var/log/foo.log` unchanged; without an exclusion it still rewrites to `rtk read /var/log/foo.log --tail-lines 400`

Security note: this changes Tier 1 rewrite logic in `src/discover/registry.rs`; the diff received isolated full-suite testing and an independent senior security/quality review before publication.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -996,6 +996,10 @@ fn rewrite_segment_inner(
     if context == RewriteContext::Normal
         && (cmd_part.starts_with("head -") || cmd_part.starts_with("tail "))
     {
+        let normalized = strip_absolute_path(cmd_part.trim());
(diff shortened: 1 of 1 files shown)


juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/rtk-ai/rtk/pull/3719

PR #3119, opened 2026-07-21
fix(grep): accept GNU grep flags -E/-r/-R/-i/-w/-A/-B/-C (#2614), and stop -E/-r from silently eating the next arg (#2253)

RichelynScott:
Fixes the clap-layer rejection described in #2614, and closes a silent-corruption path that is the likely mechanism behind #2253.

## Problem

`rtk grep` rejected the most common real-world grep invocations, so the hook's rewrite **guaranteed** a parse failure followed by a silent raw fallback at 0 tokens saved.

From a 5-week local tracking DB (11,932 commands): **2,142 parse failures, 1,970 of them `grep` (92%)**, every one with `fallback_succeeded=1` — they read as success while saving nothing. This independently reproduces @-reporter's 4,416/90-day figure in #2614.

Root cause: the rejected flags arrive **before** the `pattern`/`path` positionals, so `trailing_var_arg` never collected them into `extra_args`. `grep_cmd` already had `-r`-skip and format-flag passthrough logic, but clap failed first, so that code was unreachable.

## Two classes of flag, handled differently

**Dropped** — accepted by clap, never forwarded to rg. Their rg short forms collide with grep's *and take a value*, so forwarding silently consumes the following argument:

| flag | GNU grep | ripgrep |
|---|---|---|
| `-E` | extended regex (bool) | `--encoding=VALUE` |
| `-r` | recursive (bool) | `--replace=VALUE` |

This is very likely the mechanism behind **#2253** ("matches silently rewritten `def` → `n`"): a forwarded `-E`/`-r` eats the pattern. Dropping them is semantically correct — rg is recursive and extended-regex by default, so grep's intent is already satisfied.

**Forwarded** — identical semantics in both tools: `-i`, `-w`, `-A/-B/-C <NUM>`.

## Not changed (deliberate)

`-l` still means `max_len`. Aligning it with GNU's `--files-with-matches` would break existing CLI behavior for anyone scripting `rtk grep -l`, so it is left alone and called out in a test. Happy to do it in a follow-up if maintainers want the breaking change.

## Verification

Against `/usr/bin/grep` (not the rtk-rewritten shell alias):

| case | real grep | rtk grep |
|---|---|---|
| `-E 'fn run\|fn filter'` | 1 | 1 |
| `-ci LAZY_STATIC` | 1 | 1 |
| `-w run` | 3 (L12, 298, 404) | 3 (L12, 298, 404) |

All five previously-failing forms now parse: `-nE`, `-rn --include`, `-c`, `-n -A2`, `-i`.

## Tests

4 regression tests, including a guard that dropping a flag does not consume the following token (the #2253 shape). Suite **2147 → 2151 passing**.

The 3 failures in `hooks::rewrite_cmd::tests::unattestable_passthrough` are **pre-existing on HEAD** — verified by stashing this change and re-running (identical 3 failures, identical 2147 pass count). Unrelated to this PR.

`cargo fmt --all` clean, `cargo clippy --all-targets` 0 warnings.

### DIFF
--- a/src/cmds/system/grep_cmd.rs
+++ b/src/cmds/system/grep_cmd.rs
@@ -43,8 +43,7 @@ pub fn run(
     }
 
     for arg in extra_args {
-        // Fix: skip grep-ism -r flag (rg is recursive by default; rg -r means --replace)
-        if arg == "-r" || arg == "--recursive" {
+        if is_rg_incompatible_grep_flag(arg) {
             continue;
         }
         rg_cmd.arg(arg);
@@ -180,6 +179,26 @@ fn parse_match_line(line: &str) -> Option<(String, usize, &str)> {
         let line_num: usize = line_num.parse().ok()?;
         Some((file.to_string(), line_num, content))
     })
+}
+
+/// GNU-grep flags that must NOT be forwarded to ripgrep.
+///
+/// `rg`'s short forms for these collide with grep's and, critically, **take a
+/// value** — so forwarding them silently consumes the following argument
+/// (the mechanism behind rtk-ai/rtk#2253, where a pattern got eaten):
+///
+/// | flag | GNU grep          | ripgrep              |
+/// |------|-------------------|----------------------|
+/// | `-E` | extended regex    | `--encoding=VALUE`   |
+/// | `-r` | recursive         | `--replace=VALUE`    |
+///
+/// Dropping them is semantically correct: `rg` is recursive and
+/// extended-regex by default, so grep's intent is already satisfied.
+fn is_rg_incompatible_grep_flag(arg: &str) -> bool {
+    matches!(
+        arg,
+        "-r" | "--recursive" | "-R" | "--dereference-recursive" | "-E" | "--extended-regexp"
+    )
 }
 
 fn has_format_flag(extra_args: &[String]) -> bool {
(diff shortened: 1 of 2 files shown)


RichelynScott:
Closing — superseded by upstream, and no longer mergeable.

Since this PR was filed, upstream `develop` landed a more comprehensive grep fix that this PR's approach cannot match:

- `eafadce` — run the invoked engine verbatim (grep→grep, rg→rg), dropping the rg-substitution architecture entirely (`grep_cmd.rs` → `search.rs`). This is the root fix: BRE `\{n,m\}`/`\(...\)` patterns now reach GNU grep unchanged.
- `0adfae6` — general flag-shape parsing + raw-output fallback.
- `d727db3` — surface errors on exit ≥2 instead of a false "0 matches".

This PR only added individual clap fields for `-E/-r/-R/-i/-w/-A/-B/-C` and reconstructed ripgrep args, while still substituting rg for grep — so it left the BRE-dialect defect unfixed. Upstream's engine-fidelity approach is strictly better. Verified: a from-source build of current upstream returns correct matches for BRE interval/group patterns where the fork binary returned "0 matches".

Thanks to the maintainers for the more thorough fix. Closing in favor of it. (Refs #2614, #2253.)

PR #3121, opened 2026-07-21
fix(git): filter git show <rev>:<path> blobs instead of passthrough

### LINKED ISSUE #3116: rtk git show <rev>:<path> is a file read but routes to passthrough (0% savings) instead of the read filter
`git show <rev>:<path>` is not a diff — it's a **file read**. But rtk routes it to passthrough and emits the entire file uncompressed, while `rtk read` exists specifically to filter file content.

## Evidence

From a 5-week tracking DB (11,932 commands, `~/.local/share/rtk/history.db`), grouping by `rtk_cmd`:

| Invocation | Calls | Input tok | Output tok | Saved |
|---|---|---|---|---|
| `rtk git show origin/main:rust/…/main.rs (passthrough)` | 12 | 987.6K | 987.6K | **0.0%** |
| `rtk git show origin/main:rust/…/lib.rs (passthrough)` | 5 | 381.8K | 381.8K | **0.0%** |
| `rtk git show <sha>:rust/…/window.rs (passthrough)` | 4 | 273.5K | 273.5K | **0.0%** |
| `rtk git show origin/main:…/window.rs (passthrough)` | 3 | 189.0K | 189.0K | **0.0%** |
| `rtk git show master:CLAUDE.md (passthrough)` | 7 | 81.6K | 81.6K | **0.0%** |
| `rtk git show <sha> -- crates/…/policy.rs` (diff form) | 4 | 27.9K | 13.9K | 50.1% |

Aggregate for `rtk git show`: **436 calls, 3.82M input → 3.79M output, 0.6% saved.** A single `git show origin/main:rust/limux-cli/src/main.rs` costs **84,657 tokens**, and it was called 6 times.

Note the contrast in the last row: the `git show <sha> -- <path>` **diff** form filters fine (50.1%). Only the `<rev>:<path>` **blob** form falls through.

## Reproduction

```console
$ git show HEAD:src/cmds/system/grep_cmd.rs | wc -c
17798
$ rtk git show HEAD:src/cmds/system/grep_cmd.rs | wc -c
17798      # identical — no filtering
```

## Why this matters

`git show <rev>:<path>` is the standard way an agent reads a file at another revision (comparing against `origin/main`, inspecting a file on another branch, reviewing a file pre-merge). It is a very common agent pattern, and each call can dump an entire source file into context at full cost.

## Suggested fix

Detect the `<rev>:<path>` (blob) form in the `git show` handler and route it through the existing `read` filter pipeline — using the path's extension for language detection — instead of marking it passthrough. The `<rev> -- <path>` and plain `<rev>` diff forms should keep their current diff-filter behavior.

A conservative first cut: only reroute when the argument matches `^[^:]+:.+` and the resolved blob exceeds some threshold, so tiny files stay passthrough.

_rtk 0.42.2, Linux (WSL2)_

MannXo:
## Summary

- `git show <rev>:<path>` prints a file at a revision (a file read), but it was routed to passthrough and emitted uncompressed at 0% savings, while the `<rev> -- <path>` diff form already compacts.
- Route the blob form through the same filtering `rtk read` applies: minimal comment/blank-line stripping with language detected from the path extension, capped at the same 500-line default the diff form uses, guarded by `never_worse` so small files pass through unchanged.
- `git show --stat` and `--format` variants keep their previous behavior.

Fixes #3116

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` (2444 lib tests pass, zero clippy warnings)
- [x] New unit tests in `src/cmds/git/git.rs`: blob-path extraction, comment stripping, 500-line cap with truncation marker, unknown-extension faithfulness, `never_worse` on tiny input, and a ≥60% savings check
- [x] Manual testing against a real repo:
  - `rtk git show HEAD:Cargo.toml` — small file, byte-identical to raw (faithful)
  - `rtk git show HEAD:src/cmds/git/git.rs` — 3325-line file capped to 500 lines with a `[2692 more lines]` marker

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1,6 +1,7 @@
 //! Filters git output — log, status, diff, and more — keeping just the essential info.
 
 use crate::core::args_utils;
+use crate::core::filter::{self, FilterLevel, Language};
 use crate::core::guard::never_worse;
 use crate::core::runner::{self, RunOptions};
 use crate::core::stream::{
@@ -13,6 +14,7 @@ use crate::core::utils::{
 };
 use anyhow::{Context, Result};
 use std::ffi::OsString;
+use std::path::Path;
 use std::process::Command;
 use std::process::Stdio;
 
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into this. The blob form is filtered on develop now: #3265 (merged 2026-09-11) routes `git show <rev>:<path>` into a size-capped preview window instead of raw passthrough, so the 0% savings case you reported is gone. Measured on current develop, a 259 KB blob comes back as 8 KB. The merged version keeps raw bytes rather than stripping comments, deliberately, so that non-UTF-8 and binary blobs survive the window, and it also handles the --stat and --format combinations you left on the old path. Closing as covered, but please comment here or reopen if you hit a blob shape the new path still misses.

PR #3125, opened 2026-07-21
docs(config): update example to include `max_file_size`

derrik-fleming:
## Summary

- updates the configuration example in the documentation, it was incomplete and missing an expected value, `max_file_size`.

## Test plan

adding the value to my config allowed rtk to successfully parse it

### DIFF
--- a/docs/guide/getting-started/configuration.md
+++ b/docs/guide/getting-started/configuration.md
@@ -42,6 +42,7 @@ ignore_files = ["*.lock", "*.min.js", "*.min.css"]
 enabled = true              # save raw output on failure
 mode = "failures"           # "failures" (default), "always", "never"
 max_files = 20              # rotation: keep last N files
+max_file_size = 1048576     # 1 MB in bytes
 # directory = "/custom/tee/path"  # optional override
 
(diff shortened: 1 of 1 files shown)


pszymkowiak (approved):
Verified this is a real bug, not just a cosmetic doc gap: copying the current docs example verbatim (with a `[tee]` section, no `max_file_size`) produces a hard TOML parse failure:

```
rtk: TOML parse error at line 17, column 1
   |
17 | [tee]
   | ^^^^^
missing field `max_file_size`
```

Root cause: `TeeConfig::max_file_size` (`src/core/tee.rs`) has no `#[serde(default)]`, unlike the top-level `tee` field on `RtkConfig` which defaults only when the whole `[tee]` section is absent. So: no `[tee]` section → fine; `[tee]` section present but missing `max_file_size` → hard parse error, exactly what the previous docs example triggered.

With the PR's added line (`max_file_size = 1048576`, matching `DEFAULT_MAX_FILE_SIZE` in `tee.rs`), the same config parses cleanly. Minimal, accurate, one-line fix. Approving.

Issue #3122, opened 2026-07-21
feat: add sips and soffice (LibreOffice) filters — compress image inspection and headless conversion output


gustavoadosanjos7-dot:
**Context**: Workflow generates PPTX decks (python-pptx) with embedded images, then uses `sips` to check image dimensions before embedding and `soffice --headless --convert-to pdf` to render a PDF preview for client review.

**Commands seen repeatedly** (17x `sips`, 12x `soffice` in 30 days of one user's session history):
```
sips -g pixelWidth -g pixelHeight "$file"
soffice --headless --convert-to pdf --outdir "$dir" "$file.pptx"
```

**Problem**: Both are noisy for LLM context.
- `sips -g` prints multi-line key/value blocks per file even when only 1-2 values are needed.
- `soffice --headless` prints startup banners, temp profile warnings, and progress noise; only the final `convert ... -> ... using filter` / success-or-failure line matters.

**Ask**: `rtk sips` (strip to just the requested keys, one line per file) and `rtk soffice` (suppress banner/profile noise, keep only the conversion result line) — similar in spirit to existing `rtk vitest`/`rtk tsc` grouped-output filters.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟢 **Priority** | `low` |
| 🎯 **Confidence** | 90% |

### Summary

Request for new rtk output filters for `sips` (macOS image inspection) and `soffice --headless` (LibreOffice conversion) to reduce noisy CLI output, following the pattern of existing vitest/tsc filters; no existing duplicate found.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

MannXo:
Hi, I started to work on this issue.

PR #3126, opened 2026-07-21
feat(soffice): compact LibreOffice headless conversion output

MannXo:
## Summary

- Adds an `rtk soffice` filter for `soffice`/`libreoffice --headless --convert-to`. LibreOffice surrounds its one useful line (`convert <in> -> <out> using filter : <filter>`) with startup/profile banner noise: `javaldx` warnings on Java-less hosts, macOS `Task policy` messages, safe-mode banners. The filter keeps only the conversion result line and any genuine `Error:` line, dropping the rest.
- soffice writes both the banner noise **and** its errors to stderr, and exits `0` even when a conversion fails. So the filter processes the **combined** stream rather than stdout-only, which means a load error is surfaced instead of hidden. If no result/error line is matched, the raw output is passed through verbatim, and the runner's `never_worse` guard keeps output no larger than raw.
- Registered as a hook rewrite for `soffice … --convert-to` / `libreoffice … --convert-to`, added to the subcommand classification list, and covered by unit tests.

This addresses the **soffice** half of #3122. I left out the requested `sips` filter deliberately: `sips -g` output is almost entirely the file path plus the key/value data you asked for (there is no removable noise), so reformatting it saves only ~5–7% by rtk's token estimate — well under the project's savings bar. Happy to revisit if you'd like it purely for line-count tidiness, but it did not seem worth a filter on the token-savings metric.

**A note on savings honesty:** the win scales with how noisy the host is. On a Java-less headless Linux/WSL2 run (the reported environment) the `javaldx` banner pushes this comfortably past 60%. On a quiet macOS run with few extra lines it saves less; `never_worse` guarantees it is never a regression. Output formats were verified against a real LibreOffice 26.2.4 headless run.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` (2444 tests pass, zero clippy warnings)
- [x] Unit tests in `src/cmds/system/soffice_cmd.rs`: keeps the convert line while dropping javaldx/banner noise, drops the macOS `Task policy` line, surfaces an `Error:` line, falls back to raw on unrecognized output, and a ≥60% savings check on a representative Java-less run
- [x] Manual testing against real LibreOffice 26.2.4:
  - `rtk soffice --headless --convert-to pdf --outdir out deck.txt` produces the PDF and prints only the `convert … -> … using filter` line
  - a failed conversion (`Error: source file could not be loaded`) is surfaced, not swallowed

### DIFF
--- a/src/cmds/system/soffice_cmd.rs
+++ b/src/cmds/system/soffice_cmd.rs
@@ -1,0 +1,147 @@
+//! Compact filter for `soffice`/`libreoffice` headless conversion.
+//!
+//! `soffice --headless --convert-to <fmt> ...` emits startup/profile noise
+//! (javaldx warnings on Java-less hosts, macOS `Task policy` messages, safe-mode
+//! banners) around a single line that actually matters:
+//!
+//! ```text
+//! convert /in/deck.pptx as a Impress presentation -> /out/deck.pdf using filter : impress_pdf_Export
+//! ```
+//!
+//! This keeps only that conversion result line and any genuine `Error:` line,
+//! dropping the surrounding banner. soffice writes both the noise and its errors
+//! to stderr and exits 0 even on failure, so we filter the combined stream (not
+//! stdout only) to avoid hiding a failure. If nothing meaningful is matched, the
+//! raw output is kept verbatim so unexpected output is never swallowed.
+//!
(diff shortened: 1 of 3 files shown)


MannXo:
Closing this, since it has fallen behind main without being picked up for review. Happy to rebase and reopen if it's still wanted.

PR #3132, opened 2026-07-22
fix(read): bound memory for --tail-lines instead of loading the whole file

### LINKED ISSUE #3107: `rtk read` buffers whole file (3.9x filesize RSS); `--tail-lines`/`--max-lines` bound output but not memory; `exclude_commands` cannot disable the `tail -n` rewrite

albatrossflyon-coder:
Fixes #3107.

## The bug

`rtk read` always called `fs::read_to_string` on the *entire* file before applying `--tail-lines`/`--max-lines`, so a bounded `tail -n 3000` on a 161GB log materialized the whole file (plus later `.clone()`s in the line-numbers/raw-comparison path) in memory before the tail window was ever applied — the reported 49GB RSS OOM, with no config escape hatch.

## The fix

Added `read_tail_lines_bounded`: seeks backward from the end of the file in 64KB chunks, stopping once enough newlines have been seen to satisfy the requested tail. Memory use is bounded by the tail content itself, not file size. Used only for the `--tail-lines` path.

`--max-lines` (`smart_truncate`) is **intentionally left unbounded** in this PR: it scans the *entire* file for structurally important lines (function signatures, imports) anywhere in the document, so a prefix-only read would silently change its output — that's a separate, harder problem, flagged here as a follow-up rather than claimed as fixed.

## Testing

- 6 new unit tests covering the boundary cases that actually matter for a hand-rolled backward reader: normal tail, tail exceeding file length, no trailing newline, empty file, `tail_lines == 0`, and a tail that spans multiple 64KB chunk reads.
- Manual verification against a real 136MB / 2,000,000-line file: **41ms runtime**, byte-for-byte identical output to coreutils `tail -n 3000` on the same file (previously this unconditionally read the full 136MB).
- `cargo test --bin rtk`: 2463 passed, 8 ignored (pre-existing, require a built binary).

## Not addressed here

- `--max-lines` unbounded read (see above) — separate follow-up.
- `exclude_commands` not suppressing the `tail -n` hook-rewrite — didn't touch `hook/rewrite.rs`/`config.rs` in this PR since the underlying crash this issue reports is now fixed regardless of whether that rewrite fires; happy to take that on separately if useful.

### DIFF
--- a/src/cmds/system/read.rs
+++ b/src/cmds/system/read.rs
@@ -39,6 +39,22 @@ pub fn run(
             .with_context(|| format!("Failed to read file: {}", file.display()))?
     };
 
+    // Stateful filters (`--level minimal|aggressive`) track context (open
+    // block comments, in-progress docstrings) as they scan lines from the
+    // top of the file. `--tail-lines` only reads the tail slice, so the
+    // filter would start mid-context with no way to know it — e.g. a block
+    // comment opened before the tail window would leak its closing
+    // fragment as if it were code. Force `none` in that combination instead
+    // of risking silently corrupted output.
+    let effective_level = effective_filter_level(level, tail_lines);
+    if effective_level != level {
+        eprintln!(
+            "rtk: warning: --level {} has no effect with --tail-lines (the tail slice doesn't carry enough file context for stateful filtering)",
+            level
+        );
+    }
+    let level = effective_level;
+
     // Detect language from extension
     let lang = file
         .extension()
@@ -173,6 +189,32 @@ fn format_with_line_numbers(content: &str) -> String {
     out
 }
 
+/// `--tail-lines` reads only the tail slice of a file, so a stateful filter
+/// (`minimal`/`aggressive`) has no way to see context from earlier in the
+/// file. Returns `none` in that combination; returns `level` unchanged
+/// otherwise.
+fn effective_filter_level(level: FilterLevel, tail_lines: Option<usize>) -> FilterLevel {
+    if tail_lines.is_some() && level != FilterLevel::None {
+        FilterLevel::None
+    } else {
+        level
+    }
+}
+
+/// Splits `text` into lines like `str::lines()`, but keeps any trailing `\r`
+/// on each line intact. `str::lines()` strips `\r` before `\n`, which is
+/// exactly right for display but corrupts CRLF files when the split-and-join
+/// round trip is used to extract a byte-for-byte tail window.
+fn split_lines_preserve_cr(text: &str) -> Vec<&str> {
+    let mut lines: Vec<&str> = text.split('\n').collect();
+    // split('\n') on a trailing newline yields one extra "" element that
+    // str::lines() doesn't produce; drop it to keep the same line count.
+    if text.ends_with('\n') {
+        lines.pop();
+    }
+    lines
+}
+
 /// Reads roughly the last `tail_lines` lines of `path` without loading the
 /// whole file into memory: seeks backward from the end in fixed-size chunks,
 /// stopping once enough newlines have been seen (or the start of the file is
@@ -213,7 +255,7 @@ fn read_tail_lines_bounded(path: &Path, tail_lines: usize) -> Result<String> {
     let bytes: Vec<u8> = chunks.into_iter().flatten().collect();
     let text = String::from_utf8_lossy(&bytes).into_owned();
 
-    let lines: Vec<&str> = text.lines().collect();
+    let lines = split_lines_preserve_cr(&text);
     let start = lines.len().saturating_sub(tail_lines);
     let mut result = lines[start..].join("\n");
     if text.ends_with('\n') {
@@ -232,7 +274,7 @@ fn apply_line_window(
         if tail == 0 {
             return String::new();
         }
-        let lines: Vec<&str> = content.lines().collect();
+        let lines = split_lines_preserve_cr(content);
         let start = lines.len().saturating_sub(tail);
         let mut result = lines[start..].join("\n");
         if content.ends_with('\n') {
@@ -277,6 +319,15 @@ mod tests {
         write!(file, "a\nb\nc")?;
         let bounded = read_tail_lines_bounded(file.path(), 2)?;
         assert_eq!(bounded, "b\nc");
+        Ok(())
+    }
+
+    #[test]
+    fn test_read_tail_lines_bounded_preserves_crlf() -> Result<()> {
+        let mut file = NamedTempFile::new()?;
+        write!(file, "a\r\nb\r\nc\r\n")?;
+        let bounded = read_tail_lines_bounded(file.path(), 2)?;
+        assert_eq!(bounded, "b\r\nc\r\n");
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


pszymkowiak (changes requested):
The memory fix itself is solid — verified on a compiled build that a tiny `--tail-lines` on 200–300 MB files stays at ~16.6 MB RSS (OOM #3107 genuinely fixed), the backward-chunk reassembly is correct, and it's byte-exact with `tail -n N` across LF fuzzing (3141/3141), UTF-8/emoji straddling 64 KB chunk boundaries (6/6), and all the count edge cases (N=0/1/=lines/>lines, empty, no-trailing-newline). Nice work on the off-by-one margin and byte-level UTF-8 handling.

Two correctness regressions on the `--tail-lines` path, though — both should be fixed before merge:

**1. CRLF line endings get normalized to LF (tail-lines only).**
`read_tail_lines_bounded` does `text.lines().join("\n")`, which drops the `\r`:
```
file:  a\r\nb\r\nc\r\n
tail -n 1                → c\r\n   (63 0d 0a)   ✓
rtk read --tail-lines 1  → c\n      (63 0a)      ✗  \r lost
rtk read --max-lines 1   → CRLF preserved
rtk read (whole file)    → CRLF preserved
```
Only the new path corrupts CRLF; the whole-file and `--max-lines` paths preserve it, so this is a regression, not pre-existing.

**2. Stateful filters lose context when combined with `--level minimal|aggressive`.**
The bounded reader now feeds the filter only the tail slice instead of the whole file, and `MinimalFilter`/`AggressiveFilter` track block-comment state line-by-line:
```rust
fn main() {}
/* long comment
spanning several
lines
*/
fn foo() {}
```
`rtk read file.rs --level minimal`                → strips the comment ✓
`rtk read file.rs --level minimal --tail-lines 3` → `lines\n*/\nfn foo() {}` ✗ (comment fragment leaks; the filter never saw the opening `/*`)
Latent because the default level is `none`, but it's silent output corruption when both flags are used. The PR comment flags this context-loss risk for `max_lines`/`smart_truncate` but the same class now applies to `tail_lines` + stateful filtering.

**Suggested fix:** operate on raw bytes — locate the byte offset of the N-th-from-last `\n` and emit that slice verbatim (no `.lines().join()`), which fixes CRLF for free; and either skip stateful filtering on the tail-bounded slice or document that `--tail-lines` disables it. A CRLF test and a block-comment-straddling-the-tail-boundary test would lock both down.

Minor: `from_utf8_lossy` on the tail path silently succeeds on invalid UTF-8 while the whole-file `read_to_string` path errors — worth a deliberate call on which behavior wins.

albatrossflyon-coder:
Both issues fixed in `f8a9a81`:

1. **CRLF preservation** — replaced the `.lines().join("\n")` round trip (which strips `\r`) with a `split_lines_preserve_cr` helper that splits on `\n` only, so any trailing `\r` survives. Applied in both `read_tail_lines_bounded` and `apply_line_window`. Added a regression test for each.
2. **Stateful filter context loss** — `--tail-lines` now forces the effective filter level to `none` when combined with `--level minimal|aggressive`, with a stderr warning explaining why (the tail slice can't carry the file's prior block-comment/docstring state). Added tests for the override logic.

Live-verified against your exact repro cases — CRLF now round-trips byte-for-byte, and the block-comment-straddling-the-tail-boundary case now warns and shows raw content instead of leaking the comment fragment as code. Thanks for the thorough review.

albatrossflyon-coder:
Both issues from the review are addressed in `f8a9a81`: `split_lines_preserve_cr` now keeps `\r` intact on the `--tail-lines` path (matches whole-file/`--max-lines` behavior), and stateful filters are skipped on the tail-bounded slice to avoid the dangling-comment-fragment case. Added `test_read_tail_lines_bounded_preserves_crlf` to lock in the CRLF fix. Ready for another look whenever you get a chance.

PR #3134, opened 2026-07-22
fix(search): route rtk grep/rg through the already-bounded capture path

### LINKED ISSUE #3106: rtk rg can consume over 50 GB by buffering and duplicating large search output
RTK version 0.43.0 on Windows consumed more than 50 GB of private memory while processing one large rg result.

The offending command line was captured from the running process. Paths are normalized below.

    rtk rg -n "load_workspace_dependencies|workspace_dependencies|models/list|/models|model_provider|wire_api|supports_tools|tool server" src docs issues "%USERPROFILE%\.codex" "%APPDATA%\npm\node_modules\@openai\codex"

Observed behavior

- Working set was initially 21.111 GB and later reached 37.236 GB
- Private memory reached 50.829 GB
- Virtual memory reached 54.855 GB
- The rg child had already exited when this was measured
- rtk.exe had one thread and continued using CPU while post-processing the output
- The parent process had also exited

I ran the same underlying rg argument shape separately and discarded stdout through a fixed 1 MB buffer.

- Native rg stdout was 4,431,986,055 bytes, or 4.128 GiB
- Native rg peak working set was 1,118.4 MB
- Native rg peak private memory was 1,409.7 MB
- RTK peak private memory was about 12.3 times the raw output size

The large output came from Codex session JSONL files under .codex. The relevant corpus contained about 3,450 JSONL files totaling 14.796 GiB. About 3,494 files and 49,000 long JSONL lines matched.

The current implementation appears to amplify memory in several places.

- exec_capture_stdin uses Command::output, so all child stdout is buffered before filtering
- String::from_utf8_lossy followed by into_owned creates an owned String copy
- search.rs clones the complete stdout into raw_output
- all matches are processed before the 200 result limit is applied
- clean_line lowercases the complete line and may collect the complete line into Vec char, even when only a short result is retained
- a full plain representation is rebuilt before RTK decides whether to use the capped output

Relevant source locations in commit 5a7880d404db8364d602f2ecdc41dd790f64013f

- src/core/stream.rs lines 545 to 552
- src/cmds/system/search.rs lines 397 to 405
- src/cmds/system/search.rs lines 427 to 480
- src/cmds/system/search.rs lines 629 to 670

This seems different from issue 897. That issue involved grep child processes remaining alive. In this case no child remained and rtk.exe itself held 50.829 GB while post-processing one command.

I also checked develop at f9d8c775b1e7f94f449c400f4130410170e590ad. The relevant capture and duplication behavior is still present.

Expected behavior

RTK should keep memory bounded when filtering large search output. The result limit should apply while output is being read, or raw output should be streamed or spooled instead of retained and duplicated in memory.

Limitations

- The exact offending command line was captured, but its original working directory was lost after the parent process exited. The relative src, docs, and issues paths therefore could not be reproduced exactly.
- The .codex session files continue to change, so this is not a byte-for-byte replay of the original input.
- I did not rerun the full command through RTK because it had already reached 50.829 GB and repeating it risked another system OOM. I reproduced the underlying 4.128 GiB native output with bounded streaming instead.
- No heap profiler was attached, so the exact share of the 50.829 GB attributable to each allocation is not known. The source path and overall unbounded buffering behavior are confirmed.

albatrossflyon-coder:
Fixes #3106.

## The bug

`engine_capture()` and `passthrough()` in `search.rs` called `exec_capture_stdin`, which wraps `Command::output()` and buffers the entire child process's stdout/stderr in memory with **no limit** — on a large search corpus this could consume 50+ GB before the 200-result cap in `search.rs` ever gets a chance to apply.

## The fix

rtk already has a memory-safe capture path for exactly this: `run_streaming` with `FilterMode::CaptureOnly` reads line-by-line through the existing `RAW_CAP` (10 MiB), already covered by `test_run_streaming_raw_cap_at_10mb` in `stream.rs`. It just wasn't being used by `grep`/`rg`.

Added `exec_capture_stdin_bounded` as a thin `CaptureResult`-shaped wrapper around `run_streaming`, so both call sites in `search.rs` needed a one-line swap — no changes to `search.rs`'s grouping/capping logic at all. Removed the now-unused `exec_capture_stdin` (the unbounded function this replaces).

## Testing

- Full suite: 2457 passed, 8 ignored (unchanged — the bounded-capture behavior itself is already covered by `test_run_streaming_raw_cap_at_10mb`, so this PR doesn't need new unit tests to prove that mechanism works; it only needed to prove `search.rs` is actually using it).
- Manual verification: `rtk grep` against a real 36MB file where every line matches. Completes cleanly (exit 0) and stderr shows `[rtk] warning: output exceeds 10 MiB — filter input truncated` — proof the cap is genuinely active now (this warning could not have fired through the old unbounded path).

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -7,7 +7,8 @@
 //! `file:line:content` shape.
 
 use crate::core::stream::{
-    self, exec_capture, exec_capture_stdin, CaptureResult, FilterMode, StdinMode, StreamFilter,
+    self, exec_capture, exec_capture_stdin_bounded, CaptureResult, FilterMode, StdinMode,
+    StreamFilter,
 };
 use crate::core::tracking;
 use crate::core::utils::{resolved_command, strip_ansi};
@@ -303,7 +304,7 @@ fn engine_capture<T: AsRef<str>>(
     paths: &[String],
 ) -> Result<CaptureResult> {
     let mut cmd = engine_command(engine, extra_args, patterns, paths, false);
-    exec_capture_stdin(&mut cmd).context("search failed")
+    exec_capture_stdin_bounded(&mut cmd).context("search failed")
 }
 
 fn engine_command<T: AsRef<str>>(
@@ -460,7 +461,7 @@ fn passthrough<T: AsRef<str>>(
             .context("search failed")?
             .exit_code
     } else {
-        let result = exec_capture_stdin(&mut cmd).context("search failed")?;
+        let result = exec_capture_stdin_bounded(&mut cmd).context("search failed")?;
         print!("{}", strip_ansi(&result.stdout));
         if !result.stderr.is_empty() {
             eprint!("{}", result.stderr);
--- a/src/core/stream.rs
+++ b/src/core/stream.rs
@@ -565,10 +565,29 @@ pub fn exec_capture(cmd: &mut Command) -> Result<CaptureResult> {
     capture(cmd)
 }
 
-/// Like [`exec_capture`] but inherits stdin so a wrapped engine can read a piped stdin.
+/// Like [`exec_capture`] but inherits stdin so a wrapped engine can read a
+/// piped stdin. Unbounded (buffers the full child output via [`capture`]) —
+/// fine for small, known-bounded output like `git commit`'s summary line.
+/// For anything that can produce large output, use
+/// [`exec_capture_stdin_bounded`] instead.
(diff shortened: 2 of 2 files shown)


KuSh:
Closing in favour of #3849, which bounds the same unbounded search capture and goes further: it caps per line as well as per stream, reaps the child on a broken pipe, and adds regression tests for both cases. Please take any further discussion of the capture path to that PR.

PR #3135, opened 2026-07-22
fix(git): don't assume '[' is byte 0 in commit-output parser

### LINKED ISSUE #3100: `git commit` output parser still panics when first line starts with a multibyte char and contains `]` (parse_commit_output, v0.43.0)

albatrossflyon-coder:
Fixes #3100.

## The bug

`parse_commit_output` hardcoded `line[1..bracket_end]` to strip the leading `[` from `git commit`'s summary line, assuming `[` is always the first (1-byte) character. When a multibyte character precedes it — e.g. a localized prefix — byte offset 1 lands mid-character and panics: `byte index 1 is not a char boundary; it is inside '初'`.

## The fix

Locate `[` with `line.find('[')` instead of assuming position 0. `[` is ASCII, so `bracket_start + 1` is always a valid boundary regardless of what precedes it. This also means the hash still gets extracted correctly in that case, not just "doesn't panic."

## Testing

- Confirmed the old code's exact panic first, not just asserted a fix works: compiled and ran its logic standalone against the reported input shape — `start byte index 1 is not a char boundary; it is inside '初' (bytes 0..3 of string)`, matching the issue exactly.
- 2 new regression tests: multibyte prefix before a valid bracket (hash still extracted correctly), and multibyte prefix with no bracket at all (safe fallback to `"ok"`).
- Full suite: 2459 passed, 8 ignored (2457 baseline + 2 new tests).

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1009,15 +1009,23 @@ fn build_commit_command(args: &[String], global_args: &[String]) -> Command {
 /// Handles: `[main abc1234def] message`, `[main (root-commit) abc1234def] msg`,
 /// localized variants, and multibyte branch names.
 fn parse_commit_output(line: &str) -> String {
-    if let Some(bracket_end) = line.find(']') {
-        let bracket_content = &line[1..bracket_end];
-        let hash = bracket_content.split_whitespace().next_back().unwrap_or("");
-        if !hash.is_empty() && hash.len() >= 7 {
-            let short_hash: String = hash.chars().take(7).collect();
-            format!("ok {}", short_hash)
-        } else {
-            "ok".to_string()
-        }
+    // Locate '[' by search rather than assuming it's byte 0: a multibyte
+    // character before the bracket (e.g. a localized prefix) would make the
+    // old hardcoded `line[1..]` cut mid-character and panic. `[` is ASCII
+    // (1 byte), so `bracket_start + 1` is always a valid boundary regardless
+    // of what precedes it.
+    let Some(bracket_start) = line.find('[') else {
+        return "ok".to_string();
+    };
+    let after_open = &line[bracket_start + 1..];
+    let Some(bracket_end) = after_open.find(']') else {
+        return "ok".to_string();
+    };
+    let bracket_content = &after_open[..bracket_end];
+    let hash = bracket_content.split_whitespace().next_back().unwrap_or("");
+    if !hash.is_empty() && hash.len() >= 7 {
+        let short_hash: String = hash.chars().take(7).collect();
+        format!("ok {}", short_hash)
     } else {
         "ok".to_string()
     }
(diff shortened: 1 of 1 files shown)


pszymkowiak (approved):
Verified locally on a compiled build. The bug is real (byte inspection: `初め[` = `E5 88 9D E3 82 81 5B…`, so the old `line[1..]` starts on a UTF-8 continuation byte `0x88` → guaranteed panic). Ran the old vs new logic side-by-side on the exact issue input:

```
[main abc1234def] msg       old=ok abc1234   new=ok abc1234
初め[main abc1234def] add    old=PANIC        new=ok abc1234
初めsome output              old=ok           new=ok
[初め-branch abc1234def]     old=ok abc1234   new=ok abc1234
```

The fix eliminates the panic AND extracts the hash correctly in the multibyte-before-bracket case, byte-identical to the old behavior everywhere else (no regression). Also confirmed end-to-end: 9 parse_commit_output unit tests pass (incl. the 3 new multibyte ones), and real `rtk git commit` works on both a normal branch and a multibyte-named branch (`初め-branch` → `ok 12c3e78`). Locating `[` by search + slicing at the ASCII `+1` boundary is exactly right.

LGTM 👍

albatrossflyon-coder:
This looks superseded, closing rather than merging. `f496f59` (2026-08-15, part of the decode/exec_capture refactor addressing #2717) already fixed `parse_commit_output` to locate `[`/`]` by search instead of assuming byte 0, same underlying bug this PR targets. This PR predates that fix (opened 2026-07-22) and now conflicts with it. No action needed on the newer code, it already handles the multibyte case.

PR #3136, opened 2026-07-22
fix(hooks): honor CLAUDE_CONFIG_DIR when loading permission rules

### LINKED ISSUE #3133: rtk init -g should respect CLAUDE_CONFIG_DIR Environment variable
Changing you CLAUDE_CONFIG_DIR is common on setups where multiple accounts are being used (work vs personal). `rtk init -g` should respect working directories for Claude profiles based on the existence or absence of the environment variable instead of defaulting to `~/.claude`

quinnjr:
## Summary

Fixes #3133.

`get_settings_paths()` in `src/hooks/permissions.rs` built the global
`settings.json`/`settings.local.json` paths directly from
`dirs::home_dir()`, bypassing `resolve_claude_dir()` (which already
checks `CLAUDE_CONFIG_DIR` before falling back to `$HOME/.claude`).
Every other Claude Code path resolution in the codebase goes through
`resolve_claude_dir()` — this was the one spot that didn't.

Practical effect: with multiple Claude profiles (e.g. via a tool like
`claude-profile` that switches accounts by setting
`CLAUDE_CONFIG_DIR`), RTK's auto-allow/deny/ask permission checks kept
reading the default `~/.claude/settings.json` instead of the active
profile's settings — so permission decisions didn't match whichever
profile Claude Code itself was actually using.

## Changes

- `get_settings_paths()` now resolves the global settings location via
  `resolve_claude_dir()` instead of hand-rolling `home_dir().join(".claude")`.
- Added `test_get_settings_paths_honors_claude_config_dir` verifying the
  override is picked up and the default `~/.claude` path is excluded
  when `CLAUDE_CONFIG_DIR` is set.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo clippy --all-targets` clean
- [x] `cargo test --all` — 2518 passed, 8 ignored
- [x] Manual check: pointed `CLAUDE_CONFIG_DIR` at a temp dir with its
      own `settings.json` and confirmed rtk's permission rules are now
      read from it

### DIFF
--- a/src/hooks/permissions.rs
+++ b/src/hooks/permissions.rs
@@ -2,6 +2,7 @@ use super::constants::{
     CLAUDE_DIR, CURSOR_DIR, DROID_DIR, DROID_HOME_ENV, DROID_SETTINGS_FILE, GEMINI_DIR,
     SETTINGS_JSON, SETTINGS_LOCAL_JSON,
 };
+use super::init::resolve_claude_dir;
 use crate::core::stream::exec_capture;
 use crate::discover::lexer::split_for_permissions;
 use serde_json::Value;
@@ -174,15 +175,25 @@ fn append_bash_rules(rules_value: Option<&Value>, target: &mut Vec<String>) {
 
 /// Return the ordered list of Claude Code settings file paths to check.
 fn get_settings_paths() -> Vec<PathBuf> {
+    get_settings_paths_from(find_project_root(), resolve_claude_dir().ok())
+}
+
+/// Assemble the settings paths for a project root and a resolved Claude config dir.
+///
+/// `claude_dir` is already resolved, so it honors `CLAUDE_CONFIG_DIR` when set.
+fn get_settings_paths_from(
+    project_root: Option<PathBuf>,
+    claude_dir: Option<PathBuf>,
+) -> Vec<PathBuf> {
     let mut paths = Vec::new();
 
-    if let Some(root) = find_project_root() {
+    if let Some(root) = project_root {
         paths.push(root.join(CLAUDE_DIR).join(SETTINGS_JSON));
         paths.push(root.join(CLAUDE_DIR).join(SETTINGS_LOCAL_JSON));
     }
-    if let Some(home) = dirs::home_dir() {
-        paths.push(home.join(CLAUDE_DIR).join(SETTINGS_JSON));
-        paths.push(home.join(CLAUDE_DIR).join(SETTINGS_LOCAL_JSON));
+    if let Some(claude_dir) = claude_dir {
+        paths.push(claude_dir.join(SETTINGS_JSON));
+        paths.push(claude_dir.join(SETTINGS_LOCAL_JSON));
     }
 
     paths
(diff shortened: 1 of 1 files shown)


pszymkowiak (approved):
Verified on a compiled build, including an old-vs-new contrast with a controlled $HOME to isolate the two code paths:

```
scenario: HOME=<fake, empty ~/.claude>, CLAUDE_CONFIG_DIR=<profileA with deny Bash(git status:*)>
old (develop)  → rtk rewrite "git status" → exit 3   # profileA IGNORED, read ~/.claude (the #3133 bug)
new (this PR)  → rtk rewrite "git status" → exit 2   # profileA honored (deny)
default (no CLAUDE_CONFIG_DIR) → exit 3 on both        # ~/.claude path unchanged
```

So the fix genuinely routes permission-rule loading through the active profile without changing the default `~/.claude` behavior. Also: full suite 2518 passed / 0 failed, 61 permissions tests green (incl. the new `get_settings_paths_honors_claude_config_dir`). Routing this last spot through `resolve_claude_dir()` — the same helper every other Claude-dir lookup uses — is exactly right.

Minor test-hygiene note (non-blocking): the new test mutates `CLAUDE_CONFIG_DIR` via `set_var` under a local `Mutex`; fine as-is, just be aware other env-mutating tests elsewhere don't share that lock.

LGTM 👍

KuSh (approved):
Correct fix: `get_settings_paths()` now routes the global scope through `resolve_claude_dir()`, matching every other Claude-path resolution in the crate. Project scope rightly stays on `root.join(".claude")`, the union semantics of `load_permission_rules()` are unchanged, and the no-home case still yields no global paths.

Pushed one test-only fixup to the branch. The test mutated process-wide `CLAUDE_CONFIG_DIR` under a module-private `CLAUDE_CONFIG_DIR_LOCK`, which didn't serialize against `init.rs`'s own `CLAUDE_DIR_LOCK` guarding the same variable in the same test binary. On the losing interleaving, `with_claude_dir_override`'s restore raced `run_default_mode()`, which re-resolves through `resolve_claude_dir()` — so `cargo test --all` could write `RTK.md` into and patch the developer's real `~/.claude`.

Rather than share the lock, `get_settings_paths()` is split the way `resolve_claude_dir_from` already is:

```rust
fn get_settings_paths() -> Vec<PathBuf> {
    get_settings_paths_from(find_project_root(), resolve_claude_dir().ok())
}
```

Path assembly is now tested as a pure function — no env mutation, no lock to get wrong. That also clears two smaller points: the negative assertion on `~/.claude/settings.json` wasn't a property of the code under test (`find_project_root()` walks up from the CWD, and when it lands on `$HOME` the project path is byte-identical to the default), and the `tempdir()` was never written to. Exact-equality assertions now cover ordering and both scopes, plus a second case for the resolution-failure branch.

One tradeoff worth naming: the wiring itself — that `get_settings_paths()` resolves via `resolve_claude_dir()` rather than `dirs::home_dir()` — is now a single visible expression with no test behind it. Covering it would mean mutating env again, which is what we just removed.

`cargo fmt --all`, `cargo clippy --all-targets` and `cargo test --all` (2461 tests) are green on the amended commit.

PR #3139, opened 2026-07-22
fix(sbt): filter selective test tasks (testOnly, testQuick)

randomBrainstormer:
## What

Follow-up to #752. `sbt testOnly <spec>` and `sbt testQuick` — the commands you run to execute a single test or the just-changed tests, the tightest inner loop for a Scala dev — were passed through **unfiltered**, even though they emit the exact same ScalaTest/munit output that `filter_sbt_test` already handles.

Reported by @greenhost87 on #752:
> Minor problem - single test execution via `sbt testOnly` falls into `rtk sbt <other>`

## Why it slipped through

Two layers, both missing the task:

1. **Rewrite rule** (`src/discover/rules.rs`): the subcommand alternation listed `test` but not `testOnly`/`testQuick`. With the `(?:\s|$)` anchor, `test` cannot match `testOnly`, so `sbt testOnly …` was never rewritten to `rtk sbt …`.
2. **Routing** (`src/cmds/scala/sbt_cmd.rs`): even as `rtk sbt testOnly …`, `run_other` only recognized integration-test tasks (`:test` / `/test` suffixes) and fell through to `run_passthrough`.

## Change

- Add `testOnly`/`testQuick` to the rewrite alternation, **keeping the `(?:\s|$)` anchor** so custom tasks like `testify` are still not rewritten.
- New `is_test_cmd()` helper recognizes `testOnly`/`testQuick` and their scoped forms (`Test/testOnly`, `it:testQuick`, …) in addition to the existing integration-test tasks.
- `run_other` routes all test-producing tasks through `runner::run_filtered` with `filter_sbt_test`, so the `never_worse` cap and tee hint apply (Transparency) — same shared-runner shape as the rest of the module.

No new output format; the filter itself is unchanged and already covered by fixtures.

## Tests

- `test_is_test_cmd` — bare + scoped `testOnly`/`testQuick`, integration tasks, and negatives (`compile`, `assembly`, `update`, `testable`).
- `test_rewrite_sbt_test_only` / `test_rewrite_sbt_does_not_match_unrelated_tasks`.

`cargo fmt --all --check` ✅ · `cargo clippy --all-targets` (no warnings) ✅ · `cargo test --all` (2520 passed) ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ rtk ruff check                  # Python linting (JSON, -80%)
 rtk golangci-lint run           # Go linting (JSON, -85%)
 rtk rubocop                     # Ruby linting (JSON, -60%+)
 rtk sbt test                    # ScalaTest output (-90%)
+rtk sbt testOnly com.example.X  # Single/selective tests, same filter (-90%)
 rtk sbt compile                 # Compilation errors only (-75%)
 rtk sbt run                     # Strip SBT preamble noise
 ```
--- a/src/cmds/scala/sbt_cmd.rs
+++ b/src/cmds/scala/sbt_cmd.rs
@@ -60,6 +60,21 @@ fn is_integration_test_cmd(subcommand: &str) -> bool {
     ) || (subcommand.ends_with(":test") || subcommand.ends_with("/test"))
 }
 
+/// Selective/single-test task names that emit ScalaTest/munit output just like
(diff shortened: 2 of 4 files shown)


randomBrainstormer:
closing in favor of https://github.com/rtk-ai/rtk/pull/3140

PR #3140, opened 2026-07-22
fix(sbt): filter testOnly/testQuick like sbt test

aeppling:
Follow-up to #752, reported by @greenhost87: `sbt testOnly` fell through to passthrough (0% savings) and the rewrite rule never matched it.

Routes `testOnly`/`testQuick` (bare, scoped, and quoted forms like `sbt "testOnly com.example.MySpec"`) through the existing test filter, and extends the rewrite rule to match them.

### DIFF
--- a/src/cmds/scala/sbt_cmd.rs
+++ b/src/cmds/scala/sbt_cmd.rs
@@ -63,7 +63,7 @@ fn is_integration_test_cmd(subcommand: &str) -> bool {
 fn is_test_task(subcommand: &str) -> bool {
     let task = subcommand.split_whitespace().next().unwrap_or(subcommand);
     let task = task.rsplit(['/', ':']).next().unwrap_or(task);
-    matches!(task, "testOnly" | "testQuick" | "test-only" | "test-quick")
+    matches!(task, "testOnly" | "testQuick")
 }
 
 /// Returns true if `s` is a scoped SBT task (e.g. `Test/test`, `it/Test/compile`).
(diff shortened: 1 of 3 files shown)


max-leuthaeuser:
Did you also test these?
```
sbt "testOnly *MySuite -- -n Tag" // select tests with a specific tag
sbt "testOnly *MySuite -- -z foo" // run only the tests whose name includes the substring "foo".
sbt "testOnly *MySuite -- -t foo" // run only the test whose name matches "foo" exactly.
```

randomBrainstormer:
Nice one @aeppling — I'd taken a run at this too (#3139, now closed) and you got there with a cleaner version. Really glad you handled the quoted form; `sbt "testOnly ..."` is how most people actually pass a spec, so it's much better having that covered here than left as a follow-up.

The one thing mine had that this doesn't is a small README touch, so `testOnly` shows up next to `sbt test` in the supported-commands list:

```
rtk sbt test                    # ScalaTest output (-90%)
rtk sbt testOnly com.example.X  # Single/selective tests, same filter (-90%)
rtk sbt compile                 # Compilation errors only (-75%)
```

No worries if you'd rather keep this PR focused, I'm happy to send it as a tiny docs PR once this lands.

aeppling:
@max-leuthaeuser yes verified and test added
@randomBrainstormer you can create a pr, not a blocker this is not really user facing docs since agent run this seamless

pszymkowiak (approved):

PR #3147, opened 2026-07-22
fix(test): accept both Ask and Allow verdicts in rewrite tests

### LINKED ISSUE #3146: Rewrite tests fail when local permission settings allow git commands
## Summary

Two tests in `hooks::rewrite_cmd::tests::unattestable_passthrough` fail on machines that have `Bash(git *)` (or similar) in their `.claude/settings.local.json` allow rules:

- `test_plain_command_still_rewrites`
- `test_fd_dup_redirect_still_rewrites`

## Root Cause

Both tests call `evaluate()` which invokes `check_command()` → `load_permission_rules()`, reading real permission files from `~/.claude/settings.json` and `.claude/settings.local.json`. When those files contain allow rules matching git commands, `check_command("git status")` returns `Allow` instead of `Default`, making `evaluate()` return `Allow(...)` instead of `Ask(...)`.

The tests assert exactly `RewriteOutcome::Ask(_)`, but the actual intent is "command gets rewritten" (i.e., not `Passthrough` or `Deny`).

## Reproduction

```bash
# 1. Add git allow rule to local settings
# In .claude/settings.local.json, add: "Bash(git *)" to permissions.allow

# 2. Run the tests
cargo test "still_rewrites"

# Result: 2 failures
# hooks::rewrite_cmd::tests::unattestable_passthrough::test_fd_dup_redirect_still_rewrites
# hooks::rewrite_cmd::tests::unattestable_passthrough::test_plain_command_still_rewrites
```

Tests pass on clean machines (CI) and when `settings.local.json` is absent.

## Fix Direction

Accept both `Ask` and `Allow` as valid rewrite outcomes, since both indicate successful rewriting (the tests verify the command is NOT treated as Passthrough):

```rust
let result = evaluate("git status", &[], &[]);
assert!(
    matches!(result, RewriteOutcome::Ask(_) | RewriteOutcome::Allow(_)),
    "plain command should still be rewritten, got {:?}", result
);
```

guyoron1:
Heyaaa! : )

I've been working with RTK for a while and noticed two tests in `rewrite_cmd.rs` kept failing on my machine but passed on fresh checkouts. Tracked it down to my local `.claude/settings.local.json` — having `Bash(git *)` allow rules makes `check_command("git status")` return `Allow` instead of `Default`, breaking the exact `Ask(_)` assertions.

Fix: accept both `Ask(...)` and `Allow(...)` since the real intent is "command was rewritten, not Passthrough."

Closes #3146

## Test plan
- [x] Tests pass WITH `settings.local.json` containing `Bash(git *)` allow rules
- [x] Tests pass WITHOUT `settings.local.json` (clean machine)
- [x] Full test suite passes (`cargo test --all`)

### DIFF
--- a/src/hooks/rewrite_cmd.rs
+++ b/src/hooks/rewrite_cmd.rs
@@ -104,25 +104,46 @@ mod tests {
         );
     }
 
+    /// The verdict still drives the outcome: an allow rule yields `Allow`.
+    /// Pinning both directions keeps the mapping covered without depending
+    /// on which rules the developer happens to have configured.
+    #[test]
+    fn test_allow_verdict_yields_allow() {
+        assert!(matches!(
+            evaluate_with_verdict("git status", PermissionVerdict::Allow, &[], &[]),
+            RewriteOutcome::Allow(_)
+        ));
+    }
+
+    #[test]
+    fn test_deny_verdict_yields_deny() {
+        assert_eq!(
+            evaluate_with_verdict("git status", PermissionVerdict::Deny, &[], &[]),
+            RewriteOutcome::Deny
+        );
+    }
+
+    /// Commands with an unattestable construct are always a passthrough,
+    /// regardless of permission verdict.
+    ///
+    /// The verdict is pinned to `Default` rather than going through `evaluate`,
+    /// which reads the developer's own `.claude/settings.local.json`: a
+    /// `Bash(git *)` allow rule turns the expected `Ask` into `Allow` and these
+    /// tests fail on that machine only (#3146). Pinning the verdict keeps the
+    /// assertions about the rewrite logic and nothing about the host.
     mod unattestable_passthrough {
         use super::super::{evaluate_with_verdict, RewriteOutcome};
         use crate::hooks::permissions::PermissionVerdict;
 
-        /// Evaluate with the verdict a machine with no permission rules produces.
-        ///
-        /// Calling `evaluate` here instead would read the developer's own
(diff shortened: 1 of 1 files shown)


KuSh:
Hi @guyoron1, and thanks for the PR. Unfortunately this isn’t the right fix. Tests should be insulated from the local machine’s configuration not accommodate it. That could be a great contribution if you'd like to work on it (at least those that affect you at first)

guyoron1:
You're right — accommodating the local config was the wrong direction. Reworked to insulate instead.

`evaluate()` called `check_command()`, which reads the machine's settings files, so the test outcome tracked whatever the developer happened to have configured. It now has the same verdict-injection seam `permissions.rs` already exposes via `check_command_with_rules`: `evaluate_with_verdict()` holds the decision logic and takes the verdict as a parameter, `evaluate()` stays the thin wrapper that looks it up. The tests pin `PermissionVerdict::Default` and read no settings at all, so the strict `Ask(_)` assertions are back.

Worth noting the four `Passthrough` tests in that module had the same latent problem in the other direction — a `Bash(git *)` **deny** rule would have turned them into `Deny`. They're pinned now too, and I added Allow/Deny cases so the verdict mapping stays covered in both directions rather than losing coverage.

Verified by running the test binary against HOME directories containing deny, ask, allow, and no rules — 8/8 in all four, plus with this repo's own `settings.local.json` in place. Full suite: 2662 passed, fmt/clippy clean.

KuSh (approved):
LGTM thanks!

PR #3149, opened 2026-07-22
fix(hooks): emit permissionDecision allow for simple Copilot CLI rewrites

### LINKED ISSUE #3037: rtk hook copilot emits permissionDecision "ask", forcing a permission prompt on every command in Copilot CLI >= 1.0.66

guyoron1:
Heyaaa! : )

I've been testing RTK with Copilot CLI and since version 1.0.66, every rewrite triggers a permission prompt — even safe stuff like `cargo test` → `rtk cargo test`. Copilot changed how it handles missing `permissionDecision`: now treats it as "always ask." Since `AskRewrite` omitted the field, every rewrite became manual approval.

Fix: emit `permissionDecision: "allow"` for simple single-segment commands. Compound commands (newlines, `&`, `;`, pipes) still omit it so Copilot shows its own prompt. Uses existing `split_for_permissions()` — all CVE bypass security tests still pass.

Closes #3037

## Test plan
- [x] All 17 Copilot CLI tests pass (including security CVE tests)
- [x] Simple commands (`cargo test`) now get `permissionDecision: "allow"`
- [x] Compound commands (`git status\nrm -rf /`) still omit `permissionDecision`
- [x] Full test suite passes (`cargo test --all`)

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -127,7 +127,7 @@ fn get_rewritten(cmd: &str) -> Option<String> {
 
 enum HookDecision {
     AllowRewrite(String),
-    AskRewrite(String),
+    AskRewrite { rewritten: String, explicit: bool },
     Defer,
     Deny,
 }
@@ -141,7 +141,10 @@ fn decide_from_verdict(cmd: &str, verdict: PermissionVerdict) -> HookDecision {
     }
     match get_rewritten(cmd) {
         Some(r) if verdict == PermissionVerdict::Allow => HookDecision::AllowRewrite(r),
-        Some(r) => HookDecision::AskRewrite(r),
+        Some(r) => HookDecision::AskRewrite {
+            rewritten: r,
+            explicit: verdict == PermissionVerdict::Ask,
+        },
         None => HookDecision::Defer,
     }
 }
@@ -158,7 +161,7 @@ fn handle_vscode(cmd: &str) -> Result<()> {
         }
         HookDecision::Defer => return Ok(()),
         HookDecision::AllowRewrite(r) => ("allow", r),
-        HookDecision::AskRewrite(r) => ("ask", r),
+        HookDecision::AskRewrite { rewritten: r, .. } => ("ask", r),
     };
 
     audit_log("rewrite", cmd, &rewritten);
@@ -202,9 +205,12 @@ fn copilot_cli_response_from_decision(
         }
         HookDecision::Defer => return None,
         HookDecision::AllowRewrite(r) => (r, true),
-        HookDecision::AskRewrite(r) => {
+        HookDecision::AskRewrite {
+            rewritten: r,
+            explicit,
+        } => {
             let is_simple = crate::discover::lexer::split_for_permissions(cmd).len() <= 1;
-            (r, is_simple)
+            (r, !explicit && is_simple)
         }
     };
 
@@ -261,7 +267,7 @@ pub fn run_gemini() -> Result<()> {
             audit_log("rewrite", cmd, rewritten);
             print_gemini("allow", Some(rewritten));
         }
-        HookDecision::AskRewrite(ref rewritten) => {
+        HookDecision::AskRewrite { ref rewritten, .. } => {
             audit_log("ask", cmd, rewritten);
             print_gemini("ask_user", Some(rewritten));
         }
@@ -366,7 +372,7 @@ fn process_claude_payload(v: &Value) -> PayloadAction {
             }
         }
         HookDecision::AllowRewrite(r) => (r, true),
-        HookDecision::AskRewrite(r) => (r, false),
+        HookDecision::AskRewrite { rewritten: r, .. } => (r, false),
     };
 
     let updated_input = {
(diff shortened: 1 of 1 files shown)


pszymkowiak (approved):
Verified on a compiled build, both unit and end-to-end. The security boundary (`split_for_permissions(cmd).len() <= 1`) holds — I drove the real `rtk hook copilot` with Copilot-CLI-shaped JSON on stdin:

```
cargo test / git status          → {"modifiedArgs":{"command":"rtk cargo test"},"permissionDecision":"allow"}
git status && git push           → {"modifiedArgs":{...}}                 (no permissionDecision → Copilot prompts)
cargo test; rm -rf x             → {"modifiedArgs":{...}}                 (no allow — the rm -rf isn't auto-approved) ✓
cargo test $(rm -rf /tmp/x)      → (no response → defer)
cargo test && echo hi > /etc/passwd → (no response → defer)
```

So only simple, single-segment rewrites get `allow`; anything compound or dangerous keeps Copilot's own prompt, and substitution/redirect forms defer entirely. Also confirmed `decide_from_verdict` handles `Deny` before `AskRewrite`, so this never auto-allows a denied command. 44 copilot/hook tests pass, including the CVE ones (`newline_bypass`/`background_bypass` never auto-allow).

One semantic note (non-blocking): `AskRewrite` maps to verdict `Ask`, which covers both the default (no rule) and an explicit user `ask` rule — so a simple rewritable command a user explicitly marked "ask" would now auto-allow under Copilot. Given this is scoped to Copilot CLI's 1.0.66 behavior, rtk only rewrites a curated command set, and the verdict doesn't expose default-vs-explicit, it's a reasonable trade-off — maybe worth a line in the docs.

LGTM 👍

KuSh (approved):
LGTM thanks

src/hooks/hook_cmd.rs
```diff
@@ -202,7 +202,10 @@ fn copilot_cli_response_from_decision(
         }
         HookDecision::Defer => return None,
         HookDecision::AllowRewrite(r) => (r, true),
-        HookDecision::AskRewrite(r) => (r, false),
+        HookDecision::AskRewrite(r) => {
+            let is_simple = crate::discover::lexer::split_for_permissions(cmd).len() <= 1;
+            (r, is_simple)
+        }
```
KuSh: As noted by Patrick and as the fix is simple, change `AskRewrite(String)` to `AskRewrite { rewritten: String, explicit: bool }` in the `HookDecision` enum, setting `explicit = verdict == PermissionVerdict::Ask` in `decide_from_verdict`. Update the match arms accordingly and here return `(r, !explicit && is_simple)`.

KuSh (changes requested):

KuSh (approved):
LGTM!

PR #2378, opened 2026-06-10
fix(rewrite): route direct biome invocations to rtk biome instead of rtk lint

feliperugai:
## Summary

- `biome check x` / `npx biome check x` were rewritten to `rtk lint check x` — the **ESLint adapter**. The rewrite strips the `biome` token, so `rtk lint` falls back to ESLint; in Biome-only projects the subprocess dies in ~0.2s and rtk prints a misleading `[warn] Linter process terminated abnormally (possibly out of memory)`.
- Direct invocations (`biome`, `npx biome`, `pnpm exec biome`, `pnpm dlx biome`, `npm exec biome`, `npm x biome`, `pnpx biome`) now rewrite to `rtk biome`, which runs the Biome CLI through the built-in `src/filters/biome.toml` filter.
- Script-runner forms (`npm run biome`, `pnpm run biome`) fall through to the generic npm/pnpm rules (`rtk npm run biome` / `rtk pnpm run biome`), preserving package.json script semantics — same philosophy as #1491.

Refs #1489 (same root cause, `npm run lint` symptom). Complements #1491, which routes `npm`/`npx` lint scripts but explicitly left direct `biome` invocations out of scope — this PR covers that gap.

### Repro (before)

```
$ rtk rewrite "npx biome check src/x.ts"
rtk lint check src/x.ts        # ESLint adapter — wrong tool
```

### After

```
$ rtk rewrite "npx biome check src/x.ts"
rtk biome check src/x.ts       # Biome TOML filter
$ rtk rewrite "npm run biome"
rtk npm run biome              # script semantics preserved
$ rtk rewrite "npx eslint src/"
rtk lint src/                  # eslint/lint routing unchanged
```

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — 2154 passed
- [x] Manual testing: `rtk <command>` output inspected
  - `cargo run -- rewrite "npx biome check src/x.ts"` → `rtk biome check src/x.ts`
  - `cargo run -- rewrite "biome check --write ."` → `rtk biome check --write .`
  - `cargo run -- rewrite "npm run biome"` → `rtk npm run biome`
  - `cargo run -- rewrite "npx eslint src/"` → `rtk lint src/` (unchanged)
  - E2E on a real Biome project: `rtk biome check src/.../review.service.ts` → `biome: ok`
[...]

PR #3150, opened 2026-07-22
fix(lint): add first-class Biome support to rtk lint

### LINKED ISSUE #3117: rtk hook claude rewrites Biome commands to ESLint and can mask execution failures
### Environment

- RTK: `rtk 0.38.0` (Windows binary, `C:\Users\<user>\AppData\Local\rtk\rtk`)
- OS: Windows 11 Home Single Language, build 10.0.26200
- Shell used for repro: Git Bash (MSYS2), `GNU bash, version 5.2.37(1)-release (x86_64-pc-msys)`
- Package manager: pnpm 11.0.8
- Node: v22.21.0
- Project: Next.js repo using Biome (via Ultracite) exclusively — no ESLint installed
  (`node_modules/.bin/eslint` absent, no eslint config file), triggered via the
  Claude Code `PreToolUse` hook (`rtk hook claude`, configured in
  `~/.claude/settings.json` as `{"matcher": "Bash", "hooks": [{"type": "command",
  "command": "rtk hook claude"}]}`).

This report covers two distinct, separable bugs found while investigating the same
symptom. Both were reproduced directly against the `rtk` binary (not just through the
Claude Code hook), using the JSON stdin format the hook itself sends.

---

### Bug 1 — the auto-rewrite matcher substitutes an ESLint-only filter for Biome invocations

`rtk hook claude` decides to rewrite a Bash command by scanning it for known keywords.
For every pattern actually tested below, a command containing `biome` (as its own
token, or embedded in `pnpm exec biome ...` / `npx biome ...`), or the package-script
name `lint` in `pnpm run lint`, gets rewritten to `rtk lint` — a subcommand documented
as "ESLint with grouped rule violations". There is no Biome-specific subcommand
(`rtk biome --help` reports `Failed to resolve 'biome' via PATH, falling back to
direct exec: Binary 'biome' not found on PATH`), and no detection of which linter the
target project actually has configured before choosing the ESLint filter.

Patterns confirmed to trigger this rewrite (piping the hook's own JSON stdin format
directly into `rtk hook claude`):

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"biome check app/api/user/sessions/route.ts"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk lint check app/api/user/sessions/route.ts"}}}

$ echo '{"tool_name":"Bash","tool_input":{"command":"pnpm exec biome check app/api/user/sessions/route.ts"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk lint check app/api/user/sessions/route.ts"}}}

$ echo '{"tool_name":"Bash","tool_input":{"command":"npx biome check app/api/user/sessions/route.ts"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk lint check app/api/user/sessions/route.ts"}}}

$ echo '{"tool_name":"Bash","tool_input":{"command":"pnpm run lint"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk lint"}}}
```

(In this project, `pnpm run lint` is defined in `package.json` as
`ultracite check --config-path biome.jsonc` — a Biome wrapper, not ESLint.)

Patterns confirmed NOT to trigger the rewrite (untested beyond these — not a claim
about the general matching rule):

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"node_modules/.bin/biome check app/api/user/sessions/route.ts"}}' | rtk hook claude
(no output — command passed through unmodified)

$ echo '{"tool_name":"Bash","tool_input":{"command":"rtk proxy pnpm exec biome check app/api/user/sessions/route.ts"}}' | rtk hook claude
(no output — command passed through unmodified)

$ echo '{"tool_name":"Bash","tool_input":{"command":"pnpm exec ultracite check --config-path biome.jsonc"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk pnpm exec ultracite check --config-path biome.jsonc"}}}
```
(this last one is a generic passthrough-shaped rewrite, not the broken ESLint
substitution — `ultracite` as a literal token is not treated as a `lint`/`biome`
keyword.)

**Consequence:** for the four triggering patterns above, the real Biome/Ultracite
check is never executed. RTK substitutes its own ESLint invocation instead, in a
project that has no ESLint installed.

---

### Bug 2 — when the substituted `rtk lint` call fails, the reported result is
inconsistent and can be a false pass (including exit 0)

Running `rtk lint` directly (bypassing the hook, to isolate this from Bug 1) shows
what happens once the substitution from Bug 1 has occurred:

```
$ rtk lint app/api/user/sessions/route.ts -vvv
ESLint output (JSON parse failed: expected value at line 1 column 1)
undefined
[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command "eslint" not found
$ echo $?
1
```

This confirms `rtk lint` shells out to `eslint` via pnpm, that ESLint is not
resolvable in this project, and that RTK's ESLint-JSON output parser fails to parse
the resulting plaintext pnpm error (`JSON parse failed: expected value at line 1
column 1`).

The two argument shapes actually produced by the Bug-1 rewrites behave differently
after that parse failure:

```
$ rtk lint check app/api/user/sessions/route.ts
Lint: No issues found
$ echo $?
1
```
(what `biome check <file>` / `pnpm exec biome check <file>` / `npx biome check
<file>` become — false "No issues found" message, but the exit code does correctly
propagate as failure)

```
$ rtk lint
ESLint output (JSON parse failed: expected value at line 1 column 1)
undefined
[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command "eslint" not found
$ echo $?
0
```
(what bare `pnpm run lint` becomes — the raw failure text is shown, but the process
exits 0: a caller that checks only the exit code is told the check passed)

**Consequence:** the same underlying failure (ESLint not found → JSON parse failure)
produces two different outcomes depending on how the command reached `rtk lint`, and
in the `pnpm run lint` case the exit code itself is wrong, not just the message.

---

### Related but separate: `exclude_commands` doesn't reliably opt a command out of
Bug 1's rewrite

While investigating a local workaround, adding `"biome"` to `[hooks].exclude_commands`
in `config.toml` was found to stop the rewrite only when `biome` is the first token of
the command:

```
$ rtk config | grep -A1 biome
    "biome",
]
$ echo '{"tool_name":"Bash","tool_input":{"command":"biome check app/api/user/sessions/route.ts"}}' | rtk hook claude
(no output — correctly not rewritten)

$ echo '{"tool_name":"Bash","tool_input":{"command":"pnpm exec biome check app/api/user/sessions/route.ts"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk lint check app/api/user/sessions/route.ts"}}}
(still rewritten — exclude_commands did not apply)
```

The rewrite matcher (Bug 1) scans the whole command string, but `exclude_commands`
appears to check only argv[0]. This means `exclude_commands` cannot be used to
reliably protect a tool that's normally invoked via `pnpm exec`/`npx`/a package
script, which is the common case for `biome`. (This config change was reverted after
testing — it offered only partial protection and is a global, per-developer setting
that would have affected other, ESLint-using projects.)

---

### Suggested fixes

1. Before routing a `lint`/`biome`-keyword command to the ESLint filter, detect what
   the target project actually has configured (presence of `biome.json(c)` vs an
   ESLint config / `eslint` resolvable in `node_modules/.bin`), or add a first-class
   Biome filter (`rtk biome`, analogous to `rtk prettier`/`rtk ruff`) and route to it
   instead when Biome is what's present.
2. When the underlying linter process fails to spawn/resolve (as opposed to running
   and reporting zero violations), that should be surfaced as an execution error —
   non-zero exit, clear "underlying linter not found/failed to run" message — every
   time, regardless of which argument shape reached the parser. It should never look
   the same as a clean pass.
3. Make `exclude_commands` matching consistent with the rewrite matcher (both
   argv[0]-only, or both whole-string) so it can reliably opt out `pnpm exec <tool>`,
   `npx <tool>`, and `pnpm run <script-that-wraps-tool>` forms, not just the bare
   `<tool> ...` form.

This is reported as a correctness/reliability bug — the tool changes what actually
runs and what result is reported, silently, without exposing this in its own output.

guyoron1:
Heyaaa! : )

I've been digging around the hook rewrite system and found Biome users getting a rough experience. `biome check src/` correctly rewrites to `rtk lint biome check src/` and `detect_linter()` identifies "biome" — but then it falls through to implicit `_ => {}` match arms, so output goes through the generic filter that just scans for "error"/"warning" words. No ESLint flags got injected (correct by accident), but filtering was basically throwing away Biome's structured diagnostics.

Fix: added a proper `"biome"` arm and dedicated `filter_biome_output()` that parses Biome's diagnostic headers, groups by rule/file, with 60%+ token savings.

Fixes #3117

## Test plan
- [x] `cargo test lint_cmd` — all 20 tests pass (4 new biome tests)
- [x] `cargo test --all` — 2465/2467 pass (2 pre-existing env-sensitivity failures, not related)
- [x] Token savings ≥60% verified with realistic Biome output fixture

### DIFF
--- a/src/cmds/js/lint_cmd.rs
+++ b/src/cmds/js/lint_cmd.rs
@@ -119,6 +119,10 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         "mypy" => {
             // mypy uses default text output (no special flags)
         }
+        "biome" => {
+            // Biome uses its own subcommands (check, lint, format) and
+            // native diagnostics — no format flags needed.
+        }
         _ => {
             // Other linters: no special formatting
         }
@@ -135,6 +139,8 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
             1
         }
     } else {
+        // For biome and other explicit linters: skip the linter name (index 0),
+        // but preserve subcommands (check, lint, format, etc.)
         1
     };
 
@@ -196,6 +202,7 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         }
         "pylint" => filter_pylint_json(&result.stdout),
         "mypy" => mypy_cmd::filter_mypy_output(&raw),
+        "biome" => filter_biome_output(&raw),
         _ => filter_generic_lint(&raw),
     };
 
@@ -436,6 +443,110 @@ fn filter_pylint_json(output: &str) -> String {
             crate::core::tee::force_tee_tail_hint(&all_file_lines, "pylint-files", MAX_FILES + 1)
         {
             result.push_str(&format!("  {}\n", hint));
+        }
+    }
+
+    result.trim().to_string()
+}
+
+/// Filter Biome diagnostic output — groups violations by file and rule.
+///
+/// Biome outputs diagnostics in a structured text format with file paths,
+/// rule names, and suggestions. This filter extracts diagnostics and groups
+/// them by rule for a compact summary.
+fn filter_biome_output(output: &str) -> String {
+    use lazy_static::lazy_static;
+    use regex::Regex;
+
+    lazy_static! {
+        // Matches diagnostic headers like: "path/to/file.ts:10:5 lint/suspicious/noDoubleEquals"
+        // or "path/to/file.ts lint/style/useConst"
+        static ref DIAG_RE: Regex =
+            Regex::new(r"(?m)^[^\s].*?\s+(lint/\S+|a11y/\S+|nursery/\S+|style/\S+|suspicious/\S+|correctness/\S+|complexity/\S+|performance/\S+|security/\S+)")
+                .expect("biome diagnostic regex");
+
+        // Matches "Found N errors" / "Found N warnings" summary lines
+        static ref SUMMARY_RE: Regex =
+            Regex::new(r"(?i)Found\s+(\d+)\s+(error|warning|diagnostic)")
+                .expect("biome summary regex");
+
+        // Matches file paths at the start of diagnostic blocks
+        static ref FILE_RE: Regex =
+            Regex::new(r"(?m)^(\S+\.\w{1,6})(?::\d+(?::\d+)?)?")
+                .expect("biome file regex");
+    }
+
+    // Extract rule names
+    let mut by_rule: HashMap<String, usize> = HashMap::new();
+    for cap in DIAG_RE.captures_iter(output) {
+        let rule = cap[1].to_string();
+        *by_rule.entry(rule).or_insert(0) += 1;
+    }
+
+    // Extract unique file paths
+    let mut files: std::collections::HashSet<String> = std::collections::HashSet::new();
+    for cap in FILE_RE.captures_iter(output) {
+        let path = cap[1].to_string();
+        if path.contains('/') || path.contains('.') {
+            files.insert(path);
+        }
+    }
+
+    // Try to parse summary counts from biome's own summary
+    let mut total_errors = 0usize;
+    let mut total_warnings = 0usize;
+    for cap in SUMMARY_RE.captures_iter(output) {
+        let count: usize = cap[1].parse().unwrap_or(0);
+        let kind = cap[2].to_lowercase();
+        if kind.starts_with("error") {
+            total_errors += count;
+        } else {
+            total_warnings += count;
+        }
+    }
+
+    // If we couldn't parse structured data, fall back to generic filter
+    if by_rule.is_empty() {
+        return filter_generic_lint(output);
+    }
+
+    // If biome didn't emit summary counts, use rule match count
+    if total_errors == 0 && total_warnings == 0 {
+        total_errors = by_rule.values().sum();
+    }
+
+    let mut result = String::new();
+    result.push_str(&format!(
+        "Biome: {} issues in {} files\n",
+        total_errors + total_warnings,
+        files.len()
+    ));
+
+    // Show top rules
+    let mut rule_counts: Vec<_> = by_rule.iter().collect();
+    rule_counts.sort_by(|a, b| b.1.cmp(a.1));
+
+    result.push_str("Rules:\n");
+    for (rule, count) in rule_counts.iter().take(10) {
+        result.push_str(&format!("  {} ({}x)\n", rule, count));
+    }
+
+    if rule_counts.len() > 10 {
+        result.push_str(&format!("  … +{} more rules\n", rule_counts.len() - 10));
+    }
+
+    // Show top files
+    if !files.is_empty() {
+        result.push('\n');
+        result.push_str("Files:\n");
+        let mut file_list: Vec<_> = files.iter().collect();
+        file_list.sort();
+        const MAX_FILES: usize = CAP_WARNINGS;
+        for f in file_list.iter().take(MAX_FILES) {
+            result.push_str(&format!("  {}\n", compact_path(f)));
+        }
+        if file_list.len() > MAX_FILES {
+            result.push_str(&format!("  … +{} more files\n", file_list.len() - MAX_FILES));
         }
     }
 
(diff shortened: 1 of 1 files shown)


pszymkowiak:
The filter itself is genuinely nice — I tested it against real Biome (v2.5.5) output and `rtk lint biome check src/` compresses a 8620-byte report down to 242 bytes (~97%), grouping cleanly by rule and file:

```
Biome: 12 issues in 2 files
Rules:
  lint/suspicious/noDoubleEquals (4x)
  lint/style/useConst (4x)
  lint/correctness/noUnusedVariables (2x)
Files: App.ts, utils.ts
```

But there's a gap that stops it from working through the hook, which is the normal path. The rewrite drops the `biome` name (it's treated as a strippable prefix in `rules.rs`, same as `eslint`):

```
biome check src/  → rtk lint check src/     # "biome" gone
eslint src/       → rtk lint src/           # fine — eslint is rtk lint's default
```

The difference is fatal: eslint is the default, biome isn't. So the command the hook actually produces, `rtk lint check src/`, fails at runtime — `detect_linter(["check", "src/"])` no longer sees `biome` and tries to run `check` as the linter:

```
rtk: Failed to run check. Is it installed?
```

So `filter_biome_output` only runs if someone manually types `rtk lint biome check …` (keeping the name) — never via the hook. The PR description says `biome check src/` rewrites to `rtk lint biome check src/`, but on current develop it rewrites to `rtk lint check src/` (verified on a compiled build).

To make this work end-to-end, the routing needs to preserve the linter name for biome (delegate to `rtk lint biome check …`) rather than stripping it — that's in `rules.rs`/the lint rewrite path, outside this PR's current diff. An end-to-end test (`biome check` → rewrite → biome filter applied) would catch it. The filter is solid; it just needs the routing half to land with it. Happy to help wire that up.

guyoron1:
@pszymkowiak thanks for catching the routing gap, that was the missing half. It's closed in this push:

- `rewrite_segment_inner` now keeps the tool name when the matched `rtk lint` prefix is `biome`, so `biome check src/` becomes `rtk lint biome check src/` (also `npx biome ...`, `pnpm biome ...`, `pnpm exec biome ...`, `npm run biome ...`). eslint is still stripped since it's the default. It's 8 added lines reusing `tool_portion`.
- `tests/biome_hook_roundtrip_test.rs` is the end-to-end test you suggested: rewrite, then run exactly that command against a stand-in `biome` printing a real 1110-byte report, asserting the grouped summary, more than 2x compression and that biome's exit code 1 survives. It fails on develop with your exact finding.
- The red clippy job was mine: `lazy_static` isn't a dependency. The three regexes are `LazyLock` statics now.
- Rebased on current develop, and the description no longer claims a rewrite that didn't happen.

One question, since you offered to help wire this up: #2378 routes direct biome invocations to a separate `rtk biome` command instead of `rtk lint biome`. Which destination do you want? This PR follows what `rtk lint` already does for the other linters.

Also, this and #2860 both edit the `test_rewrite_lint` list in `registry.rs`; the conflict is confined to that one test.

PR #3164, opened 2026-07-22
fix(discover): count hook-rewritten commands as coverage, not misses

### LINKED ISSUE #3148: rtk discover counts hook-rewritten commands as misses
# `rtk discover` counts hook-rewritten commands as misses

**Version:** rtk 0.42.3 · macOS 14 (Darwin 24.6.0) x86_64 · Claude Code

## Summary

When the PreToolUse hook (`rtk hook claude`) is installed, `rtk discover` reports
a large undercount of RTK usage — and lists commands the hook *successfully
rewrote* under **MISSED SAVINGS**, as if they had run raw.

Measured over 359 transcripts / 30 days on my machine: discover reports **25.3%**
"already using RTK", actual effective coverage is **44.8%**. A 19.5-point
undercount, and the accompanying "~24.6K tokens saveable" is mostly savings
already banked.

## Cause

`discover` reads the Claude Code transcripts under `~/.claude/projects/*/*.jsonl`,
which record the command **as the model emitted it** — before the PreToolUse hook
rewrote it. The rewrite is therefore invisible to `discover`, and the original
(raw) form is what gets classified.

## Reproduction

1. Install the PreToolUse hook (`rtk hook claude`) in Claude Code settings.
2. Have the agent run a raw `grep -n pattern file.py`.
3. Observe the output is RTK-shaped (`N matches in M files: [+K more]`) — the hook
   rewrote it to `rtk grep` and it ran through RTK.
4. Run `rtk discover`.
5. **Expected:** that call counted under "Already using RTK".
   **Actual:** it appears under MISSED SAVINGS as `grep -n`, with an estimated
   token saving that is not actually available.

On my 30-day window discover listed `grep -n` 37× and `ls -la` 34× as missed. Both
are rewritten by the hook on every invocation — confirmed with:

```
$ rtk hook check 'grep -n foo bar.py'
rtk grep -n foo bar.py

$ rtk hook check 'cd /tmp && grep -n foo bar.py'
cd /tmp && rtk grep -n foo bar.py
```

## Suggested fix

`rtk hook check` already exposes the rewrite engine as a pure function. Applying
those same rules to each transcript command before classification would make
`discover` agree with reality, with no new logging or storage:

- if `hook check <cmd>` yields a rewrite → count as covered, not missed
- if it does not → genuine miss, report as today

An alternative would be having the hook record its rewrites, but that adds runtime
state for something the dry-run can already answer retroactively.

Two details that mattered when I implemented this correction locally:

- `cd X && rtk …` and pipelines should count as covered — matching only a leading
  `rtk` token undercounts significantly.
- `rtk proxy <cmd>` should **not** count as coverage; it deliberately runs the raw
  command, so counting it lets the audit flatter itself via its own escape hatch.

## Secondary: `$(cat …)` command substitution is unhandled

147 occurrences in my 30-day window, the largest single unhandled shape. These come
from a python-shim pattern:

```
$(cat graphify-out/.graphify_python) -c "..."
```

The hook can't see past the substitution to the interpreter, so nothing is
rewritten. Lower priority than the above, but it is the top entry on my
"genuinely unhandled" list once the false misses are removed.

---

*Happy to share the local correction script (~260 lines of Python that reproduces
the corrected figures from transcripts) if it's useful as a reference.*

albatrossflyon-coder:
## Summary

Fixes #3148.

`rtk discover` reads commands from Claude Code transcripts as the model emitted them — before the PreToolUse hook rewrites them. It classified every `Supported` command as a missed opportunity regardless of whether the hook was actually installed and had already rewritten that exact command at runtime, producing a real undercount of RTK usage (the issue reports ~19.5 points on a 30-day/359-transcript window).

**Fix:** re-derive per-command whether an installed hook would have covered this exact instance, using the same rewrite engine the hook itself calls (`registry::rewrite_command`, same function backing `rtk hook check`). This respects `exclude_commands`/`transparent_prefixes` config and the same unattestable-construct defer the hook applies (heredocs, command substitution, etc.) — so commands the hook genuinely couldn't/wouldn't touch still report as misses, only commands it actually would have rewritten move to "already using RTK."

**Secondary fix, same issue:** `rtk proxy <cmd>` deliberately runs the raw command unfiltered, but was being counted as coverage just because it starts with `"rtk "`. Excluded explicitly — the escape hatch shouldn't flatter the audit it's supposed to be exempt from.

Did not address the issue's third (explicitly lower-priority) finding — `$(cat …)` command substitution shim detection — to keep this PR scoped to the two miscounts the issue calls the actual bug.

## Test plan

- [x] Added `covered_by_hook`/`is_already_rtk` as small pure functions (matching the existing pattern in this codebase — `hook_check.rs`, `rewrite_cmd.rs`) with 7 new unit tests covering: hook not installed, rewritable command covered, unattestable construct still a miss, config-excluded command still a miss, plain `rtk` prefix counted, `rtk proxy` excluded, unrelated command not counted
- [x] `cargo test` — 2478 passed, 0 failed
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --all-targets` — no issues

### DIFF
--- a/src/discover/mod.rs
+++ b/src/discover/mod.rs
@@ -17,6 +17,7 @@ use registry::{
 use report::{DiscoverReport, SupportedEntry, UnsupportedEntry};
 
 use crate::discover::registry::prefix_contains_rtk_disabled;
+use crate::hooks::hook_check::{status as hook_status, HookStatus};
 
 /// Aggregation bucket for supported commands.
 struct SupportedBucket {
@@ -72,6 +73,17 @@ pub fn run(
         }
     }
 
+    // Transcripts record commands as the model emitted them, before the PreToolUse
+    // hook rewrites them. If the hook is installed, a command it would rewrite was
+    // already routed through RTK at runtime — it's coverage, not a missed opportunity.
+    // Uses the same rewrite engine the hook itself calls (`rtk hook check`), so a
+    // command excluded via config or containing an unattestable construct (the hook
+    // would defer on it) still counts as a genuine miss below.
+    let hook_installed = hook_status() != HookStatus::Missing;
+    let (excluded, transparent_prefixes) = crate::core::config::Config::load()
+        .map(|c| (c.hooks.exclude_commands, c.hooks.transparent_prefixes))
+        .unwrap_or_default();
+
     let mut total_commands: usize = 0;
     let mut already_rtk: usize = 0;
     let mut parse_errors: usize = 0;
@@ -115,6 +127,18 @@ pub fn run(
                 }
 
                 match classify_command(part) {
+                    Classification::Supported { .. }
+                        if covered_by_hook(
+                            part,
+                            hook_installed,
+                            &excluded,
+                            &transparent_prefixes,
+                        ) =>
+                    {
+                        // Hook is installed and would rewrite this exact command —
+                        // it already ran through RTK, not a missed opportunity.
+                        already_rtk += 1;
+                    }
                     Classification::Supported {
                         rtk_equivalent,
                         category,
@@ -169,8 +193,7 @@ pub fn run(
                         bucket.count += 1;
                     }
                     Classification::Ignored => {
-                        // Check if it starts with "rtk "
-                        if part.trim().starts_with("rtk ") {
+                        if is_already_rtk(part) {
                             already_rtk += 1;
                         }
                         // Otherwise just skip
@@ -274,6 +297,31 @@ pub fn run(
     Ok(())
 }
 
+/// Whether a `Supported` command was already routed through RTK by an installed
+/// PreToolUse hook, using the same rewrite engine the hook itself calls (`rtk hook
+/// check`). A transcript records the command as the model emitted it — before any
+/// hook rewrite — so this re-derives whether the hook would have intercepted this
+/// exact instance, respecting excludes/transparent-prefixes and the same
+/// unattestable-construct defer the hook applies (heredocs, substitutions, etc.).
+fn covered_by_hook(
+    cmd: &str,
+    hook_installed: bool,
+    excluded: &[String],
+    transparent_prefixes: &[String],
+) -> bool {
+    hook_installed
+        && !lexer::contains_unattestable_construct(cmd)
+        && registry::rewrite_command(cmd, excluded, transparent_prefixes).is_some()
+}
+
+/// Whether an already-`rtk`-prefixed command counts as coverage. `rtk proxy <cmd>`
+/// deliberately runs the raw command unfiltered, so it must not count — that would
+/// let the audit flatter itself via its own escape hatch.
+fn is_already_rtk(cmd: &str) -> bool {
+    let trimmed = cmd.trim();
+    trimmed.starts_with("rtk ") && !trimmed.starts_with("rtk proxy")
+}
+
 /// Extract the subcommand from a command string (second word).
 fn extract_subcmd(cmd: &str) -> &str {
     let parts: Vec<&str> = cmd.trim().splitn(3, char::is_whitespace).collect();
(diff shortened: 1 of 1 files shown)


pszymkowiak (approved):
Verified on a compiled build — unit tests (7) cover all branches of `covered_by_hook` (hook-not-installed → miss, rewritable → covered, unattestable-construct → miss, config-excluded → miss) and `is_already_rtk` (incl. the `rtk proxy` exclusion). Also ran it end-to-end with a synthetic transcript via `CLAUDE_CONFIG_DIR`:

```
hook absent   → Already using RTK: 0 (0%)    ; grep / git status counted as MISS
hook present  → Already using RTK: 2 (50%)   ; git status $(whoami) stays a MISS, rtk proxy not counted
```

So rewritable commands correctly move from "missed" to "covered" when the hook is installed, while the unattestable-construct case still reports as a genuine miss (the hook would defer) and `rtk proxy` never flatters the count. Reusing `registry::rewrite_command` — the same engine the hook calls — is the right call for consistency with `rtk hook check`.

One minor semantic note (non-blocking): `hook_installed` reflects the hook's *current* state, not whether it was installed when each transcript was recorded, so very old pre-install transcripts could be over-counted as covered. Given discover is a heuristic and defaults to a 30-day window (where hook state is usually stable), and the pre-rewrite form is all the transcript preserves, this is a reasonable trade-off — maybe worth a line in the help/docs.

LGTM 👍

src/discover/mod.rs
KuSh, line 121: it seems this one should also be filtered by `covered_by_hook` even if less important

KuSh:
Nice catch on the root cause (transcripts recording pre-rewrite commands), and reusing the real rewrite engine instead of the `starts_with("rtk ")` heuristic is the right call. But I think this reintroduces the same *class* of bug in the other direction:

`hook_installed` (`src/discover/mod.rs:80`) and the excluded/transparent-prefix config are snapshotted once from **current** state and then applied uniformly to every transcript entry across the whole `since_days` window (up to 30 days). Two concrete failure modes:

1. **Hook installed after the history was recorded.** If someone installs the hook today and runs `rtk discover --since-days 30`, every rewritable command from the prior 29 days — which genuinely ran unfiltered, since no hook existed yet — gets classified as `already_rtk` instead of a miss. That's not a hypothetical edge case; running `discover` right after installing rtk is probably the single most common trigger for this command.
2. **Registry support added after the history was recorded.** `covered_by_hook()` calls `registry::rewrite_command()` against whatever rtk version is installed *now*. If command support for X shipped after some of the scanned history was generated, those historical invocations get retroactively marked as covered even though the hook (if it existed at all at that point) had no rewrite for X yet.

Both stem from the same thing: `ExtractedCommand` (`src/discover/provider.rs:13`) carries no per-entry timestamp, and there's no stored record of when the hook was installed or which rtk version was active at any given point in the scan window — so there's no way to bound the "would this have been covered" check to the state that actually existed at that time.

Given the PR's own motivation is fixing a systematic ~19.5-point miscount, I'd want this to not trade one systematic miscount for another. Suggestions, roughly in order of effort:
- Minimal: bound the `covered_by_hook` check by hook-install mtime (if resolvable) so history predating install still counts as a miss, and document that command-registry-support drift isn't handled.
- Better: have the hook itself log its rewrite decision (allow/ask/deny/passthrough + rtk version) at the moment it runs, so `discover` reads ground truth for any window covered by that log instead of re-deriving it from current state. Happy to sketch this separately if useful — it'd also incidentally let us report whether "ask" rewrites were actually accepted, which today isn't tracked anywhere.

Not blocking if there's appetite to land the transcript-parsing fix now and treat the time-travel issue as a fast-follow, but I think it should at least be called out in the PR description / a tracking issue so it doesn't get read as "coverage stats are now trustworthy" when they're only trustworthy for windows where the hook + registry haven't changed.

KuSh (changes requested):

albatrossflyon-coder:
Good catch, and you're right that this is a real time-travel bug, not just a theoretical edge case — `hook_status()` (`src/hooks/hook_check.rs`) only checks whether a hook is installed *right now*, there's genuinely no stored record of when it was installed or which rtk version was active at any point in the past.

For a minimal fix I can bound the scan by `settings.json`'s mtime (`resolve_claude_dir().join(SETTINGS_JSON)`) — treat any transcript entry older than that as unknown/pre-hook and count it as a miss regardless of what `covered_by_hook()` says today. It's an approximation (mtime moves on *any* settings edit, not just the hook being added, and it does nothing for the registry-version-drift case you raised — a command whose rewrite support shipped after the fact still gets misclassified) but it directly fixes failure mode #1, which you called the most common trigger.

I'd rather not try to solve #2 (registry drift) in this PR — that really needs the hook to log its own rewrite decision + rtk version at runtime, which is a bigger persistence change than a bug-fix PR should carry. I'll open a tracking issue for that and link it here, and call out both limitations explicitly in this PR's description so nobody reads the merged stats as "fully time-accurate."

Will push the mtime bound + doc update to this branch — let me know if you'd rather see it split into its own follow-up PR instead of amending this one.

KuSh:
@albatrossflyon-coder I've pushed a complete fix in https://github.com/rtk-ai/rtk/pull/3206 that should superseed your PR. Could you check it before spending time updating your PR?

albatrossflyon-coder:
Thanks for pushing through on this — #3206 is the right fix, not just a good-enough one. It solves both gaps you originally flagged (pre-install history via the real hook_decisions log, and registry/permission drift via measuring the actual PermissionVerdict at the moment the hook ran) instead of the mtime-bound approximation I was about to add here, which only ever addressed the first. The tool_use_id join and the hook-decision logging table are a materially better foundation than re-deriving coverage from current state, and the validation (hook-absent vs hook-installed-with-real-decisions, direct sqlite checks) is thorough.

Closing this in favor of #3206 — no need to duplicate review effort. Thanks for catching the gap and doing the deeper fix.

PR #3165, opened 2026-07-22
docs: add root AGENTS.md router for contributing agents

### LINKED ISSUE #3141: Add a root AGENTS.md to guide coding agents contributing to rtk
### Context

A growing share of contributions to rtk come through coding agents (Claude Code, Codex, etc.). rtk already has excellent, well-thought-out guidance for contributors — the design philosophy, the runner-standardization pattern, the Transparency/`never_worse` contract, the Rust conventions — but it's spread across `CONTRIBUTING.md`, `docs/contributing/CODING_PRACTICES.md`, `ARCHITECTURE.md`, and `.claude/rules/`. There's no single entry point that surfaces the key rules to an agent (or a new human contributor) at the *start* of a task, so it's easy to start coding before encountering them.

### Proposal

Add a root **`AGENTS.md`** — the emerging cross-tool convention, symmetric to `CLAUDE.md` — as a terse **router**: trigger → one-line rule → link to the authoritative doc. Deliberately *not* a copy of the rules (so they can't drift out of sync); it just points contributors at the right doc up front. Pairing it with an `@AGENTS.md` import line in `CLAUDE.md` means Claude Code picks it up automatically too.

### Not to be confused with

This is **distinct** from rtk's existing `AGENTS.md` *integration* for Codex/Kimi (`rtk init --codex`), which writes RTK-awareness into a *user's* `AGENTS.md`. This proposal is a file in *rtk's own repo root* that guides people (and agents) contributing to rtk itself.

### Scope

One new `AGENTS.md` + a one-line `CLAUDE.md` import. Pure docs, no code. I've been working in the codebase recently and have a working draft ready — happy to open the PR if there's interest.

albatrossflyon-coder:
## Summary

Fixes #3141.

A growing share of contributions to rtk come through coding agents. The existing contribution guidance is thorough — design philosophy, the filter-authoring checklist, Rust conventions, testing strategy — but it's spread across `CONTRIBUTING.md`, `docs/contributing/*.md`, and `.claude/rules/`, with no single entry point an agent (or new human contributor) encounters at the *start* of a task.

Adds a root `AGENTS.md` as a terse router: trigger → one-line rule → link to the authoritative doc. Deliberately **not** a copy of the rules — it just points contributors at the right doc up front, so the router can't drift out of sync with the real thing. `CLAUDE.md` now imports it via `@AGENTS.md` so Claude Code picks it up automatically.

**Not to be confused with:** the existing `rtk init --codex`/`rtk init --agent kimi` integration, which writes RTK-awareness into a *user's own project* `AGENTS.md`. This PR adds a file in rtk's own repo root, for people (and agents) contributing to rtk itself.

## Test plan

- [x] Pure docs change, no Rust source touched
- [x] Verified every linked file/anchor exists and resolves (`CONTRIBUTING.md`, `src/cmds/README.md`, `docs/contributing/ARCHITECTURE.md`, `.claude/rules/*.md`)
- [x] `scripts/validate-docs.sh` passes
- [x] Confirmed `CLAUDE.md` is otherwise identical to `upstream/develop` before editing, so the diff is scoped to exactly the one added import line


randomBrainstormer:
Nice, this reads really cleanly, and calling out the difference from `rtk init --codex` up front is spot on.

One thing before it lands: a lot of these rules already live inline in `CLAUDE.md` (no `unwrap()`, `lazy_static!` for regex, no async, the whole fmt/clippy/test gate). So now they're written in two places, which is the drift the router was meant to avoid, just moved to CLAUDE.md ↔ AGENTS.md. Might be cleaner to let `AGENTS.md` hold the shared stuff and trim `CLAUDE.md` down to the Claude-only bits. Bonus: the other agents would then see the Plan / Rabbit-Holes / Working-Dir sections that are Claude-only today.

One small add for the "add a filter" row too: the thing people actually trip on here is routing through `core/runner` + `guard::never_worse` + returning `Result<i32>` instead of rolling their own `.output()`/exit. A one-line pointer to that would save a review round or two.

Either way, thanks for getting this moving!

albatrossflyon-coder:
Pushed the trim — `fef7971`. `CLAUDE.md`'s "Coding Rules" and "Build Verification" sections now point at `AGENTS.md` for the shared rules instead of restating them, keeping only the Claude-specific enforcement language and the hyperfine perf-check steps. Also added the `core/runner` + `guard::never_worse` + `Result<i32>` pointer to the "add a filter" row.

`scripts/validate-docs.sh` still passes.

albatrossflyon-coder:
Checking in on this one — it's been open since 07-22, with the requested trim (duplicate-rules cleanup) pushed on 07-27. Happy to make any further changes, or let me know if anything's still blocking review. Thanks for maintaining rtk!

randomBrainstormer:
Thanks for the trim, that reads better. A few things have come up since though, partly because `develop` has moved.

Coming back to what I had in mind in #3141: I was hoping for a single source of truth, with `CLAUDE.md` ending up as just the one line `@AGENTS.md`. Right now it still has seven sections of its own (Project Overview, Development Commands, Architecture, Proxy Mode, Build Verification, Working Directory, Rabbit Holes, Plan Protocol), so we've ended up with two files that both give instructions instead of one.

And the drift is already showing. The pre-commit gate now exists in three versions: `CLAUDE.md` says `cargo fmt --all && cargo clippy --all-targets && cargo test --all`, `CONTRIBUTING.md` says `cargo fmt --all --check && cargo clippy --all-targets && cargo test`, and the new `AGENTS.md` row copies the first one but links to `CONTRIBUTING.md`. Not just cosmetic either — `--check` fails instead of rewriting your files, and `--all` changes what gets tested.

Same thing with the `lazy_static!` row: `5269df7` swapped it for `LazyLock`, so `develop` has zero `lazy_static` left in `src/` and 30 uses of `LazyLock`, and both `rust-patterns.md` and `CLAUDE.md` already say `LazyLock`. The rule is a bit narrower now too (fixed patterns reused across calls go in `LazyLock`, runtime-dependent ones stay local).

That one's on the format I asked for in the issue, not on you — a "one-line rule" column is exactly what invites restating things that change. Might be better if that column only carries stuff that doesn't move (no async, fallback on failure, exit code propagation) and leaves the rest to the link.

One thing worth sorting out before deciding the layout: there's no `settings.json` here, so Claude Code picks up `.claude/rules/*.md` on its own. Claude gets those three files for free, while Codex and Kimi have to go follow links into a `.claude/` folder. So the `@AGENTS.md` import on its own doesn't really put the agents on equal footing. If we want that, the rules probably need to live in `docs/contributing/` with `AGENTS.md` pointing at them.

I don't want to dictate the file layout though — maintainers, how far this should go is your call. Happy to help move things around if that's useful.

Also needs a rebase, the conflict is in the `Coding Rules` block that `a1673f7` touched too.

Small stuff while I'm here: `Build Verification (Mandatory)` now ends on a colon with nothing after it, and "all 3 checks" doesn't have anything left to refer to. Working Directory / Rabbit Holes / Plan Protocol are still Claude-only and not mentioned in `AGENTS.md`. The "Search the codebase" row sends everyone to `search-strategy.md`, which is written in Claude Code tool names (Grep/Glob/Read/Explore agent). And there are two rows starting with "Commit", with the gate one last, after "Open a PR".

Last one, for the maintainers: I think there's a real collision here, not just a naming one. `rtk init --codex` in local mode writes to `./AGENTS.md` (`src/hooks/init.rs:2377`) and appends `@RTK.md` to it, and `--agent kimi` injects a whole RTK block. So anyone dogfooding rtk inside this repo would end up modifying a tracked file. Probably worth deciding before this lands.

Thanks again for pushing this along!

albatrossflyon-coder:
Thanks for the detailed re-review, this was a real gap-check and I've pushed two more commits addressing it.

**Rebased onto current `develop`** (`f597fcb`), resolving the conflict in the "Coding Rules" section in favor of the pointer-to-AGENTS.md direction.

**On the single-source-of-truth ask:** I moved Project Overview, Development Commands, Architecture, and Proxy Mode out of `CLAUDE.md` into `AGENTS.md`, so Codex/Kimi/etc. get the same reference Claude Code does instead of a router pointing at Claude-only content. `CLAUDE.md` now imports `AGENTS.md` and keeps only three sections I'm treating as genuinely Claude Code session-management behavior rather than rtk coding rules: working-directory confirmation, rabbit-hole avoidance, and the plan-execution protocol. That's not literally the one line you asked for, flagging that directly rather than let it look like an oversight. Happy to go all the way to one line if you (or a maintainer) still want that, my read was that those three aren't "rules that could drift," they're Claude-specific behavior, so keeping them adjacent felt more honest than either deleting them or forcing them onto agents they don't apply to. Genuinely your call.

**Content drift fixed:**
- `LazyLock` instead of `lazy_static!` in the router row, matching `rust-patterns.md`'s current (narrower) rule
- The pre-commit gate command matches `CONTRIBUTING.md` everywhere it's stated now (`cargo fmt --all --check && cargo clippy --all-targets && cargo test`), and after a `/simplify` pass caught me re-duplicating it three times inside my own fix, it's now stated once in `AGENTS.md` with the router row linking to it instead of repeating it

**Equal footing:** moved `rust-patterns.md`, `cli-testing.md`, and `search-strategy.md` from `.claude/rules/` to `docs/contributing/`, so Claude Code doesn't get free auto-loaded context that Codex/Kimi contributors following the same `AGENTS.md` links don't get. You also specifically flagged that `search-strategy.md`'s content is written in Claude Code tool names (Grep/Glob/Read) even after the move, that's real, the move fixed access but not usability. Added an explicit note plus generic equivalents (`rg`/`fd`) rather than a full rewrite, since the Claude-specific syntax is still directly useful for Claude Code itself.

**Small stuff:** the two "Commit" rows are one now, before "Open a PR". The dangling "Build Verification (Mandatory):" colon is gone, that section got folded into `AGENTS.md`'s Development Commands entirely, along with the `hyperfine` performance-verification block that was oddly stranded under it.

**On the collision you flagged** (`rtk init --codex`/`--agent kimi` writing to a user's own `./AGENTS.md` vs. this PR adding a tracked one at rtk's own repo root): not resolving this myself, agreed it needs a maintainer decision. Wanted to flag it's now slightly more concrete: if a maintainer dogfoods `rtk init --codex` inside a checkout of this repo, `git status` would show `AGENTS.md` as modified rather than untracked, since it's tracked now. Whether that's an acceptable "well, don't do that" or needs an actual guard is genuinely not mine to call.

**One thing I found and didn't fix here, opened as its own issue instead:** #3626. `AGENTS.md`'s Development Commands/Architecture content still duplicates `.github/copilot-instructions.md` almost line-for-line, and `INSTALL.md`'s name-collision warning is now also independently duplicated in `copilot-instructions.md`. That's a pre-existing pattern across those two files predating this PR, bigger scope than #3141 asked for, so I linked `AGENTS.md`'s collision warning to `INSTALL.md` instead of triple-duplicating it, but didn't touch `copilot-instructions.md` itself.

**On CI:** the current run's `benchmark` job failure (golangci-lint filter returning empty output, one filter producing more tokens than raw) is pre-existing on `develop` itself as of the commit I rebased onto (`29f9bb7`, same failure signature, checked directly), not something this PR introduced.

Ran `vuln-hunter`'s diff scanner and a 4-angle simplification review against both commits (clean after the second one), and `scripts/validate-docs.sh` passes. Every link in the touched files resolves to a real file.

PR #3166, opened 2026-07-22
fix(read): head -N truncation must be a literal contiguous prefix

### LINKED ISSUE #3046: head -N rewritten to whole-file compressed view folds structure without inline markers — false-positive "orphaned }" nearly triggered a wrong edit
## Summary

`head -30 <file>` executed through the rtk rewrite hook (Claude Code PreToolUse) returned what appears to be a **whole-file compressed view** instead of the first 30 lines: the output ended with `[124 more lines]` for a 154-line file, and content from beyond line 30 was folded into the visible portion. Function bodies were elided **without any inline marker at the fold point**, which caused a dangerous false positive.

## What happened

a 154-line `~/.zshrc` containing (structure sanitized):

```
L10-L32   various export PATH=... lines
L61       git_branch() {
L62         git branch 2>/dev/null | ...
L63       }
L67       PROMPT='...'
...
L154      EOF
```

Command issued by the agent: `head -30 ~/.zshrc`

Output received (abridged): a mixed view where the closing brace from **line 63** appeared directly between two `export PATH` lines, with the function header `git_branch() {` (line 61) folded away:

```
export PATH="/opt/homebrew/opt/mysql-client/bin:$PATH"
}
export PATH="..."
...
[124 more lines]
```

## Consequence

The LLM agent reading this output concluded the file contained an **orphaned `}` (syntax error)** and proposed editing the user's shell config to remove it. The file was in fact perfectly valid (`zsh -n` clean); the brace was a normal function closer whose opening line had been silently folded. The edit was only avoided because a later `Read`/`grep -n` of the raw file contradicted the compressed view.

## Why this matters

Related open issues (#2204 silent truncation without marker, #2806 rg false negatives, #1296/#2665 git log silent truncation) all describe **omission** — the LLM misses data. This case is the more dangerous direction: **the rewritten view fabricated an apparent defect that triggered a concrete (wrong) modification action** on a user's config file. Silent truncation misleads scope; silent folding misleads *content*.

## Suggestions

1. Preserve `head -N` semantics — return only the first N lines; a bounded read is already token-cheap and should not be replaced by a whole-file summary.
2. Wherever compression folds lines, emit an **inline elision marker at the fold point** (e.g. `[... 2 lines folded]`), not just a trailing total count. A trailing `[124 more lines]` does not tell the reader that *interior* lines are missing.

rtk 0.43.0 (homebrew), macOS, invoked via Claude Code PreToolUse hook rewrite.

albatrossflyon-coder:
## Summary

Fixes #3046.

`rtk read --max-lines N` — what `head -N` (and `cat`/`tail` via the same rule) rewrites to — is supposed to carry the exact same contract as `head` itself: the literal first N lines, unchanged. Instead, `smart_truncate` scanned the **entire** file and pulled in any line matching an "important" pattern (function signatures, imports, `pub`/`export` prefixes, or a bare `}`/`{`) regardless of how far past the requested window it actually was, capping total shown lines at N with no marker indicating where lines were skipped from the middle.

Concretely, for a 154-line file with `head -30`: a closing `}` from line 63 got pulled in (it matched the bare-brace "important" rule) while its own function header on line 61 — which didn't match any of the importance patterns — got silently dropped. The output showed that stray `}` sitting directly after an early, unrelated line with nothing indicating a gap existed. The agent reading it concluded the file had a genuine orphaned brace and nearly edited the user's real shell config to "fix" a defect that was actually just missing context from `head`'s own output.

This is the same category of bug the project's own design principle calls out directly (`CONTRIBUTING.md` § Transparency): RTK's output must be "a valid, useful subset of the original tool's output, not a different format the LLM wouldn't expect." A `head`-shaped request answered with a non-contiguous "greatest hits" sample is exactly that kind of invented format.

**Fix:** replace the importance-based scan with a plain contiguous slice of the first `max_lines - 1` lines plus the existing single trailing marker — matching the `tail-lines` path in the same function, which was already implemented correctly as a genuine contiguous slice.

## Test plan

- [x] Updated `test_smart_truncate_no_annotations` for the corrected line count (was asserting the buggy `max_lines/2` behavior)
- [x] Added `test_smart_truncate_does_not_pull_in_later_important_lines` — directly reproduces the #3046 shape (a bare `}` past the window) and asserts it's no longer pulled in
- [x] `test_smart_truncate_overflow_count_exact` (kept-count + reported-more == total invariant) needed no changes — still holds under the new contiguous logic
- [x] `cargo test` — 2472 passed, 0 failed
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --all-targets` — no issues

### DIFF
--- a/src/core/filter.rs
+++ b/src/core/filter.rs
@@ -326,33 +326,15 @@ pub fn smart_truncate(content: &str, max_lines: usize, _lang: &Language) -> Stri
         return content.to_string();
     }
 
-    let mut result = Vec::with_capacity(max_lines + 1);
-    let mut kept_lines = 0;
-
-    for line in &lines {
-        let trimmed = line.trim();
-
-        // Prioritize structurally important lines so the visible window stays useful.
-        // The old approach interleaved "// ... N lines omitted" markers which AI agents
-        // treated as code, causing parsing confusion and extra retry loops.
-        let is_important = FUNC_SIGNATURE.is_match(trimmed)
-            || IMPORT_PATTERN.is_match(trimmed)
-            || trimmed.starts_with("pub ")
-            || trimmed.starts_with("export ")
-            || trimmed == "}"
-            || trimmed == "{";
-
-        if is_important || kept_lines < max_lines / 2 {
-            result.push((*line).to_string());
-            kept_lines += 1;
-        }
-        // Non-important lines beyond max_lines/2 are silently skipped —
-        // no inline markers that could be mistaken for file content.
-
-        if kept_lines >= max_lines - 1 {
-            break;
-        }
-    }
+    // Bounded reads (head -N, rtk read --max-lines) promise the literal first N
+    // lines, unchanged — the same contract `head` itself has. Pulling in
+    // "important-looking" lines from later in the file (rtk-ai/rtk#3046) breaks
+    // that contract silently: a closing `}` from far past the window can land
+    // right after an early line with no marker, reading as an adjacent syntax
+    // error that isn't real. Keep it a plain contiguous slice, like the
+    // tail-lines path already does.
+    let kept_lines = max_lines.saturating_sub(1);
+    let mut result: Vec<String> = lines[..kept_lines].iter().map(|l| l.to_string()).collect();
 
     // Single end-of-output marker: not code syntax, unambiguous to AI agents.
     // Invariant: kept_lines + N == lines.len() (N = lines not shown)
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for tracking this down with such a clear reproduction. #3941, merged on 2026-09-14, fixed the reported symptom a different way: `head -N` is no longer rewritten to `rtk read --max-lines N` at all, it now goes through a new `--head-lines N` that slices a byte-exact first-N-lines window, and `--max-lines` was deliberately kept as the structural preview. Its help text on develop now reads "keeps signatures and imports; not the first N lines, use --head-lines for that", so making smart_truncate a contiguous slice would remove behaviour develop now documents on purpose, which is why this cannot land as written. What develop still misses is the other half of open issue #3421: the structural preview drops lines from the middle with no gap marker, only a trailing `[N more lines]` at src/core/filter.rs:320-363, and a narrow PR that adds gap markers there would be very welcome. If a case your patch covers has been missed here, please comment and this can be reopened.

Issue #2989, opened 2026-07-14
Title: Request for digitally signed Windows releases


broll:
# Request: Digitally Signed Windows Releases

## Background

We are evaluating this project for use in a corporate environment that uses application allowlisting controls.

Currently, the Windows executable appears to be unsigned, which means we cannot create publisher-based trust rules. Instead, each release must be individually approved using file hashes, creating additional operational overhead and making upgrades more difficult.

## Request

Would it be possible to provide digitally signed Windows releases using a code-signing certificate with a consistent publisher identity?

## Business Impact

Providing signed releases would allow organizations like ours to:

- Create publisher-based trust rules rather than hash-based rules.
- Reduce the operational effort required to approve new releases.
- Simplify application upgrades and maintenance.
- Improve confidence that binaries are official releases from the project maintainers.
- Better support deployment in enterprise environments using application allowlisting solutions such as Airlock Digital.

## Additional Notes

We understand that obtaining and maintaining a code-signing certificate involves both cost and administrative effort. However, signed releases would significantly improve the ability for enterprise customers to adopt and maintain the software.

Thank you for considering this request and for maintaining the project.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟢 **Priority** | `low` |
| 🎯 **Confidence** | 90% |

### Summary

Enterprise user requests digitally signed Windows release binaries to support publisher-based application allowlisting.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

PR #3167, opened 2026-07-22
build(windows): embed VERSIONINFO into the compiled .exe

### LINKED ISSUE #3124: v0.43.0 detected as virus by Microsoft defender
I received a threat alert when I updated to v0.43.0

Image

albatrossflyon-coder:
## Summary

Relates to #3124 (v0.43.0 flagged as a virus by Microsoft Defender).

An unsigned binary shipping with zero embedded version metadata — no CompanyName, ProductName, FileVersion, publisher identity of any kind — is a documented contributing factor to AV heuristic false positives for Rust/Go CLI tools generally. This adds a Windows-only build.rs step that compiles `rtk.rc` via `embed-resource` and links the result into the binary, giving the compiled `.exe` real version/publisher metadata instead of shipping "faceless."

**I want to be upfront about what this does and doesn't accomplish**, rather than overclaim a fix for something I can't fully verify:

- **Verified working:** `FileVersion`/`ProductVersion` (`0.42.4.0`) and `Language` (English - United States) are correctly embedded and readable — checked directly against the built binary via .NET's `FileVersionInfo` API.
- **Not working, and I don't know why:** the `StringFileInfo` string values (`CompanyName`, `ProductName`, `FileDescription`, `LegalCopyright`, `OriginalFilename`) still come back empty on inspection, even after getting the `FIXEDFILEINFO` block (`FILEVERSION`/`PRODUCTVERSION`/etc.) correct and confirming the langID/codepage block matches the `VarFileInfo` `Translation` value. I suspect an RC codepage/encoding mismatch, but I didn't pin it down, and at that point I was guessing at `rc.exe` syntax variations without a cheap way to verify each guess — didn't want to keep committing time to that without a clearer lead. Left the `StringFileInfo` block in place since it's harmless even non-functional, and someone with more Windows resource-compiler experience may spot the issue at a glance.
- **This is not a replacement for code signing** (tracked separately in #2989) — that remains the real fix for enterprise allowlisting, and it's a cost/infrastructure decision for the maintainers, not something a contributor PR can resolve.
- **Whether this measurably affects the actual Defender false-positive rate is not something I can verify** without a real signed release build and live detection telemetry over time. AV heuristics are opaque; I'm not claiming this closes #3124, just that it's a real, low-risk improvement over the current state.

Happy to have someone else pick up the StringFileInfo gap, or to keep digging myself if there's a specific lead worth chasing.

## Test plan

- [x] `cargo build` — compiles clean on Windows, `rc.exe` runs without error, resulting `.lib` links into the binary
- [x] Verified via `[System.Diagnostics.FileVersionInfo]::GetVersionInfo()` against the built `target\debug\rtk.exe`: `FileVersionRaw`/`ProductVersionRaw` = `0.42.4.0`, `Language` = `English (United States)` — both correct
- [x] `cargo fmt --all`, `cargo clippy --all-targets`, `cargo test` — all clean, 2472 passed, 0 failed
- [ ] StringFileInfo values (CompanyName/ProductName/etc.) — not resolved, documented above

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -343,6 +343,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "embed-resource"
+version = "3.0.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fbfdaacccebec3b28e4866b8973543c7647797db5ada1bdab552e48fe665fbbd"
+dependencies = [
+ "cc",
+ "memchr",
+ "rustc_version",
+ "toml 1.1.3+spec-1.1.0",
+ "vswhom",
+ "winreg",
+]
+
+[[package]]
 name = "encoding_rs"
 version = "0.8.35"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -939,6 +953,7 @@ dependencies = [
  "codepage",
  "colored",
  "dirs",
+ "embed-resource",
  "encoding_rs",
  "flate2",
  "getrandom 0.4.2",
@@ -952,7 +967,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "tempfile",
- "toml",
+ "toml 0.8.23",
  "ureq",
  "walkdir",
  "which",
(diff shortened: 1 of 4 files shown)


Soap-141:
Any updates?

PR #3168, opened 2026-07-23
feat: add --secure, --no-telemetry, --no-tracking flags + config for corporate compliance

sigsegv0x0b:
## Summary

- Add `--secure`, `--no-telemetry`, `--no-tracking` CLI flags for corporate compliance
- Add `secure = true` config.toml field (MDM-deployable)
- Respect existing `[tracking] enabled = false` config (was previously ignored)
- All flags/config/env follow priority: CLI > env > config > defaults

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — 2470 pass
- [x] Manual: `rtk --no-telemetry telemetry status` shows "secure flag: blocked"
- [x] Manual: `rtk --no-tracking gain` shows "Tracking is disabled"
- [x] Manual: `rtk --secure git log -1` runs without creating history.db
- [x] Manual: `RTK_DISABLE_TRACKING=1 rtk gain` shows disabled message
- [x] Unit tests for secure.rs init/setters (5 tests)
- [x] Unit test for gain run disabled message

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,9 +463,17 @@ rtk telemetry disable    # Withdraw consent — stops all collection immediately
 rtk telemetry forget     # Withdraw consent + delete all local data + request server-side erasure
 ```
 
+**Override via CLI flag (per-invocation):**
+```bash
+rtk --no-telemetry git log -5       # Skip telemetry for this command
+rtk --no-tracking cargo build       # Skip local history.db for this command
+rtk --secure git log -5             # Disable both (overrides config.toml)
+```
+
 **Override via environment:**
 ```bash
-export RTK_TELEMETRY_DISABLED=1   # Blocks telemetry regardless of consent
+export RTK_TELEMETRY_DISABLED=1    # Blocks telemetry regardless of consent
+export RTK_DISABLE_TRACKING=1      # Blocks local tracking regardless of consent
 ```
 
 ## Star History
--- a/docs/TELEMETRY.md
+++ b/docs/TELEMETRY.md
@@ -127,6 +127,12 @@ rtk telemetry disable    # Withdraw consent
 rtk telemetry forget     # Withdraw consent + delete local data + request server erasure
 ```
 
+CLI flag override (blocks telemetry regardless of consent, per invocation):
+```bash
+rtk --no-telemetry git log -5
+rtk --secure git log -5          # also disables local tracking
+```
+
 Environment variable override (blocks telemetry regardless of consent):
 ```bash
 export RTK_TELEMETRY_DISABLED=1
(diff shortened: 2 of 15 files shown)


pszymkowiak:
Thanks for tackling this — the `--secure`/`--no-telemetry`/`--no-tracking` flags and Clap wiring in `main.rs` are clean, and the `[tracking] enabled = false` fix is a real, verified bug fix (confirmed `Tracker::new()` on `develop` never checked `cfg.tracking.enabled` at all).

Found two issues while testing the compiled binary, though:

**1. `secure.rs::init()`'s raw `env::args()` scan produces false positives on unrelated passthrough arguments.** It looks for the literal tokens `--secure`/`--no-telemetry`/`--no-tracking` anywhere in `std::env::args()`, independent of the (correct) Clap parsing in `main.rs`. Reproduced:
```
rtk git commit -m "--secure"
```
This silently disables tracking for that invocation — even though no rtk flag was passed, it's just a commit message that happens to match. Confirmed the same with `--no-tracking`; ran a control case with an unrelated message to rule out flakiness. Not a security issue (fails toward less tracking, not more), but it defeats the "CLI > env > config > defaults" priority the PR describes, since here there's no CLI flag at all. Suggest dropping the raw-argv fallback in `init()` entirely and relying solely on the already-correctly-parsed `cli.secure`/`cli.no_telemetry`/`cli.no_tracking` booleans from `main.rs`.

**2. The test suite doesn't actually pass.** `cargo test --all` on this branch: 6 failures, 100% reproducible (checked twice, and single-threaded to rule out a parallelism race):
```
core::tracking::tests::test_tracker_record_and_recent
core::tracking::tests::test_timed_execution_records_time
core::tracking::tests::test_timed_execution_passthrough
core::tracking::tests::test_track_passthrough_no_dilution
core::tracking::tests::test_parse_failure_roundtrip
core::tracking::tests::test_parse_failure_recovery_rate
```
Root cause: `TELEMETRY_FORCE_DISABLED`/`TRACKING_FORCE_DISABLED` are process-global `AtomicBool`s that only ever get set to `true`, never reset outside `#[cfg(test)] reset_for_tests()` — which is only called from one place (`src/analytics/gain.rs:796`). Once any `secure.rs` test (e.g. `test_secure_disables_both`) runs in the same test binary, it poisons global state for every later `tracking.rs` test in the same process, and `Tracker::new()` starts bailing with "tracking disabled by --secure".

Note the CI workflows haven't run on this PR yet (only the CLA check shows up — looks like it needs a maintainer to approve running Actions for first-time contributors), so this wasn't caught there either.

Happy to take another look once these are addressed.

PR #3170, opened 2026-07-23
fix(benchmark): use deterministic curl and wget responses

TaKO8Ki:
## Summary

- Use MockHTTP’s indexed `/json/1` endpoint for the curl and wget benchmarks.
- Prevent false regressions caused by comparing different randomly selected JSON responses.

The previous `/json` endpoint selects a random response for each request. This meant the raw command and RTK could receive responses with different sizes.

The indexed endpoint returns the same response for every request. This PR should be merged after `/json/1` is deployed to MockHTTP.

## Test plan

- [x] `bash -n scripts/benchmark.sh`
- [ ] Manual testing after `/json/1` is deployed

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -346,8 +346,7 @@ bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 # ===================
 section "curl"
 if command -v curl &> /dev/null; then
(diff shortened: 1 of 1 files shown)


KuSh:
Following https://github.com/jaredwray/mockhttp/issues/180, mockhttp added stable JSON output (and other format too). It would also test JSON compaction

KuSh (approved):
We can already merge it but it won't be fixed until mockhttp release a new version

Issue #2424, opened 2026-06-12
Windows and GitHub Copilot


BillChirico:
I've installed everything for windows and GitHub Copilot and it does not seem to be using RTK at all. How can I get it to use RTK more to save tokens?

<img width="701" height="108" alt="Image" src="https://github.com/user-attachments/assets/a375cc8d-eaaa-495d-a4ba-d661d4627c6b" />

<img width="417" height="489" alt="Image" src="https://github.com/user-attachments/assets/5712210a-1387-4d3b-9593-644752d0134b" />

I see this in the `README.md`, but unsure what it actually means: _"On native Windows (cmd.exe / PowerShell), RTK filters work but the hook does not auto-rewrite commands"_

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 📋 **Category** | `question` |
| 🟢 **Priority** | `low` |
| 🎯 **Confidence** | 78% |

### Summary

User asking how to configure RTK with GitHub Copilot on Windows; likely a usage/setup question referencing a known Windows hook limitation documented in the README

<details>
<summary>📁 Relevant files</summary>

- `README.md`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

romain-cureau:
The README line you quoted has a concrete cause, at least for Copilot CLI on Windows: `rtk hook copilot` only recognizes the native Copilot payload when `toolName == "bash"`, but Copilot CLI names its shell tool `powershell` on Windows. The payload therefore falls through to `PassThrough` and the native path never runs.

Captured from a live session (Copilot CLI 1.0.73, Windows 11, RTK 0.43.0):

```json
{"sessionId":"...","cwd":"C:\Windows\System32","toolName":"powershell","toolArgs":"{\"command\":\"rtk gain\",...}"}
```

`toolArgs` is already the JSON-encoded string RTK expects — only `toolName` differs.

Details in #3178, one-line fix + regression test in #3179.

[...]

PR #3179, opened 2026-07-23
fix(hook): detect Copilot CLI shell tool on Windows

### LINKED ISSUE #3178: hook(copilot): on Windows, detect_format never matches Copilot CLI's shell tool (toolName: "powershell")
## Summary

`rtk hook copilot` recognizes the native Copilot CLI payload only when
`toolName == "bash"`. On Windows, Copilot CLI names its shell tool
**`powershell`**, so `detect_format` returns `PassThrough` and the
`HookFormat::CopilotCli` branch — the one that emits `modifiedArgs` and
deliberately omits `permissionDecision` — is never reached.

What answers instead is the Claude-compat `PreToolUse` entry, which routes to
`HookFormat::VsCode` and emits `permissionDecision: "ask"`. Since Copilot CLI
1.0.66+ treats that field as authoritative, **every rewritten command prompts**
(this is #3037), and the allowlist (`--allow-tool "shell(rtk:*)"`,
`permissions-config.json`) cannot help: only `allow`/`deny` short-circuit the
permission flow, `ask` forces the prompt.

This also explains the Windows limitation documented in the README
(*"On native Windows (cmd.exe / PowerShell), RTK filters work but the hook does
not auto-rewrite commands"*) and reported in #2424.

## Environment

- **RTK:** 0.43.0 (official release asset `rtk-x86_64-pc-windows-msvc.zip`,
  SHA256 `A715E989BCEBFC208F388CF5ADAAA9953CBF1127B081BC09C4EF02E7D7FEA39F`)
- **Copilot CLI:** 1.0.73
- **OS:** Windows 11 Pro 26200
- Hook file: the one generated by `rtk init -g --copilot`
  (`~/.copilot/hooks/rtk-rewrite.json`, declaring both `PreToolUse` and
  `preToolUse`)

## Evidence

Payload actually delivered to the **camelCase** `preToolUse` entry, captured
from a live Copilot CLI session with a logging hook:

```json
{"sessionId":"c1b3f427-...","timestamp":1784807355614,"cwd":"C:\\Windows\\System32",
 "toolName":"powershell","toolArgs":"{\"command\":\"rtk gain\",\"description\":\"...\"}"}
```

`toolArgs` **is** the JSON-encoded string RTK expects — only `toolName` differs
(`powershell`, not `bash`).

Payload delivered to the **PascalCase** `PreToolUse` entry in the same session:

```json
{"hook_event_name":"PreToolUse","session_id":"...","cwd":"C:\\Windows\\System32",
 "tool_name":"Bash","tool_input":{"command":"git status --short"}}
```

Resulting behavior, reproducible with the shipped binary:

```console
$ echo '{"toolName":"powershell","toolArgs":"{\"command\":\"git status\"}"}' | rtk hook copilot
                                     # (empty — PassThrough, native path skipped)

$ echo '{"tool_name":"Bash","tool_input":{"command":"git status"}}' | rtk hook copilot
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"ask",
 "permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk git status"}}}
```

## Root cause

`src/hooks/hook_cmd.rs`, `detect_format`:

```rust
// Copilot CLI: camelCase keys, toolArgs is a JSON-encoded string
    if tool_name == "bash" {        // ← never true on Windows
```

## Suggested fix

Accept Copilot CLI's Windows shell tool name:

```rust
if matches!(tool_name, "bash" | "powershell") {
```

A patch with a regression test (`test_detect_copilot_cli_powershell`) is
attached / in the linked PR.

## Note on interaction with #3037

The two are complementary. With this fix the native path is reached on Windows
and returns `modifiedArgs` without `permissionDecision`, but the Claude-compat
`PreToolUse` entry generated by `rtk init` still answers `ask` for the same
tool call. Unless #3037 is fixed as well (or the generated hook file stops
declaring both entries), Windows users keep getting a prompt on every command.

## Current workaround

Wrap the hook and rewrite the decision before Copilot sees it:

```json
"command": "rtk hook copilot | sed 's/\"permissionDecision\":\"ask\"/\"permissionDecision\":\"allow\"/'"
```

On Windows the same thing is done from a small PowerShell wrapper, which must
also handle the fact that Windows PowerShell 5.1 prepends a UTF-8 BOM to piped
stdin (RTK already handles that itself — see
`test_copilot_bom_prefixed_payload_is_recognized`).

romain-cureau:
## Summary

- `detect_format` matched the native Copilot CLI payload only on `toolName == "bash"`. Copilot CLI names its shell tool **`powershell`** on Windows, so the payload fell through to `PassThrough` and the `HookFormat::CopilotCli` branch was unreachable there.
- Consequence: on Windows only the Claude-compat `PreToolUse` entry answers, via `HookFormat::VsCode`, which emits `permissionDecision: "ask"` — a permission prompt on **every** rewritten command (see #3037), and one the allowlist cannot suppress since only `allow`/`deny` short-circuit Copilot's permission flow.
- This also matches the Windows limitation documented in the README (*"On native Windows (cmd.exe / PowerShell), RTK filters work but the hook does not auto-rewrite commands"*) and reported in #2424.

Fixes #3178.

## Test plan

- [x] `cargo fmt --all -- --check` ✅ · `cargo clippy --all-targets` ✅ (no warnings) · `cargo test --all` — `hooks::` subset 396/396 green, new `test_detect_copilot_cli_powershell` passes

  Run locally on rustc 1.97.1 / stable-x86_64-pc-windows-msvc. `fmt` caught a width overflow in my own patch, fixed in `10ca886`. `cargo test --all` reports 16 failures, all in `core::stream::tests` (`program not found` — they spawn Unix binaries); verified identical on unmodified `origin/develop`, so pre-existing on Windows and unrelated. Details in the comments below.

- [x] Manual verification against a live Copilot CLI session

  Captured with a logging hook wired to the camelCase `preToolUse` entry, Copilot CLI 1.0.73 / Windows 11, RTK 0.43.0 (official release asset):

  ```json
  {"sessionId":"c1b3f427-...","timestamp":1784807355614,"cwd":"C:\\Windows\\System32",
   "toolName":"powershell","toolArgs":"{\"command\":\"rtk gain\",\"description\":\"...\"}"}
  ```

  `toolArgs` is already the JSON-encoded string RTK expects — only `toolName` differed. And with the shipped binary:

  ```console
  $ echo '{"toolName":"powershell","toolArgs":"{\"command\":\"git status\"}"}' | rtk hook copilot
                                       # empty: PassThrough, native path skipped

  $ echo '{"tool_name":"Bash","tool_input":{"command":"git status"}}' | rtk hook copilot
  {"hookSpecificOutput":{...,"permissionDecision":"ask",...,"updatedInput":{"command":"rtk git status"}}}
  ```

## Note

This is complementary to #3037, not a replacement. With this fix the native path is reached on Windows and returns `modifiedArgs` without a `permissionDecision`, but the `PreToolUse` entry that `rtk init` also writes still answers `ask` for the same tool call. Unless #3037 lands as well (or the generated hook file stops declaring both entries), Windows users keep getting a prompt on every command.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -87,9 +87,10 @@ fn detect_format(v: &Value) -> HookFormat {
         return HookFormat::PassThrough;
     }
 
-    // Copilot CLI: camelCase keys, toolArgs is a JSON-encoded string
+    // Copilot CLI: camelCase keys, toolArgs is a JSON-encoded string.
+    // The shell tool is "bash" on Unix and "powershell" on Windows.
     if let Some(tool_name) = v.get("toolName").and_then(|t| t.as_str()) {
-        if matches!(tool_name, "bash" | "run_in_terminal") {
+        if matches!(tool_name, "bash" | "powershell" | "run_in_terminal") {
             if let Some(tool_args_str) = v.get("toolArgs").and_then(|t| t.as_str()) {
                 if let Ok(tool_args) = serde_json::from_str::<Value>(tool_args_str) {
                     if let Some(cmd) = tool_args
@@ -724,9 +725,9 @@ mod tests {
         })
     }
 
-    fn copilot_cli_input(cmd: &str) -> Value {
+    fn copilot_cli_input(tool: &str, cmd: &str) -> Value {
         let args = serde_json::to_string(&json!({ "command": cmd })).unwrap();
-        json!({ "toolName": "bash", "toolArgs": args })
+        json!({ "toolName": tool, "toolArgs": args })
     }
 
     fn copilot_ide_input(cmd: &str) -> Value {
@@ -758,7 +759,7 @@ mod tests {
     #[test]
     fn test_detect_copilot_cli_bash() {
         assert!(matches!(
-            detect_format(&copilot_cli_input("git status")),
+            detect_format(&copilot_cli_input("bash", "git status")),
             HookFormat::CopilotCli { .. }
         ));
     }
@@ -768,6 +769,15 @@ mod tests {
         assert!(matches!(
             detect_format(&copilot_ide_input("git status")),
             HookFormat::CopilotIde { .. }
+        ));
+    }
+
+    #[test]
+    fn test_detect_copilot_cli_powershell() {
+        // Copilot CLI names its shell tool "powershell" on Windows, not "bash".
+        assert!(matches!(
+            detect_format(&copilot_cli_input("powershell", "git status")),
+            HookFormat::CopilotCli { .. }
         ));
     }
 
@@ -783,8 +793,11 @@ mod tests {
         // (confirmed for Cursor). run_copilot strips them before parsing;
         // verify both Copilot formats still parse after the same handling.
         for raw in [
-            format!("\u{feff}{}", copilot_cli_input("git status")),
-            format!("\u{feff}\u{feff}{}", copilot_cli_input("git status")),
+            format!("\u{feff}{}", copilot_cli_input("bash", "git status")),
+            format!(
+                "\u{feff}\u{feff}{}",
+                copilot_cli_input("bash", "git status")
+            ),
         ] {
             let cleaned = strip_leading_bom(&raw).trim();
             let v: Value = serde_json::from_str(cleaned).expect("BOM-stripped JSON must parse");


src/hooks/hook_cmd.rs
```diff
@@ -706,6 +707,19 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn test_detect_copilot_cli_powershell() {
+        // Copilot CLI on Windows names its shell tool "powershell", not "bash".
+        // Without this the payload falls through to PassThrough and only the
+        // Claude-compat PreToolUse entry answers, which prompts on every command.
+        let args = serde_json::to_string(&json!({ "command": "git status" })).unwrap();
+        let input = json!({ "toolName": "powershell", "toolArgs": args });
```
KuSh: nitpick: update `copilot_cli_input` to take a tool like vscode_input does and use it here

KuSh (changes requested):
Hi @romain-cureau, thanks for your contribution, LGTM apart from one minor nitpick!

pszymkowiak:
Thanks for tracking this down — since you've got Copilot CLI running on Windows already, could you help confirm something that affects whether this stays needed long-term?

Right now `rtk init --copilot` registers *two* hook entries for Copilot CLI: a PascalCase `PreToolUse` and a camelCase `preToolUse`. Copilot CLI apparently fires both, independently, on every tool call (that's the redundant-double-hook bug #3212 separately fixes). Your fix here targets the camelCase one (`toolName: "powershell"`).

Question: on your Windows setup, was the *other* hook (`PreToolUse`, PascalCase) already receiving and correctly handling your terminal commands before this fix, in parallel with the broken camelCase one? Per [GitHub's hooks reference](https://docs.github.com/en/copilot/reference/hooks-reference), that payload remaps both `bash` and `powershell` to a single `tool_name: "Bash"` — a value rtk already recognizes there. If that hook was already firing correctly for you, it'd mean the prompt-every-time symptom you saw was likely coming only from the camelCase side being silently ignored (your fix is still correct either way) rather than Windows being entirely unhandled.

Asking because #3212 removes the camelCase registration entirely in favor of the PascalCase one — if that one already works standalone on Windows, this fix's code path becomes legacy-only once #3212 lands and someone re-runs `rtk init --copilot`. Not asking for changes here, just want to get the full picture before the two PRs land.

romain-cureau:
@KuSh nitpick addressed — `copilot_cli_input` now takes a `tool` argument like `vscode_input`, existing call sites pass `"bash"`, and the new test uses it. Also rebased onto current `develop`, so the `run_in_terminal` arm added by #3149/`fix/2443-copilot-ide-terminal` is preserved: `matches!(tool_name, "bash" | "powershell" | "run_in_terminal")`.

@pszymkowiak — yes, and I tested it rather than inferring it. **The PascalCase hook was already firing and being honored on Windows, in parallel with the ignored camelCase one.** Two pieces of evidence, both from Copilot CLI 1.0.73 on Windows 11.

**1. It receives the call.** Both entries fire for the same tool call; the payloads differ exactly as you describe:

```jsonc
// camelCase preToolUse — toolName is the real Windows tool
{"sessionId":"...","cwd":"C:\\Windows\\System32",
 "toolName":"powershell","toolArgs":"{\"command\":\"rtk gain\",...}"}

// PascalCase PreToolUse — same session, remapped to "Bash"
{"hook_event_name":"PreToolUse","session_id":"...","cwd":"C:\\Windows\\System32",
 "tool_name":"Bash","tool_input":{"command":"git status --short"}}
```

**2. Its `updatedInput` is actually applied.** I registered a throwaway PascalCase hook that rewrites one marker command to another and returns `permissionDecision: "allow"`:

```jsonc
// hook sees:      echo PROBE_INPUT_5521
// hook returns:   {"hookSpecificOutput":{...,"permissionDecision":"allow",
//                   "updatedInput":{"command":"echo REWRITE_APPLIED_9137"}}}
```

Session output was `REWRITE_APPLIED_9137`. So on Windows the PascalCase path rewrites end-to-end, and `updatedInput` is honored there — the symptom I hit was purely `handle_vscode` asserting `ask`, which matches your analysis in #3212.

**Correction to my own issue text:** I wrote in #3178 that the unreachable native path "explains the Windows limitation documented in the README" and cross-posted that on #2424. That was wrong for Copilot CLI and I've retracted it there — the rewrite *does* happen on Windows, via the Claude-compat entry. Whatever #2424 is hitting is something else. Sorry for the noise.

So yes: once #3212 lands and users re-run `rtk init --copilot`, this arm is legacy-only — it still matters for anyone who doesn't re-init, and it keeps `detect_format` honest about what Copilot CLI actually sends, but it's no longer load-bearing. Happy to close it in favour of #3212 if you'd rather keep the surface small.

Test plan status unchanged and worth restating: `cargo fmt/clippy/test` still **not run** on my side (no Rust toolchain on the Windows box this was diagnosed on, and CI here is `action_required` pending maintainer approval). The change is one match arm plus a test helper signature, but please don't take my word for the build.

romain-cureau:
Retracting my earlier "test plan not run" caveat — I installed a Rust toolchain and ran the CI commands locally. Pushed as `10ca886`.

`cargo fmt --all -- --check` caught a real issue in my own patch: adding the `tool` argument pushed the second `copilot_cli_input` call in `test_copilot_bom_prefixed_payload_is_recognized` over the width limit. Fixed and folded into the commit.

| CI command | Result |
|---|---|
| `cargo fmt --all -- --check` | ✅ exit 0 |
| `cargo clippy --all-targets` | ✅ exit 0, no warnings |
| `cargo test --all` | ⚠️ 2483 passed / 16 failed — **all pre-existing on Windows, none related** (see below) |
| `hooks::` subset | ✅ 396 passed, 0 failed |
| new `test_detect_copilot_cli_powershell` | ✅ passes |

The 16 failures are all in `core::stream::tests`, failing with `Failed to spawn process / program not found` — they shell out to Unix binaries that don't exist on Windows. To be sure they aren't mine, I ran the same subset on a detached checkout of unmodified `origin/develop` (`1d6798d`):

```console
# origin/develop, not a line of this PR applied
$ cargo test --bin rtk core::stream
test result: FAILED. 18 passed; 16 failed; ...

# this branch
$ cargo test --bin rtk core::stream
test result: FAILED. 18 passed; 16 failed; ...
```

Identical, so `cargo test --all` is simply not green on a Windows host today. Worth knowing given this is the Windows-facing code path — happy to file that separately if it's not already known.

The security-lint job's `expect()`/`panic!` hits are also pre-existing and none are in `hook_cmd.rs` (that step is `::warning::`-only, non-blocking).

Environment: rustc 1.97.1, stable-x86_64-pc-windows-msvc, Windows 11.

KuSh (approved):

PR #3181, opened 2026-07-23
chore: declare RtkRule default to avoid repeating unecessary attributes

KuSh:
## Summary
Simplify RtkRule declarations by using default values

## Test plan

### DIFF
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -12,11 +12,33 @@ pub struct RtkRule {
     pub subcmd_status: &'static [(&'static str, RtkStatus)],
 }
 
+impl RtkRule {
+    // `Default::default()` isn't a const fn on stable, so `RULES` (a const array) can't call
+    // it via `..Default::default()` — hence this associated const as the const-context
+    // workaround. Once `const_trait_impl` stabilizes, drop this and derive/impl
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Verified this is behavior-preserving on a compiled build:

- **Static diff check**: only default-valued fields were dropped — every removed `savings_pct` is `60.0`, every removed `pipeline_final_safe` is `false`, and no non-empty `category`/`rewrite_prefixes`/`subcmd_*` was removed. Crucially, the `pipeline_final_safe: true` rules (grep/rg) are kept explicit.
- **Rewrite parity**: a battery of 20 varied commands (grep/rg final-safe, `cat|grep`, `cat|wc` passthrough, cargo/npm/docker/uv/go/kubectl…) rewrites **byte-identically** to `origin/develop`.
- **Tests**: full suite **2543 passed / 0 failed**, rules (12) + discover (471) green, clippy clean. The `const DEFAULT` associated-const trick (since `Default::default()` isn't const) compiles fine and the `..RtkRule::DEFAULT` call sites all resolve.

Nice cleanup — cuts a lot of repetition in `rules.rs` without touching routing or savings metadata. LGTM from my side.

TaKO8Ki (approved):
🚀

TaKO8Ki:
The CI failure is not related to this pr.

PR #3188, opened 2026-07-24
feat(rewrite): rewrite each line of multi-line Bash blocks

revopsrocks:
## Summary

- Multi-line Bash blocks (sequential commands on separate lines — a shape coding agents emit constantly) were rewritten as if they were one command: the `rtk` prefix landed on the first line and every following line ran raw and unfiltered. Implements the line-by-line rewrite proposed in #1243.
- Splits at the newline tokens the existing quote-aware lexer emits (`tokenize_inner(_, true)`, already used by `split_for_permissions`), so newlines inside quoted strings — e.g. multi-line commit messages — are never split points. Each line is then rewritten through the existing single-line path; blank lines, comments, indentation, and CRLF separators are preserved verbatim.
- Deliberately conservative, per the concerns in #2792/#2902: the whole block passes through untouched when any line opens control flow (`for`/`if`/`while`/`case`/…), a list or pipeline continues across the line break (`&&`, `||`, `|`, `|&` at a line edge), a subshell/group spans lines, or the block contains a heredoc (existing gate). The heredoc *output-filtering* idea in #1243 is out of scope here.

Example:

```
$ rtk hook check 'git status
# check history
  grep -rn foo src'
rtk git status
# check history
  rtk grep -rn foo src
```

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` (2559 passed)
- [x] Manual testing: `rtk hook check` output inspected for sequential lines, indented/comment/blank lines, CRLF, for/if blocks, cross-line `&&` and `|`/`|&` continuations, quoted multi-line strings, heredocs
- [x] 16 new unit tests (`registry::tests::multiline_blocks`, lexer newline-token test) covering the rewrite and every passthrough gate

Ref #1243

### DIFF
--- a/src/discover/lexer.rs
+++ b/src/discover/lexer.rs
@@ -26,6 +26,13 @@ pub fn tokenize(input: &str) -> Vec<ParsedToken> {
     tokenize_inner(input, false)
 }
 
+/// Like [`tokenize`] but emits a `\n` operator token for each newline that
+/// sits outside quotes. Newlines inside quoted strings stay part of their
+/// argument, so callers can use the emitted offsets as safe line-split points.
+pub fn tokenize_with_newlines(input: &str) -> Vec<ParsedToken> {
+    tokenize_inner(input, true)
+}
+
 fn tokenize_inner(input: &str, emit_newline: bool) -> Vec<ParsedToken> {
     let mut tokens = Vec::new();
     let mut current = String::new();
@@ -1341,5 +1348,18 @@ mod tests {
     fn test_split_perms_empty() {
         assert!(split_for_permissions("").is_empty());
         assert!(split_for_permissions("   ").is_empty());
+    }
+
+    #[test]
+    fn test_tokenize_with_newlines_emits_operator_outside_quotes_only() {
+        let newline_ops = |input: &str| {
+            tokenize_with_newlines(input)
+                .iter()
+                .filter(|t| t.kind == TokenKind::Operator && t.value == "\n")
+                .count()
+        };
+        assert_eq!(newline_ops("git status\ngit log"), 1);
+        assert_eq!(newline_ops("echo 'line1\nline2'"), 0);
+        assert_eq!(newline_ops("git status\r\ngit log"), 2);
     }
 }
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -634,6 +634,9 @@ fn line_breaks_independence(line: &str) -> bool {
         || line.ends_with('{')
         || line.starts_with(')')
         || line.starts_with('}')
+        || line.starts_with("((")
+        || line.ends_with("))")
+        || line.ends_with("((")
 }
 
 /// Rewrite each line of a multi-line block independently (issue #1243).
@@ -644,6 +647,14 @@ fn line_breaks_independence(line: &str) -> bool {
 /// line is an independent command (see [`line_breaks_independence`]); blank
 /// lines and comment lines are preserved verbatim, as is indentation and the
 /// original separator bytes (`\n` vs `\r\n`).
+///
+/// If any newline byte was swallowed by quote state, the block passes through
+/// untouched. The lexer has no comment awareness, so an apostrophe in a `#`
+/// comment opens quote state and hides the rest of the block — rewriting (or
+/// prefixing) such a block would act on lines no permission verdict was
+/// computed for. Passthrough hands the original command to the agent's native
+/// permission handling instead. Genuine quoted newlines (multi-line commit
+/// messages) also land here; forgoing that rewrite is the safe trade.
 fn rewrite_multiline_block(
     cmd: &str,
     excluded: &[ExcludePattern],
@@ -655,9 +666,9 @@ fn rewrite_multiline_block(
         .map(|t| t.offset)
         .collect();
 
-    // Newlines only inside quotes — a single logical command.
-    if newline_offsets.is_empty() {
-        return rewrite_single(cmd, excluded, transparent_prefixes);
+    let raw_breaks = cmd.chars().filter(|c| matches!(c, '\n' | '\r')).count();
+    if raw_breaks != newline_offsets.len() {
+        return None;
     }
 
     let mut segments = Vec::with_capacity(newline_offsets.len() + 1);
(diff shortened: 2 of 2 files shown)


revopsrocks:
Pushed a hardening commit (aaa123c) after running an adversarial review against the branch. Two gaps closed:

- **Swallowed newlines now force passthrough.** The lexer has no comment state, so an apostrophe inside a `#` comment opens quote state and swallows the following newline — hiding later lines from the splitter. The block now passes through whenever the raw `\n`/`\r` count differs from the emitted newline-token count, so native permission handling always sees the original command. (This check is deliberately global, not per-segment.) Quoted multi-line strings — e.g. multi-line commit messages — also take this path now; forgoing that rewrite is the conservative trade.
- **Cross-line arithmetic bails.** `(( x =` / `ls ))` previously had `rtk` spliced into arithmetic context; `((`/`))` at a line edge now triggers passthrough.

Both cases have tests (`multiline_blocks::test_comment_apostrophe_*`, `test_arithmetic_spanning_lines_passes_through`), plus a positive test that a comment with balanced quotes still rewrites. Full gate re-run: `cargo fmt --check` / `clippy --all-targets` / `cargo test` (2563 passed).

aeppling:
hey, i think this is addressing same enhancement as https://github.com/rtk-ai/rtk/pull/2781

revopsrocks:
Second adversarial review round (two independent reviewers) — three more gaps closed in 882d561:

- **Array literals / function bodies spanning lines**: `arr=(one` ... `two)` slipped past the edge checks (the parens are mid-line), so the inner line — an array *element* — got an `rtk` prefix. Now any line whose unquoted `(`/`)` or `{`/`}` don't balance bails the whole block; this also catches `foo() {` function definitions and subsumes the earlier single-char edge checks.
- **Continuation operators behind trailing comments**: `git log | # keep pipeline` ends in comment text, so the `ends_with("|")` check missed that bash continues the pipeline across the newline — and rewriting the next line could rebuild a pipeline whose final stage the pipeline-safety guard specifically rejects (e.g. `grep -f`). Independence checks now run against the line with any trailing unquoted comment stripped.
- **ANSI-C `$'...'` quoting**: inside `$'...'` bash treats `\'` as a literal quote that does *not* close the string; the lexer thinks it does, and emits a split point bash would never honor — the exact inverse of the swallowed-newline case, invisible to the count check. `rtk` could be spliced into the *contents* of a string literal. Any block containing `$'` now passes through.

The review also re-probed everything previously fixed (comment-apostrophe swallowed newlines, cross-line arithmetic, CRLF/lone-CR byte round-trips, UTF-8 offset slicing, `${HOME}` brace balance, `time`/`!` prefixes) — all hold. Tests added for each new case, including positives that balanced grouping and quoted/unquoted `${...}` still rewrite. Gate re-run: fmt / clippy / `cargo test` (2568 passed).

One divergence found during review is pre-existing and outside this branch's diff (`collapse_line_continuations` treats `\\<newline>` — escaped backslash + separator newline — as a continuation); filing separately.

revopsrocks:
@aeppling You're right, thanks — same territory (#1243), and I'd missed #2781 when I searched for prior art (I grepped for "multiline", your title says "bundles" — my bad).

The two PRs differ mainly in risk posture. #2781 goes for coverage: statement-level rewriting inside bundles that also contain pipes/substitutions/redirects, with the #1155 ask-clamp as the safety net (39.3% → 71.8% is a big win). #3188 is deliberately narrow: strictly line-level, and any construct that makes lines non-independent — or that the lexer can't attest — forces whole-block passthrough. Lower coverage, but the rewrite can never touch a line the splitter misjudged.

Since we'd just built an adversarial corpus for #3188 (two independent review rounds), I ran it against the #2781 head (2c89559). Four inputs currently rewrite in ways worth a look — all pass `bash -n`, which is why differential fuzz keyed on syntax validity wouldn't flag them:

1. **Comment apostrophe swallows the newline** — bash's word-start `#` comment rule vs a lexer with no comment state:
   ```
   git status # don't
   rm -rf /tmp/x
   ```
   → `rtk git status # dont'` + the second line preserved raw. The apostrophe opens quote state, so the splitter never "saw" line 2 — worth checking how the ask-clamp treats lines that quote-state hid from `split_for_permissions`.
2. **Continuation operator behind a trailing comment**:
   ```
   git log | # keep pipeline
   grep -f patterns.txt
   ```
   → `rtk grep -f patterns.txt` becomes the pipeline's final stage — which develop's own pipeline-final guard specifically rejects for `grep -f` (pattern file vs stdin).
3. **ANSI-C quoting** — inside `$'...'`, `\'` is a literal quote that does *not* close the string:
   ```
   x=$'foo\'
   git status
   '
   ```
   → `rtk git status` is spliced into the *value of the string literal* (silent data corruption; `git status` here is content, not a command).
4. **Function bodies** get rewritten (`foo() { rtk git status; }`) — possibly intended given the coverage philosophy, but it changes captured output for `x=$(foo)` call sites.

#3188 now passes all of these by bailing to passthrough (the fix commits document each one). Happy for these to just become fuzz-corpus additions to #2781 if maintainers prefer the higher-coverage direction — the tests transfer either way, and I'd rather one good implementation land than two compete. Maintainers' call.

revopsrocks:
The failing `CI / benchmark` check is the known non-deterministic live-HTTP case (#2844): `curl json` against `mockhttp.org/json/1` came back at 26 raw tokens this run, so the filter's 61-token output tripped the never-worse guard (`-134%`). This branch doesn't touch the curl filter (or any filter) — the diff is confined to the rewrite path (`registry.rs`) and lexer, and all `rewrite` benchmark cases pass. All other checks are green. Could a maintainer re-run the benchmark job when convenient?

aeppling:
Hey @revopsrocks , thanks for addressing this, here is a review, some of those case are not likely to be ran by agent, but if possible to have a solid implementation that cover most of case

### Narrow the $' bail

One unrelated line currently disables the whole block:

```
kubectl get pods
git status
echo $'a\tb'          -> disables the other two
```

Only \' inside a $'…' span is actually hazardous, the rest rewrite normally.

### Narrow the cross-line && / || / | bail

develop rewrites this correctly and completely today; the PR declines it:

```
git status &&
git log -3
```

develop → rtk git status && rtk git log -3
this PR → passthrough

### comment_start should recognise "#" after an operator

Bash starts a comment wherever # begins a word, including right after |, &&, ;, (, ) — not only after whitespace:
Comment starts when the preceding non-space byte is whitespace, start-of-line, or one of | & ; ( ).

### Guard the heredoc gate against line continuations

collapse_line_continuations joins \<NL> with a space where bash joins with nothing, so << stops existing and has_heredoc() misses it:

```
cat <\
<EOF
git status
EOF
```
= cat < <EOF / rtk git status / EOF → syntax error near '<'

No real-world occurrences produced by agents, but the fix is one line: run has_heredoc() on the raw input, before collapse_line_continuations.

```rust
#[test]
fn test_heredoc_split_by_line_continuation_passes_through() {
    assert_eq!(
        rewrite_command_no_prefixes("cat <\\\n<EOF\ngit status\nEOF", &[]),
        None
    );
}
```

KuSh:
Thanks for building this out. The line-by-line multiline rewrite landed on develop in #3319 (merged 2026-08-03), covering the same split-at-newline-tokens approach and the same passthrough gates for control flow, cross-line lists and pipelines, grouping and heredocs, along with a few extra ones. Running the develop binary on the example from this PR gives exactly the output shown here, and the lexer test added here is present on develop under the same name, so nothing from this PR looks missing. Please comment or reopen if you find a block shape develop still mishandles.

PR #3203, opened 2026-07-25
fix(init): strip UTF-8 BOM when parsing hand-edited JSON config files

kylehgc:
## Bug

`rtk init -g --auto-patch` aborts with exit 1 on a `settings.json` that begins with a UTF-8 BOM:

```
rtk: Failed to parse C:\...\settings.json as JSON: expected value at line 1 column 1
```

Notepad and PowerShell 5.1's `Out-File -Encoding utf8` / `>` both write a BOM by default, so any Windows user who hand-edits `~/.claude/settings.json` loses RTK hook installation — with an error that blames their (valid) JSON. Two aggravating details:

- init prints "RTK hook registered (global)" *before* failing, so the output claims success while the hook is absent.
- The statusline patch path (`serde_json::from_str(&content).unwrap_or(json!({}))` in `init.rs`) swallows the parse failure and silently **wipes existing settings** on write-back.
- `permissions.rs` silently drops the user's deny/ask/allow rules, and `hook_check.rs` misreports the hook as not installed.

## Repro

```powershell
$sandbox = Join-Path $env:TEMP "rtk-bom"; New-Item -ItemType Directory -Force $sandbox | Out-Null
$env:CLAUDE_CONFIG_DIR = $sandbox
$json = '{"hooks":{"PreToolUse":[{"matcher":"Bash","hooks":[{"type":"command","command":"echo hi"}]}]}}'
[System.IO.File]::WriteAllText((Join-Path $sandbox "settings.json"), $json, (New-Object System.Text.UTF8Encoding($true)))
rtk init -g --auto-patch   # before: EXIT=1, hook not installed; after: EXIT=0, hook added, echo-hi entry preserved
```

Verified before/after with real binaries on Windows 11.

## Fix

Reuse, not new code: `strip_leading_bom` already exists in `hook_cmd.rs`, guarding the hook **stdin** entry points (Cursor on Windows ships BOM-prefixed payloads). The file-read paths never got the same treatment. This PR:

- Promotes the helper to `src/core/utils.rs` (`pub`), moves its unit test along; stdin behavior unchanged (existing BOM stdin tests still pass).
- Applies it at every site that parses JSON a human may have hand-edited: 12 `settings.json`/`hooks.json` sites in `init.rs`, `hook_check.rs`, `integrity.rs`, `permissions.rs` (×2), plus `composer.json` (`core/utils.rs`) and `global.json` (`dotnet_cmd.rs`).

Deliberately untouched:
- **TOML sites** (`config.rs`, filter DSL): the `toml` crate tolerates a leading BOM natively — pinned by a new regression test so a crate upgrade can't silently regress it.
- **BOM on write-back is dropped**: RFC 8259 §8.1 forbids adding a BOM to JSON.

## Tests

- `test_patch_settings_json_tolerates_utf8_bom` — BOM + valid JSON: init succeeds, existing keys survive (written red-first; failed with the exact repro error)
- `test_patch_settings_json_bom_only_file` — U+FEFF is not whitespace, so the `trim().is_empty()` guard alone misses a BOM-only file
- `test_patch_settings_json_bom_plus_invalid_json_still_errors` — stripping must not mask genuinely broken JSON
- `test_toml_parse_tolerates_utf8_bom` — pins the toml crate behavior
- `test_strip_leading_bom_helper` — moved with the helper (empty, single/double/triple BOM, mid-string BOM preserved)

## Quality gate

On this branch (off `develop`), Windows x86_64-pc-windows-msvc:

```
cargo fmt --all            → clean
cargo clippy --all-targets → zero warnings
cargo test --all           → 2493 passed, 0 failed (8 ignored)
```

### DIFF
--- a/src/core/utils.rs
+++ b/src/core/utils.rs
@@ -418,7 +418,7 @@ fn composer_bin_dirs_from(env_bin_dir: Option<&str>, composer_json: Option<&str>
 }
 
 fn read_composer_bin_dir(composer_json: &str) -> Option<PathBuf> {
-    let parsed: Value = serde_json::from_str(composer_json).ok()?;
+    let parsed: Value = serde_json::from_str(strip_leading_bom(composer_json)).ok()?;
     let bin_dir = parsed.get("config")?.get("bin-dir")?.as_str()?.trim();
     if bin_dir.is_empty() {
         None
--- a/src/core/utils.rs
+++ b/src/core/utils.rs
@@ -68,6 +68,12 @@ pub fn strip_leading_bom(input: &str) -> &str {
     s
 }
 
+/// BOM-tolerant `serde_json::from_str`. Prefer this over `serde_json::from_str`
+/// anywhere the JSON may have been written by a human, an editor, or another tool.
+pub fn from_json_str<'a, T: serde::Deserialize<'a>>(s: &'a str) -> serde_json::Result<T> {
+    serde_json::from_str(strip_leading_bom(s))
+}
+
 /// Executes a command and returns cleaned stdout/stderr.
 ///
 /// # Arguments
@@ -418,7 +424,7 @@ fn composer_bin_dirs_from(env_bin_dir: Option<&str>, composer_json: Option<&str>
 }
 
 fn read_composer_bin_dir(composer_json: &str) -> Option<PathBuf> {
-    let parsed: Value = serde_json::from_str(strip_leading_bom(composer_json)).ok()?;
+    let parsed: Value = from_json_str(composer_json).ok()?;
     let bin_dir = parsed.get("config")?.get("bin-dir")?.as_str()?.trim();
     if bin_dir.is_empty() {
         None
@@ -485,6 +491,30 @@ mod tests {
         assert_eq!(strip_leading_bom("\u{feff}\u{feff}\u{feff}hello"), "hello");
         // BOM in the middle is preserved (not "leading").
         assert_eq!(strip_leading_bom("a\u{feff}b"), "a\u{feff}b");
+    }
+
+    #[test]
+    fn test_from_json_str_strips_bom() {
+        let v: Value = from_json_str("\u{feff}{\"foo\": 1}").unwrap();
+        assert_eq!(v["foo"], 1);
+    }
+
+    #[test]
+    fn test_from_json_str_no_bom() {
+        let v: Value = from_json_str("{\"foo\": 1}").unwrap();
+        assert_eq!(v["foo"], 1);
+    }
+
+    #[test]
+    fn test_from_json_str_borrowed_type() {
+        // Deserialize<'a> (not DeserializeOwned) must keep zero-copy
+        // borrowing working through the wrapper.
+        #[derive(serde::Deserialize)]
+        struct Borrowed<'a> {
+            name: &'a str,
+        }
+        let b: Borrowed = from_json_str("\u{feff}{\"name\": \"rtk\"}").unwrap();
+        assert_eq!(b.name, "rtk");
     }
 
     #[test]
--- a/src/core/utils.rs
+++ b/src/core/utils.rs
@@ -55,6 +55,19 @@ pub fn strip_ansi(text: &str) -> String {
     ANSI_RE.replace_all(text, "").to_string()
 }
 
+/// Strip one or more leading UTF-8 BOMs (`EF BB BF`, sometimes doubled),
+/// which serde_json refuses to parse. Windows editors (Notepad, PowerShell
+/// 5.1 `Out-File -Encoding utf8`) prepend one to hand-edited config files,
+/// and Cursor on Windows ships hook payloads with them. Strip defensively
+/// wherever RTK parses JSON a human or another tool may have written.
+pub fn strip_leading_bom(input: &str) -> &str {
+    let mut s = input;
+    while let Some(rest) = s.strip_prefix('\u{feff}') {
+        s = rest;
+    }
+    s
+}
+
 /// Executes a command and returns cleaned stdout/stderr.
 ///
 /// # Arguments
@@ -460,6 +473,19 @@ pub fn human_bytes(bytes: u64) -> String {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn test_strip_leading_bom_helper() {
+        // Direct unit test on the helper so future refactors can't
+        // regress the loop semantics without a clear failure signal.
+        assert_eq!(strip_leading_bom(""), "");
+        assert_eq!(strip_leading_bom("hello"), "hello");
+        assert_eq!(strip_leading_bom("\u{feff}hello"), "hello");
+        assert_eq!(strip_leading_bom("\u{feff}\u{feff}hello"), "hello");
+        assert_eq!(strip_leading_bom("\u{feff}\u{feff}\u{feff}hello"), "hello");
+        // BOM in the middle is preserved (not "leading").
+        assert_eq!(strip_leading_bom("a\u{feff}b"), "a\u{feff}b");
+    }
 
     #[test]
     fn test_truncate_short_string() {
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -7,6 +7,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 use tempfile::NamedTempFile;
 
+use crate::core::utils::strip_leading_bom;
 use crate::hooks::constants::{
     CONFIG_DIR, COPILOT_HOME_ENV, COPILOT_HOOK_FILE, COPILOT_INSTRUCTIONS_FILE, COPILOT_USER_DIR,
     CURSOR_DIR, GEMINI_DIR, GITHUB_DIR, OPENCODE_PLUGIN_FILE, OPENCODE_SUBDIR, PLUGIN_SUBDIR,
@@ -595,12 +596,13 @@ fn remove_hook_from_settings(ctx: InitContext) -> Result<bool> {
 
     let content = fs::read_to_string(&settings_path)
         .with_context(|| format!("Failed to read {}", settings_path.display()))?;
+    let content = strip_leading_bom(&content);
 
     if content.trim().is_empty() {
         return Ok(false);
     }
 
-    let mut root: serde_json::Value = serde_json::from_str(&content)
+    let mut root: serde_json::Value = serde_json::from_str(content)
         .with_context(|| format!("Failed to parse {} as JSON", settings_path.display()))?;
 
     let removed = remove_hook_from_json(&mut root);
@@ -965,11 +967,12 @@ fn patch_settings_json_command(
     let mut root = if settings_path.exists() {
         let content = fs::read_to_string(&settings_path)
             .with_context(|| format!("Failed to read {}", settings_path.display()))?;
+        let content = strip_leading_bom(&content);
 
         if content.trim().is_empty() {
             serde_json::json!({})
         } else {
-            serde_json::from_str(&content)
+            serde_json::from_str(content)
                 .with_context(|| format!("Failed to parse {} as JSON", settings_path.display()))?
         }
     } else {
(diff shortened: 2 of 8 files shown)


src/hooks/init.rs
```diff
@@ -4269,7 +4279,7 @@ fn patch_gemini_settings(
     let mut settings: serde_json::Value = if settings_path.exists() {
         let content = fs::read_to_string(&settings_path)
             .with_context(|| format!("Failed to read {}", settings_path.display()))?;
-        serde_json::from_str(&content).unwrap_or(serde_json::json!({}))
+        serde_json::from_str(strip_leading_bom(&content)).unwrap_or(serde_json::json!({}))
```
KuSh: **Medium — the data-loss half of this bug is still open.**

The PR description calls this site out for swallowing the parse failure and *silently wiping existing settings* on write-back, but only the BOM half got fixed. `unwrap_or(serde_json::json!({}))` still turns **any** other parse error into an empty object: a trailing comma, a `//` comment, a truncated file. The function then writes `to_string_pretty(&settings)` over `~/.gemini/settings.json` via `NamedTempFile::persist` (L4374-4377) with **no backup**.

Concrete failure: a user with a hand-edited `~/.gemini/settings.json` containing a trailing comma runs `rtk init -g --agent gemini --auto-patch`. Their `model`, `mcpServers` and `theme` keys are destroyed and replaced by a file containing only the RTK `BeforeTool` hook — no error printed, and nothing to restore from.

The Claude path in this same file already does the right thing: it propagates the parse error with context (L975) and takes a `.json.bak` before overwriting (L1036). Suggest matching that shape here:

```rust
serde_json::from_str(strip_leading_bom(&content))
    .with_context(|| format!("Failed to parse {}", settings_path.display()))?
```

…or, if silently continuing is deliberate for Gemini, at minimum back the file up before the `persist`. As it stands the BOM fix makes the common case work while leaving the destructive case intact.

(For what it's worth, `uninstall_gemini` just below is fine — `if let Ok(...)` skips rather than overwrites, so it's non-destructive.)

src/core/utils.rs
```diff
@@ -55,6 +55,19 @@ pub fn strip_ansi(text: &str) -> String {
...
+/// 5.1 `Out-File -Encoding utf8`) prepend one to hand-edited config files,
+/// and Cursor on Windows ships hook payloads with them. Strip defensively
+/// wherever RTK parses JSON a human or another tool may have written.
+pub fn strip_leading_bom(input: &str) -> &str {
+    let mut s = input;
+    while let Some(rest) = s.strip_prefix('\u{feff}') {
+        s = rest;
+    }
+    s
+}
```
KuSh, line 69: **Low / follow-up — three JSON parse sites still uncovered now that this helper is `pub`.**

The stated goal is "every site that parses JSON a human may have hand-edited", and these three are outside this diff but fit that description exactly:

- `src/cmds/system/deps.rs:137` — `summarize_package_json_str` does `serde_json::from_str(&content)?` on `package.json`, and the `?` at the call site (`deps.rs:46`) aborts the whole command. On Windows a BOM-prefixed `package.json` is at least as likely as the `composer.json` / `global.json` cases this PR *did* add — so `rtk deps` hard-fails on an otherwise valid project instead of degrading.
- `src/cmds/system/json_cmd.rs:95` and `:186` — `filter_json_compact` / `filter_json_string`, where `rtk json <file>` errors out on a BOM-prefixed file. `rtk json` is arguably the single most likely command to be pointed at a file a human just saved from an editor.

Neither is a regression from this diff, and I'd rather see them in a follow-up than balloon this PR — just flagging them while the helper is fresh, since each is a one-line call to the function you just made `pub`.

KuSh (changes requested):
Reviewed the full diff and verified locally on a worktree of `submit/bom-config-reads` (`0341ed5`): `cargo check --all-targets` clean, `cargo clippy --all-targets` zero warnings, `cargo test bom` 8/8 green including the 5 new tests.

The change itself is mechanical and low-risk, and the reasoning behind it holds up:

- The TOML claim is genuine — `filter_parse_error` calls `toml::from_str` directly and that *does* tolerate a leading BOM, so deliberately skipping the TOML sites is correct, and the test now pins it.
- The `let content = strip_leading_bom(&content);` shadowing at all 9 sites is sound: the borrow of the still-live `String` is fine and no site reads the pre-strip value afterwards.
- Coverage inside `src/hooks/init.rs` checks out — the remaining uncovered `read_to_string` calls there are Markdown (CLAUDE.md / AGENTS.md / copilot-instructions) or YAML (Hermes `config.yaml`), where BOM stripping doesn't apply.

Requesting changes for one thing: the PR description names `patch_gemini_settings` as silently wiping existing settings, but only the BOM half of that bug is fixed — the `unwrap_or(json!({}))` data-loss path is still there for every other parse failure. Details inline. The second comment is a lower-priority coverage note on three sites outside this diff; happy to see it split into a follow-up rather than grown into this PR.

KuSh:
Separate from the change request — a suggestion on the shape of the fix, not a blocker.

Rather than having every call site remember to wrap its argument, it might be worth putting the policy in one place:

```rust
/// BOM-tolerant `serde_json::from_str`. Prefer this over `serde_json::from_str`
/// anywhere the JSON may have been written by a human, an editor, or another tool.
pub fn from_json_str<'a, T: serde::Deserialize<'a>>(s: &'a str) -> serde_json::Result<T> {
    serde_json::from_str(strip_leading_bom(s))
}
```

Two details that make it a clean drop-in: `Deserialize<'a>` rather than `DeserializeOwned` preserves zero-copy deserialization for borrowed types (so no site silently gains an allocation), and returning `serde_json::Result<T>` keeps all three call-site shapes already in the diff working untouched — `?`, `.unwrap_or(json!({}))`, and `if let Ok(...)`. Converting a site becomes a textual swap rather than an error-handling rewrite.

`strip_leading_bom` is still worth keeping `pub` as-is — it's the primitive, and there are non-serde consumers that want the raw text (the Hermes `config.yaml` read, `from_slice`, anything hashing or diffing file contents). The wrapper is just the policy built on top of it.

The reason I'd lean this way: there are ~135 `serde_json::from_str` call sites across 32 files in `src/`, and this PR covers 9 of them. That's not a criticism of the scope — the 9 you picked are the ones that matter — but a convention that says "remember to wrap the argument" has no way to reach the other 126, and the next person parsing a config file won't know the rule exists. A named helper at least makes the correct call the discoverable one.

I'd suggest adding it here and converting only the sites this PR already touches. No need to sweep the rest in this PR.

Worth noting for whoever picks it up: the "needs BOM tolerance" line is about *provenance* (human/editor/tool-written vs. our own subprocess stdout), and it doesn't follow the directory layout. `src/hooks/hook_cmd.rs` has 38 sites that are all hook stdin — and it already documents Cursor shipping BOMs there — while `src/cmds/cloud/aws_cmd.rs` has 28 that are all `aws` CLI output and need nothing.

Optional, and definitely a separate PR if anyone wants it: once the helper exists, a `clippy.toml` `disallowed-methods` entry for `serde_json::from_str` with per-module `#![allow(...)]` opt-outs on the subprocess-parsing modules would make the convention enforced rather than remembered. Out of scope here — just flagging it while the context is fresh.

kylehgc:
Thanks for the review — all three points addressed as of dd867b2.

**The blocker (Gemini data-loss path):** `patch_gemini_settings` now mirrors the Claude path's shape — empty/whitespace-only content still yields `{}`, but any real parse failure propagates with `.with_context()` naming the file, so the destructive write-back is unreachable on bad input. A backup file turned out unnecessary once the error propagates: the only route to the `persist` is a successful parse. Regression test (`test_patch_gemini_settings_malformed_json_errors_without_overwriting`) pins both the error context and that the file on disk stays byte-identical.

**The `from_json_str` suggestion:** adopted verbatim, including the `Deserialize<'a>` zero-copy detail (pinned by a borrowed-type test) and the `serde_json::Result` return. Every JSON parse site in this PR's diff now routes through it — the nine shadowing-pattern sites plus the inline sites in `init.rs`, `hook_check.rs`, `permissions.rs`, `integrity.rs`, `dotnet_cmd.rs`, and the composer.json helper — so the discoverable helper is the convention, not a per-site habit. `strip_leading_bom` stays `pub` as the primitive for the non-serde consumers you listed.

**The three out-of-diff sites** (`deps.rs`, `json_cmd.rs` ×2) are split into #3751 as suggested, stacked on this branch so it can pick up the wrapper.

Gate: fmt clean, clippy zero warnings, 2497 tests passing.

src/hooks/init.rs
KuSh: Verified on dd867b2 — the destructive path is gone. Parse failures now propagate with `.with_context()` instead of collapsing into `{}`, so `persist` is only reachable after a successful parse, and the write-back is a superset of what was read. Agreed that a backup is unnecessary once the error propagates.

The explicit `strip_leading_bom` before the `trim().is_empty()` check is a nice catch — U+FEFF is not Unicode whitespace, so a BOM-only file would otherwise miss the empty branch and error out. The regression test pins both the context string and the file staying byte-identical. Resolving.

src/core/utils.rs
KuSh: Confirmed split into #3751, which covers `deps.rs:137` and both `json_cmd.rs` sites and is stacked on this branch so it picks up `from_json_str`. That is exactly the split I was hoping for — resolving here and following up there.

src/hooks/init.rs
```diff
@@ -4269,7 +4278,14 @@ fn patch_gemini_settings(
...
         let content = fs::read_to_string(&settings_path)
             .with_context(|| format!("Failed to read {}", settings_path.display()))?;
-        serde_json::from_str(&content).unwrap_or(serde_json::json!({}))
+        let content = strip_leading_bom(&content);
+
+        if content.trim().is_empty() {
+            serde_json::json!({})
+        } else {
+            from_json_str(content)
+                .with_context(|| format!("Failed to parse {} as JSON", settings_path.display()))?
```
KuSh, line 4287: **Non-blocking** — a consequence of the fix I asked for, so flagging the trade-off rather than requesting a change.

This now hard-errors where it previously fell back to `{}`, which is the right call. But it lands at the *end* of `run_gemini`: the hook script, `GEMINI.md`, and the `integrity::store_hash` baseline are all written before this point. So a user whose `.gemini/settings.json` has `//` comments or a trailing comma — both of which Gemini CLI's own loader tolerates — goes from exit 0 to exit 1 with a half-installed state, where previously the install completed (having silently reset their settings).

Strictly better than the silent wipe, so no objection to shipping as is. But if you'd rather not regress the exit code, warning and skipping just the settings patch would keep the rest of the install coherent:

```rust
match from_json_str(content) {
    Ok(v) => v,
    Err(e) => {
        eprintln!("rtk: skipping {} patch: {e}", settings_path.display());
        return Ok(false); // or whatever "not patched" is here
    }
}
```

Either way the destructive path is gone, which was the actual bug. Your call.

KuSh (approved):
Approving — `dd867b2` addresses everything from the previous review.

Re-verified on the new head:

- **Gemini data-loss path is closed.** Parse failures propagate with `.with_context()` instead of collapsing into `{}`, so `persist` is only reachable after a successful parse and the write-back is strictly a superset of what was read. Agreed that a backup file becomes unnecessary once the error propagates. The regression test pins both the error context and the file staying byte-identical.
- **`from_json_str` adopted** with the `Deserialize<'a>` signature (zero-copy preserved, pinned by the borrowed-type test) and the `serde_json::Result` return, so every converted site kept its prior error/fallback semantics.
- **The three out-of-diff sites** are split into #3751, which I confirmed converts `deps.rs:137` and both `json_cmd.rs` sites to the wrapper.

Worth calling out explicitly for future readers, because it looks redundant and isn't: pairing `strip_leading_bom(&content)` with `from_json_str(content)` at the nine guarded sites is deliberate. `U+FEFF` is not Unicode whitespace, so without the explicit strip a BOM-only file would leave `"\u{feff}"`, miss the `trim().is_empty()` branch, and hard-error instead of yielding `{}`. Please don't collapse those to a bare `from_json_str` in a later cleanup.

One non-blocking observation inline — a behavioural consequence of the fix I asked for, worth a moment's thought but not something I'd hold the PR on.

Small nit on the description rather than the code: PowerShell 5.1's `>` writes UTF-16LE, not a UTF-8 BOM, so the repro as written wouldn't actually reproduce (`read_to_string` would reject it as invalid UTF-8 first). `Out-File -Encoding utf8` is the accurate repro. Doesn't affect the fix.

KuSh:
Superseded by #3751, which carries this work forward (the BOM stripping is promoted to a shared `core::utils::from_json_str` helper and applied to the remaining JSON parse sites). Closing this one in favour of that PR — thanks for the original fix, it's fully preserved there.

Review feedback now lives on #3751.

PR #3206, opened 2026-07-25
fix(discover): log real hook decisions instead of guessing coverage retroactively

### LINKED ISSUE #3148: rtk discover counts hook-rewritten commands as misses

KuSh:
## Summary

`rtk discover` re-derives whether a historical transcript command would have been covered by the PreToolUse hook using *today's* hook-install state and registry, applied uniformly across the whole scan window (default 30 days). PR #3164 attempted a fix along these lines but has the same gap: history that predates the hook being installed, or a registry/permission change, gets mis-classified — plus it misses `permissions.deny` and doesn't gate the `RTK_DISABLED=` bypass bucket the same way.

This PR logs the *real* PreToolUse decision (allow/ask/deny/defer) at the moment the Claude Code hook actually runs, keyed by `tool_use_id` — the same id Claude Code stores on the transcript's `tool_use`/`tool_result` blocks (verified empirically, not assumed) — so `discover` can join real hook outcomes back to transcript entries instead of guessing. Falls back to a heuristic estimate only for history that predates logging (or exceeds the tracking DB's retention window — see below), and labels that portion of the report as an estimate rather than blending it in silently.

- New `hook_decisions` SQLite table + `HookOutcome` enum (`Display`/`FromStr`), shared between the writer (`hooks::hook_cmd`) and reader (`discover`). Logging happens after the hook's real stdout response is written, never in front of it — it's a best-effort side channel and must never add latency to the tool call it's recording.
- `ExtractedCommand` now carries `tool_use_id` so transcript entries can join against the log.
- `Coverage::Measured`/`Coverage::Estimated`: prefers the real log; the estimate fallback checks `PermissionVerdict::Deny` against the *full* raw command (not an isolated chain segment — a deny rule matching a sibling segment must still deny the whole chain), and stops crediting `RTK_DISABLED=` as a "bypass" when the hook wouldn't actually have covered the command anyway (that case now shows as a genuine missed-savings opportunity instead of disappearing from the report).
- Report gains `already_rtk_estimated`/`rtk_disabled_estimated` + `measured_since`, with a text-mode caveat explaining when/why part of a coverage number is an estimate. `measured_since` reflects the oldest currently-retained log entry, not an install date — `hook_decisions` is pruned by the same retention window as the rest of the tracking DB, so the estimate fallback is a permanent fact of scanning far enough back, not a transitional one.
- `record_hook_decision()` skips the full 3-table `cleanup_old()` sweep on its hot path (every Bash tool call, not just RTK-covered ones), but still gets a lightweight sampled retention sweep of its own so a user whose commands are mostly denied/deferred doesn't grow the table unbounded.
- Permission rules, exclude-pattern regexes, and hook-install status are loaded/compiled once per `discover` run instead of once per scanned command, and config is cached per-process for the hook's hot path — a large history (tens of thousands of commands) was previously dominated by this redundant I/O.
- `rtk discover --since <very large>` no longer crashes; the cutoff computation is clamped instead of overflowing.

Fixes #3148

## End-to-end validation

Beyond `cargo test --all` (2749 passed) and `cargo clippy --all-targets` (clean), ran the built release binary against a synthetic Claude Code transcript to validate the fix's actual runtime behavior, isolating state via `CLAUDE_CONFIG_DIR` (Claude config dir) and `RTK_DB_PATH` (tracking DB) so nothing touched real user data.

**Setup**: one synthetic session transcript with four Bash tool calls, each with its own `tool_use_id`:
- `grep foo bar.py` — Supported, cleanly rewritable
- `git status` — Supported, cleanly rewritable
- `git status $(whoami)` — Supported but contains an unattestable construct (command substitution), which the hook always defers on
- `rtk proxy grep foo bar.py` — already `rtk`-prefixed, but via the escape hatch that must not count as coverage

**Scenario 1 — hook never installed** (no `settings.json` in the synthetic `CLAUDE_CONFIG_DIR`, empty `hook_decisions` table):
```
Already using RTK: 0 commands (0.0%)
MISSED SAVINGS -- Commands RTK already handles
Command                  Count    RTK Equivalent     Status        Est. Savings
git status                   2    rtk git            existing      ~4 tokens
grep foo                     1    rtk grep           existing      ~2 tokens
```
`git status` (both instances, including the unattestable one) and `grep` correctly show as missed; `rtk proxy` doesn't appear in either bucket.

**Scenario 2 — hook installed, with real decisions logged**: wrote a `settings.json` registering `rtk hook claude` as the PreToolUse hook, then piped the matching real PreToolUse JSON payload for each of the four commands (same `tool_use_id`s as the transcript) through the actual built `rtk hook claude` binary, populating `hook_decisions` for real. Verified via direct sqlite read that all four rows landed with the expected decisions (`grep`→ask, `git status`→ask, the `$(whoami)` variant→defer, `rtk proxy`→defer). Then ran `rtk discover` against the same DB/transcript:
```
Already using RTK: 2 commands (50.0%)
MISSED SAVINGS -- Commands RTK already handles
Command                  Count    RTK Equivalent     Status        Est. Savings
git status                   1    rtk git            existing      ~2 tokens
```
- Coverage jumped to 2/4 (50%), matching the two commands that were actually rewritable.
- `git status $(whoami)` still correctly shows as a genuine MISS (the hook really did defer on it — this isn't just "it was estimated as uncovered", it's a *measured* defer).
- `rtk proxy` still doesn't inflate either bucket.
- JSON output (`--format json`) confirmed `already_rtk_estimated: 0` and `measured_since` set to the log's timestamp — proving the *measured* path drove the result end-to-end, not the heuristic fallback.

**Additional scenarios verified since**: an RTK_DISABLED= bypass on a command the hook wouldn't have covered anyway now shows as missed savings rather than vanishing from the report; a chain-wide deny rule (`cd /sensitive && grep -rn foo .`) matching only one segment correctly denies the whole chain in the estimate; `rtk discover --since 100000000` no longer panics; a real PreToolUse payload piped through the built binary still writes a correct `hook_decisions` row with the response written before the log entry.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` (2749 passed, 0 failed)
- [x] Manual e2e run against a synthetic transcript, hook-absent vs hook-installed-with-real-decisions (see above)
- [x] Verified `hook_decisions` rows persist correctly via direct sqlite inspection
- [x] Merged and resolved conflicts against `upstream/develop`

### DIFF
--- a/src/analytics/session_cmd.rs
+++ b/src/analytics/session_cmd.rs
@@ -1,6 +1,7 @@
 //! Compares RTK-routed vs raw commands in a coding session.
 
 use crate::core::utils::format_tokens;
+use crate::discover::is_already_rtk;
 use crate::discover::provider::{ClaudeProvider, ExtractedCommand, SessionProvider};
 use crate::discover::registry::{classify_command, split_command_chain, Classification};
 use anyhow::{Context, Result};
@@ -27,7 +28,9 @@ impl SessionSummary {
 
 /// Count RTK-covered commands from extracted commands.
 /// A command is "covered" if it either:
-/// - starts with "rtk " (explicit rtk invocation), or
+/// - is an explicit rtk invocation, excluding `rtk proxy` (see `is_already_rtk` —
+///   proxy deliberately runs the raw command unfiltered, so it must not count as
+///   adopted usage), or
 /// - would be rewritten by the hook (classify_command returns Supported)
 ///
 /// Chained commands (e.g. "cd ./path && rtk ls") are split so each part
@@ -39,7 +42,7 @@ fn count_rtk_commands(cmds: &[ExtractedCommand]) -> (usize, usize, usize) {
         let parts = split_command_chain(&c.command);
         for part in &parts {
             total += 1;
-            if part.starts_with("rtk ")
+            if is_already_rtk(part)
                 || matches!(classify_command(part), Classification::Supported { .. })
             {
                 rtk += 1;
@@ -271,6 +274,20 @@ mod tests {
         let (total, rtk, _) = count_rtk_commands(&cmds);
         assert_eq!(total, 3);
         assert_eq!(rtk, 0);
+    }
+
+    #[test]
+    fn test_count_rtk_proxy_not_counted_as_adopted() {
+        // `rtk proxy <cmd>` deliberately runs the raw command unfiltered — it must
+        // not count as adopted rtk usage, or the session report would flatter
+        // itself via its own escape hatch (mirrors discover::is_already_rtk).
+        let cmds = vec![
+            make_cmd("rtk proxy git log -20", Some(2000)),
+            make_cmd("rtk git status", Some(200)),
+        ];
+        let (total, rtk, _) = count_rtk_commands(&cmds);
+        assert_eq!(total, 2);
+        assert_eq!(rtk, 1, "only the explicit non-proxy rtk invocation counts");
     }
 
     #[test]
--- a/src/analytics/session_cmd.rs
+++ b/src/analytics/session_cmd.rs
@@ -200,6 +200,7 @@ mod tests {
             command: command.to_string(),
             output_len,
             session_id: "test".to_string(),
+            tool_use_id: "toolu_test".to_string(),
             output_content: None,
             is_error: false,
             sequence_index: 0,
--- a/src/core/config.rs
+++ b/src/core/config.rs
@@ -150,6 +150,17 @@ pub fn limits() -> LimitsConfig {
     Config::load().map(|c| c.limits).unwrap_or_default()
 }
 
+/// Get `(exclude_commands, transparent_prefixes)` for hook-rewrite decisions.
+/// Falls back to empty (no exclusions/prefixes) if config can't be loaded.
+/// Shared by every place that decides whether/how to rewrite a command
+/// (`hooks::hook_cmd`, `hooks::rewrite_cmd`, `discover`, `rtk rewrite`'s CLI
+/// entry point in `main.rs`) so they can't drift from each other.
+pub fn hook_rewrite_params() -> (Vec<String>, Vec<String>) {
+    Config::load()
+        .map(|c| (c.hooks.exclude_commands, c.hooks.transparent_prefixes))
+        .unwrap_or_default()
+}
+
 impl Config {
     pub fn load() -> Result<Self> {
         let path = get_config_path()?;
--- a/src/core/config.rs
+++ b/src/core/config.rs
@@ -155,10 +155,33 @@ pub fn limits() -> LimitsConfig {
 /// Shared by every place that decides whether/how to rewrite a command
 /// (`hooks::hook_cmd`, `hooks::rewrite_cmd`, `discover`, `rtk rewrite`'s CLI
 /// entry point in `main.rs`) so they can't drift from each other.
+///
+/// Reads the process-wide cached config (see `cached_config`), not a fresh
+/// `Config::load()`: this is on the PreToolUse hook's hot path, and
+/// `tracking::get_db_path` (called via `Tracker::new()` for `hook_decisions`
+/// logging, right after this in the same hook invocation) also reads config —
+/// without caching, that's two full disk-read-plus-TOML-parse round trips per
+/// single Bash tool call instead of one.
 pub fn hook_rewrite_params() -> (Vec<String>, Vec<String>) {
-    Config::load()
-        .map(|c| (c.hooks.exclude_commands, c.hooks.transparent_prefixes))
-        .unwrap_or_default()
+    let c = cached_config();
+    (
+        c.hooks.exclude_commands.clone(),
+        c.hooks.transparent_prefixes.clone(),
+    )
+}
+
+/// Process-wide cached `Config::load()` result, populated on first use.
+///
+/// Safe for read-only callers on hot paths that may load config multiple times
+/// within a single `rtk` invocation (a `rtk` process is short-lived and exits
+/// after one subcommand, so there's no cross-invocation staleness to worry
+/// about) — but NOT used by any path that mutates and saves config within the
+/// same process run (e.g. `hooks::init::save_telemetry_consent`'s load-mutate-save),
+/// since those must always observe a fresh read. Only reach for this from a
+/// caller that never itself writes config.toml.
+pub(crate) fn cached_config() -> &'static Config {
+    static CACHE: std::sync::OnceLock<Config> = std::sync::OnceLock::new();
+    CACHE.get_or_init(|| Config::load().unwrap_or_default())
 }
 
 impl Config {
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -33,6 +33,7 @@ use anyhow::{Context, Result};
 use chrono::{DateTime, Utc};
 use rusqlite::{params, Connection};
 use serde::Serialize;
+use std::collections::HashMap;
 use std::ffi::OsString;
 use std::path::PathBuf;
 use std::time::Instant;
@@ -105,6 +106,72 @@ pub struct CommandRecord {
     pub saved_tokens: usize,
     /// Savings percentage ((saved / input) * 100)
     pub savings_pct: f64,
+}
+
+/// The real outcome a PreToolUse hook reached for one Bash call.
+///
+/// Shared by both ends of the `hook_decisions` log: `hooks::hook_cmd` writes it at
+/// the moment the hook actually runs, and `discover` reads it back to know whether
+/// a historical command was truly covered. `Display`/`FromStr` are the only
+/// string boundary — everywhere else this stays a typed enum.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum HookOutcome {
+    /// Rewritten and auto-allowed.
+    Allow,
+    /// Rewritten, but Claude Code still prompts the user.
+    Ask,
+    /// A deny rule matched — the hook never rewrites, defers to native deny handling.
+    Deny,
+    /// An unattestable construct (substitution, etc.) — the hook defers, no rewrite.
+    Defer,
+}
+
+impl HookOutcome {
+    /// Whether this outcome means the command was actually routed through RTK.
+    pub fn is_covered(self) -> bool {
+        matches!(self, HookOutcome::Allow | HookOutcome::Ask)
+    }
+}
+
+impl std::fmt::Display for HookOutcome {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(match self {
+            HookOutcome::Allow => "allow",
+            HookOutcome::Ask => "ask",
+            HookOutcome::Deny => "deny",
+            HookOutcome::Defer => "defer",
+        })
+    }
+}
+
+impl std::str::FromStr for HookOutcome {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "allow" => Ok(HookOutcome::Allow),
+            "ask" => Ok(HookOutcome::Ask),
+            "deny" => Ok(HookOutcome::Deny),
+            "defer" => Ok(HookOutcome::Defer),
+            _ => Err(()),
+        }
+    }
+}
+
+/// A single PreToolUse hook decision, logged at the moment the hook actually ran.
+///
+/// Keyed by `tool_use_id` so it can be joined 1:1 against the same id Claude Code
+/// stores on the transcript's `tool_use`/`tool_result` blocks — giving `rtk discover`
+/// ground truth about historical hook coverage instead of re-deriving a guess from
+/// today's hook-install state and registry.
+///
+/// Only carries what `discover` actually consumes (the decision). The
+/// `hook_decisions` table itself also stores `timestamp`/`raw_cmd`/`rewritten_cmd`/
+/// `rtk_version` for direct inspection (`sqlite3 tracking.db`), but nothing in Rust
+/// reads those back today — add fields here if/when something does.
+#[derive(Debug, Clone, Copy)]
+pub struct HookDecisionRecord {
+    pub decision: HookOutcome,
 }
 
 /// Aggregated statistics across all recorded commands.
@@ -336,6 +403,29 @@ impl Tracker {
             [],
         )?;
 
+        conn.execute(
+            "CREATE TABLE IF NOT EXISTS hook_decisions (
+                id INTEGER PRIMARY KEY,
+                timestamp TEXT NOT NULL,
+                session_id TEXT NOT NULL,
+                tool_use_id TEXT NOT NULL,
+                project_path TEXT DEFAULT '',
+                raw_cmd TEXT NOT NULL,
+                decision TEXT NOT NULL,
+                rewritten_cmd TEXT,
+                rtk_version TEXT NOT NULL
+            )",
+            [],
+        )?;
+        conn.execute(
+            "CREATE INDEX IF NOT EXISTS idx_hook_decisions_tool_use_id ON hook_decisions(tool_use_id)",
+            [],
+        )?;
+        conn.execute(
+            "CREATE INDEX IF NOT EXISTS idx_hook_decisions_timestamp ON hook_decisions(timestamp)",
+            [],
+        )?;
+
         restrict_db_files(&db_path);
 
         Ok(Self { conn })
@@ -387,6 +477,28 @@ impl Tracker {
         )?;
         self.conn.execute(
             "CREATE INDEX IF NOT EXISTS idx_pf_timestamp ON parse_failures(timestamp)",
+            [],
+        )?;
+        self.conn.execute(
+            "CREATE TABLE IF NOT EXISTS hook_decisions (
+                id INTEGER PRIMARY KEY,
+                timestamp TEXT NOT NULL,
+                session_id TEXT NOT NULL,
+                tool_use_id TEXT NOT NULL,
+                project_path TEXT DEFAULT '',
+                raw_cmd TEXT NOT NULL,
+                decision TEXT NOT NULL,
+                rewritten_cmd TEXT,
+                rtk_version TEXT NOT NULL
+            )",
+            [],
+        )?;
+        self.conn.execute(
+            "CREATE INDEX IF NOT EXISTS idx_hook_decisions_tool_use_id ON hook_decisions(tool_use_id)",
+            [],
+        )?;
+        self.conn.execute(
+            "CREATE INDEX IF NOT EXISTS idx_hook_decisions_timestamp ON hook_decisions(timestamp)",
             [],
         )?;
         Ok(())
@@ -461,16 +573,21 @@ impl Tracker {
             "DELETE FROM parse_failures WHERE timestamp < ?1",
             params![cutoff.to_rfc3339()],
         )?;
+        self.conn.execute(
+            "DELETE FROM hook_decisions WHERE timestamp < ?1",
+            params![cutoff.to_rfc3339()],
+        )?;
         Ok(())
     }
 
-    /// Delete all tracked data (commands + parse_failures), resetting all stats to zero.
+    /// Delete all tracked data (commands + parse_failures + hook_decisions), resetting all stats to zero.
     pub fn reset_all(&self) -> Result<()> {
         self.conn
             .execute_batch(
                 "BEGIN;
                  DELETE FROM commands;
                  DELETE FROM parse_failures;
+                 DELETE FROM hook_decisions;
                  COMMIT;",
             )
             .context("Failed to reset tracking database")?;
@@ -496,6 +613,88 @@ impl Tracker {
         )?;
         self.cleanup_old()?;
         Ok(())
+    }
+
+    /// Record a PreToolUse hook decision at the moment the hook actually ran.
+    ///
+    /// `decision` must be one of `"allow"`, `"ask"`, `"deny"`, `"defer"`.
+    #[allow(clippy::too_many_arguments)]
+    pub fn record_hook_decision(
+        &self,
+        session_id: &str,
+        tool_use_id: &str,
+        project_path: &str,
+        raw_cmd: &str,
+        decision: HookOutcome,
+        rewritten_cmd: Option<&str>,
+        rtk_version: &str,
+    ) -> Result<()> {
+        self.conn.execute(
+            "INSERT INTO hook_decisions (timestamp, session_id, tool_use_id, project_path, raw_cmd, decision, rewritten_cmd, rtk_version)
+             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8)",
+            params![
+                Utc::now().to_rfc3339(),
+                session_id,
+                tool_use_id,
+                project_path,
+                raw_cmd,
+                decision.to_string(),
+                rewritten_cmd,
+                rtk_version,
+            ],
+        )?;
+        self.cleanup_old()?;
+        Ok(())
+    }
+
+    /// Bulk-fetch every hook decision at or after `cutoff`, keyed by `tool_use_id`.
+    ///
+    /// Meant to be called once per `rtk discover` run (mirroring the existing
+    /// single `Config::load()`/`hook_status()` snapshot pattern), so per-command
+    /// lookups during the scan loop are in-memory instead of one query each.
+    pub fn hook_decisions_since(
+        &self,
+        cutoff: DateTime<Utc>,
+    ) -> Result<HashMap<String, HookDecisionRecord>> {
+        let mut stmt = self.conn.prepare(
+            "SELECT tool_use_id, decision
+             FROM hook_decisions
+             WHERE timestamp >= ?1",
+        )?;
+        let rows = stmt
+            .query_map(params![cutoff.to_rfc3339()], |row| {
+                Ok((row.get::<_, String>(0)?, row.get::<_, String>(1)?))
+            })?
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let mut out = HashMap::with_capacity(rows.len());
+        for (tool_use_id, decision) in rows {
+            // Skip rows with an unrecognized decision string (e.g. written by a
+            // future rtk version with a decision this build doesn't know) rather
+            // than failing the whole query — `discover` just falls back to its
+            // estimate heuristic for that command.
+            if let Ok(decision) = decision.parse::<HookOutcome>() {
+                out.insert(tool_use_id, HookDecisionRecord { decision });
+            }
+        }
+        Ok(out)
+    }
+
+    /// Earliest timestamp any hook decision was logged, if any exist.
+    ///
+    /// Used to tell `rtk discover` where the "measured" window starts — any scan
+    /// range before this point falls back to the estimate-based heuristic.
+    pub fn earliest_hook_decision_timestamp(&self) -> Result<Option<DateTime<Utc>>> {
+        let ts: Option<String> =
+            self.conn
+                .query_row("SELECT MIN(timestamp) FROM hook_decisions", [], |row| {
+                    row.get(0)
+                })?;
+        Ok(ts.and_then(|t| {
+            DateTime::parse_from_rfc3339(&t)
+                .ok()
+                .map(|dt| dt.with_timezone(&Utc))
+        }))
     }
 
     /// Get parse failure summary for `rtk gain --failures`.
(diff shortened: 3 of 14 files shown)


pszymkowiak:
## Perf note: `log_hook_decision` adds a SQLite round-trip to every PreToolUse call

Nice fix for the retroactive-guessing problem — measuring the real decision instead of re-deriving it is clearly the right approach.

One thing worth flagging before merge: `log_hook_decision()` now calls `Tracker::new()` on **every single PreToolUse hook invocation** (both the `Rewrite` and `Skip` paths) — i.e. every Bash tool call Claude Code makes, not just the ones rtk ends up covering. `hook_cmd.rs` never touched the tracking DB before this PR. `Tracker::new()` opens a SQLite connection, sets `PRAGMA journal_mode=WAL`, and runs several `CREATE TABLE/INDEX IF NOT EXISTS` + an `ALTER TABLE` + a `SELECT EXISTS` — all of that on the hook's hot path, which is explicitly called out elsewhere in this repo as being under a <10ms budget.

I benchmarked it with hyperfine (`rtk hook claude` fed a realistic payload with `session_id`/`tool_use_id`/`cwd`, warm DB, 200+ runs, `--warmup 20`):

| | mean |
|---|---|
| `develop` (no DB write) | ~12.6 ms |
| this PR as-is (SQLite write/call) | ~13.5 ms (**+0.9 ms, ~7%**) |
| prototype below (jsonl append) | ~12.9 ms (**+0.3 ms, ~2%**) |

Confirmed with `ls`/`sqlite3` that `develop` and the prototype never create a `.db` file on this path at all, while this PR's binary does, on every call.

### Suggested alternative

Instead of a SQLite table, log to a plain append-only `hook_decisions.jsonl` (same cost class as the existing plaintext audit log — just an `OpenOptions::append` write, no connection/DDL), and have `discover` read that file back directly instead of querying SQL:

```rust
// src/core/tracking.rs
fn hook_decisions_log_path() -> Result<PathBuf> {
    Ok(get_db_path()?.with_file_name("hook_decisions.jsonl"))
}

pub fn append_hook_decision(
    session_id: &str,
    tool_use_id: &str,
    project_path: &str,
    raw_cmd: &str,
    decision: HookOutcome,
    rewritten_cmd: Option<&str>,
    rtk_version: &str,
) -> Result<()> {
    let path = hook_decisions_log_path()?;
    if let Some(parent) = path.parent() {
        std::fs::create_dir_all(parent)?;
    }
    let mut file = std::fs::OpenOptions::new().create(true).append(true).open(&path)?;
    let line = serde_json::json!({
        "timestamp": Utc::now().to_rfc3339(),
        "session_id": session_id,
        "tool_use_id": tool_use_id,
        "project_path": project_path,
        "raw_cmd": raw_cmd,
        "decision": decision.to_string(),
        "rewritten_cmd": rewritten_cmd,
        "rtk_version": rtk_version,
    });
    use std::io::Write;
    writeln!(file, "{line}")?;
    Ok(())
}

pub type HookDecisionsLog = (HashMap<String, HookDecisionRecord>, Option<DateTime<Utc>>);

pub fn read_hook_decisions_since(cutoff: DateTime<Utc>) -> Result<HookDecisionsLog> {
    let path = hook_decisions_log_path()?;
    let content = match std::fs::read_to_string(&path) {
        Ok(c) => c,
        Err(_) => return Ok((HashMap::new(), None)),
    };
    let mut map = HashMap::new();
    let mut earliest: Option<DateTime<Utc>> = None;
    for line in content.lines() {
        let Ok(v) = serde_json::from_str::<serde_json::Value>(line) else { continue };
        let Some(ts) = v.get("timestamp").and_then(|t| t.as_str())
            .and_then(|s| DateTime::parse_from_rfc3339(s).ok())
            .map(|dt| dt.with_timezone(&Utc)) else { continue };
        earliest = Some(match earliest { Some(e) if e <= ts => e, _ => ts });
        if ts < cutoff { continue; }
        let (Some(tool_use_id), Some(decision)) = (
            v.get("tool_use_id").and_then(|t| t.as_str()),
            v.get("decision").and_then(|d| d.as_str()).and_then(|s| s.parse::<HookOutcome>().ok()),
        ) else { continue };
        map.insert(tool_use_id.to_string(), HookDecisionRecord { decision });
    }
    Ok((map, earliest))
}
```

And the two call sites become:

```diff
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ fn log_hook_decision(v: &Value, cmd: &str, decision: HookOutcome, rewritten: Option<&str>) {
-    let Ok(tracker) = crate::core::tracking::Tracker::new() else {
-        return;
-    };
-    if let Err(e) = tracker.record_hook_decision(
+    if let Err(e) = crate::core::tracking::append_hook_decision(
         session_id,
         tool_use_id,
         project_path,
```

```diff
--- a/src/discover/mod.rs
+++ b/src/discover/mod.rs
@@ pub fn run(
-        Tracker::new()
-            .map(|t| {
-                let log = t.hook_decisions_since(cutoff).unwrap_or_default();
-                let measured_since = t.earliest_hook_decision_timestamp().ok().flatten();
-                (log, measured_since)
-            })
-            .unwrap_or_default();
+        crate::core::tracking::read_hook_decisions_since(cutoff).unwrap_or_default();
```

The tradeoff: `discover`'s `cutoff` filtering and `measured_since` become a linear scan over the jsonl file instead of an indexed SQL query. For the actual data volume here (one line per Bash tool call, and `discover` already runs once per invocation, not in a hot loop) that's negligible — but worth being explicit about the tradeoff since it's a real one.

I have this fully working as a prototype on top of this branch (builds clean, `cargo clippy --all-targets` clean, full `cargo test --all` passes — including rewritten versions of the 4 tests that exercised `record_hook_decision`/`hook_decisions_since`/`earliest_hook_decision_timestamp` directly, now driven against a temp file instead of the shared `RTK_DB_PATH` env var to avoid the parallel-test race already flagged elsewhere in `tracking.rs`). Happy to open it as a PR against this branch, or just paste the full diff here, whichever is more useful — let me know.

(Small unrelated nit I noticed while in there: `record_hook_decision`'s doc comment is currently duplicated — two `///` blocks with `#[allow(clippy::too_many_arguments)]` sandwiched in between. Worth a quick cleanup regardless of which direction this perf question goes.)

KuSh:
@pszymkowiak Thanks for the benchmark — you're right that `log_hook_decision()` shouldn't be re-running the full schema-init sequence on every PreToolUse call. I don't want to add a second persistence mechanism alongside SQLite though (a jsonl log plus SQLite means two places to reason about retention, concurrent-write safety, and schema evolution going forward). Instead I have a POC that keeps everything in SQLite but removes the redundant per-call work — proposing this for discussion before I push it.

### Proposed fix: gate schema migration behind `PRAGMA user_version`

`Tracker::new()` currently re-runs all 13 DDL statements (2×`CREATE TABLE IF NOT EXISTS`, 2×`ALTER TABLE ADD COLUMN`, a `SELECT EXISTS` scan, 5×`CREATE INDEX IF NOT EXISTS`, plus the two table/index blocks for `hook_decisions`) on *every* call — not just from `log_hook_decision`, but from the two pre-existing hot paths `TimedExecution::track`/`track_passthrough` too (every single `rtk <cmd>` invocation already paid this before this PR).

The fix: store a `SCHEMA_VERSION` constant, stamp it into SQLite's built-in `user_version` pragma once migrations run, and check that pragma (a single scalar read, no `sqlite_master` lookup) before deciding whether to run the DDL block at all:

```rust
const SCHEMA_VERSION: i64 = 1;

fn run_schema_migrations(conn: &Connection) -> Result<()> {
    // ...unchanged CREATE TABLE/INDEX/ALTER statements, byte-for-byte...
    conn.pragma_update(None, "user_version", SCHEMA_VERSION)?;
    Ok(())
}

impl Tracker {
    pub fn new() -> Result<Self> {
        let db_path = get_db_path()?;
        if let Some(parent) = db_path.parent() {
            std::fs::create_dir_all(parent)?;
        }
        let conn = Connection::open(&db_path)?;
        let _ = conn.execute_batch("PRAGMA journal_mode=WAL; PRAGMA busy_timeout=5000;");

        let current_version: i64 = conn.query_row("PRAGMA user_version", [], |row| row.get(0))?;
        if current_version < SCHEMA_VERSION {
            run_schema_migrations(&conn)?;
        }
        Ok(Self { conn })
    }
}
```

The first `Tracker::new()` after a fresh install or a version bump pays the full migration cost once and stamps `user_version`; every call after that — every hook invocation, every `rtk <cmd>` — is just `Connection::open` + 2 pragmas + 1 scalar read. Bumping `SCHEMA_VERSION` is how future migrations get added (mirrors how the existing `ALTER TABLE` migrations already no-op safely once applied — same idea, just gated instead of unconditional).

I also added a pre-warm call (`let _ = Tracker::new();`, best-effort, matching this repo's fallback pattern) at the end of `rtk init`'s install path, so the one-time migration happens synchronously during install/upgrade rather than on whichever Bash call happens to run first afterward.

`new_in_memory()` (the `#[cfg(test)]`-only in-memory path) is untouched — every test DB is fresh, so there's no repeated-call cost there to optimize.

### Benchmark

Same methodology as your comment: `hyperfine` on `rtk hook claude` fed the same realistic payload shape, warm DB, `--warmup` before measuring. I built three release binaries — `develop` (pre-#3206, no `Tracker` touch at all in `run_claude`), this PR as-is (unconditional schema-init), and the POC — and ran them side by side:

```
Benchmark 1: develop (no DB write)
  Time (mean ± σ):  18.5 ms ± 1.1 ms   (1000 runs)

Benchmark 2: PR as-is (unconditional schema-init)
  Time (mean ± σ):  19.3 ms ± 1.2 ms   (1000 runs)

Benchmark 3: POC (user_version gate)
  Time (mean ± σ):  19.2 ms ± 1.5 ms   (1000 runs)
```

Full transparency: my environment is noisier than yours (baseline here is ~18.5ms vs your ~12.6ms — sounds like a more virtualized/shared box), and at that noise level (σ ~1.2-1.5ms) a ~0.1-0.3ms gap between "PR as-is" and "POC" isn't something I can call statistically conclusive from a single machine. I re-ran it three separate times (200/500/1000 samples each) and POC's mean landed at or slightly below "PR as-is" every time, consistent with the gate doing less work, but I'd rather be honest that the effect size here is close to my noise floor than round up.

What I can confirm precisely (no timing noise involved): with the gate, the on-disk DB's `PRAGMA user_version` is stamped to `1` after the first call and stays there — I verified via a direct SQLite read that a second `Tracker::new()` against the same file does not re-touch `sqlite_master` for the DDL statements (only `hook_decisions` row count changes, no schema churn). So the *mechanism* is confirmed working as designed even where the wall-clock delta is within noise on this box. Given your own measurement showed a clean, real ~0.9ms gap on your machine, I'd expect this gate to recover most or all of it there too — happy to have you (or anyone else) re-run your existing methodology against the POC if you want a tighter number than I can produce here.

Also, per your nit: the doc-comment duplication on `record_hook_decision` is fixed in the POC too.

Not pushing this yet — wanted to get eyes on the approach first, per feedback from whoever's driving this fix. If the `user_version` gate + `rtk init` pre-warm direction looks right, I'll commit and push it as a follow-up on this branch.

KuSh:
Follow-up with a cleaner measurement: I re-ran the same three binaries/methodology on quieter, better hardware (my earlier numbers had σ ~1.2-1.5ms of noise, which was too close to the effect size to draw a real conclusion from).

```
Benchmark 1: develop (no DB write)
  Time (mean ± σ):   6.7 ms ± 0.5 ms   (500 runs)

Benchmark 2: PR as-is (unconditional schema-init)
  Time (mean ± σ):   7.1 ms ± 0.4 ms   (500 runs)

Benchmark 3: POC (user_version gate)
  Time (mean ± σ):   7.0 ms ± 0.4 ms   (500 runs)
```

So: +0.4ms for the PR as it stands, +0.3ms with the `user_version` gate. That +0.3ms lines up with the +0.3ms you measured for the jsonl prototype — same practical hot-path cost, without a second persistence mechanism.

Dug into where that remaining +0.3ms actually comes from, since the gate only removes the DDL re-run (12 statements down to 1 `PRAGMA user_version` read) and I wanted to know if there was more to get: it isn't the DDL. On a warm, page-cached DB those `CREATE TABLE/INDEX IF NOT EXISTS` checks are already cheap (~0.1ms for all 12 combined). The rest is `Connection::open()` + the per-connection pragmas + the actual `INSERT` into `hook_decisions` — costs that are identical between "PR as-is" and "POC" because both open a connection and log a row. I checked whether any single `rtk` invocation opens more than one `Tracker` (which would make an in-process cache worth adding) — it doesn't; every call site (`log_hook_decision`, `discover`, `gain`, `cc_economics`) opens exactly one `Tracker` per process. So there's no redundant open to dedupe within a process, and the remaining cost is the inherent floor of touching SQLite from a fresh CLI invocation at all — the same floor any persistence approach (including jsonl) pays. Removing it further would mean a resident daemon process holding a connection across hook calls, which is a much bigger architectural change than this fix warrants (and cuts against this repo's single-threaded, one-shot-CLI design) — not proposing that here.

Net: the `user_version` gate gets this to the same hot-path cost as the jsonl idea while keeping one storage mechanism, and there's no further low-risk optimization available on top of it.

KuSh:
@pszymkowiak rebased with the schema migration gated behind PRAGMA user_version + some more iteration on it:
1. hook_cmd.rs audit prefix regression — fixed: restored skip:deny_rule/skip:defer audit-log reasons instead of bare HookOutcome::Display.
2. Chained-command coverage attribution — investigated deeply, found the "fix" was based on a wrong assumption (rewrite is per-segment, not all-or-nothing); reverted the premature change and documented the real limitation instead, since it naturally fades out as hook_decisions backfills history.
3. rtk proxy counted as adopted — fixed: session_cmd.rs now reuses discover::is_already_rtk instead of its own diverging starts_with("rtk ") check.
4. Self-healing schema migration — discussed with you, landed on your simpler idea: rtk init now unconditionally re-runs migrations (bypassing the user_version gate), healing a dropped/corrupted table without a new flag; added a diagnostic pointing at rtk init on write failure. Also fixed an unrelated pre-existing test-isolation bug (per-function static ENV_LOCKs that didn't actually share a lock) surfaced while testing this.
5. Duplicated Config::load() boilerplate — fixed: extracted core::config::hook_rewrite_params(), used by all 4 call sites.
6. Test schema duplication — fixed: new_in_memory() now calls run_schema_migrations directly instead of hand-duplicated DDL.
7. TimedExecution test flakiness from shared on-disk DB races under parallel cargo test — fixed. Root cause: several tests called Tracker::new() with no RTK_DB_PATH override, so they all hit the same real default DB file concurrently — get_recent(N) then raced against other tests' concurrent inserts, occasionally missing the test's own row (or erroring outright under heavy concurrent file access). Switched the tests that don't need real on-disk behavior to Tracker::new_in_memory(), and isolated the two TimedExecution tests (which hardcode Tracker::new() internally) via a private temp-file RTK_DB_PATH under the shared lock. Verified stable across 14 repeated runs, including 6 at 16 threads, with zero failures.

aeppling:
Hey @KuSh 

Since no news from Patrick i've went through the PR so we can go forward with it.
Before going deeper in code review, some first observations and questions.


The fix works. On my 30d history (473 sessions, ~43k commands) coverage goes 2.2% → 27.9%, and the ~3.19M "saveable" tokens drop to ~201K. Top base misses were `grep -n` ×4251 and `git diff` ×2448, both hook-rewritten every time.

Pain on the estimate path (not the measured one):

- 80s vs 2.5s on base. Filled the log with my real `tool_use_id`s and it drops to 1.2s. It's `check_command_for` reloading settings per command, and the verdict is computed eagerly so the `hook_installed` short-circuit never fires.
- 100s with no hook installed, and that never improves, since no hook means no rows.
- RTK_DISABLED section goes empty once the log fills, because the hook always `defer`s on a bypassed command.

Also observed a crash with "rtk discover --since 100000000"

### Questions on direction:

`tool_use_id` is Claude-only (Droid/Vibe send `session_id`, nobody else sends `tool_use_id`), so this can't extend to other agents.
RTK is agent agnostic everywhere else, so tying the fix to a Claude-only field feels like it'll be a problem later.

Also now measured and estimated get blended into one number, so you can't tell which part actually happened and which part is a guess. The fallback re-derives old history from today's config, which is the same kind of guessing the PR set out to remove.

KuSh:
Thanks for the deep review, @aeppling — pushed fixes for everything concrete you found:

**Perf (80s vs 2.5s / the 100s-with-no-hook case)**: two compounding bugs. `estimate_hook_coverage_with_verdict`'s `hook_installed &&` check couldn't short-circuit anything because the caller computed the permission verdict *before* calling it, unconditionally. And `check_command_for` was re-reading every settings file from disk on every single command. Fixed both: permission rules are now loaded once per `discover` run, and `hook_installed` is checked before touching rules/lexer/registry at all. Found and fixed the same "settings-file-per-command" and "no short-circuit" pattern didn't leave any other hot loop behind.

**RTK_DISABLED= section going empty**: root cause was that a bypassed command's *real* logged hook decision is always `Defer` (RTK_DISABLED= makes `get_rewritten` return `None` unconditionally), so once real `hook_decisions` rows existed, checking the measured log for that bucket made it collapse toward zero — the log correctly says "the hook stepped aside," which isn't the question being asked ("would this have been covered *absent* the bypass?"). That bucket now always uses the estimate, since the measured log can never answer the counterfactual.

**Crash on `rtk discover --since 100000000`**: was `Utc::now() - Duration::days(...)` overflowing. Fixed, with regression tests for the huge-value and `u64::MAX` cases. While hunting this down I found the identical overflow pattern in two sibling spots — `discover_sessions_in_projects_dir`'s `days * 86400` and `rtk hook audit --since`'s own cutoff computation — both would've hit the same crash, so I fixed those too and pulled the clamping logic into one shared `core::utils::days_ago_cutoff` helper.

Also found (separately from your review) that the branch didn't compile its own test suite — `HookDecision::AskRewrite` is a tuple variant but a test still constructed it as a struct literal. Fixed.

Ran `/code-review high` twice more after the above. No further correctness issues surfaced. It did flag:
- A pre-existing (not from this PR) redundant lexer scan in the estimate path — filed as #3650 rather than folded in here, since a real fix touches `permissions::check_command_with_rules`'s contract and its other callers.
- Two SQLite-write-on-every-hook-call / unbounded-`hook_decisions`-growth observations that are really inherent tradeoffs of ground-truth logging itself (writing on every call is the whole point; growth is bounded in practice by the existing `cleanup_old()` cadence unless a user's commands are almost entirely denied/deferred) — noting these for you rather than unilaterally changing the logging behavior.
- One maintainability nit (too many positional bool/Vec params threaded through `hook_coverage`/`estimate_hook_coverage`) — bundled into a `CoverageContext` struct, no behavior change.

On your direction questions — not mine to settle, but for what it's worth on current state: the `already_rtk_estimated`/`rtk_disabled_estimated` counters aren't silently blended into the totals — they're tracked separately and the text/JSON report surfaces an explicit "~N estimated from current hook/config state" caveat whenever any estimate contributes, so a reader can tell measured from guessed. The `tool_use_id`-is-Claude-only limitation is real and unaddressed by this round — `Estimated` remains the permanent fallback for other hosts, by design, not just a transitional one for those.

All tests pass (`cargo test --all`), clippy clean.

KuSh:
Following up on #3657 (closed — agreed it belongs here, not deferred): both points are now fixed in this PR.

- **Per-hook-call SQLite write cost**: \`Tracker::new()\`'s \`open_and_prepare()\` was paying a full \`create_private_dir\` + a redundant pre-create \`open()\`/\`close()\` (on top of the real \`Connection::open\` right after it) + an unconditional chmod across the db/-wal/-shm files, on every single call — this now runs on every PreToolUse hook invocation, not just RTK-covered ones, where before this PR that path did at most a JSON audit-log append. \`set_owner_only\` now skips its chmod syscall once permissions already match (still self-heals a genuinely-wrong-permission legacy file, just doesn't reconfirm it every call), and the pre-create open/close is skipped once the DB file already exists. No behavior change on the first-run/healing paths, fewer syscalls after that.
- **Unbounded \`hook_decisions\` growth**: \`record_hook_decision\` deliberately skips the full 3-table \`cleanup_old()\` (hot-path budget), and a user whose commands are mostly \`Deny\`/\`Defer\`'d rarely triggers \`record()\`'s/\`record_parse_failure()\`'s own cleanup either. Added a sampled (~1-in-500 calls) single-table \`hook_decisions\` sweep as a backstop, via a pure/testable sampling core (\`should_sample_cleanup\`) so it doesn't need a real 1-in-500 chance to test.

Verified end-to-end with a real PreToolUse payload (row lands correctly in \`hook_decisions\`) and \`cargo test --all\` (2693 passed, clippy clean). Local \`hyperfine\` before/after on fast local storage (ext4) showed no measurable difference, as expected at this syscall scale — the win is on slower/network-backed home directories, same caveat this file already notes elsewhere for WAL mode.

KuSh:
Ran another `/code-review high` pass (multi-angle, 227 tool calls) over the full branch. It found real issues this time — fixed all of them, plus merged and resolved conflicts against \`upstream/develop\` (base had drifted, 39 commits):

- **Security regression in my own previous fix**: \`set_owner_only\`'s fast-path masked the mode with \`& 0o777\` before comparing, so a file with setuid/setgid/sticky set (e.g. \`0o4600\`) read as "already \`0o600\`" and permanently skipped the chmod that would clear it — defeating the self-heal for exactly the tampered/restored-backup files it exists to catch. Mask is now \`0o7777\`.
- **Exposure-window regression, also mine**: skipping the pre-create dance for an already-existing DB also skipped the only chmod that ran *before* \`Connection::open\` — a file with drifted permissions would get opened/migrated before \`restrict_db_files\` healed it at the end. Now re-chmods (cheaply, via the same optimized path) before opening, for the existing-file case too.
- **Sampling reliability, also mine**: \`should_sample_cleanup\`'s 1-in-500 sampling read \`Utc::now().timestamp_subsec_nanos()\`, which silently breaks on any clock source whose resolution doesn't hit every residue mod 500 (coarse virtualized/Windows timers) — could sample ~100% or ~0% of calls depending on tick size. Now hashes \`tool_use_id\` instead.
- **Silent error swallowing**: \`discover::run\` was masking \`hook_decisions\` read failures instead of warning like every write path does — could report "no hook-decision log yet" for an actually-corrupted DB. Fixed.
- **RTK_DISABLED= permission-check command mismatch**: the bypass path was checking permission rules against the env-prefix-*stripped* command, but the real hook checks against the full raw command (prefix intact) — a rule anchored on the literal prefixed string would behave differently. \`estimate_hook_coverage\` now takes separate permission-check and rewrite-check command params (they must differ here, for opposite reasons), and a new \`would_be_covered_without_bypass\` wrapper makes the correct call reachable by name rather than only by doc comment — a prior round already had to catch the hook_coverage-vs-estimate mixup once, so this closes the "doc-comment-only guard" gap too.
- **Perf**: \`registry::rewrite_command\` recompiled its exclude-pattern regexes on every call; \`estimate_hook_coverage\` now calls it once per historical command in \`discover\`'s scan loop, turning a previously-rare cost into a hot-loop one — the same "recompute per command" class this PR's own \`PermissionRules\`/\`hook_status\` caching already fixed elsewhere. Split into \`rewrite_command\` (compiles, thin wrapper) and \`rewrite_command_precompiled\`, with patterns precompiled once per \`discover\` run.
- Fixed a stale comment in \`provider.rs\` referencing a function name that had been refactored away in an earlier commit.

\`cargo test --all\`: 2746 passed (post-merge with develop). \`cargo clippy\`: clean. Verified end-to-end with a real PreToolUse payload and the \`--since\` crash repro.

KuSh:
Ran \`/code-review max\`. 10 finder angles + 6 verification agents + a completeness sweep, ~365K tokens. This one found real issues, several of them in code I'd already touched in prior rounds:

**Fixed:**
- **Latency**: \`run_claude()\` was calling \`log_hook_decision\` (a synchronous SQLite write, 5s \`busy_timeout\`) *before* writing the hook's stdout response Claude Code is synchronously blocked on. Reordered — the response goes out first, logging is genuinely best-effort now (as its own doc comment already claimed but the code didn't honor). Worth noting: SQLite's WAL mode only serializes writer-vs-writer, not reader-vs-writer, so hitting the full 5s in practice would need real pathological contention — but the fix is free and removes the dependency entirely regardless.
- **Correctness**: the coverage estimate's permission-verdict check ran against an isolated chain segment instead of the full raw command, so a \`Deny\` rule matching only a *sibling* segment (\`cd /sensitive && grep -rn foo .\`) was invisible when evaluating the other segment alone. Threaded the full unsplit command through for the permission check specifically, while keeping the per-segment command for the rewrite check (they need to differ, for opposite reasons — see the updated doc comments).
- **Correctness**: \`format_text\`'s "no missed savings, looks good!" early return didn't account for \`rtk_disabled_count\`, so a user whose *every* command ran under \`RTK_DISABLED=\` got a false "all good" message.
- **Correctness**: the RTK_DISABLED branch always \`continue\`d after resolving the bypass question, even when the command genuinely wouldn't have been covered anyway (hook not installed / excluded) — such a command vanished from the whole report instead of showing as a real missed-savings opportunity.
- **Self-inflicted design flaw**: the sampled retention sweep I added in an earlier round prunes \`hook_decisions\` by the same 90-day window as the rest of the tracking DB — so \`earliest_hook_decision_timestamp()\` (and the report's "measured logging began {date}" text) never reflects an install date, just a rolling ~90-day boundary, directly contradicting an earlier comment of mine claiming Estimated "fades out" over time. Corrected the text and comments to state this honestly: Estimated is a *permanent* fallback beyond the retention window, not transitional.
- **Correctness**: \`record_hook_decision\` propagated the retention sweep's error via \`?\` as if it were the INSERT's own failure, even though the INSERT had already committed. Now best-effort or, per rust-patterns.md's anti-pattern rule.
- \`rtk init\`'s schema self-heal call silently discarded its \`Result\` (the exact anti-pattern the project's own rules call out) — now warns.
- **Perf**: \`get_rewritten\`/\`Tracker::new()\` each independently called \`Config::load()\` per hook invocation — two disk-read-plus-TOML-parse round trips instead of one. Added a process-wide cache for the read-only hot-path callers (never used by config-mutating paths, which must keep seeing fresh writes).
- A few minor ones: an avoidable per-call allocation, a two-pass loop that didn't need to be two passes, a test that only exercised session_id-missing when it claimed to test tool_use_id-missing.

**Deliberately not fixed here** (filed as separate issues rather than silently dropped, since they're either pre-existing or a different command's scope):
- #3683 — \`split_command_chain\`'s pipe-truncation blinds the whole classification loop (not just the coverage estimate) to anything after \`|\`. Predates this PR.
- #3684 — \`rtk session\`'s adoption-counting doesn't use the same hook-coverage machinery \`rtk discover\` now does, so they can disagree on the same transcript. Different command, different scope.
- #3685 — \`rtk gain\`'s read paths don't get the same "run \`rtk init\`" self-heal hint the write paths (and now \`discover\`'s reads) do. Pre-existing gap.

\`cargo test --all\`: 2749 passed. clippy clean. Verified end-to-end (synthetic transcripts + real hook payloads): the RTK_DISABLED section no longer hides itself, a bypassed-but-uncovered command now shows as missed savings instead of vanishing, hook_decisions still writes correctly.

aeppling:
Thanks for the quick turnaround. Ran a fuzz pass against develop before going further.

~28k differential invocations: `hook claude` payloads plus 17 hostile DB states, rewrite decisions across `rewrite`/`hook check`/`hook claude` with 43 config and 40 permission variants, `discover` args and malformed transcripts, and 60 concurrent hooks on one DB.

Zero behavioural divergence, no new panic or hang. Rewrite behaviour is unchanged, which is the invariant that mattered here.

It also fixes two develop bugs for free: `rtk hook-audit --since 100000000` SIGABRTs on develop, and `rtk discover --since 2^63` reports `Scanned: 0 sessions ... RTK usage looks good!` because `days * 86400` wraps.

LGTM on the behaviour side.

aeppling (approved):

PR #3212, opened 2026-07-26
fix(hooks): stop Copilot from silently deciding permission on unconfigured commands

### LINKED ISSUE #1425: `[Bug] rtk hook copilot` does not trigger for VS Code Copilot Chat — `"run_in_terminal"` not recognized in `detect_format()`
## Summary

`rtk hook copilot` produces no output (PassThrough) when invoked from VS Code
Copilot Chat, because `detect_format()` does not recognize `"run_in_terminal"`
as a valid tool name. As a result, the hook never fires and all commands pass
through without rewriting.

---

## Environment

- RTK version: 0.37.2 (Homebrew)
- AI tool: VS Code Copilot Chat extension (not Copilot CLI)
- OS: Ubuntu (WSL2)
- No `copilot-instructions.md` was active during hook verification (isolated hook behavior only)

---

## Root Cause

VS Code Copilot Chat sends `tool_name: "run_in_terminal"` in the PreToolUse
hook payload, but `detect_format()` only matches `"runTerminalCommand"`,
`"Bash"`, and `"bash"`:

```rust
// src/hooks/hook_cmd.rs
if matches!(tool_name, "runTerminalCommand" | "Bash" | "bash") {
```

This causes all commands from VS Code Copilot Chat to fall through to
`HookFormat::PassThrough`, and the hook returns no output.

---

## Evidence

Actual JSON received by the hook (captured via debug wrapper script):

```json
{
  "tool_name": "run_in_terminal",
  "tool_input": {
    "command": "git status",
    "explanation": "Check current git status",
    "mode": "sync",
    "timeout": 60000
  }
}
```

VS Code Output panel (GitHub Copilot Chat Hooks channel):

```
[PreToolUse] Input: {"tool_name":"run_in_terminal","tool_input":"..."}
[PreToolUse] Completed (Success) in 26ms, no output
```

---

## Fix

Add `"run_in_terminal"` to the match arm in `detect_format()`:

```rust
if matches!(tool_name, "runTerminalCommand" | "Bash" | "bash" | "run_in_terminal") {
```

Verified locally with a source build — the hook correctly rewrites commands
after this change.

---

## Additional Context: Allow loop after fix (updatedInput not honored)

After applying the fix above, a secondary issue appears in VS Code Copilot Chat:

When the hook returns `permissionDecision: "ask"` with `updatedInput`, VS Code
Copilot Chat enters an infinite Allow-dialog loop:

1. Hook fires → returns `"ask"` + `updatedInput: {"command": "rtk git status"}`
2. Copilot Chat shows an Allow dialog
3. User clicks Allow
4. `updatedInput` is **ignored** — the original command (`git status`) is retried
5. Hook fires again → loop repeats

This is consistent with the `updatedInput` problems reported in Copilot CLI:

- **github/copilot-cli#2013** — `updatedInput` was completely ignored in CLI
  (fixed in v1.0.24)
- **github/copilot-cli#2643** — even after the fix, `permissionDecision: "allow"`
  + `updatedInput` still shows a confirmation dialog on every rewrite. The issue
  notes that deny-with-suggestion was less disruptive because denials are handled
  gracefully by the agent without surfacing a modal to the user.

The same class of problem appears to affect VS Code Copilot Chat. Resolution is
expected on the Copilot side.

**Workaround (verified working in VS Code Copilot Chat, 2026-04-21):**

Use `permissionDecision: "deny"` with a suggestion message — the same approach
that the Copilot CLI community has recognized as the practical workaround:

```json
{
  "hookSpecificOutput": {
    "permissionDecision": "deny",
    "permissionDecisionReason": "Token savings: use `rtk git status` instead (rtk saves 60-90% tokens)"
  }
}
```

With this approach, Copilot Chat reads the reason message and automatically
retries with `rtk git status` without surfacing a dialog to the user.

KuSh:
## Summary

Fixes the root cause of #3037 properly, instead of the workaround `84aa4d6`/`0df6929` shipped, and closes two related gaps found while validating it live against real Copilot CLI (1.0.75) and VS Code Copilot Chat.

- **`permissionDecision` parity** — `handle_vscode` (`rtk hook copilot`'s PascalCase path, shared by VS Code Copilot Chat and Copilot CLI's Claude-compat entry) always asserted `permissionDecision: "allow"` or `"ask"`. Asserting `"ask"` is exactly what `#3037` reported: Copilot CLI 1.0.66+ treats it as authoritative and forces a distinct, memory-less "Hook permission request" dialog on *every* call. `84aa4d6`/`0df6929` patched the camelCase native path instead (which never had this problem) with an auto-allow heuristic — the actual bug in `handle_vscode` was never touched. Now `rtk hook copilot` only ever asserts `permissionDecision: allow` for an explicit, user-configured Allow rule — exactly the same rule Claude Code's own hook (`process_claude_payload`) and Factory Droid's hook already use. Everything else omits the field, per GitHub's own docs ("empty output uses default behavior" — hook silence is a supported, intentional no-op, not a workaround), letting Copilot's native trust-on-first-use dialog (which *does* have a working "don't ask again" option) take over.
- **Dual hook registration removed** — `rtk init --copilot` registered both a PascalCase `PreToolUse` and a camelCase `preToolUse` entry in the same file. Live testing (raw stdin capture, both hook-order and key-order reversed) showed Copilot CLI treats them as two independent, sequentially-run hooks regardless of declaration order — camelCase always runs first and does the real rewrite, PascalCase always runs second on the already-rewritten command and just defers. A redundant process spawn with no behavioral benefit, since Copilot CLI honors the PascalCase-only schema perfectly well on its own. Dropped the camelCase entry; existing installs upgrade on the next `rtk init --copilot` / `rtk init --global --copilot` re-run (`write_if_changed` overwrites unconditionally on content diff — verified with dedicated upgrade tests, not just fresh-install tests).
- **VS Code Copilot Chat tool-name gap** — found while testing the above live in actual VS Code: `detect_format` only recognized `tool_name` values `"runTerminalCommand"`/`"Bash"`/`"bash"`. A real captured payload from VS Code Copilot Chat showed it actually sends `"run_in_terminal"` — unrecognized, so the hook fell through to `PassThrough` and never fired at all for VS Code Copilot Chat (no rewrite, no decision, nothing). Added `"run_in_terminal"` to the recognized set. Fixes #1425

## Live validation

All three changes were validated against the real, currently-released binaries and both real hosts (not just unit tests):

- Built the actual `develop` HEAD (`23d1e89`) fresh as the true "before" baseline (an initially-used stale local install turned out to predate both `84aa4d6` and `0df6929` entirely).
- Confirmed via raw stdin capture that Copilot CLI 1.0.75 genuinely sends the PascalCase (`tool_name`/`tool_input.command`) payload shape when only that key is registered — not just a VS-Code-relayed artifact.
- Reproduced the exact `#3037` bug live: `develop` HEAD's literal `permissionDecision: "ask"` produces a distinct "Hook permission request" / "RTK auto-rewrite" dialog with **no** remember option, repeating on every call.
- Confirmed the fix produces Copilot's normal generic trust dialog instead, with a working "don't ask again" option — one prompt per repo, then silent, matching Claude Code's own UX.
- Confirmed real VS Code Copilot Chat (freshly installed) now correctly rewrites a compound `cd <dir> && git status --short --branch` command to `cd <dir> && rtk git status --short --branch`, with no forced `permissionDecision`.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` (2498 tests, all green)
- [x] New unit tests: `vscode_response_from_decision` decision matrix (allow/ask/deny/defer), dual-schema-upgrade tests (project + global scoped), `run_in_terminal` format detection
- [x] Live end-to-end validation against real `copilot` CLI 1.0.75 and VS Code Copilot Chat, described above

### DIFF
--- a/docs/contributing/TECHNICAL.md
+++ b/docs/contributing/TECHNICAL.md
@@ -314,7 +314,7 @@ Start here, then drill down into each README for file-level details.
 |-----------|-------|-------------------------------|
 | [`hooks/`](../hooks/README.md) | _(parent)_ | **All JSON formats**, rewrite registry overview, exit code contract, override controls |
 | [`claude/`](../hooks/claude/README.md) | Claude Code | Shell hook mechanism, `PreToolUse` JSON, test script |
-| [`copilot/`](../hooks/copilot/README.md) | GitHub Copilot | Rust binary hook, VS Code Chat vs Copilot CLI dual format |
+| [`copilot/`](../hooks/copilot/README.md) | GitHub Copilot | Rust binary hook, single `PreToolUse` schema shared by VS Code Chat and Copilot CLI |
 | [`cursor/`](../hooks/cursor/README.md) | Cursor IDE | Shell hook, empty JSON response requirement |
 | [`cline/`](../hooks/cline/README.md) | Cline / Roo Code | Rules file (prompt-level, no programmatic hook) |
 | [`windsurf/`](../hooks/windsurf/README.md) | Windsurf / Cascade | Rules file (workspace-scoped) |
@@ -331,7 +331,7 @@ RTK supports the following LLM agents through hook integrations:
 |-------|-----------|-----------|---------------------|
 | Claude Code | Shell hook | `PreToolUse` in `settings.json` | Yes (`updatedInput`) |
 | GitHub Copilot (VS Code) | Rust binary | `rtk hook copilot` reads JSON | Yes (`updatedInput`) |
-| GitHub Copilot CLI | Rust binary | `rtk hook copilot` reads JSON | No (deny + suggestion) |
+| GitHub Copilot CLI | Rust binary | `rtk hook copilot` reads JSON | Yes (`updatedInput`) |
 | Cursor | Rust binary | `rtk hook cursor` reads JSON | Yes (`updated_input`) |
 | Gemini CLI | Rust binary | `rtk hook gemini` reads JSON | Yes (`hookSpecificOutput`) |
 | Cline/Roo Code | Rules file | Prompt-level guidance | N/A (prompt) |
--- a/docs/guide/getting-started/supported-agents.md
+++ b/docs/guide/getting-started/supported-agents.md
@@ -30,7 +30,7 @@ Agent runs "cargo test"
 |-------|-----------------|---------------------------|
 | Claude Code | Shell hook (`PreToolUse`) | Yes |
 | VS Code Copilot Chat | Shell hook (`PreToolUse`) | Yes |
-| GitHub Copilot CLI | Shell hook (`preToolUse` `modifiedArgs`) | Yes |
+| GitHub Copilot CLI | Shell hook (`PreToolUse`) | Yes |
 | Cursor | Shell hook (`preToolUse`) | Yes |
 | Gemini CLI | Rust binary (`BeforeTool`) | Yes |
 | OpenCode | TypeScript plugin (`tool.execute.before`) | Yes |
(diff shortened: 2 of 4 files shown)


pszymkowiak:
Nice catch on the dual-hook-invocation — that's a real bug and the fix (plus the upgrade test) is clean.

One thing worth double-checking before merge, since it interacts with #3179 (still open): once `rtk init --copilot` stops registering the camelCase `preToolUse` entry, Copilot CLI on Windows will only ever send the PascalCase `PreToolUse` payload. Per [GitHub's own hooks reference](https://docs.github.com/en/copilot/reference/hooks-reference), that payload remaps the runtime tool name for terminal execution:

> `bash`, `powershell` → `Bash`

So both platforms report `tool_name: "Bash"` under this schema — a value already in the `matches!` list here. If that mapping is accurate for the CLI version people are actually running, `toolName: "powershell"` (the case #3179 adds) only ever existed in the camelCase schema this PR removes — meaning after this merges and someone re-runs `rtk init --copilot`, that code path becomes legacy-only (still correct for un-upgraded installs, just not exercised by new ones).

Did the live validation on 1.0.75 happen to include a Windows run? If the PascalCase `Bash` mapping already covers it, might be worth a one-line note here (or a Windows-flavored test) so #3179 doesn't get merged under a premise this PR quietly invalidates, and so whoever reads `HookFormat::CopilotCli`'s updated doc comment later knows the `powershell` case is transition-only.

KuSh:
No windows machine on my side to check that part unfortunately :confused:

BillChirico:
So this will fix RTK to work on Windows?

KuSh:
> So this will fix RTK to work on Windows?

It depends what's your exact problem with RTK on Windows but it should help for sure

BillChirico:
> > So this will fix RTK to work on Windows?
> 
> 
> 
> It depends what's your exact problem with RTK on Windows but it should help for sure

What problems does this fix and when can we expect it to be released?

pszymkowiak (approved):
LGTM — this fixes the root cause in `handle_vscode` cleanly instead of extending the previous workaround, and the removed dual-hook registration + `run_in_terminal` detection gap are both real fixes found through solid live testing.

Re: my earlier question about Windows and the camelCase hook removal — confirmed and closed by @romain-cureau's live test on Windows 11 / Copilot CLI 1.0.73: https://github.com/rtk-ai/rtk/pull/3179#issuecomment-5088268495. The PascalCase `PreToolUse` path fires and its `updatedInput` is honored end-to-end on Windows, so dropping the camelCase entry here doesn't regress Windows behavior.

CI green across the board (including windows-latest and benchmark). Approving.

PR #3213, opened 2026-07-26
fix(hook): allow simple Cursor rewrites

### LINKED ISSUE #3210: rtk hook cursor (v0.43.0, native binary path) discards successful rewrite due to exit code 3 — regression of #2719 in the non-script hook path
## Summary

`rtk hook cursor` returns `{}` (no rewrite) for commands that `rtk rewrite` successfully rewrites, because the rewrite exits with code 3 on success and the hook subcommand appears to treat any non-zero exit as "no rewrite." This looks like the same root cause as #2719, but reproduced in the native `rtk hook cursor` binary path (introduced after the hook moved away from the standalone `hooks/cursor/rtk-rewrite.sh` script), not the old shell wrapper.

## Environment

- rtk version: 0.43.0
- OS: Ubuntu (Linux)
- Editor: Cursor
- Hook config (`~/.cursor/hooks.json`):

```json
{
  "version": 1,
  "hooks": {
    "preToolUse": [
      { "command": "rtk hook cursor", "matcher": "Shell" }
    ]
  }
}
```

## Steps to reproduce

1. Confirm rewrite works at the CLI level:

```bash
$ rtk rewrite "git status"
rtk git status
$ echo "exit code: $?"
exit code: 3
```

2. Simulate what Cursor sends to the hook:

```bash
$ echo '{"tool_name": "Bash", "tool_input": {"command": "git status"}}' | rtk hook cursor
{}
```

## Expected behavior

```json
{"permission": "allow", "updated_input": {"command": "rtk git status"}}
```

## Actual behavior

```json
{}
```

## Impact

Shell commands run by the Cursor agent are never rewritten to their `rtk`-prefixed equivalents, so `rtk gain` shows no tracked savings even though the hook is correctly registered and firing. This defeats the entire purpose of the Cursor integration.

## Notes

This appears related to #2719, which fixed the same exit-code-3 handling bug in the old `hooks/cursor/rtk-rewrite.sh` shell script. It looks like the fix either didn't carry over to the newer native `rtk hook cursor` binary implementation, or a regression was introduced when the hook logic moved from the shell script into the binary itself.

cyphercodes:
## Summary
- Fix Cursor native hook simple default rewrites so `updated_input` is emitted with `permission: "allow"`, which Cursor applies instead of discarding.
- Keep explicit/compound ask paths as `permission: "ask"` so broader commands are not silently auto-allowed.
- Add regression coverage for the default `git status` Cursor hook path.

Closes #3210

## Test plan
- [x] `cargo test`
- [x] Manual testing: `printf '{"tool_name":"Bash","tool_input":{"command":"git status"}}' | cargo run --quiet -- hook cursor` outputs `{"continue":true,"permission":"allow","updated_input":{"command":"rtk git status"}}`

> **Important:** This PR targets `develop`.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -496,9 +496,16 @@ pub fn run_cursor() -> Result<()> {
             audit_log("rewrite", &cmd, &rewritten);
             cursor_allow(&rewritten)
         }
-        HookDecision::AskRewrite { rewritten, .. } => {
+        HookDecision::AskRewrite {
+            rewritten,
+            explicit,
+        } => {
             audit_log("ask", &cmd, &rewritten);
-            cursor_ask(&rewritten)
+            if !explicit && crate::discover::lexer::split_for_permissions(&cmd).len() <= 1 {
+                cursor_allow(&rewritten)
+            } else {
+                cursor_ask(&rewritten)
+            }
         }
         other => {
             if matches!(other, HookDecision::Deny) {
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for digging into the Cursor hook path. #3212, merged 2026-08-02, reworked how RTK decides permissions and removed the explicit flag on AskRewrite that this patch keys on, so the branch it adds has nothing left to hang off, and the auto-allow heuristic it restores is the one that merge deliberately dropped in favour of letting the host prompt. One gap is still real: a default-verdict Cursor rewrite is emitted with permission ask, and Cursor applies updated_input only for allow or deny, so the rewrite can be dropped; that is worth an issue or a narrow PR against the current decision model, and open issue #3634 covers the related rewrite exit-code contract. If a part of this still applies and I have missed it, please comment and the PR can be reopened.

PR #3216, opened 2026-07-26
fix(build): use GNU-correct stack linker flag on windows-gnu target

### LINKED ISSUE #3215: build.rs hardcodes MSVC-only /STACK: linker flag, breaks builds on the GNU/mingw Windows target
## Summary

`build.rs` unconditionally emits `cargo:rustc-link-arg=/STACK:8388608` inside `#[cfg(windows)]`, assuming the linker is always MSVC's `link.exe`. On the `x86_64-pc-windows-gnu` target the linker is GNU `ld` (via `x86_64-w64-mingw32-gcc`), which doesn't understand `/STACK:` syntax, so the build fails at the link step. This breaks `cargo install --git ...` (and any local build) for anyone whose default/active Rust toolchain targets `*-pc-windows-gnu` — not just MSVC-less setups, since `rustup`'s per-project or `rustup default` toolchain selection can point at the GNU target even on a machine that otherwise reports `x86_64-pc-windows-msvc` as its default host.

## Environment

- OS: Windows 11 Pro (10.0.26100)
- Rust: `rustc 1.93.0 (254b59607 2026-01-19)`, host `x86_64-pc-windows-gnu`
- Active toolchain: `1.93.0-x86_64-pc-windows-gnu`
- rtk commit: `23d1e899ce89bef54dfa7f3b6f66374c24f311e7` (build.rs last touched in `12276c174378d5fdbd7deb37f921bd8a46806cd4`, 2026-03-26)

## Steps to reproduce

1. On Windows, have an active/default Rust toolchain targeting `x86_64-pc-windows-gnu` (e.g. `rustup default stable-x86_64-pc-windows-gnu`).
2. Run `cargo install --git https://github.com/rtk-ai/rtk`.

## Expected

Build succeeds and installs `rtk.exe`.

## Actual

Link step fails:

```
error: linking with `x86_64-w64-mingw32-gcc` failed: exit code: 1
  |
  = note: "x86_64-w64-mingw32-gcc" ... "/STACK:8388608"
  = note: ...\ld.exe: cannot find /STACK:8388608: No such file or directory
          collect2.exe: error: ld returned 1 exit status

error: failed to compile `rtk v0.42.4 (...)`, intermediate artifacts can be found at `...`
```

`ld` treats `/STACK:8388608` as a positional filename argument (not recognizing the MSVC flag syntax), so it errors out looking for a file by that name.

## Root cause

https://github.com/rtk-ai/rtk/blob/23d1e899ce89bef54dfa7f3b6f66374c24f311e7/build.rs#L6-L13

```rust
#[cfg(windows)]
{
    // Clap + the full command graph can exceed the default 1 MiB Windows
    // main-thread stack during process startup. Reserve a larger stack for
    // the CLI binary so `rtk.exe --version`, `--help`, and hook entry
    // points start reliably without requiring ad-hoc RUSTFLAGS.
    println!("cargo:rustc-link-arg=/STACK:8388608");
}
```

The `/STACK:` syntax is specific to MSVC's `link.exe`. GNU `ld` (used on the `*-pc-windows-gnu` target) expects `--stack=<size>` / `-Wl,--stack,<size>`.

## Suggested fix

Branch on `CARGO_CFG_TARGET_ENV` (set by Cargo for build scripts; `"msvc"` or `"gnu"` on Windows targets) and emit the linker-appropriate flag:

```rust
#[cfg(windows)]
{
    // Clap + the full command graph can exceed the default 1 MiB Windows
    // main-thread stack during process startup. Reserve a larger stack for
    // the CLI binary so `rtk.exe --version`, `--help`, and hook entry
    // points start reliably without requiring ad-hoc RUSTFLAGS.
    // MSVC's link.exe and GNU/mingw's ld use different flag syntax for this.
    if std::env::var("CARGO_CFG_TARGET_ENV").as_deref() == Ok("msvc") {
        println!("cargo:rustc-link-arg=/STACK:8388608");
    } else {
        println!("cargo:rustc-link-arg=-Wl,--stack,8388608");
    }
}
```

I applied exactly this patch locally, built with `cargo +stable-x86_64-pc-windows-gnu install --path .`, and confirmed `rtk.exe` builds and runs correctly (`rtk --version`, `rtk hook claude`, `rtk gain` all work as expected) on the GNU target. Happy to open a PR with this change if useful.

Derviis:
## Summary

Fixes #3215.

`build.rs` unconditionally emitted the MSVC-only `/STACK:8388608` linker flag on any Windows build. On the `x86_64-pc-windows-gnu` target the linker is GNU `ld`, which doesn't understand `/STACK:` syntax — it treats it as a missing file operand and the link step fails. This breaks `cargo install --git ...` (and any local build) for anyone whose active Rust toolchain targets `*-pc-windows-gnu`.

## Change

Branch on `CARGO_CFG_TARGET_ENV` (set by Cargo for build scripts) so:
- MSVC (`link.exe`) keeps `/STACK:8388608`
- GNU (`ld`, via mingw) gets the equivalent `-Wl,--stack,8388608`

## Testing

Built and ran successfully on `x86_64-pc-windows-gnu`:

```
cargo +stable-x86_64-pc-windows-gnu install --path .
rtk --version
rtk hook claude   # confirmed via simulated PreToolUse payload
rtk gain
```

All worked as expected; no change in behavior on MSVC (flag/logic unchanged for that branch).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -9,7 +9,12 @@ fn main() {
         // main-thread stack during process startup. Reserve a larger stack for
         // the CLI binary so `rtk.exe --version`, `--help`, and hook entry
         // points start reliably without requiring ad-hoc RUSTFLAGS.
-        println!("cargo:rustc-link-arg=/STACK:8388608");
+        // MSVC's link.exe and GNU/mingw's ld use different flag syntax for this.
+        if std::env::var("CARGO_CFG_TARGET_ENV").as_deref() == Ok("msvc") {
+            println!("cargo:rustc-link-arg=/STACK:8388608");
+        } else {
+            println!("cargo:rustc-link-arg=-Wl,--stack,8388608");
+        }
     }
 
     let filters_dir = Path::new("src/filters");


Derviis:
Closing this PR — I'm not in a position to sign the CLA, so I'd rather not leave it sitting in the queue.

The bug itself is still present on `develop`, and the full patch, reproduction steps, and environment details are in #3215. It's a seven-line change to `build.rs`, so please feel free to apply it directly — no attribution needed.

PR #3221, opened 2026-07-26
fix(tsc): respect failed compiler exit status

### LINKED ISSUE #3220: tsc filter: "TypeScript: No errors found" is printed on a non-zero exit, so a tsc that never ran reads as a clean typecheck

nightcityblade:
## Summary

- Prevent the `tsc` streaming filter from reporting "No errors found" when the compiler exits non-zero without parseable diagnostics.
- Add regression coverage for an unparsed TypeScript compiler failure.
- Fixes #3220

## Test plan

- [x] Manual testing: `target/debug/rtk tsc --definitely-invalid-option` reports `TypeScript: compilation failed (exit 1)` and preserves exit code 1.

### DIFF
--- a/src/cmds/js/tsc_cmd.rs
+++ b/src/cmds/js/tsc_cmd.rs
@@ -78,8 +78,14 @@ impl BlockHandler for TscHandler {
         line.starts_with("  ") || line.starts_with('\t')
     }
 
-    fn format_summary(&self, _exit_code: i32, _raw: &str) -> Option<String> {
+    fn format_summary(&self, exit_code: i32, _raw: &str) -> Option<String> {
         if self.error_count == 0 {
+            if exit_code != 0 {
+                return Some(format!(
+                    "TypeScript: compilation failed (exit {})\n",
+                    exit_code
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the fix. This landed on develop in #2232 (merged 2026-08-29), which makes the tsc summary use the exit code so a failed compiler with no parsed diagnostics no longer reports "No errors found", and additionally prints the head and tail of the raw output so the cause is visible. The regression case in this PR, a non-zero exit with unparseable output, is covered there by several tests, so closing as already fixed. Please comment or reopen if you see a case that develop still gets wrong.

PR #3223, opened 2026-07-26
docs(hook): distinguish typed and rewritten commands

### LINKED ISSUE #3186: RTK guidance confuses agents and sidesteps allowlist
## User Story

As a Claude Code user with RTK's auto-rewrite hook installed, I want RTK's own docs/RTK.md to avoid modeling literal `rtk <cmd>` invocations for hook-covered commands, so Claude doesn't start typing rtk-prefixed commands itself and trigger avoidable permission prompts.

## Background

Claude Code's `permissions.allow` check runs against the literal Bash command the model writes, *before* the `PreToolUse` hook rewrites it. So `git status` (plain) matches an existing allow rule and runs silently — the hook then rewrites it to `rtk git status` after the fact, invisibly. But if Claude itself writes `rtk git status` (already prefixed), that string doesn't match the user's plain-command allow rules, and the user gets a permission prompt every time — defeating the "zero-friction" promise of the hook.

In practice, Claude sessions frequently do type the prefixed form themselves. The likely cause is `RTK.md`: it has a "Meta Commands (always use rtk directly)" section, immediately followed by an example — `git status → rtk git status` — meant to illustrate the hook's *automatic* rewrite. The visual pattern is easy for an LLM to key on as "the command to type" rather than "what the hook produces for you," especially since the doc never explicitly says "do not type this form yourself." The generated RTK.md ships with no distinction in formatting or emphasis between the two categories.  

The potential confusion for agents interferes with the user's configured command allow-list, because e.g. `rtk git` might not be on the allow list  even if `git` is.  

## Acceptance Criteria

- [ ] `RTK.md` (and any other agent-facing rtk docs) clearly distinguishes "commands you must type as `rtk ...`" (meta commands) from "commands the hook rewrites for you — type the plain form" (git/gh/cargo/etc.)
- [ ] The illustrative `X → Y` example is reworded or annotated so it can't be read as an instruction to type `Y`
- [ ] No regression to existing meta-command guidance (`rtk gain`, `rtk discover`, `rtk proxy`, etc. still explicitly called out as requiring the `rtk` prefix)

## Technical Notes

- Relevant file: the `RTK.md` template generated by `rtk init -g` (installed to `~/.claude/RTK.md`)
- Per the hook docs, an already-`rtk`-prefixed command passes through unmodified ("commandes deja prefixees par rtk passent sans modification") — so this isn't a hook bug, it's a documentation/prompting issue causing Claude to bypass the hook unnecessarily

nightcityblade:
## Summary

- Clarify that only RTK meta commands should be typed with the `rtk` prefix.
- Tell Claude to type hook-covered commands in their plain form, and test the generated `RTK.md` contract.

Fixes #3186

## Test plan

- [x] Manual testing: generated `RTK.md` with `rtk init -g` in a temporary Claude config directory and inspected the command guidance

### DIFF
--- a/hooks/claude/rtk-awareness.md
+++ b/hooks/claude/rtk-awareness.md
@@ -2,7 +2,7 @@
 
 **Usage**: Token-optimized CLI proxy (cuts up to 90% of bash output)
 
-## Meta Commands (always use rtk directly)
+## Meta Commands (type these with `rtk` directly)
 
 ```bash
 rtk gain              # Show token savings analytics
@@ -21,9 +21,10 @@ which rtk             # Verify correct binary
 
 ⚠️ **Name collision**: If `rtk gain` fails, you may have reachingforthejack/rtk (Rust Type Kit) installed instead.
 
-## Hook-Based Usage
+## Hook-Rewritten Commands
 
-All other commands are automatically rewritten by the Claude Code hook.
-Example: `git status` → `rtk git status` (transparent, 0 tokens overhead)
+For git, gh, cargo, and other hook-covered tools, type the plain command without an `rtk` prefix. The Claude Code hook rewrites it after permission checks.
+
+Example: type `git status`. The hook transparently runs `rtk git status`; do not type the rewritten form yourself.
 
 Refer to CLAUDE.md for full command reference.
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -4872,6 +4872,13 @@ mod tests {
     }
 
     #[test]
+    fn test_claude_rtk_md_distinguishes_meta_and_rewritten_commands() {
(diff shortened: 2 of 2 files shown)


KuSh:
Thanks for the careful wording here. The awareness rework in #3372, merged 2026-09-07, replaced both files this touches: hooks/claude/rtk-awareness.md is gone and the text installed by default is now hooks/rtk-awareness.md, which no longer mentions typing an rtk prefix at all, so the confusion reported in #3186 has nothing left to feed on. The prefix instruction survives only in the full awareness level, which is meant for agents without hook support and where typing the prefix is the right thing to do, and the RTK_SLIM constant the new test asserts on was removed in the same change. If you still see agents typing the prefixed form on the current default text, please comment and this can be reopened.

PR #3227, opened 2026-07-26
feat(tracking): record project_path on parse failures

bharvey42:
## Problem

`commands` has been stamped with the cwd since project-scoped stats landed, but `parse_failures` was not. A failure could therefore only be related to other telemetry by bare timestamp, which is ambiguous on a machine running concurrent sessions in different projects — so "which project do rtk's filters keep breaking in?" was not answerable from the data.

## Changes

- Add `parse_failures.project_path TEXT DEFAULT ''`, stamped by `record_parse_failure` via the same `current_project_path_string()` helper `record` uses. Call sites are unchanged.
- Add `get_parse_failure_summary_filtered()` and wire `--project` through `rtk gain --failures`, using the same exact-or-subdirectory GLOB scoping as the `commands` queries. The unscoped failure log now labels each recent row with its project directory, since that view spans several projects.
- Index `(project_path, timestamp)` to match `idx_project_path_timestamp`.

## Migration handling

Migrations move into a shared `migrate_schema()` that both `Tracker::new()` and the in-memory test schema call. Previously the test schema was a hand-maintained copy of the migrated one, so a new column could be tested against a shape production never has; now in-memory tests run the real migration path.

Existing rows take the column DEFAULT rather than NULL, so readers never have to tell "old row" from "unknown project".

## Privacy

No new field is reported upstream: telemetry sends `parse_failures_24h` and `projects_count`, both counts, and no raw path leaves the machine.

## Verification

Verified on a copy of a 27 MB production `history.db`: 8846 pre-existing rows migrate to `''`, newly written rows carry the cwd, and `--failures --project` scopes to the project and its subdirectories.

### DIFF
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -44,7 +44,7 @@ pub fn run(
     }
 
     if failures {
-        return show_failures(&tracker);
+        return show_failures(&tracker, project_scope.as_deref()); // changed: pass project scope
     }
 
     // Handle export formats
@@ -429,6 +429,17 @@ fn resolve_project_scope(project: bool) -> Result<Option<String>> {
     Ok(Some(canonical.to_string_lossy().to_string()))
 }
 
+/// Final path component of a project path, for compact per-row labelling. // added
+/// Returns `None` for rows recorded before `project_path` existed (empty string).
+fn project_dir_name(project_path: &str) -> Option<&str> {
+    if project_path.is_empty() {
+        return None;
+    }
+    project_path
+        .rsplit(['/', '\\'])
+        .find(|part| !part.is_empty())
+}
+
 /// Shorten long absolute paths for display. // added
 fn shorten_path(path: &str) -> String {
     let path_buf = PathBuf::from(path);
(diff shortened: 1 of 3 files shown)


bharvey42:
Rebased onto current `develop` (5e0f92c) — conflict-free again, and simplified in the process.

`develop` has since done the schema unification this PR originally carried: `Tracker::new()` and `new_in_memory()` both run one `run_schema_migrations()`, gated by `SCHEMA_VERSION` in `Tracker::new()`. So this branch no longer ships its own `migrate_schema()` helper. Instead:

- the `parse_failures.project_path` column + `idx_pf_project_path_timestamp` index are added inside `run_schema_migrations()`, and
- `SCHEMA_VERSION` is bumped 1 → 2 so existing databases actually pick the migration up on their next open (with the gate, a new `ALTER TABLE` without a bump would never run against an already-stamped DB).

Recording (`record_parse_failure` stamps cwd), the `--project` filter for `rtk gain --failures`, the `ParseFailureRecord.project_path` field, and the README are as before. The legacy-schema migration test now drives `run_schema_migrations()` directly (renamed to `test_schema_migration_adds_parse_failures_project_path`).

Full gate on Windows 11: fmt clean, clippy zero warnings, 3599 passed / 0 failed. The 3 remaining failures (`find_cmd`, `stream`, `hook_cmd`) reproduce identically on pristine `develop` under Windows — pre-existing, unrelated.

PR #3228, opened 2026-07-26
fix(tracking): stop tests writing into the real history.db

bharvey42:
## Problem

`cargo test` — which CONTRIBUTING instructs contributors to run — inserted rows into the developer's live `<data-dir>/rtk/history.db`. On one machine a single full run wrote **25 `commands` rows and 8 `parse_failures` rows**, which then show up in `rtk gain` output and in anything downstream that treats that database as real telemetry.

## Why the fix is at `tracker_db_path()` and not per-test

Every write path funnels through `Tracker::new()`, so the leak is not limited to the few tests in this module that construct a `Tracker` directly: command tests that exercise the real code path via `TimedExecution` leak too, and in practice those (`find`, `git`, `read`) accounted for the **majority** of the observed rows. Converting individual tests to `new_in_memory()` would have missed them — I counted 10 rows attributable to the direct-`Tracker` tests out of 25 total.

`tracker_db_path()` diverts to a per-process temp file under `cfg(test)`, so no present or future test can reach the real database regardless of which code path it exercises.

## Preserved behaviour

- An explicit `RTK_DB_PATH` still takes precedence.
- `get_db_path()` stays a pure configuration resolver, so `test_db_path_env_and_default` continues to assert the real precedence rules instead of the test shim.

## Verification

Full suite on `x86_64-pc-windows-msvc` leaves **zero** test rows in the real database and writes to `rtk-test-history-<pid>.db` in the temp dir instead. `cargo fmt --check` and `cargo clippy --all-targets` clean.

### DIFF
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -247,7 +247,7 @@ impl Tracker {
     /// # Ok::<(), anyhow::Error>(())
     /// ```
     pub fn new() -> Result<Self> {
-        let db_path = get_db_path()?;
+        let db_path = tracker_db_path()?;
         if let Some(parent) = db_path.parent() {
             std::fs::create_dir_all(parent)?;
         }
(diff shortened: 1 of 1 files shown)


bharvey42:
Fresh data point for this, from today: I ran `cargo test` on a different branch (a rebase of #3227), and before doing so backed up my real `history.db` specifically because of this issue. Afterwards, querying the live database:

```sql
SELECT COUNT(*) FROM parse_failures
WHERE (raw_command GLOB 'cmd_ok*' OR raw_command GLOB 'cmd_fail*'
       OR raw_command GLOB '*test_[0-9]*');
```

returns **32 all-time, 8 of them written today.** Samples, PID-suffixed straight out of the fixtures:

```
cmd_ok1_116780                    | err
cmd_fail_116780                   | err
git -C /path status test_116780   | unrecognized subcommand
```

Those sit in the same table as real telemetry and are indistinguishable from it downstream unless you know the fixture naming. My `commands` table happened to come out clean this run, so on this machine the leak is `parse_failures`-only — a slightly narrower blast radius than the 25/8 split in the description, but the same root cause.

Practical effect: anyone following CONTRIBUTING's instruction to run `cargo test` silently contaminates their own `rtk gain` numbers, and there's no signal that it happened. Backing up the database beforehand shouldn't be the workaround. Would like to see this land.

bharvey42:
Rebased onto current `develop` (5e0f92c) — conflict-free again.

`develop` has since moved DB opening into a shared `open_and_prepare()` (with `SCHEMA_VERSION`-gated migrations), so the change is now a single call-site swap inside that helper plus the `tracker_db_path()` shim — still 34 lines, same intent.

Re-measured today, since `develop` also gained `Tracker::new_in_memory()` and serialized `RTK_DB_PATH` handling and I wanted to know whether those had closed the leak on their own. Method: count rows in the real `history.db` before/after a full `cargo test --all` on a **pristine `develop` checkout**, Windows 11.

- `parse_failures`: 20077 → 20077. That table no longer leaks — good.
- `commands`: 63504 → 63527. ~20 of those rows landed inside the same second, straight out of the test fixtures: `find C:\Users\...\Temp\.tmp9T02TM -name '*'`, `find /definitely/missing/xyz -mtime +0`, `find . -name secret.txt`, `git stash show`, `git worktree list`, `cat ...\.tmpHth7z4.rs`, etc. They sit next to real telemetry in `rtk gain` and are indistinguishable from it.

Same full run with this branch applied: zero fixture rows written. So the blast radius moved from `parse_failures` to `commands` since the original description, but the root cause — every write path funnels through `Tracker::new()`, which opens the real DB under `cfg(test)` — is unchanged, and this is still the O(1) fix for the class.

Full gate on Windows: fmt clean, clippy zero warnings, 3595 passed / 0 failed (the 3 Windows-only failures on `find_cmd`, `stream`, and `hook_cmd` reproduce identically on pristine `develop`).

KuSh:
Closing in favour of #3758, which fixes the same leak at the root: it redirects rtk's whole data directory under cfg(test), so the tracking database, the tee spool, the telemetry salt, the trust store and the hook markers are all covered, and it adds a guard test for children spawned by the suite. This change covers the tracking database only, so merging #3758 makes it redundant. Review and comments are very welcome on #3758, or reply here if you had a case in mind that it does not cover.

Issue #1785, opened 2026-05-08
Prevent telemetry low_savings_commands from including arguments


LionartBR:
## Summary

Telemetry documentation says full command lines, arguments, file paths, environment variables, source code, and secrets are not collected. In `v0.39.0`, most telemetry fields are aggregate or reduced, but `low_savings_commands` can still include command arguments because it sends the first three whitespace-separated words of the stored `rtk_cmd`.

This is a privacy hardening issue, not an accusation of malware.

## Version / tag / commit

- Release: `v0.39.0`
- Commit/tag target: `2fbc7514f6964acabcfac65501b8bb6b525e3aa8`
- Producing Actions run: https://github.com/rtk-ai/rtk/actions/runs/25445836840

## Evidence

Telemetry payload:

- `src/core/telemetry.rs:106-140` includes `low_savings_commands` in the JSON payload.
- `src/core/telemetry.rs:271-278` populates it from `tracker.low_savings_commands(5)` and formats each entry as `"{cmd}:{pct}%"`.
- `src/core/tracking.rs:1053-1067` implements `low_savings_commands`; it selects full `rtk_cmd`, then returns `cmd.split_whitespace().take(3).join(" ")`.

Stored command sources can include arguments:

- `src/main.rs:1199-1204` records TOML-filtered fallback as `rtk:toml {raw_command}`.
- `src/main.rs:1227` records passthrough as `rtk fallback: {raw_command}`.
- `src/main.rs:2341-2344` records proxy usage as `rtk proxy {cmd_name} {cmd_args.join(" ")}`.

Therefore a command such as a proxy or fallback invocation with a URL/path/token-like first argument can flow into the first three words of `low_savings_commands`, depending on savings stats and ranking.

Docs:

- `docs/TELEMETRY.md:65` documents `low_savings_commands` as a sent telemetry field.
- `docs/TELEMETRY.md:109-114` says RTK does not collect full command lines or arguments, file paths, secrets, API keys, or environment variable values.

Other telemetry fields look better reduced:

- `src/core/tracking.rs:976-986` extracts only the tool name for `top_commands`.
- `src/core/tracking.rs:1024-1038` groups passthrough telemetry by first word only.
- `src/core/tracking.rs:1141-1172` maps commands to ecosystem categories for `ecosystem_mix`.
- `src/core/tracking.rs:1188-1195` sends only a count of distinct project paths.

## Impact

[...]

PR #3233, opened 2026-07-27
fix(tracking): drop argument data from telemetry low_savings_commands

kobihikri:
## Summary

- `low_savings_commands` sent the first three whitespace words of the stored `rtk_cmd`. Because the proxy and fallback record sites store the raw command line, the third word can be a path, URL, or database connection string instead of a subcommand — so argument data could reach the telemetry payload.
- This is the issue **@LionartBR** reported in #1785. All credit for finding and diagnosing it is theirs; their write-up already identified the exact call path and proposed this mitigation. This PR just implements it.
- The label is now reduced rather than truncated: rtk's own leading marker is kept, then only bare subcommand words, and a third word only when the second names a tool that actually has subcommands. `rtk git log` and `rtk docker ps` are preserved; `rtk psql postgres://user:pass@host/db -c ...` becomes `rtk psql`.

Why it matters: `docs/TELEMETRY.md:112-113` says full command lines, arguments, and file paths are never collected, and the contributor checklist at `docs/TELEMETRY.md:189` requires new telemetry fields to carry *"only aggregate counts or anonymized names — no raw paths, arguments, or user data"*. This field was the one that did not meet that bar; the sibling fields (`top_commands` via `.nth(1)`, `top_passthrough` via SQL `SUBSTR`) already reduce correctly.

I confirmed the leak is reachable rather than theoretical. With a stored command whose savings land in the field's `0 < avg < 30` window, the credential survived into the value that `send_ping` puts in the JSON body:

```
TELEMETRY low_savings_commands PAYLOAD = [("rtk psql postgres://demo_user:demo_pass@db.internal:5432/appdb", 15.0)]
```

(synthetic credentials, run in a container). On my own machine's history DB, 129 distinct stored commands have a path or URL in that third word.

Kept deliberately small and in-scope: one file, one behaviour change, no drive-by edits. If you would rather go further — e.g. @LionartBR's alternative of reducing to the tool name only, or a broader per-tool allowlist — I'm happy to change it to whatever you prefer, or to close this if you'd sooner fix it yourselves.

## Test plan

- [x] Five tests added in `src/core/tracking.rs` covering: the DSN regression from #1785, paths/flags/patterns, preserved real subcommands (`git log`, `docker ps`, `cargo build`), the `rtk:toml` / `rtk proxy` markers and empty input, plus a property test asserting no label ever contains `/`, `@`, `=`, or `.`
- [x] Full suite green: **2505 passed, 0 failed** (run in a `rust:1.91-slim` container with `git` installed — without `git` present, 5 unrelated pre-existing git tests fail for environmental reasons)

---

*Disclosure: I used AI assistance to help investigate this and draft the change. I verified every line myself, reproduced the behaviour, and ran the full test suite before opening this.*

### DIFF
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -48,6 +48,56 @@ fn current_project_path_string() -> String {
         .unwrap_or_default()
 }
 
+/// Tools whose second word is a fixed subcommand (`git log`, `docker ps`) rather than a
+/// user-supplied argument. Only for these is a third word safe to report.
+const SUBCOMMAND_TOOLS: &[&str] = &[
+    "artisan", "aws", "cargo", "docker", "dotnet", "gh", "git", "glab", "go", "gradlew", "gt",
+    "kubectl", "mvn", "npm", "pip", "pnpm", "rake", "uv", "yarn",
+];
+
+/// True for a bare tool/subcommand word — letters, digits, `_`, `-`, `:` only.
+///
+/// Rejects anything carrying `/`, `.`, `=`, `@` or a leading `-`, which is how paths,
+/// URLs, DSNs, filenames and flags are excluded.
+fn is_bare_command_word(word: &str) -> bool {
+    !word.is_empty()
+        && !word.starts_with('-')
+        && word
+            .chars()
+            .all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-' || c == ':')
+}
+
+/// Reduce a stored `rtk_cmd` to a label that is safe to put in a telemetry payload.
+///
+/// `docs/TELEMETRY.md` states that full command lines, arguments, file paths and secrets
+/// are never collected. Stored `rtk_cmd` values do contain them: the proxy and fallback
+/// record sites embed the raw command line, so `rtk psql postgres://user:pass@host/db` is
+/// a real stored value. Taking a fixed number of leading words therefore leaks argument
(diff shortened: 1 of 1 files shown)


kobihikri:
Closing this — I should not have opened it.

@YOMXXX already has [#2578](https://github.com/rtk-ai/rtk/pull/2578) open since 2026-06-24, fixing the same thing in the same file, and it is mergeable and just waiting on review. Theirs came first and is the smaller change (+61/-4 against my +123/-2). Mine adds nothing that justifies a competing PR, and a second one only makes the queue harder to read.

My mistake was in how I checked for prior work: I searched issues, found @LionartBR's #1785 and credited it, then went straight to writing a patch without searching open pull requests. #2578 was linked right under the issue title the whole time. Entirely avoidable, and I'm sorry for the noise.

Maintainers: **[#2578](https://github.com/rtk-ai/rtk/pull/2578) is the one to review.** @LionartBR reported the issue, @YOMXXX wrote the fix.

The one thing I'd still offer is the reproduction I left on [#1785](https://github.com/rtk-ai/rtk/issues/1785#issuecomment-5088024408): with a stored command whose savings land in the field's `0 < avg < 30` window, a connection-string password survives the `take(3)` reduction into the POSTed payload. That applies to whichever fix you land, and it may be worth a regression test in #2578 — but that is @YOMXXX's call, not mine.

*Disclosure: I used AI assistance on the original PR and on this comment.*

PR #3239, opened 2026-07-27
fix(build): use the GNU linker's stack flag on *-pc-windows-gnu

kamusmg:
## Summary

- `rtk` does not build at all on `x86_64-pc-windows-gnu`:

  ```
  error: linking with `x86_64-w64-mingw32-gcc` failed: exit code: 1
    = note: ld.exe: cannot find /STACK:8388608: No such file or directory
  ```

- `build.rs` emits `cargo:rustc-link-arg=/STACK:8388608`, which is **MSVC `link.exe` syntax**. GNU ld reads `/STACK:8388608` as a **file path** and aborts. CI only covers `windows-latest` (msvc), so the gnu target has been silently unbuildable.
- Emit `-Wl,--stack,8388608` there instead, keeping `/STACK:` for msvc so the original intent (a larger main-thread stack for clap's command graph) is preserved on both.
- Also switches the gate from `#[cfg(windows)]` to `CARGO_CFG_TARGET_OS`/`CARGO_CFG_TARGET_ENV`. Inside a build script `#[cfg(windows)]` describes the **host** compiling `build.rs`, not the target, so today it fires when cross-compiling *away* from Windows and misses cross-compiling *toward* it.

## Test plan

- [x] `cargo test --all` on `x86_64-pc-windows-gnu` — **2511 passed, 0 failed** (this target could not produce a binary before this change)
- [x] Manual: `rtk --version`, `rtk grep`, `rtk find`, `rtk git status` all start and run — the larger stack is still applied, just spelled for the right linker

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,13 +3,28 @@ use std::fs;
 use std::path::Path;
 
 fn main() {
-    #[cfg(windows)]
-    {
-        // Clap + the full command graph can exceed the default 1 MiB Windows
-        // main-thread stack during process startup. Reserve a larger stack for
-        // the CLI binary so `rtk.exe --version`, `--help`, and hook entry
-        // points start reliably without requiring ad-hoc RUSTFLAGS.
-        println!("cargo:rustc-link-arg=/STACK:8388608");
+    // Clap + the full command graph can exceed the default 1 MiB Windows
+    // main-thread stack during process startup. Reserve a larger stack for
+    // the CLI binary so `rtk.exe --version`, `--help`, and hook entry
+    // points start reliably without requiring ad-hoc RUSTFLAGS.
+    //
+    // The spelling is linker-specific: MSVC's link.exe takes `/STACK:<bytes>`,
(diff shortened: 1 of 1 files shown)


bharvey42:
Confirming this from the user side, and it's worse than a nicety: current `develop` (aa40853) does not link **at all** on `x86_64-pc-windows-gnu`. `build.rs` emits the MSVC-only `/STACK:8388608` under `#[cfg(windows)]`; GNU ld reads it as a file path and aborts. CI only covers `windows-latest` (msvc), so the target has been silently unbuildable on mainline.

This patch fixes it. I've been carrying it in a local fork build since 2026-08-08 and building rtk on that target daily since, with no issues. It still merges cleanly with `develop` as of today.

Worth noting for triage priority: this also blocks *validating* any other Windows PR from a gnu toolchain, since you can't build the baseline to compare against. I hit exactly that today and had to apply this patch locally just to run `cargo test` on `develop`.

PR #3244, opened 2026-07-27
refactor: replace `lazy_static` with `LazyLock`

TaKO8Ki:
## Summary

Since `LazyLock` has been stable since Rust 1.80 and RTK's MSRV is Rust 1.91, there is no longer a project-specific reason to use the external lazy_static macro. In addition, #3249 removed nested lazy_static dependency by updating colored crate.

- Replace `lazy_static!` declarations with `std::sync::LazyLock` without intended behavior changes.
- Remove `lazy_static` dependency and update related documentation.
- Remove three unused regex definitions exposed by `dead_code` checks.

## Test plan


src/cmds/cloud/aws_cmd.rs
```diff
@@ -1325,9 +1325,8 @@ fn filter_eks_cluster(json_str: &str) -> Option<FilterResult> {
     Some(FilterResult::new(text))
 }
 
-lazy_static! {
-    static ref S3_TRANSFER_RE: Regex = Regex::new(r"^(upload|download|delete|copy|move):").unwrap();
-}
+static S3_TRANSFER_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^(upload|download|delete|copy|move):").unwrap());
```
KuSh, line 1329: This might be a good opportunity to replace "unwrap" with "expect.

TaKO8Ki: So is it better to replace all uses of `unwrap` with `expect` in this pr?

aeppling (changes requested):
just some agent docs in .claude that will misslead some coding agent if used (still enforcing lazy_static)

TaKO8Ki:
@aeppling I have updated the agent docs.

aeppling (approved):

PR #3251, opened 2026-07-27
fix(ci): skip CD pipeline entry jobs on forks

### LINKED ISSUE #3237: CD pipeline fails on any fork push — pre-release job requires tags that forking doesn't copy
## What happens

Any push to `develop`/`master` on a **fork** of this repo triggers `cd.yml`, and the `pre-release` job fails immediately:

```
LATEST_TAG=$(git tag -l 'v[0-9]*.[0-9]*.[0-9]*' --sort=-version:refname | grep -v '-' | head -1)
::error::No stable release tag found
```

## Why

Forking a repo doesn't copy tags, and a plain `git fetch`/`gh repo sync` on a fork doesn't pull them either — so any fork ends up with zero `v*.*.*` tags. The `pre-release` job's version-computation step hard-requires at least one, so it can never succeed on a fork, regardless of what the push actually contains. I hit this simply by adding an unrelated workflow file to my fork's `develop` branch — first push, immediate red X.

The rest of the pipeline (`build-prerelease`, `release-please`, `build-release`, `update-latest-tag`) already cascades to `skipped` correctly once its `needs.*.outputs` are empty — only the two entry-point jobs (`pre-release`, `release-please`) actually attempt to run and fail on a fork.

## Suggested fix

Guard both entry-point jobs in `cd.yml` with `github.repository_owner == 'rtk-ai'` (or equivalent), so the release/CD pipeline only ever runs on this repo, not on forks. Verified locally on my own fork that this fully resolves it — the same push that previously failed reports `skipped` instead, with no behavior change on the real repo (the added condition is already true there).

Happy to open a PR with this change if that's the preferred way to receive it — wanted to check first since I wasn't sure whether CI/workflow changes go through issues first here.

Nithin078:
## Summary

- Guard `pre-release` and `release-please` in `cd.yml` with `github.repository_owner == ''rtk-ai''` so the CD/release pipeline only runs on the canonical repo.
- Forks do not inherit release tags, so the pre-release version step always fails with `No stable release tag found` on any develop/master push. Downstream jobs already skip when entry outputs are empty; only these two entry points were failing.

Fixes #3237

## Test plan

- [x] Verify YAML conditions: both entry jobs require `github.repository_owner == ''rtk-ai''` in addition to existing branch/event gates
- [ ] On a fork push to `develop`: CD workflow runs but `pre-release` is **skipped** (no red X)
- [ ] On `rtk-ai/rtk` push to `develop`: `pre-release` still runs as before
- [ ] On `rtk-ai/rtk` push to `master`: `release-please` still runs as before

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -19,9 +19,11 @@ jobs:
   # ═══════════════════════════════════════════════
 
   pre-release:
+    # Forks don't inherit tags; skip CD on non-canonical repos (see #3237)
     if: >-
-      github.ref == 'refs/heads/develop'
-      || (github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/master')
+      github.repository_owner == 'rtk-ai'
+      && (github.ref == 'refs/heads/develop'
+      || (github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/master'))
     runs-on: ubuntu-latest
     outputs:
       tag: ${{ steps.tag.outputs.tag }}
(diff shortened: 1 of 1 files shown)


Nithin078:
Hey @pszymkowiak, could you review and verify this pr please? Thank you.

PR #3254, opened 2026-07-27
refactor: simplify regex initialization

TaKO8Ki:
## Summary

This is a follow-up to #3244.

- Replace the fixed `OnceLock<Regex>` in `AggregatedTestResult::parse_line()` with `LazyLock<Regex>`, allowing repeated calls to reuse one compiled regex.
- Replace the two function-local static regex caches in `filter_cargo_nextest()` with ordinary local regexes.

## Test plan

### DIFF
--- a/src/cmds/rust/cargo_cmd.rs
+++ b/src/cmds/rust/cargo_cmd.rs
@@ -10,7 +10,7 @@ use serde::Deserialize;
 use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::ffi::OsString;
-use std::sync::OnceLock;
+use std::sync::LazyLock;
 
 #[derive(Debug, Clone)]
 pub enum CargoCommand {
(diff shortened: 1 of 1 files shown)


src/cmds/rust/cargo_cmd.rs
```diff
@@ -620,18 +620,12 @@ fn flush_failure_block(header: &mut String, body: &mut Vec<String>, failures: &m
...
-        regex::Regex::new(
-            r"Summary \[\s*([\d.]+)s\]\s+(\d+) tests? run:\s+(\d+) passed(?:,\s+(\d+) failed)?(?:,\s+(\d+) skipped)?"
-        ).expect("invalid nextest summary regex")
-    });
-
-    static STARTING_RE: OnceLock<regex::Regex> = OnceLock::new();
-    let starting_re = STARTING_RE.get_or_init(|| {
-        regex::Regex::new(r"Starting \d+ tests? across (\d+) binar(?:y|ies)")
-            .expect("invalid nextest starting regex")
-    });
```
TaKO8Ki, line 634: They are used once in a process, so it seems unnecessary to use static for them.

KuSh: Not sure it is worth it to avoid them. If something reuse them elsewhere it will be hard to catch before we get report that something is wrong.

KuSh (approved):

src/cmds/rust/cargo_cmd.rs
TaKO8Ki: Thanks for the feedback. I think `filter_cargo_nextest` is only expected to run once per process, so the static cache provides little benefit here. Local variables better match the actual lifetime and avoid keeping the regexes alive until process exit, even if the memory impact is very small.

On the other hand, `RE` defined in parse_line below is useful.

PR #3259, opened 2026-07-27
fix(grep): free -m for GNU --max-count instead of rtk --max

breisnerlopez:
## Summary

- `rtk grep`'s `-m` short (derived from `--max`, a display cap) collided with GNU grep where **`-m N` = `--max-count`** (stop after N matches per file). `rtk grep -m N <pat>` set RTK's display cap instead of forwarding `--max-count`, so grep never actually stopped early — a silent semantic mismatch (no parse error).
- Fix: drop the short. `-m` now flows to `extra_args` and is forwarded verbatim to the real `grep`/`rg` (it's not in `has_format_flag`, so it takes the normal engine path, not passthrough), which applies genuine `--max-count` while RTK still compacts. This makes `Grep` consistent with `Rg`, which never had the short and already forwarded `-m` correctly. `--max` keeps its long form and `default_value = "200"`.
- Docs: drop the `-m` short from the grep options table and correct its stale default (`50` → `200`, matching the code).

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — fmt clean, clippy 0 warnings, **2502 passed / 0 failed**.
- [x] New tests:
  - `test_try_parse_grep_dash_m_is_max_count` — clap parses `grep -m 5 <pat> <file>`, keeps `max` at 200, puts `-m 5` in `extra_args`.
  - `dash_m_max_count_matches_grep_n` — end-to-end: `rtk grep -m N` is byte-identical to `grep -n -m N` (`-m` leading, trailing, and N ≥ total).
  - `dash_m_max_count_is_per_file_like_grep_n` — multi-file, since `--max-count` is per file.
- [x] Manual: before the fix `rtk grep -m 2 . <file>` showed every line (`-m` swallowed); after, it stops at 2 like `grep -m 2`.

> Same class of GNU short-flag collision as the `-l` fix in #3258 (submitted separately per the single-focus PR rule). A related `-t` short (`--file-type`) also shadows ripgrep's `-t`/`--type`; left out of scope here and worth a follow-up.

### DIFF
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -216,7 +216,7 @@ rtk grep <pattern> [chemin] [options]
 | Option | Court | Defaut | Description |
 |--------|-------|--------|-------------|
 | `--max-len` | `-l` | 80 | Longueur maximale de ligne |
-| `--max` | `-m` | 50 | Nombre maximum de resultats |
+| `--max` |  | 200 | Nombre maximum de resultats (pas de raccourci, `-m` est reserve a `grep --max-count`) |
 | `--context-only` |  | non | Afficher uniquement le contexte du match (pas de raccourci, `-c` est reserve a `grep --count`) |
 | `--file-type` | `-t` | tous | Filtrer par type (ts, py, rust, etc.) |
 | `--line-numbers` | `-n` | oui | Numeros de ligne (toujours actif) |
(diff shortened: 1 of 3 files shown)


KuSh (approved):
Hi, LGTM thanks!

PR #3266, opened 2026-07-27
fix(tee): hash long recovery-file slugs to prevent collisions and shorten hints

breisnerlopez:
## Summary

- **`fix(tee)`**: recovery-file slugs (an embedded path that duplicates the command) are now hashed when long. This fixes a latent collision where two sibling paths truncated in the same second overwrote each other's recovery file, and it shortens the truncation hint by ~12 tokens.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — clean on current `develop`.
- [x] Unit tests cover slug hashing and the sibling-path collision case (distinct recovery files for paths that previously collided).

> Scope note: split out of the earlier bundled PR #3064 as its own focused PR (per the single-focus rule), rebased onto current `develop`. It is independent of the `git show` blob-filter feat (submitted separately).

### DIFF
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -14,8 +14,10 @@ const DEFAULT_MAX_FILES: usize = 20;
 const DEFAULT_MAX_FILE_SIZE: usize = 1_048_576;
 
 /// Sanitize a command slug for use in filenames.
-/// Replaces non-alphanumeric chars (except underscore/hyphen) with underscore,
-/// truncates at 40 chars.
+/// Replaces non-alphanumeric chars (except underscore/hyphen) with underscore.
+/// Long slugs (usually an embedded file path that duplicates the command the LLM
+/// already issued) collapse to a short readable prefix plus a collision-resistant
+/// hash, keeping recovery filenames unique but compact — fewer tokens per tee hint.
 fn sanitize_slug(slug: &str) -> String {
     let sanitized: String = slug
         .chars()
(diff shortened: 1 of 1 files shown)


src/core/tee.rs
```diff
@@ -27,11 +29,25 @@ fn sanitize_slug(slug: &str) -> String {
...
+/// tag. Distinct slugs get distinct tags (~1-in-16M collision), so shortening never
+/// makes two different commands share a tee filename (the epoch already scopes writes
+/// to the same second, exactly as before).
+fn short_hash(s: &str) -> String {
+    use sha2::{Digest, Sha256};
+    Sha256::digest(s.as_bytes())
+        .iter()
+        .take(3)
+        .map(|b| format!("{:02x}", b))
+        .collect()
```
KuSh:
```suggestion
    format!("{:x}", Sha256::digest(s.as_bytes()))[..6].to_string()
```

src/core/tee.rs
```diff
@@ -27,11 +29,25 @@ fn sanitize_slug(slug: &str) -> String {
...
+    const MAX_READABLE: usize = 24;
+    if sanitized.len() <= MAX_READABLE {
+        return sanitized;
     }
+    let prefix: String = sanitized.chars().take(8).collect();
+    format!("{}_{}", prefix, short_hash(&sanitized))
+}
+
+/// First 6 hex chars (24 bits) of the SHA-256 of `s` — a compact, collision-resistant
+/// tag. Distinct slugs get distinct tags (~1-in-16M collision), so shortening never
```
KuSh: I wouldn’t call that collision-resistant. With only 24 bits, you have about 2^24 possible outputs, and the birthday bound says a collision becomes likely after around 1.2 * 2^12 samples, roughly 5,000.
For your use it's still safe as it will also require the same exact prefix and second epoch, but you clearly can't say  "collision-resistant"

KuSh (changes requested):
Hi, LGTM except for some nitpicks

src/core/tee.rs
breisnerlopez: Applied in 9ee1532 — took the suggestion verbatim.

src/core/tee.rs
breisnerlopez: Fair — dropped the "collision-resistant" wording. The doc now states the 24-bit reality (birthday collision after a few thousand distinct slugs) and that safety here comes from also requiring the identical readable prefix and the same epoch second, not from the hash width alone. 9ee1532.

KuSh (approved):
LGTM thanks!

PR #3269, opened 2026-07-28
fix(git-diff): emit hunk lines at column 0 so `^-` anchors again

iliaal:
A grep anchored to the start of a line finds nothing in `rtk git diff` output:

```
raw:  git diff | grep -c "^-"   -> 3
rtk:  git diff | grep -c "^-"   -> 0
```

`compact_diff` indents every hunk-body line two spaces to nest it under the filename (`format!("  {}", line)`), which moves the content off column 0 and breaks the `^` anchor.

Two things worth stating precisely, because both were initially diagnosed wrong on our side:

- **Nothing is dropped.** The content is on screen the whole time, and `--stat` plus the per-file `+N -M` tally report accurately. `grep -c DELETED_LINE_A` (unanchored) returns 1 both raw and filtered. It is a rendering issue, not data loss.
- **It is not deletion-specific.** `^+` breaks identically — measured raw `^-`=3 / `^+`=2 versus rtk `^-`=0 / `^+`=0. There is no `+`/`-` asymmetry.

The cost is a silent false negative for any anchored grep over a diff: auditing "was a test expectation weakened" with `git diff tests/ | grep "^-"` returns empty while `--stat` reports deletions in the same range, and empty is exactly the answer that ends the investigation.

**Fix:** hunk bodies emit at column 0 in git's own unified shape. rtk's own annotations keep the indent, so the same greps never count them as diff lines.

After:

```
rtk:  git diff | grep -c "^-"   -> 2
```

2 is the exact deletion count; raw's 3 includes the `--- a/f.txt` header, so the piped grep was never a correct count even unfiltered.

## What the anchored audit does and does not promise

Review surfaced three holes in the original "`grep -c '^-'` gives you the exact deletion count" claim. All three are closed, two by fixing the code and one by stating the limit here.

**Content that looks like a file header.** The `+++` / `---` guards ran inside the hunk-body branch, where they can only match content: git emits the file headers before the first `@@`. So an added `++i;` or a removed `-- sql comment` was dropped from the body *and* left out of the tally, and a diff doing both rendered as a bare hunk header with nothing under it. The guards are gone.

**Combined diffs.** `diff --cc` sections, which git emits for unmerged paths during a merge, rebase, cherry-pick, or stash pop, never matched the `diff --git` reset, so `in_hunk` stayed set across sections and the `---` / `+++` headers of every file after the first reached the body. Once hunk bodies emit at column 0, those headers anchor too: a two-file conflict with zero deletions returned `^-` = 1. The reset fires on any `diff --` prefix now, so the headers land before the first `@@` again and the same repo returns 0.

**Combined diffs put the marker in column 2.** `diff --cc` carries one marker column per parent, so a line changed against only one of them (` +ours`, ` -theirs`) does not anchor. It is counted in the `+N -M` tally now, but `grep "^+"` cannot see it without rewriting the line, so `docs/usage/FEATURES.md` states the limit next to the others. Against `git diff --numstat` on three complete commit ranges the tally is exact.

**Past 100 change lines in a hunk, the count plateaus.** A 300-deletion hunk yields `grep -c "^-"` = 100 while the tally correctly says `+300 -300`. The truncation note carries the remainder and is split by sign — `... (200 deletions truncated)` — but it is indented by design, so the anchored grep cannot see it. If you gate CI on `grep -c "^-"`, gate on the `+N -M` tally instead, or pass `--no-compact`.

Two notes on the budgets. Exempting leading context from `max_lines` makes it a soft cap: output can exceed it by up to 10%, so the `git stash show -p` call site that passes 100 can see 110. Separately, and pre-existing on `develop`, the previous file's tally and the next file's header are pushed before the budget is checked, so a many-file diff can carry two entries past `max_lines`. And a hunk now ends when the line counts in its `@@` header are consumed, so trailing text after a diff — an mbox envelope, a `-- ` signature, prose — is outside every hunk rather than read as its body. `gh pr diff --patch` additionally passes through unfiltered: GitHub's `.patch` is an mbox, not a unified diff.

**A hunk's declared length is per parent, and a combined hunk spends one budget per marker column.** In a combined diff a space means opposite things depending on the line: on a change line the line is in that parent unchanged, on a removal line it is absent from it. Collapsing the columns into one add/delete pair lost that, so the first parent's budget never converged on real conflict output and the hunk never closed by count — everything above about keeping envelopes and signatures out of the body silently did not apply to `diff --cc`, which is what git emits for every unmerged path. The reverse case closed early and dropped the body: `@@@ -1,2 -1,4 +1,2 @@@` over two removals from the second parent alone rendered as a bare hunk header, with no tally and no truncation note. A header whose range count disagrees with its column count is reconciled against the columns, since only a column can be charged.

**Word diffs pass through unfiltered.** `--word-diff`, `--word-diff-regex` and `--color-words` drop the marker column and put `[-removed-]` / `{+added+}` inline, so a body line is arbitrary content in the position the compactor reads as a marker. Slicing those bytes as bytes rather than as a `str` fixes the crash on a leading multi-byte character — `rtk git diff --word-diff` exited 101 with empty stdout — but not the misparse: a row starting `+` counts as an addition and spends the wrong budget, one starting `\` is dropped as a no-newline annotation, and a real `git show --word-diff=plain` row reading `diff --git [-a/src/main.rs...` opens a new file section. There is no faithful compaction of a word diff, so `run_diff`, `run_show` and `git stash show -p` hand these modes straight through. `rtk pipe --filter git-diff` has no flags to read and still misparses word-diff text piped into it; `compact_diff` reads a unified or combined diff, and that is now what its callers give it.

**Section paths are read without guessing at the separator.** Under the default `core.quotepath` git escapes a non-ASCII path and wraps it in `"`, which removes the ` b/` the plain form splits on, so the fallback returned both paths as the section header. And ` b/` is not a reliable separator anyway: it appears inside the path for a file under a directory named `x b`, and `--no-prefix` or a custom `--src-prefix` leaves none at all. Anything but a rename names the same path twice, so the halves are the same length and the separating space sits dead centre; splitting there covers all of those and does not care what the prefixes are. A rename's halves differ in length and fall through to the ` b/` split, which names the destination.

Leading context is also emitted now. It was silently dropped before the first `+` or `-` of each hunk, which contradicted the "nothing is dropped" claim above. It keeps the last three lines before the change rather than the first, so the body is contiguous with it under `-U10` or `--function-context`, and is capped at `max_lines / 10` across the diff and exempt from `max_lines` so it cannot displace change lines. On a five-commit diff of this repo it costs 0.1 points of byte reduction against master (71.7% versus 71.8%) while taking anchorable change lines from 0 to 201.

This does not make the output an applyable patch — rtk still drops the `diff --git` / `---` / `+++` headers, so a patch parser rejects it regardless. Use `rtk proxy git diff` when you need a faithful patch. The indent was never what made that case fail.

`docs/usage/FEATURES.md` is updated to real captured output; its illustration showed change lines indented under a `src/main.rs (+5/-2)` header that no version of rtk emits.

## Testing

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all` are clean on this branch, apart from two `hooks::rewrite_cmd::tests::unattestable_passthrough` failures that reproduce with this branch's changes stashed and that it does not touch. `cmds::git::git` is 119 tests.

Ordinary unified diffs come out byte-identical to the previous head: checked on a 26,561-line real diff of this repo, on a 390-line one, and on a 300-case generated corpus covering additions, deletions, replacements, empty rows, Unicode content, leading `+` / `-` / `\`, multiple hunks and multiple files. The one intentional divergence is a hunk that ends without a change line, which now emits its buffered context instead of a bare header.

### DIFF
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -332,17 +332,72 @@ rtk git diff [args...]    # Supporte --stat, --cached, --staged, etc.
 ```
 
 **Avant / Apres :**
+
+`git diff` brut, 21 lignes :
 ```
-# git diff (~100 lignes)                    # rtk git diff (~25 lignes)
-diff --git a/src/main.rs b/src/main.rs      src/main.rs (+5/-2)
-index abc123..def456 100644                    +  let config = Config::load()?;
---- a/src/main.rs                              +  config.validate()?;
-+++ b/src/main.rs                              -  // old code
-@@ -10,6 +10,8 @@                              -  let x = 42;
-   fn main() {                               src/git.rs (+1/-1)
-+    let config = Config::load()?;              ~  format!("ok {}", branch)
-...30 lignes de headers et contexte...
+diff --git a/git.rs b/git.rs
+index 50f7a19..225f918 100644
+--- a/git.rs
++++ b/git.rs
+@@ -1,3 +1,3 @@
+ fn helper(branch: &str) -> String {
+-    format!("ko {}", branch)
++    format!("ok {}", branch)
+ }
+diff --git a/main.rs b/main.rs
+index 765936a..7abfb4f 100644
+--- a/main.rs
++++ b/main.rs
+@@ -1,5 +1,5 @@
+ fn main() {
+-    let x = 42;
+-    // old code
++    let config = Config::load()?;
++    config.validate()?;
+     println!("start");
+ }
 ```
+
+`rtk git diff`, 15 lignes :
+```
+ git.rs  | 2 +-
+ main.rs | 4 ++--
+ 2 files changed, 3 insertions(+), 3 deletions(-)
+
+Changes:
+
+git.rs
+@@ -1,3 +1,3 @@
+ fn helper(branch: &str) -> String {
+-    format!("ko {}", branch)
++    format!("ok {}", branch)
+ }
+  +1 -1
+
+main.rs
+@@ -1,5 +1,5 @@
+ fn main() {
+-    let x = 42;
+-    // old code
++    let config = Config::load()?;
++    config.validate()?;
+     println!("start");
+ }
+  +2 -2
+```
+
+Les lignes de hunk et les en-tetes `@@` sortent en colonne 0, dans la forme
+unifiee de git, donc `rtk git diff | grep "^-"` fonctionne. Les annotations
+propres a rtk restent indentees de deux espaces pour que ces memes greps ne
+les comptent pas : le total par fichier (`  +2 -2`) et la note de troncature
+(`  ... (30 deletions, 20 additions truncated)`).
+
+Deux limites de l'audit ancre. Au-dela de 100 lignes de changement par hunk,
+`grep -c "^-"` plafonne a ce qui est affiche ; la note de troncature indique
+le reste, par signe, mais elle est indentee et donc invisible au meme grep.
+Et la sortie n'est pas un patch applicable : rtk supprime les en-tetes
+`diff --git` / `---` / `+++`. Utilisez `rtk proxy git diff` pour un patch
+fidele.
 
 ---
 
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -358,9 +358,9 @@ index 765936a..7abfb4f 100644
  }
 ```
 
-`rtk git diff`, 15 lignes :
+`rtk git diff`, 24 lignes :
 ```
- git.rs  | 2 +-
+git.rs  | 2 +-
  main.rs | 4 ++--
  2 files changed, 3 insertions(+), 3 deletions(-)
 
@@ -386,18 +386,29 @@ main.rs
   +2 -2
 ```
 
+Sur un diff aussi petit, rtk economise peu : 440 octets bruts contre 392
+filtres, soit 11 %. La compression vient des diffs ou les en-tetes par fichier
+et les plafonds par hunk pesent. Sur un diff de 4 fichiers dont un hunk de 300
+lignes changees, la sortie passe de 335 a 140 lignes et de 13 262 a 4 880
+octets, soit 63 % de reduction, avec la note de troncature et le rappel
+`[full diff: rtk git diff --no-compact]`.
+
 Les lignes de hunk et les en-tetes `@@` sortent en colonne 0, dans la forme
 unifiee de git, donc `rtk git diff | grep "^-"` fonctionne. Les annotations
 propres a rtk restent indentees de deux espaces pour que ces memes greps ne
 les comptent pas : le total par fichier (`  +2 -2`) et la note de troncature
 (`  ... (30 deletions, 20 additions truncated)`).
 
-Deux limites de l'audit ancre. Au-dela de 100 lignes de changement par hunk,
+Trois limites de l'audit ancre. Au-dela de 100 lignes de changement par hunk,
 `grep -c "^-"` plafonne a ce qui est affiche ; la note de troncature indique
 le reste, par signe, mais elle est indentee et donc invisible au meme grep.
-Et la sortie n'est pas un patch applicable : rtk supprime les en-tetes
-`diff --git` / `---` / `+++`. Utilisez `rtk proxy git diff` pour un patch
-fidele.
+Sur un diff combine (`diff --cc`, ce que git produit pour chaque chemin en
+conflit pendant un merge, rebase, cherry-pick ou stash pop), le marqueur peut
+occuper la deuxieme colonne : ` +ours` est compte dans le total `+N -M` mais
+reste invisible a `grep "^+"`. Et la sortie n'est pas un patch applicable :
+rtk supprime les en-tetes `diff --git` / `---` / `+++`. Utilisez
+`rtk proxy git diff` pour un patch fidele, ou `rtk gh pr diff --patch`, qui
+passe sans filtrage.
 
 ---
 
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -126,7 +126,7 @@ fn run_diff(
         .any(|arg| arg == "--stat" || arg == "--numstat" || arg == "--shortstat");
 
     // Check if user wants compact diff (default RTK behavior)
-    let wants_compact = !args.iter().any(|arg| arg == "--no-compact");
+    let wants_compact = !args.iter().any(|arg| arg == "--no-compact") && !emits_word_diff(args);
 
     if wants_stat || !wants_compact {
         // User wants stat or explicitly no compacting - pass through directly
@@ -236,7 +236,7 @@ fn run_show(
     // pass through directly to avoid duplicated output from compact-show steps.
     let wants_blob_show = args.iter().any(|arg| is_blob_show_arg(arg));
 
-    if wants_stat_only || wants_format || wants_blob_show {
+    if wants_stat_only || wants_format || wants_blob_show || emits_word_diff(args) {
         let mut cmd = git_cmd(global_args);
         cmd.arg("show");
         for arg in args {
@@ -330,6 +330,26 @@ fn run_show(
     Ok(0)
 }
 
+/// Whether these args make git emit a word diff rather than a line diff.
+///
+/// `compact_diff` reads a unified or combined diff: a body line's first column
+/// (or columns) is a marker and the rest is content. A word diff drops the
+/// marker entirely and puts `[-removed-]` / `{+added+}` inline, so its body
+/// lines are arbitrary content in the marker position. A line starting with `+`
+/// then counts as an addition, one starting with `\` is dropped as a
+/// no-newline annotation, and one whose content happens to start `diff --`
+/// opens a new file section. There is nothing to compact faithfully, so these
+/// modes pass through.
+fn emits_word_diff(args: &[String]) -> bool {
+    args.iter().any(|arg| {
+        arg == "--word-diff"
+            || arg.starts_with("--word-diff=")
+            || arg.starts_with("--word-diff-regex")
+            || arg == "--color-words"
+            || arg.starts_with("--color-words=")
+    })
+}
+
 fn is_blob_show_arg(arg: &str) -> bool {
     // Detect `rev:path` style arguments while ignoring flags like `--pretty=format:...`.
     !arg.starts_with('-') && arg.contains(':')
@@ -353,16 +373,50 @@ fn diff_header_path(line: &str) -> String {
     {
         return path.to_string();
     }
-    match line.split(" b/").nth(1) {
+    let rest = match line.splitn(3, ' ').nth(2) {
+        Some(rest) => rest,
+        None => return "unknown".to_string(),
+    };
+    if let Some(path) = same_path_twice(rest) {
+        return path.to_string();
+    }
+    match rest.split(" b/").nth(1) {
         Some(path) => path.to_string(),
-        None => {
-            let rest = line.splitn(3, ' ').nth(2).unwrap_or("unknown");
-            // A single-path header (`diff --cc`) carries the quoting too.
-            rest.strip_prefix('"')
-                .and_then(|r| r.strip_suffix('"'))
-                .unwrap_or(rest)
-                .to_string()
-        }
+        // A single-path header (`diff --cc`) carries the quoting too.
+        None => rest
+            .strip_prefix('"')
+            .and_then(|r| r.strip_suffix('"'))
+            .unwrap_or(rest)
+            .to_string(),
+    }
+}
+
+/// The path a `diff --git` header names twice, split at the midpoint.
+///
+/// Anything but a rename names the same path on both sides, so the two halves
+/// are the same length and the separating space sits dead centre. Splitting
+/// there instead of on the first ` b/` keeps a path that contains that
+/// substring — a file under a directory named `x b` — and it does not care what
+/// the prefixes are, so `--no-prefix` and a custom `--src-prefix` work too.
+///
+/// `None` for a rename, whose halves differ in length, and for a single-path
+/// header; both fall through to the ` b/` split.
+fn same_path_twice(rest: &str) -> Option<&str> {
+    if rest.len().is_multiple_of(2) {
+        return None;
+    }
+    let mid = rest.len() / 2;
+    // A space at the midpoint is a char boundary, so both halves are valid.
+    if rest.as_bytes().get(mid) != Some(&b' ') {
+        return None;
+    }
+    let (left, right) = (&rest[..mid], &rest[mid + 1..]);
+    if left == right {
+        return Some(left);
+    }
+    match (left.strip_prefix("a/"), right.strip_prefix("b/")) {
+        (Some(p1), Some(p2)) if p1 == p2 => Some(p1),
+        _ => None,
     }
 }
 
@@ -434,7 +488,7 @@ fn parse_hunk_header(line: &str) -> Option<HunkHeader> {
     }
     let body = line[at_run..].split('@').next()?;
 
-    let mut parents = Vec::new();
+    let mut parents: Vec<usize> = Vec::new();
     let mut new = None;
     for group in body.split_whitespace() {
         let Some(rest) = group.strip_prefix(['-', '+']) else {
@@ -453,11 +507,22 @@ fn parse_hunk_header(line: &str) -> Option<HunkHeader> {
         }
     }
 
+    // `@@` has one marker column, `@@@` two, and so on for more parents.
+    let prefix_width = at_run - 1;
+    // A well-formed header lists exactly one range per marker column. When it
+    // does not, only the columns can be charged, so trust them: an untracked
+    // parent would otherwise sit at its declared count forever and the hunk
+    // would never close, while a parent with no column of its own would be
+    // charged against nothing. A missing range gets `usize::MAX`, which keeps
+    // the hunk open to the next header rather than dropping its body.
+    if parents.len() != prefix_width {
+        parents.resize(prefix_width, usize::MAX);
+    }
+
     Some(HunkHeader {
         parents,
         new: new.unwrap_or(0),
-        // `@@` has one marker column, `@@@` two, and so on for more parents.
-        prefix_width: at_run - 1,
+        prefix_width,
     })
 }
 
@@ -2190,8 +2255,10 @@ fn run_stash(
                 return Ok(result.exit_code);
             }
 
-            let filtered = if patch_mode {
+            let filtered = if patch_mode && !emits_word_diff(args) {
                 compact_diff(&result.stdout, 100)
+            } else if patch_mode {
+                result.stdout.clone()
             } else {
                 compact_stash_stat(&result.stdout)
             };
@@ -2898,6 +2965,81 @@ mod tests {
             diff_header_path(r#"diff --git "a/Ã©t.txt" b/plain.txt"#),
             "plain.txt"
         );
+    }
+
+    #[test]
+    fn test_emits_word_diff_detects_every_form() {
+        for flag in [
+            "--word-diff",
+            "--word-diff=plain",
+            "--word-diff=porcelain",
+            "--word-diff-regex=.",
+            "--color-words",
+            "--color-words=.",
+        ] {
+            assert!(
+                emits_word_diff(&[flag.to_string()]),
+                "{} must pass through",
+                flag
+            );
+        }
+        assert!(!emits_word_diff(&["--stat".to_string()]));
+        assert!(!emits_word_diff(&["-U10".to_string()]));
+        assert!(!emits_word_diff(&[]));
+    }
+
+    #[test]
+    fn test_parse_hunk_header_reconciles_ranges_with_marker_columns() {
+        // A well-formed header lists one range per marker column. When it does
+        // not, only the columns can be charged. A missing range must not leave
+        // an untracked parent holding the hunk open forever, and an extra one
+        // must not sit at its declared count with no column to spend it.
+        let h = parse_hunk_header("@@@ -1 +1 @@@").expect("two columns, one range");
+        assert_eq!(h.prefix_width, 2);
+        assert_eq!(h.parents, vec![1, usize::MAX]);
+
+        let h = parse_hunk_header("@@@ -1 -1 -1 +0,0 @@@").expect("two columns, three ranges");
+        assert_eq!(h.prefix_width, 2);
+        assert_eq!(h.parents, vec![1, 1]);
+    }
+
+    #[test]
+    fn test_compact_diff_extra_range_does_not_strand_a_hunk() {
+        // With the third range untracked, `--x` left it at 1 forever, so the
+        // hunk never closed and the mbox signature became its content.
+        let out = compact_diff("diff --cc f\n@@@ -1 -1 -1 +0,0 @@@\n--x\n-- \n2.40.0\n", 100);
+        assert!(out.contains("--x"), "got:\n{}", out);
+        assert!(!out.contains("2.40.0"), "got:\n{}", out);
+        assert!(!out.contains("-- "), "got:\n{}", out);
+        // One line, removed from both parents, is one deletion.
+        assert!(out.contains("+0 -1"), "got:\n{}", out);
+    }
+
+    #[test]
+    fn test_compact_diff_missing_range_keeps_the_body() {
+        // One range for two columns: the untracked parent gets `usize::MAX`, so
+        // the hunk stays open to the next header rather than closing early and
+        // dropping ` -lost`.
+        let out = compact_diff("diff --cc f\n@@@ -1 +1 @@@\n +kept\n -lost\n", 100);
+        assert!(out.contains(" +kept"), "got:\n{}", out);
+        assert!(out.contains(" -lost"), "got:\n{}", out);
+        assert!(out.contains("+1 -1"), "got:\n{}", out);
+    }
+
+    #[test]
+    fn test_diff_header_path_splits_the_pair_at_its_midpoint() {
+        // A file under a directory named `x b` puts the ` b/` separator inside
+        // the path, so the first match is the wrong one.
+        assert_eq!(diff_header_path("diff --git a/x b/y b/x b/y"), "x b/y");
+        // `--no-prefix` and custom prefixes leave no ` b/` at all.
+        assert_eq!(diff_header_path("diff --git x x"), "x");
+        assert_eq!(
+            diff_header_path("diff --git src/main.rs src/main.rs"),
+            "src/main.rs"
+        );
+        // A rename's halves differ in length, so the ` b/` split still names
+        // the destination.
+        assert_eq!(diff_header_path("diff --git a/old.txt b/new.txt"), "new.txt");
     }
 
     #[test]
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -374,12 +374,21 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
             hunk_shown = 0;
             // Preserve the full unified diff hunk header, including trailing
             // function / symbol context after the second @@ marker.
-            result.push(format!("  {}", line));
+            //
+            // Hunk-body lines below are emitted at column 0, in git's own
+            // unified shape. An earlier revision indented them two spaces to
+            // nest them under the filename, which silently broke every
+            // `git diff | grep '^-'` audit: the content was all there, but the
+            // `^` anchor no longer matched, so "was anything removed?" answered
+            // a confident, wrong "no". rtk's own annotations (the `+N -M` tally
+            // and `... (N lines truncated)`) keep the indent precisely so they
+            // are not mistaken for diff lines by the same greps.
+            result.push(line.to_string());
         } else if in_hunk {
             if line.starts_with('+') && !line.starts_with("+++") {
                 added += 1;
                 if hunk_shown < max_hunk_lines {
-                    result.push(format!("  {}", line));
+                    result.push(line.to_string());
                     hunk_shown += 1;
                 } else {
                     hunk_skipped += 1;
@@ -387,15 +396,15 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
             } else if line.starts_with('-') && !line.starts_with("---") {
                 removed += 1;
                 if hunk_shown < max_hunk_lines {
-                    result.push(format!("  {}", line));
+                    result.push(line.to_string());
                     hunk_shown += 1;
                 } else {
                     hunk_skipped += 1;
                 }
             } else if hunk_shown < max_hunk_lines && !line.starts_with("\\") {
-                // Context line
+                // Context line (git already prefixes it with a space).
                 if hunk_shown > 0 {
-                    result.push(format!("  {}", line));
+                    result.push(line.to_string());
                     hunk_shown += 1;
                 }
             }
@@ -2410,6 +2419,38 @@ mod tests {
         let result = compact_diff(diff, 100);
         assert!(result.contains("foo.rs"));
         assert!(result.contains("+"));
+    }
+
+    #[test]
+    fn test_compact_diff_hunk_lines_are_grep_anchorable() {
+        // Regression: hunk-body lines were indented two spaces, so
+        // `git diff | grep '^-'` matched nothing and an audit of "was anything
+        // removed?" got a confident, wrong "no". The content was never missing;
+        // the `^` anchor was what broke. Added lines were equally affected, so
+        // this is not a deletions-only defect.
+        let diff = "diff --git a/f.txt b/f.txt\n\
+                    --- a/f.txt\n\
+                    +++ b/f.txt\n\
+                    @@ -1,5 +1,4 @@\n\
+                    \x20keep1\n\
+                    -DELETED_A\n\
+                    \x20keep2\n\
+                    -DELETED_B\n\
+                    +ADDED\n";
+        let result = compact_diff(diff, 100);
+
+        let removed: Vec<&str> = result.lines().filter(|l| l.starts_with('-')).collect();
+        let added: Vec<&str> = result.lines().filter(|l| l.starts_with('+')).collect();
+
+        assert_eq!(removed, vec!["-DELETED_A", "-DELETED_B"], "`^-` must anchor");
+        assert_eq!(added, vec!["+ADDED"], "`^+` must anchor");
+
+        // rtk's own tally stays indented so these same greps never count it as
+        // a diff line. Without this, `^+` would pick up the "+1 -2" summary.
+        assert!(result.contains("  +1 -2"), "tally must stay indented");
+
+        // Context lines keep git's leading space, so they are not `^-`/`^+`.
+        assert!(result.lines().any(|l| l == " keep2"));
     }
 
     #[test]
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -338,12 +338,64 @@ fn is_blob_show_arg(arg: &str) -> bool {
 /// Path named by a diff section header.
 ///
 /// `diff --git a/p b/p` carries the path twice; `diff --cc p` and
-/// `diff --combined p` carry it once, at the end.
+/// `diff --combined p` carry it once, as the whole remainder of the line. git
+/// does not quote spaces in these headers, so the remainder is taken as-is.
 fn diff_header_path(line: &str) -> String {
     match line.split(" b/").nth(1) {
         Some(path) => path.to_string(),
-        None => line.rsplit(' ').next().unwrap_or("unknown").to_string(),
+        None => line.splitn(3, ' ').nth(2).unwrap_or("unknown").to_string(),
     }
+}
+
+/// Line budget a hunk header declares, and how wide its body prefix is.
+struct HunkHeader {
+    /// Lines the hunk spans in the first parent.
+    old: usize,
+    /// Lines the hunk spans in the result file.
+    new: usize,
+    /// Marker columns: 1 for `@@`, and one per parent for a combined `@@@`.
+    prefix_width: usize,
+}
+
+/// Parse `@@ -a,b +c,d @@` and the combined `@@@ -a,b -c,d +e,f @@@`.
+///
+/// The counts bound the hunk body, which is what lets the body end where the
+/// hunk ends rather than running on until the next header. Anything after it —
+/// an mbox envelope, a `--` signature, trailing prose — is then outside every
+/// hunk and cannot be read as diff content. A count is 1 when the header omits
+/// it (`@@ -1 +1 @@`).
+fn parse_hunk_header(line: &str) -> Option<HunkHeader> {
+    let at_run = line.len() - line.trim_start_matches('@').len();
+    if at_run < 2 {
+        return None;
+    }
+    let body = line[at_run..].split('@').next()?;
+
+    let mut old = None;
+    let mut new = None;
+    for group in body.split_whitespace() {
+        let Some(rest) = group.strip_prefix(['-', '+']) else {
+            continue;
+        };
+        let count = match rest.split_once(',') {
+            Some((_, c)) => c.parse::<usize>().ok()?,
+            None => 1,
+        };
+        if group.starts_with('-') {
+            // A combined header lists one range per parent; the first is the
+            // one whose line numbers the `-` column tracks.
+            old.get_or_insert(count);
+        } else {
+            new = Some(count);
+        }
+    }
+
+    Some(HunkHeader {
+        old: old.unwrap_or(0),
+        new: new.unwrap_or(0),
+        // `@@` has one marker column, `@@@` two, and so on for more parents.
+        prefix_width: at_run - 1,
+    })
 }
 
 /// Render the note for change lines dropped past `max_hunk_lines`, split by
@@ -365,11 +417,11 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
     let mut current_file = String::new();
     let mut added = 0;
     let mut removed = 0;
-    let mut in_hunk = false;
+    let mut hunk: Option<HunkHeader> = None;
     let mut hunk_shown = 0;
     let mut skipped_add = 0usize;
     let mut skipped_del = 0usize;
-    let mut leading_context = 0usize;
+    let mut leading_context: Vec<String> = Vec::new();
     let mut leading_context_total = 0usize;
     let max_hunk_lines = 100;
     // Context before a hunk's first change, up to three lines per hunk and
@@ -382,12 +434,10 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
     let mut was_truncated = false;
 
     for line in diff.lines() {
-        // `diff --cc` / `diff --combined` head the sections git emits for
-        // unmerged paths. Matching only `diff --git` left `in_hunk` set from
-        // the previous section, so the `---` / `+++` headers of every combined
-        // section after the first were counted and printed as hunk content.
+        // Every diff section header (`--git`, `--cc`, `--combined`) opens a new
+        // file and closes any open hunk, so the `---` / `+++` headers that
+        // follow it are never read as hunk content.
         if line.starts_with("diff --") {
-            // Flush hunk truncation before starting a new file
             if let Some(note) = hunk_truncation_note(skipped_del, skipped_add) {
                 result.push(note);
                 was_truncated = true;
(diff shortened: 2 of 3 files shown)


src/cmds/git/git.rs
```diff
@@ -374,28 +374,37 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
...
+            // unified shape. An earlier revision indented them two spaces to
+            // nest them under the filename, which silently broke every
+            // `git diff | grep '^-'` audit: the content was all there, but the
+            // `^` anchor no longer matched, so "was anything removed?" answered
+            // a confident, wrong "no". rtk's own annotations (the `+N -M` tally
+            // and `... (N lines truncated)`) keep the indent precisely so they
+            // are not mistaken for diff lines by the same greps.
+            result.push(line.to_string());
         } else if in_hunk {
             if line.starts_with('+') && !line.starts_with("+++") {
```
KuSh: These two guards (`!line.starts_with("+++")` here and `!line.starts_with("---")` at line 396) are wrong inside the `in_hunk` branch, and they silently delete content.

File headers `--- a/f` and `+++ b/f` only ever appear *before* the first `@@` — `in_hunk` is reset to `false` on every `diff --git`. So inside a hunk, a line starting `+++` or `---` is always **content**, never a header.

The consequence: any added line whose text starts with `++` (`++i;` in C, C++, Java, PHP) or removed line whose text starts with `--` (SQL comments, Markdown/YAML `---` rules, mail signatures) is dropped from the output *and* never counted in `added`/`removed`.

Verified end-to-end on your branch — a diff that deletes `-- sql comment` and adds `++i;` renders as a bare `@@ -1,4 +1,4 @@` with **no body and no tally at all**, and `grep -c '^-'` returns 0 while the file really did lose a line. That is precisely the "confident, wrong no" this PR sets out to kill, arriving by a different route.

Fix is two lines: drop both `starts_with` guards in the `in_hunk` branch, or gate them on `!in_hunk`.

For context, this is pre-existing — you are not introducing it — but it lands squarely inside the contract this PR establishes, and it's the same class kylehgc documented for the sibling `condense_unified_diff` in #3769. Worth fixing here while the lines are already open.

src/cmds/git/git.rs
```diff
@@ -374,28 +374,37 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
...
-                    result.push(format!("  {}", line));
+                    result.push(line.to_string());
                     hunk_shown += 1;
                 } else {
                     hunk_skipped += 1;
                 }
             } else if hunk_shown < max_hunk_lines && !line.starts_with("\\") {
-                // Context line
+                // Context line (git already prefixes it with a space).
                 if hunk_shown > 0 {
```
KuSh: `if hunk_shown > 0` means context lines appearing *before* the first `+`/`-` of a hunk are never emitted, which contradicts the PR body's "nothing is dropped".

Verified: for a hunk ` a / -b / +B / c / -d / e / +f`, rtk omits ` a` entirely.

It compounds with the `+++`/`---` issue above: when every `+`/`-` line in a hunk gets swallowed by those guards, `hunk_shown` stays 0 and the *entire* hunk body, context included, disappears.

The new test at line 2237 asserts on `keep2` and never on `keep1`, so as written it encodes the loss rather than catching it — worth adding `keep1` to that assertion whichever way you resolve this.

Heads-up: #1855 is an open PR that removes this guard outright. That turns out to have its own problem (leading context then eats the 100-line hunk budget and can crowd the actual change out entirely), so please don't just delete the condition — leading context needs its own budget. Happy for that to stay #1855's job if you'd rather keep this PR to the de-indent; in that case just fix the test assertion.

src/cmds/git/git.rs
```diff
@@ -374,28 +374,37 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
             hunk_shown = 0;
             // Preserve the full unified diff hunk header, including trailing
             // function / symbol context after the second @@ marker.
-            result.push(format!("  {}", line));
+            //
+            // Hunk-body lines below are emitted at column 0, in git's own
+            // unified shape. An earlier revision indented them two spaces to
```
KuSh: This comment is written as history — "An earlier revision indented them two spaces…", "answered a confident, wrong 'no'" — which the repo's comment convention rules out: comments should state what the code does and why, not narrate how it got here. The same narrative is then repeated almost verbatim in the test at line 2239.

It's also misplaced: it sits in the `else if line.starts_with("@@")` branch but describes the `else if in_hunk` branch below, so someone editing the hunk-body code won't see it.

Suggest replacing both with a single line on the `in_hunk` branch stating the invariant, e.g.:

```rust
// Hunk bodies emit at column 0; rtk's annotations stay indented so `^+`/`^-` never match them.
```

The reasoning is valuable — it just belongs in the commit message and the PR body, where you've already written it well.

KuSh (changes requested):
Requesting changes. The de-indent itself is correct and verified — I built the branch and confirmed `rtk git diff | grep -c '^-'` returns the true deletion count where it previously returned 0, that rtk's own `+N -M` tally and truncation notes correctly stay indented so they aren't miscounted, and that no caller re-parses this output (`pipe_cmd.rs:54`, `gh_cmd.rs:952`, `glab_cmd.rs:490` all just print it). Token cost goes down slightly, so `never_worse` is unaffected. Full suite green.

This is the version of the column-0 fix being taken forward — the competing #3646 is closed in its favour.

What needs another pass is the *contract* the PR body sells. "`grep -c '^-'` gives you the exact deletion count" is the promise, and there are three holes in it, two of them on lines this PR edits. Details inline; the third is here because it isn't tied to a single line:

**Past `max_hunk_lines` (100) the count silently stops being exact.** Verified: a 300-deletion hunk yields `grep -c '^-'` = 100 while the tally correctly says `+300 -300`. The `  ... (500 lines truncated)` note is deliberately indented — correctly, so the same grep doesn't count it — which means the anchored audit you're recommending has no way to see that it was truncated at all, and the note merges additions and deletions into one number. Either a per-sign note (`... (N- deletions, N+ additions truncated)`) or an explicit caveat in the PR body would close it honestly.

None of this makes the PR worse than the status quo — it's strictly better. It's just that the moment an audit is documented as trustworthy, people write CI gates against it.

src/cmds/git/git.rs
```diff
@@ -363,40 +383,53 @@ pub(crate) fn compact_diff(diff: &str, max_lines: usize) -> String {
...
-            result.push(format!("  {}", line));
+            result.push(line.to_string());
         } else if in_hunk {
-            if line.starts_with('+') && !line.starts_with("+++") {
+            // Hunk bodies emit at column 0 in git's own unified shape, so
+            // `^+` / `^-` anchor. rtk's own annotations stay indented so those
+            // same anchors never match them. Inside a hunk every `+`/`-` line
+            // is content: the `---` / `+++` file headers only ever appear
+            // before the first `@@`, where `in_hunk` is still false.
+            if line.starts_with('+') {
```
KuSh: `diff --cc` breaks the invariant this comment asserts.

The claim is that `---` / `+++` only ever appear before the first `@@`, where `in_hunk` is false. That holds for `diff --git` sections. It does not hold for **combined diffs**, which git emits for unmerged paths during a merge / rebase / cherry-pick / stash pop:

```
diff --cc a.txt
index ba2906d,e45c9c2..0000000
--- a/a.txt
+++ b/a.txt
@@@ -1,1 -1,1 +1,5 @@@
++<<<<<<< HEAD
 +main
...
diff --cc z.txt
index ba2906d,e45c9c2..0000000
--- a/z.txt
+++ b/z.txt
@@@ -1,1 -1,1 +1,5 @@@
```

`diff --cc z.txt` does not match the `line.starts_with("diff --git")` branch, so `in_hunk` is never reset between combined sections — the `@@@ ... @@@` header of the *previous* file left it `true`. The `--- a/z.txt` / `+++ b/z.txt` headers of the second and every later file therefore reach this branch with `in_hunk == true`.

Reproduced against a build of this branch, on a repo with two conflicted files and zero actual deletions:

```
$ rtk git diff | grep -c "^-"   ->  1
$ rtk git diff | grep -c "^+"   ->  9
```

The output carries a literal `--- a/z.txt` and `+++ b/z.txt` at column 0. On master these fell through to the context branch and were emitted as `  --- a/z.txt`, so `^-` did not match. They also increment `removed` / `added` now, inflating the `+N -M` tally by one each.

A `!current_file.is_empty()` style guard would not help here — `current_file` is only assigned in the `diff --git` branch too. The state reset needs to fire on any `diff --` prefix (`--git`, `--cc`, `--combined`), e.g.:

```rust
if line.starts_with("diff --") {
```

with the filename extraction falling back to the trailing path for the `diff --cc <path>` shape. With `in_hunk` reset there, the `---` / `+++` headers land before the first `@@` again and this comment's invariant genuinely holds. Worth a test with a two-file `diff --cc` fixture asserting `^-` matches nothing.

src/cmds/git/git.rs
```diff
@@ -335,24 +335,44 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
     let mut in_hunk = false;
     let mut hunk_shown = 0;
-    let mut hunk_skipped = 0usize;
+    let mut skipped_add = 0usize;
+    let mut skipped_del = 0usize;
+    let mut leading_context = 0usize;
     let max_hunk_lines = 100;
+    // Context before a hunk's first change gets its own budget so it cannot
+    // crowd the change itself out of `max_hunk_lines`.
+    let max_leading_context = 3;
```
KuSh: This justification does not describe a problem that existed: before this PR, leading context was never emitted and never incremented `hunk_shown`, so it could not crowd anything out of `max_hunk_lines`. The budget bounds a behaviour this PR itself introduces.

The cost is real, but it lands on the *global* `max_lines` budget, not `max_hunk_lines`: every leading-context `result.push` counts against `result.len() >= max_lines` (500 default, 200 for the pipe wrapper, 100 for `git stash show -p`).

Measured on a 5-commit diff of this repo (30 hunks, `max_lines = 500`): **367 change lines shown before this PR vs 347 after** — 20 fewer real `+` / `-` lines, with the output still ending in `... (more changes truncated)`. On a 20-hunk `git stash show -p` (budget 100), up to 60% of the budget can go to leading context.

I asked for the leading-context drop to be fixed, so I am not asking you to revert it blindly — but as written it trades one silent omission for another, and this comment does not surface that trade. Either exempt leading context from the `max_lines` count so it cannot displace change lines, or drop it here (correcting the PR body's "nothing is dropped" claim) and land it separately with a savings measurement against the repo's >=60% floor. Rewriting the comment to state the actual trade-off is needed either way.

KuSh (changes requested):
Thanks for the rework — the core fix is right, and I verified it: on ordinary `git diff` output `^-` / `^+` anchor again, and dropping the `+++` / `---` guards also fixes the silent content loss I flagged earlier (a deleted line starting with `--`, an added `++i;`). `hunk_truncation_note` splitting by sign is a good call, and the flush points all match the previous behaviour. The new tests cover the right things, `keep1` included.

One blocking issue, one open question, one docs item.

**Blocking — combined diffs (`diff --cc`) still emit false `^-` / `^+` lines.** Details inline. This one is introduced by the PR, not pre-existing: on master those header lines came out indented, so `^-` did not match them.

**Open question — the leading-context budget.** Details inline and on the existing thread; it addresses what I asked for, but it now competes with change lines for the global `max_lines` budget.

**Docs — `docs/usage/FEATURES.md:333-341`.** The `rtk git diff` *Avant / Apres* block shows change lines indented with `+ ` / `- ` / `~ ` under a `src/main.rs (+5/-2)` header. That illustration was already drifting before this PR, but this PR changes the emitted shape again (hunk bodies and `@@` headers now at column 0, `  +5 -2` as a trailing indented line). Since the whole point of this PR is that the emitted shape is greppable, please update that block to match what `compact_diff` now produces — otherwise the one page a user would read to learn the format still teaches the wrong anchors.

src/cmds/git/git.rs
KuSh: Leading context is emitted now and the test asserts `keep1`, so both halves of this are addressed — thanks.

Leaving the thread open because the fix has a cost I had not anticipated: the leading-context lines count against the global `max_lines` budget, so on a large diff they displace real change lines (measured 367 -> 347 change lines on a 30-hunk diff at `max_lines = 500`). Continued at the `max_leading_context` declaration in my new review.

src/cmds/git/git.rs
KuSh: Fixed — the `+++` / `---` guards are gone and `test_compact_diff_keeps_content_starting_with_plus_or_minus` covers both the body and the tally. Resolving.

src/cmds/git/git.rs
KuSh: Fixed — the comment is rewritten as an invariant, moved onto the `in_hunk` branch, and the history narration is gone from the test too. Resolving.

iliaal:
Thanks — all three inline points and the truncation gap are fixed in 4efc0d9.

**`+++` / `---` guards.** You're right that they can only match content inside a hunk, and you're right about the consequence. Both guards are gone from the `in_hunk` branch. `test_compact_diff_keeps_content_starting_with_plus_or_minus` covers your exact case: a diff removing `-- sql comment` and adding `++i;` now renders both lines and reports `+1 -1`.

**Leading context.** I didn't want to ship a PR whose body says "nothing is dropped" while `hunk_shown > 0` silences the context before a hunk's first change, so I fixed it here rather than deferring — but not by deleting the condition, for the reason you gave. Leading context gets its own budget of three lines, matching git's default `-U3`, and it doesn't touch `hunk_shown`, so a long run of context can't crowd out the change it frames. `test_compact_diff_leading_context_has_its_own_budget` puts 20 context lines ahead of 100 deletions and asserts all 100 deletions still render. The anchorable test now asserts on `keep1` as well. If #1855 wants to generalize the budget, this shouldn't get in its way.

**The comment.** Correct on both counts — it narrated history and it sat on the `@@` branch while describing the branch below. Replaced with a single line stating the invariant, on the `in_hunk` branch. The narration in the test is gone too; the reasoning lives in the commit message and the PR body.

**Truncation past `max_hunk_lines`.** Took the per-sign note: `... (30 deletions, 30 additions truncated)`, or the single-sign form when only one side is cut. It stays indented, so the anchored greps still skip it. That's the honest version of the contract — the note can't be seen by `grep '^-'` by design, so it has to tell a reader who does look how many deletions they didn't see, and one merged count couldn't. `test_compact_diff_truncation_note_splits_by_sign` pins it.

`cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`, and `cargo test --all` are clean on the branch, apart from two `hooks::rewrite_cmd::tests::unattestable_passthrough` failures I reproduced with my changes stashed — they predate this PR and it doesn't touch `rewrite_cmd.rs`.

I left `condense_unified_diff` alone: same class of bug, but it belongs with #3769 rather than in a PR about `compact_diff`'s indent.

iliaal:
Thanks — the `diff --cc` catch is right, and it's mine: on master those headers came out indented, so this branch is what let them anchor. Fixed in 201b550.

**Combined diffs.** Reproduced your exact numbers on a two-file conflict with zero deletions — `^-` = 1, `^+` = 9, with a literal `--- a/z.txt` and `+++ b/z.txt` at column 0. The reset now fires on `line.starts_with("diff --")`, so `--git`, `--cc`, and `--combined` all clear `in_hunk`, and the path extraction falls back to the trailing token when there's no ` b/` to split on. Same repo now gives `^-` = 0, `^+` = 8, the 8 being the real `++`-prefixed conflict lines. `test_compact_diff_combined_diff_headers_are_not_hunk_content` uses a two-section `diff --cc` fixture and asserts no `^-` match and no `+++ b/` in the body.

**Leading context.** You're right that the comment justified a problem that never existed — before this branch leading context was never emitted, so it couldn't crowd anything out of `max_hunk_lines`. The comment now states the real trade-off.

I took the exemption rather than dropping the feature, with one change to your suggestion. A pure exemption has a hole: a diff of many small hunks spends three exempt lines on each, so the output can overrun `max_lines` several times over. So leading context is exempt from `max_lines` *and* capped at `max_lines / 10` across the whole diff. Change lines are never displaced; the overrun is bounded at 10%.

Measured on a five-commit diff of this repo, `max_lines = 500`:

| build | change lines | bytes | reduction vs raw |
|---|---|---|---|
| master (`1c5f164`) | 0 | 34,693 | 71.8% |
| `4efc0d9` (context counted) | 144 | 31,858 | 74.1% |
| `201b550` (exempt + capped) | 201 | 34,920 | 71.7% |

So the exemption buys back 57 change lines for 0.1 points of reduction against master, which showed zero anchorable change lines at almost exactly the same byte count. Well clear of the 60% floor. Two tests pin it: one builds the same 30-hunk diff with and without leading context and asserts the change-line count is identical, the other builds 200 single-change hunks and asserts the context stays within `max_lines / 10`.

**`docs/usage/FEATURES.md`.** Rewritten with real captured output rather than the illustration — the old block showed change lines indented under a `src/main.rs (+5/-2)` header that no version of rtk has emitted. It now also states both limits of the anchored audit: past 100 change lines per hunk `grep -c "^-"` plateaus at what's shown, with the per-sign note carrying the rest where the same grep can't see it, and the output is not an applyable patch because rtk drops the `diff --git` / `---` / `+++` headers. `rtk proxy git diff` for that.

`cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test --all` at 2506 passed, with the same two pre-existing `hooks::rewrite_cmd::tests::unattestable_passthrough` failures I reproduced with the branch's changes stashed.

src/cmds/git/git.rs
```diff
@@ -335,82 +335,138 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
-            result.push(format!("  {}", line));
+            result.push(line.to_string());
         } else if in_hunk {
-            if line.starts_with('+') && !line.starts_with("+++") {
+            // Hunk bodies emit at column 0 in git's own unified shape, so
+            // `^+` / `^-` anchor. rtk's own annotations stay indented so those
+            // same anchors never match them. Inside a hunk every `+`/`-` line
+            // is content: the `---` / `+++` file headers only ever appear
+            // before the first `@@`, where `in_hunk` is still false.
+            if line.starts_with('+') {
```
KuSh: **Blocking.** The invariant does not hold for every producer that reaches `compact_diff`, and the failure lands on this PR's headline use case.

`gh pr diff --patch` is routed through this function: `has_non_diff_format_flag` (`src/cmds/git/gh_cmd.rs:919`) lists only `--name-only` / `--name-status` / `--stat` / `--numstat` / `--shortstat`, so `--patch` falls through to the filtered path. GitHub's `.patch` is an mbox of `format-patch` output — every patch carries a bare `---` between the commit message and the diffstat, and ends with a `-- ` signature line, both at column 0 while `in_hunk` is still true from the previous patch's last hunk.

Reproduced by calling `compact_diff` directly with a two-commit mbox containing **two** real deletions:

```
^- = 5   ["-old1", "-- ", "---", "-old2", "-- "]
first file tally:  +1 -3
```

The mbox envelope leaks into the body too — `From def ...` and `Subject: [PATCH 2/2] two` are emitted verbatim at column 0 inside the first file's hunk.

Master counted `-- ` in the tally as well, so that part is not new, but the anchoring is: on master none of these lines matched `^-`. As it stands `rtk gh pr diff --patch | grep -c "^-"` reports 5 deletions where there are 2. The same shape reaches `glab mr diff` (`glab_cmd.rs:492`) and any patch text piped in.

Two ways out, and they compose:

- Cheap and immediate: add `--patch` to `has_non_diff_format_flag` so mbox passes through unfiltered. An mbox is not a unified diff, and `compact_diff` has no business parsing one.
- Principled: end the hunk when it is over. `@@ -a,b +c,d @@` declares its own line counts, so `in_hunk` can be cleared once `b` / `d` lines have been consumed. That also stops *any* trailing non-diff text being read as hunk content, rather than special-casing the shapes we happen to know about.

Note that a heuristic guard on `---` / `-- ` cannot work here: a deleted line whose content is `- ` renders as `-- `, indistinguishable from the signature. The hunk-length approach avoids that ambiguity entirely.

Worth a test with a two-patch mbox asserting `^-` counts only the real deletions.

src/cmds/git/git.rs
```diff
@@ -335,82 +335,138 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
+/// Path named by a diff section header.
+///
+/// `diff --git a/p b/p` carries the path twice; `diff --cc p` and
+/// `diff --combined p` carry it once, at the end.
+fn diff_header_path(line: &str) -> String {
+    match line.split(" b/").nth(1) {
+        Some(path) => path.to_string(),
+        None => line.rsplit(' ').next().unwrap_or("unknown").to_string(),
+    }
+}
```
KuSh, line 497: The fallback truncates any path containing a space.

git does not quote spaces in these headers. On a repo with a conflicted `my file.txt`:

```
diff --cc my file.txt
--- a/my file.txt
+++ b/my file.txt
```

`rsplit(' ').next()` yields `file.txt`, so the section header rtk prints for that file names the wrong path. All the shapes:

```
'diff --git a/my file.txt b/my file.txt'  ->  'my file.txt'   ok
'diff --cc my file.txt'                   ->  'file.txt'      wrong
'diff --combined my file.txt'             ->  'file.txt'      wrong
```

`line.splitn(3, ' ').nth(2)` keeps the whole remainder and covers both `diff --cc <path>` and `diff --combined <path>`.

The `" b/"` split above is fine for ordinary spaces, for what it's worth — it only misparses a path that literally contains `" b/"`, which is pre-existing and not worth chasing here.

Also: `unwrap_or("unknown")` is unreachable, since `rsplit` on a non-empty `&str` always yields at least one item. Not a problem, just dead.

src/cmds/git/git.rs
```diff
@@ -335,82 +335,138 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
+                        result.push(line.to_string());
+                        hunk_shown += 1;
+                    }
+                } else if leading_context < max_leading_context
+                    && leading_context_total < leading_context_cap
+                {
+                    result.push(line.to_string());
+                    leading_context += 1;
+                    leading_context_total += 1;
                 }
```
KuSh: Keeping the *first* lines of the hunk's leading context fabricates adjacency.

With `-U10`, `-W` or `--function-context` — all common — a hunk that opens with ten context lines renders as:

```
@@ -1,14 +1,14 @@
 ctx1
 ctx2
 ctx3
-old
+new
```

A reader concludes `ctx3` is the line immediately preceding the deletion. It is actually `ctx10`. The `@@` header still advertises 14 lines, nothing marks the gap, and `was_truncated` is not set — so the `[full diff: rtk git diff --no-compact]` recovery hint does not fire either. There is no signal at all that context was elided.

Keeping the *last* `max_leading_context` lines instead — a 3-slot ring buffer over the pre-change context, flushed when the first `+`/`-` arrives — makes the emitted body contiguous with the change at the same cost and the same line count.

The diff-wide `max_lines / 10` cap compounds this: two identical hunks render with 3 context lines early in the diff and 0 later, with nothing to explain the difference.

src/cmds/git/git.rs
```diff
@@ -2225,6 +2281,202 @@ mod tests {
...
+                    \x20+main\n";
+        let result = compact_diff(diff, 500);
+
+        let removed: Vec<&str> = result.lines().filter(|l| l.starts_with('-')).collect();
+        assert!(
+            removed.is_empty(),
+            "no deletions in this diff, got {:?} in:\n{}",
+            removed,
+            result
+        );
```
KuSh: This assertion locks in an under-count for combined diffs.

`diff --cc` uses two prefix columns: a line changed relative to only one parent carries its marker in column 2 (` +ours`, ` -theirs`). Those land in the context branch, so they are neither counted in `+N -M` nor reachable by `^+` / `^-`.

Verified against a build of this branch on a real two-file conflict: ` +main` — present in HEAD only — is emitted as context and excluded from the tally, which prints `  +4 -0`. A deletion relative to a single parent (` -x`) would be equally invisible.

The fixture here contains ` +main` and asserts `removed.is_empty()`, which reads as correct but only holds because the column-2 lines are invisible to the check. Since `diff --cc` is what `git diff` emits for every unmerged path during a merge, rebase, cherry-pick or stash pop, the anchored audit is wrong precisely on conflict diffs — the case this PR just added support for.

Either widen the prefix check to the first two columns when the hunk header is `@@@` (a combined diff of N parents uses N columns), or keep the current behaviour, assert it here deliberately, and document it — see my note on the *limites de l'audit ancre* paragraph.

src/cmds/git/git.rs
```diff
@@ -335,82 +335,138 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
+    let max_leading_context = 3;
+    let leading_context_cap = max_lines / 10;
     let mut was_truncated = false;
 
     for line in diff.lines() {
-        if line.starts_with("diff --git") {
+        // `diff --cc` / `diff --combined` head the sections git emits for
+        // unmerged paths. Matching only `diff --git` left `in_hunk` set from
+        // the previous section, so the `---` / `+++` headers of every combined
+        // section after the first were counted and printed as hunk content.
```
KuSh: This narrates the bug instead of stating the rule: "Matching only `diff --git` **left** `in_hunk` set from the previous section, so the `---` / `+++` headers ... **were** counted and printed as hunk content."

That is the same objection as the earlier thread on the `@@`-branch comment, which you fixed — the convention here is that a comment says what the code does and why, not how it got here. Someone reading this in six months needs the invariant, not the regression it replaced.

Something like: *"Every diff section header (`--git`, `--cc`, `--combined`) opens a new file and closes any open hunk, so the `---` / `+++` headers that follow are never read as hunk content."*

Same shape in the test comment at lines 2370-2373: "`in_hunk` **stayed** true from the previous section ... **were** printed at column 0 and counted".

docs/usage/FEATURES.md
```diff
@@ -330,17 +330,72 @@ rtk git diff [args...]    # Supporte --stat, --cached, --staged, etc.
...
+ fn main() {
+-    let x = 42;
+-    // old code
++    let config = Config::load()?;
++    config.validate()?;
+     println!("start");
+ }
+```
+
+`rtk git diff`, 15 lignes :
```
KuSh: `15 lignes` is wrong — this block is **24** lines: 3 stat lines, blank, `Changes:`, blank, 8 for `git.rs`, blank, 10 for `main.rs`. I reproduced the example and the content is byte-accurate, so it is only the count that is off.

With the raw block correctly labelled 21 lines, the section as written shows rtk emitting *more* lines than `git diff`. It does still save tokens — 440 bytes raw against 392 filtered, so `never_worse` keeps the filtered form and the block really is what gets printed — but ~11% is a thin illustration in a document that advertises 60-90%. An example with enough hunks for the per-hunk cap or the truncation note to appear would demonstrate the feature and make both counts worth printing.

Minor: the first line here is ` git.rs  | 2 +-` with a leading space. Real output has none, since the stat is `.trim()`ed before `\n\nChanges:\n` is appended.

docs/usage/FEATURES.md
```diff
@@ -330,17 +330,72 @@ rtk git diff [args...]    # Supporte --stat, --cached, --staged, etc.
...
+propres a rtk restent indentees de deux espaces pour que ces memes greps ne
+les comptent pas : le total par fichier (`  +2 -2`) et la note de troncature
+(`  ... (30 deletions, 20 additions truncated)`).
+
+Deux limites de l'audit ancre. Au-dela de 100 lignes de changement par hunk,
+`grep -c "^-"` plafonne a ce qui est affiche ; la note de troncature indique
+le reste, par signe, mais elle est indentee et donc invisible au meme grep.
+Et la sortie n'est pas un patch applicable : rtk supprime les en-tetes
+`diff --git` / `---` / `+++`. Utilisez `rtk proxy git diff` pour un patch
+fidele.
```
KuSh: Worth a third limit in this paragraph: the anchored audit under-counts combined diffs.

`diff --cc` puts the marker in one of two prefix columns and rtk reads only column 1, so a line changed relative to a single parent (` +ours`, ` -theirs`) is invisible to both `grep "^-"` and the `+N -M` tally — details on the combined-diff test.

`git diff` emits `diff --cc` for every unmerged path during a merge, rebase, cherry-pick or stash pop, so this is a case a user auditing a conflict hits directly. It belongs right next to the 100-lines-per-hunk caveat.

KuSh (changes requested):
Both items from my previous review are fixed, and I checked them against a build of `201b550` rather than by reading the diff:

- **`diff --cc` header leak** — my earlier repro now gives `^-` = 0 (was 1) and `^+` = 8 (was 9), with the per-file tally back to `+4 -0`.
- **Leading context displacing changes** — 354 change lines on a real 5-commit diff of this repo, identical to master's 354, where the previous head gave 347. The exemption plus the diff-wide cap does exactly what it says.
- **FEATURES.md** — I reproduced the example verbatim; the block content is byte-accurate.

`cargo test --all` on this head: 2508 passed, 0 failed. I have resolved those threads.

A second pass turned up four things, one of which I would block on:

1. **Blocking** — `compact_diff` still misreads patch/mbox input, and it lands on this PR's headline use case (`gh pr diff --patch | grep -c "^-"`). Inline.
2. Leading context keeps the *first* lines of a hunk, so the emitted body is not contiguous with the change and nothing says so. Inline.
3. Combined diffs under-count by design — on exactly the format this PR just added support for. Inline, plus a docs note.
4. `diff_header_path` truncates any path containing a space. Inline.

Plus two comment-convention nits and a docs count, all inline.

One thing not worth an inline comment: exempting leading context from `max_lines` makes it a soft cap — output can now exceed it by up to 10% (`git stash show -p` passes 100, so up to 110). That is deliberate and the code comment explains it well; it just deserves a line in the PR description, since the two call sites chose those numbers as hard budgets.

Unrelated heads-up: `src/cmds/git/git.rs` is being renamed to `src/cmds/git/git_cmd.rs` on another branch. If that lands first this PR will need a rebase.

iliaal:
All four addressed in dc44930, and the mbox one is a good catch — it is exactly the case this PR was written for.

**Patch and mbox input.** Took both of your routes, since they close different halves. `--patch` joins `has_non_diff_format_flag`, so `gh pr diff --patch` passes through whole: someone asking for a patch wants one that applies, and `compact_diff` has no business parsing an mbox. And the hunk now ends where it says it ends — `@@ -a,b +c,d @@` declares its line counts, so the body stops when they are consumed and everything after it is outside every hunk. Your two-patch mbox with two real deletions now gives `grep -c "^-"` = 2, and `Subject:` / `From` no longer appear in any body. You were right that a heuristic could not work here: a deleted line whose content is `- ` renders as `-- `, so only the length approach distinguishes them. `test_compact_diff_mbox_signature_is_not_a_deletion` covers it, and `test_parse_hunk_header_counts` pins the parsing, including the `@@ -1 +1 @@` omitted-count form and the combined `@@@`.

The general benefit is bigger than the mbox: any trailing text after a diff, and any truncated input, now stops being read as hunk content.

**Leading context adjacency.** Fixed with the ring buffer you described — it keeps the last three lines, so with `-U10` the body shows `ctx8 ctx9 ctx10` rather than `ctx1 ctx2 ctx3`. Same line count, and the adjacency it implies is real. I also moved the diff-wide budget so it is charged when the buffer drains rather than when a line enters it, so a line the ring evicted costs nothing. `test_compact_diff_leading_context_is_adjacent_to_the_change` asserts the exact three.

**Combined diff under-count.** Widened rather than documented-as-is, at least for the tally: the marker check now reads the first `prefix_width` columns, which `parse_hunk_header` derives from the `@` run, so ` +ours` and ` -theirs` are counted. Verified against `git diff --numstat` on three complete commit ranges — the tally is now exact on all three, where master under-counted one by `+2 -1`.

The grep half I cannot fix without rewriting the line, so it is documented instead: `docs/usage/FEATURES.md` now lists it as a third limit next to the 100-lines-per-hunk cap, naming merge, rebase, cherry-pick and stash pop. The test asserts the tally at `+5 -0` and the anchored count at 4 per file deliberately, with a comment saying which is which.

**`diff_header_path` and spaces.** Fixed with `splitn(3, ' ').nth(2)`, and the dead `unwrap_or` is gone from the reasoning — `test_diff_header_path_keeps_spaces` covers all three header shapes with `my file.txt`.

**Comment convention.** Both rewritten to state the invariant. Fair repeat: I fixed this on the `@@` branch and then wrote the same shape two comments later.

**FEATURES.md.** The count is 24, and the stat's first line has no leading space. On the thin-illustration point you are right, and the honest version is that a two-file diff saves 11% because git's own `-U3` output is already compact — the compression comes from the per-file headers and the per-hunk cap. So the block now states that 11% directly and adds a measured larger case: 4 files with a 300-line hunk, 335 lines and 13,262 bytes down to 140 and 4,880, 63% reduction, with the truncation note and the recovery hint both visible.

**Soft cap.** Added to the PR description, with the `git stash show -p` call site named: 100 can now yield up to 110.

Noted on the `git.rs` to `git_cmd.rs` rename; happy to rebase whenever it lands.

`cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test --all` at 2511 passed, with the same two pre-existing `hooks::rewrite_cmd::tests::unattestable_passthrough` failures that reproduce with this branch's changes stashed.

src/cmds/git/git.rs
```diff
@@ -335,82 +335,223 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
+                // occupies no line in either file.
+                continue;
+            }
+
+            let width = header.prefix_width.min(line.len());
+            let markers = &line[..width];
+            // A combined diff carries one marker column per parent, so a line
+            // changed against only one of them has its marker in column 2.
+            let is_add = markers.contains('+');
+            let is_del = markers.contains('-');
```
KuSh: **Crash.** This slices by *bytes*: `width` is `prefix_width.min(line.len())`, so on a unified hunk it is 1, and any body line beginning with a multi-byte character splits mid-character.

`--word-diff` / `--color-words` / `--word-diff-regex` emit body lines with no marker column at all, so their content lands here directly. On a file containing `école ancienne ligne`:

```
$ rtk git diff --word-diff
thread 'main' panicked at src/cmds/git/git.rs:482:32:
end byte index 1 is not a char boundary; it is inside 'é' (bytes 0..2)
$ echo $?
101          # stdout empty
```

The installed pre-PR rtk handles the same command fine and exits 0, so this is a regression from this commit — `starts_with` was byte-safe.

`run_diff` (`git.rs:198`), `run_show` (`git.rs:315`) and `git stash show -p` (`git.rs:1907`) call `compact_diff` with no `catch_unwind`, so the panic is fatal on those paths. `rtk pipe --filter git-diff` (`src/cmds/system/pipe_cmd.rs:54`) and `src/core/stream.rs` do catch it, but the user still gets a panic dump on stderr before the raw fallback:

```
printf 'diff --git a/f b/f\n@@ -1,3 +1,3 @@\n-old\n+new\nété\n' | rtk pipe --filter git-diff
```

The markers are ASCII by construction, so comparing bytes avoids the boundary question entirely:

```rust
let markers = &line.as_bytes()[..header.prefix_width.min(line.len())];
let is_add = markers.contains(&b'+');
let is_del = markers.contains(&b'-');
```

Worth a test with a non-ASCII unprefixed body line — that is exactly the case no existing fixture covers.

src/cmds/git/git.rs
```diff
@@ -335,82 +335,223 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
+            // Present in the result file unless it is a pure deletion, and in
+            // the first parent unless it is a pure addition.
+            header.new = header.new.saturating_sub(if is_del && !is_add { 0 } else { 1 });
+            header.old = header.old.saturating_sub(if is_add && !is_del { 0 } else { 1 });
```
KuSh: The `old` accounting is inverted for combined diffs, which breaks the termination guarantee this commit is built on.

In a combined diff `' '` means opposite things depending on the line. On a change line it means *present and unchanged* in that parent; on a removal line it means *absent* from that parent. `is_add` / `is_del` collapse all columns, so both readings are lost:

- ` +main` — space in column 1, so the line **is** in parent 1 — hits `is_add && !is_del` and does not decrement `old`.
- ` -res` — space in column 1 on a removal line, so the line is **not** in parent 1 — decrements `old` anyway.

Both directions bite:

**`old` never converges on real conflict output.** On an actual `git diff` of a conflicted merge (`@@@ -1,1 -1,1 +1,5 @@@` over `++<<<<<<<`, ` +main`, `++=======`, `+ side`, `++>>>>>>>`) the hunk ends with `old == 1`, so it never closes by count. Everything this commit adds to keep mbox envelopes, `-- ` signatures and trailing prose out of the hunk body silently does not apply to combined sections — and combined sections are what `git diff` emits for every unmerged path.

**The reverse case closes early and drops the body.** Reproduced:

```rust
compact_diff("diff --cc f.txt\n@@@ -1,2 -1,4 +1,2 @@@\n  a\n  b\n -x\n -y\n", 100)
```

returns nothing but

```
\nf.txt\n@@@ -1,2 -1,4 +1,2 @@@
```

Both removals are gone, there is no `+N -M` tally, and no truncation note — nothing signals the loss. That is precisely the silent false negative this PR set out to eliminate.

The rule wants the column, not the aggregate: a line counts against `old` iff it is present in parent 1, i.e. column 1 is `'-'` on a removal line, or `' '` on a non-removal line. `new` is already right (`!is_del`).

src/cmds/git/git.rs
```diff
@@ -335,82 +335,223 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
+                // file does not have.
+                if leading_context.len() == max_leading_context {
+                    leading_context.remove(0);
                 }
+                leading_context.push(line.to_string());
+            }
+
+            if header.old == 0 && header.new == 0 {
+                hunk = None;
             }
```
KuSh, line 790: The ring buffer only drains on the first `+` / `-` line (line 508 or so), so when a hunk ends without one the buffered context is discarded. A context-only hunk renders as a bare header:

```rust
compact_diff("diff --git a/g.txt b/g.txt\n@@ -1,3 +1,3 @@\n ctx1\n ctx2\n ctx3\n", 100)
//  -> "\ng.txt\n@@ -1,3 +1,3 @@"
```

Master dropped those lines too, so this is not new against the base — but it is a regression against the previous head, which emitted them, and it is inconsistent with the buffering this commit introduces. It also compounds the early-close case in my comment above: the dropped hunk body leaves nothing behind at all.

Flushing whatever the buffer holds when the hunk closes (here at the `old == 0 && new == 0` reset, and at end of input) keeps it consistent, at the cost of at most `max_leading_context` lines already inside the diff-wide budget.

src/cmds/git/git.rs
```diff
@@ -335,82 +335,223 @@ fn is_blob_show_arg(arg: &str) -> bool {
...
 
+/// Path named by a diff section header.
+///
+/// `diff --git a/p b/p` carries the path twice; `diff --cc p` and
+/// `diff --combined p` carry it once, as the whole remainder of the line. git
+/// does not quote spaces in these headers, so the remainder is taken as-is.
+fn diff_header_path(line: &str) -> String {
+    match line.split(" b/").nth(1) {
+        Some(path) => path.to_string(),
+        None => line.splitn(3, ' ').nth(2).unwrap_or("unknown").to_string(),
```
KuSh: The claim in this doc comment is true for spaces but not for the case that actually reaches the fallback.

git quotes any path with non-ASCII bytes under the default `core.quotepath=true`, and a quoted header contains no ` b/`, so it falls through to `splitn(3, ' ').nth(2)` — which then returns the whole remainder, both paths included. On a file named `été.txt`:

```
raw:  diff --git "a/\303\251t\303\251.txt" "b/\303\251t\303\251.txt"
rtk:  "a/\303\251t\303\251.txt" "b/\303\251t\303\251.txt"     <- the section header
```

Master printed `unknown` here, which is not good either, but this reads as a broken parse rather than a missing name.

Handling the quoted form explicitly — strip a leading `"`, split on `" b/` as well as ` b/` — covers it, and the doc comment then wants to say that git quotes non-ASCII paths but not spaces, since that is the distinction the code has to make. The same fallback also mis-splits a `diff --cc` path that happens to contain ` b/`, though that one is pathological enough to leave alone.

KuSh (changes requested):
All seven comments from the last round are fixed, and I checked each against a build of `dc44930` rather than by reading:

- A **real `git format-patch --stdout`** mbox now gives `^-` = 2 for its 2 real deletions (was 5), with the `---` / `-- ` / `From:` / `Subject:` envelope gone entirely. Ending the hunk at its declared length is the right shape for this, and the `--patch` passthrough is a good belt-and-braces.
- `-U10` now emits `ctx8 ctx9 ctx10` — contiguous with the change.
- A conflicted `my file.txt` keeps its full name; the combined tally counts ` +main` (`+5 -0`, was `+4 -0`).
- The docs example re-measures correctly (21 lines / 440 bytes raw against 24 / 392 filtered), and the third limit is accurate.

Change-line parity holds against master (390 = 390 on a real 5-commit diff of this repo) and `cargo test --all` gives 2508 -> 2513 passing. I have resolved all of those threads.

The rework introduces four new problems, one of which is a crash:

1. **`&line[..width]` panics on non-ASCII.** `rtk git diff --word-diff` exits 101 with empty stdout on this branch and works on master.
2. **Combined-diff `old` accounting is wrong in both directions**, so declared-length termination — the mechanism this commit is built around — silently does not apply to `diff --cc`, and the reverse case drops change lines with no marker.
3. Buffered leading context is dropped when a hunk ends without a change line.
4. `diff_header_path` mishandles git's quoted (non-ASCII) paths.

1 and 2 are related but independent: fixing 2 does not fix 1, since `--word-diff` hunks are ordinary `@@` whose counts mismatch by construction. Leaving 2 unfixed means the hunk-length guard does not protect conflict diffs, which is where this PR started.

src/cmds/git/git.rs
iliaal: Fixed in `99c3a92`, and the rationale behind it did not survive `affcae1` — see below.

The markers are sliced as bytes now, so the boundary question is gone. Both your repros:

```
printf 'diff --git a/f b/f\n@@ -1,3 +1,3 @@\n-old\n+new\nété\n' | rtk pipe --filter git-diff
  before: exit 134, core dumped, empty stdout
  after:  exit 0, 6 lines

rtk git diff --word-diff   (unchanged line starting `été`)
  before: exit 134, empty stdout
  after:  exit 0
```

`test_compact_diff_non_ascii_body_line_without_a_marker_does_not_panic` covers it.

But you were right to point at word diff as the reachable path, and the byte fix only stopped the panic. It does not stop the misparse: a markerless row starting `+` counts as an addition and consumes the wrong budget, one starting `\` is dropped as a no-newline annotation, and a real `git show --word-diff=plain` row reading `diff --git [-a/src/main.rs...` opens a new file section. Claim 1's "ASCII by construction" was false on exactly the path the commit cited. `affcae1` passes these modes through instead, the way `--patch` already does for mbox. `rtk pipe --filter git-diff` has no flags to read and still misparses word-diff text piped into it; `compact_diff` reads a unified or combined diff, and that is now what its callers give it.

src/cmds/git/git.rs
iliaal: Fixed in `99c3a92`. The rule is per column, as you set out: a line spends one of parent `i + 1`'s lines when column `i` is `-` on a removal line, or a space on a line that is not one.

`HunkHeader` tracks a budget per parent and ends the hunk when every one of them and the result are spent. That was the missing half — your `-1,2 -1,4 +1,2` case closes on the first parent and the result while the second parent still has budget, which is why the removals vanished. Both directions:

```
real conflict + trailing `-- ` / signature / prose
  before: `-- ` emitted as a diff line, tally +5 -1
  after:  envelope gone, tally +5 -0, hunk closes by count

@@@ -1,2 -1,4 +1,2 @@@ over `  a  b -x -y`
  before: bare hunk header, no tally, no note
  after:  all four body lines, +0 -2
```

So declared-length termination applies to combined sections now, which is what the previous commit was built around.

`affcae1` adds the case where the header and the columns disagree. `@@@ -1 +1 @@@` left the second parent untracked, so ` -lost` drove the tracked budgets to zero and the hunk closed with the removal discarded; `@@@ -1 -1 -1 +0,0 @@@` declared a third parent with no column to spend it, so that budget never reached zero and the mbox signature became hunk body. Ranges are reconciled with columns: a missing one gets `usize::MAX`, keeping the hunk open to the next header rather than dropping content, and an extra one is dropped.

src/cmds/git/git.rs
iliaal: Fixed in `99c3a92`. `flush_leading_context` is called wherever a hunk closes — on `diff --`, on a new `@@`, when the budgets are spent, and at end of input — not only at the first change line.

```
compact_diff("diff --git a/g.txt b/g.txt\n@@ -1,3 +1,3 @@\n ctx1\n ctx2\n ctx3\n", 100)
  before: "\ng.txt\n@@ -1,3 +1,3 @@"
  after:  the three context lines under the header
```

The lines are still charged against the same diff-wide budget, so nothing escapes `leading_context_cap`.

src/cmds/git/git.rs
iliaal: Fixed in `99c3a92`, and `affcae1` covers two more shapes in the same function.

```
diff --git "a/\303\251t\303\251.txt" "b/\303\251t\303\251.txt"  →  \303\251t\303\251.txt
diff --cc "\303\251t\303\251.txt"                              →  \303\251t\303\251.txt
diff --git a/plain.txt "b/\303\251t.txt"                        →  \303\251t.txt
```

The doc comment names the distinction the code makes: git quotes a non-ASCII byte, a control character or a quote, and not a space.

You said the ` b/`-inside-a-path case was pathological enough to leave alone, and I fixed it anyway because the same change fixes `--no-prefix`, which is not. `diff --git a/x b/y b/x b/y` reported `y`, and `diff --git x x` reported both paths as the section header. Anything but a rename names the same path twice, so the halves are the same length and the separating space sits dead centre; splitting there handles all three and does not care what the prefixes are. A rename's halves differ in length and still fall through to the ` b/` split.

iliaal:
One thing I looked at and left: `max_lines` can overshoot by two entries at a file boundary, because the previous file's tally and the next file's header are pushed before the budget check. At `max_lines = 200` over 100 single-hunk files the output carries 201 non-context entries. It is byte-identical on master and on `dc44930`, so it is not this PR's, and it is a second source of overshoot alongside the context exemption the description already documents. Say the word if you want it in scope here rather than on its own.

Non-regression: ordinary unified diffs are byte-identical to `dc44930`, verified on a 26,561-line real diff of this repo and on a 390-line one, plus a 300-case generated corpus covering additions, deletions, replacements, empty rows, Unicode, leading `+`/`-`/`\`, multiple hunks and multiple files. The intentional divergence is the context-only hunk. `cargo fmt --all --check` and `cargo clippy --all-targets` clean, `cmds::git::git` 119/119.

KuSh (approved):
Approving. Everything raised across the four rounds is addressed, and I verified this head by building and running it rather than by reading the diff.

**What I checked on this head**

- `--word-diff`, `--color-words`, `--word-diff=porcelain` pass through untouched (exit 0); `--word-diff=none` compacts again — 25,307 bytes against 101,259 raw on a five-commit diff of this repo, so the saving that flag used to cost is back.
- A real `git format-patch --stdout` mbox gives `^-` = 2 for its two deletions, envelope and `-- ` signature gone.
- A real `diff --cc` conflict: `^-` = 0, `+5 -0` counting the column-2 marker, `my file.txt` keeping its space.
- `core.quotepath` path renders as `été.txt`, so `rtk git diff | grep été` finds it.
- `--dst-prefix=w/` names `git.rs`; `diff --cc dup dup` names `dup dup`.
- `-U10` leading context comes out adjacent to the change (`ctx8 ctx9 ctx10`).
- Change-line parity against develop: 453 = 453 on a real diff, so nothing is lost relative to the old behaviour.
- `cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test --all` — clean, 2527 passing.

**The two commits I pushed** (thanks for the maintainer-edit box): `e224cf4` decodes git's quoting rather than only unwrapping it, gates the midpoint split on the header kind so a single-path `diff --cc` is not split, matches prefixes against each other so `--dst-prefix` reads like any other pair, and honours `--word-diff=none` last-one-wins. `af01931` is a one-line plural fix — a hunk truncated at exactly one line over the cap said `1 deletions truncated`.

**Two shapes consciously accepted**, both documented in the code:

- A `--no-prefix` rename between two directories of equal path length reads as the shared trailing path. Space-separated paths with no prefix are ambiguous by construction; there is no signal to disambiguate on.
- Lines between a hunk's declared end and the next header are dropped. That is the mbox fix working as intended, and it means `git log -p | rtk pipe --filter git-diff` no longer shows `commit <sha>` / `Author:` lines as indented context. The old rendering was accidental and `rtk git log -p` does not route through `compact_diff`, so this is not worth holding the PR for — but it is a real difference for that one pipe.

Actually blocked, waiting for https://github.com/rtk-ai/rtk/pull/3793 to be merged

PR #3265, opened 2026-07-27
feat(git): filter large git show blob dumps with recovery and Latin-1 decoding

breisnerlopez:
## Summary

- **`feat(git)`**: `git show <rev>:<path>` blob dumps (previously passed through at 0% savings) are now filtered — large text blobs are capped to an 8 KiB preview with a `tail -n +N` recovery pointer (the full blob is tee'd); small blobs, tree listings, binary, and blobs too large to recover intact pass through unchanged.
- **Encoding-aware capture**: blobs are captured as raw bytes and decoded encoding-aware — ISO-8859 / Latin-1 files (e.g. Oracle PL/SQL `.pck`) are transcoded losslessly instead of being corrupted into U+FFFD by lossy UTF-8, which also unblocks their compression. Ambiguous single-byte encodings (CP1252) and binary pass through raw. Measured on a real 77 KB Latin-1 blob: **0% → 89% savings** with the `É`→`�` corruption removed. (Same class of fix as #1087.)

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` — clean; blob/decode tests green on current `develop`.
- [x] Unit + fixture tests cover decode (UTF-8 / Latin-1 / CP1252 / binary), byte-cap windowing, recovery-offset math, and edge cases (tree passthrough, binary passthrough, UTF-8 boundary no-panic, giant single line, exceeds-recoverable passthrough, no-trailing-newline).
- [x] Manual testing:
  - `git show HEAD:src/main.rs` → ~120 KB → 8 KB, `tail -n +N` recovery reconstructs byte-for-byte
  - `git show HEAD:<latin1.pck>` → 89% savings, no `�`, recovery byte-identical to `iconv -f ISO-8859-1`
  - `git show HEAD:<dir>` (tree) and `HEAD:<binary>` → passthrough unchanged

> Scope note: this is the `feat(git)` part of the earlier bundled PR #3064, **split into a focused PR** (per the single-focus rule) and **rebased onto current `develop`**. The independent `fix(tee)` recovery-slug fix from that bundle is submitted separately; the subjective advisory-hints refactor was dropped. #3064 will be closed in favor of these focused PRs.

PR #3274, opened 2026-07-28
fix(hook): split compound commands on newlines for rewriting

breisnerlopez:
### Problem

The hook rewriter splits a command line on `&&`, `||`, `;` and `|` before rewriting each clause, but **not on a bare newline**. So a multi-line command such as

```
cd /some/worktree
grep foo bar.txt
```

is treated as a single unrewritable segment and escapes rewriting entirely — `grep`/`git`/`find`/… on the following lines never become `rtk grep`/… This is common in `cd <dir>`-prefixed and worktree-oriented workflows, where it drops rtk coverage on those commands to ~0%.

### Fix

- Add `tokenize_with_newlines` (thin wrapper over the existing `tokenize_inner(_, true)`) and use it in `rewrite_compound`, so a bare `\n` is treated as a clause separator like `;`. The `\n` is re-emitted verbatim in the rejoin (no surrounding spaces), preserving the original layout.
- Include `\n`/`\r` in the `has_compound` fast-path so a command that *starts* with `rtk` but has more lines (`rtk gain\nls -la`) still falls through and rewrites the later lines.
- Collapse `\r\n`/lone `\r` to a single `\n` separator so CRLF input doesn't emit a spurious blank line.

### Safety

Command-substitution safety is unaffected: the permission/hook layer already defers on `$(...)`/backticks (and other unattestable constructs) *before* rewriting, and the allow-per-segment permission verdict already splits on newlines (the existing newline-bypass CVE guards). This change only touches the rewriter, which runs after and independently of the permission verdict — it can't elevate a decision or change how a dangerous command is treated. Verified end-to-end that `git status\nrm -rf …` is byte-identical to before (no auto-allow) and that multi-line `$(...)` defers with no rewrite.

### Tests

New unit tests for: `cd`-prefixed newline, both-sides rewrite, newline+pipe, trailing newline, `rtk`-prefixed multi-line, CRLF collapse; plus a hook-level regression test that multi-line command substitution defers. `cargo fmt --check` and `cargo clippy --all-targets` clean; full suite green.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -2,7 +2,6 @@
 
 use crate::core::utils::composer_bin_dirs;
 use regex::{Regex, RegexSet};
-use std::borrow::Cow;
 use std::path::Path;
 use std::sync::LazyLock;
 
@@ -702,15 +701,6 @@ fn rewrite_compound(
     excluded: &[ExcludePattern],
     transparent_prefixes: &[String],
 ) -> Option<String> {
-    // Normalize CRLF/lone-CR to LF so one logical newline yields exactly one
-    // separator (avoids emitting a blank line for `\r\n`). Both the tokens and the
-    // slices below are taken from `cmd`, so they stay consistent.
-    let normalized: Cow<str> = if cmd.contains('\r') {
-        Cow::Owned(cmd.replace("\r\n", "\n").replace('\r', "\n"))
-    } else {
-        Cow::Borrowed(cmd)
-    };
-    let cmd = normalized.as_ref();
     // Newline-aware: a bare `\n` separates clauses just like `;`, so multi-line
     // commands (e.g. `cd /x\ngrep foo`) get each clause rewritten instead of being
     // swallowed as a single unrewritable segment.
@@ -3964,6 +3954,23 @@ mod tests {
         assert_eq!(
             rewrite_command_no_prefixes("cd /x\r\ngrep foo", &[]),
             Some("cd /x\nrtk grep foo".into())
+        );
+    }
+
+    #[test]
+    fn test_rewrite_compound_cr_inside_quotes_not_a_separator() {
+        // A raw CR *inside* a quoted argument must not be treated as a clause
+        // separator: it stays part of the (single) segment and is preserved
+        // verbatim. Regression guard for over-broad CRLF normalization that used
+        // to `replace('\r', "\n")` across the whole command, splitting quoted CRs.
+        assert_eq!(
+            rewrite_command_no_prefixes("grep 'a\rb' file.txt", &[]),
+            Some("rtk grep 'a\rb' file.txt".into())
+        );
+        // Same for a CRLF sequence embedded in a quoted argument.
+        assert_eq!(
+            rewrite_command_no_prefixes("grep \"a\r\nb\" file.txt", &[]),
+            Some("rtk grep \"a\r\nb\" file.txt".into())
         );
     }
 
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -2,10 +2,14 @@
 
 use crate::core::utils::composer_bin_dirs;
 use regex::{Regex, RegexSet};
+use std::borrow::Cow;
 use std::path::Path;
 use std::sync::LazyLock;
 
-use super::lexer::{shell_split, split_on_operators, tokenize, ParsedToken, PipeKind, TokenKind};
+use super::lexer::{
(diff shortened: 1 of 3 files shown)


breisnerlopez:
This is the highest-impact of my open PRs on the token side: without it, compound commands split across newlines (`cd …`⏎`grep …` on separate lines) skip rewriting entirely, so their output streams through unfiltered — a big chunk of savings lost in parallel/worktree flows. CI's green and it's mergeable. Would appreciate a review when you have a moment, @aeppling 🙏

breisnerlopez:
Friendly nudge — this one's been open ~10 days with green CI and CLA signed. @aeppling @KuSh, when you have a spare cycle, would you mind taking a look?

It's the highest token-impact of my open PRs: without it, compound commands split across newlines (`cd …`⏎`grep …`) skip rewriting entirely and stream their output raw, so rtk saves nothing on them. The fix is scoped to the lexer/rewriter (newline splitting), keeps the substitution/permission guards untouched, and adds 7 tests.

For batch review, my other single-focus PRs on `develop` are: #3265 (git-show blob filter), #3266 (tee slug collisions), #3258 / #3259 (grep flag collisions), #3275 (pnpm global flags). No rush on any — happy to rebase or split further if that helps. Thanks for the project!

src/discover/registry.rs
```diff
@@ -693,7 +702,19 @@ fn rewrite_compound(
...
-    let tokens = tokenize(cmd);
+    // Normalize CRLF/lone-CR to LF so one logical newline yields exactly one
+    // separator (avoids emitting a blank line for `\r\n`). Both the tokens and the
+    // slices below are taken from `cmd`, so they stay consistent.
+    let normalized: Cow<str> = if cmd.contains('\r') {
+        Cow::Owned(cmd.replace("\r\n", "\n").replace('\r', "\n"))
+    } else {
+        Cow::Borrowed(cmd)
+    };
+    let cmd = normalized.as_ref();
```
KuSh, line 713: This seems a bit too broad and could replace content inside quoted arguments, for example `grep 'a\rb' file.txt` with a raw CR byte inside the single quotes.

Suggestion: collapse CR/CRLF only on the newline tokens the tokenizer already isolates, since it already skips quoted content correctly.

It would also be worth adding a test for that.

KuSh (changes requested):

src/discover/registry.rs
breisnerlopez: Good catch, thank you — fixed exactly as you suggested in 33d4823. Instead of the global `replace()`, the collapse now happens on the newline token the tokenizer already isolates: `tokenize_inner` consumes a `\r\n` pair (and a lone `\r`) as a single `Operator("\n")` token, so CRLF still yields exactly one separator with no spurious blank clause, and a CR/LF *inside* quotes is never touched — it's consumed by the quote branch before the newline arm is ever reached. The global normalization and its now-orphan `Cow` import are gone.

Added the test you asked for plus a couple more:
- `test_rewrite_compound_cr_inside_quotes_not_a_separator` — raw CR and CRLF inside single/double quotes stay a single rewritten segment.
- On the shared-tokenizer / permission-gating side: `test_split_perms_crlf`, `test_split_perms_cr_inside_quotes_not_split`, and a CRLF case in `test_attestable_subshell_and_separators`, confirming a command hidden after a CRLF is still split into its own segment while a quoted CR can't smuggle one past the per-segment check.

`cargo fmt --check` + `clippy --all-targets` clean, full suite green (2511/0). CI's rerunning on the push.

KuSh (approved):
Thanks, LGTM!

KuSh:
Hi @breisnerlopez, unfortunately fc8054e already solves the exact same problem your PR targets.

The only thing your PR still adds is support for bare `\r` without `\n`, which is a narrow edge case.

You have two options:
1. Rebase onto current `develop` and diff your intended behavior against `rewrite_multiline_block`.
2. Close this PR and open a small follow-up just for the lone-`\r` separator gap, if you think it’s worth covering, using `fc8054e`’s safer per-line, guarded approach.

Sorry I didn’t catch that sooner

breisnerlopez:
Thanks @KuSh — makes sense, fc8054e covers the main case cleanly and I hadn't seen it land. Going with option 2: closing this in favor of a small, focused follow-up for just the lone-`\r` separator gap, built on fc8054e's per-line guarded path (`rewrite_multiline_block`) rather than my broader rewrite: #3600. Appreciate the pointer.

PR #3278, opened 2026-07-28
feat(recall): SQLite recall system

aeppling:
Sqlite recall store replaces tee files as the default recovery mechanism.

## What changed

- Recovery hints become `[full output: rtk recall <hash>]` / `[+N hidden: rtk recall <hash>]` — the default recall returns only what the agent didn't see.
- `rtk recall <hash>` (prefix ok) with `--full`, `--from N`, `--lines N`, `--grep`, `--list` (human inspection; the hint is the whole agent interface). Byte-faithful store (gzip BLOB; input is the captured command text, as with tee files), 10MB cap, FIFO 200 entries / 30 days, content-hash dedup.
- `rtk gain --recalls` — per-filter rate of elided output the agent goes back for, with a `Verdict` column naming an over-recalled filter (`too aggressive` above 30%, `-` under 5 elisions) and pointing at the issue tracker. A recall costs an extra API turn plus the full output, so it cancels the savings on that command. Exact in sqlite, lower bound in tee mode.
- `rtk gain --reset` now clears the recall counters too (stored outputs are kept).
- `rtk config recall <sqlite|tee|disabled>` — switch modes without editing files (preserves comments, migrates `[tee]` keys).
- Telemetry (existing opt-in consent): per-filter recall counters (`recall_stats`, `recall_mode`) — filter names from a fixed allowlist, anything else folded into `other`; no hashes, paths, arguments or content.
- `recall.db` created owner-only (0600/0700, WAL/SHM included), corrupted/unavailable DB or disabled mode stays silent — no extra token is ever emitted into the agent session, and command output/exit code are never affected.

## Migration

- No config → sqlite on first post-upgrade failure; old `.log` files untouched.
- Legacy `[tee]` section → auto-mapped to `mode = "tee"` (behavior preserved) or `disabled` if `enabled = false`.
- One-shot notices in `rtk init`/`rtk gain` for legacy users, pointing to `rtk config recall`.
- Legacy `[tee] mode = "always"` keeps its behavior, mapped to `tee_on_success = true`. Kill switches: `RTK_RECALL=0` (legacy `RTK_TEE=0`), `RTK_RECALL_DB`.

## Validation

3395 unit tests + e2e suites (byte fidelity incl. NUL/ANSI/CRLF, 15MB truncation, concurrent writes, corrupted DB, migration scenarios) + real-agent fuzz. CI green on ubuntu/macos/windows (rusqlite bundled).

PR #3280, opened 2026-07-28
feat(stash): content-addressed output stash + rtk retrieve

texas697:
## What & why

RTK compresses tool output brilliantly, but compression is **lossy** — when a filter drops detail, that detail is gone. The one recovery path today is the tee file (`[full output: ~/path]`), which is failures-only, rotates at 20 files, and has no addressable handle.

This PR adds the missing **retrieve** half: park full output behind a short content-addressed handle so the agent reads the compact preview and pulls the full bytes — or re-interrogates them — only when it actually needs to.

Opening as a **draft for discussion** per CONTRIBUTING's "discuss core features with maintainers before" guidance. Happy to split, trim, or re-shape.

## The one-line integration win

Rather than touch the 50+ `force_tee_hint` / `tee_and_hint` call sites, stash is wired into the **three** functions they all funnel through. So every existing recovery hint automatically gains a recall handle — zero changes at the call sites:

```
[full output: ~/…/x.log] [recall: rtk retrieve d51141d7]
```

## Surface

```bash
some-cmd | rtk stash                 # park stdin → prints a handle
rtk stash --list                     # what's parked
rtk stash --gc                        # prune (retention + LRU + missing)

rtk retrieve <handle>                 # full parked output
rtk retrieve <handle> --grep TS2345   # re-interrogate without pulling it whole
rtk retrieve <handle> --lines 40-90   # a line range
rtk retrieve <handle> --head/--tail N
rtk retrieve <handle> --meta          # command, size, age
```

## Design

- **Blobs**: SHA-256 content-addressed, gzip-compressed, git-style sharded at `~/.local/share/rtk/stash/<xx>/<hash>.gz`. Identical content dedups to one blob.
- **Index**: a `stash` table in the existing `history.db` (CREATE-IF-NOT-EXISTS, additive — no migration, older binaries ignore it).
- **Lifecycle**: retention-day + LRU byte-cap GC + missing-blob pruning, run opportunistically on write.
- **Deps**: reuses `sha2` + `flate2`, already in `Cargo.toml`. No new dependencies.
- **Speed**: the critical path is unchanged for anything below `min_bytes` (default 2 KB); the automatic path is config-gated and error-swallowing — a stash failure can never break a tee hint.

## Config (`[stash]`)

```toml
[stash]
enabled = true
auto = true            # record recall handles from tee hints
min_bytes = 2048
retention_days = 7
max_bytes = 536870912  # 512 MiB, LRU-evicted
handle_len = 8
```

Env overrides: `RTK_STASH=0` (disable), `RTK_STASH_DIR` (relocate blobs).

## Tests

28 new tests (store put/resolve/retrieve/dedup/gc/LRU, retrieve filters, formatting). Full suite: **2587 pass, 0 fail**. `clippy` + `rustfmt` clean. `rtk init` docs updated for both templates.

## Open questions for maintainers

- Naming: `stash` vs `park`/`cache` — `git stash` overlap is only at `rtk git stash` (different enum), no collision, but happy to rename.
- Whether the automatic tee-hint integration should default `auto = true` or ship opt-in.
- Compression choice — went with `flate2` (already vendored); `zstd` would compress tighter if you're open to the dep.

— Jean Claude Damme

https://claude.ai/code/session_01VA9etjKxKpHeM26Y1haR6C

### DIFF
--- a/src/cmds/system/retrieve_cmd.rs
+++ b/src/cmds/system/retrieve_cmd.rs
@@ -1,0 +1,169 @@
+//! `rtk retrieve` — pull a parked stash blob back by handle, optionally sliced
+//! or grepped so the agent can re-interrogate large output without paying for
+//! the whole thing again.
+
+use anyhow::{anyhow, Result};
+
+use crate::core::stash::StashStore;
+
+/// Entry point for the `rtk retrieve` subcommand.
+///
+/// Filters compose in a fixed order over the parked lines: grep → lines → head
+/// → tail. `meta` short-circuits to a metadata dump.
+pub fn run(
+    handle: &str,
+    lines: Option<String>,
+    grep: Option<String>,
+    head: Option<usize>,
+    tail: Option<usize>,
+    meta: bool,
+) -> Result<()> {
+    let store = StashStore::open()?;
+
+    if meta {
+        let row = store.resolve(handle)?;
+        crate::cmds::system::stash_cmd::print_meta(&row, store.handle_len());
+        return Ok(());
+    }
+
+    let (_row, content) = store.retrieve(handle)?;
+
+    let filtered = apply_filters(&content, lines.as_deref(), grep.as_deref(), head, tail)?;
+    print!("{filtered}");
+    if !filtered.ends_with('\n') && !filtered.is_empty() {
+        println!();
+    }
+    Ok(())
+}
+
+fn apply_filters(
+    content: &str,
+    lines: Option<&str>,
+    grep: Option<&str>,
+    head: Option<usize>,
+    tail: Option<usize>,
+) -> Result<String> {
+    let mut current: Vec<&str> = content.lines().collect();
+
+    if let Some(pattern) = grep {
+        let re = regex::Regex::new(pattern)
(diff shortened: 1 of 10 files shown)


KuSh:
Thanks for prototyping the retrieve half, and for opening it for discussion. #3278 (merged 2026-09-08) landed a recall system over the same ground: a gzip blob store in SQLite with content-hash dedup, hints of the form "[full output: rtk recall <hash>]" wired through the same funnel your PR hooks, and rtk recall with --full, --from, --lines, --grep and --list. It rewrote src/core/tee.rs, removing about 660 lines, so the three functions this PR taps are no longer the ones that produce the hints and the stash layer cannot be rebased onto them. Several ideas here are genuinely absent from develop and would be worth separate issues or narrow PRs on the recall model: parking output from a manual pipe such as `some-cmd | rtk stash`; --head/--tail N and --lines START-END ranges, where recall today takes only --lines N and --from N; a --meta view of a stored entry; and a total-store byte cap with a gc command, since develop retains 200 entries for 30 days with a 10MB per-entry limit. Please comment or reopen if I have read the overlap wrong.

Issue #2317, opened 2026-06-07
Tool filters mask hard failures with benign summaries (pytest collection error -> "No tests collected")


danscMax:
**Repo:** rtk-ai/rtk · **Version:** rtk 0.42.0 · **Platform:** Windows, used as the Claude Code Bash hook

## Summary
RTK's per-tool filters (`pytest`, and likely `test`/`tsc`/`mypy`/`cargo`/`vitest`/`jest`) can replace a **hard failure** (collection error, `ImportError`, compile/load error) with a benign-looking summary that **misrepresents** the failure. The full output is correctly saved to the tee file, but the **inline summary** that reaches the LLM/dev is misleading — and the agent acts on the inline text, not the tee file.

## Concrete repro
1. Project where `tests/conftest.py` raises `ImportError` during collection (e.g. an optional dep like `weasyprint` imported transitively at import time and not installed).
2. Run (via the hook, i.e. `rtk pytest tests/foo.py`).

**Observed:**
```
Pytest: No tests collected
[full output: ~/.local/share/rtk/tee/..._pytest.log]
```

**Expected (something like):**
```
Pytest: COLLECTION FAILED — ImportError: No module named 'weasyprint' (conftest.py:34)
[full output: ~/.local/share/rtk/tee/..._pytest.log]
```

## Why it matters
`No tests collected` reads as *"there are no tests / nothing to run"* — a benign state. An agent/dev then debugs **test discovery / paths** instead of the real cause (a broken import in `conftest`). In real use this cost a wrong-direction round-trip. RTK's value is condensing output — but on a hard failure the condensed form must not change the **meaning**.

## Proposed fix (priority order)
1. **Minimum:** on non-zero exit, the tool-filter summary must surface the actual error (tail of stderr / the exception line), not a count-only summary. The data is already in the tee file (`[tee] mode = "failures"`) — just reflect it inline.
2. **Better:** distinguish failure CLASSES for pytest — `no tests matched` vs `collection error` vs `N failed` — and label them differently.
3. **General principle ("raw-on-failure"):** for tool filters, when `exit != 0`, prefer passing the relevant error lines through verbatim over a heuristic summary. This aligns with the existing `rtk err` ("show only errors/warnings") philosophy.

## Affected filters (likely)
`pytest`, `test`, `tsc`, `mypy`, `cargo`, `vitest`, `jest` — any "show only failures / grouped errors" filter that can also be hit by an error **before** the run proper (collection / compile / load phase).

---

[...]

PR #3295, opened 2026-07-29
fix(tsc): guard "No errors found" behind exit code

### LINKED ISSUE #3220: tsc filter: "TypeScript: No errors found" is printed on a non-zero exit, so a tsc that never ran reads as a clean typecheck

NekomyaDev:
Closes #3220. (Re-targeted to `develop` per CONTRIBUTING.md — supersedes #3294.)

## Problem

`TscHandler::format_summary` ignored the process exit code: whenever zero parseable `TS####` errors were captured, it returned `TypeScript: No errors found` — even when `tsc` exited non-zero without emitting parseable diagnostics (e.g. missing `tsconfig.json`, invalid compiler flags, fatal config errors like TS5058). The filter could therefore report success on a failed run.

## Fix

`format_summary` now guards on `exit_code`:

- `exit_code == 0` + no parseable errors → unchanged: `TypeScript: No errors found`
- `exit_code != 0` + no parseable errors → reports `TypeScript: failed (exit code N) with no parseable TS errors` and appends the last 10 non-empty lines of raw output so the fatal diagnostic (which the streaming filter otherwise drops) stays visible

## Tests

Two new regression tests:

- `test_tsc_stream_failure_without_parseable_errors` — non-zero exit with only a fatal TS5058 line must **not** contain "No errors found", must surface exit code and the fatal line
- `test_tsc_stream_success_still_reports_no_errors` — zero exit with empty output still reports "No errors found"

All existing tsc streaming/filter tests are unaffected (the success path is unchanged).

### DIFF
--- a/src/cmds/js/tsc_cmd.rs
+++ b/src/cmds/js/tsc_cmd.rs
@@ -77,9 +77,33 @@ impl BlockHandler for TscHandler {
         line.starts_with("  ") || line.starts_with('\t')
     }
 
-    fn format_summary(&self, _exit_code: i32, _raw: &str) -> Option<String> {
+    fn format_summary(&self, exit_code: i32, raw: &str) -> Option<String> {
         if self.error_count == 0 {
-            return Some("TypeScript: No errors found\n".to_string());
+            if exit_code == 0 {
+                return Some("TypeScript: No errors found\n".to_string());
+            }
+
+            // tsc exited non-zero without emitting parseable TS errors
+            // (e.g. missing tsconfig, invalid compiler flags, fatal config
+            // errors). Claiming "No errors found" here is a false positive,
+            // so surface the failure and the raw output tail instead.
+            let tail: Vec<&str> = raw
+                .lines()
+                .filter(|l| !l.trim().is_empty())
+                .rev()
+                .take(10)
+                .collect::<Vec<_>>()
+                .into_iter()
+                .rev()
+                .collect();
+            let mut result = format!(
+                "TypeScript: failed (exit code {}) with no parseable TS errors\n",
+                exit_code
+            );
+            if !tail.is_empty() {
+                result.push_str(&format!("Output:\n{}\n", tail.join("\n")));
+            }
+            return Some(result);
         }
 
         let mut result = format!(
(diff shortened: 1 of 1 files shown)


HetCreep:
Two notes for whoever reviews this, from an external audit of **v0.44.2 (`700bdde`)** that landed on the same defect.

*Provenance: AI-assisted source audit, human-directed. Source and `git log` only — I did not build either branch.*

### There are currently three PRs for this one bug

- **#3221** — minimal: `exit_code != 0` → `TypeScript: compilation failed (exit N)`.
- **#3295** (this one) — same guard, plus the last 10 non-empty raw lines appended.
- **#3294** — superseded by this PR per its own description.

They are not conflicting approaches so much as the same change at two levels of detail, which may be why both are sitting. For what it is worth from the outside, appending the raw tail is the part that matters most to the actual consumer here — an agent that sees only `compilation failed (exit 2)` will usually re-run the command to find out why, which costs more than the tokens the filter saved. But that is the reviewer's call, and either is strictly better than the current behaviour.

### The pattern this restores already exists in the tree, one directory over

`src/cmds/rust/cargo_cmd.rs:99` implements the same `BlockHandler::format_summary` contract and already gates its success line on the exit code:

```rust
if self.error_count == 0 && self.warnings == 0 && exit_code == 0
```

That guard was added in **`39a4d28` (2026-07-03)**, *"fix(cargo): …"*. `tsc_cmd.rs:80`'s unguarded `return Some("TypeScript: No errors found\n")` has been unchanged since **`b82fa85` (2026-01-29)** — so this is a fix that landed for one handler and did not propagate to its sibling, rather than a new class of bug. Worth citing in the PR body: it gives the reviewer an in-tree precedent to compare against instead of evaluating the change cold.

The same shape recurs beyond `tsc` — `go vet` is the one I looked at closely (`src/cmds/go/go_cmd.rs:117`, where the `build` path got `run_filtered_with_exit` in `f69ad6e` and `vet` did not); I have written that up on #2281. #2317 tracks the general case.

No objection to this PR — posting only so the duplicate-PR situation and the `39a4d28` precedent are visible in one place.

AlvinManojAlex:
Does this PR solve the bug, or is still a work-in-progress?

KuSh:
Thanks for this. Develop already guards the summary on the exit code: #2232 (merged 2026-08-29) reworked TscHandler::format_summary so a non-zero exit with no parseable diagnostics reports "compiler exited with code N, but RTK parsed no diagnostics" and prints the raw output instead of "No errors found", and it closed the same issue #3220. It also strips ANSI from the dumped lines and keeps both ends of long output rather than only the tail, and carries regression tests for the non-zero-exit and clean-exit paths. Closing as covered, but please comment or reopen if you see a case that still slips through.

Issue #3241, opened 2026-07-27
Add aarch64-unknown-linux-musl release build (static, glibc-independent)


damovsky:
## Problem

Release assets for aarch64-linux only ship `rtk-aarch64-unknown-linux-gnu.tar.gz` (dynamically linked against glibc). x86_64-linux gets both `-gnu` and `-musl` variants, but aarch64-linux has no musl option.

This means the aarch64-gnu binary has a hard floor on the host's glibc version. Concretely: v0.44.0's aarch64-gnu build requires GLIBC 2.39, but Ubuntu 22.04 LTS (a very common Graviton/ARM64 EC2 target, still receiving security updates until 2027) ships GLIBC 2.35 -- the prebuilt binary fails to run at all:

```
/home/ubuntu/.local/bin/rtk: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by rtk)
```

## Suggested fix

Add `aarch64-unknown-linux-musl` to the release matrix, matching the existing x86_64-unknown-linux-musl target. Since the project already builds a musl target for x86_64, the toolchain/cross setup presumably already exists -- this should mainly be a matter of adding the target triple to the release build matrix.

## Workaround in use today

Building from source with `cargo build --release` works fine (rustc 1.97 on Ubuntu 22.04 arm64, no issues), so this is purely a release-artifact gap, not a code compatibility problem. But requiring a source build defeats the point of having a one-line installer script for users on LTS-vintage distros.

## Environment
- OS: Ubuntu 22.04.5 LTS (Jammy), aarch64 (AWS Graviton, m8g.xlarge)
- glibc: 2.35
- rtk version attempted: v0.44.0 (release binary), works fine when built from source at HEAD (0.42.4 per Cargo.toml, may be behind latest tag)

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 92% |

### Summary

Add aarch64-unknown-linux-musl to the release build matrix so aarch64-linux users on older glibc (e.g. Ubuntu 22.04) aren't forced to build from source.

[...]

PR #3318, opened 2026-07-30
fix(ci)/cicd add aarch64 musl release

stavrosfilippidis:
## Summary
- Adds `aarch64-unknown-linux-musl` to the release build matrix, enabling ARM64 users on glibc-constrained systems (Ubuntu 22.04 LTS, Graviton EC2) to use prebuilt binaries
- Updates Homebrew formula to prefer musl (static linking) for Linux ARM64, matching the existing x86_64-musl approach for cross-platform compatibility
- Updates checksum parsing and documentation to include the new aarch64-musl binary in releases

Addresses: https://github.com/rtk-ai/rtk/issues/3241

## Test plan
- [x] Tested locally: `docker run ubuntu:latest` with full `cargo build --release --target aarch64-unknown-linux-musl` ✓ Binary created successfully (~5.2MB)
- [x] Verified YAML syntax: `yamllint .github/workflows/release.yml` ✓
- [x] Pre-commit gate passed: `cargo fmt --all --check && cargo clippy --all-targets && cargo test --all` ✓ (2565 tests pass)
- [ ] CI validation: Release workflow will build all 5 targets (macOS x86/ARM, Linux x86/ARM-musl/ARM-gnu, Windows)
- [ ] Homebrew integration: Verify formula generates correctly with aarch64-musl checksums

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,6 +48,10 @@ jobs:
             os: ubuntu-latest
             archive: tar.gz
             musl: true
+          - target: aarch64-unknown-linux-musl
+            os: ubuntu-latest
+            archive: tar.gz
+            musl: true
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-latest
             archive: tar.gz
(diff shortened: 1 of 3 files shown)


k-k1:
Another data point for this, from a platform that isn't a hobbyist board: the same failure hits a **Debian 12 container running on AWS Graviton4** (EC2 `m8g`), so it isn't limited to Raspberry Pi / RHEL as the earlier reports might suggest.

```
rtk: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by rtk)
```

rtk 0.45.0, `rtk-aarch64-unknown-linux-gnu.tar.gz`, Debian 12 (bookworm) = glibc 2.36.

What makes this expensive downstream is *where* it fails: **the download and the `checksums.txt` verification both succeed** — only execution does not. An installer that pins by sha256 (ours does) therefore reports a clean, verified install and ships a binary that cannot start, so the failure surfaces at first use instead of at install time. We ended up having to run `rtk --version` after installing and delete the binary when it fails, which is a strange thing to have to do to a checksum-verified artifact.

For what it's worth on the approach here: `rtk-x86_64-unknown-linux-musl.tar.gz` works fine for us in the *same* image, so the musl path is already proven on the sibling target, and this PR keeps `aarch64-unknown-linux-gnu` in the matrix so existing consumers of that asset aren't affected.

Happy to test an `aarch64-unknown-linux-musl` build on Graviton if that would help move this along.

KuSh:
Closing in favour of #2089, which makes the same aarch64 musl switch and goes further: it updates install.sh so the installer actually downloads the renamed artifact, and adds Cross.toml so the telemetry vars reach the cross container. That PR is already under review. Please take any release workflow feedback there.

PR #3319, opened 2026-07-30
feat(rewrite): rewrite multiline blocks

aeppling:
## Summary

Multi-line Bash blocks (a shape coding agents emit constantly) were treated as one command: the `rtk` prefix landed on the first line and every following line ran raw and unfiltered. This PR splits blocks at the newline tokens the quote-aware lexer emits and rewrites each line through the existing single-line path. Implements the line-by-line rewrite proposed in #1243.

```
$ rtk rewrite 'git status
# check history
  grep -rn foo src'
rtk git status
# check history
  rtk grep -rn foo src
```

- Lines continued by a trailing `&&` / `||` / `|` / `|&` are joined and rewritten as **one logical command** (`git status &&` ⏎ `git log -3` → `rtk git status && rtk git log -3`), exactly what develop produced for the single-line form.
- Quoted multi-line strings (multi-line commit messages) keep rewriting as one prefixed command, as develop always did.
- Blank lines, comment lines, indentation, and CRLF separators are preserved verbatim.
- **Conservative by design** — the whole block passes through untouched (native permission handling sees the original) whenever a line can't be attested: control flow keywords, heredocs (including `<<` split across a `\<NL>` continuation), grouping/arithmetic spanning lines, swallowed newlines (quote state opened by an apostrophe in a comment), a continuation operator hidden behind a trailing comment, `$'…'` containing `\'` (the one ANSI-C case where the lexer's view diverges from bash), and `|&` stderr pipes.

## How this was built

This PR supersedes #3188 (@revopsrocks), taken over after the author became unresponsive; their three commits are cherry-picked with authorship preserved.

1. **#3188 as submitted** — line-by-line rewrite + two adversarial review rounds by the original author (swallowed newlines, cross-line arithmetic, array literals/function bodies, comment-stripped independence checks, ANSI-C bail).
2. **Maintainer review fixes** (review on #3188, applied here):
   - narrowed the `$'` bail to the actual hazard (`\'` inside a `$'…'` span) so an unrelated `echo $'a\tb'` line no longer disables the block;
   - cross-line `&&`/`||`/`|` lists are now joined and rewritten instead of declined (develop already handled the joined form);
   - `comment_start` recognizes `#` after operator bytes (`|&;()`), so `git log |# keep` ⏎ `grep -f patterns.txt` passes through instead of rebuilding a guarded pipeline;
   - the heredoc/`$((` gate runs on the bash-faithful continuation join (`\<NL>` → nothing) before the space-join normalization erases the operator.
3. **Differential fuzz round** (below) — found one genuine regression vs develop (quoted multi-line strings declined) and fixed it: zero emitted newline tokens + balanced quote state at EOF proves the block is one logical command, which is rewritten whole; unbalanced quote state (the apostrophe-comment hazard) still declines.

## Differential fuzz validation

971-case corpus (real per-ecosystem commands, compounds, multiline blocks, cross-line continuations, control flow, deep quoting, heredoc/redirect matrix, permission-bypass attempts, plus the #3188 adversarial cases), run 3-way through the real `rtk rewrite` hook path — develop vs this PR vs #2781 — with every rewrite validated by `bash -n`:

| binary | rewrote | invalid rewrites | crashes |
|---|---|---|---|
| develop | 598/971 (61.6%) | 0 | 0 |
| **this PR** | **684/971 (70.4%)** | **0** | 0 |
| #2781 (2c89559) | 717/971 (73.8%) | 1 | 0 |

- **105 blocks newly rewritten**; **208 upgraded from partial to full** (of 579 cases both rewrote, develop had prefixed only the first line in 208 — silent incomplete coverage the percentage never showed).
- **19 deliberately declined** that develop "rewrote": apostrophe-in-comment blocks (develop's output left lines 2+ unfiltered anyway), degraded comment-in-continuation joins, no-op trims.
- **Pipe-consumer protection verified**: zero cases where any stage feeding a pipe gained an `rtk` prefix; cross-line pipes join into the same `rewrite_compound` machinery (final-stage-only, `pipeline_final_command_is_safe`, `grep -f` rejected, `|&` raw).
- #2781's coverage lead is partly mis-rewrites its own `bash -n` check can't see; this PR declines all of them: `cat <\` ⏎ `<EOF` → `cat < <EOF` (the corpus's only syntax error), `$(\` ⏎ `(1+2))` arithmetic corrupted into command substitution, `rtk` spliced inside a `$'…'` literal's value, `grep -f` rebuilt as the final stage of a pipeline behind a `|# comment`, and function bodies prefixed (changes `x=$(foo)` captured output).

## Permission verdicts

Verified live against Claude Code allow/deny rules (`.claude/settings.json`): a block auto-allows (exit 0) only when **every** line independently matches an allow rule (#1213 aggregation); one unmatched line downgrades to ask; a deny match on any line wins outright; and a block whose newline was swallowed by quote state can never auto-allow — the same condition that blinds `split_for_permissions` forces the rewrite to decline, and exit 0 is impossible without a rewrite. `split_for_permissions` and the multiline splitter share the same newline-emitting tokenizer, so a verdict is always computed for exactly the lines being rewritten.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` (2609 passed, clippy clean)
- [x] 971-case 3-way differential fuzz (above), every rewrite `bash -n`-validated
- [x] ~45 unit tests on the multiline path covering the rewrite and every passthrough gate

## Follow-ups (out of scope here)

- Replace the hand-rolled attestation checks with a real Bash grammar (tree-sitter-bash or a pure-Rust parser) — the apostrophe-comment cluster (18 declined cases, a very common agent shape) needs comment-aware parsing to rewrite safely. Being discussed with maintainers; startup-latency and binary-size budgets to be measured first.
- Statement-level rewriting when a *different* line holds a redirect/substitution (#2781's remaining legitimate coverage lead).
- Check the fuzz corpus + runner into the repo as the regression suite for the above.

Ref #1243. Supersedes #3188. Related: #2781.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -637,6 +637,9 @@ fn line_breaks_independence(line: &str) -> bool {
         || line.ends_with('{')
         || line.starts_with(')')
         || line.starts_with('}')
+        || line.starts_with("((")
+        || line.ends_with("))")
+        || line.ends_with("((")
 }
 
 /// Rewrite each line of a multi-line block independently (issue #1243).
@@ -647,6 +650,14 @@ fn line_breaks_independence(line: &str) -> bool {
 /// line is an independent command (see [`line_breaks_independence`]); blank
 /// lines and comment lines are preserved verbatim, as is indentation and the
 /// original separator bytes (`\n` vs `\r\n`).
+///
+/// If any newline byte was swallowed by quote state, the block passes through
+/// untouched. The lexer has no comment awareness, so an apostrophe in a `#`
+/// comment opens quote state and hides the rest of the block — rewriting (or
+/// prefixing) such a block would act on lines no permission verdict was
+/// computed for. Passthrough hands the original command to the agent's native
+/// permission handling instead. Genuine quoted newlines (multi-line commit
+/// messages) also land here; forgoing that rewrite is the safe trade.
 fn rewrite_multiline_block(
     cmd: &str,
     excluded: &[ExcludePattern],
@@ -658,9 +669,9 @@ fn rewrite_multiline_block(
         .map(|t| t.offset)
         .collect();
 
-    // Newlines only inside quotes — a single logical command.
-    if newline_offsets.is_empty() {
-        return rewrite_single(cmd, excluded, transparent_prefixes);
+    let raw_breaks = cmd.chars().filter(|c| matches!(c, '\n' | '\r')).count();
+    if raw_breaks != newline_offsets.len() {
+        return None;
     }
 
     let mut segments = Vec::with_capacity(newline_offsets.len() + 1);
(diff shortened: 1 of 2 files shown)


pszymkowiak (approved):

src/discover/registry.rs
```diff
@@ -595,7 +622,296 @@ pub fn rewrite_command(
...
+    rewrite_compound(trimmed, excluded, transparent_prefixes)
+}
+
+/// Shell keywords that open or close a multi-line construct. A line inside a
+/// loop, conditional, case arm, function body, or group is not an independent
+/// command, so the whole block passes through untouched.
+const BLOCK_KEYWORDS: &[&str] = &[
+    "for", "while", "until", "if", "then", "else", "elif", "fi", "do", "done", "case", "esac",
+    "select", "function", "coproc", "{", "}", "(", ")",
+];
```
KuSh, line 634: Seems you can have multiline without backslash between `[[` and `]]` too

src/discover/registry.rs
```diff
@@ -595,7 +622,296 @@ pub fn rewrite_command(
...
+
+/// Byte offset where an unquoted `#` at the start of a word begins a trailing
+/// comment, if any. The lexer has no comment state, so the independence checks
+/// must ignore comment text themselves: `git log | # keep pipeline` continues
+/// the pipeline across the newline even though the line ends in comment text.
+fn comment_start(line: &str) -> Option<usize> {
```
KuSh, line 690: comment_start, line_has_unbalanced_grouping, ansi_c_quote_defeats_lexer, and quotes_balanced are four near-identical hand-rolled single/double-quote-tracking byte scanners with the same backslash-skip idiom. I know we want to switch to a real bash/sh parser in the future but it could be worth, at least, to have a shared low-level "quote-state walker" helper

src/discover/registry.rs
```diff
@@ -595,7 +622,296 @@ pub fn rewrite_command(
...
+        .iter()
+        .filter(|t| t.kind == TokenKind::Operator && t.value == "\n")
+        .map(|t| t.offset)
+        .collect();
+
+    if ansi_c_quote_defeats_lexer(cmd) {
+        return None;
+    }
+
+    let raw_breaks = cmd.chars().filter(|c| matches!(c, '\n' | '\r')).count();
```
KuSh, line 857: This one is non-obvious. Could be great to have a one-line comment on why raw_breaks/newline_offsets count \r and \n separately.

src/discover/registry.rs
```diff
@@ -595,7 +622,296 @@ pub fn rewrite_command(
...
+
+/// Rewrite each line of a multi-line block independently (issue #1243).
+///
+/// Split points are the newline tokens the quote-aware lexer emits, so a
+/// newline inside a quoted string (e.g. a multi-line commit message) never
+/// becomes a boundary. Lines continued by a trailing `&&`/`||`/`|`/`|&` are
+/// joined and rewritten as one logical command through the single-line path;
+/// any line [`classify_line`] marks unsafe passes the whole block through.
+/// Blank lines and comment lines are preserved verbatim, as is indentation
+/// and the original separator bytes (`\n` vs `\r\n`).
```
KuSh, line 831: minor: Would be great to document the caveat that joined continuation lines (`&&` / `||` / `|`) are not preserved (tested by `test_blank_line_inside_continuation_joins`)

Bonus point for another comment at the reconstruction site itself too (line 900).

KuSh (changes requested):

KuSh (approved):
LGTM, thanks

PR #3320, opened 2026-07-30
fix(git): preserve status trailing newline

### LINKED ISSUE #3303: `git status --short/--porcelain` output loses its trailing newline
rtk drops the final `\n` of `git status --short` (and `--porcelain`). Content is
byte-identical to native git otherwise — only the trailing newline is missing.

Repro (verified on rtk 0.43.0 and 0.44.1, x86_64-linux):

```
git init -b main repo && cd repo
echo a > tracked && git add tracked && git commit -m init
mkdir -p src/pkg && echo x > src/pkg/new.py
rtk git status --short > rtk.out
git status --porcelain  > raw.out
diff rtk.out raw.out
# 2c2
# < ?? src/pkg/new.py
# \ No newline at end of file
```

Impact: the last entry gets glued to whatever prints next (`?? new.pyrc=0`), and
`rtk git status --short | while read -r line; do ...; done` silently drops it.
We hit this for real — a newly added module was missed while staging a commit.

Cause: `src/cmds/git/git.rs::run_status` — the non-compact path builds the output
with `join("\n")` and emits it with `print!`, so no trailing newline is written.
Other commands (`git status` without flags, `git diff`, `git log`, `wc`, `grep`,
`ls`) are unaffected — verified byte-level.

Fix: print with a trailing newline when the filtered output doesn't already end in one.

```rust
if filtered.is_empty() || filtered.ends_with('\n') {
    print!("{}", filtered);
} else {
    println!("{}", filtered);
}
```

Happy to send this as a PR if useful.

lntutor:
## Summary

- preserve the source trailing newline after filtering explicit `git status` output
- add regression coverage for porcelain output consumed by line-oriented shell loops

Closes #3303

## Testing

- `cargo test`

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -814,11 +814,17 @@ fn filter_status_with_args(output: &str) -> String {
         result.push(line.to_string());
     }
 
-    if result.is_empty() {
+    let mut filtered = if result.is_empty() {
         "ok".to_string()
     } else {
         result.join("\n")
+    };
+
+    if output.ends_with('\n') {
+        filtered.push('\n');
     }
+
+    filtered
 }
 
 fn run_status(args: &[String], verbose: u8, global_args: &[String]) -> Result<i32> {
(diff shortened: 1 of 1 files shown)


lntutor:
Closing as a duplicate of #3307, which was opened first for the same issue and fix. Sorry for the overlap.

PR #3321, opened 2026-07-30
fix(discover): rewrite supported dotnet commands

### LINKED ISSUE #3300: dotnet test never rewritten: rule matches ^dotnet build only, though rtk dotnet supports test
## Summary

`rtk dotnet` advertises `test` support, but the rewrite rule only matches `dotnet build`, so `dotnet test` is never rewritten by the hook.

```
$ rtk dotnet --help
dotnet    .NET commands with compact output (build/test/restore/format)
```

Hook behaviour (`rtk hook claude` on `PreToolUse`/`Bash`), same shape of input for both:

```
$ echo '{...,"tool_input":{"command":"dotnet build"}}' | rtk hook claude
{"hookSpecificOutput":{...,"updatedInput":{"command":"rtk dotnet build"}}}

$ echo '{...,"tool_input":{"command":"dotnet test"}}'  | rtk hook claude
(empty output — no rewrite)
```

The rule in `src/discover/rules.rs` matches `build` only:

```rust
pattern: r"^dotnet\s+build\b",
rtk_cmd: "rtk dotnet",
rewrite_prefixes: &["dotnet"],
```

Consistent with that, `rtk discover` lists `dotnet build` under commands it handles and `dotnet test` under "TOP UNHANDLED COMMANDS".

## Why it matters

`dotnet test` was the second-highest unhandled command in my last 30 days (179 calls). Test output is long, repetitive, and mostly noise until something fails — the case RTK is best at, and `rtk dotnet` seems to already implement the filter.

Suggested change: widen the pattern to the subcommands the filter already supports, e.g. `^dotnet\s+(build|test|restore|format)\b`, with per-subcommand savings entries as done for `git`/`gh`.

## Two smaller questions

- `pgrep -f` / `pkill -f` (102 calls in 30 days here) aren't covered. Most uses are existence checks whose output could collapse to a count or one line. Worth a rule, or intentionally out of scope?
- `sed ` is on the ignore list in `rules.rs` alongside `awk`/`cut`/`tr`, which makes sense for stream editing — but `sed -n 'A,Bp' file` used as a range reader was 105 calls in my post-hook history and is really a `rtk read` case. Would you take a narrow rule for the `-n '<range>p' <file>` form only, or do you prefer to keep all of `sed` ignored?

## Environment

- rtk 0.43.0 (installed binary), Claude Code 2.1.220
- Hook: `PreToolUse` / matcher `Bash` / `rtk hook claude`
- Fedora Linux

Note: `main` reads `version = "0.42.4"`, older than the 0.43.0 binary I tested, so line references are approximate.

lntutor:
## Summary

- expand the dotnet discovery rule to build, test, restore, and format
- report the higher expected savings for test invocations
- add regression coverage for every supported dotnet subcommand

Closes #3300

## Testing

- `cargo test`

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1877,6 +1877,17 @@ mod tests {
     }
 
     #[test]
+    fn test_rewrite_dotnet_supported_subcommands() {
+        for subcommand in ["build", "test", "restore", "format"] {
+            assert_eq!(
+                rewrite_command_no_prefixes(&format!("dotnet {subcommand}"), &[]),
+                Some(format!("rtk dotnet {subcommand}")),
+                "dotnet {subcommand} should use the dotnet filter"
+            );
+        }
+    }
+
+    #[test]
     fn test_rewrite_playwright() {
         let commands = vec![
             "npm exec playwright",
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -687,11 +687,12 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^dotnet\s+build\b",
+        pattern: r"^dotnet\s+(build|test|restore|format)\b",
         rtk_cmd: "rtk dotnet",
         rewrite_prefixes: &["dotnet"],
         category: "Build",
         savings_pct: 70.0,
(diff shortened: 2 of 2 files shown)


lntutor:
Closing as a duplicate of #3309, which was opened first for the same issue and implementation. Sorry for the overlap.

PR #3322, opened 2026-07-30
docs(init): trim always-loaded awareness templates

### LINKED ISSUE #3306: RTK.md template carries one-time setup content and a dangling reference
# User Story

As a developer running `rtk init`, I want the generated `RTK.md` to contain
only content that's useful every session, so that the always-loaded context
`rtk init` installs into my Claude Code config doesn't carry one-time setup
instructions and dead references on every single session.

# Background

`rtk init` installs `~/.claude/RTK.md` from the template at
[`hooks/claude/rtk-awareness.md`](hooks/claude/rtk-awareness.md), and users
are expected to `@RTK.md`-import it into their global `CLAUDE.md` — meaning
its full contents are loaded into context on every session, not just once.

`claude doctor` surfaced two issues with this template:

1. **"Installation Verification" section is one-time content in an
   always-loaded file** (lines 14–22):

   ```
   ## Installation Verification

   ​```bash
   rtk --version         # Should show: rtk X.Y.Z
   rtk gain              # Should work (not "command not found")
   which rtk             # Verify correct binary
   ​```

   ⚠️ **Name collision**: If `rtk gain` fails, you may have
   reachingforthejack/rtk (Rust Type Kit) installed instead.
   ```

   This is useful exactly once — right after install, to confirm `rtk` is on
   `PATH` and isn't shadowed by the unrelated `reachingforthejack/rtk` (Rust
   Type Kit) package. It has no relevance to any session after that, but
   because it lives in the always-imported `RTK.md`, every session pays the
   token cost of loading it.

2. **Dangling reference to a nonexistent section** (line 29):

   ```
   Refer to CLAUDE.md for full command reference.
   ```

   The generated `CLAUDE.md` has no "full command reference" section for RTK
   — this line points at content that doesn't exist, so it's dead weight at
   best and misleading at worst.

Both were confirmed directly against the source template in this repo
(`hooks/claude/rtk-awareness.md`), copied via `src/hooks/init.rs` — this
isn't local drift in one user's file, it's present in what `rtk init` ships
today.

# Proposed Change

- Remove the "Installation Verification" section from
  `hooks/claude/rtk-awareness.md` (and any other RTK.md templates that carry
  it, e.g. `hooks/codex/`, if applicable).
  - Alternative worth considering instead of outright deletion: move this
    content to `INSTALL.md`, or print it as part of `rtk init`'s own command
    output, so the verification steps are still discoverable right after
    install without living in the always-loaded file. Leaving the exact
    approach to maintainer judgment.
- Remove the dangling "Refer to CLAUDE.md for full command reference." line
  (line 29), or replace it with an accurate pointer if such a reference
  section is intended to exist elsewhere.

# Acceptance Criteria

- [ ] `hooks/claude/rtk-awareness.md` no longer contains the "Installation
      Verification" section (or it has been relocated out of the
      always-loaded template, per maintainer preference).
- [ ] The dangling "Refer to CLAUDE.md for full command reference." line is
      removed or corrected.
- [ ] A fresh `rtk init` regenerates `~/.claude/RTK.md` without the removed
      section/line.
- [ ] No other docs (README variants, `hooks/*/README.md`, etc.) reference
      the removed section in a way that would also go stale.

lntutor:
Fixes #3306

Remove one-time installation checks from the Claude and Codex awareness templates, and drop the dangling Claude command-reference pointer. The generated files now contain only guidance that remains useful in every session.

Validation:
- Added a regression test for both embedded templates
- `cargo fmt --all --check`
- `cargo test` (2512 passed, 8 ignored)
- `git diff --check`

### DIFF
--- a/hooks/claude/rtk-awareness.md
+++ b/hooks/claude/rtk-awareness.md
@@ -11,19 +11,7 @@ rtk discover          # Analyze Claude Code history for missed opportunities
 rtk proxy <cmd>       # Execute raw command without filtering (for debugging)
 ```
 
-## Installation Verification
-
-```bash
-rtk --version         # Should show: rtk X.Y.Z
-rtk gain              # Should work (not "command not found")
-which rtk             # Verify correct binary
-```
-
-⚠️ **Name collision**: If `rtk gain` fails, you may have reachingforthejack/rtk (Rust Type Kit) installed instead.
-
 ## Hook-Based Usage
 
 All other commands are automatically rewritten by the Claude Code hook.
 Example: `git status` → `rtk git status` (transparent, 0 tokens overhead)
-
-Refer to CLAUDE.md for full command reference.
--- a/hooks/codex/rtk-awareness.md
+++ b/hooks/codex/rtk-awareness.md
@@ -22,11 +22,3 @@ rtk gain            # Token savings analytics
 rtk gain --history  # Recent command savings history
 rtk proxy <cmd>     # Run raw command without filtering
 ```
-
-## Verification
-
-```bash
-rtk --version
-rtk gain
-which rtk
-```
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -4872,6 +4872,15 @@ mod tests {
     }
 
     #[test]
+    fn test_awareness_templates_omit_one_time_setup() {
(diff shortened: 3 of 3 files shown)


lntutor:
Closing as a duplicate of #3308, which was opened first for the same issue and template cleanup. Sorry for the overlap.

Issue #3185, opened 2026-07-23
bug(windows): redirect character in quoted arguments creates stray empty files (quoting lost on re-execution)


Solvaray:
## Environment

- rtk 0.43.0 (standalone binary from GitHub releases, `rtk-x86_64-pc-windows-msvc.zip`)
- Windows 11 Home 10.0.26200
- Invoked via the Claude Code PreToolUse hook (`rtk hook claude`), shell is Git Bash
- Reproduced in a git repository working directory

## Summary

When a command is routed through rtk and an argument contains a `>` character
inside a properly quoted string (very common: `->` arrows in git commit
messages), rtk's internal re-execution loses the quoting. cmd.exe then
interprets the `>` as an output redirection and creates an empty (0-byte)
file in the working directory, named after the token following the arrow.

The proxied command itself still succeeds with the correct arguments (the
commit message is intact in `git log`), so the bug is silent — users just
find mystery files accumulating in their repos.

## Reproduction

```bash
git commit --allow-empty -m "probe commit -> strayprobeCOMMIT test"
```

Run through the rtk hook (rewritten to `rtk git commit ...`, confirmed via
`rtk gain --history`).

**Expected:** commit is created, nothing else happens.

**Actual:** commit is created correctly, **and** an empty file named
`strayprobeCOMMIT` appears in the working directory.

## Additional observations

- Windows filename rules shape the stray names: a token like `overfit.`
  becomes file `overfit` (trailing dot stripped); a token containing `:`
  such as `float:` becomes file `float` (the rest is treated as an NTFS
  alternate data stream).
- Real-world stray files we found in one project over a single day:
  `6.6%`, `overfit`, `float`, `{t['exit_date'].date()}` — all 0 bytes, all
  traceable to `->` sequences in commit messages or command content.
- Possibly related: piped commands can return wrong results through rtk on
  Windows — `ls -la .git/hooks/ | grep -v sample` returned empty output
  although the directory contains 14 sample files that should have been
  filtered line-by-line, not dropped entirely. That half looks like it is
  addressed by #3128 (pipe-rewrite safety); the quoting/redirect issue
  described here is explicitly *not* covered by that PR per its notes.

## Impact

Anyone using rtk as a Claude Code / agent hook on Windows gets junk files
silently created in their repositories whenever an AI-generated commit
message or command contains `->` — which is extremely common in generated
text ("A -> B", "18.3% -> 6.6%", Python return annotations, etc.).

[...]

PR #3324, opened 2026-07-30
fix(hooks): honour exclude_commands for head and tail

### LINKED ISSUE #2363: hooks.exclude_commands does not apply to head/tail rewrites

### LINKED ISSUE #2823: `[hooks] exclude_commands` is bypassed for `head`/`tail` (line-range fast path skips the exclusion check)

### LINKED ISSUE #3371: `[hooks] exclude_commands` silently ignored for `head`/`tail` rewrites (honored for `cat`)
## Environment

- Config: `~/Library/Application Support/rtk/config.toml`

## Summary

With `head` and `tail` listed in `[hooks] exclude_commands`, the hook engine still rewrites `head -N` / `tail -N` to `rtk read`. The same list entry works for `cat` (and for `ls`, `find`, `git diff`, `docker logs`). No warning is emitted — the binary contains `warning: invalid exclude_commands pattern` / `ignoring trivial exclude_commands pattern` strings, but nothing is printed for `head`/`tail`.

## Repro

```toml
# config.toml
[hooks]
exclude_commands = ["cat", "head", "tail"]
```

```console
$ rtk hook check 'cat x.txt'
No rewrite for: cat x.txt              # exclusion honored

rtk read big.log --max-lines 50        # exclusion ignored, no warning

$ rtk hook check 'tail -100 app.log'
rtk read app.log --tail-lines 100      # exclusion ignored, no warning
```

Also tried as list entries: `"read"`, `"rtk read"`, `"head *"`, `"head -50"` — none suppress the rewrite.

## Expected

Either `head`/`tail` in `exclude_commands` suppresses the rewrite (consistent with `cat`), or a startup/stderr warning states the pattern is unsupported. Silent partial application makes the config untrustworthy — the operator believes the exclusion is active when it is not.

## Why we needed this

We wanted to opt out of the `head` rewrite because of the `--max-lines` halving bug (filed separately) while keeping the rest of the hook active. Currently the only workaround is that the `head -n N` spelling happens not to be rewritten — which is itself an inconsistency (`head -N` rewritten, `head -n N` not; `tail -n N` rewritten).

katspa-sc:
Fixes #2823. Fixes #3371. Fixes #2363.

Partially addresses #3107 — this closes the `exclude_commands` half only. The memory half
(`rtk read` buffering the whole file) is unaffected and tracked separately.

## Problem

`head -N <file>` and `tail -N <file>` are rewritten to `rtk read`, but listing `head` or `tail` in `[hooks] exclude_commands` has no effect — they are still rewritten.

Root cause is ordering in `rewrite_segment_inner` (`src/discover/registry.rs`): the `head -` / `tail ` branch returns early, before the `is_excluded()` check in the `Classification::Supported` arm below it.

Reproduce with `exclude_commands = ["head", "tail"]`:

```
$ rtk hook check "head -20 src/main.rs"
rtk read src/main.rs --max-lines 20     # expected: no rewrite
```

This leaves no way to opt out of routing file contents through `rtk read` — relevant when you want a file read verbatim and not filtered.

## Fix

Applies the exclusion check inside that branch, mirroring how the `Supported` arm strips an env prefix before matching, so behaviour is consistent between the two paths.

## Tests

Two regression tests in `src/discover/registry.rs`:

- `test_head_tail_honour_exclude_commands` — `head`/`tail` suppressed when excluded, `git status` unaffected
- `test_head_tail_rewrite_when_not_excluded` — normal rewrite still happens when they are not in the list

## Pre-commit gate

`cargo fmt --all --check` and `cargo clippy --all-targets` are clean.

`cargo test` passes except 3 pre-existing failures in `hooks::rewrite_cmd::tests::unattestable_passthrough` (`test_plain_command_still_rewrites`, `test_fd_dup_redirect_still_rewrites`, `test_dollar_substitution_passthrough`). These fail identically on a clean checkout of `develop` with this branch stashed — they assert `evaluate("git status", &[], &[])` is `Ask`, but `check_command` consults the developer's real agent permission settings, so the outcome is `Allow` on a machine where `git status` is pre-allowed. Happy to file that separately.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -992,6 +992,12 @@ fn rewrite_segment_inner(
             if !routable {
                 return None;
             }
+            // The inner command may have been dropped because it is excluded.
+            // Re-testing the wrapped form would route it through the wrapper's
+            // own filter, defeating the exclusion.
+            if is_excluded(ENV_PREFIX.replace(rest, "").trim(), excluded) {
+                return None;
+            }
             break;
         }
     }
@@ -2280,6 +2286,27 @@ mod tests {
         assert_eq!(
             rewrite_command_no_prefixes("git status", &excluded),
             Some("rtk git status".into())
+        );
+    }
+
+    #[test]
+    fn test_routable_wrapper_honours_exclude_commands() {
+        // `uv run` is a routable wrapper: when the inner rewrite is dropped it
+        // falls through and re-tests `uv run <cmd>` as a `uv` invocation. That
+        // fall-through must not resurrect a command the user excluded.
+        let excluded = vec!["head".to_string(), "tail".to_string()];
+        assert_eq!(
+            rewrite_command_no_prefixes("uv run head -20 src/main.rs", &excluded),
+            None
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("uv run cat src/main.rs", &["cat".to_string()]),
+            None
+        );
+        // A non-excluded inner command still rewrites through the wrapper.
+        assert_eq!(
+            rewrite_command_no_prefixes("uv run head -20 src/main.rs", &["cat".to_string()]),
+            Some("uv run rtk read src/main.rs --max-lines 20".into())
         );
     }
 
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1020,6 +1020,12 @@ fn rewrite_segment_inner(
     if context == RewriteContext::Normal
         && (cmd_part.starts_with("head -") || cmd_part.starts_with("tail "))
     {
+        // head/tail rewrite to `rtk read`, so honour exclude_commands here too —
+        // this branch returns before the checks below and used to ignore the list.
+        let stripped = ENV_PREFIX.replace(cmd_part, "");
+        if is_excluded(stripped.trim(), excluded) {
+            return None;
+        }
         return rewrite_line_range(cmd_part).map(|r| format!("{}{}", r, redirect_suffix));
     }
 
@@ -2247,6 +2253,33 @@ mod tests {
     }
 
     // --- P0.2: head -N rewrite ---
+
+    #[test]
+    fn test_head_tail_honour_exclude_commands() {
+        // head/tail rewrite to `rtk read`; excluding them must suppress that.
+        let excluded = vec!["head".to_string(), "tail".to_string()];
+        assert_eq!(
+            rewrite_command_no_prefixes("head -20 src/main.rs", &excluded),
+            None
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("tail -20 src/main.rs", &excluded),
+            None
+        );
+        // ...and must not affect unrelated commands.
+        assert_eq!(
+            rewrite_command_no_prefixes("git status", &excluded),
+            Some("rtk git status".into())
+        );
+    }
+
+    #[test]
+    fn test_head_tail_rewrite_when_not_excluded() {
+        assert_eq!(
+            rewrite_command_no_prefixes("head -20 src/main.rs", &["cat".to_string()]),
+            Some("rtk read src/main.rs --max-lines 20".into())
+        );
+    }
 
     #[test]
     fn test_rewrite_head_numeric_flag() {
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1022,8 +1022,8 @@ fn rewrite_segment_inner(
     {
         // head/tail rewrite to `rtk read`, so honour exclude_commands here too —
         // this branch returns before the checks below and used to ignore the list.
-        let stripped = ENV_PREFIX.replace(cmd_part, "");
-        if is_excluded(stripped.trim(), excluded) {
+        // Any sudo/env prefix has already been peeled by strip_disabled_prefix above.
+        if is_excluded(cmd_part, excluded) {
             return None;
         }
         return rewrite_line_range(cmd_part).map(|r| format!("{}{}", r, redirect_suffix));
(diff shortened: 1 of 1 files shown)


KuSh:
Picked this one out of the seven PRs open against the `head`/`tail` fast path — it's the one that strips the env prefix before matching, so `sudo head -20 f` and `FOO=1 head -20 f` are covered, and mirroring the `Supported` arm is the right call given `head`/`tail` resolve to `rtk read`. I've added the `Fixes` lines to the description; the others are closed pointing here.

One thing to add before this lands: also match the path-stripped form, the way the `Unsupported` arm does with `strip_absolute_path`. Right now `/usr/bin/head -20 f` still rewrites past the exclusion. #3719 had that piece and nothing else did, so it's worth carrying over — roughly:

```rust
if is_excluded(cmd_clean, excluded)
    || is_excluded(&strip_absolute_path(cmd_clean), excluded)
{
    return None;
}
```

plus a test for the path-prefixed form.

Worth knowing about the fast path while you're in there: it's guarded on `context == RewriteContext::Normal`, so inside a pipeline `head -20` already falls through to `classify_command` and gets exclusion-checked correctly today. Same command, opposite outcomes:

```
["head","tail","read"]   head -20 f.txt        →  rtk read f.txt --max-lines 20
                         cat f.txt | head -20  →  EXCLUDED
```

So this is removing a special case rather than adding one, which is a good sign for the approach.

On the three failures you flagged in `hooks::rewrite_cmd::tests::unattestable_passthrough` — your diagnosis is right, and it's a known separate problem: those tests read the developer's real agent permission settings. No need to file it, #3638 is already open against it. Nothing for you to do here.

bharvey42:
Confirming this is still needed as of `develop` @ aa40853 — worth stating explicitly, because #3749 landed a few hours ago and *looks* like it might have subsumed this one. It doesn't. The fast path still returns before any exclusion check:

- `src/discover/registry.rs:1364-1368` — the `head -` / `tail ` branch returns `rewrite_line_range(...)` directly
- `src/discover/registry.rs:1385-1387` — the `is_excluded` check, in the `Classification::Supported` arm *below* it

So #3371 is live on mainline today, and this PR is still the fix.

One thing that may have shifted under the review comment above, since #3749 merged after it was written: the `Supported` arm this PR mirrors no longer reads the way it did on 2026-08-30. It is now

```rust
if !excluded.is_empty()
    && (is_excluded(cmd_clean, excluded)
        || is_excluded(&tool_form(cmd_clean, rtk_equivalent), excluded))
{
    return None;
}
```

and `tool_form` (`registry.rs:1516-1520`) already applies `strip_absolute_path` internally. If that reads right to you, mirroring the arm as it now stands would cover the `/usr/bin/head -20 f` case that was asked for, without hand-rolling a second `strip_absolute_path` call.

It looks like it would also pick up something the review's own example assumed — that `read` in `exclude_commands` suppresses `head`/`tail`, since `rules.rs:105-108` maps `cat|head|tail` to `rtk_cmd: "rtk read"` and `tool_form` normalizes to that. `cat` gets that today via the classify path; `head`/`tail` don't, because of the same early return. I haven't run that case, so treat it as a thing to confirm rather than a claim — but it seems to fall out of the mirror for free rather than needing extra code.

User-side context: I've been carrying the equivalent fix (from the now-closed #3378) in a local fork build since 2026-08-08 on Windows 11. `exclude_commands` is load-bearing for me — I exclude `grep`/`rg` there to work around #3185/#3161 — so an exclusion that silently doesn't apply is a sharp edge. Would be glad to see this land.

KuSh (approved):
Reviewed at high effort. No correctness bugs — the fix is minimal, correctly placed before `rewrite_line_range`, and consistent with the `Supported`/`Unsupported` arms.

**Verification**
- All rewrite entry points (`src/hooks/rewrite_cmd.rs`, `src/hooks/hook_cmd.rs`, `src/main.rs`) funnel through `registry::rewrite_command`, so the single-site fix covers every hook/agent path. No other `head`/`tail` rewrite site exists.
- PR head: `cargo fmt --check`, `cargo clippy --all-targets` and `cargo test --all` all clean.
- Merged latest `develop`: no conflicts, full suite green.
- Probes with `exclude_commands = ["head", "tail"]`: `head -20 f`, `tail -20 f`, `tail -n 20 f`, `sudo head -20 f`, `RUST_LOG=x head -20 f`, `head -20 f 2>&1`, `nocorrect head -20 f` all pass through. Controls hold: `cat f` still rewrites, and `head -20 f` still rewrites when only `cat` is excluded. Pipelines unchanged.
- The `Fixes` claims check out — #2823, #3371 and #2363 are three duplicates of this bug, and #3107's `exclude_commands` half is genuinely closed.

**Two follow-up commits pushed to this branch**

1. `refactor: drop the no-op env strip in the head/tail exclude check` — `ENV_PREFIX` is anchored on `sudo`/`env`/`VAR=val`, and the branch is only reached once `cmd_part` starts with `head -` or `tail `, so the `replace` could never fire. `strip_disabled_prefix` already peels those prefixes at the top of `rewrite_segment_inner` and recurses. Matching `cmd_part` directly is equivalent; added tests for `sudo head …` and `RUST_LOG=… tail …` so the recursion is what carries the exclusion.

2. `fix: keep exclude_commands honoured under routable wrappers` — the exclusion was still bypassed under `uv run`. Because `uv run` is a routable wrapper, dropping the inner rewrite fell through and re-tested `uv run head -20 f.rs` as a `uv` invocation, yielding `rtk uv run head -20 f.rs`; the file contents went through an RTK filter despite the exclusion. Pre-existing (excluded `cat` behaved the same way), but it left a hole in exactly what this PR promises, so I closed it here rather than filing a follow-up. Stopping the fall-through when the wrapped command is excluded fixes both.

Both are separate commits, so drop either if you'd rather keep the scope tight. Full suite green on the branch and after merging `develop`.

Thanks for the clean fix and the two-directional tests — nice catch on a bug that had three open duplicates.

PR #3329, opened 2026-07-30
fix(discover): preserve command builtin bypass

### LINKED ISSUE #3230: Hook rewrites the `command` bypass prefix into `command rtk …`, inverting the one prefix used to escape wrappers
## Summary

The rewriter treats the POSIX `command` builtin as a transparent wrapper prefix, so `command grep …` becomes `command rtk grep …`. `command` is the one prefix whose entire purpose is to *bypass* shell functions and aliases and run the real binary, so rewriting it inverts the user's explicit intent: the escape hatch routes back into the thing it was used to escape.

## Affected version

`rtk 0.43.0` (Homebrew, macOS arm64). Source inspected at `develop` `1d6798d`.

## Root cause

`src/discover/registry.rs:829`:

```rust
/// Shell keywords that wrap a command without changing which one runs. They are
/// not spawnable, so they must never fall through: `rtk exec foo` cannot run.
const SHELL_KEYWORD_PREFIXES: &[&str] = &["noglob", "command", "builtin", "exec", "nocorrect"];
```

The doc comment is the bug. For `noglob`, `nocorrect`, `builtin` and `exec` it holds — they change *how* a command runs. `command` is different: per POSIX it changes *which* executable runs, by suppressing function and alias lookup. It belongs with `sudo` (#2810) in the "pass through untouched" category, not here.

## Reproduction

```console
$ rtk rewrite 'command grep -rn X a'
command rtk grep -rn X a
$ rtk rewrite 'command ls -la'
command rtk ls -la
$ rtk rewrite 'command git status'
command rtk git status
```

Observed end to end, on a file with 80 matching lines:

```console
$ command grep -rn NEEDLE many.txt | tail -3
/.../many.txt:24:line 24 NEEDLE filler
/.../many.txt:25:line 25 NEEDLE filler
  +55 more in /.../many.txt [see remaining: tail -n +26 ~/…/rtk/tee/…log]
```

`command grep` returns rtk's capped, path-elided output. The user asked for the real binary and got the filter. `\grep` and an absolute `/usr/bin/grep` are both correctly left alone, so the inconsistency is specific to `command`.

`env` shows the same shape (`env grep -rn X a` → `env rtk grep -rn X a`) though it is not in the list above; I have not traced which path produces that one.

## It also defeats rtk's own pipeline carve-out

`find` is deliberately not rewritten when it feeds a pipe, because rtk's tree output is incompatible with `xargs` (#439, restated in #2962). Adding `command` bypasses that protection and breaks a working pipeline:

```console
$ rtk rewrite 'find . -name "*.ts" -print0 | xargs -0 tool'
(not rewritten)                      # carve-out works

$ rtk rewrite 'command find . -print0 | xargs -0 tool'
command rtk find . -print0 | xargs -0 tool
```

Run for real, the second form fails outright:

```
rtk: rtk find does not support compound predicates or actions (e.g. -not, -exec). Use `find` directly.
```

So a user who adds `command` — the conventional way to signal "I want the real tool, not a wrapper" — turns a working command into a broken one.

## Why this matters for AI-agent automation

Under Claude Code, `grep`, `find` and `rg` are *already* shell functions: the session's shell snapshot defines `function grep { … ARGV0=ugrep … }`, pointing at binaries embedded in the CLI. In that environment `command grep` is the standard, widely-documented way to reach the real `grep` — it is what an agent reaches for the moment it suspects its search results are being filtered.

That makes this failure self-concealing in a specific way: the agent forms the correct hypothesis ("something is transforming my output"), applies the correct remedy, gets output that is still transformed, and concludes the hypothesis was wrong. In our case that closed off the truncation explanation for two full sessions and sent us chasing a non-existent pattern-mangling bug instead. `command X` is a bypass primitive; when a tool redefines it to mean its opposite, the usual way of establishing ground truth stops working.

## Expected vs actual

| | |
|---|---|
| **Expected** | `command foo …` passes through unrewritten, like `sudo` (#2810) and `\foo`. |
| **Actual** | Rewritten to `command rtk foo …`, so the bypass prefix routes back into rtk. |

## Suggested fix

Drop `"command"` from `SHELL_KEYWORD_PREFIXES` and skip rewriting for any segment starting with `command `, matching the `sudo` treatment. The "not spawnable, must never fall through" concern in that comment does not apply — the fix is to leave the whole segment alone, not to route it.

Worth deciding the same question for `env`, which behaves identically in testing: `env VAR=1 foo` is a legitimate transparent wrapper, but bare `env foo` is often used as a bypass too.

lntutor:
## Summary

- leave commands prefixed with the POSIX `command` builtin unrewritten
- preserve bypass semantics within compound commands and pipelines
- add regression coverage for grep, git, find/xargs, assignments, and mixed compound segments

Closes #3230

## Test plan

- [x] Manual testing: `rtk rewrite "command grep -rn X src"` produced no rewrite (empty output, exit 1)

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -830,7 +830,7 @@ const ROUTABLE_WRAPPER_PREFIXES: &[&str] = &["uv run"];
 
 /// Shell keywords that wrap a command without changing which one runs. They are
 /// not spawnable, so they must never fall through: `rtk exec foo` cannot run.
-const SHELL_KEYWORD_PREFIXES: &[&str] = &["noglob", "command", "builtin", "exec", "nocorrect"];
+const SHELL_KEYWORD_PREFIXES: &[&str] = &["noglob", "builtin", "exec", "nocorrect"];
 
 /// Every built-in transparent wrapper, paired with whether it may fall through.
 /// Derived from the two lists above so they cannot drift apart.
@@ -975,6 +975,12 @@ fn rewrite_segment_inner(
             depth + 1,
         )?;
         return Some(format!("{}{}", env_prefix, rewritten));
+    }
+
+    // The POSIX `command` builtin explicitly bypasses shell functions and aliases.
+    // Rewriting its operand would invert that escape-hatch behavior.
+    if strip_word_prefix(trimmed, "command").is_some() {
+        return None;
     }
 
     for (prefix, routable) in builtin_transparent_prefixes() {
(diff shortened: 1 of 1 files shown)


lntutor:
Closing as a duplicate of #3246, which was opened first for the same issue and preserves the same command-bypass behavior. Sorry for the overlap.

PR #3330, opened 2026-07-30
fix(opencode): use cross-platform process API

### LINKED ISSUE #3326: `rtk init --opencode` plugin silently does nothing on Desktop (`$ is not a function`) and on Windows (`which` not found)
## Environment

- Windows 10
- opencode Desktop 1.18.9

## What happens

The `rtk init --opencode` plugin uses Bun's `$` tagged-template helper and `which rtk` as an availability probe.

- **opencode Desktop** (Electron) does not expose Bun's `$` → `TypeError: $ is not a function` → plugin silently does nothing
- **Windows** (any runtime) → `which` is not available → availability check throws → plugin disabled

The upstream issue on Opencode (anomalyco/opencode#25878) was closed as "not planned" — Desktop will never receive `$`.

## Steps to reproduce

1. Run `rtk init --opencode` to generate the plugin
2. Open Opencode Desktop
3. Ask the agent to run any shell command, like "run git status"
4. Plugin does nothing — commands are not rewritten (OK on Opencode CLI)

## Expected behavior

Plugin rewrites commands on all runtimes (CLI + Desktop) and all platforms (Windows, macOS, Linux).

## Actual behavior

- Desktop: `TypeError: $ is not a function`
- Windows CLI: availability check fails (`which` not found), plugin disables itself

## Fix

Replace `$` and `which` with `node:child_process.execFile`, which works on Bun, Node.js/Electron, Windows, macOS, and Linux:

```typescript
import type { Plugin } from "@opencode-ai/plugin"
import { execFile } from "node:child_process"

function runRtk(args: string[]): Promise<string> {
    execFile("rtk", args, { windowsHide: true }, (_error, stdout) => {
      // `rtk rewrite` can exit non-zero even on a successful rewrite
      // (observed: exit code 3) — stdout is what matters, not the exit code.
      resolve(String(stdout ?? "").trim())
    })
  })
}

export const RtkOpenCodePlugin: Plugin = async () => {
  const version = await runRtk(["--version"])
  if (!version) {
    console.warn("[rtk] rtk binary not found in PATH — plugin disabled")
    return {}
  }

  return {
    "tool.execute.before": async (input, output) => {
      const tool = String(input?.tool ?? "").toLowerCase()
      const args = output?.args
      if (!args || typeof args !== "object") return

      const command = (args as Record<string, unknown>).command

      const rewritten = await runRtk(["rewrite", command])
        ;(args as Record<string, unknown>).command = rewritten
      }
    },
  }
}

export default RtkOpenCodePlugin
```

## Notes

- `windowsHide: true` is silently ignored on macOS/Linux — no side effect
- `execFile("rtk")` resolves `rtk.exe` on Windows via `CreateProcess` — no shell needed
- `rtk --version` as probe is cross-platform (replaces Unix-only `which`)

## Tested on

- Windows 10 — opencode Desktop ✅ — opencode CLI ✅ 1.18.9

lntutor:
## Summary

- replace the Bun-only `$` helper and Unix-only `which` probe with Node `execFile`
- pass rewrite commands as literal subprocess arguments without invoking a shell
- keep successful rewrite stdout even when the rewrite protocol exits nonzero
- update OpenCode integration docs and add embedded-plugin regression coverage

Closes #3326

## Test plan

- [x] `cargo test`
- [x] Bundle the TypeScript plugin with Bun and invoke it against a fake `rtk` binary, verifying literal command arguments and nonzero-exit stdout handling

### DIFF
--- a/docs/contributing/TECHNICAL.md
+++ b/docs/contributing/TECHNICAL.md
@@ -319,7 +319,7 @@ Start here, then drill down into each README for file-level details.
 | [`cline/`](../hooks/cline/README.md) | Cline / Roo Code | Rules file (prompt-level, no programmatic hook) |
 | [`windsurf/`](../hooks/windsurf/README.md) | Windsurf / Cascade | Rules file (workspace-scoped) |
 | [`codex/`](../hooks/codex/README.md) | OpenAI Codex CLI | Awareness document, AGENTS.md integration |
-| [`opencode/`](../hooks/opencode/README.md) | OpenCode | TypeScript plugin, zx library, in-place mutation |
+| [`opencode/`](../hooks/opencode/README.md) | OpenCode | TypeScript plugin, Node `execFile`, in-place mutation |
 
 ---
 
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -39,7 +39,7 @@ Each agent subdirectory has its own README with hook-specific details:
 - **[`cline/`](cline/README.md)** — Rules file (prompt-level), `.clinerules` project-local installation
 - **[`windsurf/`](windsurf/README.md)** — Rules file (prompt-level), `.windsurfrules` workspace-scoped
 - **[`codex/`](codex/README.md)** — Awareness document, `AGENTS.md` integration, `$CODEX_HOME` or `~/.codex/` location
-- **[`opencode/`](opencode/README.md)** — TypeScript plugin, `zx` library, `tool.execute.before` event, in-place mutation
+- **[`opencode/`](opencode/README.md)** — TypeScript plugin, Node `execFile`, `tool.execute.before` event, in-place mutation
 - **[`pi/`](pi/README.md)** — TypeScript extension, `tool_call` event, `isToolCallEventType` guard, in-place mutation, `~/.pi/agent/extensions/`
 - **[`hermes/`](hermes/README.md)** — Python plugin, `pre_tool_call` hook, in-place terminal command mutation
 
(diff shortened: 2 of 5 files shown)


amnesiaof:
### Independent confirmation (Windows 11, OpenCode CLI, rtk 0.47.0)

I reproduced the silent failure from #3326 on OpenCode **CLI** on Windows — two stacked bugs:

1. The plugin context provides **no Bun `$` shell**, so `` $`which rtk` `` throws `TypeError: $ is not a function`, the plugin factory crashes silently, and no hooks are registered — nothing surfaced to the user.
2. The `which` probe is Unix-only and doesn't exist on Windows.

After working around both, I hit a third subtlety that this PR already handles correctly: **`rtk rewrite` uses non-zero exit codes as status signals** (exit 3 with valid stdout such as `rtk git status`), so a naive try/catch treats successful rewrites as failures. The stdout must be recovered from the exception object.

I independently arrived at the same approach as this PR (Node `child_process`, literal arguments, no shell, stdout-from-exception, `windowsHide`) and it works reliably in daily use: hooks fire, commands get rewritten, `rtk gain` tracks savings.

My working plugin, matching this PR's approach:

```ts
import { execFileSync } from "node:child_process"
import type { Plugin } from "@opencode-ai/plugin"

const run = (args: string[], timeout = 5000): string | null => {
  try {
    return execFileSync("rtk", args, { encoding: "utf8", timeout, windowsHide: true })
  } catch (e) {
    // rtk uses non-zero exit codes as status signals; stdout may still be valid
    const out = (e as { stdout?: string | Buffer }).stdout
    return typeof out === "string" || Buffer.isBuffer(out) ? String(out) : null
  }
}

export const RtkOpenCodePlugin: Plugin = async () => {
  if (run(["--version"]) === null) return {}

  return {
    "tool.execute.before": async (input, output) => {
      const tool = String(input?.tool ?? "").toLowerCase()
      const args = output?.args
      if (!args || typeof args !== "object") return

      const command = (args as Record<string, unknown>).command

      const rewritten = run(["rewrite", command])
      if (rewritten === null) return
      const trimmed = rewritten.trim()
      if (trimmed && trimmed !== command) {
        ;(args as Record<string, unknown>).command = trimmed
      }
    },
  }
}
```

Notes:

- Probing `rtk --version` directly (as suggested in #2426) also removes the `which` dependency entirely — worth considering here.
- Related: #3742 reports the plugin module never loading on Windows CLI; merging a cross-platform template should shrink that class of reports.

Would be great to see this merged so `rtk init --opencode` works out of the box on Windows.

amnesiaof:
I opened #3837 targeting current `develop` — same root causes and fix as this PR, with the exit-code-3 stdout handling and a regression test. This comment is just to link the two; happy to have either merged, whichever is easier for maintainers.

PR #3333, opened 2026-07-30
fix(core): cap exec_capture buffer to prevent OOM on unbounded linter output

FlorianBruniaux:
## Problem

`pnpm lint` gets hook-rewritten to `rtk lint` (see the classification
table in `src/discover/registry.rs`). Without arguments, `rtk lint`
defaults to eslint and scans `.` (the whole repo) with `-f json`
(`src/cmds/js/lint_cmd.rs::run`). On a large repo with many violations,
that JSON payload can reach hundreds of MB.

It gets captured via `core::stream::exec_capture()`, which used
`Command::output()` with no size limit, unlike `run_streaming()`, which
already caps captured output at `RAW_CAP` (10 MiB). Result: the rtk
process itself can OOM before any filter runs.

`exec_capture` is shared by 15 other filter modules (golangci, pip,
prisma, vitest, playwright, dotnet, git, gt, container, ...). This wasn't
lint-specific, it was a gap in the shared capture primitive.

## Fix

`exec_capture`/`exec_capture_stdin` now read each pipe (stdout/stderr)
capped at `RAW_CAP` (10 MiB), still draining (and discarding) anything
past the cap so the child process never blocks writing to a full pipe.
Behavior is unchanged below 10 MiB; above it, output is bounded instead
of crashing the process.

## Testing

- `cargo fmt --all && cargo clippy --all-targets`: clean
- `cargo test --all`: full suite green except 2 pre-existing, unrelated
  failures in `hooks::rewrite_cmd` (permission-config-dependent, verified
  to fail identically on `develop` without this patch)
- New regression test `test_exec_capture_raw_cap_at_10mb`: proves the cap
  holds and the child process exits cleanly (no pipe deadlock)

## Test plan

- [ ] CI green (fmt, clippy, test)
- [ ] Confirm the 2 pre-existing `rewrite_cmd` failures are also present on
      `develop` HEAD without this change

### DIFF
--- a/src/core/stream.rs
+++ b/src/core/stream.rs
@@ -1,5 +1,5 @@
 use anyhow::{Context, Result};
-use std::io::{self, BufRead, BufReader, BufWriter, Write};
+use std::io::{self, BufRead, BufReader, BufWriter, Read, Write};
 use std::process::{Command, Stdio};
 use std::sync::mpsc;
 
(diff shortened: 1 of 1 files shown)


bharvey42:
Carried in a local fork build since 2026-08-08 (Windows 11, `x86_64-pc-windows-gnu`, daily use) with no regressions observed across the filter modules that share `exec_capture`.

Heads-up for the author: as of `develop` @ aa40853 this branch no longer merges cleanly — `git merge-tree` reports conflicts against current mainline, so it probably needs a rebase before a reviewer can act on it. Three PRs merged into `develop` today (#3798, #3749, #3470).

PR #3340, opened 2026-07-31
test(benchmark): use stable curl JSON fixture

lntutor:
## Summary

- replace the rotating `https://mockhttp.org/json/1` benchmark input with the repository `oc_pods.json` fixture
- keep raw and RTK curl commands pointed at the same deterministic payload
- avoid false token-growth failures caused by two requests receiving unrelated JSON documents

## Why

`mockhttp.org/json/1` rotates among different response bodies. The benchmark performs separate raw and RTK requests, so it can compare different documents and report apparent token growth even when RTK behavior is unchanged. This is reproducible on `develop`.

Using a local file URL makes the comparison deterministic and requires no runtime code changes.

## Validation

- `bash -n scripts/benchmark.sh`
- `git diff --check`
- release build
- curl benchmark row: `2523 -> 2523 (+0%)`

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -12,6 +12,7 @@ else
 fi
 BENCH_DIR="$(pwd)/scripts/benchmark"
 RTK_ROOT="$(pwd)"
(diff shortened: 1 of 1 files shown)


KuSh:

Issue #3339, opened 2026-07-31
Inflated savings metric, unreachable truncation log on Windows, and passthrough that doesn't pass through


juanre7:
## Environment

- rtk 0.43.0
- Windows 11 (10.0.26200), Git Bash / MINGW64, bash 5.2.37
- Claude Code, hook: `PreToolUse` → `matcher: "Bash"` → `rtk hook claude`

Filed together because all three surfaced in one session and are related to
how failures are reported. Happy to split if you prefer.

---

## 1. `rtk gain` counts failed parses as savings

Two `playwright test` runs failed to parse and returned nothing usable:

```
[RTK:PASSTHROUGH] playwright parser: All parsing tiers failed
[full output: ~/AppData\Local\rtk\tee\1785452444_playwright.log]
```

Both appear in `rtk gain` as the highest-efficiency entries of the session:

```
 9.  rtk playwright test -...      1    221   92.9%    1.9s
10.  rtk playwright test -...      1    221   92.9%    2.4s
```

Output that was never delivered is not savings — it is a command that had to
be re-run, costing more than it saved. Counting discarded output inflates the
headline number precisely when the tool is failing.

Suggestion: exclude runs whose parser failed from the savings aggregate, or
report them in a separate "failed" bucket.

## 2. Truncation log path is unreachable on Windows

When output is truncated, rtk points at the full log:

```
+36 more in src/App.tsx [see remaining: tail -n +26 ~/AppData\Local\rtk\tee\1785452647_grep_0_src_App_tsx.log]
```

Two problems:

- The `~` does not resolve to the Windows user profile under Git Bash. It
  expands to `C:/SPB_Data/AppData/...` and the file is not found:
  ```
  cat: C:/SPB_Data/AppData/Local/rtk/tee/....log: El sistema no puede encontrar la ruta especificada. (os error 3)
  ```
- The path mixes separators: `~/AppData\Local\rtk\tee\...` (forward slash then
  backslashes).

Net effect: the escape hatch to the full output cannot be opened, so
truncation becomes unrecoverable in practice.

Suggestion: emit an absolute path with consistent separators for the shell in
use, or print `$LOCALAPPDATA`-based paths on Windows.

## 3. `[RTK:PASSTHROUGH]` does not pass anything through

When the parser fails, the label says PASSTHROUGH but the caller receives no
command output — only the notice and the (unreachable, see #2) log path. If
parsing fails, the useful behaviour is to emit the raw output unmodified.

[...]

PR #3344, opened 2026-07-31
fix: parse-failure tracking, Windows tee path, passthrough output loss (#3339)

SomSamantray:
Fixes the three related bugs reported in #3339, all rooted in how rtk's Tier-3 "parsing failed" fallback (`ParseResult::Passthrough`) is tracked and displayed.

## 1. `rtk gain` counted parse failures as savings

`timer.track(...)` was called unconditionally, even for Passthrough-tier runs. Since the passthrough content is a truncated view of the raw output, the size difference between "raw" and "shown" was counted as savings — even though nothing was parsed or summarized, and truncation only happened because parsing failed.

Fix: route genuine Tier-3 failures through the existing `record_parse_failure_silent` + `TimedExecution::track_passthrough` plumbing (already used by `rtk gain --failures`) instead of the generic savings timer. Applied to `playwright test`, `pnpm list`, `pnpm outdated`, and `vitest`/`jest run` — the four call sites sharing this pattern. Explicit user-requested passthrough (e.g. vitest's `--reporter` override) is untouched — it isn't a failure.

## 2. Tee hint path broken on Windows

`[full output: ~/path/to/log]` mixed a forward-slash `~` with native backslash separators, and could resolve `~` against a different home directory than the invoking Git Bash/MSYS shell's own `~` expansion uses.

Fix, in `display_path`:
- Normalize the home-relative portion to forward slashes unconditionally (previously only the whitespace-quoting branch did this).
- Prefer `$HOME` over the OS home-directory API when stripping the home prefix, converting Git Bash/MSYS's POSIX-style `HOME` (`/c/Users/name`) to a native path first — otherwise it would never match via `strip_prefix` and would silently defeat the `~` shorthand.
- Reject an empty `HOME` value (an empty `PathBuf`'s `strip_prefix` trivially matches everything).

This keeps the existing `~/` shorthand (per #2325's direction) rather than reverting to always-absolute paths, an approach previously tried in #2187/#2701.

## 3. Passthrough could silently emit nothing

When parsing failed and a command's real output was empty on stdout (diagnostic on stderr instead), the passthrough fallback was itself empty. `never_worse` then preferred that empty result over the real combined output, discarding it.

Fix: build the passthrough fallback content from the combined stdout+stderr instead of stdout alone, for all four call sites above.

## Notes

- No local rtk checkout existed when this was written — verified against a fresh `develop` clone before implementing.
- `cargo fmt --all`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --all` all pass (2523 tests; the only 2 failures in this environment — `hooks::rewrite_cmd::tests::unattestable_passthrough::*` — are pre-existing and unrelated, reproduced identically on unmodified `develop`).
- `pnpm list`'s Tier-1/Degraded guard and tracking stay anchored to stdout alone (unchanged) — only the Tier-3 arm uses the combined stream, so a stderr notice alongside a successful parse doesn't inflate the reported savings ratio.

### DIFF
--- a/src/cmds/js/playwright_cmd.rs
+++ b/src/cmds/js/playwright_cmd.rs
@@ -288,6 +288,7 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     // Parse output using PlaywrightParser
     let parse_result = PlaywrightParser::parse(&result.stdout);
     let mode = FormatMode::from_verbosity(verbose);
+    let tier = parse_result.tier();
 
     let filtered = match parse_result {
         ParseResult::Full(data) => {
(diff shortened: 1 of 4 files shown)


bharvey42:
Carried in a local fork build since 2026-08-08 (Windows 11, Git Bash / `x86_64-pc-windows-gnu`, daily use), no regressions.

The tee-hint half fixes a real day-to-day annoyance on Windows: the `[full output: ~/...]` hint mixed a forward-slash `~` with native backslash separators, and could resolve `~` against a different home directory than the invoking MSYS shell's own `~` expansion uses — so the path it printed wasn't reliably pasteable back into the same shell. Preferring `HOME` over the OS home-directory API is the right call there; that is the root Git Bash's own expansion resolves against.

Still merges cleanly with `develop` as of today (aa40853).

PR #3347, opened 2026-07-31
fix(init): install Pi extension under .omp so OMP discovers it

manateeJr:
## Problem

`rtk init --agent pi` installs the TypeScript extension to `.pi/extensions/rtk.ts` (project) / `~/.pi/agent/extensions/rtk.ts` (global). The current Oh My Pi (OMP) runtime auto-discovers extensions **only** from `<cwd>/.omp/extensions` and `~/.omp/agent/extensions`; `.pi/extensions` is not a discovery root. The extension is therefore written but never loaded, so no commands are rewritten and `rtk gain` shows no OMP activity.

## Root cause

The install directory literals — not the extension code. OMP still honors the `PI_CODING_AGENT_DIR` env var, and its default agent dir is `~/.omp/agent`, so pointing the paths at `.omp` aligns install, uninstall, and OMP's discovery.

## Fix

- `src/hooks/constants.rs`: `PI_DIR` `.pi/agent` → `.omp/agent`, `PI_LOCAL_DIR` `.pi` → `.omp`
- `src/hooks/init.rs`: matching doc comments + test assertions
- docs: `hooks/README.md`, `hooks/pi/README.md`, `src/hooks/README.md`, `docs/guide/getting-started/supported-agents.md`

The extension file (`hooks/pi/rtk.ts`), the `--agent pi` flag, the `PI_*` constant names, and `PI_CODING_AGENT_DIR` are unchanged. Paths-only, minimal.

## Verification

Verified end-to-end against a live `omp v17.0.6`:

- With the extension at `~/.omp/agent/extensions/rtk.ts`, OMP loads it on startup via native discovery and its `tool_call` handler rewrites bash commands through `rtk rewrite`.
- Example: `ls -la /tmp/…` was rewritten to `rtk ls -la /tmp/…`, executed, and logged in rtk's history at 94.4% savings — `rtk gain` now reflects OMP usage.

Gate: `cargo fmt --all --check`, `cargo clippy --all-targets`, and full `cargo test` all pass (the `test_run_pi_mode_*` / `test_pi_*` tests now assert `.omp`).

### DIFF
--- a/docs/guide/getting-started/supported-agents.md
+++ b/docs/guide/getting-started/supported-agents.md
@@ -109,7 +109,7 @@ rtk init --agent pi
 rtk init --agent pi --global
 ```
 
-Creates `.pi/extensions/rtk.ts` (local) or `~/.pi/agent/extensions/rtk.ts` (global). Pi auto-discovers extensions from both paths on startup.
+Creates `.omp/extensions/rtk.ts` (local) or `~/.omp/agent/extensions/rtk.ts` (global). OMP auto-discovers extensions from both paths on startup.
 
 Uninstall:
 
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -40,7 +40,7 @@ Each agent subdirectory has its own README with hook-specific details:
 - **[`windsurf/`](windsurf/README.md)** — Rules file (prompt-level), `.windsurfrules` workspace-scoped
 - **[`codex/`](codex/README.md)** — Awareness document, `AGENTS.md` integration, `$CODEX_HOME` or `~/.codex/` location
 - **[`opencode/`](opencode/README.md)** — TypeScript plugin, `zx` library, `tool.execute.before` event, in-place mutation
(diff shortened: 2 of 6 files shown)


akhy:
This PR seems replacing the extension to OMP completely instead of allowing it to adapt to either PI or OMP runtime

KuSh:
Thanks for digging into where OMP actually looks for extensions. Develop now installs to those paths through a dedicated OMP agent added in #3707 (merged 2026-09-05): rtk init --agent omp writes .omp/extensions/rtk.ts locally or ~/.omp/agent/extensions/rtk.ts globally, sharing the same rtk.ts file with Pi and tracking ownership in a .rtk-agents sidecar. The .pi paths were kept on purpose so existing Pi installs keep working, which is the one part of this PR that was not taken. If you still hit a case where OMP cannot find the extension after rtk init --agent omp, please comment here or reopen.

PR #3352, opened 2026-08-01
fix(find): preserve native visibility and bracket globs

ricorna:
## Summary

- Preserve native `find` visibility when native flags are used by disabling hidden, `.ignore`, and git-ignore pruning.
- Keep the existing ignore-aware behavior for compact RTK syntax.
- Support positive, ranged, and negated bracket classes such as `[123]`, `[0-9]`, and `[!0-9]`.
- Support `]` as the first class member, including patterns such as `[]a]` and `[!]]`.
- Preserve unmatched `[` as a literal instead of introducing a malformed-pattern regression.

This addresses the hidden/gitignored and bracket-class false negatives documented in #2995 and its follow-up report. It overlaps the visibility portion of draft #2996, while also covering the bracket-class cases that PR does not address.

## Root cause

The `ignore::WalkBuilder` always pruned hidden and ignored entries, even when RTK was transparently rewriting native `find ... -name/-type/-maxdepth` syntax. The custom glob matcher only implemented `*` and `?`, so valid native bracket expressions silently returned no matches with exit 0.

## Verification

- Added end-to-end tests for hidden and gitignored matches under native syntax.
- Added end-to-end tests for ranged, negated, and literal-`]` bracket classes.
- Added a regression test for an unclosed literal bracket.
- `cargo clippy --locked --all-targets -- -D warnings`
- `cargo test --locked`

The complete locked suite passed locally: 2,512 unit tests passed, 8 ignored, plus all integration targets.

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -7,7 +7,7 @@ use ignore::WalkBuilder;
 use std::collections::HashMap;
 use std::path::Path;
 
-/// Match a filename against a glob pattern (supports `*` and `?`).
+/// Match a filename against a find-style glob pattern.
 fn glob_match(pattern: &str, name: &str) -> bool {
     glob_match_inner(pattern.as_bytes(), name.as_bytes())
 }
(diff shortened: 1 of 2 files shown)


KuSh:
Thanks for the bracket-class work and the fixture-backed tests. Bracket patterns have been handled since #3603 (merged 2026-08-26): the subset rtk walks itself now refuses any -name value containing "[" and hands the command to real find, so on current develop "a[0-9].txt", "a[!0-9].txt" and "[]a]*.txt" all return the right files without rtk implementing fnmatch itself. The same merge removed has_native_find_flags and rewrites the compact rtk syntax into find syntax before dispatch, so the native-versus-compact distinction this PR keys its visibility carve-out on no longer exists at that point, and #3854 (merged 2026-09-03) then chose to disclose pruned entries as "... (N filtered: hidden or gitignored)" with a tee file rather than return them. That last decision is still contested: hidden and gitignored matches are indeed missing from the results, and open issue #2995 tracks it, so a narrow PR against the current walker would be the way back in. Please comment or reopen if a case here was missed.

PR #3353, opened 2026-08-01
fix(test): preserve diagnostics from generic runner failures

### LINKED ISSUE #2420: rtk test shows only the last 5 output lines (not failures) for unrecognized runners

ricorna:
## Summary

- Pass the child process exit code into generic test summarization and treat it as authoritative.
- Bypass the shared never-worse raw fallback on nonzero generic-test exits so short success-looking or empty raw output cannot erase the failure banner.
- Recover failure diagnostics from the complete raw output rather than the displayed tail.
- Keep bounded head/tail context plus anchor-centered context around detected errors and recognized specialized failures, including diagnostics displaced by cleanup tails.
- Ignore zero-valued counters such as `0 failed` when locating real diagnostics.
- Keep standalone `FAIL:` recognition scoped to `rtk test` so `rtk err` behavior is unchanged.
- Add subprocess and unit regressions for early failures, heuristic runner-name collisions, contradictory zero-exit text, success-looking nonzero summaries, zero-failure counters, recognized failures with nearby causal diagnostics, and unknown failure formats.

## Safety contract

- Exit code `0` never receives a failure banner solely because output contains words such as `FAILED`.
- A nonzero exit always receives a failure presentation and retains bounded actionable context.
- Full raw output remains available to RTK tee handling.
- Never-worse output capping remains active.

## Verification

- `cargo clippy --locked --all-targets -- -D warnings`
- `cargo test --locked`
- `cargo test --locked --test generic_test_runner_test`

The complete locked suite passed locally: 2,515 unit tests passed, 8 ignored, plus all integration targets.

Fixes #2420

This overlaps earlier work in #2427, #2683, and #2996, but adds full-output diagnostic recovery, authoritative exit-status handling, and regression coverage for success-looking nonzero output and zero-exit scary words.

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -34,6 +34,15 @@ static ERROR_PATTERNS: LazyLock<Vec<Regex>> = LazyLock::new(|| {
     ]
 });
 
+static TEST_FAIL_PATTERN: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"(?i)^.*\bfail\b.*$").unwrap());
+static ZERO_OUTCOME_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(
+        r"(?i)\b(?:0\s+(?:test\s+)?(?:failed|failures?|errors?|warnings?)|(?:failures?|errors?|warnings?)\s*:\s*0)\b",
+    )
+    .unwrap()
+});
+
 struct ErrorStreamFilter {
     in_error_block: bool,
     blank_count: usize,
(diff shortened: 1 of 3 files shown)


KuSh:
Closing in favour of #2683, which fixes the same generic runner fallback in extract_test_summary for the same issue and keeps the shared error pattern list untouched. Review and comments are welcome there. If part of the scope here is not covered by #2683, say so on that PR or reply here.

PR #3354, opened 2026-08-01
fix(bench): serve curl/wget fixtures locally instead of fetching a random payload

kylehgc:
The `curl` and `wget` benchmark cases fetch `https://mockhttp.org/json/1`, which returns a **randomly chosen body per request**:

```
call 1: 152 bytes  {"metrics":...
call 2: 244 bytes  {"config":...
call 3: 102 bytes  {"message":...
call 4: 138 bytes  {"metrics":...
```

`bench()` invokes the raw command and the rtk command as two separate requests, so the two sides receive different payloads and the ratio between them is noise. When the rtk side happens to draw a larger body than the raw side, the case is recorded as `NEG` and the build fails.

## This is visible in CI history

Over the last 40 CI runs the `benchmark` job executed 17 times and **failed 4 of them** — every failure on `curl json`, each with different numbers:

```
35 → 38    (138 → 152 bytes)
38 → 61    (152 → 244 bytes)
```

Roughly a one-in-four chance of failing a build for reasons unrelated to any change under test. As far as I can tell it is currently the only recurring CI failure in the repo.

## Fix

Serve the fixtures from `python3 -m http.server` on a loopback port. Deterministic, offline, and still exercising HTTP.

`file://` URLs were considered and rejected: wget does not support the file scheme at all (the raw side silently returns nothing and the case degrades to `SKIP`, which does not fail the build — so the case would look healthy while measuring nothing), and `file://` would bypass the HTTP path these cases exist to measure.

## Verification

Three successive raw fetches return 470 bytes each, and rtk returns 470 — the case now ties at 0% and cannot produce a spurious `NEG`.

Because rtk passes JSON through unchanged, `curl json` and `curl text` will report 0% in `WARN` from here on, rather than a fabricated saving that depended on which random payload the raw side happened to draw.

If `python3` is unavailable the server never starts and the curl and wget cases are skipped with an explicit message, rather than silently reporting zero.

## Notes

- No Rust changes; the diff is `scripts/benchmark.sh` plus two fixture files.
- Your CI runs `./scripts/benchmark.sh` directly, so the PR checks validate the fix end to end.
- Fixture payloads are modelled on the shapes mockhttp.org was returning, so the measured sizes stay in the same range as before.


KuSh:

PR #3369, opened 2026-08-03
feat(hooks): add Kiro IDE and CLI integration

RogerFelipeNsk:
## Summary

- Adds **Kiro** (kiro.dev) as a supported agent: steering file (primary, prompt-level) + PreToolUse hook (`rtk hook kiro`, deny-with-suggestion reinforcement), reusing the shared decision flow (`decide_hook_action` / `get_rewritten` / `registry::rewrite_command`)
- Kiro's PreToolUse contract has no transparent-rewrite field, so the hook exits `2` with the `rtk` suggestion on stderr; Kiro forwards it to the agent, which re-issues the command — already-`rtk` commands pass through, so the retry cannot loop, and every failure path exits `0` (a broken hook never blocks the user)
- Install scopes mirror what Kiro reads: project install (`.kiro/`) covers IDE + CLI; `--global` writes both artifacts to `~/.kiro/` (global hooks are loaded by **Kiro CLI v3+** only — the IDE reads workspace hooks; `rtk init` and `--show` explain this)

## Test plan

- [x] `cargo fmt --all -- --check && cargo clippy --all-targets && cargo test --all` (all green, warnings-as-errors)
- [x] Manual hook contract testing via stdin: rewritable command → exit 2 + single stderr suggestion line; `rtk`-prefixed retry → exit 0 silent; snake_case `execute_bash` tool name → exit 2; malformed JSON → exit 0 + parse note; empty stdin / non-shell tool → exit 0 silent
- [x] Installer round-trips in temp dirs (project) and temp `$HOME` (`--global`): install → `--show` reports `[ok]` per scope → uninstall removes only the two RTK-managed files, preserving user files in `.kiro/steering/` and `.kiro/hooks/`
- [x] Verified live in Kiro IDE (workspace hook fires, agent re-issues the `rtk` form) and Kiro CLI

## Notes for reviewers

- `src/hooks/init.rs` is a flagged critical file — changes are additive (new `run_kiro_mode` / `uninstall_kiro` / `show_kiro_config` section + one dispatch arm); no existing agent flow was modified
- `hook_check.rs` / `integrity.rs` / `verify_cmd.rs` deliberately untouched, following the Factory Droid precedent (#2267)
- `Host::Kiro` loads no Kiro-native permission source yet (empty rule sets → `Default` verdict); documented in `src/hooks/README.md` as a known limitation
- `read_limited_from` is a pure extraction from `read_stdin_limited` (shared by all hook handlers, behavior unchanged) so the 1 MiB stdin cap is unit-testable
- Hook file format verified against the current Kiro docs (v1 `hooks` array, `trigger: PreToolUse`, `matcher: execute_bash`, command action, exit-2-blocks semantics)


awskrishnak:
I validated the deny-with-suggestion approach independently in Kiro CLI using a global PreToolUse hook matching both `execute_bash` and `shell`.

Observed behavior:
- raw `git status` is blocked with exit 2 and a suggestion to rerun `rtk git status`;
- an already-prefixed `rtk git status` exits 0 silently, so the retry does not loop;
- unrelated commands pass through unchanged.

This matches the approach in this PR and confirms it works with the Kiro hook contract in a real Kiro CLI setup. The `execute_bash` matcher is important because Kiro emits that snake_case tool name in some configurations.

RogerFelipeNsk:
> I validated the deny-with-suggestion approach independently in Kiro CLI using a global PreToolUse hook matching both `execute_bash` and `shell`.
> 
> Observed behavior:
> 
> * raw `git status` is blocked with exit 2 and a suggestion to rerun `rtk git status`;
> * an already-prefixed `rtk git status` exits 0 silently, so the retry does not loop;
> * unrelated commands pass through unchanged.
> 
> This matches the approach in this PR and confirms it works with the Kiro hook contract in a real Kiro CLI setup. The `execute_bash` matcher is important because Kiro emits that snake_case tool name in some configurations.

Thanks for validating this independently in Kiro CLI — that's exactly the contract the PR relies on (exit 2 + stderr suggestion, silent pass-through for already-prefixed rtk commands, no loop).

Your setup surfaced two gaps, both now addressed in the updated commit:
- Global hook: rtk init --agent kiro --global now installs the hook to ~/.kiro/hooks/ alongside the steering file (it previously wrote the hook to the project only). Since Kiro IDE reads hooks from the workspace, rtk init and rtk init --show note that the global hook applies to Kiro CLI.
- Matcher: the hook config now matches both execute_bash and shell, as in your setup, so either tool name triggers it.

PR #3372, opened 2026-08-03
fix(awareness): slim awareness text 

aeppling:
## Summary

- built on top of #3319 

## Test plan


hooks/claude/rtk-awareness.md
```diff
@@ -1,29 +1,9 @@
-# RTK - Rust Token Killer
+# Command output
 
-**Usage**: Token-optimized CLI proxy (cuts up to 90% of bash output)
+Command output in this environment may be shown in a condensed form that
+preserves all errors, warnings, test results, diffs, and other actionable
+content. Treat condensed output as the complete, authoritative result of the
+command — there is no hidden or truncated information you need to recover.
```
KuSh: Perhaps add a "unless explicitly stated" for tee hints?

hooks/claude/rtk-awareness.md
KuSh, line 1: This file content is no longer associated with Claude at all and has become generic. It could perhaps be moved outside of the claude dedicated folder (CodeBuddy PR is reusing it with content substitution for example)

hooks/claude/rtk-awareness.md
```diff
@@ -1,29 +1,9 @@
...
 
-## Meta Commands (always use rtk directly)
-
-```bash
-rtk gain              # Show token savings analytics
-rtk gain --history    # Show command usage history with savings
-rtk discover          # Analyze Claude Code history for missed opportunities
-rtk proxy <cmd>       # Execute raw command without filtering (for debugging)
```
KuSh, line 11: This and `RTK_DISABLED` bypass documentation seems legit to keep as a sanctioned way to recover from a bad/incomplete filter result. Don't you think?

KuSh (commented):
You seem to have accidentally included the commit from #3343

aeppling:
@KuSh 

Thanks for review, it was intentionally stacked on top of #3343 for benchmarking purpose.
Now cleaned.

hooks/claude/rtk-awareness.md
aeppling: done

hooks/claude/rtk-awareness.md
aeppling: This is debatable of course, but why would an agent need to run rtk gain, rtk discover or whatever rtk commads ?

The only thing agent should know is "rtk proxy" then, but others are useless, and it mostly perturbate the agent in their behavior.

In this case, maybe we can just keep "rtk proxy" with a clear explanation ?

hooks/claude/rtk-awareness.md
aeppling: Updated, i tried to rework the text for a more clear usage of recovery and proxy

KuSh (approved):
LGTM!

aeppling:
@KuSh updated with awareness level configuration (agent without hooks are given the full instructions, else it is default / slim, user can still configure with toml)

src/hooks/init.rs
```diff
@@ -109,6 +111,41 @@ pub enum PatchResult {
...
+    };
+    println!(
+        "{prefix}Awareness: full — {agent} has no command hook, so the agent is told to prefix rtk itself{ignored}"
+    );
+}
+
+/// Wrap an awareness file in the `<!-- rtk-instructions -->` markers used by
+/// `write_rtk_block`, so it can be upserted into a shared file like AGENTS.md.
+fn rtk_block(body: &str) -> String {
+    format!("{RTK_BLOCK_START} v2 -->\n{body}{RTK_BLOCK_END}\n")
```
KuSh: nitpick: set current RTK_BLOCK_VERSION in a const?

src/hooks/init.rs
```diff
@@ -1951,6 +2017,7 @@ fn run_kimi_mode_at(base_dir: &Path, ctx: InitContext) -> Result<()> {
     if !ctx.dry_run {
         println!("\nRTK configured for Kimi AI.\n");
         println!("  AGENTS.md: {}", agents_md_path.display());
+        print_rules_only_awareness_note("Kimi AI", ctx);
```
KuSh: For other agents, this doesn't depend on ctx.dry_run value. For Kimi and Codex dry-run users never see the "awareness: full (config level ignored)" note

src/hooks/init.rs
```diff
@@ -23,16 +23,18 @@ use super::constants::{
...
 
 // Embedded OpenCode plugin (auto-rewrite)
 const OPENCODE_PLUGIN: &str = include_str!("../../hooks/opencode/rtk.ts");
 
 // Embedded Pi extension (auto-rewrite)
 const PI_PLUGIN: &str = include_str!("../../hooks/pi/rtk.ts");
 
-// Embedded slim RTK awareness instructions
-const RTK_SLIM: &str = include_str!("../../hooks/claude/rtk-awareness.md");
-const RTK_SLIM_CODEX: &str = include_str!("../../hooks/codex/rtk-awareness.md");
```
KuSh, line 113: Those files (antigravity, cline, codex, kilocode, windsurf) should be removed are they're now unused.

src/main.rs
```diff
@@ -2017,6 +2033,7 @@ fn run_cli() -> Result<i32> {
             let ctx = hooks::init::InitContext {
                 verbose: cli.verbose,
                 dry_run,
+                awareness: configured_awareness_level(),
```
KuSh: nitpick: This runs unconditionally, including for commands such as `--show` and `--uninstall`, which don’t use `ctx.awareness`. Could we limit it to the cases where it’s needed?

src/hooks/init.rs
KuSh:
```suggestion
            println!("              replaced with @RTK.md (awareness: {})", ctx.awareness);
```

KuSh (approved):
LGTM except for some nitpicks

KuSh (commented):
The three-level model is the right shape, and routing rules-only agents (Codex, Cline, Windsurf, Kilo Code, Antigravity, Kimi) unconditionally to `full` with `print_rules_only_awareness_note` is the correct call — those agents genuinely have to type `rtk` themselves.

One path is left behind, though: **`rtk init` without `-g` never reaches `awareness_content`.**

`run_default_mode` early-returns for local init (`src/hooks/init.rs:1196-1201`):

```rust
if !global {
    // Local init: inject CLAUDE.md + generate project-local filters template
    run_claude_md_mode(false, install_opencode, ctx)?;
    generate_project_filters_template(ctx)?;
    return Ok(());
}
```

and `run_claude_md_mode` still writes the legacy constant (`src/hooks/init.rs:1656-1662`):

```rust
let action = write_rtk_block(
    &path,
    RTK_INSTRUCTIONS,
    "rtk instructions",
    recovery_cmd,
    ctx,
)?;
```

`RTK_INSTRUCTIONS` (`src/hooks/init.rs:157`) is still the v2 block with:

```
## Golden Rule

**Always prefix commands with `rtk`**. If RTK has a dedicated filter, it uses it. If not, it passes through unchanged.

**Important**: Even in command chains with `&&`, use `rtk`:
...
rtk git add . && rtk git commit -m "msg" && rtk git push
```

So on a Claude Code install that *does* have the transparent rewrite hook, a project-local `rtk init` writes prefix-everything guidance into `./CLAUDE.md` and `awareness.level` has no effect on it. `--claude-md` opting into that is intentional and fine; plain `rtk init` inheriting it is the part that looks unintended, since the PR's whole premise is that hook-based agents should default to being told nothing about RTK.

Practical consequence is the one #3186 describes: Claude reads the Golden Rule and types `rtk git status` itself. Claude Code matches `permissions.allow` against the literal command *before* the PreToolUse hook rewrites it, so the prefixed form misses the user's plain-command allow rules and prompts every time — while the unprefixed form would have been rewritten transparently anyway.

Two ways out, both small:

1. Thread the level through — give `run_claude_md_mode` an explicit content argument and pass `awareness_content(ctx.awareness)` from the local branch, keeping `RTK_INSTRUCTIONS` only for the explicit `--claude-md` opt-in.
2. Have local init follow the same shape as global: write the awareness file and add the `@RTK.md` reference, rather than inlining a block.

Worth noting the docs added here already imply (1): the configuration table says `default` means "Nothing about RTK", and "Re-run `rtk init -g` (or your agent's init command) to rewrite the file" reads as though every init path honours the setting.

Also minor: `run_claude_md_mode`'s doc comment still says "full 137-line injection" — worth a look while the constant is in scope, since the block is now ~4KB and the test at `src/hooks/init.rs:5424` asserts `RTK_INSTRUCTIONS.len() > 4000`.

src/hooks/init.rs
KuSh: Addressed in 4bbf778897 — `RTK_BLOCK_VERSION` is now a named const used by `rtk_block()`.

src/hooks/init.rs
KuSh: Fixed in 4bbf778897 — `print_rules_only_awareness_note` now runs unconditionally (moved outside `if !ctx.dry_run`) for both Kimi and Codex, matching the other agents.

src/hooks/init.rs
KuSh: Removed in 4bbf778897 (antigravity/cline/codex/kilocode/windsurf rules.md all deleted).

src/main.rs
KuSh: Fixed in 4bbf778897 — `--show`/`--uninstall` now use `AwarenessLevel::default()` directly and skip `configured_awareness_level()`'s config-file read entirely.

src/hooks/init.rs
KuSh: Applied verbatim in 4bbf778897.

hooks/claude/rtk-awareness.md
KuSh: Looks resolved by the current text — `hooks/rtk-awareness.md` keeps only `rtk proxy <cmd>` with a clear trigger condition, no `rtk gain`/`discover`/`RTK_DISABLED` mentions, matching your last comment. Reopen if `RTK_DISABLED` should be called out too.

KuSh:
Following up on the local-init review from Aug 29: fixed in 4bbf778897. `run_default_mode`'s local branch now calls `run_claude_md_mode_with(false, install_opencode, &rtk_block(awareness_content(ctx.awareness)), ctx)` instead of the legacy `RTK_INSTRUCTIONS` Golden Rule block; `RTK_INSTRUCTIONS` is now only used by the explicit `--claude-md` opt-in. Covered by the new `test_local_init_block_follows_awareness_level_not_legacy` test, and verified against the built binary: plain `rtk init` now writes the slim awareness block to `CLAUDE.md`, while `rtk init --claude-md` still gets the full legacy block.

PR #3376, opened 2026-08-03
fix(telemetry): stop low_savings_commands from sending command arguments

### LINKED ISSUE #1785: Prevent telemetry low_savings_commands from including arguments

lutzkuen:
Closes #1785

## Problem

`low_savings_commands` is sent in every telemetry ping. It formatted each entry from the **first three whitespace-separated words** of the stored `rtk_cmd` ([`src/core/tracking.rs`](https://github.com/rtk-ai/rtk/blob/develop/src/core/tracking.rs)).

Stored labels embed raw user input, so the third word is frequently an argument rather than a subcommand:

| Stored `rtk_cmd` | Old telemetry value | Leak |
|---|---|---|
| `rtk proxy mysql -pS3cr3t-passw0rd` | `rtk proxy mysql` | binary name |
| `rtk:toml DB_PASSWORD=hunter2 psql` | `rtk:toml DB_PASSWORD=hunter2 psql` | **credential** |
| `rtk grep AWS_SECRET_ACCESS_KEY .env` | `rtk grep AWS_SECRET_ACCESS_KEY` | **search pattern** |
| `rtk read /home/alice/.ssh/id_ed25519` | `rtk read /home/alice/.ssh/id_ed25519` | **path** |
| `rtk playwright test tests/login.spec.ts` | `rtk playwright test` | ok |

Label sources that carry raw arguments: `rtk:toml {raw_command}` and `rtk fallback: {raw_command}` ([`src/main.rs`](https://github.com/rtk-ai/rtk/blob/develop/src/main.rs)), `rtk proxy {cmd} {args}`, plus the many `format!("rtk <tool> {}", args.join(" "))` labels across `src/cmds/`.

This contradicts [`docs/TELEMETRY.md`](https://github.com/rtk-ai/rtk/blob/develop/docs/TELEMETRY.md), under **What is NOT collected**:

> - Full command lines or arguments (only tool names like "git", "cargo")
> - File paths or directory structures
> - Secrets, API keys, or environment variable values

`low_savings_commands` was the only remaining payload field with this gap — `top_commands`, `top_passthrough`, `ecosystem_mix`, `projects_count` and `meta_usage` are all already reduced.

## Fix

A deny-by-default sanitizer, `telemetry_label()`, applied before any label leaves the process:

- Keeps the RTK **route marker** (`proxy`, `fallback:`, `rtk:toml`) — these are literals from RTK's own source, never user input, so they preserve useful routing signal.
- Keeps the **tool name** only if it looks like a plain command name (ASCII alphanumeric start; only `.`, `_`, `+`, `-` besides alphanumerics; ≤32 chars). That rejects paths, URLs, flags, `KEY=value` env prefixes and credential-shaped strings.
- **Drops everything after the tool name**, unconditionally.
- Emits `<other>` when the tool token fails validation, so the "some filter underperforms" signal survives without transmitting the value.

This is the same tool-name-only reduction `top_commands` and `top_passthrough` already apply, which is what issue #1785 recommends as the primary mitigation.

Because many stored labels now collapse onto one, the averages are re-aggregated **weighted by occurrence count**. Previously `rtk git log --oneline`, `rtk git log -20` and `rtk git log --stat` could occupy three of the five slots with near-duplicate entries; they now form one weighted `rtk git` entry, so the top 5 covers five distinct tools. The SQL scan window is capped at 500 distinct labels to keep the query bounded.

### Trade-off

Subcommand granularity is lost: `rtk git log` becomes `rtk git`. Keeping the subcommand safely would need a curated allowlist of known subcommand words, since the token in that position is user-controlled for several routes (`rtk grep <pattern>`, `rtk npx <args>`, `rtk playwright <args>`, `rtk proxy <cmd> <args>`) — a shape check alone does not distinguish `log` from a password. I went with the issue's primary suggestion rather than introduce a list to maintain, but happy to add one if you'd prefer the granularity.

## Tests

Five new tests in `src/core/tracking.rs`, including the URL/path/token-like argument cases the issue asks for:

- `test_telemetry_label_keeps_tool_only`
- `test_telemetry_label_keeps_route_markers`
- `test_telemetry_label_drops_arguments` — asserts no credential/URL/path/pattern survives, for six leaky stored labels
- `test_telemetry_label_rejects_argument_shaped_tools` — `<other>` fallback, incl. the length cap
- `test_low_savings_commands_never_reports_arguments` — end-to-end through the DB
- `test_low_savings_commands_aggregates_reduced_labels` — weighted average + ordering

## Verification

```
cargo fmt --all          # clean
cargo clippy --all-targets  # no warnings
cargo test --all         # 2631 passed, 0 failed
```

## Docs

`docs/TELEMETRY.md` and `docs/guide/resources/telemetry.md` updated to state that the field is reduced to the tool name and arguments are never sent. The `docs/TELEMETRY.md` example changes from `rtk docker ps:25%` to `rtk docker:25%` to match actual output.

## Scope

No behavior change outside the telemetry payload. `rtk gain --history` still shows full local labels — the reduction happens only at the telemetry boundary.

### DIFF
--- a/docs/TELEMETRY.md
+++ b/docs/TELEMETRY.md
@@ -62,7 +62,7 @@ This data directly drives our roadmap. For example, if telemetry shows that 40% 
 |-------|---------|---------|
 | `passthrough_top` | `["git:15", "npm:8"]` | Top 5 commands with 0% savings — these need filters |
 | `parse_failures_24h` | `3` | Filter fragility — high count means filters are breaking |
-| `low_savings_commands` | `["rtk docker ps:25%"]` | Commands averaging <30% savings — filters to improve |
+| `low_savings_commands` | `["rtk docker:25%"]` | Commands averaging <30% savings — filters to improve. Reduced to the tool name only (arguments are dropped, never sent) |
 | `avg_savings_per_command` | `68.5` | Unweighted average (vs global which is volume-biased) |
 
 ### Ecosystem distribution
--- a/docs/guide/resources/telemetry.md
+++ b/docs/guide/resources/telemetry.md
@@ -67,7 +67,7 @@ This data directly drives our roadmap. For example, if telemetry shows that 40% 
 |-------|---------|---------|
 | `passthrough_top` | `["git:15", "npm:8"]` | Top 5 commands with 0% bash output reduction — these need filters |
 | `parse_failures_24h` | `3` | Filter fragility — high count means filters are breaking |
-| `low_savings_commands` | `["rtk <cmd>:25%"]` | Commands averaging <30% bash output reduction — filters to improve. The example is a placeholder, not a measured value |
+| `low_savings_commands` | `["rtk <cmd>:25%"]` | Commands averaging <30% bash output reduction — filters to improve. Reduced to the tool name only (arguments are dropped, never sent). The example is a placeholder, not a measured value |
 | `avg_savings_per_command` | `68.5` | Unweighted average (vs global which is volume-biased) |
 
 ### Ecosystem distribution
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -1065,21 +1065,49 @@ impl Tracker {
     }
 
     /// Count commands with low savings (<30%) — filters that need improvement.
+    ///
+    /// Labels are reduced with [`telemetry_label`] before being returned, so the
+    /// result never carries user-supplied arguments (see that function's docs).
+    /// Several stored commands collapse onto the same reduced label, so averages
+    /// are re-computed weighted by occurrence count.
     pub fn low_savings_commands(&self, limit: usize) -> Result<Vec<(String, f64)>> {
         let mut stmt = self.conn.prepare(
-            "SELECT rtk_cmd, AVG(savings_pct) as avg_sav FROM commands
+            "SELECT rtk_cmd, AVG(savings_pct) as avg_sav, COUNT(*) as cnt FROM commands
              WHERE input_tokens > 0
              GROUP BY rtk_cmd
              HAVING avg_sav < 30.0 AND avg_sav > 0.0
-             ORDER BY COUNT(*) DESC LIMIT ?1",
+             ORDER BY cnt DESC LIMIT ?1",
         )?;
-        let rows = stmt.query_map(params![limit as i64], |row| {
+        let rows = stmt.query_map(params![LOW_SAVINGS_SCAN_LIMIT], |row| {
             let cmd: String = row.get(0)?;
             let sav: f64 = row.get(1)?;
-            let short = cmd.split_whitespace().take(3).collect::<Vec<_>>().join(" ");
-            Ok((short, sav))
+            let count: i64 = row.get(2)?;
+            Ok((cmd, sav, count))
         })?;
-        Ok(rows.filter_map(|r| r.ok()).collect())
+
+        let mut totals: std::collections::HashMap<String, (f64, i64)> =
+            std::collections::HashMap::new();
+        for (cmd, sav, count) in rows.filter_map(|r| r.ok()) {
+            if count <= 0 {
+                continue;
+            }
+            let entry = totals.entry(telemetry_label(&cmd)).or_insert((0.0, 0));
+            entry.0 += sav * count as f64;
+            entry.1 += count;
+        }
+
+        let mut aggregated: Vec<(String, f64, i64)> = totals
+            .into_iter()
+            .map(|(label, (sum, count))| (label, sum / count as f64, count))
+            .collect();
+        // Most-used first; ties broken by label so the payload is deterministic.
+        aggregated.sort_by(|a, b| b.2.cmp(&a.2).then_with(|| a.0.cmp(&b.0)));
+        aggregated.truncate(limit);
+
+        Ok(aggregated
+            .into_iter()
+            .map(|(label, avg, _)| (label, avg))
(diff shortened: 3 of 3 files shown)


lutzkuen:
Closing this — I should not have opened it.

@YOMXXX has [#2578](https://github.com/rtk-ai/rtk/pull/2578) open since 2026-06-24 fixing the same field in the same function, and it is the smaller change (+61/-4 against my +246/-8). Theirs came first. @kobihikri also hit this and closed [#3233](https://github.com/rtk-ai/rtk/pull/3233) for exactly the same reason a week ago, so this is now the third PR on one 20-line function — clearly not what the review queue needs.

My duplicate check was the problem: I searched issues, found @LionartBR's #1785, then listed open PRs with a limit that only covered the most recent ~100. With 900+ PRs open, #2578 was nowhere near that window. Querying PRs that reference the issue number would have found it immediately. My mistake, and I'm sorry for the noise.

**Maintainers: [#2578](https://github.com/rtk-ai/rtk/pull/2578) is the one to review.** @LionartBR reported it, @YOMXXX wrote the fix. It currently shows as CONFLICTING against `develop`, which is probably the main thing standing between it and a review.

Two things from my version that may be worth folding into #2578 if @YOMXXX wants them — entirely their call, and neither is a security concern:

1. **Reduced labels can crowd each other out.** Once labels collapse to the tool name, `rtk git log --oneline`, `rtk git log -20` and `rtk git status` all become `git`, and each still occupies its own slot in the top 5 with its own average. Re-aggregating weighted by occurrence count would make the five slots cover five distinct tools. This is a metrics-quality point, not part of the leak fix, and is better as a follow-up PR than as scope added to #2578.

2. **A regression test for the exact reported path.** @kobihikri's reproduction on [#1785](https://github.com/rtk-ai/rtk/issues/1785#issuecomment-5088024408) — a connection-string password surviving into the POSTed payload when average savings land in the `0 < avg < 30` window — makes a good assertion to pin the behaviour, whichever fix lands.

Happy to open the aggregation follow-up separately once #2578 is merged, if that's useful.

*Disclosure: I used AI assistance (Claude Code) to investigate and write the change and this comment.*

PR #3377, opened 2026-08-03
fix(read): honor --max-lines N as exact head count

### LINKED ISSUE #3370: `rtk read --max-lines N` outputs only N/2 lines — `head -N` hook rewrite silently under-delivers half the requested lines
## Environment

- Claude Code, hook: `PreToolUse` → `matcher: "Bash"` → `rtk hook claude`

## Summary

`--max-lines N` emits exactly `N/2` content lines (plus a `[K more lines]` marker), not `N`. Because the Claude Code hook rewrites `head -N <file>` to `rtk read <file> --max-lines N`, every agent `head -N` receives half the lines it asked for.

## Repro

```console
$ python3 -c "open('long.txt','w').write(''.join(f'line {i}: some content here\n' for i in range(3000)))"
$ for n in 5 10 20 50 100 200; do
    echo "max-lines $n -> $(rtk read long.txt --max-lines $n | grep -c '^line ')"
  done
max-lines 5 -> 2
max-lines 10 -> 5
max-lines 20 -> 10
max-lines 50 -> 25
max-lines 100 -> 50
max-lines 200 -> 100
```

Expected (matching `head -50`): 50 lines. Actual: 25 lines + `[2975 more lines]`.

Hook path that makes this user-visible:

```console
rtk read big.log --max-lines 50        # → agent gets 25 lines
```

`--tail-lines N` is exact (verified against `tail -N`, byte-identical); only `--max-lines` halves.

## Impact

- Any agent running `head -N file` silently gets `N/2` lines. The truncation marker is visible in standalone use, but the count contradicts what was requested.
- On 0.38.0 this also poisoned pipelines (`head -200 f | grep X` → grep searched only 100 lines and the marker was consumed by grep → silent false negative). 0.44.2 no longer rewrites `head` inside pipelines, which contains that failure mode — but the standalone halving remains.

## Additional observation (savings accounting)

`rtk read ... --max-lines/--tail-lines` invocations are logged to `history.db` with `original_cmd` synthesized as plain `cat <file>` (flags erased), `input_tokens` = the full file, and the difference booked as savings. Example measured entry: a `--max-lines 20` read of a 1.27 MB file logged as `cat <file>`, input=318,445 tokens, output=119, savings 99.96%. A raw `head -20` would have produced the same small output without rtk, so the full-file baseline overstates real savings. In our 3-month ledger, 205 such entries account for 71.5M of 79.4M total `rtk read` "saved" tokens (~90%). Related in spirit to #3339 (different mechanism). Happy to split this into its own issue if you prefer.

mohammedelkarsh:
## Summary
- `--max-lines N` now keeps exactly N content lines (plus a recoverable `[K more lines]` marker), matching `head -N` rewrite expectations
- Root cause: `smart_truncate` only retained `max/2` plain-text lines, so Claude/hook rewrites of `head -N` silently under-delivered
- Removed unused `smart_truncate` and added regression coverage from the #3370 repro

Fixes #3370

## Test plan
- [x] `cargo test` (includes `test_head_truncate_plain_text_exact_count` for N in 5..200)
- [ ] Manual: `rtk read long.txt --max-lines 50` should print 50 content lines + marker


src/cmds/system/read.rs
```diff
@@ -179,12 +178,32 @@ fn apply_line_window(
     }
 
     if let Some(max) = max_lines {
-        return filter::smart_truncate(content, max, lang);
+        // Honor --max-lines exactly (head -N parity). Do not use smart_truncate:
+        // it keeps only max/2 plain-text lines, which silently under-delivers when
+        // hooks rewrite `head -N file` → `rtk read file --max-lines N` (#3370).
+        return head_truncate(content, max);
```
KuSh, line 184: This branch should keep `filter::smart_truncate`.

`--max-lines` is the structure-aware budget — it's what makes `-l minimal|aggressive --max-lines N` return an overview of signatures and imports rather than the literal first N lines. The `head -N` parity fix belongs in a *new* `head_lines` branch above this one, mirroring the `tail_lines` branch at the top of the function.

KuSh (commented):
Thanks for digging into this — the repro in #3370 is real, but I think the fix is pointed at the wrong flag, and I'd like to redirect the PR before it goes further.

### The flag pair is already half-built

`--tail-lines` (`read.rs:165-177`) is an exact, faithful window: slice, join, preserve the trailing newline, no marker. A one-to-one mirror of `tail -N`.

`--max-lines` is a different thing. It routes to `filter::smart_truncate`, a **structure-aware budget** that keeps signatures / imports / `pub` / `export` / braces plus `max/2` ordinary lines and appends `[K more lines]`. Halving isn't a bug there — it's what a budget does, and it's the mechanism that makes `rtk read big.rs -l aggressive --max-lines 100` return an overview instead of the literal first 100 lines.

So the actual defect is at `src/discover/registry.rs:1150`:

```rust
return Some(format!("rtk read {} --max-lines {}", file, n));
```

`head -N` is rewritten onto the budget flag because **no head-side window flag exists**. The tail side got the pairing right; the head side never did. Redefining `--max-lines` fixes the rewrite by trading one wrong behavior for another — it makes `head -N` exact at the cost of deleting the only structure-aware truncation path in the tree.

### Proposal: add `--head-lines`, leave `--max-lines` alone

```rust
// main.rs — long-only, symmetric with tail_lines
#[arg(long, conflicts_with_all = ["max_lines", "tail_lines"])]
head_lines: Option<usize>,

// read.rs, apply_line_window — mirror of the tail_lines branch
if let Some(head) = head_lines {
    if head == 0 { return String::new(); }
    let lines: Vec<&str> = content.lines().collect();
    let end = lines.len().min(head);
    let mut result = lines[..end].join("\n");
    if content.ends_with('\n') { result.push('\n'); }
    return result;
}

// registry.rs:1150
return Some(format!("rtk read {} --head-lines {}", file, n));
```

Plus `docs/contributing/TECHNICAL.md:167` and the flag table at `docs/usage/FEATURES.md:133`.

`rtk read` defaults to `-l none` (`main.rs:110`), so `head -10 f` → `rtk read f --head-lines 10` stays byte-exact.

### Why this is worth an extra flag

I built the branch and exercised the release binary. `cargo test --all` and `cargo clippy --all-targets` are green, but the headline contract doesn't hold end to end:

1. **The guard eats the truncation.** `never_worse(&raw, &rtk_output)` at `read.rs:77` discards the truncated output whenever the ~15-byte `[K more lines]` marker outweighs the omitted content. On a 7-line file of short lines, `rtk read f --max-lines 3` prints **all 7 lines**.
2. **The marker carries no trailing newline.** `rtk read a.txt b.txt --max-lines 3` prints `[97 more lines]line number that is reasonably long 1`, because `main.rs` loops `read::run` per file and `run()` uses `print!`.

Both are consequences of emitting a marker at all. A faithful head window emits none — which is precisely why `--tail-lines` has neither bug today — so the `--head-lines` shape makes them disappear instead of requiring a `never_worse` exemption.

### Reframing

Under this reading, #3370's title ("`--max-lines` halving") names the symptom rather than the cause; something like "`head -N` is rewritten to the wrong `rtk read` flag" fits better, and the same goes for this PR's title. Worth retitling both once the direction is settled — I haven't touched either.

Inline notes below on the specific hunks.

src/cmds/system/read.rs
```diff
@@ -179,12 +178,32 @@ fn apply_line_window(
...
+    }
+    let lines: Vec<&str> = content.lines().collect();
+    if lines.len() <= max {
+        return content.to_string();
+    }
+    let omitted = lines.len() - max;
+    let mut result = lines[..max].join("\n");
+    result.push('\n');
+    result.push_str(&format!("[{} more lines]", omitted));
+    result
```
KuSh, line 204: Two separate bugs live in these lines, and both come from the marker:

- **`result` never ends in `\n`.** `run()` prints with `print!` and `main.rs` loops `read::run` per file, so `rtk read a.txt b.txt --max-lines 3` emits `[97 more lines]line number that is reasonably long 1`. It also leaves the shell prompt on the same line in the single-file case.
- **The marker costs ~15 bytes.** When that exceeds the omitted content, `never_worse` at line 77 throws this entire result away and prints the whole file — `rtk read f --max-lines 3` on a 7-line file of short lines prints all 7. Verified against the built release binary.

A faithful head window drops the marker entirely, the way the `tail_lines` branch does, and both go away without needing a `never_worse` exemption.

src/core/filter.rs
```diff
@@ -317,47 +317,6 @@ pub fn get_filter(level: FilterLevel) -> Box<dyn FilterStrategy> {
...
-            break;
-        }
-    }
-
-    // Single end-of-output marker: not code syntax, unambiguous to AI agents.
-    // Invariant: kept_lines + N == lines.len() (N = lines not shown)
-    result.push(format!("[{} more lines]", lines.len() - kept_lines));
-
-    result.join("\n")
-}
```
KuSh, line 359: I'd keep this.

It's the only structure-aware truncation path in the tree — after removal, `rtk read big.rs -l aggressive --max-lines 100` returns the literal first 100 lines instead of an overview.

Nothing forces the deletion either: the `head -N` rewrite at `registry.rs:1150` emits no `--level`, so it runs at `level = none` and never reaches this code. And `FUNC_SIGNATURE` / `IMPORT_PATTERN` remain in use at `filter.rs:257,264`, so this isn't dead-code cleanup.

src/cmds/system/read.rs
```diff
@@ -217,23 +236,64 @@ fn main() {{
...
+                .lines()
+                .filter(|l| l.starts_with("line "))
+                .count();
+            assert_eq!(kept, n, "max-lines {n} should keep {n} content lines, got {kept}");
+            assert!(
+                output.contains(&format!("[{} more lines]", 3000 - n)),
+                "missing recovery marker for n={n}: {output}"
+            );
+        }
     }
```
KuSh, line 297: This is the test that should have caught the `never_worse` clobber, and it can't: it calls `head_truncate` directly, so it never passes through `run()` where the guard lives. Same for the other four new tests.

Worth closing that gap whichever direction the fix takes — a binary-level test (`rtk_bin()` is defined right below) asserting an exact line count on a short-line file would have failed on this branch.

KuSh:
Closing in favour of #3941, which fixes the same head -N under-delivery along the direction asked for in the review here: a separate --head-lines window that is exact, with --max-lines left as the structure-aware budget. It also adds the missing head -n N and head --lines N matchers, preserves CRLF line endings, and arrives with green CI and much wider test coverage. Review and comments are very welcome on #3941, and if you read your scope as different from that, please reply here and say so.

PR #3378, opened 2026-08-03
fix(hooks): honor exclude_commands for head/tail rewrites

### LINKED ISSUE #3371: `[hooks] exclude_commands` silently ignored for `head`/`tail` rewrites (honored for `cat`)

mohammedelkarsh:
## Summary
- `head`/`tail` rewrites now consult `hooks.exclude_commands` before the `rewrite_line_range` short-circuit
- Makes `exclude_commands = [cat, head, tail]` consistent: all three stay unre written
- Adds regression coverage for `head -N`, `tail -N`, and `tail -n N`

Fixes #3371

## Test plan
- [x] `cargo test exclude_head_and_tail` / `head_rewrites_when_not_excluded`
- [ ] Manual: `rtk hook check 'head -50 big.log'` with `exclude_commands = [head]` prints `No rewrite`

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1361,9 +1361,14 @@ fn rewrite_segment_inner(
         return Some(trimmed.to_string());
     }
 
+    // head/tail short-circuit (flag reordering for `rtk read`) must still honor
+    // exclude_commands — otherwise `head`/`tail` exclusions are silently ignored (#3371).
     if context == RewriteContext::Normal
         && (cmd_part.starts_with("head -") || cmd_part.starts_with("tail "))
     {
+        if is_excluded(cmd_part, excluded) {
+            return None;
+        }
         return rewrite_line_range(cmd_part).map(|r| format!("{}{}", r, redirect_suffix));
     }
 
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as a duplicate of #3324, which fixes the same gap; #3324 will carry `Fixes #3371`.

There were seven PRs open against this one fast path. #3324 was picked on shape: it strips the env prefix before matching, so `sudo head -20 f` and `FOO=1 head -20 f` are covered too. This version matches `cmd_part` raw, so those still rewrite past the exclusion.

Thanks for the report and the fix — #3371 is what made the cluster findable.

PR #3379, opened 2026-08-03
fix(git): suppress blank line on empty git log range

### LINKED ISSUE #3365: fix(git): empty `git log` range emits a blank line instead of empty output — breaks `| wc -l` counting
## Summary

When a `git log` range contains **zero commits**, rtk's filter emits a single blank line (`\n`, 1 byte) where raw git emits nothing (0 bytes). Any consumer that counts lines — the classic `| wc -l` — reads **1** instead of **0**.

## Minimal repro

```console
$ rtk git log --oneline HEAD..HEAD | od -c
0000000   \n
0000001

$ rtk proxy git log --oneline HEAD..HEAD | od -c
$            # empty, as raw git
```

Non-empty ranges are filtered correctly.

## Why it matters for rtk's audience

rtk's primary users are LLM coding agents, and `git log <upstream>..HEAD | wc -l` is the idiomatic "how many unpushed commits" check they emit constantly. Under the hook, a compound command starting with a rewritten command (e.g. `git push … ; git log … | wc -l`) goes through the filter and the agent concludes **"1 unpushed commit" on a fully pushed branch** — then wastes turns re-pushing or investigating phantom state. That's how we found it.

Workarounds we use meanwhile: `git rev-list --count <range>` (emits a number, not lines) or `rtk proxy git log …`.

## Environment

- rtk 0.42.3 (Homebrew), macOS (Darwin 27.0.0)
- Changelog through v0.44.1 shows no related fix, and no existing issue matches (closest: #2305, #3028 — commit-dropping in non-empty logs, different defect).

## Expected

Empty range in → empty output out (0 bytes), matching raw git.

Hydr0gen19:
## Summary

- `rtk git log` on an empty commit range (e.g. `HEAD..HEAD`) printed a single blank line instead of producing no output at all, because `println!("{}", filtered)` always emits a trailing newline even when `filtered` is empty.
- This breaks the common `git log <upstream>..HEAD | wc -l` idiom used by coding agents to count unpushed commits: they see "1" instead of "0" and can waste turns re-investigating or re-pushing already-pushed branches.
- Fix: only call `println!` when the filtered output is non-empty, matching the pattern already used elsewhere in this file (e.g. `run_add`) for optional/empty output.

## Test plan

- [x] `cargo test --all --no-fail-fast` — 3781 passed, 1 failed: `copilot_selfheal_test::unwritable_hooks_dir_never_breaks_the_hook`, which relies on a read-only directory and cannot fail as root in a container; it does not go through `git log`. The two timestamp-based `dotnet_trx` tests are flaky under the container's mtime resolution and fail on `develop` as well.
- [x] Regression test in `tests/git_log_empty_range_test.rs`: builds a real temp repo, runs the `rtk` binary cargo builds for the suite (`CARGO_BIN_EXE_rtk`) against `HEAD..HEAD`, and asserts stdout is empty. Runs under a plain `cargo test`. Verified on current `develop`: fails without the fix (`"\n"`), passes with it.

Fixes #3365

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -498,7 +498,11 @@ fn run_log(
     // Post-process: truncate long messages, cap lines only if RTK set the default
     let filtered = filter_log_output(&result.stdout, limit, user_set_limit, has_format_flag);
     let filtered = never_worse(&result.stdout, &filtered).to_string();
-    println!("{}", filtered);
+    // Empty range (e.g. HEAD..HEAD) must produce empty output, not a bare
+    // newline — `println!` on an empty string still breaks `| wc -l` (#3365).
+    if !filtered.is_empty() {
+        println!("{}", filtered);
+    }
 
     timer.track(
         &format!("git log {}", args.join(" ")),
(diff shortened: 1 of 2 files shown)


Hirakhyzer:
The production change correctly avoids emitting a newline for an empty range while preserving existing non-empty output.

One test-coverage concern: `test_git_log_empty_range_produces_no_output` is marked `#[ignore]` and depends on a manually built `target/debug/rtk`, so the normal CI suite cannot catch this regression returning.

Could this be moved into an integration test under `tests/` and use `env!("CARGO_BIN_EXE_rtk")`, following the existing CLI integration-test pattern? That would remove the manual `cargo build` requirement and let ordinary `cargo test` and CI run the regression automatically.

Hydr0gen19:
Oh i see. It now runs under a plain cargo test with no manual build step.

Hydr0gen19:
Still current on `develop` as of today: the bare `println!` this PR guards sits at `src/cmds/git/git.rs:508` — it has drifted a few lines since I opened this — and the branch still merges cleanly against it, so there is nothing to rebase.

The same defect turned up independently in a fork adoption batch, [kylehgc/rtk#101](https://github.com/kylehgc/rtk/issues/101), described there the way it is here: `git log A..HEAD | wc -l` returning 1 instead of 0. Flagging it as a second person hitting the bug, not to hurry the queue along.

Issue #3408, opened 2026-08-04
`yadm <subcommand>` is rewritten to `rtk git`, so yadm commands act on the current project repo instead of the dotfiles repo — and bare `git add` becomes `git add .`


HetCreep:
Two rewrite-layer defects that combine into a data-integrity problem: commands aimed at one repository are executed against another, and a no-op form of `git add` is turned into a whole-tree stage.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Both behaviours below were reproduced by running the shipped `rtk 0.44.2` binary — commands and outputs are verbatim.*

### 1. `yadm` is rewritten to `git`

`src/discover/rules.rs:40-42` has one rule whose `rewrite_prefixes` are `["git", "yadm"]` while its `rtk_cmd` is `"rtk git"`. `rtk git` then executes the literal `git` binary (`src/cmds/git/git.rs:39`, `resolved_command("git")`) in the current working directory.

Observed:

```
$ rtk rewrite yadm status        →  rtk git status
$ rtk rewrite yadm commit -m x   →  rtk git commit -m x
$ rtk rewrite yadm add .         →  rtk git add .
$ rtk rewrite yadm list          →  rtk yadm list      ← correct
```

The whole point of `yadm` is that it manages a *separate* dotfiles repository whose work-tree is `$HOME`, while `git` in the same shell manages whatever project you are standing in. Rewriting `yadm commit` to `git commit` silently redirects the operation to the wrong repository.

`yadm list` going to `rtk yadm list` shows the intended path exists — `src/filters/yadm.toml` (`match_command = "^yadm\b"`) handles it. The rule at `rules.rs:40` shadows that filter for exactly the subcommands it enumerates, which include `commit`, `push`, `checkout` and `add`.

### 2. Bare `git add` gains a `.`

`src/cmds/git/git.rs:938-939`:

```rust
if args.is_empty() {
    cmd.arg(".");
}
```

Reproduced in a scratch repository containing an untracked `.env`:

```
$ git add                      # real git
Nothing specified, nothing added.
$ git diff --cached --name-only
                               # (nothing staged)

$ rtk git add
$ git diff --cached --name-only
.env
tracked.txt
```

Real `git add` with no pathspec is a deliberate no-op that prints guidance. Through rtk it stages everything, including files the user had not tracked.

### 3. Together

[...]

PR #3424, opened 2026-08-04
fix(filters): repair never-activating filters and make the test harness check match_command

### LINKED ISSUE #3402: `gradle`, `dotnet-build` and `uv-sync` filters can never activate — and `rtk verify` reports them green because the test harness never exercises `match_command`
Three of the 63 TOML filters cannot fire. Their inline tests all pass, so nothing surfaces the problem — the test harness validates the transformation but never the matching.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. All 63 filter files were read; the regex claims were verified by executing the patterns. No wrapped tool was run.*

### Why the tests cannot catch it

`src/core/toml_filter.rs:761`:

```rust
for test in tests {
    let actual = apply_filter(compiled, &test.input);
```

The harness feeds `test.input` straight into `apply_filter`. **`match_command` is never evaluated.** So a filter whose pattern can never match a real command still reports every test green, in `rtk verify` and in CI.

### 1. `gradle.toml:3` — the pattern requires the literal text twice

```toml
match_command = "^(gradle|gradlew|\./)gradlew?\b"
```

The alternation consumes `gradle`/`gradlew`/`./`, and then `gradlew?` demands *another* `gradle` or `gradlew`. Verified by executing the pattern:

- `gradle build` → **no match**
- `gradlew build` → **no match**
- `./gradlew build` → matches

And even that last case does not reach the filter: `src/main.rs:1306-1309` takes `file_name()` before lookup, which strips `./`, and `Commands::Gradlew` (`main.rs:814` → `2411`) routes it to the Rust module first anyway.

Net: 35 lines of configuration and 3 always-green tests for a filter that cannot activate. The intended pattern is presumably `^(\./)?gradlew?\b`.

### 2 & 3. `dotnet-build.toml` and `uv-sync.toml` are shadowed by clap subcommands

`rtk dotnet build` is handled by `Commands::Dotnet` (`main.rs:282` → `1899-1900`, `dotnet_cmd::run_build`), and `rtk uv …` by `Commands::Uv` (`main.rs:780` → `2383`). Clap routes both before `run_fallback` ever consults the TOML filters, so neither definition is reachable — 64 and 37 lines respectively, plus 5 green tests.

### The guard that exists for exactly this is stale

`src/core/toml_filter.rs:315-326` already warns about this situation:

```rust
// Shadow warning: if match_command matches a Rust-handled command, this filter
// will never activate (Clap routes before run_fallback). Warn the author.
for cmd in RUST_HANDLED_COMMANDS {
    if match_regex.is_match(cmd) { ... }
}
```

But `RUST_HANDLED_COMMANDS` (`:250-300`) has drifted from the clap enum. Comparing the two, these subcommands exist in `Commands` but are **absent** from the list:

`dotnet`, `uv`, `gradlew`, `mvn`, `gt`, `sbt`, `oc`, `rg`, `jest`, `rake`, `rspec`, `rubocop`, `pest`, `ecs`, `pint`, and the `php*` family.

That is why `dotnet-build` and `uv-sync` were not caught by the mechanism written to catch them.

### Suggested directions (untested)

- **Fix the harness first** — have the test runner check `match_command` against a sample command (either a new `test.command` field, or by asserting the pattern matches at least one plausible invocation). Without that, this class recurs silently.
- **Derive `RUST_HANDLED_COMMANDS` from the clap enum** rather than maintaining it by hand, so it cannot drift again. If that is awkward, a test asserting every `Commands` variant name appears in the list would catch the drift.
- Then decide per filter: fix `gradle.toml`'s pattern, and delete `dotnet-build.toml` / `uv-sync.toml` (or keep them as documentation with a comment saying they are shadowed).

### One more in the same family

`src/filters/gcc.toml:3` is `^g(cc|\+\+)\b`. `\b` cannot follow `+`, so `g++ -o a main.cpp` does **not** match (verified). The `gcc` half works; the `g++` half is dead while `gcc.toml:2` describes the filter as handling "gcc/g++". Harmless — it falls through to raw passthrough — but the description overstates what it does.

RawNuke:
Closes #3402

## What was wrong

Three TOML filters could never fire, and the test harness validated only the transform (`apply_filter`) — never `match_command` — so `rtk verify` and CI reported them green.

1. **gradle.toml**: `^(gradle|gradlew|\./)gradlew?\b` demanded the literal text twice (`gradle build` and `gradlew build` never matched; only `./gradlew` did, which `file_name()` strips and which clap routes to the Rust `Gradlew` module before fallback anyway).
2. **gcc.toml**: `^g(cc|\+\+)\b` could never match `g++` — there is no word boundary after `+`.
3. **dotnet-build.toml / uv-sync.toml**: shadowed by clap (`Commands::Dotnet`, `Commands::Uv` route before `run_fallback` consults filters).

## The fix

- **gradle.toml**: `^gradle\b` — the half that actually reaches the TOML fallback (`gradlew` is clap-routed; verified with `RTK_TOML_DEBUG=1`: `rtk gradle build` matches the filter, `rtk gradlew build` never reaches it).
- **gcc.toml**: `^(gcc|g\+\+)(\s|$)` — `gcc -c main.c` and `g++ -o main main.cpp` both verified to match.
- **dotnet-build.toml / uv-sync.toml**: deleted — they could never activate; their inline tests were always-green noise.

## The harness now checks activation

`[[tests.X]]` accepts an optional `command` field. When present, the harness asserts the filter's `match_command` actually matches that command; a dead pattern fails instead of passing transform-only tests. The gradle and gcc tests carry activation commands, so this class can't recur silently.

## The drift guard

`RUST_HANDLED_COMMANDS` had drifted from the clap enum — that is why the existing shadow warning never caught dotnet-build and uv-sync. It is now completed (dotnet, uv, gradlew, mvn, gt, sbt, oc, rg, jest, rake, rspec, rubocop, pest, ecs, pint, php, phpunit, phpstan, paratest, cc-economics, session, telemetry, learn, run, hook, hook-audit, untrust, glab), and a new test derives every clap subcommand name from `Cli::command()` and asserts each is reserved — it flagged `telemetry` and `untrust` during development, and will flag any future variant the moment it is added.

## Verification

- `rtk verify`: 150/150 tests pass.
- `cargo test`: 2571 pass, 0 fail.
- `cargo clippy --all-targets`: clean.
- Manual: `gradle build`, `gcc -c main.c`, `g++ -o main main.cpp` all activate their filters; no spurious shadow warning.


KuSh:
I re-verified this PR's claims against current `develop` (`005eb9c5`) with stand-in tools, and the three filter findings all still hold:

```
$ rtk g++ -o main main.cpp
raw 137 bytes -> rtk 137 bytes          # gcc.toml never matches g++; \b cannot follow +

$ rtk dotnet build                       # stdout: clean build, "0 Warning(s) / 0 Error(s)"
fail dotnet build: 1 projects, 0 errors, 0 warnings (00:00:02.34)   # Rust module, not "ok (build succeeded)"

$ rtk uv sync
Resolved 42 packages in 123ms
Audited 42 packages in 0.05ms            # raw, not "ok (up to date)"
```

The `g++` half of `gcc.toml` is the one I'd have missed reading the pattern quickly, and the `Cli::command()` drift guard is the part of this PR I value most — a stale `RUST_HANDLED_COMMANDS` is exactly why `dotnet-build` and `uv-sync` were never warned about, and deriving it from the enum closes that for good. The optional `command` field on `[[tests.X]]` is the other one: it turns "a dead pattern passes its transform tests" into a failing test, which is the class rather than the instances.

Where I'd ask you to narrow, so the verified parts can land without waiting on the rest:

**1. Drop the `gradle.toml` hunk.** #2709 deletes that file instead of repairing it, which will give you a modify/delete conflict here. There is also a reason not to keep it alive: filters are ordered by filter *name* (`BTreeMap<String, TomlFilterDef>`) and first match wins, and `gradle` sorts before `spring-boot`. So any pattern that matches `gradle bootRun` takes that command away from `spring-boot.toml`. On #2709's earlier `^(\./)?(gradlew?)\b` head I measured that as 240 bytes containing the `ERROR … Application run failed` and `BUILD FAILED` lines becoming 6779 bytes containing neither — the head cut from `max_lines = 50` fills up with INFO first. `^gradle\b` matches `gradle bootRun` too, so I expect the same outcome, though I have not run it against this head. `^gradle\b` also matches `gradle-profiler benchmark` (`\b` sits happily before `-`), which pulls an unrelated tool into that same 50-line head cut.

**2. The counts then become 60 and 61.** This PR currently sets 61/62 (63 − 2, keeping gradle). With #2709 in, `src/filters/` starts at 62, so removing `dotnet-build` and `uv-sync` leaves 60, and the concat test 61.

**3. Consider splitting out the two fixes that aren't in the title.** `append_add_pathspecs` (bare `git add` staying a no-op, #3408) and dropping `yadm` from the git rewrite rule both look like real bugs to me — the `yadm` one especially, since rewriting `yadm status` to `rtk git` acts on the wrong repository. They're independent of the filter work and would be quicker to review and land on their own than bundled here.

That leaves this PR as `gcc.toml` + the two deletions + the harness `command` field + the drift guard, which is a tight, well-evidenced change. It'll need a rebase either way — it currently shows as conflicting.

Happy to do the rebase and the count arithmetic as a fixup on your branch if that's easier than redoing it yourself; equally happy to leave it with you. I've added the two extra holes I found in the same class to #3402, including the fact that `test_builtin_all_expected_filters_present` cannot see either `gcc` or `spring-boot`.

PR #3430, opened 2026-08-04
fix(benchmark): avoid negative curl/cargo cases that fail the benchmark job

ousamabenyounes:
## Problem

The shared `benchmark` CI job (`scripts/benchmark.sh`) fails with `BENCHMARK FAILED: N filter(s) produced more tokens than raw output` on `develop`:

- `cargo build`: `18 → 25` tokens for an already-small successful build.
- `curl json`: negative on and off against `https://mockhttp.org/json/1` — the endpoint is a network dependency and its payload varies, so the result is not reproducible.

Because this lives in the shared benchmark script, every branch based on `develop` inherits the red job regardless of its own diff.

## Fix

- `scripts/benchmark.sh`: serve deterministic local fixtures from a loopback `python3 -m http.server` for both curl and wget — fully offline, and the JSON is pretty-printed and served with a real `Content-Type: application/json` so `rtk wget` actually exercises JSON minification (`47 → 12`, +74%).
  - Binds `http.server 0` (kernel-picked port) and reads the chosen port back from the server log, so a busy port cannot fail the run.
  - `cleanup_net_fixtures` is failure-tolerant: it runs from an EXIT trap under `set -e`, where `[ -n "$PID" ] && kill …` aborted the handler whenever `kill` failed and leaked the fixture dir and downloads.
  - The wget case prints an explicit skip line instead of silently disappearing when no local server is available (`wget` rejects `file://`, so there is no offline fallback URL).
- `src/cmds/rust/cargo_cmd.rs`: guard the streamed and non-streamed cargo build success summaries, and the `cargo test` summary, with `core::guard::never_worse(raw, summary)`. The test-summary path extracts a private `compute_test_summary` and applies the guard at the call site.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets -- -D warnings && cargo test --all`
- [x] `./scripts/benchmark.sh`

**RED** — new cargo guard test on upstream `develop` (`32c83cc`):

```text
---- cmds::rust::cargo_cmd::tests::test_cargo_test_summary_uses_raw_when_summary_is_larger stdout ----
  left: Some("cargo test: 1 errors, 0 warnings (0 crates)\nerror[E0433]: failed to resolve: use of undeclared type `Foo`\n")
 right: Some("error[E0433]: failed to resolve: use of undeclared type `Foo`\n")
```

**RED** — benchmark script asks, driven against the pre-fix script:

```text
=== cleanup must complete under 'set -e' even when kill fails ===
RED: /tmp/tmp.jOVSnnPPUe NOT removed
=== server must come up even when port 8899 is taken ===
RED: server unusable
```

**RED** — benchmark job on upstream `develop` (`32c83cc`, [run 31976715084](https://github.com/rtk-ai/rtk/actions/runs/31976715084)):

```text
🔴 find --max 10   40 → 48 (-20%)     <- pre-existing, untouched by this PR
🔴 curl json       26 → 38 (-46%)
BENCHMARK FAILED: 2 filter(s) produced more tokens than raw output
```

**GREEN** — this branch:

```text
test_filter_cargo_build_success_uses_raw_when_summary_is_larger ... ok
test_cargo_build_stream_success_uses_raw_when_summary_is_larger ... ok
test_cargo_test_summary_uses_raw_when_summary_is_larger ... ok
2612 unit tests + 78 integration tests passed, 0 failed
cargo fmt --all -- --check: passed
cargo clippy --all-targets -- -D warnings: passed
test presence: passed
benchmark: 0 negative caused by this PR
=== cleanup must complete under 'set -e' even when kill fails ===
GREEN: fixture dir removed
=== server must come up even when port 8899 is taken ===
GREEN: server reachable at http://127.0.0.1:33793 (ephemeral port, 8899 still busy)
```

Remote CI on this branch matches the `develop` baseline except that the two negatives this PR targets are gone. The single remaining benchmark negative (`find --max 10`, `40 → 48`) is pre-existing and reproduces identically on `develop`'s own run for the same base commit.

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -346,7 +346,12 @@ bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 # ===================
 section "curl"
 if command -v curl &> /dev/null; then
-  bench "curl json" "curl -s https://mockhttp.org/json/1" "$RTK curl https://mockhttp.org/json/1"
+  CURL_JSON_FIXTURE=$(mktemp)
+  cat > "$CURL_JSON_FIXTURE" << 'JSONEOF'
+{"message":"Hello from RTK benchmark","status":"success","code":200}
+JSONEOF
+  bench "curl json" "curl -s file://$CURL_JSON_FIXTURE" "$RTK curl file://$CURL_JSON_FIXTURE"
+  rm -f "$CURL_JSON_FIXTURE"
   bench "curl text" "curl -s https://mockhttp.org/robots.txt" "$RTK curl https://mockhttp.org/robots.txt"
 fi
 
@@ -355,8 +360,8 @@ fi
 # ===================
 if command -v wget &> /dev/null; then
   section "wget"
-  bench "wget" "wget -qO- https://mockhttp.org/json/1" "$RTK wget https://mockhttp.org/json/1"
-  rm -f 1 2>/dev/null
+  bench "wget" "wget -qO- https://mockhttp.org/json" "$RTK wget https://mockhttp.org/json"
+  rm -f json 2>/dev/null
 fi
 
 # ===================
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -342,26 +342,73 @@ section "wc"
 bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 
 # ===================
-# curl
+# curl / wget — fully offline. mockhttp.org was both a network dependency and
+# non-deterministic (its /json output is random), which made the benchmark flaky.
+# Serve fixed fixtures locally instead. The JSON is pretty-printed on purpose so
+# `rtk` actually exercises JSON minification (a pre-minified body leaves nothing
+# to compact).
 # ===================
-section "curl"
-if command -v curl &> /dev/null; then
-  CURL_JSON_FIXTURE=$(mktemp)
-  cat > "$CURL_JSON_FIXTURE" << 'JSONEOF'
-{"message":"Hello from RTK benchmark","status":"success","code":200}
+NET_FIXTURE_DIR="$(mktemp -d)"
+readonly NET_HTTP_PORT=8899   # loopback-only server shared by curl + wget
+NET_HTTP_PID=""
+cleanup_net_fixtures() {
+  [ -n "$NET_HTTP_PID" ] && kill "$NET_HTTP_PID" 2>/dev/null
+  rm -rf "$NET_FIXTURE_DIR"
+  # `rtk wget <url>/data.json` saves to ./data.json (default basename); remove
+  # that download and any numbered duplicates from repeated runs.
+  rm -f data.json data.json.* 2>/dev/null
+}
+trap cleanup_net_fixtures EXIT
+
+cat > "$NET_FIXTURE_DIR/data.json" << 'JSONEOF'
+{
+    "message": "Hello from RTK benchmark",
+    "status": "success",
+    "code": 200,
+    "items": [
+        { "id": 1, "name": "first" },
+        { "id": 2, "name": "second" }
+    ]
+}
 JSONEOF
-  bench "curl json" "curl -s file://$CURL_JSON_FIXTURE" "$RTK curl file://$CURL_JSON_FIXTURE"
-  rm -f "$CURL_JSON_FIXTURE"
-  bench "curl text" "curl -s https://mockhttp.org/robots.txt" "$RTK curl https://mockhttp.org/robots.txt"
+cat > "$NET_FIXTURE_DIR/robots.txt" << 'TXTEOF'
+User-agent: *
+Disallow: /private/
+Allow: /
+Sitemap: https://example.com/sitemap.xml
+TXTEOF
+
+# Bring up a loopback HTTP server once so both curl and wget get real response
+# headers (`Content-Type: application/json`) — that's what lets rtk detect JSON
+# and minify it. file:// carries no headers, so it can't demonstrate that path.
+NET_HTTP_URL=""
+if command -v python3 &> /dev/null; then
+  ( cd "$NET_FIXTURE_DIR" && exec python3 -m http.server "$NET_HTTP_PORT" --bind 127.0.0.1 ) >/dev/null 2>&1 &
+  NET_HTTP_PID=$!
+  for _ in $(seq 1 25); do
+    curl -s -o /dev/null "http://127.0.0.1:$NET_HTTP_PORT/data.json" 2>/dev/null \
+      && { NET_HTTP_URL="http://127.0.0.1:$NET_HTTP_PORT"; break; }
+    sleep 0.2
+  done
 fi
 
-# ===================
-# wget
-# ===================
-if command -v wget &> /dev/null; then
+section "curl"
+if command -v curl &> /dev/null; then
+  if [ -n "$NET_HTTP_URL" ]; then
+    bench "curl json" "curl -s $NET_HTTP_URL/data.json" "$RTK curl $NET_HTTP_URL/data.json"
+    bench "curl text" "curl -s $NET_HTTP_URL/robots.txt" "$RTK curl $NET_HTTP_URL/robots.txt"
+  else
+    # No python3 to host a local server — fall back to file:// (no headers, so
+    # no JSON minification, but still fully offline and deterministic).
+    bench "curl json" "curl -s file://$NET_FIXTURE_DIR/data.json" "$RTK curl file://$NET_FIXTURE_DIR/data.json"
+    bench "curl text" "curl -s file://$NET_FIXTURE_DIR/robots.txt" "$RTK curl file://$NET_FIXTURE_DIR/robots.txt"
+  fi
+fi
+
+# wget — rtk doesn't accept file://, so it needs the loopback server above.
+if command -v wget &> /dev/null && [ -n "$NET_HTTP_URL" ]; then
   section "wget"
-  bench "wget" "wget -qO- https://mockhttp.org/json" "$RTK wget https://mockhttp.org/json"
-  rm -f json 2>/dev/null
+  bench "wget" "wget -qO- $NET_HTTP_URL/data.json" "$RTK wget $NET_HTTP_URL/data.json"
 fi
 
 # ===================
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -349,14 +349,20 @@ bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 # to compact).
 # ===================
 NET_FIXTURE_DIR="$(mktemp -d)"
-readonly NET_HTTP_PORT=8899   # loopback-only server shared by curl + wget
+# Server log lives outside the served directory so it is never itself served.
+NET_HTTP_LOG="$(mktemp)"
 NET_HTTP_PID=""
+# Every step is failure-tolerant: this runs from an EXIT trap under `set -e`, so
+# a single failing command (e.g. `kill` on a server that already died) would
+# otherwise abort the trap and leak the fixture dir and downloaded files.
 cleanup_net_fixtures() {
-  [ -n "$NET_HTTP_PID" ] && kill "$NET_HTTP_PID" 2>/dev/null
-  rm -rf "$NET_FIXTURE_DIR"
+  if [ -n "$NET_HTTP_PID" ]; then
+    kill "$NET_HTTP_PID" 2>/dev/null || true
+  fi
+  rm -rf "$NET_FIXTURE_DIR" "$NET_HTTP_LOG" || true
   # `rtk wget <url>/data.json` saves to ./data.json (default basename); remove
   # that download and any numbered duplicates from repeated runs.
-  rm -f data.json data.json.* 2>/dev/null
+  rm -f data.json data.json.* 2>/dev/null || true
 }
 trap cleanup_net_fixtures EXIT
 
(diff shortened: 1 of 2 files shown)


aeppling:
Hey @ousamabenyounes , thanks for this, actually @KuSh already opened an issue on mockhttp repo and this should be solved now that it is merged , https://github.com/jaredwray/mockhttp/pull/181  , so apparently we should get a stable /json/1

KuSh:
Still missing a release on mockhttp side for the stable endpoint to be available though

ousamabenyounes:
Rebased this PR onto the latest `develop` and rechecked the suggested upstream endpoint fix. The endpoint is now stable, but the real benchmark still reproduces the negative curl case (`38 → 61` tokens), so the deterministic local fixture remains necessary. The cargo guard also fixes the independent `18 → 25` negative case.

Validation: upstream local benchmark RED with the two affected negatives; focused cargo regressions GREEN (2/2); full local replay GREEN with 2,593 unit tests plus integrations, fmt, clippy, test-presence, benchmark (0 negatives), and changed-line coverage. Remote CI passes every check except the pre-existing unrelated `find --max 10` benchmark negative (`40 → 48`), which is identical on current `develop` run 31653356311 and this PR run 31701940437.

Force-pushed the rebased branch (`74e8221` → `6c7df92`).

KuSh:
mockhttp responses still aren't stable as there wasn't any release since the fix was merged

scripts/benchmark.sh
```diff
@@ -346,7 +346,12 @@ bench "wc" "wc Cargo.toml src/main.rs" "$RTK wc Cargo.toml src/main.rs"
 # ===================
 section "curl"
 if command -v curl &> /dev/null; then
-  bench "curl json" "curl -s https://mockhttp.org/json/1" "$RTK curl https://mockhttp.org/json/1"
+  CURL_JSON_FIXTURE=$(mktemp)
+  cat > "$CURL_JSON_FIXTURE" << 'JSONEOF'
+{"message":"Hello from RTK benchmark","status":"success","code":200}
```
KuSh: This doesn't exercise rtk minimizing JSON content as current (when stable) does

scripts/benchmark.sh
```diff
@@ -355,8 +360,8 @@ fi
 # ===================
 if command -v wget &> /dev/null; then
   section "wget"
-  bench "wget" "wget -qO- https://mockhttp.org/json/1" "$RTK wget https://mockhttp.org/json/1"
-  rm -f 1 2>/dev/null
+  bench "wget" "wget -qO- https://mockhttp.org/json" "$RTK wget https://mockhttp.org/json"
+  rm -f json 2>/dev/null
```
KuSh: This won't work, https://mockhttp.org/json output is random

KuSh (changes requested):
Two things:

1. The `mockhttp` part should not be needed even if frustrating atm. I hope they’ll release a new `mockhttp` version soon. But if not, `benchmark.sh` should be made fully offline, not just for JSON, and preserve everything that was tested before.

2. The cargo part looks OK to me but:
  - You should apply the same guards elsewhere in the file (CargoTestHandler for example)
  - It would also be great to fix the fact that we don’t compact tiny cargo success build output; that doesn’t seem too difficult.

ousamabenyounes:
Pushed a follow-up addressing both review points:

**Fully offline & deterministic** — dropped the remaining `mockhttp.org` calls (curl `robots.txt`, wget `/json`). Both curl and wget now hit a loopback `python3 -m http.server` serving fixed local fixtures, so there is no network dependency and no random response. curl falls back to `file://` if `python3` is unavailable.

**Exercises JSON minification** — the fixture is pretty-printed and served with a real `Content-Type: application/json`, so rtk actually compacts it: `wget 47 → 12 (+74%)`. That was the path the earlier `file://` fixture skipped.

Local replay against `target/release/rtk`:
```
✅ 35 good  ⚠️ 27 warn  🔴 0 negative  ❌ 0 fail    (exit 0)
```
The 56% good line is a non-fatal WARNING; the job only fails on negatives or empty output, both now 0.

Also applied the same `never_worse` guard to `CargoTestHandler::format_summary` so a tiny `cargo test` run can never produce a summary larger than its raw output (matching `CargoBuildHandler`). `cargo fmt --check` clean, cargo tests green.

scripts/benchmark.sh
```diff
@@ -342,21 +342,73 @@ section "wc"
...
-# wget
-# ===================
-if command -v wget &> /dev/null; then
+# wget — rtk doesn't accept file://, so it needs the loopback server above.
+if command -v wget &> /dev/null && [ -n "$NET_HTTP_URL" ]; then
   section "wget"
-  bench "wget" "wget -qO- https://mockhttp.org/json/1" "$RTK wget https://mockhttp.org/json/1"
-  rm -f 1 2>/dev/null
+  bench "wget" "wget -qO- $NET_HTTP_URL/data.json" "$RTK wget $NET_HTTP_URL/data.json"
 fi
```
KuSh, line 437: No fallback?

scripts/benchmark.sh
```diff
@@ -342,21 +342,73 @@ section "wc"
...
+# non-deterministic (its /json output is random), which made the benchmark flaky.
+# Serve fixed fixtures locally instead. The JSON is pretty-printed on purpose so
+# `rtk` actually exercises JSON minification (a pre-minified body leaves nothing
+# to compact).
+# ===================
+NET_FIXTURE_DIR="$(mktemp -d)"
+readonly NET_HTTP_PORT=8899   # loopback-only server shared by curl + wget
+NET_HTTP_PID=""
+cleanup_net_fixtures() {
+  [ -n "$NET_HTTP_PID" ] && kill "$NET_HTTP_PID" 2>/dev/null
```
KuSh: this runs under `set -e`, if anything fails, the cleanup won't be complete

src/cmds/rust/cargo_cmd.rs
```diff
@@ -195,6 +196,11 @@ impl BlockHandler for CargoTestHandler {
     }
 
     fn format_summary(&self, _exit_code: i32, raw: &str) -> Option<String> {
+        // Same never-worse guard as CargoBuildHandler (#3430 review): if the
+        // compacted summary ends up larger than the raw output (e.g. a tiny
+        // `cargo test` run), keep the raw output instead of "compacting" it
+        // into something bigger.
+        let summary = (|| -> Option<String> {
```
KuSh: this fails fmt and an IIFE closure shouldn't be needed.
Just extract the body in a private fn and apply the guard at the call site:

```rs
fn format_summary(&self, _exit_code: i32, raw: &str) -> Option<String> {
    let summary = self.compute_test_summary(raw)?;
    Some(crate::core::guard::never_worse(raw, &summary).to_string())
}

fn compute_test_summary(&self, raw: &str) -> Option<String> {
  ...

File: src/cmds/rust/cargo_cmd.rs [hunk 6]
```diff
@@ -250,6 +256,8 @@ impl BlockHandler for CargoTestHandler {
         }
 
         None
+        })();
+        summary.map(|s| crate::core::guard::never_worse(raw, &s).to_string())
```
KuSh: This isn't tested

scripts/benchmark.sh
```diff
@@ -342,21 +342,73 @@ section "wc"
...
-# curl
-# ===================
+# curl / wget — fully offline. mockhttp.org was both a network dependency and
+# non-deterministic (its /json output is random), which made the benchmark flaky.
+# Serve fixed fixtures locally instead. The JSON is pretty-printed on purpose so
+# `rtk` actually exercises JSON minification (a pre-minified body leaves nothing
+# to compact).
+# ===================
+NET_FIXTURE_DIR="$(mktemp -d)"
+readonly NET_HTTP_PORT=8899   # loopback-only server shared by curl + wget
```
KuSh: This will fail needlessly if port 8899 is already in use. Why not use `http.server 0` and then use the port it picked instead?

KuSh (changes requested):
Hi, this is going in the right direction, but it needs a bit more work

scripts/benchmark.sh
ousamabenyounes: Done — the server now binds `python3 -u -m http.server 0` and the benchmark reads the chosen port back from the server log ("Serving HTTP on 127.0.0.1 port NNNNN", printed once the socket is bound), so a busy port can no longer fail the run. `-u` is needed because that line would otherwise sit in a block buffer when stdout is a file.

Verified with 8899 deliberately occupied: before, `NET_HTTP_URL` stayed unusable; now the server comes up on an ephemeral port (`http://127.0.0.1:33793` in my run).

scripts/benchmark.sh
ousamabenyounes: Good catch. `[ -n "$NET_HTTP_PID" ] && kill ...` returns 1 whenever `kill` fails (server already dead), which under `set -e` aborted the trap before `rm -rf`. Every step is failure-tolerant now:

```sh
cleanup_net_fixtures() {
  if [ -n "$NET_HTTP_PID" ]; then
    kill "$NET_HTTP_PID" 2>/dev/null || true
  fi
  rm -rf "$NET_FIXTURE_DIR" "$NET_HTTP_LOG" || true
  rm -f data.json data.json.* 2>/dev/null || true
}
```

Reproduced with a PID whose process had already exited: before, the fixture dir was left behind; after, cleanup completes.

scripts/benchmark.sh
ousamabenyounes: There is no offline URL to fall back to here: `wget` rejects `file://` outright (`file:///tmp/x: Unsupported scheme.`), so unlike curl there is nothing local to point it at without the loopback server.

What I did fix is the silent disappearance — the block was `if command -v wget && [ -n "$NET_HTTP_URL" ]`, so the whole case vanished from the report. It now always prints a section and an explicit skip line, matching the style already used for cargo:

```
⏭️  wget (no local HTTP server available, skipped)
```

Happy to do something else if you had a specific fallback in mind.

src/cmds/rust/cargo_cmd.rs
ousamabenyounes: Done, exactly as suggested — `compute_test_summary` is a private fn on the existing `impl CargoTestHandler` and `never_worse` is applied at the call site:

```rs
fn format_summary(&self, _exit_code: i32, raw: &str) -> Option<String> {
    let summary = self.compute_test_summary(raw)?;
    Some(crate::core::guard::never_worse(raw, &summary).to_string())
}
```

(For the record `cargo fmt --all -- --check` passed on the IIFE version with rustfmt 1.93.1, but the closure was ugly either way — it is gone now.)

src/cmds/rust/cargo_cmd.rs
ousamabenyounes: Added `test_cargo_test_summary_uses_raw_when_summary_is_larger`. It uses a one-line compile failure, where the compacted form is genuinely bigger than the raw output:

- raw: `error[E0433]: failed to resolve: use of undeclared type ` + "`Foo`" + ` ` (62 B)
- unguarded summary: `cargo test: 1 errors, 0 warnings (0 crates)` + the same error line (105 B)

The test asserts both halves — that the unguarded summary really is larger (so the case still exercises the guard if the header ever changes), and that `format_summary` returns the raw output. It fails on `develop` (returns the 105 B form) and passes here.

ousamabenyounes:
Thanks for the detailed review @KuSh — all five points are addressed in `7f6156f` + `d003920` (rebased onto `develop` `32c83cc`, force-push `c4ac631` → `d003920`). Replies are in each thread; summary:

**`scripts/benchmark.sh`**
- **Fixed port 8899** → `python3 -u -m http.server 0`, and the benchmark reads the kernel-picked port back from the server log. `-u` is required or that line stays block-buffered when stdout is a file.
- **Cleanup under `set -e`** → `[ -n "$PID" ] && kill …` returned 1 whenever `kill` failed, aborting the trap before `rm -rf`. Every step is now `|| true` / `if`-guarded.
- **wget "no fallback"** → `wget` rejects `file://` outright (`Unsupported scheme.`), so there is no offline URL to fall back to. What I did fix is that the case silently vanished from the report; it now always prints a section plus `⏭️  wget (no local HTTP server available, skipped)`. Left that thread open in case you had a different fallback in mind.

**`src/cmds/rust/cargo_cmd.rs`**
- **IIFE** → replaced by a private `compute_test_summary` on the existing `impl CargoTestHandler`, guard applied at the call site, exactly as you sketched.
- **Untested guard** → `test_cargo_test_summary_uses_raw_when_summary_is_larger`: a one-line compile failure whose unguarded summary (105 B) is larger than the raw output (62 B). It asserts both that the unguarded summary really is larger and that `format_summary` returns raw. Fails on `develop`, passes here.

**Validation**

RED (before), on the shell asks:
```
=== cleanup must complete under 'set -e' even when kill fails ===
RED: /tmp/tmp.jOVSnnPPUe NOT removed
=== server must come up even when port 8899 is taken ===
RED: server unusable
```
RED (before), on the cargo ask, run on upstream `develop`:
```
---- test_cargo_test_summary_uses_raw_when_summary_is_larger stdout ----
  left: Some("cargo test: 1 errors, 0 warnings (0 crates)\nerror[E0433]: ...")
 right: Some("error[E0433]: ...")
```
GREEN (after), full local CI replay:
```
PASS  test presence
PASS  fmt
PASS  clippy
PASS  test           2612 unit + 78 integration, 0 failed
PASS  build release
benchmark            0 negative (develop baseline on 32c83cc: 2 negative)
```
Remote CI on `d003920`: everything green except `benchmark`, which is down to the single **pre-existing** `find --max 10` negative (`40 → 48`) — identical on develop's own run for the same base commit ([run 31976715084](https://github.com/rtk-ai/rtk/actions/runs/31976715084), which additionally had the `curl json` negative this PR removes).

**Unrelated heads-up** while I was in there: `benchmark.sh` sets `BENCH_DIR="$(pwd)/scripts/benchmark"` and does `rm -rf "$BENCH_DIR"` when `$CI` is unset — so running it locally deletes the tracked `scripts/benchmark/*.ts` VM-harness. Not touching it here to keep this PR to one concern, but happy to open a separate PR if useful.

KuSh (approved):
LGTM, thanks! Don’t hesitate to push a follow-up PR to fix the broken `BENCH_DIR` cleanup

PR #3435, opened 2026-08-04
fix(find/git): avoid UTF-8 mid-char byte-slice panics

### LINKED ISSUE #3415: `rtk find` panics on a non-ASCII directory path >50 bytes, and `rtk git commit` panics when the first output line does not start with `[`
I swept the repository for the byte-slice class discussed in #2954 (`&str` sliced at a computed index behind a byte-length guard). Two live sites remain, both reproduced.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Both reproduced by compiling the two functions verbatim with `rustc 1.97.1` and running them.*

### 1. `src/cmds/system/find_cmd.rs:326` — non-ASCII directory paths

```rust
let dir_display = if dir.len() > 50 {
    format!("...{}", &dir[dir.len() - 47..])
} else {
    dir.clone()
};
```

`dir.len()` is bytes; `dir.len() - 47` is a byte index. For any path longer than 50 bytes whose byte at that offset is inside a multibyte character, the slice panics.

```
ascii 60-byte dir      : ok
"ก" x20  (60 bytes)    : PANIC
"日" x20 (60 bytes)    : PANIC
```

Twenty CJK or Thai characters is a perfectly ordinary directory name, and 50 bytes is short — any non-ASCII path a little way down a tree reaches it. This is a display-formatting line, so the whole `rtk find` invocation dies for a cosmetic truncation.

### 2. `src/cmds/git/git.rs:1013` — first line of `git commit` output

```rust
/// … localized variants, and multibyte branch names.
fn parse_commit_output(line: &str) -> String {
    if let Some(bracket_end) = line.find(']') {
        let bracket_content = &line[1..bracket_end];
```

Two unchecked assumptions: that byte 1 is a character boundary, and that `bracket_end >= 1`.

```
"[main abc1234def] msg" : ok
"] done"                : PANIC   (range 1..0)
"日本] x"                : PANIC   (byte 1 not a char boundary)
```

The doc comment directly above says the function handles "multibyte branch names", but `&line[1..]` is exactly what cannot. The four unit tests all feed lines beginning with `[`.

Reachability: this runs on the first stdout line of a **successful** `git commit` (via `classify_commit_outcome`, `git.rs:1078`). Anything a `pre-commit` hook prints to stdout lands there first. I did not manage to trigger it through the CLI on this machine — git routed my test hook's output to stderr — so treat the reachability as plausible rather than demonstrated; the panic itself is demonstrated.

Note the correct idiom is already two lines below, at `git.rs:1016`: `hash.chars().take(7)`.

### Suggested fixes (untested — no programmer has reviewed these)

Both have an in-tree model:

- `find_cmd.rs:326` → `floor_char_boundary` as used at `src/analytics/session_cmd.rs:113` and `src/analytics/gain.rs:730`, or count characters instead of bytes as `src/core/utils.rs:29` does.
- `git.rs:1013` → check `bracket_end > 1` and use a character-aware slice; or skip the leading `[` with `strip_prefix('[')` rather than a byte index.

### On the class as a whole

I checked the other ~30 `&str` slice sites in `src/` and they are safe by construction — `floor_char_boundary` (`gain.rs:730`, `session_cmd.rs:113`, `json_cmd.rs:112`, `pipe_cmd.rs:171`), explicit boundary walks (`tee.rs:135-140`, `curl_cmd.rs:141-143`), regex `Match::start()/end()` (`ls.rs:157-159`), ASCII-only offsets (`wc_cmd.rs:252`, `pnpm_cmd.rs:161`), or input pre-mapped to ASCII (`search.rs:116-120`). So this is not a widespread pattern; it is two remaining sites.

Two further sites are byte-length-guarded but reachable only through data rtk wrote itself (`analytics/gain.rs:464`, `core/display_helpers.rs:205,210` — all fed by `%Y-%m-%d` SQLite columns). They panic on non-ASCII input in isolation but I could not construct a real path to them, so I am listing rather than filing them.

As mentioned in #3394, `clippy::indexing_slicing` / `clippy::string_slice` are not configured anywhere, which is why new instances are not caught at review time.

mohammedelkarsh:
## Summary
- Fix `rtk find` display truncation so paths longer than 50 bytes truncate on char boundaries (Thai/CJK paths no longer panic).
- Fix `parse_commit_output` to require a leading `[` via `strip_prefix` instead of `&line[1..]`, so hook noise / multibyte first lines cannot panic.
- Add regression tests for ASCII truncation parity and the two panic shapes from #3415.

Fixes #3415

## Test plan
- [x] `cargo test --bin rtk truncate_path_display`
- [x] `cargo test --bin rtk test_parse_commit_output`
- [x] `cargo fmt --all`
- [x] `cargo clippy --bin rtk --all-targets -- -D warnings`

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -1009,15 +1009,20 @@ fn build_commit_command(args: &[String], global_args: &[String]) -> Command {
 /// Handles: `[main abc1234def] message`, `[main (root-commit) abc1234def] msg`,
 /// localized variants, and multibyte branch names.
 fn parse_commit_output(line: &str) -> String {
-    if let Some(bracket_end) = line.find(']') {
-        let bracket_content = &line[1..bracket_end];
-        let hash = bracket_content.split_whitespace().next_back().unwrap_or("");
-        if !hash.is_empty() && hash.len() >= 7 {
-            let short_hash: String = hash.chars().take(7).collect();
-            format!("ok {}", short_hash)
-        } else {
-            "ok".to_string()
-        }
+    // Require a leading `[` via strip_prefix so we never byte-slice at index 1
+    // when the first character is multibyte, or when `]` appears before any `[`
+    // (e.g. hook noise). See #3415.
+    let Some(after_open) = line.strip_prefix('[') else {
+        return "ok".to_string();
+    };
+    let Some(bracket_end) = after_open.find(']') else {
+        return "ok".to_string();
+    };
+    let bracket_content = &after_open[..bracket_end];
+    let hash = bracket_content.split_whitespace().next_back().unwrap_or("");
+    if !hash.is_empty() && hash.len() >= 7 {
+        let short_hash: String = hash.chars().take(7).collect();
+        format!("ok {}", short_hash)
     } else {
         "ok".to_string()
     }
@@ -2867,6 +2872,18 @@ no changes added to commit (use "git add" and/or "git commit -a")
     #[test]
     fn test_parse_commit_output_empty() {
         assert_eq!(parse_commit_output(""), "ok");
+    }
+
+    /// Regression #3415: `]` with no leading `[` used to slice `1..0` and panic.
+    #[test]
+    fn test_parse_commit_output_leading_close_bracket() {
+        assert_eq!(parse_commit_output("] done"), "ok");
+    }
+
+    /// Regression #3415: multibyte text before `]` must not byte-slice at index 1.
+    #[test]
+    fn test_parse_commit_output_multibyte_without_open_bracket() {
+        assert_eq!(parse_commit_output("日本] x"), "ok");
     }
 
     // --- commit outcome classification (issue #2494) ---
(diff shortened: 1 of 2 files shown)


KuSh:
Thanks for chasing both panics and for pinning them with tests. Issue #3415 is closed as completed: parse_commit_output was rewritten in #2717 (merged 2026-08-13) so both bracket indices come from str::find and can never land mid-character, and it now carries eleven tests including the exact inputs from the report, among them "] stray [main abc1234def]" and the multibyte prefixes. That version also keeps the short hash when git prints hook output ahead of its summary, a case the strip_prefix approach here would report as a bare "ok". The find side was rewritten in #3603 (merged 2026-08-26): the directory label now truncates at 50 characters rather than 50 bytes, so the mid-character slice is gone, and there is an end-to-end test for a long non-ASCII directory name. What develop still lacks is a unit test pinning the shape of that truncated label, as the issue reporter noted, and a small test there would be welcome. Please comment or reopen if something was missed.

PR #3439, opened 2026-08-04
feat(rewrite): translate PowerShell cmdlets to rtk subcommands

make0uthill:
## Summary

The `RULES` table keys off the command word, so PowerShell cmdlets — which share no name with their GNU counterparts — never match it. On Windows the agent reaches for `Get-ChildItem` and `Select-String` rather than `ls` and `grep`, so those calls reach the model as unfiltered output and the hook saves nothing.

Translates the three cmdlets whose semantics map cleanly onto an rtk subcommand, plus their common aliases (`gc`, `gci`, `dir`, `sls`):

```
Get-Content foo.txt -TotalCount 5      -> rtk read foo.txt --max-lines 5
Get-ChildItem src -Recurse             -> rtk ls -R src
Select-String -Path a.txt -Pattern rtk -> rtk grep -i rtk a.txt
```

Two details worth flagging for review:

- The `-i` on `Select-String` is not cosmetic: the cmdlet matches case-insensitively by default while `grep` does not, so omitting it would silently narrow the result set.
- Argument splitting cannot reuse `shell_split`, which treats a backslash as an escape and would turn `D:\Code\x` into `D:Codex`. `ps_split` keeps backslashes literal and preserves quoting.

Deliberately narrow. A cmdlet carrying any parameter without a faithful rtk equivalent (`-Filter`, `-Raw`, `-File`, `-ErrorAction`) returns `None` and runs unrewritten — mistranslating a flag changes what the command *means*, which is worse than forgoing the saving. Cmdlets are also only rewritten in `RewriteContext::Normal`: as a pipeline stage a cmdlet hands .NET objects to the next stage, and rtk's text output would not survive that.

Known gap, left for a follow-up: `Get-ChildItem -Recurse -Filter *.rs` could map faithfully onto `rtk find . -name '*.rs'` (and `-File`/`-Directory` onto `-type f`/`-type d`). Happy to add it here instead if you'd prefer it in one PR.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — 2562 passed, 0 failed
- [x] 8 new tests covering the translations, the pass-through cases, Windows paths, and pipeline stages
- [x] Manual testing: verified end-to-end through `rtk hook claude` with a `PowerShell` tool payload; rewritten and passed-through cases both behave as the tests describe.

Targets `develop` as required.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -88,6 +88,268 @@ static TAIL_LINES_EQ: LazyLock<Regex> =
 static TAIL_LINES_SPACE: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^tail\s+--lines\s+(\d+)\s+(\S+)$").unwrap());
 
+/// Splits a PowerShell argument list on whitespace, honouring single and double
+/// quotes. Unlike [`shell_split`] a backslash is a literal character, not an
+/// escape — on Windows it is a path separator, and consuming it would turn
+/// `D:\Code\x` into `D:Codex`. Quotes are kept in the returned tokens so a
+/// rewrite can re-emit the argument exactly as the caller wrote it.
+fn ps_split(input: &str) -> Vec<String> {
+    let mut tokens = Vec::new();
+    let mut current = String::new();
+    let mut quote: Option<char> = None;
+
+    for c in input.chars() {
+        match quote {
+            Some(q) => {
+                current.push(c);
+                if c == q {
+                    quote = None;
+                }
+            }
+            None if c == '"' || c == '\'' => {
+                quote = Some(c);
+                current.push(c);
+            }
+            None if c.is_whitespace() => {
+                if !current.is_empty() {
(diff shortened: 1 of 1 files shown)


make0uthill:
Pushed the follow-up: `-Filter`/`-Include`/`-File`/`-Directory` now route to `rtk find` instead of passing through, so the known gap noted above is closed in this PR rather than a separate one.

```
Get-ChildItem src -Recurse -Filter *.rs -> rtk find src -name '"*.rs"'
Get-ChildItem -Recurse -File            -> rtk find . -type f
```

Without `-Recurse` the search is depth-limited with `-maxdepth 1`, since `Get-ChildItem` only descends when asked while `find` recurses by default. `-Include` with a comma-separated list still passes through.

Full suite green, 2571 passed.

make0uthill:
Correction to my note above: the suite is **2563 passed, 0 failed** (I quoted 2563 as 2571 by mistake). Nothing else in that comment changes.

PR #3449, opened 2026-08-05
fix(hooks): copilot self heal dual hooks (drop camelCase entry)

aeppling:
## Summary

- Self-heal stale Copilot hook configs left by pre-#3212 `rtk init --copilot`: on a legacy camelCase `preToolUse` invocation, remove rtk's own duplicate camelCase entry from `rtk-rewrite.json` (project + global) so upgrades need no manual re-init
- Surgical and fail-safe: only the exact stock rtk entry is removed, user hooks/keys/order preserved, PascalCase entry required before touching anything; any other shape, IO failure, or malformed file is a silent no-op
- Atomic write (temp + rename), no effect on healthy installs; code path only reachable via the stale registration itself

## Test plan

- [x] `tests/copilot_selfheal_test.rs`: 13 real-binary e2e tests (sandboxed HOME/COPILOT_HOME); heal correctness, 10-case refusal matrix, response byte-identical pre/post heal, concurrency, unwritable dir, exit 0 + silent stderr everywhere
- [x] Manual testing: stale/healthy/JetBrains payloads piped to `rtk hook copilot`, configs and responses inspected


src/hooks/hook_cmd.rs
```diff
@@ -145,6 +150,87 @@ fn detect_format(v: &Value) -> HookFormat {
...
+            healed.push(global);
+        }
+    }
+    healed
+}
+
+// Exact camelCase entry written by pre-b754b85 `rtk init --copilot`; that
+// stale registration is the only thing routing invocations into the
+// CopilotCli arm above. Only this entry is removed — user additions stay.
+fn legacy_camelcase_entry() -> Value {
```
KuSh, line 175: Nitpick: use OnceLock for lazy init

src/hooks/hook_cmd.rs
```diff
@@ -145,6 +150,87 @@ fn detect_format(v: &Value) -> HookFormat {
...
+    }
+    let pascalcase_still_registered = hooks
+        .get("PreToolUse")
+        .and_then(|p| p.as_array())
+        .is_some_and(|entries| {
+            entries
+                .iter()
+                .any(|e| e.get("command").and_then(|c| c.as_str()) == Some("rtk hook copilot"))
+        });
+    if !pascalcase_still_registered {
```
KuSh, line 206: So in that case instead of fixing the hook we use the legacy one? I'm not sure this could happen but it's a strange behaviour, no?

KuSh (approved):
LGTM, just a nitpick and a question (non blocking)

src/hooks/hook_cmd.rs
aeppling: if there's no PascalCase entry and we drop the camelCase it left no hook at all, and if this state exist it may be because a user modified the file itself so maybe better to not touch at all ?

KuSh: Yes that's perhaps the best move

KuSh (approved):

PR #3151, opened 2026-07-22
feat(hooks): add OMP (Oh My Pi) agent support

DavidHLP:
## Summary

- Add `rtk init --agent omp` to install RTK transparent rewrite support for [OMP (Oh My Pi)](https://omp.sh) coding agent
- TypeScript extension loaded from `hooks/pre/` that intercepts `tool_call` events and calls `rtk rewrite` for in-place command mutation — same architecture as the existing Pi extension
- Adds OMP as the 16th supported agent across `README.md`, `hooks/README.md`, and `docs/guide/getting-started/supported-agents.md`

Closes #591.

> ℹ️ **Note on #1365**: This PR is an independent OMP implementation that supersedes #1365 (stale since June 2025). Both close #591. This PR uses `--agent omp` (consistent with Pi/Hermes/Droid) rather than #1365's `--omp` flag.

> ℹ️ **Conflict note for maintainers**: This PR is intentionally left **DIRTY (conflicting)**. The conflict is `README.md` only — `develop` reworded the "Supported AI Tools" intro while this PR bumped the count 15 → 16. Resolution: keep **16**, adopt `develop`'s wording (`reducing the bash output the agent reads`). The other two overlapping files auto-merge cleanly. `develop` was not merged into this branch to keep the commit history focused on the OMP feature.

---

## Install / Uninstall

```bash
# Install
rtk init -g --agent omp          # global → ~/.omp/agent/hooks/pre/rtk.ts
rtk init --agent omp             # project → .omp/hooks/pre/rtk.ts

# Uninstall
rtk init --uninstall -g --agent omp   # global
rtk init --uninstall --agent omp      # project

# Preview
rtk init --agent omp --dry-run
```

Both install and uninstall honor the `OMP_AGENT_DIR` environment variable (mirrors `PI_CODING_AGENT_DIR`) for custom install locations.

---

## Test plan

- [x] `cargo fmt --all --check` — pass
- [x] `cargo clippy --all-targets` — no issues
- [x] `cargo test` — 2535 passed, 0 failed (14 new OMP tests)
- [x] Manual testing: `rtk init --agent omp` install/uninstall/dry-run verified
- [x] End-to-end: OMP 17.0.8 loads the extension and rewrites commands

### Runtime evidence

Environment: `rtk 0.42.4`, `omp 17.0.8`, Linux x86_64.

**1. `rtk rewrite` on OMP-rewritable commands:**

```
$ rtk rewrite "git status"     → rtk git status
[...]

PR #3453, opened 2026-08-06
fix(hook): return revised input from Pi tool_call handler

thinhngotony:
## Summary

- Fixes the merged Pi extension (`hooks/pi/rtk.ts`, #1741, released in rtk 0.42) so `rtk rewrite` results actually take effect — today the extension never rewrites a single command.
- Root cause: the Pi/OMP `tool_call` dispatcher builds the args a tool actually executes with from a separate copy of `event.input`; it only substitutes a handler's **returned** `{ input }` (`ToolCallEventResult`). The handler mutated `event.input.command` in place and returned nothing, so the rewrite was silently discarded on every call.
- Fix: `return { input: { ...event.input, command: rewritten } }` instead of mutating, preserving the other `BashToolInput` fields via spread.
- Updates `hooks/pi/README.md`'s "Specifics" section, which documented the broken mutate-in-place mechanism as the intended design.
- Adds a regression test (`test_pi_plugin_returns_revised_input_instead_of_mutating_in_place`) asserting the shipped plugin source returns the revision and does not rely on mutating `event.input` in place.

## Why this matters

This is the same file discussed in the [Oh My Pi integration](https://github.com/can1357/oh-my-pi) and [rtk#591](https://github.com/rtk-ai/rtk/issues/591). That integration's review discussion assumed "mutate `event.input.command`" was a working mechanism. It isn't in the affected OMP runtime, per its `ExtensionAPI` behavior and confirmed empirically: after applying this exact fix locally and running a real `omp -p --auto-approve "git status"` session, `rtk`'s own `history.db` recorded `git status → rtk git status`; before the fix, nothing was ever rewritten despite the extension loading and probing successfully.

## Test plan

- [x] `cargo test` (2563 passed, 0 failed, incl. new regression test)
- [x] `bun --eval 'await import("./hooks/pi/rtk.ts")'`
- [x] Manual end-to-end: built a fresh `omp` session (Oh My Pi, the extension consumer for #591), installed this exact fixed source as its `tool_call` handler, ran `git status` through the agent's bash tool. `rtk`'s local history.db recorded the command as `original_cmd: git status`, `rtk_cmd: rtk git status` — confirming the **executed** shell command was rewritten, not just displayed.

Related: #401, #591, #1741

### DIFF
--- a/hooks/pi/README.md
+++ b/hooks/pi/README.md
@@ -16,7 +16,7 @@ with other Pi extensions.
 
 - TypeScript extension using Pi's `ExtensionAPI` (not a shell hook, no `zx` dependency)
 - Subscribes to `tool_call` event, narrows to `bash` tool via `isToolCallEventType`
-- Calls `rtk rewrite` via `pi.exec`; mutates `event.input.command` in-place if rewrite differs
+- Calls `rtk rewrite` via `pi.exec`; returns a revised `{ input }` (not an in-place mutation of `event.input`) when the rewrite differs, per `ToolCallEventResult`
 - All error paths return `undefined` (pass through); RTK never blocks execution
 - Version guard at load time: checks `rtk >= 0.23.0`; warns and registers no-op if too old or missing
 - Installed to `.pi/extensions/rtk.ts` by `rtk init --agent pi` (project-local) or `~/.pi/agent/extensions/rtk.ts` by `rtk init --agent pi --global`
--- a/hooks/pi/rtk.ts
+++ b/hooks/pi/rtk.ts
@@ -6,9 +6,9 @@
 // To add or change rewrite rules, edit the Rust registry — not this file.
 //
 // Exit code contract for `rtk rewrite`:
-//   0 + stdout  Rewrite found → mutate command
+//   0 + stdout  Rewrite found → revised command returned to the caller
 //   1           No RTK equivalent → pass through unchanged
-//   3 + stdout  Rewrite (advisory) → mutate command
+//   3 + stdout  Rewrite (advisory) → revised command returned to the caller
 
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent"
 import { isToolCallEventType } from "@earendil-works/pi-coding-agent"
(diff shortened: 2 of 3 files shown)


thinhngotony:
@aeppling @xavierpestel-ai Can you guys help me to review this?

zktaiga:
This is a no-op on pi. `ToolCallEventResult` has no `input` field, and the doc comment says what to do instead:

https://github.com/earendil-works/pi/blob/709aa03194301afd008a07d64ff1bf12e4f7ece6/packages/coding-agent/src/core/extensions/types.ts#L1071-L1074

The dispatcher only ever reads `.block` off the handler result — anything else on the returned object is discarded:

https://github.com/earendil-works/pi/blob/709aa03194301afd008a07d64ff1bf12e4f7ece6/packages/coding-agent/src/core/extensions/runner.ts#L932-L952

`event.input` is the same object that reaches execution, which is why mutating it works today:

https://github.com/earendil-works/pi/blob/709aa03194301afd008a07d64ff1bf12e4f7ece6/packages/coding-agent/docs/extensions.md?plain=1#L759

thinhngotony:
Thanks for the correction. You’re right about upstream Pi: `ToolCallEventResult` has no `input` field, and upstream executes the same `event.input` object, so mutating `event.input.command` is the correct approach there.

This PR targets [Oh My Pi](https://github.com/can1357/oh-my-pi), whose dispatcher appears to differ. In the OMP version tested, mutating `event.input.command` did not change the executed command, while returning `{ input: ... }` did. I’ll verify this against the exact OMP commit and update the PR to target the correct runtime behavior.

I’ll also fix the misleading `#1365` reference. The relevant integration context is [rtk#591](https://github.com/rtk-ai/rtk/issues/591).

madsmadsen:
Wouldn't it then make sense to get the fix into https://github.com/rtk-ai/rtk/pull/3151 and help get that into rtk?

thinhngotony:
Closing in favor of #3151, which adds the dedicated OMP integration. The returned `{ input }` rewrite should be applied to `hooks/omp/rtk.ts` there, not `hooks/pi/rtk.ts`, because upstream Pi still uses in-place mutation.

PR #3456, opened 2026-08-06
feat(grep): add agent-safe and JSON output modes

isink17:
## Summary

- add opt-in agent-safe grep output with explicit total-match, per-file, context-row, and line-length limits
- add files-only, count-by-file, top-files, and stable JSON result modes
- preserve existing grep defaults, native rg behavior, argument ordering, stdin handling, and child exit codes

## Compatibility

- default `rtk grep` behavior remains unchanged unless a new mode, explicit limit, environment opt-in, or config opt-in is used
- `rtk rg --json` remains native ripgrep JSON
- explicit CLI limits override agent-safe, environment, and config defaults
- `--all` and `--full-lines` independently disable match caps and line clipping
- default piped match output remains live; piped structured and JSON modes emit after EOF
- `--context-only` remains an RTK option and is supported before or after the pattern/path
- native help and version behavior remains available without enabling structured processing

## Agent-safe output

`--agent-safe` defaults to:

- 80 total matches
- 5 matches per file
- 240 Unicode scalar values per line

Match and context rows are accounted for separately. Context output remains associated with displayed matches and cannot bypass configured bounds.

## JSON

- emits one deterministic newline-terminated `rtk.grep.v1` object
- supports matches, files-only, count-by-file, top-files, zero-match, capped, clipped, context, partial-result, and error cases
- preserves machine-openable paths, exact forwarded recovery argv, and native grep exit codes
- emits no human text or ANSI on stdout
- partial output plus an engine error preserves parseable results and includes a non-null error
- `rtk rg --json` remains native ripgrep JSON

## Parser safety

- NUL-separated grep output is preferred through `--null` or `-Z`
- Windows grep without NUL support uses a platform-limited ordinary parser
- native context rows require NUL-safe grep output; ordinary Windows fallback rejects context modes before child execution rather than guessing around valid -digits- filenames
- native grep operands are tracked by argv role; incomplete required native options fail closed before child execution in RTK-owned modes
- POSIX grep without NUL-safe filename output fails closed for RTK-owned structured modes
- parser-unavailable JSON emits a typed error object and exit 2 without executing the child
- parser-unavailable non-JSON structured modes return a clear error without passthrough
- default legacy grep behavior remains unchanged

## Safety

- structured capture uses fixed-size reads and bounds retained stdout and stderr to 10 MiB per stream
- capture overflow fails closed with a partial-result error
- ambiguous ordinary colon-separated POSIX paths are never parsed as structured records
- forced ANSI output and incompatible native shape modes are rejected before structured execution
- active RTK processing modes cannot silently fall through to native output

## Validation

- focused search tests: 122 passed, 2499 filtered
- config tests: 12 passed, 2609 filtered
- grep integration tests: 1 passed, 10 filtered
- exact no-match guard: 1 passed, 10 filtered
- `cargo check --all-targets`: passed
- `cargo fmt --all -- --check`: passed
- `cargo clippy --all-targets`: passed
- full suite: 2636 passed, 8 ignored
- manual checks passed for native operand-role preservation, incomplete native operand rejection, separated forced-color handling, shape-flag rejection, Windows filenames and labels, ordinary-context fail-closed behavior, bounded capture and live-stream decoding, child draining/reaping, piped `--context-only`, partial-result JSON, invalid regex, ANSI rejection, Unicode handling, PowerShell JSON parsing, and native `rtk rg --json`


MiVi29 (approved):

MiVi29:
sorry, did by mistake.

PR #3461, opened 2026-08-06
fix(hooks): stop unit tests reading the developer's config.toml

David-Martel:
## Problem

`cargo test` is **red for contributors who use rtk and green in CI**, which is why this hasn't
surfaced upstream. Two places in the hook decision path do file I/O, so unit tests inherit the
contributor's machine configuration.

**1. `get_rewritten` calls `Config::load()`** — reads `$CONFIG_DIR/rtk/config.toml`. With a
configured

```toml
[hooks]
exclude_commands = ["git", "ls", "grep", ...]
```

commands the tests expect to be rewritten are excluded, `get_rewritten` returns `None`, and the
decision comes back `Defer`. `test_claude_rewrite_git_status` — which asserts `git status` →
`rtk git status` — fails, along with 22 others spanning every host handler (claude, cursor,
gemini, droid, vibe, copilot), since they all funnel through `decide_from_verdict`.

**2. `evaluate` calls `check_command`** — reads `~/.claude/settings.json` and
`$PROJECT_ROOT/.claude/settings.json`. The `unattestable_passthrough` tests pass explicit empty
exclusions but still inherit that ambient verdict, so with a matching deny rule the outcome is
`Deny` instead of `Ask` and 2 more tests fail.

Excluding `git` is a reasonable real-world config — it's what you'd set to keep exact `git`
output for a coding agent — so this is not an exotic setup.

### Measured

Pristine checkout of `develop`, no source changes, only the config file moved aside:

| condition | `hooks::hook_cmd` |
|---|---|
| `config.toml` present with `[hooks] exclude_commands` | **23 of 108 fail** |
| same file moved aside | **108 pass, 0 fail** |

## Fix

Split file I/O out of the decision path, mirroring the convention `permissions` already
established with `check_command_with_rules` — whose doc comment reads *"Internal implementation
allowing tests to inject rules without file I/O."* The seam existed; these paths just didn't
use it.

| commit | change |
|---|---|
| `stop unit tests reading the developer's config.toml` | `get_rewritten_with` / `decide_from_verdict_with` take exclusions as parameters; `hook_exclusions()` becomes the single loader and is `#[cfg(test)]`-stubbed to empty so no test path reaches the filesystem. |
| `stop rewrite tests reading the host's permission settings` | `evaluate_with_verdict` takes the verdict as a parameter; `evaluate` loads it and delegates; the test module supplies `PermissionVerdict::Default`. |

**Production behavior is unchanged** in both cases — the loading wrappers still load. The only
behavioral difference is inside `#[cfg(test)]`.

Added `test_get_rewritten_with_honors_exclusions` so the exclusion behavior the other tests now
bypass stays pinned rather than silently untested.

## Verification

Run **with a populated `config.toml` and `settings.json` in place** — the condition that used to
fail:

- unit test binary: **25 failures → 0** (2571 passed, 8 ignored)
- `cargo clippy --release --all-targets -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean

## Note for maintainers

Two related issues found while investigating, **not** addressed here:

1. `tests/copilot_selfheal_test.rs` (added in #3449) fails **10 of 12 on a pristine `develop`
   checkout** on a machine with existing Copilot hook config — same class of ambient-state
   dependency, in brand-new code. Happy to extend this PR to cover it if you'd like.
2. `core::tracking` tests are flaky under full-suite parallelism — they share one real SQLite
   tracking DB. They pass 14/14 in isolation, both parallel and single-threaded.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -231,22 +231,46 @@ fn heal_legacy_hook_file(path: &std::path::Path) -> bool {
         .is_ok()
 }
 
-fn get_rewritten(cmd: &str) -> Option<String> {
+/// Rewrite `cmd` using caller-supplied hook exclusions.
+///
+/// Kept free of file I/O so unit tests can exercise the decision path without
+/// reading the developer's `config.toml` — mirrors `check_command_with_rules`.
+fn get_rewritten_with(
+    cmd: &str,
+    excluded: &[String],
+    transparent_prefixes: &[String],
+) -> Option<String> {
     if has_heredoc(cmd) {
         return None;
     }
 
-    let (excluded, transparent_prefixes) = crate::core::config::Config::load()
-        .map(|c| (c.hooks.exclude_commands, c.hooks.transparent_prefixes))
-        .unwrap_or_default();
-
-    let rewritten = rewrite_command(cmd, &excluded, &transparent_prefixes)?;
+    let rewritten = rewrite_command(cmd, excluded, transparent_prefixes)?;
 
     if rewritten == cmd {
         return None;
     }
 
     Some(rewritten)
+}
+
+/// The user's `[hooks]` exclusions and transparent prefixes.
+#[cfg(not(test))]
+fn hook_exclusions() -> (Vec<String>, Vec<String>) {
+    crate::core::config::Config::load()
+        .map(|c| (c.hooks.exclude_commands, c.hooks.transparent_prefixes))
+        .unwrap_or_default()
+}
+
+/// Unit tests never read the developer's `config.toml`.
+///
+/// Reading it made `cargo test` environment-dependent: on a machine with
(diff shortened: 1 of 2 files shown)


KuSh:
Hi @David-Martel,

Thanks for your PR. Could you extend it with copilot fix? The flaky DB tests are already fixes in #3206 with commit 2aef8e8556b5415b00aa4ed2e1f5fb838b7416b4

KuSh:
Closing as a duplicate of #3638, which refactors the same two functions in `hook_cmd.rs` — `get_rewritten` and `decide_from_verdict` — for the same reason: keeping the unit tests off the developer's real `config.toml`.

#3638 is CLA-cleared and this one isn't, so it can't merge as-is.

The underlying problem is real and worth having found — thanks for spotting it.

PR #3462, opened 2026-08-07
fix(rewrite): preserve quoted assignment data

### LINKED ISSUE #3262: bug(hook): rewrite edits inside quoted string literals — 'shellcheck' in a single-quoted sed replacement became 'rtk shellcheck' (data corruption)
## Summary

The Claude `PreToolUse` rewrite path rewrites command names **inside quoted string literals**. A single-quoted string assigned to a variable had its contents edited: `shellcheck` became `rtk shellcheck` *inside the string*, so the corrupted text — not a command — was written into files by the surrounding `sed`.

Like #3197, this is a corruption bug, not a filtering one: the rewritten string is not equivalent to the input, and the data the command operates on is silently altered.

## Reproduction (rtk 0.43.0, macOS)

```bash
python3 - <<'EOF' > /tmp/rtk-probe.json
import json
cmd = """D='# shellcheck disable=SC2034  # comment'; for f in hooks/*.sh; do sed -i '' -e "s|^TS_BACKUP=|${D}\\nTS_BACKUP=|" "$f"; done"""
print(json.dumps({"session_id":"t","transcript_path":"/tmp/x","cwd":"/tmp",
  "hook_event_name":"PreToolUse","tool_name":"Bash","tool_input":{"command":cmd}}))
EOF
rtk hook claude < /tmp/rtk-probe.json
```

Actual (`updatedInput.command`, abbreviated):

```
D='# rtk shellcheck disable=SC2034  # comment'; for f in hooks/*.sh; do ...
```

The token `shellcheck` sits inside a **single-quoted literal** — it is data, not a command — yet it was rewritten. Expected: passthrough (or at minimum, no edits inside quote spans).

A second probe shows the same machinery inserting a space inside a sed script:

```bash
printf '%s' '{"tool_input":{"command":"D=# shellcheck disable=SC2034; sed -e s|a|$D| f.sh"}}' | rtk hook claude
# → "command":"D=# rtk shellcheck disable=SC2034; sed -e s |a|$D| f.sh"
#                                                        ^ space inserted: `s|a|$D|` split at the `|`
```

`s|a|$D|` is one sed argument; after the rewrite it becomes `s` + a pipeline, which is a different program.

## Real-world incident

An agent session ran a loop inserting a shellcheck directive into eight hook scripts:

```bash
D='# shellcheck disable=SC2034  # ...' && for f in hooks/*.sh; do sed -i '' -e "s|^TS_BACKUP=...|${D}\n...|" "$f"; done
```

All eight files ended up containing `# rtk shellcheck disable=SC2034` — which shellcheck does not recognize as a directive, so the disable was silently inert. From the model's side nothing failed: exit 0, files written, corruption discovered only by re-running shellcheck. This is the worst failure mode — no error, wrong bytes on disk.

## Analysis

Argument-position occurrences are *not* rewritten (`echo 'run shellcheck later'` passes through untouched), so this is not a blind substring replace. The lexer appears to lose the quote span when the opening quote is attached to an assignment word (`D='...`), leaving the following words in what it considers command position. The rejoin step then also normalizes spacing (the `s|a|$D|` → `s |a|$D|` split above), which is the same "rejoin loses original bytes" class as #3197's suggested fix — splicing rewritten segments back by byte offset and leaving all untouched bytes verbatim would fix both.

## Related

- #3197 — `;;` → `; ;` (same corruption class, different lexer gap)
- #1560 — pipes rewritten on the left-hand side
- #3111 (closed) — commands with double-quoted strings were previously *skipped*; the current behavior rewrites them, which trades a missed optimization for corruption
- #3171 — safe pipeline rewriting tracking issue

None cover quoted-literal contents.

## Environment

- rtk 0.43.0 (Homebrew), also checked the v0.44.0 changelog — no lexer/quoting fix listed
- Claude Code, hook `rtk hook claude`, `PreToolUse` / matcher `Bash`
- macOS (Darwin 25.5.0), bash

zhiwuyazhe-fjr:
## Summary

- prevent the environment-prefix regex from backtracking into quoted assignment values and exposing literal data as a command
- preserve valid shell values composed from quoted and unquoted fragments
- add regressions for the original `shellcheck` corruption and the latest reported single-/double-quoted PR-title and prose cases
- rebase the branch onto current `develop`

## Root cause

The lexer kept quote boundaries and byte offsets correctly. The corruption happened later in `ENV_PREFIX`: when a complete quoted assignment did not satisfy the trailing-whitespace requirement at a compound-command boundary, the regex could fall back to the broad unquoted alternative and match only the start of the assignment. The remaining literal text then looked like a command and was rewritten.

The unquoted fragment now excludes quote characters, while the environment value accepts a sequence of complete quoted or unquoted fragments. This prevents partial reinterpretation without regressing valid forms such as `FOO='bar baz'qux git status`.

This is independent of #3202's case-terminator lexer fix.

## Security behavior

The original #3262 payload now passes through unchanged. A supported command outside the quoted assignment is still rewritten:

```text
D='# shellcheck | git status; $(whoami)' ; cargo test
  -> D='# shellcheck | git status; $(whoami)'; rtk cargo test
```

The latest reported shapes are covered too:

- `T="... make ..."; echo "len=${#T}"` preserves the double-quoted value
- `T='please make a cake'; printf '%s' "$T"` preserves the single-quoted value
- the longer `T=...; echo ...; [ ... ] && gh pr edit ... --title "$T"` form preserves `T` while rewriting only the external `gh` command
- a simple `gh pr edit --title "... make ..."` still takes the normal rewrite path and preserves its title bytes

## Test plan

- `cargo fmt --all`
- `cargo test --all`: 2583 passed, 0 failed, 8 ignored; all integration suites passed
- targeted regressions for the four latest reported command shapes

Closes #3262

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -5684,6 +5684,45 @@ mod tests {
     }
 
     #[test]
+    fn test_rewrite_compound_preserves_double_quoted_prose_assignment() {
+        let command =
+            r#"T="chore(skills): make every Pocock skill model-invocable"; echo "len=${#T}""#;
+
+        assert_eq!(rewrite_command_no_prefixes(command, &[]), None);
+    }
+
+    #[test]
+    fn test_rewrite_compound_preserves_single_quoted_prose_assignment() {
+        let command = r#"T='please make a cake'; printf '%s' "$T""#;
+
+        assert_eq!(rewrite_command_no_prefixes(command, &[]), None);
+    }
+
+    #[test]
+    fn test_rewrite_single_preserves_inline_title_prose() {
+        let command =
+            r#"gh pr edit 1476 --title "chore(skills): make every Pocock skill model-invocable""#;
+
+        assert_eq!(
+            rewrite_command_no_prefixes(command, &[]),
+            Some(
+                r#"rtk gh pr edit 1476 --title "chore(skills): make every Pocock skill model-invocable""#
+                    .into()
+            )
+        );
+    }
+
+    #[test]
+    fn test_rewrite_compound_preserves_title_before_gh_edit() {
+        let command = r#"T="chore(skills): make every Pocock skill model-invocable"; echo "len=${#T}"; [ ${#T} -le 100 ] && gh pr edit 1476 --title "$T""#;
+
+        assert_eq!(
+            rewrite_command_no_prefixes(command, &[]),
+            Some(r#"T="chore(skills): make every Pocock skill model-invocable"; echo "len=${#T}"; [ ${#T} -le 100 ] && rtk gh pr edit 1476 --title "$T""#.into())
+        );
+    }
+
+    #[test]
     fn test_rewrite_compound_pipe_raw_filter() {
         // Producers stay raw; only a pipeline-safe final stage is rewritten.
         assert_eq!(
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -63,8 +63,11 @@ static COMPILED: LazyLock<Vec<Regex>> = LazyLock::new(|| {
 static ENV_PREFIX: LazyLock<Regex> = LazyLock::new(|| {
     let double_quoted = r#""(?:[^"\\]|\\.)*""#;
     let single_quoted = r#"'(?:[^'\\]|\\.)*'"#;
-    let unquoted = r#"[^\s]*"#;
-    let env_value = format!("(?:{}|{}|{})", double_quoted, single_quoted, unquoted);
+    // Quotes must be handled by the complete quoted alternatives above.
+    // Otherwise regex backtracking can reinterpret a quoted assignment as a
+    // partial unquoted value and expose literal data as a command (#3262).
+    let unquoted = r#"[^\s'"]+"#;
+    let env_value = format!("(?:{}|{}|{})*", double_quoted, single_quoted, unquoted);
     let env_assign = format!(r#"[A-Z_][A-Z0-9_]*={}"#, env_value);
     // NOTE: `sudo` is intentionally NOT stripped here. Rewriting `sudo docker ps`
     // to `sudo rtk docker ps` breaks at runtime because `rtk` is not on root's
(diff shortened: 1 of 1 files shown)


bharvey42:
Carried in a local fork build since 2026-08-08 (Windows 11, daily use), no regressions.

Still merges cleanly with `develop` as of today (aa40853) — including after #3749 landed a few hours ago and reworked `exclude_commands` matching in this same file (`src/discover/registry.rs`, ~230 lines changed). `git merge-tree` reports no conflict against that, but given the overlap it may be worth a second look before merge.

zhiwuyazhe-fjr:
Thanks @bharvey42 for the report and for carrying this in a daily-use fork build since 2026-08-08 — glad to hear no regressions.

Re: the overlap with #3749 — rebased onto `develop` aa40853 (which includes it) and took that second look:

- **Rebase was clean**, no conflicts, as `merge-tree` predicted.
- **Semantics**: this PR only changes the `ENV_PREFIX` regex (how quoted env-assignment prefixes are peeled), which runs *before* #3749's `tool_form()` / `exclude_commands` matching. The fix ensures the peeled form of e.g. `FOO='bar baz' npx playwright test` is exactly `npx playwright test`, so the exclusion sees the same well-formed input #3749 canonicalizes. No interaction beyond that — the wrapper-peeling and the quoted-literal fix compose on `cmd_clean` without touching each other's logic.

Verification on Windows after the rebase:

- Full suite: **2910 tests, 0 failed** (8 ignored) — includes all quoted-literal tests from this PR and the wrapper-form exclusion tests from #3749, passing together
- Note: `copilot_selfheal_test` fails 10/12 on Windows, but I reproduced the identical failures on a clean `develop`@aa40853 checkout, so that's pre-existing (#3449) and unrelated to this change.

Ready for review.

zhiwuyazhe-fjr:
Fresh rebase onto the current `develop` tip (0f8b26c) pushed — still conflict-free, including against the ~365 commits that landed since the last check (the arg_tokenizer work touches `src/discover/registry.rs` heavily, and this rebase stayed clean).

Verification on Windows after the rebase:

- Full suite: **3532 passed, 0 failed** (8 ignored) — the quoted-literal tests from this PR and the wrapper-form `exclude_commands` tests from #3749 all pass together
- Only failure is `find_cmd::run_from_args_propagates_find_exit_status`, which fails identically on a clean `develop` checkout under Windows (find.exe vs GNU find exit semantics) — pre-existing and unrelated

One heads-up for reviewers: workflow runs from this fork are approval-gated, so the CI run for this push is sitting in `action_required` — a maintainer approval from the Actions tab is needed for the checks to actually execute.

PR #3464, opened 2026-08-07
feat(opencode): add v2 plugin support (--opencode-v2)

umi008:
## Summary

- Add `hooks/opencode/rtk-v2.ts` — OpenCode v2 plugin using `Plugin.define` + `ctx.tool.hook('execute.before', ...)` with `child_process.spawn` (v2 doesn't provide the bun-shell `$` helper)
- Add `--opencode-v2` flag to `rtk init`, mutually exclusive with `--opencode`
- Embed `OPENCODE_V2_PLUGIN` via `include_str!`; thread through all sub-functions

## Test plan

- [x] Manual testing: plugin loads and registers hook with bun (opencode2 runtime)
- [x] Manual testing: `rtk rewrite` rewrites commands correctly (exit code 3 handled)
- [x] Manual testing: `rtk init -g --opencode-v2` installs the v2 plugin

### DIFF
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -187,7 +187,7 @@ Returns `{}` when no rewrite (Cursor requires JSON for all paths).
 
 ### OpenCode (TypeScript Plugin)
 
-Mutates `args.command` in-place via the zx library:
+**v1 (default)**: Mutates `args.command` in-place via the zx `$` helper:
 
 ```typescript
 const result = await $`rtk rewrite ${command}`.quiet().nothrow()
@@ -196,6 +196,16 @@ if (rewritten && rewritten !== command) {
   (args as Record<string, unknown>).command = rewritten
 }
 ```
+
+**v2**: Uses `Plugin.define` + `ctx.tool.hook("execute.before", ...)` — spawns `rtk rewrite` directly since v2 context does not provide `$`:
+
+```typescript
+const proc = spawn("rtk", ["rewrite", command])
+proc.stdout.on("data", (d) => { out += d })
+proc.on("close", () => resolve(out.trim()))
+```
+
+Installed with `rtk init -g --opencode` (v1) or `rtk init -g --opencode-v2` (v2).
 
 ### Hermes (Python Plugin)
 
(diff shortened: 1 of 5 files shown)


Sn0wo2:
`import { Plugin } from "@opencode-ai/plugin"` fails to resolve in the v2 plugin runtime (even with the package installed next to the file). Dropping the import and exporting a plain `{ id, setup }` object fixes it.
`Plugin.define` is just a type wrapper. Everything else works as-is (verified on next-17444):

```ts
import { spawn } from "child_process"

// RTK OpenCode v2 plugin — rewrites commands to use rtk for token savings.
// Requires: rtk >= 0.23.0 in PATH.
//
// OpenCode 2.0 uses a new plugin API (Plugin.define + ctx.tool.hook) and does
// not provide the v1 bun-shell `$` helper, so this variant spawns `rtk`
// directly. `rtk rewrite` exits non-zero even on success, so we read stdout
// from the child instead of relying on exit codes.
//
// This is a thin delegating plugin: all rewrite logic lives in `rtk rewrite`,
// which is the single source of truth (src/discover/registry.rs).
// To add or change rewrite rules, edit the Rust registry — not this file.

function rtkRewrite(command: string): Promise<string> {
    const proc = spawn("rtk", ["rewrite", command], { windowsHide: true })
    let out = ""
    proc.stdout.on("data", (d) => {
      out += d
    })
    proc.on("error", () => resolve(""))
    proc.on("close", () => resolve(out.trim()))
  })
}

export default {
  id: "rtk.rewrite",
  setup: async (ctx) => {
    await ctx.tool.hook("execute.before", async (event) => {
      const tool = String(event.tool ?? "").toLowerCase()
      if (!event.input || typeof event.input !== "object") return

      const command = (event.input as Record<string, unknown>).command

      const rewritten = await rtkRewrite(command)
        ;(event.input as Record<string, unknown>).command = rewritten
      }
    })
  },
}
```

KuSh:
Closing this one in favour of #3899, which adds the same OpenCode 2.0 support inside the existing plugin file so both plugin formats are handled automatically and no separate --opencode-v2 flag is needed. If the separate-file approach still matters for your setup, a comment on #3899 is the best place to raise it.

PR #3470, opened 2026-08-07
feat(ls): cap listing with tee tail hint and standard dotfile semantics

aeppling:
rtk ls silently dropped content: no display cap on large listings, noise-filtered dirs vanished without a trace, and dotfiles were shown by default via a forced -a.

Now:
- Listings are capped; everything RTK hides (overflow + filtered noise) is recoverable with a single tee tail command that returns only unseen content.
- Dotfiles follow standard ls behavior, shown only when -a is passed.
- The TTY-only summary footer is removed; agents never saw it.

### DIFF
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -125,7 +125,7 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
                     raw.len(),
                     out.len(),
                     if !raw.is_empty() {
-                        100 - (out.len() * 100 / raw.len())
+                        100usize.saturating_sub(out.len() * 100 / raw.len())
                     } else {
                         0
                     }
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -2,11 +2,10 @@
 
 use super::constants::NOISE_DIRS;
 use crate::core::runner::{self, RunOptions};
-use crate::core::truncate::{reduced, CAP_WARNINGS};
+use crate::core::truncate::CAP_INVENTORY;
 use crate::core::utils::resolved_command;
 use anyhow::Result;
 use regex::Regex;
-use std::io::IsTerminal;
 use std::sync::LazyLock;
 
 /// Matches the date+time portion in `ls -la` output, which serves as a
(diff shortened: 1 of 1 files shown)


pszymkowiak:
Tested current head (67a3576). Two real bugs confirmed locally, two cleanup points.

**1. `-A` silently drops dot-prefixed noise dirs with zero trace** (`ls.rs:303`)
```
mkdir -p .git node_modules && touch README.md
/bin/ls -A .        # → .git  node_modules  README.md
rtk ls -A .         # → README.md, "1 filtered" (node_modules only)
```
The teed recovery file (the one the "N filtered" hint points to) only contains `node_modules/` — `.git` isn't in the listing, isn't counted in "filtered", and isn't in the recovery file. It's just gone. `show_all` detection (`a.contains('a')`) doesn't recognize uppercase `-A`, and the flag isn't stripped before forwarding to the real `ls`, so the child actually runs with `-A` (dotfiles included) while rtk's own noise filter still assumes `show_all == false` and drops them — the exact "filtered dir vanishes without a trace" bug this PR is meant to fix.

**2. Tracking label hardcodes `-la` even when `-a` wasn't passed**
```
rtk ls .   # no -a flag at all
```
`rtk gain --history` records this as `rtk ls -la .` regardless — the label at line 87 is a fixed string, not built from the actual flags used. Misrepresents what was actually run.

Two smaller cleanup points, not bugs:
- Same concept named 3 different ways across the file: `hidden` (the Vec in `compact_ls`), `'filtered'` (its own doc comment), `noise` (in `run()`), `filtered` again (as `hidden_hint`'s param name).
- `hidden_hint()` hand-rolls its own two-category overflow text instead of extending the existing shared `join_with_overflow` helper (`core/utils.rs:151`).

aeppling:
@pszymkowiak updated

pszymkowiak (approved):
Verified locally (merged onto #3798 since develop itself doesn't compile): fmt/clippy clean, 2975 tests pass (+10 new). Smoke-tested the binary: dotfiles now hidden by default and shown with `-a`, filtered noise dirs produce the `... (N filtered)` note plus a one-shot tee tail command containing only the dropped entries — recovery works as advertised. Good call removing the TTY-only summary agents never saw, and recording unparsable lines as hidden instead of silently dropping them. LGTM — merge after #3798.

PR #3471, opened 2026-08-07
fix(diff): compare files byte-accurately instead of by normalized lines

### LINKED ISSUE #3469: rtk diff reports "Files are identical" (exit 0) for byte-different files
# `rtk diff` reports "Files are identical" (exit 0) for byte-different files

## Summary

`rtk diff` (the built-in "ultra-condensed diff") gives **false-equal** results in two
independent ways, both dangerous when `diff` is used to *prove* two files are equal:

1. **Whitespace/line-ending normalization → false "identical".** Files whose only
   differences are trailing whitespace, CRLF-vs-LF, a missing/extra trailing newline,
   or leading indentation are reported as `[ok] Files are identical` even though the
   files are **not** byte-identical (different md5, and `cmp`/`diff` report a
   difference).

2. **Exit code is always 0.** Even when `rtk diff` *does* detect and display a real
   content change, it exits `0`. GNU/BSD `diff` exits `1` when files differ. This
   breaks `if diff a b; then ...` and `diff a b && echo same`.

Either defect silently defeats any verification of the form "diff proves these two
files are equal."

## Environment

- `rtk 0.42.4` (the token-killer `rtk`; `rtk gain` works — not the unrelated
  `reachingforthejack/rtk`). Current Homebrew stable is `0.44.2` — the behavior
  below should be re-verified against the latest release in case it is already
  addressed.
- Installed via Homebrew
- Triggered in practice through a Claude Code `PreToolUse` hook (`rtk hook claude`)
  that rewrites bare `diff a b` → `rtk diff a b`, but the defects reproduce by
  calling `rtk diff` directly (see below), so the hook is not required.

### Defect 1 — whitespace-only difference reported as identical

```bash
printf 'foo\nbar\n'  > A      # md5: f47c75614087a8dd938ba4acff252494
printf 'foo \nbar\n' > B      # md5: 259f776914a6f16abb525d25d8c84aa1  (trailing space)

md5 A B            # -> two DIFFERENT hashes
cmp A B; echo $?   # -> "A B differ: char 4, line 1"   exit 1
diff A B; echo $?  # (real diff) -> "1c1 < foo / > foo "   exit 1

rtk diff A B; echo $?
# ACTUAL:   [ok] Files are identical
#           0
# EXPECTED: a reported difference, non-zero exit
```

Same false "identical" reproduces for:

| Case | file A | file B | md5 A | md5 B |
|------|--------|--------|-------|-------|
| trailing space | `foo\nbar\n` | `foo \nbar\n` | f47c7561… | 259f7769… |
| CRLF vs LF | `alpha\nbeta\n` | `alpha\r\nbeta\r\n` | 852e77b4… | def13413… |
| trailing newline | `abc` | `abc\n` | 90015098… | 0bee89b0… |
| indentation | `x\n    indented\n` | `x\nindented\n` | 08990ce7… | 8a4dbe59… |

All four: `rtk diff` → `[ok] Files are identical`, exit `0`.
Ground truth for all four: different md5, `cmp` exit `1`, real `diff` exit `1`.

### Defect 2 — exit code 0 even when a real change is shown

```bash
printf 'a\nworld\n' > A
printf 'a\nWORLD\n' > B

rtk diff A B; echo $?
# ACTUAL:
#   A → B
#      +1 added, -1 removed, ~0 modified
#   -   2 world
#   +   2 WORLD
#   0            <-- exit 0 despite a displayed difference
# EXPECTED: exit 1
```

## Impact

`diff` is commonly used as an equality oracle in scripts and deploy verifications
(`diff a b && echo "unchanged"`, `if diff old new; then skip; fi`). Both defects turn
a genuine difference into a "they match" verdict. The whitespace case is especially
risky because whitespace/line-ending changes (e.g. a file silently converted to CRLF,
or a lost trailing newline) are exactly the kind of drift a verification is meant to
catch.

## Suggested fix

- `rtk diff` should exit **non-zero when the files differ** (match `diff`'s convention:
  0 = same, 1 = differ, 2 = error), on both the "identical" path and the
  changes-displayed path.
- Whitespace/line-ending/trailing-newline differences should **not** be collapsed to
  "identical" by default. If line-normalized comparison is desirable, gate it behind an
  explicit opt-in flag (e.g. `-w` / `--ignore-whitespace`) rather than making it the
  default, and even then report a non-zero exit for byte-level differences.
- At minimum, when the byte contents differ but the normalized view is equal, print a
  distinct message (e.g. "files differ only in whitespace/line endings") and exit
  non-zero — never `[ok] Files are identical`.

## Workaround

Add `diff` to `exclude_commands` in `config.toml` so the hook passes it through to the
real binary:

```toml
[hooks]
exclude_commands = ["diff"]
```

Verified: with this set, `rtk`-hooked `diff` runs the system `diff` and correctly
reports differences with exit 1 for all cases above. `rtk proxy diff a b` also bypasses
the filter. For equality proofs we now standardize on `cmp`/`md5` (which have no `rtk`
subcommand and pass through untouched) rather than `diff`.

Hydr0gen19:
## Summary

- `rtk diff a b` compared files with `str::lines()`, which drops the line terminator. CRLF vs LF and a missing final newline produced an empty change list, so byte-different files were reported as `[ok] Files are identical` and exited `0`. Comparison is now byte-accurate: only strictly equal contents exit `0`.
- When the bytes differ but every line matches, the terminators are the only possible cause, so the output says that explicitly instead of claiming the files are identical.
- Added `--ignore-whitespace` to keep the lenient comparison available on request (trims trailing whitespace, ignores line endings and the final newline).

`Fixes #3469`

### Why this matters beyond `rtk diff`

`src/discover/rules.rs` rewrites plain `diff <args>` to `rtk diff`, so an agent
running `diff a b` got the false-equal answer and the `0` exit code without ever
asking for rtk. That is exactly the case the issue describes as breaking
diff-based equality checks.

### Behaviour

| Input | Before | After |
|-------|--------|-------|
| identical bytes | `[ok] Files are identical`, exit 0 | unchanged |
| CRLF vs LF | `[ok] Files are identical`, exit 0 | `identical line content; files differ in line endings or the final newline`, exit 1 |
| final newline present vs absent | `[ok] Files are identical`, exit 0 | same as above, exit 1 |
| line content differs | change list, exit 1 | unchanged |
| `--ignore-whitespace`, whitespace-only difference | n/a | `[ok] Files are identical (ignoring whitespace)`, exit 0 |

Trailing-space differences were already reported as changes before this PR; they
still are, and `--ignore-whitespace` now suppresses them.

Note: for very small files `guard::never_worse` still prefers the raw content
over the condensed message, as it does for every other `rtk diff` output. The
exit code — the part scripts rely on — is correct in every case.

### On the added flag

`--ignore-whitespace` is the one part of this PR that is not strictly the bug
fix. It is here because the triage comment on the issue asks for the lenient
comparison to stay available behind an explicit flag, and without it this change
removes a behaviour and leaves nothing in its place. It is about ten isolated
lines — happy to drop it and keep the fix minimal if you would rather.

It deliberately has no `-w` short form: in GNU `diff`, `-w` means "ignore all
white space", which is a broader thing than what this does.

## Test plan

      (clippy clean; see note below on two pre-existing flaky tests)
- [x] Unit tests added in `src/cmds/git/diff_cmd.rs`: CRLF vs LF, missing final
      newline, trailing space, byte-identical CRLF files, and both
      `--ignore-whitespace` paths. They fail on `develop` and pass here.
- [x] Manual testing:

```
$ printf 'alpha\r\nbeta\r\n' > crlf.txt; printf 'alpha\nbeta\n' > lf.txt

# before
$ rtk diff crlf.txt lf.txt
[ok] Files are identical
$ echo $?
0

# after
$ rtk diff crlf.txt lf.txt
crlf.txt → lf.txt
   identical line content; files differ in line endings or the final newline
$ echo $?
1

$ rtk diff --ignore-whitespace crlf.txt lf.txt
[ok] Files are identical (ignoring whitespace)
$ echo $?
0
```

Two tests (`core::tracking::tests::test_timed_execution_*` and
`cmds::dotnet::dotnet_trx::tests::test_parse_trx_files_in_dir_since_ignores_older_files`)
fail intermittently on `cargo test --all` in my environment and pass when run on
their own. They are unrelated to this change — different runs fail on different
tests, and nothing here touches tracking or dotnet.

## Docs

`docs/usage/FEATURES.md` — added the `--ignore-whitespace` option and a note on
the byte-accurate comparison and exit codes.

### DIFF
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -246,6 +246,14 @@ rtk diff <fichier1> <fichier2>
 rtk diff <fichier1>              # Stdin comme second fichier
 ```
 
+**Options :**
+
+| Option | Defaut | Description |
+|--------|--------|-------------|
+| `--ignore-whitespace` | non | Ignore les espaces en fin de ligne, les fins de ligne (CRLF/LF) et le saut de ligne final |
+
+La comparaison se fait au niveau des octets : deux fichiers ne sont declares identiques (code de sortie 0) que si leur contenu est strictement egal. Toute difference renvoie le code 1.
+
 ---
 
 ### `rtk wc` -- Comptage compact
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -8,7 +8,7 @@ use std::path::Path;
 
 /// Ultra-condensed diff - only changed lines, no context.
 /// Returns the diff-convention exit code: 0 if identical, 1 if files differ.
-pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
+pub fn run(file1: &Path, file2: &Path, ignore_whitespace: bool, verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
 
     if verbose > 0 {
@@ -19,7 +19,7 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
     let content2 = fs::read_to_string(file2)?;
     let raw = format!("{}\n---\n{}", content1, content2);
 
-    let (rtk, exit_code) = render_file_diff(file1, file2, &content1, &content2);
+    let (rtk, exit_code) = render_file_diff(file1, file2, &content1, &content2, ignore_whitespace);
 
     let shown = never_worse(&raw, &rtk);
     print!("{}", shown);
@@ -32,15 +32,45 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
     Ok(exit_code)
 }
 
+/// Splits into comparable lines. Line terminators are always dropped, so byte
+/// equality is checked separately by the caller.
+fn split_lines(content: &str, ignore_whitespace: bool) -> Vec<&str> {
+    if ignore_whitespace {
+        content.lines().map(str::trim_end).collect()
+    } else {
+        content.lines().collect()
+    }
+}
+
 /// Renders the condensed file comparison and returns it with the
 /// diff-convention exit code (0 = identical, 1 = differences found).
-fn render_file_diff(file1: &Path, file2: &Path, content1: &str, content2: &str) -> (String, i32) {
-    let lines1: Vec<&str> = content1.lines().collect();
-    let lines2: Vec<&str> = content2.lines().collect();
+fn render_file_diff(
+    file1: &Path,
+    file2: &Path,
+    content1: &str,
+    content2: &str,
+    ignore_whitespace: bool,
+) -> (String, i32) {
+    if content1 == content2 {
+        return ("[ok] Files are identical\n".to_string(), 0);
+    }
+
+    let lines1 = split_lines(content1, ignore_whitespace);
+    let lines2 = split_lines(content2, ignore_whitespace);
     let diff = compute_diff(&lines1, &lines2);
 
     if diff.changes.is_empty() {
-        return ("[ok] Files are identical\n".to_string(), 0);
+        if ignore_whitespace {
+            return ("[ok] Files are identical (ignoring whitespace)\n".to_string(), 0);
+        }
+        // Bytes differ but every line matches: only the line terminators can
+        // be responsible, and those never survive into the change list.
+        let rtk = format!(
+            "{} → {}\n   identical line content; files differ in line endings or the final newline\n",
+            file1.display(),
+            file2.display()
+        );
+        return (rtk, 1);
     }
 
     let mut rtk = String::new();
@@ -320,6 +350,7 @@ mod tests {
             Path::new("two.yaml"),
             "a: 1\n",
             "a: 2\n",
+            false,
         );
         assert!(
             !out.contains("identical"),
@@ -339,6 +370,7 @@ mod tests {
             Path::new("j2.json"),
             "{\"a\": 1}\n",
             "{\"a\": 2}\n",
+            false,
         );
         assert!(
             !out.contains("identical"),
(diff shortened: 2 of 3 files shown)


KuSh:
Closing as a duplicate of #3519, which adds the same byte-equality guard for #3469.

Your version is the more thorough of the two on the guard itself — checking `content1 == content2` up front, before `split_lines`, is the cleaner placement, and the `test_render_byte_identical_exit_zero_with_crlf` case (both files CRLF, must still be identical) is one the others miss.

What tipped it to #3519 is scope. This PR also adds a `--ignore-whitespace` CLI flag, changes `diff_cmd::run`'s signature, touches `main.rs` and edits `docs/usage/FEATURES.md` — none of which #3469 asks for, and each of which is a separate review surface. It also only ever ran `check-target` (skipped), so the flag's behaviour was never CI-verified. #3519 is 57 lines in one file with a full green matrix.

If you'd still like `--ignore-whitespace`, it's worth its own PR on top — it's a reasonable feature, just not one that should ride along with a correctness fix.

Hydr0gen19:
Fair call, and the scope point lands. I flagged `--ignore-whitespace` in the description as the part that was not the bug fix, and then shipped it in the same PR anyway — saying so does not make it a smaller review surface. A correctness fix and a new CLI option should not share one.

The CI note is the more useful half for me. I ran the suite in a container and treated that as covered, without checking whether the matrix had actually exercised the flag. It had not.

If it is any use: `test_render_byte_identical_exit_zero_with_crlf` — both files CRLF, identical bytes, must still exit 0 — is self-contained on this branch and depends on nothing else in it. Lift it into #3519 as-is if you want it, no need to wait on me or credit anything.

If `--ignore-whitespace` still looks worth having once #3519 is in, I will open it on its own.

Hydr0gen19:
Saw #3519 land with the CRLF case in it — thanks for routing the test over rather than letting it go with the PR. Better outcome than merging mine would have been: the fix stayed one focused change, and the coverage still made it in.

The scope point stuck, for what it is worth. Obvious in hindsight, and not something I was going to see from inside my own diff.

PR #3475, opened 2026-08-08
docs: fix Cursor init flag and document Windows paths

### LINKED ISSUE #3429: RuboCop autocorrect aliases missed, `rtk init --cursor` in troubleshooting does not exist, Windows config path undocumented — plus what I could not verify and why
Closing out the audit with the items I could verify from source but not exercise end to end. Each one says exactly what I checked and what I did not, so you can weigh them accordingly rather than taking a uniform confidence from me.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Where an item says "source only", no tool was run — the reason is stated inline.*

### 1. RuboCop autocorrect aliases are missed — SOURCE ONLY

`src/cmds/ruby/rubocop_cmd.rs:56-58`:

```rust
let is_autocorrect = args
    .iter()
    .any(|a| a == "-a" || a == "-A" || a == "--auto-correct" || a == "--auto-correct-all");
```

RuboCop also accepts `--autocorrect`, `--autocorrect-all`, `-x` (`--fix-layout`) — the hyphenless spellings are the ones its own docs use in current versions. Those miss this check, so `has_format` stays false and `--format json` is injected into a run that rewrites files (`:63`), routing it to the JSON filter.

There is a second-order effect visible in the struct: `RubocopOffense` (`:28-34`) has `correctable` but **no `corrected` field**, so offences that the same run just fixed are still counted as outstanding — which is what produces advice to re-run the autocorrect that already ran.

**What I did not do:** Ruby is not installed on the audit machine and I did not build a stand-in for `rubocop` specifically. So the flag list and the missing struct field are read from source; the user-visible output I described is inferred, not observed. If the hyphenless aliases are not actually accepted by the RuboCop versions you support, this is not a bug — I could not check that either.

### 2. `rtk init --global --cursor` does not exist

`docs/guide/resources/troubleshooting.md:49` gives:

```bash
rtk init --global --cursor    # Cursor
```

`rtk init --help` lists these long flags:

```
--opencode --gemini --agent --show --claude-md --hook-only --auto-patch --no-patch
--trust-filters --no-trust-filters --uninstall --codex --copilot --dry-run
--ultra-compact --skip-env
```

There is no `--cursor`. Cursor is reachable only as `--agent cursor`. Note the inconsistency this documents: `--codex`, `--copilot`, `--gemini` and `--opencode` have bare flags; Cursor, Droid, Hermes, Pi, Windsurf, Cline, Kilo and Antigravity do not — so a user reasonably generalises from the four that exist.

### 3. Windows config path is not documented anywhere

`docs/guide/getting-started/configuration.md:12-15` lists config-file locations for Linux and macOS only. `README.md:346` says rtk "works fully on native Windows". On Windows the path is `%APPDATA%\rtk\config.toml` (from `dirs::config_dir()`) — observed as `C:\Users\<user>\AppData\Roaming\rtk\config.toml`. `docs/guide/analytics/gain.md:125` has the same Linux/macOS-only listing for the database path.

A Windows user following the docs cannot find the file they are told to edit.

### What I could not check at all, and why

Rather than leave these as silent gaps:

- **`rtk cc-economics` behaviour.** I reported arithmetic defects in that path (halved monthly savings, weekly keys that cannot merge) from source in #3418. I never ran the command: it invokes `npx --yes ccusage`, which downloads and executes an unpinned package, and it reads real spend data. Running it on a machine I do not own to verify a report seemed like the wrong trade. Everything in #3418 is therefore an arithmetic argument you can check by reading, not an observed output.
- **`rtk init` and `rtk init --uninstall`.** All the `init.rs` findings (#3400, #3403, #3404, #3405, #3406) are source-derived for the same reason — running them writes into the real home directory. The Hermes YAML cases in #3405 were reproduced by extracting the pure functions and running them under `rustc`, which is as close as I could get without touching a real `~/.hermes`.
- **`rtk learn --write-rules`.** Same: there is no environment override for the transcript root, so a faithful run would read my own agent transcripts, which contain live credentials from unrelated work.
- **Live tool behaviour** for `aws`, `kubectl`, `docker`, `psql`, `mvn`, `gradle`, `sops`, `helm`, `terraform`, `ansible`, `php`, `ruby`, `golangci-lint`. Where a finding touches these I used controlled stand-ins and said so in the issue. Installing the binaries would not have closed the gap — each needs credentials, a cluster, a daemon or a database I do not have.

This is the same list as #3399, updated with what the later rounds added. The short version: **claims about rtk's own code are checkable by reading; claims about how a wrapped tool behaves are not, and I have tried to keep those two labelled separately throughout.**

### Suggested fixes (untested)

1. Add the hyphenless aliases and `-x` to `is_autocorrect`; add `corrected` to `RubocopOffense` and exclude corrected offences from the count.
2. Correct `troubleshooting.md:49` to `--agent cursor`, or add the bare `--cursor` flag for consistency with the other four.
3. Add a Windows row to the config and database path tables.

ivanclarksoft:
## Summary
Fixes documentation gaps from #3429 (docs items only):

1. 	roubleshooting.md: 
tk init --global --cursor does not exist — use --agent cursor (consistent with supported-agents docs).
2. configuration.md: add Windows config path %APPDATA%\rtk\config.toml.
3. gain.md: document Windows/macOS/Linux history.db locations.

## Notes
- RuboCop autocorrect alias / struct field fix from #3429 left for a separate code PR.
- Docs-only.

Fixes #3429

### DIFF
--- a/docs/guide/analytics/gain.md
+++ b/docs/guide/analytics/gain.md
@@ -122,7 +122,10 @@ Savings %     = (Saved / Input) × 100
 
 Savings data is stored locally in SQLite:
 
-- **Location**: `~/.local/share/rtk/history.db` (Linux / macOS)
+- **Location**:
+  - Linux: `~/.local/share/rtk/history.db`
+  - macOS: `~/Library/Application Support/rtk/history.db`
+  - Windows: `%LOCALAPPDATA%\rtk\history.db` (e.g. `C:\Users\<you>\AppData\Local\rtk\history.db`)
 - **Retention**: 90 days (automatic cleanup)
 - **Scope**: Global across all projects and Claude sessions
 
--- a/docs/guide/getting-started/configuration.md
+++ b/docs/guide/getting-started/configuration.md
@@ -13,6 +13,7 @@ sidebar:
 |----------|------|
 | Linux | `~/.config/rtk/config.toml` |
 | macOS | `~/Library/Application Support/rtk/config.toml` |
+| Windows | `%APPDATA%\rtk\config.toml` (e.g. `C:\Users\<you>\AppData\Roaming\rtk\config.toml`) |
 
 ```bash
 rtk config            # show current configuration
--- a/docs/guide/resources/troubleshooting.md
+++ b/docs/guide/resources/troubleshooting.md
@@ -45,9 +45,9 @@ rtk gain    # should now show token savings stats
 
 2. Initialize the hook:
    ```bash
-   rtk init --global    # Claude Code
-   rtk init --global --cursor    # Cursor
-   rtk init --global --opencode  # OpenCode
+   rtk init --global              # Claude Code
+   rtk init --global --agent cursor  # Cursor (no bare --cursor flag)
+   rtk init --global --opencode   # OpenCode
    ```
 
 3. Restart your AI assistant.


ivanclarksoft:
recheck

ivanclarksoft:
recheck

PR #3496, opened 2026-08-09
fix(format): default ruff to check mode

### LINKED ISSUE #3417: `rtk format` with no arguments runs `ruff format .` (in-place rewrite) in a ruff project but `black --check .` (dry run) in a black project
The two formatter arms of `rtk format` inject different modes: black gets a read-only `--check`, ruff gets the write-mode `format` subcommand. Both then get a `.` appended. So the same bare command dry-runs in one project and rewrites the whole tree in another.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Reproduced by running the shipped `rtk 0.44.2` with PATH stand-ins that print the argv they receive.*

### Observed

```
$ cd plain-project && rtk format      # no config → ruff is the fallback
RUFF-ARGV: format .

$ cd black-project && rtk format      # pyproject.toml has [tool.black]
BLACK-ARGV: --check .
```

The user typed neither `format` nor `.` in either case.

### The code

`src/cmds/system/format_cmd.rs:89-95`:

```rust
// Inject --check if not present for check mode
"black" if !user_args.iter().any(|a| a == "--check" || a == "--diff") => {
    cmd.arg("--check");
}
// Add "format" subcommand if not present
"ruff" if user_args.is_empty() || !user_args[0].starts_with("format") => {
    cmd.arg("format");
}
```

The black arm's own comment says "check mode" — that is clearly the intent for the command as a whole. The ruff arm adds `format`, which for ruff is the **write** subcommand; there is no `--check` alongside it. (`ruff format --check` is ruff's dry-run form.)

Then `:104-106`:

```rust
// Default to current directory if no path specified
if user_args.iter().all(|a| a.starts_with('-')) {
    cmd.arg(".");
}
```

`Iterator::all` is vacuously true on an empty iterator, so bare `rtk format` — no arguments at all — appends `.`.

Compounding it, ruff is also the fallback when no formatter is detected (`:54`), so a project with no formatter config at all takes the write path.

### Why this is worth fixing rather than documenting

A wrapper whose purpose is reducing output should not change whether a command writes to disk. Someone running `rtk format` to *see* what would change gets a dry run on one project and a whole-tree rewrite on the next — and in the ruff case the scope is also widened from "whatever ruff's config resolves" to an explicit `.`.

### Suggested fix (untested — no programmer has reviewed this)

Make the ruff arm match the black arm's intent: inject `format --check` (or `check` if that is the desired semantic), so both formatters are read-only by default. If write mode is wanted, it should be something the user asks for explicitly — `src/cmds/php/pint_cmd.rs:41` and `src/cmds/python/ruff_cmd.rs:46-50` both already distinguish read-only from write invocations and are the pattern to follow.

Separately, `if user_args.iter().all(...)` should probably be `if user_args.is_empty() || user_args.iter().all(...)` written explicitly, so the empty case is a deliberate decision rather than a property of `all`.

### Related

Same family as the `eslint --fix` path injection reported in #3409 — a default path added to a command that writes. Filing separately because this one also gets the mode wrong, not just the scope.

nightcityblade:
Fixes #3417

## Summary

- Make implicit Ruff formatting default to `ruff format --check`, matching Black’s read-only behavior.
- Preserve explicit `rtk format ruff format` write mode and `--diff` behavior.

## Test plan

  - N/A: command construction is covered by `test_ruff_defaults_to_check_mode`; Ruff execution is not needed for this argument-only change.

### DIFF
--- a/src/cmds/system/format_cmd.rs
+++ b/src/cmds/system/format_cmd.rs
@@ -55,6 +55,22 @@ fn detect_formatter_in_dir(args: &[String], dir: &Path) -> String {
     "ruff".to_string()
 }
 
+fn formatter_prefix_args(formatter: &str, user_args: &[String]) -> Vec<&'static str> {
+    match formatter {
+        "black" if !user_args.iter().any(|arg| arg == "--check" || arg == "--diff") => {
+            vec!["--check"]
+        }
+        "ruff" if user_args.is_empty() || !user_args[0].starts_with("format") => {
+            let mut args = vec!["format"];
+            if !user_args.iter().any(|arg| arg == "--check" || arg == "--diff") {
+                args.push("--check");
+            }
+            args
+        }
+        _ => Vec::new(),
+    }
+}
+
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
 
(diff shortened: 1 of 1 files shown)


HetCreep:
Confirmed — built this branch and drove `format_cmd::run()` directly with fake `ruff.bat`/`black.bat` stand-ins that echo their argv:

```
ruff project, bare `rtk format`         -> format --check .   (was write mode — fixed)
ruff project, `rtk format --diff`       -> format --diff .    (still read-only)
ruff project, `rtk format format`       -> format             (escape hatch preserved)
black project, bare `rtk format`        -> --check .          (unchanged, no regression)
```

All four match the intended behavior. Thanks for picking this up.

Issue #3395, opened 2026-08-04
`rtk psql` parses psql's aligned table output without `-X`, so a user's `~/.psqlrc` silently changes what rtk reads


HetCreep:
`rtk psql` parses psql's human-readable aligned table — `|` column separators, `^[-+]+$` rules, and a `(N rows)` footer — but it does not pass `-X`, so psql still executes the user's `~/.psqlrc` first. Any `\pset` in there changes the format rtk is parsing.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Source read only — psql is not installed on this machine, so I have not run the failing cases.*

### Source

- `src/cmds/cloud/psql_cmd.rs:29-45` (`run`) builds the command and injects **no** flags.
- `:111` and `:19`/`:63` implement the parse: `is_table_format` looks for the `|` separator and the `[-+]+` rule line, and the summary keys off `(N rows)`.

### What breaks it

1. **`~/.psqlrc` is honoured.** `\pset border 0` or `\pset border 2`, `\pset format wrapped|csv|html`, `\pset fieldsep`, or `\pset linestyle unicode` all change the output shape. `linestyle unicode` emits `│` and `─` instead of `|` and `-`, so `is_table_format` stops recognising the output as a table at all.
2. **`(N rows)` is translated.** It goes through NLS, so on a server or client with a non-English `lc_messages` the footer text differs and the row-count summary misses.
3. **A literal `|` inside a cell shifts the columns.** The parse splits the line on `|`, so a value containing a pipe produces an extra column — the row is rendered with fields shifted rather than failing to parse. That is the case I would worry about most: a wrong answer rather than a fallback.

### Suggested directions (untested — no programmer has reviewed these)

psql documents machine-readable modes that do not depend on user configuration:

- `--csv` (PostgreSQL 12+), or
- `-A -t -F$'\t'` (unaligned, tuples-only, explicit field separator)

plus `-X` to skip the startup file regardless of which output mode is chosen. rtk already injects machine-format flags with user-override detection for many other tools (`go test -json`, `kubectl -o json`, `docker --format`, eslint `-f json`, ruff/pylint `--output-format`, `glab ci list -F json`), so this would follow the existing pattern rather than introduce one.

If parsing the aligned format is deliberate — for example to preserve exactly what a user would have seen — then `-X` alone would still remove the `~/.psqlrc` variable.

pszymkowiak:
[...]

PR #3497, opened 2026-08-09
fix(psql): disable startup file loading

nightcityblade:
Addresses the startup-file portion of #3395.

## Summary

- Invoke `psql` with `-X` so startup configuration cannot alter the aligned output that RTK parses.
- Avoid duplicating the option when users already pass `-X` or `--no-psqlrc`.

## Test plan

  - N/A: exact command arguments are covered by `test_build_psql_command_disables_startup_file`; a live PostgreSQL server is not required for this argument-only change.

### DIFF
--- a/src/cmds/cloud/psql_cmd.rs
+++ b/src/cmds/cloud/psql_cmd.rs
@@ -8,6 +8,7 @@ use crate::core::truncate::CAP_LIST;
 use crate::core::utils::resolved_command;
 use anyhow::Result;
 use regex::Regex;
+use std::process::Command;
 use std::sync::LazyLock;
 
 const MAX_TABLE_ROWS: usize = CAP_LIST;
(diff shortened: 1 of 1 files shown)


HetCreep:
Confirmed the `-X` fix — built this branch, the unit test passes, and it's the right call for the `.psqlrc` case.

One flag before this merges: the original report had three separate sub-issues, and this closes only the first. Since the PR says "Fixes #3395", merging into `develop` will auto-close the whole issue.

Still open, neither touched by this diff:

- **NLS translation of the `(N rows)` footer** — `ROW_COUNT`'s regex is English-only (`^\(\d+ rows?\)`); a non-English `lc_messages` produces different footer text and the row-count summary misses it. `-X` doesn't affect this, it's a locale setting, not a startup-file setting.
- **A literal `|` inside a cell value shifts columns** — `filter_table`'s `trimmed.split('|')` still splits unconditionally. I fed a row with a text cell containing `a|b` through the current `filter_table` and got 3 columns back instead of 2 (`["1", "a", "b"]`), silently shifting every field after it. This is the one I'd worry about most since it produces a wrong answer rather than a visible failure.

Not asking for either in this PR — just flagging so they get a follow-up issue instead of disappearing when #3395 closes.

nightcityblade:
Thanks for flagging this. I changed the linkage so #3395 will remain open for the NLS-footer and literal-pipe issues; this PR stays scoped to the startup-file fix.

PR #3498, opened 2026-08-09
fix(git): keep `git status --porcelain` empty on a clean tree

asem89:
## Summary

- `rtk git status --porcelain` prints `ok` on a clean tree instead of the zero bytes git emits.
- For `--porcelain`, `--porcelain=v1/v2`, `-z` and `--null`, emptiness *is* the result — consumers parse "no output" as "no changes", so the `ok` compaction breaks them.
- `filter_status_with_args` now takes the user's args and skips the substitution for those four machine-readable formats. Human-readable output, including short `-s`, is unchanged.

## Why it bites

The standard clean-tree check inverts:

```bash
$ cd /tmp/clean-repo
$ git status --porcelain | wc -c
0
$ rtk git status --porcelain | wc -c
3          # "ok"

$ [ -z "$(rtk git status --porcelain)" ] && echo clean || echo dirty
dirty      # on a clean tree
```

`--porcelain` isn't routed through the compact path — `uses_compact_status_path` returns false for it — so it reaches `filter_status_with_args` and gets rewritten.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets -- -D warnings && cargo test --all` — 2657 passed, 0 failed
- [x] Two regression tests added: `test_filter_status_porcelain_clean_stays_empty` (all four flags) and `test_filter_status_short_still_collapses_to_ok` (guards the compaction that should stay)
- [x] Manual, against a clean throwaway repo: `rtk git status --porcelain` → 0 bytes; `rtk git status` → still `clean — nothing to commit`

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -784,8 +784,16 @@ fn extract_detached_head(raw: &str) -> Option<String> {
         .map(str::to_string)
 }
 
-/// Minimal filtering for git status with user-provided args
-fn filter_status_with_args(output: &str) -> String {
+/// Minimal filtering for git status with user-provided args.
+///
+/// When `args` requests machine-readable output (`--porcelain`, `--porcelain=v1/v2`,
+/// `-z`, `--null`), empty stdout is the documented "clean tree" contract — we must
+/// NOT substitute "ok", because callers parse the output programmatically.
+fn filter_status_with_args(output: &str, args: &[String]) -> String {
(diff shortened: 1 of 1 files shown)


asem89:
Closing — I'm not in a position to sign the CLA, so this can't be merged as-is. Leaving the analysis here in case a maintainer wants to reimplement it; both are small and self-contained, and the reproduction steps in the description stand on their own. No response needed.

PR #3499, opened 2026-08-09
fix(init): resolve migrate_old_hook_script through $CLAUDE_CONFIG_DIR

asem89:
## Summary

- `migrate_old_hook_script` builds its paths from `dirs::home_dir()`, ignoring `$CLAUDE_CONFIG_DIR`.
- Its own sibling `remove_legacy_settings_entries` already uses `resolve_claude_dir()`, so the two disagree about which directory they operate on.
- Fixed to resolve through `resolve_claude_dir()`, with the Cursor root derived from that directory's parent so the same override sandboxes it too.

## Why it matters

**1. It deletes from the wrong directory.** A user who points `CLAUDE_CONFIG_DIR` at a non-default location has `rtk init` remove `hooks/rtk-rewrite.sh` and `hooks/.rtk-hook.sha256` from the real `$HOME/.claude` — the directory they deliberately moved away from — while the configured one is left alone.

**2. The test suite does it to whoever runs it.** Nine existing tests call `run_default_mode` / `run_hook_only_mode` inside a `with_claude_dir_override` sandbox:

```
src/hooks/init.rs:7239, 7261, 7278, 7279, 7298, 7316, 7334, 7356, 7378
```

Every one reaches `migrate_old_hook_script`, which escapes the fixture and deletes the live hook from the developer's real home.

I hit this for real: running `cargo test --all` on an unrelated branch removed my installed `~/.claude/hooks/rtk-rewrite.sh` while `settings.json` still referenced it. RTK was silently disabled — the hook simply stopped rewriting — until I restored the file by hand. The `.rtk-hook.sha256` baseline goes with it, which is also why `rtk verify` starts reporting `no baseline hash found` after a test run.

## Test plan

- [x] `cargo fmt --all --check && cargo clippy --all-targets -- -D warnings && cargo test --all` — 2656 passed, 0 failed
- [x] Added `test_migrate_old_hook_respects_claude_config_dir`. It fails on current `develop`: the sandboxed hook survives while the real one is deleted.
- [x] Verified the restored hook still rewrites correctly afterwards.

Dry-run behaviour is unchanged.

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -1233,51 +1233,62 @@ fn run_default_mode(
 /// can be registered.
 fn migrate_old_hook_script(ctx: InitContext) {
     let InitContext { verbose, dry_run } = ctx;
-    if let Some(home) = dirs::home_dir() {
-        let old_hook = home
-            .join(CLAUDE_DIR)
-            .join(HOOKS_SUBDIR)
-            .join(REWRITE_HOOK_FILE);
-        if old_hook.exists() {
-            if dry_run {
-                println!(
-                    "[dry-run] would migrate legacy hook script: {}",
-                    old_hook.display()
-                );
-            // nosemgrep: filesystem-deletion
-            } else if let Err(e) = std::fs::remove_file(&old_hook) {
+    // Resolve through resolve_claude_dir() so $CLAUDE_CONFIG_DIR applies.
+    // dirs::home_dir() ignores that override, so a user who points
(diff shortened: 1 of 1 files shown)


asem89:
Closing — I'm not in a position to sign the CLA, so this can't be merged as-is. Leaving the analysis here in case a maintainer wants to reimplement it; both are small and self-contained, and the reproduction steps in the description stand on their own. No response needed.

PR #3502, opened 2026-08-09
fix: use line.len() - pos fallback in prisma migration slice

### LINKED ISSUE #3445: prisma_cmd.rs migration-name slice panics with out-of-bounds index when token is short and line-final
Split off from #2954 (per @HetCreep's audit at v0.44.2), same session as #3444 — a different failure mode than the byte-boundary family: an out-of-bounds slice rather than a mid-character one.

`src/cmds/js/prisma_cmd.rs:319-321`:

```rust
let end = line[pos..].find(|c: char| c.is_whitespace()).unwrap_or(20);
latest_migration = line[pos..pos + end].to_string();
```

When the `202…` migration token is the last thing on the line (no trailing whitespace to find) and shorter than 20 bytes, `.unwrap_or(20)` hands back a fixed length that overruns the actual remaining string, and `pos + end` slices past the end.

Note `prisma_cmd.rs:252`, ~60 lines earlier in the same file, already has the correct fallback: `.unwrap_or(line.len() - pos)`. Line 319 just didn't get the same fix.

**Live repro:**

```rust
fn prisma_slice(line: &str) -> String {
    let pos = line.find("202").unwrap();
    let end = line[pos..].find(|c: char| c.is_whitespace()).unwrap_or(20);
    line[pos..pos + end].to_string()
}
fn main() {
    let line = "applied 20240115_x"; // token from pos to end-of-line is 11 bytes, no trailing whitespace
    println!("{}", prisma_slice(line));
}
```
```
thread 'main' panicked at repro.rs:8:9:
end byte index 28 is out of bounds for string of length 18
```

Confirmed exit code 101, matches current `develop` (3044911).

**Suggested fix:** mirror line 252's fallback:

```rust
let end = line[pos..].find(|c: char| c.is_whitespace()).unwrap_or(line.len() - pos);
```

(Credit: @HetCreep's source audit on #2954 first flagged this while confirming the original issue was fixed — filing separately per their suggestion since it's a distinct call site/bug from the original report.)

mmustafasenoglu:
Fixes #3445

Line 319 uses `.unwrap_or(20)` as the fallback when no whitespace is found after the migration token. If the token is shorter than 20 bytes and at end-of-line, this causes an out-of-bounds slice panic.

Line 252 in the same file already has the correct fallback: `.unwrap_or(line.len() - pos)`. This PR applies the same fix to line 319.

### DIFF
--- a/src/cmds/js/prisma_cmd.rs
+++ b/src/cmds/js/prisma_cmd.rs
@@ -316,7 +316,7 @@ fn filter_migrate_status(output: &str) -> String {
             applied_count += 1;
             if latest_migration.is_empty() && line.contains("202") {
                 if let Some(pos) = line.find("202") {
-                    let end = line[pos..].find(|c: char| c.is_whitespace()).unwrap_or(20);
+                    let end = line[pos..].find(|c: char| c.is_whitespace()).unwrap_or(line.len() - pos);
                     latest_migration = line[pos..pos + end].to_string();
                 }
             }


KuSh:
Closing in favour of #3836, which applies the identical line.len() - pos fallback and adds the regression test that filter_migrate_status was missing, covering both the end of line id and the multibyte case. Review and comments are welcome there.

PR #3505, opened 2026-08-09
fix(exec): stop re-parsing already-parsed argv through a second shell

### LINKED ISSUE #3185: bug(windows): redirect character in quoted arguments creates stray empty files (quoting lost on re-execution)

bharvey42:
**Rescoped 2026-09-01.** This PR previously carried nine commits across seven subsystems. Eight of those were cherry-picks of other contributors' still-open PRs that I had been carrying in a local fork build (#3239, #3333, #3344, #3462) — re-proposing them here was duplicative and made the PR unreviewable. They are dropped. What remains is **one commit touching only the exec path.**

## The bug

Four subcommands take clap `trailing_var_arg` vectors, rejoin them with spaces, and hand the resulting string to a second shell (`cmd /C` on Windows, `sh -c` elsewhere):

```
main.rs   Err (1850)   Test (1855)   Summary (1978)   Run (2496)
```

Those varargs have already been split by the shell that invoked rtk, so any operator it recognized was consumed before rtk ran. Re-joining and re-parsing makes the second shell reinterpret metacharacters that were plain data — the mechanism behind #3185, where an argument containing `>` becomes a redirection and silently creates a stray empty file.

Reproduced on 0.43.0 and 0.45.0, all four creating a stray empty file:

```
rtk run     echo "a > b"             -> file 'b'
rtk summary echo "a > stray_summary" -> file 'stray_summary'
rtk err     echo "a > stray_err"     -> file 'stray_err'
rtk test    echo "a > stray_test"    -> file 'stray_test'
```

## The fix

The discriminator is arity, matching the convention `Commands::Proxy` already uses (`main.rs:2528`, *"If a single quoted arg contains spaces, split it respecting quotes (#388)"*):

- **one part** — the user quoted a whole command line, so shell syntax inside it is intentional; keep running it through a shell. `rtk err 'cargo build && cargo test'` is unchanged.
- **many parts** — the invoking shell already parsed it; spawn argv directly via `resolved_command` so every argument reaches the child verbatim and no second shell ever sees it.

`rtk run -c` is documented as the shell-like invocation, so it stays a shell string by contract.

Adds `utils::user_command` and `utils::shell_command`, the latter collapsing the three copies of the platform-shell split that had accumulated in `runner.rs` and `summary.rs`. Joining is now used only to build display and tracking labels, never to build a process.

## Testing

Rebased onto `develop` @ aa40853; applies cleanly.

`cargo test` on the rebased branch: **2894 passed, 2 failed.** The two failures are

- `core::stream::tests::test_read_lines_lossy_preserves_lines_after_invalid_utf8`
- `hooks::hook_cmd::tests::test_claude_pipeline_rewrites_only_safe_final_stage`

and both fail **identically on pristine `develop` without this commit** (verified: 0 passed, 2 failed on a baseline branch), so they are pre-existing and unrelated to this change. Neither touches a file this PR modifies.

One local-build note, not a request: on `x86_64-pc-windows-gnu` current `develop` does not link at all, because `build.rs` emits the MSVC-only `/STACK:8388608` under `#[cfg(windows)]`. I applied #3239 locally to run the suite and deliberately kept it out of this PR. CI only covers `windows-latest` (msvc), so the gnu target is silently unbuildable on `develop` today.

Ref #3185. This is **not** a fix for #3727 — that one is about quote handling when spawning MSYS/Cygwin children, a distinct mechanism; this PR only removes the re-parse.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,13 +3,28 @@ use std::fs;
 use std::path::Path;
 
 fn main() {
-    #[cfg(windows)]
-    {
-        // Clap + the full command graph can exceed the default 1 MiB Windows
-        // main-thread stack during process startup. Reserve a larger stack for
-        // the CLI binary so `rtk.exe --version`, `--help`, and hook entry
-        // points start reliably without requiring ad-hoc RUSTFLAGS.
-        println!("cargo:rustc-link-arg=/STACK:8388608");
+    // Clap + the full command graph can exceed the default 1 MiB Windows
+    // main-thread stack during process startup. Reserve a larger stack for
+    // the CLI binary so `rtk.exe --version`, `--help`, and hook entry
+    // points start reliably without requiring ad-hoc RUSTFLAGS.
+    //
+    // The spelling is linker-specific: MSVC's link.exe takes `/STACK:<bytes>`,
+    // while `*-pc-windows-gnu` links with GNU ld, which reads `/STACK:...` as a
+    // FILE PATH and aborts with `cannot find /STACK:8388608` — the crate does not
+    // link at all on that target.
+    //
+    // Gate on the TARGET, not the host: inside a build script `#[cfg(windows)]`
+    // describes the machine compiling build.rs, so it also fires when
+    // cross-compiling away from Windows and misses cross-compiling toward it.
+    // `CARGO_CFG_*` is the target-accurate source.
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    let target_env = std::env::var("CARGO_CFG_TARGET_ENV").unwrap_or_default();
+    if target_os == "windows" {
+        if target_env == "msvc" {
+            println!("cargo:rustc-link-arg=/STACK:8388608");
+        } else {
+            println!("cargo:rustc-link-arg=-Wl,--stack,8388608");
+        }
     }
 
     let filters_dir = Path::new("src/filters");
--- a/src/cmds/js/playwright_cmd.rs
+++ b/src/cmds/js/playwright_cmd.rs
@@ -288,6 +288,7 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     // Parse output using PlaywrightParser
     let parse_result = PlaywrightParser::parse(&result.stdout);
     let mode = FormatMode::from_verbosity(verbose);
+    let tier = parse_result.tier();
 
     let filtered = match parse_result {
         ParseResult::Full(data) => {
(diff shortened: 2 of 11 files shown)


bharvey42:
Re-ran the repro today as a straight A/B against current `develop` @ aa40853, in an empty directory, since the description only cited 0.43.0/0.45.0:

**Control — `develop` @ aa40853, this commit not applied:**

```
$ rtk run echo "a > stray_control"
$ ls
stray_control          # 4B, and the text never reached stdout
```

**Same command, this commit applied:**

```
$ rtk run echo "a > stray_check"
a > stray_check        # printed verbatim
$ ls
                       # empty
```

So the redirect is still being reinterpreted on mainline today, and the argument reaches the child verbatim with the change. Windows 11, `x86_64-pc-windows-gnu`.

bharvey42:
Rebased onto current `develop` (5e0f92c) — conflict-free again.

Since the last push, `develop` reshaped `src/cmds/rust/runner.rs` to hand a shell-built `Command` to the shared `run_err_cmd`/`run_test_cmd` helpers in core, but the `Command` was still built as `cmd /C <joined string>` / `sh -c <joined string>`, so the re-parse bug was intact. This branch now builds that `Command` from the already-split argv via `user_command(parts)` and keeps upstream's native-test-expression passthrough in `main.rs` untouched.

Verification on Windows 11 (`x86_64-pc-windows-gnu`):

- `cargo test --all`: 3600 passed, 0 failed, 8 ignored. The 3 remaining failures (`find_cmd::run_from_args_propagates_find_exit_status`, `stream::test_read_lines_lossy_preserves_lines_after_invalid_utf8`, `hook_cmd::test_claude_pipeline_rewrites_only_safe_final_stage`) fail identically on a pristine `develop` checkout under Windows — pre-existing, unrelated.
- A/B in an empty directory: the official **v0.49.0** release binary still produces the stray file for `rtk run echo "a > stray_control"` (4 bytes, text never reaches stdout); this branch's build prints `a > stray_check` verbatim and creates nothing.

Note for anyone testing on the `*-pc-windows-gnu` toolchain: `develop` currently cannot link test binaries there (`/STACK:8388608` handed to GNU ld — #1814, fix in #2539). I applied that fix locally for the test run only; it is not part of this PR.

bharvey42:
Some upstream context that landed after this PR was opened, which narrows what it is asking for.

#3728 merged on 18 Sep and fixes the sibling half of this problem — arguments containing a `"` being mangled on their way to MSYS/Cygwin children. Its description draws the line explicitly:

> `rtk err`, `rtk run` and `rtk summary` hand a whole string to `cmd /C` instead of an argument vector, which is a separate problem (#2985, #2389, #3704).

That is this PR. #2389 is still open, and the four subcommands named there (`err`, `run`, `summary`, plus `test`) are exactly the ones this changes. So the argument-vector half is now the only part of the Windows quoting story still unfixed on `develop`, and upstream has documented it as such in a merged PR rather than it being my characterisation.

Two things I re-verified today against `develop` at 0924356:

- **Still merges clean.** #3728 touched `src/core/utils.rs` and `src/main.rs`, which this PR also touches, so I expected a conflict. `git merge-tree --write-tree origin/develop <this branch>` returns a tree with zero conflicts.
- **CI has never run on this PR.** The run for 8957c8e is parked at `conclusion=action_required` — the fork-PR gate waiting on a maintainer to approve workflows — so `statusCheckRollup` is empty and the PR reads as BLOCKED without a single check having executed. Same on #3227. I mention it only because it is not something I can clear from this side; one click would let it prove itself green or red.

Happy to rebase, split it further, or add cases if any of that would help.

PR #3510, opened 2026-08-10
fix(tracking): weight by-command savings percentage

### LINKED ISSUE #3506: `gain`: by-command table puts `SUM(saved_tokens)` next to `AVG(savings_pct)`, producing rows that can't be read together
### What happens

`rtk gain`'s **By Command** table prints a `Saved` column and an `Avg%` column side by side,
but they come from different aggregations, so the two numbers in a row don't describe the
same quantity. On my install the top row reads:

```
 #  Command      Count   Saved    Avg%
 1. rtk grep      1536  478.8M   24.0%
```

478.8M is a `SUM`. 24.0% is an `AVG` of per-command percentages. A reader naturally combines
them — "24% off 478.8M" — and gets nonsense. Against the reported total input of 609.1M, a
478.8M saving is ~78%, not 24%.

### Where

`src/tracking.rs`, `get_by_command`:

```rust
"SELECT rtk_cmd, COUNT(*), SUM(saved_tokens), AVG(savings_pct), AVG(exec_time_ms)
 FROM commands
 GROUP BY rtk_cmd
 ORDER BY SUM(saved_tokens) DESC
 LIMIT 10",
```

`AVG(savings_pct)` weights a command that saved 12 tokens exactly as heavily as one that
saved 12 million. With a long-tailed distribution — which command output always is — the
average percentage is dominated by the many tiny commands while the sum is dominated by the
few huge ones. The columns end up describing disjoint populations.

### Suggested fix

Make the percentage consistent with the sum in the same row:

```rust
"SELECT rtk_cmd, COUNT(*), SUM(saved_tokens),
        100.0 * SUM(saved_tokens) / NULLIF(SUM(input_tokens), 0),
        AVG(exec_time_ms)
 FROM commands GROUP BY rtk_cmd ORDER BY SUM(saved_tokens) DESC LIMIT 10",
```

If the mean-of-percentages is wanted too, it's genuinely useful — but as its own column with
a distinct header (`Avg%` vs `Total%`), because they answer different questions.

### Environment

rtk 0.45.0, Linux x86-64, 10,321 commands over the 90-day window.

cakeni:
﻿## Summary

- derive each by-command savings percentage from aggregate saved and input token totals
- label the weighted value as Total% instead of Avg%
- cover weighted long-tail command data and zero-input passthrough records

Closes #3506

## Test plan

- [x] cargo fmt --all --check
- [x] cargo clippy --all-targets
- [x] cargo test --all -- --test-threads=1
- [x] targeted regression test confirmed red before the fix (expected 89.2079%, got 50%) and green after it
- [x] manual testing with an isolated RTK_DB_PATH: rtk proxy followed by rtk gain displayed the Total% column

AI assistance was used during investigation or implementation.

### DIFF
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -204,7 +204,7 @@ pub fn run(
             println!("{}", "─".repeat(table_width));
             println!(
                 "{:>3}  {:<cmd_width$}  {:>count_width$}  {:>saved_width$}  {:>6}  {:>time_width$}  {:<impact_width$}",
-                "#", "Command", "Count", "Saved", "Avg%", "Time", "Impact",
+                "#", "Command", "Count", "Saved", "Total%", "Time", "Impact",
                 cmd_width = cmd_width, count_width = count_width,
                 saved_width = saved_width, time_width = time_width,
                 impact_width = impact_width
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -127,7 +127,7 @@ pub struct GainSummary {
     pub total_time_ms: u64,
     /// Average execution time per command (milliseconds)
     pub avg_time_ms: u64,
-    /// Top 10 commands by tokens saved: (cmd, count, saved, avg_pct, avg_time_ms)
+    /// Top 10 commands by tokens saved: (cmd, count, saved, total_pct, avg_time_ms)
     pub by_command: Vec<(String, usize, usize, f64, u64)>,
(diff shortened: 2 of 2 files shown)


KuSh:
Closing in favour of #891, which makes the same weighted rate fix to get_by_command, has already been approved, and adds a changelog entry and a warning on the CommandStats alias. The in-memory tests here, especially the zero input case, are a good addition, so please offer them on #891 if you would like them kept.

PR #3517, opened 2026-08-10
fix(discover): suppress false savings in Claude hook mode

### LINKED ISSUE #3472: rtk discover reports already-hook-rewritten commands as missed savings (inflated estimate in Claude Code hook mode)
## Summary

When rtk is installed as a Claude Code **PreToolUse hook** (the `rtk hook claude` mode that `rtk init` sets up), `rtk discover` cannot see any of the rewrites that hook performs. It reads session transcripts, which record the command the *model emitted*, never the hook's `updatedInput`. The result is that already-optimized commands are reported as "missed savings", and the headline estimate is inflated by a large factor.

On my install it reported **~19.2M tokens saveable** across commands that were, in fact, already being rewritten on every invocation.

## Mechanism

The Claude Code PreToolUse hook contract is: the hook receives the tool input on stdin and may return `hookSpecificOutput.updatedInput` to replace it. `rtk hook claude` does exactly this:

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"grep -n foo bar.txt"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite",
 "updatedInput":{"command":"rtk grep -n foo bar.txt"}}}
```

The transcript, however, retains the pre-hook `tool_use.input.command` (`grep -n foo bar.txt`). The rewritten form is never written back. So any analysis keyed on transcript text sees 100% unrouted, regardless of how much the hook actually rewrote.

## Evidence

**1. The two datasets do not overlap.** Same executions, 30-day window:

```
transcripts (what `discover` reads)     76,906 Bash commands
  ...that begin with "rtk "                 20   (0.026%)
history.db (what actually executed)     42,738 rtk-routed rows
```

Those 20 are commands I typed with an explicit `rtk` prefix by hand. If transcripts recorded `updatedInput`, these two numbers would overlap heavily. They do not overlap at all.

**2. Most of the "missed savings" list is already routed.** I took every entry `rtk discover` reported and fed its example command back through the live `rtk hook claude`:

| discover says "missed" | count | actually rewritten by the hook? |
|---|---|---|
| `grep -n` | 23,009 | yes, to `rtk grep` |
| `git show` | 16,264 | yes, to `rtk git` |
| `gh pr` | 10,739 | yes, to `rtk gh` |
| `ls -la` | 8,706 | yes, to `rtk ls` |
| `cat -n` | 6,241 | yes, to `rtk read` |
| `find .` | 2,350 | yes, to `rtk find` |
| `wc -l` | 2,043 | yes, to `rtk wc` |
| ... | | |

**27 of the 36** entries in the "MISSED SAVINGS" table were already being rewritten. The same applies to the "TOP UNHANDLED" section: its #1 entry, `cat` at 13,749 calls, rewrites to `rtk read`.

## Impact

- The headline savings estimate is not usable for prioritisation. Acting on it means building filters for commands that are already filtered.
- It reports worst on the installs that are configured best. The more thoroughly the hook is wired up, the more inflated the number becomes.
- `Already using RTK: 72 commands (0.0%)` sits directly above it, contradicting `rtk session`, which reported 21% average adoption for the same period. Two rtk subcommands disagreeing by that margin makes it hard to know which to trust.

## Suggested fixes

1. **Detect hook mode and say so.** `rtk init --show` already knows whether the hook is registered. When it is, `discover` could print a banner that transcript-derived numbers cannot account for hook rewrites, rather than presenting them as actionable.
2. **Reconcile against `history.db` before reporting.** history.db has the executed form (`rtk_cmd`) and a timestamp. Subtracting commands with a corresponding routed execution in the same window would remove most of the phantom entries.
3. **Failing both, drop the token estimate** in hook mode and report only the unhandled-command list, which stays useful.

## Minor, same command

`rtk discover` defaults to the current project and silently reports `Scanned: 0 sessions, 0 Bash commands / No missed savings found. RTK usage looks good!` when run from a directory that is not a known project (a scratch dir, for instance). That reads as a clean bill of health rather than "I had nothing to look at". A distinct message for the zero-sessions case would help.

## Environment

```
rtk 0.45.0 (Homebrew, rtk-ai/tap)
macOS 15 (Darwin 25.5.0), Apple Silicon
Claude Code, PreToolUse hook: `rtk hook claude`
~42.7K routed commands in history.db, 62.4M tokens saved (31.2%) per `rtk gain`
```

zhiwuyazhe-fjr:
## Summary
- detect active Claude RTK `PreToolUse` hooks in global and project settings
- stop presenting pre-hook transcript commands as missed savings or adoption data in text and JSON output
- keep unsupported-command discovery actionable and report zero-session scans honestly
- document the transcript visibility limitation

## Test plan
- `cargo fmt --all --check`
- `cargo test discover::report::tests`
- `cargo test --all -- --test-threads=1` (2582 unit tests plus integration suites)

Fixes #3472

Written with Codex (GPT-5).

### DIFF
--- a/src/discover/README.md
+++ b/src/discover/README.md
@@ -48,6 +48,11 @@ When a hook sends `cargo fmt --all && cargo test 2>&1 | tail -20`:
 3. Classifies each command against the same rules used for live rewriting
 4. Aggregates results: which commands could have been rewritten, estimated token savings, adoption rate
 
+Claude Code transcripts store the model's original command, not the command returned by a
+`PreToolUse` hook. When `discover` detects an active Claude RTK hook, it therefore suppresses
+missed-savings and adoption estimates instead of presenting pre-hook commands as missed. The
+unsupported-command list remains available because those commands are not rewritten by RTK.
+
 The classification logic is shared between discover and rewrite — same patterns, same rules, different consumers.
 
 ## Env Prefix Handling
--- a/src/discover/README.md
+++ b/src/discover/README.md
@@ -63,6 +63,11 @@ The permission gate, discover/analytics classification, and rewrite each segment
 3. Classifies each command against the same rules used for live rewriting
 4. Aggregates results: which commands could have been rewritten, estimated token savings, adoption rate
 
+Claude Code transcripts store the model's original command, not the command returned by a
+`PreToolUse` hook. When `discover` detects an active Claude RTK hook, it therefore suppresses
+missed-savings and adoption estimates instead of presenting pre-hook commands as missed. The
+unsupported-command list remains available because those commands are not rewritten by RTK.
+
 The classification logic is shared between discover and rewrite — same patterns, same rules, different consumers.
 
 ## Env Prefix Handling
--- a/src/discover/report.rs
+++ b/src/discover/report.rs
@@ -1,9 +1,11 @@
 //! Data types for reporting which commands RTK can and cannot optimize.
 
 use crate::hooks::constants::{
-    COPILOT_HOOK_FILE, CURSOR_DIR, GITHUB_DIR, HERMES_DIR, HERMES_PLUGINS_SUBDIR,
-    HERMES_PLUGIN_MANIFEST_FILE, HERMES_PLUGIN_NAME, HOOKS_SUBDIR, REWRITE_HOOK_FILE,
+    CLAUDE_DIR, COPILOT_HOOK_FILE, CURSOR_DIR, GITHUB_DIR, HERMES_DIR, HERMES_PLUGINS_SUBDIR,
+    HERMES_PLUGIN_MANIFEST_FILE, HERMES_PLUGIN_NAME, HOOKS_SUBDIR, PRE_TOOL_USE_KEY,
+    REWRITE_HOOK_FILE, SETTINGS_JSON, SETTINGS_LOCAL_JSON,
 };
+use crate::hooks::is_claude_hook_command;
 use serde::Serialize;
 use std::path::Path;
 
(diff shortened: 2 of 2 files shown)


zhiwuyazhe-fjr:
Rebased onto the current `develop` tip (0f8b26c) — this also clears the merge conflict GitHub was reporting.

The conflict was in `src/discover/report.rs`: since this PR branched off, develop added the `already_rtk_estimated` disclosure block and the `RTK_DISABLED` early-return guard in exactly the region this PR reworks. Resolution keeps both semantics:

- the `sessions_scanned == 0` early return and the Claude-hook branch from this PR are preserved;
- the estimated-coverage caveat is now scoped to the non-hook branch, so adoption estimates stay fully suppressed when a Claude PreToolUse hook is detected;
- the `RTK_DISABLED` guard stays in the empty-report early return, so the hook-mode variant of that message can't hide a 100%-bypassed setup either.

Verified locally on Windows:

- `discover` tests: 609 passed, 0 failed
- Full suite: 3525 passed / 1 failed (8 ignored) — the single failure is `find_cmd::run_from_args_propagates_find_exit_status`, which fails identically on a clean `develop` checkout under Windows (find.exe vs GNU find exit semantics), unrelated to this change

Heads-up for reviewers: workflow runs from this fork are approval-gated, so the CI run for this push is in `action_required` — it needs a maintainer approval from the Actions tab before the checks will execute.

KuSh:
Thanks for the report and the patch, the inflated adoption numbers you saw were real. #3206, merged on 2026-09-02 and shipped in v0.48.0, fixed the underlying problem by joining transcript entries against the hook decision log by tool_use_id, so commands the Claude PreToolUse hook already rewrote are counted as already using RTK instead of missed savings, and the part of the number that can only be inferred is labelled as estimated. That approach deliberately replaced reading settings.json to guess today's hook state, which is what this PR's suppression is built on, and the hunk that adds the zero-sessions message also removes the new disclosure block, so it cannot be rebased as written. One thing you found is still true on develop: a scan that finds no sessions at all still prints "No missed savings found. RTK usage looks good!", and a small standalone PR adding a sessions_scanned == 0 guard above src/discover/report.rs:187 would be welcome. If something here was missed, please comment and this can be reopened.

PR #3519, opened 2026-08-10
fix(diff): don't report byte-different files as identical

### LINKED ISSUE #3469: rtk diff reports "Files are identical" (exit 0) for byte-different files

ousamabenyounes:
## Summary

- `rtk diff` could report `[ok] Files are identical` with exit 0 for byte-different files whose only differences are hidden by `str::lines()`, such as CRLF-vs-LF or a trailing newline (#3469).
- Fix: pass the original file contents into the renderer and only use the identical branch when `content1 == content2`. If the line diff is empty but the file contents differ, `rtk diff` now prints an explicit whitespace/line-ending message and exits 1.
- This is rebased onto current `develop` and keeps the newer `select_file_diff_output` path, so the empty-diff message bypasses the tiny-output `never_worse` fallback that swallowed the reviewer-requested small-file case.
- Added Hydr0gen19's byte-identical CRLF guard test from #3471, adapted to this PR's current renderer signature with no fifth `render_file_diff`/renderer argument.

## Test verification (RED -> GREEN)

New coverage:

- Unit tests for CRLF-vs-LF, trailing-newline-only, and byte-identical CRLF content.
- Run-level integration test for a small CRLF-vs-LF file pair through `rtk diff`, covering the review feedback path.

**RED on unmodified `origin/develop`**

```
assertion `left == right` failed: [ok] Files are identical

  left: Some(0)
 right: Some(1)
```

**GREEN on this branch**

```
test small_crlf_vs_lf_diff_prints_whitespace_message ... ok
```

Local CI replay:

- Baseline `run-ci.sh` on `origin/develop`: test-presence, fmt, clippy, and `cargo test --all` passed; unit suite `2929 passed; 0 failed; 8 ignored`.
- Final `run-ci.sh` on this branch: test-presence, fmt, clippy, and `cargo test --all` passed; unit suite `2932 passed; 0 failed; 8 ignored`; new integration test `1 passed`.
- Coverage: `cargo llvm-cov --all` passed, and the Rust changed-line LCOV check found covered executable changed lines only (`changed executable lines with coverage data: 80`, PASS).

## Test plan

- [x] `cargo test --all`
- [x] `cargo llvm-cov --lcov --output-path ... --all`
- [x] RED proof on unmodified `origin/develop` for the new run-level regression test

> All PRs target the `develop` branch.

Fixes #3469

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -27,7 +27,7 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
     let fallback = format_classic_diff(&diff);
     let both_files = format!("{}\n---\n{}", content1, content2);
 
-    let (rtk, exit_code) = render_diff(file1, file2, &diff, &content1, &content2);
+    let (rtk, exit_code) = render_diff(file1, file2, &diff, content1 == content2);
 
     let shown = select_file_diff_output(&diff, &fallback, &rtk);
     print!("{}", shown);
@@ -44,15 +44,9 @@ fn render_file_header(file1: &Path, file2: &Path) -> String {
     format!("{} → {}\n", file1.display(), file2.display())
 }
 
-fn render_diff(
-    file1: &Path,
-    file2: &Path,
-    diff: &DiffResult,
-    content1: &str,
-    content2: &str,
-) -> (String, i32) {
+fn render_diff(file1: &Path, file2: &Path, diff: &DiffResult, bytes_equal: bool) -> (String, i32) {
     if diff.changes.is_empty() {
-        if content1 == content2 {
+        if bytes_equal {
             return (IDENTICAL_FILES_MESSAGE.to_string(), 0);
         }
         // `str::lines()` strips `\r` and drops a trailing newline, so these
@@ -401,8 +395,7 @@ mod tests {
             Path::new(file1),
             Path::new(file2),
             &diff,
-            content1,
-            content2,
+            content1 == content2,
         )
     }
 
@@ -587,13 +580,8 @@ mod tests {
     fn test_never_worse_fallback_is_a_classic_diff() {
         let diff = compute_diff(&["alpha beta"], &["alpha zzzz"]);
         let fallback = format_classic_diff(&diff);
-        let (rendered, code) = render_diff(
-            Path::new("before"),
-            Path::new("after"),
-            &diff,
-            "alpha beta",
-            "alpha zzzz",
-        );
+        let (rendered, code) =
+            render_diff(Path::new("before"), Path::new("after"), &diff, false);
         let shown = select_file_diff_output(&diff, &fallback, &rendered);
 
         assert_eq!(code, 1);
@@ -622,8 +610,7 @@ mod tests {
             Path::new("a"),
             Path::new("b"),
             &diff,
-            &old_content,
-            &new_content,
+            old_content == new_content,
         );
         let shown = select_file_diff_output(&diff, &fallback, &rendered);
         let baseline = tracking_baseline(&diff, &fallback, &both_files, shown);
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -6,6 +6,10 @@ use anyhow::Result;
 use std::fs;
 use std::path::Path;
 
+const IDENTICAL_FILES_MESSAGE: &str = "[ok] Files are identical\n";
+const WHITESPACE_ONLY_DIFF_DETAIL: &str =
+    "   files differ only in whitespace or line endings (no line-content change)\n";
+
 /// Ultra-condensed diff - only changed lines, no context.
 /// Returns the diff-convention exit code: 0 if identical, 1 if files differ.
 pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
@@ -23,7 +27,7 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
     let fallback = format_classic_diff(&diff);
     let both_files = format!("{}\n---\n{}", content1, content2);
 
-    let (rtk, exit_code) = render_diff(file1, file2, &diff);
+    let (rtk, exit_code) = render_diff(file1, file2, &diff, &content1, &content2);
 
     let shown = select_file_diff_output(&diff, &fallback, &rtk);
     print!("{}", shown);
@@ -36,13 +40,35 @@ pub fn run(file1: &Path, file2: &Path, verbose: u8) -> Result<i32> {
     Ok(exit_code)
 }
 
-fn render_diff(file1: &Path, file2: &Path, diff: &DiffResult) -> (String, i32) {
+fn render_file_header(file1: &Path, file2: &Path) -> String {
+    format!("{} → {}\n", file1.display(), file2.display())
+}
+
+fn render_diff(
+    file1: &Path,
+    file2: &Path,
+    diff: &DiffResult,
+    content1: &str,
+    content2: &str,
+) -> (String, i32) {
     if diff.changes.is_empty() {
-        return ("[ok] Files are identical\n".to_string(), 0);
+        if content1 == content2 {
+            return (IDENTICAL_FILES_MESSAGE.to_string(), 0);
+        }
+        // `str::lines()` strips `\r` and drops a trailing newline, so these
+        // byte-level differences can leave no line changes to render.
+        return (
+            format!(
+                "{}{}",
+                render_file_header(file1, file2),
+                WHITESPACE_ONLY_DIFF_DETAIL
+            ),
+            1,
+        );
     }
 
     let mut rtk = String::new();
-    rtk.push_str(&format!("{} → {}\n", file1.display(), file2.display()));
+    rtk.push_str(&render_file_header(file1, file2));
     rtk.push_str(&format!(
         "   +{} added, -{} removed, ~{} modified\n\n",
         diff.added, diff.removed, diff.modified
@@ -367,6 +393,19 @@ fn condense_unified_diff(diff: &str) -> String {
 mod tests {
     use super::*;
 
+    fn render_test_diff(file1: &str, file2: &str, content1: &str, content2: &str) -> (String, i32) {
+        let lines1: Vec<&str> = content1.lines().collect();
+        let lines2: Vec<&str> = content2.lines().collect();
+        let diff = compute_diff(&lines1, &lines2);
+        render_diff(
+            Path::new(file1),
+            Path::new(file2),
+            &diff,
+            content1,
+            content2,
+        )
+    }
+
     // --- similarity ---
 
     #[test]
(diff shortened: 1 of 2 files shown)


src/cmds/git/diff_cmd.rs
```diff
@@ -40,7 +40,22 @@ fn render_file_diff(file1: &Path, file2: &Path, content1: &str, content2: &str)
...
+        // `str::lines()` strips `\r` and drops a trailing newline, so CRLF-vs-LF
+        // and trailing-newline-only differences leave no line-level change.
+        // Claiming "identical" for byte-different files is a false-equal result
+        // that defeats `rtk diff` used to verify equality (issue #3469). Report
+        // identical only when the bytes truly match; otherwise surface the
+        // whitespace/line-ending difference and exit 1 per the diff convention.
+        if content1 == content2 {
+            return ("[ok] Files are identical\n".to_string(), 0);
+        }
+        let msg = format!(
```
KuSh: This message is ~92 bytes (≈23 estimated tokens) plus both paths, and `run()` prints `never_worse(&raw, &rtk)` where `raw = format!("{}\n---\n{}", content1, content2)`. So whenever the two files together are under roughly 90 bytes, the guard discards this message and prints the raw concatenation instead — which is precisely the small-file case #3469 reports.

Measured on your own build:

```
$ rtk diff /tmp/lf.txt /tmp/crlf.txt      # "alpha\nbeta\n" vs "alpha\r\nbeta\r\n"
alpha
beta

---
alpha
beta
exit=1                                    # message never printed

$ rtk diff big_lf.txt big_crlf.txt        # same content, 5 longer lines
/tmp/big_lf.txt → /tmp/big_crlf.txt
   files differ only in whitespace or line endings (no line-content change)
exit=1
```

Two ways out, either is fine:

1. Shorten the message enough to survive the guard, e.g. `"{} != {}: whitespace/EOL only\n"`.
2. Return early past `never_worse` for this branch — the raw fallback is actively misleading here, since it renders two byte-different files as two identical-looking blocks.

I'd lean towards (2) with a short comment saying why, because (1) leaves the behaviour size-dependent and someone will lengthen the string later.

Either way this needs a test through `run()` rather than `render_file_diff()`. Both new unit tests call `render_file_diff` directly, so they pass regardless of what the guard does — the bug is invisible to them.

KuSh (changes requested):
Requesting changes on one point. The fix itself is right and I want it in — this is about it actually reaching the user.

**Verified:** `diff.changes.is_empty()` is equivalent to "the `lines()` vectors are element-wise equal", so the only byte differences that can survive that check are exactly `\r\n`-vs-`\n` and a trailing newline. The guard's message is therefore accurate, not a guess, and the exit code correctly becomes 1. `cargo fmt --check`, `cargo clippy --all-targets` and the 25 `diff_cmd` tests all pass on your head. I also ran the binary end-to-end against the cases in #3469.

**The problem:** for small files the message never gets printed. Details inline.

Worth knowing, since it explains why this is worth another round: I confirmed on 0.46.0 that the *pre-existing* behaviour already prints a raw dump for these inputs —

```
$ printf 'a\r\nb\r\nc\r\n' > x; printf 'a\nb\nc\n' > y
$ rtk diff x y; echo "exit=$?"
a
b
c

---
a
b
c
exit=0
```

— so a user comparing two small files today sees two visually identical blobs and exit 0. Your PR fixes the exit code half of that. If the message is also swallowed, they get two visually identical blobs and exit 1, which is better but still leaves them to work out *why* on their own. That's the gap worth closing while you're in here.

KuSh:
Also, since @Hydr0gen19 has already approved it, could you include the `test_render_byte_identical_exit_zero_with_crlf` test from #3471 in your PR?

Hydr0gen19:
Yes, please take it — copy it straight across, no attribution needed.

One adjustment when you do: the version on #3471 passes a fifth argument to `render_file_diff` (`false`), which belongs to an `--ignore-whitespace` flag that this PR rightly does not carry. Drop that argument and it compiles against your signature unchanged. Your guard sits inside the `diff.changes.is_empty()` branch rather than ahead of the split, but the case still reaches your identical path, so the assertions hold either way.

Why it is worth having next to your two: both of yours pin the "must differ" direction. This one pins the other side — two files that are both CRLF and byte-identical must still report identical and exit 0. A guard that over-triggers whenever it sees a CRLF would pass `test_render_crlf_vs_lf_not_identical` and `test_render_trailing_newline_not_identical`, and fail only this one.

ousamabenyounes:
Thanks @KuSh and @Hydr0gen19. Rebased onto current `develop` and pushed `f4325bc`: the whitespace/EOL-only branch now uses the existing empty-diff output path to bypass the size guard, so small CRLF/LF files print the explanatory message and exit 1 instead of being swallowed into `[ok] Files are identical`.

I also included Hydr0gen19's byte-identical CRLF guard test from #3471, adapted to this PR's current renderer signature with no fifth `render_file_diff`/renderer argument.

Validation: the new run-level regression is RED on `origin/develop` (`[ok] Files are identical`, exit 0) and GREEN here. Local `run-ci.sh` baseline/final passed test-presence, fmt, clippy, and `cargo test --all` (`2929 -> 2932` unit tests, plus the new integration test). `cargo llvm-cov` and the Rust changed-line coverage check passed.

Remote CI on `f4325bc`: test presence, fmt, doc review, clippy, Ubuntu/macOS/Windows tests, security scan, semgrep, and CLA passed. `benchmark` failed on the existing `golangci-lint` empty-output filter; the current `develop` run `33541678720` fails the same benchmark step with the same `BENCHMARK FAILED: 1 filter(s) returned empty output` summary.

KuSh (approved):
The message reaches the user now — clearing my earlier request for changes.

Rebasing onto `develop` was the right move and it solved this structurally rather than by special-casing. `select_file_diff_output` returns the rendered output unguarded when `changes.is_empty()`, which is exactly the whitespace branch, so the message survives at any file size instead of only above ~90 bytes:

```console
$ rtk diff lf.txt crlf.txt          # 11 bytes each — previously printed a raw dump
lf.txt → crlf.txt
   files differ only in whitespace or line endings (no line-content change)
exit=1
```

And `tests/diff_byte_accuracy_test.rs` is the run()-level test I asked for — it spawns the real binary on the *small* fixture, which is precisely the case the unit tests were blind to.

### Verified

`cargo fmt --all --check` and `cargo clippy --all-targets` clean, 3016 tests passing. Exit codes match `/usr/bin/diff` across the edge cases, and each branch is genuinely reachable:

| | verdict | exit (rtk / GNU) |
|---|---|---|
| `a\n` vs `a` | whitespace message | 1 / 1 |
| CRLF vs LF | whitespace message | 1 / 1 |
| `\n` vs `\r\n` | whitespace message | 1 / 1 |
| `a\r\n` vs `a\r\r\n` | real diff | 1 / 1 |
| `a\r` vs `a` | real diff | 1 / 1 |
| byte-identical | identical | 0 / 0 |

The last two matter: they confirm the message isn't over-claiming. `str::lines()` loses exactly two things — whether the final line ends in `\n`, and whether a `\r` preceded each `\n` — and both are line terminators, so "differ only in whitespace or line endings" is exact by construction rather than a best guess.

I also mutation-checked the new tests rather than trusting that they pass:

- byte check → line check (the original bug): both new unit tests fail.
- whitespace branch returning 0 instead of 1: both fail.
- removing the guard bypass so the message is swallowed again: `small_crlf_vs_lf_diff_prints_whitespace_message` fails with empty stdout.

That third one is the important one — the regression I flagged is now genuinely pinned, not passing by accident.

### One small change pushed

I pushed `428af61` on top: `render_diff` was taking `content1` and `content2` only to compare them, which gave a rendering function five parameters and an input comparison to do. The caller already holds both strings, so it now passes `content1 == content2` as a `bytes_equal` verdict. Behaviour is identical — same tests, same output, same exit codes, re-verified after the change. Revert it if you'd rather keep the contents in the signature.

Issue #3486, opened 2026-08-09
Regression: du filter (fixed in v0.30.0, PRs #349/#351/#386, #284) is absent again in v0.45.0


nitrocode:
## Summary

`du` support appears to have regressed. Issue #284 requested a `du` rewrite/filter and was closed as completed on 2026-03-10, with maintainer confirmation:

> Resolved — RTK now ships a built-in TOML filter for `du` (see `src/filters/du.toml`, landed in PRs #349, #351, #386). Tested savings: 96.8% on `du -h .` (2602 → 84 tokens).
> — @pszymkowiak, [#284](https://github.com/rtk-ai/rtk/issues/284)

On rtk **v0.45.0**, `du` is not a recognized subcommand and `rtk rewrite` refuses to rewrite `du ...` invocations. This is the third instance of this exact pattern found in the same audit — see the parallel `helm` (https://github.com/rtk-ai/rtk/issues/3484) and `gcloud` (https://github.com/rtk-ai/rtk/issues/3485) regressions filed alongside this issue — suggesting a shared root cause (e.g. a batch of built-in TOML filters dropped from the dispatcher in some release between v0.30.0 and v0.45.0) rather than three independent regressions.

## Version

```
$ rtk --version
rtk 0.45.0
```

## Reproduction

```
$ rtk --help | grep -i du
$ echo $?
1

$ rtk rewrite "du -sh ."
$ echo $?
1
```

Expected (per #284 resolution): `rtk rewrite "du -sh ."` routes through the built-in `du.toml` filter. Actual: no `du` subcommand exists, exits 1 with no output, falls back to raw passthrough (0 filtered invocations across local history).

## Ask

Given this is the third confirmed instance (alongside #182/#183 for helm/gcloud) of a "fixed via built-in TOML filter, now missing" pattern, worth checking whether `src/filters/du.toml`, `helm.toml`, and `gcloud.toml` were all removed together in the same commit/release, rather than debugging each individually.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 90% |

### Summary

[...]

Issue #3485, opened 2026-08-09
Regression: gcloud filter support (fixed in v0.30.0, #183) is absent again in v0.45.0


nitrocode:
## Summary

`gcloud` support appears to have regressed. Issue #183 tracked adding a gcloud filter and was closed as completed on 2026-03-16 with the comment:

> Fixed in v0.30.0 — gcloud.toml TOML filter. Tested locally: `rtk gcloud compute instances list` strips blank lines, keeps results.
> — @pszymkowiak, [#183](https://github.com/rtk-ai/rtk/issues/183)

On rtk **v0.45.0** (15 minor versions later), `gcloud` is not a recognized subcommand and `rtk rewrite` refuses to rewrite `gcloud ...` invocations. Either the v0.30.0 fix was reverted, or it never shipped in a released binary that reached users. Same pattern as the parallel `helm` regression (#182, see https://github.com/rtk-ai/rtk/issues/3484) filed alongside this issue.

## Version

```
$ rtk --version
rtk 0.45.0
```

## Reproduction

```
$ rtk --help | grep -i gcloud
$ echo $?
1

$ rtk rewrite "gcloud compute instances list"
$ echo $?
1

$ rtk rewrite "gcloud dns record-sets list --zone=example"
$ echo $?
1
```

Expected (per #183 resolution): `rtk rewrite "gcloud compute instances list"` routes through the `gcloud.toml` filter; `rtk --help` lists `gcloud` among supported subcommands.

Actual: no `gcloud` subcommand exists, `rtk rewrite` exits 1 with no output for any `gcloud ...` input, and every `gcloud` call falls back to raw passthrough (0 filtered invocations across local history).

## Impact

Any user who added `gcloud` to `[hooks] exclude_commands` as a workaround is getting zero token savings on gcloud output, which was the entire point of #183.

## Ask

- Confirm whether `gcloud.toml` still exists in the current tree and, if so, why the CLI dispatcher does not route to it.
- If it was removed, note that in the changelog so downstream users know to stop expecting it.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 72% |

### Summary

[...]

Issue #3484, opened 2026-08-09
Regression: helm filter support (fixed in v0.30.0, #182) is absent again in v0.45.0


nitrocode:
## Summary

`helm` support appears to have regressed. Issue #182 tracked adding a helm filter and was closed as completed on 2026-03-16 with the comment:

> Fixed in v0.30.0 — helm.toml TOML filter. Tested locally: `rtk helm install` strips blank lines, keeps results.
> — @pszymkowiak, [#182](https://github.com/rtk-ai/rtk/issues/182)

On rtk **v0.45.0** (15 minor versions later), `helm` is not a recognized subcommand and `rtk rewrite` refuses to rewrite `helm ...` invocations. Either the v0.30.0 fix was reverted, or it never shipped in a released binary that reached users.

## Version

```
$ rtk --version
rtk 0.45.0
```

## Reproduction

```
$ rtk --help | grep -i helm
$ echo $?
1

$ rtk rewrite "helm dependency build"
$ echo $?
1

$ rtk rewrite "helm template mychart ."
$ echo $?
1
```

Expected (per #182 resolution): `rtk rewrite "helm dependency build"` prints a rewrite that routes through the `helm.toml` filter; `rtk --help` lists `helm` among supported subcommands.

Actual: no `helm` subcommand exists, `rtk rewrite` exits 1 with no output for any `helm ...` input, and every `helm` call falls back to raw passthrough (0 filtered invocations across local history).

## Impact

Any user who added `helm` to `[hooks] exclude_commands` as a workaround is getting zero token savings on helm output, which was the entire point of #182.

## Ask

- Confirm whether `helm.toml` still exists in the current tree and, if so, why the CLI dispatcher does not route to it.
- If it was removed, note that in the changelog so downstream users know to stop expecting it.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 72% |

### Summary

helm subcommand/filter (previously added in v0.30.0 per #182) is missing again in v0.45.0 — verify whether helm.toml still exists and why the dispatcher doesn't route to it, or fix if regressed.

<details>
<summary>📁 Relevant files</summary>

[...]

Issue #284, opened 2026-02-28
Feature request: add rtk rewrite for `du`


oysi2025:
## Context

`rtk discover --all` shows `du` as an unhandled command:

- **84 invocations** in the last 30 days
- Typical usage: checking disk usage of directories and log files

## Request

Add an RTK rewrite/filter for `du` output. Recursive directory listings can produce hundreds of lines that could be summarized.

## Example commands

```
sudo du -sh /var/log/* 2>/dev/null
du -sh /home/ubuntu/services/*
du -h --max-depth=1 /home/ubuntu
```

Found via `rtk discover --all`.

PR #3538, opened 2026-08-11
fix(du): keep summary and depth-limited output raw

veltri-23:
## Summary

- `du` summary and depth-limited output now runs raw instead of being cut to 40 lines. Every row of `du -s` (or `-d N`) is an independent total for its own root, so the cap silently dropped the largest consumers while reporting a confident answer. A 312-row cache survey came back 42 rows with the 4.4 GB entry missing; exit code stayed 0.
- Added a small option to the TOML filter engine, `pass_through_if_args`, so any declarative filter can opt out of filtering for specific invocation shapes. `du.toml` uses it for `-s`, `--summarize`, `-d`, `--max-depth`. Plain tree output (`du -h .`) keeps the existing aggressive compression.
- Also checked the #3486 report that `du` disappeared from the rewrite path: on `develop` (and in the v0.45.0 tag) the rewrite rule, `rtk du` subcommand, and `du.toml` are all present. The reachable defect was the shape-blind line cap, fixed here.

## Test plan

- [x] Manual verification (60 roots, sizes 5 KB to 200 KB):

| Invocation | Before | After |
|---|---|---|
| `rtk du -sk <60 roots>` | 40 rows + `... (20 lines truncated)` + tee hint | 60 rows, identical to raw `du` |
| `rtk du -h -d1 <tree>` | capped at 40 | raw |
| `rtk du -s <tree>` | capped at 40 | raw |
| `rtk du -h <tree>` | compressed | still compressed (savings kept) |
| `rtk rewrite "du -sh ."` | `rtk du -sh .` | unchanged |

- Unit tests: 2572 passed, 0 failed (2 new for the arg gate). The `copilot_selfheal_test` suite fails 10 tests, confirmed identical on clean `develop` before this change (Windows environment), unrelated.

## Details

Reproduction and mechanism. `du.toml` caps output at 40 lines. For a single-root tree (`du -h .`) interior rows are redundant with their parent totals, so the cap reads correctly. For `du -sk <glob>` every line is an independent answer and the drop is arbitrary (listing order, not size order), so a survey can understate the largest consumer by 17x while exiting 0.

Why this shape. Options considered: dropping the cap entirely would kill the 96.8% savings on `du -h .` that #284 shipped; special-casing `du` in the runner would be a per-path hack. The TOML engine had no flag awareness at all, so the option lives in the filter definition, default empty, backwards compatible with the existing `deny_unknown_fields` schema.

Edge cases covered by prefix matching: combined short flags (`-sk`, `-sh`), value-bearing long flags (`--max-depth=1`), space-separated forms (`-d 1`, `--max-depth 1`). Flags that stay filtered: `-h`, `-a`, `-c`, no-flag invocations.

Scope note. #3484 (helm) and #3485 (gcloud) report the same "filter missing" pattern. Their rewrite rules and `.toml` files are also present on `develop`, so like the `du` rewrite half they do not reproduce from source. They are `needs-reproduction`; left untouched.

Refs #3486

### DIFF
--- a/src/core/toml_filter.rs
+++ b/src/core/toml_filter.rs
@@ -167,11 +167,30 @@ pub struct CompiledFilter {
 impl CompiledFilter {
     /// True when any invoked arg matches (exact or prefix) an entry in
     /// `pass_through_if_args`. Prefix matching covers combined short flags
-    /// (`-sk`) and value-bearing long flags (`--max-depth=1`).
+    /// (`-sk`) and value-bearing long flags (`--max-depth=1`). Short-option
+    /// clusters where the target flag is not first (`-hs`, `-hd1`) are matched
+    /// by scanning the cluster for the single-char needle.
     pub fn should_pass_through(&self, args: &[String]) -> bool {
         self.pass_through_if_args.iter().any(|needle| {
-            args.iter()
-                .any(|arg| arg == needle || arg.starts_with(needle))
+            args.iter().any(|arg| {
+                if arg == needle || arg.starts_with(needle) {
+                    return true;
+                }
+                // Short-option cluster: a single-dash token carrying the
+                // needle's char somewhere (du -hs, -hd1). Long flags (--) are
+                // handled by the exact/prefix branch above.
+                let single = needle
+                    .strip_prefix('-')
+                    .filter(|c| !needle.starts_with("--") && c.chars().count() == 1);
+                match single {
+                    Some(c) => {
+                        arg.starts_with('-')
+                            && !arg.starts_with("--")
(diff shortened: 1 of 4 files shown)


veltri-23:
Hi, author of PR #3538 (fix(du): keep summary and depth-limited output raw). Opened Aug 11, still no response. The latest push merged develop into the branch and is sitting at "workflow awaiting approval", so CI has not run on it. Could you approve the workflow run when you get a chance? For what its worth the earlier CI failures (checkout tests plus one compile error) do not reproduce locally on the same toolchain, so I suspect they were stale, but I want the fresh run to confirm.

Issue #3434, opened 2026-08-04
`rtk ls` silently omits build/dist/target/coverage/venv/node_modules — no count, no marker — and `rtk tree` feeds Unix flags to Windows' `tree.com`, reporting "No subfolders exist" as success


HetCreep:
Two findings from comparing rtk against the real tools.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Both reproduced on Windows with the shipped binary; outputs verbatim.*

### 1. `rtk ls` hides directories without saying so

A directory containing `build coverage dist node_modules normaldir target venv`:

```
$ ls
build coverage dist node_modules normaldir target venv

$ rtk ls
normaldir/
```

Six of seven directories are gone. No count, no `+N more`, nothing.

They do exist, and rtk knows it:

```
$ rtk ls -a
build/ coverage/ dist/ node_modules/ normaldir/ target/ venv/

$ rtk ls dist
out.js  2B
```

`src/cmds/system/ls.rs:266` filters against `NOISE_DIRS` (`src/core/constants.rs:1-27`).

Suppressing `node_modules` from a listing is a reasonable default. Doing it with no marker is the problem — a reader concludes the build has not run, the venv was never created, or `dist/` does not exist. That conclusion is wrong and nothing in the output hints at it.

Contrast rtk's own behaviour elsewhere: the result cap in `rtk find` prints `+450 more`, and `rtk read` prints `[N more lines]`. The convention exists; this path just does not use it.

A single line — `(6 build/tooling directories hidden; -a to show)` — would preserve the compression and remove the false impression.

### 2. `rtk tree` on Windows invokes `tree.com` with Unix flags

```
$ rtk tree
Too many parameters - node_modules|.git|target|__pycache__|.next|dist|build|.cache|…
exit=0
```

`src/cmds/system/tree.rs:15` checks `tool_exists("tree")`, which matches Windows' built-in `tree.com` — an unrelated program that takes no `-I` flag. `:25-32` then passes the ignore pattern, and `tree.com` rejects the whole invocation.

With `-a`, the same review observed `Invalid path - …\-A` followed by **`No subfolders exist`** in a directory that has eight — an affirmatively wrong statement, also at exit 0.

Both cases exit 0, so a caller sees a successful command that produced an error message as its "output".

`.claude/rules/cli-testing.md` lists Windows as a supported platform, and the release workflow ships a Windows binary.

[...]

PR #3540, opened 2026-08-11
fix(ls): report hidden directory entries

uuzzrm:
## Summary

- report how many noise directories are omitted by default in `rtk ls`
- keep `(empty)` for genuinely empty directories, and point users to `-a` when every entry is hidden
- leave `rtk ls -a` output unchanged

This addresses the `rtk ls` portion of #3434. The Windows `tree.com` behavior described in that issue is separate and is not changed here.

## Test plan

- `cargo fmt --all --check`
- `cargo test cmds::system::ls::tests`
- manual `rtk ls`, piped output, and `rtk ls -a` against a temporary tree containing build/dist/node_modules/target/venv

The full local `cargo test` run reaches 2 existing hook tests that depend on permission rules; this checkout's Claude settings include `Bash(*)`, so those tests resolve to `Allow` while their assertions expect `Ask`. The changed `ls` tests and all other tests passed.

### DIFF
--- a/src/cmds/system/ls.rs
+++ b/src/cmds/system/ls.rs
@@ -247,6 +247,7 @@ fn compact_ls(raw: &str, show_all: bool, show_long: bool) -> (String, String, us
     let mut lines_seen: usize = 0;
     let mut parsed_count: usize = 0;
     let mut dotdirs: usize = 0;
+    let mut hidden_count: usize = 0;
 
     for line in raw.lines() {
         if line.starts_with("total ") || line.is_empty() {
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for this. The reporting gap it targets was closed on develop by #3470 (merged 2026-09-01), which made compact_ls collect every entry it drops and emit a hint for them, and by #3278 (merged 2026-09-08), which turned that hint into a one-shot recall command returning the hidden entries themselves. Develop now prints "... (N filtered)" with that retrieval command, and the all-filtered directory keeps the hint instead of collapsing silently, so the count this PR adds is already there in a form that also hands back the names. One thing develop still does not do is point at -a as the simpler way to see those noise dirs; that would make a good small issue or a narrow PR against the current hidden_hint. If some part of this was missed, please comment or reopen.

PR #3542, opened 2026-08-11
fix(gh): deduplicate watched PR checks

### LINKED ISSUE #3541: gh pr checks --watch reports duplicated pass/fail counts when checks already finished
## Summary

`rtk gh pr checks <PR> --watch` reports check counts inflated by a multiple of the real total (e.g. 18 reported instead of 9) when the PR's checks have already finished by the time `--watch` starts polling.

## Root cause

`gh pr checks --watch`, when stdout is not a TTY (piped/captured — the normal case for an agent-driven shell), does not clear the screen between polls. It just appends the full checks table again on every polling cycle until it detects the run is complete. If the checks are already resolved, `gh` still does at least one extra poll cycle before exiting, so the raw output contains the same 9-row table printed 2+ times.

`rtk`'s filtered summary for `--watch` appears to count pass/fail occurrences across the **entire captured output stream** instead of deduplicating by check name / using only the final snapshot, so N poll cycles × real check count gets reported as the total.

## Reproduction

```sh
# Raw gh output — same table printed twice within one --watch invocation
gh pr checks 9776 --watch
# (9-row table, then the same 9-row table again ~10s later, 18 lines of "pass" total)

# rtk filtered mode, same PR
rtk gh pr checks 9776 --watch
# CI Checks Summary:
#   [ok] Passed: 18      <- real total is 9
#   [FAIL] Failed: 0

# Without --watch, rtk is correct on the same PR:
rtk gh pr checks 9776
# CI Checks Summary:
#   [ok] Passed: 9
#   [FAIL] Failed: 0
```

## Environment

- rtk 0.45.0 (also reproduced on 0.37.2 before upgrading — not fixed by the upgrade)
- gh version: 2.97.0 (2026-07-31)
- macOS, invoked via Claude Code's Bash tool (non-TTY stdout)

## Impact

An agent relying on `rtk gh pr checks --watch`'s summary to decide whether CI passed gets a plausible-looking but wrong number. In this case it wasn't misleading (18/18 still reads as "all passed, 0 failed"), but the count itself is fabricated and would misrepresent partial-failure states too if, say, 1 of 9 real checks failed but got counted twice as failed among 18 — the ratio would still look coherent but wouldn't match GitHub's actual check count.

## Suggested fix

For `--watch`, either:
- parse only the last complete table snapshot in the captured output before summarizing, or
- deduplicate by check name (keep the most recent status per name) before counting.

uuzzrm:
## Summary

`gh pr checks --watch` can append each polling snapshot when its output is captured instead of sent to a TTY. The formatter was counting every `pass` and `fail` row, so repeated snapshots inflated the totals.

This change parses the tab-separated check rows, keeps the latest status for each check identity, and summarizes the deduplicated results. Checks with the same name but different links remain separate, and the final failed rows are still shown.

Fixes #3541

## Testing

- `cargo fmt --all --check`
- `cargo test --bin rtk format_pr_checks` (3 passed)
- `cargo run --quiet -- gh pr checks 3540 --repo rtk-ai/rtk`

The full `cargo test` suite was also run. It produced 2,557 passing tests, with 16 existing process-test failures because their Unix command dependencies are unavailable in this Windows environment; none of those failures involve `gh_cmd.rs`.

### DIFF
--- a/src/cmds/git/gh_cmd.rs
+++ b/src/cmds/git/gh_cmd.rs
@@ -469,21 +469,43 @@ fn pr_checks(args: &[String], _verbose: u8, _ultra_compact: bool) -> Result<i32>
 }
 
 fn format_pr_checks(stdout: &str) -> String {
-    let mut passed = 0;
-    let mut failed = 0;
-    let mut pending = 0;
-    let mut failed_checks = Vec::new();
-
+    // `gh pr checks --watch` appends each poll's table when stdout is not a TTY.
+    // Keep the latest row for each check so status transitions are reflected
+    // without inflating the summary.
+    let mut checks: Vec<(String, String, PrCheckStatus, String)> = Vec::new();
     for line in stdout.lines() {
-        if line.contains("[ok]") || line.contains("pass") {
-            passed += 1;
-        } else if line.contains("[x]") || line.contains("fail") {
-            failed += 1;
-            failed_checks.push(line.trim().to_string());
-        } else if line.contains('*') || line.contains("pending") {
-            pending += 1;
+        let Some((name, link, status)) = parse_pr_check_line(line) else {
+            continue;
+        };
+
+        if let Some(check) = checks
+            .iter_mut()
+            .find(|check| check.0 == name && check.1 == link)
+        {
+            check.2 = status;
+            check.3 = line.trim().to_string();
+        } else {
+            checks.push((
+                name.to_string(),
+                link.to_string(),
+                status,
+                line.trim().to_string(),
+            ));
         }
     }
+
+    let passed = checks
+        .iter()
+        .filter(|check| check.2 == PrCheckStatus::Passed)
+        .count();
+    let failed = checks
+        .iter()
+        .filter(|check| check.2 == PrCheckStatus::Failed)
+        .count();
+    let pending = checks
+        .iter()
+        .filter(|check| check.2 == PrCheckStatus::Pending)
+        .count();
 
     let mut out = String::new();
     out.push_str("CI Checks Summary:\n");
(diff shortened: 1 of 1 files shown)


KuSh (approved):
## Round 1 of 3 — APPROVED, follow-up filed

**Claim**: dedupe `gh pr checks --watch` poll snapshots so each check is counted once, at its latest status.

**Scope**: accept _(round 1, frozen)_. Two other open PRs touch `gh_cmd.rs` nearby — #1122 (the `RunOptions` line just above) and #4000 (the identifier parse inside `pr_checks`). Neither touches `format_pr_checks`, and both merge cleanly with this; #3542 can land independently.

**Ran**: real `gh` 2.46.0 locally and 2.100.0 in Docker, `LC_ALL=C` — green / red / live-pending / missing-PR / `--required` / `--json` / `--watch`, develop `5626e94a` vs this PR merged onto it.

The decisive case is a **live 25-poll `--watch`**, captured against #3857 while its CI was actually running, so the pending→pass transitions are real rather than synthetic (347 lines, 30,794 B):

```
develop:  [ok] Passed: 254   [pending] Pending: 43     <- 11 checks
this PR:  [ok] Passed: 11    [FAIL] Failed: 0          <- correct
```

That is 99.8% against the 20% floor (CONTRIBUTING.md). Non-watch output is byte-identical to develop, exit codes 0/1/8 all propagate, and the error paths are unchanged.

**CI** @`83f87f92`: green.

Worth recording: the two earlier CI runs on `656a8640` showed as `failure`, but both had **zero jobs and no logs** — they were `action_required` runs that expired without ever executing. This PR had never been built. The fixup push created a fresh run, which I approved; that is the first real CI this branch has had.

### Blocks merge (0, frozen at round 1)

Nothing. No content loss, no wrong answer, no regression against develop.

### Optional — closed by fixup `83f87f92`

The transition test used `"Lint\t*\t\t…"`. Real `gh` spells a pending bucket `pending`, never `*`, and emits **five** tab-separated fields with the description last — I checked 92 real rows across 9 PRs plus the 347-line watch capture: zero `*`, zero empty link fields. The `"*"` arm itself is harmless dead code and I left it alone; the fixup only corrects the fixture and adds a case with the real five-field shape. Mutation-checked: disabling the dedup fails 3 of the 4 tests, the new one included.

### Filed as follow-up

- #4021 — `format_pr_checks` drops every bucket that is not `pass`/`fail`/`pending`, so `skipping` and `cancelled` rows vanish and the totals do not add up (real #3982: 8+2 reported for 15 checks). **Pre-existing** — the old substring implementation dropped the same rows, verified on #1122 — so it is not a regression here and does not hold this up. This PR is the natural base for the fix.

- Output format is byte-identical between `gh` 2.46.0 and 2.100.0, so the new exact-match parser is not brittle across versions.
- `--json` is passed through untouched (`has_json_flag` routes it to passthrough at `gh_cmd.rs:194`) — confirmed by replaying real `gh` 2.100 JSON through both binaries.
- Each of the three added tests bites under targeted mutation.

- **Format drift on newer `gh`** — the exact-match parser looked brittle next to the old substring matching, but 2.100 emits the identical format.
- **Empty `link` field** splitting one check into two entries across polls — 0 of 92 real rows have one.
- **Red PRs showing the fix** — they cannot today: `early_exit_on_failure` sends a non-zero exit down the raw path, so the benefit on failing runs only appears once #1122 lands. Both are needed for the red case; neither blocks the other.

### Next

Nothing — approving. Fixup `83f87f92` pushed: corrects the test fixtures to `gh`'s real row shape and adds a five-field case.

PR #3552, opened 2026-08-12
feat(hooks): add direct Codex command rewriting

### LINKED ISSUE #1003: Add hook to Codex

### LINKED ISSUE #1812: Codex now support PreToolUse hook

yeahjack:
## Summary

- Add native Codex `PreToolUse` integration that transparently rewrites supported shell commands to `rtk ...` through `updatedInput`.
- Codex now supports input mutation and applies the rewritten command before its normal approval and sandbox flow. This removes the historical blocker without a deny-and-retry model round trip.
- Add project/global installation, status, backup, and uninstall support while preserving existing user hooks.

## Implementation

- Add `rtk hook codex` with fail-open parsing and preservation of other tool input fields.
- Wire the hook into `rtk init --codex` and update Codex guidance and integration docs.
- Cover command rewriting, unsupported inputs, JSON merging, installation, and removal.

## Benefits

Codex users get transparent RTK token savings without relying on prompt compliance or paying for another model inference, while retaining Codex's native approval and sandbox checks.

This integration is now mature because the required `PreToolUse.updatedInput` capability is available in Codex and has been verified end to end against Codex CLI `0.147.0-alpha.6.6`.

I would appreciate review and merge if this direction aligns with the project.

## Test plan

- [x] Full suite: 2,673 tests passed, 0 failed
- [x] Wire-level tests for rewrite and fail-open behavior
- [x] Local install, idempotency, status, and uninstall tested
- [x] Real Codex end-to-end test confirmed that raw `ls -la` executed as `rtk ls -la` and returned compact output

## Related issues

Closes #1003
Closes #1812
Addresses #2921


yeahjack:
Hi @aeppling @KuSh @TaKO8Ki, Codex has been gaining a lot of traction lately, and I think this integration could be really useful for its users. It lets RTK work transparently in Codex, so command output takes up less context without adding an extra model round trip.

Whenever you have some bandwidth, I would really appreciate a review. I am happy to respond quickly and keep improving anything that needs work. Thanks!

SplittyDev:
AFAIK this is already open in https://github.com/rtk-ai/rtk/pull/2557, sadly not yet merged.

yeahjack:
> AFAIK this is already open in #2557, sadly not yet merged.

Thanks! I checked #2557. There is an important behavioral difference between the two implementations.

#2557 intentionally skips rewriting in default and ask modes, and only applies `updatedInput` under `bypassPermissions`, because it treats `permissionDecision: "allow"` as potentially bypassing Codex's normal approval flow.

Our PR targets the current Codex hook semantics. The official Codex hooks documentation now explicitly specifies `permissionDecision: "allow"` together with `updatedInput` as the way to rewrite a `PreToolUse` call without blocking it. Approval is handled separately through the `PermissionRequest` lifecycle, which falls back to the normal approval prompt when no hook makes a decision.

This means RTK can transparently rewrite the command in normal modes while still preserving Codex's native approval and sandbox flow. I also verified this end to end with Codex CLI `0.147.0-alpha.6.6`, where the rewritten command is what ultimately executes.

So while the two PRs overlap in purpose, the behavior is meaningfully different. Happy to consolidate with #2557 if the maintainers prefer.

src/hooks/hook_cmd.rs
```diff
@@ -670,6 +670,115 @@ fn run_claude_inner(input: &str) -> Option<String> {
...
+        obj.insert("command".into(), Value::String(rewritten.clone()));
+    }
+
+    PayloadAction::Rewrite {
+        cmd: cmd.to_string(),
+        rewritten,
+        output: json!({
+            "hookSpecificOutput": {
+                "hookEventName": PRE_TOOL_USE_KEY,
+                "permissionDecision": "allow",
```
KuSh: This should be gated by decide_hook_action/PermissionVerdict

If you want rtk to allow commands on its own and Codex permits it, you should handle Codex in permissions.rs check_command_for

src/hooks/init.rs
KuSh: now that you write project-scoped codex configs you should not bail when uninstalling them

KuSh:
Hi @yeahjack, I’ll review your PR once #2557 is merged, since we should build on it.

yeahjack:
Thanks! One thought after looking more closely at #2557 and the review discussion: Our PR may be better considered as an alternative implementation rather than necessarily a follow-up.

It keeps RTK responsible only for `PreToolUse` rewriting via `allow + updatedInput`, while Codex remains responsible for approval and sandboxing. This avoids the permission-layer complexity currently being discussed in #2557 and matches the current Codex hook semantics.

I’m happy to follow whichever direction you prefer, whether that means reviewing Our PR directly or rebasing it on top of #2557 and incorporating the relevant parts.

KuSh:
Replying to https://github.com/rtk-ai/rtk/pull/3552#issuecomment-5301322704 —

On "alternative implementation vs follow-up": I think that framing is fair on the protocol point specifically. I traced Codex's own source (`codex-rs/core/src/tools/registry.rs` → `handlers/shell.rs`) and confirmed `updatedInput` is applied *before* Codex's native approval/sandbox check runs, and that check evaluates the *rewritten* command — so `permissionDecision: "allow"` here really is a protocol requirement to get the substitution accepted, not RTK vouching for the command's safety. Codex remaining authoritative over approval is a defensible position, not a shortcut around #2557's discussion.

That said, digging further turned up a real gap in that argument, plus some independent bugs — I'd want these addressed regardless of which PR becomes the reference implementation:

**1. `permission_mode` is received but never read.** Codex computes and threads this field deliberately through every hook event (`hook_permission_mode()` derives it from `approval_policy()`), so it reads as a signal hook authors are meant to use, not an inert echo. `process_codex_payload` rewrites identically in `default`/`ask`/`bypassPermissions` — unlike #2557, which only rewrites when `bypassPermissions` is active. Even if the "Codex re-evaluates post-rewrite" argument holds today, discarding this field is a forward-compatibility risk that's cheap to hedge against now.

**2. Wrapping through `rtk` is opaque to Codex's own safety heuristics — in both directions.** Codex's `is_known_safe_command`/`dangerous_command_match_for_exec` (in `codex-rs/shell-command/src/command_safety/`) key off the literal first token and only unwrap `sudo`/`env`/`trap` as known wrappers — not `rtk`. So rewriting defeats Codex's safe-command allowlist (more prompts than before, under `UnlessTrusted`/restricted-sandbox policies) *and* its dangerous-command denylist (potentially fewer, if RTK ever rewrites something Codex would flag). Today's blast radius is narrow only because RTK doesn't currently wrap `sudo`/`env`/`trap`/`rm --force` — RTK's git rewrite already covers `commit`/`push`/`pull`, not just reads, so this isn't hypothetical scope, just a currently-narrow one. Worth a line in the code comment near the `allow` output making this explicit, since "Codex still checks" undersells the caveat.

**3. Local install has no working uninstall.** `rtk init --codex` (no `-g`) writes a live entry into `.codex/hooks.json`, but `uninstall_codex(global, ctx)` unconditionally bails for `!global` with a message that doesn't even mention `hooks.json` — so a local install has no CLI path to remove the hook it created.

**4. Two hook-detection parser regressions in `src/hooks/mod.rs`.** The new hand-rolled binary-basename parser mishandles backslash-escaped-space paths (`/Users/jane/My\ Apps/rtk hook claude` misdetects), and the quoted-binary extractor doesn't check the closing quote is followed by whitespace/EOF (`"rtk"evil hook codex` gets misidentified as bare `rtk`). Both cause false negatives/positives in install idempotency and integrity checks.

**On duplication** (`run_codex`/`run_claude`, and the `hooks.json` patch/remove functions for Codex/Claude/Cursor/Droid): I don't think collapsing these into one shared function is the right fix — each agent's stdin quirks (e.g. Codex needing `strip_leading_bom`, which Claude apparently doesn't) and response shape will keep diverging as more hosts are added, and a shared dispatcher just reintroduces host branching inside itself. What I'd suggest instead: keep the per-agent orchestration functions separate, but extract the mechanical scaffolding that's identical across them — stdin read/parse/fail-open in `hook_cmd.rs`, and the `hooks.PreToolUse[].hooks[].command` find/insert/remove/backup/atomic-write walk in `init.rs` — into shared low-level helpers that each agent's function calls into. That keeps agent-specific behavior explicit while stopping the drift already visible between the Codex and Claude copies.

Happy to help verify any of the above against a specific approach if useful.

yeahjack:
Thanks for the detailed review. I rebased onto the latest `develop` and pushed `e85985e` with fixes for the four concrete issues you found:

1. `permission_mode` is now validated against the documented Codex modes. Missing or unknown modes fail open and keep the original command unchanged.

2. The code and docs now explicitly explain that Codex classifies the rewritten argv and does not currently unwrap `rtk`. This limitation remains, but unsupported destructive commands now have regression coverage to ensure they stay on the passthrough path.

3. Project-scoped uninstall now works. `rtk init --codex --uninstall` removes the RTK hook, `RTK.md`, and the `AGENTS.md` reference while preserving unrelated user content and hooks.

4. Hook command detection now handles backslash-escaped spaces correctly and rejects cases such as `"rtk"evil hook codex`.

I added regression tests for each case. `cargo fmt --all --check`, `cargo clippy --all-targets`, and the full test suite pass. That includes 2,628 unit tests, with 8 ignored as expected, plus all 78 integration tests. I also verified project-scoped install and uninstall end to end using the built CLI.

I kept the agent-specific orchestration separate as suggested. I did not include the broader low-level scaffolding extraction in this update, but I am happy to follow up with that if you would prefer it in this PR.

Would appreciate another look when you have time. Happy to keep iterating.

KuSh:
Rechecked against the updated branch — the four issues from the previous round are genuinely fixed, verified against source rather than just the diff:

- **`permission_mode`**: not a bug after all. I pulled Codex's actual `hook_permission_mode()` (`codex-rs/core/src/hook_runtime.rs`) — it only ever emits `"bypassPermissions"` or `"default"` for `PreToolUse`, both of which are now handled. The extra `acceptEdits`/`plan`/`dontAsk` schema values are unreachable dead code (carried over from Claude Code's mode set), but harmless.
- **Local uninstall**: works and is tested (`test_local_codex_install_can_be_uninstalled`), and does back up before writing (`.bak` via `fs::copy`).
- **Parser regressions** (backslash-escaped-space paths, `"rtk"evil hook codex`): both fixed and tested; the fix is a strict superset of the prior matching behavior, no regression.
- **Classifier-blindness caveat**: addressed exactly as asked — a documentation note, not a mechanism change.

Three minor things left, none of them merge blockers:

1. `src/hooks/init.rs:4308` and `:4340` — the new `fs::read_to_string()` calls in `show_codex_config()` skip `.context()`, unlike this same PR's other new helpers (`patch_codex_hooks_json`, `remove_codex_hook_from_file`), and unlike the repo's "always context, always anyhow" rule. A permission error or a delete-between-check-and-read race would surface as a bare OS error with no indication of which file failed.
2. `src/hooks/init.rs:494` — the Codex `hooks.json` patch/remove/idempotency-check pipeline still duplicates the Cursor/Droid versions instead of sharing low-level helpers, which was asked for last round. Worth calling out concretely: the copies have already drifted — Droid's swallows backup errors with `.ok()` while Codex's propagates with `?`. Not asking to merge the per-agent orchestration (that split is correct), just the mechanical read/backup/atomic-write scaffolding.
3. `src/hooks/hook_cmd.rs:1668` — the `codex_input()` test helper now hardcodes `permission_mode: "default"` everywhere, so only one dedicated test still exercises the missing-`permission_mode` fail-open path. Worth keeping that test around deliberately since nothing else in the suite covers it anymore.

Nice work turning the previous round around — this is in good shape.

yeahjack:
Thanks for checking this again. I rebased onto the latest `develop` and pushed `cff625a` with the remaining cleanup:

1. Added path-aware error context to all four Codex configuration reads in `show_codex_config()`.

2. Extracted shared JSON read and backup plus atomic-write helpers for the Codex, Cursor, and Droid hook paths. Their agent-specific orchestration and JSON structures remain separate. Backup failures now propagate consistently instead of being silently ignored.

3. Kept the missing `permission_mode` fail-open test and added a comment making its purpose explicit.

4. Added the Semgrep annotation for the intentional project-local `RTK.md` removal.

I also added regression coverage for missing and empty JSON files, path-aware parse and read errors, successful backups, backup failures, and preservation of the original file when a backup fails.

Local verification is green: formatting, Clippy, 2,634 unit tests with 8 expected ignores, all 78 integration tests, the real Codex install and uninstall flow, and a workflow-equivalent Semgrep scan with zero findings.

The new CI run is currently waiting for maintainer approval before GitHub will start any jobs. One unrelated note: the benchmark job on the current `develop` branch is already failing on the existing `find --max 10` token regression, so I left that code untouched.

Happy to address anything else that comes up after the CI run is approved.

KuSh:
Re-verified `cff625a` — all three items from the last round are genuinely done:

- `show_codex_config()`'s reads now carry `.with_context(...)`.
- `read_json_file`/`backup_and_atomic_write` are extracted and shared across the Codex/Cursor/Droid patchers (12+ call sites), while orchestration stays split per agent as intended.
- The missing-`permission_mode` fail-open test is still there, now with an explicit comment.

One more pass at high effort surfaced one actionable item, posted inline below.

Two other things it turned up are **not** asks for this PR:

- The lack of a `Host::Codex` permission-engine check (unconditional `allow` on rewrite) is the wrapper-opacity trade-off already discussed and now documented in the code comment above `updated_input` — no action needed, just flagging that it's still the same known trade-off, not a new one.
- `raw_first_token`'s escaped-quote handling and the underlying `shell_split` Windows-backslash-stripping it works around are both root-caused in the shared lexer, which is being consolidated in #3704. That's out of scope here — tracking it there rather than asking for changes in this PR. If #3704 lands first, we'll ask for a follow-up to drop the `raw_first_token` workaround; if this PR lands first, #3704 will account for it.

src/hooks/hook_cmd.rs
```diff
@@ -670,6 +670,135 @@ fn run_claude_inner(input: &str) -> Option<String> {
...
+
+    // Codex requires `permissionDecision: allow` alongside `updatedInput`.
+    // Its runtime applies the replacement before native approval and sandbox
+    // checks, so this is a protocol-level allow rather than RTK approving the
+    // command. Those checks inspect the rewritten argv, however, and Codex's
+    // safe/dangerous-command classifiers do not currently unwrap `rtk`. This
+    // can add prompts for known-safe commands and obscure classifier signals
+    // for wrapped mutating commands such as git push. Keep the passthrough gates
+    // above conservative and revisit this boundary whenever coverage expands.
+    let mut updated_input = v.get("tool_input").cloned().unwrap_or_else(|| json!({}));
```
KuSh: `process_codex_payload` duplicates `process_claude_payload`'s `tool_input`-merge + `hookSpecificOutput` envelope construction (lines 601-607 above) by hand instead of sharing a helper. They've already diverged once (Claude conditionally omits `permissionDecision` on Ask; Codex always sets it to `allow`) — a future fix or field addition to one envelope-building path is easy to forget in the other since nothing ties them together. Could you extract a small shared helper for the `tool_input`-clone-and-merge step (and ideally the envelope shape), similar to what was just done for the hooks.json read/backup/write scaffolding?

KuSh (changes requested):
Requesting one small change (inline below) — the rest of the previous review round is resolved, see the follow-up comment.

src/hooks/mod.rs
```diff
@@ -12,15 +12,50 @@ pub mod rewrite_cmd;
...
+}
+
+fn raw_first_token(command: &str) -> Option<&str> {
+    let command = command.trim_start();
+    let quote = match command.as_bytes().first() {
+        Some(b'"') => Some('"'),
+        Some(b'\'') => Some('\''),
+        Some(_) => return command.split_whitespace().next(),
+        None => return None,
+    }?;
```
KuSh: Two notes on `raw_first_token`, both "can't happen" rather than breakage — worth pinning down so a future reader doesn't have to re-derive them.

**The single-quote branch (line 24) can never change the outcome.** `shell_split` doesn't process escapes inside `'...'`, so a single-quoted first token already survives intact through the primary path:

```
shell_split(r"'C:\Users\jane\rtk.exe' hook claude")
  → ["C:\\Users\\jane\\rtk.exe", "hook", "claude"]
```

If that token is an rtk path, `is_rtk_binary(parsed_binary)` matched and the `||` short-circuited before the fallback ran; if it isn't, the fallback returns the same string and still doesn't match. The backslash-eating this fallback exists to work around is confined to unquoted text and `"..."`, so only the `"` branch and the unquoted `split_whitespace()` branch are load-bearing.

**The `?` on line 27 is a no-op.** Every arm either yields `Some(_)` or returns early, so it can never short-circuit. Dropping the `Option` round-trip says the same thing:

```rust
let quote = match command.as_bytes().first() {
    Some(b'"') => '"',
    Some(b'\'') => '\'',
    Some(_) => return command.split_whitespace().next(),
    None => return None,
};
```

Neither needs fixing for this PR to be correct. One thing explicitly *not* worth adding here, since it looks like a gap: `quoted.find(quote)` is blind to a backslash-escaped quote, but there's no input that reaches it with one — Windows forbids `"` in a path outright, and any POSIX string with a legitimately escaped quote is parsed correctly by `shell_split`, so the fallback is never consulted for it.

src/hooks/hook_cmd.rs
KuSh: Fixed in `e4b74fd3` — `pre_tool_use_rewrite_output()` is now shared by Claude, Codex, and Droid, with its own test. Resolving.

src/hooks/init.rs
```diff
@@ -3811,33 +3953,23 @@ fn remove_cursor_hooks(ctx: InitContext) -> Result<Vec<String>> {
...
-                            hooks_json_path.display()
-                        );
-                    } else {
-                        let backup_path = hooks_json_path.with_extension("json.bak");
-                        fs::copy(&hooks_json_path, &backup_path).ok();
-
-                        let serialized = serde_json::to_string_pretty(&root)
-                            .context("Failed to serialize hooks.json")?;
-                        atomic_write(&hooks_json_path, &serialized)?;
+    if let Some(mut root) = read_json_file(&hooks_json_path)? {
```
KuSh: This is a behavioral regression from the `read_json_file`/`backup_and_atomic_write` refactor in `53fc1b61`. Before that commit, a `hooks.json` parse error was silently swallowed here (`if let Ok(mut root) = serde_json::from_str(...) { ... }`), so `remove_cursor_hooks` always reported success. Now `read_json_file(&hooks_json_path)?` propagates the parse error as a hard `Err`. Step 1 above (deleting the Cursor hook script) has already run and committed to disk by the time this executes, so a corrupted `hooks.json` now leaves partial, inconsistent state (script gone, stale RTK entry still in `hooks.json`) and reports failure where it used to succeed cleanly. Could this step catch/log a parse error the way the old code did, rather than propagating it, so uninstall stays best-effort here?

src/hooks/init.rs
```diff
@@ -3276,12 +3448,9 @@ fn remove_droid_hook_from_file(file: &DroidHookFile, ctx: InitContext) -> Result
             path.display()
         );
     } else {
-        let backup_path = path.with_extension("json.bak");
-        fs::copy(path, &backup_path).ok();
-
         let serialized =
             serde_json::to_string_pretty(&root).context("Failed to serialize Droid hook file")?;
-        atomic_write(path, &serialized)?;
+        backup_and_atomic_write(path, &serialized)?;
```
KuSh, line 3774: This `?` means a backup failure on this file aborts `uninstall_droid_at`'s whole loop over `droid_hook_file_candidates` before the remaining candidate files are ever touched — even though `test_remove_droid_hook_propagates_backup_failure_and_preserves_file` covers the single-file fail-closed behavior, this loop-level consequence (one file's backup problem, e.g. a stray `<file>.json.bak` directory, blocking cleanup of unrelated files) isn't tested or documented. Worth collecting errors per-candidate in `uninstall_droid_at` and continuing the loop, surfacing a combined error/warning at the end, so an unrelated file's backup issue doesn't block cleanup of the others?

src/hooks/init.rs
```diff
@@ -2434,11 +2512,126 @@ fn run_codex_mode_with_paths(
...
         }
+        println!(
+            "\n  Restart Codex. For a project hook, approve it when Codex asks you to trust it."
+        );
     }
 
     Ok(())
 }
 
+fn codex_hook_already_present(root: &serde_json::Value) -> bool {
```
KuSh, line 2817: `codex_hook_already_present`/`patch_codex_hooks_json`/`remove_codex_hook_from_json` (this block, ~130 lines) reimplement the same read/check/patch/remove-hooks.json pattern already present for Claude (`hook_already_present`/`insert_hook_entry` near line 1147) and Cursor. This PR did unify the low-level file I/O via `read_json_file`/`backup_and_atomic_write`, but the higher-level presence-check/insert/remove logic is now a 3rd near-identical implementation — a future fix to that logic has to be found and applied in 3+ places. Not blocking, just flagging as a follow-up worth extracting at some point (could be a separate PR).

src/hooks/init.rs
```diff
@@ -4124,8 +4258,31 @@ fn show_codex_config() -> Result<()> {
         println!("[--] Global RTK.md: not found");
     }
 
+    if global_hooks_json.exists() {
```
KuSh, line 5378: This global-hooks.json status block and its local counterpart further down are copy-pasted 16-line blocks differing only in the path variable and the "Global"/"Local" label. A `print_codex_hook_status(label, path)` helper would collapse both and avoid the wording/error-handling drifting apart if one copy gets updated and the other doesn't. Not blocking.

src/hooks/hook_cmd.rs
```diff
@@ -571,6 +571,28 @@ enum PayloadAction {
     Ignore,
 }
 
+fn pre_tool_use_rewrite_output(
```
KuSh, line 600: `vscode_response_from_decision` (further down, around the Copilot/VS Code handlers) still hand-rolls the `hookSpecificOutput`/`updatedInput` JSON that this new helper exists to consolidate. Since `handle_vscode`/`vscode_response` only thread a bare `cmd: &str` rather than the full payload, that path also can't preserve extra `tool_input` fields (`timeout`, `description`) the way this helper does for Codex/Claude/Droid — worth migrating it too, either in this PR or a quick follow-up, so the consolidation is complete.

src/hooks/init.rs
```diff
@@ -433,6 +434,45 @@ fn atomic_write(path: &Path, content: &str) -> Result<()> {
...
+        return Ok(Some(serde_json::json!({})));
+    }
+
+    serde_json::from_str(&content)
+        .map(Some)
+        .with_context(|| format!("Failed to parse {} as JSON", path.display()))
+}
+
+/// Back up an existing JSON file before replacing it atomically.
+fn backup_and_atomic_write(path: &Path, content: &str) -> Result<Option<PathBuf>> {
```
KuSh, line 644: Claude's own settings.json installer, `patch_settings_json_command` (further down, ~line 1014), is the most-used path in this file but wasn't migrated to `read_json_file`/`backup_and_atomic_write` — Droid and Cursor were. That leaves two divergent failure-handling styles (that function's older ad hoc read/backup vs. the new hard-fail-on-backup-error everywhere else) between Claude's install path and everyone else's, undermining the consistency this refactor otherwise establishes. Worth a follow-up to bring it in line too.

KuSh (changes requested):
Two real regressions from the `read_json_file`/`backup_and_atomic_write` refactor in `53fc1b61` (inline below), plus a handful of non-blocking cleanup notes on where the duplication this PR set out to reduce is still present.

yeahjack:
Thanks for the detailed follow-up. I pushed `e260715` with fixes for the two regressions:

1. Cursor uninstall now catches only malformed `hooks.json` parse errors, logs a warning, leaves the malformed file unchanged, and continues the best-effort cleanup. Read and permission errors still propagate normally.

2. Droid uninstall now attempts every candidate hook file, collects per-file errors, and returns a combined error after the loop. A backup failure in one candidate no longer prevents later unrelated candidates from being cleaned.

I added regression coverage for the malformed Cursor JSON path, Cursor I/O error propagation, and the Droid multi-candidate failure case.

Local verification is green: formatting, Clippy, 2,675 unit tests with 8 expected ignores, all 79 integration tests, a workflow-equivalent Semgrep scan with zero findings, and a real project-scoped Droid install and uninstall round trip.

I kept the non-blocking cleanup suggestions out of this change so this round stays focused on the two behavioral regressions.

The new CI run is waiting for maintainer approval before GitHub will start the jobs.

src/hooks/mod.rs
yeahjack: Thanks, that makes sense. I will leave `raw_first_token` unchanged in this PR. The single-quote branch and the `?` are non-behavioral, and #3704 is the right place to simplify this once the shared lexer consolidation lands.

src/hooks/init.rs
yeahjack: Fixed in `e260715`. `remove_cursor_hooks_at()` now catches only `serde_json::Error`, logs a warning, leaves the malformed `hooks.json` unchanged, and continues the best-effort cleanup. Read and permission errors still propagate normally.

I added `test_remove_cursor_hooks_keeps_malformed_json_best_effort` and `test_remove_cursor_hooks_still_propagates_read_errors` to cover both boundaries.

src/hooks/init.rs
yeahjack: Fixed in `e260715`. `uninstall_droid_at()` now attempts every candidate file, collects per-file failures, and returns a combined error after the loop.

The new regression test verifies that a backup failure leaves the affected file unchanged without preventing a later candidate from being cleaned and backed up.

src/hooks/init.rs
yeahjack: Agreed. This is worth a separate follow-up. I kept this PR scoped to the Codex integration and the concrete behavioral regressions, so I did not expand this round into a higher-level hooks.json traversal refactor.

src/hooks/init.rs
yeahjack: Agreed. A `print_codex_hook_status(label, path)` helper would be a straightforward cleanup. I kept it as a follow-up since it is non-blocking and does not affect the current integration behavior.

src/hooks/hook_cmd.rs
yeahjack: Agreed. A proper migration needs to thread the full payload through `handle_vscode` and `vscode_response`, rather than only replacing the envelope builder, so that extra `tool_input` fields are preserved.

I kept this out of the current PR and would cover it with dedicated field-preservation tests in a follow-up.

src/hooks/init.rs
yeahjack: Agreed. The Claude installer should use the shared JSON I/O helpers as well. Since it is the most-used path, I would handle that in a focused follow-up with coverage for existing settings, backup failures, dry-run behavior, and idempotency. I left its current behavior unchanged in this PR.

KuSh:
> I pushed `e260715` with fixes for the two regressions

Thanks
> I kept the non-blocking cleanup suggestions out of this change so this round stays focused on the two behavioral regressions.

Would you mind tackling these points in a follow-up PR?

patrickstratznig:
Command rewriting works, but RTK history remains broken in Codex workspace sandboxes: SQLite cannot open its database directory (`SQLITE_CANTOPEN`, error 14), so `rtk gain --history` does not record commands.

Safe follow-up: during `rtk init --codex`, resolve RTK’s actual tracking-data directory and print an opt-in minimal Codex `workspace-write` configuration that adds only that directory to `writable_roots`.

yeahjack:
@KuSh Yes, happy to take those cleanup items in a separate follow-up PR. Thanks for consolidating the related Codex PRs here. I’ve also prepared a local rebase onto the latest develop and removed the lexer workaround now that #3704 has landed.

yeahjack:
@patrickstratznig Thanks for reporting this. I pushed `293a50f` with the opt-in setup guidance you suggested.

`rtk init --codex` now prints a workspace-write configuration snippet using the actual tracking database directory, respecting `RTK_DB_PATH`, `tracking.database_path`, and the platform default. It adds only the database parent directory, which SQLite also needs for journal/WAL files. Users can review and merge the snippet into their existing Codex configuration; RTK does not change sandbox permissions automatically.

I verified custom paths, TOML escaping, and the real init flow, including that no `config.toml` is created. Formatting, Clippy, 3,435 unit tests, all 110 integration tests, and the baseline Semgrep scan pass. I have not reproduced your exact sandbox environment, so I would appreciate confirmation that this resolves history recording in your setup.

src/hooks/hook_cmd.rs
```diff
@@ -783,6 +784,125 @@ fn run_claude_inner(input: &str) -> Option<String> {
...
+    }
+
+    if crate::discover::lexer::contains_unattestable_construct(cmd) {
+        return PayloadAction::Skip {
+            decision: HookOutcome::Defer,
+            cmd: cmd.to_string(),
+        };
+    }
+
+    let rewritten = match get_rewritten(cmd) {
```
KuSh: **This is the one thing blocking merge, and it isn't your doing.**

`get_rewritten` no longer exists on `develop`, so the PR no longer compiles against it. Merging `293a50f9` into `develop` (`79b96a44`) in a throwaway worktree gives 5 errors, all from this call:

```
src/hooks/hook_cmd.rs:803:27: error[E0425]: cannot find function `get_rewritten` in this scope
src/hooks/hook_cmd.rs:803:9:  error[E0277]: the size for values of type `str` cannot be known at compilation time
src/hooks/hook_cmd.rs:804:14: error[E0277]: ...
src/hooks/hook_cmd.rs:804:9:  error[E0277]: ...
src/hooks/hook_cmd.rs:825:9:  error[E0308]: mismatched types: expected `String`, found `str`
```

(line numbers from the merged tree). `develop`'s `d6ce8f71 refactor(hooks): extract the shared hook decision` removed `get_rewritten`; the rewrite now only comes out of `decide_hook_action(cmd, Host) -> HookDecision`. That commit is **not** an ancestor of your merge-base, so it landed after you branched — this is `develop` moving under you. GitHub still reports `MERGEABLE` because there's no textual conflict; it simply doesn't build.

It isn't a mechanical rename, which is why I'm asking instead of pushing a fixup. `get_rewritten` was precisely the primitive that let the Codex path skip the permission engine — the documented trade-off in the comment just below. There is still no `Host::Codex`. So:

- **(a)** Add `Host::Codex` to `permissions::Host` and route Codex through `decide_hook_action`, like every other handler. This closes the deny/ask gap we've discussed across rounds — your own comment says to revisit this boundary whenever coverage expands, and `develop` deleting the bypass is arguably that moment. The wrinkle is protocol: Codex needs `permissionDecision: "allow"` alongside `updatedInput`, so you'd have to decide what an `AskRewrite` verdict emits.
- **(b)** Reintroduce a bypass helper to preserve today's behaviour exactly, accepting that it re-adds a primitive the refactor just removed.

I lean (a), but it's your design call and I've built neither, so I'm not going to pretend it's a one-liner. Either way this needs a rebase — the branch is 126 commits behind.

KuSh (changes requested):
## Round 5 (landing) — CHANGES REQUESTED

**Claim**: Add a native `rtk hook codex` PreToolUse processor that rewrites Bash commands to `rtk …` for Codex CLI, plus `rtk init --codex` install/uninstall/status tooling.
**Scope**: accept (frozen).
**Ran**: the 3 new regression tests, plus 3 mutations — each fix reverted individually to confirm the matching test actually fails; the hook-detection suite including the Windows-path case; a merge into latest `develop` (`79b96a44`) in a throwaway worktree; `cargo fmt --all --check` clean on the merged tree.
**CI** @`293a50f9`: **never run** — only `license/cla` is present. No workflow has ever built this PR.

### Blocks merge (1 — new, caused by `develop` moving)

The PR no longer compiles against current `develop` (`get_rewritten` was removed upstream). Details and the two ways out are in the line comment on `src/hooks/hook_cmd.rs`.

### Fold into that same push (not blocking on its own)

`src/hooks/hook_cmd.rs:889` — the upstream adaptation collapsed the Codex skip reasons:

```rust
-        PayloadAction::Skip { reason, cmd } => audit_log(reason, &cmd, ""),
+        PayloadAction::Skip { cmd, .. } => audit_log("skip:defer", &cmd, ""),
```

At `e4b74fd3` this recorded three distinct reasons (`skip:unsupported_permission_mode`, `skip:defer`, `skip:no_rewrite`); now all three log `skip:defer`. `rtk hook audit` groups its skip breakdown by that suffix (`hook_audit_cmd.rs:144-152`), and `process_claude_payload`'s arm keeps the distinction with a comment explaining exactly why. The label is never *wrong* today — all three Codex sites are genuinely `Defer` — but it loses the one signal that would diagnose the failure mode we discussed earlier: if Codex ever changes its `permission_mode` vocabulary, every command skips silently and the audit log would now just say "defer". Since `decision` alone can't separate the three, the reason needs to travel on `Skip`.

### Verified fixed this round — no need to revisit

- **Cursor malformed `hooks.json`** (`remove_cursor_hooks_at`): parse errors warn and continue best-effort; read/permission errors still propagate. Confirmed by mutation from both sides — narrowing the catch fails `test_remove_cursor_hooks_keeps_malformed_json_best_effort`, widening it to swallow everything fails `test_remove_cursor_hooks_still_propagates_read_errors`. Both arms are genuinely pinned.
- **Droid multi-candidate cleanup** (`uninstall_droid_at`): collects per-file errors and continues. Confirmed by mutation — restoring `?` fails `test_uninstall_droid_continues_after_candidate_failure`.
- **`raw_first_token`**: gone entirely now that #3704 landed; `is_rtk_hook_command` goes through the shared `shell_split`, and `codex_hook_command_matches_bare_absolute_and_windows_rtk` still passes. Exactly the outcome we planned for.
- **New sandbox guidance** (`293a50f9`): print-only and opt-in, with coverage for path-limiting and TOML escaping. Scope growth this late is normally something I'd push back on, but it's additive, it answers a real user report, and splitting it would cost more than it saves — keeping it.

### Filed as follow-ups

#4002 tracks the four cleanup threads you offered to take separately. I searched open and closed issues on both the symptom and the mechanism first — nothing existing covered them. Resolving those threads here; the detail lives in the issue. One extra detail landed in it that wasn't its own thread: `show_codex_config` still parses with raw `read_to_string` rather than `read_json_file`, so an empty `hooks.json` reports as invalid JSON where every other Codex path treats it as benign.

### Your questions

- *"happy to take those cleanup items in a separate follow-up PR"* → yes please, that's the right split — tracked in #4002.
- *"I would appreciate confirmation that this resolves history recording in your setup"* → that one's for @patrickstratznig; not gating this PR on it.

### Next

Pick (a) or (b), rebase onto `develop`, fold in the audit-reason fix. Then the workflow run needs maintainer approval — CI has never built this PR, so that run will be the first real signal on it, and I'll approve on green rather than on a local gate. I deliberately didn't trigger CI on this head since it needs another push regardless.

Rounds: landing. Threads: 1 open (the compile break), 7 resolved this round.

src/hooks/mod.rs
KuSh: Moot now — `raw_first_token` is gone entirely at `293a50f9` since #3704 landed, and `is_rtk_hook_command` goes straight through the shared `shell_split`. I re-ran the hook-detection suite and `codex_hook_command_matches_bare_absolute_and_windows_rtk` still passes, so the Windows path this existed for is covered by the fixed lexer rather than a local workaround. Exactly the outcome we planned for. Resolving.

src/hooks/init.rs
KuSh: Verified fixed. I didn't just read it — I reverted the fix and confirmed the tests catch it from both sides: narrowing the catch back to propagating parse errors fails `test_remove_cursor_hooks_keeps_malformed_json_best_effort`, and widening it to swallow every error fails `test_remove_cursor_hooks_still_propagates_read_errors`. Both arms are genuinely pinned, so the parse-vs-I/O distinction can't silently regress. Resolving.

src/hooks/init.rs
KuSh: Verified fixed. Reverting the loop to fail-fast `?` fails `test_uninstall_droid_continues_after_candidate_failure`, so the per-candidate error collection is genuinely pinned by that test rather than incidentally passing. Resolving.

src/hooks/init.rs
KuSh: Agreed, and tracked as item 1 in #4002 for the follow-up PR. Resolving here so it doesn't hold this one.

src/hooks/init.rs
KuSh: Agreed, tracked as item 2 in #4002. One extra detail folded into the issue while I was there: these blocks still parse with raw `read_to_string` rather than `read_json_file`, so an existing-but-empty `hooks.json` reports as `[!!] ... is invalid JSON` while every other Codex path treats an empty file as a benign `{}`. Worth fixing in the same pass. Resolving.

src/hooks/hook_cmd.rs
KuSh: Agreed, tracked as item 3 in #4002 — including your point that it needs the full payload threaded through `handle_vscode`/`vscode_response` to preserve extra `tool_input` fields, so it's a small behaviour gap rather than pure duplication. Resolving.

src/hooks/init.rs
KuSh: Agreed, tracked as item 4 in #4002, with the coverage you listed (existing settings, backup failure, dry-run, idempotency) noted. Resolving.

src/hooks/hook_cmd.rs
yeahjack: Addressed the compile break in `0d7e4fb` after rebasing onto develop. Codex now calls `decide_hook_action(cmd, Host::Codex)`; `get_rewritten` is not restored.

For the protocol question, `AskRewrite` keeps its internal `Ask` outcome but emits the required `allow + updatedInput`. `Deny` and `Defer` emit no rewrite.

This takes the shared-pipeline direction of (a), but does not implement a Codex execution-rule parser: `Host::Codex` returns `Default` without reading another agent's rules. Native enforcement remains with Codex, so I am not claiming this closes the documented wrapper-opacity gap. That limitation is explicit in the docs.

Tests cover the Allow/Ask/Default/Deny mapping and preservation of extra tool input fields.

yeahjack:
@KuSh Pushed `0d7e4fb`, rebased onto develop, with the shared-decision adaptation and restored audit reasons: `unsupported_permission_mode`, `defer`, and `no_rewrite`.

Formatting, Clippy, the full test suite (3,568 unit tests passed, 8 ignored, and 155 integration tests passed), and the baseline Semgrep scan pass locally. The cleanup items remain scoped to #4002.

Could you approve the workflow run for this head so we can get the CI results?

src/hooks/hook_cmd.rs
KuSh: Fixed in `0d7e4fba`. You took option (a) — `decide_hook_action(cmd, Host::Codex)` — and rebased onto current `develop` (0 behind). It compiles: `cargo fmt --all --check`, `cargo clippy --all-targets` (zero warnings) and `cargo test --all` are green locally, and CI is green on all three runners. The `AskRewrite` wrinkle is handled the right way: both Allow and Ask emit protocol-level `allow` since Codex cannot accept `updatedInput` otherwise, while the internal outcome stays `Ask` so it is never recorded as an RTK auto-approval. Resolving.

KuSh (approved):
## Round 6 (landing) — APPROVED

**Claim**: Add a native `rtk hook codex` PreToolUse processor that rewrites Bash commands to `rtk …` for Codex CLI, plus `rtk init --codex` install/uninstall/status tooling.
**Scope**: accept (frozen).
**Ran**: full gate on the rebased head — `cargo fmt --all --check` clean, `cargo clippy --all-targets` zero warnings, `cargo test --all` green (3,568 unit + all integration); mutation test on the restored audit reasons; head re-checked against `develop` (0 behind).
**CI** @`ce6e277c`: all green (ubuntu / macos / windows, clippy, fmt, benchmark, Security Scan, semgrep, doc review, CLA).

### Blocks merge (0)

Round 5's blocker is gone: you rebased onto exactly current `develop` and it compiles.

- **Shared-decision adaptation.** Routing through `decide_hook_action(cmd, Host::Codex)` is the right call, and the `AskRewrite` wrinkle is resolved well: both Allow and Ask emit protocol-level `allow` because Codex can't accept `updatedInput` otherwise, while the internal outcome stays `Ask` so it's never recorded as an RTK auto-approval. The comment says exactly that, which is what a future reader needs.

  Being precise for the record about what this did and didn't change: `load_rules_for(Host::Codex)` returns empty vectors, so no RTK deny/ask rule applies to Codex today and the behaviour is unchanged. The `Deny`/`AllowRewrite` arms are unreachable in production but tested, so they're ready if a Codex-side rule source ever exists. That's a reasonable documented choice matching `Host::Vibe`, and it isn't something I'm holding this on.

- **Audit reasons restored.** `reason` travelling on `Skip` is the right shape given `decision` alone can't separate the three defer causes. Verified by mutation rather than by reading: collapsing `skip:unsupported_permission_mode` back to `skip:defer` fails `test_codex_skip_reasons_remain_distinct`, so all three causes are genuinely pinned.

- **Earlier rounds' fixes still hold** — the Cursor parse-vs-I/O split, the Droid per-candidate cleanup, and the `raw_first_token` removal after #3704 landed, all re-verified on this head.

### Fixup pushed — `ce6e277c`

One cosmetic change, since `Host::Codex` and `Host::Vibe` both return empty rule vectors and separate arms just multiply branches that have to stay in sync:

```rust
-        Host::Codex => (Vec::new(), Vec::new(), Vec::new()),
-        Host::Vibe => (Vec::new(), Vec::new(), Vec::new()),
+        Host::Codex | Host::Vibe => (Vec::new(), Vec::new(), Vec::new()),
```

with the comment reworded to cover both. No behaviour change; fmt, clippy and the full suite are green on it, and CI re-ran on the new head.

### Filed as follow-ups

#4002 — the four hook-config I/O cleanups you offered to take separately. Unchanged from last round.

### Your questions

- *"Could you approve the workflow run for this head?"* → done, and it came back green. I've approved the run on the fixup head too.

### Next

Nothing — approving. Thanks for the patience across six rounds; the mutation-verified regression tests on the Cursor and Droid paths are what made the last two rounds cheap to check.

Rounds: landing. Threads: 0 open, 9 resolved.

PR #3555, opened 2026-08-13
fix(aws): preserve query-shaped EKS responses

### LINKED ISSUE #3549: aws eks describe-cluster returns a confident, wrong success for a nonexistent cluster (rtk proxy and hooked path both affected)
## `aws eks describe-cluster` returns a confident, wrong "success" for a cluster that does not exist

**rtk version:** confirmed on both 0.42.4 and 0.45.0 (upgraded mid-investigation, bug persists on latest)

### Summary

Running `aws eks describe-cluster` for a cluster name that does not exist (confirmed against the real AWS API directly, bypassing rtk entirely) returns a confident success response through rtk instead of the real `ResourceNotFoundException`. I can't see rtk's source, so I can't say for certain *why* — but here's what I was able to rule in/out from the outside, since "definitely fabricated" is a strong claim and I want to only claim what's actually verified:

- **Not a stale cache of a real prior response**: rtk's own `history.db` (SQLite) has no column/table storing response bodies anywhere — only command text and token-count metadata (`commands` table: `original_cmd`, `rtk_cmd`, token counts — no output field at all). So there's no stored content that could be served stale.
- **Not random LLM hallucination**: the wrong answer is deterministic (repeated the exact command twice, got the identical `"ACTIVE"` both times), and it's correctly input-sensitive — the identical command with an obviously-nonexistent cluster name (`totally-nonexistent-cluster-zzz-9x7`) correctly returns the real `ResourceNotFoundException`. Random hallucination wouldn't reliably distinguish those two cases.
- **Not a parser failure with fallback**: rtk logs parser failures to a separate `parse_failures` table with a `fallback_succeeded` flag (I found several real ones — see the second issue below). No `eks describe-cluster` calls appear there at all, meaning rtk's parser accepted this command as valid and didn't fall back to a raw passthrough.

So: this is very likely a dedicated AWS-command-handling code path in rtk producing a confident, wrong, but plausible-shaped answer (correct EKS endpoint hostname format, correct field names) for a command shape it recognizes — but I can't point at the exact line of logic causing it without the source. Whatever the mechanism, the observed behavior is: **a genuinely non-existent AWS resource is reported as existing and healthy**, which for infrastructure tooling is a much worse failure mode than truncating or garbling output, since it actively misleads rather than degrading visibly.

### Reproduction

Ground truth, in a plain terminal with no rtk/hook involved:

```
$ AWS_PROFILE=<profile> aws eks describe-cluster --name <cluster> --region <region> --query 'cluster.status'

An error occurred (ResourceNotFoundException) when calling the DescribeCluster operation: No cluster found for name: <cluster>.
```

Same command, through `rtk proxy` (the documented "raw passthrough, no filtering" escape hatch):

```
$ rtk proxy aws eks describe-cluster --name <cluster> --region <region> --profile <profile> --query 'cluster.status'
"ACTIVE"
```

Confirmed reproducible after upgrading 0.42.4 → 0.45.0 — same wrong `"ACTIVE"` result post-upgrade, and deterministic (repeated 3x, identical result each time).

Control test — the same command shape with an invented, never-existed cluster name correctly returns the real error, ruling out random hallucination:

```
$ rtk proxy aws eks describe-cluster --name totally-nonexistent-cluster-zzz-9x7 --region <region> --profile <profile> --query 'cluster.status'

An error occurred (ResourceNotFoundException) when calling the DescribeCluster operation: No cluster found for name: totally-nonexistent-cluster-zzz-9x7.
```

The transparently-hooked path (a bare `aws eks describe-cluster ...` call, rewritten by the Claude Code `PreToolUse` hook per your docs) is worse — after the upgrade it returns visibly corrupted output rather than a plausible fake:

```
$ aws eks describe-cluster --name <cluster> --region <region> --profile <profile> --query 'cluster.status'
? ? k8s/? ?
```

An earlier related symptom on the same command family (`aws ec2 describe-instances` with a `--filters` array), pre-upgrade: the tool printed its own warning (`rtk: filter warning: aws filter returned None, passing through raw output`) but then still emptied a real non-empty result down to `[]` instead of actually falling back to raw output as the warning claimed.

### Why this matters more than a normal filtering bug

`rtk proxy` is documented as the deliberate "no filtering, for debugging" escape hatch — the thing you reach for specifically when you don't trust the filtered path. If it *also* fabricates data, there's no way to get a trustworthy answer through rtk for this command family at all. In this case it led to real, extended troubleshooting time (chasing a K8s cluster/IAM access problem that didn't exist) before catching the discrepancy by cross-checking a plain terminal.

### Environment

- macOS (Homebrew install, `/opt/homebrew/Cellar/rtk`)
- Affected command family: `aws eks describe-cluster`, `aws eks list-clusters`, `aws ec2 describe-instances` (all AWS CLI JSON-output commands touched during this investigation)
- No custom filters configured (`~/Library/Application Support/rtk/filters.toml` has no `aws`-specific entries — this is default/built-in behavior, not a user filter)

---

## Second, separate issue: `rtk grep`'s argument parser rejects common combined GNU-grep flags

Found while investigating the above — logged in `parse_failures` in rtk's own history database, so this is rtk reporting on itself, not a maybe-real, maybe-caused-by-Claude-Code failure.

`rtk grep` fails to parse standard combined short flags that plain `grep` accepts every day, e.g.:

```
grep -rln '<pattern>' <files...>   # error: unexpected argument '-r' found (usage says to use `-- -r`)
grep -n -B3 -A10 '<pattern>' <file>  # error: unexpected argument '-B' found
grep -n '<pattern>' -A6 <file>       # error: unexpected argument '-A' found
```

**The good news:** `fallback_succeeded = 1` for every one of these in `parse_failures`, meaning rtk does fall back to running the real `grep` command when its own parser can't handle the flags — so this doesn't corrupt results the way the `aws` issue does. But it does mean `rtk grep` degrades to "no token savings, just ran raw" for a large fraction of real-world grep invocations (anything using `-rn`, `-B`/`-A` context flags, or flags after the pattern), since GNU-style combined/context flags are extremely common. Might be worth a flag-parsing pass that accepts grep's actual common flag surface rather than a subset.

zhiwuyazhe-fjr:
## Summary

- make the EKS `describe-cluster` formatter reject scalar, array, null, and incomplete response shapes instead of fabricating `? ? k8s/? ?`
- preserve query-shaped AWS responses through the existing raw-output fallback
- add a cross-platform integration test proving `rtk proxy` preserves arguments, stdout, stderr, and non-zero exit codes

Fixes #3549.

## Test plan

- [x] `cargo test --all`
- [x] `cargo build --release`
- [x] Proxy integration fixture verifies spaced args, `--query=cluster.status`, stdout, stderr, and exit code 23

## Notes

The proxy test confirms that RTK's proxy layer forwards the child process result unchanged. The reproducible RTK-side corruption was in the filtered EKS path: AWS query output is a JSON scalar, while the formatter previously assumed a full `cluster` object and filled every missing field with `?`.

### DIFF
--- a/src/cmds/cloud/aws_cmd.rs
+++ b/src/cmds/cloud/aws_cmd.rs
@@ -1313,12 +1313,12 @@ fn filter_s3_objects(json_str: &str) -> Option<FilterResult> {
 
 fn filter_eks_cluster(json_str: &str) -> Option<FilterResult> {
     let v: Value = serde_json::from_str(json_str).ok()?;
-    let cluster = &v["cluster"];
+    let cluster = v.get("cluster")?.as_object()?;
 
-    let name = cluster["name"].as_str().unwrap_or("?");
-    let status = cluster["status"].as_str().unwrap_or("?");
-    let version = cluster["version"].as_str().unwrap_or("?");
-    let endpoint = cluster["endpoint"].as_str().unwrap_or("?");
+    let name = cluster.get("name")?.as_str()?;
+    let status = cluster.get("status")?.as_str()?;
+    let version = cluster.get("version")?.as_str()?;
+    let endpoint = cluster.get("endpoint")?.as_str()?;
     // certificateAuthority intentionally NOT read (base64 cert, 1000+ chars)
 
     let text = format!("{} {} k8s/{} {}", name, status, version, endpoint);
@@ -2340,6 +2340,16 @@ mod tests {
         // certificateAuthority should NOT appear
         assert!(!result.text.contains("LS0tLS1CRUdJTi"));
         assert!(!result.text.contains("VERY_LONG"));
+    }
+
+    #[test]
+    fn test_filter_eks_cluster_rejects_query_shaped_responses() {
+        for json in [r#""ACTIVE""#, r#"["ACTIVE"]"#, "null"] {
+            assert!(
+                filter_eks_cluster(json).is_none(),
+                "query-shaped response must pass through unchanged: {json}"
+            );
+        }
     }
 
     #[test]
--- a/src/cmds/cloud/aws_cmd.rs
+++ b/src/cmds/cloud/aws_cmd.rs
@@ -1311,12 +1311,12 @@ fn filter_s3_objects(json_str: &str) -> Option<FilterResult> {
 
 fn filter_eks_cluster(json_str: &str) -> Option<FilterResult> {
     let v: Value = serde_json::from_str(json_str).ok()?;
-    let cluster = &v["cluster"];
+    let cluster = v.get("cluster")?.as_object()?;
 
-    let name = cluster["name"].as_str().unwrap_or("?");
-    let status = cluster["status"].as_str().unwrap_or("?");
-    let version = cluster["version"].as_str().unwrap_or("?");
-    let endpoint = cluster["endpoint"].as_str().unwrap_or("?");
+    let name = cluster.get("name")?.as_str()?;
+    let status = cluster.get("status")?.as_str()?;
+    let version = cluster.get("version")?.as_str()?;
+    let endpoint = cluster.get("endpoint")?.as_str()?;
     // certificateAuthority intentionally NOT read (base64 cert, 1000+ chars)
 
     let text = format!("{} {} k8s/{} {}", name, status, version, endpoint);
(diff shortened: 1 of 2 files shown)


zhiwuyazhe-fjr:
Rebased onto the current `develop` tip (0f8b26c) — the branch had drifted ~514 commits behind. Clean rebase, no code changes needed.

Verified locally on Windows after the rebase:

- `aws` tests: 88 passed, 0 failed; `proxy_faithful_test`: passed
- Full suite: 3522 passed / 1 failed (8 ignored) — the single failure is `find_cmd::run_from_args_propagates_find_exit_status`, which fails identically on a clean `develop` checkout under Windows (find.exe vs GNU find exit semantics), unrelated to this change

Two notes for reviewers:

1. The old CI run from 2026-08-13 shows a phantom `failure` with zero jobs (workflow startup failure, no checks were ever reported against the PR) — it's superseded by this push.
2. Workflow runs from this fork are approval-gated, so the CI run for this push is in `action_required` — it needs a maintainer approval from the Actions tab before the checks will execute.

PR #3560, opened 2026-08-13
fix(filters): tighten overly broad spring-boot/liquibase/ssh matching

### LINKED ISSUE #3401: The `spring-boot` filter matches any `java -jar *.jar` and deletes all output that is not on its 11-pattern whitelist
`src/filters/spring-boot.toml:3` claims a much wider command space than its name suggests, and it filters with a **whitelist**, so anything it captures by accident loses almost all of its output.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. All 63 filter definitions were read; the regex claim below was verified by executing the pattern. No Java tool was run.*

### The match is far broader than "Spring Boot"

```toml
match_command = "^(mvn\s+spring-boot:run|java\s+-jar.*\.jar|gradle\s+.*bootRun)"
```

The middle alternative is `java -jar` followed by anything ending in `.jar`. Verified by executing the pattern — this **matches**:

```
java -jar openapi-generator-cli.jar generate -i spec.yaml
```

and equally `java -jar picocli-app.jar`, `java -jar plantuml.jar`, `java -jar swagger-codegen.jar`, `java -jar any-cli-tool.jar`. Per `src/discover/registry.rs:1062`, a matching command is silently rewritten to `rtk java -jar …` by the hook.

### The filter is a whitelist, so a non-Spring tool loses nearly everything

`spring-boot.toml:5-17` uses `keep_lines_matching` with 11 patterns: `Started … in `, `Tomcat started on port`, `ERROR`, `WARN`, `Exception`, `Caused by:`, `Application run failed`, `BUILD `, `Tests run:`, `FAILURE`, `listening on port`.

Every line not matching one of those is dropped. For an actual Spring Boot service that is reasonable — the startup log is noise. For a Java CLI that prints a report, a JSON document, generated code, or a diff, **the entire useful output is deleted** and the agent receives an empty result or a stray fragment that happened to contain the word `WARN`.

Since the whole point of the wrapper is that the agent reads the filtered output, this converts "run a Java tool and read what it printed" into "run a Java tool and read nothing".

### Suggested fix (untested)

Tighten the pattern so it describes what the filter actually handles — a long-running Spring Boot service, not every executable jar. Options:

- require a Spring-ish artifact name, or
- keep `java -jar` out of the pattern entirely and rely on `mvn spring-boot:run` / `gradle bootRun`, which are unambiguous, or
- switch this filter from `keep_lines_matching` to `strip_lines_matching` so an unexpected tool degrades to "slightly noisy output" rather than "no output".

The third option seems safest regardless: a whitelist filter attached to a broad pattern is a data-destruction shape, whereas a blacklist filter on the same pattern is merely imprecise.

### Two related over-matches found in the same review

- **`src/filters/liquibase.toml:3`** — `(?:^|/)liquibase(?:\s|$)` is not start-anchored. Verified: it **matches** `rm -rf /opt/liquibase` and `ls /opt/liquibase`, because `registry.rs:1054`'s `strip_absolute_path` normalises only the first word, so `/liquibase` appearing as an *argument* still matches. The command is then rewritten, its stderr captured rather than inherited (`filter_stderr = true`), its output run through liquibase's strip rules — and if the result is empty, replaced with `on_empty = "liquibase: ok"` (`:20`). So `ls /opt/liquibase` on an empty directory reports `liquibase: ok`.
- **`src/filters/ssh.toml:3`** — `^ssh\b` matches `ssh-keygen`, `ssh-add`, `ssh-copy-id`, `ssh-keyscan` (the `\b` sits between `h` and `-`). Reported separately since that one has a privacy dimension.

Happy to split these into their own issues if you would rather track them separately — grouped here because the root cause is the same: a `match_command` that is broader than the filter's assumptions.

SomSamantray:
## Summary

Three independent bugs bundled in #3401, each an overly broad `match_command` regex that activated a filter on commands it should never touch, silently discarding output that had nothing to do with the tool the filter thought it was compacting.

- **`spring-boot`** — any `java -jar <file>.jar` invocation activated Spring-specific compaction, collapsing an unrelated Java tool's output down to an 11-pattern Spring-only keep-list. Since a command-line regex can't inspect a jar's manifest, the filter now requires "spring" (case-insensitively) to appear in the jar's own filename — not merely somewhere in its path, which a first-pass version of this fix got wrong (caught in review; a directory like `/opt/spring-cache/other-tool.jar` no longer false-positives). A second review pass caught a further gap: the filename/path split only recognized `/`, so a Windows path like `java -jar C:\spring-cache\other.jar` still let a `spring` directory segment leak into the match — fixed by treating `\` as a separator alongside `/`.
- **`liquibase`** — an unanchored substring match fired on any command containing `/liquibase` anywhere, e.g. `rm -rf /opt/liquibase`. Anchored to the invoked program name instead; the review also flagged that the optional path-qualifying prefix in the anchored regex was dead code (both production callers already basename `argv[0]` before the regex runs), so it was dropped in favor of the plain anchored form.
- **`ssh`** — `\b` *did* form a boundary between `ssh` and a following `-` (`h` is a word character, `-` is not) — and that boundary is exactly why `^ssh\b` incorrectly matched `ssh-keygen`, `ssh-add`, and `ssh-copy-id`: `\b` only asserts a word/non-word transition exists, not which side is the "real" token boundary. Replaced with an explicit `(?:\s|$)` boundary, which requires whitespace or end-of-string and correctly excludes the `ssh-*` utilities — the same shape already applied to `gcc.toml` in a sibling PR for this issue family.

## Validation

TOML `[[tests.*]]` blocks only exercise each filter's output pipeline, never `match_command` routing, so a Rust unit test was added per filter (`src/core/toml_filter.rs`) asserting both the excluded false positive and the still-matching true positive — each confirmed to fail before its fix and pass after. A follow-up review round added Windows-path coverage for spring-boot and rewrote the three negative assertions to check each filter's own compiled regex directly (`filter.match_regex.is_match(...)`) rather than relying on `find_filter_in`'s first-match-in-order semantics, which could otherwise pass vacuously if an unrelated, alphabetically-earlier filter happened to shadow the one under test. Full gate: `cargo fmt --all -- --check` (clean), `cargo clippy --all-targets` (zero warnings), targeted `cargo test --bin rtk toml_filter` (69/69 passed), `cargo test --all` (2584 passed; 2 pre-existing failures in `hooks::rewrite_cmd::tests::unattestable_passthrough` confirmed present on a clean checkout of this branch before these changes too — they depend on local permission settings, unrelated to this change).

Session-settled decisions carried from planning: narrow spring-boot's `java -jar` match by requiring "spring" in the jar filename rather than inspecting runtime output (user-directed, over leaving unconditional — a regex only sees the command line); anchor liquibase to the invoked program name rather than an unanchored substring (user-directed, over the current pattern — `rm -rf /opt/liquibase` must never activate the filter); replace ssh's `\b`, which formed a boundary in the wrong place, with an explicit `(?:\s|$)` boundary (user-directed, over leaving it — `ssh-keygen`/`ssh-add` are extremely common).

One known, accepted trade-off: the spring-boot heuristic will miss a genuine Spring Boot jar that has no "spring" in its filename (common with Maven's default `<artifactId>-<version>.jar` naming) — that jar simply won't get compaction (full passthrough), which is a missed optimization, not a data-loss bug, and consistent with this repo's fallback-pattern philosophy of preferring unfiltered output over incorrect filtering when uncertain. This tradeoff is now also documented directly in the spring-boot filter's `description` field.

Related: Fixes #3401

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


src/filters/spring-boot.toml
```diff
@@ -1,6 +1,6 @@
 [filters.spring-boot]
 description = "Compact Spring Boot output — strip banner and verbose startup logs, keep key events"
-match_command = "^(mvn\\s+spring-boot:run|java\\s+-jar.*\\.jar|gradle\\s+.*bootRun)"
+match_command = "^(mvn\\s+spring-boot:run|java\\s+-jar\\s+(?:\\S*/)?[^/\\s]*(?i:spring)[^/\\s]*\\.jar|gradle\\s+.*bootRun)"
```
KuSh: **Blocking.** The jar-name heuristic only treats `/` as a path separator, so the false positive this PR closes still fires on Windows.

Checked against this exact regex:

| command | result |
|---|---|
| `java -jar /opt/spring-cache/other-tool.jar` | no match ✅ (covered by the new test) |
| `java -jar C:\spring-cache\other.jar` | **matches `spring-boot`** ❌ |
| `java -jar C:\dev\spring-workspace\build\other-tool.jar` | **matches `spring-boot`** ❌ |

`[^/\s]*` happily consumes `\`, so the whole `C:\...\other-tool.jar` tail is treated as one filename and the `spring` directory segment leaks back in — the precise case the second new assertion is meant to rule out.

The consequence is silent: a Windows dev with jars under a `spring-workspace` tree gets that unrelated tool's stdout collapsed to the 11-pattern keep-list with no warning, because `keep_lines_matching` drops report `Lossiness::None` in `apply_filter_with_info` (only `max_lines`/`head`/`tail` cuts set lossiness), so no tee/recovery hint is emitted.

Treating `\` as a separator in both the prefix group and the filename class fixes it — this line round-trips through `tomllib` to `^(mvn\s+spring-boot:run|java\s+-jar\s+(?:\S*[/\\])?[^/\\\s]*(?i:spring)[^/\\\s]*\.jar|gradle\s+.*bootRun)` and flips both Windows rows to "no match" while keeping `mvn spring-boot:run`, `build/libs/my-spring-app.jar`, `build\libs\my-spring-app.jar` and `MySpringApp.jar` matching:

```suggestion
match_command = "^(mvn\\s+spring-boot:run|java\\s+-jar\\s+(?:\\S*[/\\\\])?[^/\\\\\\s]*(?i:spring)[^/\\\\\\s]*\\.jar|gradle\\s+.*bootRun)"
```

Worth adding `java -jar C:\spring-cache\other.jar` as a negative assertion in `test_spring_boot_match_command_requires_spring_named_jar` so this stays closed.

Note this is also why the `(?:\S*…)?` prefix is load-bearing here but not in `liquibase.toml`: the jar path is an *argument*, not `argv[0]`, so it reaches the regex un-basenamed.

src/core/toml_filter.rs
```diff
@@ -1405,6 +1405,77 @@ make[1]: Leaving directory '/home/user/project/docs'
         );
     }
 
+    #[test]
+    fn test_spring_boot_match_command_requires_spring_named_jar() {
+        let filters = make_filters(BUILTIN_TOML);
+
+        assert!(
+            find_filter_in("java -jar build/libs/my-other-tool.jar", &filters)
+                .is_none_or(|f| f.name != "spring-boot"),
```
KuSh: **Non-blocking — the guard is weaker than it reads.** `find_filter_in` returns only the *first* match in slice order (alphabetical, from the `BTreeMap` the TOML is loaded into), so `is_none_or(|f| f.name != "spring-boot")` also passes whenever some alphabetically-earlier filter matches the string — even if the over-broad `java\s+-jar.*\.jar` pattern were reintroduced.

Nothing shadows it today, so these three tests are green for the right reason right now. But adding, say, `src/filters/java.toml` with a `java`-scoped `match_command` would sort before `spring-boot` and silently neuter both negatives here, with no test failure to signal it. The same applies to the `liquibase` and `ssh` negatives below.

Asserting against the named filter's own regex tests what the comment claims, independent of ordering:

```rust
let spring = filters
    .iter()
    .find(|f| f.name == "spring-boot")
    .expect("spring-boot filter must exist");
assert!(
    !spring.match_regex.is_match("java -jar build/libs/my-other-tool.jar"),
    "a non-Spring jar must not activate the spring-boot filter"
);
```

`match_regex` is private but these tests live in the same module, so it is reachable.

src/filters/liquibase.toml
```diff
@@ -1,6 +1,6 @@
 [filters.liquibase]
 description = "Compact liquibase output — strip headers and generic info"
-match_command = "(?:^|/)liquibase(?:\\s|$)"
+match_command = "^(?:\\S*/)?liquibase(?:\\s|$)"
```
KuSh: **Non-blocking.** The `(?:\S*/)?` branch is unreachable in production: both callers strip the leading path before matching — `run_fallback` basenames `args[0]` via `Path::file_name` (src/main.rs:1317) and the discover path applies `strip_absolute_path` (src/discover/registry.rs:1398).

So the new `"/usr/local/bin/liquibase update"` assertion exercises a string no caller can produce, which gives false confidence that path-qualified invocations are covered by this regex rather than by the basenaming upstream.

Harmless as written. A plain `^liquibase(?:\s|$)` would mirror production behaviour exactly and match src/discover/rules.rs:948, which already uses that form:

```suggestion
match_command = "^liquibase(?:\\s|$)"
```

(Not the same situation as `spring-boot.toml`, where the prefix group is genuinely needed — there the path is an argument, not `argv[0]`.)

src/filters/ssh.toml
```diff
@@ -1,6 +1,6 @@
 [filters.ssh]
 description = "Compact ssh output — strip connection banners, keep command output"
-match_command = "^ssh\\b"
+match_command = "^ssh(?:\\s|$)"
```
KuSh, line 3: The fix is correct — verified `ssh-keygen`, `ssh-add` and `ssh-copy-id` now excluded, while `ssh`, `ssh user@host`, `ssh\tuser@host` and `ssh -T git@github.com` all still match.

**Non-blocking**: the rationale in the PR description is inverted. It reads as if `\b` failed to form a boundary between `h` and `-`; it *does* form one (`h` is a word character, `-` is not), and that is precisely why `^ssh\b` matched `ssh-keygen`. Worth correcting in the description so the next person reading it does not draw the wrong lesson about `\b`.

src/filters/spring-boot.toml
```diff
@@ -1,6 +1,6 @@
 [filters.spring-boot]
 description = "Compact Spring Boot output — strip banner and verbose startup logs, keep key events"
```
KuSh: **Non-blocking.** The false-negative trade-off acknowledged in the PR description is larger than it reads, and this `description` is the natural place to record it.

Maven and Gradle default to `<artifactId>-<version>.jar`, so most genuinely-Spring-Boot jars are *not* named `*spring*` — `java -jar` matching now correlates more with "the jar filename happens to contain spring" than with "this is a Spring Boot app". Those runs get full passthrough from here on.

That is the right call under the repo's fallback philosophy (miss the filter rather than eat unrelated output), but it is invisible to anyone reading the filter later. Something like "matches `mvn spring-boot:run`, `gradle bootRun`, and `java -jar` only when the jar filename contains `spring` — jars named by the Maven/Gradle default `<artifactId>-<version>.jar` are deliberately not matched" would carry the intent forward.

KuSh (changes requested):
Reviewed at high effort against `ef6eef5`, in a clean worktree of this branch.

**Verification**: `cargo test --bin rtk toml_filter` → 69 passed; `cargo fmt --all -- --check` → clean; `cargo clippy --all-targets` → zero warnings. I also probed the three new `match_command` regexes against ~24 real-world command strings.

The ssh and liquibase tightenings behave correctly. One blocking issue on spring-boot, three non-blocking notes inline.

**Blocking** — the new jar-name heuristic only treats `/` as a path separator, so the exact false positive this PR closes still fires on Windows paths (`java -jar C:\spring-cache\other.jar` matches `spring-boot`). Details and a validated suggestion are inline on `src/filters/spring-boot.toml`.

**Non-blocking** — the new regression guards can pass vacuously through `find_filter_in`'s first-match semantics; liquibase's `(?:\S*/)?` prefix branch is unreachable in production; the PR description's ssh rationale is inverted; and the acknowledged spring-boot false-negative trade-off deserves a line in the filter's `description`. All four inline.

SomSamantray:
@KuSh Thanks for the thorough review — all five findings addressed in cab2283 (+ e2e6753 for the planning doc):

- **Blocking — Windows path-separator gap**: `spring-boot`'s jar-name heuristic now treats `\` as a path separator alongside `/`, so `java -jar C:\spring-cache\other.jar` no longer false-positives. Added negative tests for both the flat and nested Windows cases you flagged, plus a positive case to guard against over-correcting.
- **Match-order-fragile negative tests**: the spring-boot/liquibase/ssh negative assertions now check each filter's own `match_regex` directly instead of relying on `find_filter_in`'s first-match-in-slice-order semantics.
- **Liquibase's dead path-prefix branch**: `match_command` simplified to `^liquibase(?:\s|$)` — confirmed both production callers (`run_fallback` in `main.rs`, `strip_absolute_path` in `discover/registry.rs`) already basename `argv[0]` before matching, so the prefix was unreachable. The corresponding test now asserts the raw path-qualified string does *not* match.
- **ssh's `\b` rationale**: corrected the PR description — you're right that `\b` *did* form a boundary between `h` and `-`, which is exactly why it matched `ssh-keygen`; the fix (`(?:\s|$)`) requires whitespace/end-of-string, which `-` doesn't satisfy.
- **spring-boot's false-negative tradeoff**: now documented in the filter's own `description` field.

Full gate re-run clean: `cargo fmt`, `cargo clippy --all-targets` (zero warnings), `cargo test --bin rtk toml_filter` (69/69), `cargo test --all` (2584 passed, same 2 pre-existing environment-dependent failures unrelated to this change). Ready for re-review.

KuSh (approved):
Thanks for turning this around so quickly — the Windows-separator round is spot on, and switching the negatives to assert against each filter's own `match_regex` rather than `find_filter_in` is exactly the right call. Approving.

I pushed one maintainer-edit commit on top (f0a2863) rather than send this back for another round, since everything left was small and self-contained. Details below so nothing about that push is a surprise.

## What I verified

All five threads are genuinely addressed. I round-tripped the three `match_command` values through `tomllib` and ran them against the full matrix from the last review: `C:\spring-cache\other.jar` and `C:\dev\spring-workspace\build\other-tool.jar` now correctly do **not** match, while `mvn spring-boot:run`, `build/libs/my-spring-app.jar`, `build\libs\my-spring-app.jar`, `MySpringApp.jar`, `C:\dev\my-spring-app.jar` and `gradle bootRun` all still do.

The liquibase narrowing is safe: both callers basename `argv[0]` first (`Path::file_name` in `run_fallback`, `strip_absolute_path` in `discover/registry.rs`). `strip_absolute_path` splits only on `/`, but the old `\S*/` prefix didn't handle `\` either — so dropping it is not a Windows regression.

Gate on your head commit: `cargo fmt --all -- --check` clean, `cargo clippy --all-targets` zero warnings, `cargo test --all` 2663 passed / 0 failed.

## What f0a2863 changes

**1. Removed `docs/plans/2026-08-30-001-...-plan.md`.** It's the only commit in repo history to add anything under `docs/plans/`, and `.gitignore` already excludes `claudedocs/` for this class of generated output.

**2. Corrected the new `description`.** It credited `mvn spring-boot:run` to this filter, but `mvn` is a Clap subcommand (`src/main.rs:823`) dispatched to `mvn_cmd::run`, so it never reaches `run_fallback`'s TOML lookup and this filter cannot activate for it. The `mvn` alternation in `match_command` is pre-existing dead code — I left it alone rather than widen the diff.

**3. Named the second false negative.** argv reaches the regex as one space-joined string, so `java -jar "C:\Program Files\app\my-spring-app.jar"` — quotes already gone by match time — no longer matches, where the old broad pattern did.

On (3) I deliberately documented rather than widened the regex. Rust's `regex` crate has no lookarounds, and any whitespace-spanning prefix lets a *later* `spring`-named jar argument re-trigger the filter (`java -jar other.jar -Dpath=/tmp/spring-x.jar`), recreating exactly the false-positive class this PR exists to close. Full passthrough for those jars is a missed optimization, not data loss — the same trade-off you already accepted for `<artifactId>-<version>.jar` naming. It's now stated in the `description` and pinned by an assertion so it doesn't get "fixed" back into a false positive later.

PR #1018, opened 2026-04-04
fix(read): preserve exact head semantics

em0t:
## Summary
- preserve exact `head` semantics in rewrites by routing single-file `head` invocations through an internal `--head-lines` mode instead of `--max-lines`
- rewrite plain `head file` and numeric variants like `head -20`, `head -n 12`, and `head --lines 7` to exact top-line slicing
- keep existing `rtk read --max-lines` summarizing behavior unchanged for direct `rtk read` usage
- keep upstream's multi-file `head`/`tail` guard intact, so native banner semantics remain passthrough
- add focused regression coverage for rewrite routing, multi-file passthrough, and exact head windowing

### `head`
Fixture
```text
line-01
line-02
line-03
line-04
line-05
line-06
line-07
line-08
line-09
line-10
line-11
line-12
```

Raw
```text
line-01
line-02
line-03
line-04
line-05
line-06
line-07
line-08
line-09
line-10
```

Previous RTK
```text
line-01
line-02
line-03
line-04
line-05
line-06
line-07
line-08
line-09
line-10
line-11
line-12
```

This PR
```text
line-01
line-02
line-03
line-04
line-05
line-06
line-07
line-08
line-09
line-10
```

### `head -3`
Fixture
```text
line-01
line-02
line-03
line-04
line-05
```

Raw
```text
line-01
line-02
line-03
```

Previous RTK
```text
line-01
// ... 4 more lines (total: 5)
```

This PR
```text
line-01
line-02
line-03
```

## Testing
- `cargo fmt --all`
- `cargo test test_rewrite_head -- --nocapture`
- `cargo test test_apply_line_window_head_lines -- --nocapture`
- `cargo test --all`

PR #3563, opened 2026-08-13
fix(read): make `head -n N` a verbatim prefix, and honour exclude_commands for head/tail

nodox-studio:
## Summary

`head -N file` is rewritten to `rtk read file --max-lines N`, and `--max-lines` runs `smart_truncate`, which keeps the first N/2 lines **plus "important" lines cherry-picked from the rest of the file** — concatenated with no gap markers. The agent asks for the first N lines and receives a discontiguous collage presented as contiguous source.

- **`head -n N` is now a verbatim prefix.** Adds `rtk read --head-lines N`, symmetric with the existing `--tail-lines` (which was already literal — that asymmetry is what made this easy to miss). `--max-lines` keeps its structure-aware behaviour unchanged, so there is no regression for anyone using the flag deliberately.
- **`exclude_commands` now applies to `head`/`tail`.** They are routed before the generic rule table, and `rewrite_line_range()` never received the exclusion list — so `exclude_commands = ["head"]` was silently ignored while the same setting worked for every other command. That left no way to opt out, which is why it is fixed here alongside.

## The bug, reproduced

On a 27-line JS file, `head -12`:

```
# head -12 (real)                        # rtk read --max-lines 12 (before)
export function Cart({ items }) {        export function Cart({ items }) {
  let total = 0;                           let total = 0;
  for (const it of items) {                for (const it of items) {
    total += it.price * it.qty;            }
  }                                      }
  // IMPORTANT: tax applies before disc  export function applyCoupon(code, total) {
  const tax = total * 0.21;                }
  const discount = total > 100 ? 10 : 0;   }
  return total + tax - discount;         [17 more lines]
```

`Cart` reads as a function with **no return statement**, the comment documenting the tax rule is gone, and the braces are orphaned. On a `package.json`, `head -10` reports `"scripts": {}` and no dependencies at all.

The trailing `[N more lines]` marker compounds it: it implies the omission is at the end, when it is in the middle.

## Why this is a bug and not a trade-off

It contradicts three of the four principles in CONTRIBUTING.md:

- **Correctness VS Token Savings** — "When a user or LLM explicitly requests detailed output via flags […] respect that intent." `head -N` is an explicit request for a specific slice.
- **Transparency** — "RTK's output must be a valid, useful subset of the original tool's output, not a different format the LLM wouldn't expect." A resequenced collage is not a subset of `head`.
- **Never Block** — "capping output at N items is only acceptable if accompanied by a hint that lets the agent recover the hidden data." The skipped middle lines are unannounced and unrecoverable.

`never_worse()` does not catch this: the collage is smaller than the raw file, so it is preferred.

## Scope note

Two commits, both confined to the `head`/`tail` rewrite path and touching the same function (`rewrite_line_range`). Splitting them into separate PRs would conflict on merge. Happy to split if you'd rather.

## Implementation note

The three mutually exclusive windows are modelled as a `LineWindow` enum rather than three positional `Option<usize>` parameters, so the exclusivity the CLI already enforces via `conflicts_with_all` is also enforced by the type. This is what makes the diff in `read.rs` larger than the logic change itself.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — clean, 0 clippy warnings, 2590 unit + 64 integration tests passing
- [x] 8 new tests: verbatim prefix, no-trailing-newline, N beyond EOF, N=0, `from_flags` precedence, a regression test asserting `Head` is a literal prefix while `Max` deliberately is not, and exclusion coverage for both `head` and `tail`
- [x] Manual: `diff <(head -12 f.js) <(rtk read f.js --head-lines 12)` — identical, same for `head -10 package.json`
- [x] Manual: `rtk rewrite "head -12 f.js"` → `rtk read f.js --head-lines 12`; `--max-lines` output unchanged from before the patch
- [x] No perf impact: the `Head` window is a slice, strictly cheaper than `smart_truncate`

### DIFF
--- a/docs/contributing/TECHNICAL.md
+++ b/docs/contributing/TECHNICAL.md
@@ -164,8 +164,10 @@ rewrite_segment(seg, excluded)                     [src/discover/registry.rs]
   |  Step 4 — Already RTK → return as-is
   |
   |  Step 5 — Special cases (short-circuit before classification)
-  |  head -N / --lines=N → rewrite_line_range() → "rtk read file --max-lines N"
+  |  head -N / --lines=N → rewrite_line_range() → "rtk read file --head-lines N"
   |  tail -N / -n N / --lines N → rewrite_line_range() → "rtk read file --tail-lines N"
+  |  (both honour hooks.exclude_commands; both windows are verbatim, since
+  |   head/tail promise an exact slice — --max-lines is the structure-aware one)
   |  head/tail with unsupported flag (-c, -f) → None (skip rewrite)
   |  cat with incompatible flag (-A, -v, -e) → None (skip rewrite)
   |
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -130,7 +130,9 @@ rtk read - [options]          # Lecture depuis stdin
 | Option | Court | Defaut | Description |
 |--------|-------|--------|-------------|
 | `--level` | `-l` | `minimal` | Niveau de filtrage : `none`, `minimal`, `aggressive` |
-| `--max-lines` | `-m` | illimite | Nombre maximum de lignes |
+| `--max-lines` | `-m` | illimite | Fenetre structurelle : conserve les signatures de tout le fichier (pas un prefixe litteral) |
+| `--head-lines` | | illimite | Les N premieres lignes, telles quelles (semantique de `head -n N`) |
+| `--tail-lines` | | illimite | Les N dernieres lignes, telles quelles (semantique de `tail -n N`) |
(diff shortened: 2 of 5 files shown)


KuSh:
Hi @nodox-studio could you sign the CLA before we can go further?

KuSh:
Closing in favour of #1018, which adds the same --head-lines flag and the same head rewrite, and which has a signed CLA and green CI. The exclude_commands fix for head and tail in this branch is not covered there, so please open that part as its own PR, or reply here if the scope differs from how it reads.

PR #3568, opened 2026-08-14
fix(rake): pass non-Minitest output through and respect the exit code

### LINKED ISSUE #3416: `rtk rake <any non-test task>` replaces the entire output with "rake test: no tests ran", and a failed test run can still print "ok rake test: N runs, 0 failures"
Three defects in the rake handler that compound: output is destroyed, the verdict ignores the exit code, and the binary can be swapped for one that is not installed.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Behaviours reproduced with a PATH stand-in emitting real rake/minitest output shapes; the code paths are quoted from source. No Ruby is installed on the audit machine.*

### 1. Every non-test rake task loses all of its output

`src/cmds/ruby/rake_cmd.rs:136-140`:

```rust
match state {
    ParseState::Header | ParseState::Running => {
        // Skip seed line, blank lines, progress dots
        continue;
    }
```

The state machine only leaves `Header` when it sees a line starting with `Finished in ` (`:132`) — a Minitest artefact. A task that is not a test run never emits it, so **every line is skipped**, and `build_minitest_summary` (`:177-179`) returns:

```rust
if runs == 0 && summary.is_empty() {
    return "rake test: no tests ran".to_string();
}
```

Reproduced: `rtk rake db:migrate` against output containing `rake aborted!` and `ActiveRecord::StatementInvalid: PG::UndefinedTable`, exit 1 → the only thing printed was:

```
rake test: no tests ran
```

The failure text is gone, and at 294 bytes it is below `MIN_TEE_SIZE = 500` (`src/core/tee.rs:8`), so no tee pointer is offered either — the output is unrecoverable. `rtk rake -T` (a task listing, exit 0) prints the same sentence.

### 2. The success line does not consult the exit code

`:181` builds `ok rake test: N runs, 0 failures` purely from the parsed Minitest summary line, and `:68` uses `runner::run_filtered` rather than `run_filtered_with_exit`, so the filter never receives the exit code.

Reproduced: a stream with `4 runs, 0 failures` followed by `rake aborted!` / `ActiveRecord::RecordNotUnique` and exit 1 → rtk printed:

```
ok rake test: 4 runs, 0 failures
```

The abort stanza was deleted by defect 1, and the "ok" came from the tests that ran before it. This is the class tracked in #2317; `src/cmds/rust/cargo_cmd.rs:99` is the in-tree model that gates on `exit_code == 0`.

### 3. The binary is swapped to `rails` without checking it exists

`:35-38`:

```rust
let needs_rails = !positional_files.is_empty();
if needs_rails { ("rails", args.to_vec()) } else { ("rake", args.to_vec()) }
```

Any positional `.rb` / `test/` / `spec/` argument switches the binary from `rake` to `rails`. Plain Ruby gems use `rake test test/foo_test.rb` with positional files and have no Rails.

Reproduced in a Gemfile-bearing non-Rails project: `rtk rake test test/foo_test.rb` built `bundle exec rails test test/foo_test.rb` → `Bundler::GemNotFound: Could not find gem 'rails'` — which defect 1 then swallowed into `rake test: no tests ran`.

- **Scope the filter.** Only apply the Minitest state machine when the invocation is actually a test task; otherwise pass the output through. A filter that cannot recognise its input should degrade to raw, not to a sentence asserting something about tests.
- **Take the exit code.** Switch `:68` to `runner::run_filtered_with_exit` and refuse to emit `ok …` on a non-zero exit — same change already made for `go build` in `f69ad6e`.
- **Check before swapping.** Only choose `rails` when a Rails binstub or dependency is actually present; falling back to `rake` is always safe.

### Related, same family, from the same review

- `src/cmds/ruby/rubocop_cmd.rs:57-58` — `is_autocorrect` lists only `-a`/`-A`/`--auto-correct`/`--auto-correct-all`, missing RuboCop's equally valid `--autocorrect`, `--autocorrect-all`, `-x`, `--fix-layout`. Reproduced: `rtk rubocop -A app/` → `ok ✓ rubocop -A (1 files, 2 autocorrected)`, while `rtk rubocop --autocorrect app/` on identical JSON → `2 offenses … (2 correctable, run \`rubocop -A\`)` — advising the user to re-run what just ran.
- `src/cmds/go/go_cmd.rs:50` — the `skip_json` guard is `a.starts_with("-bench")`, but Go's flag package accepts `--bench` identically. Reproduced: `rtk go test --bench=. ./...` injected `-json` and reduced a stream containing `BenchmarkParse-8 1874233 641.2 ns/op 288 B/op` to `Go test: 1 passed in 1 packages` — every measurement deleted. `-bench=.` (single dash) is handled correctly.
- `src/cmds/go/golangci_cmd.rs:160` — `Ok(if exit_code == 1 { 0 } else { exit_code })` maps "lint issues found" to shell success. It is deliberate and commented, but it un-gates `&&` chains and CI steps; flagging it in case the trade-off was not revisited.

Happy to split any of these into their own issues.

guyoron1:
## Summary

- `rtk rake <any non-test task>` (db:migrate, -T, an aborted run) replaced the entire output with `rake test: no tests ran`: the Minitest state machine never left its header state, so every line was skipped — including `rake aborted!` stanzas, which were often under the tee threshold and therefore unrecoverable. The filter now tracks whether the output actually shows Minitest structure (`# Running:` / `Finished in` / a summary line) and passes it through raw when it does not.
- The `ok rake test: N runs, 0 failures` verdict was built purely from the parsed summary line via `runner::run_filtered`, so a green suite followed by `rake aborted!` + exit 1 still printed `ok`. Switched to `runner::run_filtered_with_exit` (the same shape `cargo_cmd.rs` uses): the `ok` line requires exit 0, and a non-zero exit with nothing failure-shaped parsed degrades to raw output so the real diagnostic survives.
- Any positional test file switched the binary from `rake` to `rails`, which fails with `Bundler::GemNotFound` in plain Ruby gem projects. The swap now additionally requires the conventional `bin/rails` binstub; falling back to `rake` is always safe.

## Test plan

- [x] `cargo test` — 2665 passed, 8 ignored
- [x] New regression tests: non-test task failure passes through (`rake aborted!` / `PG::UndefinedTable` preserved), `rake -T`-style listing passes through, green summary + late abort + exit 1 is not `ok`, non-zero exit with no parsed failures is not `ok`, and positional files without `bin/rails` stay on `rake`.

Fixes #3416

### DIFF
--- a/src/cmds/ruby/rake_cmd.rs
+++ b/src/cmds/ruby/rake_cmd.rs
@@ -18,7 +18,7 @@ const MAX_RAKE_FAILURES: usize = CAP_WARNINGS;
 /// file args. When any positional test file paths are detected, we switch to
 /// `rails test` which handles single files, multiple files, and line-number
 /// syntax (`file.rb:15`) natively.
-fn select_runner(args: &[String]) -> (&'static str, Vec<String>) {
+fn select_runner(args: &[String], rails_available: bool) -> (&'static str, Vec<String>) {
     let has_test_subcommand = args.first().is_some_and(|a| a == "test");
     if !has_test_subcommand {
         return ("rake", args.to_vec());
@@ -32,7 +32,10 @@ fn select_runner(args: &[String]) -> (&'static str, Vec<String>) {
         .filter(|a| looks_like_test_path(a))
         .collect();
 
-    let needs_rails = !positional_files.is_empty();
+    // Only swap to `rails` when the project actually has Rails (#3416):
+    // plain Ruby gems use `rake test` with positional files and have no
+    // `rails` binary, so the swap would fail with Bundler::GemNotFound.
+    let needs_rails = !positional_files.is_empty() && rails_available;
 
     if needs_rails {
         ("rails", args.to_vec())
@@ -50,8 +53,14 @@ fn looks_like_test_path(arg: &str) -> bool {
         || path.contains("_spec.rb")
 }
 
+/// A Rails project is detected by its conventional binstub.
+// ponytail: bin/rails covers the Rails convention; check Gemfile.lock if binstub-less apps show up
+fn rails_available() -> bool {
+    std::path::Path::new("bin/rails").exists()
+}
+
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
-    let (tool, effective_args) = select_runner(args);
+    let (tool, effective_args) = select_runner(args, rails_available());
     let mut cmd = ruby_exec(tool);
     for arg in &effective_args {
         cmd.arg(arg);
@@ -65,7 +74,7 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         );
     }
 
-    runner::run_filtered(
+    runner::run_filtered_with_exit(
         cmd,
         "rake",
         &args.join(" "),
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased on current develop; only rustfmt (style edition 2024) needed touching. I added two tests that pin the parts the original set didn't: a run killed mid-way (raw output kept, no `no tests ran`), and the raw pass-through on a non-zero exit. Each of the three behaviours now has a test that fails when it is reverted (checked by mutation).

Repro from #3416 with `rake` stand-ins: before, `rake test: no tests ran`; after, the `rake aborted!` stanza. I couldn't run a real Ruby/Rake here, so the demos use stand-ins.

PR #3569, opened 2026-08-14
fix(rewrite): never rewrite sudo-prefixed commands

### LINKED ISSUE #3412: The rewrite preserves a `sudo` prefix, so rtk's filtering, SQLite writes and tee writes execute with root privileges the original command did not require for them
`sudo` and `env` prefixes are re-attached in front of the rewritten command, so `sudo <tool>` becomes `sudo rtk <tool>`. The user elevated one program; rtk inserts itself into that elevation.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. The rewrites below were produced by the shipped binary. **This is a privilege-hygiene report, not a vulnerability claim** — see "What I am not claiming".*

### Observed

```
$ rtk rewrite sudo iptables -L     →  sudo rtk iptables -L
$ rtk rewrite sudo du -sh /var     →  sudo rtk du -sh /var
$ rtk rewrite sudo docker ps       →  sudo rtk docker ps
$ rtk rewrite env FOO=1 git status →  env FOO=1 rtk git status
```

`ENV_PREFIX` (`src/discover/registry.rs:60-67`) matches a leading run of `sudo `, `env ` and `VAR=value ` and re-attaches it ahead of the rewrite (`registry.rs:959-977`).

### Why the placement matters

`sudo iptables -L` elevates exactly one binary. `sudo rtk iptables -L` elevates rtk, which then does its own work as root before and after running the real tool:

- opens and writes the SQLite history database
- writes tee files
- loads and parses configuration
- fires the telemetry ping (`main.rs:1578`, on every invocation)
- resolves the target binary through `PATH`

None of that needed root. The user's intent was "run `iptables` as root", not "run a wrapper, its database layer and its config parser as root".

Two concrete consequences worth considering:

1. **File ownership.** Files rtk creates during an elevated run are created by a root process. On a normal desktop that can leave root-owned entries in a user-owned data directory, which the subsequent non-elevated runs then have to write to.
2. **`PATH` resolution under sudo.** rtk resolves the wrapped tool itself rather than letting `sudo` do it. Distributions that set `secure_path` in `sudoers` do so specifically to control which binaries a root invocation can reach; a wrapper that performs its own lookup inside the elevated process is working against that intent.

### What I am not claiming

I checked, and want to be explicit about the negative:

- **The project-filter trust gate still applies.** `src/core/toml_filter.rs:191-211` routes project-local filter files through `hooks::trust::gated_filter_paths()` and `check_trust_with_content`, so the `.rtk/filters.toml` path fixed in GHSA-fvvm-949w-qj4w is **not** re-opened by this. An untrusted repository file does not get parsed as root.
- I have **not** demonstrated privilege escalation, and I do not believe this is one. It is a widening of what runs with elevated privilege, not a way to gain it.

The cleanest fix keeps rtk outside the elevation: rewrite `sudo <tool> …` to `rtk sudo <tool> …` rather than `sudo rtk <tool> …`, so rtk runs as the user and `sudo` elevates only the tool — the same shape the user typed. If that breaks output capture in ways I have not thought through, the alternatives are to skip rewriting `sudo`-prefixed commands entirely (they are rare and usually short-output), or to detect an elevated run and disable the database/tee/telemetry side effects for it.

Note there is currently no root awareness anywhere in the codebase — grepping `src/` for `geteuid`, `getuid` or similar returns nothing — so option three would be new machinery, whereas the first two are changes to existing logic.

### Not checked

The audit machine is Windows, so I verified the **rewrite** but not the runtime consequences: file ownership after an elevated run, `secure_path` interaction, and whether `sudo` is even in a typical rtk user's path for these commands are all unverified. If the elevated-run consequences turn out to be harmless in practice, the rewrite shape is still worth a look, but the priority drops.

guyoron1:
## Summary

- The rewrite re-attached a leading `sudo` in front of the rewritten command (`sudo iptables -L` → `sudo rtk iptables -L`), inserting rtk into the elevation the user granted to exactly one binary. An elevated rtk then writes the SQLite history db and tee files, parses config and fires the telemetry ping as root (leaving root-owned files in user-owned data directories), and resolves the target binary itself instead of letting `sudo`'s `secure_path` do it.
- `rewrite_segment_inner` now returns `None` when the stripped env prefix contains a `sudo` token, so sudo-prefixed commands pass through unrewritten — the same mechanism already used for `RTK_DISABLED=`. Plain `env` / `VAR=val` prefixes still rewrite as before, and the suppression also covers compositions like `sudo noglob git status` and `env FOO=1 sudo …`.
- Chose the pass-through option over rewriting to `rtk sudo <tool>` (no such passthrough exists today) — sudo-prefixed commands are rare and typically short-output, so the token cost of not filtering them is negligible next to running rtk's side effects as root.

## Test plan

- [x] `cargo test` — 2661 passed, 8 ignored
- [x] Updated/added unit tests: `sudo docker ps` → no rewrite, `env FOO=1 sudo iptables -L` → no rewrite, `sudo noglob git status` → no rewrite, `FOO=1 noglob git status` still rewrites.
- [x] Verified with the built binary: `rtk rewrite "sudo docker ps"` exits 1 (no rewrite); `rtk rewrite "docker ps"` and `rtk rewrite "FOO=1 git status"` still rewrite (exit 3).

Fixes #3412

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1311,6 +1311,13 @@ fn rewrite_segment_inner(
             );
             return None;
         }
+        // #3412: never insert rtk inside an elevated invocation. `sudo <tool>`
+        // elevates exactly one binary; rewriting to `sudo rtk <tool>` would run
+        // rtk's SQLite/tee/config/telemetry side effects as root and resolve
+        // the target binary outside sudo's secure_path. Pass through untouched.
+        if env_prefix.split_whitespace().any(|t| t == "sudo") {
+            return None;
+        }
         let rewritten = rewrite_segment_inner(
             rest_after_env,
             excluded,
@@ -4594,11 +4601,18 @@ mod tests {
 
     // --- sudo / env prefix + rewrite ---
 
+    // #3412: sudo-prefixed commands are never rewritten — inserting rtk into
+    // the elevation would run its database/tee/telemetry side effects as root.
     #[test]
-    fn test_rewrite_sudo_docker() {
+    fn test_rewrite_sudo_docker_skipped() {
+        assert_eq!(rewrite_command_no_prefixes("sudo docker ps", &[]), None);
+    }
+
+    #[test]
+    fn test_rewrite_sudo_with_env_assign_skipped() {
         assert_eq!(
-            rewrite_command_no_prefixes("sudo docker ps", &[]),
-            Some("sudo rtk docker ps".into())
+            rewrite_command_no_prefixes("env FOO=1 sudo iptables -L", &[]),
+            None
         );
     }
 
@@ -5147,9 +5161,16 @@ mod tests {
 
     #[test]
     fn test_env_prefix_composed_with_builtin() {
+        // #3412: the sudo prefix suppresses the rewrite even through a
+        // builtin transparent prefix like noglob.
         assert_eq!(
             rewrite_command_no_prefixes("sudo noglob git status", &[]),
-            Some("sudo noglob rtk git status".into())
+            None
+        );
+        // A plain env-assignment prefix still composes with builtins.
+        assert_eq!(
+            rewrite_command_no_prefixes("FOO=1 noglob git status", &[]),
+            Some("FOO=1 noglob rtk git status".into())
         );
     }
 


KuSh:
Thank you for this, and sorry it took a while to get a decision. Your analysis of the problem is correct and well argued — the comment you wrote (`sudo <tool>` elevates exactly one binary; `sudo rtk <tool>` would run rtk's SQLite/tee/telemetry side effects as root and resolve the target outside `secure_path`) is precisely the reasoning we settled on, and it's now the rationale carried in the code. The `env FOO=1 sudo iptables -L` and `sudo noglob git status` cases you identified are the sharpest ones in this area, and they were genuinely not covered before you raised them.

I'm closing this in favour of **#2810**, which reaches the same passthrough behaviour but at a different point in the pipeline — and that difference matters for what comes next.

**Why the location matters.** This PR guards inside `rewrite_segment_inner`, after the env prefix has already been peeled, and leaves `sudo\s+` in the `ENV_PREFIX` regex (`registry.rs:61-69`). #2810 instead removes `sudo\s+` from `ENV_PREFIX` itself.

Behaviourally the two are equivalent today. They diverge on #146 (`rtk sudo`), which is the direction this is headed: rather than passing `sudo` through forever, `rtk sudo docker ps` will filter the inner command while keeping rtk itself *outside* the elevation — the child rtk runs unprivileged and hands the wrapper down as an internal transport, so none of the side effects you listed ever execute as root.

That design needs a `^sudo(\s|$)` rule in `rules.rs`. Rules are matched **after** `ENV_PREFIX` strips the prefix, so as long as `sudo\s+` stays in that regex, the leading `sudo` is gone before any rule is consulted and a `^sudo` rule can never fire — it would be dead code. Removing it from `ENV_PREFIX` is the prerequisite; a guard placed downstream is not. The same applies to `classify_command`, which needs to see `sudo` as a leading word for `rtk discover` to categorise it.

**Nothing you raised is being dropped.** All four of your motivating cases — `FOO=1 sudo …`, `env FOO=1 sudo …`, bare `sudo`, and `sudoedit …` — pass through correctly under #2810 (I verified each against a build of that branch), but they were pinned by nothing there either. I've asked for exactly those assertions to be added to #2810's `test_rewrite_sudo_passthrough` before it merges, so the coverage your PR contributed survives the swap. `sudo noglob git status` is covered too, in its own test.

Two follow-ups also came out of this area, both filed: #3756 and #3757.

Closing as superseded rather than rejected — the fix is landing, just one layer up. Thanks again for the report and the diagnosis.

Issue #3282, opened 2026-07-28
RTK init want to install in .claude when using another agent


pierre-skyvisor:
When I run `rtk init --global --agent cursor`, I have this error.

```
rtk: Failed to write RTK.md: /home/.../.claude/RTK.md: Failed to create temp file in /home/.../.claude: No such file or directory (os error 2) at path "/home/.../.claude/.tmp8bS3b2"
```

I am using WSL and RTK v0.44.1

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 88% |

### Summary

rtk init --agent cursor still tries to create RTK.md inside a nonexistent .claude directory instead of the target agent's directory; fails with ENOENT on WSL (v0.44.1)

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

- `src/init.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

KuSh:
Closing as a duplicate of #2097, which tracks the same problem with the more complete discussion. Please follow up there; if your case differs in a way that issue does not cover, comment here and it can be reopened.

PR #3570, opened 2026-08-14
fix(init): migrate Cursor hooks.json before deleting the legacy hook script

### LINKED ISSUE #3465: rtk init --agent cursor deletes ~/.cursor/hooks/rtk-rewrite.sh, leaves hooks.json referencing it, and reports success
`rtk init --global --agent cursor` deletes `~/.cursor/hooks/rtk-rewrite.sh`, leaves `~/.cursor/hooks.json` still referencing it, and prints "Cursor hook registered (global)".

**rtk 0.44.2, Windows 11 x86_64, Git Bash.** Reproduced twice.

The net result is a registered Cursor `preToolUse` hook whose script does not exist. Cursor hooks that cannot start are a denial path, so this can break shell tool calls until the user notices the dangling reference — and nothing in the command output suggests anything went wrong.

### Repro

Starting state — the shell-script hook installed by an earlier rtk version, registered in `~/.cursor/hooks.json`:

```
$ ls -l ~/.cursor/hooks/rtk-rewrite.sh
-rw-r--r-- 644  2.2K  rtk-rewrite.sh

$ cat ~/.cursor/hooks.json
{
  "version": 1,
  "hooks": {
    "preToolUse": [
      { "command": "./hooks/rtk-rewrite.sh", "matcher": "Shell" }
    ]
  }
}
```

Run:

```
$ rtk init -g --agent cursor

RTK hook registered (global).

  Command:   rtk hook claude
  RTK.md:    C:\Users\aaron\.claude\RTK.md (10 lines)

Cursor hook registered (global).

  Command:    rtk hook cursor
```

After:

```
$ ls -la ~/.cursor/hooks/
(empty)

$ cat ~/.cursor/hooks.json
{
  "version": 1,
  "hooks": {
    "preToolUse": [
      { "command": "./hooks/rtk-rewrite.sh", "matcher": "Shell" }
    ]
  }
}
```

`grep -rl "rtk" ~/.cursor --include="*.json"` finds no config anywhere containing `rtk hook cursor`.

### Three separate problems

**1. The script is deleted but the registration is not updated.** Migrating from the shell-script hook to the native `rtk hook cursor` path is reasonable, but `hooks.json` must be rewritten to invoke `rtk hook cursor` in the same operation. Deleting the script while leaving `"command": "./hooks/rtk-rewrite.sh"` in place produces a broken hook.

**2. "Cursor hook registered (global)" is a false success message.** Nothing was registered. The command exits 0. A user has no signal that their hook is now broken.

**3. `--agent cursor` still runs the Claude installer first**, writing `~/.claude/RTK.md` and modifying `~/.claude/CLAUDE.md`. On my first run this replaced a 133-line user-authored CLAUDE.md with a 3-line `@RTK.md` reference; the generated RTK.md then ends with "Refer to CLAUDE.md for full command reference", pointing at content the same migration had just removed. This is the Windows/`.claude`-exists counterpart of #3282, which reports the same `--agent cursor` → `.claude` targeting but surfaces as a write error on a system with no `.claude` directory. There it fails loudly; here it succeeds and silently rewrites user config the user did not ask it to touch.

### Suggested fix

- When removing `~/.cursor/hooks/rtk-rewrite.sh`, rewrite `~/.cursor/hooks.json` to `"command": "rtk hook cursor"` in the same operation, or leave the script in place.
- Only print "Cursor hook registered" after verifying the registration was actually written.
- Do not run the Claude Code installer or touch `~/.claude/` when `--agent cursor` is passed.

I had backups and restored both files. A user without them loses their global CLAUDE.md content and gets a silently broken Cursor hook.

guyoron1:
## Summary

- `rtk init -g --agent cursor` deleted `~/.cursor/hooks/rtk-rewrite.sh` first and updated `hooks.json` in a separate, error-swallowed pass; `cursor_hook_already_present` then accepted the stale legacy entry as "RTK hook already present". Net result on any failure of the cleanup pass: a registered Cursor preToolUse hook whose script no longer exists, reported as success (#3465).
- `patch_cursor_hooks_json` now strips legacy `rtk-rewrite.sh` entries and registers `rtk hook cursor` in the same atomic write, and the presence check only accepts the new binary command — a legacy entry can never satisfy it, since its script is about to be deleted.
- The legacy script is deleted only after `hooks.json` was successfully patched, so a parse failure propagates and leaves the still-working script hook in place. The now-redundant `remove_legacy_cursor_hooks_json_entries` pass is removed.
- The third problem in the issue (`--agent cursor` also running the Claude installer and touching `~/.claude/`) is the scope of #3290 / #3282 and is not changed here.

## Test plan

- [x] `cargo test` — 2661 passed, 8 ignored
- [x] New regression test `test_patch_cursor_hooks_json_migrates_legacy_entry`: a hooks.json referencing the legacy script is rewritten to a single `rtk hook cursor` entry in one operation. Updated `test_cursor_hook_already_present_legacy_script*` to assert a legacy entry does not count as present.

Fixes #3465

### DIFF
--- a/src/hooks/init.rs
+++ b/src/hooks/init.rs
@@ -3567,6 +3567,14 @@ fn install_cursor_hooks(ctx: InitContext) -> Result<()> {
     let InitContext { verbose, dry_run } = ctx;
     let cursor_dir = resolve_cursor_dir()?;
 
+    // Register the binary command in hooks.json FIRST (this also drops any
+    // legacy rtk-rewrite.sh entries in the same write). Only once that has
+    // succeeded is it safe to delete the legacy script: deleting the script
+    // while hooks.json still references it leaves Cursor with a registered
+    // hook whose command does not exist (#3465).
+    let hooks_json_path = cursor_dir.join(HOOKS_JSON);
+    let patched = patch_cursor_hooks_json(&hooks_json_path, ctx)?;
+
     // Migrate old hook script if present
     let old_hook = cursor_dir.join("hooks").join(REWRITE_HOOK_FILE);
     if old_hook.exists() {
@@ -3584,18 +3592,7 @@ fn install_cursor_hooks(ctx: InitContext) -> Result<()> {
                 );
             }
         }
-        // Clean stale hooks.json entry pointing to the deleted script
-        let hooks_json_path = cursor_dir.join(HOOKS_JSON);
-        if let Err(e) = remove_legacy_cursor_hooks_json_entries(&hooks_json_path, ctx) {
-            if verbose > 0 {
-                eprintln!("  [warn] Failed to clean legacy Cursor hooks.json entry: {e}");
-            }
-        }
     }
-
-    // Create or patch hooks.json with binary command
-    let hooks_json_path = cursor_dir.join(HOOKS_JSON);
-    let patched = patch_cursor_hooks_json(&hooks_json_path, ctx)?;
 
     // Report (skip in dry-run)
     if !dry_run {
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased on current develop; two conflicts in `src/hooks/init.rs`, both the same shape: develop kept `remove_legacy_cursor_hooks_json_entries` and ported it to the new JSON I/O helpers, whereas this PR deletes it because the migration now happens in the same atomic write, so I took the PR side.

The bug is worse on develop than when I wrote this up: an earlier cleanup step ("Remove Cursor legacy hook") deletes the script before `install_cursor_hooks` runs, so the migration block never fires at all. `hooks.json` keeps pointing at the deleted script while rtk prints "RTK preToolUse entry already present". Before/after with the real binaries is in the description.

PR #3571, opened 2026-08-14
fix(prettier): read check results from stderr, never report a failing check as success

guyoron1:
## Summary
- Prettier >= 1.19 writes its `--check` results to **stderr** as `[warn] <file>` lines, but the filter ran with `RunOptions::stdout_only()` — so it never saw the file list — and additionally skipped any line containing `[warn]`
- On a failing check the list of files to fix was silently discarded, and the filter's check-mode branch produced `Prettier: All files formatted correctly` next to exit code 1 (truth inversion; in practice the `never_worse` guard often reduced the output to just `Checking formatting...` — still zero signal)
- Fix: filter the combined stdout+stderr stream, recognize `[warn] <file>` lines, and add an exit-aware guard — a check that exited non-zero with no parsable file list passes the real output through instead of claiming success

## Reproduction
```
$ printf 'const  x=1\n' > bad.js
$ npx prettier@3 --check bad.js        # stdout: 'Checking formatting...' / stderr: '[warn] bad.js', exit 1
$ rtk prettier --check bad.js          # before: 'Checking formatting...' (file list lost), exit 1
$ rtk prettier --check bad.js          # after:  'Prettier: 1 files need formatting / 1. bad.js', exit 1
```

## Test plan
- [x] Added regression tests (`test_filter_check_failure_warn_prefixed_files`, `test_filter_check_failure_never_reports_success`, `test_filter_check_success_exit_zero_unchanged`)
- [x] Verified end-to-end against real prettier 3.x output
- [x] `cargo fmt && cargo clippy --all-targets && cargo test` passes (2663 passed)

### DIFF
--- a/src/cmds/js/prettier_cmd.rs
+++ b/src/cmds/js/prettier_cmd.rs
@@ -16,17 +16,24 @@ pub fn run(args: &[String], verbose: u8) -> Result<i32> {
         eprintln!("Running: prettier {}", args.join(" "));
     }
 
-    runner::run_filtered(
+    // Prettier >= 1.19 logs its `--check` results ("[warn] <file>") to stderr,
+    // not stdout — the filter must see the combined output, and the exit code,
+    // to avoid reporting a failing check as success.
+    runner::run_filtered_with_exit(
         cmd,
         "prettier",
         &args.join(" "),
-        filter_prettier_output,
-        RunOptions::stdout_only(),
+        filter_prettier_output_with_exit,
+        RunOptions::default(),
     )
 }
 
 /// Filter Prettier output - show only files that need formatting
 pub fn filter_prettier_output(output: &str) -> String {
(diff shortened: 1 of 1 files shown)


guyoron1:
Reworked after the stdout-only stderr forwarding landed: `rtk prettier` still filters stdout only and forwards stderr (`tests/stderr_only_failure_test.rs` pins that), so this no longer merges the two streams. What still reproduces with prettier 3.9.8: a syntax error or a non-matching pattern (exit 2) prints "All files formatted correctly", and `rtk format prettier --check` reports a failing check as formatted and drops the diagnostics.

The filter now takes the exit code and refuses to summarise a failed run, and `rtk format` passes it. I've retitled the PR accordingly. This overlaps #2410, which currently fails `tests/stderr_only_failure_test.rs` and doesn't cover `rtk format`; glad to coordinate on whichever the maintainers prefer.

PR #3572, opened 2026-08-14
fix(playwright): consume the value of a space-separated --reporter flag

guyoron1:
## Summary
- `rtk playwright test` forces `--reporter=json` and strips the user's own reporter flag — but for the space-separated form (`--reporter list`) it removed only the `--reporter` token and kept `list`
- Playwright treats the leftover bare word as a positional test filter, so the run silently executes the wrong (usually empty) test set: 0 tests run, exit 1, "No tests found matching …"
- Fix: strip both `--reporter=<name>` and `--reporter <name>` (flag **and** its value) before injecting the forced JSON reporter

## Reproduction
```
# project with one passing spec
$ npx playwright test --reporter=json         # stats.expected = 1, exit 0
$ npx playwright test --reporter=json list    # what rtk used to run for 'rtk playwright test --reporter list'
                                              # stats.expected = 0, exit 1 (no tests match 'list')
```

## Test plan
- [x] Added regression tests (`test_strip_reporter_space_form_consumes_value`, `test_strip_reporter_equals_form`, `test_strip_reporter_keeps_unrelated_args`, `test_strip_reporter_trailing_flag_without_value`)
- [x] Verified against a real Playwright project (leaked positional runs 0 of 1 tests and exits 1)
- [x] `cargo fmt && cargo clippy --all-targets && cargo test` passes (2664 passed)

### DIFF
--- a/src/cmds/js/playwright_cmd.rs
+++ b/src/cmds/js/playwright_cmd.rs
@@ -235,6 +235,31 @@ fn extract_failures_regex(output: &str) -> Vec<TestFailure> {
     failures
 }
 
+/// Strip the user's `--reporter` flag so it can't conflict with the injected
+/// `--reporter=json`. Handles both `--reporter=<name>` and the space-separated
+/// `--reporter <name>` form — the latter must also consume the value, otherwise
+/// the bare reporter name (`list`, `dot`, ...) is left behind as a positional
+/// test-filter and Playwright silently runs the wrong (usually empty) test set.
+fn strip_reporter_args(args: &[String]) -> Vec<String> {
+    let mut out = Vec::new();
+    let mut skip_value = false;
+    for arg in args {
+        if skip_value {
+            skip_value = false;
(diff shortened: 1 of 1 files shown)


guyoron1:
Rebased on current develop, no conflicts. I added a test that pins the argv Playwright receives for `--reporter list` (the value no longer survives as a test filter); the unit tests from the original PR call a helper that doesn't exist on develop, so on their own they couldn't show the bug.

Heads-up on overlap: #3060 fixes the same leak differently, by leaving an explicit reporter untouched, like vitest and deno already do. I built #3060 on develop and ran it against a real Playwright 1.63 project: `--reporter list` prints the native list output, `--reporter=dot` the dot output, no flag still gives the compact summary, and exit codes are preserved. This PR keeps the forced-JSON summary and only stops the leak. Happy to close it in favour of #3060 if that's the direction you prefer.

PR #3573, opened 2026-08-14
fix(curl): use -sS so failures keep curl's error message

guyoron1:
## Summary
- `rtk curl` injects a bare `-s`, which silences the progress bar but **also suppresses curl's error messages**
- Any failure — DNS, connection refused, TLS — printed only `FAILED: curl ` with no reason, forcing the agent to rerun the command outside rtk to learn what went wrong
- Fix: inject `-sS` — `-S` re-enables error messages while `-s` keeps progress output off

## Reproduction
```
$ rtk curl http://nonexistent.invalid/
FAILED: curl                                          # before: no reason, exit 6
FAILED: curl curl: (6) Could not resolve host: …      # after
```

## Test plan
- [x] Added regression test (`curl_failure_surfaces_reason`, offline-safe: connects to 127.0.0.1:1)
- [x] Verified the test fails on the previous `-s`-only behavior and passes with `-sS`
- [x] `cargo fmt && cargo clippy --all-targets && cargo test` passes (2661 passed)

### DIFF
--- a/src/cmds/cloud/curl_cmd.rs
+++ b/src/cmds/cloud/curl_cmd.rs
@@ -23,14 +23,16 @@ const MAX_RESPONSE_SIZE: usize = 500;
 pub fn run(args: &[String], verbose: u8) -> Result<i32> {
     let timer = tracking::TimedExecution::start();
     let mut cmd = resolved_command("curl");
-    cmd.arg("-s"); // Silent mode (no progress bar)
+    // -s silences the progress bar but ALSO suppresses error messages; -S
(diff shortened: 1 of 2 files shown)


guyoron1:
Rebased on current develop. Re-checking with real curl, I found that `-sS` alone made the FAILED line drop the response body for `--fail-with-body` (stderr won over stdout), so it now prints curl's reason and then the body. Tests: real curl on a closed port for the reason, and a stand-in curl for the `--fail-with-body` case.

It merges cleanly together with #2657 in either order.

PR #3575, opened 2026-08-15
fix(git): don't misdetect a value-taking option's argument as a patch  flag

KuSh:
## Summary

Follow-up on #2951.

`requests_raw_log_output()` matched bare `-p`/`-u`/`--patch` tokens anywhere in the args, without checking whether the previous token was an option like `--grep` or `-S` that consumes the next token as its value. `git log --grep -p` searches commit messages for the literal string `-p` (verified against git 2.53.0: no diff output, same as `--grep=-p`), but RTK treated it as a patch request and skipped its own filtering/limit, dumping raw uncapped `git log` output for what is actually a plain grep search.

Skip the value token after any known value-taking `git log`/`git diff` option before checking for the patch flags.

## Test plan
- [x] Manual testing: `rtk git log --grep -p` output inspected

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -534,6 +534,8 @@ fn consumes_next_token_as_value(arg: &str) -> bool {
             | "--date"
             | "--decorate-refs"
             | "--decorate-refs-exclude"
+            | "--diff-algorithm"
+            | "--diff-filter"
             | "--diff-merges"
             | "--dst-prefix"
             | "--encoding"
@@ -2875,7 +2877,15 @@ A  added.rs
     fn test_patch_flag_as_value_of_grep_is_not_misdetected() {
         // `git log --grep -p` searches commit messages for the literal
         // string "-p"; git does not treat it as the patch flag.
-        for opt in ["--grep", "--author", "--committer", "-S", "-G"] {
+        for opt in [
+            "--author",
+            "--committer",
+            "--diff-algorithm",
+            "--diff-filter",
+            "--grep",
+            "-G",
+            "-S",
+        ] {
             let args = vec![opt.to_string(), "-p".to_string()];
             assert!(
                 !requests_raw_log_output(&args),
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -431,6 +431,12 @@ fn run_log(
     verbose: u8,
     global_args: &[String],
 ) -> Result<i32> {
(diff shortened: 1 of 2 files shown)


src/cmds/git/git.rs
```diff
@@ -517,15 +517,80 @@ fn run_log(
...
+            | "--color-moved-ws"
+            | "--committer"
+            | "--date"
+            | "--decorate-refs"
+            | "--decorate-refs-exclude"
+            | "--diff-merges"
+            | "--dst-prefix"
+            | "--encoding"
+            | "--exclude"
+            | "--expand-tabs"
```
aeppling: git log --expand-tabs 4 fails with fatal: ambiguous argument '4', the value does not seems to be a separate token. So in git log --expand-tabs -p, -p is a real patch flag

src/cmds/git/git.rs
```diff
@@ -517,15 +517,80 @@ fn run_log(
...
+            | "--rotate-to"
+            | "--since"
+            | "--since-as-filter"
+            | "--skip"
+            | "--skip-to"
+            | "--src-prefix"
+            | "--stat-count"
+            | "--stat-name-width"
+            | "--stat-width"
+            | "--unified"
```
aeppling: same as expand-tabs

src/cmds/git/git.rs
```diff
@@ -517,15 +517,80 @@ fn run_log(
...
+            | "--until"
+            | "--word-diff-regex"
+            | "--ws-error-highlight"
+            | "-G"
+            | "-I"
+            | "-L"
+            | "-O"
+            | "-S"
+            | "-U"
```
aeppling: same

aeppling: then we can add a test to avoid regression on those ( eg. "git log -U -p" still work as expected)

aeppling (changes requested):

KuSh:
@aeppling removed the three problematic args and added a test to make sure they don’t get added back. Two more problematic args were removed as well: `--max-parents` and `--min-parents`, and two missed args were added to the list: `--diff-algorithm` and `--diff-filter`. Checked against git 2.53.0.

KuSh:
The same issue was also present in `has_limit_flag`, `has_format_flag`, `wants_merges`, and `parse_user_limit`, which scanned args without accounting for values or the `--` pathspec separator. That’s now fixed.

aeppling:
@KuSh LGTM :)

aeppling (approved):

PR #3585, opened 2026-08-16
fix(benchmark): compare find --max against full find, not head -N

### LINKED ISSUE #3584: benchmark job fails on spurious `find --max 10` negative (head -N truncates the baseline)
The `benchmark` CI job fails on `develop` because of one spurious negative in `scripts/benchmark.sh`:

```
🔴 find --max 10   find . -not -path ./'target/*' -not -path ./'.git/*' -type f | head -10   rtk find '*' --max 10   40 → 48 (-20%)
```

A single negative makes the whole script `exit 1`, so the job is red on every push/run.

### Root cause

`rtk find --max N` does **not** reduce scanning — it walks and summarizes the *whole* tree and only caps how many filenames are **displayed**:

```
356F 63D:
./ CHANGELOG.md CLAUDE.md ... README_es.md
+346 more
ext: .rs(130) .md(72) .toml(64) .txt(33) .sh(17)
```

The benchmark baseline pipes through `head -10`, which early-terminates the raw scan to 10 truncated paths (~174 chars). That is a *different operation* — a truncated slice, not the full listing rtk compacts. On a repo this size the 10-path slice is smaller than rtk’s summary header (~191 chars), so `--max 10` is flagged negative even though rtk’s own `never_worse` guard confirms it is smaller than the full `find` output it actually replaces.

`--max 100` only passes by luck (head -100 is large enough).

### Fix

Compare `rtk find --max N` against the full `find` output a user would otherwise read (same `target/`/`.git` exclusions), without `head -N`. Both `--max` rows then compare like-for-like and are stable across repo sizes:

```
✅ find --max 10    3500 → 48  (+98%)
✅ find --max 100   3533 → 466 (+86%)
```

Local replay after the fix: `✅ 39 good  ⚠️ 23 warn  🔴 0 negative  ❌ 0 fail  → exit 0`.

PR incoming.

ousamabenyounes:
Fixes #3584.

The `benchmark` CI job fails on a single spurious negative that trips the whole `exit 1`:

```
🔴 find --max 10   … | head -10   rtk find '*' --max 10   40 → 48 (-20%)
```

### Why it is spurious

`rtk find --max N` caps how many names are **displayed** but still scans and summarizes the whole tree (counts, `+N more`, extension histogram). The baseline piped the raw scan through `head -N`, truncating it to a *different, early-terminated* operation. On a repo this size `head -10` (~174 chars) is smaller than rtk’s summary header (~191 chars), so `--max 10` is marked negative — even though rtk’s own `never_worse` guard already proves the output is smaller than the full `find` it replaces. `--max 100` passed only because `head -100` happened to be large enough.

### Fix

Drop `head -N`: compare `rtk find --max N` against the full `find` a user would otherwise read (same `target/`/`.git` exclusions). Both rows now compare like-for-like and are stable across repo sizes.

Local replay against a `--release` build.

**RED — upstream `develop` (`scripts/benchmark.sh` unchanged):**
```
🔴 find --max 10   find … | head -10   rtk find '*' --max 10   40 → 48 (-20%)
  🔴 1 negative   BENCHMARK FAILED: 1 filter(s) produced more tokens than raw output   (exit 1)
```

**GREEN — with this patch:**
```
✅ find --max 10    find … (full)   rtk find '*' --max 10    3500 → 48  (+98%)
✅ find --max 100   find … (full)   rtk find '*' --max 100   3533 → 466 (+86%)
  ✅ 39 good  ⚠️ 23 warn  🔴 0 negative  ❌ 0 fail   (exit 0)
```

Docs-free, benchmark-only change; no rtk behavior is modified.

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -190,8 +190,12 @@ bench "read -n" "cat -n src/main.rs" "$RTK read src/main.rs -n"
 section "find"
 bench "find *" "find . -type f" "$RTK find '*'"
 bench "find *.rs" "find . -name '*.rs' -type f" "$RTK find '*.rs'"
-bench "find --max 10" "find . -not -path './target/*' -not -path './.git/*' -type f | head -10" "$RTK find '*' --max 10"
-bench "find --max 100" "find . -not -path './target/*' -not -path './.git/*' -type f | head -100" "$RTK find '*' --max 100"
+# `rtk find --max N` caps how many names are DISPLAYED but still scans and
+# summarizes the whole tree, so the honest baseline is the full `find` a user
+# would otherwise read — not a `head -N`-truncated slice, which measures a
+# different (early-terminated) operation and made `--max 10` spuriously negative.
+bench "find --max 10" "find . -not -path './target/*' -not -path './.git/*' -type f" "$RTK find '*' --max 10"
+bench "find --max 100" "find . -not -path './target/*' -not -path './.git/*' -type f" "$RTK find '*' --max 100"
 
 # ===================
 # git
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -192,10 +192,64 @@ bench "find *" "find . -type f" "$RTK find '*'"
 bench "find *.rs" "find . -name '*.rs' -type f" "$RTK find '*.rs'"
 # `rtk find --max N` caps how many names are DISPLAYED but still scans and
 # summarizes the whole tree, so the honest baseline is the full `find` a user
-# would otherwise read — not a `head -N`-truncated slice, which measures a
-# different (early-terminated) operation and made `--max 10` spuriously negative.
-bench "find --max 10" "find . -not -path './target/*' -not -path './.git/*' -type f" "$RTK find '*' --max 10"
+# would otherwise read, not a `head -N`-truncated slice — that measures a
+# different, early-terminated operation.
+# A single row carries the savings: the baseline is identical for every N, so a
+# second row would only add the same token count to the totals twice.
 bench "find --max 100" "find . -not -path './target/*' -not -path './.git/*' -type f" "$RTK find '*' --max 100"
+
+# That baseline no longer scales with N, so it cannot detect a --max that stops
+# limiting. Assert the cap directly instead: `rtk find --max N` displays exactly
+# min(N, total) names, so a --max that is ignored (falling back to the default
+# cap) fails whether N sits below or above that default.
+count_find_names() {
+  awk '
+    /^[0-9]+F [0-9]+D:$/ { next }   # "356F 63D:" summary header
+    /^\+[0-9]+ more$/    { next }   # "+256 more" truncation marker
+    /^ext: /             { next }   # extension histogram footer
+    NF == 0              { next }
+    { n += NF - ($1 ~ /\/$/ ? 1 : 0) }   # grouped lines lead with "dir/"
+    END { print n + 0 }
+  '
+}
+
+# Total tree size as rtk itself reports it: the "NNNF" header when it groups,
+# otherwise the displayed names plus whatever "+N more" hides.
+count_find_total() {
+  awk '
+    /^[0-9]+F [0-9]+D:$/ { total = $1 + 0; next }
+    /^\+[0-9]+ more$/    { more = substr($1, 2) + 0; next }
+    /^ext: /             { next }
+    NF == 0              { next }
+    { shown += NF - ($1 ~ /\/$/ ? 1 : 0) }
+    END { print (total ? total : shown + more) + 0 }
+  '
+}
+
+bench_find_max() {
+  local max="$1"
+  local out displayed total expected
+
+  out=$(eval "$RTK find '*' --max $max" 2>/dev/null || true)
+  displayed=$(printf '%s\n' "$out" | count_find_names)
+  total=$(printf '%s\n' "$out" | count_find_total)
+  expected=$([ "$max" -lt "$total" ] && echo "$max" || echo "$total")
+
+  TOTAL_TESTS=$((TOTAL_TESTS + 1))
+
+  if [ "$displayed" -eq "$expected" ] && [ "$expected" -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


mhcoen:
Independent confirmation, plus one more piece of evidence that the baseline is the wrong
comparison rather than the filter being too fat: rtk never rewrites the piped form, so no
agent ever pays those 48 tokens.

On current `develop`:

```
$ rtk rewrite 'find . -type f | head -10'
(no output, exit 1)

$ rtk rewrite "find . -not -path './target/*' -type f | head -10"
(no output, exit 1)

$ rtk rewrite 'find . -type f'
rtk find . -type f
```

The baseline in `bench "find --max 10"` is a command rtk leaves untouched. That pipeline runs
raw and costs its 40 tokens whether or not rtk is installed. The 48-token summary only ever
appears in place of a full `find . -type f`, which is exactly what this PR compares against.

The numbers reproduce here as well: `rtk find '*' --max 10` is 191 chars, so 48 tokens under
the script's `chars/4` estimate, of which the extension histogram is 49 chars, or 12 tokens.
Suppressing that histogram on small result sets would put the case under the baseline too,
but that trades real output quality for a measurement artifact, so fixing the baseline looks
like the right call.

Worth noting for prioritization: one negative fails the whole `benchmark` job, so every open
PR against `develop` currently shows a red CI badge for a reason unrelated to its own diff.
The develop runs 3820, 3882, 3934, 3936 and 3938 are all red on this single case.

scripts/benchmark.sh
```diff
@@ -190,8 +190,12 @@ bench "read -n" "cat -n src/main.rs" "$RTK read src/main.rs -n"
 section "find"
 bench "find *" "find . -type f" "$RTK find '*'"
 bench "find *.rs" "find . -name '*.rs' -type f" "$RTK find '*.rs'"
-bench "find --max 10" "find . -not -path './target/*' -not -path './.git/*' -type f | head -10" "$RTK find '*' --max 10"
-bench "find --max 100" "find . -not -path './target/*' -not -path './.git/*' -type f | head -100" "$RTK find '*' --max 100"
+# `rtk find --max N` caps how many names are DISPLAYED but still scans and
+# summarizes the whole tree, so the honest baseline is the full `find` a user
+# would otherwise read — not a `head -N`-truncated slice, which measures a
+# different (early-terminated) operation and made `--max 10` spuriously negative.
+bench "find --max 10" "find . -not -path './target/*' -not -path './.git/*' -type f" "$RTK find '*' --max 10"
```
KuSh: Non-blocking, but worth flagging: both `--max` rows now run the identical `unix_cmd`, so the baseline no longer scales with `N` and the two rows differ only in rtk's own output size. Neither can detect a `--max` regression any more.

Concretely — if `--max` became a no-op (say the `displayed >= max_results` break in `find_cmd.rs` were dropped), `rtk find '*' --max 10` would print the full grouped listing: ~1487 tokens against the ~3041-token baseline = 51% savings → `⚠️ WARN`, `NEGATIVE_TESTS` stays 0, benchmark exits 0. Under the old `head -N` baselines that same regression measured 1487 vs 40 tokens → `🔴 NEG` → `exit 1`.

The savings measurement itself is now honest and that's the point of the PR, so this isn't a reason to hold it. But it'd be nice to keep coverage for `--max` itself — e.g. a cheap assertion alongside the bench that `rtk find '*' --max 10` prints no more than ~10 file names. The exact-match `bench_rewrite`-style helper at the bottom of this script would fit.

scripts/benchmark.sh
```diff
@@ -190,8 +190,12 @@ bench "read -n" "cat -n src/main.rs" "$RTK read src/main.rs -n"
...
+bench "find --max 100" "find . -not -path './target/*' -not -path './.git/*' -type f" "$RTK find '*' --max 100"
```
KuSh, line 199: FYI / cosmetic: now that both rows share the same `unix_cmd`, the same ~3041-token baseline gets added to `TOTAL_UNIX` twice, so the headline `Tokens: X → Y` line double-counts it.

Immaterial in practice — the `find *` row alone contributes ~530k tokens because `find . -type f` includes `target/` — so no need to change anything unless you want the totals to be exact.

KuSh (approved):
Approving — the root cause is right. `rtk find --max N` (`run()` in `src/cmds/system/find_cmd.rs`) walks and summarizes the *whole* tree: it emits the `{total_files}F {dirs}D:` header and the `+N more` line regardless of `max_results`, which only caps how many names get **displayed**. So `head -N` was benchmarking a different, early-terminated operation.

Verified locally against `target/release/rtk`: the full `find` baseline is ~3041 tokens, vs ~32 for `--max 10` and ~466 for `--max 100`. Under the old baseline `--max 10` scored 32 vs 40 tokens — one file added to the repo flips that row negative, which matches the flakiness this PR describes.

Benchmark-only change, no runtime behavior touched, and no stale references to the old baseline elsewhere in the repo. Two non-blocking notes inline.

scripts/benchmark.sh
KuSh: Addressed in 9b16854 — I pushed the fix rather than leaving it to you, since it was my comment.

`bench_find_max` now asserts the cap directly: `rtk find --max N` displays exactly `min(N, total)` names. I checked the invariant holds exactly for N = 1, 3, 5, 10, 17, 37, 63, 99, 100, 200, 355, 356 and 400 (the last clamping to 356), across both output shapes — the flat one-name-per-line list at small N and the grouped `356F 63D:` listing at larger N.

Worth noting why it asserts equality rather than `<= N`: with `--max` stubbed out entirely, `rtk find` falls back to its own default cap of 56 names here, so a `<= N` rule would pass vacuously for `--max 100`. The equality rule catches the no-op on both rows (56 != 10, 56 != 100) and fails the benchmark via `FAIL_TESTS`.

scripts/benchmark.sh
KuSh: Also addressed in 9b16854. Since the baseline is identical for every N, the two rows were measuring the same scenario, so I kept a single savings row (`--max 100`, the more conservative of the two) and moved `--max 10` into the cap assertion above.

`TOTAL_UNIX` for the find section is now 1107528 + 6138 + 358641 — three distinct baselines, each counted once.

PR #3590, opened 2026-08-17
fix(core): run the caller's binary when RTK_BIN names it

mhcoen:
## Summary

- Add `rtk_bin_override()`, consulted by `resolved_command()` and `tool_exists()`, so a tool reached through an explicit path is the one rtk runs
- Honored only when the `RTK_BIN` basename equals the tool being resolved, so a stale value can never redirect an unrelated binary
- Inert as it stands: nothing in rtk sets `RTK_BIN` yet, so this lands as plumbing with no behavior change

Part 1 of 3 for #1053.

`resolved_command()` and `tool_exists()` always re-resolve a tool through PATH. That is correct when the agent typed `pytest`, and wrong when it typed `.venv/bin/pytest`: rtk runs whichever pytest PATH happens to find, which is usually a different interpreter than the one the user asked for, or none at all. This adds the lookup point; part 2 is what starts setting the variable.

`tool_exists()` additionally requires the path to be on disk, since handlers branch on availability and a virtualenv or `node_modules/.bin` binary is normally absent from PATH.

Security note: the override cannot broaden what rtk can execute. It only selects a different path for the tool rtk was already about to run, and anyone able to set `RTK_BIN` in rtk's environment can already set `PATH`.

## Test plan

- [x] `test_rtk_bin_override_targets_only_the_named_tool` covers a matching basename winning, a different tool being unaffected, and a path that is not on disk not counting as present. The environment is restored before any assertion, so a failure cannot leak `RTK_BIN` into the rest of the suite.
- [x] Full suite: 2600 passed on `develop`, 2601 with this branch, with the same 9 failures in both (`cmds::cloud::curl_cmd`, `core::tracking`), which fail on unmodified `develop` in my environment.

Follow-ups in this series: part 2 makes the rewrite emit `RTK_BIN` for path-prefixed commands, part 3 fixes `exclude_commands` matching for the same invocation forms.

### DIFF
--- a/src/core/utils.rs
+++ b/src/core/utils.rs
@@ -365,6 +365,23 @@ pub fn resolve_binary(name: &str) -> Result<PathBuf> {
     which::which(name).context(format!("Binary '{}' not found on PATH", name))
 }
 
+/// Resolve the caller's own binary path from `RTK_BIN`, when it names `name`.
+///
+/// A user who runs `.venv/bin/pytest` means that pytest, not whichever one PATH
+/// happens to resolve first. The rewrite layer records the path it was given in
+/// `RTK_BIN`, and this is where handlers pick it back up (#1053).
+///
+/// The basename must match the requested tool, so a stale `RTK_BIN` left over
+/// from another command can never redirect an unrelated binary.
+fn rtk_bin_override(name: &str) -> Option<PathBuf> {
+    let path = PathBuf::from(std::env::var_os("RTK_BIN")?);
+    if path.file_name()? == std::ffi::OsStr::new(name) {
+        Some(path)
+    } else {
+        None
+    }
+}
+
 /// Create a `Command` with PATHEXT-aware binary resolution.
 ///
 /// Drop-in replacement for `Command::new(name)` that works on Windows
@@ -379,6 +396,10 @@ pub fn resolve_binary(name: &str) -> Result<PathBuf> {
 /// # Returns
 /// A `Command` configured with the resolved binary path.
 pub fn resolved_command(name: &str) -> Command {
+    if let Some(path) = rtk_bin_override(name) {
+        return Command::new(path); // nosemgrep: dynamic-command-execution
+    }
+
     match resolve_binary(name) {
         Ok(path) => Command::new(path),
         Err(e) => {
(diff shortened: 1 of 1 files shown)


mhcoen:

```
```

series touches `find`.

mhcoen:
@aeppling @TaKO8Ki

rtk stops filtering whenever the agent happens to spell a command with a path, and nobody
controls that choice. The model writes `pytest -v` on one turn and `.venv/bin/pytest -v` on
the next, same task, same repo, same session. Both are correct commands. Only the first is
rewritten.

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"pytest -v"}}' | rtk hook claude
{"hookSpecificOutput":{...,"updatedInput":{"command":"rtk pytest -v"},"permissionDecision":"allow"}}

$ echo '{"tool_name":"Bash","tool_input":{"command":".venv/bin/pytest -v"}}' | rtk hook claude
(no output)
```

Same for `./venv/bin/pytest`, `/usr/bin/pytest`, `.venv/bin/python -m pytest` and
`./node_modules/.bin/vitest`.

So coverage is not a property of the user's setup that they can configure or verify. It is
decided per command by a model, and it fails silently: no error, no warning, the raw output
just lands in the context. The user cannot work around it by typing the command differently,
because the user is not the one typing. And it makes the savings numbers unstable, since the
same workload measures differently depending on how the model phrased things that day. Any
project whose `CLAUDE.md` or `AGENTS.md` tells the agent to use the project virtualenv, which
is the standard instruction, loses filtering on precisely the highest-volume commands: test
runs, linters, type checkers.

Internally this is an inconsistency, not a gap: `classify_command` already normalizes the
path and reports these commands as supported. The rewrite loop then matches the literal
prefixes against the raw token and drops them, so rtk recognizes the command and throws it
away. #1053, open since 6 April.

The fix is in three parts because stripping the path alone makes rtk run PATH's `pytest`
instead of the venv's: wrong interpreter, wrong dependencies, no diagnostic, which is worse
than not rewriting. So the caller's binary travels with the rewrite in `RTK_BIN`. This PR is
the 71-line piece that consumes it, inert until #3592 sets it. #3591 fixes exclusions for the
same invocation forms and is independent of both.

The rewrite stays conservative: the new matching runs only after the existing literal
prefixes fail, so no command that rewrites today changes its output, and the Composer tools
keep their own normalization. Full suite green, 8 new tests.

Review here unblocks the other two.

(The red `benchmark` check is the pre-existing `find --max 10` negative, #3584 / #3585.)

PR #3591, opened 2026-08-17
fix(hook): apply exclude_commands to path-prefixed invocations

mhcoen:
## Summary

- Match `exclude_commands` against the path-stripped command as well as the raw one, so an exclusion covers the tool rather than one spelling of it
- Excluding `gradlew` now stops `./gradlew`, and excluding `phpunit` now stops `vendor/bin/phpunit`; both currently rewrite straight through the exclusion
- Brings the supported-command branch in line with the TOML-only branch a few lines below, which already normalizes before matching

Part 3 of 3 for #1053, independent of the other two: it stands on its own and can merge in any order.

`exclude_commands` is compared against the raw command, so the setting only works for the spelling the user happened to type. Anyone who configures an exclusion for a tool reasonably expects it to hold however that tool is invoked, and today a `./`-prefixed or `vendor/bin`-prefixed invocation slips past it and gets rewritten anyway.

## Test plan

- [x] `test_exclude_path_prefixed_command` covers `./gradlew` under an exclusion of `gradlew`, `vendor/bin/phpunit` under `phpunit`, and an unrelated exclusion still leaving the command rewritten
- [x] Full suite: 2600 passed on `develop`, 2601 with this branch, with the same 9 failures in both (`cmds::cloud::curl_cmd`, `core::tracking`), which fail on unmodified `develop` in my environment

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1382,7 +1382,13 @@ fn rewrite_segment_inner(
         Classification::Supported { rtk_equivalent, .. } => {
             let stripped = ENV_PREFIX.replace(cmd_part, "");
             let cmd_clean = stripped.trim();
-            if is_excluded(cmd_clean, excluded) {
+            // Match exclusions against the path-stripped form as well, the way
+            // the TOML branch below already does. A user who excludes "gradlew"
+            // means the tool, not one spelling of it, and `./gradlew` would
+            // otherwise rewrite straight through the exclusion.
+            if is_excluded(cmd_clean, excluded)
+                || is_excluded(&strip_absolute_path(cmd_clean), excluded)
+            {
                 return None;
             }
             rtk_equivalent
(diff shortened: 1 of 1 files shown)


mhcoen:

```
```

series touches `find`.

KuSh:
Closing in favour of #3749, which covers this plus the wrapper-invocation forms — but your approach is the one being adopted, so this wasn't wasted.

I checked both against a build. #3749's target check already excludes `./gradlew assembleDebug` and `vendor/bin/phpunit tests/` under `["gradlew"]` / `["phpunit"]`, because the resolved target has no path left to strip. What it *doesn't* cover is your case combined with a subcommand pattern — `["git push"]` against `/usr/bin/git push origin`, where the target is just `git`.

More importantly, substituting the bare target name breaks end-anchored regexes (`["^ls$"]` starts excluding `ls -la`); yours doesn't, because you keep the arguments. So #3749 is being reworked to strip the prefix and match the remainder — your shape, applied to a wider input set. It'll carry `Fixes #1053`.

KuSh:
Correction to what I said above: I stated #3749 would carry `Fixes #1053`. It won't, and I was wrong to promise that.

Reading #1053 properly, its ask is that path-prefixed commands *rewrite* — while preserving the caller's binary, which needs a `--rtk-bin` style override plus handler changes. The issue explicitly calls the naive fix dangerous, since RTK handlers re-resolve via PATH and would run the wrong binary out of an unactivated venv. #3749 only makes the *exclusion* honour path forms, which is fail-safe in the opposite direction: a missed rewrite is passthrough.

So #1053 stays open, and #3749 says so in its scope section.

What did come from this PR: #3749 has been reworked to peel the wrapper/interpreter/path off the command and match the remainder, rather than substituting the resolved target — your argument-preserving shape, applied to a wider input set. That turned out to matter more than I first credited. Substituting the bare target breaks anchored entries outright (`["^ls$"]` starts excluding every `ls`), and it also misses `["eslint"]` against `npx eslint .`, because `eslint` and `biome` both resolve to `rtk lint`. Keeping the arguments, the way you did, avoids both.

`./gradlew assembleDebug` and `vendor/bin/phpunit tests/` — your two test cases — are covered and now have a regression test each in `test_exclude_covers_interpreter_and_path_forms`.

PR #3592, opened 2026-08-17
fix(rewrite): rewrite path-prefixed commands and keep the caller's binary

mhcoen:
## Summary

- Retry the rewrite match against the path-stripped command once the literal prefixes have failed, so `.venv/bin/pytest`, `./node_modules/.bin/vitest` and `/usr/bin/grep` stop falling out of the hook unfiltered
- Carry the caller's path in `RTK_BIN` so the handler runs the binary the user named instead of whatever PATH resolves
- Runs only as a fallback, so every command that rewrites today produces byte-identical output

Part 2 of 3 for #1053, and the part that fixes the reported bug. **Depends on #3590**, which must land first: without it the preserved path is emitted but ignored, which is the wrong-binary failure the issue warns about.

`rtk hook claude` rewrites `pytest -v`, but not `.venv/bin/pytest -v`. `classify_command` already normalizes the path when it decides the command is supported, and then the rewrite loop matches the literal prefixes against the raw token, which cannot start with `pytest`. Every path-prefixed invocation therefore falls out of the hook unfiltered: virtualenv binaries, `./node_modules/.bin`, and absolute `/usr/bin` paths alike.

This generalizes the normalization already merged for the Composer tools. `strip_target` handles the `php`/`vendor/bin` spellings today; `rewrite_path_prefixed()` does the same job for the other ecosystems. The Composer tools are explicitly left out of the new path, since they accept a deliberately narrow set of spellings (`bin/pint` is rejected on purpose) and widening that from here would break `test_rewrite_php_tool_invocation_forms`.

Two deliberate limits keep the blast radius at zero for commands that already work:

- The new matching runs **only after** the existing literal prefixes have failed, so any command that rewrites today is unchanged. `./gradlew` and `./vendor/bin/phpunit` are unaffected.
- The path is preserved **only when its basename is the tool rtk re-invokes**. `.venv/bin/python -m pytest` names python but routes to `rtk pytest`, so handing that path to pytest would resolve the wrong binary; those forms keep today's behavior of not rewriting. Covering them properly means teaching the python handlers about the interpreter, which belongs in its own PR.

The preserved path is single-quoted with POSIX `'\''` escaping, so a directory name containing a quote cannot break out of the assignment.

## Before / after

| command | before | after |
| --- | --- | --- |
| `pytest -v` | `rtk pytest -v` | unchanged |
| `.venv/bin/pytest -v` | no rewrite | `RTK_BIN='.venv/bin/pytest' rtk pytest -v` |
| `/usr/bin/grep -rn foo` | no rewrite | `RTK_BIN='/usr/bin/grep' rtk grep -rn foo` |
| `./node_modules/.bin/vitest run` | no rewrite | `RTK_BIN='./node_modules/.bin/vitest' rtk vitest` |
| `./gradlew assembleDebug` | `rtk gradlew assembleDebug` | unchanged |
| `.venv/bin/python -m pytest -v` | no rewrite | unchanged (see above) |

## Test plan

- [x] 6 new tests covering venv binaries, absolute and `./` paths, bare commands staying free of `RTK_BIN`, the interpreter form being left alone, quote escaping, and an unsupported tool still not rewriting
- [x] Full suite: 2600 passed on `develop`, 2607 with this branch (which includes #3590), with the same 9 failures in both (`cmds::cloud::curl_cmd`, `core::tracking`), which fail on unmodified `develop` in my environment. No existing expectation changed, which is the point of running the new matching only as a fallback.
- [x] Manual testing through the production hook: `echo '{"tool_name":"Bash","tool_input":{"command":".venv/bin/pytest -v"}}' | rtk hook claude` now returns `updatedInput.command` of `RTK_BIN='.venv/bin/pytest' rtk pytest -v`, where it previously returned nothing

Supersedes #1057 (same author, rebased onto current develop and split for review). Overlaps #1378, which takes the same `RTK_BIN` approach across more handlers; happy to fold this into that one if a maintainer prefers its scope.

### DIFF
--- a/src/core/utils.rs
+++ b/src/core/utils.rs
@@ -365,6 +365,23 @@ pub fn resolve_binary(name: &str) -> Result<PathBuf> {
     which::which(name).context(format!("Binary '{}' not found on PATH", name))
 }
 
+/// Resolve the caller's own binary path from `RTK_BIN`, when it names `name`.
+///
+/// A user who runs `.venv/bin/pytest` means that pytest, not whichever one PATH
+/// happens to resolve first. The rewrite layer records the path it was given in
+/// `RTK_BIN`, and this is where handlers pick it back up (#1053).
+///
+/// The basename must match the requested tool, so a stale `RTK_BIN` left over
+/// from another command can never redirect an unrelated binary.
+fn rtk_bin_override(name: &str) -> Option<PathBuf> {
+    let path = PathBuf::from(std::env::var_os("RTK_BIN")?);
+    if path.file_name()? == std::ffi::OsStr::new(name) {
+        Some(path)
+    } else {
+        None
+    }
+}
+
 /// Create a `Command` with PATHEXT-aware binary resolution.
 ///
 /// Drop-in replacement for `Command::new(name)` that works on Windows
@@ -379,6 +396,10 @@ pub fn resolve_binary(name: &str) -> Result<PathBuf> {
 /// # Returns
 /// A `Command` configured with the resolved binary path.
 pub fn resolved_command(name: &str) -> Command {
+    if let Some(path) = rtk_bin_override(name) {
+        return Command::new(path); // nosemgrep: dynamic-command-execution
+    }
+
     match resolve_binary(name) {
         Ok(path) => Command::new(path),
         Err(e) => {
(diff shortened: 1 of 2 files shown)


mhcoen:

```
```

series touches `find`.

PR #3596, opened 2026-08-17
fix(exec): preserve argv in rtk test

wh1teee:
## Summary
- preserve Clap-parsed `rtk test` argv instead of flattening it with `join(" ")` and re-parsing through `sh -c` / `cmd /C`
- reuse the existing `resolved_command(...).args(...)` execution path while keeping the existing test filtering, tee, tracking, environment, and exit-code handling
- add integration coverage for `bash -lc` grouped commands, child paths/literal args with spaces, metacharacters, quotes/backslashes, direct non-shell commands, exit codes, and `pnpm -C <target> run <script>`

## Reproduction
With installed `rtk 0.43.0`, `rtk test bash -lc "cd <child> && <absolute-probe>"` re-runs the probe from the caller CWD. The release candidate built from this branch preserves the grouped `-lc` argument and the probe observes `<child>` via `/proc/<pid>/cwd`.

## Verification
- `cargo fmt --all --check`
- `cargo test --all` — 2696 passed, 8 ignored
- `cargo build --release`
- CodeRabbit CLI review — 0 findings

Related: #3505 (same argv re-parsing root-cause class; this PR is intentionally scoped to `rtk test`).

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -130,16 +130,23 @@ pub fn run_err(command: &str, verbose: u8) -> Result<i32> {
 }
 
 /// Run tests and show only failures
-pub fn run_test(command: &str, verbose: u8) -> Result<i32> {
+pub fn run_test(command: &[String], verbose: u8) -> Result<i32> {
+    let command_label = command.join(" ");
     if verbose > 0 {
-        eprintln!("Running tests: {}", command);
+        eprintln!("Running tests: {}", command_label);
     }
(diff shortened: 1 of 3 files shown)


KuSh:
Closing in favour of #3002, which fixes the same argv flattening for both rtk err and rtk test by quoting each argument before the shell re-parses it, and which has a signed CLA. Your bash -lc repro is covered there. If the direct exec approach for rtk test still matters after that change, please raise it on #3002.

PR #3597, opened 2026-08-17
fix(test): preserve exact command argv

wh1teee:
## Summary

- pass `rtk test` trailing argv directly to the existing `resolved_command(...).args(...)` execution path
- keep the joined command string only for display, tracking, and test-summary heuristics
- reject an empty `rtk test` command instead of retaining a second empty `sh -c` executor path
- add public-CLI regression coverage for grouped `bash -lc`, exact CWD, spaces, metacharacters, quotes/backslashes, non-shell exit codes, and `pnpm -C`

## Root cause

`Commands::Test` flattened `Vec<String>` with `join(" ")`, then `runner::run_test` executed that reconstructed string through `sh -c`/`cmd /C`. A command such as `bash -lc "cd <child> && <probe>"` therefore lost the grouped script argv boundary and could execute the probe from the caller CWD.

The first regression test was run before the fix and failed with `/proc/$$/cwd` evidence: caller CWD was observed where the child CWD was expected.

## Scope / related PRs

This is intentionally narrower than the existing open approaches:

- #2393 direct-spawns but re-tokenizes the single-argument form
- #3002 reconstructs shell quoting and still executes through a shell
- #3057 also addresses argv preservation but combines it with a much broader shell-wrapper/hook rewrite and is currently conflicting

This PR changes only `rtk test`; `rtk err`, `rtk run`, hooks, rewrite logic, filtering, tee behavior, environment, and stdio plumbing are left untouched.

## Verification

- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --all` — 2698 passed, 8 ignored
- `cargo build --release` — passed
- `git diff --check` — passed
- exact-checkout CodeGraph sync/status — up to date

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -2,7 +2,7 @@
 
 use crate::core::stream::StreamFilter;
 use crate::core::truncate::{CAP_LIST, CAP_WARNINGS};
-use anyhow::Result;
+use anyhow::{bail, Result};
 use regex::Regex;
 use std::process::Command;
 use std::sync::LazyLock;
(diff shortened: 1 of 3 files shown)


wh1teee:
Closing this upstream contribution because the fix has been landed and deployed in the maintained fork: wh1teee/rtk#1 (merge commit f2094f90ffb7b0652837365af3a39c7a9a3bbe1c) and vendor release v0.45.0-0fiax.1. The current account has READ-only permission on rtk-ai/rtk and cannot complete maintainer merge/review; CLA remains an external legal gate. The fork deployment is fully qualified and does not depend on this PR. The upstream change can be resubmitted/reopened later if desired.

PR #3607, opened 2026-08-19
Fix/git diff external driver

poelzi:
## Summary
Unbreaks rtk when a git diff viewer is configured

-

## Test plan
Testsuite + my system with difftastic configured


src/cmds/git/git.rs
```diff
@@ -1774,8 +1814,7 @@ fn run_stash(
         Some("show") => {
             let patch_mode = args.iter().any(|a| a == "-p" || a == "--patch");
 
-            let mut cmd = git_cmd(global_args);
-            cmd.args(["stash", "show"]);
+            let mut cmd = git_diff_cmd(global_args, &["stash", "show"]);
```
KuSh, line 1817: `git stash show` changes behaviour when it is given diff options: per git's documentation, passing any diff option switches it out of its default diffstat mode into showing the patch.

So this line makes `rtk git stash show` (no `-p`) receive a full patch, which is then handed to `compact_stash_stat()` — a function that expects a diffstat. It dumps the whole patch verbatim, and where a hunk line happens to match `path | <digit>` it fabricates bogus file rows.

Verified: `git stash show` gives a diffstat; `git stash show --no-ext-diff` gives a full patch.

This one is documented git behaviour rather than a bug, so it is the PR's to handle. The new test `test_git_diff_cmd_keeps_globals_before_subcommand` (line 2186) asserts this exact invocation, so as written it pins the regression rather than catching it.

Passing an explicit `--stat` when the user hasn't asked for `-p` restores the current behaviour.

src/cmds/git/git.rs
```diff
@@ -54,6 +54,37 @@ fn git_cmd_c_locale(global_args: &[String]) -> Command {
...
+/// before diffing, and `diff.noprefix` / `diff.mnemonicPrefix` change the `a/` `b/`
+/// path prefixes `compact_diff()` reads filenames from. Any of them makes the parse
+/// silently yield nothing.
+///
+/// These are subcommand flags, not git global options -- `git --no-ext-diff status`
+/// is an error -- so they are applied per-subcommand, never in `git_cmd()`.
+const DIFF_HARDENING: &[&str] = &[
+    "--no-ext-diff",
+    "--no-textconv",
+    "--src-prefix=a/",
```
KuSh, line 71: Separate from the diffstat-mode problem above, and **this half is a bug in git, not in this PR**. I'm recording it here only because it decides which flags `rtk` should pass.

`git stash show` with explicit `--src-prefix=` / `--dst-prefix=` emits corrupted prefixes — git reads freed memory, so the output varies between runs:

```
git 2.49.1   diff --git a/f.txt b/f.txt          # correct
git 2.52.0   diff --git ributesf.txt 4c/f.txt
git 2.53.0   diff --git Uf.txt Uf.txt            # and Vf.txt on the next run
git 2.54.0   diff --git 4c/f.txt bjectmodef.txt
```

The garbage is recognisable as fragments of other strings on the heap (`ributes` from `attributes`, `bjectmode` from `objectmode`).

Scope, from testing across versions: `git diff --src-prefix=a/ --dst-prefix=b/` is fine everywhere; only `stash show` is affected; `git stash show -p` without prefix flags is fine; and `--no-ext-diff --no-textconv` on `stash show` are fine. The regression arrived in **git 2.52.0** (commit `3ea35c64b`, "stash: tell setup_revisions() to free our allocated strings") and is still present in 2.54.0. It doesn't appear to have been reported upstream; that's being raised with the git project separately.

`compact_diff()`'s `split(" b/")` finds no match against the corrupted output and yields `unknown` for every filename in `rtk git stash show -p`.

**rtk shouldn't be in the business of compensating for bugs in the tools it calls**, and it doesn't need to here: `stash show` has no reason to be given prefix flags in the first place, since RTK is pinning prefixes to defend against `diff.noprefix`/`mnemonicPrefix` in the *user's* config, and `stash show` is not a path where that configuration reaches the parse in a way `--no-ext-diff --no-textconv` doesn't already cover. Dropping the two prefix flags for this subcommand sidesteps it without RTK taking on a version check or a workaround it would have to carry forever.

Worth noting the two problems are independent: dropping only the prefix flags leaves the diffstat mode broken, and adding only an explicit `--stat` leaves the prefixes corrupt.

KuSh (changes requested):
Requesting changes for one thing, and flagging a second that is **not yours to fix**.

**What's verified and right.** I checked the hardening flags against git 2.53 for `git diff`, `git diff --stat`, `git show`, `git show --no-patch`, `git show --stat --pretty=format:` and `git diff --cached --stat --shortstat` — all behave identically to before. The subcommand-versus-global distinction is correct and it matters (`git --no-ext-diff status` really is an error). Globals stay ahead of the subcommand, hardening goes behind it, and applying it before the user's own args so an explicit `--ext-diff` still wins is exactly right for git's last-flag-wins semantics. Leaving passthrough paths unhardened is the correct call. The `compact_diff` empty-result fallback is protected at every call site by `never_worse`, and the difftastic fixture is real captured output, which is what makes that test worth having.

`git stash show` is the one subcommand where routing through `git_diff_cmd()` goes wrong, for two unrelated reasons — one is a genuine bug in this PR, the other is a bug in git itself. Both are inline.

**Suggested fix, covering both:** don't route `stash show` through `git_diff_cmd()`. Build it explicitly as `stash show` + (`--stat`, or the user's `-p`) + `--no-ext-diff --no-textconv`, **without** the prefix flags. I verified that `--no-ext-diff --no-textconv` on `stash show` are safe on every version I tested (2.49.1, 2.52.0, 2.53.0, 2.54.0), so you keep the protection this PR is actually about and simply don't pass two flags that `stash show` never needed. Worth a test asserting `rtk git stash show` still emits the compact stat form, since nothing currently pins that.

One process note: CI has never run on this branch — only `check-target` fired, and it skipped. A rebase or an empty push to trigger the full matrix would help.

kylehgc:
Thanks for this. I adopted the first commit (`5a212a1`, the `compact_diff` raw-text fallback) into my fork with your authorship preserved; it fixes the empty `Changes:` section under `diff.external` end to end, and a real-git integration test now guards it.

I could not take the second commit (`379a34e`, `DIFF_HARDENING`). A base-vs-head A/B across several git config states turned up two regressions, both reproduced by hand on git 2.54:

1. **`rtk git stash show` with no args, default config, flips from stat to full patch.** `git stash show` only applies its `--stat` default when it receives no diff options, so any of the four hardening flags puts it into patch mode. `compact_stash_stat` then finds no stat lines and the guard falls back to raw: a 17-line patch where develop prints `f.txt 1 +` / `1 changed 1 +`. The prefix flags also get mangled there (`diff --git f.txt f.txt`). Hardening buys nothing on that path either: `git stash show -p` never runs `diff.external`.

   ```
   $ git stash show
    f.txt | 1 +
    1 file changed, 1 insertion(+)
   $ git stash show --no-ext-diff
   diff --git a/f.txt b/f.txt
   ...
   ```

2. **`--no-textconv` drops hunks that textconv users previously got.** `textconv` output is ordinary unified diff (`diff --git` + `@@`), so it never triggered the empty-diff bug. With the flag, a `*.pdf diff=pdf` or `od -c` style driver now yields `Binary files a/x and b/x differ`, which `compact_diff` renders as a bare filename with nothing under it.

Two smaller notes: `--src-prefix`/`--dst-prefix` only change rename headers under `diff.noprefix` (non-rename prefixes are already stripped by matching the two halves), and `git show` never invokes external drivers without `--ext-diff`, so the raw-pass comment describes a case that cannot occur.

What still looks right to me from that commit: `--no-ext-diff` alone, on the `run_diff` parsed path only, leaving `stash show` on plain `git_cmd`. That would give difftastic/delta users a compacted unified diff instead of the raw driver rendering the fallback now passes through.

PR #3612, opened 2026-08-20
fix(lint): scope two dead_code false positives to test builds

newblacc:
## Problem

`cargo clippy --all-targets -- -D warnings` fails with two `dead_code` errors:

- `FILTERS_TOML` (`src/core/constants.rs`)
- `TomlFilterRegistry::load` (`src/core/toml_filter.rs`)

Both are false positives. The chain is live in the real binary:

`TomlFilterRegistry::load()` → the `REGISTRY` lazy_static (`toml_filter.rs:410`) → `find_matching_filter` (`toml_filter.rs:678`) → `main.rs:1240`

`--all-targets` additionally builds a test binary that never links main's call path, so the whole chain reads as dead in that target, and `[lints.rust] warnings = "deny"` in Cargo.toml promotes it to a hard error.

## Fix

`#[cfg_attr(test, allow(dead_code))]` on both, rather than an unconditional `#[allow]` — production builds are still held to the deny, so genuinely dead code keeps failing the lint.

## Verification

- `cargo clippy --all-targets -- -D warnings` → exit 0
- `cargo test` → 11 passed, 0 failed

Surfaced while running `scripts/release.sh` in the token-diet superproject, which is the only place this lint currently runs.


newblacc:
Closing — opened against the wrong repository by mistake. This was meant for a downstream fork, not upstream. Apologies for the noise.

Issue #3196, opened 2026-07-24
Add rtk ssh command — compress remote command output


zdravac:
`rtk discover` flagged `ssh` as my top unhandled command by a wide margin — 634 combined invocations (`ssh` + `ssh <host>`) across 30 days, well ahead of any other gap on the list. Running a homelab with several remote Macs/Linux boxes managed over SSH from an AI coding agent session, and every remote command's raw stdout goes straight into context uncompressed.

Requesting an `rtk ssh <args>` wrapper, same shape as `rtk git`/`rtk gh`: pass through the SSH connection itself untouched, but apply rtk's existing output filters to whatever the remote command prints — e.g. `rtk ssh host 'git status'` gets the same compaction as local `rtk git status`, `rtk ssh host 'ls -la'` gets `rtk ls`-style formatting, etc. Ideally dispatches to the same per-command filter rtk already has locally, keyed off the remote command string rather than needing a whole new filter set.

Also flagging `sqlite3` (46 uses) as a secondary case worth considering in the same pass, if useful — table-formatted query output compresses well with the same "strip padding/borders" approach rtk already uses elsewhere (psql already gets this treatment — sqlite3 output is structurally identical).

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 90% |

### Summary

Add `rtk ssh` wrapper that passes the SSH connection through untouched but dispatches remote command output to rtk's existing per-command filters (keyed off the remote command string); also consider `sqlite3` filter reuse from `psql`.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

KuSh:
Closing as a duplicate of #333, which tracks the same problem with the more complete discussion. Please follow up there; if your case differs in a way that issue does not cover, comment here and it can be reopened.

PR #1972, opened 2026-05-19
feat(sqlite): add sqlite3 filter with table output compression

prolifel:
## Summary

- Adds sqlite3 command filter in src/cmds/cloud/sqlite_cmd.rs
- Detects and compresses list, column, and line output formats
- Caps output at 50 data rows with overflow indicator
- Dot-commands (.tables, .schema) and errors passthrough unchanged
- Registered as rtk sqlite with hook-based rewriting from sqlite3

## Changes
- New: src/cmds/cloud/sqlite_cmd.rs - filter module with 15 tests
- Modified: src/main.rs - Command enum, routing, import, operational check
- Modified: src/discover/rules.rs - sqlite3 rewrite rule
- Modified: src/discover/registry.rs - classification and rewrite tests

## Test plan
- [ ] cargo fmt --all && cargo clippy --all-targets && cargo test --all
- [ ] Manual: rtk sqlite mydb.db "SELECT * FROM users;" verifies tab-separated output
- [ ] Manual: rtk sqlite mydb.db ".tables" passes through unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1605, opened 2026-04-29
Feature request: rtk wrapper for sqlite3


mshogin:
Real-world numbers from `rtk discover --all --since 30`:

| Command   | Calls (30d) | Notes                                  |
|-----------|------------:|----------------------------------------|
| `sqlite3` |         470 | top entry under "unhandled" commands   |

Use cases driving the call volume: querying local SQLite stores with frequent `SELECT`, `.schema`, `.tables` calls — patterns where an LLM iteratively explores or polls a small database.

Two main sources of per-call bloat:

1. **Default row dumps from `SELECT`**: with wide TEXT columns, even a few dozen rows expand to several KB of context per call. Most of it is repeated padding/whitespace or long unreadable bodies.
2. **`.schema` output**: full `CREATE TABLE` with all constraints, triggers, and index DDL — far more than an LLM typically needs, which is column names and types.

**Suggested wrapper behavior:**

- For `SELECT` queries:
  - Apply a default row limit (e.g. 50) with a one-line truncation marker
  - Truncate long text cells to ~200 chars with a length marker (`…(N chars)`)
  - Strip trailing whitespace/padding
- For `.schema` / `.tables`:
  - One line per table: `table_name(col1 type, col2 type, ...)` with PK/FK markers, no constraint bodies
  - Index list summarised separately
- For `.dump` and similar export commands: pass through unchanged (intent is the full payload)
- Pass through unchanged when output is already short (e.g. < 5 lines)

**Methodology:** counts come from `rtk discover --all --since 30` against the local Claude Code transcript directory. Token estimates use the 4-chars/token approximation.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| ✨ **Category** | `feature` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 92% |

### Summary

Feature request to add an rtk wrapper for sqlite3 that truncates large SELECT output, compresses schema output, and passes through dump/short output unchanged.

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

KuSh:
[...]

Issue #1253, opened 2026-04-12
feat: add sqlite3 filter — compress query output, .schema dumps, and .explain plans


Flipsi85:
## Summary

Add a `rtk sqlite3` filter analogous to the existing `rtk psql` — strip column-padding, borders, repeated headers, and sqlite system tables from sqlite3 CLI output so query results, schema dumps, and query plans consume less LLM context.

## Use case

`sqlite3` is the de-facto system database on macOS (TCC, Quicklook, Launchservices, Chrome/Firefox/Slack state, Apple Mail index, many Electron apps) and the most common embedded DB for CLI/desktop tools. It's a frequent target in Claude Code sessions that need to inspect app state without installing a server DB. It is also a gap in the current registry: `rtk psql` exists, but `rtk sqlite3` does not.

`rtk discover` on a 30-day window of my Claude Code history lists sqlite3 under "TOP UNHANDLED COMMANDS":

```
sqlite3    277    sqlite3 "/Library/Application Support/..."
```

277 invocations in 30 days is comparable to the volume of `rtk prisma` (138) and higher than many existing filters — a first-class handler is justified.

## What is wasteful in raw sqlite3 output

1. **`.mode column` + `.headers on`** — fixed-width padding inflates every cell; a 10-column result with long VARCHARs can multiply row size 3–5×.
2. **Repeated column headers** when the caller runs multiple queries in one invocation with `.headers on`.
3. **`.schema` output** — trailing commas, indentation, blank lines, and system tables (`sqlite_sequence`, `sqlite_stat1..4`, `sqlite_master`) that are almost never what the user asked for.
4. **`.dump`** — `BEGIN TRANSACTION;` / `COMMIT;` boilerplate and leading `PRAGMA foreign_keys=OFF;` / `PRAGMA defer_foreign_keys=ON;` lines.
5. **`.explain query plan`** — the "detail" column is whitespace-heavy and the "id/parent/notused" columns are rarely useful to an LLM.

## Proposed behavior

Mirror `rtk psql`: force a compact output mode, strip borders, cap row count with a truncation notice on large result sets.

### Query output

```console
# Before (native sqlite3 with .mode column, .headers on)
$ sqlite3 db.sqlite \".headers on\" \".mode column\" \"SELECT id, name, email FROM users LIMIT 3\"
id    name           email
----  -------------  --------------------
1     Alice Johnson  alice@example.com
2     Bob Miller     bob@example.com
3     Carol Chen     carol@example.com

[...]

PR #3629, opened 2026-08-21
feat(filters): add sqlite3 TOML filter

saddestmartian:
## Summary

Adds a built-in TOML filter for the `sqlite3` CLI (`src/filters/sqlite3.toml`, `match_command = "^sqlite3\b"`), same shape as the existing `jq.toml` / `ssh.toml` filters.

Strips only always-uninteresting decoration:
- column-mode header underlines (`--  -----`)
- empty lines
- the sqliterc `Loading resources from` banner

Default list-mode query rows (`1|alice`) pass through untouched. No `max_lines` / `truncate_lines_at` (position-based cuts).

## Why TOML, not a Rust module

Open PR #1972 (`feat(sqlite): add sqlite3 filter with table output compression`) is a Rust `sqlite_cmd.rs` wrapper with a 50-row cap — currently `DIRTY` / `CONFLICTING`. This PR is the smaller TOML slice, matching how `jq` and `ssh` already ship, so it can land independently. Happy to close this if #1972 is the preferred path.

Related: #1253, #1605, #3196 (mentions sqlite3 as a secondary case).

## Proof against stock rtk 0.45.0

Project-local copy of this filter, trusted vs untrusted, on macOS, binary `/Users/Michael/.local/bin/rtk` = rtk 0.45.0:

**Representative command** (kit corpus uses `sqlite3 -header -column`, e.g. hum-rail):

```
sqlite3 -header -column :memory: "CREATE TABLE t(id INTEGER, name TEXT);" \
  "INSERT INTO t VALUES (1,'alice'),(2,'bob');" "SELECT * FROM t;"
```

| Arm | hook rewrite | bytes | output |
|---|---|---|---|
| untrusted | `No rewrite for: sqlite3 ...` | 40 | keeps `--  -----` underline |
| trusted | `rtk sqlite3 -header -column ...` (rewrite rc=0) | 30 | underline stripped; header + rows kept |

Default list-mode `SELECT` is identical on both arms (14 bytes, `1\|alice\n2\|bob\n`) — already compact; the filter is a no-op there by design.

`rtk verify --filter sqlite3` → `4/4 tests passed` once trusted.

## Test plan

- [ ] `cargo test --lib test_builtin_filter_count test_builtin_all_expected_filters_present test_new_filter_discoverable_after_concat test_builtin_all_filters_have_inline_tests`
- [ ] Manual: `rtk sqlite3 -header -column db.sqlite "SELECT * FROM t LIMIT 5"` drops the underline
- [ ] Manual: `rtk sqlite3 db.sqlite "SELECT 1"` passthrough

### DIFF
--- a/src/core/toml_filter.rs
+++ b/src/core/toml_filter.rs
@@ -1864,6 +1864,7 @@ match_command = "^make\\b"
             "shellcheck",
             "shopify-theme",
             "sops",
+            "sqlite3",
             "swift-build",
             "systemctl-status",
             "terraform-plan",
@@ -1892,8 +1893,8 @@ match_command = "^make\\b"
         let filters = make_filters(BUILTIN_TOML);
         assert_eq!(
             filters.len(),
-            63,
-            "Expected exactly 63 built-in filters, got {}. \
+            64,
+            "Expected exactly 64 built-in filters, got {}. \
              Update this count when adding/removing filters in src/filters/.",
             filters.len()
         );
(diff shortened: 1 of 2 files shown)


saddestmartian:
signed

PR #3638, opened 2026-08-22
fix(test): stop hook tests reading the machine's permission and rewrite config

xantorres:
Hook unit tests resolved permission verdicts through `decide_hook_action`, which reads `.claude/settings.json` from the home directory and from the project — the latter located by walking up the process-global CWD. A local `Bash(cargo test)` or `Bash(git status)` deny rule turns an expected rewrite into a skip, so six Claude-path tests fail on that machine and pass everywhere else. The CWD walk also races the `init.rs` tests that call `set_current_dir`. `get_rewritten` had the same problem one layer down, loading `hooks.exclude_commands` from `config.toml`, where a local `exclude_commands = ["grep"]` already broke `test_claude_pipeline_rewrites_only_safe_final_stage`.

## Changes

- Both inputs now reach `decide_from_verdict` as parameters, matching `evaluate_with_verdict` in `rewrite_cmd.rs`, which already takes this exact `(excluded, transparent_prefixes)` pair. `decide_hook_action` becomes the thin wrapper that reads them off the machine, so a seam can no longer inherit ambient state by omission.
- `process_claude_payload` splits into a wrapper plus `claude_action_from_decision`, letting `run_claude_inner_with_rules` inject the verdict — the seam `run_cursor_inner_with_rules` and `run_droid_inner_with_rules` already have. Copilot CLI tests go through `copilot_cli_response_with_rules`; `run_vibe_inner` takes its settings from `run_vibe`.
- `end_to_end` passed the unextracted `"Bash(git:*)"` as an allow rule, but rules arrive at `check_command_with_rules` already unwrapped by `extract_bash_pattern`, so it matched nothing. `test_copilot_cli_cve_newline_bypass_never_auto_allows` and its background-`&` sibling rely on that rule to exercise the per-segment allow invariant, and bare `\n`/`&` are not unattestable constructs — so both were unfalsifiable. Reverting the invariant to its pre-fix "any matching segment allows the chain" form leaves both green with the old literal and fails both with the corrected one. Their asserts are unconditional now too.
- New coverage: the Allow and Deny directions through the Claude seam, the deny-versus-defer audit reason, a non-string command payload, `exclude_commands` suppressing a rewrite, and the `Config` to `(excluded, transparent_prefixes)` field mapping.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` — clean, 2632 pass.
- [x] Test binary run with `HOME` and CWD pointed at settings files holding deny, allow and mixed-scope rules alongside a `config.toml` excluding grep/git/cargo, with and without `RTK_HOOK_AUDIT=1`: all 430 `hooks::` tests pass in every combination, against 27 failures before the fix.

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -231,16 +231,32 @@ fn heal_legacy_hook_file(path: &std::path::Path) -> bool {
         .is_ok()
 }
 
-fn get_rewritten(cmd: &str) -> Option<String> {
+/// `(exclude_commands, transparent_prefixes)` from the machine's `config.toml`.
+fn hook_rewrite_settings() -> (Vec<String>, Vec<String>) {
+    crate::core::config::Config::load()
+        .map(settings_from_config)
+        .unwrap_or_default()
+}
+
+/// Split from the load so the field mapping stays reachable from a test —
+/// `Config::load` only ever reads the developer's own file.
+fn settings_from_config(config: crate::core::config::Config) -> (Vec<String>, Vec<String>) {
+    (
+        config.hooks.exclude_commands,
+        config.hooks.transparent_prefixes,
+    )
+}
+
+fn get_rewritten(
+    cmd: &str,
+    excluded: &[String],
+    transparent_prefixes: &[String],
+) -> Option<String> {
     if has_heredoc(cmd) {
         return None;
     }
 
-    let (excluded, transparent_prefixes) = crate::core::config::Config::load()
-        .map(|c| (c.hooks.exclude_commands, c.hooks.transparent_prefixes))
-        .unwrap_or_default();
-
-    let rewritten = rewrite_command(cmd, &excluded, &transparent_prefixes)?;
+    let rewritten = rewrite_command(cmd, excluded, transparent_prefixes)?;
 
     if rewritten == cmd {
         return None;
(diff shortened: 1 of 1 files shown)


KuSh:
Keeping this one as the fix for hook tests reading the machine's real config — #3461 covered the same two functions and is closed pointing here.

Two things before it can land:

The branch is in conflict and needs a rebase on `develop`.

More importantly: you have #3645 open against the same file, also touching the hook entry points, and also conflicting. Rather than guess at the intended order, which of the two supersedes the other? Happy to take them as a stack if they're meant to be sequential, or as one consolidated PR if #3645 grew out of this one — whichever matches what you had in mind. Just not both landing independently against `hook_cmd.rs` in their current state.

For context on why this matters beyond test hygiene: #3324 hit exactly this problem from the other side — three unrelated tests failing on the contributor's machine because `check_command` consults real agent permission settings, which cost them a round of investigation to rule out as their own regression.

KuSh:
Thanks for chasing this one down. Develop now does what this PR set out to do: #3955 (merged 2026-09-11) extracted src/hooks/decision.rs with decide_with_params, which takes exclude_commands and transparent_prefixes as arguments so a decision under test no longer picks up the developer's own config.toml, and #3206 (merged 2026-09-02) split process_claude_payload into process_claude_payload_from_decision so the Allow/Ask/Deny/Defer matrix is testable without real permission files. The verdict itself was already passed in rather than looked up, so that part of the seam was in place before. Two things you found are still true on develop and would be welcome as an issue or a narrow PR against the new decision module: end_to_end in src/hooks/hook_cmd.rs still passes the unextracted "Bash(git:*)" as an allow rule, which leaves the newline and background-& bypass tests unfalsifiable, and run_vibe_inner still resolves its verdict and rewrite settings off the machine with no rules seam of its own. If I have missed a part of this that develop still lacks, please say so or reopen.

Issue #3634, opened 2026-08-22
rtk rewrite exits 3 with a valid rewrite on stdout, contradicting its documented 0/1 contract, silently disabling every hook


yonatangross:
## Summary

`rtk rewrite` exits **3** while printing a valid rewrite on stdout. Its own `--help` documents exit **0** for that case. Every hook that follows the documented contract therefore discards a correct rewrite and silently falls through.

Measured on `rtk 0.45.0` (Homebrew, macOS arm64):

```
$ rtk rewrite "git status"
rtk git status
$ echo $?
3
```

`--help` states:

> Exits 0 and prints the rewritten command if supported. Exits 1 with no output if the command has no RTK equivalent.

So `3` is neither documented value, and it arrives together with correct output.

## Why this is worse than a wrong exit code

The shipped Claude Code hook (`rtk-rewrite.sh`, `rtk-hook-version: 2`) follows the documented contract exactly, and the README recommends the same idiom to other integrators:

```sh
REWRITTEN=$(rtk rewrite "$CMD") || exit 0
```

Under `|| exit 0`, any non-zero exit reads as "no RTK equivalent" and the hook passes through. Nothing errors. Nothing logs. Commands simply stop being rewritten, and the failure is indistinguishable from normal operation.

## Impact, from rtk's own telemetry

`rtk discover` on this machine, after the hook had been silently inert:

```
Scanned: 871 sessions (last 30 days), 34182 Bash commands
Already using RTK: 9 commands (0.0%)

Command       Count    RTK Equivalent   Est. Savings
grep -n        2394    rtk grep         ~588.2K tokens
git log        2310    rtk git          ~462.5K tokens
cat -n         1115    rtk read         ~426.0K tokens
gh pr          1569    rtk gh           ~371.4K tokens
```

Roughly 1.8M tokens missed on the top four patterns alone, at **0.0% adoption across 34,182 commands**. `discover` had been reporting this accurately the whole time; nothing surfaced it, because the integration path fails closed and quiet.

## Suggested fix

Restore the documented contract: exit `0` when a rewrite is produced, `1` when there is none. That needs no change from any existing integrator.

If exit `3` is intentional and carries meaning, please document it in `rtk rewrite --help` and update the README idiom, since `|| exit 0` is actively wrong against the current binary.

## Workaround for other integrators

Branch on the output rather than the exit code. This is forward compatible either way:

```sh
REWRITTEN=$(rtk rewrite "$CMD" 2>/dev/null)
[...]

PR #3639, opened 2026-08-22
fix(cli): document rewrite exit-code contract

tandede:
## Summary

- document all four exit statuses returned by `rtk rewrite`
- remove integration guidance that discards valid rewrites returned with exit 3
- align the feature documentation and add regression coverage for the public CLI help contract

## Context

Exit 3 intentionally includes the rewritten command on stdout and indicates that user approval is still required. The existing CLI help only documents exits 0 and 1 and recommends a shell idiom that treats every nonzero result as a passthrough.

The shipped hooks already distinguish exit 3 correctly, so this change does not alter runtime permission behavior. It makes the public integration contract match the existing implementation.

## Testing

- `cargo fmt --all --check`
- `cargo test` — 2705 passed, 8 ignored
- manually verified exit 3 with rewrite output and exit 1 without output

Addresses #3634.

### DIFF
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -1175,12 +1175,18 @@ rtk init -g --uninstall         # Desinstaller
 
 ### `rtk rewrite` -- Recriture de commande
 
-Commande interne utilisee par le hook. Imprime la commande reecrite sur stdout (exit 0) ou sort avec exit 1 si aucun equivalent RTK n'existe.
+Commande interne utilisee par le hook. Le code de sortie preserve la decision de permission :
+
+- `0` : imprime une commande reecrite explicitement autorisee
+- `1` : aucun equivalent RTK, sans sortie
+- `2` : une regle de refus correspond, sans sortie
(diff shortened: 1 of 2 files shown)


jondmarien:
Confirmed on RTK 0.49.0 with Codex CLI 0.155.1 on Linux: `rtk rewrite 'git status'` prints `rtk git status` but exits 3; an unsupported command such as `pwd` exits 1 with no output.

Workaround on my setup: route Codex `PreToolUse` through TokSave’s `toksave rtk-hook codex` adapter. It uses the rewrite output instead of treating every nonzero status as “no rewrite,” then returns the changed command through `hookSpecificOutput.updatedInput`. End-to-end, Codex’s `ls -la` ran as `rtk ls -la` and exited 0; output went from 5,130 to 864 characters (about 83% fewer bytes). A generic adapter can likewise capture stdout and apply a non-empty rewrite even when the process exits 3.

Thanks for documenting the exit-code contract here; this workaround kept my Codex integration working without relying on the misleading `|| exit 0` pattern.

PR #3648, opened 2026-08-23
fix: claude command detecting issue for windows

hosseinmobarakian:
This fix is according to issue #3632
I changed the shell_split function structure to make it easier to detect commands and better for both Windows and Unix systems:

```rust 
pub fn shell_split(input: &str) -> Vec<String> {
    input
        .replace('"', "")
        .split_whitespace()
        .map(|s| s.to_string())
        .collect()
}
```

I updated the test for is_claude_hook_command() at src/hooks/mod.rs:26 to check Windows commands too.

### DIFF
--- a/src/discover/lexer.rs
+++ b/src/discover/lexer.rs
@@ -456,41 +456,11 @@ pub fn strip_quotes(s: &str) -> String {
 }
 
 pub fn shell_split(input: &str) -> Vec<String> {
-    let mut tokens = Vec::new();
-    let mut current = String::new();
-    let mut chars = input.chars().peekable();
-    let mut in_single = false;
-    let mut in_double = false;
-
-    while let Some(c) = chars.next() {
-        match c {
-            '\\' if !in_single => {
-                if let Some(next) = chars.next() {
-                    current.push(next);
-                }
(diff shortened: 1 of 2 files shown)


KuSh:
Thanks for digging into this. The rtk.exe half of this landed on develop in #3552 (merged 2026-09-13) and was extended in #3008 (merged 2026-09-18): the shared is_rtk_binary helper in src/hooks/mod.rs now matches both rtk and rtk.exe, with Windows positive and negative tests alongside it. The shell_split half can no longer apply, because #3704 (merged 2026-09-04) rebuilt shell_split on the shared tokenize and coalesce_words primitives that the rewrite and discover paths also use, so the hand-rolled scanner this PR replaces is no longer in the file. The case from #3632 is still real on develop: an unquoted Windows path such as C:\Users\x\rtk.exe hook claude has its backslashes consumed as escapes by resolve_word_text, so the hook command is not recognised, and a narrow fix inside the shared lexer (or a platform-aware binary-name compare) would be very welcome as a fresh PR. Closing this one as it stands, but please comment or reopen if you think a part of it was missed.

PR #3660, opened 2026-08-23
fix(find): newline-terminate the output

### LINKED ISSUE #3659: `rtk find` output has no trailing newline, so line counters are off by one
**Type:** Bug · **Severity:** Low (output hygiene) · **Area:** `src/cmds/system/find_cmd.rs`

## Summary

`run` ends with `print!("{}", shown)`. `find` newline-terminates its last result; RTK does not, so
the final entry is not a complete line.

## Reproduction

```bash
printf 'x\n' > one.txt; printf 'x\n' > two.txt

find . -name '*.txt' | tail -c 1 | od -An -tx1   #  0a
rtk  find . -name '*.txt' | tail -c 1 | od -An -tx1   #  74  ('t')

find . -name '*.txt' | wc -l        # 2
rtk  find . -name '*.txt' | wc -l   # 1
```

## Impact

- `wc -l` undercounts by one.
- `while read -r path; do …; done` silently drops the final entry.
- The next command's output, or the shell prompt, runs on to the same line.

None of these are catastrophic on their own, but `rtk find` exists to be piped, and a
non-terminated last line is a trap for every consumer that assumes line-oriented output.

## Expected behaviour

Non-empty output ends with `\n`. Empty output stays completely empty — a lone newline for "no
results" would be its own small lie.

## Environment

- GNU findutils 4.9.0, coreutils wc 9.1, Linux

dwin-gharibi:
Closes #3659.

## Summary

- `rtk find` now newline-terminates its output, matching `find`, so `wc -l` counts correctly and
  `while read` no longer drops the last entry.
- Empty output stays byte-for-byte empty — no lone newline for "no results".
- Implemented as a small `terminated()` helper returning `Cow`, so the common already-terminated
  case allocates nothing.

## Test plan

### Unit test (added)

`src/cmds/system/find_cmd.rs` — `output_is_newline_terminated`: appends `\n` when missing, leaves
an already-terminated block alone, and keeps the empty case empty.

### Behavioural check against real find

| | real `find` | `rtk find` before | `rtk find` after |
|---|---|---|---|
| last byte of output | `0x0a` | `0x74` (`t`) | `0x0a` |
| `\| wc -l` for 2 results | 2 | 1 | 2 |
| no-results output | empty | empty | empty |

### Suite

`cargo test --all`: 2626 passed, 0 failed (2625 on `develop`, +1 from this PR).
`cargo clippy --all-targets`: clean. `cargo fmt --all`: clean.

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -7,6 +7,19 @@ use ignore::WalkBuilder;
 use std::collections::HashMap;
 use std::path::Path;
 
+/// Ensure the block ends with a newline.
+///
+/// `find` newline-terminates its last result. Without it `wc -l` undercounts by
+/// one, `while read` drops the final entry, and the next line of output runs on
+/// to the same line.
+fn terminated(body: &str) -> std::borrow::Cow<'_, str> {
+    if body.is_empty() || body.ends_with('\n') {
+        std::borrow::Cow::Borrowed(body)
+    } else {
+        std::borrow::Cow::Owned(format!("{body}\n"))
+    }
+}
+
 /// Match a filename against a glob pattern (supports `*` and `?`).
 fn glob_match(pattern: &str, name: &str) -> bool {
     glob_match_inner(pattern.as_bytes(), name.as_bytes())
@@ -373,7 +386,7 @@ pub fn run(
     }
 
     let shown = never_worse(&raw_output, &body);
-    print!("{}", shown);
+    print!("{}", terminated(shown));
     timer.track(
         &format!("find {} -name '{}'", path, effective_pattern),
         "rtk find",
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the careful writeup and the before/after byte comparison. This is fixed on develop: the print! call this PR wraps is gone, and both output paths in find now go through the shared emit_guarded helper added by #2554 and adopted by find in #3603 (merged 2026-08-26, shipped in v0.46.0), which newline-terminates. Checking the develop binary with your own commands, the output ends with 0a, wc -l matches find's count, and a run with no matches still emits zero bytes, so the empty case you protected is also already correct. Issue #3659 was closed as completed on 2026-09-13 for the same reason. Closing this, but please comment or reopen if you still see a missing newline on a case we did not cover.

PR #3666, opened 2026-08-23
fix(search): treat -f/--file as a pattern source, not a positional pa…

### LINKED ISSUE #3665: `rtk grep -f patterns.txt file` searches the wrong input and reports no matches
**Type:** Bug · **Severity:** Critical (silent wrong result) · **Area:** `src/cmds/system/search.rs`

## Summary

`-f FILE` / `--file FILE` tells grep and ripgrep to read *patterns from a file*. It is the
file-based twin of `-e`. RTK classifies it as an ordinary value-taking flag, so the first
remaining positional — the file the user wanted to search — is consumed as the pattern and the
path list ends up empty. The engine then reads stdin instead of the requested file.

The command exits `1` ("no matches") with no output and no warning, which is indistinguishable
from a genuine no-match.

## Reproduction

```bash
printf 'ERROR db timeout\nINFO ok\nWARN disk full\nERROR oom\n' > app.log
printf 'ERROR\nWARN\n' > patterns.txt

grep -f patterns.txt app.log        # 3 matches, exit 0
rtk  grep -f patterns.txt app.log   # no output,  exit 1
```

`rtk -v` shows the misparse directly — the search path has become the pattern and no path is left:

```
$ rtk -v grep -f patterns.txt app.log
grep: 'app.log' in
```

Affects every spelling: `-f FILE`, `--file FILE`, `--file=FILE`, `-fFILE`.

### Worse than empty with two files

```bash
printf 'ERROR two\n' > second.log

grep --file patterns.txt app.log second.log
# app.log:ERROR db timeout
# app.log:WARN disk full
# app.log:ERROR oom
# second.log:ERROR two

rtk grep --file patterns.txt app.log second.log
# ERROR two          <- the wrong file, presented as the whole answer
```

Here RTK does not return nothing; it returns a plausible-looking answer drawn from the wrong
file, because `app.log` was used as the pattern and only `second.log` remained as a path.

## Root cause

`extract_pattern_path()` splits args into `(patterns, paths, flags)`. `-e`/`--regexp` are special
cased: when either is present, *all* positionals are paths. `-f`/`--file` carry exactly the same
meaning but are listed among the generic value-taking flags (`VALUE_FLAGS_SHORT = b"ABCMTdfgjmt"`,
`VALUE_FLAGS_LONG` contains `"--file"`), so the "first positional is the pattern" branch still runs.

## Expected behaviour

When a pattern source is present — `-e`, `--regexp`, `-f`, or `--file` — every positional is a
path, and RTK's output matches the underlying engine's exactly.

## Impact

An agent grepping with a pattern file gets "no matches" for a file that does match, or matches
from a file it did not ask about. Both are silent: exit codes and stderr are indistinguishable
from correct behaviour, so nothing downstream can detect the substitution.

## Environment

- GNU grep 3.8, Linux
- Reproduced on both debug and release builds

dwin-gharibi:
Closes #3665.

## Summary

- Treat `-f` / `--file` as a **pattern source** in `extract_pattern_path`, exactly like
  `-e` / `--regexp`: when one is present every positional is a path, not a pattern.
- Fixes `rtk grep -f patterns.txt app.log` silently returning no matches (exit 1) where real
  grep returns matches — and, with two files, returning matches from the wrong file.
- The flag is tracked during the arg scan rather than by inspecting `flags` afterwards, so
  another option's *value* that happens to read `-f` cannot be mistaken for it.

With no positional pattern left, `run()` takes its existing `patterns.is_empty()` passthrough
branch and executes the user's command verbatim — correct output, no compression on this form.
Grouping `-f` searches is a follow-up; correctness first.

## Test plan

`src/cmds/system/search.rs`:

- `pattern_file_flag_leaves_positionals_as_paths` — all four spellings (`-f F`, `--file F`,
  `--file=F`, `-fF`) yield no positional pattern and keep the search path.
  Fails on `develop` with: `["-f", "pats.txt", "hay.txt"]: a pattern file must not turn a
  positional into the pattern, got ["hay.txt"]`.
- `pattern_file_flag_is_forwarded_to_the_engine` — `-f` and its value still reach the engine.
- `positional_pattern_still_wins_without_a_pattern_source` — guards the ordinary form.

### Behavioural check against real grep

Every case below now produces byte-identical output and exit code to GNU grep:

| Command | Before | After |
|---|---|---|
| `grep -f patterns.txt app.log` | no output, exit 1 | 3 lines, exit 0 — identical |
| `grep --file patterns.txt app.log` | no output, exit 1 | identical |
| `grep --file=patterns.txt app.log` | no output, exit 1 | identical |
| `grep -fpatterns.txt app.log` | no output, exit 1 | identical |
| `grep -f patterns.txt app.log second.log` | wrong file's match | 4 lines, identical |
| `grep -n -f patterns.txt app.log` | no output, exit 1 | identical |

Controls confirming ordinary grep is untouched: `grep -n ERROR`, `grep -c ERROR`,
`grep -e ERROR -e WARN`, `grep ERROR a b` — all identical before and after.

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -146,6 +146,10 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
     let mut e_patterns: Vec<String> = Vec::new();
     let mut positionals: Vec<String> = Vec::new();
     let mut flags: Vec<String> = Vec::new();
+    // -f/--file read patterns from a file, so like -e they leave every
+    // positional as a path. Tracked during the scan rather than by inspecting
+    // `flags` afterwards, so another flag's *value* can't be mistaken for it.
+    let mut has_pattern_file = false;
     let mut past_dashdash = false;
     let mut i = 0;
 
@@ -177,6 +181,7 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
             }
             // Other long value-taking flags: consume next token as value.
             if VALUE_FLAGS_LONG.contains(&arg) {
+                has_pattern_file |= arg == "--file";
                 flags.push(arg.to_string());
                 if i + 1 < args.len() {
                     flags.push(args[i + 1].as_ref().to_string());
@@ -186,6 +191,7 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
                 }
                 continue;
             }
+            has_pattern_file |= arg.starts_with("--file=");
             flags.push(arg.to_string());
             i += 1;
             continue;
@@ -219,6 +225,7 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
                             i += 1;
                         }
                     } else {
+                        has_pattern_file |= flag == 'f';
                         flags.push(format!("-{}", flag));
                         if !inline.is_empty() {
                             flags.push(inline);
@@ -239,9 +246,10 @@ fn extract_pattern_path<T: AsRef<str>>(args: &[T]) -> (Vec<String>, Vec<String>,
         }
     }
 
-    // If -e/--regexp was used: all positionals are paths.
+    // If a pattern source was used (-e/--regexp, or -f/--file reading patterns
+    // from a file): all positionals are paths.
     // Otherwise: first positional is the pattern, rest are paths.
-    let (patterns, paths) = if !e_patterns.is_empty() {
+    let (patterns, paths) = if !e_patterns.is_empty() || has_pattern_file {
         (e_patterns, positionals)
     } else {
         let paths = positionals.iter().skip(1).cloned().collect();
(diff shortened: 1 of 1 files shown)


devnulled:
This also fixes the `rg` engine, where the bug is worse than the grep case
described here — I built the branch and measured it. Flagging because the rg
half isn't mentioned in the description or pinned by a test, and it's the half
that fails silently.

On grep, `-f` stealing the path gives an empty result and exit 1. On rg it gives
a **larger** wrong answer at **exit 0**, because rg recurses the working
directory when the path list comes back empty. Same exit code as a correct run,
plausible output, wrong result set.

Fixture — three files; the question asked is "which lines of `hay.txt` match the
patterns in `pats.txt`":

```
./hay.txt        alpha FOO one / beta bar / gamma FOO three
./pats.txt       FOO
./sub/other.txt  unrelated FOO in another file
```

Three-way, all four `-f` spellings, `rg` engine:

```console
$ rg -f pats.txt hay.txt                    # real ripgrep 15.2.0
alpha FOO one
gamma FOO three
exit=0

$ rtk rg -f pats.txt hay.txt                # v0.48.0
FOO                                         <-- from pats.txt, not searched
alpha FOO one
gamma FOO three
unrelated FOO in another file               <-- from sub/, not searched
exit=0

$ rtk rg -f pats.txt hay.txt                # this branch
alpha FOO one
gamma FOO three
exit=0
```

The extra lines come from files the caller never named — and note it matched
inside `pats.txt` itself, so the pattern file contaminated its own results.

Every form reaches parity with the real engine on this branch, and every one of
them is broken on v0.48.0:

| form | engine | v0.48.0 | this branch |
|---|---|---|---|
| `-f pats.txt hay.txt` | rg | 4 lines, exit 0 | parity |
| `--file pats.txt hay.txt` | rg | 4 lines, exit 0 | parity |
| `--file=pats.txt hay.txt` | rg | 4 lines, exit 0 | parity |
| `-fpats.txt hay.txt` | rg | 4 lines, exit 0 | parity |
| `-f pats.txt hay.txt` | grep | empty, exit 1 | parity |
| `--file pats.txt hay.txt` | grep | empty, exit 1 | parity |
| `-fpats.txt hay.txt` | grep | empty, exit 1 | parity |

`cargo test --all` on the branch: 2707 passed, 0 failed.

Measured on debug builds of `88a50e9` (this branch) and `fde0a8f` (v0.48.0),
macOS, against `/opt/homebrew/bin/rg` and `/usr/bin/grep` by absolute path. That
last detail matters for anyone re-running it: in some agent shells `grep` is a
shell function routing to a bundled ugrep, and `rg` gets rewritten by rtk's own
hook, so a differential using bare names quietly compares rtk against rtk.

The one gap: `pattern_file_flag_leaves_positionals_as_paths` exercises
`extract_pattern_path` directly, so nothing pins the rg-side end-to-end
behavior. Since the rg failure is exit 0 with plausible output, a regression
there can't be caught by watching for an error — a case asserting that
`rtk rg -f pats.txt hay.txt` returns exactly the engine's own lines would close
it. Killing mutation for such a test: revert the condition on line 252 to
`!e_patterns.is_empty()`.

KuSh:
Thanks for the precise diagnosis and the four-spelling test table. This landed on develop in #3681 (merged 2026-09-12): extract_pattern_path now tracks a patterns_from_file flag during the scan and, as in your patch, treats every positional as a path when -f or --file is present, with a test covering -f F, --file F, -fF and --file=F. That PR also rebuilt the function on the shared argument tokenizer, so the hand-rolled flag scan your hunks edit is no longer in the file and they can no longer apply. Issue #3665 is still open only because it has not been verified against a release yet. Closing this, but please comment or reopen if you find a form that develop still gets wrong.

Issue #2809, opened 2026-07-04
rtk pnpm typecheck incorrectly rewrites command to pnpm tsc


tanakalucky:
## Description
When running `rtk pnpm typecheck` in the root of a project, `rtk` incorrectly rewrites the command to `pnpm tsc` (or runs `tsc` directly), executing a TypeScript build rather than running the `typecheck` script defined in `package.json`.

In the reproduction project's `package.json`, `typecheck` is defined to use Turborepo:
```json
"scripts": {
  "typecheck": "turbo run typecheck"
}
```

Instead of running `pnpm run typecheck` (which would trigger `turbo run typecheck`), `rtk` rewrites and executes `pnpm tsc`.

## Reproduction Repository
A minimal reproduction repository is available here:
https://github.com/tanakalucky/rtk-pnpm-typecheck-rewrite-test

## Steps to Reproduce
1. Clone the reproduction repository and navigate into it:
   ```bash
   git clone https://github.com/tanakalucky/rtk-pnpm-typecheck-rewrite-test.git
   cd rtk-pnpm-typecheck-rewrite-test
   ```
2. Install dependencies:
   ```bash
   pnpm install
   ```
3. Run the reproduction command:
   ```bash
   rtk pnpm typecheck
   ```

## Expected Behavior
`rtk` should execute `pnpm typecheck`, which in turn runs the script defined in `package.json` (`turbo run typecheck`).

## Actual Behavior
`rtk` rewrites the command to `pnpm tsc` (or runs `tsc` directly) and starts a compiler build, ignoring the custom script definition in `package.json`.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 88% |

### Summary

rtk's command-rewrite logic incorrectly maps `pnpm typecheck` to `pnpm tsc`/`tsc` instead of passing through to the package.json script (e.g. `turbo run typecheck`); fix the pnpm rewrite rule to respect custom script definitions.

> This looks like a **simple fix** that could be auto-resolved. Repo maintainers can use `/wshm fix` to attempt it.

<details>
<summary>📁 Relevant files</summary>

- `src/rewrite/pnpm.rs`

</details>

---
[...]

PR #3676, opened 2026-08-23
feat(discover): route pnpm's bare script forms

### LINKED ISSUE #3675: `pnpm test` / `pnpm build` are not routed, only `pnpm run <script>`
**Type:** Enhancement (coverage gap) · **Severity:** Low–Medium · **Area:** `src/discover/rules.rs`

## Summary

pnpm runs a bare script name directly — `pnpm test` is `pnpm run test` — and that shorter form is
what people actually type. The rule only matched `^pnpm\s+(exec|i|install|list|ls|outdated|run|run-script)`,
so `pnpm test`, `pnpm build`, `pnpm lint` fell through while `pnpm run build` was routed.

## A second, smaller problem

The rule had no trailing word boundary, so `pnpm installer` matched on the `install` alternative
and was rewritten. Adding `(\s|$)` fixes that.

## Not every script is safe to route

`pnpm dev`, `pnpm start`, `pnpm serve` and `pnpm watch` do not terminate. rtk captures child
output rather than streaming it, so routing a long-running script would leave the agent with a
blank screen until it gives up. Those are deliberately excluded.

## Environment

dwin-gharibi:
Closes #3675.

## Summary

- Routes pnpm's bare script forms: `test`, `build`, `typecheck`, `format`.
- Adds the missing `(\s|$)` word boundary — `pnpm installer` used to match the `install`
  alternative and get rewritten.
- **`lint` is deliberately excluded**: `pnpm lint` already routes to the dedicated eslint/biome
  filter (`rtk lint`), which parses the tool's output rather than just stripping package-manager
  boilerplate. Sending it to `rtk pnpm` would have been a downgrade.
- **Deliberately excludes `dev`, `start`, `serve`, `watch.`** Those never terminate, and rtk
  captures rather than streams, so routing them would hang the agent with no output. This is the
  one case where *not* adding coverage is the correct call.

## Test plan

### Unit tests (added)

`src/discover/registry.rs`, module `pnpm_bare_scripts`:

- `bare_script_names_are_rewritten` — the five safe scripts.
- `long_running_scripts_are_left_alone` — `dev`, `start`, `serve`, `watch` must return `None`.
  This is the test that keeps the hazard closed.
- `lint_keeps_its_dedicated_filter` — asserts `pnpm lint` still resolves to `rtk lint`.
- `existing_subcommands_still_rewrite` — `install`, `run build`, `outdated`, `ls`.
- `similar_names_are_not_captured` — `pnpm installer`, `pnpm testify`, `pnpm buildkite`.

### Behavioural check

Fake `pnpm` on `PATH`; failing output preserved and exit codes propagated:
`pnpm test` keeps `a.test.ts` + `1 failed` (exit 1); `pnpm build` keeps `dist/index.js` (exit 0).

Savings on these samples are ~0%: pnpm's own output is already terse, and the filter strips
package-manager boilerplate that small fixtures don't have. The value here is coverage
consistency and tracking, not compression — flagging that honestly rather than claiming a win.

## What the tests caught

The first draft included `lint` in the rule. The unit test failed with
`left: Some("rtk lint")`, `right: Some("rtk pnpm lint")` — the dedicated linter filter was already
winning, and adding `lint` here would only have created ambiguity between two rules. The
expectation was wrong, not the code; `lint` was dropped and a test now pins the existing routing.

### DIFF
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -79,7 +79,14 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^pnpm\s+(exec|i|install|list|ls|outdated|run|run-script)",
+        // pnpm accepts a bare script name (`pnpm test` == `pnpm run test`), and
+        // rtk pnpm forwards args verbatim, so these route correctly.
+        // Deliberately NOT dev/start/serve/watch: those don't terminate, and rtk
+        // captures rather than streams, so the agent would see nothing and hang.
+        // `lint` is absent on purpose: `pnpm lint` already routes to the dedicated
+        // eslint/biome filter (`rtk lint`), which knows the tool's output format.
+        // The trailing (\s|$) is new -- without it `pnpm installer` matched.
+        pattern: r"^pnpm\s+(exec|i|install|list|ls|outdated|run|run-script|test|build|typecheck|format)(\s|$)",
         rtk_cmd: "rtk pnpm",
         rewrite_prefixes: &["pnpm"],
         category: "PackageManager",
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1499,6 +1499,64 @@ mod tests {
     use super::super::report::RtkStatus;
     use super::*;
 
+    mod pnpm_bare_scripts {
+        use super::rewrite_command_no_prefixes;
+
+        /// pnpm runs a bare script name directly (`pnpm test` == `pnpm run test`).
+        #[test]
+        fn bare_script_names_are_rewritten() {
+            for cmd in ["pnpm test", "pnpm build", "pnpm typecheck", "pnpm format"] {
+                assert_eq!(
+                    rewrite_command_no_prefixes(cmd, &[]).as_deref(),
+                    Some(format!("rtk {cmd}").as_str()),
+                    "{cmd} should route to rtk pnpm"
+                );
+            }
+        }
+
+        /// Long-running scripts must NOT be routed: they never terminate, and rtk
+        /// captures output rather than streaming it, so the agent would hang with
+        /// nothing on screen.
+        #[test]
+        fn long_running_scripts_are_left_alone() {
+            for cmd in ["pnpm dev", "pnpm start", "pnpm serve", "pnpm watch"] {
+                assert_eq!(
+                    rewrite_command_no_prefixes(cmd, &[]),
+                    None,
+                    "{cmd} must not be captured by rtk"
+                );
+            }
+        }
+
+        /// `pnpm lint` is deliberately NOT in the pnpm rule: it already routes to
+        /// the dedicated eslint/biome filter, which understands the tool's output
+        /// format rather than just stripping package-manager boilerplate.
+        #[test]
+        fn lint_keeps_its_dedicated_filter() {
+            assert_eq!(
+                rewrite_command_no_prefixes("pnpm lint", &[]).as_deref(),
+                Some("rtk lint"),
+                "pnpm lint should reach the eslint/biome filter, not rtk pnpm"
+            );
+        }
+
+        #[test]
+        fn existing_subcommands_still_rewrite() {
+            for cmd in ["pnpm install", "pnpm run build", "pnpm outdated", "pnpm ls"] {
+                assert!(rewrite_command_no_prefixes(cmd, &[]).is_some(), "{cmd}");
+            }
+        }
+
+        /// The rule had no word boundary, so a longer word starting with a listed
+        /// subcommand matched.
+        #[test]
+        fn similar_names_are_not_captured() {
+            for cmd in ["pnpm installer", "pnpm testify", "pnpm buildkite"] {
+                assert_eq!(rewrite_command_no_prefixes(cmd, &[]), None, "{cmd}");
+            }
+        }
+    }
+
     fn rewrite_command_no_prefixes(cmd: &str, excluded: &[String]) -> Option<String> {
         super::rewrite_command(cmd, excluded, &[])
     }


src/discover/rules.rs
```diff
@@ -79,7 +79,14 @@ pub const RULES: &[RtkRule] = &[
...
     RtkRule {
-        pattern: r"^pnpm\s+(exec|i|install|list|ls|outdated|run|run-script)",
+        // pnpm accepts a bare script name (`pnpm test` == `pnpm run test`), and
+        // rtk pnpm forwards args verbatim, so these route correctly.
+        // Deliberately NOT dev/start/serve/watch: those don't terminate, and rtk
+        // captures rather than streams, so the agent would see nothing and hang.
+        // `lint` is absent on purpose: `pnpm lint` already routes to the dedicated
+        // eslint/biome filter (`rtk lint`), which knows the tool's output format.
+        // The trailing (\s|$) is new -- without it `pnpm installer` matched.
+        pattern: r"^pnpm\s+(exec|i|install|list|ls|outdated|run|run-script|test|build|typecheck|format)(\s|$)",
```
KuSh, line 89: **Blocking: `typecheck` is the one added name that is not a passthrough, and routing it writes files into the project.**

`PnpmCommands::Typecheck` is a real clap variant. On current `develop`, `main.rs:2096` dispatches it to `tsc_cmd::run(Some("pnpm"), &args, …)`, which runs `tsc` with only the args *after* `typecheck`. For a bare `pnpm typecheck` that arg list is empty, so the project's `typecheck` script never runs and any flags it carried are dropped.

Verified on the head merged into latest `develop`, real project, `"typecheck": "tsc --noEmit"`, `tsconfig.json` with `outDir: "./dist"`:

```
# pnpm 12.4.1
pnpm typecheck       → $ tsc --noEmit      exit 0, no dist/
rtk pnpm typecheck   → Running: pnpm tsc   exit 0, dist/index.js WRITTEN

# pnpm 9
pnpm typecheck       → exit 0, no dist/
rtk pnpm typecheck   → Running: tsc        exit 0, dist/index.js WRITTEN
```

`--noEmit` is dropped, so a read-only type check acquires a **write side effect** and emits compiled JavaScript into the working tree. On a project whose `typecheck` script is not tsc at all, the same path turned exit 0 with 1392 bytes of real output into exit 1 with a fabricated `TypeScript: No errors found`.

Broken by the same mechanism: `vue-tsc --noEmit`, `svelte-check`, `tsc -b`, `turbo run typecheck`, and `tsc -p tsconfig.app.json` (checks the wrong file set, so green is a false pass). With `tool_exec(…, MissingTool::Fetch)`, a project without `tsc` also gets a network fetch.

This is a regression against `develop`: today `pnpm typecheck` is not rewritten, so the hook leaves it alone. This change makes the hook do it automatically for everyone, turning the already-filed #2809 from opt-in into default.

The other three names are fine — `test`, `build` and `format` all reach `PnpmCommands::Other` → `run_passthrough`, byte-identical to raw pnpm with exit codes propagated, verified on pnpm 9 and 12.4.1.

```suggestion
        pattern: r"^pnpm\s+(exec|i|install|list|ls|outdated|run|run-script|test|build|format)(\s|$)",
```

`"pnpm typecheck"` also needs dropping from `bare_script_names_are_rewritten`, which currently asserts the broken routing is correct. With both changes the suite is green and `pnpm typecheck` goes back to being left alone.

KuSh (changes requested):

**Claim**: Route pnpm's bare script forms (`test`, `build`, `typecheck`, `format`) through `rtk pnpm`, and add the missing `(\s|$)` word boundary.

**Scope**: **accept if narrowed to** "route pnpm's bare `test`, `build` and `format`, and add the missing `(\s|$)` word boundary" — `typecheck` must come out. (round 1, frozen)

**Ran**: real pnpm **9** and **12.4.1** (the Rust rewrite) against two real projects; `rtk rewrite` × 28 command forms × `develop` 79b96a44 vs head f0f6042c; flag forms (`--silent`, `--prod`, `-- --watch=false`, `--help`); a streaming probe; two mutation tests; `fmt`/`clippy`/`cargo test --all` on the head (2630 pass) and on the head **merged into latest develop** (3541 pass, clean merge). Savings on the newly-routed forms: **0%**, byte-identical — which your PR body states honestly.

**CI** @f0f6042c: **never run**. Only `license/cla` is green; `check-target` is SKIPPED because the fork's workflow is waiting for maintainer approval. I'll approve the run once the blocker is addressed.

Thank you for the PR description — the `lint` exclusion, the long-running exclusion and the honest 0%-savings note all made this much faster to review, and the "what the tests caught" section is exactly the right instinct.

1. `src/discover/rules.rs:89` — **routing a script name into a non-passthrough rtk subcommand**. `typecheck` is the one added name that collides with a real clap variant, so `rtk pnpm typecheck` runs `tsc` instead of the project's script, drops `--noEmit`, and writes `dist/index.js` into the working tree. Verified on pnpm 9 and 12.4.1. Regression against `develop`, and it makes #2809 fire by default. Details and a one-word fix in the line comment.

2. **`rules.rs:93` — `test`/`build`/`format` report 80% savings but deliver 0%.** All three reach `run_passthrough`, which inherits stdio and filters nothing. With `subcmd_status: &[]` they classify as `RtkStatus::Existing` ("dedicated handler with filtering") and `discover` credits them the rule's 80%. They also leave the *unsupported* list — erasing the exact signal that #3876 depends on. The fix mirrors the cargo rule one entry above: `subcmd_status: &[("test", Passthrough), ("build", Passthrough), ("format", Passthrough)]`, which zeroes the savings automatically.

3. **`rules.rs:84` — the stated rationale for excluding dev/start/serve/watch is not accurate.** "rtk captures rather than streams, so the agent would see nothing and hang" isn't true of this path: `run_passthrough` inherits all three stdio handles. Verified — a script printing then sleeping 10s, killed at 3s, showed its early line. Excluding those four is still a fine conservative call; it's the *reason* that would mislead the next maintainer. The same text is repeated in the test doc at `registry.rs:1517`.

4. **`rules.rs:88` — changelog text in a comment.** "The trailing `(\s|$)` is new -- without it `pnpm installer` matched" describes a prior revision; house style is to state the invariant instead, e.g. "anchors the subcommand at a word boundary so `pnpm installer` does not match".

5. **`registry.rs:1544` — `existing_subcommands_still_rewrite` asserts only `is_some()`**, so it can't distinguish "rewrote correctly" from "rewrote to the wrong rtk command". The sibling test two functions up already uses `assert_eq!(…, Some("rtk …"))`.

- `test`, `build`, `format` → `PnpmCommands::Other` → passthrough. Byte-identical to raw pnpm on **both pnpm 9 and 12.4.1**, exit codes propagated, `--silent` / `--prod` / `-- --watch=false` / `--help` all forwarded verbatim with no clap interception.
- The `(\s|$)` fix is real: `pnpm installer` rewrote to `rtk pnpm installer` on `develop`, no longer does.
- `pnpm lint` → `rtk lint` holds. The mechanism is `matches.last()` at `registry.rs:169` ("last match wins"), with the eslint/biome rule at a higher index — your reasoning is right, though note it's the index and not the alternation that decides it.
- No later rule shadows the new names (`next` / `prettier` / `vitest` / `jest` all require their literal tool name).
- Capture-group numbering is safe: `(\s|$)` is group 2; `subcmd_savings` / `subcmd_status` read group 1.
- Both new tests are genuine, proven by mutation: adding `dev` to the pattern fails `long_running_scripts_are_left_alone`; removing `(\s|$)` fails `similar_names_are_not_captured`.
- pnpm 12 specifics: `test` and `start` are genuine built-ins there, while `build` / `typecheck` / `format` resolve through the bare-script fallback, which v12 still supports — so your premise holds on the Rust rewrite too. (Unrelated to this PR: `run-script` no longer exists in pnpm 12, so that alternative is dead on v12.)
- Merges cleanly into latest `develop`; full suite green after merge.

### Filed as follow-ups

- **#2809** — absorbed: [comment added](https://github.com/rtk-ai/rtk/issues/2809#issuecomment-5649144451) with the `dist/index.js` emission transcript and the dropped-`--noEmit` mechanism there rather than opening a duplicate.
- **#4009** — filed: 23 rules on `develop` end in a bare subcommand alternation with no boundary; this PR fixes the pnpm one, 22 remain, with 16 demonstrated over-matches including `git branchless` (a real tool) → `rtk git branchless`. Untouched code, so an issue rather than a finding here — nice catch surfacing the class.
- `pnpm --filter app test` / `-F` stay unrouted — absorbed by existing #3858, not refiled.

- *The 20% savings floor blocks these 0%-savings forms.* Dropped — the floor governs filters; these are passthroughs that add tracking.
- *`pnpm format` is shadowed by the prettier rule.* Dropped — that rule requires the literal `prettier`.
- *`pnpm test:unit` should route.* Dropped — `:` correctly falls outside `(\s|$)`.
- *`long_running_scripts_are_left_alone` is vacuous because it also passes on the parent commit.* Dropped — mutation shows it does catch a future addition of `dev`, which is what it claims to guard.

### Next

Drop `typecheck` from the alternation and from `bare_script_names_are_rewritten`; optionally take items 2–5 in the same push. Then I'll approve the CI run and, once it's green, approve the PR. That is the whole remaining set.

Rounds: 1/3. Threads: 1 open (the `typecheck` blocker), 0 resolved this round.

PR #3680, opened 2026-08-23
fix(vitest): don't bury vitest's subcommands under an injected run

### LINKED ISSUE #3679: `rtk vitest` prepends `run` in front of vitest's own subcommands
**Type:** Bug · **Severity:** Low–Medium (wrong tests run) · **Area:** `src/cmds/js/vitest_cmd.rs`

## Summary

`build_vitest_effective_args` unconditionally starts the argument vector with `run`. vitest has
subcommands of its own (`related`, `bench`, `list`, `typecheck`, `init`), and placing them after
`run` changes what they mean: `vitest run related src/a.ts` reads `related` as a test-file filter,
not as the `related` subcommand.

## Reproduction

With a shim on `PATH` recording vitest's argv:

| Typed | Executed before |
|---|---|
| `vitest related src/a.ts` | `vitest run --reporter=json related src/a.ts` |
| `vitest bench` | `vitest run --reporter=json bench` |
| `vitest --watch` | `vitest run --reporter=json` (`--watch` dropped, silently) |

The rewrite rule (`src/discover/rules.rs`) routes bare `vitest …` to `rtk vitest`, so this reaches
production.

## Impact

`vitest related <files>` is the "run the tests affected by my change" command. Under `run` it
becomes "run tests whose path matches the string `related`" — usually nothing. The agent sees a
green, empty run and concludes the change is covered.

Dropping `--watch` is the right *outcome* — a watcher would hang a non-interactive agent — but
doing it silently means the agent cannot tell that its request was changed.

## Expected behaviour

- A real vitest subcommand stays the leading token.
- Plain invocations still get `run`, so an agent never lands in watch mode.
- When a watch request is rewritten, say so.

## Environment

dwin-gharibi:
Closes #3679.

## Summary

- `rtk vitest` no longer buries vitest's own subcommands under `run`. `related`, `bench`, `list`,
  `typecheck` and `init` stay the leading token; everything else still gets `run` so an agent
  never lands in watch mode.
- Watch requests (`watch`, `dev`, `--watch`) are still rewritten to `run` — a watcher would hang a
  non-interactive agent — but the substitution is now reported on stderr instead of being silent.

## Test plan

`src/cmds/js/vitest_cmd.rs`:

- `vitest_subcommands_are_not_buried_under_run` — each subcommand stays leading and no `run` is
  injected. Red on the old behaviour with
  `related must stay the leading subcommand, got ["run", "--reporter=json", "related", "src/a.ts"]`.
- `plain_invocation_still_gets_run` — the plain and file-argument forms keep `run`.
- `explicit_run_is_not_duplicated`.
- `watch_is_replaced_by_run_and_reported` — watch still becomes `run`, and the substitution is
  now observable via `dropped_watch`.

### Behavioural check with a `PATH` shim recording vitest's argv

| Typed | Before | After |
|---|---|---|
| `vitest related src/a.ts` | `run --reporter=json related src/a.ts` | `related --reporter=json src/a.ts` |
| `vitest bench` | `run --reporter=json bench` | `bench --reporter=json` |
| `vitest list` | `run --reporter=json list` | `list --reporter=json` |
| `vitest typecheck` | `run --reporter=json typecheck` | `typecheck --reporter=json` |
| `vitest` | `run --reporter=json` | unchanged |
| `vitest src/a.ts` | `run --reporter=json src/a.ts` | unchanged |
| `vitest run src/a.ts` | `run --reporter=json src/a.ts` | unchanged |
| `vitest --watch` | `run --reporter=json`, silent | same argv **+ stderr notice** |
| `vitest --coverage` | `run --reporter=json --coverage` | unchanged |
| `vitest -t mytest` | `run --reporter=json -t mytest` | unchanged |

### DIFF
--- a/src/cmds/js/vitest_cmd.rs
+++ b/src/cmds/js/vitest_cmd.rs
@@ -209,6 +209,9 @@ pub fn run_test(command: &Commands, args: &[String], verbose: u8) -> Result<i32>
             let mut cmd = package_manager_exec(framework);
             let effective_args = build_vitest_effective_args(args);
             passthrough_requested = effective_args.passthrough;
+            if effective_args.dropped_watch {
+                eprintln!("rtk: vitest watch mode replaced with `run` (non-interactive)");
+            }
             cmd.args(effective_args.args);
             (framework, cmd)
         }
@@ -269,6 +272,9 @@ pub fn run_test(command: &Commands, args: &[String], verbose: u8) -> Result<i32>
 struct EffectiveVitestArgs {
     args: Vec<String>,
     passthrough: bool,
+    /// Set when a watch request was rewritten to `run`, so the substitution can
+    /// be reported rather than applied silently.
+    dropped_watch: bool,
 }
 
 struct FormattedTestOutput {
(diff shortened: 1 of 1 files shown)


amandeavor (commented):
I think `typecheck` should be removed from `VITEST_SUBCOMMANDS`.

In Vitest 4.1.11, the documented commands are `run`, `related`, `watch`, `dev`, `bench`, `init`, and `list`. Typechecking is enabled with the `--typecheck` option; a positional `typecheck` is still treated as a test-file filter. I confirmed this against the current CLI help.

With this patch, `vitest typecheck` changes from `vitest run --reporter=json typecheck` to `vitest typecheck --reporter=json`. That drops the non-interactive `run` mode even though no real subcommand was selected, which can put the command back into Vitest's default watch behavior.

Could you remove `typecheck` from the list and add a case asserting that it remains under `run` as a positional filter? The handling for the actual subcommands looks good.

Issue #3663, opened 2026-08-23
`rtk grep` claims `-h`, `-l`, `-m` and `-t`, which belong to grep/rg


dwin-gharibi:
**Type:** Bug · **Severity:** Critical (silent wrong result / fabricated output) · **Area:** `src/main.rs`, `src/cmds/system/search.rs`

## Summary

`Commands::Grep` declares its own short options that collide with the engine it proxies:

| rtk option | short | what the letter means to the proxied tool |
|---|---|---|
| `--max-len` | `-l` | grep `--files-with-matches` |
| `--max` | `-m` | grep `--max-count` |
| `--file-type` | `-t` | rg `--type` |
| *(clap's automatic help)* | `-h` | grep `--no-filename` |

When clap can parse the following token as the option's value, it consumes both silently and the
engine never sees the flag. When it cannot, clap errors and `run_fallback` runs the raw command —
which is why the bug looks intermittent.

### `-h` returns rtk's help instead of searching

```bash
printf 'GET /a 404\nGET /a 200\n' > a.log
printf 'GET /b 404\n' > b.log

grep -h 404 a.log b.log
# GET /a 404
# GET /b 404

rtk grep -h 404 a.log b.log
# Compact grep - strips whitespace, truncates, groups by file
#
# Usage: rtk grep [OPTIONS] [EXTRA_ARGS]...
#   ... exit 0
```

The same applies to `rtk ls -h`, where `-h` is `ls --human-readable`.

Clusters (`-hn`, `-lh`) are unaffected — only the bare flag is claimed by clap.

### `-l` disappears when the pattern is numeric

```bash
grep -l 404 a.log b.log       # a.log \n b.log   (exit 0)
rtk  grep -l 404 a.log b.log  # no output        (exit 1)
```

`-l 404` is parsed as `--max-len 404`, so the remaining positionals shift: `a.log` becomes the
pattern and `b.log` the only path. Searching `b.log` for the literal string `a.log` finds nothing.

Control: `grep -l GET a.log b.log` is correct, because `GET` fails to parse as a `usize`, clap
errors, and the fallback runs the raw command. The bug fires exactly when the operand looks like a
number — status codes, years, ports, PIDs, error codes.

### `-m N` is dropped

```bash
printf 'hit1\nhit2\nhit3\nhit4\n' > m.txt
grep -m 1 hit m.txt       # 1 line
rtk  grep -m 1 hit m.txt  # 4 lines
```

[...]

PR #3681, opened 2026-08-23
feat(core): add arg_tokenizer and migrate git/grep/dotnet/golangci arg parsing onto it

### LINKED ISSUE #3710: `git worktree repair` is routed to list mode: prints the worktree list, exits 0, repairs nothing
## Summary

`run_worktree`'s `has_action` set is missing `repair`, so `git worktree repair <path>` falls through to list mode. rtk runs `git worktree list` instead, prints the filtered list, and returns 0. The repair never runs and nothing says so.

Same masking class as #2497, but the cause is different: there the read path swallowed a git failure, here a write subcommand never reaches git at all.

Found through the Claude Code PreToolUse hook, which rewrites `git worktree repair ...` to `rtk git worktree repair ...`. The agent read the list output as a successful repair and moved on with two still-broken worktrees.

## Reproduce

```console
$ git init -q A && cd A && git commit -q --allow-empty -m init
$ git worktree add -q -b wt A/.worktrees/wt
$ cd .. && mv A B && cd B          # moving the repo breaks both gitdir pointers

$ cat B/.worktrees/wt/.git
gitdir: /abs/path/A/.git/worktrees/wt          # stale, points at the old location

$ rtk git worktree repair /abs/path/B/.worktrees/wt
/abs/path/B                       9d3c6f8 [main]
/abs/path/A/.worktrees/wt         9d3c6f8 [wt] prunable
$ echo $?
0
$ cat B/.worktrees/wt/.git
gitdir: /abs/path/A/.git/worktrees/wt          # unchanged

$ git worktree repair /abs/path/B/.worktrees/wt
repair: gitdir incorrect: /abs/path/B/.git/worktrees/wt/gitdir
$ cat B/.worktrees/wt/.git
gitdir: /abs/path/B/.git/worktrees/wt          # repaired
```

rtk 0.45.0, macOS.

## Cause

`src/cmds/git/git.rs:2143`

```rust
let has_action = args.iter().any(|a| {
    a == "add" || a == "remove" || a == "prune" || a == "lock" || a == "unlock" || a == "move"
});
```

git has eight `worktree` subcommands: `add`, `list`, `lock`, `move`, `prune`, `remove`, `repair`, `unlock`. Six are listed, `list` is the fallback, `repair` is the only one missing.

I verified the other six empirically: `add`, `remove`, `remove --force`, `lock`, `unlock`, `move` and `prune` all take effect. `repair` is the only broken one.

## Two more defects at the same site

**List mode drops the caller's flags.** Line 2179 hardcodes the invocation:

```rust
```

So `git worktree list --porcelain` returns the compact filtered format, not porcelain. Anything parsing that output gets a shape it never asked for:

```console
$ rtk git worktree list --porcelain
/abs/path/B    9d3c6f8 [main]        # not porcelain
```

This is arguably worse than the `repair` case, since porcelain exists precisely to be machine-read.

**`has_action` matches the token in any position.** `args.iter().any(...)` tests every argument, not the subcommand slot. A worktree whose directory is named `move` or `remove` selects the action branch regardless of the actual subcommand.

## Fix shape

- [ ] Add `repair` to the `has_action` set, or invert it: treat everything except `list` (and no subcommand) as pass-through, so the next git release does not reopen this.
- [ ] In list mode, forward the caller's arguments instead of hardcoding `["worktree", "list"]`, and skip filtering when a format flag such as `--porcelain` or `-z` is present.
- [ ] Match the subcommand at `args[0]` rather than anywhere in `args`.

## Workaround

`exclude_commands = ["git worktree"]` in the rtk config keeps the rest of the git proxying while removing the hazard.

KuSh:
## What this adds

`src/core/arg_tokenizer.rs` — one place where RTK turns an argv slice into typed tokens, replacing the ad hoc `starts_with('-')` scans and `find("--")` walks that each command module had grown its own copy of.

A `Token` carries its kind (`Long`, `Short`, `Positional`, `DashDash`), its flag name, its attached value (`--flag=v`), a link to its separate-token value (`--flag v`), and its index in the original argv, so a caller can always rebuild the exact arg a token came from.

The tokenizer never guesses which flags take a value. The caller supplies the grammar as a `takes_value(kind, name) -> Option<ValueSpec>` predicate scoped to *that tool and that subcommand*, and everything else follows from it: a flag's value is linked to the flag instead of counted as a positional, `--` is a real boundary, and a positional that looks like a flag stays a positional.

One table answers every question the tokenizer has about a flag, rather than one predicate per question — that shape is how two lists start drifting apart, which is a bug class this PR fixes three times elsewhere.

- `ValueSpec::value()` — `--flag=v` or `--flag v`, and a literal `--` is the boundary. The common case.
- `ValueSpec::attached_only()` — `--flag=v` only; the next argument is never the value (git's `-M`/`-U`/`-C`/`-B` take an optional attached number and nothing else).
- `ValueSpec::solo_only()` — a `Short` flag takes a separate value only when it is the whole argument: `git log -n 2` does, `git log -pn 2` does not.
- `.claiming_dash_dash()` — lets a literal `--` be this flag's value. A per-*tool* split, not per-flag: grep and rg let any value-taking flag swallow it, git and cargo reject it whichever flag is asking.

Two entry points, split by whether the caller has a grammar at all: `tokenize(args)` for the callers asking only which arguments are flags and where `--` is, and `tokenize_grammar(args, takes_value, dialect)` for the rest. Two dialects — `Posix` (`-abc` clusters, `--flag`) and `Msbuild` (`/flag:value`, for `dotnet`).

Helpers on top: `before_dashdash`, `dashdash_index`, `injection_point`, `is_digit_run`, `Token::value`, `Token::is_free_positional`.

## What it changes

`git.rs`, `search.rs` (grep/rg), `dotnet_cmd.rs` and `golangci_cmd.rs` are migrated onto it. Every fix below has a regression test that fails without it, and every claim about a tool's own behaviour was checked against that tool (git 2.53, GNU grep 3.12, ripgrep 15.2, .NET SDK 9 and golangci-lint 2.13.1 via Docker) rather than assumed.

**A `--` boundary is a boundary.** `restore_double_dash` now runs once at git's dispatcher instead of per-subcommand, and every flag lookup that should stop at `--` does. A pathspec, worktree path or project file spelled like a flag is no longer parsed as one.

**A flag's value is not a positional.** `dotnet test --filter X` no longer reads `X` as a project path; `git branch -u origin/main` links its upstream; `rg -r X TODO f` no longer loses the pattern to `--replace`; `grep -f patterns.txt` reads its pattern file instead of searching for it.

**One grammar per tool, and per subcommand.** grep and rg get separate value-taking tables transcribed from each engine's own `--help`, because the overlap is smaller than it looks: `-r` is `--recursive` in grep and `--replace` in rg, `-h` is `--no-filename` in grep and `--help` in rg, `-I` is the reverse, and `-N`/`-T`/`-z`/`-L`/`-E` all diverge. Within git, `log` and `diff` likewise get their own: `git log -cl 2` is an error while `git diff -wl 100` is valid, and `-u` means `-p` in `git log` but `--include-untracked` in `git stash show`.

**Output-shape detection reads tokens, not reconstructed strings.** `git log -U<n>`, `--cc`, `-c` and `--remerge-diff` request a patch and now pass through raw instead of being compacted into `[+N lines omitted]`; `-W` does not request one and no longer dumps the unfiltered history. `--check`, `--exit-code` and `--quiet` keep their reports instead of being thrown away by RTK's own stat probe. `git worktree repair` reaches git instead of being answered with a worktree listing, and `worktree list --porcelain` passes through instead of being replaced.

**Injection checks recognise every spelling of what they look for.** `dotnet --report-trx:true`, `/l:trx` and `-l trx` all count as "already present", so RTK stops double-injecting a logger; `/flag:value` MSBuild switches can no longer swallow the next token as a value; golangci's output-flag check covers the full set of sink flags and is scoped to run-level ones.

## Design rule

The rule the migration kept running into is documented in `src/core/README.md`: **one grammar per tool and per subcommand — never reuse a sibling's predicate because it looks close enough.** `.claude/rules/rust-patterns.md` gains a matching rule 6, so new commands, new flags and fixes to arg handling go through the tokenizer rather than adding another bespoke scan.

## What is not in this PR

- **Modules still on raw string scans**: `src/cmds/system/ls.rs`, `src/cmds/rust/cargo_cmd.rs`, `src/cmds/go/go_cmd.rs`, `src/cmds/python/ruff_cmd.rs`, `src/discover/registry.rs`. Same bug class, pre-existing, each its own follow-up. `go_cmd.rs` needs an `OsString` conversion first, since the tokenizer is deliberately `AsRef<str>`-only.
- **`ValueSpec` carries no semantic tag for what a value *means*** — only how it attaches. A `kind` field (`Path`, or a `MaybePath` that asks for a filesystem check) drops in as one more field when a consumer appears, which is the headroom the struct was shaped for. Nothing reads such a tag today: every path-ish use in these four modules is keyed on a *role* — golangci's output sinks, search's pattern file, dotnet's results directory — and a generic `Path` tag would be strictly less precise than the name-keyed checks it would replace.
- **`Dialect::Msbuild` folds ASCII case for every flag name**, but its only consumer mixes two grammars: MSBuild switches really are case-insensitive, dotnet's own options are not. A wrong-case spelling is rejected by dotnet either way, so the only cost is RTK suppressing an injection for a command that will fail regardless.
- **`Token` exposes the prefix as two bools** (`double_dash`, `slash`) rather than the three-state enum it holds internally, so consumers spell "single dash" as a double negative.
- **Some modules tokenize twice per invocation.** Known waste, not a regression — startup is 3.8 ms release.
- **MSBuild single-segment absolute paths** (`/app`, `/tmp`) are structurally indistinguishable from a switch name. Real MSBuild resolves this with a filesystem `stat()`; this tokenizer is a pure function and does not.
- **`dotnet_takes_value`'s allowlist is subcommand-unaware** and is what `explicit_projects` relies on: a value-taking flag missing from it whose value ends in `.csproj` would be read as an explicit project path.
- **`--diff-merges`' routing is static.** `log` sends it to the raw passthrough on the flag alone, because a predicate cannot know whether the revision walk contains a merge; on a merge-free walk that costs compaction for a flag which provably changes git's output by zero bytes. The alternative — compacting a combined diff whose two marker columns `compact_diff` misreads — is silent corruption rather than verbosity, so this is the deliberate side to err on.
- **`git show` repeats its summary line** under `--oneline`, and once per parent under `--diff-merges`' single-column formats, because `show_cmd` puts RTK's flags ahead of the user's and git repeats the header per parent even under `--no-patch`. Both are pre-existing and byte-identical on the merge-base; both are cosmetic against a compacted body.

## Issues

- Fixes #3710 — `git worktree repair` is routed to list mode. Verified: the stale gitdir pointer is now actually repaired.
- Refs #2275 — bare `git log -p` already passed through before this branch, but `-U<n>`, `--cc`, `-c` and `--remerge-diff` were still being compacted into `[+N lines omitted]`. They now pass through too.
- Refs #2391 / #3663 — the display side of `-h`/`--no-filename` (and rg's `-I`) is implemented here and works for every spelling clap does not intercept. The bare `-h` still never reaches the module; that needs `disable_help_flag` on the `Grep`/`Rg` variants, which is a separate UX decision.

## Test plan

- [x] `cargo fmt --all --check`, `cargo clippy --all-targets` (zero warnings), `cargo test --all` (3074 unit tests plus every integration suite)
- [x] Every fix has a regression test verified to fail before it and pass after
- [x] Differential testing against a binary built at the merge base, repeatedly, across ~90–430 `git`/`grep`/`rg` invocations per round, including fuzzed flag combinations compared byte-for-byte against the real engines
- [x] Flag semantics verified against the real tools listed above, not from documentation or memory
- [x] Focused fuzz campaign on the two ecosystems that needed a container to test: 1,890 argument vectors compared against a merge-base binary for what RTK forwards (every divergence is one of this PR's intended fixes), then real-tool matrices in Docker under `LC_ALL=C` — golangci-lint 2.13.1 on clean and issue-producing modules, .NET SDK 9 on passing and failing test projects, and .NET SDK 10 for MTP-native. Exit codes match the real tool in every case, including failures; no regressions found


src/cmds/git/git.rs
```diff
@@ -1890,6 +1907,29 @@ fn format_stash_message(subcommand: Option<&str>, result: &CaptureResult) -> Str
...
+/// false-positive (misread some option's own value as -p/-u), never false-negative the way
+/// reusing log's grammar just did (silently swallowing a real following -p as --author's value,
+/// then rendering full patch content through the stat-only compact_stash_stat path instead of
+/// compact_diff). Only the patch-mode flags matter here, not the full requests_raw_diff_shape
+/// list: --stat et al. correctly stay on the compact_stash_stat branch, not compact_diff.
+fn stash_show_wants_patch(args: &[String]) -> bool {
+    let tokens = arg_tokenizer::tokenize(args, &|_kind, _name| false);
+    tokens.iter().any(|t| match t.kind {
+        TokenKind::Long => t.text == "patch",
+        TokenKind::Short => matches!(t.text, "p" | "u"),
```
aeppling: "u" is --include-untracked in stash show, not patch

https://git-scm.com/docs/git-stash#Documentation/git-stash.txt-show-u--include-untracked--only-untrackeddiff-optionsstash

aeppling (changes requested):

aeppling:
Hey @KuSh 

Thanks for working on this which is a pain for commands rewirtting.
Some regressions found:

#### repo with a stash that includes an untracked file:

rtk git stash show -u
develop: f.txt 1 + / u.txt 1 + / 2 changed 2 + exit 0
PR: (empty) exit 0
git: 2-file stat

#### file f containing a -- b:

rtk grep -e -- f </dev/null
develop / grep: a -- b exit 0
PR: (empty) exit 1

#### directory with 4 files containing foo:

rtk rg -z foo </dev/null
develop / rg: f:foo one … (filenames)
PR: foo one … (filenames dropped)

src/core/arg_tokenizer.rs
```diff
@@ -0,0 +1,1030 @@
...
+    ///
+    /// The still-unseen boundary `--` can never be swallowed as a value (verified against real
+    /// git: `git log --grep -- pattern` fails with "Option '--grep' requires a value" rather
+    /// than treating `--` as the pattern) -- once the boundary has already been emitted, a later
+    /// `--` is just ordinary text and fair game (`git log -- -- pattern` works).
+    fn link_next_value(&mut self, flag_index: usize, value_index: usize) -> bool {
+        let Some(next) = self.args.get(value_index) else {
+            return false;
+        };
+        if next.as_ref() == "--" && !self.emitted_dash_dash {
```
aeppling: GNU grep does not follow that rule. grep -e -- file means "pattern is the literal string --".

(man grep):  -e PATTERNS, --regexp=PATTERNS — Use PATTERNS as the patterns. … This option can be used to protect a pattern beginning with "-".

This should be scoped to commands so each respect its own behavior

src/cmds/dotnet/dotnet_cmd.rs
```diff
@@ -739,59 +790,134 @@ fn detect_test_runner_mode(args: &[String]) -> TestRunnerMode {
...
-                    return true;
-                }
-            }
-        }
-    }
+/// Strict lookup (the safe default -- prefer this name): only a literal `--flag` matches, never
+/// `-flag`/`/flag`. Required for every dotnet_cmd.rs flag that isn't a legacy MSBuild.exe
+/// passthrough switch -- confirmed against a real dotnet 9 SDK (Docker) that `dotnet format`'s
+/// `--verify-no-changes`/`--report` and `dotnet test`'s `--logger`/`--results-directory` are
+/// double-dash-only, and a single-dash or slash spelling doesn't just get rejected, it gets
```
aeppling: in many places there is more comment than code, please remove most, code should be explicit by design and if not, a single or few lines are often enough.

comments add maintenance and also can misslead ai agent, so we prefer to avoid much

aeppling:
Also we may want to document this system in respective README.md

src/cmds/git/git.rs
KuSh: fixed

src/core/arg_tokenizer.rs
KuSh: that was not just -e but any grep/rg arg that takes value. fixed

src/cmds/dotnet/dotnet_cmd.rs
KuSh: I've tried to reduce comments. Let me know if this isn't enough

KuSh:
@aeppling took all of this into account

aeppling:
I re-ran a fuzz test batch against raw and develop, some new findings.

Each row reads: real tool / develop / PR.

Most of these look like the same shape: the tokenizer is now correct, and some consumers were tuned to the old parse. B1 is separate, the stat/patch split leaves the patch child's exit status unchecked.

## B1 · `diff`/`show` exit 0 on commands git refuses
`git.rs:306`, `:466`, `:478`

```
rtk git diff --unified=x      129 / 129 / 0
rtk git diff -Uabc            129 / 129 / 0
rtk git diff -mp              129 / 129 / 0
rtk git show -pq HEAD         128 / 128 / 0
rtk git show -mp HEAD         128 / 128 / 0
```

Dirty tree also prints a well formed diffstat alongside that exit 0.

Stat child at `:280` is checked and succeeds, because `args_without_patch_shape` (`:171`) strips the offending flag from it. Patch child at `:306` is gated on `!stdout.is_empty()` only.

Control, still correct on both: `diff --stat=x` 129/129/129, `diff --color-moved=bogus` 129/129/129, `diff --dirstat=bogus` 128/128/128.

## B2 · `--line-prefix <value>` drops the patch body
`git.rs:1140`, render at `:819`

```
rtk git diff --line-prefix --stat HEAD~1 HEAD    333B / 333B / 95B
```

PR emits a diffstat plus an empty `Changes:` section. Attached form `--line-prefix=XX` is 277B/83B/83B, so that one is pre-existing and not this.

## B3 · `grep -T` on stdin leaks injected flags
`search.rs:475`, guard at `:673`

```
printf 'alpha x\n' | rtk grep -T alpha
  real     alpha x
  develop  alpha x
  PR       (standard input)^@                  1:	alpha x
```

Leaks the injected `-H`, a raw NUL, and `-n`. `--initial-tab` leaks on both, so the PR widens the trigger to short `-T`.

## B4 · `checkout -B<branch>` glued drops `(new)`
`git.rs:1988` feeding `:1913`

```
rtk git checkout -Bfoo    "Switched to a new branch" / ok foo (new) / ok foo
```

## B5 · golangci short cluster ending in `c` loses filtering
`golangci_cmd.rs:39`

```
rtk golangci-lint -Egosec run
  develop  rc 0   forwards: -Egosec run --out-format=json
  PR       rc 1   forwards: -Egosec run
```

`c` is read as `--config`, which swallows `run`. Argv stays faithful, and the PR propagates the exit 1 that develop swallowed.

KuSh:
Thanks — that batch was worth a lot. All five are fixed, and chasing them turned up three more of the same shape. Head is now `f3593ecd`.

## Your five

**B1 — `diff`/`show` exit 0 on commands git refuses.** Your diagnosis was exact: the stat child succeeds because `args_without_patch_shape` strips the offending flag from it, and the patch child's status was never read. The child that runs the user's arguments verbatim is git's verdict on the command, so its failure now propagates with its stderr. All five of your cases match raw git again, and your three controls (`--stat=x`, `--color-moved=bogus`, `--dirstat=bogus`) still do.

**B2 — `--line-prefix` drops the patch body.** It prefixes every line including the `diff --git` and `@@` markers the compaction keys on, so nothing parsed. It now takes the raw route, which also fixes the attached spelling you measured as pre-existing (`--line-prefix=XX` was 83B against raw 277B; both are byte-equal now).

**B3 — `grep -T` leaks the injected flags.** Same shape — `--initial-tab` re-shapes every match line, so the forced `-H --null -n` parse read nothing back. Both spellings are now treated as shape flags for grep only, so the long form you saw failing on develop is fixed too, and rg's `-T` stays `--type-not`.

**B4 — `checkout -B` loses `(new)`.** Real, and worth a note on why it took a second look: I could not reproduce your develop result at first, because this machine runs a French git. Develop only printed `ok foo (new)` for you because it *failed* to parse the glued `-Bfoo` and fell through to the English output scan, which hits on your machine and misses on mine. Parsing `-Bfoo` correctly is what moved it onto the argument path, where an early return claimed the name before reading what git actually did. `-B` creates *or* resets and only git knows which, so the arguments are now the fallback rather than a short-circuit past the scan. The underlying locale fragility is not specific to checkout — `run_checkout` is one of the unpinned spawns in #3839 — so I added the details there rather than fixing it here.

**B5 — golangci `-Egosec run`.** Confirmed against 2.13.1 in Docker: it accepts `-Egosec run` and rejects `-vc <file> run` with `unknown shorthand flag: 'c' in -c`, so a value-taking shorthand never takes a separate value inside a cluster. The tokenizer already had a way to say that (`solo_only`, added for `git log -n`), so this was a one-line grammar correction rather than new machinery.

## Three more of the same shape, found while verifying yours

- The raw-shape `git log` passthrough capped every walk at 10 — including one the user had already bounded, so `git log -p HEAD~15..HEAD` returned 10 of the 15 commits it named. The cap stays (that route streams to the terminal and an unbounded `-p` printed 1,800 commits) but an explicit revision range is now left alone, and when the cap does apply it says so on stderr.
- `run_show` read `-s`/`--no-patch`/`--quiet` order-independently where git resolves them against `-p` by last flag wins, so `git show -s -p` printed no diff. `never_worse` then kept the smaller summary, so the patch vanished without trace.
- `rtk git show --oneline` printed its summary line twice: `--oneline` outranks the stat step's `--pretty=format:`, whose job is suppressing that header.

## Not fixed here, filed instead

Things this branch does not cause and that are better fixed on develop:

- **#3870** — `rtk golangci-lint` swallows the failure three ways: `exit 1` (issues found) mapped to 0, stderr dropped by `stdout_only()`, and a hard error printed as `No issues found`. Measured against 2.13.1 in Docker; identical on develop.
- **#3843** — the same failure-path defect in `run_log`, `run_fetch`, `run_branch` and `run_status`, in the other direction: the exit code survives and the already-produced stdout is discarded.
- **#3842** — `--output-indicator-{new,old,context}` belongs to the same family as `--color=always` and `--line-prefix`: it replaces the marker column the compaction counts, so the body survives and only the numbers lie.
- **#3792** — still reproduces on develop and 0.46.0: `git log --graph` drops the merge commit while keeping its `|\` and `|/` rails, so the graph contradicts itself.
- **#3839** — the `run_checkout` locale details from B4.

## Verification

Since golangci-lint and dotnet are the two ecosystems that need a container, I ran a focused campaign on them rather than only reading:

- 1,890 argument vectors compared against a merge-base binary for what RTK forwards. 300 diverge, and every one is one of this PR's intended fixes — boundary preservation, duplicate-injection suppression, scoped verbosity, mode-aware results directory. No exit-code changes, no dropped boundaries.
- Real-tool matrices under `LC_ALL=C`: golangci-lint 2.13.1 on clean and issue-producing modules; .NET SDK 9 on passing *and failing* test projects; .NET SDK 10 for MTP-native, which nothing had exercised before. Exit codes match the real tool in every case including failures, and there is no TRX left in the project directory.

That last one also caught a regression of this branch's own that your batch did not reach: in Classic/VSTest mode, `dotnet test -- --results-directory X` was reading the app's flag as dotnet's own, suppressing RTK's injection and leaving the TRX that RTK itself asked for sitting in the user's project, uncleaned. It is now read per runner mode.

`cargo fmt --all && cargo clippy --all-targets && cargo test --all` clean, 3182 tests.

aeppling:
Re-ran the fuzz batch on `f3593ecd`. Previous five blockers all fixed. Three regressions against develop below, one of them new this round.

Setup:

```
git init -q -b main . && git config user.email t@t.t && git config user.name T
printf 'l1\nl2\nl3\n' > a.txt && git add -A && git commit -qm c1
printf 'l1\nl2 CHANGED\nl3\n' > a.txt
```

Rows read: real / develop / PR.

**M1 · stderr notice on the raw log path. New in `f3593ecd`.** `git.rs:1083`

```
git log --stat 2>&1 >/dev/null
  (nothing) / (nothing) / [rtk] showing 10 commits; pass -n <count> for more

git log -pq
  rc 128 "fatal: unrecognized argument: -pq"
  PR: same rc, notice printed first
```

Repo has 2 commits. Only `--stat` and `-p` trigger it; plain `log`, `--oneline`, `-n2`, `--graph`, `--format` do not.

**M2 · `git diff --color` plus another flag drops the patch body.**

```
git diff --color --unified=0 | grep -c '@@'
  1 / 1 / 0
```

PR prints the stat header and an empty `Changes:` section. `--color` alone loses it on both.

**M3 · wrong exit code and wrong error when a second argument also fails.**

```
git diff -Uabc nonexistent-ref
  129 error: --unified expects a numerical value
  129 same
  128 fatal: ambiguous argument 'nonexistent-ref'
```

`-Uabc` alone is correct on all three.

KuSh:
@aeppling all three fixed, plus a fourth that chasing them turned up. Head is `72722904`, CI green.

## M3 · wrong exit code and wrong error when a second argument also fails

Your row was the whole diagnosis. `run_diff` ran its stat probe first and let *that* report failure — but the probe runs with the patch-shape flags stripped, so it answers a different command than the one typed: `-Uabc` is gone by then, and the probe got as far as the unknown ref and said `ambiguous argument` (128) instead of git's `--unified expects a numerical value` (129).

The user's own arguments now run first and their result is the verdict; the probe is decoration, and a failure there costs the header rather than the command. Your row matches raw git again, each argument alone still reports its own error (129 and 128 respectively), and the three controls from your earlier batch — `--stat=x`, `--color-moved=bogus`, `--dirstat=bogus` — are unchanged.

## M2 · `--color` plus another flag drops the patch body

Colour puts an escape at column 0, which is exactly where the compaction looks for `diff --git` and `@@`, so nothing matched and the body came back empty. Fixed by stripping ANSI before the compaction parses — RTK renders its own output, so the colour was never going to survive compaction anyway and nothing is lost by removing it first.

That covers more than the row you filed. `--color` alone, which you noted loses the body on develop too, is fixed; so is `--color=always`; and so is the `color.ui = always` config form, where the escapes arrive with no flag on the command line at all and no amount of argument inspection could have seen them coming. That closes #3842.

## M1 · the stderr notice, new in `f3593ecd`

Mine, and wrong in both the ways you show: it announced a truncation that had not happened on a two-commit repo, and announced it ahead of commands git then rejected outright.

The notice exists because that route streams, so there is no footer or tee to notice missing commits from. But when the limit applies the output is *bounded by it*, which makes that run cheap to capture — so it is captured, and the notice prints afterwards, only when git really produced the full limit, and never when git failed. A non-default `--format` makes the commit count unreadable; RTK then says nothing rather than guessing.

## A fourth, from fuzzing `show` against raw git

Since this area has taken most of the findings on this branch, I fuzzed it harder rather than only fixing your three: 2,000 combinations across five seeds, comparing real git / develop / branch on exit code, first stderr line and body presence. It found a regression I introduced last round:

```
git show --quiet -p    -> patch
git show -p --quiet    -> patch
git show -s -p         -> patch
git show -p -s         -> no patch
```

`--quiet` loses to a patch request from *either* side; only `-s`/`--no-patch` genuinely fold last-flag-wins. I had modelled all three as one group, which dropped a patch the user had asked for — silently, because `never_worse` then kept the smaller summary. Thirteen rows measured against git 2.53 now pin the difference, including the mixed spellings (`--quiet -s -p` prints, `-p -s --quiet` does not) where the two rules meet.

After the fix, the only thing those 2,000 combinations still flag is develop's `--name-only` / `--name-status` / `--check` failure, which this branch fixes.

## Tests

All four are now integration tests against the real binary in a temp repo with `LC_ALL=C`, in CI rather than behind `#[ignore]` — every one of them was a regression that the suite as it stood reported as green.

## One structural note, filed rather than done here

Probing git properly for the `--quiet` fix turned up something worth writing down, so it is in #3910 with the measurements rather than in this PR.

RTK routes by asking per-token predicates "does any token match"; git resolves output shape by two different rules. Only suppressors are order-sensitive — 40 ordered pairs per subcommand change shape when reversed and every one has `-s`/`--no-patch` on one side, identical across diff, show and log. Among shape flags order never matters, which means a fixed precedence rather than recency; testing "last flag wins" against 343 random combinations predicts only 207, with rows like `--name-only --compact-summary` → NAMES and `--check --raw --shortstat` → NONE.

Neither rule is expressible as set membership, which is why each combination in this family has had to be found empirically. I deliberately did not stack that work on this PR — it would keep this branch in flux, and most of the predicates it would rewrite are this branch's own, so it wants `develop` afterwards. The precedence lattice needs measuring before anyone implements it; #3910 says so explicitly.

aeppling:
LGTM.

Re-ran the full fuzz batch against raw commands and develop tip (f849ef3c). 30,716 binary invocations plus 4,400,008 tokenizer property cases.

| Surface | Cases | New regressions | Fixes |
|---|---|---|---|
| git render (raw / develop / PR) | 8,652 | 0 | 8 exit-code, 58 bloat |
| search render (raw / develop / PR) | 5,592 | 0 | 43 |
| argv passthrough vs raw exec | 15,704 | 0 | 153 |
| crash / hang / perf | 768 | 0 | 2 |
| tokenizer property | 4,400,008 | 0 | n/a |

Panics 0, hangs 0, timeouts 0, signal loss 0, bloat 0. fmt clean, clippy 0 warnings, 3,214 unit tests pass.

aeppling (approved):

PR #3687, opened 2026-08-24
fix(diff): keep +/- markers at column 0 in condense_unified_diff

kylehgc:
Companion to #3646, which fixed the same class in `compact_diff`.

## Bug

The stdin diff path (`git diff | rtk diff -`, `run_stdin` → `condense_unified_diff`) emitted changed lines as `format!("  {}", c)`. The `+`/`-` markers landed at column 2, so anchored greps (`grep -E '^[+-]'`) over the output matched nothing — a silent false negative for scripts and CI gates built on diff conventions.

## Fix

Emit change lines at column 0 by moving them with `result.append(&mut changes)` — they already carry their markers, and the move also drops a per-line `String` clone. `[file] name (+N -M)` headers and the `... +N more` overflow line are unchanged; neither starts with `+`/`-` on real git output, so anchored greps see the change lines.

## Verification

- Repro before: `printf '<unified diff>' | rtk diff - | grep -cE '^[+-]'` → **0**. After: **4** on a two-file diff, all markers at column 0.
- The regression test uses a **two-file** fixture deliberately: the function flushes changes at two separate sites (once per `+++` for the preceding file, once after the loop for the last). Mutation testing showed a single-file fixture lets the first site revert with the suite still green; the two-file fixture kills both mutants.
- The anti-indent assertion is scoped to change lines (`" +"`/`" -"`) because the `... +N more` trailer is legitimately indented.
- Quality gate green: `cargo fmt --all && cargo clippy --all-targets && cargo test --all`, 0 failures.

## Not in scope

Deeper classification issues in this function (e.g. a removed line whose content is `-- note` arrives as `--- note` and is dropped as a header) are pre-existing, reproducible, and left untouched here — this PR only fixes the emission indentation and its test coverage.

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -164,7 +164,7 @@ fn condense_unified_diff(diff: &str) -> String {
     let mut current_file = String::new();
     let mut added = 0;
     let mut removed = 0;
-    let mut changes = Vec::new();
+    let mut changes: Vec<String> = Vec::new();
 
     // Never truncate diff content — users make decisions based on this data.
     // Only strip diff metadata (headers, @@ hunks); all +/- lines shown in full.
@@ -174,7 +174,8 @@ fn condense_unified_diff(diff: &str) -> String {
                 if !current_file.is_empty() && (added > 0 || removed > 0) {
                     result.push(format!("[file] {} (+{} -{})", current_file, added, removed));
                     for c in &changes {
-                        result.push(format!("  {}", c));
+                        // Column 0: anchored greps (`^[+-]`) must match these.
+                        result.push(c.clone());
                     }
                 }
                 current_file = line
@@ -198,7 +199,8 @@ fn condense_unified_diff(diff: &str) -> String {
     if !current_file.is_empty() && (added > 0 || removed > 0) {
         result.push(format!("[file] {} (+{} -{})", current_file, added, removed));
         for c in &changes {
-            result.push(format!("  {}", c));
+            // Column 0: anchored greps (`^[+-]`) must match these.
+            result.push(c.clone());
         }
     }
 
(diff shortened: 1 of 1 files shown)


kylehgc:
Note for reviewers on the amendment (`b7f389b`), pushed after self-review of the original commit:

**What changed and why:**

1. **The regression test now uses a two-file fixture.** `condense_unified_diff` flushes a file's changes at two separate sites — per `+++` for the preceding file, and after the loop for the last one. The original single-file fixture only exercised the second: mutation testing (reverting the in-loop site alone back to the indenting `format!`) left the entire suite green. With two files, that mutant fails the test.
2. **The third assertion is scoped to change lines.** The blanket `!l.starts_with(' ')` form was false in general — the `... +N more` trailer is indented, so the assertion only held for diffs of ten or fewer changes.
3. **`result.append(&mut changes)` replaces the per-line `clone()`**, and the `Vec<String>` annotation is gone. The original PR description justified that annotation as replacing inference lost with the removed `format!` — that claim was wrong (`format!` accepts any `Display` and contributes no type information; inference comes from `changes.push(line.to_string())`). The description above has been corrected accordingly.

No behavior change to the fix itself; base-vs-head output is identical to the original commit on every input we tested.

KuSh (approved):
Approving. Verified in depth; the change is correct and the one thing I found was in the test rather than the fix.

**The fix:** `result.append(&mut changes)` is behaviourally equivalent to the old borrow-and-`format!` loop minus the indent — `append` drains `changes`, and the retained `changes.clear()` still covers the branch-not-taken path (empty `current_file`, or zero added/removed). Type inference resolves `changes` to `Vec<String>` via `push(line.to_string())`, so dropping the annotation is fine. The `[file] … (+N -M)` header never starts with `+`/`-`, so anchored greps see exactly the change lines and nothing else. `never_worse` is unaffected — output shrinks by two bytes per line. `condense_unified_diff` has one caller (`run_stdin`), which prints it, so nothing downstream re-parses this.

Confirmed live on 0.46.0 that this is a real user-facing failure, not a theoretical one:

```
$ cat u.diff | rtk diff - | grep -c '^-'
0                     # a diff with one deletion
```

**Two commits pushed to your branch**, shout if you'd rather have written either differently:

1. **Rebased onto `develop`** — the branch was 73 commits behind. One conflict, in both flush sites: #2652 landed in the meantime and removed the `... +N more` trailer that your `append` hunks carried as context. Resolved by keeping your `append` and dropping the trailer, which is what both changes independently wanted.

2. **`fc2cc72` — made the anti-indent assertion able to fail.** `l.starts_with(" +") || l.starts_with(" -")` matches a *one*-space indent, but the form it guards against is two spaces (`"  -fn old() {}"`), so it passed on exactly the code it rejects. Only the `l == want` equality loop above it was killing the mutant. It now matches on leading whitespace generally. Verified by neutralising the equality loop and reverting both flush sites: the assertion fails on its own, at `diff_cmd.rs:420`.

   Your comment carved the assertion narrowly because the `  ... +N more` trailer was legitimately indented — correct reasoning at the time, and it's what made the one-space form look deliberate. #2652 removed that trailer, so the carve-out is no longer needed.

I also dropped the `(#3646)` reference from the test comment: that PR is closed in favour of #3269, which makes the same fix to `compact_diff`. #3269 is still open, so `rtk git diff` keeps the two-space indent for now — this PR fixes the stdin path only, and the two are independent (different functions, different files, no ordering constraint between them).

Thanks for the two-file fixture, and for spelling out in the PR comment *why* one file wasn't enough. Reasoning about which flush site a fixture actually reaches is the part most regression tests skip.

PR #3688, opened 2026-08-24
fix(discover): preserve pipelines when splitting chains

### LINKED ISSUE #3683: discover: split_command_chain's stop-at-pipe truncation blinds classification to everything after |
Found by \`/code-review max\` while iterating on #3206 (pre-existing behavior, not introduced by that PR — it inherits the limitation, doesn't cause it — so flagged separately rather than folded in).

\`discover::registry::split_command_chain\` calls \`lexer::split_on_operators(cmd, true)\` with \`stop_at_pipe=true\`, which returns only the segment(s) *before* the first \`|\` — by design, per its own doc comment ("used by command rewriting — only the left side of a pipe gets rewritten"). \`discover::run()\`'s classification loop uses this same function to split transcript commands into parts for counting/classification, so a command like \`cargo test | tail -50\` becomes just \`["cargo test"]\` — the \`| tail -50\` half is silently dropped from the scan entirely (not counted in \`total_commands\`, not classified, invisible to every reporting bucket including the new hook-decision coverage estimate).

Separately, this also means \`estimate_hook_coverage\`'s registry-rewrite check can diverge from the real hook's own pipe-safety veto (\`RewriteContext::PipelineFinal\` in the real \`rewrite_command\` path) for the *first* segment too, since discover evaluates \`"cargo test"\` in isolation rather than the full \`"cargo test | tail -50"\` line the real hook sees — the real hook may correctly refuse to rewrite because of what follows the pipe, while discover's isolated-segment view has no way to know that.

**Likely fix**: discover's classification loop needs its own pipe-aware handling — either treat a piped command as a single opaque unit for counting purposes (matching what the real hook actually evaluates), or explicitly model the "rewrite only applies pre-pipe, but the whole line is one attestable unit" distinction the real hook already has. This touches how \`rtk discover\` counts/classifies commands generally, not just the hook_coverage estimate, so it's a bigger change than a coverage-estimate fixup.

k4its1t:
## Summary

- keep a complete pipeline as one discovery counting/classification unit instead of truncating at the first pipe
- continue splitting independent clauses on `&&`, `||`, and `;`, including clauses after pipelines
- cover plain pipelines, mixed chains, quoted operators, `|&`, and the shared session-counting path

Closes #3683

## Latest upstream sync

Merged develop at 5e10aa6 normally. Preserved upstream permission handling and parser helpers, retaining pipeline-as-one-unit analytics. Updated the new segmenter consistency test only for the intended analytics behavior, keeping its permission and rewrite assertions intact. Updated the corresponding module documentation.

Validation on Rust 1.98.0: cargo fmt --all --check, cargo clippy --all-targets, focused regressions (including 69 permission tests), and cargo test passed: 3,210 passed / 8 ignored.

## Test plan

- [x] `cargo test --all`
- [x] Targeted pipeline regressions (RED before the implementation, green after)
- [ ] Manual testing: not needed for this lexer/counting-only change

## AI assistance

OpenAI Codex assisted with issue analysis, implementation, tests, and PR preparation.


k4its1t:
Resolved conflicts by merging develop in d3e4c43; no rebase or force-push. Pipeline counting is retained alongside upstream parser helpers and conservative permission segmentation. The newly added upstream pipe_then_and characterization still expected truncation at the first pipe, so only its analytics expectation was updated; permission and rewrite assertions are unchanged. Format, Clippy, all 3,210 tests (8 ignored), and targeted permission/pipeline regressions passed. Final PR scope is three original code files plus the corresponding module README. Codex assisted with resolution and verification.

k4its1t:
Post-push check: the branch is mergeable and license/cla is successful. GitHub CI for this update is action_required, awaiting maintainer approval to run; I have not bypassed or approved that gate. The local format, Clippy and full test results above are complete.

k4its1t:
Merged current develop (f749546) with a normal merge, resolving the sole import conflict by retaining upstream redirect_has_file_target alongside split_on_chain_operators. Permission/rewrite behavior is unchanged; the PR remains scoped to the original four files. Rust 1.98: cargo fmt --all --check, cargo clippy --all-targets, and cargo test passed (3,330 unit tests plus 108 integration tests; 8 ignored). The PR diff against develop passes git diff --check. AI-assisted conflict resolution; no force push.

k4its1t:
Merged latest develop (b748a5f7) normally in e58b03df and resolved the import-only conflict. Pipeline counting remains separate from permission/rewrite segmentation; the PR still touches only its original four files. Format, Clippy, and 3,735 unit tests passed (8 ignored). The full run reproduced the two existing Git integration failures documented in the updated description; no unrelated fixes were added. AI-assisted conflict resolution and verification.

PR #3689, opened 2026-08-24
fix(read): preserve context across truncated source

### LINKED ISSUE #3421: `rtk read --max-lines N` (and the `head` rewrite) drops lines from the middle with no gap marker, producing code that compiles-looking but means the opposite
`smart_truncate` keeps structurally interesting lines and discards the rest, but it joins the survivors with no marker where the gap was, and appends a single `[N more lines]` at the end — which implies the omission is at the end. The result is code an agent reads as complete and correct while it says something different.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Reproduced with the shipped binary; input and output below are verbatim.*

### Reproduction

`auth.rs`, 15 lines:

```rust
 1  fn can_admin(user: &User) -> bool {
 2      if !user.is_admin {
 3          return false;
 4      }
 5      if user.is_banned {
 6          return false;
 7      }
 8      if user.mfa_ok {
 9          return true;
10      }
11      false
12  }
13  fn wipe_database() {
14      drop_all();
15  }
```

`rtk read auth.rs --max-lines 10`:

```rust
fn can_admin(user: &User) -> bool {
    if !user.is_admin {
        return false;
    }
    if user.is_banned {
    }
    }
}
fn wipe_database() {
[6 more lines]
```

Two things happened:

1. **The ban check was inverted.** `if user.is_banned {` survived; its `return false;` did not. What the agent reads is a guard that does nothing — the opposite of what the file says. Nothing marks the removal.
2. **`fn wipe_database()` was spliced in** directly after, with no indication that lines were skipped between them. The only marker, `[6 more lines]`, sits at the bottom and reads as "the rest of the file follows", not "content was removed from the middle".

An agent asked to review this file would conclude that banned users pass the check.

### Why this is on a default path

This is not an exotic flag. `rtk rewrite head -50 file` emits `rtk read file --max-lines 50`, so the hook routes ordinary `head` invocations through the same code (`src/core/filter.rs:342`, reached from `src/cmds/system/read.rs`). The default filter level is `none`, and this happens at that level.

### Related, same function

- **`src/cmds/system/read.rs:157`** — with `-n`, the line numbers count the **kept** lines 1..N, not the file's real line numbers. In the run above, the row labelled 7 is raw line 9, and the `[N more lines]` marker itself gets numbered as though it were content. An agent told to "fix line 8" edits a different line than the one it was shown.
- **`src/core/filter.rs:349`** — `kept_lines >= max_lines - 1` underflows when `max_lines == 0`. `Cargo.toml:45` sets `panic = "abort"` but not `overflow-checks`, so a release build wraps to `usize::MAX` and the loop never breaks (a debug build panics with `attempt to subtract with overflow`). Reachable as `head -0 file` → `--max-lines 0`.

- **Emit a marker where the gap is**, not only at the end — e.g. `… 4 lines omitted …` inline at each discontinuity. That single change makes the output honest even if the selection heuristic stays as is.
- Consider keeping a line's closing brace only when its opening block was kept intact; dropping a body while keeping both braces is what makes the output read as valid.
- Guard `max_lines == 0` explicitly rather than relying on `- 1`.

I would rate the inline gap marker as the important one: an agent can reason correctly about "content was removed here", but it cannot detect a guard clause that was silently emptied.

### One thing I could not reproduce

An earlier note in my own review suggested `rtk read -` (stdin) could return zero bytes for a small error log. I tried it — a 147-byte tsc-style log came back as 146 bytes, exit 0. So that half does not reproduce and I am not claiming it.

k4its1t:
## Summary

- preserve smart-truncation omission markers and source line numbers
- merge current develop while retaining its byte-faithful head/tail windows and zero-budget behavior
- retain the stdin empty-filter fallback and the original four integration regressions
- add a focused regression for numbered head output

The head rewrite and zero-budget handling are now also present upstream; the remaining contribution is visible gaps, numbering, and stdin fallback. This does not replace native head/tail slicing with smart selection.

## Verification (current merge)

Passed:
- cargo fmt --all --check
- cargo clippy --all-targets --locked
- read unit tests: 28 passed, 3 ignored
- filter unit tests: 13 passed
- read_integrity_test: 4 passed

cargo test --locked was run but did not fully pass: guard_integration_test had two failures, both reproduced individually:
- git_diff_dash_dash_stat_pathspec_after_double_dash_is_not_stat_flag
- git_log_malformed_digit_run_propagates_real_git_error

The Git command sources and that test file are identical to upstream/develop. This machine uses Apple Git 2.39.5; the latter assertion expects "not an integer", while Git emits "not a non-negative integer". No unrelated Git code or assertions were changed. Suites after that failure were not reached.

## AI assistance

OpenAI Codex assisted with the implementation, conflict resolution, and local verification. The commands and results above were observed locally; they are not a claim of a fully green suite.

Closes #3421

### DIFF
--- a/src/cmds/system/read.rs
+++ b/src/cmds/system/read.rs
@@ -38,16 +38,16 @@ pub fn run(
 
     // Apply filter
     let filter = filter::get_filter(level);
-    let mut filtered = filter.filter(&content, &lang);
+    let filtered = filter.filter(&content, &lang);
 
     // Safety: if filter emptied a non-empty file, fall back to raw content
-    if filtered.trim().is_empty() && !content.trim().is_empty() {
+    let (filtered, fell_back) = preserve_nonempty_input(&content, filtered);
+    if fell_back {
         eprintln!(
             "rtk: warning: filter produced empty output for {} ({} bytes), showing raw content",
             file.display(),
             content.len()
         );
-        filtered = content.clone();
     }
 
     if verbose > 0 {
@@ -64,15 +64,16 @@ pub fn run(
         );
     }
 
-    filtered = apply_line_window(&filtered, max_lines, tail_lines, &lang);
-
     let (raw, rtk_output) = if line_numbers {
         (
             format_with_line_numbers(&content),
-            format_with_line_numbers(&filtered),
+            apply_numbered_line_window(&filtered, max_lines, tail_lines, &lang),
         )
     } else {
-        (content.clone(), filtered.clone())
+        (
+            content.clone(),
+            apply_line_window(&filtered, max_lines, tail_lines, &lang),
+        )
     };
     let shown = never_worse(&raw, &rtk_output);
     print!("{}", shown);
@@ -116,7 +117,14 @@ pub fn run_stdin(
 
     // Apply filter
     let filter = filter::get_filter(level);
-    let mut filtered = filter.filter(&content, &lang);
+    let filtered = filter.filter(&content, &lang);
+    let (filtered, fell_back) = preserve_nonempty_input(&content, filtered);
+    if fell_back {
+        eprintln!(
+            "rtk: warning: filter produced empty output for stdin ({} bytes), showing raw content",
+            content.len()
+        );
+    }
 
     if verbose > 0 {
         let original_lines = content.lines().count();
@@ -132,21 +140,30 @@ pub fn run_stdin(
         );
     }
 
-    filtered = apply_line_window(&filtered, max_lines, tail_lines, &lang);
-
     let (raw, rtk_output) = if line_numbers {
         (
             format_with_line_numbers(&content),
-            format_with_line_numbers(&filtered),
+            apply_numbered_line_window(&filtered, max_lines, tail_lines, &lang),
         )
     } else {
-        (content.clone(), filtered.clone())
+        (
+            content.clone(),
+            apply_line_window(&filtered, max_lines, tail_lines, &lang),
+        )
     };
     let shown = never_worse(&raw, &rtk_output);
     print!("{}", shown);
 
     timer.track("cat - (stdin)", "rtk read -", &raw, shown);
     Ok(())
+}
+
+fn preserve_nonempty_input(content: &str, filtered: String) -> (String, bool) {
+    if filtered.trim().is_empty() && !content.trim().is_empty() {
+        (content.to_string(), true)
+    } else {
+        (filtered, false)
+    }
 }
 
 fn format_with_line_numbers(content: &str) -> String {
@@ -183,6 +200,33 @@ fn apply_line_window(
     }
 
     content.to_string()
+}
+
+fn apply_numbered_line_window(
+    content: &str,
+    max_lines: Option<usize>,
+    tail_lines: Option<usize>,
+    lang: &Language,
+) -> String {
+    if let Some(tail) = tail_lines {
+        if tail == 0 {
+            return String::new();
+        }
+        let lines: Vec<&str> = content.lines().collect();
+        let start = lines.len().saturating_sub(tail);
+        let width = lines.len().to_string().len();
+        let mut output = String::new();
+        for (index, line) in lines.iter().enumerate().skip(start) {
+            output.push_str(&format!("{:>width$} │ {}\n", index + 1, line));
+        }
+        return output;
+    }
+
+    if let Some(max) = max_lines {
+        return filter::smart_truncate_numbered(content, max, lang);
+    }
+
+    format_with_line_numbers(content)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


HetCreep (commented):
Built `fix/read-honest-truncation` at `b93dd68` and ran the full verification independently (release build, `cargo test --bin rtk`, `cargo test --all`, plus the original `#3421` reproducer by hand).

**All three core problems from #3421 are genuinely fixed:**

- Gap markers now appear at every real discontinuity, not just at the end. Re-running the `auth.rs` / `--max-lines 10` example from the issue: the `is_banned` guard's dropped `return false;` now shows `[1 line omitted]` right where it was removed, instead of silently vanishing into `if user.is_banned { }`.
- `-n` now prints the file's real line numbers for kept lines (omission markers left unnumbered), not a sequential kept-line count.
- `--max-lines 0` returns cleanly (empty output, exit 0) — no underflow, no hang.
- `cargo test --bin rtk`: 2623 passed, 0 failed, 8 ignored. `cargo test --all`: 17 passed, 0 failed, including the new `read_integrity_test.rs` (`truncation_marks_removed_guard_body_at_the_gap`, `numbered_truncation_uses_source_line_numbers`, `zero_line_window_is_empty_without_panicking`, `stdin_falls_back_when_filter_removes_nonempty_input`).

Nice work — the inline-marker approach is exactly what the issue asked for, and the regression coverage is real (I ran it, not just read it).

**One thing worth flagging for @rtk-ai maintainers before merge, not a correctness objection:** this PR's diff is wider than its own description states ("keep the change scoped to `read`/`smart_truncate`"). Alongside the `read.rs`/`filter.rs` changes, it also touches:

- `src/cmds/git/git.rs` (+466/-…) — a `git log -- <path>` fix citing #1215 (a real, closed issue — the `trailing_var_arg`/`--` bug apparently never got the same fix `git log` that `git diff` got). Comes with its own regression test (`git_log_dash_p_pathspec_after_double_dash_is_not_patch_flag`), which I also ran — passes.
- `src/core/tee.rs` (+45/-…) — reworks slug truncation for tee recovery filenames (readable-prefix + hash instead of a hard 40-char cut, to avoid collisions on long embedded paths).
- `scripts/benchmark.sh` (+8/-…) — fixes `rm -rf "$BENCH_DIR"` deleting tracked benchmark-harness files on every local run.

Each of those looks like a legitimate, independently-motivated fix with its own justification in the diff — I'm not saying any of them is wrong. But bundled together with #3421 in one PR, a maintainer reviewing this has to sign off on four unrelated concerns at once, and if any one of them needs discussion the whole PR (including the already-verified #3421 fix) sits blocked. Might be worth splitting `git.rs`/`tee.rs`/`benchmark.sh` into their own PRs so the #3421 fix — which is fully verified and ready — isn't held hostage to review of the other three.

*(Disclosure, matching this PR's own AI-assistance note: this review was carried out with AI (Claude) assistance, human-directed — I built the branch and ran every command above myself rather than trusting the PR description.)*

k4its1t:
Thanks for independently running the reproducer and full test suites. I rechecked the scope concern against the actual PR head and base: commit `b93dd68` has parent `29f9bb7`, and both `git show 29f9bb7..b93dd68 --stat` and GitHub’s PR Files view contain exactly three paths: `src/cmds/system/read.rs`, `src/core/filter.rs`, and `tests/read_integrity_test.rs`. The PR currently reports 3 changed files (359 additions, 64 deletions). `src/cmds/git/git.rs`, `src/core/tee.rs`, and `scripts/benchmark.sh` are not part of this commit or PR diff; they are already present in the base history, so there is nothing from those concerns to split out here.

HetCreep:
You're right, and I was wrong — sorry for the noise. Root cause on my end: my local `develop` branch was stale at `ba7a9ce` (from an earlier sync in this session) while `origin/develop` had moved to `29f9bb7`. I diffed against the stale local branch instead of the real tip, so the diff pulled in everything that landed on `develop` between those two points — none of which is yours.

Re-ran it properly against `origin/develop` (`29f9bb7`, which is in fact `b93dd68`'s actual parent): 3 files, 359 insertions, 64 deletions — `src/cmds/system/read.rs`, `src/core/filter.rs`, `tests/read_integrity_test.rs`. Matches what you posted exactly. The `git.rs` change I flagged came from PR #3575 (`fix/git-log-grep-value-misdetected-as-patch-flag`), already merged into `develop` before your branch forked — nothing to do with this PR. Same story for `tee.rs`/`benchmark.sh`.

Retracting the scope note entirely — this PR is scoped exactly as described. The functional verification (build, `cargo test --bin rtk` 2623/0 failed, `cargo test --all` 17/0 failed, and the three #3421 repros) stands unchanged; that part didn't depend on the bad diff.

k4its1t:
Merged develop b748a5f7 normally in 9ab008fe. Kept upstream chunked-head reading alongside the existing numbered truncation function. The only extra test adjustment is the numbered-tail expectation (source line 4, not renumbered line 1). Format, Clippy, 3,742 unit tests, and all 9 targeted read integration tests passed. The full run still hits the two previously documented, unrelated Git integration assertions; details are in the refreshed description. No new tests or refactors added. AI-assisted resolution and verification.

PR #3698, opened 2026-08-24
fix: emit trailing newline from rtk find output

### LINKED ISSUE #3659: `rtk find` output has no trailing newline, so line counters are off by one

VedantMadane:
## Summary

fix: emit trailing newline from rtk find output

## Changes

- print! left the last path without a terminating newline, so wc -l under-counted by one.
- Match find(1) by ensuring non-empty output ends with a newline.

Fixes #3659

### DIFF
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -373,7 +373,13 @@ pub fn run(
     }
 
     let shown = never_worse(&raw_output, &body);
-    print!("{}", shown);
+    if shown.is_empty() {
+        // no matches
+    } else if shown.ends_with('\n') {
+        print!("{}", shown);
+    } else {
+        println!("{}", shown);
+    }
     timer.track(
         &format!("find {} -name '{}'", path, effective_pattern),
         "rtk find",


KuSh:
Closing this in favour of #3660, which makes the same newline fix in src/cmds/system/find_cmd.rs and adds a unit test covering the missing, already terminated and empty cases. Review or comments are welcome over there. If you had a case in mind that #3660 does not cover, say so here and it can be reopened.

PR #3704, opened 2026-08-25
refactor(lexer): consolidate raw-command lexing behind shared primitives

KuSh:
## Summary

The lone-CR bug found by `/code-review` on #3600 appeared in two independent places (`registry.rs::rewrite_multiline_block` and `permissions.rs::command_matches_pattern`) because there was no single "parse a raw shell command string" implementation — at least five (`tokenize_inner`, `QuoteScan`, `shell_split`, `split_token_spans`, `permissions.rs`'s ad hoc `split_whitespace()`), plus three compound-command segmenters on top. This consolidates layer 1 (raw string → quote/operator/newline-aware words and segments) so such a fix lands once. Layer 2 (`core/arg_tokenizer.rs`, argv → flags/values) is untouched — that's #3681.

### What changed

1. **Root fix**: word boundaries follow bash's default `$IFS` (exactly space/tab/newline) via one shared `is_word_boundary_whitespace`, not `char::is_whitespace()` (Unicode `White_Space`, wrongly includes NBSP) or a per-caller `split_whitespace()`. A lone `\r` is no longer a word boundary, nor a command boundary on the bash-accurate path. `permissions.rs::command_matches_pattern` shares the predicate — it previously collapsed an embedded CR into a space, letting an allow-rule for `git status` auto-approve `git status\rrm -rf ~`.
2. **One quote/escape rule**: `tokenize_inner`, `QuoteScan` and `shell_split` each tracked "am I inside a quote"; all three now drive `advance_quote_state`.
3. **One CRLF rule**: `tokenize_inner`'s newline guard and `registry.rs`'s `raw_breaks` parity check share `is_crlf_at`.
4. **One word-coalescing rule**: `split_token_spans` (golangci-lint flags) and `shell_split` only need "was there a space here", not full shell-syntax tokens (`*.yml` is one bash word but `Shellism("*")` + `Arg(".yml")`). New `coalesce_words()` merges gapless adjacent tokens; both functions are now thin wrappers over `tokenize()` instead of bespoke scanning loops.
5. **Newline policy is explicit, not implicit**: `tokenize_inner` takes a `NewlineMode` (`None`/`Bash`/`Conservative`). Only `split_for_permissions` uses `Conservative`, where a lone `\r` *is* a segment boundary — keeping the gate's segmentation identical to `develop`'s. Living inside the shared char loop, it stays quote-aware for free.
6. **Second root fix, same shape as the first**: inside double quotes bash only lets `\` escape `` $ ` " \ `` or a newline — before anything else it is a literal character. `shell_split` treated `\` as an escape everywhere outside single quotes, so a quoted Windows path lost its separators (`"C:\Program Files\rtk.exe"` → `C:Program Filesrtk.exe`). Now bash-accurate, which fixes that path for every caller at once (`hooks/mod.rs` hook detection, `rtk proxy` argv construction, `registry.rs::search_uses_pattern_file`) rather than per call site. Unquoted backslashes still escape, as bash does. Found while reviewing #3552.
7. **Documented and pinned** the three deliberately-different compound-command segmenters (permission gate / analytics / rewrite): cross-referencing doc comments, a comparison table, a `segmenter_consistency` test module, and a `discover/README.md` section — the lexer is consumed by `hooks/permissions.rs`, `hooks/mod.rs` and `main.rs`, so it's shared infrastructure, not a `discover/` internal.

### Not in scope

- `core/arg_tokenizer.rs` and its call sites — layer 2, #3681.
- Unifying `QuoteScan` onto the token stream: it needs quote state at an arbitrary byte, including *inside* a token (is the `#` in `foo#bar` a comment?), which `Vec<ParsedToken>` doesn't carry — it would re-scan every token anyway. It already shares the rule that was actually duplicated.
- Unifying the three segmenters further: they already consume the same token stream and differ only in boundary *policy*, which is deliberate (item 7).
- `hooks/rewrite_cmd.rs` vs `hooks/hook_cmd.rs` decision-flow duplication — real, but a separate follow-up at a different level.

## Test plan

- [x] `cargo fmt --all --check`, `cargo clippy --all-targets` (zero warnings), `cargo test --all` (2724 tests)
- [x] Regression test per behavior: lone-CR word splitting (`tokenize`, `shell_split`, `command_matches_pattern`), NBSP splitting, permission-gate lone-CR segmentation inside and outside quotes, quoted and unquoted golangci-lint flag values, backslash handling in and out of double quotes, the segmenter-consistency table, coalesced-word `shell_split`

### DIFF
--- a/src/discover/README.md
+++ b/src/discover/README.md
@@ -39,6 +39,21 @@ When a hook sends `cargo fmt --all && cargo test 2>&1 | tail -20`:
 
 **Result**: `rtk cargo fmt --all && rtk cargo test 2>&1 | tail -20`. Bash handles the `&&` and `|` at execution time — each `rtk` invocation is a separate process.
 
+## Shared Lexer Toolkit
+
+`lexer.rs` is layer 1 of RTK's command parsing (raw string → quote/operator-aware tokens and words) — layer 2, already-split argv → flags/values, is `core/arg_tokenizer.rs`. It's not private to this module: `hooks/permissions.rs`, `hooks/mod.rs`, and `main.rs`'s `rtk proxy` all build on it instead of re-scanning commands themselves.
+
+| Function | Purpose | Used outside `discover/` by |
+|---|---|---|
+| `tokenize(cmd)` | Full shell-syntax tokens: quotes, escapes, operators, pipes, redirects, shellisms | — |
+| `tokenize_with_newlines(cmd)` | Like `tokenize`, plus a `\n` `Operator` token per unquoted newline (a lone `\r` stays glued to its word, matching real bash) | — |
+| `shell_split(cmd)` | Quote-aware split into argv-ready words (quotes stripped, escapes resolved) | `hooks/mod.rs::is_claude_hook_command`, `main.rs`'s `rtk proxy '...'` |
+| `split_for_permissions(cmd)` | Segments a compound command for the **permission gate** — deliberately the most conservative of three segmenters (see its doc comment for the full comparison table) | `hooks/permissions.rs::check_command_with_rules` |
+| `split_on_operators(cmd, stop_at_pipe)` | Segments for classification only — not safe for permission/security decisions | `registry.rs::split_command_chain` |
+| `contains_unattestable_construct(cmd)` | True for command/process substitution or a file-target redirect — constructs the permission gate can't decompose and must never auto-allow | `hooks/permissions.rs::check_command_with_rules` |
+
+The permission gate, discover/analytics classification, and rewrite each segment compound commands (`&&`, `;`, `|`, background `&`, subshells) slightly differently on purpose — the gate must never under-segment (a hidden command could evade a deny rule), while rewrite and analytics only need to reproduce or classify the command's actual shape. Don't reuse `split_on_operators` or `rewrite_compound`'s segmenting for a permission/security decision; use `split_for_permissions`.
+
 ## How History Analysis Works
 
 `rtk discover` reads Claude Code JSONL session files. Each file contains `tool_use`/`tool_result` pairs for every command the LLM ran. The module:
--- a/src/discover/lexer.rs
+++ b/src/discover/lexer.rs
@@ -68,6 +68,36 @@ pub(crate) fn is_crlf_at(bytes: &[u8], i: usize) -> bool {
     bytes.get(i) == Some(&b'\r') && bytes.get(i + 1) == Some(&b'\n')
 }
 
+/// Merges lexer tokens that are directly adjacent in `cmd` (no whitespace or
+/// other gap between them) into single words — the gap between "one lexer
+/// token" and "one bash word": `tokenize()` splits `*.yml` into a
+/// `Shellism("*")` and an `Arg(".yml")` for shell-syntax purposes, but bash
+/// sees one word since nothing separates them. Callers that only need
+/// "was there a space here" (not full shell-operator awareness) build on
+/// this instead of re-scanning `cmd` themselves.
+pub(crate) fn coalesce_words<'a>(cmd: &'a str, tokens: &[ParsedToken]) -> Vec<(&'a str, usize)> {
+    let mut words = Vec::new();
+    let mut run_start: Option<usize> = None;
+    let mut run_end: usize = 0;
+
+    for tok in tokens {
+        if let Some(start) = run_start {
+            if tok.offset != run_end {
+                words.push((&cmd[start..run_end], start));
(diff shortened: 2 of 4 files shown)


aeppling:
Tested locally against develop. Rewrite output identical on CR/CRLF inputs, suite green. Two regressions and one cleanup before merge.

**1. `shell_split` now splits on NBSP**

`is_word_boundary_whitespace` is Unicode whitespace minus CR, not bash IFS. Old `shell_split` split on space/tab only. `a<NBSP>b` was one word on develop, now two, bash keeps it one. Breaks argv for `rtk proxy '...'`, `is_claude_hook_command`, `search_uses_pattern_file` on pasted commands. Restrict to space/tab/newline and add a test.

**2. Permission gate loosened on lone CR, description says no policy change**

On develop `git status<CR>rm -rf ~` with deny `rm:*` returns Deny. On this branch it is one segment and returns Default. Not exploitable, git just errors, but it is a policy change in the gate you document as the most conservative. Fix the description and add a test pinning the deny result.

**3. Comments**

Doc comments carry review history ("confirmed via /code-review high", "flagged during PR #3600 review", "the way they had before this was extracted"), and `coalesce_words` has a dozen lines defending an allocation. That belongs in the PR, not the source. Keep the why, drop the rest.

KuSh:
> Tested locally against develop. Rewrite output identical on CR/CRLF inputs, suite green. Two regressions and one cleanup before merge.
> 
> **1. `shell_split` now splits on NBSP**
> 
> `is_word_boundary_whitespace` is Unicode whitespace minus CR, not bash IFS. Old `shell_split` split on space/tab only. `a<NBSP>b` was one word on develop, now two, bash keeps it one. Breaks argv for `rtk proxy '...'`, `is_claude_hook_command`, `search_uses_pattern_file` on pasted commands. Restrict to space/tab/newline and add a test.
> 
> **2. Permission gate loosened on lone CR, description says no policy change**
> 
> On develop `git status<CR>rm -rf ~` with deny `rm:*` returns Deny. On this branch it is one segment and returns Default. Not exploitable, git just errors, but it is a policy change in the gate you document as the most conservative. Fix the description and add a test pinning the deny result.
> 
> **3. Comments**
> 
> Doc comments carry review history ("confirmed via /code-review high", "flagged during PR #3600 review", "the way they had before this was extracted"), and `coalesce_words` has a dozen lines defending an allocation. That belongs in the PR, not the source. Keep the why, drop the rest.

Everything has been taken into account. Regarding point 2, I wasn’t sure whether the better approach was to get back to the old behavior or update the documentation to describe the new one. I opted for the latter, but please let me know if you’d prefer to restore the previous behavior.

aeppling:
@KuSh 

Your code already restores the previous gate behaviour via NewlineMode::Conservative and the new deny test pins it. That is the right call, just align the PR description and code comments with it.

KuSh:
> Your code already restores the previous gate behaviour via NewlineMode::Conservative and the new deny test pins it. That is the right call, just align the PR description and code comments with it.

PR description updated. Didn't found comments that contradict the code

aeppling:
Approving. Fuzzed this against develop: four refs (merge-base, PR head, develop, and a local merge into current develop), identical deterministic corpora fed to each. Roughly 540k generated inputs per ref at unit level, 4,679 commands through the real `rtk hook claude` surface with a sandboxed HOME, and 32,725 inputs scored against real bash 5.2.21 as ground truth.

No regressions. Zero unexplained diffs, zero panics, zero malformed hook JSON, zero token offset or slice violations, and no case where develop passed a command and this branch blocks or errors. Every diff traced to a documented intended fix.

No savings impact. The diff touches no filter code. On 84 realistic commands, 74 rewrite on both branches with byte-identical rewritten text, and real filter output is byte-identical across `git log`, `git status`, `git diff`, `grep`, `ls`, `cargo tree`, `wc`.

Measurably better than develop. Bash conformance on `shell_split` goes from 57.4% to 99.99%, with zero inputs where develop matches bash and this branch does not. It closes a real permission hole: develop returned `allow` for shapes like `git status<NBSP>rm -rf ~`, because its whitespace normalization collapsed the separator into the allow rule. It also stops develop from mutating command bytes, for example rewriting `golangci-lint<NBSP>run ./...` with the separator replaced by a space. Windows paths in double quotes and empty quoted arguments now survive argv intact.

On the permission gate: zero Deny or Ask to Allow transitions across 5,127 hook invocations plus 40k rule-matrix cases, and the lone-CR deny path is fully preserved by Conservative mode.

aeppling (approved):

PR #3716, opened 2026-08-26
fix(search): neutralize the `rg -rn` grep-habit footgun

ninthhousestudios:
## Problem

ripgrep is **recursive by default** and reads `-r` as `--replace`, consuming the rest of the short cluster as replacement text. So `rg -rn 'Foo' .` — typed constantly from `grep -rn` muscle memory — means *"replace every match of `Foo` with the literal `n`"*. The searched identifiers, URLs, etc. all silently become `n` in the output.

rtk forwarded the flag through to ripgrep unchanged, so the corruption surfaced in filtered output and looked like an rtk problem. It isn't — raw `rg -rn` corrupts identically — but since rtk is in the pipeline it takes the blame, and the output is silently wrong either way.

**Repro (any ripgrep, with or without rtk):**
```
$ printf 'class ChatPanel extends X {}\n' > f.rs
$ rg -rn 'ChatPanel' .        # grep habit
./f.rs:1:class n extends X {}  # <-- match replaced with "n"
```

## Fix

For the **rg engine only**, strip `-r`/`-R`/`--recursive` (all no-ops in ripgrep, which recurses by default) before the command reaches the engine, and warn once on stderr so the correction is visible:

```
$ rtk rg -rn 'ChatPanel' .
rtk: ripgrep reads -r as --replace (it is recursive by default); dropped -r/-R
     to avoid replacing matches. Use --replace for deliberate substitution.
./f.rs:1:class ChatPanel extends X {}
```

Scoped tightly to avoid surprises:
- **grep is untouched** — its `-r`/`-R` genuinely mean recursion.
- The deliberate long form `--replace` is left alone.
- Only boolean short clusters are rewritten; `-e`/`--regexp` values, value-flag values (`-A 3`), and positionals after `--` are never touched, so a literal pattern like `-rn` is never rewritten.
- `real_cmd` (tracking/tee label) is captured before sanitization, so metrics still reflect what the agent actually typed.

## Tests

Adds `sanitize_rg_recursive` + unit tests: the `-rn` footgun, cluster variants (`-nr`, `-rln`, `-Rn`, lone `-r`), context-flag preservation, deliberate `--replace`, `--recursive` drop, and pattern-safety (`--`/`-e`). Full suite green (`cargo test --all`), clippy clean, `cargo fmt` applied.

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -136,7 +136,8 @@ fn match_block(path: &str, entries: &[(usize, bool, String)]) -> String {
 ///
 /// - `patterns`: positional pattern + all `-e`/`--regexp` values. Empty → error.
 /// - `paths`: subsequent non-flag positionals. Empty → caller defaults to `["."]`.
-/// - `flags`: other flags forwarded to rg (`-r`/`-R`/`--recursive` stripped).
+/// - `flags`: other flags, forwarded verbatim. (For the rg engine the `-r`/`-R`
+///   grep-habit footgun is neutralized earlier by `sanitize_rg_recursive`.)
 ///
 /// Short clusters are scanned left-to-right; the first value-taking letter
 /// terminates the cluster — everything after it is its inline value, not a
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the clear repro and for pinning down that ripgrep reads -r as --replace. #3681, merged on 2026-09-12, rebuilt src/cmds/system/search.rs on the shared arg_tokenizer: parse_cluster, ClusterResult and the VALUE_FLAGS_LONG table that sanitize_rg_recursive is built from were all removed, and the engine difference now lives in is_recursive_token and rg_takes_value, so this patch and its tests no longer apply. The behaviour is still missing, though. Running rtk rg -rn 'ChatPanel' . against a current develop build still replaces the match with n, so a narrow follow-up that drops the flag and warns from inside the tokenizer grammar would be very welcome, either as an issue or a small PR on the new model. Please comment here or reopen if a part of this was missed.

PR #3719, opened 2026-08-26
fix(hook): honor exclusions for tail line ranges

juliosuas:
## Summary

- honor `hooks.exclude_commands` before the special `head`/`tail` line-range rewrite
- add literal and regex regression coverage for the `tail -n` escape hatch
- address the configuration bypass reported in #3107 without expanding this PR into the separate streaming/memory work

## Test plan

- [x] Manual testing: an isolated config with `tail` or `^tail -n` leaves `tail -n 400 /var/log/foo.log` unchanged; without an exclusion it still rewrites to `rtk read /var/log/foo.log --tail-lines 400`

Security note: this changes Tier 1 rewrite logic in `src/discover/registry.rs`; the diff received isolated full-suite testing and an independent senior security/quality review before publication.

---
Restored after an accidental fork deletion. Same commits as #3109.

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -996,6 +996,10 @@ fn rewrite_segment_inner(
     if context == RewriteContext::Normal
         && (cmd_part.starts_with("head -") || cmd_part.starts_with("tail "))
     {
+        let normalized = strip_absolute_path(cmd_part.trim());
+        if is_excluded(&normalized, excluded) {
+            return None;
+        }
         return rewrite_line_range(cmd_part).map(|r| format!("{}{}", r, redirect_suffix));
     }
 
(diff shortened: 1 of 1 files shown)


KuSh:
Closing as a duplicate of #3324, which fixes the same fast-path gap.

Two things from yours are being carried over: the path-stripped match — #3324 currently only strips the env prefix, so `/usr/bin/head -20 f` would still slip through, and I've asked for yours to be added — and the regex-pattern test case. Of the seven PRs open against this one branch, yours is the only one that handles path-prefixed invocations.

One note on the issue link: #3107 bundles three problems — `rtk read` buffering the whole file, output-vs-memory bounds, and the exclusion gap. This fixes only the third, so #3324 will reference it as a partial fix rather than closing it. The memory half is tracked by #3132.

PR #3724, opened 2026-08-27
fix(hooks): rewrite producer stage of `producer | cat` pipelines

### LINKED ISSUE #3722: Hook rewrite (rtk hook claude) skips commands containing a pipe (|)
**Version:** rtk 0.46.0 (Windows x86_64-pc-windows-msvc)

**Summary:**
`rtk hook claude` (the PreToolUse Bash hook used for Claude Code auto-rewrite) silently skips rewriting whenever the command contains a pipe (`|`), even when a rewritable subcommand (e.g. `git status`, `find`) appears before the pipe. No rewrite happens and no error is emitted — the hook just passes the original command through unchanged.

**Repro:**

```
echo '{"tool_name":"Bash","tool_input":{"command":"git status"}}' | rtk hook claude
# -> rewrites to: rtk git status  (correct)

echo '{"tool_name":"Bash","tool_input":{"command":"git status | cat"}}' | rtk hook claude
# -> no output / no rewrite (bug)

echo '{"tool_name":"Bash","tool_input":{"command":"git status --porcelain | head -20"}}' | rtk hook claude
# -> no output / no rewrite (bug)

echo '{"tool_name":"Bash","tool_input":{"command":"find . -name \"*.txt\" | head -20"}}' | rtk hook claude
# -> no output / no rewrite (bug)
```

Chained commands with `&&` work fine and rewrite the trailing subcommand correctly:

```
echo '{"tool_name":"Bash","tool_input":{"command":"git rev-parse HEAD && git status"}}' | rtk hook claude
# -> rewrites to: git rev-parse HEAD && rtk git status  (correct)
```

**Expected:** Commands piped to another program (`| head`, `| cat`, `| wc -l`, etc.) should still have their rewritable segment(s) converted to the `rtk` equivalent, the same way `&&`-chained commands already are.

**Impact:** Found via `rtk discover --all` on real Claude Code session history — dozens of `git status`, `find`, etc. invocations went unhandled specifically because they were piped to `head`/`wc`/`cat`, undercutting the tool's token-savings hook for a very common shell pattern.

javython999:
A bare `cat` (no flags, no files) is byte-transparent: it copies stdin to stdout unchanged. `producer | cat` and `producer` are therefore equivalent on stdout, so the producer segment can be rewritten the same way a standalone command would be, without risking a change to what a downstream consumer reads.

Previously the hook left pipelines entirely raw except for an allowlisted final-stage rewrite (grep/rg), so `git status | cat` never got the rtk rewrite even though nothing consumes the pipe's output structurally.

The exception is scoped to a single stdout pipe with no further pipe stages, so pipelines with real intermediate processing (`a | b | cat`) keep the existing raw-producer contract.

Fixes #3722

## Summary

- Rewrite the producer stage of a `producer | cat` pipeline (e.g. `git status | cat` -> `rtk git status | cat`) since a bare `cat` sink is byte-transparent
- Added `is_transparent_cat_sink` / `rewrite_producer_before_transparent_sink` in `src/discover/registry.rs`, wired as a fallback when the existing final-stage rewrite finds nothing
- Scoped narrowly: only a single stdout pipe with no further pipe stages qualifies; `a | b | cat`, `cat -n`, and `|&` stderr pipes are unaffected

## Test plan

- [x] Added 5 new unit tests in `src/discover/registry.rs` covering the happy path, `&&` continuation after the pipe, `cat -n` (non-transparent), a real intermediate stage before `cat`, and the stderr-pipe case
- [x] Manual testing: `echo '{"tool_name":"Bash","tool_input":{"command":"git status | cat"}}' | rtk hook claude` now rewrites to `rtk git status | cat`

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -1031,6 +1031,72 @@ fn rewrite_pipeline_final_stage(
     })
 }
 
+/// A plain `cat` invocation (no flags, no files) is byte-transparent: it
+/// copies stdin to stdout unchanged. `producer | cat` and `producer` are
+/// therefore equivalent on stdout, so the producer can be rewritten exactly
+/// like a standalone command without risking a change to what the pipe's
+/// consumer sees. Any argument (`cat -n`, `cat -`, `cat file`) changes that
+/// contract, so only the bare word qualifies.
+fn is_transparent_cat_sink(stage: &str) -> bool {
+    stage.trim() == "cat"
+}
+
+/// Issue #3722: pipelines never rewrite their producer (see
+/// [`rewrite_compound`]'s doc comment) because a filter's reformatted output
+/// could change what a downstream consumer reads. That contract holds even
+/// for [`rewrite_pipeline_final_stage`]'s allowlisted final-stage rewrite —
+/// it only replaces the *sink*, never the producer feeding it.
+///
+/// `producer | cat` is the one shape where rewriting the producer is
+/// provably safe: a bare `cat` sink is byte-transparent (see
+/// [`is_transparent_cat_sink`]), so nothing downstream of the pipe can
+/// observe the difference between `producer` and `rtk producer`. The
+/// exception is scoped tightly — a single stdout pipe, no further pipe
+/// stages — so a pipeline with real intermediate processing (`a | b | cat`)
+/// keeps the general raw-producer contract.
+fn rewrite_producer_before_transparent_sink(
+    cmd: &str,
+    tokens: &[ParsedToken],
+    seg_start: usize,
+    first_pipe_offset: usize,
+    end_offset: usize,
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the tight scoping and for the tests that came with it. #3343, merged on 2026-09-07, made the hook rewrite the producer stage whenever the consumer is safe, so git status | cat already becomes rtk git status | cat on develop, and it goes further than this PR by also accepting cat -n while still leaving a | b | cat and stderr pipes alone. I checked each of those cases against a current develop build. The helpers here, is_transparent_cat_sink and rewrite_producer_before_transparent_sink, sit on the pipeline shape that #3343 and #3704 replaced, so there is nothing left to carry over. Please comment here or reopen if a case you had in mind is still not rewritten.

Issue #348, opened 2026-03-05
feat: Automated Issue & PR triage system (templates, labeling, classification)


FlorianBruniaux:
## Context

With 88 open issues and 51 open PRs (as of March 2026), 86% of issues and 72% of PRs have no labels, forcing manual triage every time. This issue tracks a 3-sprint plan to automate the full triage pipeline.

The real bottleneck is PRs: 54% are 500+ lines, 32% have active discussions. Issues close fast (median 1.9 days) but arrive unlabeled.

**Estimated time saved**: ~3-5h/week of triage once fully implemented.

---

## Target Contributor Experience

Max 2 automated messages per item. No "wall of bots" effect.

### Issue Flow
```
Contributor opens issue
  -> Structured form (Bug/Feature/Question)              [Sprint 1]
  -> Labels auto-applied by template                     [Sprint 1]
  -> Action checks fields:
     - Missing version/OS -> label "needs-info" + 1 comment
     - Duplicate detected (Jaccard > 0.6) -> label "possible-duplicate" + mention
     - Complete -> labels applied, zero comment          [Sprint 2]
  -> If "needs-info" unanswered 14d -> 1 reminder        [Sprint 2]
  -> If still nothing 28d -> close                       [Sprint 2]
```

### PR Flow
```
Contributor opens PR
  -> Template with RTK checklist                         [Sprint 1]
  -> Labels by path (area:git, area:go, etc.) + size     [Sprint 1]
  -> security-check.yml runs + RTK quality patterns
     = 1 combined "Security & Quality Scan" comment      [Sprint 2]
  -> CODEOWNERS auto-requests maintainer review          [Sprint 1]
```

---

### P1 — Issue Templates (YAML Forms)
3 templates with required fields:
- **Bug Report**: version, OS, shell, command, repro steps (all required)
- **Feature Request**: target command, use case, expected token savings
- **Question**
Auto-labeling at creation. Fixes the #1 problem: 86% of issues without labels.

### P2 — PR Template
RTK checklist embedded:
- [ ] `lazy_static!` for regex (not inline `Regex::new`)
- [ ] `anyhow` + `.context()` on all `?` operators
- [ ] Fallback to raw command on filter failure
- [ ] Exit codes preserved
- [ ] Token savings ≥60% (with test assertion)
- [ ] No `async fn` / `tokio`

### P3 — PR Auto-Labeling by path
`actions/labeler` mapping all modules. Fixes 72% of PRs without labels.

[...]

PR #3743, opened 2026-08-28
ci: add stale bot for issues and PRs (90d stale, +7d close)

pszymkowiak:
## Summary
- Adds `.github/workflows/stale.yml` using `actions/stale@v9`, mirroring facebook/react's config: 90 days of inactivity → `stale` label + comment, +7 more days with no reaction → auto-close.
- Applies to both issues and PRs. Exempt labels: `bug`, `enhancement`, `area:security`, `P0-critical`, `P1-high`, `DONE_REVIEW`, `good first issue`, `help wanted` — mirrors React's approach of protecting anything that's already been triaged.

## Context
#348 originally scoped a stale bot **out** of its triage-automation plan: *"Oldest item is 17 days. No staleness problem yet"* — accurate in March 2026 at 51 open PRs / 88 open issues.

That's no longer the case. At current volume (1049 open PRs):
- **253 PRs** (24%) have had zero activity in 90+ days
- **182 of those** have no label at all — never triaged, not just idle

This doesn't replace the broader triage system #348 proposes (templates, auto-labeling by path, `needs-info` follow-up, etc.) — it's a narrow, independent piece that can land on its own while the rest of that plan is still pending.

## Notes
- Thresholds (90d/+7d) are React's numbers, used as a reasonable starting point — open to tuning for rtk's actual velocity (happy to go tighter, e.g. Bun's 30d/+14d for PRs specifically, if that fits better).
- `operations-per-run: 500` to actually make a dent in the current backlog size; default is 30.
- Daily cron (`0 6 * * *`) rather than hourly, since this isn't time-sensitive.

## Test plan
- [ ] `workflow_dispatch` dry run on a fork or via `actions/stale`'s `debug-only: true` before merge, to sanity-check the label/message output against a handful of real PRs
- [ ] Confirm `exempt-*-labels` list against the actual label taxonomy (`gh api repos/rtk-ai/rtk/labels`)

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -8,20 +8,27 @@
 name: Manage stale issues and PRs
 on:
   schedule:
-    # Run daily
-    - cron: '0 6 * * *'
+    # Run weekly
+    - cron: '0 6 * * 1'
   workflow_dispatch:
 
-permissions:
-  issues: write
-  pull-requests: write
+permissions: {}
 
 jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        id: app-token
+        with:
+          client-id: ${{ secrets.APP_CLIENT_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          permission-issues: write
+          permission-pull-requests: write
+
       - uses: actions/stale@v9
         with:
+          repo-token: ${{ steps.app-token.outputs.token }}
           # --- Issues & PRs ---
           days-before-stale: 90
           days-before-close: 7
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,55 @@
+# Configuration for stale action workflow - https://github.com/actions/stale
+#
+# Mirrors facebook/react's config (days-before-stale: 90, days-before-close: 7),
+# adapted with rtk's own label taxonomy for exemptions. See #348 for the wider
+# triage-automation context this is a piece of; #348 originally scoped a stale
+# bot out ("oldest item is 17 days, no staleness problem yet") — that was true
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,55 @@
...
+# bot out ("oldest item is 17 days, no staleness problem yet") — that was true
+# in March 2026 at 51 open PRs, no longer true at current volume.
+name: Manage stale issues and PRs
+on:
+  schedule:
+    # Run daily
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
+permissions:
```
aeppling: permissions: {}


Then use the git app token like others ci does (one step to get token, then use it in the stale action)
Otherwise lgtm

aeppling (changes requested):

.github/workflows/stale.yml
```diff
@@ -0,0 +1,55 @@
...
+# Mirrors facebook/react's config (days-before-stale: 90, days-before-close: 7),
+# adapted with rtk's own label taxonomy for exemptions. See #348 for the wider
+# triage-automation context this is a piece of; #348 originally scoped a stale
+# bot out ("oldest item is 17 days, no staleness problem yet") — that was true
+# in March 2026 at 51 open PRs, no longer true at current volume.
+name: Manage stale issues and PRs
+on:
+  schedule:
+    # Run daily
+    - cron: '0 6 * * *'
```
aeppling: Daily may not be needed here ? Maybe weekly is enough ?

.github/workflows/stale.yml
pszymkowiak: Done in 3ffc297 — permissions: {} at job level, actions/create-github-app-token step (same pattern as cd.yml/release.yml/pr-target-check.yml), token passed to actions/stale via repo-token.

.github/workflows/stale.yml
pszymkowiak: Done in 3ffc297 — switched to weekly (Mondays 06:00 UTC).

aeppling (approved):

PR #3749, opened 2026-08-29
fix(hooks): match exclude_commands against the peeled command form

### LINKED ISSUE #243: Support exclude_commands config for auto-rewrite hook

### LINKED ISSUE #3035: [hooks] exclude_commands is matched against the raw command, not the resolved tool — "pytest" fails to exclude `python -m pytest`

KuSh:
Fixes #243. Fixes #3035.

## The problem

`exclude_commands` entries name a tool, but a command can spell that tool with a wrapper (`npx`, `pnpm exec`, `pnpm dlx`), an interpreter (`python3 -m`), or a path (`vendor/bin/`, `./`). Those spellings are absorbed by each rule's own pattern rather than stripped beforehand, so by the time `is_excluded` runs the string is still `npx playwright test` — and the anchored `^playwright($|\s)` never matches.

That produced an invisible asymmetry:

```toml
[hooks]
exclude_commands = ["playwright"]
```

| Command | before | after |
|---|---|---|
| `playwright test` | excluded ✅ | excluded ✅ |
| `npx playwright test` | **rewritten** ❌ | excluded ✅ |
| `pnpm exec playwright test` | **rewritten** ❌ | excluded ✅ |

Same tool, opposite outcomes, no warning. `README.md` shipped `exclude_commands = ["curl", "playwright"]` as *the* example — and `playwright` is a tool almost nobody invokes bare, so the documented example silently failed for the standard invocation. That's what @nhumrich hit in #243.

The same gap covers `python3 -m pytest` with `["pytest"]` (#3035), where the consequence is sharper: `-m` runs the current interpreter with CWD on `sys.path`, while the bare shim may be a different Python entirely, so the rewrite silently swaps interpreters — exactly what the exclusion existed to prevent.

## The fix

Peel the wrapper off the command and match what remains, OR'd with the existing check on the typed command.

Peeling uses the rule's own `rewrite_prefixes`: take the longest prefix the command matches, then the shortest token-suffix of that prefix which is itself a prefix of the same rule. That drops `npx` and `python3 -m` while keeping a subcommand the rule treats as part of the tool, so `golangci-lint run` does not collapse to `run`.

**The arguments are kept.** This is what makes anchored entries keep meaning what they say — `"^ls$"` excludes a bare `ls` without swallowing `ls -la`.

### Why not match the resolved `rtk` target

Substituting the rule's `rtk_cmd` is a one-liner, and it's wrong in three ways. Recording them because it's the obvious shortcut:

1. **It breaks anchored regexes.** `compile_exclude_patterns` passes `^`-prefixed entries through verbatim, so handing `is_excluded` a bare token turns an anchor written to *narrow* a pattern into one that widens it: `["^ls$"]` starts excluding every `ls`, `["^cargo$"]` every `cargo`.
2. **It misses tools whose target differs from the binary.** `eslint` and `biome` both resolve to `rtk lint`, so `["eslint"]` would still rewrite `npx eslint .` — the exact class of bug this PR is meant to close.
3. **It leaks across tools sharing a target.** `["read"]` would exclude `cat`, `["git"]` would exclude `yadm`, `["rake"]` would exclude `rails`.

Peeling has none of these. All three are covered by tests.

## Verified

Built binary, isolated `XDG_CONFIG_HOME`. Covered forms:

```
["playwright"]     playwright test / npx / pnpm exec / pnpm dlx   -> all excluded
["pytest"]         python3 -m pytest tests/ -q                    -> excluded
["eslint"]         npx eslint .                                   -> excluded
["gradlew"]        ./gradlew assembleDebug                        -> excluded
["phpunit"]        vendor/bin/phpunit tests/                      -> excluded
["rspec"]          bundle exec rspec                              -> excluded
["phpunit"]        php vendor/bin/phpunit tests/                  -> excluded
```

Exactness preserved:

```
["^ls$"]           ls -la                -> rewritten   ("ls" alone -> excluded)
["go"]             golangci-lint run     -> rewritten
["py"]             python3 -m pytest     -> rewritten
["git push"]       git status            -> rewritten
["read"]           cat foo.txt           -> rewritten
["lint"]           eslint . / biome ...  -> rewritten
["git"]            yadm status           -> rewritten
```

## Tests

Seven in `src/discover/registry.rs`, each mapped to a claim above:

- `test_exclude_matches_wrapper_invoked_form` — the README example across all four forms
- `test_exclude_covers_interpreter_and_path_forms` — `python -m`, `./`, `vendor/bin/`, `bundle exec`
- `test_exclude_covers_wrapper_when_tool_name_differs_from_target` — the eslint/biome case
- `test_exclude_keeps_arguments_so_anchored_regex_still_narrows` — `^ls$` and `^pytest `
- `test_exclude_does_not_widen_across_tools_sharing_a_target` — read/cat, lint/eslint, git/yadm
- `test_exclude_peeled_form_is_exact_token` — `["go"]` vs `golangci-lint`, and `golangci-lint run` kept whole
- `test_exclude_subcommand_pattern_stays_narrow` — `["git push"]` vs `git status`
- `test_exclude_covers_php_wrapper_forms` — `php vendor/bin/phpunit`, `php bin/phpunit`, `php vendor/bin/phpstan`

`cargo fmt --all --check`, `cargo clippy --all-targets` and `cargo test --all` are clean (2657 passed).

## Limits

Peeling is driven by each rule's `rewrite_prefixes`, so it only reaches tools RTK has a filter
for. A tool RTK sees only through a generic wrapper rule is matched as typed: with
`["my-tool"]`, `npx my-tool` still rewrites to `rtk npx my-tool` — exclude `"npx"` for that.
The docs state this rather than promising blanket coverage.

## Scope

Two adjacent gaps are deliberately **not** addressed here:

- **#1053** — path-prefixed commands failing to *rewrite*. This PR makes the exclusion honour path forms, which is fail-safe (a missed rewrite is passthrough), but the issue's actual ask is to make them rewrite while preserving the caller's binary, which needs a `--rtk-bin` override and handler changes. Leaving it open.
- **#2823 / #3371 / #2363** — the `head`/`tail` line-range fast path returns before any exclusion check. Different code path; #3324 covers it.

## Docs

- `docs/guide/getting-started/configuration.md` — documents peeling, the argument-preserving guarantee, and the exactness limits.
- `docs/usage/FEATURES.md` — same, matching the section's existing French.
- `README.md` — annotates the example so the `npx` case is stated where people copy it from.

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,7 +411,7 @@ For per-agent setup details, override controls, and graceful degradation, see th
 
 ```toml
 [hooks]
-exclude_commands = ["curl", "playwright"]  # skip rewrite for these
+exclude_commands = ["curl", "playwright"]  # skip rewrite for these (matches `npx playwright` too)
 
 [tee]
 enabled = true          # save raw output on failure (default: true)
--- a/docs/guide/getting-started/configuration.md
+++ b/docs/guide/getting-started/configuration.md
@@ -95,6 +95,22 @@ Patterns match against the full command after stripping env prefixes (`sudo`, `V
 
 Subcommand patterns work too: `"git push"` excludes `git push origin main` but not `git status`.
 
+An entry names a tool RTK has a filter for, and that tool is covered however it is invoked.
+Before matching, RTK peels the wrapper, interpreter or path off the command and matches what is
+left, so `"playwright"` excludes `playwright test`, `npx playwright test` and
+`pnpm exec playwright test` alike; `"pytest"` also covers `python3 -m pytest tests/`, and
+`"phpunit"` covers `vendor/bin/phpunit` and `php vendor/bin/phpunit`.
+
+A tool RTK has no filter of its own for is matched as typed, because RTK only sees the wrapper:
+with `["my-tool"]`, `npx my-tool` still rewrites to `rtk npx my-tool`. Exclude `"npx"` to stop
+that.
+
+The arguments are kept when peeling, so an anchored pattern still narrows the way you wrote it:
+`"^ls$"` excludes a bare `ls` without swallowing `ls -la`. Matching stays exact — `"go"` never
+excludes `golangci-lint`, subcommand patterns stay literal (`"git push"` does not widen to all of
+`git`), and an entry never leaks to a different tool that happens to share an RTK filter: `"read"`
+does not exclude `cat`, and `"eslint"` does not exclude `biome`.
+
 Patterns starting with `^` are treated as regex:
 
 ```toml
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -1230,6 +1230,12 @@ Pour empecher certaines commandes d'etre reecrites, ajoutez-les dans `config.tom
 exclude_commands = ["curl", "playwright"]
 ```
 
+Le motif est ancre en debut de commande : `"curl"` exclut `curl https://...` mais pas `curl-config`.
+Avant la comparaison, RTK retire le wrapper, l'interpreteur ou le chemin, donc `"playwright"` couvre
+`playwright test` comme `npx playwright test` ou `pnpm exec playwright test`, et `"pytest"` couvre
+aussi `python3 -m pytest tests/`. Les arguments sont conserves : `"^ls$"` exclut `ls` seul sans
+englober `ls -la`.
+
 ---
 
 ## Configuration
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -9,7 +9,7 @@ use super::lexer::{
     shell_split, split_on_operators, tokenize, tokenize_with_newlines, ParsedToken, PipeKind,
     TokenKind,
 };
-use super::rules::{IGNORED_EXACT, IGNORED_PREFIXES, RULES};
+use super::rules::{RtkRule, IGNORED_EXACT, IGNORED_PREFIXES, RULES};
 
 const PHP_TOOL_NAMES: [&str; 6] = ["phpunit", "phpstan", "ecs", "pest", "paratest", "pint"];
 
@@ -1382,7 +1382,10 @@ fn rewrite_segment_inner(
         Classification::Supported { rtk_equivalent, .. } => {
             let stripped = ENV_PREFIX.replace(cmd_part, "");
             let cmd_clean = stripped.trim();
-            if is_excluded(cmd_clean, excluded) {
+            if !excluded.is_empty()
+                && (is_excluded(cmd_clean, excluded)
+                    || is_excluded(&tool_form(cmd_clean, rtk_equivalent), excluded))
+            {
                 return None;
             }
             rtk_equivalent
@@ -1448,20 +1451,12 @@ fn rewrite_segment_inner(
     // classify_command does, so a small canonical prefix list matches every
     // invocation form instead of enumerating each literal spelling.
     let php_normalized;
-    let strip_target: &str = if rule
-        .rtk_cmd
-        .strip_prefix("rtk ")
-        .is_some_and(|t| PHP_TOOL_NAMES.contains(&t))
-    {
-        // Peel `php ` then a leading `./` (normalize_php_tool_command only
-        // strips `./` for paths that resolve to a Composer tool, so a plain
-        // `./bin/<tool>` would otherwise survive and miss the prefix match).
-        let unwrapped = strip_php_wrapper(cmd_part);
-        let unwrapped = unwrapped.strip_prefix("./").unwrap_or(unwrapped);
-        php_normalized = normalize_php_tool_command(unwrapped);
-        &php_normalized
-    } else {
-        cmd_part
+    let strip_target: &str = match php_tool_form(cmd_part, rule.rtk_cmd) {
+        Some(normalized) => {
+            php_normalized = normalized;
+            &php_normalized
+        }
+        None => cmd_part,
     };
 
     // Try each rewrite prefix (longest first) with word-boundary check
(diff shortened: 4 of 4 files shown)


aeppling:
Hey @KuSh , reviewed. No regressions, logic is correct as written.

Just one doc fix , non blocking

### configuration.md over-promises

New line 97: "that tool is covered however it is invoked." Stated limits cover only TOML-only tools and exactness. Three invocation forms of filtered tools aren't covered:

["head"]           head -20 f          -> rtk read f --max-lines 20    (head f IS excluded)
["gradlew"]        gradlew.bat build   -> rtk gradlew build            (./gradlew and gradlew both excluded)
["golangci-lint"]  golangci run ./...  -> rtk golangci-lint run ./...  (golangci-lint run IS excluded)

Repro: put the entry in ~/.config/rtk/config.toml under [hooks], then rtk rewrite '<cmd>'.

Row 1: line-range fast path returns before the exclusion check #3324.
Row 2: strip_absolute_path handles /, not \ or .bat/.exe; Windows asymmetry vs CODING_PRACTICES "Portability" #3617 closes it.
Row 3: the rule's own alias, tool_portion keeps it whole.

Suggest scoping the sentence to the wrapper/interpreter/path forms actually peeled and naming these as known gaps.

Line 97, replace:

"An entry names a tool RTK has a filter for, and that tool is covered however it is invoked."

with:

"An entry names a tool RTK has a filter for, and covers the wrapper, interpreter and path spellings of it."

aeppling (approved):

aeppling:
Oopsi i merged, my bad, you can revert if you want to still work on this

KuSh:
No problem at all — thanks for the review, and the nitpick was right.

Follow-up in #3803, docs only. I took your suggested sentence verbatim and tabled the three forms as known gaps rather than leaving them implied.

Two things I confirmed while writing it up. `mvnw.cmd test` under `["mvn"]` is a fourth instance of your row 2 — same `.bat`/`.cmd` cause as `gradlew.bat` — so the table names both tools. And the `golangci` workaround isn't symmetric: `["golangci"]` excludes `golangci run ./...` but *not* `golangci-lint run ./...`, so both entries are needed to cover the tool. That's documented.

On row 3 specifically — I left it as a documented gap rather than fixing it. `golangci run` is one of the rule's own `rewrite_prefixes` aliases, and `tool_portion` keeps a rule-owned subcommand whole on purpose, so `golangci-lint run` doesn't collapse to `run`. Mapping aliases onto a canonical name is exactly the resolved-target matching this PR rejected — it's what broke `["^ls$"]` and missed `npx eslint`. Not worth reopening for an alias form.

Rows 1 and 2 are already covered by #2823/#3324 and #3617, so they'll drop out of the table as those land.

PR #3751, opened 2026-08-29
fix(json,deps): tolerate UTF-8 BOM at the remaining JSON parse sites

kylehgc:
**Stacked on #3203** (`submit/bom-config-reads`) — only the last commit (`fix(json,deps): tolerate UTF-8 BOM at the three remaining JSON parse sites`) is new; the first two commits are #3203 itself and will drop out of the diff once it merges. Happy to rebase onto develop after #3203 lands.

Follow-up covering the three parse sites @KuSh flagged in the #3203 review as needing the same BOM tolerance, which were scoped out of that PR:

- **`src/cmds/system/deps.rs` — `summarize_package_json_str`**: parsed `package.json` with a raw `serde_json::from_str(&content)?` (also missing `.context()`). A BOM'd package.json — written by Notepad or PowerShell 5.1 `Out-File -Encoding utf8` — made `rtk deps` hard-fail with a bare `expected value at line 1 column 1` that didn't name the file. Now strips the BOM via `strip_leading_bom` and carries the file path in both read and parse errors.
- **`src/cmds/system/json_cmd.rs` — `filter_json_compact` and `filter_json_string`**: `rtk json <file>` (compact and `--keys-only`) errored on BOM-prefixed files — the single most likely command to be pointed at human-saved JSON. Both entry points (file and stdin) route through these functions, so both are covered.

## Repro

```powershell
# PowerShell 5.1: Out-File -Encoding utf8 writes a BOM
'{"name": "demo", "version": "1.0.0"}' | Out-File -Encoding utf8 package.json
rtk json package.json   # before: "Failed to parse JSON: expected value at line 1 column 1" (exit 1) — after: prints the JSON
rtk deps .              # before: "expected value at line 1 column 1" (exit 1) — after: summarizes normally
```

Verified before/after on all three paths, including `--keys-only`.

## Tests

Per site: BOM-prefixed input parses and produces output identical to the plain input; a plain-input test guards existing behavior; the invalid-package.json error now contains the path.

`cargo fmt --all && cargo clippy --all-targets && cargo test --all` — clean.


kylehgc:
Updated: merged the current #3203 branch (which now carries the `from_json_str` wrapper from its review round) and converted this PR's three sites from the raw `serde_json::from_str(strip_leading_bom(..))` combination to the wrapper, so the whole series states the policy one way. Stacked on #3203 — the first two commits here are its; the diff net of that base is the three-site fix plus the wrapper conversion. Gate: fmt clean, clippy zero warnings, 2502 tests passing.

src/hooks/init.rs
```diff
@@ -4269,7 +4278,14 @@ fn patch_gemini_settings(
...
             .with_context(|| format!("Failed to read {}", settings_path.display()))?;
-        serde_json::from_str(&content).unwrap_or(serde_json::json!({}))
+        let content = strip_leading_bom(&content);
+
+        if content.trim().is_empty() {
+            serde_json::json!({})
+        } else {
+            from_json_str(content)
+                .with_context(|| format!("Failed to parse {} as JSON", settings_path.display()))?
+        }
```
KuSh, line 4323: **Aborting here leaves a half-installed state.**

Not wiping the user's settings on a parse error is the right call, but the `?` now fires *after* `run_gemini` has already written the hook script, `GEMINI.md`, and the integrity baseline (init.rs:4227-4247).

Concrete case: a `~/.gemini/settings.json` containing `//` comments, a form Gemini CLI's own settings loader accepts since it strips JSON comments before parsing. `rtk init -g --gemini` then exits 1 with `Failed to parse ... as JSON`, prints no "Gemini CLI hook installed" summary, and leaves those files on disk with no guidance on how to proceed.

Suggestion: treat an unparseable settings file the same as `PatchMode::Skip` (warn, print the manual-setup instructions, let the rest of the install report success) rather than aborting the whole command.

src/hooks/hook_cmd.rs
```diff
@@ -9,6 +9,7 @@ use anyhow::{Context, Result};
 use serde_json::{json, Value};
 use std::io::{self, Read, Write};
 
+use crate::core::utils::strip_leading_bom;
 use crate::discover::registry::{has_heredoc, rewrite_command};
```
KuSh, line 13: **The two main agent hook entry points still don't call this.**

`run_gemini` (hook_cmd.rs:232) and `run_claude` (hook_cmd.rs:407) still do a bare `serde_json::from_str(&input)` on the stdin payload, while `run_cursor` (448), the Copilot path (50) and `run_droid` (613) all strip first. The PR promotes the helper to `core::utils` with a doc comment saying it belongs "wherever RTK parses JSON a human or another tool may have written", but the "remaining parse sites" framing leaves these two out.

Failure, and it is the exact one `run_cursor` exists to prevent: a Windows agent ships a BOM-prefixed PreToolUse payload. In `run_gemini` the parse returns `Err`, so `rtk hook gemini` exits non-zero and the tool call is blocked. `run_claude` degrades more quietly: `str::trim` does not strip U+FEFF, so the parse fails, it logs to stderr and returns `Ok(())`, and every command silently stops being rewritten.

One added call per site.

src/cmds/system/json_cmd.rs
```diff
@@ -92,7 +93,8 @@ pub fn run_stdin(max_depth: usize, schema_only: bool, verbose: u8) -> Result<()>
 /// Parse a JSON string and return compact representation with values preserved.
 /// Long strings are truncated, arrays are summarized.
 pub fn filter_json_compact(json_str: &str, max_depth: usize) -> Result<String> {
-    let value: Value = serde_json::from_str(json_str).context("Failed to parse JSON")?;
+    let value: Value =
+        from_json_str(json_str).context("Failed to parse JSON")?;
```
KuSh, line 121: **`never_worse` still compares against the un-stripped input, so the raw fallback re-emits the BOM.**

`run` (json_cmd.rs:56) and `run_stdin` call `never_worse(&content, &output)` with the raw `content`; the BOM only disappears inside `filter_json_*`.

Concrete case: a minified BOM-prefixed `package.json`. `compact_json` pretty-prints with indentation, so the filtered form is *larger*, `never_worse` returns the raw string, and `rtk json package.json | jq .` still fails with a parse error, the exact command this PR advertises as fixed.

Suggestion: strip the BOM once at the file/stdin read, before `never_worse`, instead of only inside the two filter functions.

KuSh (changes requested):
Reviewed the full cumulative diff (5 commits, 10 files). Verified on a clean worktree at `beee9df`: `cargo fmt --all -- --check` clean, `cargo clippy --all-targets` zero warnings, `cargo test --all` 2568 passed / 0 failed.

The core change is sound. `from_json_str` keeps the zero-copy `<'a, T: Deserialize<'a>>` signature, and the `strip_leading_bom` -> `trim().is_empty()` ordering is correct at every site that needed it: U+FEFF is `Cf`, not whitespace, so the empty-file guard has to run *after* stripping, and the new BOM-only / BOM-plus-invalid-JSON tests pin exactly that. Panic safety on the write paths is intact, and the remaining `serde_json::from_str` call sites in the tree are command-*output* parsers (aws, pnpm, vitest, ruff, ...) or RTK-written state (`trust.rs`), correctly left alone.

Three things to address before merge, see the inline comments. The `hook_cmd.rs` one matters most, given the PR is scoped as "the remaining parse sites".

kylehgc:
All three addressed as of b6f354c.

**Half-installed state:** `patch_gemini_settings` now treats an unparseable settings file the way `PatchMode::Skip` does — warns with the file path and parse error, prints the same manual-setup instructions (factored into a shared helper), and lets the rest of the install report success. The no-overwrite guarantee is unchanged: the parse-failure path returns before any mutation, and the regression test now asserts `Ok` + byte-identical file + the warn path.

**Hook stdin:** `run_gemini` and `run_claude` both strip the BOM now, matching the copilot/cursor/droid paths — all five entry points are uniform. `run_gemini`'s parse logic moved into a `run_gemini_inner` that production and the new BOM test share, so the test exercises the real path (a dropped strip breaks the actual hook, not a mirrored copy).

**`never_worse` fallback:** the BOM is stripped once at the read, in a shared `render_json` used by both the file and stdin paths, before the compare — so the raw fallback for an already-compact file emits BOM-free output and `rtk json foo.json | jq .` works on exactly the case you named. Pinned by a test forcing the fallback path with a minified BOM'd input.

Gate: fmt clean, clippy zero warnings, 2516 tests passing.

src/hooks/init.rs
KuSh: Fixed in b6f354c. `patch_gemini_settings` now warns with the file and the parse error, prints the same manual-setup instructions as `PatchMode::Skip` (extracted into the shared `print_gemini_manual_setup`), and returns `Ok(())` instead of aborting — so the install still reports its summary and the unparsed file is left untouched. The existing test was updated to pin the new contract. Resolving.

src/hooks/hook_cmd.rs
KuSh: Fixed in b6f354c. `run_claude_inner` and the new `run_gemini_inner` both call `strip_leading_bom` before parsing, and `run_claude` strips before `trim()` (right order — U+FEFF is `Cf`, so `trim` would not have removed it). Nice touch extracting `run_gemini_inner` and routing production `run_gemini` through it rather than mirroring the parse logic in the test; `gemini_json("allow", None)` renders byte-identical to the old `print_allow()`, so the refactor is behaviour-preserving. Both paths now have a BOM test. Resolving.

src/cmds/system/json_cmd.rs
KuSh: Fixed in b6f354c. The new `render_json` helper strips once up front and both filters *and* `never_worse` now see the stripped content, so the raw fallback no longer carries the BOM into piped output. `timer.track` still measures against the original `content`, which is right for the savings accounting. The added test drives `render_json` with the raw, un-stripped input and asserts the fallback output parses as JSON. Resolving.

src/hooks/hook_cmd.rs
```diff
@@ -1487,6 +1473,30 @@ mod tests {
...
+        let payload = json!({
+            "tool_name": "run_shell_command",
+            "tool_input": { "command": "git status" }
+        })
+        .to_string();
+        let with_bom = format!("\u{feff}{payload}");
+        let result = run_gemini_inner(&with_bom).expect("BOM-prefixed payload must parse");
+        let v: Value = serde_json::from_str(&result).unwrap();
+        // No config in the test environment: default-to-ask semantics.
+        assert_eq!(v["decision"], "ask_user");
```
KuSh: **Blocking: this test is non-hermetic and fails on real developer machines.**

"No config in the test environment" doesn't hold. This is the only test in the module that calls `run_gemini_inner`, which goes `decide_hook_action(cmd, Host::Gemini)` -> `permissions::load_gemini_rules()` -> `gemini_settings()` -> `global_config(GEMINI_DIR, SETTINGS_JSON)` — so it reads the *real* `~/.gemini/settings.json` (and the project `.gemini/settings.json`). Every other Gemini test here deliberately routes through the hermetic `gemini_render` / `decide_with_rules` helpers with explicit rule slices.

Reproduced: with `~/.gemini/settings.json` = `{"tools":{"allowed":["run_shell_command(git status)"]}}`, `cargo test` fails here with `left: "allow", right: "ask_user"` — and it is the *only* failure in the whole 2507-test suite under that HOME. Any contributor who actually uses Gemini CLI with `git status` allow-listed gets a spurious red test.

The BOM behaviour is what this test is for, and that part is environment-independent. Suggestion: drop the `decision` assertion and keep the one below it (the payload parsed and `hookSpecificOutput.tool_input.command == "rtk git status"`), or thread injected rules through an inner helper the way `gemini_render` does.

src/hooks/hook_cmd.rs
```diff
@@ -227,49 +228,46 @@ fn copilot_cli_response_from_decision(
...
+}
 
-    let json: Value = serde_json::from_str(&input).context("Failed to parse hook input as JSON")?;
+/// Parse the Gemini BeforeTool stdin payload, decide, and render the
+/// response JSON — no stdin/stdout I/O. Used by `run_gemini` itself (not
+/// just tests), so a regression here (e.g. dropping the BOM strip) fails
+/// for real rather than only in a duplicate test copy.
+fn run_gemini_inner(input: &str) -> Option<String> {
+    let input = strip_leading_bom(input);
+    let json: Value = serde_json::from_str(input).ok()?;
```
KuSh: Non-blocking: `.ok()?` discards the serde error, so the `.context("Failed to parse hook input as JSON")` at line 231 no longer has a source attached.

Before this refactor a malformed payload reported the serde detail (`expected value at line 1 column 1`, etc.). Now, when a Gemini host ships JSON RTK can't parse, `rtk hook gemini` exits non-zero — blocking the tool call — with only the bare generic message and no hint of *where* the payload is bad.

`run_gemini_inner` returning `Result<String>` keeps both properties: the tests only need `.ok()` on it, and production keeps the diagnostic.

src/hooks/init.rs
```diff
@@ -4269,7 +4289,29 @@ fn patch_gemini_settings(
...
+                    // GEMINI.md, and the integrity baseline by this point).
+                    // Treat it like PatchMode::Skip: warn, tell the user how
+                    // to patch it themselves, and leave the file untouched.
+                    eprintln!(
+                        "Warning: failed to parse {} as JSON: {}",
+                        settings_path.display(),
+                        e
+                    );
+                    print_gemini_manual_setup(&settings_path);
+                    return Ok(());
```
KuSh: Non-blocking, but worth a thought: `return Ok(())` here means control falls back into `run_gemini`, which unconditionally prints `"Gemini CLI hook installed (global)."` and `rtk init` exits 0 — even though the BeforeTool hook was never registered and no command will ever be rewritten.

That's defensible for `PatchMode::Skip` (the user asked to skip) but less so for an unexpected parse failure. Compounding it: this warning goes to **stderr** while `print_gemini_manual_setup` and the "installed" line go to **stdout**, so anyone capturing stdout — or scripting on the exit code — sees only the success text.

Either a non-zero exit, or threading a flag through so the summary reads "hook script installed, settings.json NOT patched", would make the outcome honest. Not a merge blocker; the data-loss fix this replaced was the important part.

src/cmds/system/json_cmd.rs
```diff
@@ -78,21 +74,32 @@ pub fn run_stdin(max_depth: usize, schema_only: bool, verbose: u8) -> Result<()>
...
+        filter_json_string(content, max_depth)?
     } else {
-        filter_json_compact(&content, max_depth)?
+        filter_json_compact(content, max_depth)?
     };
-    let shown = never_worse(&content, &output);
-    println!("{}", shown);
-    timer.track("cat - (stdin)", "rtk json -", &content, shown);
-    Ok(())
+    Ok(never_worse(content, &output).to_string())
```
KuSh: Non-blocking: `never_worse` returns a borrowed `&str` and the previous code printed that borrow directly, so `.to_string()` here adds a full extra copy of the input on the fallback path.

Concretely: `rtk json` on a large minified file, where the pretty-printed form is bigger than the raw, now holds file content + filtered output + a third full copy at peak. Against the repo's <5MB memory target that's worth avoiding — returning `Cow<'_, str>` from `render_json` (or keeping the borrow by inlining the strip at the two call sites) preserves the dedup you just gained without the allocation.

src/cmds/system/json_cmd.rs
```diff
@@ -48,18 +49,13 @@ pub fn run(file: &Path, max_depth: usize, schema_only: bool, verbose: u8) -> Res
...
-    let shown = never_worse(&content, &output);
+    let shown = render_json(&content, max_depth, schema_only)?;
     println!("{}", shown);
     timer.track(
         &format!("cat {}", file.display()),
         "rtk json",
         &content,
-        shown,
+        &shown,
     );
```
KuSh, line 60: Nitpick, take it or leave it: `timer.track` still passes the original un-stripped `content` as the input baseline while `shown` is measured post-strip, so a BOM-prefixed file reports 3 bytes of "savings" that are really just the BOM.

Arguably correct as-is — the BOM genuinely was in the bytes RTK read — so this is only worth changing if you'd rather the ratio reflect filtering alone. Same in `run_stdin` below.

KuSh (changes requested):
Thanks for the quick turnaround — I re-reviewed `b6f354c` and **all three earlier findings are properly fixed**; I've resolved those threads. The `run_gemini_inner` extraction in particular is the right shape: routing production `run_gemini` through it (rather than mirroring the parse in a test) means a dropped BOM strip fails for real.

Re-ran the gate on the head: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --all` all clean on a default environment.

One blocker and three non-blocking notes, all on the new commit:

- **Blocking — `test_gemini_strips_utf8_bom` is non-hermetic.** It reads the developer's real `~/.gemini/settings.json`; I reproduced a red test on a machine that uses Gemini CLI. Details inline.
- Non-blocking: the lost serde diagnostic in `run_gemini`, the "installed" summary printed after a failed settings patch, and an avoidable full copy of the input in `render_json`. Plus one nitpick on the tracking baseline.

Happy to see the rest land as-is if you'd rather keep the PR tight — only the first one needs to change before merge.

kylehgc:
All addressed as of ed8b0eb — the blocker plus the three notes; one nit deliberately declined, reasoning below.

**The hermetic test (blocker):** went with your second option — rules injected through a shared `run_gemini_inner_impl`, the same pattern as `gemini_render`/`decide_with_rules`, with production `run_gemini_inner` and the test-only `run_gemini_inner_with_rules` both routing through the one parse/strip/render implementation. So the test is deterministic under any HOME while still exercising the identical production BOM path — dropping the strip inside the shared impl fails it. Reproduced your exact scenario (a HOME with `run_shell_command(git status)` allow-listed) and ran the full suite under it: green, including this test.

**Serde diagnostic:** `run_gemini_inner` returns `serde_json::Result<String>`, so the line/column detail is chained under the context message again; pinned by a test asserting the diagnostic survives.

**Honest summary:** `patch_gemini_settings` signals the parse-failure outcome distinctly, and the summary now prints `settings.json: NOT patched (existing file could not be parsed; see warning above)` instead of the unconditional installed line. Kept exit 0 — the point of the graceful-degrade was that the rest of the install genuinely succeeded — but stdout no longer claims more than happened. `PatchMode::Skip` is untouched.

**Fallback copy:** `render_json` returns `Cow<'_, str>`; the raw-fallback branch borrows, only the genuinely-filtered branch allocates, and the test asserts `Cow::Borrowed` on the fallback path.

**The `timer.track` baseline — acknowledged and deliberately left as-is:** the baseline is meant to be the bytes the raw command would have emitted, and `cat` on a BOM'd file genuinely emits those bytes; counting them keeps the metric consistent with how every other filter accounts. Your framing ("arguably correct as-is") matches our read, so we kept the current semantics rather than special-casing 3 bytes.

Gate: fmt clean, clippy zero warnings, 2517 tests passing — under both a clean HOME and your dirty-HOME reproduction.

src/hooks/hook_cmd.rs
KuSh: Fixed in ed8b0eb, and I re-ran your reproduction to confirm: with `~/.gemini/settings.json` = `{"tools":{"allowed":["run_shell_command(git status)"]}}`, `test_gemini_strips_utf8_bom` now passes and the gemini test group is 13/13 green. Splitting `run_gemini_inner_impl` out and injecting the decision closure is a better fix than what I suggested — the test keeps exercising the identical parse/strip/render path production uses, and only the permission lookup is substituted. Resolving.

src/hooks/hook_cmd.rs
KuSh: Fixed in ed8b0eb. `run_gemini_inner` now returns `serde_json::Result<String>` and propagates with `?`, so `run_gemini`'s `.context(...)` has the serde error chained as its source again. `test_gemini_inner_preserves_serde_diagnostic` pins the line/column detail so it can't silently regress. Resolving.

src/hooks/init.rs
KuSh: Fixed in ed8b0eb. `patch_gemini_settings` returning `Ok(true)` only on the parse-failure path (every other non-patch outcome — already patched, Skip, declined at the Ask prompt, dry-run — stays `false`) is exactly the distinction that was missing, and the summary now prints `settings.json: NOT patched`. That also closes the stderr/stdout split I raised: a stdout-only capture no longer reads as unqualified success. Keeping exit 0 is a reasonable call for a graceful-degrade path. Resolving.

src/cmds/system/json_cmd.rs
KuSh: Fixed in ed8b0eb. `render_json` returning `Cow<'a, str>` keeps the fallback a borrow, and the test now asserts `matches!(shown, Cow::Borrowed(_))` so the allocation can't creep back in unnoticed. Resolving.

src/cmds/system/json_cmd.rs
KuSh: Leaving this as-is is the right call — the BOM genuinely was in the bytes RTK read, so counting it in the input baseline is defensible, and it was flagged as optional. Resolving so it does not linger.

KuSh (approved):
Approving. Three rounds in, everything raised is addressed and the BOM coverage is now genuinely complete: all six hook stdin entry points strip, and every non-test `serde_json::from_str` over a human- or tool-written file routes through `from_json_str`. The sites deliberately left raw — the `trust.rs` trust store, analytics JSONL, nested `toolArgs` strings, fixtures — are RTK/machine-written and correctly untouched.

Worth restating what the `patch_gemini_settings` change actually buys: the old `unwrap_or(json!({}))` would silently overwrite an unparseable `~/.gemini/settings.json` with a fresh `{}` plus the hook, destroying user config. That's a real data-loss fix, not just BOM tolerance.

Two things I liked in the review responses: routing production `run_gemini` through `run_gemini_inner_impl` rather than mirroring the parse in a test, so a dropped strip fails for real; and returning `Ok(true)` from `patch_gemini_settings` *only* on the parse-failure path, so the summary can distinguish it from Skip, already-patched, declined-at-prompt and dry-run.

**Disclosure — the last three commits are mine, pushed via maintainer edit rather than sent back for another round:**

- `9d78af2` — the three round-3 findings: `std::ptr::eq` on the `&str` fat pointers instead of `as_ptr()` (an address-only compare also matches a sub-slice, so a future truncating `never_worse` would silently re-expand the output); `run_droid` and its test entry point now share a `droid_payload` helper, closing a gap where droid's BOM strip had *zero* coverage; and `.with_context(...)` on the four remaining bare reads in `deps.rs`.
- `3865a84` — merge of `develop`. The conflict was in `hook_cmd.rs` only: `develop` added the Vibe hook exactly where this branch refactored `run_gemini`'s tail. Kept both sides, dropped the now-unused `print_allow`/`print_gemini`.
- `ac3389a` — `run_vibe` landed on `develop` mid-review as the sixth hook stdin entry point and the only one not stripping, so merging `develop` in would have quietly broken the invariant this PR exists to establish. One line plus a test.

Both new tests are mutation-verified: deleting the strip fails `test_droid_strips_utf8_bom` / `test_vibe_strips_utf8_bom` and nothing else. Gate on `ac3389a`: `cargo fmt --all -- --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --all` 2735 passed / 0 failed.

Please sanity-check those three commits — happy to drop or rework any of them if you'd rather they were separate.

Two follow-ups filed from this review, both pre-existing and out of scope here: #3767 (one malformed manifest aborts the whole `rtk deps` scan instead of degrading) and #3768 (`test_resolved_command_executes_known_command` asserts on rustup's exit status and fails whenever `HOME` is redirected).

kylehgc:
Sanity-checked all three commits — keep them all, and thanks for pushing directly rather than spending another round.

- `9d78af2`: the fat-pointer compare is strictly better than the `as_ptr()` version we wrote — comparing the whole slice means a future sub-slice-returning `never_worse` can't masquerade as the raw input. And `droid_payload` closes exactly the mirrored-copy test trap we'd just fixed for gemini; fair catch that we didn't sweep the other agents for the same pattern after that lesson.
- `3865a84`: verified the merge kept everything intact on our side — `run_gemini_inner`, the "NOT patched" summary path, and all strip sites present post-merge.
- `ac3389a`: right call — landing the merge without it would have broken the six-entry-point invariant this PR exists to establish.

Independently re-ran the gate on `ac3389a` here: fmt clean, clippy zero warnings, full suite 0 failed. Nothing to drop or rework — and thanks for the thorough rounds on this one; the review made it a much better change than the one we started with.

PR #3752, opened 2026-08-29
fix(git): preserve worktree repair and list semantics

### LINKED ISSUE #3710: `git worktree repair` is routed to list mode: prints the worktree list, exits 0, repairs nothing

arkgum:
## Summary

- route every non-list worktree subcommand to git unchanged, fixing repair and future action subcommands
- preserve explicit list arguments and return native output for --porcelain and -z while retaining filtering for human-readable lists
- add regression coverage for subcommand-slot detection, argument forwarding, machine-readable list modes, and repairing pointers after a repository move

Closes #3710

## Test plan

- [x] cargo fmt --all --check
- [x] cargo clippy --all-targets
- [ ] cargo test (not run locally; the full suite will run in CI)
- [ ] Manual testing: rtk git worktree repair output inspected

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -2132,35 +2132,64 @@ fn diffstat_row(line: &str) -> Option<String> {
     Some(format!("{} {}{}", path, count, sign))
 }
 
-fn run_worktree(args: &[String], verbose: u8, global_args: &[String]) -> Result<i32> {
-    let timer = tracking::TimedExecution::start();
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum WorktreeMode {
+    Action,
+    FilteredList,
+    NativeList,
+}
 
-    if verbose > 0 {
-        eprintln!("git worktree list");
+fn worktree_mode(args: &[String]) -> WorktreeMode {
+    if !matches!(args.first().map(String::as_str), None | Some("list")) {
+        return WorktreeMode::Action;
     }
 
-    // If args contain "add", "remove", "prune" etc., pass through
-    let has_action = args.iter().any(|a| {
-        a == "add" || a == "remove" || a == "prune" || a == "lock" || a == "unlock" || a == "move"
-    });
+    if args
+        .iter()
+        .any(|arg| matches!(arg.as_str(), "--porcelain" | "-z"))
+    {
+        WorktreeMode::NativeList
+    } else {
+        WorktreeMode::FilteredList
+    }
+}
 
-    if has_action {
-        let mut cmd = git_cmd(global_args);
-        cmd.arg("worktree");
-        for arg in args {
-            cmd.arg(arg);
-        }
+fn build_worktree_command(args: &[String], global_args: &[String]) -> Command {
+    let mut cmd = git_cmd(global_args);
+    cmd.arg("worktree");
+    if args.is_empty() {
+        cmd.arg("list");
+    } else {
+        cmd.args(args);
+    }
+    cmd
+}
+
+fn run_worktree(args: &[String], verbose: u8, global_args: &[String]) -> Result<i32> {
+    let timer = tracking::TimedExecution::start();
+    let mode = worktree_mode(args);
+    let display_args = if args.is_empty() {
+        "list".to_string()
+    } else {
+        args.join(" ")
+    };
+    let original_cmd = format!("git worktree {display_args}");
+    let rtk_cmd = format!("rtk git worktree {display_args}");
+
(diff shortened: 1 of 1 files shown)


amandeavor (approved):
Tested this on Windows and walked through the routing changes. All six worktree-related tests pass here, including the integration test that moves a repository and verifies `git worktree repair` updates the linked worktree pointer.

The mode split also looks right: only the actual `list` subcommand is filtered, unknown or future subcommands pass through as actions, and `--porcelain` / `-z` keep native output. `cargo fmt --all --check` and `cargo clippy --all-targets` are clean on this branch.

Looks good to me.

arkgum:
Hi maintainers — a gentle follow-up on this fix for #3710. It preserves native behavior for worktree actions and machine-readable list output. The PR has an approving review with Windows verification, but the Actions run is still waiting for maintainer approval.

Could someone approve the CI run and review it when time permits? It also covers the lost `prune --dry-run` report mentioned in #3838; that issue's separate exclusion-matching fix remains out of scope. Happy to address any feedback. Thanks!

KuSh:
Closing in favour of #3681, which fixes the same #3710 behaviour: repair and any other non list subcommand pass through to git unchanged, and list with --porcelain or -z returns native output. It also carries the shared argument tokenizer that the rest of that fix is built on. Review and comments are very welcome on #3681.

PR #3753, opened 2026-08-29
fix(test): trust Node test status over names

### LINKED ISSUE #3741: rtk test mislabels a fully-passing suite as [FAIL] when a test name contains the substring FAIL
## Bug: `rtk test` mislabels a fully-passing suite as `[FAIL]` when a test **name** contains the substring `FAIL`

`rtk test <cmd>` appears to classify pass/fail by substring-matching the word `FAIL` against the raw output text (which includes test *names*), instead of parsing the actual pass/fail status glyph (`✔`/`✖`) or the reporter's structured summary counters (`ℹ pass N` / `ℹ fail N`) that `node --test`'s spec reporter prints.

Any test whose **name** happens to contain the uppercase substring `FAIL` (e.g. a regression test that is itself *about* failure-handling, a very common naming pattern: `"...rejected: FAIL..."`, `"...FAILURES section..."`) gets the entire run wrongly summarized as `[FAIL] FAILURES:`, even though every individual test passed (`✔`) and the reporter's own summary says `fail 0`.

This is dangerous for the exact use case `rtk test` exists for: an agent/human skimming compacted output would read `[FAIL] FAILURES:` and believe the suite is broken, when it is 100% green.

### Repro (minimal, 2 files)

`package.json`:
```json
{
  "name": "rtk-bug-repro",
  "version": "1.0.0",
  "private": true,
  "scripts": { "test": "node --test \"*.test.js\"" }
}
```

`repro.test.js`:
```js
const { test } = require("node:test");
const assert = require("node:assert");

test("contains uppercase word FAIL in the name", () => {
  assert.ok(true);
});
```

### Commands

```
$ npm test
> rtk-bug-repro@1.0.0 test
> node --test "*.test.js"

✔ contains uppercase word FAIL in the name (0.45ms)
ℹ tests 1
ℹ suites 0
ℹ pass 1
ℹ fail 0
...
```

```
$ rtk test npm test
[FAIL] FAILURES:
  ✔ contains uppercase word FAIL in the name (0.45ms)

OUTPUT (last 5 lines):
  ℹ fail 0
  ℹ cancelled 0
  ℹ skipped 0
  ℹ todo 0
  ℹ duration_ms 57.6223
```

Note the contradiction in `rtk`'s own output: it prints `[FAIL] FAILURES:` in the header, lists a test marked `✔` (pass) as the "failure," and then shows `ℹ fail 0` in its own OUTPUT tail — three signals in the same block that disagree with each other.

### Isolating the trigger

- `"...contains lowercase word failure..."` → **not** misclassified (correct).
- `"...contains uppercase word FAIL..."` → **misclassified** as `[FAIL]`.

So the match appears to be a case-sensitive substring check for `FAIL` against the full command output (which includes test names as printed by the reporter), not the actual reporter-emitted status per test or the `ℹ fail N` summary line.

### Real-world trigger (not just a synthetic case)

Found via a regression test whose name is legitimately about testing failure-shape handling:
```
✔ REGRESSION-FIXTURE: real captured 48-failure selftest shape (cloudflare skills) collapses to fully-passing areas hidden, failing area + every FAIL line kept verbatim (0.45ms)
```
This test passes; the whole suite (196 other passing tests, 0 failures per `node --test`'s own summary) gets relabeled `[FAIL] FAILURES:` by `rtk test`, purely because the test's own name contains `FAIL`.

### Environment

- `rtk 0.42.4` (installed via `cargo install --git https://github.com/rtk-ai/rtk`)
- Windows 11, PowerShell 7
- `node --test` (Node v24.14.1) as the underlying test runner

### Expected

`rtk test` should classify pass/fail using the reporter's structured summary (`ℹ pass N` / `ℹ fail N`) or the actual per-test status glyph, never a substring match against free-text test names — the exact anti-pattern already flagged in #3534 (`rtk log` misclassifying JSONL records as errors by prose substring match) and #2545 (false positives from text matching), just hitting `rtk test` this time.

### Suggested fix

Parse `node --test`'s TAP/spec summary block (`ℹ pass N`, `ℹ fail N`) as the source of truth for the `[FAIL]`/`[PASS]` header, and use each line's leading `✔`/`✖` glyph (not its trailing free-text name) to decide which individual lines belong in the `FAILURES:` section.

amandeavor:
## Summary

- recognize Node's structured `ℹ fail N` summary instead of treating every `FAIL` substring as a failed test
- require a complete numeric Node summary block before switching away from the Jest fallback
- use the leading `✕` / `✖` status glyph to identify actual Node test failures
- preserve the existing Jest fallback when no complete Node summary is present
- include Node summary counters in the compact output

Closes #3741

## Tests

- passing Node test whose name contains `FAIL`
- real failing Node test whose name also contains `FAIL`
- partial Node-like log line followed by a genuine Jest failure
- `cargo fmt --check`
- `cargo test node_test_`
- `cargo test partial_node_summary_log_does_not_hide_jest_failure`

### DIFF
--- a/src/cmds/rust/runner.rs
+++ b/src/cmds/rust/runner.rs
@@ -178,6 +178,41 @@ fn filter_errors(output: &str) -> String {
     result.join("\n")
 }
 
+fn node_test_summary_key(line: &str) -> Option<&str> {
+    let summary = line.trim_start().strip_prefix('ℹ')?.trim_start();
+    let mut fields = summary.split_whitespace();
+    let key = fields.next()?;
+    let value = fields.next()?;
+
+    if fields.next().is_some() {
+        return None;
+    }
+
+    let is_valid_value = if key == "duration_ms" {
+        value.parse::<f64>().is_ok()
+    } else {
+        value.parse::<u64>().is_ok()
+    };
+
+    if is_valid_value
+        && matches!(
+            key,
+            "tests"
+                | "suites"
+                | "pass"
+                | "fail"
+                | "cancelled"
+                | "skipped"
+                | "todo"
+                | "duration_ms"
+        )
+    {
+        Some(key)
+    } else {
+        None
+    }
+}
+
 fn extract_test_summary(output: &str, command: &str) -> String {
     let mut result = Vec::new();
     let lines: Vec<&str> = output.lines().collect();
@@ -189,10 +224,10 @@ fn extract_test_summary(output: &str, command: &str) -> String {
         || command.contains("yarn test")
         || command.contains("node --test");
     let is_go = command.contains("go test");
-    let has_node_test_summary = lines.iter().any(|line| {
-        line.trim_start()
-            .strip_prefix('ℹ')
-            .is_some_and(|summary| summary.trim_start().starts_with("fail "))
+    let has_node_test_summary = ["tests", "pass", "fail"].iter().all(|expected| {
+        lines
+            .iter()
+            .any(|line| node_test_summary_key(line) == Some(expected))
     });
 
     let mut failures = Vec::new();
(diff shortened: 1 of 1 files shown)


amandeavor:
I tightened the Node summary detection after spotting one more edge case: a partial log containing something like \ℹ fail 0\ could previously make us ignore a real Jest \FAIL\ line. It now only switches to Node summary handling when a complete numeric summary block is present. I added a regression test for that case and reran the focused tests, formatting, and clippy checks locally. Everything passes. The latest commit is \5712b5d\.

KuSh:
Thanks for working out the Node summary format and for the test cases around it. #988, merged on 2026-09-03, moved the err and test runners out of src/cmds/rust/runner.rs into src/core/runner.rs and replaced the command sniffing inside extract_test_summary with a TestEcosystem enum carrying one branch per runner, so the function this PR patches no longer exists in that shape and the tests cannot apply. The report in #3741 is still open: the Jest branch at src/core/runner.rs:765 still counts any line containing FAIL as a failure, so a passing test whose name holds FAIL is still reported as failing. A follow-up adding a Node branch to TestEcosystem, or anchoring the Jest match, would be welcome there. Please comment here or reopen if a part of this was missed.

PR #3758, opened 2026-08-29
fix(tests): stop the test suite writing to the developer's own rtk data

KuSh:
## Problem

Running `cargo test` writes into the contributor's own `~/.local/share/rtk/`, in several ways.

It inserts rows into `history.db`, which `rtk gain` then counts as real usage. And it fills the tee spool with its own output — the spool is capped at `DEFAULT_MAX_FILES = 20` and `write_tee_file` rotates after every write, so a run evicts the raw output the developer was keeping there for recovery. Measured on this branch's parent: **16 rows inserted, 16 tee files written, 16 of theirs deleted**, every run.

The file count alone hides the second one, since rotation pins it at the cap; it only shows up by diffing the filenames.

It is not only those two. The telemetry salt, the trust store and the hook-warning marker each resolved the data directory independently, so a contributor who had never run rtk had `cargo test` create their data dir and mint `.device_salt` — the permanent identity rtk uses for telemetry from then on — and a contributor with an outdated hook had the suite consume their once-a-day warning, silencing it for the rest of the day.

This is not only about polluted statistics. Several rtk children writing the same database concurrently corrupted a 35 MB one outright:

```
$ sqlite3 ~/.local/share/rtk/history.db "PRAGMA integrity_check;"
*** in database main ***
Tree 2 page 8287 cell 449: 2nd reference to page 5617
wrong # of entries in index idx_pf_timestamp
```

After that the failure is self-sustaining: `rtk gain` stops working and the tracking tests themselves fail with `database disk image is malformed`, so the suite cannot pass again until the user repairs it by hand. Recovery via `.recover` cost 57 of 55828 rows.

## Fix

The data directory is reached by two routes, and they need different fixes.

**One place that answers where rtk's data lives.** Seven call sites resolved `<data_local_dir>/rtk` independently, and only two of them were covered. `constants::data_dir()` is now that place, with the `cfg(test)` redirect in it, and the tracking database, tee spool, telemetry salt and marker, trust store and hook-warning marker all go through it. `data_dir_under(root)` serves the three callers that carry on when the platform gives no answer, so the fallback path cannot lose rtk's own directory segment.

**Spawned children.** A test that runs the rtk binary cannot redirect the child from inside it — the child is built without `cfg(test)` and resolves its data directory exactly as a normal invocation does, so the environment is the only channel. `test_isolation::rtk_command`, and its `tests/` twin `common::rtk_command` (rtk is a binary crate, so `tests/` cannot reach into it), set `RTK_DB_PATH` and `RTK_TEE_DIR` for the two files a filter run writes, plus `XDG_DATA_HOME`, `XDG_CONFIG_HOME` and `HOME` for everything that resolves through `dirs` instead. Every spawn site now uses one of them.

Windows resolves its known folders through the shell API rather than the environment, so there only the two named files are redirected — noted in the helper rather than papered over.

**In-process.** A test that exercises a *command* path instead reaches `TimedExecution::track` and `tee_raw`, which resolve their own destinations and take no argument — so nothing at the call site can reach them. `open_and_prepare` now goes through `resolve_db_path`, and `get_tee_dir` has the matching branch: in a test build both land in the same scratch directory unless the run names a destination through `RTK_DB_PATH` or `RTK_TEE_DIR`.

Neither consults the config file, and that is deliberate on both sides: `Config::load()` reads the developer's real `~/.config/rtk/config.toml`, so honouring `[tee] directory` would rotate the very spool this is protecting. Verified by setting `[tee] directory` to a populated directory and running the full suite — all 20 files still present, byte-identical listing, suite green. `get_db_path` keeps its own semantics so `test_db_path_env_and_default` still pins the real default resolution, and production compiles the branches away.

**A vacuous test, found on the way.** `test_rewrite_rtk_disabled_subprocess_warns` skipped itself on every run. It compared the binary's mtime against the test harness's and bailed if the binary looked older — but cargo links the harness *after* the bin, so it always did. Measured here: bin `1788517902`, harness `1788517930`. CI runs `cargo test --all` with no separate `cargo build`, so it had never executed there. `rtk_binary_is_built()` compares the binary against the newest change under `src/` instead. That comparison still earns its keep: `cargo test --all` rebuilds and uplifts the bin because the integration targets need it, but `cargo test --bin rtk` builds only the harness, so without it the test would spawn whatever an earlier build left behind. Confirmed in both directions by breaking the expected string: with a stale binary the test skips, and after a rebuild it fails on the child's actual stderr — so it is genuinely spawning.

**Keeping it that way** is mostly the compiler's job. `rtk_bin` is private, so nothing outside `test_isolation` can reach the binary's path at all — and it now takes the target and profile directory from the running harness (`current_exe()`, which lives one level below the binary), so it finds it under a custom target directory, a `--target` triple, a named profile and on Windows. Verified by building under `CARGO_TARGET_DIR` and confirming the spawn test genuinely runs there. A test that is not `#[ignore]`d and has to skip quietly when the binary has not been built asks `rtk_binary_is_built()`; `rtk_command` asserts on it, which also folds the four hand-written copies of that existence check into one.

A scan covers the channel visibility cannot reach: `tests/`, where cargo supplies the binary through `CARGO_BIN_EXE_rtk` rather than through this module. It fails the suite on any file outside the two constructors containing that literal, and asserts both constructor files were actually visited — the sources are reached through a compile-time path, so a run that cannot see them would otherwise scan nothing and report success.

The scan earned its place: it caught `diff_byte_accuracy_test.rs`, which reaches the binary through a `const RTK_BIN` that a search-and-replace over the inline form had missed, and four unit tests in `src/` (`main.rs`, `cmds/git/git.rs`, `cmds/system/read.rs`, `discover/registry.rs`) that built `target/debug/rtk` by hand.

An earlier revision also matched hand-written `target/debug/rtk` paths in their various spellings. That is deliberately gone: it was an open-ended pattern list that kept needing another entry, and with `rtk_bin` private a `src/` test would have to write the path out by hand to get past the compiler — a thing review catches. The exact rule that remains is complete for the channel it guards.

**The scratch directory.** A directory of its own rather than a file placed straight in the temp root, because rtk chmods its database's parent to 0700 via `create_private_dir` — which would otherwise land on the shared root. `tempfile` picks the name, so nothing can pre-empt the path.

It outlives every test in the process, so it is held in a `static` — and Rust never drops statics, which leaves `TempDir`'s own guard unable to remove it. An `atexit` handler does that instead, and the harness leaves through `exit(3)` whether the run passed or failed. Two cases leave the directory in place: a run killed by a signal, and any non-Unix platform, where there is no `atexit` to hand this to.

The two `rtk_command`s are unavoidably separate — they construct the binary path differently, and rtk has no `lib.rs` for `tests/` to import from. Everything behind them is not: `test_isolation::scratch` names nothing from the crate, so `tests/common/mod.rs` compiles that one file by path instead of carrying a copy.

## Verification

Over a full `cargo test --all`, by max rowid and by diffing the tee filenames:

| | rows into the real DB | its mtime | tee files written / evicted | mode of the temp root | scratch dirs left |
|---|---|---|---|---|---|
| before this PR | 16 | modified | 16 / 16 | — | — |
| after | **0** | **unchanged** | **0 / 0** | **1777, unchanged** | **0** |

The broader check, since the database and the spool turned out not to be the only things at risk: comparing **every entry under `~/.local/share/rtk` and `~/.config/rtk` by mtime, mode, size and file set**, a full run now changes nothing at all, and leaves nothing named `rtk*` in the temp root.

A control measurement over the same interval confirmed the database does not grow on its own, so those 16 rows were attributable to the suite rather than to concurrent activity. `cargo test --all -- --ignored` also inserts 0 rows, and leaves 0 scratch directories despite exiting non-zero.

The guard was checked against live evasions as well as its own unit test: a file rebuilding the `target/debug/rtk` path by hand is reported, and so is one that calls `rtk_command()` and *also* spawns raw — while a file building a `target/debug` path for something other than the binary is not.

- `cargo fmt --all --check`, `cargo clippy --all-targets` (zero warnings), `cargo test --all` (2999 + integration) all pass.
- Rebased onto current `develop` (`f83c0cf`).

## Related issues

- **#2106** — *telemetry tests write a real `.device_salt` to the user's data dir* — is fixed by this PR. Its three named tests (`test_device_hash_is_stable`, `test_device_hash_is_valid_hex`, `test_salt_is_persisted`) reach `salt_file_path()`, which now resolves through `constants::data_dir()`. Verified with that issue's own repro: pointing the data dir at an empty tree and running the telemetry tests leaves it empty — no `rtk/` directory, no salt. Left without a closing keyword so the maintainer decides whether it closes on merge.
- **#3874** — the Windows half of this, split out rather than bundled in. `dirs` resolves Windows known folders through the shell API, so a spawned child there is isolated for `RTK_DB_PATH` and `RTK_TEE_DIR` but not the four `dirs`-based paths. Documented in `redirect_rtk_data` rather than implied.

Unrelated pre-existing failure noticed while running `--ignored`: `test_git_status_not_a_repo_exits_nonzero` asserts on the English text of git's "not a git repository" message, so it fails under a non-English locale. Left alone here.


KuSh:
Needs a rebase before merging: a new integration test is landing that this branch's guard will reject.

#3519 is approved and adds `tests/diff_byte_accuracy_test.rs`, which spawns the binary via `CARGO_BIN_EXE_rtk` and does not set `RTK_DB_PATH`. Merging it into this branch locally and running the guard:

```
---- core::db_isolation_tests::tests_that_spawn_rtk_isolate_the_database stdout ----
these files spawn the rtk binary without setting RTK_DB_PATH, so the child writes to
the user's real ~/.local/share/rtk/history.db — ...
  tests/diff_byte_accuracy_test.rs
```

So the rebase needs to convert that file to `common::rtk_command()` alongside the eight it already covers, or the suite fails on first merge. Worth noting the guard did its job here — it caught the new file automatically, which is the whole point of scanning sources rather than listing them.

Second one to fold in while rebasing: #3781 adds `tests/diff_exit_code_test.rs`, which sets `RTK_DB_PATH` inline on the `Command`. The guard passes it, since it only requires the string to appear in the file, but it should use the shared helper too — otherwise the one convention this PR establishes has an exception in the same release.

Neither is a change to the approach here, just coverage that moved while this was open.

bharvey42:
Reviewed this on Windows 11 (`stable-x86_64-pc-windows-gnu`), since #3228 was closed in its favour and the leak I measured there was Windows-side. Short version: **the isolation works** — it closes the leak completely on this machine, including the spawned-child path. Three things I hit that you may not see on Linux/macOS.

## It works

Method: PR head (e6e066c) vs its own parent (5e0f92c), counting rows in the real `~/AppData/Local/rtk/history.db` before/after `cargo test --all`. This machine runs other Claude sessions against the same database, so a raw total is meaningless here — I filtered on the `project_path` column and counted only rows attributable to the worktree under test.

| | rows written to the real `history.db` |
|---|---|
| base 5e0f92c | **56** |
| PR head | **0** |

The base rows are the fixtures, unmistakable: `find /definitely/missing/xyz -mtime +0`, `find . -name secret.txt`, `git stash show`, `git worktree list`, `find <temp>\.tmpKrq4GF -name '*.txt'`, `cat <temp>\.tmp9SKZ6a.rs`. By filter: 33 `rtk read`, 15 `rtk find`, 4 `rtk git stash/status/worktree`, 2 `rtk json`, 2 `rtk diff`. On the PR head, zero — and `.device_salt`, `.hook_warn_last` and `trusted_filters.json` were all untouched too. `--no-fail-fast`, so the integration targets that spawn the binary actually ran.

3678 passed / 3 failed (the base run: 3676 / 3 — the +2 are this PR’s own guard tests), the 3 being the Windows-only failures that reproduce identically on pristine `develop` (`find_cmd::run_from_args_propagates_find_exit_status`, `core::stream::test_read_lines_lossy_preserves_lines_after_invalid_utf8`, `hooks::hook_cmd::test_claude_pipeline_rewrites_only_safe_final_stage`). No regressions.

## 1. Windows CI is red, and it is a class rather than a single error

`test (windows-latest)` currently fails with:

```
error: unused import: `Command`
 --> tests\read_window_bytes_test.rs:3:20
error: could not compile `rtk` (test "read_window_bytes_test")
```

I reproduced that byte-for-byte locally. Cause: the PR moves the other four tests in that file onto `common::rtk_command()`, and the only remaining user of `Command` is `rewritten_head_spellings_match_native_on_non_utf8_files`, which is `#[cfg(unix)]`. On Windows the import is genuinely unused, and `[lints.rust] warnings = "deny"` makes it fatal.

Fixing only that uncovers the next one, which is the more general shape:

```
error: function `rtk_command` is never used        --> tests\common\mod.rs:19:8
error: function `redirect_rtk_data` is never used  --> src\core\test_isolation\scratch.rs:25:8
error: function `scratch_dir` is never used        --> src\core\test_isolation\scratch.rs:39:8
error: function `remove_at_exit` is never used     --> src\core\test_isolation\scratch.rs:60:4
error: could not compile `rtk` (test "dotnet_double_dash_test")
```

Of the 20 integration targets that now carry `mod common;`, about a dozen have their sole `rtk_command()` call inside a `#[cfg(unix)]` test — `dotnet_double_dash`, `copilot_selfheal`, `gh_pr_checks_failure`, `grep_context`, `hook_warning_scope`, `native_test_expression`, `search_compress`, `search_error`, `search_faithful`, `search_stdin`, `signal_flush`, `stderr_only_failure`. Each compiles `common` (and `scratch.rs` through it) with nothing using it, so each is a separate `-D dead-code` failure surfacing one at a time.

Two edits made the whole suite compile here:

```rust
// tests/read_window_bytes_test.rs
use std::process::{Output, Stdio};
#[cfg(unix)]
use std::process::Command;
```

```rust
// tests/common/mod.rs — each integration target compiles this separately
// and uses only the part it needs.
#![allow(dead_code)]
```

The numbers above are from a tree with exactly those two edits applied and nothing else.

## 2. One call site the sweep missed — and it deletes

`telemetry_cmd.rs` is not in this PR, and `run_forget` still resolves the database itself rather than through `constants::data_dir()`:

```rust
// src/core/telemetry_cmd.rs:220
let db_path = dirs::data_local_dir()
    .unwrap_or_else(|| std::path::PathBuf::from("."))
    .join(super::constants::RTK_DATA_DIR)
    .join(super::constants::HISTORY_DB);
if db_path.exists() {
    match std::fs::remove_file(&db_path) { ... }
}
```

Every sibling got converted — `trust.rs:51`, `hook_check.rs:140`, `tee_file.rs:111`, `telemetry.rs:222`, `telemetry.rs:552`, `tracking.rs:1562` — which is what makes this one stand out. Nothing reaches it today (`TelemetrySubcommand::Forget` has no test), so it is latent rather than a live leak. But it is the one path that *removes* the developer's database instead of appending to it, and `tests_that_spawn_rtk_isolate_their_data` scans for `CARGO_BIN_EXE_rtk`, so it would not catch a future in-process test for `telemetry forget`.

Since the guard test is already a source scan, it could cover this shape cheaply: reject `dirs::data_local_dir()` anywhere under `src/` except `constants.rs`. That turns "remember to use `data_dir()`" into something the suite enforces.

## 3. On Windows the scratch directories are never reclaimed

`remove_at_exit` is `#[cfg(unix)]`, and `tempfile`'s guard is dropped by `.keep()`, so nothing removes the directory on Windows. It is per test *binary*, not per run — a single `cargo test --all --no-fail-fast` here went from 2 to 11 `rtk-test-*` directories under `%TEMP%`, each with its own `rtk/history.db`. The doc comment says as much, so I assume this is known; flagging the magnitude since it accrues every run on every Windows contributor's machine, and `%TEMP%` is not cleared as reliably as `/tmp`.

Not a blocker for what the PR sets out to do — just the last thing standing between this and "the suite leaves nothing behind".

---

Also worth noting `semgrep security scan` is red (2 findings on the changed files); I did not dig into which rules.

Happy to re-run any of this, or to push the two compile fixes as a PR against your branch if that is easier than carrying them here.

KuSh:
Thank you @bharvey42! The `project_path` filter to get an attributable count on a machine that runs rtk concurrently is a better measurement than the one I had, and it is the half I could not do from Linux at all. Both compile fixes were exactly right; I have pushed them along with the rest.

**1. Windows CI.** Fixed as you described, and your framing as a class rather than a single error is what made it worth fixing properly:

```rust
// tests/read_window_bytes_test.rs
use std::process::{Output, Stdio};
#[cfg(unix)]
use std::process::Command;
```

`tests/common/mod.rs` now carries `#![allow(dead_code)]` with the reason above it, since every integration target compiles the module separately and uses only the part it needs.

I also set up a cross-check so this class stops reaching CI: `cargo check --target x86_64-pc-windows-gnu --all-targets` in a container, a couple of seconds once warm. It reproduced both of your errors byte-for-byte before the fix and is clean after. Running it against the rebased branch turned up two more files that had landed on `develop` since my branch point and spawn the binary directly — `tests/trae_hook_test.rs` and `tests/windows_child_quoting_test.rs` — so those are converted too.

**2. `telemetry_cmd.rs`.** Converted, and I took the enforcement suggestion: there is now a second scan rejecting `dirs::data_local_dir()` anywhere under `src/` outside `constants.rs`. It found one more caller while I was at it — `core/retriever.rs`, added to `develop` after my branch point, which had grown its own hand-rolled `cfg(test)` redirect into the temp root. It now goes through `data_dir()` like everything else, which also gets it the scratch cleanup.

Worth being precise about scope, since the two overlap but are not the same: this makes `run_forget` resolve through `data_dir()`, so it is test-isolated and consistent with its siblings. It does **not** make it honour `RTK_DB_PATH` or `config.tracking.database_path` — that is #2956, and still open.

**3. Windows scratch directories.** Recorded on #3874 with your measurement, rather than guessed at here. Your 2 → 11 figure is the useful part: the doc comment says the behaviour is known, but not that it accrues per test binary per run.

I sketched a shape there — sweep sibling `rtk-test-*` directories at creation, which suits Windows precisely because it refuses to delete a file with an open handle, so a live run's directory would be skipped and the rest reclaimed. I have deliberately not implemented it: that safety argument rests on file-locking semantics I can only type-check from Linux, not exercise, and this PR has already had one round where something I could not test turned out to behave differently than I assumed. If you have the machine and the inclination it would be a good standalone change; equally happy to leave it sitting on the issue.

**semgrep.** Two findings, both mine: `unsafe-block` on the `libc::atexit` call and `dynamic-command-execution` on `Command::new(bin)` in `rtk_command`. Suppressed with justifications in the repo's existing `// nosemgrep:` style. I reproduced the CI job locally with the same `--baseline-commit --error` invocation to confirm it comes back zero rather than assuming the suppressions landed — the first attempt did not, because I had split the justification across two lines and only the immediately preceding line counts.

The branch is rebased onto current `develop` (it had gone conflicting), which took one conflict in `hook_warning_scope_test.rs` where `develop`'s newer `isolating_env` helper supersedes what I had inline — I took `develop`'s, it pins `XDG_CONFIG_HOME` too.

On the offer to push a PR against the branch: genuinely appreciated, and no need this time since the fixes are small and now in. If you would rather review the Windows-side work as its own change in future, that suits me fine too — you have the machine and I do not, and that has been the binding constraint on this PR throughout.

Issue #3761, opened 2026-08-29
ci.yml security jobs: Rust steps on shell-only PRs, unguarded semgrep baseline


KuSh:
> **Partly blocked.** Item 1 describes `ci.yml` as it exists on branch
> `ci/path-filtering-189` (#3763) — the `changes` job, `.github/security-paths.conf`,
> the `Shell script scan` step and `needs.changes.outputs.build` are not on
> `develop` today. Item 2 applies to `develop` as it stands. Both are held here
> until #3763 merges so they land as one pass over the security jobs.

## 1. Security Scan runs the full Rust toolchain for shell-only changes

`ci.yml`'s `Security Scan` job runs for script-only changes as well as code
changes, but it always executes its Rust-specific steps — including
`cargo install cargo-audit` and a full `cargo clippy --all-targets` build. For a
PR that touches only `.claude/hooks/*.sh`, that is roughly five minutes of work
whose output cannot say anything about the change.

`.github/security-paths.conf` routes `.claude/hooks/` to this job even when the
build gate skips the Rust pipeline, because those shell scripts run on
contributors' machines and semgrep is Rust-only. That routing is correct, but
the job was written for one caller and now has two.

### Step breakdown for a shell-only change

| # | Step | Useful here? |
|---|------|--------------|
| 1 | `dtolnay/rust-toolchain@stable` | no |
| 2 | `Swatinem/rust-cache@v2` | no |
| 3 | `Install cargo-audit` | no |
| 4 | `Cargo Audit (CVE check)` | no — dependencies unchanged |
| 5 | `Critical files check` | yes, cheap |
| 6 | `Shell script scan` | **yes — the reason the routing exists** |
| 7 | `Dangerous patterns scan` | no — Rust-only regex |
| 8 | `New dependencies check` | no — `Cargo.toml` unchanged |
| 9 | `Clippy security lints` | no |
| 10 | `Summary verdict` | yes, cheap |

### Proposed fix

Add a step-level condition to steps 1–4 and 7–9:

```yaml
        if: needs.changes.outputs.build == 'true'
```

`needs` is available in step contexts, and the `security` job already has
`needs: [changes, clippy]`. Steps 5, 6 and 10 stay unconditional, so a shell-only
PR gets the scan it was routed here for in seconds instead of minutes.

Worth pairing with a check that the job still reports something useful when
every Rust step is skipped, so the summary does not read as a clean pass that
never looked at anything.

## 2. The semgrep baseline is unguarded

[...]

Issue #3760, opened 2026-08-29
Two diverged rtk-rewrite.sh hooks both claim rtk-hook-version 3


KuSh:
## Problem

There are two implementations of the Claude Code rewrite hook in this repo, and
they have diverged while both declaring `# rtk-hook-version: 3`.

| | `hooks/claude/rtk-rewrite.sh` (authoritative) | `.claude/hooks/rtk-rewrite.sh` (what this repo dogfoods) |
|---|---|---|
| sha256 | `742418d7…` | `100bfefe…` |
| missing `jq`/`rtk` | warns on stderr | silent exit |
| `rtk >= 0.23.0` guard | yes, cached | **absent** |
| heredoc skip | — | yes |
| `set -euo pipefail` | — | yes |
| audit logging (`RTK_HOOK_AUDIT=1`) | — | yes |
| `jq` spawns per call | 5 | 6 |

`src/hooks/hook_check.rs:53` treats anything with `rtk-hook-version >= 3` as
`HookStatus::Ok`, so RTK's own staleness check cannot distinguish them. Both
report healthy while running materially different logic.

## The copy is stale, not ahead

The authoritative file's two most recent commits are `fix(hooks): reduce jq
invocations in rtk-rewrite (#968)` and `fix(hook): reduce rewrite hook spawn
overhead`. `.claude/hooks/rtk-rewrite.sh` last changed in #893, before both,
and still uses the pre-#968 shape:

```bash
CMD=$(echo "$INPUT" | jq -r '.tool_input.command // empty')
ORIGINAL_INPUT=$(echo "$INPUT" | jq -c '.tool_input')
UPDATED_INPUT=$(echo "$ORIGINAL_INPUT" | jq --arg cmd "$REWRITTEN" '.command = $cmd')
```

This hook runs on every Bash tool call, so the extra process spawns work
directly against the <10ms / Zero Overhead goal — and the version most likely to
be exercised while developing RTK is the slow one.

The three features unique to the copy were never in the authoritative file
(`git log -S` across all refs finds no commit adding or removing them there), so
this is not a case of something being deliberately dropped.

## Proposed fix

1. Port into `hooks/claude/rtk-rewrite.sh`, keeping its jq-spawn reductions,
   version guard and dependency warnings:
   - `set -euo pipefail`
   - the heredoc early-skip
   - opt-in audit logging
2. Bump `# rtk-hook-version:` to `4` and `CURRENT_HOOK_VERSION` in
[...]

Issue #3759, opened 2026-08-29
Stable releases are 'Latest' for ~6-7 minutes before their assets exist, breaking install.sh


KuSh:
## Problem

Every stable release opens a window of roughly 6–7 minutes during which
`https://github.com/rtk-ai/rtk/releases/latest` resolves to a **published
release that has no assets yet**. Any `curl -fsSL … | sh` install started in
that window fails.

`install.sh` resolves the version from exactly that redirect and then downloads
by name:

```bash
# install.sh:53
VERSION=$(curl -sI "https://github.com/${REPO}/releases/latest" ... )
# install.sh:92
DOWNLOAD_URL="https://github.com/${REPO}/releases/download/${VERSION}/${BINARY_NAME}-${TARGET}.tar.gz"
```

So during the window the version resolves fine and the tarball 404s.

## Evidence

Release publish time vs. first asset upload, for the last six stable releases:

| Tag | Published | First asset | Window |
|-----|-----------|-------------|--------|
| v0.46.0 | 19:07:43 | 19:14:55 | 7m12s |
| v0.45.0 | 09:41:43 | 09:48:19 | 6m36s |
| v0.44.2 | 16:47:18 | 16:53:01 | 5m43s |
| v0.44.1 | 13:51:32 | 13:59:04 | 7m32s |
| v0.44.0 | 12:39:04 | 12:45:55 | 6m51s |
| v0.43.0 | 08:55:20 | 09:02:07 | 6m47s |

Reproduce:

```bash
gh release view v0.46.0 --json publishedAt,assets \
  --jq '"\(.publishedAt) vs \(.assets | map(.createdAt) | sort | .[0])"'
```

## Root cause

`release-please` creates the GitHub Release as **published** (not a draft) the
moment the release PR merges. The multi-platform build in `release.yml` only
starts afterwards and takes ~6–7 minutes to upload assets. GitHub points
`/releases/latest` at the newest published, non-prerelease release, so it
switches to the new tag immediately.

## Proposed fix

Have release-please create a draft and promote it only once assets are
verified:

1. Add `"draft": true` to `release-please-config.json`.
2. After the `verify-assets` job in `release.yml` passes, publish it:
   `gh release edit "$TAG" --draft=false`.

Drafts are excluded from `/releases/latest`, so the redirect keeps pointing at
the previous complete release until the new one is genuinely installable. The
failure mode also inverts usefully: if the build breaks, the release stays a
draft rather than becoming a broken "Latest".

[...]

Issue #189, opened 2026-02-18
Release pipeline should skip binary builds for non-code changes


pszymkowiak:
## Problem

The release pipeline currently triggers full multi-platform binary builds for **every** release, even when no Rust source code changed. This wastes CI resources and pollutes the release history with identical binaries.

### Example: v0.20.1 (2026-02-17)

PR #161 (`fix: install to ~/.local/bin`) changed only `install.sh` + `README.md` — **zero Rust code changes**. Yet it triggered:

- 5 platform builds (macOS Intel/ARM, Linux Intel/ARM, Windows)
- DEB package build
- RPM package build
- Homebrew formula update
- Version bump in `Cargo.toml` + `Cargo.lock`

The resulting binaries are **functionally identical** to v0.20.0 (only the embedded version string differs).

### Impact

On Feb 17 alone, 4 releases were created (v0.19.0 → v0.21.1). At ~35 CI minutes per release (7 build jobs), that's over 2 hours of CI time. At least one release (v0.20.1) was entirely unnecessary.

## Root Cause

`release-please` determines version bumps based on **conventional commit prefixes** only (`fix:` → patch, `feat:` → minor). It has no awareness of which files changed. So `fix: install to ~/.local/bin` triggers a patch release even though no compiled code was affected.

The `build-release` job in `release-please.yml` runs unconditionally whenever `release_created == 'true'`:

```yaml
build-release:
  needs: release-please
  if: ${{ needs.release-please.outputs.release_created == 'true' }}
  uses: ./.github/workflows/release.yml
```

## Proposed Solution

Add a source-code change detection step between release-please and the build job:

```yaml
check-source-changes:
  name: Check if source code changed
  needs: release-please
  if: ${{ needs.release-please.outputs.release_created == 'true' }}
  runs-on: ubuntu-latest
  outputs:
    code_changed: ${{ steps.check.outputs.code_changed }}
  steps:
    - uses: actions/checkout@v4
      with:
        fetch-depth: 0
    - name: Check for Rust source changes since last release
      id: check
      run: |
        # Get the previous release tag
        CURRENT_TAG="${{ needs.release-please.outputs.tag_name }}"
        PREV_TAG=$(git tag --sort=-v:refname | grep -v "$CURRENT_TAG" | head -1)
        
        # Check if any Rust source files changed
[...]

PR #3763, opened 2026-08-29
ci: skip builds when a change touches no build inputs

KuSh:
## What

Adds path-based change filtering to `.github/`, which had none. Every push to
`develop` ran `pre-release` plus 8 build jobs (5 platforms + DEB + RPM + Create
Release) and published a `dev-*-rc.N` prerelease; every PR ran 10 job instances.
Replaying the last 100 merges into `develop`, 8 were pure markdown and paid for
all of it.

`.github/inert-paths.conf` holds **git pathspecs** for paths that cannot change
the binary or CI. `changed-scope.sh` subtracts them from a diff range and
answers through its exit status. Git does the matching, so there is no glob
engine of ours to get wrong.

## About #189 — read before merging

**This does not implement what #189 proposed, and that is deliberate.** The
issue asked to gate the binary build on the *release* path. Two problems, both
verified:

1. **The proposed diff can never return false.** It checks
   `git diff PREV_TAG..HEAD -- '*.rs' 'Cargo.toml' 'Cargo.lock' 'build.rs'`, but
   release-please's own release commit always bumps the version in `Cargo.toml`
   and `Cargo.lock`. `code_changed` would be `true` on every release.
2. **Skipping the build on `master` breaks installs.** `install.sh` resolves the
   version by following the redirect on `/releases/latest`, then downloads that
   tag's tarballs by name. A stable release published without assets becomes
   "Latest" and every fresh `curl | sh` fails.

The savings #189 was after are taken on the develop and PR side instead, where
they are roughly 8x larger (369 RCs vs ~46 stable tags) and carry no such risk.
The master path is moved to its own workflow and left ungated.

Closing #189 on merge for that reason — if you would rather keep it open to
track the release-path idea, drop the `Fixes` line before merging.

## Design

Every failure direction leads to building rather than to shipping a stale
artifact:

- The list is inverted, so an unrecognised path builds.
- The skip verdict is exit status **3**, not 1, because bash hands out the low
  statuses itself (1 for `set -u`, 2 for syntax, 127 for missing). An accident
  cannot be mistaken for a skip.
- Gated jobs skip only on an explicit `build=false`, and the gate's own guards
  run *before* it writes that, so a red gate runs CI rather than waving a PR
  through green.
- `--no-renames` throughout: rename detection reports only a rename's
  destination, so moving an embedded file into `docs/` would otherwise look
  inert on a commit that no longer compiles.

### The `,glob` trap

`hooks/**` is deliberately **not** excluded. `src/hooks/init.rs` `include_str!`s
ten files out of `hooks/`, six of them `.md`. A bare `:(exclude)*.md` would
swallow them, because without `,glob` a pathspec `*` also matches `/` — and a
docs-only commit would then ship a stale binary.

`check-embed-scope.sh` enforces this mechanically: it resolves every
`include_str!`/`include_bytes!` target under all tracked `*.rs` (collapsing
newlines first, since rustfmt wraps long paths, then splitting each macro onto
its own line so `grep -o`'s non-overlapping windows cannot let one embed hide
the next) and fails if any is excluded, missing, unresolvable, or written in a
form it cannot verify. It runs ungated, so it still fires on the docs-only PRs
that skip everything else.

### Other changes

- **`stable-release.yml`** — the master path moves out of `cd.yml` so a bug in
  the develop gate cannot reach a release. It is not path-gated.
- **`verify-assets`** in `release.yml` — asserts all 8 assets uploaded non-empty
  and that `checksums.txt` names nothing that failed to upload. On failure it
  demotes the release to prerelease, taking it out of `/releases/latest` so
  installs fall back to the last good tag. Discord and the Homebrew tap wait on
  it. It detects and demotes; it cannot prevent — see #3759.
- **`.claude/hooks/`** is excluded from the build gate but routed to the
  security job via `.github/security-paths.conf`. Those shell scripts run on
  contributors' machines and semgrep is Rust-only; that job's pattern scan was
  Rust-only too, so it gains a `Shell script scan` step.
- **`cd.yml`** diffs from the last RC rather than `github.event.before`: develop
  runs are `cancel-in-progress`, and a docs push landing on a cancelled source
  build would otherwise skip and strand that source change with no RC.
- **Security scan diff range.** Three existing steps diffed
  `origin/master...HEAD`, which on a PR into `develop` spans every merge since
  the last release rather than the PR. On this PR that was 88 files against 10,
  and it reported `Cargo.toml`, `tracking.rs` and `init.rs` as "critical files
  modified" by a PR touching none of them. All four scans now share one
  `BASE_SHA` resolved once, so they report the PR's own changes. Resolving it
  once also lets each scan separate its `git` call from the `grep` that needs
  `|| true` for a no-match — a failing `git` now fails the step instead of
  passing as a clean verdict.
- **`pr-target-check.yml`** gains `branches: [master]`, matching what its job
  already tested; every PR into `develop` was starting a run that only skipped
  (all 20 most recent runs did nothing), and it holds a `pull-requests:write`
  app token on `pull_request_target`. It also stops flagging release-please:
  its only exemption was the `develop` → `master` promotion, so release-please's
  own version-bump PR has been labelled `wrong-base` and told to target
  `develop` on every release since 0.37.1. Both exemptions are keyed by branch
  (the account behind them is not stable — the promotion has been opened by a
  human and by two different apps) and require the branch to live in this repo,
  since a fork's head branch name is the contributor's to choose.

## Verification

- `changed-scope.sh --self-test` — 40 assertions, wired into CI.
- `check-embed-scope.sh` — 49 embed sites, verified non-vacuous by injecting a
  `docs/` embed (flagged DRIFT) and by removing `,glob` (flagged the six
  embedded `hooks/*.md`).
- **Historical replay over the last 100 merges into `develop`:** 92 build /
  8 skip, and every file in all 8 skipped merges is markdown — no `.rs`, no
  `hooks/`, no `Cargo.*`.
- `verify-assets` logic run against the real `v0.46.0` release: clean on the
  good release, catches a dropped tarball in both checks, and aborts without
  demoting when `gh` is unavailable.
- **`pr-target-check.yml` replayed over the last 100 PRs into `master`:** 17
  verdicts flip, all of them release-please's, 0 unintended exemptions, 65
  contributor PRs still flagged. The same-repo clause changes no verdict.
- Security scans replayed against the real `refs/pull/3763/merge`: the summary
  now lists exactly the files this PR touches, and with `BASE_SHA` unset every
  scan fails hard rather than reporting clean.

## Known limitations

- The `Security Scan` job still runs the full Rust toolchain for a shell-only
  change (~5 min for a shell grep). Fix is step-level conditions; deferred
  because it needs testing on two real PR shapes. Tracked in #3761, closed
  until this merges.
- Stable releases are "Latest" for ~6-7 minutes before their assets exist. That
  is pre-existing and unrelated to this gate — #3759.

## Follow-ups opened while working on this

- #3759 — stable releases are "Latest" before their assets exist
- #3760 — two diverged `rtk-rewrite.sh` hooks both claim `rtk-hook-version: 3`


amandeavor (approved):
Reviewed the path classifier, embedded-file guard, CI dependency graph, prerelease range selection, and stable-release asset verification. The classifier passes all 40 self-test cases, the PR checks pass across Linux, Windows, and macOS, and the gate consistently fails open when it cannot prove a change is inert. I also checked the release demotion and downstream notification/Homebrew dependencies. I did not find a correctness blocker.

pszymkowiak:
Solid PR overall — the design write-up and the `,glob` trap section in particular are great, and the self-test/replay verification approach is exactly right.

One correctness bug worth fixing before merge, on the `notify-discord`/`homebrew` gate:

```yaml
notify-discord:
  needs: [verify-assets]
  if: ${{ !inputs.prerelease }}
```

`needs:` changed from `[release]` to `[verify-assets]`, but `if:` didn't change with it. GitHub Actions only applies an implicit `success()` check on `needs` when a job has **no** custom `if:` — the moment you add your own condition, that implicit gate is gone and you have to write the success check yourself:

> "The `success()` function ... this status check function will cause a job to be skipped unless a previous job's status is success. It is recommended to use this expression whenever you use a custom condition alongside `needs`."
> — [GitHub Docs: Using conditions to control job execution](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution)

So as written, both jobs run as soon as `verify-assets` *completes* — success or failure — as long as `!inputs.prerelease` holds. Since `verify-assets` demotes a broken release via `gh release edit --prerelease` and then exits 1 (rather than flipping `inputs.prerelease`, which is just the original workflow input), the demotion doesn't stop either downstream job: Discord announces the release and Homebrew's formula gets updated pointing at assets that were just flagged incomplete — the exact scenario "Discord and the Homebrew tap wait on it" in the PR description says this prevents.

Fix:
```yaml
if: ${{ !inputs.prerelease && needs.verify-assets.result == 'success' }}
```
on both `notify-discord` and `homebrew`.

(Rest of the review — duplicated merge-ref guard, hardcoded asset list in `verify-assets`, inconsistent diff ranges between the two security-job steps, `changed-scope.sh` reimplementing what `dorny/paths-filter` already does — are cleanup/simplification notes, not blockers.)

KuSh:
Thanks — two of the cleanup notes turned out to be worth more than the blocker, and both are now fixed in the PR. On the blocker itself I think the rule is the other way round, and I'd rather not add the condition.

## The implicit `success()` is not removed by a custom `if:`

The quote is real, but it says the opposite of the reading. From [Expressions → status check functions](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions):

> A default status check of `success()` is applied unless you include one of these functions.

"These functions" is the status-check set — `success()`, `always()`, `cancelled()`, `failure()`. What drops the default is *including one of those*, not writing a custom condition. [`jobs.<job_id>.needs`](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs) says the same from the other side:

> If a job fails or is skipped, all jobs that need it are skipped unless the jobs use a conditional expression that causes the job to continue.

and its worked example of such an expression is `always()`.

`if: ${{ !inputs.prerelease }}` contains no status check function, so `success()` still applies and both jobs skip when `verify-assets` fails.

The strongest evidence is inside the PR. `verify-assets` is the job that *does* have to run on failure, and the only reason it can is the `!cancelled()`:

```yaml
verify-assets:
  needs: [release]
  if: ${{ !cancelled() && !inputs.prerelease }}
```

If a custom `if:` already removed the implicit `success()`, that `!cancelled()` would be dead weight — `!inputs.prerelease` alone would run it on a failed `release`. It isn't dead weight: without it the job is skipped exactly when a release is incomplete, which was a HIGH finding in an earlier round here. The two facts are one rule seen from both ends — `!cancelled()` on `verify-assets` is load-bearing *because* the plain `if:` downstream is sufficient.

So `needs.verify-assets.result == 'success'` would be a no-op: both checks would have to pass and the implicit one already does the work. I'd rather leave it out, because a redundant success check sitting beside a deliberate `!cancelled()` reads as if the implicit gate were unreliable — which is what would invite someone to "fix" `verify-assets` the same way and quietly break the demotion path.

Happy to be shown wrong by a run: if you can point at one where a job with `needs` and a status-function-free `if:` ran after its dependency failed, I'll take the change.

## Diff ranges — you were right, and it was worse than inconsistent

Fixed in the PR. Measured on this PR's own merge ref: `origin/master...HEAD` resolved to **88 files** against the PR's **10**, because on a PR into `develop` it spans every merge since the last release. The critical-files scan was reporting

```
.github/workflows/stale.yml
Cargo.toml
src/core/tracking.rs
src/hooks/init.rs
```

as modified by this PR, which touches none of them, and then demanding "Manual security review by 2 maintainers". Misattribution rather than noise — and a scan that cries wolf every time is one people learn to click past.

All four scans now share one `BASE_SHA` resolved once. It is also target-aware, which the old range was not: on a PR into `develop` it is the PR alone, and on a PR into `master` it is the whole release delta. (On the last release PR the old range even pulled in the *previous* release's `CHANGELOG.md` and `.release-please-manifest.json`, since `merge-base(master, develop)` predates release-please's own commits on master.)

That surfaced a second thing: `grep` returns 1 on no-match, so each step needed `|| true`, which also swallowed a failing `git diff` and turned a broken scan into "No dangerous patterns detected". Each scan now separates its `git` call from the `grep`, so git failure fails the step. Verified by running them with `BASE_SHA` unset — all four exit 128 instead of reporting clean.

## Duplicated merge-ref guard — your note is what made it worth doing

I had this as not worth factoring: one line of plumbing across two jobs, with opposite responses. Fixing the diff ranges changed the arithmetic — four steps needed the same base — so it now lives once, in a `Resolve the PR diff base` step placed before the toolchain install so a broken assumption costs seconds rather than minutes. The shell scan lost its private copy.

## Hardcoded asset list — deliberate, and now commented

Deriving `want.txt` from the build matrix would drop a vanished matrix entry from *both* sides of the `comm`, so the check would agree nothing was missing. The independence is the point. Adding a target only grows `have`, which is ignored, so it cannot cause a false demotion.

## `dorny/paths-filter`

Considered and rejected in the design, for a reason specific to this use: a file matching no filter yields *all filters false*, so an unclassified new path reads as "nothing to build" and skips. The fail-safe direction is inverted from what this gate needs. `changed-scope.sh` subtracts an inert list instead, so anything unlisted builds — which is also why the classification lives in one file both workflows read rather than in either workflow.

PR #3772, opened 2026-08-30
fix(core): stdout-only filters silently swallow a tool's stderr

KuSh:
Every `RunOptions::stdout_only()` filter drops the tool's stderr. When a tool reports on stderr and leaves stdout empty, `rtk` prints nothing at all — on either stream.

`src/cmds/README.md` documents the constructor as *"stdout-only to filter, **stderr passthrough**, no tee"*, but the passthrough half was never implemented: `run_captured_filter` only emits stderr on the `skip_filter_on_failure` path, which most callers do not enable.

## Impact

Around 30 call sites: `ruff`, `pytest`, `rspec`, `rubocop`, `prettier`, `phpunit`, `phpstan`, `gh`, `glab`, `go test`, `golangci-lint`, `tree`, `wc`, `psql`.

Partially addresses #2878, which documents the prettier case in detail. That issue's first defect — stderr discarded before the filter ever sees it — is fixed here, so the `[warn]` lines now reach the user. Its second defect (`filter_prettier_output` excluding `[warn]` lines) is prettier-specific and is **not** touched by this PR, so #2878 should stay open.

## How it broke

Two commits compound:

- `078ad1d` moved `golangci_cmd` onto `runner::run_filtered` and deleted a block whose own comment read `// Always forward stderr (config errors, missing linters, etc.)`.
- `861a46d` added `guard::never_worse`. Under `stdout_only` the baseline is stdout, so with stdout empty any filter output counts as "worse than nothing" and collapses to the empty string.

Together: silence on both streams.

## The fix

Forward the captured stderr, and count it. The guard keeps comparing against **stdout**, which is the correct baseline: stderr is forwarded verbatim on both sides, so it cancels, and never-worse reduces to `filtered <= stdout`. Widening the baseline to the combined output would let rtk emit more than the command it replaces.

Tracking now measures against the full raw output and counts what was actually emitted — comparing stdout to stdout while printing stderr booked a passed-through stream as if the filter had removed it, overstating savings in `rtk gain`.

## Tests

Six integration tests over a fake tool on PATH, pinning both halves of the contract: the tool's stderr must reach the user, and rtk must not invent a stdout message in its place. The prettier case covers the second — it writes its report to stderr and nothing to stdout *even on a successful run*, so a filter's empty-input placeholder would otherwise be printed over a run that worked.

`cargo test --all`: 2927 passing. fmt clean, 0 clippy warnings.

### DIFF
--- a/src/core/runner.rs
+++ b/src/core/runner.rs
@@ -2,6 +2,7 @@
 
 use anyhow::{Context, Result};
 use regex::Regex;
+use std::borrow::Cow;
 use std::process::Command;
 use std::sync::LazyLock;
 
@@ -138,6 +139,8 @@ where
         raw
     };
 
+    // stderr is forwarded unchanged below, so it costs the same whether or not rtk is
+    // in the way. What the guard must not inflate is the stdout it replaces.
     let shown = if let Some(label) = opts.tee_label {
         print_with_hint(&filtered, raw, raw_for_tracking, label, exit_code)
     } else {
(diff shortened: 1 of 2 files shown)


aeppling:
Hey @KuSh 

Looks sound, stderr should be forwarded. For most tools it's a few lines of pure signal, so passing it through uncompressed is fine for now

One case to flag: prettier writes its whole report to stderr, so rtk prettier --check . on a large repo goes from 1 token to ~26k. Not a blocker for this PR

aeppling (approved):

KuSh:
One thing worth flagging, though not a blocker: `filter_prettier_output` only runs on stdout, and prettier's `[warn] <file>` report lives entirely on stderr. Once this lands, `rtk prettier --check .` will forward that report raw and unfiltered — on a large repo with many unformatted files that's roughly 1 token today going to ~26k. This is the same gap #2878 already tracks (extending the filter to handle stderr's `[warn]` lines); this PR correctly leaves that half for #2878, just calling out that the failure mode changes from silent to verbose in the interim.

PR #3773, opened 2026-08-30
fix(ci): unbreak the golangci-lint benchmark row

KuSh:
The `benchmark` job has been the only failing CI job since 2026-08-20. Nothing in the repo changed — GitHub runners moved from go1.26.6 to go1.27.0 in a 4h42m window (last green run `32325235750`, first red `32343365031`).

## Cause

`go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest` resolves on the **unversioned** module path, which stops at v1.64.8 (March 2025, the last v1). v2 is published under `/v2`, so `@latest` never moved.

Under Go 1.27 that v1.64.8 fails before it lints anything — its vendored `go/types` rejects the newer export data:

```
could not load export data: internal error in importing "internal/goarch"
(export data version 4 is greater than maximum supported version 2)
```

It writes 628 bytes to stderr, leaves stdout empty and exits 3 — the 157-token row that reddens the job. Reproduced byte-for-byte in `golang:1.27`.

## Why these four commits ship together

Switching to the `/v2` path alone does **not** fix CI. On the old fixture a working golangci-lint prints `0 issues.` (3 tokens) while rtk prints `golangci-lint: No issues found` (8), which scores `NEG` and still exits 1. Verified in Docker both symmetrically and asymmetrically.

The fixture has to have findings for the row to measure anything. It never did — it read `0 -> 8 SKIP` for as long as it was green, so this row has never actually exercised the filter.

Measured in `golang:1.27` against v2.13.2, on the fixture exactly as it appears here:

```
golangci-lint   220 -> 35   GOOD (84%)
go test          30 ->  8   GOOD (73%)   unchanged
go build          0 ->  0   SKIP         unchanged
go vet            0 ->  0   SKIP         unchanged
```

`go build`, `go vet` and `go test` stay clean, so the neighbouring rows keep measuring what they measured before.

## Also here

- The synthetic `golangci_v2_json.txt` fixture is pretty-printed across 144 lines; golangci-lint emits its JSON on a single line. Its savings assertion counted whitespace-separated words, so indentation was doing the work — against real single-line output the same helper reads 40 "words" for 4 KB and reports 30% for a filter that is doing its job. Replaced with output captured from v2.13.2 and measured with `estimate_tokens`, which puts the real figure at 95.3%.
- `parse_major_version` required a bare integer, so a version reported as `v2.13.2` fell through to the v1 fallback and rtk would send v2 `--out-format=json`, a flag v2 removed. The prefix is not stable across builds: a `go install` of v1.64.8 reports `has version v1.64.8` while the same build of v2.13.2 reports `has version 2.13.2`. Today's install takes the unprefixed form and detects correctly; this closes the case that does not.

`cargo test --all`: 2925 passing. fmt clean, 0 clippy warnings.


aeppling:
Hey @KuSh , ci green and implementation is correct, thanks :)

aeppling (approved):

Issue #3226, opened 2026-07-26
Windows release binaries are unsigned, so Smart App Control blocks every new version for its first ~week


mIBRAHIM707:
## Summary

`rtk-x86_64-pc-windows-msvc.zip` ships an unsigned `rtk.exe`. On Windows 11 with **Smart App Control** enabled, a freshly released build is refused outright:

```
Program 'rtk.exe' failed to run: An error occurred trying to start process '...\rtk.exe'.
An Application Control policy has blocked this file.
```

Because SAC trusts binaries by *reputation* rather than signature, this recurs on **every release** — each new version is a new hash with zero reputation, so Windows users with SAC on are locked out of every update for roughly its first week.

## Evidence

Checked both the working installed version and the new one, on the same machine:

| Version | `Get-AuthenticodeSignature` | Runs under SAC |
|---|---|---|
| 0.43.0 (installed, working) | `NotSigned` | yes |
| 0.44.0 (downloaded 5h after release) | `NotSigned` | blocked |

Checksum verified against `checksums.txt` before testing — `f5824278f7f668a8638a6b0ec6286fd31190165b1f2bc7b031d53cb2dfa6ead5`. The download is intact; it simply isn't trusted.

0.43.0 being equally unsigned yet running is the point: it works only because it has accrued enough Intelligent Security Graph reputation over time, not because it is trusted by identity.

## Why users can't work around it

- **`Unblock-File` does not help.** This is not the mark-of-the-web zone flag — clearing it changes nothing, because SAC evaluates reputation independently.
- **Smart App Control cannot be turned off and back on.** Disabling it is a one-way switch; re-enabling requires reinstalling Windows. Asking users to disable it is not a reasonable workaround for installing a CLI tool.
- The only remaining options are "wait an unknown number of days and retry" or "stay on the old version" — neither of which a user can plan around, since Microsoft publishes no timeline for reputation accrual.

## Impact

This is worse than a slow update path, because of what rtk does. `0.44.0` contains `python: stop reporting a failed tool run as clean` — a filter reporting a failed run as clean is precisely the bug class users most need fixed promptly. SAC guarantees the affected users are the last to receive it.

Note this also affects anyone installing rtk for the *first* time on a SAC-enabled machine shortly after a release: there is no older, reputable build to fall back to.

## Request

[...]

Issue #3214, opened 2026-07-26
Security report: Defender detection on release asset rtk-x86_64-pc-windows-msvc.zip (dev-0.44.0-rc.331)


shauncroton:
Sorry id this turns out to be a false positive but........

A Microsoft Defender detection was reported for the release asset rtk-x86_64-pc-windows-msvc.zip on the release page: https://github.com/rtk-ai/rtk/releases/tag/dev-0.44.0-rc.331

Detection details reported by a user:
- Detection name: Trojan:Script/Wacatac.H!ml
- Asset: rtk-x86_64-pc-windows-msvc.zip

Recommended immediate actions for maintainers:
1. Temporarily remove or unpublish the release asset to prevent further downloads until the file is investigated.
2. If you have the artifact locally or in CI, compute and publish SHA256 (and other) checksums for the ZIP so users can verify whether their copy matches yours. Example commands:
   - PowerShell: Get-FileHash .\rtk-x86_64-pc-windows-msvc.zip -Algorithm SHA256
   - Linux/macOS: sha256sum rtk-x86_64-pc-windows-msvc.zip
3. Rebuild the Windows artifact from the exact commit/tag used to create this release and compare the hash to the distributed asset. If hashes differ, treat the release as compromised.
4. Scan your rebuilt artifact with multiple AV engines (VirusTotal) and compare results.
5. Inspect the ZIP contents for suspicious scripts or unsigned installers (.ps1, .js, .vbs, .hta, installer stubs) and for any unexpected files.
6. If you confirm a false positive, publish the verified hashes and an explanation in the release notes and re-upload the asset (or provide a replacement signed build). If you confirm tampering, delete the compromised asset, investigate CI/secret exposure, rotate keys, and create a new release from a safe artifact.
7. Consider submitting the file/hash to Microsoft Security Intelligence for analysis: https://www.microsoft.com/en-us/wdsi/filesubmission

Information requested from maintainers to help users and security triage:
- Please confirm whether you see the same detection internally.
- If possible, provide the SHA256 (or other) checksum of the published artifact and whether a rebuilt artifact matches that checksum.
- Any CI/build logs or provenance that show how the artifact was produced (builder image, timestamps, signings).

Notes:
- Affected release: dev-0.44.0-rc.331 (prerelease) — https://github.com/rtk-ai/rtk/releases/tag/dev-0.44.0-rc.331
- Reported by: user shauncroton (via issue report to maintainers)

[...]

PR #3789, opened 2026-08-31
feat(release): sign Windows binaries via SignPath

pszymkowiak:
## Problem

Windows release binaries are unsigned:

- **#3226** — Smart App Control blocks every new release for ~a week (reputation-based, each release is a fresh hash)
- **#2989** — enterprise allowlisting can't use publisher-based rules
- Recurring Defender false positives: #3124, #3214, #2694, #2070, #1749, #1782

## Change

New `sign-windows` job in `release.yml`, between `build` and `release`:

1. The windows build leg uploads the bare `rtk.exe` as artifact `rtk-windows-exe-unsigned`
2. `sign-windows` submits it to SignPath (`signpath/github-action-submit-signing-request@v1.2`, `wait-for-completion: true`)
3. The signed exe is re-zipped and **overwrites** the `rtk-x86_64-pc-windows-msvc` artifact — the release job and `checksums.txt` pick up the signed binary with zero changes downstream

Failure semantics:
- `SIGNPATH_ORGANIZATION_ID` repo variable absent → job **skipped**, release ships unsigned (previous behaviour; forks unaffected)
- Signing attempted and failed → release job does **not** run (never silently ship unsigned when signing is expected)

Docs added to `.github/workflows/CICD.md`.

## Required setup (not in this PR)

| Where | Name | Kind |
|-------|------|------|
| GitHub repo | `SIGNPATH_API_TOKEN` | Actions secret |
| GitHub repo | `SIGNPATH_ORGANIZATION_ID` | Actions variable |
| SignPath | project `rtk`, signing policy `release-signing` | |
| SignPath | artifact configuration: single `pe-file` `rtk.exe` | |
| SignPath | GitHub Actions trusted build system linked to the project | |

For SAC immediate trust an EV-grade cert is ideal; SignPath's OSS program certificate still gives a stable publisher identity for allowlisting and greatly reduces Defender FPs.

Refs #3226, #2989

https://claude.ai/code/session_01DyGoVR59QcBsJ8WfnYwCe7

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,9 +32,6 @@ jobs:
   build:
     name: Build ${{ matrix.target }}
     runs-on: ${{ matrix.os }}
-    outputs:
-      # Set only by the windows-msvc leg (see "Expose Windows exe artifact id").
-      windows-exe-artifact-id: ${{ steps.win-artifact.outputs.id }}
     strategy:
       fail-fast: false
       matrix:
@@ -113,17 +110,10 @@ jobs:
       # by the sign-windows job below.
       - name: Upload unsigned Windows exe (for signing)
         if: matrix.os == 'windows-latest'
-        id: upload-exe
         uses: actions/upload-artifact@v4
         with:
           name: rtk-windows-exe-unsigned
           path: target/${{ matrix.target }}/release/rtk.exe
-
-      - name: Expose Windows exe artifact id
-        if: matrix.os == 'windows-latest'
-        id: win-artifact
-        shell: bash
-        run: echo "id=${{ steps.upload-exe.outputs.artifact-id }}" >> "$GITHUB_OUTPUT"
 
   # Signs rtk.exe via SignPath, then replaces the unsigned windows zip with a
   # signed one under the same artifact name so the release job needs no change.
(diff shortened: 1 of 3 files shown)


aeppling:
Hey @pszymkowiak 

### Blocking: windows-exe-artifact-id can arrive empty

Set from a 5-leg matrix; the four non-Windows legs evaluate it to ''. Per GitHub's docs:

"Actions does not guarantee the order that matrix jobs will run in. Ensure that the output name is unique, otherwise the last matrix job that runs will override the output value."

Empty → getInput('github-artifact-id', {required: true}) throws → sign-windows fails → release blocked. Windows is usually the slowest leg, so it'll mostly work. Intermittent, not always-broken.

### Also

Prefer using a sha instead of the version tag which can move.
Pin a tag always better for prod.

wait-for-completion defaults to 600s could times out and blocks the release if release-signing has approvers.

.github/workflows/release.yml
```diff
@@ -105,6 +105,77 @@ jobs:
...
+          if [ -z "$artifact_id" ]; then
+            echo "::error::rtk-windows-exe-unsigned artifact not found for run ${{ github.run_id }}"
+            exit 1
+          fi
+          echo "id=$artifact_id" >> "$GITHUB_OUTPUT"
+
+      - name: Submit signing request
+        # Pinned by commit SHA (not the mutable v1.2 tag) per SignPath repo tag
+        # refs/tags/v1.2 = 4f13d373e8f0cd8d3c0465ff4877feff27aed2ae.
+        uses: signpath/github-action-submit-signing-request@4f13d373e8f0cd8d3c0465ff4877feff27aed2ae # v1.2
```
aeppling, line 153: Latest version is https://github.com/[SignPath/github-action-submit-signing-request](https://github.com/SignPath/github-action-submit-signing-request)

This would be quickly deprecated because current 1.2 run node 20, current is 2.X with Node 24

.github/workflows/release.yml
```diff
@@ -105,6 +105,77 @@ jobs:
...
+  # SIGNPATH_ORGANIZATION_ID repository variable is not configured.
+  sign-windows:
+    name: Sign Windows binary (SignPath)
+    needs: build
+    if: vars.SIGNPATH_ORGANIZATION_ID != ''
+    runs-on: ubuntu-latest
+    # Least-privilege: this job only reads the artifact list and talks to the
+    # SignPath API — it never checks out or writes repo contents.
+    permissions:
+      actions: read
```
aeppling, line 130: Missing contents: read

aeppling (changes requested):

PR #3799, opened 2026-09-01
fix: honor configured grep maximum results

### LINKED ISSUE #3229: [config] limits.grep_max_results is declared and documented but never read; the --max clap default of 200 is an unconfigurable ceiling
## Summary

`limits.grep_max_results` is declared in `LimitsConfig`, documented in two READMEs, and serialized into every generated `config.toml` — but **no code ever reads it**. `rtk grep`'s total-result cap comes exclusively from the `--max` clap flag, whose `default_value = "200"` therefore acts as a hard ceiling that no configuration can move.

This is the same defect class as #2437 (`status_max_files` / `status_max_untracked` declared but never read), for a different field. #2437 enumerates the orphaned fields it covers and `grep_max_results` is not among them, so this is not a duplicate.

Its sibling `grep_max_per_file` **is** wired up correctly, which is what makes the resulting behavior actively misleading rather than merely inert — see "The perverse interaction" below.

## Affected version

Reproduced on `rtk 0.43.0` (Homebrew, macOS arm64, Darwin 25.5.0). Source inspected at `develop` `1d6798d` (Cargo.toml `0.42.4`).

## Evidence

`grep_max_results` — declared, defaulted, documented, never read:

```
$ rg -n "grep_max_results" src/
src/core/config.rs:125:    pub grep_max_results: usize,
src/core/config.rs:139:            grep_max_results: 200,
src/core/README.md:88:grep_max_results = 200
src/cmds/system/README.md:8:  ... reading `core/config` for `limits.grep_max_results` and `limits.grep_max_per_file`.
```

`grep_max_per_file` — declared *and consumed*:

```
$ rg -n "config::limits\(\)" src/cmds/system/search.rs
633:    let per_file = config::limits().grep_max_per_file;
1337:    let per_file = config::limits().grep_max_per_file;
```

Those two are the only `config::limits()` call sites in `search.rs`. The value actually enforced at `search.rs:642`/`:652` (`if shown >= max_results`) is a plain function parameter threaded down from the CLI arg defined at `src/main.rs:318`:

```rust
/// Max results to show
#[arg(short, long, default_value = "200")]
max: usize,
```

Note that `src/cmds/system/README.md:8` explicitly claims `search.rs` reads **both** fields, so the repo's own documentation is currently wrong on this point.

## Reproduction

```sh
D=$(mktemp -d); mkdir -p "$D/solo"
for i in $(seq 1 300); do printf 'line %d SOLOHIT x\n' "$i"; done > "$D/solo/one.txt"

cfg=~/Library/Application\ Support/rtk/config.toml   # or ~/.config/rtk/config.toml
# set: grep_max_results = 100000, grep_max_per_file = 100000

rtk grep -n SOLOHIT "$D/solo/one.txt" | grep -c SOLOHIT   # config path
rtk grep --max=300 -n SOLOHIT "$D/solo/one.txt" | grep -c SOLOHIT   # flag path
```

Sweeping `grep_max_results` with `grep_max_per_file` pinned at `100000` (ground truth: 300 matches):

| `grep_max_results` | match lines shown |
|---|---|
| 50 | 200 |
| 150 | 200 |
| 199 | 200 |
| 200 | 200 |
| 201 | 200 |
| 500 | 200 |
| 100000 | 200 |

The value has no effect in *either* direction — it cannot tighten the cap or loosen it. The equivalent flag works correctly: `--max=300` → 300 lines shown, confirming the cap itself is functional and only the config plumbing is missing.

For contrast, `grep_max_per_file` behaves exactly as documented on a single file of 80 matches: `10 → 10`, `25 → 25`, `50 → 50`, `100 → 80` (all).

## The perverse interaction

Because the per-file cap is honored while the global cap is frozen at 200, raising `grep_max_per_file` to get more faithful output **reduces how many files you see**. On a fixture of 12 files × 30 matches (360 total):

| `grep_max_per_file` | match lines | distinct files shown |
|---|---|---|
| 10 | 120 | **12 of 12** |
| 25 | 200 | **8 of 12** |
| 100 | 200 | **7 of 12** |

The 200-line budget is consumed by the first few files, and the remainder collapse into a single trailing `+5 more files [see remaining: ...]` line. A user who reads the docs, concludes their sweeps are being truncated, and raises both limits will make coverage strictly *worse* while believing they fixed it. That is how I found this.

## Expected vs actual

| | |
|---|---|
| **Expected** | `limits.grep_max_results` sets the default total cap for `rtk grep`/`rtk rg`; `--max` overrides it per invocation. |
| **Actual** | The field is inert. The clap default of 200 is an unconfigurable ceiling, while the documentation states the opposite. |

## Why this bites AI-agent automation specifically

`rtk grep` is the workhorse for agent verification sweeps — "does any file still contain X?", "did I convert every call site?". The header count (`360 matches in 12 files:`) stays honest, but the *body* is where an agent reads which files and lines actually matched, and 5 of 12 files can be absent from it.

The failure is quiet in the way that matters: there is no error, exit status is `0`, and the elision markers sit at the very end of a 200-line block that an agent has often already stopped attending to. An operator who notices the caps and tries to raise them has no way to succeed and no feedback that their setting was discarded — the natural remedy (raise both limits) silently makes file coverage worse. The result is a confident, incomplete "all call sites converted" verdict.

This is adjacent to but distinct from #2988 (grep truncation is silent and the tee log doesn't restore it) and #1313 (request for a lossless mode). Those argue about what the *default* behavior should be; this issue is that the escape hatch the config already advertises does not exist.

## Suggested fix

Mirror whatever lands for #2437. Concretely, make the clap arg optional and fall back to config:

```rust
#[arg(short, long)]
max: Option<usize>,
// ...
let max_results = max.unwrap_or_else(|| config::limits().grep_max_results);
```

so precedence reads flag → config → built-in default. Failing that, remove the field and both README references so the knob isn't advertised, and consider warning at config-load time when a user has set a key that is never consumed — the same reasoning as Option B in #2437.

mvanhorn:
## Summary

Change the `rtk grep` display-cap argument to an optional CLI override while preserving `-m` as a native grep/rg argument, then pass that optional value through the existing dispatch in `src/main.rs`; the `rtk rg` dispatch should pass no override instead of a hard-coded cap. In `src/cmds/system/search.rs`, resolve the effective maximum once at the production entry point using CLI override first and `config::limits().grep_max_results` otherwise, so downstream streaming and grouped-output branches retain their existing concrete cap and behavior. Add parsing assertions for omitted and explicit `--max` values and end-to-end coverage using an isolated temporary config to prove both `grep` and `rg` honor the configured default while an explicit `rtk grep --max` still wins.

`limits.grep_max_results` is defined, serialized, and documented with a default of 200, but the search execution path never reads it. `rtk grep` instead receives Clap's unconditional 200 default, while `rtk rg` passes a hard-coded 200 directly to the shared search runner. As a result, changing the configuration cannot tighten or raise the global result cap, although the sibling per-file limit works. The thread confirms the same behavior on current macOS and Windows releases and identifies the shared call path for both commands.

Fixes #3229

## Test plan

  Not verified: this needs a person on the named hardware or environment.
- With an isolated config setting `grep_max_results` below the number of fixture matches and no RTK `--max`, `rtk grep` emits only the configured number of results and its cap/elision output reports that configured limit. - With the same isolated config and ripgrep available, `rtk rg` uses the configured global cap rather than the former hard-coded 200 value. - With a configured cap and an explicit long-form `rtk grep --max N`, the explicit value takes precedence and controls the emitted-result limit.

AI was used for assistance.

### DIFF
--- a/src/cmds/go/golangci_cmd.rs
+++ b/src/cmds/go/golangci_cmd.rs
@@ -139,20 +139,12 @@ fn run_filtered(original_args: &[String], invocation: &RunInvocation, verbose: u
         );
     }
 
-    let exit_code = runner::run_filtered(
+    let exit_code = runner::run_filtered_with_exit(
         cmd,
         "golangci-lint",
         &original_args.join(" "),
-        |stdout| {
-            // v2 outputs JSON on first line + trailing text; v1 outputs just JSON
-            let json_output = if version >= 2 {
-                stdout.lines().next().unwrap_or("")
-            } else {
-                stdout
-            };
-            filter_golangci_json(json_output, version)
-        },
-        crate::core::runner::RunOptions::stdout_only(),
+        |output, exit_code| filter_golangci_run(output, version, exit_code),
+        crate::core::runner::RunOptions::default(),
     )?;
 
     // golangci-lint: exit 0 = clean, exit 1 = lint issues found (not an error),
@@ -259,6 +251,23 @@ fn format_command(base: &str, args: &[String]) -> String {
     } else {
         format!("{} {}", base, args.join(" "))
     }
+}
+
+fn filter_golangci_run(output: &str, version: u32, exit_code: i32) -> String {
+    if exit_code >= 2 {
+        return output.to_string();
+    }
+
+    // v2 outputs JSON on first line + trailing text; v1 outputs just JSON.
+    let json_output = if version >= 2 {
+        output.lines().next().unwrap_or("")
+    } else {
+        output
+    };
+    if exit_code != 0 && serde_json::from_str::<GolangciOutput>(json_output).is_err() {
+        return output.to_string();
+    }
+    filter_golangci_json(json_output, version)
 }
 
 /// Filter golangci-lint JSON output - group by linter and file
@@ -398,6 +407,18 @@ mod tests {
         let result = filter_golangci_json(output, 1);
         assert!(result.contains("golangci-lint"));
         assert!(result.contains("No issues found"));
+    }
+
+    #[test]
+    fn test_filter_golangci_run_preserves_command_errors() {
+        let output = "Error: no go files to analyze\n";
+        assert_eq!(filter_golangci_run(output, 2, 7), output);
+    }
+
+    #[test]
+    fn test_filter_golangci_run_preserves_non_json_lint_failure() {
+        let output = "Error: no go files to analyze\n";
+        assert_eq!(filter_golangci_run(output, 2, 1), output);
     }
 
     #[test]
--- a/src/cmds/system/find_cmd.rs
+++ b/src/cmds/system/find_cmd.rs
@@ -391,6 +391,10 @@ fn build_capped_listing(files: &[String], max_results: usize) -> String {
     listing
 }
 
+fn build_explicit_listing(files: &[String], max_results: usize) -> String {
+    build_capped_listing(&display_ordered(files), max_results)
+}
+
 #[allow(clippy::too_many_arguments)]
 pub fn run(
     pattern: &str,
@@ -511,6 +515,14 @@ fn render(
 
     if files.is_empty() {
         timer.track(track_cmd, "rtk find", raw_output, "");
+        return;
+    }
+
+    if max_explicit {
+        let listing = build_explicit_listing(&files, max_results);
+        let listing = listing.trim_end_matches('\n');
+        let shown = crate::core::runner::emit_guarded(listing, None, listing);
+        timer.track(track_cmd, "rtk find", raw_output, &shown);
         return;
     }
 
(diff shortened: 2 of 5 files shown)


mvanhorn:
Fixed the `benchmark` failure in ee415cc.

Root cause was in the golangci filter, not in the grep max-results change. `filter_golangci_json` returns an empty string whenever `serde_json` can't parse its input, and the filter only ever saw stdout — never the exit code. So any golangci-lint run that errored out, or that reported a lint failure without emitting JSON, produced no output at all, which is what the benchmark reported as "1 filter(s) returned empty output".

The filter now takes the exit code through `run_filtered_with_exit` and passes raw output through in two cases: the command itself errored (exit >= 2), or a non-zero exit carried unparseable output. Clean runs and JSON-carrying lint failures are filtered exactly as before. Two regression tests cover both passthrough paths.

Issue #3757, opened 2026-08-29
Unanchored `match_command` in `liquibase.toml` rewrites any command containing `/liquibase`


KuSh:
## Summary

`src/filters/liquibase.toml:3` is the only one of the 63 shipped filters whose `match_command` is not anchored at `^`:

```toml
match_command = "(?:^|/)liquibase(?:\\s|$)"
```

Because the TOML branch matches against the **whole command string**, any unrecognised leading word followed by a path containing `/liquibase` gets rewritten to `rtk <whole thing>`.

## Reproduction

Verified against a locally built binary (`rtk rewrite` — exit 3 + stdout means "rewritten", exit 1 means "passed through"):

```
timeout 5 /usr/bin/liquibase update    → 3  rtk timeout 5 /usr/bin/liquibase update
nohup /opt/tools/liquibase update      → 3  rtk nohup /opt/tools/liquibase update
liquibase update                       → 3  rtk liquibase update                     ← correct
echo /liquibase update                 → 1                                           ← `echo` is ignored, so it escapes
```

This is **pre-existing and independent of any sudo work** — `timeout` and `nohup` reproduce it on `develop` today. #2810 simply adds `sudo` to the set of leading words that expose it:

```
sudo /usr/bin/liquibase update   → 3  rtk sudo /usr/bin/liquibase update
sudo ./liquibase update          → 3  rtk sudo ./liquibase update
sudo liquibase update            → 1                                     ← correct
```

## Cause

`src/discover/registry.rs:1398`, in the `Classification::Unsupported` (TOML-only) branch:

```rust
let normalized = strip_absolute_path(cmd_part.trim());
…
if crate::core::toml_filter::command_matches_filter(&normalized) {
    return Some(format!("rtk {}{}", cmd_part, redirect_suffix));
}
```

`strip_absolute_path` only peels a *leading* path. When the first word is `timeout` / `nohup` / `sudo`, `normalized` is the untouched full string, and the filter's `(?:^|/)liquibase` alternation happily matches the `/liquibase` in the middle of it. The `is_rtk_reserved_command(base)` check just above (`:1403`) only inspects the first word, so it doesn't help here.

Note that `RULES` cannot misfire the same way: all 86 patterns in `rules.rs` are `^`-anchored. This is specific to TOML filters, and today specific to this one file.

## Not a security issue

Confirmed, so this can be triaged as correctness rather than as a vulnerability:

[...]

PR #3807, opened 2026-09-01
test(filters): assert all builtin toml match_command patterns are anchored

amandeavor:
## Summary

Pins that every built-in TOML filter `match_command` is anchored on **every top-level branch**, so a pattern cannot match a path component or subcommand mid-line. A filter selects on argv[0], but its regex runs against the whole command line, so an unanchored branch makes `timeout 5 /usr/bin/liquibase update` activate the liquibase filter and rewrite the wrapper.

The anchoring rule is now enforced at load time by a shared `is_fully_anchored` predicate applied to **both** loaders (`match_patterns_in` and `TomlFilterRegistry::parse_and_compile`), so it also covers project-local and global filters. Those are user-authored and were previously unguarded — which is where #3757 said the durable fix belonged. Unanchored filters are skipped with a warning rather than silently misfiring.

Refs #3757

## Validation

- `test_builtin_all_filters_match_command_anchored` — checks every built-in filter, reading **raw** patterns from `BUILTIN_TOML` so the load-time gate cannot hide an unanchored built-in from the assertion. Accepts per-branch anchoring (`^pnpm\b|^npm\b`) and inline flags (`(?i)^tool\b`); rejects `^liquibase\b|/liquibase\b` and `(?:^|/)liquibase(?:\s|$)`.
- `test_unanchored_user_filter_is_rejected_by_both_loaders` — pins that a user-authored unanchored filter is dropped by both load paths, keeping `MATCH_SET` and the compiled registry in agreement.
- `test_liquibase_match_command_ignores_path_substring` — extended with the wrapper cases (`timeout`, `nohup`) alongside the existing path-substring assertions and their rationale comment.
- `test_toml_filter_rewrites_bare_command_but_not_wrapped_invocations` — exercises the rewrite path end to end via `jj`, which is covered only by a TOML filter and never by the native `RULES` table. The bare case acts as a positive control so the wrapper assertions cannot pass vacuously.
- `test_path_qualified_liquibase_is_not_rewritten` — pins the current `None` for `/usr/bin/liquibase update`, with a comment recording that this diverges from #3757's requested behavior.

`cargo fmt --all`, `cargo clippy --all-targets` (clean), `cargo test --all` (all passing), verified both on this branch and merged onto the latest `develop`.

## Note on #3757

Changed `Closes` to `Refs`. That issue's third required-coverage bullet asks for `/usr/bin/liquibase update` to still be rewritten; this branch pins the current `None` as a characterization test instead.

That `None` is itself a bug, tracked in #3757 rather than fixed here. `classify_command` correctly reports the command as `Supported` — it is a liquibase invocation and RTK can filter it — but the rewrite path matches `RULES.rewrite_prefixes` against the un-normalized `cmd_part`, finds no prefix, and drops it. So every path-qualified invocation of every supported command silently skips filtering (`/usr/bin/grep`, `/usr/bin/make` behave the same), and `rtk discover` advertises an optimization the hook then declines.

The fix is to normalize `cmd_part` before that match, which is a behavior change beyond this PR's test-and-guard scope. When it lands, `test_path_qualified_liquibase_is_not_rewritten` should be updated to assert the rewrite rather than deleted, so the path-qualified case stays covered. #3757 stays open until then.

### DIFF
--- a/src/core/toml_filter.rs
+++ b/src/core/toml_filter.rs
@@ -2032,19 +2032,64 @@ match_command = "^make\\b"
         );
     }
 
-    /// Verify that every built-in filter's match_command starts with `^`.
-    /// Prevents substring misfires where an unanchored match matches a subcommand or path in the middle of a command line.
+    /// Verify that every built-in filter is anchored across every top-level branch.
     #[test]
     fn test_builtin_all_filters_match_command_anchored() {
+        fn has_top_level_alternation(pattern: &str) -> bool {
+            let mut depth = 0;
+            let mut escaped = false;
+            let mut in_class = false;
+
+            for character in pattern.chars() {
+                if escaped {
+                    escaped = false;
+                    continue;
+                }
+                match character {
+                    '\\' => escaped = true,
+                    '[' if !in_class => in_class = true,
+                    ']' if in_class => in_class = false,
+                    '(' if !in_class => depth += 1,
+                    ')' if !in_class => depth -= 1,
+                    '|' if !in_class && depth == 0 => return true,
+                    _ => {}
+                }
+            }
+
+            false
+        }
+
         let filters = make_filters(BUILTIN_TOML);
         for filter in &filters {
+            let pattern = filter.match_regex.as_str();
             assert!(
-                filter.match_regex.as_str().starts_with('^'),
-                "Filter '{}' has match_command '{}' which does not start with '^'",
+                pattern.starts_with('^') && !has_top_level_alternation(pattern),
+                "Filter '{}' has match_command '{}' with an unanchored top-level branch",
                 filter.name,
-                filter.match_regex.as_str()
+                pattern
             );
         }
+
+        assert!(has_top_level_alternation(r"^liquibase\b|/liquibase\b"));
+        assert!(!has_top_level_alternation(r"^(liquibase\b|/liquibase\b)"));
+    }
+
+    #[test]
+    fn test_builtin_liquibase_filter_does_not_match_wrapped_or_path_commands() {
+        let filters = make_filters(BUILTIN_TOML);
+        let liquibase = filters
+            .iter()
+            .find(|filter| filter.name == "liquibase")
+            .expect("built-in liquibase filter should exist");
+
+        assert!(liquibase.match_regex.is_match("liquibase update"));
+        assert!(!liquibase
+            .match_regex
+            .is_match("timeout 5 /usr/bin/liquibase update"));
+        assert!(!liquibase
+            .match_regex
+            .is_match("nohup /opt/tools/liquibase update"));
+        assert!(!liquibase.match_regex.is_match("/usr/bin/liquibase update"));
     }
 
     /// Verify that every built-in filter has at least one inline test.
--- a/src/core/toml_filter.rs
+++ b/src/core/toml_filter.rs
@@ -2032,6 +2032,21 @@ match_command = "^make\\b"
         );
     }
 
+    /// Verify that every built-in filter's match_command starts with `^`.
+    /// Prevents substring misfires where an unanchored match matches a subcommand or path in the middle of a command line.
+    #[test]
+    fn test_builtin_all_filters_match_command_anchored() {
+        let filters = make_filters(BUILTIN_TOML);
+        for filter in &filters {
+            assert!(
+                filter.match_regex.as_str().starts_with('^'),
+                "Filter '{}' has match_command '{}' which does not start with '^'",
+                filter.name,
+                filter.match_regex.as_str()
+            );
+        }
+    }
+
     /// Verify that every built-in filter has at least one inline test.
     /// Prevents shipping filters with zero test coverage.
     #[test]
--- a/src/core/toml_filter.rs
+++ b/src/core/toml_filter.rs
@@ -235,6 +235,15 @@ impl TomlFilterRegistry {
 
         let mut compiled = Vec::new();
         for (name, def) in file.filters {
+            if !is_fully_anchored(&def.match_command) {
+                eprintln!(
+                    "[rtk] warning: filter '{}' in {}: match_command '{}' has a top-level branch \
+                     that does not start with '^'; it would match a path component mid-command. \
+                     Filter ignored.",
+                    name, source, def.match_command
+                );
+                continue;
+            }
             match compile_filter(name.clone(), def) {
                 Ok(f) => compiled.push(f),
                 Err(e) => eprintln!("[rtk] warning: filter '{}' in {}: {}", name, source, e),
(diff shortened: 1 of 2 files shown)


src/core/toml_filter.rs
```diff
@@ -2032,6 +2032,21 @@ match_command = "^make\\b"
         );
     }
 
+    /// Verify that every built-in filter's match_command starts with `^`.
+    /// Prevents substring misfires where an unanchored match matches a subcommand or path in the middle of a command line.
+    #[test]
+    fn test_builtin_all_filters_match_command_anchored() {
+        let filters = make_filters(BUILTIN_TOML);
+        for filter in &filters {
+            assert!(
```
KuSh, line 2123: `starts_with('^')` is not equivalent to "anchored" — a top-level alternation walks straight through it.

A future `match_command = "^liquibase\\b|/liquibase\\b"` starts with `^`, passes this assertion, and yet its second branch is unanchored: `timeout 5 /usr/bin/liquibase update` gets rewritten to `rtk timeout 5 ...` again, with a green suite. That is #3757, reintroduced.

This isn't a styling hypothetical — `src/filters/spring-boot.toml` already ships a top-level alternation (`^(mvn\s+spring-boot:run|java\s+-jar\s+...|gradle\s+.*bootRun)`). It happens to be parenthesized, but nothing in this guard requires that, so the next contributor who drops the parens gets no warning.

#3757 asked for a guard so "filter 64 can't silently reintroduce this". As written it only catches the one spelling that was already fixed. Two options:

- reject a `|` occurring at nesting depth 0, in addition to the `^` check; or
- assert behaviorally: for each filter, no `"zzwrapper " + <string the filter matches>` should match.

The second is stronger — it tests the property you actually care about rather than a proxy for it.

src/discover/registry.rs
```diff
@@ -6012,4 +6012,20 @@ mod tests {
...
+        assert_eq!(
+            rewrite_command_no_prefixes("timeout 5 /usr/bin/liquibase update", &[]),
+            None
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("nohup /opt/tools/liquibase update", &[]),
+            None
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("liquibase update", &[]),
```
KuSh: This assertion never reaches the TOML filter path it's meant to cover.

`liquibase` has a native RULES entry (`src/discover/rules.rs:979`, `pattern: r"^liquibase(?:\s|$)"`, `rtk_cmd: "rtk liquibase"`), so `classify_command("liquibase update")` returns `Classification::Supported` and `rewrite_segment_inner` returns from the RULES branch long before `toml_filter::command_matches_filter` is consulted.

Verified: with `RTK_NO_TOML=1` this assertion still passes. Delete `src/filters/liquibase.toml` tomorrow and this line stays green — so the PR description's "bare invocation rewrites for TOML filters" isn't what's being exercised here.

src/discover/registry.rs
```diff
@@ -6012,4 +6012,20 @@ mod tests {
             "pest"
         );
     }
+
+    #[test]
+    fn test_rewrite_liquibase_wrapper_and_paths() {
```
KuSh: The whole test degrades to a no-op under `RTK_NO_TOML=1` — verified passing with it set.

`toml_disabled()` short-circuits the `Unsupported` branch, which returns `None` unconditionally at `registry.rs:1396`, so both `None` assertions are satisfied for the wrong reason; the third is satisfied by RULES (see other comment). Anyone running the suite with TOML filters off gets a green regression guard that guards nothing.

Suggest either pinning the env explicitly in the test, or asserting through `find_filter_in` / `command_matches_filter` against an explicitly-constructed registry, the way the `src/core/toml_filter.rs` tests do.

src/discover/registry.rs
```diff
@@ -6012,4 +6012,20 @@ mod tests {
...
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("nohup /opt/tools/liquibase update", &[]),
+            None
+        );
+        assert_eq!(
+            rewrite_command_no_prefixes("liquibase update", &[]),
+            Some("rtk liquibase update".into())
+        );
+    }
```
KuSh, line 6044: Missing the third case #3757 explicitly requires — and the real behavior contradicts the issue.

#3757's "Test coverage (required)" section lists `/usr/bin/liquibase update` → *still rewritten*, as the guard for the path-prefixed case that motivated the original `(?:^|/)` alternation. The test is named `..._wrapper_and_paths` but asserts nothing for a bare path-qualified invocation.

I probed it: `rewrite_command_no_prefixes("/usr/bin/liquibase update", &[])` returns `None` today — as do `/usr/bin/grep -rn foo src` and `/usr/bin/make build`. Cause: `classify_command` basenames via `strip_absolute_path` (`registry.rs:133`) and reports `Supported`, but the rewrite branch then runs `strip_word_prefix(cmd_part, "liquibase")` against the *un-normalized* string (`registry.rs:1474`), finds no prefix, and falls through to `None`.

So the anchor fix didn't regress liquibase specifically — RTK uniformly declines to rewrite path-qualified argv[0]. But merging with "Closes #3757" silently closes that bullet. Please add:

```rust
assert_eq!(rewrite_command_no_prefixes("/usr/bin/liquibase update", &[]), None);
```

to pin the actual behavior, with a comment noting it diverges from the issue text (or open a follow-up if path-qualified rewriting is wanted).

KuSh (changes requested):
Test-only diff (31 lines) — the anchor fix itself already landed in `cab2283`. I validated it in a throwaway worktree: merges cleanly onto latest `develop`, full suite green (2931 passed), and both new tests are non-vacuous — reverting `src/filters/liquibase.toml` to the unanchored `(?:^|/)liquibase(?:\s|$)` makes both fail. So the guards do catch the exact bug from #3757.

No correctness issues. Requesting changes on four coverage gaps, three of them empirically verified — the anchor assertion is weaker than the class of pattern it's meant to prevent, the `discover` test largely tests a path other than the one it claims, and #3757's required `/usr/bin/liquibase update` case is missing (and behaves opposite to what the issue states).

amandeavor:
Added direct built-in TOML coverage so the Liquibase tests cannot be bypassed by the native registry or RTK_NO_TOML. The anchor check now rejects top-level alternation, and the current path-qualified behavior is pinned explicitly. Formatting, Linux clippy, and all three focused regression tests pass.

src/core/toml_filter.rs
KuSh: Fixed and verified on `e521e8d`, in a dedicated worktree.

`has_top_level_alternation` is the right shape — depth counting with escape and character-class handling, so `^(a|b)` passes and `^a|b` doesn't, and the two self-asserts pin both directions.

I mutation-tested the exact case this comment described. Setting `src/filters/liquibase.toml` to `match_command = "^liquibase\\b|/liquibase\\b"` (starts with `^`, would have passed the old guard) now fails:

```
test core::toml_filter::tests::test_builtin_all_filters_match_command_anchored ... FAILED
panicked at src/core/toml_filter.rs:2065
```

And `test_builtin_liquibase_filter_does_not_match_wrapped_or_path_commands` fails alongside it on `timeout 5 /usr/bin/liquibase update`. The guard now catches the class of pattern, not just the one spelling. Resolving.

src/discover/registry.rs
KuSh: Fixed on `e521e8d` — the RULES-shadowed assertion is gone, and the coverage moved to where it actually exercises the filter.

`test_builtin_liquibase_filter_does_not_match_wrapped_or_path_commands` asserts directly against `liquibase.match_regex` from `make_filters(BUILTIN_TOML)`, so `src/discover/rules.rs` can no longer satisfy it by proxy. Verified: reverting the filter to the unanchored `(?:^|/)liquibase(?:\s|$)` makes it fail. Delete or break `src/filters/liquibase.toml` now and the test notices. Resolving.

src/discover/registry.rs
KuSh: Fixed on `e521e8d`. The real coverage no longer goes through `rewrite_command_no_prefixes`, so it's env-independent by construction.

Verified explicitly — with the filter reverted to the unanchored pattern and `RTK_NO_TOML=1` set:

```
test core::toml_filter::tests::test_builtin_all_filters_match_command_anchored ... FAILED
test core::toml_filter::tests::test_builtin_liquibase_filter_does_not_match_wrapped_or_path_commands ... FAILED
```

Both guards still fire with TOML filtering disabled. No vacuous green. Resolving.

src/discover/registry.rs
KuSh: Fixed on `e521e8d` — `test_path_qualified_liquibase_is_not_rewritten` pins `/usr/bin/liquibase update` → `None`, and the comment documents why this diverges from #3757's text (registry normalization classifies the basename without rewriting the original argv[0]).

That's exactly what I was after: the behavior is pinned, and the divergence is recorded rather than silently closed. Resolving.

KuSh (approved):
Verified in a dedicated worktree. All four round-one comments are fixed, each confirmed by mutation rather than by reading:

- Reverting `src/filters/liquibase.toml` to the unanchored `(?:^|/)liquibase(?:\s|$)` fails the guards.
- Setting it to `^liquibase\b|/liquibase\b` — the top-level-alternation case that passed the old `starts_with('^')` check — now fails too.
- Both still fail with `RTK_NO_TOML=1`, so the coverage is no longer env-dependent.

`cargo fmt --all`, `cargo clippy --all-targets` (clean) and `cargo test --all` pass, both on the branch and merged onto the latest `develop` (2981 passing).

I pushed one commit (3522477) addressing the follow-up review findings:

- The anchoring rule moved into a shared `is_fully_anchored` predicate applied at load time in **both** loaders, so it now covers project-local and global filters. Those are user-authored and were unguarded — which is where #3757 said the durable fix belonged. Unanchored filters are skipped with a warning instead of misfiring.
- The built-in invariant test now reads **raw** patterns from `BUILTIN_TOML`, so the new gate cannot hide an unanchored built-in from the assertion.
- Per-branch anchoring (`^pnpm\b|^npm\b`) and inline flags (`(?i)^tool\b`) are accepted rather than rejected with a misleading message.
- The rewrite path is covered end to end through `jj` (TOML-only, never in the `RULES` table), with the bare case as a positive control so the wrapper assertions cannot pass vacuously.
- The wrapper cases are folded into the existing `test_liquibase_match_command_ignores_path_substring`, next to its rationale comment.

PR #3808, opened 2026-09-01
fix(discover): count sudo RTK_DISABLED=1 commands as bypass in analytics

### LINKED ISSUE #3756: discover/gain: `sudo RTK_DISABLED=1 …` is not counted as an RTK bypass
## Summary

Once #2810 lands, `sudo RTK_DISABLED=1 <cmd>` stops being counted as an RTK bypass by `rtk discover` and `rtk gain`. The command still correctly runs unfiltered — this is an **analytics blind spot**, not a runtime bug.

## Cause

`strip_disabled_prefix` (`src/discover/registry.rs:484`) peels the env prefix using the shared `ENV_PREFIX` regex. #2810 removes `sudo\s+` from that regex (deliberately, and correctly — see #146). `ENV_PREFIX` is anchored at `^`, so for `sudo RTK_DISABLED=1 docker ps` nothing matches at offset 0 and the returned `env_prefix` is empty — the `RTK_DISABLED=` never gets seen.

Probed on #2810's head (`713b448`):

```
sudo RTK_DISABLED=1 docker ps      prefix=""                 rest="sudo RTK_DISABLED=1 docker ps"  has_disabled=false
sudo -E RTK_DISABLED=1 docker ps   prefix=""                 rest="sudo -E RTK_DISABLED=1 docker…"  has_disabled=false
RTK_DISABLED=1 sudo docker ps      prefix="RTK_DISABLED=1 "  rest="sudo docker ps"                  has_disabled=true
RTK_DISABLED=1 docker ps           prefix="RTK_DISABLED=1 "  rest="docker ps"                       has_disabled=true
```

Only the `sudo`-first ordering is affected; `RTK_DISABLED=1 sudo …` still works.

## Impact

Two consumers:

- `src/discover/mod.rs:101` — the `RTK_DISABLED BYPASS` section of `rtk discover`. The command is silently reclassified: it drops out of the bypass list and shows up under `TOP UNHANDLED` with base command `sudo RTK_DISABLED=1`.
- `src/analytics/gain.rs:674` (`check_rtk_disabled_bypass`, via `cmd_has_rtk_disabled_prefix` at `registry.rs:478`) — undercounts the bypass rate that drives the >10% warning.

Measured on a synthetic 8-command session with a fake `$HOME`:

|  | before #2810 | after #2810 |
|---|---|---|
| `RTK_DISABLED BYPASS` section | 3 commands — `docker ps (2x)`, `git status (1x)` | 1 command — `git status (1x)` |
| `sudo RTK_DISABLED=1 docker ps` | counted as a bypass of `docker ps` | lands in `TOP UNHANDLED` as base `sudo RTK_DISABLED=1` |

The hook path is unaffected: `rtk rewrite "RTK_DISABLED=1 sudo docker ps"` still prints the warning and exits 1, because that ordering is matched by the `VAR=val` alternation.

## Fix — and one thing the fix must *not* do

**Do not put `sudo` back into `strip_disabled_prefix`.** That function is shared: it is also called from `rewrite_segment_inner` at `registry.rs:1303`, so re-adding `sudo` there would immediately reintroduce the broken `sudo rtk …` rewrite that #2810 exists to remove.

The peel has to be analytics-only — a small local helper that skips a leading `sudo` (plus its flags) before delegating to `strip_disabled_prefix`, applied at the two consumer sites (`discover/mod.rs:101`, `gain.rs:674`) and nowhere else.

**`rtk sudo` (#146) does not fix this on its own.** Even with a `^sudo` rule in place, `strip_disabled_prefix` still won't peel `sudo`, so this stays broken until it is addressed directly.

## Test coverage (required)

There is currently **no test at all** for `sudo` combined with `RTK_DISABLED` — `test_cmd_has_rtk_disabled_prefix` (`registry.rs:4796`) and `test_strip_disabled_prefix` (`registry.rs:4810`) only cover the `VAR=val` orderings, which is why CI is green through #2810. Any fix must add:

- `sudo RTK_DISABLED=1 docker ps` and `sudo -E RTK_DISABLED=1 docker ps` → detected as a bypass of `docker ps`
- `RTK_DISABLED=1 sudo docker ps` → still detected (regression guard)
- `sudo docker ps` → not a bypass
- a guard asserting `strip_disabled_prefix("sudo docker ps")` still returns an empty prefix, pinning that the shared helper was not the thing that changed

## Context

Found while reviewing #2810. Not a blocker for it — the bypass hole is strictly smaller than the `sudo rtk …` breakage that PR removes.

amandeavor:
## Summary

Adds an analytics-only prefix stripper (\strip_disabled_prefix_for_analytics\) that skips a leading \sudo\ (and optional sudo flags like \-E\) before delegating to the shared \strip_disabled_prefix\. This ensures that commands such as \sudo RTK_DISABLED=1 docker ps\ are counted as RTK bypasses in \
tk discover\ and \
tk gain\ without modifying the shared \strip_disabled_prefix\ used by rewrite routines.

Closes #3756

## Validation

- Added unit tests in \	est_cmd_has_rtk_disabled_prefix\ covering:
  - \sudo RTK_DISABLED=1 docker ps\ -> detected as bypass
  - \sudo -E RTK_DISABLED=1 docker ps\ -> detected as bypass
  - \RTK_DISABLED=1 sudo docker ps\ -> detected as bypass
  - \sudo docker ps\ -> not a bypass
- Added \	est_strip_disabled_prefix_for_analytics\ verifying prefix extraction
- Pinned \	est_strip_disabled_prefix\ to ensure the shared helper behavior remains unchanged
- Ran test suite with \cargo test --bin rtk\ (all tests passing)

### DIFF
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -470,13 +470,27 @@ fn strip_absolute_path(cmd: &str) -> String {
     }
 }
 
+static SUDO_PREFIX_ANALYTICS: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"^(?:sudo(?:\s+-[A-Za-z0-9_-]+(?:\s+\S+)?)*\s+)").unwrap()
+});
+
 pub fn prefix_contains_rtk_disabled(prefix_part: &str) -> bool {
     prefix_part.contains("RTK_DISABLED=")
 }
 
+/// Strip RTK_DISABLED=X and other env prefixes for analytics, skipping leading `sudo` and flags if present.
+pub fn strip_disabled_prefix_for_analytics(cmd: &str) -> (&str, &str) {
+    let trimmed = cmd.trim();
+    let after_sudo = SUDO_PREFIX_ANALYTICS.replace(trimmed, "");
+    let sudo_len = trimmed.len() - after_sudo.len();
+    let (_env_part, rest) = strip_disabled_prefix(&trimmed[sudo_len..]);
+    let full_prefix_len = trimmed.len() - rest.len();
+    (&trimmed[..full_prefix_len], rest)
+}
+
 /// Check if a command has RTK_DISABLED= prefix in its env prefix portion.
 pub fn cmd_has_rtk_disabled_prefix(cmd: &str) -> bool {
-    let (prefix_part, _) = strip_disabled_prefix(cmd);
+    let (prefix_part, _) = strip_disabled_prefix_for_analytics(cmd);
     prefix_contains_rtk_disabled(prefix_part)
 }
 
@@ -5137,9 +5151,19 @@ mod tests {
         assert!(cmd_has_rtk_disabled_prefix(
             "RTK_DISABLED=true git log --oneline"
         ));
+        assert!(cmd_has_rtk_disabled_prefix(
+            "sudo RTK_DISABLED=1 docker ps"
+        ));
+        assert!(cmd_has_rtk_disabled_prefix(
+            "sudo -E RTK_DISABLED=1 docker ps"
+        ));
+        assert!(cmd_has_rtk_disabled_prefix(
+            "RTK_DISABLED=1 sudo docker ps"
+        ));
         assert!(!cmd_has_rtk_disabled_prefix("git status"));
         assert!(!cmd_has_rtk_disabled_prefix("rtk git status"));
         assert!(!cmd_has_rtk_disabled_prefix("SOME_VAR=1 git status"));
+        assert!(!cmd_has_rtk_disabled_prefix("sudo docker ps"));
     }
 
     #[test]
@@ -5153,6 +5177,22 @@ mod tests {
             ("FOO=1 RTK_DISABLED=1 ", "cargo test")
         );
         assert_eq!(strip_disabled_prefix("git status"), ("", "git status"));
+    }
+
+    #[test]
+    fn test_strip_disabled_prefix_for_analytics() {
+        assert_eq!(
+            strip_disabled_prefix_for_analytics("sudo RTK_DISABLED=1 docker ps"),
+            ("sudo RTK_DISABLED=1 ", "docker ps")
+        );
+        assert_eq!(
+            strip_disabled_prefix_for_analytics("sudo -E RTK_DISABLED=1 docker ps"),
+            ("sudo -E RTK_DISABLED=1 ", "docker ps")
+        );
+        assert_eq!(
+            strip_disabled_prefix_for_analytics("RTK_DISABLED=1 sudo docker ps"),
+            ("RTK_DISABLED=1 sudo ", "docker ps")
+        );
     }
 
     // --- #485: absolute path normalization ---
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -470,22 +470,60 @@ fn strip_absolute_path(cmd: &str) -> String {
     }
 }
 
-static SUDO_PREFIX_ANALYTICS: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new(r"^(?:sudo(?:\s+-[A-Za-z0-9_-]+(?:\s+\S+)?)*\s+)").unwrap()
-});
-
 pub fn prefix_contains_rtk_disabled(prefix_part: &str) -> bool {
     prefix_part.contains("RTK_DISABLED=")
 }
 
-/// Strip RTK_DISABLED=X and other env prefixes for analytics, skipping leading `sudo` and flags if present.
+/// Strip an RTK_DISABLED prefix for analytics, including sudo options around it.
 pub fn strip_disabled_prefix_for_analytics(cmd: &str) -> (&str, &str) {
     let trimmed = cmd.trim();
-    let after_sudo = SUDO_PREFIX_ANALYTICS.replace(trimmed, "");
-    let sudo_len = trimmed.len() - after_sudo.len();
-    let (_env_part, rest) = strip_disabled_prefix(&trimmed[sudo_len..]);
-    let full_prefix_len = trimmed.len() - rest.len();
-    (&trimmed[..full_prefix_len], rest)
+    let tokens = tokenize(trimmed);
+
+    if let Some(disabled_index) = tokens
+        .iter()
+        .position(|token| token.value.starts_with("RTK_DISABLED="))
+    {
+        let mut saw_sudo = false;
+        let mut saw_sudo_flag = false;
+        let prefix_is_valid = tokens[..disabled_index].iter().all(|token| {
+            let value = token.value.as_str();
+            if value == "env" || value.contains('=') {
+                true
+            } else if value == "sudo" {
+                saw_sudo = true;
+                true
+            } else if saw_sudo && value.starts_with('-') {
+                saw_sudo_flag = true;
+                true
+            } else {
+                saw_sudo && saw_sudo_flag
+            }
+        });
+
+        if prefix_is_valid {
+            for token in &tokens[disabled_index + 1..] {
+                if token.kind != TokenKind::Arg {
+                    break;
+                }
+                if token.value == "sudo"
+                    || token.value == "env"
+                    || token.value.starts_with('-')
+                    || token.value.contains('=')
+                {
+                    continue;
+                }
+                let candidate = trimmed[token.offset..].trim();
+                if matches!(
+                    classify_command(candidate),
+                    Classification::Supported { .. }
+                ) {
+                    return (&trimmed[..token.offset], candidate);
+                }
+            }
+        }
+    }
+
+    strip_disabled_prefix(trimmed)
 }
 
 /// Check if a command has RTK_DISABLED= prefix in its env prefix portion.
@@ -5151,14 +5189,16 @@ mod tests {
         assert!(cmd_has_rtk_disabled_prefix(
             "RTK_DISABLED=true git log --oneline"
         ));
-        assert!(cmd_has_rtk_disabled_prefix(
-            "sudo RTK_DISABLED=1 docker ps"
-        ));
+        assert!(cmd_has_rtk_disabled_prefix("sudo RTK_DISABLED=1 docker ps"));
         assert!(cmd_has_rtk_disabled_prefix(
             "sudo -E RTK_DISABLED=1 docker ps"
         ));
         assert!(cmd_has_rtk_disabled_prefix(
-            "RTK_DISABLED=1 sudo docker ps"
+            "sudo -E -u root RTK_DISABLED=1 docker ps"
+        ));
+        assert!(cmd_has_rtk_disabled_prefix("RTK_DISABLED=1 sudo docker ps"));
+        assert!(cmd_has_rtk_disabled_prefix(
+            "RTK_DISABLED=1 sudo -E docker ps"
         ));
         assert!(!cmd_has_rtk_disabled_prefix("git status"));
         assert!(!cmd_has_rtk_disabled_prefix("rtk git status"));
@@ -5190,8 +5230,16 @@ mod tests {
             ("sudo -E RTK_DISABLED=1 ", "docker ps")
         );
         assert_eq!(
+            strip_disabled_prefix_for_analytics("sudo -E -u root RTK_DISABLED=1 docker ps"),
+            ("sudo -E -u root RTK_DISABLED=1 ", "docker ps")
+        );
+        assert_eq!(
             strip_disabled_prefix_for_analytics("RTK_DISABLED=1 sudo docker ps"),
             ("RTK_DISABLED=1 sudo ", "docker ps")
+        );
+        assert_eq!(
+            strip_disabled_prefix_for_analytics("RTK_DISABLED=1 sudo -E docker ps"),
+            ("RTK_DISABLED=1 sudo -E ", "docker ps")
         );
     }
 
--- a/src/discover/mod.rs
+++ b/src/discover/mod.rs
@@ -11,8 +11,8 @@ use std::collections::HashMap;
 
 use provider::{ClaudeProvider, SessionProvider};
 use registry::{
-    category_avg_tokens, classify_command, split_command_chain, strip_disabled_prefix,
-    Classification,
+    category_avg_tokens, classify_command, split_command_chain,
+    strip_disabled_prefix_for_analytics, Classification,
 };
 use report::{DiscoverReport, SupportedEntry, UnsupportedEntry};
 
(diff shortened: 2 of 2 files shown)


src/discover/registry.rs
```diff
@@ -470,13 +470,27 @@ fn strip_absolute_path(cmd: &str) -> String {
     }
 }
 
+static SUDO_PREFIX_ANALYTICS: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"^(?:sudo(?:\s+-[A-Za-z0-9_-]+(?:\s+\S+)?)*\s+)").unwrap()
```
KuSh: The flag group `(?:\s+-[A-Za-z0-9_-]+(?:\s+\S+)?)*` treats *every* sudo flag as value-taking, so a value-less flag eats the following token. That produces a false negative for the exact case this PR is fixing:

`sudo -E -u root RTK_DISABLED=1 docker ps`
- the `-E` iteration consumes `-u`, then the group stops at the non-flag `root`
- match = `sudo -E -u `, remainder = `root RTK_DISABLED=1 docker ps`
- `strip_disabled_prefix` can't strip a prefix starting with `root`, so the returned prefix is `sudo -E -u `, `prefix_contains_rtk_disabled` is false, and the bypass is **not** counted (it classifies as `Unsupported { base_command: "root RTK_DISABLED=1" }`)

Note the `sudo -E RTK_DISABLED=1 docker ps` test only passes by accident: `-E` swallows `RTK_DISABLED=1` as its "value", so the token happens to land inside the prefix anyway.

Simplest in-scope fix: only consume the token after a flag when it is neither another flag nor a `KEY=VALUE` assignment — or don't consume it at all. Value-less (`-E -H -k -n -i -s -b -A --`) vs value-taking (`-u -g -p -C -D -h`) arity shouldn't be hardcoded here (see the scope note in the review body).

src/discover/registry.rs
```diff
@@ -470,13 +470,27 @@ fn strip_absolute_path(cmd: &str) -> String {
...
 pub fn prefix_contains_rtk_disabled(prefix_part: &str) -> bool {
     prefix_part.contains("RTK_DISABLED=")
 }
 
+/// Strip RTK_DISABLED=X and other env prefixes for analytics, skipping leading `sudo` and flags if present.
+pub fn strip_disabled_prefix_for_analytics(cmd: &str) -> (&str, &str) {
+    let trimmed = cmd.trim();
+    let after_sudo = SUDO_PREFIX_ANALYTICS.replace(trimmed, "");
+    let sudo_len = trimmed.len() - after_sudo.len();
+    let (_env_part, rest) = strip_disabled_prefix(&trimmed[sudo_len..]);
```
KuSh: The sudo skip runs only *before* `strip_disabled_prefix`, so the mirrored ordering with flags is still broken:

`RTK_DISABLED=1 sudo -E docker ps`
- the sudo regex doesn't match at position 0, and `strip_disabled_prefix` stops at `-E`
- prefix = `RTK_DISABLED=1 sudo ` (detected ✅) but `actual_cmd = "-E docker ps"` → `Unsupported { base_command: "-E docker" }`
- in `src/discover/mod.rs:103-113` the non-`Supported` arm falls through to `continue`, so the bypass is detected and then silently dropped from `rtk_disabled_count` / `rtk_disabled_cmds`

Same class of bug as #3756. The tests here only cover the flagless `RTK_DISABLED=1 sudo docker ps`; the flagged variant is worth adding in both orderings.

src/discover/registry.rs
```diff
@@ -470,13 +470,27 @@ fn strip_absolute_path(cmd: &str) -> String {
...
+    let full_prefix_len = trimmed.len() - rest.len();
+    (&trimmed[..full_prefix_len], rest)
```
KuSh: Same greediness, seen on the returned tuple: `strip_disabled_prefix_for_analytics("sudo -E docker ps")` returns `("sudo -E docker ", "ps")`, and `classify_command("ps")` then yields `Supported { rtk_equivalent: "rtk ps" }` — a docker command reported as `ps`.

It's latent today because `src/discover/mod.rs:101` uses the second element only inside the `prefix_contains_rtk_disabled` branch, but the fn is `pub` and its doc comment promises `(env_prefix, actual_command)`. `sudo -E RTK_DISABLED=1 -H docker ps` mis-attributes already.

src/discover/registry.rs
```diff
@@ -470,13 +470,27 @@ fn strip_absolute_path(cmd: &str) -> String {
     }
 }
 
+static SUDO_PREFIX_ANALYTICS: LazyLock<Regex> = LazyLock::new(|| {
```
KuSh: `cargo fmt --all -- --check` fails on this branch: rustfmt wants this static collapsed to two lines, and wants two of the new `assert!(cmd_has_rtk_disabled_prefix(...))` calls in `test_cmd_has_rtk_disabled_prefix` on a single line.

The description mentions only `cargo test --bin rtk` — the repo gate is `cargo fmt --all && cargo clippy --all-targets && cargo test --all`. Clippy and the tests are green, only fmt is failing.

KuSh (changes requested):
Thanks — the bug is real and the direction is right, but the new sudo prefix regex makes a flag-arity assumption that reintroduces the same false negative it's meant to fix, so requesting changes.

Three correctness items inline, plus a `cargo fmt` failure.

**Design note**: rather than a second regex sitting next to `ENV_PREFIX`, consider teaching `ENV_PREFIX` (registry.rs:69) itself to tolerate sudo flags — it already alternates on `sudo\s+`. Doing it there fixes both orderings in one place *and* fixes `classify_command`, which today returns `Unsupported { base_command: "-E docker" }` for `sudo -E docker ps`. So flagged-sudo commands are currently missing from the supported/rewrite side of `rtk discover`, not just from the bypass counter.

**Scope note**: please keep the arg parsing here deliberately shallow — no sudo option table, no flag/value arity map. #3681 is landing a shared CLI-arg tokenizer for exactly that job, and duplicating a mini-parser here will only have to be torn out. This fix just needs to stop swallowing tokens it can't identify.

amandeavor:
Fixed the sudo parsing issues from the review, including flags with values and RTK_DISABLED before sudo. Added regression coverage for both command detection and analytics tuples. Formatting and clippy pass, the full Linux unit suite passes with 2,930 tests, and the focused regressions pass. The only Linux integration failure locally is the existing unwritable-directory test because this WSL run is root, so chmod does not make the directory unwritable.

src/discover/registry.rs
KuSh: Fixed in b7a642c — the regex is gone and the prefix is walked with the existing `lexer::tokenize`, so no flag-arity assumption is made any more. Verified on the new head: `sudo -E -u root RTK_DISABLED=1 docker ps` → prefix `sudo -E -u root RTK_DISABLED=1 `, actual `docker ps`, `Supported { rtk docker }`. `sudo -E RTK_DISABLED=1 -H docker ps` is right too. Resolving.

src/discover/registry.rs
KuSh: Fixed in b7a642c — `RTK_DISABLED=` is now located anywhere in the prefix rather than only after a leading sudo, so both orderings work. Verified: `RTK_DISABLED=1 sudo -E docker ps` → prefix `RTK_DISABLED=1 sudo -E `, actual `docker ps`, `Supported { rtk docker }`, and a test now covers the flagged variant in both orders. Resolving.

src/discover/registry.rs
KuSh: Fixed in b7a642c — `strip_disabled_prefix_for_analytics("sudo -E docker ps")` now returns `("sudo ", "-E docker ps")` instead of `("sudo -E docker ", "ps")`, so the `ps` mis-attribution is gone. (The `classify_command` gap from the review body is still open — that one gives `Unsupported { base_command: "-E docker" }` — but that is out of this thread’s scope.) Resolving.

src/discover/registry.rs
KuSh: Fixed in b7a642c — full gate is green on the new head: `cargo fmt --all -- --check`, `cargo clippy --all-targets`, and `cargo test --all` (2930 + 83 tests) all pass. Resolving.

src/discover/registry.rs
```diff
@@ -474,9 +474,61 @@ pub fn prefix_contains_rtk_disabled(prefix_part: &str) -> bool {
...
+                    continue;
+                }
+                let candidate = trimmed[token.offset..].trim();
+                if matches!(
+                    classify_command(candidate),
+                    Classification::Supported { .. }
+                ) {
+                    return (&trimmed[..token.offset], candidate);
+                }
+            }
```
KuSh, line 522: This loop returns the first *remaining token* whose tail classifies as `Supported`, not the command word. So any wrapper command gets skipped and one of its arguments is reported as the bypassed command. Verified on this head:

- `RTK_DISABLED=1 ssh host docker ps` → `("RTK_DISABLED=1 ssh host ", "docker ps")`, counted as a `docker` bypass with `rtk docker ps` recommended — but the command runs on a *remote* host, where RTK isn't even installed. Pre-PR this classified as unsupported and wasn't counted.
- `RTK_DISABLED=1 xargs -n1 git show` → attributed to `git show`.
- `RTK_DISABLED=1 watch -n1 git status` → attributed to `git status`.
- `sudo -E RTK_DISABLED=1 ./deploy.sh docker ps` → attributed to `docker ps`, i.e. the script's arguments.

The `continue` filter on lines 508-513 makes it worse, because an unrecognised or quoted command word is silently skipped rather than ending the search: `sudo -E RTK_DISABLED=1 'docker' ps` → `("… 'docker' ", "ps")`, reporting `ps`.

Stopping at the first plausible command word — and giving up if it doesn't classify — would keep all of these out.

src/discover/registry.rs
```diff
@@ -474,9 +474,61 @@ pub fn prefix_contains_rtk_disabled(prefix_part: &str) -> bool {
...
+            } else if value == "sudo" {
+                saw_sudo = true;
+                true
+            } else if saw_sudo && value.starts_with('-') {
+                saw_sudo_flag = true;
+                true
+            } else {
+                saw_sudo && saw_sudo_flag
+            }
+        });
```
KuSh, line 501: Once `sudo` plus any single `-flag` has been seen, `saw_sudo && saw_sudo_flag` accepts *every* later token as part of the env prefix — so an `RTK_DISABLED=` buried anywhere in the command line reads as an RTK bypass prefix.

`sudo -E docker run -e RTK_DISABLED=1 myimage npm run build` → prefix `sudo -E docker run -e RTK_DISABLED=1 myimage `, actual `npm run build`, detected ✅. But that `RTK_DISABLED=1` is a **container** env var passed through `docker run -e`, the real command is `docker run`, and npm runs inside the container. Pre-PR: not detected.

What pins the catch-all as the culprit: the same line *without* `-E` (`sudo docker run -e RTK_DISABLED=1 img …`) is correctly rejected, because `saw_sudo_flag` never gets set. So the detection depends on an unrelated sudo flag being present.

Accepting only `sudo` / `env` / `KEY=VALUE` / `-flags` and bailing on anything else is stricter than what's here and drops the false positive.

src/discover/registry.rs
```diff
@@ -474,9 +474,61 @@ pub fn prefix_contains_rtk_disabled(prefix_part: &str) -> bool {
...
 
+/// Strip an RTK_DISABLED prefix for analytics, including sudo options around it.
+pub fn strip_disabled_prefix_for_analytics(cmd: &str) -> (&str, &str) {
+    let trimmed = cmd.trim();
+    let tokens = tokenize(trimmed);
+
+    if let Some(disabled_index) = tokens
+        .iter()
+        .position(|token| token.value.starts_with("RTK_DISABLED="))
+    {
```
KuSh, line 485: `position()` scans the whole token vector, and the validity closure below accepts `Operator` / `Pipe` tokens through the same catch-all (`&&` is not `env`, contains no `=`, doesn't start with `-`, so it lands in the `else` arm). The `token.kind != TokenKind::Arg` guard only exists in the loop at line 505 — i.e. *after* the match.

`src/discover/mod.rs:101` is shielded because it calls `split_command_chain` first, but `src/analytics/gain.rs:674` passes the **whole, unsplit** command line to `cmd_has_rtk_disabled_prefix`. Verified: `sudo -E ls && docker run -e RTK_DISABLED=1 img git status` → `has_disabled=true`, actual `git status`. That inflates the `rtk gain` "used RTK_DISABLED=1 unnecessarily" percentage that gates the >10% warning.

Bailing out on any non-`Arg` token before `disabled_index` closes this one.

KuSh (changes requested):
Round 2 on `b7a642c`. The four threads from the first round are fixed and resolved — the flag-arity guess is gone, both orderings work, the returned tuple is no longer mangled, and the full gate is green on my machine (`cargo fmt --all -- --check`, `cargo clippy --all-targets`, `cargo test --all`). Thanks for reusing `lexer::tokenize` rather than growing a second regex, and for keeping the sudo option table out of it.

The new heuristic overshoots in the other direction, though. It never *misses* a bypass relative to `strip_disabled_prefix` — the returned prefix always spans the `RTK_DISABLED=` token, and it falls back to the old helper otherwise — but it now *invents* bypasses that aren't there. Three inline, each verified by running `strip_disabled_prefix_for_analytics` / `cmd_has_rtk_disabled_prefix` on this head:

| input | reported as | reality |
|---|---|---|
| `RTK_DISABLED=1 ssh host docker ps` | `docker` bypass, suggests `rtk docker ps` | runs on a remote host, RTK can't apply |
| `sudo -E docker run -e RTK_DISABLED=1 img npm run build` | `npm` bypass | container env var, npm runs inside the container |
| `sudo -E ls && docker run -e RTK_DISABLED=1 img git status` | `git` bypass | detection crossed a `&&` |

The third one is the one with teeth: `src/discover/mod.rs:101` is protected because it calls `split_command_chain` first, but `src/analytics/gain.rs:674` passes the whole unsplit line to `cmd_has_rtk_disabled_prefix`, so this inflates the `rtk gain` "used RTK_DISABLED=1 unnecessarily" percentage that gates the >10% warning.

**On scope / #3681**: the loop is now doing the two things a real arg tokenizer does — deciding whether a token is a flag's value or a command word, and where the argument list ends. That's exactly #3681's job, and I'd rather not have a second implementation of it here. But #3681 is still a draft with changes requested, so I don't think this should sit and wait on it. My suggestion is to make this round deliberately *narrower* rather than smarter:

- bail out entirely if any token before `disabled_index` is not `TokenKind::Arg`;
- drop the `saw_sudo && saw_sudo_flag` catch-all — accept only `sudo`, `env`, `KEY=VALUE` and `-flags` in the prefix, and give up on anything else;
- stop at the *first* command-word candidate after the prefix instead of scanning forward until something classifies as `Supported`.

That fixes the bug you set out to fix, keeps every false positive above out, and leaves a small enough surface that swapping in #3681's tokenizer later is a rewrite of ten lines rather than an untangling. Happy to be argued out of it if you see a case the narrow version misses.

PR #3668, opened 2026-08-23
feat(discover): rewrite head -n N and head --lines N

dwin-gharibi:
Closes #3667.

## Summary

- Adds `HEAD_N_SPACE` (`head -n N file`) and `HEAD_LINES_SPACE` (`head --lines N file`), mirroring
  the `tail` arms that already exist, so head and tail are rewritten through the same four
  spellings.
- Forms `rtk read` cannot express — `head -c N` (bytes), multi-file, stdin — still fall through to
  the native binary.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [x] Manual testing: `rtk <command>` output inspected

### Unit tests (added, TDD — red before green)

`src/discover/registry.rs`:

- `head_and_tail_rewrite_the_same_spellings` — a table of all eight forms. Red on `develop` with
  `head -n 20 f.txt should rewrite to rtk read f.txt --max-lines 20`.
- `unsupported_head_forms_are_left_alone` — `head -c 500 f.txt`, `head -3 a b c`, bare `head`
  stay unrewritten, so a byte count is never silently reinterpreted as a line count.

### Behavioural check through `rtk rewrite`

All four head spellings and all four tail spellings produce the expected rewrite; `head -c 500`
and `head -3 a b c` return exit 1 (passthrough).

### Suite

`cargo test --all`: 2627 passed, 0 failed (2625 on `develop`, +2 from this PR).
`cargo clippy --all-targets`: clean. `cargo fmt --all`: clean.

PR #3828, opened 2026-09-02
fix(rewrite): rewrite inner commands of quoted sh/bash/zsh -c wrappers

SindriFr:
## Summary

Rewrites the portable commands inside an exact quoted `sh -c`, `bash -c`, or `zsh -c` wrapper, leaving the wrapper's own bytes untouched.

```
bash -c "head foo && grep -R bar ."
  → bash -c "rtk read foo && rtk grep -R bar ."
```

Agents that reach a project toolchain through a wrapper (`direnv exec . bash -c '…'`, `devenv shell -- bash -c '…'`, container and CI runners) send every command as a quoted script, so the whole invocation was opaque and no inner command was rewritten.

`shell_wrapper.rs` and most tests are @pashifika's, from #3057, rebuilt against current `develop` — the rewrite engine has since gained `RewriteContext`, `PipeKind`, `analyze_pipeline`, `pipeline_final_safe` and multi-line block handling. Credited via `Co-authored-by`.

Refs #2767. That issue also asks for `fish -c`, which this deliberately leaves out (see Passthrough cases), so it does not close it.

**Overlaps #3057**, which is still open. Landing this one should close that one; I have no write access there.

## Eligibility

Deliberately narrow. The shell must be exactly `sh`, `bash`, or `zsh` (any path, e.g. `/bin/bash`), the option exactly `-c`, and the script one balanced single- or double-quoted argument.

Only the script's byte span is replaced, so shell path, quote delimiters, horizontal spacing, suffix arguments and fd redirects come back byte-identical:

```
/bin/bash -c 'git status' command-name 'a b'
  → /bin/bash -c 'rtk git status' command-name 'a b'

bash -c 'git status' 2>&1
  → bash -c 'rtk git status' 2>&1
```

A wrapper is found at any prefix depth, so one behind a transparent prefix or a strippable `command` is reached too:

```
direnv exec . bash -c 'git status'
  → direnv exec . bash -c 'rtk git status'
```

## Passthrough cases

These return to the host unchanged:

- a leading `sudo` — the inner `rtk` would run as root, where it is off `secure_path`, the same reason a bare `sudo docker ps` is left alone (#146). Pinned by `test_shell_wrapper_behind_sudo_is_not_rewritten`.
- a script containing a newline, heredoc, or `$((`
- active outer `$` expansion, backticks, or backslash escapes in a double-quoted script
- anything `contains_unattestable_construct` rejects — command/process substitution, file-target redirects
- extra shell options: `bash -lc`, `bash -e -c`, `fish --command`
- an unquoted or unbalanced script
- a wrapper nested inside a wrapper — recursion is capped at one level
- `fish`, whose control keywords (`and`, `or`, `if`) at a command boundary need a fish-aware lexer the shared one does not model

## Permission handling

A host's allow rule matched the outer command — it never parsed the quoted script. Rewriting inside that script therefore must not widen what the rule approved, so:

- deny rules are also evaluated against the inner commands, and an inner deny outranks an outer allow
- a wrapper rewrite is never auto-allowed; its strongest verdict is `Ask`, so a `*` allow rule cannot approve a script the host never saw
- `Ask`-capable hosts prompt, others keep their native permission flow

One parser backs both rewriting and permission evaluation, so the two paths agree on the boundary. Agent protocol coverage: Claude, Cursor, Gemini, Copilot CLI, Droid.

## Test plan

- [x] `cargo test --all` — 3,137 passed, 0 failed, 8 ignored, all 12 suites green
- [x] 30 new unit tests (parser, registry, permissions, agent protocols) + 4 CLI integration tests in `tests/rewrite_shell_wrapper_test.rs`
- [x] `rtk rewrite` verified end-to-end on wrapped and prefixed forms
- [ ] CI to validate Linux and Windows

Tested on Linux x86_64 (NixOS).

### DIFF
--- a/src/hooks/permissions.rs
+++ b/src/hooks/permissions.rs
@@ -5,6 +5,7 @@ use super::constants::{
 use super::init::resolve_claude_dir;
 use crate::core::stream::exec_capture;
 use crate::discover::lexer::{is_word_boundary_whitespace, split_for_permissions};
+use crate::discover::shell_wrapper::{is_shell_wrapper_candidate, parse_shell_wrapper};
 use serde_json::Value;
 use std::path::PathBuf;
 
@@ -80,6 +81,10 @@ pub(crate) fn check_command_with_rules(
         }
     }
 
+    if let Some(verdict) = check_shell_wrapper_permissions(&segments, deny_rules) {
+        return verdict;
+    }
+
     // Can't decompose substitution / file-target redirects — never auto-allow.
     if crate::discover::lexer::contains_unattestable_construct(cmd) {
         return PermissionVerdict::Ask;
@@ -130,6 +135,36 @@ pub(crate) fn check_command_with_rules(
     } else {
         PermissionVerdict::Default
     }
+}
+
+fn check_shell_wrapper_permissions(
+    segments: &[&str],
+    deny_rules: &[String],
+) -> Option<PermissionVerdict> {
+    let mut contains_shell_wrapper = false;
+    for segment in segments {
+        let segment = segment.trim();
+        let Some(wrapper) = parse_shell_wrapper(segment) else {
+            contains_shell_wrapper |= is_shell_wrapper_candidate(segment);
+            continue;
+        };
+        let Some(script) = wrapper.script(segment) else {
+            return Some(PermissionVerdict::Ask);
+        };
+        contains_shell_wrapper = true;
+        let inner_denied = split_for_permissions(script).iter().any(|inner_segment| {
+            deny_rules
+                .iter()
+                .any(|pattern| command_matches_pattern(inner_segment.trim(), pattern))
+        });
+        if inner_denied {
+            return Some(PermissionVerdict::Deny);
+        }
+    }
+
+    // A quoted shell script hides a second parsing boundary from the host rule.
+    // Rewriting it is useful, but RTK must never turn that into auto-approval.
+    contains_shell_wrapper.then_some(PermissionVerdict::Ask)
 }
 
 /// Load deny, ask, and allow Bash rules from all Claude Code settings files.
@@ -1269,6 +1304,56 @@ mod tests {
         assert_eq!(
             check_command_with_rules("rm -rf /", &[], &[], &allow),
             PermissionVerdict::Default
+        );
+    }
+    #[test]
+    fn test_shell_wrapper_never_auto_allowed() {
+        let allow = vec!["*".to_string()];
+        assert_eq!(
+            check_command_with_rules(r#"bash -c "head foo && grep -R bar .""#, &[], &[], &allow),
+            PermissionVerdict::Ask
+        );
+    }
+
+    #[test]
+    fn test_shell_wrapper_inner_deny_wins() {
+        let deny = vec!["rm:*".to_string()];
+        let allow = vec!["*".to_string()];
+        assert_eq!(
+            check_command_with_rules(
+                "bash -c 'git status; rm -rf /tmp/example'",
+                &deny,
+                &[],
+                &allow
+            ),
+            PermissionVerdict::Deny
+        );
+    }
+
+    #[test]
+    fn test_unsupported_shell_wrapper_candidate_never_auto_allowed() {
+        let allow = vec!["*".to_string()];
+        for command in ["bash -lc 'git status'", "bash -e -c 'git status'"] {
+            assert_eq!(
+                check_command_with_rules(command, &[], &[], &allow),
+                PermissionVerdict::Ask,
+                "unsupported command-string wrapper must ask: {command:?}"
+            );
+        }
+    }
+
+    #[test]
+    fn test_shell_wrapper_inner_deny_wins_inside_outer_compound() {
+        let deny = vec!["rm:*".to_string()];
+        let allow = vec!["*".to_string()];
+        assert_eq!(
+            check_command_with_rules(
+                "bash -c 'git status; rm -rf /tmp/example' && cargo test",
+                &deny,
+                &[],
+                &allow
+            ),
+            PermissionVerdict::Deny
         );
     }
 }
--- a/README.md
+++ b/README.md
@@ -345,6 +345,19 @@ The most effective way to use rtk. The hook transparently intercepts Bash comman
 
 **Scope note:** this only applies to Bash tool calls. Claude Code built-in tools such as `Read`, `Grep`, and `Glob` bypass the hook, so use shell commands or explicit `rtk` commands when you want RTK filtering there.
 
+Simple quoted shell wrappers are also rewritten without changing the selected
+shell:
+
+```bash
+bash -c "head foo && grep -R bar ."
+# → bash -c "rtk read foo && rtk grep -R bar ."
+```
+
+This support is intentionally conservative: it covers exact `sh -c`,
+`bash -c`, and `zsh -c` wrappers with a quoted portable script. Shell expansion
+in an outer double quote, additional shell options, redirects to files, `fish`
+scripts, and nested wrappers pass through unchanged.
+
 ### Setup
 
 ```bash
--- a/docs/contributing/TECHNICAL.md
+++ b/docs/contributing/TECHNICAL.md
@@ -208,6 +208,7 @@ LLM Agent executes rewritten command
 Key design decisions:
 - **Lexer-based tokenization**: A single-pass state machine (`lexer.rs`) handles all shell constructs (quotes, escapes, redirects, operators). Used for both compound splitting and redirect stripping.
 - **Segment-level rewriting**: Compound commands are split by operators, each segment rewritten independently. Bash recombines them at execution time.
+- **Conservative shell-wrapper recursion**: Only exact quoted `sh|bash|zsh -c` wrappers enter one level of inner rewriting. The parser hands back byte ranges instead of a decoded script, so the wrapper's own bytes are never re-quoted.
 - **Pipe semantics**: Producers and intermediate stages of `|` remain raw. Only an argument-safe final stage whose rule has `pipeline_final_safe` may be rewritten; initially this is limited to ordinary `grep` and `rg` invocations. Search pattern-file forms (`-f`/`--file`) defer because they can consume pipeline stdin as configuration. `|&` is recognized separately and its complete pipeline stays raw.
 - **Double env prefix handling**: `classify_command()` strips env prefixes to match the underlying command against rules. `rewrite_segment()` extracts the same prefix separately to re-prepend it to the rewritten command.
 - **Fallback contract**: If any segment fails to match, it stays raw. `rewrite_command()` returns `None` only when zero segments were rewritten.
--- a/src/discover/README.md
+++ b/src/discover/README.md
@@ -23,6 +23,8 @@ When a hook sends `cargo fmt --all && cargo test 2>&1 | tail -20`:
 
 **Compound splitting** — The rewrite engine walks the tokens, splitting on `Operator` (`&&`, `||`, `;`) and typed `Pipe` tokens (`|`, `|&`). For normal pipelines, producers and intermediate stages stay raw, and only an argument-safe final stage marked `pipeline_final_safe` is rewritten. The initial safe set is ordinary `grep` and `rg` invocations; search pattern-file forms (`-f`/`--file`) defer because they can consume pipeline stdin as configuration. Stderr pipelines (`|&`) and pipelines containing opaque shell groups remain raw.
 
+**Quoted shell wrappers** — An exact `sh -c`, `bash -c`, or `zsh -c` segment whose script is one balanced quoted argument has that script rewritten in place, leaving the shell path, quote delimiters, spacing, suffix arguments and redirects byte-identical (`shell_wrapper.rs`). Recursion is capped at one level, so a wrapper inside a wrapper stays raw. A script carrying a newline, heredoc, `$((`, or any construct the lexer cannot attest defers unchanged, as does a `fish` script — fish control keywords (`and`, `or`, `if`) at a command boundary need a fish-aware lexer the shared one does not model.
+
 **Per-segment rewriting** — Each segment goes through:
 
 1. Strip trailing redirects (`2>&1`, `>/dev/null`) — matched via lexer tokens, set aside, re-appended after rewriting
--- a/src/discover/mod.rs
+++ b/src/discover/mod.rs
@@ -5,6 +5,7 @@ pub mod provider;
 pub mod registry;
 mod report;
 pub mod rules;
+pub(crate) mod shell_wrapper;
 
 use anyhow::Result;
 use chrono::{DateTime, Utc};
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -6,10 +6,11 @@ use std::path::Path;
 use std::sync::LazyLock;
 
 use super::lexer::{
-    advance_quote_state, coalesce_words, is_crlf_at, shell_split, split_on_operators, tokenize,
-    tokenize_with_newlines, ParsedToken, PipeKind, TokenKind,
+    advance_quote_state, coalesce_words, contains_unattestable_construct, is_crlf_at, shell_split,
+    split_on_operators, tokenize, tokenize_with_newlines, ParsedToken, PipeKind, TokenKind,
 };
 use super::rules::{RtkRule, IGNORED_EXACT, IGNORED_PREFIXES, RULES};
+use super::shell_wrapper::parse_shell_wrapper;
 
 const PHP_TOOL_NAMES: [&str; 6] = ["phpunit", "phpstan", "ecs", "pest", "paratest", "pint"];
 
@@ -642,7 +643,7 @@ fn rewrite_single(
         return Some(trimmed.to_string());
     }
 
-    rewrite_compound(trimmed, excluded, transparent_prefixes)
+    rewrite_compound(trimmed, excluded, transparent_prefixes, 0)
 }
 
 /// Shell keywords that open or close a multi-line construct. A line inside a
@@ -1036,6 +1037,7 @@ fn rewrite_pipeline_final_stage(
     analysis: PipelineAnalysis,
     excluded: &[ExcludePattern],
     transparent_prefixes: &[String],
+    shell_wrapper_depth: usize,
 ) -> Option<String> {
     let final_stage_start = analysis.final_stage_start?;
     let final_stage = cmd[final_stage_start..analysis.end_offset].trim();
@@ -1046,6 +1048,7 @@ fn rewrite_pipeline_final_stage(
         transparent_prefixes,
         RewriteContext::PipelineFinal,
         0,
+        shell_wrapper_depth,
     )
     .filter(|rewritten| rewritten != final_stage)
     .map(|rewritten| {
@@ -1066,6 +1069,7 @@ fn rewrite_compound(
     cmd: &str,
     excluded: &[ExcludePattern],
     transparent_prefixes: &[String],
+    shell_wrapper_depth: usize,
 ) -> Option<String> {
     let tokens = tokenize(cmd);
     let has_pipe = tokens
@@ -1089,8 +1093,9 @@ fn rewrite_compound(
         match tok.kind {
             TokenKind::Operator => {
                 let seg = cmd[seg_start..tok.offset].trim();
-                let rewritten = rewrite_segment(seg, excluded, transparent_prefixes)
-                    .unwrap_or_else(|| seg.to_string());
+                let rewritten =
+                    rewrite_segment(seg, excluded, transparent_prefixes, shell_wrapper_depth)
+                        .unwrap_or_else(|| seg.to_string());
                 if rewritten != seg {
                     any_changed = true;
                 }
@@ -1120,6 +1125,7 @@ fn rewrite_compound(
                     analysis,
                     excluded,
                     transparent_prefixes,
+                    shell_wrapper_depth,
                 );
 
                 if let Some(rewritten) = rewritten_pipeline {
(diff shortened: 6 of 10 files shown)


SindriFr:
Rebased onto current `develop` (was conflicting, now mergeable). Head is 3317930.

Two notes while rebasing:

**A `sudo` behaviour change on `develop` caught this branch.** #146/#3569 hardened the `sudo` passthrough after my branch point, so `sudo bash -c 'git status'` now returns `None` rather than rewriting — the sudo guard fires before the wrapper path. That is the right call (the inner `rtk` would run as root, where it is off `secure_path`), so I flipped my assertion to pin the passthrough as intentional instead of restoring the old behaviour, and corrected the PR body, which had claimed `sudo` was reached.

**Overlap with #3057.** `shell_wrapper.rs` and most of the tests here are @pashifika's work from that PR, rebuilt against current `develop` — the rewrite engine has since gained `RewriteContext`, `PipeKind`, `analyze_pipeline`, `pipeline_final_safe` and multi-line block handling, so it no longer applies as written. Credit is carried in the commit's `Co-authored-by`. #3057 is still open and untouched since July; if this lands, that one should close with it. Happy to defer entirely if @pashifika would rather refresh theirs.

Gates locally, on Linux x86_64 (NixOS):

```
cargo fmt --all --check                 clean
cargo clippy --all-targets -D warnings  clean
cargo test --all                        3,137 passed, 0 failed, 8 ignored, 12 suites green
```

Only `license/cla` has run on this PR — no build or test check has been triggered. Anything needed from me to get CI to run?

src/hooks/permissions.rs
```diff
@@ -132,6 +137,36 @@ pub(crate) fn check_command_with_rules(
...
 }
 
+fn check_shell_wrapper_permissions(
+    segments: &[&str],
+    deny_rules: &[String],
+) -> Option<PermissionVerdict> {
+    let mut contains_shell_wrapper = false;
+    for segment in segments {
+        let segment = segment.trim();
+        let Some(wrapper) = parse_shell_wrapper(segment) else {
```
KuSh, line 147: **Blocks merge (1/2) — the permission path and the rewrite path disagree about where a wrapper can start.**

`parse_shell_wrapper(segment)` and `is_shell_wrapper_candidate(segment)` both go through `supported_shell_end`, which requires the shell to be the segment's **first word**. `rewrite_segment_inner` has no such requirement: it reaches the wrapper after stripping env assignments, `SHELL_KEYWORD_PREFIXES`, `ROUTABLE_WRAPPER_PREFIXES`, `PROCESS_WRAPPERS` and user `transparent_prefixes`.

So for a prefixed wrapper `contains_shell_wrapper` stays `false`, this function returns `None`, and the ask/allow loop below hands back `Allow` — while the rewriter happily rewrites the script.

With `{"permissions":{"allow":["Bash(*)"],"deny":["Bash(rm:*)"]}}`, `rtk rewrite` exit codes (0=Allow, 1=passthrough, 2=Deny):

```
                                                    develop   head
bash -c 'git status; rm -rf /tmp/x'                     1        2   correct
command bash -c 'git status; rm -rf /tmp/x'             1        0
env     bash -c 'git status; rm -rf /tmp/x'             1        0
exec    bash -c 'git status; rm -rf /tmp/x'             1        0
builtin bash -c 'git status; rm -rf /tmp/x'             1        0
noglob  bash -c 'git status; rm -rf /tmp/x'             1        0
uv run  bash -c 'git status; rm -rf /tmp/x'             1        0
FOO=bar bash -c 'git status; rm -rf /tmp/x'             1        0
direnv exec . bash -c 'git status; rm -rf /tmp/x'       1        0
nice / nohup / timeout 30  bash -c '…'                  1        0   (after rebase, #2375)
```

End to end:

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"command bash -c '\''git status; rm -rf /tmp/x'\''"}}' | rtk hook claude
develop:  (no output — the host evaluates the original command)
head:     {"hookSpecificOutput":{…,"updatedInput":{"command":"command bash -c 'rtk git status; rm -rf /tmp/x'"},
                                 "permissionDecision":"allow"}}
```

`direnv exec . bash -c '…'` is the PR's headline example, and this inverts all three stated invariants.

Please fix at the boundary rather than by enumerating prefixes — `transparent_prefixes` is user-configured and unbounded. Two directions in the review body; I did not send a patch because the obvious one (scan every word start) also makes `echo bash -c 'rm -rf /tmp/x'` return `Deny`.

KuSh (changes requested):
**Claim**: rewrite the inner commands of an exact quoted `sh|bash|zsh -c` wrapper, byte-preserving the wrapper itself, without letting a host allow rule approve a script the host never parsed.
**Scope**: **accept** — the mechanism is right, the subset is honestly narrow, and the byte-span approach is the correct way to do this. #3057 is closed; #3911 is stacked on this branch, so this lands first. (round 1, frozen)
**Ran**: merge-base `5e10aa64`, head `7518b894` and head-merged-onto-`develop` built into three separate `CARGO_TARGET_DIR`s (md5 distinct). `rtk rewrite` and all six `rtk hook <agent>` entry points, differentially, against isolated `HOME` + `XDG_CONFIG_HOME` with real `.claude/settings.json` and `rtk/config.toml`; 611 wrapped cases generated from 321 real commands harvested from the repo's own rewrite tests; raw-vs-rewritten execution of wrapped pipeline stages. No filter output changes, so the 20% floor (CONTRIBUTING.md) does not apply.
**CI** @`7518b894`: only `license/cla`. Run `34109043724` is `action_required` — see "Your questions".

### Blocks merge (2, frozen at round 1)

**1. `src/hooks/permissions.rs:140` — the permission path and the rewrite path disagree about where a wrapper can start, so a prefixed wrapper is rewritten *and* auto-allowed with its inner deny rule skipped.**

`check_shell_wrapper_permissions` calls `parse_shell_wrapper(segment)` / `is_shell_wrapper_candidate(segment)` on the raw segment, and both require the shell to be the segment's **first word**. `rewrite_segment_inner` does not: it reaches the wrapper after stripping env assignments, `SHELL_KEYWORD_PREFIXES`, `ROUTABLE_WRAPPER_PREFIXES`, `PROCESS_WRAPPERS` and user `transparent_prefixes`. So `contains_shell_wrapper` stays `false`, the function returns `None`, and the ask/allow loop hands back `Allow`.

With `~/.claude/settings.json` = `{"permissions":{"allow":["Bash(*)"],"deny":["Bash(rm:*)"]}}`, neutral cwd, `rtk rewrite` (0=Allow, 1=passthrough, 2=Deny, 3=Ask):

```
                                                    base   head
bash -c 'git status; rm -rf /tmp/x'                   1      2   ✓ inner deny fires
command bash -c 'git status; rm -rf /tmp/x'           1      0   ✗
env     bash -c 'git status; rm -rf /tmp/x'           1      0   ✗
exec    bash -c 'git status; rm -rf /tmp/x'           1      0   ✗
builtin bash -c 'git status; rm -rf /tmp/x'           1      0   ✗
noglob  bash -c 'git status; rm -rf /tmp/x'           1      0   ✗
uv run  bash -c 'git status; rm -rf /tmp/x'           1      0   ✗
FOO=bar bash -c 'git status; rm -rf /tmp/x'           1      0   ✗
nice     bash -c 'git status; rm -rf /tmp/x'          1      0   ✗   (post-rebase only, #2375)
nohup    bash -c 'git status; rm -rf /tmp/x'          1      0   ✗   (post-rebase only, #2375)
timeout 30 bash -c 'git status; rm -rf /tmp/x'        1      0   ✗   (post-rebase only, #2375)
direnv exec . bash -c 'git status; rm -rf /tmp/x'     1      0   ✗   (transparent_prefixes)
```

End to end through the Claude hook, same settings:

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"command bash -c '\''git status; rm -rf /tmp/x'\''"}}' | rtk hook claude
develop:  (no output — RTK stays out of it, the host evaluates the original command)
head:     {"hookSpecificOutput":{...,"updatedInput":{"command":"command bash -c 'rtk git status; rm -rf /tmp/x'"},
                                 "permissionDecision":"allow"}}
```

This is a regression against `develop`, which auto-allowed nothing on these inputs, and it inverts the three invariants the PR states: "deny rules are also evaluated against the inner commands", "a wrapper rewrite is never auto-allowed", and "One parser backs both rewriting and permission evaluation, so the two paths agree on the boundary." `direnv exec . bash -c '…'` is the PR's own headline example. Every host that honours `AllowRewrite` is affected.

**Class, not instance**: the covered set is exactly one shape (bare wrapper) out of every prefix family `rewrite_segment_inner` strips — `ROUTABLE_WRAPPER_PREFIXES` (7 entries), `PROCESS_WRAPPERS` (4), the env-assignment prefix, and an unbounded user-configured `transparent_prefixes` list. Patching the four I named would leave the config-driven one open, so please fix it at the boundary rather than by enumeration.

I am **not** sending a patch for this one, because I built the obvious one and it is wrong. Scanning every word start for a wrapper does close all twelve rows above (verified), but it also makes `echo bash -c 'rm -rf /tmp/x'` return `Deny` — a command that only *mentions* a wrapper. Two directions that do not have that failure mode, your call which:

- have the permission check consume the wrapper spans the **rewriter actually used**, rather than re-deriving them (this is what "one parser backs both" would mean literally, and it needs no config in `permissions.rs`); or
- thread `transparent_prefixes` into `check_command_with_rules` and reuse `rewrite_segment_inner`'s prefix-stripping before `parse_shell_wrapper`.

**2. The branch does not build on current `develop`, and one conflict hunk silently reverts #3171 if resolved toward this PR.**

`develop` has moved 304 commits since the 09-07 rebase; `mergeStateStatus` is `DIRTY`. Three conflict hunks in `src/discover/registry.rs` and one in `src/hooks/README.md`. The one that matters is the third: `develop` added the `.or_else(|| rewrite_pipeline_producer(...))` fallback (#3171), and this branch's side of that hunk does not have it — taking "theirs" drops producer-stage rewriting with nothing failing to say so.

I resolved it and ran the gates so the rebase is transcription rather than discovery. Nine items, in order:

1. `registry.rs` import block — take `develop`'s ordering, add `contains_unattestable_construct` and `use super::shell_wrapper::parse_shell_wrapper;`.
2. `rewrite_pipeline_stage` (new on `develop`) — add `shell_wrapper_depth: usize`, thread into `rewrite_segment_inner`.
3. `rewrite_pipeline_final_stage` — `develop` reshaped the signature; pass `shell_wrapper_depth` to `rewrite_pipeline_stage`.
4. `rewrite_pipeline_producer` — add the parameter and **keep** `develop`'s `.or_else(...)` fallback.
5. `strip_process_wrapper_prefix` call site (#2375, ~`registry.rs:1740`) — a new `rewrite_segment_inner` call that needs the sixth argument. This is the one that fails to compile: `error[E0061]: this function takes 6 arguments but 5 arguments were supplied`.
6. `registry.rs:1373` — `if shell_wrapper_depth < MAX_SHELL_WRAPPER_DEPTH { if let Some(...) }` must become a let-chain, or edition 2024 fails clippy: `error: this 'if' statement can be collapsed`.
7. `cargo fmt --all` — two `assert!` blocks in `shell_wrapper.rs` reflow under style edition 2024.
8. Four assertion strings: `rtk read foo` → `rtk read foo --head-lines 10` (`develop` #3668 changed how `head` rewrites). In `registry.rs:7699`, `hook_cmd.rs:2859`, `hook_cmd.rs:2884`, `tests/rewrite_shell_wrapper_test.rs:22`. Without this, 3 unit tests fail.
9. `src/hooks/README.md` — keep `develop`'s `hook_cmd.rs` line ("maps decisions to each agent's JSON protocol, including Codex `updatedInput`"), then add the `discover/shell_wrapper.rs` line.

After those nine: `cargo fmt --all --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --all` **3,843 passed, 0 failed**. The wrapper matrix and `develop`'s producer path (`git log | head -20` → `rtk git log | head -20`) both still behave.

### Optional — will not hold merge; I will resolve these threads myself

- **`exclude_commands` cannot exempt a wrapper.** The wrapper branch sits above every `is_excluded` call in `rewrite_segment_inner`, so with `exclude_commands = ["bash"]`, `["bash -c"]`, `["^bash"]` or `["bash -c .*"]`, `bash -c 'git status'` is still rewritten to `bash -c 'rtk git status'` (control: `["git"]` correctly suppresses it). A user who hits any sharp edge has no escape hatch. `if is_excluded(command, excluded) { return None; }` at the top of `rewrite_shell_wrapper` covers it.
- **`is_shell_wrapper_candidate` stops scanning at the first non-`-` token**, so any shell option taking a *separate* argument hides the `-c` behind it. `bash -lc`, `bash -e -c` and `bash --norc … -c` correctly force `Ask`; `bash -o pipefail -c`, `bash -O extglob -c` and `bash --rcfile F -c` reach `Allow`. Not a regression — `develop` allows all of them — but it is the same claim as blocking item 1, so worth closing in the same pass.
- **`registry.rs:1322`** — the doc comment says ``sh|bash|zsh|fish -c``; `SUPPORTED_SHELLS` deliberately omits fish and a test pins that. Drop `|fish`.
- **The commit message says `Closes #2767`** while the PR body says "it does not close it" (fish is out of scope). The trailer wins when `develop` reaches the default branch. Suggest `Refs #2767`.

- Byte preservation: shell path, quote delimiters, tabs/spacing, suffix arguments and fd redirects all come back identical (`sh\t-c\t'…'\tname`, `/bin/bash -c '…' command-name 'a b'`, `… 2>&1`).
- Inner deny across `;`, `&&`, `||`, `|`, `&`, both quote styles, and `sh` / `zsh` / `/bin/bash` — all 10 bare forms return `Deny`.
- All six agent entry points end to end with real payloads, not test helpers: `claude`, `cursor`, `gemini` (`ask_user` + rewritten `tool_input`), `copilot`, `droid`, `vibe`. (`codex` is not a `rtk hook` subcommand.)
- Recursion cap: `bash -c 'git status && bash -c "cargo test"'` → `bash -c 'rtk git status && bash -c "cargo test"'`; `bash -c 'bash -c "git status"'` → passthrough.
- `sudo`, `sudo -E`, `sudo -u nobody`, `doas` all pass through.
- Your gates on the head: `cargo fmt --all --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --all` 3,243 passed / 0 failed.

- **An inner rewrite emitting a shell metacharacter that changes the wrapper's meaning.** 321 real commands harvested from the repo's rewrite tests → 611 single- and double-quoted wrapped cases → 379 rewritten. Zero introduced `'`, `"`, `$`, `` ` `` or `\`; no `rtk_cmd` in `RULES` contains one. Every inner script also matched the bare rewrite of the same command.
- **UTF-8 span safety.** Every boundary (`script_start`, `script_end`, `supported_shell_end`) lands on an ASCII `'`, `"`, space or tab, which cannot occur inside a multibyte sequence, and `replace_script` uses `.get()` throughout. Your truncation corpus test already covers it.
- **Pipeline-final guards bypassed inside a wrapper.** The bypass is real — `develop` refuses `cat pat.txt | grep -f - data.txt` (exit 1, the `-f`/`--file` deferral in TECHNICAL.md), and this PR rewrites the same thing inside `bash -c` because `rewrite_shell_wrapper` re-enters through `rewrite_compound`, which resets `RewriteContext` to `Normal`. But I could not turn it into a defect: `grep -f -`, `jq .`, `wc -l` and `grep b` as wrapped final stages all produced byte-identical stdout and exit codes raw vs rewritten. Recording it as a lost safety margin, not a finding — if you want to restore it, skipping the wrapper branch when `context == RewriteContext::PipelineFinal` is the one-line version.
- **`Allow` → `Ask` for wrappers a user had explicitly allowlisted.** Real but deliberate and documented, and it only reaches a host when RTK actually rewrote; the hooks emit nothing on a non-rewrite, so nobody gets a new prompt for a wrapper RTK left alone.

### Your questions
- *"Only `license/cla` has run on this PR — no build or test check has been triggered. Anything needed from me to get CI to run?"* → **Nothing from you.** This is a fork PR, so the `CI` workflow parks at `action_required` until a maintainer approves the run — run `34109043724` on `7518b894`, and three earlier ones on older heads, are all sitting in that state. I'll approve the run on the rebased head; approving it on this one would only fail at checkout while the PR conflicts.
- *"#3057 is still open and untouched since July; if this lands, that one should close with it. Happy to defer entirely if @pashifika would rather refresh theirs."* → **No need to defer.** #3057 is already closed, and #3911 is explicitly stacked on this branch, so this is the one that lands first. Keep the `Co-authored-by`.

### Next
Fix blocking item 1 at the boundary (your call which of the two directions), then rebase with the nine steps above — I have those verified green, so they should not cost you a debugging pass. The four Optional items are cheap enough to fold into the same push. Then ping me and I'll approve the CI run.

Rounds: 1/3. Threads: 2 open (blocking items 1 and 2), 0 resolved this round.

SindriFr:
Round 1 addressed. Head is `07e17ba`, rebased onto `develop` (`b748a5f`); `mergeable` is now `MERGEABLE`.

## Blocking 1 — the boundary

Took your first direction: the gate consumes the spans the rewriter used rather than re-deriving them. `permissions.rs` no longer mentions `parse_shell_wrapper`, `is_shell_wrapper_candidate`, or any prefix — `check_shell_wrapper_permissions` is gone.

`rewrite_segment_inner` already probes for a wrapper at every recursion level, which is exactly the set of positions one can start at. It now records what it finds on a `WrapperScan` carried through the walk, `registry::scan_shell_wrappers` replays that walk and returns the scan, and `permissions::check_with_wrapper_scan` judges those scripts. One walk, so there is nothing left to drift — and nothing enumerated, so a `transparent_prefixes` entry is covered by construction rather than by being listed.

Same settings as your matrix (`{"permissions":{"allow":["Bash(*)"],"deny":["Bash(rm:*)"]}}`, `transparent_prefixes = ["direnv exec ."]`), `rtk rewrite` exit codes:

```
                                                    develop   3828(old)   head
bash -c 'git status; rm -rf /tmp/x'                     1          2        2
command bash -c 'git status; rm -rf /tmp/x'             1          0        2
env     bash -c 'git status; rm -rf /tmp/x'             1          0        2
exec    bash -c 'git status; rm -rf /tmp/x'             1          0        2
builtin bash -c 'git status; rm -rf /tmp/x'             1          0        2
noglob  bash -c 'git status; rm -rf /tmp/x'             1          0        2
uv run  bash -c 'git status; rm -rf /tmp/x'             1          0        2
FOO=bar bash -c 'git status; rm -rf /tmp/x'             1          0        2
nice    bash -c 'git status; rm -rf /tmp/x'             1          0        2
nohup   bash -c 'git status; rm -rf /tmp/x'             1          0        2
timeout 30 bash -c 'git status; rm -rf /tmp/x'          1          0        2
direnv exec . bash -c 'git status; rm -rf /tmp/x'       1          0        2

echo bash -c 'rm -rf /tmp/x'                            1          1        1
```

That last row is the reason this shape rather than a word-start scan: `echo` is not a position the rewriter can strip to, so the walk never offers it.

End to end, same settings:

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"command bash -c '\''git status; rm -rf /tmp/x'\''"}}' | rtk hook claude
(no output — RTK stays out of it, as on develop)

$ echo '{"tool_name":"Bash","tool_input":{"command":"direnv exec . bash -c '\''git status; cargo test'\''"}}' | rtk hook claude
{"hookSpecificOutput":{...,"updatedInput":{"command":"direnv exec . bash -c 'rtk git status; rtk cargo test'"}}}
                                                                  ^ no permissionDecision
```

Pinned by `test_prefixed_shell_wrapper_inner_deny_wins` (all eleven prefixes), `test_prefixed_shell_wrapper_never_auto_allowed`, and `test_wrapper_shaped_argument_is_not_a_wrapper`.

One shape note: the scan is `&RefCell<WrapperScan>` bundled with the depth into a `Copy` `WrapperWalk`, because `rewrite_pipeline_final_stage(...).or_else(|| rewrite_pipeline_producer(...))` — your item 2 step 4 — is two branches a `&mut` could not span, and I would rather keep that call site identical to `develop`'s than reshape it. It also keeps both pipeline functions at 7 arguments, under clippy's ceiling.

## Blocking 2 — the rebase

Transcribed your nine steps; all nine were exactly as described, including the `E0061` at the `strip_process_wrapper_prefix` call site and the let-chain at the wrapper probe. `develop` had moved 338 commits by the time I got to it rather than 304, but nothing beyond your list came out of it. `develop`'s `.or_else(|| rewrite_pipeline_producer(...))` is kept — `git log | head -20` → `rtk git log | head -20` still holds, pinned by the existing suite.

Thank you for the recipe; it cost no debugging pass.

## Optional — all four folded in

- **`exclude_commands`** now sits above the wrapper branch. `exclude_commands = ["bash"]` leaves `bash -c 'git status'` alone (exit 1). It is above the scan as well as the rewrite, so an excluded wrapper raises no permission surface either — `test_excluded_shell_wrapper_raises_no_permission_surface`.
- **`is_shell_wrapper_candidate`** now skips the value of an option that takes a separate argument (`-o`, `+o`, `-O`, `+O`, `--rcfile`, `--init-file`), so `bash -o pipefail -c`, `bash -O extglob -c` and `bash --rcfile F -c` are recognised and reach `Ask`. `bash -o pipefail script.sh` is still not a wrapper. Since it is driven by the same walk it also closes behind a prefix: `command bash -lc '…'` asks.
- **`registry.rs` doc comment** — `|fish` dropped.
- **Commit trailer** — now `Refs #2767`.

## Gates

```
cargo fmt --all --check                    clean
cargo clippy --all-targets -- -D warnings  clean
cargo test --all                           3,958 passed, 0 failed, 8 ignored
```

Linux x86_64 (NixOS). Ready for the CI run whenever suits you.

src/hooks/permissions.rs
SindriFr: Fixed at the boundary, first direction. `check_shell_wrapper_permissions` is gone and `permissions.rs` no longer names `parse_shell_wrapper`, `is_shell_wrapper_candidate`, or any prefix — the rewrite walk records each wrapper script it enters on a `WrapperScan`, `registry::scan_shell_wrappers` replays that walk, and `check_with_wrapper_scan` judges exactly those spans. All twelve rows of your matrix now return 2, and `echo bash -c 'rm -rf /tmp/x'` stays at 1 because `echo` is not a position the walk can strip to.

Full matrix, the end-to-end hook output and the pinning tests are in https://github.com/rtk-ai/rtk/pull/3828#issuecomment-5764344323.

PR #3829, opened 2026-09-02
fix(npx): preserve eslint in lint routing arguments

### LINKED ISSUE #3812: rtk npx eslint <bare-path> misroutes: path taken as linter name (npx -- src)
`src/main.rs:2361` routes `npx eslint …` with `lint_cmd::run(&args[1..], …)`, dropping the `eslint` token. But `lint_cmd::run` does its **own** tool detection (`strip_pm_prefix` → `detect_linter`), so it expects the tool name to still be present. The result is a double strip: `detect_linter` then treats the *first user argument* as the linter name.

`detect_linter` classifies an arg as a path/flag only if it starts with `-`, or contains `/` or `.`. A bare directory name like `src` matches none of those, so it is taken as the linter.

## Reproducer

```console
$ rtk -v npx eslint src
Running: src with structured output          # <- "src" treated as the linter

$ rtk npx eslint src
npm error could not determine executable to run
npm error A complete log of this run can be found in: …
exit=1
```

`package_manager_exec("src")` builds `npx --no-install -- src`, so eslint never runs.

Contrast — these all work, because the first arg is recognised as a path/flag and `detect_linter` falls back to `eslint`:

```console
$ rtk -v npx eslint src/     → Running: eslint with structured output
$ rtk -v npx eslint .        → Running: eslint with structured output
$ rtk -v npx eslint --fix    → Running: eslint with structured output
$ rtk -v lint eslint src     → Running: eslint with structured output
```

That last one is the tell: `Commands::Lint` (`src/main.rs:2277`) passes `&args` **whole** and behaves correctly. Only the npx arm strips.

**Trigger condition:** first argument after `eslint` is a bare name with no `/`, no `.` and no leading `-` — `src`, `lib`, `app`, `packages`, … Common enough in real projects to hit regularly.

**Impact:** the failure isn't silent (npm errors out, exit 1 is propagated), but the message names no tool and gives no hint that rtk dropped `eslint`, so it reads as a broken eslint install rather than a routing bug.

## Scope

Checked the other arms in the same `match`:

| Arm | Status |
|---|---|
| `eslint` | **Broken** — `lint_cmd` re-detects the tool, so `&args[1..]` double-strips |
| `tsc` / `typescript` | Fine — `tsc_cmd` hardcodes the tool, args are extra flags |
| `prettier` | Fine — `prettier_cmd` hardcodes the tool |
| `playwright` | Fine — `args[0]` is the playwright subcommand, as expected |
| `next` | Broken differently (hardcoded `build`) — already tracked in #2840 and #2756, not this issue |

`biome` is not routed from `main.rs` at all — it's only reachable via `Commands::Lint`, so it is unaffected despite sharing `lint_cmd`.

Not a duplicate of #1849, #2390, #1080 or #713 — those are about how `npx` itself is resolved or rewritten; this is a routing arm dropping an argument the callee still needs.

## Fix

```diff
-                "eslint" => lint_cmd::run(&args[1..], cli.verbose)?,
+                "eslint" => lint_cmd::run(&args, cli.verbose)?,
```

`lint_cmd::run` already handles a leading tool name (that's the `Commands::Lint` path), so passing the full slice is the correct call. Worth a regression test asserting `rtk npx eslint src` resolves the linter as `eslint`.

## Environment

- rtk 0.46.0, Linux
- Confirmed present on `upstream/develop`

Originally spotted by @HenriqueSchroeder in https://github.com/rtk-ai/rtk/pull/1374#issuecomment-5499803971 while reviewing the analogous slicing in that PR.

Hubujiu:
## Summary

`rtk npx eslint src` drops the `eslint` token before calling `lint_cmd::run`, which then treats `src` as the linter name. Pass the full argument slice so the callee can consume the tool name once.

Adds a cross-platform CLI regression using a local ESLint stub. It checks bare and slash/dot paths, exact forwarded arguments, and the linter's nonzero exit status. This fixes the explicit `npx` dispatch independently of the implicit-path detection in #3567.

Fixes #3812.

## Test plan

- [x] Regression fails before the routing fix and passes after it.
- [x] `cargo test --all`
- [x] `cargo test --test npx_eslint_test`

Validated on Windows with Rust 1.98.0 (MSVC). Unix-only targets are skipped on this platform; Linux/macOS remain for CI.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2358,7 +2358,7 @@ fn run_cli() -> Result<i32> {
             // Intelligent routing: delegate to specialized filters
             match args[0].as_str() {
                 "tsc" | "typescript" => tsc_cmd::run(&args[1..], cli.verbose)?,
-                "eslint" => lint_cmd::run(&args[1..], cli.verbose)?,
+                "eslint" => lint_cmd::run(&args, cli.verbose)?,
                 "prisma" => {
                     // Route to prisma_cmd based on subcommand
                     if args.len() > 1 {
--- a/tests/npx_eslint_test.rs
+++ b/tests/npx_eslint_test.rs
@@ -1,0 +1,46 @@
+use std::fs;
+#[cfg(unix)]
+use std::os::unix::fs::PermissionsExt;
+use std::process::Command;
+
+#[test]
+fn npx_eslint_preserves_linter_and_bare_path_arguments() {
+    let dir = tempfile::tempdir().expect("tempdir");
+    #[cfg(unix)]
+    let (filename, script) = (
+        "eslint",
+        "#!/bin/sh\nprintf '%s\\n' \"$@\" > \"$ARGS_FILE\"\nprintf '[]\\n'\nexit 1\n",
+    );
+    #[cfg(windows)]
+    let (filename, script) = (
+        "eslint.cmd",
+        "@echo off\r\n(for %%a in (%*) do @echo %%~a) > \"%ARGS_FILE%\"\r\necho []\r\nexit /b 1\r\n",
+    );
+    let eslint = dir.path().join(filename);
+    fs::write(&eslint, script).expect("write eslint fixture");
+    #[cfg(unix)]
+    fs::set_permissions(&eslint, fs::Permissions::from_mode(0o755)).expect("chmod eslint");
+
+    for path in ["src", "lib", "src/", "."] {
+        let args_file = dir.path().join("args");
+        let output = Command::new(env!("CARGO_BIN_EXE_rtk"))
+            .current_dir(dir.path())
+            .env("PATH", dir.path())
+            .env("ARGS_FILE", &args_file)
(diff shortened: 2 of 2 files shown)


KuSh:
Thanks for the fix and for the careful repro. The npx routing arm was corrected on develop by #988 (merged 2026-09-03), in commit c7c1d966: it now passes the full argument slice to lint_cmd, so `rtk npx eslint src` no longer takes the path as the linter name. That same merge gave lint_cmd::run a runner argument and made it run the tool through the runner the user named, so the regression test here, which puts an eslint stub on PATH, would no longer see that stub after a rebase; develop still has no CLI-level test pinning `rtk npx eslint <bare-path>`, and a test written against the current routing would be welcome as its own PR. The second bug raised in the thread, detect_linter treating a bare directory as the linter name on `rtk lint src`, is still present and deserves its own issue. If any part of the npx misrouting still reproduces on the latest develop, please comment or reopen.

Issue #1993, opened 2026-05-20
OpenCode plugin (tool.execute.before) never fires on Windows — auto-rewrite broken


JettHayes:
## Title
OpenCode plugin (tool.execute.before) never fires on Windows — auto-rewrite broken

## Environment
- RTK: latest (binary at D:\rtk_tools\rtk.exe)
- OpenCode: latest
- OS: Windows 11
- Shell: CMD / PowerShell

## Symptom
After `rtk init -g --opencode`, the plugin file is created at `C:\Users\<user>\.config\opencode\plugins\rtk.ts`. The plugin loads successfully, but bash commands are never auto-rewritten. Running `git status` produces raw output instead of `rtk git status`.

## Root cause identified
The original `rtk.ts` uses `which rtk` to detect the binary, which does not exist on Windows. This causes the plugin to immediately return `{}` and disable itself. Replaced with `where.exe rtk` as primary check with `which rtk` as fallback — binary is confirmed in PATH — but the `tool.execute.before` hook still never fires.

## Steps to reproduce
1. Install RTK on Windows, place `rtk.exe` in a directory on PATH
2. Run `rtk init -g --opencode`
3. Open OpenCode and run any git or shell command
4. Observe: commands are never rewritten to `rtk` equivalents

## Additional context
- `where.exe rtk` returns `D:\rtk_tools\rtk.exe` correctly inside OpenCode's shell
- `rtk rewrite "git status"` correctly returns `rtk git status` when run manually
- The hook structure appears identical to the master branch `hooks/opencode/rtk.ts`
- Issue #1706 (macOS) was resolved by killing stale processes — no stale processes exist on this Windows setup, issue persists after clean restart

## Question
Is `tool.execute.before` hook dispatch broken specifically on Windows, or is there a Windows-specific plugin format required?

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 83% |

### Summary

OpenCode `tool.execute.before` hook dispatch does not fire on Windows, completely breaking auto-rewrite; investigate Windows-specific plugin execution path in hooks/opencode/rtk.ts

<details>
<summary>📁 Relevant files</summary>

[...]

PR #3837, opened 2026-09-02
fix(hooks): replace zx shell with Node child_process in OpenCode plugin

### LINKED ISSUE #3326: `rtk init --opencode` plugin silently does nothing on Desktop (`$ is not a function`) and on Windows (`which` not found)

amnesiaof:
## Summary

- replace the Bun-only `$` shell helper and Unix-only `which` probe with Node `execFile` in the OpenCode plugin
- pass rewrite commands as literal subprocess arguments, no shell involved
- keep successful rewrite stdout even when `rtk rewrite` exits non-zero (exit 3 = permission-gated rewrite)
- update the three doc references to the zx library
- add a regression test for the embedded plugin template

## Why

The current plugin fails silently in two environments:

1. **OpenCode Desktop / Node runtime:** `$` is `undefined` in the plugin context, so `` $`which rtk` `` throws `TypeError: $ is not a function`, the plugin factory crashes, and no hooks are registered — nothing is surfaced to the user.
2. **Windows CLI:** `which` does not exist.

A third subtlety: `rtk rewrite` uses non-zero exit codes as status signals (exit 3 with valid stdout such as `rtk git status`), so stdout must be resolved regardless of the exit code.

## Test plan

- [x] `cargo test` — `test_opencode_plugin*` pass (3/3). Note: 18 pre-existing failures in `core::stream::tests` on Windows ("program not found" — tests spawn Unix utils like `cat`/`echo`), present on a clean `develop` tree as well, unrelated to this change.
- [x] Functional test of the plugin under Bun (`@opencode-ai/plugin` contract): `"git status"` → `"rtk git status"`, PowerShell commands pass through unchanged, already-prefixed `rtk ...` commands are not double-prefixed
- [x] Live verification on Windows 11 + OpenCode CLI: hook fires, `rtk gain` tracks savings

Closes #3326
Related: #1993, #3330 (independent fix, same root causes; this PR targets current `develop`)

### DIFF
--- a/docs/contributing/TECHNICAL.md
+++ b/docs/contributing/TECHNICAL.md
@@ -319,7 +319,7 @@ Start here, then drill down into each README for file-level details.
 | [`cline/`](../hooks/cline/README.md) | Cline / Roo Code | Rules file (prompt-level, no programmatic hook) |
 | [`windsurf/`](../hooks/windsurf/README.md) | Windsurf / Cascade | Rules file (workspace-scoped) |
 | [`codex/`](../hooks/codex/README.md) | OpenAI Codex CLI | Awareness document, AGENTS.md integration |
-| [`opencode/`](../hooks/opencode/README.md) | OpenCode | TypeScript plugin, zx library, in-place mutation |
+| [`opencode/`](../hooks/opencode/README.md) | OpenCode | TypeScript plugin, Node `execFile`, in-place mutation |
 
 ---
 
--- a/hooks/README.md
+++ b/hooks/README.md
@@ -187,11 +187,18 @@ Returns `{}` when no rewrite (Cursor requires JSON for all paths).
 
 ### OpenCode (TypeScript Plugin)
 
-Mutates `args.command` in-place via the zx library:
+Mutates `args.command` in-place via Node's cross-platform `execFile` API:
 
 ```typescript
-const result = await $`rtk rewrite ${command}`.quiet().nothrow()
-const rewritten = String(result.stdout).trim()
+function runRtk(args: string[]): Promise<string> {
+  return new Promise((resolve) => {
+    execFile("rtk", args, { windowsHide: true }, (_error, stdout) => {
+      resolve(String(stdout ?? "").trim())
+    })
+  })
+}
+
+const rewritten = await runRtk(["rewrite", command])
 if (rewritten && rewritten !== command) {
   (args as Record<string, unknown>).command = rewritten
 }
--- a/hooks/opencode/README.md
+++ b/hooks/opencode/README.md
@@ -4,8 +4,8 @@
 
 ## Specifics
 
-- TypeScript plugin using the zx library (not a shell hook)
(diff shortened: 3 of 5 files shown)


KuSh:
Closing in favour of #3330, which makes the same execFile change to the OpenCode plugin and has a green CI run across the Linux, macOS and Windows test matrix. #3330 still needs a rebase onto current develop, which is the work done on this branch, so a note there pointing at it would be useful. Review and comments are welcome on #3330.

Issue #3558, opened 2026-08-13
Hook rewrites piped commands: compressed output feeds downstream parsers, silently wrong results


sdewell:
rtk 0.43.0, macOS, Claude Code PreToolUse hook (`rtk hook claude`).

The hook rewrites a command even when its stdout is piped into another program:

```
$ echo '{"tool_name":"Bash","tool_input":{"command":"git diff main..HEAD | grep -c \"^+\""}}' | rtk hook claude
→ updatedInput.command = rtk git diff main..HEAD | grep -c "^+"
```

The downstream program then parses compressed output instead of the real format. Reproduced on a real repo (same two-commit range):

```
$ rtk git diff HEAD~2..HEAD | grep -c '^+'
0
$ /usr/bin/git diff HEAD~2..HEAD | grep -c '^+'
50
```

grep counting zero added lines is a silent wrong answer — in an agent session it reads as "no changes", and we traced a real misdiagnosis back to exactly this.

The 0.43 never-worse guard protects the *reading* consumer, but a piped consumer parses structure (`^+` prefixes, column layouts) that compression removes. Compression is for the model's eyes; anything behind a pipe needs raw bytes.

Suggested fix, either layer works:
- rtk passthrough when `!isatty(stdout)` (same convention as git's pager/color handling), or
- the hook engine skips rewriting a command whose proxied stage is followed by a pipe.

Related earlier reports from this setup: #3025, #3026.

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟠 **Priority** | `high` |
| 🎯 **Confidence** | 90% |

### Summary

rtk hook claude rewrites piped/chained commands to a compressed format, causing downstream parsers (e.g. grep) to consume the wrong data and silently produce incorrect results; recommend passthrough when stdout isn't a tty or skipping rewrite when the proxied command is piped.

<details>
<summary>📁 Relevant files</summary>

- `src/hook/claude.rs`
- `src/hook/rewrite.rs`

</details>

---
*Triaged automatically by [wshm](https://github.com/wshm-dev/wshm)* · This is an automated analysis, not a human review.

vishae:
**Variant: the rewritten command is `grep` itself, and stdout is empty rather than compressed**

[...]

PR #2693, opened 2026-06-28
feat: hook bypass: `$(...)` subshell, backticks, and `xargs <cmd>` skip

wilmerdooley:
Resolves #2145.

The `rtk hook claude` PreToolUse rewriter correctly handles top-level commands and common compound forms (`|`, `;`, `&&`) but bypasses three runtime-resolved patterns. Result: commands that should be rewritten ship as bare invocations, leaking token savings.

## Summary

-

## Test plan

- [ ] `cargo fmt --all && cargo clippy --all-targets && cargo test`
- [ ] Manual testing: `rtk <command>` output inspected

> **Important:** All PRs must target the `develop` branch (not `master`).
> See [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) for details.

PR #3869, opened 2026-09-04
fix(hook): keep commands inside command substitutions untouched

### LINKED ISSUE #1109: Claude Code hook: `git` inside `$(...)` subshells inserts rtk as a git subcommand instead of prefixing
## Summary

The Claude Code Bash tool hook that rewrites `git <subcommand>` to `rtk git <subcommand>` mishandles `git` invocations that appear inside a `$(...)` command substitution. Instead of producing the expected `rtk git <subcommand>` prefix, the rewrite inserts `rtk` AFTER `git`, producing `git rtk <subcommand>` — which git then interprets as an unknown subcommand named `rtk`.

## Environment

- Claude Code hook: the same one as in #1108 — the transparent rewriter for `git`, `tail`, `head`, etc.

## Reproduction

Inside a Claude Code session with the rtk hook active, run a command with `git` inside `$(...)`:

```bash
echo "Branch: $(git rev-parse --abbrev-ref HEAD)"
BRANCH=$(git branch --show-current)
for f in $(git ls-files); do echo "$f"; done
cd "$(git rev-parse --show-toplevel)"
```

Observed: the hook rewrites the inner command incorrectly. Instead of `$(rtk git rev-parse ...)`, it produces something like `$(git rtk rev-parse ...)`, which git interprets as an attempt to invoke a git subcommand named `rtk`. The shell then runs `git rtk rev-parse --abbrev-ref HEAD` and git errors with:

```
git: 'rtk' is not a git command. See 'git --help'.
```

Expected: the hook should rewrite `git <subcommand>` to `rtk git <subcommand>` regardless of surrounding syntax (command substitution, pipes, `&&`, backgrounded with `&`, etc.). The prefix position is always "immediately before the word `git`," never "after the word `git`."

## Root cause (speculative)

The hook's rewrite logic appears to use a regex substitution like `s/\bgit\b/rtk git/g` — but it mishandles the word boundary around `$(`. Inside `$(git ...)`, the regex engine sees `(git ` and the `(` is a word-boundary character, so the match succeeds, but the substitution places `rtk` in the wrong position relative to the capture group. Or the substitution is positional rather than regex and it's counting tokens from the start of the line, missing the fact that the `git` inside `$(...)` is not at position 0.

## Suggested fix

Rewrite logic should tokenize the command via a shell-aware parser (or at minimum a `$(...)` / `\`...\`` / pipeline-aware pre-pass) and apply the `rtk` prefix on a per-invocation basis at the correct position, rather than relying on raw regex substitution against the full command string. Handle these cases:

- `$(git ...)` → `$(rtk git ...)`
- `` `git ...` `` → `` `rtk git ...` ``
- `cmd | git ...` → `cmd | rtk git ...`
- `git ... && other` → `rtk git ... && other`
- `git ...; git ...` → `rtk git ...; rtk git ...`

A lightweight approach: split on shell metacharacters that introduce a new command context (`|`, `&&`, `||`, `;`, `$(`, `` ` ``, `(`, `{`) and apply the rewrite to each sub-command independently.

## Impact

Medium severity. `$(...)` command substitution with `git` inside is common enough in shell scripts and ad-hoc CLI use that this surfaces frequently. When it triggers, the user sees `git: 'rtk' is not a git command` and has to manually `rtk proxy git ...` or disable the hook for that command. It is also a latent failure — the user only discovers it at execution time, not at rewrite time, so it looks like "git is broken" rather than "the hook rewrite is wrong," which delays diagnosis.

See also #1108 for a related hook-rewrite bug in `tail -N`.

jsiu93:
## Problem
The hook rewrites commands that appear inside `$(...)` or backtick substitutions. For example:

    curl -X POST "$URL" --data-binary "$(jq -cn --arg p "$P" '{payload:$p}')"

becomes `"$(rtk jq -cn ...)"`. The inner command's output is data consumed by the outer command, so the compacted output plus the tee pointer corrupt it. Same root cause as #1109 (`git` inside `$(...)`).

## Fix
Before the existing top-level rewrite, a single pass masks complete `$(...)` and paired backtick substitutions (nesting, escapes, quotes and comments respected; single-quoted text stays literal). The top-level decision runs on the masked command, then the original slices are restored byte-for-byte. No regex, no new dependency, no change to heredoc, pipeline or xargs handling.

## Tests
8 unit tests in `registry.rs`: outer rewrite with inner `$(...)` kept, unsupported outer command returns None, backtick variant, single-quoted literal, nested substitutions, operators and `)` inside a substitution comment. `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check` all pass.

Fixes #1109. Refs #3558, #1282. Supersedes the subshell part of #2693, which has conflicts.

### DIFF
--- a/src/discover/README.md
+++ b/src/discover/README.md
@@ -35,6 +35,7 @@ When a hook sends `cargo fmt --all && cargo test 2>&1 | tail -20`:
 - `gh` with `--json`/`--jq`/`--template` → skip (structured output, rtk would corrupt it)
 - `cat` with flags other than `-n` → skip (different semantics than `rtk read`)
 - `cat`/`head`/`tail` with `>` or `>>` → skip (write operation, not a read)
+- Commands inside `$(...)` and paired backtick substitutions stay byte-for-byte unchanged; only the surrounding top-level command is eligible for rewriting
 - Command in `hooks.exclude_commands` config → skip
 
 **Result**: `rtk cargo fmt --all && rtk cargo test 2>&1 | tail -20`. Bash handles the `&&` and `|` at execution time — each `rtk` invocation is a separate process.
--- a/src/discover/registry.rs
+++ b/src/discover/registry.rs
@@ -561,6 +561,177 @@ fn collapse_line_continuations(s: &str) -> std::borrow::Cow<'_, str> {
     LINE_CONTINUATION_RE.replace_all(s, " ")
 }
 
+/// A command with complete shell command substitutions replaced by unique,
+/// inert markers. The original slices are borrowed and restored after the
+/// surrounding command has gone through the normal rewrite pipeline.
+struct ProtectedSubstitutions<'a> {
+    command: String,
+    originals: Vec<&'a str>,
+    marker_prefix: String,
+}
+
+impl ProtectedSubstitutions<'_> {
+    fn restore(self, mut rewritten: String) -> String {
+        for (index, original) in self.originals.into_iter().enumerate() {
+            let marker = format!("{}{index}__", self.marker_prefix);
+            if let Some(start) = rewritten.find(&marker) {
+                rewritten.replace_range(start..start + marker.len(), original);
+            }
+        }
+        rewritten
+    }
+}
+
+#[derive(Default)]
+struct SubstitutionFrame {
+    in_single: bool,
+    in_double: bool,
+    grouping_depth: usize,
+}
+
+/// Return the exclusive end of a balanced `$(...)` substitution.
+///
+/// Each nested substitution owns its quote state, while ordinary parentheses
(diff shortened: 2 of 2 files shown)


pszymkowiak:
Closing: already fixed on develop. Since #2286 (v0.42.3) the hook defers any command containing an unquoted `$(...)`, backtick or process substitution before the registry rewrite runs, and #3343 (v0.49.0) handles the piped-producer case from #3558. Verified with `rtk rewrite` and `rtk hook claude` on the #1109, #3558 and this PR's own test commands: every substitution-bearing line is left untouched. Rewriting the outer command around a masked substitution, which this PR adds, would be unreachable behind that defer; whether nested commands should ever be rewritten is tracked in #1252. Thanks for the PR.

PR #3872, opened 2026-09-04
fix(go): propagate golangci-lint exit code instead of swallowing 1

### LINKED ISSUE #2735: bug: rtk golangci-lint run returns exit 0 when lint issues are found (exit code masked)

v01dst:
## Summary

`rtk golangci-lint run` printed the (correctly filtered) findings but **exited 0 even when golangci-lint found issues**. The filtered path rewrote exit 1 → 0, so pre-commit hooks, Makefile `&&` chains, and "did lint pass?" checks saw success for a failing tree (#2735). The same mapping existed in the `rtk go tool golangci-lint` path.

## Fix

Pass the tool's exit code through unchanged in both wrapper paths:

- `src/cmds/go/golangci_cmd.rs` (`run_filtered`)
- `src/cmds/go/go_cmd.rs` (`run_go_tool_golangci_lint`)

Exit semantics now reach callers as golangci-lint defines them: `0` = clean, `1` = issues found, `2+` = config/build error. Filtered output is unchanged.

## Test

New `tests/golangci_exit_code_test.rs` (integration, follows the existing availability-gated pattern):

- real `golangci-lint run` on a tree with an `ineffassign` finding → **exit 1** (was 0), findings still visible in filtered output
- clean tree → **exit 0**
- skipped when `golangci-lint` or `go` is not installed

Verified locally against golangci-lint 2.6.2 with the exact repro from the issue:

```
direct binary → exit=1
rtk (fixed)   → exit=1   (was 0)
```

All 50 existing `--bin rtk golangci` unit tests pass unchanged.

Fixes #2735

### DIFF
--- a/src/cmds/go/go_cmd.rs
+++ b/src/cmds/go/go_cmd.rs
@@ -286,9 +286,10 @@ fn run_go_tool_golangci_lint(args: &[OsString], verbose: u8) -> Result<i32> {
         shown,
     );
 
-    // golangci-lint: exit 0 = clean, exit 1 = lint issues found (not an error),
-    // exit 2+ = config/build error, None = killed by signal (OOM, SIGKILL)
-    Ok(if exit_code == 1 { 0 } else { exit_code })
+    // golangci-lint exit codes must reach callers unchanged: 0 = clean,
+    // 1 = lint issues found (pre-commit hooks, Makefiles and agents gate on
+    // this), 2+ = config/build error.
+    Ok(exit_code)
 }
 
 /// Parse go test -json output (NDJSON format)
--- a/src/cmds/go/golangci_cmd.rs
+++ b/src/cmds/go/golangci_cmd.rs
@@ -160,9 +160,10 @@ fn run_filtered(original_args: &[String], invocation: &RunInvocation, verbose: u
         crate::core::runner::RunOptions::stdout_only(),
     )?;
 
-    // golangci-lint: exit 0 = clean, exit 1 = lint issues found (not an error),
-    // exit 2+ = config/build error, None = killed by signal (OOM, SIGKILL)
-    Ok(if exit_code == 1 { 0 } else { exit_code })
+    // golangci-lint exit codes must reach callers unchanged: 0 = clean,
+    // 1 = lint issues found (pre-commit hooks, Makefiles and agents gate on
+    // this), 2+ = config/build error.
+    Ok(exit_code)
 }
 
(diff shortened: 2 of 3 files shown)


v01dst:
Closing this. I was batch-generating PRs across repos while focused on my own project — that wasn't fair on reviewers' time, and I've stopped. Apologies for the noise.

PR #3879, opened 2026-09-05
fix(hooks): leave git unrewritten in Claude Code managed worktrees

### LINKED ISSUE #3864: rtk hook claude refuses every git command in a worktree, always the same isolation error
## Summary

Every git command run through the documented `rtk hook claude` PreToolUse hook gets refused with the exact same error text, no matter what the command is. This isn't a filter or rewrite issue: the hook refuses the command outright, before running anything.

## Environment

- rtk 0.47.0 (Homebrew, `/opt/homebrew/bin/rtk`) — same result on 0.45.0 before I updated
- macOS 26, arm64
- Claude Code, `rtk hook claude` configured as the PreToolUse Bash hook (per the docs)
- Working directory is a git worktree, nested under `.claude/worktrees/<name>` inside the repo

## What happens

All of these, run from the worktree directory, return the identical error:

```
git status
git status --short
git -C /abs/path/to/worktree status
git log -1 --oneline
rtk git status --short
rtk git status
rtk proxy git status --short
```

Error text (verbatim):

```
This session is isolated in the worktree <path>, but this command runs rtk with a git command among its operands: what runs it, and from which directory or root, cannot be read here (name git right after the launcher and its options) in a plain command, so what it runs cannot be shown not to be git. Refusing to run it — a worktree-isolated session's git operations must target its own worktree. Run the plain command from <path>.
```

`<path>` in the message matches the actual cwd. Every variant above still gets refused, including `rtk proxy`, which the docs describe as the raw/unfiltered path.

## What I ruled out

- Command syntax: all 7 variants above give the identical error.
- Stale session state: reproduced from a brand-new Claude Code agent started in the same directory.
- General rtk breakage: `rtk --version` and `rtk gain` work fine. `rtk gain` shows this same install has 435 prior successful `rtk git status` calls and 80 `rtk git commit` calls from other projects.
- Version: same error on both 0.45.0 and 0.47.0.

## Expected

git commands should run normally in a worktree, same as the history in `rtk gain` shows happening elsewhere on this machine.

## Guess

Possibly the worktree-isolation check is tripped by the nested path shape (`<repo>/.claude/worktrees/<name>`) rather than a plain sibling worktree directory. I haven't looked at the source, so treat this as a guess, not a diagnosis.

b3hz4d:
## Summary

- `rtk hook claude` refused every git command in a worktree-isolated Claude Code session. The refusal is Claude Code's own worktree-isolation guard: it evaluates the command *after* the hook has rewritten it, and it can only verify git when spelled plainly, so every `rtk git …` was rejected as "cannot be shown not to be git".
- The PreToolUse payload carries no isolation flag, but Claude Code's managed worktrees always live at `<repo>/.claude/worktrees/<name>`. When the payload `cwd` has that shape, the Claude hook now adds `git` to the exclude list so it passes through untouched while the rest of the chain is still rewritten (`git status && cargo build` → `git status && rtk cargo build`).
- Other hosts (Gemini, Cursor, Copilot, Droid, Vibe) pass no cwd and are unchanged.

Fixes #3864

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` (clippy clean, 3110 unit tests + integration suites green)
- [x] New unit tests: path-shape detection, git excluded / others rewritten inside a worktree, payload-level skip
- [x] Manual: piped real PreToolUse payloads into `target/debug/rtk hook claude`

| Command | cwd under `.claude/worktrees/` | cwd elsewhere |
|---|---|---|
| `git status` | passthrough | `rtk git status` |
| `git status && cargo build` | `git status && rtk cargo build` | both rewritten |
| `cargo test` | `rtk cargo test` | `rtk cargo test` |

Not covered: `rtk git …` / `rtk proxy git …` typed directly by the model is still refused by Claude Code's guard (the hook only leaves already-rtk commands alone); the model's plain `git` retry now succeeds, which breaks the loop from the issue.


Andre1Becker:
+1, run in the same issue

Ilya1vanov:
+1

pixelari:
**The `cwd` path-shape test will miss a large class of worktrees.**

`<repo>/.claude/worktrees/<name>` only covers worktrees the harness provisions itself. A worktree created with plain `git worktree add` at an arbitrary path gets isolated too, and hits exactly the refusal from #3864.

Reproduced just now, entering a hand-provisioned worktree that lives outside the repo entirely (`~/workspace/<project>-worktrees/<branch>`; `.claude/worktrees/` is empty):

| command | result |
|---|---|
| `git status --porcelain` | refused |
| `rtk git status --porcelain` | refused |
| `/usr/bin/git status --porcelain` | runs |

The refusal text names rtk directly - *"this command runs rtk with a git command among its operands"* - so the rewrite is landing before the guard reads it, exactly as you describe. But with this PR as drafted the `cwd` check would never fire for that worktree, and the loop stays.

A more robust test than the path shape: ask git. In a linked worktree `--git-dir` and `--git-common-dir` differ; in the main checkout they are identical. From that same worktree:

```
$ git rev-parse --git-dir --git-common-dir
/repo/.git/worktrees/<name>
/repo/.git
```

Equal means main checkout, differing means linked worktree. Canonicalize before comparing, since both resolve relative to `cwd`. That covers harness-managed and hand-provisioned worktrees alike, and does not depend on a path convention that can change.

If you would rather not shell out to git in the hook's hot path: in a linked worktree `.git` is a *file* containing `gitdir: ...`, not a directory. Same signal, no subprocess - verified in the same worktree above.

PR #3883, opened 2026-09-05
fix(grep): recognize -D/--devices as a value-taking flag

### LINKED ISSUE #3881: rtk grep -D <action> / --devices <action> shifts the pattern and path, producing exit 2 instead of a normal search
GNU grep's `-D ACTION` / `--devices ACTION` (read|skip, how to handle device/FIFO/socket files) takes a value, exactly like `-d`/`--directories`. RTK recognizes lowercase `-d` as value-taking (`VALUE_FLAGS_SHORT = b"ABCMTdfgjmt"` in `search.rs`) but not uppercase `-D`, and recognizes neither the short nor the space-separated long form for `--devices` (`--devices` is absent from `VALUE_FLAGS_LONG`). The action word is left as an ordinary positional and consumes the slot that should hold the search path, while the real pattern gets bumped into the "path" position.

# Reproduction

```
printf 'apple pie\napple\nzebra apple juice\n' > f1.txt

grep -D skip apple f1.txt
# apple pie
# apple
# zebra apple juice
# exit 0

rtk grep -D skip apple f1.txt
# (empty output)
# exit 2

grep --devices skip apple f1.txt
# same 3 lines, exit 0

rtk grep --devices skip apple f1.txt
# (empty output)
# exit 2
```

`rtk -v grep -D skip apple f1.txt` shows the misparse directly: `-D` is treated as boolean, so skip becomes the pattern, apple becomes the (nonexistent) path, and grep exits 2 with "No such file or directory."

Both the short (`-D skip`) and long-space (`--devices skip`) forms fail identically; the inline `--devices=skip` form (single token) is unaffected.

# Root cause

extract_pattern_path() in src/cmds/system/search.rs:
- VALUE_FLAGS_SHORT (line 27) is `b"ABCMTdfgjmt"` — lowercase d only, no D.
- VALUE_FLAGS_LONG (lines 33–63) has no `--devices` entry.

So `-D`/`--devices` fall through as ordinary boolean/unknown flags instead of consuming their value token, and the next real argument (the action word) is misread as a positional.

# Expected behavior

`-D`/`--devices`/`--devices=` should consume exactly one following value token (like `-d`/`--directories` already does), leaving the actual pattern and path positionals untouched, and RTK's output should match plain grep/rg byte-for-byte.

# Impact

Any agent or script that passes `-D skip` (a standard way to silence device-file warnings during a recursive-ish search) gets a spurious exit code 2 and no output instead of real results — indistinguishable from "the target file doesn't exist" unless the user manually re-runs plain grep.

# Environment
- rtk develop @ e53ec1c (Cargo.toml 0.42.4); confirmed against installed rtk 0.48.0, whose grep arg-parsing matches this code
- GNU grep 3.12, Linux

gene1wood:
## Summary

GNU grep's `-D ACTION` / `--devices ACTION` (`read`|`skip`, how to handle
device/FIFO/socket files) takes a value, exactly like `-d`/`--directories`.
RTK's `VALUE_FLAGS_SHORT` recognized lowercase `-d` but not uppercase `-D`,
and `VALUE_FLAGS_LONG` had no `--devices` entry at all. As a result the
action word (`skip`) was left as an ordinary positional, which shifted the
real search pattern and path out of position — the shifted "path" usually
doesn't exist, so grep exited 2 with no output instead of searching.

```
grep     -D skip apple f1.txt   # 3 matches, exit 0
rtk grep -D skip apple f1.txt   # (empty), exit 2 — before this fix
```

## Fix

- Add `D` to `VALUE_FLAGS_SHORT` in `src/cmds/system/search.rs`, alongside
  its lowercase `-d` twin.
- Add `--devices` to `VALUE_FLAGS_LONG` so the space-separated long form
  (`--devices skip`) also consumes its value correctly. (`--devices=skip`,
  a single token, already worked.)

## Test plan

- Added unit tests in `search.rs` covering `-D skip`, inline `-Dskip`,
  `--devices skip`, and `--devices=skip`.
- Added an end-to-end byte-for-byte test in
  `tests/grep_faithful_format_test.rs` (`dash_capital_d_devices_matches_grep_n`)
  that runs the compiled `rtk` binary against real `grep -D`/`--devices` and
  diffs stdout + exit code, with and without `-n` — the same harness shape
  that would have caught this as a regression.

Fixes #3881

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -22,9 +22,12 @@ use std::sync::LazyLock;
 /// Short single-char flags that consume one following token (or inline remainder)
 /// as their value. `-e` is handled separately — its value goes to `patterns`.
 /// Includes all rg short flags that take a value argument except `-e` and `-r`
-/// (stripped) and `-E` (dialect, left to #2138). Failure mode for a missing
-/// entry: the value becomes a positional (visible wrong result, not silent).
-const VALUE_FLAGS_SHORT: &[u8] = b"ABCMTdfgjmt";
+/// (stripped) and `-E` (dialect, left to #2138). Also includes grep's own
+/// `-D` (`--devices=ACTION`, #3881) alongside its lowercase `-d`
+/// (`--directories`/rg `--max-depth`) twin — `D` isn't an rg flag. Failure
+/// mode for a missing entry: the value becomes a positional (visible wrong
+/// result, not silent).
+const VALUE_FLAGS_SHORT: &[u8] = b"ABCDMTdfgjmt";
 
 /// Long flags that consume the NEXT token as their value (space-separated form).
 /// Inline `=` form (`--flag=value`) is one token and passes through unchanged.
@@ -37,6 +40,7 @@ const VALUE_FLAGS_LONG: &[&str] = &[
     "--colors",
     "--context",
     "--context-separator",
+    "--devices",
     "--encoding",
     "--engine",
     "--field-context-separator",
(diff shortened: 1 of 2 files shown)


gene1wood:
Closing in favor of #3681 which should fix this.

PR #3884, opened 2026-09-05
fix(grep): make -T engine-aware (grep's boolean vs rg's value-taking)

### LINKED ISSUE #3882: rtk grep -T <pattern> misparses GNU grep's boolean --initial-tab as ripgrep's value-taking --type-not
`-T` means two different things to the two engines RTK proxies: GNU grep's `-T`/`--initial-tab` is a boolean flag (pad the line prefix with a tab), while ripgrep's `-T` is `--type-not` and takes a value. RTK's `VALUE_FLAGS_SHORT` (`b"ABCMTdfgjmt"`, search.rs) includes `T` for ripgrep's sake, so when the underlying engine is actually GNU grep, RTK wrongly consumes the very next token — the search pattern itself — as `-T`'s "value," shifting every remaining argument out of place.

# Reproduction

```
printf 'apple pie\napple\nzebra apple juice\n' > f1.txt
printf 'applesauce\napple\n' > f2.txt

grep -T apple f1.txt f2.txt
# f1.txt:     apple pie
# f1.txt:     apple
# f1.txt:     zebra apple juice
# f2.txt:     applesauce
# f2.txt:     apple
# exit 0

rtk grep -T apple f1.txt f2.txt
# (empty output)
# exit 2
```

`apple` is consumed as `-T`'s inline/next-token value instead of being forwarded as the pattern, so the pattern and path positionals shift by one, the shifted "path" doesn't exist, and grep exits 2.

Note this only reproduces when `-T` is immediately followed by the pattern with no other flag in between; e.g. `-T -n apple ...` happens to work by coincidence, because both `-T` and `-n` end up forwarded as separate tokens regardless of how they're parsed internally — which is why this is easy to miss in ad hoc testing.

# Root cause

`parse_cluster()/extract_pattern_path()` in `src/cmds/system/search.rs` treat `T` as universally value-taking (borrowed from ripgrep's `--type-not`) via the shared `VALUE_FLAGS_SHORT` list, with no distinction for which engine (grep vs rg) actually received the flag. GNU grep's `-T` takes no value at all.

# Expected behavior

When the target engine is GNU grep, `-T` should be treated as boolean (matching `--initial-tab`) the pattern and path positionals should not shift. (Ripgrep's `-T`/`--type-not` value-consuming behavior should still work when the engine is rg.)

# Impact

Any use of grep's `-T` (tab-separated output, useful for piping into tools that split on tabs) exits 2 with no output instead of searching, with nothing distinguishing it from a genuine "no such file" error.

# Environment
- rtk develop @ e53ec1c (Cargo.toml 0.42.4); confirmed against installed rtk 0.48.0, whose whose grep arg-parsing matches this code
- GNU grep 3.12, Linux

gene1wood:
## Summary

`-T` means two different things to the two engines RTK proxies:
GNU grep's `-T`/`--initial-tab` is **boolean**, while ripgrep's `-T` is
`--type-not`, which **takes a value**. RTK's `VALUE_FLAGS_SHORT` included
`T` unconditionally (borrowed from ripgrep), so when the underlying
engine was actually GNU grep, RTK wrongly consumed the very next
token — the search pattern itself — as `-T`'s value, shifting every
remaining argument out of place.

```
grep     -T apple f1.txt f2.txt   # tab-separated output, exit 0
rtk grep -T apple f1.txt f2.txt   # (empty), exit 2 — before this fix
```

## Fix

- Add `is_value_taking_short(ch, engine)` in `src/cmds/system/search.rs`,
  which wraps the existing `VALUE_FLAGS_SHORT` list and adds `T` back
  only when the engine is `rg`.
- Thread an `Engine` parameter through `parse_cluster` and
  `extract_pattern_path` so flag-arity decisions can depend on which
  engine is actually being invoked (needed for `-T`; both functions
  previously had no way to tell grep and rg apart).
- `VALUE_FLAGS_SHORT` no longer includes `T` directly; the doc comment
  explains why and points at `is_value_taking_short`.

## Test plan

- Added unit tests covering both sides: `-T` is boolean for
  `Engine::Grep` and value-taking for `Engine::Rg`, at both the
  `parse_cluster` and `extract_pattern_path` levels.
- Added an end-to-end byte-for-byte test in
  `tests/grep_faithful_format_test.rs` (`dash_capital_t_initial_tab_matches_grep_n`)
  that runs the compiled `rtk` binary against real `grep -T` and diffs
  stdout + exit code, with and without `-n`.
- Added a regression test in `tests/search_faithful_test.rs`
  (`rg_dash_capital_t_type_not_excludes_the_type`) proving `rtk rg -T rust`
  still correctly excludes files of the given type — confirming the
  grep-side fix didn't regress ripgrep's own `-T`/`--type-not`.

Fixes #3882

### DIFF
--- a/src/cmds/system/search.rs
+++ b/src/cmds/system/search.rs
@@ -22,9 +22,21 @@ use std::sync::LazyLock;
 /// Short single-char flags that consume one following token (or inline remainder)
 /// as their value. `-e` is handled separately — its value goes to `patterns`.
 /// Includes all rg short flags that take a value argument except `-e` and `-r`
-/// (stripped) and `-E` (dialect, left to #2138). Failure mode for a missing
-/// entry: the value becomes a positional (visible wrong result, not silent).
-const VALUE_FLAGS_SHORT: &[u8] = b"ABCMTdfgjmt";
+/// (stripped) and `-E` (dialect, left to #2138). Also excludes `T`: ripgrep's
+/// `-T` is value-taking (`--type-not`), but GNU grep's `-T` is the boolean
+/// `--initial-tab` — see `is_value_taking_short`, which adds `T` back only for
+/// the `rg` engine. Failure mode for a missing entry: the value becomes a
+/// positional (visible wrong result, not silent).
+const VALUE_FLAGS_SHORT: &[u8] = b"ABCMdfgjmt";
+
+/// Whether short flag `ch` takes a value for the given engine. Wraps
+/// `VALUE_FLAGS_SHORT` (shared by both engines) and adds `T` only for `rg`
+/// (`--type-not`); GNU grep's `-T` (`--initial-tab`) is boolean, so treating it
+/// as value-taking there swallowed the next token — often the actual pattern
+/// (#3882).
+fn is_value_taking_short(ch: u8, engine: Engine) -> bool {
+    VALUE_FLAGS_SHORT.contains(&ch) || (ch == b'T' && matches!(engine, Engine::Rg))
+}
 
 /// Long flags that consume the NEXT token as their value (space-separated form).
 /// Inline `=` form (`--flag=value`) is one token and passes through unchanged.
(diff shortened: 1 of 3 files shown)


gene1wood:
Closing in favor of #3681

PR #3893, opened 2026-09-06
fix(filter): smart_truncate returned only the banner for max_lines = 1

Phantom-Star829:
## Summary

- `smart_truncate` uses `max_lines / 2` as the budget for ordinary (non-important) lines. With `max_lines = 1` that budget is 0, so `rtk read <file> --max-lines 1` (what the hook rewrites `head -1 <file>` to) returned only the `[N more lines]` banner: exit 0, no error, zero of the requested content. Floor the budget at 1.
- The banner is metadata, not one of the requested lines, so it no longer consumes a content slot (`kept_lines >= max_lines - 1` → `>= max_lines`). The old expression also underflows at `max_lines = 0`: on `develop` a debug build of `rtk read --max-lines 0 <file>` panics with `attempt to subtract with overflow` at `src/core/filter.rs:349`.
- Adds a regression test (`test_smart_truncate_max_lines_one_returns_a_line`).

## Test plan

- [x] Manual testing on a 200-line file: `rtk read --max-lines 1` prints the first line plus `[199 more lines]` (before: banner only); `rtk read --max-lines 0` prints the first line plus banner (before: panic in debug builds).

### DIFF
--- a/src/core/filter.rs
+++ b/src/core/filter.rs
@@ -339,14 +339,22 @@ pub fn smart_truncate(content: &str, max_lines: usize, _lang: &Language) -> Stri
             || trimmed == "}"
             || trimmed == "{";
 
-        if is_important || kept_lines < max_lines / 2 {
+        // The ordinary-line budget must never be zero. With max_lines = 1,
+        // `max_lines / 2` is 0, so no plain line could ever qualify and
+        // `head -1 file` returned the "[N more lines]" banner and nothing
+        // else -- exit 0, no error, zero content. Floor the budget at 1.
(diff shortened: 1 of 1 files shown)


KuSh:
Closing in favour of #2964, which fixes the same smart_truncate budget bug more completely: it fills the requested line budget exactly for any N, not just N of 1, and it also handles max_lines 0 without the underflow. Comments or follow-up work are welcome on that PR. Note that #3166 proposes a different contract for the same function, a literal contiguous prefix, so it is worth reading alongside.

PR #3900, opened 2026-09-06
Recall store: review fixes on top of #3278

pr0d1r2:
# Recall store: review fixes on top of #3278

This continues **#3278** by @aeppling. His ten commits are here with their
original authorship, rebased onto `develop`; everything after them is review
work on top. Base for comparison: `rtk-ai/rtk@1c79688`, the head of
`feat/retriever-recall` at the time of review.

- 33 commits: 10 by @aeppling, 23 review fixes.
- `develop` moved on. #3278 is 18 commits behind it; this branch is rebased and
  0 behind, which is why the diff against `feat/retriever-recall` is larger than
  the review work alone.
- Green: 3294 tests in the bin suite, 3396 across all suites, clippy clean at
  upstream defaults.

## Why a second PR rather than commits on #3278

The head branch of #3278 lives in `rtk-ai/rtk` and I have no write access to it,
so I could not push there. If it is easier to review as commits on the original
branch, say so and I will hand the branch over for @aeppling to pull in — the
work is his feature plus fixes, and the history is arranged so each fix is one
revertable commit.

## Merge blockers cleared

Both were open on #3278 at review time.

**Eviction ordered by two different clocks.** The retention rule deleted by
`created_at`, which the upsert refreshes, while the count cap deleted by `rowid`,
which it does not. An entry re-stored daily was fresh to one rule and
first-to-die under the other. Both rules order by `created_at` now, `rowid`
breaking ties only. This was the defect the second review round reopened.

**`benchmark` CI red.** The counter in `benchmark.sh` skipped only the legacy
`[see remaining: …]` hint. The two new sqlite hint forms are five fields each,
so they were counted as filenames and the job overshot by five.

## Correctness and privacy fixes

**Byte fidelity through `--grep`.** `grep_bytes` split on `\n` and re-emitted
every match with an unconditional trailing newline, so `rtk recall <hash>
--grep '^'` returned the stored bytes plus one. The store advertises
byte-faithfulness; it now holds through that path too.

**`mode = "disabled"` created the database.** `stats_snapshot()` opened
unconditionally, reached from `rtk gain --recalls` and from the daily telemetry
ping, so a user who turned recovery off still got a `recall.db`.

**Telemetry shipped command arguments.** The allowlist checked only the first
`_`/`-` token of a stats key, so a short command whose head matched a family was
transmitted whole — `go_build___cmd_acmepay` carried a project path. The
allowlist is tightened, and the slug type below now bounds what can reach a
stats key at all.

**Tee reads counted before the permission verdict.** `track_tee_read()` fired
ahead of the Allow/Deny decision, and `tee_reads` dedups with
`INSERT OR IGNORE`, so a denied read permanently blocked the later legitimate
read of that file from ever counting.

**`rtk gain --recalls` was broken for 63 filters.** The TOML runner passed the
raw shell command line as the slug, opening a row per invocation.

## Bounds, races and platforms

**The stats table had no cap and no eviction.** `recall` is capped at
`max_entries` and `tee_reads` at 500; `recall_stats` had neither. What kept it
small was a coincidence — `stat_family` truncates at the first digit preceded by
an underscore, and the one slug carrying a file path happens to have an index
digit in front of it. Three call sites already defeated that:
`cargo_{subcommand}`, `bun_{subcmd}` and `deno_{subcmd}` carry user-typed text
with no digit at all. Capped at 500, least-used evicted first.

**Slugs are typed now, not conventional.** A `Slug` says where its name came
from: a literal, a family plus `&'static str` parts, a value from the user's own
config, or a family plus an unbounded detail that reaches the tee filename and
never the stats key. There is no conversion from `String` or `&str`, so a caller
holding a runtime value has to say which it means and the compiler asks. Two AWS
arms changed shape rather than gaining a conversion: they matched
`args[0] == "sync"` in a guard and then forwarded the user's string anyway.

**Eviction could cut below the cap.** `evict_by_count` read `SELECT COUNT(*)`
into Rust and passed the difference to `DELETE … LIMIT excess`, so two processes
against the same 203 rows both deleted 3 and left 197. Measured with eight
threads on a barrier: 179 rows, 15 runs out of 15. The count moves inside the
statement, where SQLite evaluates it under the write lock it already holds.

**Owner-only was false on Windows.** `set_owner_only` was
`#[cfg(not(unix))] fn(..) {}` — an empty body — while the code's own comments and
tests described the store as owner-only. The store holds unredacted stdout, full
command lines and absolute working directories, so on Windows its only stated
mitigation did not exist, and every permission test was `#[cfg(unix)]` so CI
proved nothing about it. It now writes a single-entry DACL with
`PROTECTED_DACL_SECURITY_INFORMATION`; directories inherit, files do not.

## Performance

The sqlite store breached the <10ms startup target on per-file paths: 25.9ms
against 15.6ms disabled on `rtk grep -rn "let" src/cmds`. The review had
attributed the whole gap to `open()` running per `store()`. Measured per
component, `open()` was 1.80ms of it — about 18% — and the dominant term was the
gzip codec: 6.20ms to compress 889KB against 0.81ms for lz4. Fixed with a
connection cache, a cached config read and lz4; the gap is 2.7ms.

The `compression = false` path was removed rather than kept: it was slower than
lz4 at every size above 1KB *and* 2-9x larger on disk, so it never won on either
axis it existed for.

## Testing

Every fix here has a test that was shown to fail against the defect it guards,
which caught three of my own that did not. The eviction race needed a barrier —
the sequential version passed against the broken code, because sequential
callers never interleave one caller's count with another's delete.

The Windows work is compiled and run, not just written: the tree builds and
links for `x86_64-pc-windows-gnu` against the real `windows-sys`, and the suite
runs under wine at 3229 passing. Wine answers `SetNamedSecurityInfoW` with
`ERROR_SUCCESS` and then synthesises a descriptor from Unix mode bits, so the
DACL read-back cannot be observed there; those tests detect wine through
`ntdll!wine_get_version` and assert the write path, keeping the full assertions
for the `windows-latest` job in CI.

## Found while reviewing, deliberately not fixed here

These are pre-existing and outside this branch's scope. Filed here so they are
not lost.

- **Tee files never expire.** `retention_days` is consumed only by the sqlite
  path. The tee path's only limit is a file count, so `rtk config recall sqlite`
  orphans every existing `.log` forever — raw stdout at `0600`, no age bound.
- **`tee_max_file_size` is not a cap.** The `--- truncated at N bytes ---`
  marker is appended after the cut, so files land 33-36 bytes over the number
  the docs call a maximum.
- **`max_files` is global, not per-slug.** One noisy filter evicts every other
  filter's recovery files.
- **A failed tee write can leave an orphan** that still counts toward that cap.
- **Telemetry can transmit an invoked path.** `low_savings_commands`
  (`tracking.rs`) sends the first three whitespace tokens of `rtk_cmd`, and
  passthroughs are recorded as `rtk fallback: {raw_command}` (`main.rs`).
  Together those three tokens are `rtk`, `fallback:` and the program that ran —
  a repo-local path when that is what was invoked. Verified with `bin/packwerk`.
  This is a different path from the allowlist fix above and is not covered by it.

## Not addressed, and why

`force_tee_hint` and `force_tee_tail_hint` store `exit_code = None`, so
`recall --list` shows `-` for entries that came from an elision rather than a
failure. The exit code exists one frame above, in the runner's closure, but the
hints are emitted from inside filter functions that receive text only. Threading
it means changing filter signatures across 69 call sites in 19 modules this
branch does not otherwise touch, for one column in one listing. It previously
passed a fake `0`, which was worse: that is a lie the column cannot distinguish
from a real success.


pr0d1r2:
## Test plan

The repo's template asks for one and my description replaced it — sorry, here it
is.

| Check | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy --all-targets` at clippy's own defaults | 0 warnings |
| `cargo test --all` | 3294 in the bin suite, 3396 across all suites, 0 failed |
| Cross-build `x86_64-pc-windows-gnu` against real `windows-sys` | builds and links |
| Suite under wine on that target | 3229 passed, 0 failed |

Two things that run locally here but not in CI, in case they are useful to
anyone else on this repo:

- Clippy is run **twice** — once at upstream defaults, which is what this repo's
  CI sees, and once under tighter complexity thresholds that only the recall
  modules are held to. The thresholds are not in this PR; the modules carry
  `#![deny]` attributes that resolve at clippy's defaults when no `clippy.toml`
  is in scope, so they cost this tree nothing.
- The Windows half is compiled and executed rather than reasoned about. Wine
  answers `SetNamedSecurityInfoW` with `ERROR_SUCCESS` and then synthesises a
  security descriptor from the underlying Unix mode bits, so a DACL read back is
  never the one that was written. Those tests detect wine through
  `ntdll!wine_get_version` and assert only the write path there, keeping the
  read-back assertions for the `windows-latest` job on this PR. **That job is
  the first real test of the ACL work** — worth watching before merging it.

## How this was produced, and how to review it

Every commit here carries a `Co-Authored-By: Claude Opus 5` trailer, so rather
than leave that unexplained: this was a spec-driven review, and the artifacts it
produced are visible in the PR itself.

- **One decision per commit.** 33 commits rather than a squash, because
  `git revert <sha>` should undo exactly one decision. Ten are @aeppling's,
  unchanged in authorship.
- **Reasoning in the message, not in a doc.** Each message states the mechanism,
  what was measured, what was rejected and why, and — where it applies — the
  condition under which the decision should be reversed. If a commit looks
  surprising, its message is the argument for it.
- **A guard counts only once a planted violation has failed it.** Every fix here
  has a test that was run against the defect and seen to fail. That rule exists
  because it caught three of my own guards that did not: a regression test for
  the eviction race passed against the broken code until it was rewritten to use
  a barrier, and a CI-style check reported success over a tree it had just
  failed to compile.
- **Complexity ceilings that only fall.** The recall modules are held to 15
  lines and 4 arguments. Two of the splits that produced are load-bearing rather
  than cosmetic: separating `evict_by_age` from `evict_by_count` is what made
  the two rules' disagreement about "oldest" visible, and it is the merge
  blocker this PR opens with.

The tooling is [Claude Code](https://claude.com/claude-code) driving a
spec-driven loop of my own (`cavekit`): a spec holds numbered invariants and a
bug ledger, findings are written back into it as they are made, and the compact
notation exists because that document is re-read far more often than it is
written.

## The spec itself

Published rather than described, including the parts where the process was
wrong: **https://github.com/pr0d1r2/rtk/tree/review-spec**

The bug ledger there has rows for its own misses — a finding that was
**withdrawn** after a "demonstration" turned out to be built from test data
rather than a reachable path; a gate that reported green over a broken tree
because a pipeline's exit status under `set -o pipefail` is the last non-zero
one; a collision found only by checking a claim in a commit message written an
hour earlier; and a commit subject asserting that "the docs already claim" a
guarantee that no document claims.

I would rather hand you that than a clean-looking summary. If any of it reads as
overclaimed, say so — several rows in it exist because exactly that happened.

KuSh (changes requested):

**Claim**: carry #3278's recall store forward with 23 review fixes on top of it.

**Scope**: **accept if narrowed to "the `recall_stats` cap, slug bounding, and the Windows runtime guarantees"** — the three things `develop` still lacks. Everything else in the branch has either landed already or belongs to an issue someone else has taken.

**Ran**: `rtk grep` over a 300-line producer × 4 storage caps × before/after on both binaries; tee-read counting through `rtk rewrite` for the `cat` (Ask) and `tail` (Defer) forms; 80 distinct slugs against `recall_stats`; `rtk ls` on a sensitive directory as a real producer; `cargo check --target x86_64-pc-windows-gnu --all-targets` on both trees. Savings 97.5%, identical on both, against the 20% floor.

**CI** @`96565ae3`: never ran — only `license/cla` is green and `check-target` skipped, so nothing here has been built by CI yet.

### What happened to this branch

#3278 merged on 2026-09-08 as `77ba5c9a`. This branch forked from `1c79688`, its round-2 head, and the rounds that followed on #3278 fixed much of the same ground independently before it merged. That is an unlucky collision rather than a defect in the work here — but it means most of these 23 commits now describe problems `develop` no longer has.

Already fixed on `develop`, including both merge blockers this PR declares:

- **Eviction by two clocks** — `develop` orders *both* rules by `created_at` with `rowid` breaking ties only (`retriever.rs:447`, `:458`).
- **`benchmark` red** — `benchmark.sh` already skips both sqlite hint forms (lines 218-219, 235).
- Test-store isolation, `--from` past the end of an entry, the legacy `[tee]` migration, byte-faithful `--grep`, the cached connection. `TEE_READERS` and `expand_home` are on `develop` too.

Conversely the branch predates four fixes that landed after it forked, and does not carry them: hidden-tail storing, `cfg(test)` store isolation, `map_legacy_tee`, and the cap-in-force wording. The merge produces **22 conflicted files** covering every core recall file, so a hand resolution is the likeliest way for those four to be lost without anyone noticing.

### Blocks merge (1)

**The tee-read counter reads zero on the whole canonical read path.** `develop` counts in two places that agree — `rewrite_cmd.rs::run` and `hook_cmd.rs:237` in `decide_from_verdict`, both gated on "not `Deny`". This branch narrows both to the `Allow` arm only. `Ask` is what every rewritable read becomes unless an explicit allow rule matches, so on a machine with no rules configured nothing is counted at all:

```
                          develop              this branch
rewrite "cat  …log"   exit 3 -> recalls=1   exit 3 -> recalls=0
rewrite "tail …log"   exit 1 -> recalls=1   exit 1 -> recalls=0
```

`develop` over-counts here, which is real and is tracked as #3953. This trades that for a number that is structurally zero: `rtk gain --recalls` would report tee recall as never consulted, through every delegate that shells out to `rtk rewrite`. Dropping `Defer` is the larger half — `tail -n +51 …` has no RTK equivalent, and it is the command RTK's own tail hint prints, so it is the dominant real read.

I can see this is deliberate rather than an oversight — `v17_track_after_verdict` pins it, and "count after the verdict, never before" is the right instinct. The trouble is that the hook runs *before* the host prompts, so no gate there can tell an approved `Ask` from a rejected one; narrowing to `Allow` buys accuracy on the rejected ones by discarding every approved one, plus all of `Defer`.

The ask here is to **drop this change rather than repair it** (the test with it): #3953 covers it and another contributor has offered to take it, so the two would otherwise land on the same arm twice. Worth knowing that the shape being discussed there is to count where the read executes rather than in the hook gate.

This is separable from the slug work: `bounded_half` is called only from `tee_read_slug`, which extracts the slug, while the gating lives in `run()` and `decide_from_verdict`. Keeping the former and leaving both of the latter as `develop` has them is a clean cut.

### The narrowed claim, and what is worth carrying forward

- **`recall_stats` has no cap.** Verified: 80 distinct slugs produce **81 rows**, and the only `DELETE FROM recall_stats` on `develop` is the `gain --reset` wipe. `EVICT_STATS_SQL` is the real fix and it is self-contained.
- **Slug bounding** (`DETAIL_MARKER` / `bounded_half`) — not on `develop`. One thing I could not confirm: `rtk ls` on a sensitive directory produced the slug `ls-hidden` with no path in it, so I did not reproduce a path reaching `recall_stats`. If you still have the case that produces one, naming it would make this land much faster.
- **Windows owner-only and hint path handling** — not on `develop`. Both trees type-check clean for `x86_64-pc-windows-gnu`, so the build-repair half no longer applies; what remains is a runtime guarantee I have no way to execute from here, and I would take it on the strength of the code and a test.

Three small PRs against current `develop` would each be reviewable in one pass, and none of them would conflict. The `recall_stats` cap is the one I would send first.

### Your questions

> If it is easier to review as commits on the original branch, say so and I will hand the branch over for @aeppling to pull in

Moot now — #3278 merged two days after this PR opened, so there is no branch left to hand over. The narrowed PRs above are the path that survives.

### Also useful

The findings under *Found while reviewing, deliberately not fixed here* are worth keeping — the telemetry path leak in `low_savings_commands` in particular, verified with `bin/packwerk`. If you would like them as issues, say the word and I will file them with your wording; happy to do that whether or not you pick the narrowed PRs up.

Your reasoning on `force_tee_hint` and `exit_code = None` is right, and matches where the same question landed on #3278: 69 call sites in 19 modules for one column is not a trade worth making.

### Next

Open the `recall_stats` cap as its own PR against current `develop` — verified, self-contained, no conflicts. Slug bounding and Windows follow separately, and leave `run()`'s counting arm out of all three.

Rounds: 1/3. Threads: 0 open.

PR #656, opened 2026-03-17
fix(security): restrict tee file/directory permissions to 0600/0700

rjamestaylor:
## Summary

Tee files written to `~/.local/share/rtk/tee/` can contain sensitive command output — error messages often include tokens, API keys, or credentials. Previously these files were created with the process umask (typically `0644`), making them readable by other users on the same system.

### Changes

- **Tee directory**: `chmod 0700` (owner rwx only) applied after `create_dir_all`
- **Tee files**: `chmod 0600` (owner rw only) applied after each `fs::write`
- Both changes guarded by `#[cfg(unix)]` — Windows behaviour is unchanged
- Hint line updated: `[full output: <path> — may contain sensitive data]`
- New test `test_write_tee_file_permissions` (#[cfg(unix)]) verifies both modes

### Files changed

- `src/tee.rs`: permissions logic + updated hint text + new test
- `CHANGELOG.md`: unreleased entry

## Test plan

- [x] `cargo fmt --all --check` — clean
- [x] `cargo clippy --all-targets` — zero warnings
- [x] `cargo test --all` — 935 tests pass (includes new permission test)
- [x] Manual: tee file created on macOS, `ls -l` confirms `-rw-------` and `drwx------`

## Checklist

- [x] Single scope (tee permissions only)
- [x] CHANGELOG.md updated
- [x] All tests pass
- [x] DCO `Signed-off-by` present
- [x] Targets `develop` branch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3906, opened 2026-09-07
fix(tee): note that tee files may contain sensitive data in the hint

rjamestaylor:
Tee files hold raw, unfiltered output, and a failing command's stderr routinely echoes tokens, API keys, or connection strings. The recovery hint pointed an agent at the file without saying so, so it could be quoted into a transcript or a bug report unaware.

Append "— may contain sensitive data" inside the existing brackets. The "[full output: " prefix that ctest and other filters match on is unchanged, and the path stays shell-quoted for copy-paste.

Follow-up to #656: the 0600/0700 permission work already on develop (9cf7a6d, a1bbcaf) covered everything in that branch except this wording.

## Summary

Follow-up to #656, which @KuSh closed as covered by the 0600/0700 tee permission work already on develop (9cf7a6d, a1bbcaf) — apart from the hint wording, which is all this PR carries.
	•	`format_hint` now emits `[full output: <path> — may contain sensitive data]`. Tee files are raw, unfiltered output and a failing command’s stderr routinely echoes tokens, API keys, or connection strings; the hint pointed an agent at the file without saying so.
	•	Updated the four docs that print the hint verbatim (README, docs/usage/FEATURES.md, docs/guide/getting-started/configuration.md, docs/contributing/TECHNICAL.md).
The notice goes inside the existing brackets, after the path, on purpose:
	•	The `[full output:`  prefix is unchanged, so the line filters in `ctest_cmd.rs` (starts_with("`[full output:`")) and the `vitest_cmd.rs` hint tests keep matching.
	•	The path stays shell-quoted by `display_shell_path` and remains directly copy-pasteable.
	•	Nothing parses a path back out of the hint, so the trailing clause is safe.

**NOTE**: Cost is ~6 tokens per hint, and only on the failure/truncation paths that tee at all; `guard::never_worse` still caps the emitted total at the raw output.

Deliberately out of scope: `force_tee_tail_hint ([see remaining: tail -n +N <path>]`) points at the same files but is a runnable shell command rather than prose, and #656’s ask was the `[full output: …]` line. Happy to extend it in this PR if you’d prefer the two hints to match.

## Test plan

- [x] cargo fmt --all && cargo clippy --all-targets && cargo test
	•	3108/3108 unit tests pass, clippy clean, fmt clean.
	•	One pre-existing unrelated failure in this sandbox: `copilot_selfheal_test::unwritable_hooks_dir_never_breaks_the_hook` chmods a dir to 0555 and expects the write to fail, which root bypasses. Confirmed identical on unmodified develop.
	•	New `test_format_hint_warns_about_sensitive_data` (exact-string), plus the existing `test_format_hint_avoids_backslash_escaped_whitespace` exact-equality assertion updated — it still guards against backslash-escaped whitespace.
	•	Manual testing: `rtk cargo test --test copilot_selfheal_test` against a failing suite prints

```
[full output: /…/tee/1788748449_cargo_test.log — may contain sensitive data]
```

and the file is `-rw-------`, confirming the merged permission fix.

co-authored with Claude Code Opus 5

### DIFF
--- a/src/core/tee.rs
+++ b/src/core/tee.rs
@@ -261,8 +261,14 @@ fn display_shell_path(path: &std::path::Path) -> String {
     format!("\"{}\"", escape_double_quoted_path(&display))
 }
 
+/// The tee file is raw, unfiltered output: failing commands routinely echo
+/// tokens, API keys, and connection strings, so the hint says so before an
+/// agent quotes the file into a transcript or a bug report.
 fn format_hint(path: &std::path::Path) -> String {
-    format!("[full output: {}]", display_shell_path(path))
+    format!(
+        "[full output: {} — may contain sensitive data]",
+        display_shell_path(path)
+    )
 }
 
 /// Convenience: tee + format hint in one call.
@@ -588,6 +594,15 @@ mod tests {
     }
 
     #[test]
+    fn test_format_hint_warns_about_sensitive_data() {
+        let path = PathBuf::from("/tmp/rtk/tee/123_cargo_test.log");
+        assert_eq!(
+            format_hint(&path),
+            "[full output: /tmp/rtk/tee/123_cargo_test.log — may contain sensitive data]"
+        );
+    }
+
+    #[test]
     fn test_display_shell_path_preserves_simple_paths() {
         let path = PathBuf::from("/tmp/rtk/tee/123_cargo_test.log");
         assert_eq!(display_shell_path(&path), "/tmp/rtk/tee/123_cargo_test.log");
@@ -644,7 +659,7 @@ mod tests {
 
         assert_eq!(
             hint,
-            "[full output: \"$HOME/Library/Application Support/rtk/tee/123_go_test.log\"]"
+            "[full output: \"$HOME/Library/Application Support/rtk/tee/123_go_test.log\" — may contain sensitive data]"
         );
         assert!(
             !hint.contains("\\ "),
--- a/README.md
+++ b/README.md
@@ -437,7 +437,7 @@ When a command fails, RTK saves the full unfiltered output so the LLM can read i
 
 ```
 FAILED: 2/15 tests
-[full output: ~/.local/share/rtk/tee/1707753600_cargo_test.log]
+[full output: ~/.local/share/rtk/tee/1707753600_cargo_test.log — may contain sensitive data]
 ```
 
 For the full config reference (all sections, env vars, per-project filters), see the [Configuration guide](https://www.rtk-ai.app/guide/getting-started/configuration).
--- a/docs/contributing/TECHNICAL.md
+++ b/docs/contributing/TECHNICAL.md
@@ -281,7 +281,7 @@ Analytics commands (`rtk gain`, `rtk cc-economics`, `rtk session`) query this da
 On command failure (non-zero exit code):
 
 1. Raw unfiltered output is saved to `~/.local/share/rtk/tee/{epoch}_{slug}.log`
-2. A hint line is printed: `[full output: ~/.../tee/1234_cargo_test.log]`
+2. A hint line is printed: `[full output: ~/.../tee/1234_cargo_test.log — may contain sensitive data]`
 3. LLM agents can re-read the file instead of re-running the failed command
 
 Tee is configurable (enabled/disabled, min size, max files, max file size) and never affects command output or exit code on failure.
--- a/docs/guide/getting-started/configuration.md
+++ b/docs/guide/getting-started/configuration.md
@@ -70,7 +70,7 @@ When a command fails, RTK saves the full raw output to a local file and prints t
 
 ```
 FAILED: 2/15 tests
-[full output: ~/.local/share/rtk/tee/1707753600_cargo_test.log]
+[full output: ~/.local/share/rtk/tee/1707753600_cargo_test.log — may contain sensitive data]
 ```
 
 Your AI assistant can then read the file if it needs more detail, without re-running the command.
--- a/docs/usage/FEATURES.md
+++ b/docs/usage/FEATURES.md
@@ -1376,7 +1376,7 @@ Quand une commande echoue, RTK sauvegarde automatiquement la sortie brute comple
 **Sortie :**
 ```
 FAILED: 2/15 tests
-[full output: ~/.local/share/rtk/tee/1707753600_cargo_test.log]
+[full output: ~/.local/share/rtk/tee/1707753600_cargo_test.log — may contain sensitive data]
 ```
 
 **Configuration :**


KuSh (commented):
Reviewed at `255f926` against merge-base `5e10aa6`, built both sides and compared behaviour on real producers.

**The change is correct and the contract holds.** `format_hint` is the only producer of the string, and every consumer keys on the `[full output:` *prefix* (`ctest_cmd.rs:1127/1478/1725`, the stubbed closures in `vitest_cmd.rs`), so the appended clause is invisible to them. Nothing parses a path back out. The em dash matches output rtk already emits elsewhere (`pipe_cmd.rs:246`, `pip_cmd.rs:31`). Doc coverage is complete.

Measured on a failing `cargo test` whose assertion message embeds a fake AWS key: identical output plus the notice, exactly +29 bytes, invariant from 1 to 20 failing tests (+6.2% of emitted output at the smallest hint-bearing size, +0.9% at 3KB). Exit codes preserved, tee file `-rw-------` in a `0700` dir.

**Merge and CI.** Clean merge with current `develop` (the 8 commits ahead touch `hooks/`, `src/hooks/init.rs`, `src/core/config.rs` — no overlap with `tee.rs`); 3223 tests pass on the merged tree, clippy and fmt clean, `--ignored` 8/8 under `LC_ALL=C`. Worth noting the `CI` workflow on this PR is still `action_required` and has never run — the only green check is the CLA, so the above is local verification.

Two notes below, neither blocking.

**One thing I checked and am explicitly not filing:** whether the extra ~7 estimated tokens can push `filtered + hint` over `guard::never_worse` in `runner::emit_guarded` and silently drop the hint. The asymmetry is real — `runner.rs:141` passes `guard_raw = raw_stdout` while `tee_raw = raw`, so on the `stdout_only().tee()` filters a fallback would lose the only pointer to stderr — but I could not make this PR change the outcome on any real input: 161 sizes on the pytest passthrough path, 9 pytest failure counts, 7 cargo-test failure counts, zero divergence. `MIN_TEE_SIZE = 500` puts a floor under the interesting region. Pre-existing hazard, not something this PR moves.

src/core/tee.rs
```diff
@@ -261,8 +261,14 @@ fn display_shell_path(path: &std::path::Path) -> String {
     format!("\"{}\"", escape_double_quoted_path(&display))
 }
 
+/// The tee file is raw, unfiltered output: failing commands routinely echo
+/// tokens, API keys, and connection strings, so the hint says so before an
+/// agent quotes the file into a transcript or a bug report.
 fn format_hint(path: &std::path::Path) -> String {
-    format!("[full output: {}]", display_shell_path(path))
+    format!(
+        "[full output: {} — may contain sensitive data]",
```
KuSh, line 269: The reasoning here applies just as strongly to `force_tee_tail_hint`, and I'd take up your offer to extend it in this PR.

It writes through the identical `force_tee_path`, so it points at files in the same directory with the same contents — but it has more call sites (15+ vs 8: `search.rs:682,692`, `find_cmd.rs:614,725`, `ls.rs:163`, `ruff_cmd.rs:226`, `pytest_cmd.rs:228`, `rubocop_cmd.rs:204`, `ctest_cmd.rs:588,804`, `uv_cmd.rs:213`, `main.rs`), and unlike the failure path it fires on **successful** commands.

The case that convinced me is `search.rs:682`, which tees the untruncated matched file content — precisely the content most likely to hold secrets. Against the binary built from this branch:

```
$ rtk grep AWS_SECRET .env          # exits 0
line25 AWS_SECRET_ACCESS_KEY=wJalrXUtnFEMI/K7MDENG/bPxRfiCY25
  +55 more in .env [see remaining: tail -n +26 …/tee/1788782694_grep_0__env.log]

$ rtk find "*.txt"
+350 more
[see remaining: tail -n +51 …/tee/1788782810_find.log]
```

`rtk grep` over a `.env` is the canonical secret-in-tee scenario, and after this PR it's the one hint form that says nothing — while handing back a directly runnable `tail`. An agent warned off quoting a `[full output: …]` file will quote this one instead.

Same clause after the path, before the closing bracket, keeps `[see remaining: ` matching for `ctest_cmd.rs:1128` and leaves the `tail -n +N <path>` command copy-pasteable.

src/core/tee.rs
```diff
@@ -261,8 +261,14 @@ fn display_shell_path(path: &std::path::Path) -> String {
     format!("\"{}\"", escape_double_quoted_path(&display))
 }
 
+/// The tee file is raw, unfiltered output: failing commands routinely echo
```
KuSh, line 264: While you're in here: `force_tee_hint`'s rustdoc at line 308 still documents the old shape —

```rust
/// Returns `[full output: ~/path]`, or None if tee is disabled/skipped.
```

It now returns `[full output: ~/path — may contain sensitive data]`. Same staleness in `src/core/README.md:122`, which also shows `[full output: ...]` as the produced form.

The four docs that print the hint verbatim are all correctly updated — it's just these two in-tree descriptions of the format that were missed.

placeholdr-net: Thanks — both of these are fair, and I'd taken the `force_tee_tail_hint` extension as agreed. Before I push, the base has moved in a way that changes the shape of the answer, so I'd rather ask than guess.

**The recall store landed since your review.** `src/core/tee.rs` is now a dispatcher over `[retriever] mode`, the file path moved to `src/core/tee_file.rs`, and `format_hint` — the single producer we both keyed the original reasoning on — no longer exists. There are now five:

| producer | emits |
|---|---|
| `tee.rs::store_hint` (sqlite) | `[full output: rtk recall <hash>]` |
| `tee.rs::force_tee_tail_hint` (sqlite arm) | `[+N hidden: rtk recall <hash>]` |
| `tee_file.rs::tee_and_hint` | `[full output: <path>]` |
| `tee_file.rs::force_tee_hint` | `[full output: <path>]` |
| `tee_file.rs::force_tee_tail_hint` | `[see remaining: tail -n +N <path>]` |

Two consequences:

1. **`RecoveryMode::Sqlite` is now `#[default]`.** So the hint most users actually see is the `rtk recall` one, and this PR as it stands touches only the non-default legacy tee mode. Merging it today would leave the default path unannotated, which isn't what either of us wants. I'd like to cover all five, at the three `tee.rs` entry points so it's one choke point rather than five call sites.

2. **Both doc nits are already fixed by the refactor** — `force_tee_hint` lost its rustdoc entirely, and `src/core/README.md:148` now documents the recall shapes correctly. Nothing for me to do there.

**The placement question.** That same README line calls these "a **runnable** hint" and says "the agent runs `rtk recall <hash>`". The form you approved puts the notice inside the bracket, which was safe when the payload was a bare path but now makes the runnable unit read:

```
[full output: rtk recall 3f9c2a81d4e7 — may contain sensitive data]
[see remaining: tail -n +12 ~/.local/share/rtk/tee/1707753600_grep.log — may contain sensitive data]
```

An agent that copies the bracket contents runs a broken command. Moving the notice outside keeps the bracket a clean runnable unit and still preserves every prefix consumer I can find (`ctest_cmd.rs:1127-1128,1478,1725,1778`, `ls.rs:523`, `vitest_cmd.rs:596-603` all match on the opening `[full output:` / `[see remaining:` / `[+`):

```
[full output: rtk recall 3f9c2a81d4e7] — may contain sensitive data
[see remaining: tail -n +12 ~/.local/share/rtk/tee/1707753600_grep.log] — may contain sensitive data
[+37 hidden: rtk recall 9b1e4f2c8a03] — may contain sensitive data
```

Which do you want? I'll take inside-the-bracket if you'd rather keep the shape you reviewed — your call on whether the runnability cost is worth the tighter line. Either way the PR currently conflicts with `develop` (the conflicting hunks *are* the wording change), so I'll resolve the merge and push the full five-producer version in one go once you've picked.

src/core/tee.rs
KuSh: Outside the bracket. And the runnability argument was already right before the refactor — my note claimed the inside form "leaves the `tail -n +N <path>` command copy-pasteable", which was wrong: that bracket body was already a runnable command, not a bare path. The recall store widened the problem rather than creating it.

Two things back you up beyond the README line you quoted:

- `src/core/README.md:148` calls it a runnable hint, and `:150` says "The agent runs `rtk recall <hash>`".
- `compact_blob_show` (`src/cmds/git/git_cmd.rs:1037-1051`) carries `-C`/`-c` into its hint "so it is runnable from anywhere, not just the repo root", shell-quoting each arg "for copy-paste safety" (`:1042`). Prose inside the bracket breaks precisely the property that code goes out of its way to preserve.

Every consumer I can find matches the opening prefix only — `ctest_cmd.rs:1115,1461,1708,1761`, `ls.rs:523-524`, `vitest_cmd.rs:601-608` — so outside placement costs nothing there. It also makes your choke point cleaner: the three `tee.rs` entry points append to the returned `Option<String>`, instead of splicing before a `]` produced in `tee_file.rs` or `store_hint`.

One exclusion for your sweep: `git_cmd.rs:1051` is a **sixth** `[see remaining: …]` producer, and it should *not* get the notice. Its comment at `:1033` says it deliberately avoids the tee — it stores nothing and re-runs `git show` against repo content the user already has, so there is no new exposure to warn about.

So: five producers, notice outside the bracket, `git_cmd.rs` untouched.

Heads-up on the base — it moved again since you wrote this (edition 2024, a rustfmt style-edition-2024 reformat, and `src/cmds/git/git.rs` → `git_cmd.rs`), so line numbers in both our earlier comments have shifted; the ones above are against current develop. The four doc examples are all recall-shaped now with no path form left, so those edits need rewriting rather than rebasing.

src/core/tee.rs
KuSh: Confirmed moot on current develop — `force_tee_hint` carries no rustdoc in either `tee.rs` or `tee_file.rs`, and `src/core/README.md:148` documents the recall shapes correctly. Nothing to do here.

KuSh:
Thanks for picking this up as a follow-up to #656, and for the care over keeping the `[full output: ` prefix intact for the filters that match on it. The recovery hint was rebuilt by #3278 (merged 2026-09-09), which made the sqlite recall store the default: src/core/tee.rs is now just a mode dispatcher with no format_hint, the path-based hint moved into the new src/core/tee_file.rs, and the default hint is `[full output: rtk recall <hash>]` rather than a file path. The four documentation lines this PR rewrites were replaced by that merge as well, and both new tests assert on the deleted tee.rs helper, so none of this can land as written. The underlying point still stands in two places develop does not cover: the legacy tee hints in tee_file.rs still name a raw file with no warning, and the recall hint says nothing about the stored output being unfiltered; either would be welcome as an issue or a narrow PR against the current hint code. If I have misread which part you were after, please comment or reopen.

Issue #3410, opened 2026-08-04
The rewrite rule for `find` is unconditional, so `find … -delete` / `-exec` is captured and refused — the command silently does nothing and breaks `&&` chains


HetCreep:
`find` is rewritten to `rtk find` for every invocation, but `rtk find` deliberately refuses the predicates and actions it does not implement. The result is that a working `find` command becomes a no-op with a nonzero exit.

*Provenance: AI-assisted source audit, human-directed, at **v0.44.2 (`700bdde`)**. Reproduced by running the shipped `rtk 0.44.2` binary against a scratch directory; outputs verbatim.*

### Reproduction

```
$ ls a
keep.txt  x.tmp  y.tmp

$ rtk rewrite find . -name '*.tmp' -delete
rtk find . -name *.tmp -delete

$ rtk find . -name '*.tmp' -delete
rtk: rtk find does not support compound predicates or actions (e.g. -not, -exec). Use `find` directly.
$ echo $?
1
$ ls a
keep.txt  x.tmp  y.tmp        ← nothing was deleted

$ find . -name '*.tmp' -delete   # real find
$ ls a
keep.txt                       ← deleted
```

And because the exit status is nonzero, anything chained after it does not run:

```
$ rtk find . -name '*.tmp' -delete && echo next
                               # "next" never prints
```

### The cause

`src/discover/rules.rs:139-141` matches `^find\s+` unconditionally — there is no allowlist of the predicates `rtk find` actually supports. So the rewrite captures `-delete`, `-exec`, `-not` and every other form the implementation then rejects.

The rewrite layer is deciding to handle a command that the handler has already declared it cannot handle.

### Why it matters beyond the error message

- `find -delete` and `find -exec` are the destructive forms. The user's intent was to change the filesystem; nothing changed, and the only signal is a message that reads like advice rather than a failure.
- Under an agent, this is worse: the agent asked for a cleanup, got a message about unsupported predicates, and the files it believes are gone are still there.
- The nonzero exit aborts `&&` chains, so a script or agent workflow built around `find … -delete && …` stops at the first step.

### Suggested fix (untested — no programmer has reviewed this)

Narrow `rules.rs:139` so it only matches the `find` forms `rtk find` implements — an allowlist of supported predicates, in the same spirit as the `kubectl` rule at `rules.rs:403`, which is limited to `get|logs|describe|apply` and correctly lets `kubectl delete pod web` pass through untouched.

[...]

PR #3907, opened 2026-09-07
fix(test): hint POSIX test misuse toward rtk run

antonio-orionus:
## Summary
- `rtk test` expects a test-runner command. A leading POSIX unary flag (e.g. `-d`) now emits a targeted message pointing at `rtk run "test ..."` or direct `test ...`.
- Real runner invocations (`cargo test`, `pytest -x`) are untouched; `-h` stays intercepted by clap.

Split out of #2287 per review: this is the sound, unrelated half. The `find` guard stays for follow-up work (#3410, #3681).

## Test plan
- NEW unit tests in `src/main.rs`: POSIX flag gets message, runner command does not, `-d` still parses as Test command.
- Verified locally: `cargo fmt --check` clean, `cargo clippy --all-targets` zero warnings, 3 new unit tests pass.
- Manual: `rtk test -d /tmp` emits the targeted hint (exit 1); `rtk run "test -d /tmp"` works (exit 0); `rtk test cargo test` unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,8 @@ rtk err <cmd>                   # Filter errors only from any command
 rtk test <cmd>                  # Generic test wrapper - failures only (-90%)
 ```
 
+`rtk test` expects a test-runner command, for example `rtk test cargo test`. It is not POSIX `test`; use `rtk run "test -d path"` or direct `test -d path` for shell condition checks.
+
 ### Build & Lint
 ```bash
 rtk lint                        # ESLint grouped by rule/file
--- a/docs/guide/resources/troubleshooting.md
+++ b/docs/guide/resources/troubleshooting.md
@@ -62,6 +62,18 @@ rtk gain    # should now show token savings stats
    cat ~/.claude/settings.json | grep rtk
    ```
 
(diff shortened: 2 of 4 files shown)


antonio-orionus:
Verified locally: fmt clean, clippy zero warnings, 3 new unit tests pass. Manual: `rtk test -d /tmp` emits the targeted hint (exit 1); `rtk run "test -d /tmp"` works (exit 0); `rtk test cargo test` unaffected.

KuSh:
Thanks for splitting this out. The hint is no longer the shape develop uses: #2515 (merged 2026-09-08) made `rtk test` hand a native test expression to the POSIX `test` binary, so `rtk test -d path` now returns the real condition result (exit 0 or 1) instead of an error, and the README and troubleshooting lines here would describe behaviour that has since changed. One gap is still open: an expression with no leading flag, such as `rtk test a = a`, is treated as a runner command and fails with exit 127, because is_native_test_expression in src/main.rs only accepts a leading dash. An issue or a small PR on that function would be welcome, and a doc line saying the native form works would still help. Please comment or reopen if I have missed a part of this.

PR #3909, opened 2026-09-07
feat(rewrite): add --host, stop gating OpenClaw on Claude rules

### LINKED ISSUE #3908: rewrite: host-scoped permission authority so plugin hosts stop adding a second approval gate
## Summary

`rtk rewrite` evaluates every command against Claude Code's permission files regardless of which agent called it. For OpenClaw this turns routine rewrites into blocking approval prompts on a host whose own exec policy already allows the command. Add a host parameter to `rtk rewrite`, register OpenClaw as a host whose permission authority is the host itself, and drop `requireApproval` from the OpenClaw plugin. Behavior for every existing caller is unchanged.

## Problem statement

`rewrite_cmd::run` calls `check_command(cmd)`, which `src/hooks/permissions.rs` defines as `check_command_for(cmd, Host::Claude)`. The `Host` enum already carries Claude, Cursor, Gemini, Droid, and Vibe, and `rtk rewrite` can select none of them: `Commands::Rewrite` declares `args` and nothing else.

When the OpenClaw plugin calls `rtk rewrite <command>`, RTK reads the four Claude Code settings files that `load_permission_rules` builds and applies Claude Code's least-privilege default to a runtime that is not Claude Code. With no matching allow rule the verdict is `Default`, `Default` exits 3, and `openclaw/index.ts:132` converts exit 3 into a blocking plugin approval with `timeoutBehavior: "deny"` and `allowedDecisions: ["allow-once", "deny"]`.

Reproduced on OpenClaw 2026.7.1-2 with `tools.exec.mode=full`, effective `security=full`, and `ask=off`. Every rewritable command stops for a 120-second approval derived from a different agent's config file, which blocks unattended work that OpenClaw's own exec policy permits.

## OpenClaw is the only adapter that gates

| Host | Adapter | Exit 3 | Exit 2 |
|---|---|---|---|
| Pi | `hooks/pi/rtk.ts` | rewrite, no gate | passthrough |
| Hermes | `hooks/hermes/rtk-rewrite/__init__.py` | rewrite, no gate | passthrough |
| OpenCode | `hooks/opencode/rtk.ts` | rewrite, no gate | passthrough |
| Mistral Vibe | `src/hooks/hook_cmd.rs`, `run_vibe_inner` | rewrite, no gate | deny |
| OpenClaw | `openclaw/index.ts:132` | blocking approval | block |

`hooks/pi/README.md` states the convention: "Exit codes 0 and 3 both mean 'rewrite and allow'; they are handled identically."

`vscode_response_from_decision` in `src/hooks/hook_cmd.rs` records the same failure in another host. In #3037, asserting `"ask"` made Copilot CLI 1.0.66+ force a blocking dialog with no remember option on every rewritten command, and the fix was to stop asserting and defer to the host's native flow.

Commit `487a2e7`, `fix(openclaw): handle exit code 3 from rtk rewrite` from PR #2202, added the OpenClaw gate on purpose, so this proposal revises a decision that was made deliberately. The argument for revising it is that exit 3 was read as a permission decision when OpenClaw consumes it as a rewrite decision and then applies its own exec policy.

## Strategy

Add a host parameter to `rtk rewrite` and a host whose permission authority is the host itself, mirroring the `Host::Vibe` case in `load_rules_for`.

`src/hooks/permissions.rs` gains `Host::OpenClaw`, resolving to an empty rule triple. RTK owns no OpenClaw rule source, so it asserts no verdict of its own.

`src/hooks/rewrite_cmd.rs` gains `evaluate_for(cmd, host, excluded, prefixes)`, and `evaluate` becomes the `Host::Claude` case of it. For a host-authoritative host, `Ask` and `Default` both resolve to `Allow`, so a rewritable command exits 0. `Deny` still exits 2. `contains_unattestable_construct` still short-circuits to `Passthrough` before any of this, so the substitution and redirect surface stays closed for every host.

`openclaw/index.ts` passes the host, removes the `requireApproval` block and the `"ask"` verdict arm, and keeps the exit-2 deny branch.

Docs in `openclaw/README.md` and `docs/guide/getting-started/supported-agents.md` state that OpenClaw's exec policy is the sole authority over whether a rewritten command runs.

### Interface

Preferred form is `rtk rewrite --host openclaw <command>`. The positional is declared `trailing_var_arg = true, allow_hyphen_values = true` so that `rtk rewrite ls -al` parses, and adding an option to that command needs a parse test of its own. If the option cannot be added without breaking that case, use `RTK_REWRITE_HOST=openclaw`, which carries no clap interaction.

## Security invariant

`src/hooks/rewrite_cmd.rs` carries a test module `exit_code_protocol` whose comment says `PermissionVerdict::Default` must map to exit 3 and never to exit 0, citing #1155. This proposal maps `Default` to exit 0 for one host, so the invariant needs restating as host-scoped rather than global.

Claude Code consumes the exit code as a permission decision, so collapsing `Default` to allow there would remove the only gate. OpenClaw consumes it as a rewrite decision and then applies `tools.exec.mode` and `security` to the resulting command, so the gate survives in the host.

Every existing #1155 test stays verbatim under `Host::Claude`. The OpenClaw case gets its own named invariant beside them.

## Scope

In scope: the host parameter, `Host::OpenClaw`, the plugin change, and the docs updates.

Out of scope: migrating Pi, Hermes, and OpenCode onto the host parameter. They already behave correctly through adapter code, and moving them is a follow-up that would let those adapters delete their exit-3 special cases.

Out of scope: a `permission_authority` key in `HooksConfig`. A global setting would reach the Claude Code hook path and weaken it.

## Tests

`src/hooks/rewrite_cmd.rs`:
- `Host::OpenClaw` with verdict `Default` gives `Allow`.
- `Host::Claude` with verdict `Default` still gives `Ask`, guarding #1155.
- An unattestable construct under `Host::OpenClaw` gives `Passthrough` and never `Allow`.
- A host with matching deny rules still gives `Deny`.

`src/hooks/permissions.rs`:
- `Host::OpenClaw` yields an empty rule triple, pinned beside the `Host::Vibe` case.

`src/main.rs`:
- Parse tests for the host argument.
- Regression cases proving `rtk rewrite ls -al` and `rtk rewrite head -50 file.txt` still parse with the option declared.

`openclaw/`:
- The directory has no test harness today. Either add a first test for the decision mapping, following `hooks/hermes/tests/`, or state that the Rust tests are the only coverage.

## Backward compatibility

`rtk rewrite <cmd>` with no host argument stays byte-identical, so no existing adapter needs a change.

Old plugin with new rtk gives today's behavior.

New plugin with old rtk gives exit 1 passthrough. Verified against rtk 0.45.0: `rtk rewrite --host openclaw "git status"` exits 1 with empty stdout and empty stderr, because the unknown flag is swallowed by the trailing positional. Token savings are lost until rtk is upgraded, and nothing blocks or denies. Adding a version probe next to the existing `which rtk` check in `openclaw/index.ts` would turn that silent loss into a warning.

## Known limitation

The plugin mutates `params.command` in `before_tool_call`, and OpenClaw folds hook adjustments into `executeParams` before calling `tool.execute` (observed in the installed OpenClaw 2026.7.1-2 bundle, `dist/tool-split-BONVLU1k.js`), so the exec tool receives `rtk git push` rather than `git push`. Whether OpenClaw's exec policy is evaluated before or after that adjustment is unverified. If it is evaluated after, a user deny rule written against the raw command stops matching.

This applies today on the exit-0 allow path and is not introduced by this change. Docs should tell OpenClaw users to write exec rules against the `rtk` form.

## Assumption open to review

`Host::OpenClaw` carries no deny rules, matching `Host::Vibe`. An OpenClaw user today gets `.claude/settings.json` deny rules enforced against OpenClaw as a side effect of the hardcoded host, and this change removes that. Nothing in the repo documents or claims that behavior.

The alternative keeps RTK reading Claude deny rules for OpenClaw while collapsing only `Ask` and `Default`. Choose it if any OpenClaw user is knowingly relying on `.claude/settings.json` to gate OpenClaw.

lucabattistini:
Closes #3908.

## Summary

`rtk rewrite` gains `--host <HOST>`, and OpenClaw is registered as a host whose permission authority is the host itself. For that host `Ask` and `Default` resolve to `Allow`, so a rewritable command exits 0 and the OpenClaw plugin stops raising an approval prompt of its own. `--host` defaults to `claude`, so every other adapter is byte-identical.

## Problem statement

`rewrite_cmd::run` called `check_command`, which `permissions.rs` defines as `check_command_for(cmd, Host::Claude)`, and `rtk rewrite` declared no argument that could select anything else. Every caller, whatever agent it belonged to, was evaluated against Claude Code's four settings files.

For OpenClaw this became a second approval gate. With no matching allow rule the verdict is `Default`, `Default` exits 3, and `openclaw/index.ts:132` converted exit 3 into a blocking `requireApproval` with `timeoutBehavior: "deny"`. A host running `tools.exec.mode=full` still stopped on every rewritable command, waiting on a decision derived from another agent's config file.

OpenClaw was the only adapter that gated. Pi (`hooks/pi/rtk.ts`), Hermes (`hooks/hermes/rtk-rewrite/__init__.py`), and OpenCode (`hooks/opencode/rtk.ts`) treat exit 0 and exit 3 identically, which `hooks/pi/README.md` states as the convention: "Exit codes 0 and 3 both mean 'rewrite and allow'; they are handled identically."

Commit `487a2e7` (`fix(openclaw): handle exit code 3 from rtk rewrite`, PR #2202) added the gate on purpose. It read exit 3 as a permission decision. OpenClaw consumes it as a rewrite decision and then applies its own exec policy, which is what this PR corrects.

## What changed

`src/hooks/permissions.rs` gains `Host::OpenClaw`, which loads an empty rule triple beside `Host::Vibe`, plus two methods: `Host::from_cli_name` and `Host::is_permission_authority`, which only `Host::OpenClaw` satisfies.

`src/hooks/rewrite_cmd.rs` gains `evaluate_for(cmd, host, excluded, prefixes)` and `evaluate_with_verdict_for`. Verdict handling runs in this order, and only step 4 is host-dependent:

1. `Deny` short-circuits to `Deny` for every host.
2. `contains_unattestable_construct` short-circuits to `Passthrough` for every host, so command substitutions and file-target redirects are never handed back as an allowed rewrite.
3. `Allow` yields `Allow`.
4. `Ask` and `Default` yield `Ask`, except for a permission-authority host, where they yield `Allow`.

`src/main.rs` declares `--host` on `Commands::Rewrite` and resolves it through `resolve_rewrite_host`.

`openclaw/index.ts` passes `--host openclaw`, removes `requireApproval` and the `"ask"` verdict, and keeps the exit-2 deny branch. A stray exit 3 is treated as a plain rewrite, matching the Pi, Hermes, and OpenCode convention.

`docs/contributing/TECHNICAL.md`, `docs/guide/getting-started/supported-agents.md`, `src/hooks/README.md`, and `openclaw/README.md` carry the same rule for readers arriving from each of those directions.

## The #1155 invariant

`PermissionVerdict::Default` must map to exit 3 under `Host::Claude`, and this PR keeps that. The invariant is restated as host-scoped rather than global, because it binds a host that reads the exit code as a permission decision. Claude Code does, so collapsing `Default` there would remove the only gate. OpenClaw reads it as a rewrite decision and then applies `tools.exec.mode` and `security` to the result, so the gate survives in the host.

Every test in `mod exit_code_protocol` is unchanged; only that module's doc comment gains the host-scoping paragraph. The new `mod host_scoped_permission_authority` pins both halves side by side: `Host::OpenClaw` with `Default` gives `Allow`, and the identical input under `Host::Claude` still gives `Ask`. `evaluate_with_verdict` is kept as a `#[cfg(test)]` `Host::Claude` wrapper so the older assertions read verbatim.

## Interface

`rtk rewrite --host openclaw git status`. The flag must precede the command, because the positional is `trailing_var_arg = true, allow_hyphen_values = true` and swallows everything after the first positional. `test_rewrite_clap_host_after_command_is_part_of_the_command` pins that.

Clap parses an option declared before such a positional, which the repository already relies on: `Commands::Run` carries `-c/--command` alongside the same positional shape, and `test_run_command_with_dash_c` asserts it. `RTK_REWRITE_HOST`, the fallback #3908 proposed if the flag could not be added, is not needed and is not included.

An unrecognized host warns on stderr and falls back to `claude` rather than aborting. A clap-level rejection exits 2, and the `rtk rewrite` exit-code protocol reserves 2 for "a deny rule matched", so a typo must not read as a deny. Falling back to the most restrictive host keeps the failure closed.

## Backward compatibility

`rtk rewrite <cmd>` with no `--host` is unchanged, so no existing adapter needs a change. `test_rewrite_clap_without_host_is_unchanged` asserts that `ls -al`, `head -50 file.txt`, `du -sh .`, and the quoted single-argument form all still parse with `host: None`.

Old plugin with new rtk gives today's behavior. New plugin with old rtk: `--host` is absorbed into the command string, nothing matches, rtk exits 1, and rewriting stops without blocking or denying anything. The version requirement is documented in `openclaw/README.md`. A version probe was considered and left out: it would need a hardcoded minimum version that this PR cannot know, and guessing wrong would disable the plugin for valid users.

## Behavior change for OpenClaw users

Deny rules in `.claude/settings.json` no longer apply to OpenClaw. An OpenClaw user got them enforced as a side effect of the hardcoded host, and nothing in the repository documented or claimed that behavior. `Host::OpenClaw` carries no rules, matching `Host::Vibe`, so RTK asserts no verdict of its own. Users relying on those rules should move them into OpenClaw's exec configuration. This is stated in `openclaw/README.md` and `docs/guide/getting-started/supported-agents.md`.

The alternative, keeping RTK reading Claude deny rules for OpenClaw while collapsing only `Ask` and `Default`, is a one-line change to `load_rules_for` if a reviewer wants it.

## Known limitation, carried over

The plugin replaces `params.command` in `before_tool_call`, and OpenClaw folds hook adjustments into the parameters passed to the exec tool, so the tool receives `rtk git push` rather than `git push`. A user exec rule written against the raw command does not match. This applies today on the exit-0 allow path and is not introduced here. `openclaw/README.md` now tells OpenClaw users to write exec rules against the `rtk` form.

## Tests

Added:

- `src/hooks/rewrite_cmd.rs`: `Host::OpenClaw` with `Default` and with `Ask` both give `Allow`; `Host::Claude` with `Default` still gives `Ask`; every other host still gives `Ask` on `Default`; a deny verdict still gives `Deny`; four unattestable constructs give `Passthrough` and never `Allow`; an unsupported command stays `Passthrough`.
- `src/hooks/permissions.rs`: `Host::OpenClaw` and `Host::Vibe` both load an empty rule triple; `check_command_for` with `Host::OpenClaw` returns `Default` rather than `Allow`, keeping this layer a pure rule evaluation; only `Host::OpenClaw` is a permission authority; every name in `HOST_NAMES` parses, case-insensitively; unknown names return `None`.
- `src/main.rs`: `--host` parses in both the space and `=` forms and leaves hyphenated command arguments intact; the no-flag invocations parse unchanged; `--host` after the command is part of the command; `resolve_rewrite_host` defaults to and falls back to `claude`.

`openclaw/` has no test harness, and this PR does not add one. The Rust tests above are the only coverage for the decision mapping.

## Local run

On the merge of `develop` into this branch, macOS arm64, rustc stable:

- `cargo build`: clean.
- `cargo fmt --all -- --check`: clean.
- `cargo clippy --all-targets`: no warnings, under `[lints.rust] warnings = "deny"` in `Cargo.toml`.
- `cargo test --all`: 3522 passed, 0 failed, 8 ignored.

## Merge with develop

`develop` added tee-read tracking to `rewrite_cmd::run` while this branch changed the same function's signature. The merge keeps both: `run` takes the host, calls `evaluate_for`, and still calls `track_tee_read` on every non-deny outcome. The exit-code doc comment moves down onto `run`, where inserting `const TEE_READERS` had left it stranded above the const.

## Out of scope

Migrating Pi, Hermes, and OpenCode onto `--host`: they already behave correctly through adapter code. A `permission_authority` key in `HooksConfig`: a global setting would reach the Claude Code hook path and weaken it. `CHANGELOG.md` is untouched, since `CONTRIBUTING.md:116` gives that file to release-please; the changelog entry comes from the commit message.


src/hooks/permissions.rs
```diff
@@ -57,7 +95,9 @@ pub(crate) fn load_rules_for(host: Host) -> (Vec<String>, Vec<String>, Vec<Strin
         Host::Cursor => load_cursor_rules(),
         Host::Gemini => load_gemini_rules(),
         Host::Droid => load_droid_rules(),
-        Host::Vibe => (Vec::new(), Vec::new(), Vec::new()),
+        // Vibe and OpenClaw: RTK owns no rule source for these hosts, so it
+        // asserts no verdict of its own and leaves the decision to the host.
+        Host::Vibe | Host::OpenClaw => (Vec::new(), Vec::new(), Vec::new()),
```
KuSh: **blocking — site 1 of the deny class.** `Host::OpenClaw` gets an empty rule triple here, so `check_command_for` can never *return* `Deny` for this host — your own `test_openclaw_verdict_is_default_not_allow` pins that — and `Default` then collapses to `Allow`.

With `{"permissions":{"deny":["Bash(du *)","Bash(git push *)"]}}`, measured on this head vs the merge-base:

```
                    merge-base (no flag)   --host claude   --host openclaw
du -sh .            exit 2                 exit 2          exit 0 -> rtk du -sh .
git push            exit 2                 exit 2          exit 0 -> rtk git push
```

End to end, with `requireApproval` removed by this PR:

```console
$ rtk rewrite --host openclaw "du -sh ."
rtk du -sh .          # exit 0, applied with no prompt
$ rtk du -sh src
5,0M    src           # the denied command ran
```

Compounds inherit it: `--host openclaw "git push && rm -rf /tmp/zz"` returns exit 0 and the whole string. It is a regression against `develop`, where the shipped plugin calls `rtk rewrite <cmd>` and branches on `status === 2`.

I take this as unintended, since `evaluate_with_verdict_for` short-circuits `Deny` first for every host — the intent that deny survives is visible, and this line is what defeats it. Options (a) and (b) are in the review body; either is fine by me, but the docs and the unreachable exit-2 branch need to match whichever you pick.

openclaw/README.md
```diff
@@ -59,6 +59,29 @@ In `openclaw.json`:
...
+
+RTK never rewrites a command containing a command substitution (`` ` ``, `$(...)`) or a redirect to a file, on any host. Those pass through unchanged.
+
+### Writing exec rules
+
+The plugin replaces `params.command` in `before_tool_call`, and OpenClaw folds hook adjustments into the parameters it passes to the exec tool. The tool therefore receives `rtk git push`, not `git push`. Write exec allow/deny rules against the `rtk` form. This was already true before the permission change.
+
+### rtk version requirement
+
+`rtk rewrite --host` needs an rtk build that carries [#3908](https://github.com/rtk-ai/rtk/issues/3908). On an older rtk, `--host` is absorbed into the command string RTK is asked to rewrite, nothing matches, and rtk exits 1: rewriting silently stops and no command is blocked or denied. If token savings disappear after installing this plugin, upgrade rtk.
```
KuSh: **blocking — site 2 of the deny class.** The first half of this sentence holds: on an older rtk, `--host` is absorbed into the command string. The second half does not.

Because the text reaching the permission gate becomes `--host openclaw git push origin main`, it no longer matches the user's rule. With `deny: ["Bash(git push *)"]` on the merge-base binary:

```console
$ rtk rewrite "git push origin main"                 # old plugin argv
exit 2                                                # blocked
$ rtk rewrite --host openclaw "git push origin main" # new plugin argv
exit 1                                                # passthrough -> OpenClaw runs the denied command
```

So a deny silently stops applying during the upgrade window. It fails the other way too: `--host openclaw "git push && rm -rf /tmp/zz"` on the old binary exits 2, which `index.ts` turns into a hard block. A new plugin against an older rtk both over-blocks and under-blocks depending on the command text.

This one is about binaries already installed, so it needs handling whichever way the deny question goes — a version probe before passing the flag, or wording that makes upgrade order explicit.

KuSh (changes requested):

**Claim**: `rtk rewrite` gains `--host`, and OpenClaw is registered as a host that is its own permission authority, so `Ask`/`Default` collapse to `Allow` and the plugin stops raising a second approval prompt sourced from Claude Code's settings.

**Scope**: accept. #3908 is a real problem, defaulting to `claude` keeps every other adapter untouched, and the PR body is unusually precise about what changes and why. The blocking item below is one class, and once it is closed this is mergeable.

**Ran**: `rtk rewrite` on both binaries (merge-base `8e9aa04c` vs head `320e28de`, one `CARGO_TARGET_DIR` each, `md5sum` confirmed they differ) × 21 cases — no-flag baseline, `--host claude`, `--host openclaw`, an unknown host, a hyphen-typo host, a missing flag value, flag-after-command, unattestable construct, non-rewritable command, compound commands, and deny/allow/default verdicts driven from a real `CLAUDE_CONFIG_DIR/settings.json`. Invariants: every arm of the new decision on both hosts, both sides of the version boundary, and the full chain an agent walks (ask rtk, then run what it hands back). Savings: not applicable — no filter changes, so there is no bash-output ratio to measure against the 20% floor.

**Matrix**: executed this round (first time on this PR).

**CI** @`320e28de`: never run — only `license/cla` (pass) and `check-target` (skipping). The workflow is waiting for maintainer approval, so this head has not been built or tested on any platform. My local gate is clean (`cargo fmt --check`, `clippy --all-targets` 0 warnings, 3413 tests) but that is evidence about this machine, not about Windows or macOS.

### Blocks merge (1 class, 2 sites)

**`--host openclaw` loses the deny gate — on both sides of the version boundary.** Two sites, one cause: whenever the flag is in play, a command the user explicitly denied stops being denied. Both verified end to end against real `settings.json` rules, not synthetic inputs.

**Site 1 — new rtk: `Deny` is unreachable, so it becomes auto-apply.** `load_rules_for` gives `Host::OpenClaw` an empty triple, so `check_command_for` can never *return* `Deny` for it (your own `test_openclaw_verdict_is_default_not_allow` pins that), and `Default` then collapses to `Allow`. With `{"permissions":{"deny":["Bash(du *)","Bash(git push *)"]}}`:

| command | merge-base (no flag) | head `--host claude` | head `--host openclaw` |
|---|---|---|---|
| `du -sh .` | exit 2 | exit 2 | **exit 0 → `rtk du -sh .`** |
| `git push` | exit 2 | exit 2 | **exit 0 → `rtk git push`** |

and with `requireApproval` removed the plugin applies that without a prompt:

```console
$ rtk rewrite --host openclaw "du -sh ."
rtk du -sh .            # exit 0
$ rtk du -sh src
5,0M    src             # the denied command ran
```

Compounds inherit it: `--host openclaw "git push && rm -rf /tmp/zz"` → exit 0, stdout `rtk git push && rm -rf /tmp/zz`.

This is a regression against `develop`, not only a change of policy: today's shipped plugin calls `rtk rewrite <cmd>` with no flag and branches on `status === 2`, so an OpenClaw user has deny protection now and would lose it silently.

**Site 2 — older rtk, during the upgrade window: `Deny` becomes passthrough.** `openclaw/README.md:83` says that on an older rtk "rewriting silently stops and no command is blocked or denied". The first half holds; the second does not, in both directions. With `deny: ["Bash(git push *)"]` on the merge-base binary:

```console
$ rtk rewrite "git push origin main"                    # old plugin argv
exit 2                                                   # blocked
$ rtk rewrite --host openclaw "git push origin main"    # new plugin argv
exit 1                                                   # passthrough — OpenClaw runs the original denied command
```

The flag is swallowed into the command string, so the text that reaches the permission gate is `--host openclaw git push origin main`, which no longer matches `Bash(git push *)`. The user's deny rule silently stops applying. It also fails the other way — `--host openclaw "git push && rm -rf /tmp/zz"` on the old binary exits **2**, which `index.ts` turns into a hard block — so a new plugin against an older rtk both over-blocks and under-blocks depending on the command text.

I read site 1 as unintended rather than by design, because `evaluate_with_verdict_for` short-circuits `Deny` first for *every* host — the intent is visibly that deny survives; the empty rule source is what defeats it. The fix is a design call I would rather you make, since "consult Claude's deny list for OpenClaw" is the exact coupling this PR exists to remove:

- **(a)** keep `Deny` host-independent — give `Host::OpenClaw` a deny source (Claude's, or one the plugin passes) and leave ask/allow empty, so the collapse applies only to `Ask`/`Default`; or
- **(b)** accept that RTK asserts nothing for this host, and say so everywhere it currently claims otherwise — drop the unreachable exit-2 branch from `index.ts`, and correct `supported-agents.md:168` ("a deny verdict from RTK still blocks the tool call") and the `index.ts` protocol table, which today advertise a safety net that cannot fire.

Site 2 needs handling either way, since it is about older binaries already in the wild: a version probe before passing the flag, or wording that tells users upgrade order matters. A test asserting that no host and no flag arrangement can turn a denied command into `Allow` or passthrough is what closes the class.

### Needs your answer before I can call the design settled

**`openclaw/index.ts:133` — does OpenClaw's own policy actually see the command it is gating?** The justification for removing `requireApproval` is that OpenClaw enforces its own exec policy. What I can verify is that RTK hands back the *rewritten* command — `rtk git push`, not `git push` — which `openclaw/README.md:77` also states. What I cannot verify here is how OpenClaw matches its own `security`/`ask` rules against that string. If those rules are written against the user's command, a rule naming `git push` will not match `rtk git push`, and the net effect for rule-based configs is that RTK's gate is removed *and* the replacement gate is blinded; only `tools.exec.mode=full` users would be unaffected. You know that runtime better than I do — if OpenClaw matches on the rewritten string, this concern is void and I will drop it.

- **`src/main.rs:1785` — the unknown-host fallback does not fail closed for the caller.** A typo resolves to Claude's rules, and the stderr warning changes no exit code any adapter acts on. Verified: `--host open-claw "git push && rm -rf /tmp/zz"` → exit 2, so the plugin blocks the call under a rule set the OpenClaw user never chose; with no matching rule, `--host open-claw git push` → exit 3, which the new plugin now applies silently. "No rules, not an authority" — or a dedicated exit 1 — expresses the failure better than borrowing another host's rules.
- **`src/hooks/permissions.rs:41` — the exit-2 rationale is incomplete.** The comment explains that `ValueEnum` was avoided because clap exits 2 on an unknown value and the protocol reserves 2 for deny. Clap still exits 2 on a *missing* value: `rtk rewrite --host` → exit 2, `error: a value is required for '--host <HOST>'`. A caller that builds argv from config and drops an empty value gets a spurious deny.
- **`src/hooks/permissions.rs:36` — `HOST_NAMES` is tested in one direction only.** Verified by mutation: deleting `"openclaw"` from it leaves all 3413 tests green, so the flag's own host would silently vanish from the help text and from the "expected one of" warning. There is a third hardcoded copy of the list in the `--host` help at `src/main.rs:941`. This bites at rebase — `develop`'s `Host` already carries `Codex` and `Trae`, which `from_cli_name` does not know, so `--host codex` would warn and fall back to Claude. The compiler catches the missing `load_rules_for` arm but not the missing name. `AgentPath::AGENTS` on develop already tests both directions and is the pattern to copy.

- No-flag behaviour is byte-identical to the merge-base: `git status` → 3, `ls -al` → 0 under an allow rule, `rm -rf /tmp/x` → 2 under a deny rule. The backward-compatibility claim holds.
- `--host claude` preserves the #1155 invariant: `Default` → exit 3.
- `contains_unattestable_construct` still wins for the new host: `--host openclaw "echo $(whoami)"` → exit 1, and a file-target redirect likewise.
- A non-rewritable command (`frobnicate x`) → exit 1 under the new host.
- `--host` after the positional is swallowed into the command, as `test_rewrite_clap_host_after_command_is_part_of_the_command` says; behaviour is identical on both binaries.
- No host-injection path: every shipped adapter passes the command as a single argv element, so an agent-emitted command cannot select a host.

- *"`--host openclaw` lets a denied `rm -rf` through."* Dropped: `rm -rf /tmp/x` exits 1 under every host because it has no RTK rewrite at all, so it never reaches the verdict branch. The finding above uses `du` and `git push` precisely because they are rewritable, which is what makes the deny reachable.
- *"Version skew is harmless."* Dropped as stated, and it became site 2 above. `--host openclaw "git status"` on the old binary does exit 1 with no output, which is what I checked first — but that is the no-matching-rule case, and generalising from it was wrong.

### Rebase note

`develop` has since moved this decision into `src/hooks/decision.rs` — `decide(cmd, verdict) -> HookDecision` (`AllowRewrite`/`AskRewrite`/`Deny`/`Defer`), shared by `rtk rewrite`, `rtk hook <agent>` and `rtk hook check`, with the verdict passed in precisely so each host can consult its own rules (#3146). `evaluate_for`/`evaluate_with_verdict_for` duplicate that, and all three source files conflict. Threading a host through `decide` is likely smaller than the current shape, and puts `--host` next to `AgentPath::lookup` instead of a third name vocabulary. Worth targeting current `develop` rather than carrying the old shape forward. #3900 also edits `rewrite_cmd.rs` from the same pre-refactor base, and there is unmerged work on `permissions.rs` command segmentation, so whichever lands first moves the ground under the others.

### Next

Pick (a) or (b) for the deny question and handle the upgrade window; that is the whole of what stands between this and an approval. The three optional items and the rebase can ride along. I'll re-run the deny matrix and the round-1 rows on the next push. I have not approved a CI run on this head, since the rebase means it will be replaced anyway — CI will get its first real go at the next push, and that is also the first time any of this is exercised on Windows or macOS.

Rounds: 1/3. Threads: 1 open (deny under `--host`), 0 resolved.

src/hooks/permissions.rs
lucabattistini: Fixed in `26cb233` and clarified in `e8e9fe8`.

I kept the deny gate. `rtk rewrite` still evaluates the existing permission rules, while `ApprovalOwner::Delegate` converts only `AskRewrite` to `AllowRewrite`. `Deny` remains exit 2, and the OpenClaw plugin still blocks it.

The new deny-matrix tests cover `du`, `git push`, and compound commands for both RTK-owned and delegate-owned approval.

openclaw/README.md
lucabattistini: Fixed in `26cb233`.

The plugin no longer passes the host in argv. It sets `RTK_REWRITE_HOST=openclaw` instead. Older RTK versions ignore the unknown environment variable and receive the original command unchanged, so existing deny rules keep matching. New RTK versions use the variable only to transfer approval ownership; the deny gate remains active.

The compatibility tests cover old and new behavior, including denied simple and compound commands. No minimum RTK version or upgrade order is required.

lucabattistini:
Round 1 is addressed. Head is now `bc0c06b`: same diff, squashed onto current `develop` (`0c279f5`) as one commit.

**Deny class, site 1 — option (a).** RTK keeps the deny gate; only approval moves. The rule source is untouched — `AgentPath::verdict` still calls `check_command_for(cmd, Host::Claude)` on every `ViaRewrite` path — and the whole relaxation is `ApprovalOwner::Delegate`, which matches on `AskRewrite` alone, so `Deny` and `Defer` are structurally out of its reach. Your matrix is now a test: `du -sh .`, `git push` and `git push origin main` exit 2 with no host and under all six names, and `git push && rm -rf /tmp/zz` exits 2 whole. No exit-2 branch is left advertising a gate that cannot fire.

**Site 2 — the environment, not argv.** `--host` is gone entirely; the plugin sets `RTK_REWRITE_HOST=openclaw`. An rtk that predates the variable ignores it, so the command reaches the gate untouched and your `Bash(git push *)` case keeps matching — a prompt, not a passthrough. No minimum version, no upgrade order. `the_host_is_not_an_argv_token_so_versions_cannot_disagree` pins the flag-form corruption you measured, and pins that RTK ships no argv token for the host at all.

**Your open question — does OpenClaw's own policy see the command it is gating?** For the exec tool's own checks, yes: `tools.exec.mode`, `tools.exec.security`, `tools.exec.ask` and the exec-approvals allowlist match the string the exec tool receives, which is `rtk git push`, because OpenClaw carries hook adjustments forward into the parameters it passes to exec. That was already true on the exit-0 allow path before this PR, so this makes the docs say it rather than changing behaviour (`openclaw/README.md`, "Writing exec rules"). The one gate that sees the original command is a trusted tool policy (`api.registerTrustedToolPolicy(...)`), which runs before ordinary `before_tool_call` hooks; that is documented as the exception. If the case behind your question was a rule-based config whose rules name the raw command, that guidance is the mitigation, and the exit-2 branch is unaffected either way.

The three optional items fall out by construction: the unknown-name path is the stricter default rather than another host's rules (`from_env` fails closed for unknown, empty, mixed-case and in-process names, all tested); there is no argv flag left for clap to reject with exit 2; and the names resolve through `AgentPath::lookup`, one vocabulary, which `AgentPath::AGENTS` already tests in both directions.

On `#3953`: nothing here changes recall accounting beyond what develop already does, except that a denied command still records no read — now pinned for the self-approving host too.

Two things I cannot do from here. The workflow run on this head is still waiting for maintainer approval, so `bc0c06b` has not been built on any platform yet — if you can approve the run, that would close it. And your re-run of the deny matrix is the evidence that counts here, not my tests.

Local gate on `bc0c06b` (macOS arm64, rustc 1.97.1): `cargo fmt --all -- --check` clean; `cargo clippy --all-targets` no warnings; `hook_decision_protocol_test` 33 passed, 0 failed; the `decision` unit tests 35 passed, 0 failed.

src/hooks/permissions.rs
lucabattistini: Head is now `bc0c06b`, one commit on `develop` `0c279f5`; the `--host` implementation from `26cb233` was replaced entirely. Option (a) is what landed: the rule source is unchanged, and only `AskRewrite` relaxes. Summary in the round-2 comment.

openclaw/README.md
lucabattistini: Head is now `bc0c06b`, one commit on `develop` `0c279f5`. The argv flag is gone: the plugin sets `RTK_REWRITE_HOST=openclaw`, which an older rtk ignores, so the command reaches the gate untouched and `Bash(git push *)` keeps matching. No minimum version, no upgrade order. Summary in the round-2 comment.

lucabattistini:
Full gate on `bc0c06b`: `cargo test --all` → **3934 passed, 0 failed, 8 ignored**; `cargo fmt --all -- --check` clean; `cargo clippy --all-targets` no warnings. Body updated with the same.

PR #3923, opened 2026-09-08
test(diff): pin the shape detectors, the structural index lines and the GNU roots

KuSh:
A fresh mutation run against `develop` (85 compiling mutants, 70 killed — 82%) put
the surviving gaps in three places. This pins all three, and factors out the one
that could not be tested where it stood.

### The shape detectors reject nothing

Every fixture in the diff corpus is a real capture, so the detectors only ever see
well-formed lines. Loosening `&&` to `||`:

- `is_mbox_from` — any `From `-prefixed line over 40 characters opens an mbox
  message region. Ordinary commit prose reaches that shape.
- `is_hg_echo` — a non-hex revision (`diff -r zzzzzzzzzzzz …`) reads as an hg echo.
- `is_submodule_range` — a short or non-hex side of a `<hex>..<hex>` range passes.

None failed a test. `detectors_reject_near_misses` is the near-miss table.

### git's `index` lines are structure, not prose

Collapsing the `index ` / `similarity index ` / `dissimilarity index ` chain to a
single term left every test green, because `dropped_line` is only consulted once a
GNU echo has been seen — and no fixture carries both producers. A stream that does:

```
diff -ru a/z.txt b/z.txt        →  develop:  [file] b/z.txt (+1 -1)
...                                          [file] g.txt   (+1 -1)
diff --git a/g.txt b/g.txt
index 1234567..89abcde 100644   →  collapsed: raw passthrough, whole stream
```

Safe (the fallback is raw) but a silent savings regression on a section that parses.

### The GNU roots were collected twice

The same eight lines were inlined in the header-pair arm and again in the
`Binary files` arm. Mutating either copy changed no output, because nothing reaches
that logic directly — the `Only in` split resolves through paths that do not depend
on it. Extracted as `record_gnu_roots`, which is the point of the factorisation:
the dedup and the non-empty guard become assertable, and both are now pinned. A
third arm growing a third copy is no longer the default.

### No behaviour change

Byte-identical to `develop` across the 81 diff fixtures, 300 real `git show`
streams, and GNU directory diffs over names holding `: ` and `'`. Each new test was
verified to fail against the mutation it exists for, and against nothing else.

Also rewrites `condense_stdin`'s doc comment, which named a rejected alternative,
described `guard::never_worse`'s signature from a file that does not call it, and
repeated the ANSI-view clause already carried by `condense_unified_diff_strict`'s
own doc — per CONTRIBUTING.md's "No obvious comments".

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -487,15 +487,9 @@ pub fn run_stdin(_verbose: u8) -> Result<()> {
     Ok(())
 }
 
-/// Filter a piped stream: parse strictly (the parser reads structure through
-/// an ANSI-stripped view, so a `git diff --color` stream parses instead of
-/// condensing to silence, while content lines keep their bytes) and apply
-/// the never-worse check — inlined rather than via `guard::never_worse`,
-/// which hands back the winning `&str`, where this path needs `None` to
-/// mean byte-exact fallback. `None` means the caller must emit its exact
-/// input bytes — including for non-UTF-8 input, where filtering would
-/// rewrite the user's content bytes to U+FFFD (byte fidelity outranks
-/// savings here).
+/// Filter a piped stream. `None` means the caller emits its input bytes
+/// exactly: non-UTF-8 input takes that path, so content that is not valid
+/// UTF-8 survives byte for byte.
 fn condense_stdin(bytes: &[u8]) -> Option<String> {
     let input = std::str::from_utf8(bytes).ok()?;
     // PowerShell 5.1's `>` writes a BOM; without this the first `diff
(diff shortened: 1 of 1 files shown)


aeppling:
**LGTM.**

Verified behaviour-preserving rather than taking the claim on trust: built the head and its merge base from the same tree and compared `rtk diff -` output byte-for-byte (stdout and exit code) over ~131k inputs:

- the 186 fixtures shipped in `tests/fixtures/`
- ~1,200 real `git show` streams from 400 commits of this repo, in plain, `--color=always` and `--submodule=log` form
- ~80k generated git / GNU `diff -r` / hg / svn / mbox streams, weighted toward the shapes this PR touches, mutated with CRLF, BOM, ANSI, mid-stream truncation and invalid UTF-8

Zero divergences.

The corpus has proven power over the code you touched, not just coverage of it: mutating the `index` chain to `&&` diverges 73x, `is_mbox_from` to `||` diverges 38x.

aeppling (approved):

PR #3937, opened 2026-09-08
fix(rewrite): process wrapper prefixes

### LINKED ISSUE #2375: Hook rewrite ignores process-wrapper prefixes (`timeout`, `time`) so wrapped commands run raw/unfiltered

aeppling:
Closes #2375. Supersedes #2568, whose wrapper set and peel-recurse-reprepend shape this builds on (@YOMXXX).

## What changes

`timeout 300 cargo test` now rewrites to `timeout 300 rtk cargo test`. Same for `time`, `nice`, `nohup`, and path-qualified forms like `/usr/bin/timeout`.

`PROCESS_WRAPPERS` in `registry.rs` describes each wrapper's own arguments (value options, flag options, attached values like `-k5s` and `-n10`, positionals such as timeout's duration). One generic scanner walks them in order. Adding a wrapper is a table entry.

Two rules keep it safe:

- An unknown option drops the rewrite, since it may consume the next word and shift which token looks like the command.
- Shell syntax before the command (redirect, subshell, glob, command substitution) drops the rewrite, since the argv can no longer be read from the token list.

`stdbuf` is excluded on purpose. It exists to get incremental output, and rtk buffers until the child exits, so rewriting it removes the reason to type it.

## Signal handling

Wrapping changes who gets signalled: `timeout 300 rtk cargo test` signals rtk, not cargo. Before this PR rtk died on the default disposition and every captured line was lost, which is worse than running unwrapped.

`core::stream` now relays SIGINT/SIGTERM to the child and lets the normal filter-and-print path finish. If the child does not exit, it escalates to SIGKILL after 1.5s, then restores the default disposition and re-raises after another 1.5s, so rtk can never be held open.

## How this was tested

Shim-based execution tests, with a fake `cargo` on PATH that records its argv:

- argv and exit codes are identical between the raw and rewritten forms for all four wrappers, for success, exit 101 and exit 7.
- output under a kill, shim prints 200 lines then hangs, killed at 2s:

| | exit | lines seen |
|---|---|---|
| `timeout 2 cargo clippy` | 124 | 200 |
| `timeout 2 rtk cargo clippy` before | 124 | 0 |
| `timeout 2 rtk cargo clippy` after | 124 | 9 filtered, diagnostic kept |

- `timeout --foreground` no longer orphans the child.
- stubborn child that defers SIGTERM: bounded at 3s instead of hanging 300s, output still flushed.

Differential fuzz against develop, 6,567 generated command lines through both binaries:

| bucket | count |
|---|---|
| identical | 4,555 |
| newly filtered | 1,534 |
| changed (compound, wrapped half now rewrites) | 478 |
| lost rewrite | 0 |
| crash or hang | 0 |
| shape violation | 0 |

Shape check asserts a rewrite only ever inserts `rtk` tokens, never reorders or drops shell text. Rewriting is idempotent, so the hook can re-run over its own output.

`cargo fmt --all`, `cargo clippy --all-targets`, `cargo test --all`: clean, 3341 unit tests plus all integration targets. `tests/signal_flush_test.rs` fails without the relay.

## Needs a decision

`test_toml_filter_rewrites_bare_command_but_not_wrapped_invocations` asserted that a wrapped TOML-filtered command is never rewritten. That pinned the absence of wrapper peeling, which is what this PR adds, so it now asserts the wrapper-preserving rewrite and is renamed. What keeps a filter off the wrapper itself is the `is_fully_anchored` gate on `match_command`, unchanged and still tested in `core::toml_filter`.

### DIFF
--- a/src/core/runner.rs
+++ b/src/core/runner.rs
@@ -160,6 +160,19 @@ where
 }
 
 pub fn run(
+    cmd: Command,
+    tool_name: &str,
+    args_display: &str,
+    mode: RunMode<'_>,
+    opts: RunOptions<'_>,
+) -> Result<i32> {
+    let result = run_inner(cmd, tool_name, args_display, mode, opts);
+    // #2375
+    stream::die_by_relayed_signal();
+    result
+}
+
+fn run_inner(
     mut cmd: Command,
     tool_name: &str,
     args_display: &str,
--- a/src/core/stream.rs
+++ b/src/core/stream.rs
@@ -317,6 +317,13 @@ mod signal_relay {
         }
     }
 
+    pub fn relayed() -> Option<libc::c_int> {
+        match RELAYED.load(Ordering::SeqCst) {
+            0 => None,
+            sig => Some(sig),
+        }
+    }
+
     pub struct Relay;
 
     impl Relay {
@@ -374,6 +381,23 @@ mod signal_relay {
         }
     }
 }
+
+// #2375
+#[cfg(unix)]
+pub fn die_by_relayed_signal() {
+    let Some(sig) = signal_relay::relayed() else {
+        return;
+    };
+    #[allow(unsafe_code)]
+    // nosemgrep: unsafe-block
+    unsafe {
+        libc::signal(sig, libc::SIG_DFL);
+        libc::raise(sig);
+    }
+}
+
+#[cfg(not(unix))]
+pub fn die_by_relayed_signal() {}
 
 pub fn status_to_exit_code(status: std::process::ExitStatus) -> i32 {
     if let Some(code) = status.code() {
--- a/src/core/stream.rs
+++ b/src/core/stream.rs
@@ -268,6 +268,106 @@ impl StreamResult {
     }
 }
 
+// #2375
+#[cfg(unix)]
+mod signal_relay {
+    use std::sync::atomic::{AtomicBool, AtomicI32, AtomicU32, Ordering};
+    use std::thread;
+    use std::time::Duration;
+
+    const POLL: Duration = Duration::from_millis(50);
+    const KILL_GRACE: Duration = Duration::from_millis(1500);
+    const EXIT_GRACE: Duration = Duration::from_millis(1500);
+
+    static CHILD_PID: AtomicU32 = AtomicU32::new(0);
+    static RELAYED: AtomicI32 = AtomicI32::new(0);
+    static FINISHED: AtomicBool = AtomicBool::new(false);
+
+    #[allow(unsafe_code)]
+    unsafe extern "C" fn relay(sig: libc::c_int) {
+        let pid = CHILD_PID.load(Ordering::SeqCst);
+        if pid == 0 || RELAYED.swap(sig, Ordering::SeqCst) != 0 {
+            libc::signal(sig, libc::SIG_DFL);
+            libc::raise(sig);
+            return;
+        }
+        libc::kill(pid as libc::pid_t, sig);
+    }
+
+    fn escalate(pid: u32) {
+        thread::sleep(KILL_GRACE);
+        if FINISHED.load(Ordering::SeqCst) {
+            return;
+        }
+        #[allow(unsafe_code)]
+        // nosemgrep: unsafe-block
+        unsafe {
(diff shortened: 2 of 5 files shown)


pszymkowiak (approved):
Nice work on this one. The wrapper peeling is fail-closed where it counts: an option the table doesn't describe drops the rewrite rather than guessing which token is the command, and the option-form coverage (`-k 5s`, `-k5s`, `--kill-after=5s`, `--`, `nice -10`/`+5`) is the part that usually breaks on the first real user. Leaving `stdbuf` out is the right call and the README says why.

The signal relay is the half I looked at hardest, and it holds up. The handler only touches `kill`, `signal`, `raise` and atomics, so it stays async-signal-safe, and the drop order means the handlers come off before the child is reaped. `signal_flush_test.rs` tests the behaviour rather than the wiring, which is what makes it worth having.

Approving. Three things I'd like noted rather than fixed here, unless you disagree:

1. The relay installs on every `run_streaming`, not just wrapped invocations, so Ctrl-C on any rtk command now behaves differently (child gets SIGINT twice, rtk survives to print). That's probably what we want, but it deserves a line in the CHANGELOG since the PR title doesn't hint at it.

2. `CHILD_PID` / `RELAYED` / `FINISHED` are process-wide, so this quietly assumes one streaming run at a time. True today, but nothing in the code says it has to stay true. A short comment would save someone a bad afternoon later.

3. `test_toml_filter_rewrites_bare_and_wrapped_invocations` widens the blast radius to user TOML filters, with `is_fully_anchored` as the only thing keeping a filter off the wrapper itself. You documented it in the test comment, which is the right instinct - I'm just flagging it so it's visible in the review history.

PR #3938, opened 2026-09-08
fix(diff): make stdin raw fallback byte-exact (#3924)

### LINKED ISSUE #3924: rtk diff - raw fallback is not byte-exact: a newline is appended when the input has none
`rtk diff -` falls back to raw passthrough whenever the parser cannot agree with the
stream's structure, and that fallback is documented as emitting the input bytes exactly —
`condense_unified_diff_strict`'s contract is "`None` means the caller must emit its exact
input bytes".

It does not, when the input has no trailing newline. Verified on `develop` (665b2720):

```
$ printf 'not a diff at all' > in.txt
$ wc -c < in.txt
17
$ rtk diff - < in.txt | wc -c
18
```

The extra byte is a `\n` appended by `run_stdin`:

```rust
out.write_all(&bytes)?;
if !bytes.is_empty() && !bytes.ends_with(b"\n") {
    writeln!(out)?;
}
```

Input that already ends with a newline round-trips byte-exact, so this is specific to a
stream whose last line is unterminated — a truncated capture, `printf` without `\n`, a
producer killed mid-write. Those are exactly the streams most likely to hit the raw
fallback in the first place.

Two ways to close it, and the choice is a real one:

- make the fallback byte-exact and drop the `writeln!`, which matches the documented
  contract but can leave a shell prompt mid-line;
- keep the newline and narrow the contract wording to "the input bytes, newline-terminated",
  so the guarantee stated in the parser doc and in #3788 matches what ships.

Either way it needs a test. Nothing currently covers it: mutation testing over
`diff_cmd.rs` left three mutants alive on this exact condition (`!bytes.is_empty()`,
the `&&`, and the `!`), all in `run_stdin`. A unit test cannot reach it because the
behaviour is in the write path, so it belongs in `tests/` as a real-process test asserting
the output bytes for both a newline-terminated and an unterminated input.

anujsingh-cse:
Closes #3924

## Problem

\
tk diff -\'s structural fallback appends a trailing newline when the piped input has none, violating the documented contract that the raw fallback emits the caller's exact input bytes (\condense_stdin\ returns \None\ → caller emits input bytes verbatim).

Repro:
\\\
printf 'not a diff at all' | rtk diff - | wc -c   # 18 (input was 17)
\\\

Input that already ends with \\n\ round-trips byte-exact, so the bug only affects unterminated streams — exactly the truncated captures / dead producers most likely to hit the fallback.

## Fix

Drop the \writeln!\ in \
un_stdin\ so the fallback emits the exact input bytes with no additions. This follows the issue's \yte-exact\ option, matches the doc contract in \condense_stdin\, and aligns with RTK's 'never adds bytes a native pipe would not produce' / byte-fidelity philosophy for unfiltered passthrough.

## Tests

New process-level test \	ests/diff_stdin_byte_accuracy_test.rs\ (as the issue requests — a unit test can't reach the write path):

- unterminated input round-trips byte-for-byte
- newline-terminated input round-trips byte-for-byte
- non-UTF-8 bytes pass through untouched (byte fidelity for the raw branch)

Verified the byte-exact test fails on \develop\ (appends \\n\) and passes with this change. \cargo clippy --all-targets\ and \cargo fmt --check\ are clean; all diff-related tests pass (233 unit + 5 integration + 3 new).

### DIFF
--- a/src/cmds/git/diff_cmd.rs
+++ b/src/cmds/git/diff_cmd.rs
@@ -470,15 +470,13 @@ pub fn run_stdin(_verbose: u8) -> Result<()> {
             );
         }
         None => {
-            // Structural fallback: the caller's exact bytes (plus println!
-            // parity — a terminating newline when the input lacked one).
+            // Structural fallback: emit the caller's exact input bytes,
+            // byte-for-byte. condense_stdin's None contract is "exact input
+            // bytes", and RTK never adds bytes a native pipe would not produce.
             use std::io::Write;
             let mut out = io::stdout();
             out.write_all(&bytes)
                 .context("Failed to write raw diff to stdout")?;
-            if !bytes.is_empty() && !bytes.ends_with(b"\n") {
-                writeln!(out).context("Failed to write raw diff to stdout")?;
-            }
             let raw = String::from_utf8_lossy(&bytes);
             timer.track("diff (stdin)", "rtk diff (stdin)", &raw, &raw);
         }
--- a/tests/diff_stdin_byte_accuracy_test.rs
+++ b/tests/diff_stdin_byte_accuracy_test.rs
@@ -1,0 +1,68 @@
+//! Byte-exactness of `rtk diff -`'s structural fallback.
+//!
+//! When the piped stream is not a parseable diff, `rtk diff -` falls back to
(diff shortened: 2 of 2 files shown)


anujsingh-cse:

Issue #3702, opened 2026-08-24
`rtk gain` cannot represent a regression: saturating_sub floors negative savings to zero


randomparity:
## Summary

`Tracker::record` computes savings with `saturating_sub` on `usize`, so when RTK's output
is **larger** than the baseline the result is floored to `0`. The schema has no
representation for "RTK made this worse": `saved_tokens` can never be negative, and no
aggregate — `rtk gain`, `--daily`, `--weekly`, `--graph`, `--quota`, `--format json` —
can surface a regression.

RTK does inflate some outputs, so these cases exist and are currently invisible.

## Location

`src/core/tracking.rs:425` (present on `master` @ `b34be37` and in `v0.45.0`):

```rust
let saved = input_tokens.saturating_sub(output_tokens);
let pct = if input_tokens > 0 {
    (saved as f64 / input_tokens as f64) * 100.0
} else {
    0.0
};
```

Both operands come from `estimate_tokens` at `src/core/tracking.rs:1394`.

## Evidence that regressions exist

Measured on a pinned Rust repository (457 `.rs` files), comparing raw output bytes against
`rtk <cmd>` output bytes:

| command | raw | rtk | change |
|---|---:|---:|---:|
| `ls src` | 251 | 341 | **+35%** |
| `ls -R src` | 6,926 | 8,956 | **+29%** |
| `cargo check --workspace --all-targets` | 72 | 100 | +38% |

Each of these records `saved_tokens = 0, savings_pct = 0.0` — indistinguishable from a
command where RTK correctly declined to act.

Verified against the tracking DB directly:

```console
$ sqlite3 ~/Library/Application\ Support/rtk/history.db \
    "SELECT COUNT(*) FROM commands WHERE saved_tokens < 0;"
0
$ sqlite3 ~/Library/Application\ Support/rtk/history.db \
    "SELECT COUNT(*) FROM commands WHERE output_tokens > input_tokens;"
0
```

Zero by construction, not by observation.

## Why it matters

The aggregate savings number is the project's headline metric and the basis of the
`--quota` estimate. Flooring regressions to zero biases it upward: a deployment that is
net-negative for a given user cannot show that anywhere in the tooling. It also hides the
signal maintainers would most want — which handlers regress, and on what inputs.

## Proposed fix

**1. Store savings signed.** Change `saved_tokens` to `i64` and let `savings_pct` go
negative:

```rust
let saved = input_tokens as i64 - output_tokens as i64;
[...]

Issue #2805, opened 2026-07-03
Savings accounting credits rtk with the agent's own head/tail filtering — repro: 2.1M tokens claimed '100% saved' vs ~195 real (>10,000×); likely root cause of #2762


bensonwong:
## TL;DR — this is a hard one, please read it all the way through

rtk's savings accounting credits rtk with filtering **the agent already did itself**. One sandboxed command on `develop` (31f9d43) reproduces a **>10,000× overstatement**: `rtk gain` reports **2.1M tokens saved (100.0%)** for a command whose real marginal saving is **~195 tokens**.

We're raising this assuming good faith all the way down — this looks like a classic Goodhart trap (the metric was easy to record, so it became the product), not anything intentional. But the consequence is bigger than a bug: the headline numbers — the README "Token Savings (30-min Claude Code Session)" table, the `-99%` site demo, and every `rtk gain` screenshot — inherit this baseline and are not defensible in their current form. Fixing this properly means fixing the accounting **and** remeasuring, then updating README.md and the marketing copy on rtk-ai.app to match what rtk actually delivers. We think there *is* a real, honest value prop underneath (details at the bottom), but it is much smaller than currently claimed, and the current claims are costing you trust with exactly the users who measure.

This is very likely the root cause of #2762 ("~90M tokens saved (100%), mostly `rtk read`, 0% reproducible compression").

## Reproduction (sandboxed, `develop` @ 31f9d43)

```bash
# 100k-line log file, 8.3 MB
seq -f "2026-07-03T10:00:00 INFO service=api request completed status=200 latency_ms=%.0f" 1 100000 > big.log
export RTK_DB_PATH="$PWD/rtk-repro.db"   # isolated tracking DB

# 1. The agent runs `head -20 big.log`. The hook rewrites it:
rtk rewrite "head -20 big.log"
# → rtk read big.log --max-lines 20   (exit 3: rewrite + ask)

# 2. What the agent would have consumed WITHOUT rtk (its own filter):
head -20 big.log | wc -c                   # 1591 bytes  ≈ 398 tokens

# 3. What rtk emits:
rtk read big.log --max-lines 20 | wc -c    # 809 bytes   ≈ 203 tokens

# 4. What rtk claims it saved:
rtk gain
```

```
Total commands:    1
Input tokens:      2.1M
Output tokens:     203
Tokens saved:      2.1M (100.0%)
Efficiency meter: ████████████████████████ 100.0%
```

[...]

PR #3943, opened 2026-09-08
fix(tracking): cap token estimates to stop savings-figure inflation

pszymkowiak:
## Summary

`rtk gain`'s "tokens saved" headline can be inflated by orders of magnitude by a single oversized command, because nothing bounds a per-call token estimate and the counterfactual assumed is "the agent would have consumed the full raw output" — it wouldn't have, since agent shells (Claude Code, etc.) truncate captured tool-result output well below what RTK's estimator counts.

**Confirmed against a real local `history.db`**: a single `rtk read` call recorded 135,722,224 input tokens against a 21-token real output — one row alone accounting for ~92% of that install's entire lifetime "saved" total (135.7M of 148.2M). Invisible in the by-command average (5.3%, unweighted) since 4,627 of 5,180 `rtk read` calls genuinely saved 0%, and only 158 outlier rows (3%) supplied 99.86% of the sum.

This combines and adapts several stale/unmerged community PRs (#3635, #1978, #891, #2699) that each independently targeted pieces of this — none directly mergeable (conflicting with current `develop`, zero maintainer review).

## What's in this PR

1. **Write-time token-estimate cap.** New `tracking.estimate_cap_chars` config (default `30000` chars, `0` = uncapped), overridable via `RTK_TRACK_CAP_CHARS`. `TimedExecution::track` clamps `input_tokens`/`output_tokens` to this ceiling before they reach the DB.
2. **Version-gated retroactive migration** (schema `v1` → `v2`) caps already-inflated historical rows to the same ceiling and recomputes `saved_tokens`/`savings_pct` from the capped values. Floors at 0 rather than manufacturing a fake positive saving when capping turns two oversized values into an equal pair, or when a row was already a regression pre-cap.
3. **Weighted savings rate in `get_by_command`** (`Avg%` → `Rate` column): `AVG(savings_pct)` → `SUM(saved_tokens)*100.0/SUM(input_tokens)`. Fixes per-command stats being diluted by many 0%-savings calls (the exact mechanism behind the 5.3%-vs-138.9M discrepancy above).
4. **`read.rs` baseline windowing**: `--max-lines`/`--tail-lines` now window the *raw tracking baseline* the same way the filtered output already is, via a new `tracking_baseline()` helper used only by the tracking call — content outside an explicit window is the user choosing to see less, not RTK compressing anything, and was previously counted as a fake "saving" (a single bioinformatics-file read this way reported 250M+ phantom tokens in the wild — see #2805/#1045).

## Review process (important — read this)

This PR went through **two rounds** of independent multi-agent review (5 parallel adversarial agents in one workflow, plus an independent `codex` review, plus an independent `vibe`/Mistral pass) before landing. That review caught **real bugs I introduced while adapting the stale community PRs above** — worth being transparent about, since it's exactly why the extra verification mattered:

- **codex caught a real regression** in my first adaptation of #2699: the windowed raw baseline was accidentally also feeding `never_worse()`/`print!()`, i.e. changing what gets *displayed* to the user, not just what gets tracked (comparing two same-budget windowed strings could flip the safety-net into printing an unfiltered excerpt). This bug was already present in the stale PR I adapted from — its own tests never exercised the display path. Fixed by splitting the concern: `tracking_baseline()` returns a single value used only by `timer.track`, structurally unable to reach `print!`.
- **The 5-agent workflow converged (multiple independent agents, same finding)** on two more real bugs in the retroactive migration itself:
  - The migration `UPDATE`'s `Result` was discarded (`let _ =`), then `user_version` was stamped unconditionally right after — a transient failure (SQLITE_BUSY, disk pressure) would silently mark the DB as migrated forever while the exact rows this PR exists to fix stayed uncapped, with no error surfaced. Now propagated with `?`.
  - `rtk init` → `ensure_schema_fresh()` → `MigrationMode::Always` re-invokes the whole migration function unconditionally (to self-heal a dropped table) — which also unconditionally re-ran the retroactive cap UPDATE with whatever cap is *currently* configured. Lowering `RTK_TRACK_CAP_CHARS` and then running `rtk init` again (an ordinary repair action) would silently and irreversibly re-cap — and lose, with no backup — legitimate historical data. Now guarded by reading the on-disk `user_version` directly, independent of which caller invoked the function.
  - Plus two smaller ones: `low_savings_commands` (feeds telemetry) had the identical unweighted-average flaw fixed in `get_by_command`, left unconverted; `estimate_cap_chars()` did a redundant second `config.toml` parse instead of reusing the crate's existing config cache.

All of these are fixed in the latest commit, with regression tests including one that forces a real `SQLITE_BUSY` via a second connection holding an exclusive lock, and one that reproduces the `rtk init`-with-a-new-cap data-loss scenario exactly.

## Deliberately out of scope

- **#3702** — `saturating_sub` silently floors regressions (RTK output bigger than input) to `0` instead of surfacing a negative delta. Needs a signed-delta type change across the schema/struct, not a quick fix.
- **#2805**'s `rtk read`-specific measurement issues beyond the baseline-windowing fix above (hardcoded `"rtk read"` `rtk_cmd` label blocks auditing which flag combinations are affected; a possible OOM/crash-then-stub-return false-positive path). Worth its own targeted PR in `read.rs`.

## Testing

40 new/rewritten unit tests across `tracking.rs`, `config.rs`, and `read.rs`, including:
- A direct regression test replaying the exact 135.7M-token row found in the wild.
- A test forcing a real `SQLITE_BUSY` (second connection, exclusive lock) to prove the migration fails closed and retries, not silently succeeds.
- A test proving `rtk init`-style repeated migration invocation can't re-cap already-migrated data with a new, lower cap.
- A test proving real filter savings *inside* an explicit window are still counted — only content outside the window stops counting.
- An `#[ignore]`-gated integration test running the real binary end-to-end, proving the display-path regression codex caught stays fixed.

Also verified live, beyond the unit tests:
- Direct CLI, real 300MB file, `--max-lines 5`: before the fix, `input=78,643,173 / saved=78,643,108 (99.9999%)`; after, `input=7,500 / saved=7,435 (99.13%)`.
- Through the real Claude Code `PreToolUse` hook via `claude -p` (PATH pointed at this branch's build, scratch DB): `cat huge.log` correctly rewrites to `rtk read huge.log` and records a capped row, no crash.
- 50 real invocations across 10 files (1KB–3MB) × 5 filter modes: 0 invariant violations (`input_tokens`/`output_tokens` never exceed the cap, `saved_tokens` never negative or exceeding input, `savings_pct` always in [0,100]).

```
cargo test --all   # green (3x repeated default-parallel runs, no flakiness)
cargo clippy --all-targets   # clean
cargo fmt --all -- --check   # clean
```

### DIFF
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -62,9 +62,7 @@ fn project_filter_params(project_path: Option<&str>) -> (Option<String>, Option<
     }
 }
 
-use super::constants::{
-    DEFAULT_ESTIMATE_CAP_CHARS, DEFAULT_HISTORY_DAYS, HISTORY_DB, RTK_DATA_DIR,
-};
+use super::constants::{DEFAULT_HISTORY_DAYS, HISTORY_DB, RTK_DATA_DIR};
 
 /// Main tracking interface for recording and querying command history.
 ///
@@ -316,6 +314,14 @@ type CommandStats = (String, usize, usize, f64, u64);
 /// `user_version` triggers exactly one re-run of the full migration sequence, then
 /// the pragma is updated so subsequent opens skip straight past it.
 const SCHEMA_VERSION: i64 = 2;
+/// The version the retroactive token-cap migration (see `run_schema_migrations`)
+/// was introduced at. Checked directly against the on-disk `user_version`
+/// rather than assumed from `SCHEMA_VERSION`, so this one-time data
+/// migration can't be silently re-triggered by a future schema bump that
+/// re-runs `run_schema_migrations` for an unrelated reason, nor by
+/// `MigrationMode::Always` (`rtk init`'s self-heal path) re-invoking the
+/// whole function on an already-migrated DB.
+const TOKEN_CAP_MIGRATION_VERSION: i64 = 2;
 
 /// Create all tables/indexes, run column migrations, and stamp `user_version` to
 /// `SCHEMA_VERSION` for the on-disk tracker DB.
@@ -422,28 +428,65 @@ fn run_schema_migrations(conn: &Connection) -> Result<()> {
     // saturating_sub result) so a row that was a genuine regression before
     // capping (output > input) can't be capped into a fake positive saving —
     // the MAX(..., 0) floor lands on 0 in that case, same as an uncapped
-    // regression already does. Idempotent: re-running has no further effect
-    // once every row is at or below the cap.
-    // `usize::MAX` (cap disabled, RTK_TRACK_CAP_CHARS=0) can't round-trip
-    // through `as i64` without wrapping to -1, which would clamp every row
-    // to -1 instead of skipping the migration -- guard explicitly.
-    let cap = max_tracked_tokens();
-    if cap != usize::MAX {
-        let cap = cap as i64;
-        let _ = conn.execute(
-            "UPDATE commands
-             SET input_tokens = MIN(input_tokens, ?1),
-                 output_tokens = MIN(output_tokens, ?1),
-                 saved_tokens = MAX(MIN(input_tokens, ?1) - MIN(output_tokens, ?1), 0),
-                 savings_pct = CASE
-                     WHEN MIN(input_tokens, ?1) > 0
-                     THEN (CAST(MAX(MIN(input_tokens, ?1) - MIN(output_tokens, ?1), 0) AS REAL)
-                           / MIN(input_tokens, ?1)) * 100.0
-                     ELSE 0.0
-                 END
-             WHERE input_tokens > ?1 OR output_tokens > ?1",
-            params![cap],
-        );
+    // regression already does.
+    //
+    // Two correctness requirements caught in review, both handled by the
+    // `current_version < TOKEN_CAP_MIGRATION_VERSION` guard below rather than
+    // relying on the caller:
+    //
+    // 1. Genuinely one-time, not just "not on the hot path". `rtk init`
+    //    calls `ensure_schema_fresh()` -> `MigrationMode::Always`, which
+    //    deliberately re-invokes this whole function unconditionally (to
+    //    self-heal a table dropped out-of-band) -- table/column
+    //    creation is safe to always re-run (CREATE TABLE IF NOT EXISTS /
+    //    ADD COLUMN are no-ops once present), but this UPDATE is not: if a
+    //    user lowers RTK_TRACK_CAP_CHARS/estimate_cap_chars and then runs
+    //    `rtk init` again (an ordinary repair action), an unconditional
+    //    re-run would silently re-cap -- and permanently lose, with no
+    //    backup -- legitimate historical data that was already correctly
+    //    under the OLD cap. Reading the on-disk version directly (instead
+    //    of trusting the caller's migration mode) makes this immune to
+    //    which path invoked `run_schema_migrations`.
+    // 2. Must not get silently marked "done" on a failed write. The old
+    //    version discarded the UPDATE's Result via `let _ =`, then
+    //    unconditionally stamped `user_version` right after -- so a
+    //    transient failure (SQLITE_BUSY past the busy_timeout under a
+    //    concurrent rtk/hook process, disk pressure, a read-only mount)
+    //    would permanently mark the DB as migrated while the oversized
+    //    historical rows -- the exact ones this migration exists to fix --
+    //    stayed uncapped forever, with no error surfaced anywhere. Now
+    //    propagated with `?`: on failure, `run_schema_migrations` returns
+    //    Err, the pragma update below never runs, `user_version` stays
+    //    stale, and the retroactive cap retries on the next `Tracker::new()`
+    //    (silently no-op-ing tracking for just that one invocation, the
+    //    same fallback every other schema-migration failure already causes
+    //    via the `if let Ok(tracker) = Tracker::new()` pattern call sites use).
+    let current_version: i64 = conn
+        .query_row("PRAGMA user_version", [], |row| row.get(0))
+        .unwrap_or(0);
+    if current_version < TOKEN_CAP_MIGRATION_VERSION {
+        // `usize::MAX` (cap disabled, RTK_TRACK_CAP_CHARS=0) can't round-trip
+        // through `as i64` without wrapping to -1, which would clamp every
+        // row to -1 instead of skipping the migration -- guard explicitly.
+        let cap = max_tracked_tokens();
+        if cap != usize::MAX {
+            let cap = cap as i64;
+            conn.execute(
+                "UPDATE commands
+                 SET input_tokens = MIN(input_tokens, ?1),
+                     output_tokens = MIN(output_tokens, ?1),
+                     saved_tokens = MAX(MIN(input_tokens, ?1) - MIN(output_tokens, ?1), 0),
+                     savings_pct = CASE
+                         WHEN MIN(input_tokens, ?1) > 0
+                         THEN (CAST(MAX(MIN(input_tokens, ?1) - MIN(output_tokens, ?1), 0) AS REAL)
+                               / MIN(input_tokens, ?1)) * 100.0
+                         ELSE 0.0
+                     END
+                 WHERE input_tokens > ?1 OR output_tokens > ?1",
+                params![cap],
+            )
+            .context("retroactive token-cap migration (schema v2) failed")?;
+        }
     }
 
     conn.pragma_update(None, "user_version", SCHEMA_VERSION)?;
@@ -1348,8 +1391,17 @@ impl Tracker {
 
     /// Count commands with low savings (<30%) — filters that need improvement.
     pub fn low_savings_commands(&self, limit: usize) -> Result<Vec<(String, f64)>> {
+        // Weighted rate (SUM(saved)/SUM(input)), not AVG(savings_pct) --
+        // same fix as get_by_command, for the same reason: an unweighted
+        // per-call average lets a high-volume command's real performance be
+        // buried by many small, near-0%-savings calls (or inflated by a few
+        // near-100% outliers among many genuine 0%s), which would make this
+        // telemetry signal mis-flag (or fail to flag) commands purely from
+        // the statistical artifact, not their real weighted performance.
         let mut stmt = self.conn.prepare(
-            "SELECT rtk_cmd, AVG(savings_pct) as avg_sav FROM commands
+            "SELECT rtk_cmd,
+                    CASE WHEN SUM(input_tokens) > 0 THEN SUM(saved_tokens) * 100.0 / SUM(input_tokens) ELSE 0.0 END AS avg_sav
+             FROM commands
              WHERE input_tokens > 0
              GROUP BY rtk_cmd
              HAVING avg_sav < 30.0 AND avg_sav > 0.0
@@ -1725,12 +1777,16 @@ fn estimate_cap_chars() -> usize {
     }
     // Malformed RTK_TRACK_CAP_CHARS (non-numeric, negative) falls through to
     // the config/default cap below, silently — the hook path must stay quiet.
-    static CONFIG_CAP_CHARS_FALLBACK: std::sync::OnceLock<usize> = std::sync::OnceLock::new();
-    *CONFIG_CAP_CHARS_FALLBACK.get_or_init(|| {
-        crate::core::config::Config::load()
-            .map(|c| c.tracking.estimate_cap_chars)
-            .unwrap_or(DEFAULT_ESTIMATE_CAP_CHARS)
-    })
+    // Reuses the crate's single process-wide config cache (`cached_config`)
+    // instead of a second private `Config::load()` -- an earlier version had
+    // its own OnceLock here, which meant every rtk invocation that hits this
+    // (i.e. every one, via TimedExecution::track) read and TOML-parsed
+    // config.toml a second time on top of the read `get_db_path()` already
+    // does via the same cache, on a path documented as latency-sensitive
+    // (every rtk invocation and every PreToolUse hook call).
+    crate::core::config::cached_config()
+        .tracking
+        .estimate_cap_chars
 }
 
 /// Token ceiling derived from `estimate_cap_chars()` (ceil(chars/4), same
@@ -2202,6 +2258,14 @@ mod tests {
 
         let conn = Connection::open_in_memory().expect("open in-memory db");
         run_schema_migrations(&conn).expect("initial schema");
+        // Roll the version back to simulate a real pre-upgrade DB: tables
+        // already exist (from a much older version), but the schema v2
+        // token-cap migration hasn't run yet. Without this, the migration
+        // would correctly see itself as already-applied (version stamped by
+        // the call above) and skip -- exactly the one-time guarantee this
+        // fix relies on, but not what we're testing here.
+        conn.pragma_update(None, "user_version", 1_i64)
+            .expect("reset version for test");
 
         // Simulate a pre-fix row: input far past the cap, small real output.
         conn.execute(
@@ -2245,6 +2309,8 @@ mod tests {
 
         let conn = Connection::open_in_memory().expect("open in-memory db");
         run_schema_migrations(&conn).expect("initial schema");
+        conn.pragma_update(None, "user_version", 1_i64)
+            .expect("reset version for test");
         conn.execute(
             "INSERT INTO commands (timestamp, original_cmd, rtk_cmd, input_tokens, output_tokens, saved_tokens, savings_pct)
              VALUES ('2026-07-24T14:15:33Z', 'cat huge.bin', 'rtk read', 135722224, 21, 135722203, 99.9999845272208)",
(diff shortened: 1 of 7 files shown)


src/core/tracking.rs
```diff
@@ -296,6 +296,14 @@ pub struct MonthStats {
 }
 
 /// Type alias for command statistics tuple: (command, count, saved_tokens, avg_savings_pct, avg_time_ms)
+/// Type alias for command statistics tuple: (command, count, saved_tokens, weighted_savings_rate, avg_time_ms)
```
KuSh: line duplicated instead of replaced

src/core/tracking.rs
```diff
@@ -305,7 +313,15 @@ type CommandStats = (String, usize, usize, f64, u64);
...
-const SCHEMA_VERSION: i64 = 1;
+const SCHEMA_VERSION: i64 = 2;
+/// The version the retroactive token-cap migration (see `run_schema_migrations`)
+/// was introduced at. Checked directly against the on-disk `user_version`
+/// rather than assumed from `SCHEMA_VERSION`, so this one-time data
+/// migration can't be silently re-triggered by a future schema bump that
+/// re-runs `run_schema_migrations` for an unrelated reason, nor by
+/// `MigrationMode::Always` (`rtk init`'s self-heal path) re-invoking the
+/// whole function on an already-migrated DB.
+const TOKEN_CAP_MIGRATION_VERSION: i64 = 2;
```
KuSh, line 323: Can't we just use an enum?

```rs
enum SchemaVersions {
  GatedMigration = 1, // either use it to gate the conditional migration (to avoid dead code) or just start at 2 with a comment above explaining why
  TokenCapMigration,
  // Insert new migrations above that comment
  Next
}
const SCHEMA_VERSION: i64 = DbVersion::Next as i64 - 1;
```

src/core/tracking.rs
```diff
@@ -401,6 +417,78 @@ fn run_schema_migrations(conn: &Connection) -> Result<()> {
...
+    //    same fallback every other schema-migration failure already causes
+    //    via the `if let Ok(tracker) = Tracker::new()` pattern call sites use).
+    let current_version: i64 = conn
+        .query_row("PRAGMA user_version", [], |row| row.get(0))
+        .unwrap_or(0);
+    if current_version < TOKEN_CAP_MIGRATION_VERSION {
+        // `usize::MAX` (cap disabled, RTK_TRACK_CAP_CHARS=0) can't round-trip
+        // through `as i64` without wrapping to -1, which would clamp every
+        // row to -1 instead of skipping the migration -- guard explicitly.
+        let cap = max_tracked_tokens();
```
KuSh, line 470: **Blocks merge** — the one-time migration's permanent effect is decided by one arbitrary reading of a mutable runtime knob. The `current_version < TOKEN_CAP_MIGRATION_VERSION` guard protects re-runs; the damage happens on the first run.

*Arm A* — a transient low cap destroys real history. Seeded v1 DB, `legit = 6789/2081 (69.3%)` plus the 135.7M phantom row:

```
RTK_TRACK_CAP_CHARS=400 rtk gain   ->  legit 6789 -> 100,  huge 135722224 -> 100
rtk gain   (default cap, forever)  ->  legit 100 (unrecoverable)
```

*Arm B* — `0 = uncapped` at this moment forfeits the repair permanently, because line 491 stamps `user_version = 2` even when the UPDATE is skipped:

```
RTK_TRACK_CAP_CHARS=0 rtk gain  ->  uv=2, huge 135722224 unchanged
rtk gain (cap re-enabled)       ->  uv=2, huge 135722224 still there
                                    "Tokens saved: 135.7M (100.0%)"
```

Arm B is the same failure the `?`-propagation fix was written to prevent, reached through the skip path instead of the error path. Both arms and the verified prototype are in the review body.

KuSh (changes requested):

**Claim**: `rtk gain`'s savings figures are inflated by unbounded per-call token estimates and by counting user-requested `rtk read` truncation as compression; cap the estimate, retroactively cap history, weight the per-command rate, and window the `rtk read` tracking baseline.

**Scope**: **accept**. This is the survivor for #3635, #1978, #891, #2699 and the second half of #1298 — all five are stale against `develop` and none is separately mergeable. Nothing else open touches these hunks; no maintainer branch competes. (round 1, frozen)

**Ran**: `rtk read` × 12 option forms (file + stdin) × base/head, byte-compared stdout; 7 real producers (`git log/status/diff`, `ls -R`, `grep`, `tree`, `cargo check`) × base/head; 4 migration scenarios × 2 binaries on seeded v1 DBs; 3 mutation checks against the PR's own tests. `fmt` clean, `clippy` 0 warnings, `cargo test --all` 3345+108 green, `--ignored` 8/9 (the 1 failure, `test_git_status_not_a_repo_exits_nonzero`, fails identically on `develop`). `hyperfine rtk git status`: 8.0 ms ±0.3 vs 8.2 ms ±0.4 on `develop` — the `cached_config()` reuse holds, no startup regression.

The core claim checks out. On an 11.6 MB file, `rtk read --max-lines 5` went from a recorded **3,037,038 → 32 tokens (99.999 % "saved")** to **32 → 32 (0 %)**, and stdout is byte-identical to `develop` in all 12 forms — the display-path regression codex caught stays fixed, including on the stdin path.

**1. `src/core/tracking.rs:471` — the one-time migration's permanent effect is decided by one arbitrary reading of a mutable runtime knob.** Two arms, same class; the guard added in `eb2f5b30` closes neither, because it only protects *re-runs* and the damage is done on the first run.

*Arm A — a transient low cap destroys real history.* Seeded v1 DB with `legit = 6789→2081 (69.3 %)` and the 135.7M phantom row:

```
RTK_TRACK_CAP_CHARS=400 rtk gain     # one invocation, any rtk command
  legit: 6789 -> 100    huge: 135722224 -> 100
rtk gain                             # cap back to default, forever after
  legit: 100            huge: 100
```

`RTK_TRACK_CAP_CHARS` is advertised in this PR's own docs hunk, so this is a documented knob, not a hostile input. One reading of it rewrites `input_tokens`, `output_tokens`, `saved_tokens` and `savings_pct` in place, with no backup and no recovery.

*Arm B — `0 = uncapped` at that moment forfeits the repair permanently.* Same seeded DB:

```
RTK_TRACK_CAP_CHARS=0 rtk gain   # UPDATE skipped...
  user_version = 2               # ...but line 491 stamps v2 anyway
rtk gain                         # cap re-enabled, default
  huge: 135722224                # still there. "Tokens saved: 135.7M (100.0%)"
```

`current_version < TOKEN_CAP_MIGRATION_VERSION` is false from then on, so the exact row this PR exists to delete survives for the whole retention window while the write-time clamp quietly works on new rows. This is the failure mode the `?`-propagation fix (case 2 in your comment block) was written to prevent, reached through the skip path instead of the error path.

I prototyped a fix, built it and ran all four scenarios against both binaries; your 36 tracking tests still pass unchanged:

```rust
// ceiling for the one-time repair is the constant, never a lower runtime override
let cap = match max_tracked_tokens() {
    usize::MAX => usize::MAX,
    n => n.max(DEFAULT_ESTIMATE_CAP_CHARS.div_ceil(4)),
};
...
// and don't stamp past the migration when it did not actually run
let version = if token_cap_migration_applied {
    SCHEMA_VERSION
} else {
    TOKEN_CAP_MIGRATION_VERSION - 1
};
conn.pragma_update(None, "user_version", version)?;
```

|                              | head                        | prototype                   |
|------------------------------|-----------------------------|-----------------------------|
| plain upgrade                | legit 6789, huge 7500       | legit 6789, huge 7500       |
| first open `cap=400`         | **legit 100, huge 100**     | legit 6789, huge 7500       |
| first open `cap=0`, then default | legit 6789, **huge 135722224** | legit 6789, huge 7500  |
| `cap=0` always               | uv=2, untouched             | uv=1, untouched             |

I'm sending this back rather than pushing it: whether a user-configured cap should govern a one-time historical rewrite is your call, and it destroys data either way it's decided. Take the prototype, or a marker column, or a pre-migration backup — but the first run needs the same protection you already gave the second.

### My two threads on eb2f5b30 — resolutions

Answering my own review notes here so they aren't scattered across threads.

**Duplicated doc line** (`tracking.rs:299`) → **fixed in `a7560d5a`, pushed; thread resolved.** The old `avg_savings_pct` line was left above the new `weighted_savings_rate` one, so the type alias documented itself twice and contradicted itself once.

**`SchemaVersions` enum** (`tracking.rs:324`) → **worth doing — fold it into the blocker push, with the first discriminant pinned.** The snippet in that thread uses default discriminants, which start at 0: `Next as i64 - 1` then yields `SCHEMA_VERSION = 1`, not 2, and `TOKEN_CAP_MIGRATION_VERSION` becomes 1. Every install on `develop` already reads `user_version = 1`, so `1 < 1` is false and the retroactive migration never runs for a single existing user. Built both against a seeded v1 DB holding the 135.7M phantom row:

```
GatedMigration = 1:              uv 1 -> 2,  input 135722224 -> 7500   (36/36 tracking tests pass)
default discriminants (from 0):  uv 1 -> 1,  input 135722224 -> unchanged — migration never runs
```

So `GatedMigration = 1` explicitly, with a comment saying the discriminants *are* the on-disk `user_version` values — the failure mode is silent on every existing install rather than loud at compile time.

That leaves `GatedMigration` itself unreferenced, and this repo builds clippy with `-D warnings`, so it is a hard error rather than a lint:

```
error: variant `GatedMigration` is never constructed
error: could not compile `rtk` (bin "rtk") due to 1 previous error
```

Three ways out, all built and gated here:

| | clippy | tracking tests | v1 DB upgrade |
|---|---|---|---|
| keep `GatedMigration = 1`, use it as the "migration did not run" stamp | clean, no `allow` | 36/36 | `uv 1→2`, `135722224→7500` |
| keep it + `#[allow(dead_code)]` | clean via the allow | 36/36 | same |
| omit it, start at `TokenCapMigration = 2` | clean, no `allow` | 36/36 | `uv 1→2`, `135722224→7500` |

The first if the blocker below is closed by conditional stamping, which is the shape I prototyped: that fix needs exactly "the version to stamp when the token-cap migration did not run", and `SchemaVersions::GatedMigration as i64` names it better than `TOKEN_CAP_MIGRATION_VERSION - 1`. The two then close each other's loose end. All five migration scenarios with both in place:

```
A plain upgrade                    uv=2   legit 6789 kept, huge -> 7500
B first open cap=400               uv=2   legit 6789 kept, huge -> 7500
C first open cap=0, then default   uv=1   untouched -> uv=2, huge -> 7500
D cap=0 always                     uv=1   untouched (repair still pending)
E three runs at default            uv=2   idempotent
```

If the blocker is instead closed with a marker column, nothing needs the value 1 and the third option is cleaner — omit the variant rather than keep it alive with an `allow`, which would mute the same signal for every future variant.

- **`src/core/tracking.rs:1419`, `avg_savings_per_command` is the third sibling and is unconverted.** Its inner `AVG(savings_pct)` is exactly what the new `CommandStats` warning at line 301 says never to write, and `telemetry.rs:282-289` ships it in the same payload as the now-weighted `low_savings_commands` — two contradictory figures for one command. This is the other half of #1298, which you're already superseding; converting the inner aggregate (the outer `AVG` over command names is deliberately unweighted) finishes it. Sites: 5 total — `summary.avg_savings_pct` and `overall_savings_pct` were already `SUM/SUM`, `get_by_command` and `low_savings_commands` you fixed, this one is left.
- **Comment volume and review narration.** 227 comment lines added, 29 % of the PR's non-blank additions. `tracking.rs` already runs verbose (24 blocks ≥8 lines on `develop`, longest 30), so most of that is house style — but the block at `:421` is now the file's longest at **44 lines guarding 13 lines of SQL**, and `read.rs` had **zero** blocks ≥8 lines before this PR and now has an 18-line one. Separately, seven comments narrate this PR's own review history rather than the code: "caught in review", "an earlier version computed…", "The old version discarded the UPDATE's Result via `let _ =`", "Two correctness requirements caught in review", "6i-b. Regression test for a real HIGH-severity bug caught in review". CONTRIBUTING.md:198 — comments explain *why*, not *what*; a reader six months out needs the invariant, not the diff of a version that never landed. Please rewrite those blocks whole rather than trimming around them.
- **`read.rs:79` allocates a second full copy of the file.** `tracking_baseline` calls `apply_line_window`, which returns `content.to_string()` when neither flag is set, on top of the `content.clone()` already made for `raw`. On a 300 MB read that's an extra 300 MB for a value that equals `raw`. Returning `Cow` or short-circuiting when both flags are `None` costs nothing.
- **Body says "40 new/rewritten unit tests"; the diff adds 17** (tracking 12, config 2, read 3) and rewrites none. Worth correcting before this is quoted anywhere.
- **`docs/guide/resources/savings-explained.md`** doesn't mention the consequence users will actually notice: when raw *and* filtered both exceed the cap, the recorded saving is 0 %. `rtk read --level minimal` on that 11.6 MB file goes 32.1 % → 0.0 %, and `rtk git diff HEAD~5` goes 63.7 % → 26.4 %. That's correct under the truncation model, but it will read as a regression to anyone watching their own numbers.

### Fixup pushed: `a7560d5a`

`test(tracking): make the migration-failure test fail on the bug it names` — 2 files, +23/−24, gate green. Pushed as a fast-forward on top of eb2f5b30.

`test_migration_failure_does_not_silently_mark_the_db_as_migrated` does not test what it names. `BEGIN EXCLUSIVE` on a second connection blocks readers too, so `run_schema_migrations` dies on its *first* statement and the retroactive cap `UPDATE` never executes; `is_err()` and the unchanged `user_version` are both satisfied by that earlier failure. Asserting the error text proves it — it reads `database is locked`, never `retroactive token-cap migration (schema v2) failed`.

```
mutation: `?` -> `let _ =` on the cap UPDATE   =>  test still passes
```

The fixup swaps the lock for a `BEFORE UPDATE OF input_tokens` ABORT trigger, which fails exactly that statement, and asserts the propagated error is the UPDATE's own context. It passes clean and **fails** under the same mutation. Also drops the stale duplicate `CommandStats` doc line the weighted-rate change left behind, and a stray blank line in the read tests.

For contrast, the guard test is real: mutating `current_version < TOKEN_CAP_MIGRATION_VERSION` to always-true makes `test_repeated_always_mode_migration_does_not_recap_with_a_new_lower_cap` fail with the right message.

- Migration SQL. SQLite evaluates every `SET` right-hand side against pre-update values, so `saved_tokens = MAX(MIN(input,cap) - MIN(output,cap), 0)` uses originals; two-arg `MIN`/`MAX` are the scalar forms; the formula matches `record`'s write-time `saturating_sub` + pct. Seeded regression row `20000/25000` → `7500/7500`, saved 0 — no fake positive, as claimed.
- `usize::MAX as i64` guarded before the cast; `cap = 0` skips the UPDATE.
- `read.rs`: `tracking_baseline` is genuinely separate from `raw`, so `never_worse` still compares against full content and cannot print an unfiltered excerpt. Verified by stdout comparison, not by reading: 12 forms byte-identical. The `mut`→shadowing change in `run_stdin` is behaviour-neutral; the empty-filter fallback yields 0 %, not a fake saving.
- Sibling call sites closed: `--max-lines`/`--tail-lines` exist only on `Commands::Read` in `main.rs`, and both `run` and `run_stdin` are converted. `run_stdin` verified separately (100000→31 becomes 31→31).
- Independent clamping collapsing large genuine savings to 0 % is correct under the stated model — the agent truncates both sides at the same ceiling, so the token cost really is equal. Not a bug.
- `serde(default)` parses pre-v2 configs; `cached_config()` avoids the second TOML parse, confirmed by hyperfine.

- *`rtk git diff --max-lines` has the same phantom-baseline bug as `read`.* Could not reproduce — `rtk git diff HEAD~30 --max-lines 20` records `0/0`, the flag isn't reaching that path. `git.rs` is untouched here regardless, so it would be an issue, not a finding. Not filed without a repro.
- *Capping could turn a pre-cap regression into a fake positive.* Constructed a mutation for it; the `MAX(…, 0)` floor holds for every row shape I could build. Dropped.
- *`rtk find --max` is a sibling of the same class.* Not a flag this PR touches, and my repro was eaten by shell globbing. Not pursued.
- *The `--max-lines` baseline windows raw content while the output windows filtered content, so the two cover different spans of the file* (N=50/200/1000 all record baseline slightly **below** output, floored to 0). Honest under the counterfactual "the user asked for N lines either way". Noted, not filed.

### Next

Close both arms of finding 1 in one push, and fold the enum into it with `GatedMigration = 1`. Everything else is optional or mine to push. Once that lands I'll take the fixup with it and approve.

Rounds: 1/3. Threads: 2 open (finding 1; the `SchemaVersions` thread, resolved above but left open for the author to act on), 1 resolved this round by the fixup.

src/core/tracking.rs
KuSh: Fixed in a7560d5a (just pushed): the stale `avg_savings_pct` line is gone, leaving only the `weighted_savings_rate` one. Verified by diffing this region between eb2f5b30 and the fixup.

KuSh:
Fixup pushed as a7560d5a — the review body's "prepared (not pushed)" line has been updated.

It contains only test/doc changes, no production behaviour change:

1. Removes the duplicated `CommandStats` doc line (thread resolved).
2. Replaces `test_migration_failure_does_not_silently_mark_the_db_as_migrated`. The `BEGIN EXCLUSIVE` lock blocks readers too, so `run_schema_migrations` died on its first statement and the retroactive cap `UPDATE` never executed — the propagated error read `database is locked`, never the migration'''s own context, so reverting `?` to `let _ =` left the test green. It now uses a `BEFORE UPDATE OF input_tokens` ABORT trigger, which fails exactly that statement, and asserts which statement failed. It passes clean and fails under that mutation.
3. One stray blank line in the read tests.

`cargo fmt --all -- --check` clean, `cargo clippy --all-targets` 0 warnings, `cargo test --all` green (3345 + 108).

The blocking item and the `SchemaVersions` decision are unchanged and still open for the author.

Issue #3476, opened 2026-08-08
hook check dry-run does not expose the actual host permission decision


pavelbe:
## Summary

`rtk hook check` shows the rewritten command but not the host decision that
`rtk hook claude` will emit. On `rtk 0.45.0`, two inputs that both look
successfully rewritten in the dry-run have materially different hook outputs:
one includes `permissionDecision: "allow"`; the other does not.

This makes the current dry-run insufficient for diagnosing host integration
and permission/rewrite behavior (see also #3018). The request is **not** to
auto-allow mixed blocks. It is to expose the actual decision faithfully.

## Environment

- RTK: `0.45.0`
- Linux x86_64 / WSL2
- target agent: Claude Code
- checked: 2026-08-08

## Reproduction

Fully rewritten block:

```bash
rtk hook check --agent claude $'git status\ngit diff --stat'
printf '%s' '{"tool_name":"Bash","tool_input":{"command":"git status\ngit diff --stat"}}' \
  | rtk hook claude
```

The check prints only:

```text
rtk git status
rtk git diff --stat
```

The host response includes:

```json
{"updatedInput":{"command":"rtk git status\nrtk git diff --stat"},"permissionDecision":"allow"}
```

Mixed rewritten/passthrough block:

```bash
rtk hook check --agent claude $'git status\ntrue'
printf '%s' '{"tool_name":"Bash","tool_input":{"command":"git status\ntrue"}}' \
  | rtk hook claude
```

The check again prints a successful-looking rewrite:

```text
rtk git status
true
```

But the host response has `updatedInput` and no `permissionDecision`:

```json
{"updatedInput":{"command":"rtk git status\ntrue"}}
```

The human dry-run cannot distinguish these outcomes.

## Expected

Add a stable machine-readable diagnostic mode, for example:

```bash
rtk hook check --agent claude --json $'git status\ntrue'
```

The payload should expose at least:

- target agent/host;
- original and rewritten command;
- whether any rewrite occurred;
- the exact `permissionDecision` that the host adapter would emit (nullable);
- a bounded reason/pass-through classification.

Human output may remain compact, but should not imply equivalent runtime
behavior for these two cases.

## Acceptance

1. JSON dry-run and actual `rtk hook claude` agree on rewritten input and
   permission decision.
2. Tests cover fully rewritten, mixed rewritten/passthrough, no-rewrite and
   substitution/deny cases.
3. No command is executed by the dry-run.
[...]

PR #3955, opened 2026-09-09
refactor(hooks): consolidate the hook decision behind one shared decide

### LINKED ISSUE #3936: rtk hook check disagrees with rtk hook claude on commands containing a file redirect
### Summary

`rtk hook check` reports a rewrite for commands containing a **file redirect** (`>`, `>>`, `2>`, `<`), but `rtk hook claude` — the real hook — proposes no rewrite for those same commands. The dry-run and the thing it is a dry run *of* disagree.

The pass-through behaviour itself looks deliberate and fine (redirects and pipes are both left alone). The bug is that `hook check` doesn't model it, so the one tool available for "will this command be rewritten?" gives the wrong answer for a common command shape.

### Reproduction

```console
$ mkdir /tmp/hcrepro && cd /tmp/hcrepro && printf 'alpha\nbeta\n' > f.txt

$ rtk hook check "grep -rn alpha f.txt > out.txt"
rtk grep -rn alpha f.txt > out.txt          # claims a rewrite, exit 0

$ echo '{"tool_name":"Bash","tool_input":{"command":"grep -rn alpha f.txt > out.txt"}}' | rtk hook claude
                                            # (empty — no rewrite proposed), exit 0

$ echo '{"tool_name":"Bash","tool_input":{"command":"grep -rn alpha f.txt"}}' | rtk hook claude
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk grep -rn alpha f.txt"}}}
```

### Scope

Tested 12 command shapes against both entry points. The disagreement is exactly the file-redirect cases; pipes are handled consistently by both.

| Command shape | `rtk hook check` | `rtk hook claude` | |
|---|---|---|---|
| `grep -rn X f` | rewrites | rewrites | agree |
| `grep -rn X f \| head -3` | no rewrite | no rewrite | agree |
| `git log --oneline \| wc -l` | no rewrite | no rewrite | agree |
| `grep -rn X f 2>&1` | rewrites | rewrites | agree |
| `grep -rn X f && echo done` | rewrites | rewrites | agree |
| `grep -rn X f > out.txt` | rewrites | **no rewrite** | **disagree** |
| `grep -rn X f >> out.txt` | rewrites | **no rewrite** | **disagree** |
| `grep -rn X f 2> err.txt` | rewrites | **no rewrite** | **disagree** |
| `grep -rn X f < in.txt` | rewrites | **no rewrite** | **disagree** |
| `git log --oneline > out.txt` | rewrites | **no rewrite** | **disagree** |
| `python3 -m pytest tests -q > out.txt` | rewrites | **no rewrite** | **disagree** |

`2>&1` is a file-descriptor duplication rather than a file redirect and is correctly rewritten by both, which suggests the real hook's rule is specifically "contains a redirect to/from a file".

### Why it matters

`rtk hook check` is the only way to answer "will RTK rewrite this?" without running the command, and it is the natural tool to reach for when deciding whether an `exclude_commands` entry has taken effect. We rely on it in a project rules file for exactly that. Getting a confident wrong answer from the verification tool is worse than not having it: a user who checks `rtk hook check "pytest x > log.txt"`, sees a rewrite, and concludes their exclusion is missing will add an entry that does nothing.

### Suggested fix

Route `hook check` through the same predicate the hook uses to decide whether to emit `updatedInput`, so the two cannot drift. If the pass-through is intentional, having `hook check` print something explicit would be ideal, e.g.:

```
No rewrite for: grep -rn alpha f.txt > out.txt  (contains a file redirect)
```

### Minor, related

`hook check` reports "no rewrite" on **stderr with exit 1**, and the rewrite on stdout with exit 0. That is defensible, but it is undocumented in `--help` and easy to get wrong: a script capturing stdout only sees an empty string and can misread it as "rewrites". Worth a line in the help text.

### Environment

- rtk 0.48.0 (Homebrew)
- macOS 26.4.1, arm64

KuSh:
Follow-up to #3704, which consolidated raw-command lexing and closed with:

> `hooks/rewrite_cmd.rs` vs `hooks/hook_cmd.rs` decision-flow duplication — real, but a separate follow-up at a different level.

## `rtk rewrite`'s behaviour is unchanged — deliberately

The obvious generalisation is to make `rtk rewrite` behave like the hooks and defer on a rewrite that changes nothing. **This PR does not do that**, and the divergence is kept on purpose:

| | already-`rtk` command |
|---|---|
| `rtk hook <agent>` | defers — a hook speaks an edit protocol, and an edit to the same string is meaningless |
| `rtk rewrite` | reports it (exit 0/3, command unchanged on stdout) |

That is the documented contract from #241, which created `rtk rewrite`: *"Handle already-rtk commands (exit 0, identical output)"*, still pinned by `test_run_already_rtk_returns_some`. The CLI answers *"what is the RTK form of this command"*; for an already-prefixed command that form is itself, and whether that counts as a change is the caller's question. Every delegate already asks it — `hooks/opencode/rtk.ts`, `hooks/pi/rtk.ts` (shared with omp), `hooks/hermes/rtk-rewrite/__init__.py` and `openclaw/index.ts` all gate on `rewritten != command`.

So `suppress_identity` is applied by `decide_for_agent`, which the hooks share, and not by `decide`, which the CLI renders. The exit-code protocol is untouched.

## What changed

`rtk hook <agent>` and `rtk rewrite` carried two independently written copies of the same four steps over two isomorphic enums (`HookDecision`, `RewriteOutcome`), so a fix to the gate order or a new construct to refuse had to be made twice.

- **`src/hooks/decision.rs`** — one `decide`, taking the permission verdict and the rewrite parameters rather than looking them up, so each host consults its own rules and tests stay independent of the developer's machine (#3146). `decide_for_agent` adds the no-op suppression every hook applies.
- **`rewrite_cmd::run`** is reduced to exit-code rendering; `hook_cmd`'s eight response builders are untouched.
- **`rtk hook check`** now routes through the same decision (the one intended behaviour change, below).

## Related issues

- Closes #3936 — `rtk hook check` disagrees with `rtk hook claude` on commands containing a file redirect. Fixed here; all six shapes in that report now agree, and `2>&1` still rewrites on both.
- **#3476** — `hook check` does not expose the actual host permission decision. Partially addressed: the diagnostic now consults the host's real rules instead of ignoring them, but it still collapses allow and ask into one answer. The remaining half is a follow-up, commented on that issue.

## `rtk hook check` — the one behaviour change

It called `registry::rewrite_command` directly, with no verdict and none of the gates, so it reported a rewrite for command substitutions, file redirects and heredocs that both hook paths refuse — the diagnostic disagreed with the thing it exists to diagnose, in the direction that matters.

Consulting rules makes the answer host-dependent, so `--agent` stops being discarded. `AgentPath` records what actually differs: the six agents deciding in-process use their own host's rules, the five whose plugin shells out to `rtk rewrite` get Claude's, and the six installing only a rules file have none. All 17 install targets resolve; only a typo is rejected.

On develop the diagnostic consults no rules and ignores `--agent`, so it contradicts every host. Under a Claude deny rule for `git status` it prints `rtk git status`, while `rtk hook claude` refuses the command outright.

## Verification

- `cargo fmt --all` · `cargo clippy --all-targets` clean · `cargo test --all` green; each commit builds and tests independently.
- A characterization commit lands **first**, pinning the real `(exit code, stdout)` pairs end-to-end — `rtk rewrite`'s exit codes had no Rust test at all, only a hand-copied table in `exit_code_protocol` that never called `run()`. Those tests pass **unmodified** across the refactor.
- Differential-tested against the base binary across an adversarial corpus (CRLF, NBSP, ZWSP, RTL override, 50 KB command, 200-segment chain, heredocs, substitutions): byte-identical on every path except `rtk hook check`.
- Property-fuzzed with a throwaway harness driving the built binary — not part of the diff, so it is not re-runnable by CI or a reviewer; the invariants it covers are the ones asserted in the committed tests. Across ~6k generated cases: exit codes stay in range, stdout appears exactly for the two rewrite outcomes, a deny rule is always enforced, rewriting is idempotent, no hook ever emits `updatedInput` identical to its input, and — the one that matters — a rewrite is never auto-allowed unless an allow rule covers every segment (#1155, #1213).
- Out-of-crate suites unchanged from their pre-refactor baselines: `hooks/claude/test-rtk-rewrite.sh` 58/66, hermes 18 passed, `scripts/test-all.sh` 105/13/5.
- After rebasing onto the recall feature (#3258 line), `track_tee_read` is called under exactly upstream's conditions on both paths — verified by differential-testing the recall store against the upstream binary. `rtk hook check` deliberately does not record, matching upstream.

## Not in scope

- Making `rtk rewrite` defer on the identity case (above).
- `rewrite_cmd` judges every subprocess-path agent by Claude's rules, since `rtk rewrite` cannot be told who is asking.
- `hook_cmd`'s tests read the developer's `config.toml` (26) and `~/.claude` rules (5); pre-existing, same count on the base commit, and fixing it means giving all six hosts the injection seam only three have.

### DIFF
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -1499,7 +1499,7 @@ pub(crate) fn get_db_path() -> Result<PathBuf> {
     // `config::cached_config`), not a fresh `Config::load()`: this runs inside
     // `Tracker::new()`, which `log_hook_decision` now calls on every single
     // PreToolUse hook invocation — `hook_rewrite_params()` (called earlier in the
-    // same hook invocation, via `get_rewritten`) already reads config too, so
+    // same hook invocation, via `hooks::decision::decide`) already reads config too, so
     // without caching that's two full disk-read-plus-TOML-parse round trips per
     // Bash tool call instead of one.
     if let Some(db_path) = crate::core::config::cached_config()
--- a/src/discover/mod.rs
+++ b/src/discover/mod.rs
@@ -172,7 +172,7 @@ fn estimate_hook_coverage(permission_cmd: &str, rewrite_cmd: &str, ctx: &Coverag
 
 /// Pure core of `estimate_hook_coverage`, taking the permission verdict directly so
 /// it's testable without depending on real config files. Mirrors the real hook's own
-/// decision order (`hooks::hook_cmd::decide_from_verdict`): a `Deny`-rule match or an
+/// decision order (`hooks::decision::decide`): a `Deny`-rule match or an
 /// unattestable construct means the hook would never have rewritten this, regardless
 /// of registry support. See `estimate_hook_coverage` for why `permission_cmd` and
 /// `rewrite_cmd` can differ.
--- a/src/hooks/README.md
+++ b/src/hooks/README.md
@@ -11,7 +11,8 @@ Owns: `rtk init` installation flows (6 agents via `AgentTarget` enum, now includ
 Does **not** own: the deployed hook scripts themselves (that's `hooks/`), the rewrite pattern registry (that's `discover/`), or command filtering (that's `cmds/`).
 
 Boundary notes:
-- `rewrite_cmd.rs` is a thin CLI bridge — it exists to serve hooks (hooks call `rtk rewrite` as a subprocess) and delegates entirely to `discover/registry`.
+- `decision.rs` is the single place RTK decides what a hook should do with a command — deny, defer, rewrite-and-allow, or rewrite-and-ask. All three entry points route through it: the in-process `rtk hook <agent>` hosts (`hook_cmd.rs`), the `rtk rewrite` subprocess path (`rewrite_cmd.rs`), and the `rtk hook check` diagnostic (`main.rs`). Add a gate there, not in a caller.
+- `rewrite_cmd.rs` is a thin CLI bridge — it exists to serve hooks (hooks call `rtk rewrite` as a subprocess) and renders `decision.rs`'s verdict as the exit codes those delegates branch on.
 - `trust.rs` gates project-local TOML filter execution. It lives here because the trust workflow is tied to hook-installed filter discovery, not to the core filter engine.
 
 ## Purpose
--- a/src/hooks/decision.rs
+++ b/src/hooks/decision.rs
@@ -1,0 +1,257 @@
+//! The single place RTK decides what a hook should do with a command.
+//!
+//! Two entry points ask the same question — may this command be rewritten,
+//! and may the rewrite be auto-allowed?
+//!
+//! | Entry point | Verdict source | Identity rewrite |
+//! |---|---|---|
+//! | `rtk hook <agent>` (`hook_cmd`) | `check_command_for(cmd, host)` | suppressed |
+//! | `rtk rewrite` (`rewrite_cmd`, run as a subprocess by the shell/TS/Python delegates) | `check_command` (always `Host::Claude`) | reported |
+//!
+//! [`decide`] is the shared answer. What legitimately differs between callers
+//! stays outside it: the verdict is passed in rather than looked up, so each
+//! host consults its own permission rules and tests stay independent of the
+//! machine's settings (#3146); and the no-op-rewrite policy lives in
+//! [`decide_for_agent`], which every hook shares and the CLI does not.
+
+use super::permissions::PermissionVerdict;
+use crate::discover::registry::rewrite_command;
+
+/// What a hook should do with a command.
+///
+/// `rewrite_cmd` renders these as the exit codes its delegates branch on
+/// (`AllowRewrite` → 0, `AskRewrite` → 3, `Deny` → 2, `Defer` → 1); the
+/// in-process hosts render them as their own JSON response shapes.
+#[derive(Debug, PartialEq)]
+pub(crate) enum HookDecision {
+    /// Rewrite, and the host may auto-allow it — an explicit user allow rule matched.
+    AllowRewrite(String),
+    /// Rewrite, but leave approval to the host's own prompt.
+    AskRewrite(String),
+    /// Say nothing; let the host handle the command untouched.
+    Defer,
+    /// A deny rule matched — stay out of the way of the host's native deny.
+    Deny,
+}
+
+/// Decide what to do with `cmd`, given a permission verdict for it.
+///
+/// The gate order is load-bearing:
+///
+/// 1. **Deny wins outright.** Checked before anything else so a denied command
+///    is never even considered for rewriting.
+/// 2. **Unattestable constructs are refused.** Command substitution and
+///    file-target redirects can't be decomposed into segments the permission
+///    gate can check individually, so a rewrite could smuggle an unchecked
+///    command past an allow rule. `check_command_with_rules` already forces
+///    such a command to `Ask`; refusing to rewrite it at all is the stronger
+///    guarantee. Heredocs reach the same outcome, though most of them stop at
+///    this gate only because a `<<` operand reads as a file target; what
+///    actually refuses them is `rewrite_command`'s own `has_heredoc`, which
+///    catches the forms this gate lets past (see #3980).
+/// 3. **Otherwise rewrite if a rule matches**, and auto-allow only on an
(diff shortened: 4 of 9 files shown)


aeppling (approved):

aeppling:
Ran a differential against develop: `rtk rewrite` and all six in-process hooks are byte-identical across ~140k cases (real harvested commands plus generated shapes, 4 permission profiles, every reachable host×verdict cell, audit logs compared serially to separate Deny from Defer). `hooks/claude/test-rtk-rewrite.sh` gives the same 58/66 with the same 8 failures on both binaries. `fmt`/`clippy --all-targets`/`test --all` clean.

**LGTM, approving.** The consolidation is real, and routing `hook check` through the shared decision is the right fix.

Two notes on the description, no code change needed:

**The "Before:" example doesn't reproduce.** On develop, `--agent gemini git status` under a Claude deny rule printed a *rewrite*, not "denied". That arm is `agent: _` with no permission call, and the string "Denied by a permission rule" has never existed on develop. The real contradiction is the **claude** row: develop reported a rewrite while `rtk hook claude` denied. Same wording is in b070c26's commit message, so it's worth fixing before merge, since it lands in history either way.

**The ~6k property-fuzz cases aren't in the diff.** No property-testing dep in `Cargo.toml`, so that one can't be re-run by CI or a reviewer.

Minor: `has_heredoc` is deleted rather than moved, so heredocs are now caught incidentally via the `<<` delimiter reading as a redirect target. Behavior is unchanged (verified), but `cmd << /dev/null` slips `redirect_has_file_target`'s exemption and `heredoc_defers` pins only the common form.

KuSh:
Thanks — all three checked out, and the third one turned out to be the interesting one.

**The "Before:" example.** You're right, and it was worse than one wrong row: I built develop and ran it, and *two* of the three contradictions I listed were artifacts of intermediate states on this branch rather than anything develop does.

| claim | develop actually does |
|---|---|
| `--agent gemini` "reported denied" | prints `rtk git status` — a rewrite |
| `--agent pi` "rejected outright" | exit 0, prints the rewrite |
| `rtk git status` reported as a rewrite | true |

The `--agent pi` rejection was a regression I introduced in an earlier round of this branch and fixed in a later one; it never existed on develop. Both the description and the commit message now state the claude row you identified, measured rather than recalled.

**The fuzz cases.** Correct — a throwaway harness driving the built binary, never committed. Rather than drop the claim I've labelled it as not in the diff and not re-runnable by you or CI, and noted that the invariants it covered are the ones the committed tests assert.

**`has_heredoc`.** This one I had wrong, and so did my own commit message. It isn't deleted — it lives at `registry.rs:264` and `rewrite_command` calls it at lines 601 and 616. What this PR removed is the hook path's *duplicate* of that check: `get_rewritten` called `registry::has_heredoc` and then called `rewrite_command`, which asks the same question again.

So the layering is the reverse of "caught incidentally":

- `contains_unattestable_construct` catches the common forms first — and that part *is* incidental, since `<<EOF` stops there only because `EOF` reads as a file target.
- `rewrite_command`'s `has_heredoc` is the deliberate guard, and it catches what the first gate lets past, including `<< /dev/null`.

Which is why `cmd << /dev/null` still defers rather than being rewritten — verified on both binaries across `<<EOF`, `<<'EOF'`, `<<-EOF`, `<< /dev/null` and `<</dev/null`: identical, exit 1, no output.

Your `/dev/null` observation still stands on its own, though, and I've filed it as #3980. Calling the predicate directly:

```
            git status <<EOF -> unattestable=true
     git status << /dev/null -> unattestable=false     <- heredoc, not a file
      git status > /dev/null -> unattestable=false     <- correct, the exemption's purpose
```

The exemption exists for discarded *output*; applied to a heredoc it is matching a *delimiter* that happens to be spelled `/dev/null`. No consequence today because the registry guard backstops it, but the predicate is also the permission gate's "never auto-allow" test, so it is worth being right for the right reason.

The only code change since your approval is that doc comment in `decision.rs`, which asserted the wrong owner of the guarantee — `git diff b070c263 HEAD` is those four lines and nothing else.

PR #3960, opened 2026-09-09
feat(hook): add native Codex CLI lifecycle hook

locoholy:
## Summary

- `hooks/codex/README.md` says Codex has *"Prompt-level guidance via awareness document -- no programmatic hook"*, so `rtk init --codex` only writes `AGENTS.md` + `RTK.md`. Codex ships a stable hook engine — this PR wires RTK into it.
- New `rtk hook codex`: reads the `PreToolUse` payload and answers with `permissionDecision: "allow"` + `updatedInput.command`.
- `rtk init [-g] --codex` installs the handler into `hooks.json` (`$CODEX_HOME/` or `.codex/`), and uninstall now removes it again.

Companion to #3959 (Antigravity), same shape, independent of it.

## Why I opened this

Same story as #3959: I use these CLIs daily, wanted RTK's savings in Codex, found nothing was intercepted, and traced it to an assumption in RTK's source rather than a bug.

```bash
codex features list | grep hooks     # → hooks   stable   true
```

Not experimental — stable and on by default. The engine lives in `codex-rs/hooks/`, with events `PreToolUse`, `PermissionRequest`, `PostToolUse`, `PreCompact`, `PostCompact`, `SessionStart`, `SessionEnd`, `UserPromptSubmit`, `SubagentStart`, `SubagentStop`, `Stop`, `Interrupt`, and a `/hooks` review UI.

## The good news: the parse side already works

Codex modelled its hooks on Claude Code's. A payload captured live from Codex 0.153.4 for `git status`:

```json
{
  "session_id": "01a08578-…", "turn_id": "01a08578-…",
  "cwd": "/path/to/workspace", "hook_event_name": "PreToolUse",
  "permission_mode": "bypassPermissions",
  "tool_name": "Bash", "tool_input": { "command": "git status" },
  "tool_use_id": "exec-7716f8ce-…"
}
```

`tool_name` + `tool_input.command` as a plain string — exactly what `process_claude_payload` already reads. That payload is checked into the tests verbatim.

## Why it still needs its own handler

The response side is validated strictly, and `rtk hook claude` fails it. Feeding the captured payload to the released binary gives:

```json
{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecisionReason":"RTK auto-rewrite","updatedInput":{"command":"rtk git status"}}}
```

No `permissionDecision` — because `process_claude_payload_from_decision` only sets it on the Allow branch, and `Host::Claude` returns Ask for anything not explicitly allowed. Codex rejects exactly that: *"PreToolUse hook returned updatedInput without permissionDecision:allow"*. Two more rules the handler respects: a deny needs a non-empty `permissionDecisionReason`, and `continue: false` / `stopReason` / `suppressOutput` are rejected as unsupported.

Rather than change `run_claude`'s behaviour for Claude Code users, this adds a Codex handler alongside the existing per-host ones.

## Judgment call

**A rewrite answers `allow`; no rewrite answers with silence.** `Host::Codex` carries no rules of its own (Codex gates commands through `approval_policy`/`sandbox_mode`, not a Bash allow-list), so every command lands on Ask/Defer. Since Codex only honours `updatedInput` together with `allow`, an Ask verdict cannot be expressed as "rewrite but still prompt". `allow` here means "RTK has no objection" — Codex still applies its own approval policy and sandbox to the rewritten call. Say the word if you'd rather have this behind a config toggle.

## Scope

Feature + its docs. Two things found while testing that are **not** in this diff:

- `src/hooks/hook_check.rs::status()` only inspects Claude Code, so `rtk gain` warns "No hook installed" at anyone whose agent is Codex, Antigravity, Gemini, Cursor or Droid. `other_integration_installed` already exists for this — but only under `#[cfg(test)]`, so production never calls it. Independent defect, separate PR.
- `HookCommands::Check` discards its `--agent` argument (`src/main.rs`), so `rtk hook check --agent anything` returns the generic rewrite. Also separate.

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets -- -D warnings && cargo test` — clean; 3411 tests pass, 25 of them Codex-related
- [x] `rtk hook codex` against the live captured payload → correct `permissionDecision` + `updatedInput`; non-`Bash` tools and non-rewritable commands produce no output
- [x] `rtk init --codex` appends to an existing `hooks.json` and leaves foreign `PostToolUse`/`Stop` handlers untouched; re-running does not duplicate
- [x] Uninstall removes the handler, prunes emptied containers, deletes a hooks.json that held only RTK's hook, and does not rewrite a file that never had it
- [x] End-to-end in Codex 0.153.4, interactive TUI and headless `codex exec`, with the handler installed: `git status` executed as `rtk git status`, exit 0

An earlier round of my own testing reported that Codex hangs on any hook rewrite. That did not reproduce under controlled conditions and I withdraw it — the rewrite path is fine in both TUI and headless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,7 +411,7 @@ RTK supports 17 AI coding tools. Each integration rewrites shell commands to `rt
 | **GitHub Copilot CLI** | `rtk init -g --copilot` | PreToolUse deny-with-suggestion (CLI limitation) |
 | **Cursor** | `rtk init -g --agent cursor` | preToolUse hook (hooks.json) |
 | **Gemini CLI** | `rtk init -g --gemini` | BeforeTool hook |
-| **Codex** | `rtk init -g --codex` | AGENTS.md + RTK.md instructions |
+| **Codex** | `rtk init -g --codex` (or project-scoped) | `PreToolUse` hook in `$CODEX_HOME/hooks.json` / `.codex/hooks.json` (matcher `Bash`) |
 | **Windsurf** | `rtk init -g --agent windsurf` | .windsurfrules (project-scoped) |
 | **Cline / Roo Code** | `rtk init --agent cline` | .clinerules (project-scoped) |
 | **OpenCode** | `rtk init -g --opencode` | Plugin TS (tool.execute.before) |
--- a/docs/guide/getting-started/configuration.md
+++ b/docs/guide/getting-started/configuration.md
@@ -80,8 +80,9 @@ level = "high"
 
 Re-run `rtk init -g` (or your agent's init command) to rewrite the file.
 
-Agents without a hook (Codex CLI, Cline, Windsurf, Kilo Code, Antigravity, Kimi) always get `full`,
-since the agent must type `rtk` itself. `rtk init` prints a note when it does this.
+Agents whose awareness ships as a rules file (Cline, Windsurf, Kilo Code, Antigravity, Kimi)
+always get `full`, since the rules file has to stand on its own. Codex keeps `full` too: its
+`AGENTS.md` guidance is the fallback layer behind the hook. `rtk init` prints a note when it does this.
 
 ## Environment variables
 
--- a/docs/guide/getting-started/supported-agents.md
+++ b/docs/guide/getting-started/supported-agents.md
@@ -41,7 +41,7 @@ Agent runs "cargo test"
 | Factory Droid | Shell hook (`PreToolUse`, matcher `Execute`) | Yes |
 | Cline / Roo Code | Rules file (prompt-level) | N/A |
 | Windsurf | Rules file (prompt-level) | N/A |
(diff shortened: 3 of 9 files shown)


KuSh:
Thanks for the detailed writeup, the payload capture in particular. Codex hook support landed meanwhile in #3552 (merged 2026-09-13): `rtk hook codex` answers with permissionDecision allow plus updatedInput.command, and `rtk init --codex` installs the handler into hooks.json under $CODEX_HOME or .codex with uninstall covered, along with tests for the documented permission modes and for preserving the other tool_input fields. The strict response-validation point you raised is what that handler implements, so I could not find a piece of this left to port. Please comment or reopen if you see a case the merged handler gets wrong.

PR #3962, opened 2026-09-09
fix: avoid tee recall tracking before approval

Dphoshoba:
## Summary
- Avoid recording tee-read recall for commands that still require approval
- Preserve recall tracking for allowed and passthrough/deferred commands
- Add regression coverage for Ask/Deny/Allow tracking boundaries

## Verification
- `cargo test --all`

### DIFF
--- a/src/hooks/hook_cmd.rs
+++ b/src/hooks/hook_cmd.rs
@@ -260,14 +260,19 @@ fn decide_from_verdict(cmd: &str, verdict: PermissionVerdict) -> HookDecision {
     if verdict == PermissionVerdict::Deny {
         return HookDecision::Deny;
     }
-    crate::hooks::rewrite_cmd::track_tee_read(cmd);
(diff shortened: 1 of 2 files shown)


KuSh:
Thanks for spotting this. The code it patches is gone: #3955 (merged 2026-09-11) moved the hook decision into src/hooks/decision.rs, so evaluate, evaluate_with_verdict and RewriteOutcome no longer exist and decide_from_verdict is now a three line delegate, which leaves this patch and its new test with nothing to attach to. The underlying point still holds on develop: track_tee_read is called for every verdict that is not a deny, at src/hooks/hook_cmd.rs:237 and src/hooks/rewrite_cmd.rs:87, so an Ask command the user never approves still records a tee recall. An issue, or a small PR against the shared decision path, would be very welcome. Please comment or reopen if I have missed a part of this.

PR #3965, opened 2026-09-09
fix(ruff): summarize format --check output for ruff >= 0.16 wording (#3930)

### LINKED ISSUE #3930: `filter_ruff_format` matches pre-0.12 ruff wording, so `rtk ruff format --check` compresses nothing
`filter_ruff_format` keys off two strings:

```rust
if lower.contains("would reformat:") { … }
if lower.contains("left unchanged") { … }
```

ruff stopped emitting both in check mode. On ruff 0.16.6 `format --check` prints `unformatted: File would be reformatted` per file and `N files already formatted` as the summary, so neither branch fires and the filter falls through to `result.push_str(output.trim())` — an identity function over the full ANSI diagnostic block.

## Reproduction (ruff 0.16.6, LC_ALL=C, 2 files, 1 unformatted)

```
$ ruff format --check . | wc -c        # 886
$ rtk ruff format --check . | wc -c    # 886   -> 0.0% savings
```

Against CONTRIBUTING.md's 20% floor, the format path currently delivers nothing. `ruff format` (write mode) still says `1 file left unchanged` and is still summarised correctly, so only the `--check` path is dead.

Verified identical on `develop` (665b2720) and on #3083's head — this predates that PR and is out of its scope.

## Suggested direction

Match the current wording as well: `unformatted:` for the per-file lines and `already formatted` alongside `left unchanged` for the summary. Keeping both spellings covers old and new ruff. The existing `test_filter_ruff_format_needs_formatting` / `test_filter_ruff_format_all_formatted` fixtures use the pre-0.12 strings and pass, which is why the drift went unnoticed — new fixtures should be captured from a current ruff.

anujsingh-cse:
Closes #3930

## Problem

`filter_ruff_format` keys only off pre-0.12 ruff strings: `"would reformat:"` and `"left unchanged"`. On ruff 0.16.6 `format --check` prints `unformatted: File would be reformatted` per file (with the path on a following ` --> path:row:col` pointer line) and `N files already formatted` as the summary. Neither branch matched, so the filter fell through to an identity passthrough — 0% savings on a path meant to hit the 20% floor.

Repro (ruff 0.16.6):

```
ruff format --check . | wc -c     # 886
rtk ruff format --check . | wc -c # 886  -> 0.0% savings
```

## Fix

Extend `filter_ruff_format` to recognise the current wording while keeping the old spellings:

- per-file: match `unformatted:` and extract the path from the following `--> path:row:col` line, alongside the existing `Would reformat: path`
- summary: match `already formatted` alongside the existing `left unchanged`
- check-mode detection is now driven by `!files_to_format.is_empty()` (both old and new per-file issuers populate it) plus the existing `would reformat` fallback

## Tests

Added two unit tests using the new wording (captured from a current ruff):

- `test_filter_ruff_format_all_formatted_ruff_016_wording` — `"5 files already formatted"`
- `test_filter_ruff_format_needs_formatting_ruff_016_wording` — the `unformatted:` + `--> path:row:col` two-line form

Existing old-wording tests unchanged and still pass; `cargo fmt --check` and `cargo clippy --all-targets` clean; 19 ruff tests pass. Verified end-to-end that ruff 0.16.6 `format --check` output (156 raw bytes) now flattens to a compact summary instead of passing through.

### DIFF
--- a/src/cmds/python/ruff_cmd.rs
+++ b/src/cmds/python/ruff_cmd.rs
@@ -262,11 +262,32 @@ pub fn filter_ruff_format(output: &str) -> String {
     let mut files_to_format: Vec<String> = Vec::new();
     let mut files_checked = 0;
 
+    // ruff >= 0.16 splits a `format --check` finding across two lines:
+    //   unformatted: File would be reformatted
+    //    --> path/to/file.py:2:1
+    // with the path on the line after the marker. Track that pairing.
+    let mut after_unformatted = false;
+
     for line in output.lines() {
         let trimmed = line.trim();
         let lower = trimmed.to_lowercase();
 
-        // Count "would reformat" lines (check mode) - case insensitive
+        // New ruff (>= 0.16) per-file marker; the path follows on the `-->` line.
+        if lower.contains("unformatted:") {
+            after_unformatted = true;
+        } else if after_unformatted {
+            if let Some(rest) = trimmed.strip_prefix("-->") {
+                // "path/to/file.py:2:1" — keep the path, drop the location.
+                if let Some(path) = rest.trim().rsplitn(3, ':').last() {
+                    if !path.is_empty() {
+                        files_to_format.push(path.trim().to_string());
+                    }
+                }
+            }
+            after_unformatted = false;
+        }
+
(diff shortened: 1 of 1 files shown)


anujsingh-cse:

PR #3977, opened 2026-09-11
fix(gain): report total savings percentage in by-command table (#3506)

### LINKED ISSUE #3506: `gain`: by-command table puts `SUM(saved_tokens)` next to `AVG(savings_pct)`, producing rows that can't be read together

anujsingh-cse:
Closes #3506

## Problem

`rtk gain`'s By Command table shows a `Saved` column (a `SUM`) next to an `Avg%` column (a `AVG` of per-command percentages). The two numbers in a row describe disjoint populations, so a reader naturally combines them ("24% off 478.8M") and gets nonsense — the average percentage is dominated by many tiny commands while the sum is dominated by a few huge ones.

## Fix

Make the percentage consistent with the sum in the same row by switching the aggregation from `AVG(savings_pct)` to a total-based percentage:

```sql
COALESCE(100.0 * SUM(saved_tokens) / NULLIF(SUM(input_tokens), 0), 0.0)
```

`COALESCE` guards the zero-total-input edge case (matching the `total_input > 0` guard already used for the summary's overall percentage). The column header is renamed `Avg%` → `Total%` to reflect it now answers a different question than a mean-of-percentages would.

## Tests

Added `test_get_by_command_uses_total_savings_pct`, which records a long-tailed pair (a 90%-savings 1M-token command and a 10%-savings 100-token command under the same `rtk_cmd`) and asserts the reported percentage is the total-based ~89.99% rather than the mean-of-percentages 50%.

`cargo fmt --check` and `cargo clippy --all-targets` clean; 28 tracking tests pass.

### DIFF
--- a/src/analytics/gain.rs
+++ b/src/analytics/gain.rs
@@ -220,7 +220,7 @@ pub fn run(
             println!("{}", "─".repeat(table_width));
             println!(
                 "{:>3}  {:<cmd_width$}  {:>count_width$}  {:>saved_width$}  {:>6}  {:>time_width$}  {:<impact_width$}",
-                "#", "Command", "Count", "Saved", "Avg%", "Time", "Impact",
+                "#", "Command", "Count", "Saved", "Total%", "Time", "Impact",
                 cmd_width = cmd_width, count_width = count_width,
                 saved_width = saved_width, time_width = time_width,
                 impact_width = impact_width
--- a/src/core/tracking.rs
+++ b/src/core/tracking.rs
@@ -295,7 +295,7 @@ pub struct MonthStats {
     pub avg_time_ms: u64,
 }
 
-/// Type alias for command statistics tuple: (command, count, saved_tokens, avg_savings_pct, avg_time_ms)
+/// Type alias for command statistics tuple: (command, count, saved_tokens, total_savings_pct, avg_time_ms)
 type CommandStats = (String, usize, usize, f64, u64);
 
 /// Current tracking-DB schema version, stored in the SQLite `user_version` pragma.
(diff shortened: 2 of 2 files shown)


KuSh:
Closing as a duplicate of #891, which makes the same change in get_by_command, swapping AVG(savings_pct) for a volume weighted rate over SUM(saved_tokens) and SUM(input_tokens), and renames the same column header. #891 was kept because it already has a green CI run, a maintainer approval and a CHANGELOG entry next to its regression test. Your in-memory test DB is a better fixture than the one there, so it would be welcome as a comment on #891, and please reply here if you were fixing something #891 leaves untouched.

KuSh:
Closing as a duplicate of #891, which makes the same change in get_by_command, swapping AVG(savings_pct) for a volume weighted rate over SUM(saved_tokens) and SUM(input_tokens), and renames the same column header. #891 was kept because it already has a green CI run, a maintainer approval and a regression test. Your in-memory test DB is a better fixture than the one there, so it would be welcome as a comment on #891, and please reply here if you were fixing something #891 leaves untouched.

PR #3988, opened 2026-09-11
fix(git): pass through --check/--exit-code/--quiet/--no-patch unfiltered

### LINKED ISSUE #3984: rtk git diff --check emits no output at all, silently discarding every whitespace diagnostic
# `rtk git diff --check` emits no output at all, silently discarding every whitespace diagnostic

**Version:** rtk 0.49.0
**Platform:** Linux x86_64 (musl build, installed via `install.sh`)

## Summary

`git diff --check` is a diagnostic mode whose **entire value is its stdout**: it prints one line per whitespace error and exits 2. Under `rtk`, stdout is empty while exit 2 is preserved.

An agent or pre-commit script therefore sees a failure with no reason attached. Worse, a script that reads the output to decide what to fix concludes there is nothing to fix, while the non-zero exit blocks the commit.

## Reproduction

```bash
T=$(mktemp -d); cd "$T"; git init -q .
printf 'line one\nline two\n' > f.txt
git add -A; git -c user.email=t@t -c user.name=t commit -qm init
printf 'line one\nline two   \nline three\t\n' > f.txt   # add trailing whitespace

git --no-pager diff --check; echo "raw exit=$?"
rtk git diff --check;        echo "rtk exit=$?"
```

### Actual

```
-- raw --
f.txt:2: trailing whitespace.
+line two
f.txt:3: trailing whitespace.
+line three
raw exit=2
-- rtk --
rtk exit=2
```

Raw: 4 lines of diagnostics + exit 2. rtk: **zero bytes** + exit 2.

### Expected

`--check` output passes through unfiltered. It is already minimal (one line per offending line plus the offending line), so there is nothing to compress, and every line is actionable.

## Why this one matters more than general diff reshaping

This is not a formatting difference — it is total loss of the only signal the command produces. Related but distinct from #1918 (`rtk git diff` breaking patch/`--name-only`/exit-code consumers): that issue is about generating a *usable patch*, whereas `--check` produces no patch at all and is purely a diagnostic whose output is destroyed.

## Suggested fix

Treat `--check` (and `--exit-code`, `--quiet`, `--no-patch`) as diagnostic/pass-through modes in the `git diff` filter: detect the flag and emit raw stdout. These modes are all cases where the output is either tiny or deliberately empty, so filtering has no upside.

## Discovery context

Found while auditing ~14,700 real `RTK_DISABLED=1` bypasses in local agent session history. `git diff --check` appeared as a case where an agent bypassed rtk after getting "(no output)" with exit 2 and being unable to explain the failure.

yunaremaia:
## Summary

Diagnostic modes `--check`, `--exit-code`, `--quiet`, and `--no-patch` derive their value entirely from stdout and exit code. RTK's `git diff` filter destroyed that signal — `rtk git diff --check` returned exit 2 with zero bytes of output, leaving agents and pre-commit scripts with "failure with no reason attached".

This PR adds an `is_diagnostic` gate that routes these flags through the existing passthrough path.

## Changes

- `src/cmds/git/git.rs` — added `is_diagnostic` flag detection in `run_diff`, ANDed into `wants_compact` so diagnostic modes skip compaction
- Test: `test_diagnostic_flags_skip_compaction` verifies all four flags skip compaction

## Verification

- `cargo check` passes (no new warnings)
- New test added for the gating logic
- Reproduction from #3984: `rtk git diff --check` now emits the same diagnostic lines as raw `git --no-pager diff --check`

Fixes #3984

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -125,8 +125,18 @@ fn run_diff(
         .iter()
         .any(|arg| arg == "--stat" || arg == "--numstat" || arg == "--shortstat");
 
+    // Diagnostic modes pass through unfiltered: --check, --exit-code, --quiet,
+    // --no-patch. Their value is entirely in stdout (one line per whitespace
+    // error for --check, nothing for --exit-code) and exit code, so filtering
+    // loses the only signal and produces "failure with no reason attached".
+    let is_diagnostic = args.iter().any(|arg| {
+        arg == "--check" || arg == "--exit-code" || arg == "--quiet" || arg == "--no-patch"
+    });
+
     // Check if user wants compact diff (default RTK behavior)
-    let wants_compact = !args.iter().any(|arg| arg == "--no-compact") && !emits_word_diff(args);
+    let wants_compact = !args.iter().any(|arg| arg == "--no-compact")
+        && !emits_word_diff(args)
+        && !is_diagnostic;
 
     if wants_stat || !wants_compact {
         // User wants stat or explicitly no compacting - pass through directly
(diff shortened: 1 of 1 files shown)


yunaremaia:
Thanks @rtk-wshm-sync-bot for the triage summary — the analysis matches the fix exactly:  derives its value from stdout, and RTK's compaction was destroying that signal.

CLA is pending; I'll sign it now. Maintainers: this is ready for review — one-line logic change in  with a dedicated test.

yunaremaia:
Hi @rtk-ai — gentle bump on this PR (fix(git): pass through --check/--exit-code/--quiet/--no-patch unfiltered). It's mergeable and CLA is signed. Would appreciate a review when you have a moment. Thanks!

KuSh:
Thanks for the fix and the clear reproduction. This landed on develop through #3681 (merged 2026-09-12, commit 7931993d): git diff now routes --check, --exit-code and --quiet through a raw pass-through when the tokenizer sees them, and -s/--no-patch skip compaction because there is no body to compact. Checked against git 2.53 on current develop, all four flags and their combinations print exactly what raw git prints and keep its exit code, including exit 2 for --check. Closing as already fixed; if you still see a diagnostic mode that comes back empty on develop, please comment here and it can be reopened.

PR #3989, opened 2026-09-12
fix(git): pass through diff diagnostic modes like --check unfiltered (#3984)

### LINKED ISSUE #3984: rtk git diff --check emits no output at all, silently discarding every whitespace diagnostic

anujsingh-cse:
Closes #3984

## Problem

`rtk git diff --check` emits no output at all — it silently discards every whitespace diagnostic. `git diff --check` is a diagnostic mode whose entire value is its stdout (one line per whitespace error, exit code 2). Under rtk, stdout was empty while exit 2 was preserved, so an agent or pre-commit script sees a failure with no reason attached.

Repro:

```
git --no-pager diff --check   # 4 lines of diagnostics, exit 2
rtk git diff --check          # 0 bytes, exit 2
```

## Fix

Treat `--check` (plus the sibling diagnostic modes `--exit-code`, `--quiet`, `--no-patch`) as pass-through modes in the `git diff` filter. A `wants_diagnostic_output` helper detects these flags, and `run_diff` routes them to a raw pass-through branch that emits stdout and stderr verbatim and preserves the exit code — no `--stat` rewrite and no compaction.

## Tests

- `test_wants_diagnostic_output_detects_diff_modes` — all four flags detected
- `test_wants_diagnostic_output_ignores_normal_diff_args` — `--stat`, `--numstat`, a path, and empty args are not diagnostic

`cargo fmt --check` and `cargo clippy --all-targets` clean; 478 git tests pass. Verified end-to-end: `rtk git diff --check` now prints the trailing-whitespace diagnostics and preserves exit code 2.

### DIFF
--- a/src/cmds/git/git.rs
+++ b/src/cmds/git/git.rs
@@ -128,6 +128,39 @@ fn run_diff(
     // Check if user wants compact diff (default RTK behavior)
     let wants_compact = !args.iter().any(|arg| arg == "--no-compact") && !emits_word_diff(args);
 
+    // Diagnostic modes whose entire signal is on stdout (--check, --exit-code,
+    // --quiet, --no-patch) must pass through unfiltered. They exit non-zero
+    // when they find problems, so dropping their stdout turns a whitespace
+    // report into a silent failure (issue #3984).
+    if wants_diagnostic_output(args) {
+        let mut cmd = git_cmd(global_args);
+        cmd.arg("diff");
+        for arg in args {
+            if arg == "--no-compact" {
+                continue; // RTK flag, not a git flag
+            }
+            cmd.arg(arg);
+        }
+
+        let result = exec_capture(&mut cmd).context("Failed to run git diff")?;
+
+        if !result.stdout.is_empty() {
+            print!("{}", result.stdout);
+        }
+        if !result.stderr.trim().is_empty() {
+            eprint!("{}", result.stderr);
+        }
+
+        timer.track(
+            &format!("git diff {}", args.join(" ")),
+            &format!("rtk git diff {} (diagnostic)", args.join(" ")),
+            &result.stdout,
+            &result.stdout,
+        );
+
+        return Ok(result.exit_code);
+    }
+
     if wants_stat || !wants_compact {
         // User wants stat or explicitly no compacting - pass through directly
         let mut cmd = git_cmd(global_args);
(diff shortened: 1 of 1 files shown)


KuSh:
Thanks for the fix and the repro. The same behaviour landed on develop through #3681 (merged 2026-09-12, commit 7931993d): --check, --exit-code and --quiet are classified by the tokenizer as raw-shape flags and go straight through with git's stdout, stderr and exit code, and -s/--no-patch skip compaction since there is no body to compact. Verified on current develop with git 2.53: rtk git diff --check prints the whitespace report and exits 2, and the combined forms match raw git byte for byte. Closing as already fixed; if a diagnostic mode still comes back empty for you on develop, please comment here and it can be reopened.

PR #4014, opened 2026-09-13
fix(discover): require token boundaries for rewrite rules

### LINKED ISSUE #4009: 22 rewrite rules end in a bare subcommand alternation with no word boundary (`git branchless` → `rtk git branchless`)
## Summary

22 rules in `src/discover/rules.rs` end in a bare subcommand alternation with no trailing word boundary, so any command whose second word merely *starts with* a listed subcommand is rewritten. `git branchless status` becomes `rtk git branchless status`, and `git-branchless` is a real, widely used tool.

This was found while reviewing #3676, which fixes exactly this bug for the `pnpm` rule by appending `(\s|$)`. 23 rules had the defect on `develop`; that PR fixes one, leaving 22.

## Reproduction

Against the `develop` binary (79b96a44):

```
$ rtk rewrite "git branchless"      → rtk git branchless
$ rtk rewrite "git statuses"        → rtk git statuses
$ rtk rewrite "git logger"          → rtk git logger
$ rtk rewrite "git showcase"        → rtk git showcase
$ rtk rewrite "cargo installer"     → rtk cargo installer
$ rtk rewrite "cargo builder"       → rtk cargo builder
$ rtk rewrite "gh prs"              → rtk gh prs
$ rtk rewrite "gh issues"           → rtk gh issues
$ rtk rewrite "glab mrs"            → rtk glab mrs
$ rtk rewrite "docker psql"         → rtk docker psql
$ rtk rewrite "kubectl getall"      → rtk kubectl getall
$ rtk rewrite "oc status-check"     → rtk oc status-check
$ rtk rewrite "ruff checker"        → rtk ruff checker
$ rtk rewrite "go vetting"          → rtk go vetting
$ rtk rewrite "terraform planner"   → rtk terraform planner
$ rtk rewrite "pio runner"          → rtk pio runner
```

Every one of these should be left untouched.

`git branchless` is the one with real-world bite today: it is an installed third-party subcommand, and the hook silently redirects it into `rtk git`, which has no idea what `branchless` is.

## Affected rules

All 22 end in an alternation group with nothing anchoring its right edge:

`git|yadm`, `gh`, `glab`, `cargo`, `prettier`, `next build`, `playwright`, `prisma`, `docker`, `kubectl`, `oc`, `ruff`, `sqlfluff lint`, `pip|uv pip`, `go`, `sbt`, `rake|rails test`, `pio run`, `quarto render`, `shopify theme`, `terraform plan`, `trunk build`

(`pnpm` was the 23rd and is fixed by #3676.)

Rules that already do this correctly — `^ls(\s|$)`, `^tree(\s|$)`, `^wc(\s|$)`, `^mypy(\s|$)`, `^pytest(\s|$)`, the gradlew rule, and the several `\b`-terminated ones — show the intended shape.

## Suggested fix

Append `(\s|$)` (or `\b`) to each affected pattern, then close the class with a table-wide test asserting every rule's pattern is boundary-terminated, so a new rule can't reintroduce it:

```rust
#[test]
fn every_rule_pattern_is_boundary_terminated() {
    for (i, rule) in RULES.iter().enumerate() { /* … */ }
}
```

A test that simply appends a junk suffix to each rule's own sample command and asserts no rewrite would catch all 22 at once.

## Note on blast radius

pnpm 12 adds `clean` and `purge` built-ins that delete `node_modules`. None are matched by a rule today, but the combination of destructive built-ins and unanchored alternations is worth closing before it matters.

## Related

- #3676 — fixes the `pnpm` arm and is where this was found
- #2902 — same family (over-eager rewriting) but a different mechanism: unsupported *predicate shapes* in `find`/`fd` rather than the alternation boundary. Both would need fixing independently.

axelray-dev:
## Summary
- Fix #4009 by requiring a token or shell delimiter after supported subcommands in native and TOML-backed rewrite rules.
- Add table-driven coverage for longer command names while preserving valid subcommand rewrites.

## Test plan
- [ ] `cargo fmt --all && cargo clippy --all-targets && cargo test --all` (formatting and clippy passed; the full suite has one Git-version assertion failure because Git 2.43 reports `not a non-negative integer` while the test expects `not an integer`)
- [x] `cargo test discover::registry::tests`
- [x] Manual testing: `target/debug/rtk rewrite` leaves `git branchless status`, `cargo builder`, and `terraform planner` unchanged and rewrites `git branch status`, `cargo build --release`, and `terraform plan`

### DIFF
--- a/src/discover/rules.rs
+++ b/src/discover/rules.rs
@@ -56,7 +56,7 @@ impl Default for RtkRule {
 
 pub const RULES: &[RtkRule] = &[
     RtkRule {
-        pattern: r"^(?:git|yadm)\s+(?:-[Cc]\s+\S+\s+)*(status|log|diff|show|add|commit|checkout|push|pull|branch|fetch|stash|worktree)",
+        pattern: r"^(?:git|yadm)\s+(?:-[Cc]\s+\S+\s+)*(status|log|diff|show|add|commit|checkout|push|pull|branch|fetch|stash|worktree)(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk git",
         pipeline_safety: PipelineSafety::ProducerOnly,
         rewrite_prefixes: &["git", "yadm"],
@@ -71,7 +71,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^gh\s+(pr|issue|run|repo|api|release)",
+        pattern: r"^gh\s+(pr|issue|run|repo|api|release)(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk gh",
         rewrite_prefixes: &["gh"],
         category: "GitHub",
@@ -80,7 +80,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^glab\s+(mr|issue|ci|pipeline|api|release)",
+        pattern: r"^glab\s+(mr|issue|ci|pipeline|api|release)(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk glab",
         rewrite_prefixes: &["glab"],
         category: "GitLab",
@@ -89,7 +89,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^cargo\s+(build|test|clippy|check|fmt|install)",
+        pattern: r"^cargo\s+(build|test|clippy|check|fmt|install)(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk cargo",
         pipeline_safety: PipelineSafety::ProducerOnly,
         rewrite_prefixes: &["cargo"],
@@ -241,7 +241,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?prettier",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?prettier(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk prettier",
         pipeline_safety: PipelineSafety::ProducerOnly,
         rewrite_prefixes: &[
@@ -266,7 +266,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?next\s+build",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?next\s+build(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk next",
         pipeline_safety: PipelineSafety::ProducerOnly,
         rewrite_prefixes: &[
@@ -377,7 +377,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?playwright",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?playwright(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk playwright",
         rewrite_prefixes: &[
             "npm exec playwright",
@@ -401,7 +401,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?prisma",
+        pattern: r"^((p?np(m|x)|p?npm\s+(exec|run|run-script)|npm\s+(rum|urn|x)|pnpm\s+dlx)\s+)?prisma(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk prisma",
         rewrite_prefixes: &[
             "npm exec prisma",
@@ -425,7 +425,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^docker\s+(ps|images|logs|run|exec|build|compose\s+(ps|logs|build))",
+        pattern: r"^docker\s+(ps|images|logs|run|exec|build|compose\s+(ps|logs|build))(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk docker",
         rewrite_prefixes: &["docker"],
         category: "Infra",
@@ -433,7 +433,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^kubectl\s+(get|logs|describe|apply)",
+        pattern: r"^kubectl\s+(get|logs|describe|apply)(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk kubectl",
         rewrite_prefixes: &["kubectl"],
         category: "Infra",
@@ -441,7 +441,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^oc\s+(get|logs|describe|apply|status|adm)",
+        pattern: r"^oc\s+(get|logs|describe|apply|status|adm)(?:\s|$|[;|&()<>])",
         rtk_cmd: "rtk oc",
         rewrite_prefixes: &["oc"],
         category: "Infra",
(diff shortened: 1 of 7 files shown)


src/discover/rules.rs
```diff
@@ -574,7 +574,7 @@ pub const RULES: &[RtkRule] = &[
         ..RtkRule::DEFAULT
     },
     RtkRule {
-        pattern: r"^(?:bundle\s+exec\s+)?(?:bin/)?(?:rake|rails)\s+test",
+        pattern: r"^(?:bundle\s+exec\s+)?(?:bin/)?(?:rake|rails)\s+test(?:\s|$|[;|&()<>])",
```
KuSh, line 577: The boundary excludes `:`, which is exactly how Rake namespaces tasks — so every namespaced test task loses its rewrite. Verified `0f8b26c7` → `98f8fffc`:

| command | develop | this PR |
|---|---|---|
| `rake test:unit` | `rtk rake test:unit` | *(no rewrite)* |
| `rake test:models` | `rtk rake test:models` | *(no rewrite)* |
| `rails test:system` | `rtk rake test:system` | *(no rewrite)* |
| `bundle exec rake test:unit` | `rtk rake test:unit` | *(no rewrite)* |

These are the canonical Rails invocations, and `rake_cmd::select_runner` already handles them: `args[0]` is `test:unit` rather than `test`, so it falls through to `("rake", args)` and passes the task straight to `rake`. The rewrite was working and useful.

Allowing `:` on this one rule fixes it without reopening anything:

```suggestion
        pattern: r"^(?:bundle\s+exec\s+)?(?:bin/)?(?:rake|rails)\s+test(?:[\s:]|$|[;|&()<>])",
```

Verified with that change: all four commands above rewrite again, `rake tester` / `rails tester` / `rake test-helper` stay blocked, and `cargo test --all` is green.

KuSh (changes requested):

**Claim**: fix #4009 by requiring a token or shell delimiter after the subcommand alternation in 21 native rewrite rules and 5 TOML `match_command` patterns.

**Scope**: **accept** (frozen for later rounds). The claim is worth merging as stated; one regression to fix. #3676 fixes the 22nd arm (`pnpm`) and touches the same two files but not the same lines — complementary, either order merges.

**Ran**: `0f8b26c7` (merge-base) vs `98f8fffc`, one `CARGO_TARGET_DIR` each, md5-distinct binaries. `rtk rewrite` over the issue's 22 repros, a valid-command set, a boundary-character matrix (every bash metacharacter plus NBSP / CR / VT / FF / EN SPACE), and three source mutations. Savings floor is not applicable — this changes routing, not filter output.

**CI @98f8fffc**: **never ran.** Only `license/cla` is green, `check-target` skipped, run `34740154437` is `action_required` — the fork-PR state waiting on a maintainer, so this PR has never actually been built by CI. Local gate on this PR *merged into latest develop* (`6d104308`, clean merge) is green: fmt clean, clippy 0 warnings, `cargo test --all` 3640 passed / 0 failed.

1. `src/discover/rules.rs:577` — **namespaced rake/rails test tasks lose their rewrite.** `rake test:unit`, `rake test:models`, `rails test:system` and `bundle exec rake test:unit` all rewrite on `develop` and stop on this head. Details and a verified one-character fix are in the line comment.

- **`\s` is not the boundary this repo defines.** `lexer.rs:60-65` states it outright: *"Bash's default `$IFS` is exactly space/tab/newline — not Rust's `char::is_whitespace()`, which wrongly includes non-IFS Unicode whitespace like NBSP"*. The regex crate's `\s` is `\p{White_Space}`, i.e. exactly what that comment forbids. Still false-positive on this head: `git status<NBSP>foo`, `git status<EN SPACE>foo`, `git status<VT>foo`, `git status<FF>foo` all rewrite. **Not a regression** — `develop` does it too (`uv run<NBSP>x` rewrites via the untouched `^uv\s+run(?:\s|$)`), so this is a pre-existing class the PR replicates into 21 more sites while its stated purpose is to model token boundaries. `(?:[ \t\r]|$|[;|&()<>])` is suite-green as-is; `[ \t]` additionally needs `test_lone_cr_line_gets_a_single_prefix` updated, which is arguably correct — by that test's own comment (*"bash keeps `git log` glued to the preceding word"*) the command's `argv[1]` is `status\rgit`, so under #4009's logic it shouldn't be rewritten at all.
- **The new test doesn't close the class.** `test_subcommand_rules_require_token_boundaries` is a hardcoded 24-string list. Inject a new unanchored rule — `^nomad\s+(status|job|alloc)` — and `nomad status-all` → `rtk nomad status-all`, the same defect, while the test **passes**. #4009 asked for the iterating form (`for (i, rule) in RULES.iter().enumerate()`), and that form fails and names the offending pattern. In fairness the current test does catch a regression of the rules it names: reverting the `go` boundary fails it on `go vetting`. I have the iterating version written and verified locally and will bring it along rather than spend a round on it.
- **The TOML half's metacharacter alternatives are unreachable.** `match_command` is matched against argv joined by spaces (`main.rs:1555-1564`), where a metacharacter can never be a separator; and all five commands have a `RULES` entry, so `classify_command` returns `Supported` before the TOML branch is consulted. Only `(\s|$)` can ever fire — which is what the 13 sibling filters already use (`tofu-*`, `pulumi-*`, `mix-*`, `df`, `ps`, `ssh`, `liquibase`). The added group is also capturing where the Rust side is `(?:…)`.
- **Precedence widening on version-pinned npx.** `REGEX_SET.matches(...).last()` wins, so when a tool rule stops matching, the earlier generic `^npx\s+` takes over: `npx prettier@3 --check .` → `rtk npx prettier@3 --check .`, where `develop` emitted nothing (same for `playwright@1.40`, `prisma@5`). Not a regression, and consistent with how `bunx prettier@3` already behaved — but it routes these into `npm_cmd`'s npx path, which by its own comment doesn't forward piped stdin.

### On the boundary itself

Worth recording, because the shape is easy to "simplify" later and two of these were nearly dropped in review:

- **The metacharacter half is not redundant with the segmenter.** For `rewrite_segment` it is — `;`, `|`, `&` are already gone by then, and dropping them changes not one rewrite result. But `classify_command` matches the **whole unsegmented line**, so without them `git status;ls`, `go test&` and `cargo build;make` classify `Unsupported` while the rewrite still fires, and `rtk discover` / `rtk session` under-count every compound command.
- **`(` and `)` are load-bearing in both paths.** `lexer.rs:170` classifies them as `Shellism`, not `Operator`, and `split_on_operators` splits only on `Operator`/`Pipe` — so parens survive segmentation glued to the segment, and `(git status; cargo build)` yields the literal segment `cargo build)`. Drop `()` and `segmenter_consistency::subshell_grouping` fails. The suite demands `[()]` and nothing else.
- **No quantifier is needed.** Only the first character after the subcommand is examined, so `git status  now` matches identically under `[ \t]` and `[ \t]+`.
- **Lookahead isn't an option** — the regex crate rejects look-around, and these patterns go through `.expect(...)`, so it would panic at first use rather than degrade. Consuming is safe here regardless: the only reader is `registry.rs:231` `captures(...)` → `caps.get(1)`, and nothing reads the match end.
- **No POSIX/Windows split.** The lexer is bash-modelled throughout and the hook is fed whatever the agent's Bash tool ran. cmd.exe's extra separators (`,` `=` `^`) and PowerShell's (`,` `{` `}`) would be a lexer-wide decision, not a per-rule regex one. This head correctly stops rewriting `cargo build,test` under the bash model.

That said — **this is a fix at the end of the chain**, and I'd rather say so than let the shape harden. `segmenter_consistency::subshell_grouping` already pins *three* different answers for one input: `split_for_permissions` strips the parens, `split_command_chain` glues them, and the rewrite loses the first command entirely (`(git status; cargo build)` → `(git status; rtk cargo build)`; `(git status)` isn't rewritten at all). Its own comment calls that "a real, existing quirk … pinned here, not fixed here", and #4100 independently concludes the two sides should share one notion of the boundary. `(` and `)` are control operators in the shell grammar, not shellisms. I'm taking that convergence on directly as a follow-up refactor — one lexer pass shared by classify, rewrite and the permission gate — which should also absorb #1053, #1252 and #4100. When it lands these 21 suffixes become dead weight and delete cleanly. None of that blocks this PR: it removes 22 verified false positives today, including `git branchless`, which is a real installed tool being silently misrouted.

- All 22 repros in #4009 fixed. `prettierish`, `playwrighting`, `prismax`, `next builder`, `sbt tester`, `shopify themepark` were never broken on `develop` (the sbt rule was already anchored) — harmless to keep in the table.
- Coverage is complete: across all 93 rules the only remaining unanchored pattern is `^pnpm\s+(exec|i|…)`, which is #3676's.
- Capture-group numbering intact — the suffix is `(?:…)` everywhere, so `caps.get(1)` and the `subcmd_status` / `subcmd_savings` lookups are untouched; 441 registry tests green.
- Valid commands preserved: `git -C /tmp status`, `docker compose ps`, `uv pip install flask`, `oc get pods`, `go test ./...`, `terraform plan -out=x`, `(git status; cargo build)`, `git status|head`, `git status;ls`, `git status&`.
- Newly *stopping* rewrites of `git status{a,b}`, `git status!x`, `git status*` is correct — those aren't metacharacters in bash, so each is a single word.

- **`helm statuses` / `gradle tests` are not false positives.** `^helm\b` and the gradlew rule (optional subcommand group + `(\s|$)`) are deliberate catch-alls; routing every subcommand is the design.
- **"The subshell case regresses"** — an artifact of a contaminated binary on my side. A fresh worktree and target dir show this head handling `(git status; cargo build)` exactly as `develop` does.
- **"The metacharacters are untested"** — they are, by `subshell_grouping`, just not by this PR's own test.

### Your questions

- *"the full suite has one Git-version assertion failure because Git 2.43 reports `not a non-negative integer` while the test expects `not an integer`"* → **not yours.** `tests/guard_integration_test.rs:345` carries the identical assertion on `develop`; it's pre-existing and git-version-dependent, and it passes here on git 2.53. Don't let it hold your test plan — I'll deal with it separately.

### Next

Fix the rake `:` boundary and this merges. Everything else above is optional or mine to carry.

Rounds: 1/3.

axelray-dev:
Addressed the reviewer request by treating the colon in namespaced rake and Rails test commands as a valid boundary and adding regressions for rake test:unit, rails test:system, bundle exec rake test:models, and bin/rails test:integration. The targeted token-boundary test passed. The full suite reached one unrelated existing git-log error-message assertion; no touched code was involved. Please re-review commit e5518dc.

axelray-dev:
Checked the current head (5518dc) against the requested review change. The Ruby rule already includes the : boundary, and the regression matrix covers 
ake test:unit, 
ake test:models, 
ails test:system, undle exec rake test:models, and in/rails test:integration. On the VPS, cargo fmt --all --check passed and the discover test suite passed: 606 passed, 0 failed. No additional code change was necessary; please re-review the current head.

axelray-dev:
Addressed the remaining Round 1 regression in the current head: the rake/rails rule now accepts a colon after test, so namespaced tasks such as rake test:unit, rails test:system, and bundle exec rake test:models keep their rewrite while rake tester and rake test-helper remain untouched. The focused boundary test passes. The full local suite reached 3538 passing tests and exposed one unrelated pre-existing failure in git_log_malformed_digit_run_propagates_real_git_error; it is outside this discover-rule change. Please re-review the current head.

PR #4024, opened 2026-09-13
fix(gh): count skipped and cancelled checks in pr checks summary

### LINKED ISSUE #4021: gh pr checks: skipping and cancelled checks are dropped from the summary, so the totals do not add up
## Summary

`format_pr_checks` in `src/cmds/git/gh_cmd.rs` counts only the `pass`, `fail` and `pending`
buckets. Every other bucket `gh` emits — `skipping`, `cancel` — is dropped, so the summary's
totals do not add up to the number of checks that ran, and a run that was cancelled reads as
if nothing happened.

## Reproduction

`gh` 2.46.0 and 2.100.0 (Docker), `LC_ALL=C`, against a real PR with skipped jobs:

```sh
$ gh pr checks 3982 -R rtk-ai/rtk | cut -f2 | sort | uniq -c
      2 fail
      8 pass
      5 skipping      # 15 checks total

$ rtk gh pr checks 3982
CI Checks Summary:
  [ok] Passed: 8
  [FAIL] Failed: 2     # 5 checks unaccounted for
```

A cancelled run — routine, since a new push cancels the in-progress one via the concurrency
group — is worse, because every row is dropped:

```
build           cancelled  45s  https://...
test (ubuntu)   cancelled  44s  https://...
Security Scan   skipping   0    https://...

CI Checks Summary:
  [ok] Passed: 0
  [FAIL] Failed: 0
```

Nothing in that output says the run was cancelled.

## Notes

This is pre-existing: the older substring-matching implementation dropped the same rows, so
it is not a regression from #3542, which rewrote the function to key on the tab-delimited
bucket field. #3542 is the natural place for the fix to land on top of.

A minimal fix is to count the remaining buckets rather than discarding them, so the totals
always sum to the checks that ran:

```
CI Checks Summary:
  [ok] Passed: 8
  [FAIL] Failed: 2
  [skip] Skipped/cancelled: 5
```

Verified locally against `gh pr checks 3982`: totals then reach 15, and savings stay at 81%
(well above the 20% floor).

RonaldoAleo:
## Summary

Fixes #4021.

`format_pr_checks` in `src/cmds/git/gh_cmd.rs` only counted the `pass`, `fail` and `pending` buckets. The two other buckets `gh pr checks` emits, `skipping` and `cancel` (see `cli/cli` `pkg/cmd/pr/checks/aggregate.go`), were rejected by `parse_pr_check_line`, so they never reached the summary. Totals did not add up to the checks that ran, and a run cancelled by a superseding push read as an empty run.

## Change

- `PrCheckStatus` gains a `Skipped` variant.
- `parse_pr_check_line` maps `skipping` and `cancel` to it instead of dropping the row.
- The summary prints `[skip] Skipped/cancelled: N` when `N > 0`, mirroring how `Pending` is only shown when present, so the common all-green case is byte-identical to before.

The `--watch` deduplication is untouched: a check that transitions `pending -> skipping` is still counted once, with its final status.

## Before / after

Real `gh pr checks 3982` output (8 pass, 2 fail, 5 skipping; 15 checks), fed through `rtk gh pr checks 3982`:

```
# before                              # after
CI Checks Summary:                    CI Checks Summary:
  [ok] Passed: 8                        [ok] Passed: 8
  [FAIL] Failed: 2                      [FAIL] Failed: 2
                                        [skip] Skipped/cancelled: 5
  Failed checks:                        Failed checks:
    doc review  fail  13s  ...            doc review  fail  13s  ...
    fmt         fail  7s   ...            fmt         fail  7s   ...
```

A cancelled run (2 `cancel`, 1 `skipping`):

```
# before                              # after
CI Checks Summary:                    CI Checks Summary:
  [ok] Passed: 0                        [ok] Passed: 0
  [FAIL] Failed: 0                      [FAIL] Failed: 0
                                        [skip] Skipped/cancelled: 3
```

Savings on the real 15-check PR: 1432 bytes raw -> 285 bytes filtered (80%), unchanged in practice and well above the 20% floor.

## Tests

Three unit tests added next to the existing `format_pr_checks` tests (TDD: the first two failed on assertion before the fix, all pass after):

- `test_format_pr_checks_counts_skipped_and_cancelled` — mixed table, totals add up.
- `test_format_pr_checks_reports_a_cancelled_run` — the all-cancelled case is reported, not silent.
- `test_format_pr_checks_omits_skipped_line_when_nothing_was_skipped` — guards the token savings of the all-green case.

## Checklist

- [x] Single-scope change (`gh` module only), targets `develop`
- [x] `cargo fmt --all --check && cargo clippy --all-targets && cargo test` pass
- [x] Tests added
- [x] No docs update needed: bug fix, and the summary format is not documented

https://claude.ai/code/session_01WqVBMVJVbnYH56Gama65UD


RonaldoAleo:
Closing: the same fix landed on develop via #1122 (d105227) shortly after this was opened, and it also covers the cancelled-run case. Thanks for the quick turnaround on #4021.

PR #4033, opened 2026-09-13
fix(tracking): categorize psql and mysql as cloud (#4023)

### LINKED ISSUE #4023: categorize_command buckets psql (and mysql) as "other" instead of "cloud" in rtk gain
`categorize_command` in `src/core/tracking.rs` has no arm for `psql`, so every `rtk psql` run is bucketed as `"other"` in `rtk gain`'s category breakdown rather than `"cloud"`, where `docker` and `kubectl` already sit. `mysql` (added in #3153) inherits the same gap.

```rust
"docker" | "kubectl" => "cloud",
```

There is no `psql`/`mysql` arm, and the fallthrough is `_ => "other"`.

Pre-existing — #3153 matches the existing psql behaviour rather than introducing this — so it was left out of that review. The fix is one arm:

```rust
"docker" | "kubectl" | "psql" | "mysql" => "cloud",
```

Worth checking the other `src/cmds/cloud/` entry points (`aws`, `curl`, `wget`) against the same list while touching it.

anujsingh-cse:
Closes #4023

## Problem

`categorize_command` in `src/core/tracking.rs` had no arm for `psql` (or `mysql`), so every `rtk psql` / `rtk mysql` run was bucketed as `"other"` in `rtk gain`'s category breakdown instead of `"cloud"`, where `docker` and `kubectl` already sit.

## Fix

Add `psql` and `mysql` to the `cloud` arm:

```rust
"docker" | "kubectl" | "psql" | "mysql" => "cloud",
```

I also checked the other `src/cmds/cloud/` entry points (`aws`, `curl`, `wget`) the issue flagged: `aws` is a cloud-provider CLI but `curl`/`wget` are generic network tools, and re-bucketing them would rewrite existing `rtk gain` output for users who already have those commands recorded. I kept this change scoped to the database CLIs the issue names; happy to add `aws`/`curl`/`wget` if you'd prefer them in `cloud`.

## Tests

Added `test_categorize_psql_and_mysql_as_cloud` asserting both `rtk psql` and `rtk mysql` classify as `cloud`.

`cargo fmt --check` and `cargo clippy --all-targets` clean; the two categorize tests pass.


anujsingh-cse:
Rebased onto the latest \develop\ (this branch has conflicted twice already as \#891\ and \#1298\ landed in \src/core/tracking.rs\). It's currently mergeable, tests, \mt\, and \clippy\ are all green — happy to address anything in review.

PR #4039, opened 2026-09-14
fix(gain): report hook warning on every rtk gain text view

### LINKED ISSUE #4035: rtk gain reports a missing hook on only one of its five views — never on the empty report, where it matters most
## Summary

`rtk gain` reports a missing or outdated hook on only one of its five invocations: the default view, and only when there is tracking data. `--daily`, `--weekly`, `--monthly`, `--all` and the empty report all stay silent.

The empty report is the worst case. With no hook installed nothing is tracked, so "no data" is usually *caused by* the missing hook — and that is exactly the run where the user is told nothing:

```
$ rtk gain
No tracking data yet.
Run some rtk commands to start tracking savings.
```

The advice never appears, because `rtk gain` is also exempt from the daily `[rtk] /!\ No hook installed` reminder — it is exempt on the grounds that it prints its own.

## Cause

In `src/analytics/gain.rs`, the `hook_check::status()` block sits inside the default-view branch (`if !daily && !weekly && !monthly && !all`), which is itself after the `summary.total_commands == 0` early return. Two gates, so four of five views and the empty report never reach it.

## Reproduction

Measured with an isolated `HOME` holding a `.claude` directory and no hook registered, counting `[warn] No hook installed` lines:

| invocation | warnings |
|---|---|
| `rtk gain` (with data) | 1 |
| `rtk gain --daily` | 0 |
| `rtk gain --weekly` | 0 |
| `rtk gain --all` | 0 |
| `rtk gain` (empty database) | 0 |

## Expected

Every text view reports it, including the empty one. The JSON and CSV exports return before this point and should stay free of it.

AngelVRodC:
## Summary

- `rtk gain` printed the missing/outdated hook warning only on the default view **with** tracking data. `--daily`, `--weekly`, `--monthly`, `--all` and the empty report were silent — and the empty report is the worst case, because with no hook installed nothing is tracked, so "No tracking data yet" is usually *caused by* the missing hook.
- Extract `warn_hook_issues()` and call it once in `run()` after the JSON/CSV early return, before the summary load / `total_commands == 0` early return. All five text views plus the empty report now show the warning exactly once; exports stay clean by construction (they return earlier).
- Out of scope and untouched: `check_rtk_disabled_bypass()` and the untrusted-filters warning (still default-view-gated), and the daily `[rtk] /!\ No hook installed` reminder's `gain` exemption in `main.rs` (rationale preserved — gain prints its own).

Closes #4035

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — fmt clean, 0 clippy warnings, 3218 tests pass
- [x] New integration test `tests/gain_hook_warning_test.rs` (9 cases, Unix-gated, isolated `HOME`): warning present exactly once on default/`--daily`/`--weekly`/`--monthly`/`--all`/empty-DB views; absent from `--format json` and `--format csv`; absent when no `~/.claude` exists (not-applicable state stays silent)
- [x] Manual reproduction of the issue scenario: isolated `HOME` with `.claude` and no hook, empty DB → `[warn] No hook installed` now appears on `rtk gain`
- [x] Manual testing: `rtk gain` output inspected

| File | Change |
|------|--------|
| `src/analytics/gain.rs` | Hoist hook warning out of default-view branch into `warn_hook_issues()`, single call before empty-report return |
| `tests/gain_hook_warning_test.rs` | New: 9-case integration test covering all views, empty report, exports, and N/A state |


KuSh:
DUPLICATE, recommend closing — superseded by #4038, which was opened for the same issue (#4035) about three hours earlier.

Closing this one on timing alone would waste the better half of it, so it has not been: **your nine-case matrix is now the test file in #4038**, credited to you in the file header and as `Co-Authored-By` on the commit.

It was better than what #4038 had in four ways, each of which I checked by mutating the implementation:

- it **counts occurrences** instead of asserting `contains`, so a double-print fails rather than passing
- it **seeds real tracking data** with `rtk proxy true`, so the view flags exercise the with-data path — the earlier tests ran against an empty database, where `--daily` and friends hit the early return and never reached the code under test
- it **separates stdout from stderr**, and pins that the JSON export still starts with `{`
- it pins the **not-applicable state** (no `~/.claude` → silent), which is the false-positive guard the earlier tests lacked

Three cases were added on top, for the `suppress_hook_warning` interaction. That is the one thing to flag back: both this PR and #4038's first revision **dropped the `hook_warning_suppressed()` gate** on the missing-hook arm, because both were written before #776 landed. Rebased on current `develop`, this PR's diff also reverts the `recalls` / `show_recall_stats` path and the `legacy_tee_migration_pending` notice, for the same reason — the base has moved under it. Worth knowing for next time: a diff read against the branch point hides this, and reading it against current `develop` is what surfaces it.

One housekeeping note, since it is not your fault and would have stalled this regardless: the CLA bot never reported on this PR at all — no pass, no pending. If you send another PR and see no CLA check appear, that is a bot problem worth flagging rather than waiting on.

Thanks — the matrix is the part of this that ships.

PR #4040, opened 2026-09-14
fix(tests): make dotnet TRX tests deterministic via explicit mtimes

### LINKED ISSUE #4037: Flaky test: dotnet_trx test_parse_trx_files_in_dir_since_ignores_older_files depends on file mtimes
## Summary

`cmds::dotnet::dotnet_trx::tests::test_parse_trx_files_in_dir_since_ignores_older_files` fails intermittently. It is not tied to any change — it flakes on unmodified `develop`-era trees.

```
thread '…test_parse_trx_files_in_dir_since_ignores_older_files' panicked at
src/cmds/dotnet/dotnet_trx.rs:576:75:
merged summary
```

## Reproduction

Run it repeatedly in isolation:

```
cargo test --bin rtk test_parse_trx_files_in_dir_since_ignores_older_files
```

Observed rates on this machine:

- 1 failure in 5 runs on the #1422 branch
- 2 failures in 6 runs on that PR's merge-base (`da3c4322`), which predates the change entirely

So it is pre-existing and independent of the code under test in either tree.

## Likely cause

The test name says it: it filters TRX files by "since" a timestamp, so it depends on file mtimes. Files written within the same clock tick — or a filesystem whose mtime granularity is coarser than the test assumes — make "older" and "newer" indistinguishable, and the `.expect("merged summary")` at `dotnet_trx.rs:576` then unwraps a `None`.

A fixed, explicitly-set mtime on the fixture files, rather than one derived from write order, would make it deterministic.

## Why it is worth fixing

It is a false red in CI on an unrelated PR, which is the kind of failure that trains reviewers to re-run rather than read.

AngelVRodC:
Fixes #4037

## Summary

- Replace wall-clock-dependent fixture mtimes in three flaky dotnet tests with explicit `File::set_modified` timestamps (old = base − 1 day, new = base + 1 hour), deriving every `since` cutoff from the same `base` so results are independent of filesystem mtime granularity (HFS+/FAT 1–2s flooring).
- Remove all `thread::sleep` synchronization from these tests; add a tiny `#[cfg(test)]` `write_trx_with_mtime` helper in `src/cmds/dotnet/mod.rs` (inline module — `automod::dir!` would declare a new file as a production module).
- Fixes the reported flake in `test_parse_trx_files_in_dir_since_ignores_older_files` plus two sibling latent flakes found during investigation (`test_find_recent_trx_in_dir_picks_newest_trx`, `test_merge_test_summary_from_trx_ignores_stale_fallback_file`).

## Test plan

- [x] `cargo fmt --all && cargo clippy --all-targets && cargo test` — 0 warnings, 3781 passed / 0 failed
- [x] Repeated the dotnet TRX test suites 20× consecutively — 0 flakes, and the suites no longer sleep (each run finishes in ~0.02s)
- [x] Test-only diff: 3 files in `src/cmds/dotnet/`, +37/−16; no production code, no new dependencies


KuSh:
Thanks for the fix and for finding the two sibling flakes. Issue #4037 was closed by #4048 (merged 2026-09-17), which covers the same three tests: old fixtures are pinned ten seconds before the cutoff and new ones ten seconds after through an explicit set_modified, and the sleeps are gone, so the dotnet TRX suites no longer depend on filesystem mtime granularity. The shared write_trx_with_mtime helper here is the one thing develop does not have, and with a per module set_mtime helper already in place that did not look like a reason to keep this open. Please comment or reopen if you think a part of this was missed.

PR #4048, opened 2026-09-14
fix(tests): make dotnet_trx mtime-dependent tests deterministic

### LINKED ISSUE #4037: Flaky test: dotnet_trx test_parse_trx_files_in_dir_since_ignores_older_files depends on file mtimes

TechWizard9999:
Fixes #4037

**Problem**: Three tests were flaky because they relied on filesystem mtime granularity and write ordering:

1. `dotnet_trx::tests::test_parse_trx_files_in_dir_since_ignores_older_files`
2. `dotnet_trx::tests::test_find_recent_trx_in_dir_picks_newest_trx`
3. `dotnet_cmd::tests::test_merge_test_summary_from_trx_ignores_stale_fallback_file`

All three used `std::thread::sleep` (5-10 ms) to create a time gap between file writes, then compared mtimes against a threshold. On filesystems with coarse mtime granularity (1 s on some CI filesystems, FAT 2 s buckets) the mtimes become indistinguishable, so "older" and "newer" can no longer be told apart and the assertion flips.

**Solution**: set mtimes explicitly with wide gaps (±10 seconds), which is immune to any realistic filesystem granularity.

- Old files are set to `now - 10s`, new files to `now + 10s`, with the threshold at `now`
- Uses `std::fs::File::set_modified` (stable since Rust 1.75, and the crate requires 1.91), so no new dependency is needed
- Verified: 40 consecutive runs of each test pass deterministically

**Note**: the first draft of this PR added the `filetime` crate as a dev-dependency. That introduced a second `[dev-dependencies]` table in `Cargo.toml`, which collided with the one `develop` gained in the edition-2024 series (#3782) and made the merged manifest fail with `error: duplicate key`. The maintainer fixup in 1281273e switches to the std equivalent, which drops the `Cargo.toml`/`Cargo.lock` changes entirely, and extends the same treatment to the third test above.


TechWizard9999:
@KuSh Could you please review it once ? 
Thanks!!

KuSh (approved):
## Round 1 of 3 — APPROVED

**Claim**: make the two mtime-dependent `dotnet_trx` tests deterministic by setting mtimes explicitly instead of relying on `thread::sleep` plus filesystem granularity (Fixes #4037).

**Scope**: accept. The diagnosis in #4037 is right, and it is fixed at the right level — the ordering guarantee moves from "whatever the filesystem records" to an explicit value.

**Ran**: two implementation mutations; 40× determinism loop; `cargo test --all` against the PR merged into current `develop`; `cargo check` on that same merge before the fixup.

**CI** @1281273e: all 11 green, Windows included.

### Blocks merge — resolved

1. **`Cargo.toml`: merged with current `develop`, the tree did not build.** The PR added a `[dev-dependencies]` table after `which = "8"`; `develop` grew its own (`temp-env`) in the edition-2024 series. Git merges both without a conflict marker, and cargo then rejects the manifest outright:

```
error: duplicate key
  --> Cargo.toml:60:2
   |
60 | [dev-dependencies]
```

Your CI was green and still is — it just could not have seen this. Your run finished 2026-09-14 11:05; #3782 landed on `develop` at 20:21 the same day. Green on the branch, unbuildable after merge.

### Fixup pushed — `1281273e`

Rather than reshuffle the manifest, the fixup removes the need for one. `std::fs::File::set_modified` has done this in std since Rust 1.75, comfortably under the crate's `rust-version = "1.91"`, so `filetime` was not buying anything. Dropping it takes the `Cargo.toml` and `Cargo.lock` hunks with it, and that is what makes the merge clean.

It also extends the same treatment to `test_merge_test_summary_from_trx_ignores_stale_fallback_file` in `dotnet_cmd.rs`, the one remaining test where a 5 ms sleep was the only thing separating "stale" from "fresh". Forcing that fixture's mtime 1 s forward makes it fail, so the mechanism was live, just with a narrower trigger than the two you found. There are now no sleep-ordered mtime tests left in `src/cmds/dotnet/`.

Verified on the fixup merged into current `develop`: one `[dev-dependencies]` table, `cargo fmt --check` and `cargo clippy --all-targets` clean, `cargo test --all` green (3613 unit + all integration), 40/40 on the two tests.

### Checked and correct

- **Your fix does not hollow out either assertion.** Flipping `max_by_key` to `min_by_key` in `find_recent_trx_in_dir` still fails `test_find_recent_trx_in_dir_picks_newest_trx`; disabling the `since` filter still fails `test_parse_trx_files_in_dir_since_ignores_older_files` (`left: 5, right: 3`). That was the thing worth checking — a determinism fix that quietly stops testing anything is the usual way this goes wrong, and yours does not.
- The future (`now + 10s`) mtimes are deliberate and harmless: they widen the gap on both sides of `since`, and the production path only ever reads mtimes.

### Your questions

- *"Could you please review it once?"* → Done. Good catch and a correct diagnosis; the only thing between it and merge was a manifest collision that appeared on `develop` nine hours after your CI ran.

### Next

Nothing — approving. CI is green on `1281273e`.

Issue #2519, opened 2026-06-19
bug(init): global Claude init fails when ~/.claude is missing


elite133713:
### Summary
`rtk init -g` fails on a fresh user profile when the Claude config directory does not already exist.

The failure occurs before RTK can create `RTK.md` because the atomic write implementation creates a temporary file inside the target file's parent directory, and that parent directory is missing.

### Steps to reproduce
Run global init on a machine or temp home where `~/.claude` does not exist:

```bash
rm -rf ~/.claude
rtk init -g
```

Equivalent isolated reproduction:

```bash
tmp=$(mktemp -d)
HOME="$tmp/home" CLAUDE_CONFIG_DIR="$tmp/.claude" rtk init -g --auto-patch
```

### Actual behavior

```
rtk: Failed to write RTK.md: /Users/rm-rf/.claude/RTK.md: Failed to create temp file in /Users/rm-rf/.claude: No such file or directory (os error 2)
```

### Expected behavior
`rtk init -g` should create the missing Claude config directory and then write:

- `~/.claude/RTK.md`
- `~/.claude/CLAUDE.md`
- `~/.claude/settings.json` when patching is enabled

Dry-run mode should continue to report intended writes without creating the directory.

### Cause
`atomic_write()` uses `NamedTempFile::new_in(parent)`, so callers that write into global config paths must ensure the parent directory exists before attempting the atomic write.

### Fix PR
Proposed fix: #2518

Related but separate: #1840 addresses the Cursor-only `rtk init -g --agent cursor` path and missing `~/.cursor`; this issue is for plain/global Claude init with missing `~/.claude`.

### Validation in PR #2518
- Added regression tests for missing Claude config dir in default global init, hook-only global init, and dry-run.
- `cargo fmt --all --check`
- `cargo test missing_claude_dir -- --nocapture`
- `cargo test --all` (2212 passed, 8 ignored)
- `cargo clippy --all-targets -- -D warnings`
- Manual temp-home check: `./target/debug/rtk init -g --auto-patch`

pszymkowiak:
> <img src="https://raw.githubusercontent.com/wshm-dev/wshm/main/assets/wizard-icon.png" width="48" height="48"> **wshm** · Automated triage by AI

## 🔍 Automated Triage

| | |
|---|---|
| 🐛 **Category** | `bug` |
| 🟡 **Priority** | `medium` |
| 🎯 **Confidence** | 97% |

### Summary

[...]

PR #4049, opened 2026-09-14
fix(init): make --opencode additive instead of replacing Claude setup

### LINKED ISSUE #4046: `rtk init --opencode` appears to skip Claude hook installation
When running:

```bash
rtk init -g --opencode
```

I expected it to install the Claude Code hook **and** the OpenCode plugin, since the flag says “Install OpenCode plugin (in addition to Claude Code)”.

**Actual behavior:**
- OpenCode plugin was installed globally
- No Claude hook was installed
- Running `rtk init -g` separately was still required to register the Claude hook

**Expected behavior:**
`--opencode` should add OpenCode support **in addition to** the Claude Code setup, or the help text should be updated to make it clear that it changes the install flow.

Repro:
```bash
rtk init -g --opencode
```

Demo:

```
[0] [07:14:36] ~ > rtk init --help
Initialize rtk instructions for assistant CLI usage

Usage: rtk init [OPTIONS]

Options:
  -g, --global
          Add to global assistant config directory instead of local project file

      --opencode
          Install OpenCode plugin (in addition to Claude Code)
[...]
[0] [07:14:59] ~ > rtk init -g --opencode

OpenCode plugin installed (global).

  OpenCode: /home/cecile/.config/opencode/plugins/rtk.ts
  Restart OpenCode. Test with: git status


[0] [07:15:17] ~ > rtk init -g

RTK hook registered (global).

  Command:   rtk hook claude
  RTK.md:    /home/cecile/.claude/RTK.md (10 lines)

Patch existing /home/cecile/.claude/settings.json? [y/N] 
y

  settings.json: hook added
  Backup: /home/cecile/.claude/settings.json.bak
  filters:   /home/cecile/.config/rtk/filters.toml (template, edit to add user-global filters)
```

dopaemon:
Fixes #4046

## Problem

`rtk init -g --opencode` installed only the OpenCode plugin and skipped Claude hook registration, even though the flag's help text says "Install OpenCode plugin (in addition to Claude Code)". Users had to run `rtk init -g` separately afterwards.

## Root cause

`src/main.rs`: `let install_claude = !opencode;` — the flag *replaced* the Claude setup. `hooks::init::run` already handled the combined case via its `(true, opencode, …)` match arms, so nothing else was missing.

## Changes

- `src/main.rs`: `install_claude` is always `true`; `--opencode` is purely additive.
- `src/hooks/init.rs`: removed the now-unreachable `(false, true, …)` arm and `run_opencode_only_mode()`; updated the `rtk init --show` hint.
- `README.md`: `# Claude Code hook + OpenCode plugin` instead of "instead of Claude Code".
- `tests/init_opencode_additive_test.rs`: regression test — `rtk init -g --opencode --dry-run` under a temp `$HOME` must report both the OpenCode plugin and the Claude hook (RTK.md + settings.json).

## Verification

```
$ rtk init -g --opencode --dry-run
[dry-run] would create RTK.md: .../.claude/RTK.md
[dry-run] would create OpenCode plugin: .../.config/opencode/plugins/rtk.ts
[dry-run] would add @RTK.md reference to CLAUDE.md: .../.claude/CLAUDE.md
[dry-run] would patch settings.json: .../.claude/settings.json
```

`cargo fmt --all` + `cargo clippy --all-targets` clean, lib suite (3613 tests) and all integration test binaries pass locally. Note: `tests/guard_integration_test.rs` could not be run locally (its `git` subprocesses hang in my sandbox); it is untouched by this change and left to CI.


KuSh:
Thanks for this, and sorry for the duplicated effort — closing as **DUPLICATE, recommend closing** in favour of #4052.

The reason is timing on the issue rather than anything about the code: @TechWizard9999 asked to be assigned #4046 at 09:28Z and was assigned at **11:02:21Z**; this PR was opened at **11:23:40Z**, about 20 minutes after the issue was already assigned. #4052 is the assignee's PR, so that one carries.

For the record, I reviewed both against real runs rather than assuming they were interchangeable, and they are equivalent on correctness — both share the same blocking regression:

```
fresh HOME, no ~/.claude
develop d402152f : rtk init -g --opencode --auto-patch -> EXIT=0  plugin installed
#4049   43f078e7 : rtk init -g --opencode --auto-patch -> EXIT=1  nothing installed
#4052   1ee24fe6 : rtk init -g --opencode --auto-patch -> EXIT=1  nothing installed
```

Routing `--opencode` through the Claude path exposes #2519 (`~/.claude` is never created), and since `RTK.md` is written before `ensure_opencode_plugin_installed`, the plugin never lands. Details and the verified fix are on #4052 and #2519.

One thing your PR does that #4052 doesn't, and that I'd like to keep: removing `run_opencode_only_mode` and the `(false, true, _, _)` arm once `install_claude` is unconditionally `true`. I checked, and it doesn't trip `dead_code` or clippy (the arms still reference the function), so it's cosmetic rather than required — but it's the right cleanup and I've pointed @TechWizard9999 at it to carry over, with credit to you.

Your diagnosis in the PR body was accurate, and the `--dry-run` regression test was a good instinct. Please do pick up another issue — and it's worth checking the assignee field first to avoid this happening again.

PR #4052, opened 2026-09-14
fix(init): install Claude hook with OpenCode

### LINKED ISSUE #4046: `rtk init --opencode` appears to skip Claude hook installation

TechWizard9999:
## Summary
- Make `rtk init -g --opencode` install the Claude hook and OpenCode plugin together.
- Update OpenCode help and setup documentation.
- Add end-to-end coverage for additive install, Claude-only install, and local-scope rejection.

## Verification
- `cargo fmt --all --check`
- `cargo test --test init_opencode_test`
- `cargo test --all` reaches two unrelated `search_faithful_test` timeouts because this environment's bundled ripgrep does not flush piped output.

Fixes #4046


TechWizard9999:
@KuSh Could you please review this PR?

src/main.rs
```diff
@@ -2368,7 +2368,7 @@ fn run_cli() -> Result<i32> {
                 hooks::init::run_vibe_mode(global, hook_only, patch_mode, ctx)?;
             } else {
                 let install_opencode = opencode;
-                let install_claude = !opencode;
+                let install_claude = true;
```
KuSh, line 2371: **Blocks merge (1a).** This line routes `--opencode` into the Claude install path, which writes `~/.claude/RTK.md` at `src/hooks/init.rs:1459` *before* the plugin is installed at `:1464` — and that write never `create_dir_all`s the Claude dir (pre-existing #2519).

Measured, fresh `HOME` with no `~/.claude`:

```
develop d402152f : rtk init -g --opencode --auto-patch -> EXIT=0  plugin=YES
head    1ee24fe6 : rtk init -g --opencode --auto-patch -> EXIT=1  plugin=NO
```

An OpenCode-only user who has never installed Claude Code now gets a hard failure and **no plugin at all**, where `develop` installed it. Same on `--hook-only`. This is also the `windows-latest` CI failure, because `dirs::home_dir()` ignores `$HOME` there and uses the runner's absent `C:\Users\runneradmin\.claude`.

Fix that I built and verified (fmt + clippy clean, full suite green):

```rust
let claude_dir = resolve_claude_dir()?;
if !dry_run {
    fs::create_dir_all(&claude_dir).with_context(|| {
        format!(
            "Failed to create Claude config directory: {}",
            claude_dir.display()
        )
    })?;
}
```

at `src/hooks/init.rs:1451` and again at `:1249` for the settings path. It also fixes plain `rtk init -g` on a fresh `HOME`, i.e. closes #2519.

tests/init_opencode_test.rs
```diff
@@ -0,0 +1,54 @@
...
+        "{}\n{}",
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr)
+    )
+}
+
+#[test]
+fn opencode_global_install_also_registers_claude_hook() {
+    let home = TempDir::new().expect("create temporary home");
+    std::fs::create_dir_all(home.path().join(".claude")).expect("create Claude config directory");
```
KuSh, line 28: **Blocks merge (1, evidence).** This line is load-bearing, and that's the problem: it's the only reason the additive install passes. Remove it and the test fails with

```
additive install failed:
rtk: Failed to write RTK.md: .../.claude/RTK.md: Failed to create temp file ... (os error 2)
```

For a change whose whole point is "OpenCode users also get the Claude hook", the fresh-`HOME`-without-`~/.claude` case is the one that has to be covered — it's the OpenCode-only user, exactly the population `--opencode` serves.

Two mutations I ran to check the test itself: reverting `install_claude` to `!opencode` *does* fail this test, so it genuinely checks its claim — good test. But dropping this one setup line also fails it, on the bug rather than the claim.

Once `run_default_mode` creates the directory, this line can just go, and the test passes without it. I verified that.

KuSh (changes requested):

**Claim**: `rtk init -g --opencode` installs the Claude hook *in addition to* the OpenCode plugin (Fixes #4046).

**Scope**: **accept** — this PR is the survivor over #4049 (you were assigned #4046 at 11:02:21Z; #4049 was opened at 11:23:40Z). #4049 shares the same regressions below, so it has no correctness advantage; I'm closing it as a duplicate. Its one worthwhile extra is the dead-code removal noted under Optional.

**Ran**: 4 binaries — merge-base `d402152f`, this head `1ee24fe6`, #4049, and this head merged onto current `develop` — one `CARGO_TARGET_DIR` each, all md5s distinct. Merge onto current `develop` is clean. 6-case `rtk init` matrix under an isolated `HOME` with `LC_ALL=C`, plus two mutation tests. Savings floor N/A: this is an init/hook change with no filtered command output.

**CI** @`1ee24fe6`: **2 failing** — `test (windows-latest)` is this PR's new test, same root cause as the finding below; `benchmark` is an infra flake (`sum.golang.org` stream error while fetching golangci-lint deps), unrelated to this change.

**1. `--opencode` now rides the Claude install path, so every Claude-path failure or early-exit takes the OpenCode plugin down with it.** All three `--opencode` combinations regress against `develop`:

| fresh `HOME`, no `~/.claude` | `develop` | this head |
|---|---|---|
| `init -g --opencode --auto-patch` | exit 0, plugin installed | **exit 1, nothing installed** |
| `init -g --opencode --hook-only --auto-patch` | exit 0, plugin installed | **exit 1** |
| `init -g --opencode --claude-md`, re-run after deleting `rtk.ts` | plugin restored | **exit 0, prints "already up to date", plugin still missing** |

The third one is the worst: it exits 0 and reports success while leaving the plugin missing, and re-running `rtk init` is the documented repair action.

Two mechanisms:

**(a)** `run_default_mode` (`src/hooks/init.rs:1451`) and the settings patch (`src/hooks/init.rs:1249`) resolve `claude_dir` and then write into it without `create_dir_all`. That is pre-existing #2519, but this PR newly makes it reachable from `--opencode` — and reachable on Windows CI, where `dirs::home_dir()` ignores `$HOME` and resolves the runner's real (absent) `C:\Users\runneradmin\.claude`. That is exactly the `windows-latest` failure:

```
rtk: Failed to write RTK.md: C:\Users\runneradmin\.claude\RTK.md: Failed to create temp file
in C:\Users\runneradmin\.claude: The system cannot find the path specified. (os error 3)
```

Note the ordering in `run_default_mode`: `RTK.md` is written at :1459, `ensure_opencode_plugin_installed` at :1464 — so the Claude failure aborts before the plugin is ever installed.

**(b)** `run_claude_md_mode_with` returns early on `RtkBlockUpsert::Unchanged` (`src/hooks/init.rs:~1918`) **before** the `install_opencode` block, so the plugin is only ever installed when the instruction block happened to change.

### Verified fix

Built and run, not guessed — `cargo fmt --all --check` and `cargo clippy --all-targets` clean, 3613 lib tests plus every integration binary green:

- `fs::create_dir_all(&claude_dir)`, guarded by `!dry_run`, at `src/hooks/init.rs:1451` and `:1249`
- hoist the `if global && install_opencode { … }` block above the `Unchanged` early return

With that, all six cases pass, plain `rtk init -g` on a fresh `HOME` works too (which closes #2519), `--dry-run` still creates nothing, and existing `settings.json` keys are preserved. And your own test then passes **without** its `create_dir_all` line — which is the point below.

- `tests/init_opencode_test.rs` needs `#![cfg(unix)]` (see `tests/hook_warning_scope_test.rs:8`), or Windows-aware home handling: `.env("HOME", …)` is a no-op on Windows.
- `run_rtk` doesn't isolate `XDG_CONFIG_HOME`, `XDG_DATA_HOME` or `RTK_DB_PATH`. On a machine that exports them, the test writes into the real `~/.config/rtk/filters.toml` and the real tracking DB. Clean on the runners, which is precisely the "passes here, differs elsewhere" shape.
- `run_opencode_only_mode` and the `(false, true, _, _)` / `(false, false, _, _)` arms are now unreachable, since `hooks::init::run` has one caller and it passes `true` unconditionally. I verified this does **not** trip `dead_code` or clippy, so it's cosmetic — but #4049 removes it cleanly and it's worth taking along.

Merges cleanly onto current `develop`; happy-path install and idempotence; existing `settings.json` user keys preserved; local-scope `--opencode` rejection; `--show` and `--uninstall` symmetry; sibling agent branches (cursor / windsurf / cline / codex / gemini / copilot) unaffected — they return before this block, and `cursor` already had `install_claude == true`; non-TTY stdin does not hang on the new settings prompt; all four docs wording changes accurate.

- "Leaving `run_opencode_only_mode` in place trips `dead_code` and reddens clippy" — false, clippy is clean; the arms still reference it.
- "The `benchmark` failure is this PR's" — false, Go module proxy error.
- "An `error[E0308]` in the local suite is a real compile error" — false, it's fixture text printed by a filter test.

### Your questions

- "Hi @KuSh , I have made changes please review it" → the head is still the single commit `1ee24fe6`, so nothing was pushed. Flagging it so you're not waiting on a round for a push that didn't land.

### Next

Add the two-part fix above, drop the `create_dir_all(".claude")` line from the test, and gate the test `#![cfg(unix)]`. That turns `windows-latest` green and empties the blocking set — round 2 approves.

Rounds: 1/3. Threads: 2 open (both mine, waiting on the fix), 0 resolved this round.

PR #4053, opened 2026-09-14
fix(pip): stop doubling the tool name in pip messages

### LINKED ISSUE #4050: pip error messages say "pip pip" (tool name doubled)
## Description:

When `rtk pip ...` fails, the error message repeats the tool name:

`$ rtk pip list`
 `rtk: Failed to run pip pip list: Failed to execute command: No such file or directory (os error 2)`

`$ rtk -v pip list`
`Running: pip pip list --format=json`

The same happens for `pip` outdated and passthrough commands like `pip install`.

## Cause 

In `src/cmds/python/pip_cmd.rs` the messages are built as `format!("Failed to run {} pip list", base_cmd)`. The hardcoded pip was written
for the uv case (`uv pip list`), but `base_cmd` is also "pip" on the normal path, giving `pip pip`. Six strings are affected. The `eprintln!` lines at
74, 102, 128 and the `.with_context` lines at 78, 106, 132.

The command actually run is correct (`pip list --format=json`), only the message is wrong. This is therefore just a cosmetic issue.

## Fix: 

Add a small helper that returns "uv pip" when base_cmd == "uv" and "pip" otherwise, and use it in all six strings.

Related:  #3389 (pip/uv routing rule. This issue doesn't change it).

obrienciaran:
## Summary

- `rtk pip ...` error and verbose messages printed `pip pip list` because the format strings were written for the `uv pip` branch and reused on the plain-pip path.
- Adds a `prog_label()` helper ("uv pip" or "pip") and uses it in all six affected strings in `src/cmds/python/pip_cmd.rs`. The executed argument was correct; only the message was wrong. This is just a costmetic fix.
- Fixes #4050

## Test plan

- [x] cargo fmt --all && cargo clippy --all-targets && cargo test
    - clean, 0 clippy warnings; new unit test `test_prog_label_never_doubles_pip`
  
- [x] Manual testing:
    - with a pip on PATH that can't start, `rtk -v pip list` / `pip outdated` / `pip install` now print
        - `Running: pip list --format=json` and `Failed to run pip list: ...` (was `pip pip list`). 
    - With only uv on PATH, output is unchanged: 
        - `Running: uv pip list --format=json`.


Protocol-zero-0 (commented):
Checked this locally. `base_cmd` can only be `"pip"` or `"uv"` (line 28), so `prog_label` covers every case, and no `"{} pip"` strings are left in the file. On the uv fallback the verbose line is now right:

```
pip not found — falling back to `uv pip`
Running: uv pip list --format=json
```

**The same label problem, in reverse, is still in the `timer.track` call a few lines up** (`run`, ~line 50):

```rust
timer.track(
    &format!("{} {}", base_cmd, args.join(" ")),
    &format!("rtk {} {}", base_cmd, args.join(" ")),
    ...
```

On the plain path that records `pip list` / `rtk pip list`, which is fine. On the uv path it records a command that never ran. With a stub `uv` on PATH and no `pip`, `rtk pip list` leaves this row in the tracking DB:

```
('uv list', 'rtk uv list')
```

What actually ran was `uv pip list`, and what the user typed was `rtk pip list`, so `rtk gain --history` shows a `uv list` entry that doesn't correspond to either. Using the helper there too would line it up:

```rust
&format!("{} {}", prog_label(base_cmd), args.join(" ")),   // "uv pip list"
&format!("rtk pip {}", args.join(" ")),                     // what the user ran
```

Up to you whether that belongs in this PR or a follow-up, since it's the same "wrong tool name in a label" class.

**Overlap:** your #4055 also edits `src/cmds/python/pip_cmd.rs`, and a trial merge of the two branches conflicts, so one will need a rebase after the other lands.

obrienciaran:
@Protocol-zero-0 Thanks for reviewing it. I admit, it took me a moment to follow this explanation 😅 I get it now. Good catch. We can use `prog_label()` there too, and this reversed problem will also be fixed.

Issue #3389, opened 2026-08-04
pip install silently installed nothing on 0.40.0 (uv routed without a venv); regression test + docs still describe the old rule


Minrit:
## Summary

On rtk **0.40.0**, `rtk pip install <pkg>` installed **nothing** on any host that has
both `pip` and `uv` and no active virtualenv. The unconditional `use_uv` in
`pip_cmd.rs` routed to `uv pip install`, and uv refuses a non-virtualenv target:

```
error: No virtual environment found; run `uv venv` to create an environment, or pass `--system` to install into a non-virtual environment
```

Bare `pip install` on the same host succeeds, because real pip falls back to
user-site when system `site-packages` isn't writable. uv has no such fallback and
does not support `--user`.

The root cause is **already fixed** in v0.41.0 (`use_uv = !tool_exists("pip") && tool_exists("uv")`),
so this is not a request to re-fix it. It's a field report that the fix's blast
radius was much larger than the `pip list`-shows-the-wrong-env symptom it was
written for, plus three small follow-ups that are still open on `develop`.

## What happened

An AI coding agent running in a container image (Debian, glibc 2.41, mise-managed
Python 3.13, `uv` installed alongside `pip`) was asked to query an Excel file with
DuckDB. It emitted:

```
pip install duckdb -q 2>&1 | tail -2; python3 -c "import duckdb; print(duckdb.__version__)"
```

The PreToolUse hook rewrote it to `rtk pip install duckdb -q …`, rtk ran
`uv pip install`, uv printed the venv error, nothing was installed, and the
follow-up import raised `ModuleNotFoundError`. The agent reported the dependency
step as failed and continued without the package. Every `pip install` an agent
issued on that image failed the same way.

`python3 -m pip install duckdb` was **not** rewritten by the hook, so that spelling
kept working — which made the failure look intermittent and tool-dependent rather
than systematic.

## Repro (rtk 0.40.0)

Any environment where `pip` and `uv` are both on PATH, `VIRTUAL_ENV` is unset, and
system `site-packages` is not writable by the running user:

```bash
rtk pip install duckdb -q; echo "rc=$?"
# error: No virtual environment found; run `uv venv` ...
python3 -c "import duckdb"          # ModuleNotFoundError

pip install duckdb -q; echo "rc=$?"
python3 -c "import duckdb; print(duckdb.__file__)"
# /home/<user>/.local/lib/python3.13/site-packages/duckdb/__init__.py
```

[...]

PR #4055, opened 2026-09-14
fix(pip): fall back to uv when pip is on PATH but cannot start

### LINKED ISSUE #4054: A `pip` that exists but can't run blocks the `uv` fallback
## Description:

`rtk pip ...` falls back to `uv pip` when `pip` isn't installed. But if `pip` is on `PATH` and just can't start, for example a leftover script whose first line points at a Python that's been deleted, RTK still tries to run it, fails, and never tries `uv`.

`$ head -1 $(which pip)`
`#!/usr/local/opt/python/bin/python3.7    # this Python no longer exists`

`$ rtk pip list`
`rtk: Failed to run pip list: Failed to execute command: No such file or directory (os error 2)`

With that same `pip` removed, `rtk pip list` works via `uv`. This is common on Intel Macs where old Homebrew Pythons leave stale pip scripts behind.

## Cause:

`src/cmds/python/pip_cmd.rs:27` decides with `!tool_exists("pip") && tool_exists("uv")`, and `tool_exists` (`src/core/utils.rs:606`) only checks that a file named `pip` is on PATH and marked executable. It doesn't check the file can actually run. The comment above line 27 says to fall back when pip "genuinely isn't on PATH", so this is a gap in how "genuinely" is checked.

## Suggested fix:

Either check that `pip` can actually start (not just that it exists), or fall back to `uv` when running `pip` fails with "No such
file or directory". Do not switch to always preferring `uv`. That was tried and reverted (see #3389) because it picked the wrong environment.

## Related:
- #3389: why pip must stay the default when present
- #3155: same kind of PATH-lookup problem for pytest
- #4050: the error message here; the missing program name in it is being addressed in #2634
- #3904, #1280: broader pip/uv routing discussions

obrienciaran:
use_uv only asked whether a `pip` file exists on `PATH` and doesn't check if it can actually run. Right now, a stale console script whose shebang names a deleted interpreter passes the check, so rtk exec'd it, gets ENOENT, and never tries `uv`; even though the same machine works via `uv pip` once the dead script is removed.

Keep the routing rule from #3389 (`pip` wins whenever it is present) but retry through `uv pip` when the spawn itself fails with `NotFound` and `uv` is available. Nothing ran on a spawn failure, so the retry is safe for install too. A one-line stderr notice says why the switch happened.

Fixes #4054

## Summary

  - `rtk pip ...` now retries via `uv pip` when `pip` is on `PATH` but the OS can't start it (e.g. a stale shebang). Previously it failed with a bare "No such file or directory" and never tried `uv`.
  - Routing is unchanged otherwise. A working `pip` still wins whenever present (#3389). The fallback only kicks in after a failed spawn, so the normal path costs nothing new. When it does trigger, the cost is one failed `exec` (which was going to happen anyway), one extra `PATH` lookup for `uv`, and the `uv pip` run itself; plus one extra `stderr` line in the output.
  - Moved subcommand dispatch into a `dispatch()` helper, and added `is_spawn_not_found()`. 
  - Also updated the stale "Auto-detect uv" line in `docs/contributing/ARCHITECTURE.md`.

## Test plan

    - clean, 0 clippy warnings. Three new unit tests, including a `#[cfg(unix)]` one that writes a real dead-shebang script to a tempdir, spawns it, and checks it's classified as not-found.
- [x] Manual testing with a sandbox PATH and a stub uv that echoes its arguments:
  - Dead `pip` + `uv`: prints `rtk: pip is on PATH but could not be started — falling back to uv pip`, then `pip list` and `pip install` succeed via `uv (exit 0)`.
  - No `pip` + `uv`: unchanged, falls back the same as always.
  - Dead `pip`, no `uv`: same as alway, error kept, exit 1.
  - Working `pip` + `uv`: `pip` runs, `uv` is never called


obrienciaran:
Both this and #4053 both touch `src/cmds/python/pip_cmd.rs` near `run()` and will conflict. Whichever merges second needs a rebase. Both are mine; ping me once one lands and I'll rebase the other.

PR #4066, opened 2026-09-14
fix(read): fall back to raw bytes when a plain read is not valid UTF-8

### LINKED ISSUE #4065: `rtk read <file>` exits 1 with no output on a non-UTF-8 file, while `--head-lines`/`--tail-lines` on the same file are byte-exact
A plain read of a file that isn't valid UTF-8 exits 1 and prints nothing:

```console
$ printf 'line one\nline two has a latin-1 byte: \xe9\nline three\n' > /tmp/nonutf8.log
$ rtk read /tmp/nonutf8.log
cat: /tmp/nonutf8.log: invalid utf-8 sequence of 1 bytes from index 38
$ echo $?
1
```

Same file, same binary, with a line window:

```console
$ rtk read --head-lines 3 /tmp/nonutf8.log | cmp - /tmp/nonutf8.log && echo identical
identical
```

`run()` takes the byte path only when `byte_line_window()` returns `Some`, and that needs `head_lines` or `tail_lines`. A plain read has neither, falls through to `String::from_utf8`, and dies. **The byte path is already byte-exact, it just isn't reachable without a window.**

The hook decides which path a command lands on:

| rewritten to | result |
|---|---|
| `cat f` → `rtk read f` | exit 1, no output |
| `head -2 f` → `rtk read f --head-lines 2` | works |
| `tail -1 f` → `rtk read f --tail-lines 1` | works |

So `cat` on a log with one Latin-1 byte in it hands the agent an empty result, while `head` on the same file is fine. An empty result reads like an empty file.

CONTRIBUTING's Never Block rule covers this case: fall back to raw output instead of erroring out.

PR #1019 fixed it in April with a narrow gate (unfiltered, no window, no `-n`). A maintainer confirmed the bug on 2026-05-18 and asked for a rebase, the author rebased, and the branch then sat unreviewed until the stale bot closed it on 2026-09-14. I'll send a fresh PR against current `develop`.

- `develop` @ d402152, macOS 26.6.2 arm64
- also on 0.49.0 from brew, where the message reads `stream did not contain valid UTF-8`

Thanks!

sypsyp97:
## Summary

- `rtk read <file>` writes the bytes out unchanged when nothing downstream needs a `&str`, so a file that isn't valid UTF-8 reads the way it does under `cat` instead of exiting 1 with no output.
- Filtering, `-n` and `--max-lines` still decode and still report the error, since each of them works on text.
- Closes #4065.

## Motivation

`run()` reaches the raw-byte path only when `byte_line_window()` returns `Some`, which needs `head_lines` or `tail_lines`. A plain read has neither and falls through to `String::from_utf8`. The Claude Code hook splits `cat f` (rewritten to `rtk read f`, fails) from `head -2 f` (rewritten to `rtk read f --head-lines 2`, works) on the same file.

Same gate as #1019, which the stale bot closed on 2026-09-14 after it sat unreviewed. This one is written against current `develop`.

## Test plan

3 unit tests on the new `raw_byte_window()` gate, 2 integration tests (`--ignored`) that run the built binary against a Latin-1 file. With the one-line fallback dropped and the tests kept, `cargo test read::tests:: -- --include-ignored` is red:

```text
test cmds::system::read::tests::test_raw_byte_window_passes_a_plain_read_through ... FAILED
test cmds::system::read::tests::test_read_non_utf8_file_is_byte_exact ... FAILED
test result: FAILED. 29 passed; 2 failed; 0 ignored; 0 measured; 3595 filtered out; finished in 0.78s
```

and green with it:

```text
test result: ok. 31 passed; 0 failed; 0 ignored; 0 measured; 3595 filtered out; finished in 0.59s
```

`cargo test --all` on a clean `HOME`:

```text
test result: ok. 3616 passed; 0 failed; 10 ignored; 0 measured; 0 filtered out; finished in 1.46s
```

On my own machine one test fails both before and after this change: `test_claude_pipeline_rewrites_only_safe_final_stage` reads `~/Library/Application Support/rtk/config.toml`, I have `grep` in `exclude_commands`, so the rewrite it unwraps is `None`. Same failure on unmodified `develop` — same family as #2640.

Manual:

```console
$ rtk read /tmp/nonutf8.log | cmp - /tmp/nonutf8.log && echo identical
identical
$ rtk read -l minimal /tmp/nonutf8.log; echo $?
cat: /tmp/nonutf8.log: invalid utf-8 sequence of 1 bytes from index 38
1
```


sypsyp97:
@pszymkowiak — you confirmed this one on #1019 in May, so you already have the context. Same narrow gate, written against current `develop`, with tests. CLA is signed.

sypsyp97:
Rebased onto `develop` (5e0f92c). The let-chains refactor and the 2024 style-edition reformat both touched the lines this changes, so the conflict was in `read.rs`; resolved by keeping the new `let bytes = fs::read(file)...` formatting and routing it through `raw_byte_window()`. Numbers in the description were re-run on the new base.

dopaemon (approved):
Checked this out locally, built it, and exercised the gate against a Latin-1 file (`\xe9`). The change is correct and behavior-preserving for every path that still needs text.

| case | result |
|---|---|
| `rtk read nonutf8.log` | byte-exact (`cmp` clean), exit 0 |
| `rtk read -l minimal` / `-n` | still errors, exit 1 — correct, filtering needs `&str` |
| valid UTF-8 plain read | identical output, no regression |
| no trailing newline / empty file | verbatim, exit 0 |
| `--max-lines` | unchanged |

`cargo clippy --all-targets` clean, `cargo test --bin rtk read::tests:: -- --include-ignored` → 31 passed. `byte_line_window()` is still tried first inside `raw_byte_window()`, so head/tail windows keep winning over the new pass-through exactly as before.

## One gap: `run_stdin()` still fails on non-UTF-8

`run_stdin()` carries a copy of the same gate and wasn't touched, so the sibling path is still broken on this branch:

```console
$ cat nonutf8.log | rtk read -
cat: -: invalid utf-8 sequence of 1 bytes from index 18   # exit 1
```

Same root cause, and `raw_byte_window()` is already the shared helper — it looks like a one-line swap in `run_stdin` plus one test. Since the framing here is "`rtk read` should behave like `cat`", leaving `rtk read -` failing means pipelines (`curl … | rtk read -`) still hit the original bug. Happy either way on doing it here or as a follow-up, but it's worth not losing.

## Nits (non-blocking)

- `raw_byte_window` takes 6 positional params; `(level, line_numbers)` are really one "needs decoded text" question.
- `test_raw_byte_window_keeps_line_windows` passes `max_lines` + `tail_lines` together, which clap's `conflicts_with_all` makes unreachable from the CLI. Harmless as a unit test, just noting it isn't a reachable state.
- Tracking on the raw path counts both sides via `from_utf8_lossy`, so stats for non-UTF-8 files are approximate. Pre-existing, cosmetic.

dopaemon (approved):

sypsyp97:
Thanks — you were right, and it was worth not losing. Fixed here in `7cb3a04` rather than as a follow-up, since `rtk read -` failing is the same bug and splitting it would leave the pipeline case broken between merges.

`run_stdin()` now calls `raw_byte_window()` with the same arguments the file path passes, so `byte_line_window()` has a single caller again.

| case | before | after |
|---|---|---|
| `cat nonutf8.log \| rtk read -` | exit 1, no output | byte-exact (`cmp` clean), exit 0 |
| `… \| rtk read - -l minimal` | exit 1 | exit 1 — filtering still needs `&str` |
| `… \| rtk read - -n` | exit 1 | exit 1 — same |
| `… \| rtk read - --head-lines 2` | worked | unchanged, window still wins |
| valid UTF-8 stdin | fine | identical output |
| empty stdin | fine | 0 bytes, exit 0 |

New `#[ignore]` integration test `test_read_stdin_non_utf8_is_byte_exact` writes the Latin-1 bytes into a piped child and asserts stdout is verbatim. `cargo test --bin rtk read::tests:: -- --include-ignored` → 32 passed (was 31). Full suite on a clean `HOME`: 3616 passed / 0 failed. `cargo fmt --all --check` and `cargo clippy --all-targets` clean; semgrep against the PR base is 0 findings.

On the nits:

- **`test_raw_byte_window_keeps_line_windows`** — agreed, `--max-lines` + `--tail-lines` is unreachable past `conflicts_with_all`, so the assertion was covering a state that cannot occur. Dropped the `max_lines` argument; the tail window itself is still asserted.
- **6 positional params** — left as-is deliberately. Folding `(level, line_numbers)` into a "needs decoded text" type reads better in isolation, but the argument list here is the same shape `run()` and `run_stdin()` already take from the CLI layer, so the wrapper would have to be built and destructured at both call sites to save one bool. Happy to do it if you'd rather see it.
- **`from_utf8_lossy` in tracking on the raw path** — agreed it's approximate for non-UTF-8 input, and it's pre-existing. Making it exact means `track()` taking bytes, which touches every caller; out of scope for this gate, but I can open a separate issue for it.

PR #4076, opened 2026-09-15
fix(pytest): preserve elapsed duration in summaries

### LINKED ISSUE #4071: pytest filter drops the duration line, so a 3-second run and a 10-minute run look identical
rtk 0.49.0, Homebrew, macOS arm64 (Darwin 27.0.0), Python 3.11.8, pytest 7.4.0.

## What happens

`rtk pytest` drops pytest's summary line, so the elapsed time is gone from the output.

```
$ cat test_slow.py
import time

def test_fast():
    assert 1 == 1

def test_slow():
    time.sleep(3)
    assert 2 == 2

$ rtk proxy pytest test_slow.py | tail -3
test_slow.py ..                                                          [100%]

============================== 2 passed in 3.04s ===============================

$ rtk pytest test_slow.py
Pytest: 2 passed
```

Same on the failing path:

```
$ rtk pytest test_fail.py
Pytest: 0 passed, 1 failed

Failures:
1. [FAIL] test_boom
     test_fail.py:4: in test_boom
     assert 1 == 2
     E   assert 1 == 2
```

No duration anywhere.

## Why this one hurts

I run pytest through the Claude Code hook, so the model only ever sees the filtered output. A tool result carries no timing of its own, the model cannot tell a call that took 0.2s from one that took 10 minutes. Pytest's summary line was the only place that number existed, and the filter removes it.

The practical effect is that a suite which quietly goes from 12s to 10 minutes reads exactly the same as before. Nothing in the agent's context says anything changed, so nobody looks into it.

## Other filters already keep it

This looks like an oversight in the pytest filter, not a project-wide policy. From the inline filter tests in the 0.49.0 binary:

- gradle: `expected = "BUILD SUCCESSFUL in 8s\n7 actionable tasks: 7 executed"`
- gradle (failing): `expected = "> Task :app:test\n3 tests completed, 1 failed\nBUILD FAILED in 12s"`
- cargo: `expected = "   Finished release [optimized] target(s) in 45.23s\n   Binary: target/release/my-crate (5.2MB)"`

and vitest parses `Duration\s+([\d.]+)(ms|s)`, ctest parses `Total Test time (real) = ([\d.]+) sec`, dotnet captures a `Duration:` group.

## Suggestion

Put the seconds back on the summary line:

```
Pytest: 2 passed in 3.04s
Pytest: 0 passed, 1 failed in 1.02s
```

That is roughly 3 tokens, and it is the only timing signal an agent gets for the whole run.

## Workaround does not exist today

A project-local `.rtk/filters.toml` cannot patch this, since pytest is routed to the Rust module. Trusted and confirmed applied, output unchanged:

```toml
schema_version = 1

[filters.pytest-keep-duration]
description = "keep pytest duration line"
match_command = "pytest"
strip_ansi = true
keep_lines_matching = ["passed in", "failed", "error"]
max_lines = 20
```

```
$ rtk trust -y
    pytest-keep-duration pytest
  Enabled — revoke with `rtk untrust`.
Filters will now be applied.

$ rtk pytest test_slow.py
Pytest: 2 passed
```

That part is #3904, so I am not asking for it here. Just the duration line.

igorzelaya-io:
## Summary

- preserve pytest's elapsed-duration token in compact success summaries
- preserve the same timing signal when failures, skips, xfails, or xpasses are reported
- cover wrapped and quiet pytest summary formats

## Why

Without the duration, a fast suite and a severe performance regression produce the same filtered output. This keeps pytest's original value (for example, `3.04s`) while adding only the small timing signal the agent otherwise cannot recover from the tool result.

Fixes #4071.

## Verification

- `cargo fmt --all --check`
- `cargo test --all` (3,614 unit tests plus all integration tests passed)


Protocol-zero-0 (commented):
Tried this locally with real pytest 8.4.2. Short runs look right:

```
Pytest: 2 passed in 0.01s
Pytest: 1 passed, 1 failed, 1 skipped in 0.04s
```

**The duration is dropped once a run passes 60 seconds, though.** pytest appends the wall-clock form after the seconds:

```
========================= 1 passed in 61.07s (0:01:01) =========================
```

`parse_duration` takes everything after the last `" in "`, i.e. `61.07s (0:01:01)`. That doesn't end in `s`, so `strip_suffix('s')` returns `None`, and the filtered output is just:

```
Pytest: 1 passed
```

That's the case the PR description is about (telling a fast suite apart from a slow regression), so it seems worth covering. Taking the first token after `" in "` would handle both forms:

```rust
let token = duration.split_whitespace().next()?;   // "61.07s"
token.strip_suffix('s')?.parse::<f64>().ok()?;
Some(token.to_string())
```

and a test with `"=== 1 passed in 61.07s (0:01:01) ==="` -> `Pytest: 1 passed in 61.07s` would pin it.

**Overlap:** #4192 (collect-only summaries) changes the same `PytestCounts` / `build_pytest_summary` / `parse_summary_line` code, and a trial merge of the two branches conflicts in `src/cmds/python/pytest_cmd.rs`. Whichever lands second will need a rebase; if this one goes first, the collected-only summary from #4192 could pick up the duration too.

PR #4118, opened 2026-09-19
fix(git): make the blob recovery hint work under RTK's own hook

KuSh:
Fixes one regression from the release review on #3979, introduced by #3265.

Part of a set of six, one per originating PR: #3681, #3552, #3265 (this), #3772, #3857, #3941.

## The regression

A windowed `git show <rev>:<path>` prints a head and points at the rest of the file:

```
... (+1434 lines) [see remaining: git show 'HEAD:src/core/utils.rs' | tail -n +249]
```

Under RTK's own hook, the bare `git show` in that hint is rewritten back into `rtk git show`, which windows the dump a second time — so `tail -n +249` had nothing left to print but the hint itself.

```
$ rtk git show HEAD:src/core/utils.rs | tail -1
... (+1434 lines) [see remaining: git show 'HEAD:src/core/utils.rs' | tail -n +249]

$ git show 'HEAD:src/core/utils.rs' | tail -n +249     # following it verbatim, hook active
1 line                                                 # 1434 expected
```

## The fix

Name the command the hook leaves alone. `rtk proxy` is the documented escape hatch and the `diff` size hints in `src/cmds/git/diff_cmd.rs` already use it the same way:

```
... (+1434 lines) [see remaining: rtk proxy git show 'HEAD:src/core/utils.rs' | tail -n +249]
```

248 shown lines + 1434 recovered = 1682 = the native `git show` total, byte for byte.

## Verification

`tests/git_show_blob_differential_test.rs` gained `windowed_blob_hint_command_returns_the_rest`, which **runs the command the hint names** instead of simulating it — it parses the hint's own argv, executes it, and asserts head + recovered output reconstructs the blob byte for byte. Against the old hint it fails with `the hint returned 1 line(s) -- it must return the rest of the blob, not just itself`, so it is a real guard for exactly this regression; the previous tests reconstructed the tail with native git and could never have seen it.

A unit test also asserts the emitted hint is left untouched by `discover::registry::rewrite_command`, which is the function all three hook front-ends funnel through.

`docs/usage/FEATURES.md` quoted the old hint string literally; it now shows the `rtk proxy` form and notes that a bare `git show` under the hook re-windows.


pszymkowiak (approved):
Reviewed by building the branch and following the hint the way an agent would, with the hook simulated through `rtk rewrite`.

**Verified**
- develop: the hint names bare `git show 'HEAD:src/core/utils.rs' | tail -n +250`; `rtk rewrite` turns it into `rtk git show … | tail -n +250` (exit 3); following it yields 1 line (the hint itself); head + rest ≠ native (`sha256` mismatch, 8189+84 bytes vs 65645).
- this PR: hint names `rtk proxy git show '…' | tail -n +250`; `rtk rewrite` leaves it alone (exit 1, `rtk `-prefixed segment returned unchanged at `registry.rs:1691`, `tail -n +N` cannot match `TAIL_N_SPACE`); following it yields 1593 lines; head + rest == native byte for byte (8189 + 57456 = 65645, same sha256).
- Adjacent shapes: path with a space, `<sha>:<path>`, `rtk git -C <abs> show …` from outside the repo (hint carries `'-C' '<path>'` through proxy), small blob → no hint. All correct.
- `compact_blob_show` has a single production caller (`git_cmd.rs:606`); no stale copy of the old hint string remains in `src/`, `tests/` or `docs/`.

One precision note, no code change needed: `windowed_blob_hint_command_returns_the_rest` executes the hint's `rtk proxy git show` argv and applies the `tail` in-process, but does not run the hook; the hook-survival guarantee is the unit test `test_compact_blob_show_hint_survives_rtks_own_hook` calling the real `rewrite_command`. Together they cover both halves; the doc comment on the integration test could say so.

PR #4123, opened 2026-09-19
fix(tracking): keep the user's arguments out of the telemetry command label

KuSh:
From the release review on #3979, which flagged `low_savings_commands` under #1298. Independent of the six regression PRs — it shares no file with any of them and branches off `develop` directly.

## The leak

The telemetry label was the first three words of `rtk_cmd`. That field is RTK's own, but most of what it holds is the **user's command line**, so the third word is routinely an operand:

```
rtk ls /usr/bin                           445 rows
rtk curl https://mockhttp.org/robots.txt   26
rtk tree src/                              20
rtk ls /tmp/claude-1000/-home-user-…/scratchpad/…
rtk wc src/main.rs
rtk ls --all=x
```

The TOML-filter path is worse: it stores the user's entire invocation after an `rtk:toml` prefix, so the label could carry a `jq` program and the file path it ran against.

Measured over a real history database — 329,583 rows, 26,934 distinct `rtk_cmd` — **201 of 2606 labels carried a path, a URL, a flag value or a search term.**

`top_passthrough` had exactly this leak and was fixed in `15f666dd` by grouping on the tool alone. `low_savings_commands` was left behind.

## Why "just take fewer words" doesn't work

Grouping on the first word collapses everything, since every label begins with `rtk`. Taking two words loses the granularity that makes the figure worth sending — `rtk git log` and `rtk git status` are different filters with different savings.

Taking three words when the first is `rtk` and two otherwise only moves 201 → 195: it fixes the `rtk:toml` family but not the main leak, because for a **single-level** command the third word *is* the operand.

The distinction is not word count. It is whether the word is RTK's routing or the user's data.

## What it does now

Keep the prefix, the tool as a **basename**, and a third word **only** when the tool routes by a subcommand of its own *and* that word is shaped like a subcommand (`^[a-z][A-Za-z0-9:_-]*$`).

| rule | distinct labels | carrying an argument |
|---|---|---|
| `take(3)` (before) | 2606 | **201** |
| first word `rtk` ? 3 : 2 | 2589 | 195 |
| routing-aware | 2054 | 3 |
| + basename on the tool | 2054 | **0** |

The basename step clears the last three (`rtk ./gradlew`, `rtk ./mvnw`, `rtk:toml /tmp/rtk-pr`) and *merges* `./gradlew` with `gradlew`, so the labels improve rather than degrade. Granularity is intact: `rtk git log`, `rtk git status`, `rtk git show`, `rtk gh pr` all stay distinct.

## The router list

A superset of the tools that declare subcommands in `discover::rules`, which is the authority for the ones it covers — `command_label_matches_the_discover_rules` fails if a tool gains subcommands there and is not added here. The rest are listed because that table does not describe them: it carries no subcommands for `git`, the most routed tool there is.

Cross-checked by a test rather than read at runtime, because nothing in `core` depends on `discover` (the dependency runs the other way, 8 references) and this is not worth inverting it for.

That test paid for itself three times on its first runs, each a real defect in this change:

- `sbt testOnly` — the subcommand shape rejected camelCase
- `phpstan` — declares subcommands, was missing from the list
- `bun pm ls` — a two-word subcommand, which a three-word label cannot carry; it keeps `pm`, coarser than the table and never an argument

A tool missing from the list keeps two words. That costs granularity and never leaks — the direction this has to fail in.

## Not changed

The review also filed "telemetry query now admits negative-savings rows" under the same PR. That is deliberate and documented in the function's own docstring — *"A net-regressing command (negative rate) is listed: it is the filter most in need of improvement"* — added by the same commit that changed the predicate. It is also reachable, which the review could not establish: `rtk git worktree` sits at −7.7 % over 2103 calls in a real database. Left as-is.

## Verification

6 unit tests: the real leaking shapes, the routed subcommands that must survive, an operand in the subcommand slot being dropped anyway (`HEAD~3..HEAD`, `--git-dir=…`, a quoted path, a URL), degenerate input, the `discover::rules` cross-check, and a property test asserting no label contains `/`, `=`, `"`, `'`, `@`, `~` or `..`.

`cargo fmt --all`, `cargo clippy --all-targets` and `cargo test --all` are green on current `develop`.


src/core/tracking.rs
```diff
@@ -453,6 +454,97 @@ fn should_sample_cleanup(key: &str, rate: u32) -> bool {
     hasher.finish().is_multiple_of(rate as u64)
 }
 
+/// Tools that route by a subcommand of their own, so the word after them is still RTK's
+/// routing rather than something the user typed.
+///
+/// A superset of the tools that declare subcommands in `discover::rules`, which is the
+/// authority for the ones it covers -- `command_label_matches_the_discover_rules` fails if one
```
aeppling: Hey @KuSh :)

Comments could be cleaned up here

Fine to keep the list, no clean derivation possible, but could we make it discoverable for contributors?

A note at RULES and the *_SUBCOMMANDS consts saying a tool gaining subcommands needs adding here as well.

Rest looks good

aeppling (approved):

src/core/tracking.rs
KuSh: Both done in `5fcd048e`.

The doc dropped from twelve lines to six — the paragraph explaining why this is not read from `discover::rules` was rationale for a decision nobody is being asked to revisit, so it is gone, and what remains leads with the obligation rather than the derivation.

For discoverability I put the pointer at all three sites rather than one:

- `RULES` (`discover/rules.rs`): adding `subcmd_savings` or `subcmd_status` to a rule also means adding that tool to `SUBCOMMAND_ROUTERS`, or its label stops at the tool name.
- `NPM_SUBCOMMANDS` and `RUFF_SUBCOMMANDS`: same note, since those are the two `*_SUBCOMMANDS` consts in the tree and neither sits near the rules table.

The `RULES` note claims the test fails until the tool is added, so I checked that rather than assert it: removing `"git"` from `SUBCOMMAND_ROUTERS` fails `command_label_matches_the_discover_rules` at `tracking.rs:2045`, along with two other label tests. A contributor who misses the comment still gets stopped by CI, for the tools the table covers.

Worth flagging the limit of that: the guard only reaches tools `discover::rules` describes with subcommands. The hand-listed rest — `git` is in the table but carries no subcommands there — have no mechanical check, which is why the comment now says the list is by hand and why it is worth a note at the consts.

Gate on `5fcd048e`: fmt, clippy `--all-targets`, 3850 tests, all clean.

PR #4125, opened 2026-09-19
fix(cli)!: preserve argv boundaries in generic runners

pashifika:
## Summary

**PR A of the three-way split of #3911**, per @KuSh's round-1 review. Nothing here touches fish or the rewrite path: this is the CLI execution boundary on its own, against current `develop`.

`rtk run`, `rtk err`, `rtk test` and `rtk summary` rebuilt their positional arguments into one string and executed it through `sh -c` (`cmd /C` on Windows). Argument boundaries were lost and globs, variables and operators were expanded by a shell the caller never asked for:

```bash
rtk run printf '[%s]\n' 'a b' '*' '$HOME'
# develop:  [a]/[b]/[a.rs]/[b.rs]/[/home/you]     ← sh -c re-split and expanded
# here:     [a b]/[*]/[$HOME]
```

Positional forms now execute the program directly. An intentional script is explicit:

```bash
rtk run -c '<script>'                 # platform default: sh, or cmd on Windows
rtk run --shell fish -c '<script>'    # named shell; pwsh/PowerShell get -Command
rtk err|test|summary --shell sh '<script>'
```

Shell mode takes the complete script as one argument, so RTK never reconstructs quoting after Clap has already split argv.

## What the interposed shell used to provide, and now does not have to

Blocker 3 of the review, plus the two neighbouring contracts the same switch touched:

**1. Exit 127 for an unresolvable program.** Direct execution resolves the program itself (`resolve_binary`), and a miss is an execution outcome rather than an RTK error — it renders through each runner's own failure path:

```
$ rtk err nosuchcmd123
[FAIL] Command failed (exit code: 127)
  rtk: nosuchcmd123: command not found          exit 127

$ rtk run nosuchcmd123
rtk: nosuchcmd123: command not found            exit 127   (stderr, like a shell)
```

`rtk test` and `rtk summary` do the same through their own summarizers. 127 is the POSIX "command not found" code that CI steps and `[FAIL]`-line parsers key on; before this commit the head of #3911 replaced both with an `anyhow` chain and exit 1.

**2. `rtk test ! <cmd>`.** `!` is both `test`'s negation and the shell's. A command behind it is not a native `test` expression, so it goes to the test filter — and the negation the joined `sh -c` string used to get from the shell is now applied to the child's exit status, with argument boundaries intact. `tests/native_test_expression_test.rs::bang_before_a_command_still_runs_the_command` pins it.

**3. `--shell` arity.** `rtk run` spells its rule with clap (`-c` `requires`/`conflicts_with`); `err`/`test`/`summary` take a positional, and clap cannot express "exactly one value for this positional when that flag is present". The rule is enforced at the same place clap would answer, so it is a usage error (exit 2) instead of a mid-execution failure:

```
$ rtk err --shell sh printf hello
error: --shell takes the complete command as one quoted argument
Usage: rtk [OPTIONS] <COMMAND>
```

Also from the review's optional list: tracking labels re-quote arguments that carry whitespace, so `rtk test cargo test --filter "a b"` is recorded as the command that ran rather than as `cargo test --filter a b`.

## Breaking change

Callers that relied on implicit shell expansion in positional arguments must pass the script explicitly:

```bash
rtk test cargo test                              # unchanged — direct argv
rtk test --shell sh 'cargo test && cargo clippy' # was: implicit sh -c
rtk run -c 'ls *.rs | wc -l'                     # was: rtk run ls *.rs | wc -l
```

No registry rule emits `rtk run|err|test|summary`, so the hook path is unaffected; the blast radius is direct CLI callers who passed a quoted shell string. The awareness body shipped as `RTK.md` (`RTK_INSTRUCTIONS` in `src/hooks/init.rs`) is updated to teach the argv form and `--shell`, which is the other half of blocker 3.

## Key files

| File | Purpose |
|---|---|
| `src/core/shell.rs` | Direct-argv and explicit-shell command construction, per-shell flag selection, `Launch::NotFound`, display quoting |
| `src/core/runner.rs` | Renders an unresolvable program through the err/test filters, exit 127 |
| `src/cmds/rust/runner.rs`, `src/cmds/system/summary.rs` | argv/`--shell` entry points |
| `src/main.rs` | `--shell` surface, `!` negation, `--shell` arity usage error |
| `src/hooks/init.rs` | `RTK_INSTRUCTIONS` documents argv execution and `--shell` |
| `tests/run_shell_test.rs` | argv preservation, exit codes, default and explicit shells, missing shell, split-script rejection, the 127 contract |

## Test plan

- [x] `cargo test --all` — 3,864 passed, 0 failed, 8 ignored
- [x] Manual, isolated `HOME`/`XDG_CONFIG_HOME`: argv boundaries (`'a b'`, `*`, `$HOME` unexpanded), `-c` POSIX default, `--shell fish -c` fish syntax, exit 42/7 propagation, `err|test|summary|run` on a missing program → 127, `--shell` with split arguments → usage error
- [ ] CI for Linux and Windows (`cmd /C`, `-Command` flag selection)

Tested on macOS 26.6.2 arm64, rustc 1.97.1, fish 4.9.0.

---

Follow-ups from the same split, opened separately: **B** — `fish -c` wrapper rewriting on top of #3828, with the `(…)`/control-keyword deferral scoped to the wrapper's own script; **C** — fish-script wrapping at the hook gate, where the premise can be argued on its own. #3911 is closed in favour of the three.


src/core/shell.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+/// Build a command that preserves the argument boundaries supplied by Clap.
+pub fn direct_command(args: &[String]) -> Result<Launch> {
+    let Some((program, program_args)) = args.split_first() else {
+        bail!("command is required");
+    };
+
+    if resolve_binary(program).is_err() {
+        return Ok(Launch::NotFound(program.clone()));
```
KuSh: **Blocks merge 1 — the "cannot run this program" path is fixed here and unfixed one step later.**

`resolve_binary` answers only "is this resolvable". Once `Launch::Ready` is returned, every `execve` failure is a Rust `Err` that `?` turns back into the shape this PR removed. A script checked out with CRLF line endings is enough:

```
$ rtk err ./crlf.sh                       # '#!/bin/sh\r\n'
merge-base:  [FAIL] Command failed (exit code: 127)
             sh: 1: ./crlf.sh: not found                                  exit 127
head      :  rtk: Failed to run err command: Failed to run err:
             Failed to spawn process: No such file or directory (os error 2)  exit 1
dash reference:                                                           exit 127
```

Same errno, same `Context` chain, same lost `[FAIL]` line as `rtk err nosuchcmd123` had before this PR. Reproduces on `run`, `err`, `test` and `summary`, and on ENOEXEC too; an unresolvable `--shell` name is a third arm.

And on this line specifically: `which::which` returns one error for "nothing by that name" and for "it is there but you cannot run it", so a spelling that addresses the filesystem gets the wrong code *and* a false message —

```
$ rtk run ./bindir/noexec-prog        # exists, mode 644
merge-base: exit 126 "Permission denied"     head: exit 127 "command not found"
$ rtk run ./adir2                     # a directory
merge-base: exit 126                          head: exit 127
```

`dash` and `bash` agree: 126 for non-executable, directory and ENOEXEC; 127 only for genuinely missing. A bare `PATH` name without `+x` correctly stays 127 on both, so the divergence is confined to path-y spellings.

This is my miss as much as yours — in #3911 I named the ENOENT-at-resolution instance instead of the class, and you fixed exactly what I named. Details and a working prototype for the pre-spawn half are in the review body.

src/core/shell.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    let mut command = resolved_command(program);
+    command.args(program_args);
```
KuSh: **Blocks merge 2 — this is the widest caller-supplied-argv surface in the binary, and the only one that skips `child_args`.**

`ChildArgExt`'s own doc (`src/core/utils.rs:524`) says *"Use instead of `Command::arg`/`args` for anything that arrived on rtk's own command line — a pattern, a path, a flag value."* 26 call sites obey it, including `run_passthrough`. That helper is #3728, merged as this PR's own merge-base commit, so on Windows an argument containing `"` reaches MSYS/Node children with std's encoding instead of libuv's — which is the one place the PR's headline claim is not delivered.

Separately, `cmd /C` resolved from the current directory and carried ~30 builtins (`echo`, `dir`, `type`, `set`…). Direct execution does neither, so `rtk err dir`, `rtk summary echo hi` and `rtk err mytool` with `mytool.exe` in the cwd stop working on Windows. Unix loses nothing equivalent, which is why the breaking-change section reads complete while the larger Windows delta is undocumented — and all 10 new integration tests are `#[cfg(unix)]`, so no runner sees any of it.

I could not execute this one: no Windows machine, and the green `test (windows-latest)` job proves only that nothing covers it. Narrow ask: `child_args` instead of `args`, one `#[cfg(windows)]` test, and a line in the breaking-change section.

src/main.rs
```diff
@@ -240,13 +240,19 @@ enum Commands {
 
     /// Run command and show only errors/warnings
     Err {
+        /// Execute one quoted command string with this shell instead of direct argv execution
+        #[arg(long, value_name = "SHELL")]
+        shell: Option<String>,
```
KuSh, line 245: **Blocks merge 3 — `--shell` is the first flag these surfaces have ever had, and they are wired to swallow flag errors.**

`err`/`test`/`summary` are `PASSTHROUGH`, not `RTK_META_COMMANDS` (`src/core/constants.rs:11-33`), so a clap parse error under them falls through to `run_fallback`, which tries to exec a binary named `err`/`test`/`summary`. Forgetting the value after the new flag — the likeliest mistake with it — reports a file that does not exist:

```
$ rtk err --shell        merge-base: [FAIL] Command failed (exit code: 2)   exit 2
                         head:       [rtk: No such file or directory (os error 2)]   exit 127
$ rtk summary --shell    head: [rtk: No such file or directory (os error 2)]         exit 127
$ rtk test --shell       head: <no output>                                           exit 0
$ rtk run --shell        head: error: a value is required for '--shell <SHELL>'      exit 2
```

`run` is right because it *is* in `RTK_META_COMMANDS`. The `test` arm is the one that worries me: asking for a shell and getting silent success.

Moving `err` and `summary` into `RTK_META_COMMANDS` (plus the `trailing_var_arg` skip list in `test_meta_commands_reject_bad_flags`) makes both print the clap error at exit 2 with the suite green. `test` cannot move — it deliberately shadows the POSIX `test` binary — so it needs `--shell` validated by hand, the way `require_single_script` already validates arity.

KuSh (changes requested):

**Claim**: `rtk run|err|test|summary` stop joining positional argv into a string for an implicit `sh -c`/`cmd /C` and execute the program directly; a script becomes explicit via `rtk run -c` or `--shell <shell>`; exit 127 and the `[FAIL]` contract survive the switch.

**Scope**: **accept** (round 1, frozen). This is PR A of the split, it is the piece closest to landing, and the argv work is correct as claimed on Unix. The three blockers below are completion items on the new code, not a reason to narrow. It also supersedes **#3002**, which addresses the same defect by adding `quote_shell_arg`/`shell_quote_join` to re-quote argv *before* joining it for the shell, and covers `err`/`test` only. Removing the reconstruction is the better answer than making it correct, so I have closed #3002 pointing here.

**Ran**: merge-base `0924356b` vs head `790a747a`, release-built with one `CARGO_TARGET_DIR` each (md5 `ec117312…` / `32773c71…`, confirmed distinct; the merge onto current `develop` `b748a5f7` got a third). ~60 rows across the four commands × {argv boundaries, exit codes, `-c`, `--shell` default/named/missing, `--shell` arity, `!` negation, native `test` expressions, empty argv, path spellings, builtins, piped stdin, `/dev/null` stdin, signals}. Invariants: every branch of the new decision (Ready / NotFound / shell / native-test / negation); split streams on a failing `cargo build`; under the hook (`rtk rewrite` on all four, identical base vs head, and no `rules.rs` rule emits them). Savings **60.4%** vs the 20% floor (CONTRIBUTING.md) — and the filtered bytes are *byte-identical* base vs head, so the diff forfeits nothing. `cargo fmt --all --check` clean, `clippy --all-targets` 0 warnings, `cargo test --all` **3864 passed / 0 failed** on the head and **3940 / 0** on the head merged onto current `develop` (real cargo exit status, not a `tail`-masked pipeline).

**Matrix**: never executed on this PR before this round — executed now, in full.

**CI** @`790a747a`: the run had never been created (`action_required`); I approved it. Now **11 green, 1 failing**: `semgrep security scan` — the blocking `dynamic-command-execution` rule on `src/core/shell.rs:57`, caused by this PR, and `src/core/shell.rs` is a new file so `--baseline-commit` cannot absorb it. My prepared fixup fixes it (see Next). Windows, macOS and Ubuntu test jobs all pass, which is the first real signal this branch has had.

---

**1. `src/core/shell.rs:40` — the "program cannot be run" path was fixed at resolution and left unfixed at spawn, so blocker 3 of #3911 survives verbatim one step later.** This is my miss as much as yours: I named the ENOENT-at-resolution instance and you fixed exactly that, and I should have named the class.

`resolve_binary` answers only "is this resolvable". Once `Launch::Ready` is returned, every `execve` failure is a Rust `Err` that `?` turns back into the shape the fix removed. A script checked out with CRLF line endings — not an exotic input — is enough:

```
$ rtk err ./crlf.sh                       # '#!/bin/sh\r\n'
merge-base:  [FAIL] Command failed (exit code: 127)
             sh: 1: ./crlf.sh: not found                                  exit 127
head      :  rtk: Failed to run err command: Failed to run err:
             Failed to spawn process: No such file or directory (os error 2)  exit 1
dash reference:                                                           exit 127
```

Same errno, same three-level `Context` chain, same lost `[FAIL]` line as `rtk err nosuchcmd123` had before this PR. Reproduces on `run`, `err`, `test` and `summary`, and on ENOEXEC (`+x` file that is not a valid executable) as well. An unresolvable `--shell` name is the third arm of it (`exit 1` + chain, where an unresolvable *program* is a rendered `[FAIL]` + 127).

The second arm is pre-spawn: `which::which` returns one error for "nothing by that name" and for "it is there but you cannot run it", so a path-y spelling that exists reports the wrong code **and an actively false message**:

```
$ rtk run ./bindir/noexec-prog        # exists, mode 644
merge-base:  sh: 1: ./bindir/noexec-prog: Permission denied               exit 126
head      :  rtk: ./bindir/noexec-prog: command not found                 exit 127
$ rtk run ./adir2                     # a directory
merge-base:  exit 126      head: exit 127
```
`dash` and `bash` agree: **126** for non-executable, directory and ENOEXEC; **127** only for genuinely missing. A bare `PATH` name that exists without `+x` correctly stays 127 on both — the divergence is confined to spellings that address the filesystem.

**Class**: 2 causes (pre-spawn classification, post-spawn `io::Error`) × 4 surfaces, one mechanism each. The argument this PR makes for restoring 127 — "CI steps and `[FAIL]`-line parsers key on it" — is exactly the argument for 126 and for not replacing either with an `anyhow` chain.

**Prototype for the pre-spawn arm**, so this is a measurement and not a request to go write code: `Launch::NotFound(String)` → `Launch::Unrunnable { program, code, reason }`, `EXIT_COMMAND_NOT_EXECUTABLE = 126`, and one `names_existing_path` check that only fires for path-y spellings. 123 insertions across the five files, `fmt` clean, clippy 0, `cargo test --all` **3866 passed**, and it restores 126 on all four surfaces while leaving every 127 case — including the bare-name one — untouched. Say the word and I will push it as a fixup, or paste the patch here if you would rather read it first. The post-spawn arm I deliberately did **not** prototype: the spawn sites are in `src/core/stream.rs`, shared by every rtk command, so mapping errno there has a blast radius well beyond this PR and is your call whether it belongs here or in a follow-up.

**Test that closes it**: extend `tests/run_shell_test.rs` with a 126 case and a CRLF/ENOEXEC case across the four surfaces, mirroring `test_and_summary_report_an_unresolvable_program_as_exit_127`.

**2. `src/core/shell.rs:45`, `tests/run_shell_test.rs` — on Windows the new path neither uses the repo's argv-quoting helper nor replaces what `cmd /C` provided, and nothing tests it.**

`direct_command` is now the widest caller-supplied-argv carrier in the binary, and it is the one place that calls `command.args(program_args)`. `ChildArgExt`'s own doc (`src/core/utils.rs:524`) says *"Use instead of `Command::arg`/`args` for anything that arrived on rtk's own command line"*; **26 call sites obey it**, including `run_passthrough`. That helper is #3728 — merged as this PR's own merge-base commit. So for an argument containing `"`, MSYS/Node children on Windows get std's encoding instead of libuv's, and the PR's headline claim is the one thing not delivered there.

Separately, `cmd /C` resolved from the **current directory** and carried ~30 builtins (`echo`, `dir`, `type`, `set`, `copy`, `del`…). Direct execution does neither — `which` walks `%PATH%` only — so `rtk err dir`, `rtk summary echo hi`, and `rtk err mytool` with `mytool.exe` in the cwd go from working to exit 127 on Windows. Unix loses nothing equivalent (`sh` does not search `.`, and `echo`/`test`/`pwd` all exist as binaries), which is why the Unix breaking-change section reads complete while the larger Windows delta is undocumented. **All 10 new integration tests are `#[cfg(unix)]`**, so no runner sees any of this.

I could not execute this arm — I have no Windows machine, and the green `test (windows-latest)` job proves only that nothing *covers* it. The narrow ask is: `child_args` instead of `args`, one `#[cfg(windows)]` test, and a line in the breaking-change section. I have an unmerged branch making that encoding unforgettable at the type level (the `ChildCommand` newtype, #4103) — **that is not a reason to hold this PR**, and I am not asking you to wait for it; the one-line swap now is the right move and I will rebase mine onto it.

**3. `src/main.rs:245`, `:255`, `:336` — `--shell` is the first flag these three surfaces have ever had, and they are wired to swallow flag errors.**

`err`/`test`/`summary` are `PASSTHROUGH`, not `RTK_META_COMMANDS` (`src/core/constants.rs:11-33`), so a clap parse error under them falls to `run_fallback`, which tries to exec a binary called `err`/`test`/`summary`. Forgetting the value after the new flag — the likeliest mistake with it — reports a file that does not exist:

```
$ rtk err --shell
merge-base:  [FAIL] Command failed (exit code: 2)                    exit 2
head      :  [rtk: No such file or directory (os error 2)]           exit 127

$ rtk summary --shell        head: [rtk: No such file or directory (os error 2)]   exit 127
$ rtk test --shell           head: <no output>                                     exit 0
$ rtk run --shell            head: error: a value is required for '--shell <SHELL>' but none was supplied   exit 2
```

`run` is right because it *is* in `RTK_META_COMMANDS`. The `test` arm is the one that worries me: asking for a shell and getting silent success. This is reach widening — the swallowing path is older than the PR, but nothing reached it on these commands until the PR gave them a flag. Moving `err` and `summary` into `RTK_META_COMMANDS` (plus the `trailing_var_arg` skip list in `test_meta_commands_reject_bad_flags`) makes both print the clap error at exit 2 with the suite green; `test` cannot move, since it deliberately shadows the POSIX `test` binary, so it needs `--shell` validated by hand the way `require_single_script` already validates arity.

---

- **`quote_for_display` under-quotes, so the labels this PR adds are not re-runnable** — which is most of what I asked for in #3911's optional list. It quotes on whitespace and quote characters only, leaving bare exactly the metacharacters the PR exists to stop interpreting: `rtk err /bin/echo '*' '$HOME' 'a;b' 'x&&y'` is recorded and printed as `/bin/echo * $HOME a;b x&&y`, while its doc comment says a label "reads back as the command that ran". Same string feeds `-v "Running:"`, `summary`'s `[FAIL] Command:` header and `TestEcosystem::detect`. Wants a shell-safe metacharacter set, not a whitespace set.
- **`rtk test '(' <cmd> ')'` loses its subshell**: merge-base exit 1, head exit 127 `rtk: (: command not found`. `is_native_test_expression` deliberately treats `(` as a possible native prefix, but `split_leading_negations` only handles `!`, so the two functions now disagree about `(`. Either handle `(`/`)` beside `!`, or drop `(` from `is_native_test_expression` so they agree.
- **`report_not_found` drops the `-v "Running:"` line** the normal path prints (`rtk -v err rtk-no-such` shows it on the merge-base, not on the head).
- **Both `bail!` arms of `command_from_args` are unreachable.** All three production callers run `require_single_script` first, which `.exit()`s on `len != 1`, and `run` never calls it — so "pass the shell command as one quoted argument after --shell" can only be seen from a unit test, while users always get `require_single_script`'s differently-worded message. One rule, two texts, one reachable.
- **`require_single_script` prints the top-level usage** (`Usage: rtk [OPTIONS] <COMMAND>`) where the `run` surface prints `Usage: rtk run …`. Cosmetic, but it is the error people hit while learning the flag.
- **argv[0] changes for path-y spellings**: the child of `rtk run ./pb/tool` sees `$0 = /abs/path/pb/tool` where the merge-base's `sh` gave it `./pb/tool`. Only affects multi-call binaries and tools that print usage from `argv[0]`; bare `PATH` names are unchanged.
- **`report_not_found` books a second negative-savings row in `rtk gain`** (`rtk test …` 0.0% → −60%). Pre-existing shape, mildly widened — the merge-base already books −125% for `rtk err <missing>` — so evidence went to **#3702**, which already tracks it, rather than a new issue. The obvious fix is wrong, and worth recording so nobody tries it: wrapping `rendered` in `never_worse` (as `summary.rs` does) makes it print the raw line and **drops the `[FAIL]` line your own `err_reports_an_unresolvable_program_as_exit_127` asserts** — I applied it and watched that test fail. It needs a decision, not a one-liner.
- **Doc residue from a `fix(cli)!`**: `src/hooks/init.rs:277` still reads `rtk test <cmd>` in the Test section, 60 lines above the new caveat under "Analysis & Debug"; `INSTALL.md:323` and `README_{fr,es,pt,ja,ko,zh}.md` still show the old form. None carries shell syntax, so nothing breaks — they are just stale. Nothing documents the `!`-only negation restoration, or that `--shell` disables the native `test` passthrough.

### Filed as follow-ups

- **#3702** (`rtk gain` cannot represent a regression) — absorbs the negative-savings row above; I added the base-vs-head transcript there rather than opening a duplicate.

- **Argv preservation itself**, which is the point of the PR: `'a b'`, `*`, `$HOME`, `;`, `&&`, `|`, `$(…)`, backticks and an embedded newline all arrive at the child intact, where the merge-base re-split and expanded every one of them. One win the summary undersells: an **empty** argument used to vanish entirely — `rtk run printf '[%s]' "" x -- y` gives `[x][--][y]` on the merge-base and `[][x][--][y]` here. 5000 arguments round-trip cleanly on both.
- **The 127 + `[FAIL]` contract** on all four surfaces. I broke `EXIT_COMMAND_NOT_FOUND` to 1 and watched `err_reports_an_unresolvable_program_as_exit_127`, `test_and_summary_…` and `run_reports_…` all fail — the claim is genuinely pinned, not incidentally true.
- **`rtk test ! <cmd>`.** I removed the negation and `bang_before_a_command_still_runs_the_command` failed, so that test pins it too. The `is_native_test_expression` / `--shell` / negation ordering in the `Test` arm is right: `! -d adir` still reaches the real `test`, `! false` negates, `! !` cancels.
- **`--shell` arity** → clap usage error, exit 2, on all three positional surfaces; clap's own `requires`/`conflicts_with` on `run`. The #3911 asymmetry is gone.
- **The hook path is untouched**: `rtk rewrite` output is identical base vs head for all four commands, and no `rules.rs` rule emits them. The blast radius really is direct CLI callers.
- **The filter forfeits nothing**: on a failing `cargo build`, base and head produce byte-identical filtered output, 60.4% savings, streams captured separately.
- **`display_args` re-quoting** does not disturb `TestEcosystem::detect` (it substring-matches, and the quotes sit outside the token), and it fixes the tracking label from #3911's optional list.
- **The awareness body** now teaches the argv form and `--shell` in `RTK_INSTRUCTIONS`, which was the other half of #3911's blocker 3.
- **Startup got faster**, not slower: `rtk run /bin/true` is 1.7 ms ± 0.2 against the merge-base's 2.1 ms ± 0.2 (hyperfine, 50 runs, 1.28× ± 0.16) — one less `sh` in the middle. Both well inside the 10 ms target.

- **"Direct execution changes what the child sees on stdin."** Piped input, `/dev/null` and no-TTY are identical base vs head on all four commands (`printf 'a\nb' | rtk err /bin/cat`, `rtk run /bin/cat </dev/null`). Dropped.
- **"Signal exit codes regress."** They improve: `kill -TERM $$` gives 143 on the head where the merge-base's joined string produced a `kill` usage error and exit 2.
- **"`display_args` quoting breaks ecosystem detection."** `rtk test cargo test --filter "a b"` → `cargo test --filter 'a b'` → still `Cargo`.
- **"`resolved_command`'s PATHEXT/`.CMD`/`.BAT` fallback is lost on Windows."** It is not: that resolution lives in `which::which`, which `direct_command` calls first, and which-8.0.1 reads `PATHEXT`. What the short-circuit does drop is the last-ditch `Command::new(name)` that searched rtk.exe's own directory and System32 — minor, and folded into blocker 2.
- **"A bare `PATH` name that exists without `+x` should be 126."** `dash` itself returns 127 there, and so does the head. Parity, not a gap — which is why the prototype's check is restricted to path-y spellings.
- **"The `--shell` flag collides with a wrapped tool's own `--shell`."** `trailing_var_arg` keeps everything after the first positional literal, and clap has no `infer_long_args`. No collision.

### Your questions

- *"[ ] CI for Linux and Windows (`cmd /C`, `-Command` flag selection)"* → **Answered, partly.** No CI run had ever been created for this branch; I approved it and all three `test` jobs are green, so this is the first branch signal you have had. But it is weaker than it looks: every one of the 10 new integration tests is `#[cfg(unix)]`, so the Windows job compiles the new code and exercises none of it. That gap is blocker 2, and one `#[cfg(windows)]` test closes the part of it I can ask for.

### Next

Three things, then this merges: map the spawn-side failures (and 126) through the runners' own failure path the way the resolution-side ones now are; on Windows use `child_args` plus one `#[cfg(windows)]` test and a breaking-change line about `cmd`'s builtins and cwd resolution; and make `--shell` report its own misuse on `err`/`test`/`summary`.

**Fixup `4c2da904` prepared, not pushed** — yours whenever you want it. It does the two low-hanging pieces: resolves an explicit shell without a variable `Command::new`, which clears the semgrep job without suppressing a security rule (verified with the repo's own `.semgrep.yml` in the semgrep Docker image: 1 blocking finding → 0); and adds `explicit_shell_by_name_runs_the_named_shell`, because the two tests that exercise a named shell both start with `let Ok(fish) = which::which("fish") else { return; }` and pass vacuously wherever fish is absent — which is every CI runner. I proved that gap rather than assumed it: with the named-shell branch replaced by one that resolves the shell and then ignores it, all eleven existing tests still passed; the new test is the only one that fails. `fmt` clean, clippy 0, 3865 passed, and I diffed the fixup binary against yours over 11 rows covering every `--shell` form and the missing-shell error — identical output and exit code on all of them.

The 126 prototype (`f5cc836f`, 123 insertions, 3866 passed) is separate and yours to take or leave.

Rounds: 1/3. Threads: 3 open (all waiting on you), 0 resolved this round.

KuSh:
One addendum to the "Optional" list above, now that I have closed #3002 in favour of this PR.

#3002 attacked the same defect by re-quoting argv before joining it for the shell, which this PR obsoletes — but its quoting helper is better than `quote_for_display` for the job `quote_for_display` still has, and it is worth lifting rather than rewriting:

```rust
let is_safe = !arg.is_empty()
    && arg.chars().all(|c| {
        c.is_ascii_alphanumeric() || matches!(c, '_' | '-' | '.' | '/' | ':' | '=' | '@')
    });
```

An allowlist is the right shape here. `quote_for_display` currently decides with a denylist — `[' ', '\t', '\n', '\'', '"', '\\']` — so every other metacharacter stays bare in the label, which is why `rtk err /bin/echo '*' '$HOME' 'a;b' 'x&&y'` is recorded and printed as `/bin/echo * $HOME a;b x&&y`. A denylist has to enumerate the dangerous set correctly forever; an allowlist only has to enumerate the boring set once.

The escaping below it in #3002 splits on `cfg!(target_os = "windows")`, which you will not want as-is — that is a question about a child process's encoding, not about a label, and it belongs with `ChildArgExt` rather than here. It is the predicate that is worth taking.

pashifika:
@KuSh — all three blockers addressed on `482d67d`. Round-2 head is rebased on `develop` @ `0924356` still; say the word if you want a fresh rebase before you re-run.

**Blocker 1 — fixed as a class, both arms.** `Launch::NotFound(String)` became `Launch::Unrunnable { message, code }`, so the answer carries the code rather than assuming 127:

- **pre-spawn**: `which` is still asked first, and a spelling that addresses the filesystem is then stat'd, so a directory and a file without `+x` report 126. A bare `PATH` name stays 127 even when a non-executable file of that name exists — the parity you measured against `dash`.
- **post-spawn**: `spawn_failure` maps the `io::Error` out of the runner's own `Err` — ENOENT → 127, EACCES/EISDIR → 126, ENOEXEC → 126 — and renders it through the same failure path. No `anyhow` chain reaches stderr for any of them.
- **third arm**: an unresolvable `--shell` name now answers like an unrunnable program (127 + its own line) instead of `Failed to prepare`.

```
$ rtk err ./crlf.sh                 # '#!/bin/sh\r\n'
[FAIL] Command failed (exit code: 127)
  rtk: /tmp/…/crlf.sh: command not found                        exit 127
$ rtk run ./bindir/noexec-prog      # mode 644
rtk: /tmp/…/bindir/noexec-prog: Permission denied               exit 126
$ rtk run ./adir2                   # a directory
rtk: /tmp/…/adir2: Is a directory                               exit 126
```

`tests/run_shell_test.rs` gained `an_unexecutable_path_reports_exit_126` (all four surfaces), `a_directory_reports_exit_126`, `spawn_failures_keep_the_shell_contract` (CRLF shebang + a file the kernel refuses; the second asserts 126-or-127 and *never* a spawn error, because Linux reports ENOEXEC where macOS falls back to `sh` and gets 127 itself) and `an_unresolvable_shell_reports_the_shell_contract`. I did not take the prototype — thank you for offering it; writing it against the same shape was faster than reconciling two diffs, and the errno mapping stayed out of `src/core/stream.rs` exactly as you suggested: it reads the error the four surfaces already receive.

**Blocker 2 — `child_args`, plus the Windows delta documented.** `direct_command` passes caller argv through `ChildArgExt::child_args`, and `shell_command` uses `child_arg` for the script. `tests/run_shell_test.rs` grew a `#[cfg(windows)] mod windows`: PATH resolution, an argument carrying a space and a `"` reaching the child intact, and the 127 contract. README now states what `cmd /C` carried and direct execution does not — the working-directory search and `cmd`'s ~30 builtins (`rtk run -c 'dir'` or `rtk err --shell cmd 'echo hi'` is the replacement) — and the `argv[0]` spelling change from your optional list is in the same paragraph.

**Blocker 3 — the flag reports its own misuse.** `err` and `summary` joined `RTK_META_COMMANDS` (with the `trailing_var_arg` skip in `test_meta_commands_reject_bad_flags`). `test` cannot, as you said, so `run_fallback` checks for RTK's own `--shell` in argv before falling through — a command that asked for a shell never silently execs the POSIX `test`:

```
$ rtk err --shell        error: a value is required for '--shell <SHELL>' but none was supplied   exit 2
$ rtk test --shell       (same)                                                                   exit 2
$ rtk summary --shell    (same)                                                                   exit 2
```

`a_missing_shell_value_reports_a_flag_error` covers all four surfaces.

**semgrep.** `src/core/shell.rs` no longer constructs a `Command` at all: the explicit shell is resolved and then built through `core::utils::resolved_command`, which is where RTK's PATHEXT/`.CMD` handling lives anyway. That leaves `dynamic-command-execution` nothing to flag in the new file, without a suppression. Your fixup's second half is in too — `explicit_shell_by_name_runs_the_named_shell` names `sh`, so the explicit-shell branch is exercised on every runner rather than skipping wherever `fish` is absent.

**From the optional list**: `display_args` quotes the full shell metacharacter set (`*`, `$`, `;`, `&`, `|`, `<`, `>`, backtick, `#`, …), so `rtk err /bin/echo '*' 'a;b'` records and prints `/bin/echo '*' 'a;b'`; `rtk test '(' cmd ')'` keeps its grouping (`split_leading_negations` handles `(`/`)` beside `!`, so it and `is_native_test_expression` agree again); `report_unrunnable` prints the `-v "Running:"` line; the arity rule has one wording, `core::shell::SHELL_ARITY_MESSAGE`, quoted by both the clap error and `command_from_args`; the usage line now comes from the subcommand that was asked; and the stale docs are refreshed — `RTK_INSTRUCTIONS`' Test line, `INSTALL.md`, and all six translated READMEs.

**Left alone**: the negative-savings row. You are right that `never_worse` here drops the `[FAIL]` line the 127 test asserts — I reproduced that — so it stays with #3702 where the decision belongs.

```
cargo fmt --all --check                  clean
cargo clippy --all-targets -D warnings   clean
cargo test --all                         3,875 passed, 0 failed, 8 ignored
```

(macOS 26.6.2 arm64, rustc 1.97.1.) CI needs approving again on the new head — the Windows job now has three tests that actually exercise the new path.

PR #4127, opened 2026-09-19
feat(hooks): wrap unambiguously-fish scripts as rtk run --shell fish

pashifika:
## Summary

**PR C of the three-way split of #3911**, per @KuSh's round-1 review — the part he asked to argue on its own: *"should RTK infer a command's shell?"*. Opened as a PR rather than an issue so the behaviour is reviewable alongside the question; say the word and I will move the discussion to an issue instead.

**Stacked on #4125 (A) and #4126 (B)**, whose commits are the first four in this diff — C's own change is the last two. It needs A for `rtk run --shell fish -c`, the form it emits, and B for the wrapper parser it uses to recognise a command that already delegates.

### The premise, stated plainly

A host can evaluate the Bash tool's command string with a POSIX/zsh layer while the user's shell is fish. A fish script then fails to parse *before RTK is consulted at all* — nothing downstream can recover, because the command is already gone by the time anyone sees the error. That is an agent putting fish syntax into a tool called Bash; @KuSh is right that compensating for it is a policy choice, not a bug fix. What this PR argues is that the choice is cheap, reversible and precise:

```fish
if test -d src
  git status
end
# → rtk run --shell fish -c 'if test -d src\n  git status\nend'
```

- **Precise**: a script is wrapped only when it is *provably* fish — a fish-only keyword (`end`, `begin`, `switch`, `and`, `or`, `not`) at a command boundary, and nothing a POSIX shell needs anywhere: `then`, `fi`, `do`, `done`, `esac`, `elif`, `[[`, a heredoc, or a **backtick** (added this round, from the review's optional list: fish has no backtick substitution, so `` echo `hostname`; and echo ok `` no longer classifies). @KuSh's own measurement on 494 real command lines from this repo's docs: **0** wrapped.
- **Never a promotion**: the wrap's strongest verdict is `Ask`, never `AllowRewrite`, because the script's content is not attested. A deny rule outranks it. `fish_wrap_is_never_auto_allowed` and `deny_outranks_the_wrap` pin both.
- **Reversible**: `wrap_fish_scripts = false` under `[hooks]`, and it is skipped entirely on Windows or without a resolvable `fish` binary.
- **Byte-exact**: the script travels inside one quoted argument; `\\` and `\'` — the only two sequences where fish single-quote semantics diverge from POSIX — are refused, so every wrapped script round-trips identically under `sh`, `zsh` *and* `fish`. That round-trip is tested against the real shells (`wrapped_quoting_round_trips_under_posix_and_fish_hosts`); @KuSh independently verified it in `debian:sid`.

### What changed since #3911

The global `(…)`/control-keyword deferral is **gone**, not relocated. In #3911 it was what made a bare fish script reach the wrap, and it cost every ordinary bash line containing a subshell or a keyword block its rewrite (blocker 1). The trigger is now the classifier itself, asked first in `hooks::decision::decide`, so:

```
if test -d src\n  git status\nend      → rtk run --shell fish -c '…'        (exit 3)
test -d src; and git status            → rtk run --shell fish -c '…'        (exit 3)
echo `hostname`; and echo ok             unchanged                          (exit 1)
git status; if true; then echo x; fi   → rtk git status; if true; …         (exit 3)
(cd sub && git status)                 → (cd sub && rtk git status)         (exit 3)
```

The last two are the regressions blocker 1 measured; they are gone because nothing in this PR touches the shared gate any more.

`develop` has since merged `HookDecision` and `RewriteOutcome` into `hooks/decision.rs`, so the two copies of the wrap-at-gate from #3911 collapse to one site (`decide_with_wrap`), injected the same way the verdict and rewrite parameters already are (#3146) — tests never depend on the machine's `fish` or config.

## Key files

| File | Purpose |
|---|---|
| `src/discover/fish_script.rs` | Classification, POSIX/backtick vetoes, divergent-backslash veto, gate ordering, wrap assembly |
| `src/hooks/decision.rs` | `decide_with_wrap` — one site, ahead of the rewrite rules, `Ask`-only |
| `src/core/config.rs` | `hooks.wrap_fish_scripts` (defaults to enabled) |
| `tests/fish_wrap_test.rs` | E2E wrap, opt-out, backslash veto, cross-host quoting round-trip, real-`fish` execution (`#[ignore]`) |

## Test plan

- [x] `cargo test --all` — 3,947 passed, 0 failed, 9 ignored
- [x] `rtk rewrite` transcript above on this branch's binary, isolated `HOME`/`XDG_CONFIG_HOME`
- [x] `rtk run --shell fish -c 'if test -d src; echo yes; end'` executes and exits 0
- [ ] CI for Linux and Windows

Tested on macOS 26.6.2 arm64, rustc 1.97.1, fish 4.9.0.

---

Split siblings: **A** — #4125 (argv/`--shell`, exit-127 fix), **B** — #4126 (`fish -c` wrapper rewriting, dialect-scoped). #3911 is closed in favour of the three.


src/discover/fish_script.rs
```diff
@@ -0,0 +1,404 @@
...
+                None => quote = Some(character),
+                _ => {}
+            }
+        }
+    }
+
+    quote.is_some() || escaped
+}
+
+fn classify_tokens(tokens: &[ParsedToken]) -> bool {
```
KuSh, line 222: **Blocker 1 — `classify_tokens` reads comment text as code, so a working bash command can be wrapped and then fail outright under fish.**

Neither `has_unclosed_quote_or_escape` nor the shared lexer recognises `#`. Everything after a comment marker is still tokenized, so a `;`, `|` or `&` *inside the comment* resets `command_position` and the next word is matched against `FISH_ONLY_KEYWORDS` — and `and`, `or`, `not`, `end` are ordinary English words.

```
echo $((1+2))  # x; and y
  bash    -> 3                                                      exit 0
  develop -> (defer)                                                exit 1
  head    -> rtk run --shell fish -c 'echo $((1+2))  # x; and y'    exit 3
  fish    -> exit 127, nothing runs
             "command substitutions not allowed in command position"

echo "${HOME}"  # note; and more
  bash -> /home/u   exit 0     fish -> exit 127 "Variables cannot be bracketed"

ls -la  # long listing; and hidden files
  develop -> rtk ls -la  # …                  (filtered)
  head    -> rtk run --shell fish -c '…'      (unfiltered)
```

A command that worked becomes one that does not run at all. Same shape for arrays and `"$@"`.

Class: every command whose live part uses POSIX-only syntax and whose trailing comment contains a separator followed by one of the six markers. `classify_tokens` is the single site — the fix does not need to be repeated anywhere.

Suggested fix: drop everything from an unquoted, word-initial `#` to the next newline before tokenizing. sh, bash, zsh and fish all agree that span is a comment, so nothing is lost. Test: extend `test_fish_keyword_as_argument_is_not_marker` with a comment-borne separator — that case currently passes for the wrong reason (the marker is not at command position), which is why the existing coverage does not catch this.

src/hooks/decision.rs
```diff
@@ -78,11 +78,39 @@ pub(crate) fn decide_with_params(
...
 ) -> HookDecision {
     if verdict == PermissionVerdict::Deny {
         return HookDecision::Deny;
     }
 
+    // A provably-fish script fails to parse in a POSIX host layer before RTK is
+    // ever consulted, so it is handed back as one quoted argument every layer
+    // can parse. Never auto-allowed: the script's content is not attested, so
+    // its strongest verdict is `Ask` even under an explicit allow rule.
+    if let Some(wrapped) = wrap_fish(cmd) {
```
KuSh: **Blocker 2 — the wrap precedes `rewrite_command`, so a fish chain whose leading command is rewritable loses the rewrite `develop` performs.**

```
git diff HEAD~3 HEAD; and true
  develop     b748a5f7 -> rtk git diff HEAD~3 HEAD; and true   exit 3
  #4125+#4126 5b547c99 -> rtk git diff HEAD~3 HEAD; and true   exit 3   (byte-identical)
  head        f3347cf8 -> rtk run --shell fish -c '…'          exit 3

output measured on this repo:  raw 119 277 B
  develop path   27 561 B   (76.9% saved)
  head path     119 277 B   ( 0.0% saved)
```

**27 of 28** common leading commands lose it (`git status`, `git diff`, `cargo test`, `pytest`, `ls -la`, `grep -rn`, `cat`, `docker ps`, `go build`, `npm run build`, …); **10 of 10** confirmed three-way. Since `develop` and the #4125+#4126 stack are byte-identical here, this comes from this PR's own two commits — it is not pre-existing, not inherited from the siblings, and distinct from #2767 / #1243, which ask for rewriting in cases that never worked at all.

The target form does work: `rtk run --shell fish -c 'rtk git diff HEAD~3 HEAD; and echo OK'` runs under fish 4.9.1, exit 0, tail reached, 27 569 B — 76.9% restored. But #4126's inner rewriting does not currently reach `; and` / `; or` chains (`fish -c "git status; and ls"` defers, while `fish -c "git status"` becomes `fish -c "rtk git status"`), so wiring it in is not a one-liner. That is why this is a question rather than a suggestion: close the hole, or accept the loss and say so in the docs.

src/hooks/decision.rs
```diff
@@ -78,11 +78,39 @@ pub(crate) fn decide_with_params(
...
+        return HookDecision::AskRewrite(wrapped);
```
KuSh: **Blocker 3 — the wrap is the only decision path that rewrites a command RTK could not attest.**

It sits between the Deny check and `contains_unattestable_construct`, whose own doc a few lines up calls refusing to rewrite such a command "the stronger guarantee".

```
test -d src; and cat /etc/shadow > /tmp/leak
  develop -> (defer, exit 1)  — the host evaluates the original string
  head    -> AskRewrite "rtk run --shell fish -c '…'"   exit 3
```

To be precise about what this is **not**: Deny still wins. `rm -rf /tmp/zzz`, `git status && rm -rf /tmp/zzz` and `if test -d src\n  rm -rf …\nend` all still exit 2 on this head, so it is not a deny bypass. And fish's own `(cmd)` substitution is invisible to `contains_substitution` either way.

What changes is that a script the gate declined to decompose is now emitted as an `rtk`-prefixed command, so a host's matcher sees `rtk run --shell fish -c '…'` where it previously saw the script. Whether a given host's allowlist then behaves differently is host behaviour I did not measure — but on **Codex** the code answers it itself: `AskRewrite` renders as a protocol-level `allow` (`hook_cmd.rs`), and the comment right above that call already says Codex's safe/dangerous-command classifiers "do not currently unwrap `rtk`" and asks to "revisit this boundary whenever coverage expands". This PR expands it from single rewritten commands to arbitrary multi-command scripts inside one quoted argument.

Either gate the wrap behind `contains_unattestable_construct` the way every other rewrite is, or state in `src/hooks/README.md` why it is exempt.

KuSh (changes requested):

**Claim**: a command string that is *provably* fish — a fish-only keyword at command position, no POSIX disambiguator anywhere — is handed back as `rtk run --shell fish -c '<script>'` ahead of the rewrite rules, `Ask`-only, opt-out via `hooks.wrap_fish_scripts`.

**Scope**: **accept.** This is exactly the PR C the #3911 split asked for, and opening it as a PR rather than an issue was the right call — the measurements below only exist because there was a branch to run. #3911's blocker 1 is verifiably gone: the global `(…)`/control-keyword deferral is *absent* rather than relocated, and every bash row it used to cost a rewrite matches plain `develop` byte for byte. Merge order: #4125 and #4126 first (this is stacked on both); #3909 also rewrites `src/hooks/decision.rs` (+220/−11) around the same verdict handling and is not blocked on anything, so whichever of the two goes second will rebase.

**Ran**: `rtk rewrite` and `rtk hook <host>` on three binaries — plain `develop` b748a5f7, the #4125+#4126 stack 5b547c99, this head f3347cf8 — across 16 matrix rows, 19,870 real agent-emitted Bash commands, 3,733 repo command lines and 5,818 adversarial fuzz cases; end-to-end execution against real fish 4.9.1 under sh/bash/dash/zsh/fish host layers; 7 mutations of the classifier's guards. Invariants: every branch of the new decision (Default/Allow/Deny × wrap on/off × fish present/absent), under the hook (`rtk rewrite` first, then the form it returns), split streams, over-cap input (2.7 MB and binary stdout), non-UTF-8, BOM and CRLF payloads. No filter is added, so the 20% floor (CONTRIBUTING.md) has no filter to measure — but blocker 2 *is* a savings regression against `develop`.

**CI** @f3347cf8: **never ran.** Run `35425641688` sits at `action_required`; only `license/cla` is green and `check-target` is `SKIPPED`. It needs approving before any of this can be signed off.

1. `src/discover/fish_script.rs:139` — **comment text is classified as code** (1 site). `echo $((1+2))  # x; and y` → bash prints `3`, head wraps it, fish exits 127 and nothing runs. A working command becomes a hard failure. Fix: strip unquoted word-initial `#` to newline before tokenizing.
2. `src/hooks/decision.rs:110` — **the wrap costs the rewrite `develop` performs** (27 of 28 leading commands; 10 of 10 three-way). `git diff HEAD~3 HEAD; and true`: 76.9% saved on `develop` → 0% here.
3. `src/hooks/decision.rs:111` — **the wrap is the only path that rewrites a command RTK could not attest**; on Codex that renders as a protocol-level `allow`. Deny itself still wins — this is not a deny bypass.

Details and transcripts are in the three line comments.

- **`hooks.exclude_commands` and `transparent_prefixes` are bypassed**, and `decide`'s numbered gate-order doc still lists 1. Deny / 2. unattestable / 3. rewrite with no mention of the new gate between 1 and 2. Both fields are read by `rewrite_command`, which the wrap precedes: `test -d src; and curl https://x` wraps even with `exclude_commands = ["curl"]`. `wrap_fish_scripts` is the dedicated switch, so this may well be intended — one line in that doc list either way.
- **`tests/fish_wrap_test.rs` asserts nothing on any runner without `fish`**: all five tests `return` early on `which::which("fish").is_err()`, the sixth is `#[ignore]`. Forcing `try_wrap` to return `None` leaves 5/5 green. The classification logic itself is properly covered — mutating each of the seven guards (backtick, heredoc, POSIX keyword, `[[`, `\\`/`\'`, unclosed quote, Ask-not-Allow) fails a named unit test — so this is an end-to-end gap rather than a coverage hole. A `fish` step in CI, or an assertion on the fish-absent branch, would stop the file being inert.

- **Byte-exact round trip.** 9 scripts (multiline `if`/`for`/`switch`/`begin`, `; and`, `; or`, `not`, embedded single quotes, `printf '%s\n'`) wrapped and executed under **sh, bash, dash, zsh and fish**: stdout identical to `fish -c '<script>'` in every cell.
- **IO fidelity through `rtk run --shell fish -c`.** 2 688 900 B / 400 001 lines identical; 2 000 B of `/dev/urandom` identical; large stderr identical; exit codes 0/1/7 propagated; stdin forwarded. No cap, no truncation, nothing to lose.
- **Precision on real input.** 0 wraps over **19,870** agent-emitted Bash commands from 195 Claude Code transcripts, and 0 real commands over **3,733** repo command lines (the 2 hits there are doc prose and a bare `end` in a fish snippet). Blocker 1 is a mechanism defect, not something this corpus happens to trigger.
- **Gates.** `cargo fmt --all --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test --all` on **the head merged onto latest `develop`** (merge is clean): **4 023 passed, 0 failed**; `cargo check --target x86_64-pc-windows-gnu --all-targets` clean.
- **Config.** Missing field defaults true; `wrap_fish_scripts = false` disables the wrap and survives a load-mutate-save round trip.
- **Decision branches.** Deny → `Deny`; Allow → `AskRewrite` (never promoted); Default → `AskRewrite`; wrap off → prior path; `fish` absent from `PATH` → prior path. The wrap is idempotent — the emitted form re-enters as itself and `suppress_identity` defers it.
- **Perf.** No regression: 6.81 / 6.72 / 1.35 ms against `develop`'s 7.14 / 6.77 / 1.40 ms, 200 runs each. The config read and the `fish` probe are `FnOnce` behind the classifier, so non-fish commands pay nothing.
- **Host payloads.** claude / copilot / cursor carry the wrap in `updatedInput` without asserting a permission decision; BOM, CRLF and multiline scripts all render correctly.

- **Injected-marker false positives.** 5,818 fuzz cases, 5,258 wrapped, 246 where bash exits 0 and fish diverges — every one needs a bare marker keyword as a standalone command (`command -v git || end`), which bash reaches only as dead code after a successful left side. No real command has that shape. (Blocker 1 is a different mechanism and is real.)
- **gemini / codex get no benefit from the wrap.** True, but `{"decision":"allow"}` and empty output are what **both** `develop` and this head emit for a plain `git status` as well — pre-existing per-host rendering. The Codex point that does survive is in blocker 3.
- **`has_unclosed_quote_or_escape` disagreeing with the lexer.** Escape-before-quote ordering and the `quote != Some('\'')` rule are identical in both scans; no input found where they disagree. A quote inside a comment *suppresses* classification (`ls # don't; and rm -rf /` → unclosed → no wrap), which is the safe direction.
- **`HooksConfig`'s hand-written `Default` drifting from the struct.** It is an exhaustive struct literal; a new field is a compile error, not a silent `false`.

### Your questions

- *"say the word and I will move the discussion to an issue instead"* → **keep it here.** The behaviour is what made the question answerable.
- *"should RTK infer a command's shell?"* → on this evidence, **yes**. The classifier is precise on real input, the wrap is reversible, `Ask`-only, byte-exact across five host layers, and costs nothing when it does not fire. None of the three blockers is an argument against inferring the shell; they are all about *where* the inference is placed — ahead of the comment stripper it does not have, ahead of the rewrite rules, and ahead of the attestation gate.

### Next

Strip `#` comments before classifying (blocker 1) — a correctness bug and the smallest of the three. Then decide blocker 2 (nest the rewrite, or document the loss) and blocker 3 (gate behind `contains_unattestable_construct`, or document the exemption). Everything else is green; CI still has to be approved and pass on the resulting head.

Rounds: 1/3. Threads: 3 open, 0 resolved — no prior threads on this PR.

pashifika:
@KuSh — three blockers fixed on `d87d12d`, and the branch is rebased on the round-2 heads of #4125 and #4126.

**Blocker 1 — comment text is no longer code.** `is_unambiguous_fish` drops every unquoted, word-initial `#` through the end of its line before tokenizing, exactly as you suggested; sh, bash, zsh and fish all agree on that span. Your three cases, on this head:

```
echo $((1+2))  # x; and y                  unchanged                       (exit 1)
echo "${HOME}"  # note; and more           unchanged                       (exit 1)
ls -la  # long listing; and hidden files → rtk ls -la  # long listing; …   (exit 3)
```

The third is the one that shows the fix is not just "refuse more": it rewrites as bash, byte-identical to `develop`. `test_comment_text_is_never_a_marker` pins all four (yours plus `ls # don't; and rm -rf /`), and `test_comment_stripping_leaves_code_alone` pins the two directions a naive strip would break — `rg '#end' src/` stays an argument, and a fish script with a comment of its own still classifies.

**Blocker 2 — the wrap keeps the rewrite.** The script's own commands are rewritten before it is wrapped, so the savings you measured come back:

```
git diff HEAD~3 HEAD; and true
  → rtk run --shell fish -c 'rtk git diff HEAD~3 HEAD; and true'    (exit 3)
```

It runs through `rewrite_command` with the same `excluded` / `transparent_prefixes` the unwrapped path gets, which also closes the `exclude_commands` bypass from your optional list — `the_nested_rewrite_honours_exclusions` pins that, and `rewrite_keeps_the_inner_rewrite_inside_the_wrap` covers it end to end.

**Blocker 3 — the wrap now refuses what the gate refuses.** Rather than move it behind `contains_unattestable_construct` (which would kill the feature: `; and` and `if … end` are exactly what the gate cannot decompose), it applies that gate's own rules to the script itself — command and process substitution, file-target redirects — plus fish's `(cmd)` substitution, which the bash-shaped lexer reads as a subshell and lets through:

```
test -d src; and cat /etc/passwd > /tmp/leak     unchanged   (exit 1)   ← your case
test -d src; and echo $(whoami)                  unchanged   (exit 1)
for f in (ls)\n  echo $f\nend                    unchanged   (exit 1)
test -d src; and git status                    → wrapped     (exit 3)
if test -d src\n  git status\nend              → wrapped     (exit 3)
```

What still reaches the wrap is a script whose *only* unattestable property is fish control flow — the case the feature exists for. `src/hooks/README.md` now states why that one is exempt, in your terms: deferring loses the command outright rather than leaving it unrewritten, the script travels byte-identical, the verdict is `Ask` and deny wins first, and the residual exposure is a host that treats an `rtk`-prefixed command as pre-approved — naming Codex, its protocol-level `allow`, and `hooks.wrap_fish_scripts = false` as the switch for anyone who does not want that trade. `test_unattestable_scripts_are_not_wrapped` and `rewrite_defers_unattestable_fish_scripts` pin it at both levels.

**Also from the optional list**: `decide`'s numbered gate-order doc lists the wrap as gate 2 and says where the reasoning lives; and `tests/fish_wrap_test.rs` asserts on the fish-absent branch (defer, exit 1, no stdout) instead of returning early, so the file is no longer inert on a runner without `fish`. A CI `fish` step would still be better than either branch — happy to add one if you want it in this PR rather than separately.

```
cargo fmt --all --check                  clean
cargo clippy --all-targets -D warnings   clean
cargo test --all                         3,965 passed, 0 failed, 9 ignored
```

(macOS 26.6.2 arm64, rustc 1.97.1, fish 4.9.0.) Keeping the discussion here as you asked. CI has still never been approved on this branch — that is the last thing outstanding that I cannot do myself.

PR #4145, opened 2026-09-20
fix(recall): enforce retention on open, not only on store

### LINKED ISSUE #4144: Recall retention only applies on write: a quiet machine keeps rows past retention_days
## What

`retriever.retention_days` (default 30) and `max_entries` (default 200) are enforced by `evict()`, and `evict()` is called from exactly one place: `store_inner`, immediately after inserting a new row ([src/core/retriever.rs](https://github.com/rtk-ai/rtk/blob/develop/src/core/retriever.rs), the `insert recall row` path).

So retention is a property of **continued use**, not of elapsed time. A machine where rtk stops eliding — a project finished, a laptop set aside, a filter disabled, a contractor's box left on a shelf — keeps every stored row indefinitely, while `rtk config recall` and the documented default both say 30 days.

Nothing is wrong with the eviction logic itself; it is correct on an active machine. The gap is the claim.

## Why it is worth a fix rather than a doc note

The stored rows are the *output a filter hid*, plus the command text, cwd and exit code. On a machine where that output is sensitive, "we keep it 30 days" is exactly the sentence a reviewer relies on, and it holds only while somebody keeps working. A stated retention that silently doesn't apply is the same shape as the two defects already fixed this week — #4141 (`init` exits 0 having installed nothing) and #4143 (`discover` reports "looks good" over zero sessions scanned): **the tool reports a state it did not reach.**

## Possible fixes, in increasing cost

1. **Evict on read too** — call `evict()` from the `recall` command path, and from `gain`/`config recall`. Cheap, and it makes "did anyone look at this machine" the trigger rather than "did anyone store on it".
2. **Evict on open** — a sweep in `open()` behind a cheap guard (e.g. at most once per process, or when the newest row is older than the retention window). Bounds the staleness by "rtk ran at all".
3. **Document it honestly** — change the wording to "entries are evicted when rtk next stores output", and say so in `rtk config recall`. Correct, cheapest, and worse than either of the above for anyone reasoning about data lifetime.

I would take (2) with (1) as the fallback, but this is your design call. Happy to send a PR for whichever you prefer.

## Context

Found while documenting rtk's local stores for a two-machine deployment, after checking what `rtk recall` persists and where. The store itself is well-behaved: content-addressed, compressed, size-capped, and `0700`/`0600` on Unix. Worth noting separately that `set_owner_only` is `#[cfg(not(unix))] {}`, so on Windows the files carry whatever the profile ACL grants — I can raise that as its own issue if it isn't already known.

dopaemon:
Fixes #4144

## Problem

`evict()` was called from exactly one place — `store_inner`, right after an insert — so `retention_days` measured *continued use*, not elapsed time. A machine that stops storing output (project finished, laptop shelved, filter disabled) kept every row indefinitely while `rtk config recall` and the documented default both promised 30 days.

Demonstrated on `develop` with an isolated store (`RTK_RECALL_DB`), one row aged 31 days, no new output stored:

```console
$ rtk recall --list          # before this change
HASH           COMMAND       LINES  HIDDEN  ORIG  EXIT
aaa111111111   stale cmd         1       1     1     0
$ sqlite3 recall.db 'SELECT COUNT(*) FROM recall;'
1
```

## Fix

Option (2) from the issue, which also subsumes (1): sweep from the two connection entry points rather than from a command path, so every invocation that touches the store — `rtk recall`, `gain`, `config recall`, the tee stats path — enforces retention. `open()` and `open_existing()` both call the new `expire()` right after `init_schema`, and `expire` is just `evict(conn, cfg, "")` with no exempt hash (nothing is being protected on a sweep).

Properties kept:

- **Best-effort**, like `evict` itself — a read-only or locked store keeps its rows instead of erroring.
- **The read path still never creates the database**: `open_existing` returns early when the file is absent, so `test_stats_snapshot_never_creates_db` still holds.
- **Cost**: one `DELETE` plus a `COUNT(*)` over at most `max_entries` (default 200) rows, on a connection that was being opened anyway.

After:

```console
$ rtk recall --list
HASH           COMMAND       LINES  HIDDEN  ORIG  EXIT
bbb222222222   fresh cmd         1       1     1     0
$ sqlite3 recall.db 'SELECT COUNT(*) FROM recall;'
1                                  # the 31-day-old row is gone
```

## Tests

New `test_retention_applies_without_a_new_store`: two stored rows, one aged past `retention_days`, then a read-only `open_existing()` — the stale row must be gone and the fresh one untouched. The 48 existing retriever tests are unchanged and green (several of them open the store after inserting rows, which is what would break if the sweep were too eager).

`cargo fmt --all` + `cargo clippy --all-targets` clean; `cargo test --bin rtk` → 3672 passed.

## Note

The issue also flags that `set_owner_only` is a no-op off Unix, so on Windows the store carries whatever the profile ACL grants. That's a separate concern and isn't touched here — worth its own issue as the reporter suggests.


Wilson421:
Reporter of #4144 here — I verified this independently rather than just reading it, since it is my issue and an unverified "looks good" is worth nothing.

**Built and ran it on Windows 11, rtk at `bf5747f`:**

- `cargo test --bin rtk core::retriever` — 45 passed, 0 failed.
- **Negative control**, because a test that passes against the bug proves nothing: I applied **only** your new test onto a clean `develop` (not the fix) and it fails exactly where it should — `a row past retention_days must not survive a read-only open`, at the `load_by_hash` assertion. So the test genuinely catches the reported behaviour.

**On the shape you chose:** option (2) is the one I would have picked, and for the reason you give — one sweep site instead of one per caller. I checked every path that opens the store (`stats_snapshot`, `reset_stats`, `record_tee_elision`, `run_recall`), and all four are legitimately "rtk ran" moments; there is no backup or export path where an automatic DELETE would surprise someone. Keeping it best-effort and leaving `open_existing`'s early return intact means a read-only or absent store still behaves as before.

Two observations, neither a blocker:

1. **A stale hash is now swept before it can be read.** `rtk recall <hash>` on a row past `retention_days` previously still returned content (eviction only ran on store); now the sweep in `open_existing` removes it first, so the user gets nothing instead of something stale. That is the stated policy working as intended, and I think it is right — but it is a visible behaviour change for anyone recalling a hash out of an old transcript, so it may deserve a CHANGELOG line.
2. **Precedent for sampling, if the sweep ever shows up in a profile.** `record_hook_decision` faces the same "this is on a hot path" problem and solves it with `maybe_cleanup_hook_decisions` — a sampled sweep rather than one per call. The recall store is opened far less often, so I would not add that complexity now; just noting where the pattern already lives in this codebase if it becomes necessary.

Thanks for picking this up so quickly, and for reproducing the gap on `develop` first — that is the part most fixes skip.

